PR #2987, opened 2019-07-01
typo fix in core

kaichaosun:


kaichaosun:
* The changes under `core/sr-` are all comments related, I guess it should be fine even the `check-runtime` is failed, not sure if it's worth to bump up the runtime impl_version in `node`.
* The `check-line-width` also failed, from the log in gitlab CI, it seems to be a temporary network issue.

PR #2882, opened 2019-06-17
Staking rate targeting and specific rewards.

thiolliere:
## I_NPoS

The total payout is calculated using I_NPoS function defined in http://research.web3.foundation/en/latest/polkadot/Token%20Economics/#payment-details

I_NPoS is implemented using some a piecewise linear function. This piecewise linear function has been generated with a test and is tested not to differ from actual float implementation with an error of 0.2%

## Staking api

* The previous era reward accumulated at each new session is removed as reward is now computed from P_NPoS
* a root call `add_points_to_validator` is added so ppl rewarded can call it from inherent mecanism.

TODO:
* add inherent call to add_points_to_validator so ppl are actually rewarded:
  > * [ ] In the branch of validity checking, we reward:
  >   * a parachain validator for each validity statement of the parachain block that it issues.
  >     should be made in polkadot, probably near https://github.com/paritytech/polkadot/blob/master/runtime/src/parachains.rs#L741
  > 
  > * [x] In the branch of block production, we reward:
  >   * the block producer for producing a (non-uncle) block in the relay chain,
  >   * the block producer for each reference to a previously unreferenced uncle, and
  >   * the producer of each referenced uncle block.
* [x] traitify timestamp instead of direct use, note that contract and aura are using it directly as well
* timestamp trait: what api is to be used, returning 0 if timestamp is not yet inserted ?

related to https://github.com/paritytech/substrate/issues/2781

PR #2995, opened 2019-07-02
ci: add job to test staking in srml

gabreal:
add a step to the ci that tests staking in srml

Related #2882

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -144,7 +144,7 @@ test-srml-staking:                 &test-srml-staking
       - srml/staking/*
   script:
     - cd srml/staking/
-    - time cargo test --tests tests --no-default-features --features std
+    - time cargo test --no-default-features --features std
     - sccache -s
 
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -144,7 +144,7 @@ test-srml-staking:                 &test-srml-staking
       - srml/staking/*
   script:
     - cd srml/staking/
-    - time cargo test --no-default-features --features std
+    - time cargo test --release --verbose --no-default-features --features std
     - sccache -s
 
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -126,6 +126,28 @@ test-linux-stable:                 &test-linux
     - time cargo test --all --release --verbose --locked
     - sccache -s
 
+
+test-srml-staking:                 &test-srml-staking
+  stage:                           test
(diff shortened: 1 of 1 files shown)


gabreal:
the srml staking test is failing here, is that expected?

.gitlab-ci.yml
```diff
@@ -126,6 +126,32 @@ test-linux-stable:                 &test-linux
...
+  except:
+    variables:
+      - $DEPLOY_TAG
+  only:
+    changes:
+      - .gitlab-ci.yml
+      - srml/staking/*
+  script:
+    - cd srml/staking/
+    - time cargo test --tests tests --no-default-features --features std
```
thiolliere:
```suggestion
    - time cargo test --no-default-features --features std
```
yes actually for now only relevant test of staking are in tests modules, but might be better to test for whole package in case in the future we add test in lib.rs that have different behavior depending on this feature.
(how is time important here, I mean in my current other PR I have inflation module which doesn't need to be tested with both with and without the feature, does it worth filter out my module in this test ?)

thiolliere:
no it is not expected to fail, I'll take a look, also maybe cc @Kianenigma

.gitlab-ci.yml
gabreal: I don't understand entirely. With the current version of the pr the test is only run when there are changes detected in the `srml/staking` subdirectory. Do you think it should also be run for changes to other directories of the source tree?

Kianenigma: Not really, this is completely isolated to the `srml/staking`. Though I don't see it as a big deal to run the tests twice all the time, and not only there is a change?

thiolliere: theoritically the behavior of the module could change if its dependencies changes and dependency can change without modfiying srml/staking/* files due to current versionning using path.

But changing the behavior of the version without the feature and not the version with the feature seems very very unlikely. Hmm so I would say it is fine this way.


Oh also I forget but that's probably better to run them with release I think

```suggestion
    - time cargo test --release --no-default-features --features std
```

gabreal: I cannot commit this, but sounds totally reasonable

gabreal: and do I understand correctly that this test is already run from the `cargo test` from the top-level directory but with a different "feature"?

thiolliere: yes correct, cargo test --all will run this crate tests with default features which include "equalize" feature.

this line run the test suite of this crate but without the "equalize" feature

PR #3000, opened 2019-07-02
Allow where bounds on `decl_module!` and `decl_storage!`

bkchr:
This allows to put where bounds to `decl_module!` and `decl_storage!`. It is sometimes useful and required when you want to put some where bounds on the configuration trait.

We probably need more changes, but for the start it seems to work.


srml/support/src/dispatch.rs
```diff
@@ -206,15 +206,21 @@ macro_rules! decl_module {
 	// Entry point #1.
 	(
 		$(#[$attr:meta])*
-		pub struct $mod_type:ident<$trait_instance:ident: $trait_name:ident$(<I>, I: $instantiable:path $(= $module_default_instance:path)?)?>
-		for enum $call_type:ident where origin: $origin_type:ty {
+		pub struct $mod_type:ident<
+			$trait_instance:ident: $trait_name:ident
+			$( <I>, I: $instantiable:path $( = $module_default_instance:path )? )?
+		>
+		for enum $call_type:ident where origin: $origin_type:ty $(, $other_where_bounds:ty: $other_stuff:path )* {
```
thiolliere: oh ok I see only `type: trait` is allowed here: no lifetime and no +

thiolliere:
Oh maybe we should mention it in the doc ?
I mean the peculiar signature of where clause for dispatch for example.

for decl_storage I don't know what to mention though.

gavofyork:
Documenting it seems sensible.

srml/support/src/dispatch.rs
bkchr: Lifetime should not required and `+` is not doable in declarative macros :(

PR #3003, opened 2019-07-03
on_start now returns the precise elements to request

tomaka:
Some basic refactoring.
Instead of giving `on_start` full access to the import queue link, we make it have a more precise behaviour.

The objective of that is to allow some more refactoring of the import queue.


tomaka:
I'd personally prefer to kill the `on_start` method completely, as what it does should be done in the `new` method, but doing so requires a lot of refactoring.

Issue #2416, opened 2019-04-29
Add a light client code that compiles in the browser


tomaka:
This is a long term issue that I'm opening in order to keep track of PRs moving towards this direction.

We can't directly compile `node` for `wasm32-unknown-unknown`. Instead, we should create a new library (named for example `substrate-light-wasm`) that exposes an API usable from JavaScript, thanks to `wasm-bindgen`.

Once that crate exists, we can use `wasm-pack` to generate `.js` and `.wasm` files from `substrate-light-wasm`. The `.js` can then be used from regular JavaScript code.

Some challenges include:
- We can't use TCP/IP. Libp2p already provides a `wasm-ext` transport that can be used to "import" a transport from the JavaScript world, such as WebSockets.
- The RPC needs either to be disabled, or be replaced with something that compiles for the browser.
- The telemetry needs either be disabled, or be replaced with something that compiles for the browser. I have a WIP branch that replaces the telemetry with libp2p.
- We need to remove background threads, and have a fall-back for background tasks when tokio isn't available.
- The `tokio` dependencies need to be replaced with dependencies to their sub-crates: `tokio-io`, `tokio-codec`, etc.
- `ring` and `rust-crypto` don't compile for WASM. As far as I know, only diffie-hellman exchanges aren't possible through pure-Rust-only crates at the moment. Libp2p uses WebCrypto as a replacement.

rphmeier:
> The tokio dependencies need to be replaced with dependencies to their sub-crates: tokio-io, tokio-codec, etc.

Eep. The facade crate is way nicer.

tomaka:
> Eep. The facade crate is way nicer.

The reason for not using `tokio` is https://github.com/rust-lang/cargo/issues/4866.
We need to have for example `tokio-tcp` enabled only on non-WASM platforms, but you can't disable tokio's features per-platform because of this Cargo issue.

DemiMarie-parity:
While having a light client that compiles to wasm is nice, I believe that doing anything security-critical in the browser is a bad idea, as a compromised server or an MITM attack could allow compromise of the client’s secrets.  Two solutions are:

* Use a browser extension (WebExtension) instead.  These run with enhanced privileges, so they can protect themselves, and are signed, so MITM et al are not issues.
* Require that a browser extension is installed that checks the signatures on the actual client.

gterzian:
[...]

PR #3004, opened 2019-07-03
Can now disable the keystore

tomaka:
For #2416, the keystore obviously cannot load a key from the disk.
This PR makes it possible to disable the keystore by passing `None` for `keystore_path`.

The second change I made is change the configuration to take `PathBuf` instead of `String`. I have no idea why we were using `String`, and it doesn't make sense to me.


core/cli/src/lib.rs
```diff
@@ -399,13 +399,10 @@ where
 
 	let base_path = base_path(&cli.shared_params, version);
 
-	config.keystore_path = cli.keystore_path
-		.unwrap_or_else(|| keystore_path(&base_path, config.chain_spec.id()))
-		.to_string_lossy()
-		.into();
+	config.keystore_path = Some(cli.keystore_path
```
```suggestion
	config.keystore_path = cli.keystore_path.or_else(|| keystore_path(&base_path, config.chain_spec.id()))
```

core/cli/src/lib.rs
```diff
@@ -399,13 +399,10 @@ where
...
+		.unwrap_or_else(|| keystore_path(&base_path, config.chain_spec.id())));
```
```suggestion
```

core/service/src/lib.rs
```diff
@@ -156,24 +156,40 @@ impl<Components: components::Components> Service<Components> {
...
-		// Keep the public key for telemetry
-		let public_key = match keystore.contents()?.get(0) {
-			Some(public_key) => public_key.clone(),
-			None => {
-				let key = keystore.generate(&config.password)?;
-				let public_key = key.public();
-				info!("Generated a new keypair: {:?}", public_key);
-
-				public_key
+		if let Some(keystore) = keystore.as_mut() {
```
bkchr: I think putting `let public_key = ` in front of the if is more idiomatic rust?

tomaka: This is extreme nit-picking, but I didn't do it this way because it makes it look like the primary objective of this block is to fetch the public key, while in reality it is to generate the key if there's none.

PR #3007, opened 2019-07-03
Implement Executor for Service and SpawnHandle

tomaka:
Let's implement `Executor` instead of having an inherent method on `SpawnHandle`.

Necessary to upgrade Polkadot, so it would be great to merge this quickly-ish.

### DIFF
--- a/core/service/src/lib.rs
+++ b/core/service/src/lib.rs
@@ -623,7 +623,8 @@ impl<Components> Future for Service<Components> where Components: components::Co
 }
 
 impl<Components> Executor<Box<dyn Future<Item = (), Error = ()> + Send>>
-for Service<Components> where Components: components::Components {
+	for Service<Components> where Components: components::Components
+{
 	fn execute(
 		&self,
 		future: Box<dyn Future<Item = (), Error = ()> + Send>
--- a/core/service/src/lib.rs
+++ b/core/service/src/lib.rs
@@ -124,11 +124,10 @@ impl Executor<Box<dyn Future<Item = (), Error = ()> + Send>> for SpawnTaskHandle
 		&self,
 		future: Box<dyn Future<Item = (), Error = ()> + Send>
 	) -> Result<(), futures::future::ExecuteError<Box<dyn Future<Item = (), Error = ()> + Send>>> {
-		if self.sender.is_closed() {
+		if let Err(err) = self.sender.unbounded_send(future) {
 			let kind = futures::future::ExecuteErrorKind::Shutdown;
-			Err(futures::future::ExecuteError::new(kind, future))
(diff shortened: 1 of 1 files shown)


core/service/src/lib.rs
```diff
@@ -614,6 +622,22 @@ impl<Components> Future for Service<Components> where Components: components::Co
 	}
 }
 
+impl<Components> Executor<Box<dyn Future<Item = (), Error = ()> + Send>>
+for Service<Components> where Components: components::Components {
```
bkchr: Please add a tab and move the { to the next line.

core/service/src/lib.rs
```diff
@@ -119,10 +119,18 @@ pub struct SpawnTaskHandle {
...
-	pub fn spawn_task(&self, task: impl Future<Item = (), Error = ()> + Send + 'static) {
-		let _ = self.sender.unbounded_send(Box::new(task));
+impl Executor<Box<dyn Future<Item = (), Error = ()> + Send>> for SpawnTaskHandle {
+	fn execute(
+		&self,
+		future: Box<dyn Future<Item = (), Error = ()> + Send>
+	) -> Result<(), futures::future::ExecuteError<Box<dyn Future<Item = (), Error = ()> + Send>>> {
+		if self.sender.is_closed() {
+			let kind = futures::future::ExecuteErrorKind::Shutdown;
+			Err(futures::future::ExecuteError::new(kind, future))
```
bkchr: Just map any error to shutdown, because no other error is thrown by an unbounded stream.

core/service/src/lib.rs
```diff
@@ -614,6 +622,22 @@ impl<Components> Future for Service<Components> where Components: components::Co
...
 
+impl<Components> Executor<Box<dyn Future<Item = (), Error = ()> + Send>>
+for Service<Components> where Components: components::Components {
+	fn execute(
+		&self,
+		future: Box<dyn Future<Item = (), Error = ()> + Send>
+	) -> Result<(), futures::future::ExecuteError<Box<dyn Future<Item = (), Error = ()> + Send>>> {
+		if self.to_spawn_tx.is_closed() {
+			let kind = futures::future::ExecuteErrorKind::Shutdown;
+			Err(futures::future::ExecuteError::new(kind, future))
```
bkchr: Same as above for the error.

core/service/src/lib.rs
bkchr: And that is also a possible race condition.

tomaka:
I thought that the error returned by the unbounded sender didn't provide the value back, but actually it does! I therefore changed it to not possibly have a race condition.

PR #3008, opened 2019-07-03
Generic keystore

dvc94ch:
As a first step towards setting keys via rpc this makes the keystore generic to support multiple cryptos.


core/primitives/src/crypto.rs
```diff
@@ -412,6 +431,31 @@ pub trait Pair: Sized + 'static {
...
+	fn to_raw_vec(&self) -> Vec<u8>;
+}
+
+/// An identifier for a type of cryptographic key.
+///
+/// 0-1024 are reserved.
+pub type KeyTypeId = u32;
+
+/// Constant key types.
+pub mod key_types {
```
tomaka: That looks pretty arbitrary. I assume it's used in the keystore, but then it should be IMO moved in the keystore crate.

dvc94ch: It's also going to be used in the session module, that code I copied from @rphmeier s PR

core/keystore/src/lib.rs
```diff
@@ -104,22 +120,30 @@ impl Store {
...
+			.filter_map(|(ty, public)| {
+				if *ty != TPublic::KEY_TYPE {
+					return None
+				}
+				Some(TPublic::from_slice(public))
+			})
+			.collect();
+
+		let key_type: [u8; 4] = unsafe {
+			std::mem::transmute(TPublic::KEY_TYPE.to_le())
```
rphmeier: use byteorder

core/keystore/src/lib.rs
```diff
@@ -129,9 +153,12 @@ impl Store {
 		Ok(public_keys)
 	}
 
-	fn key_file_path(&self, public: &Public) -> PathBuf {
+	fn key_file_path<TPair: Pair + TypedKey>(&self, public: &TPair::Public) -> PathBuf {
 		let mut buf = self.path.clone();
-		buf.push(hex::encode(public.as_slice()));
+		let bytes: [u8; 4] = unsafe { std::mem::transmute(TPair::KEY_TYPE.to_le()) };
```
rphmeier: use byteorder

rphmeier:
I'm not convinced this has been well thought-through in terms of how consensus engines are meant to deal with updates from the key-store.

Can you shed some light on the architecture ideas you have?

rphmeier:
(needs merge & tests fail)

tomaka:
(I'm not a keystore person and I just commented by curiosity, but looks good to me for what it's worth)

PR #3009, opened 2019-07-03
Make use of child storage for testExternalities and basicExternalities

thiolliere:
Currently all our test bypass child storage genesis config.

This change TestExternalities to be able to be build with the storage tuple `(top storage, child_storage)`
This also changes BasicExternalities to be able to handle children read and write.

I think the current API is not perfect and we should not distinguish children and top storage anymore at this level.
build_storage stuff should return only one storage struct and with_storage should by default take children storage. That would avoid dismissing child storages.
Even add_extra_genesis::build should give only one argument storage which contains both top and childs. But this is let to another PR.

@rphmeier this is a complete implementation, just use with_storage_and_children now

TODO:
* given input of basicExternalities should be checked ? like warn if a key in main storage is a valid child prefix ?


core/state-machine/src/testing.rs
```diff
@@ -72,17 +79,6 @@ impl<H: Hasher, N: ChangesTrieBlockNumber> TestExternalities<H, N> {
...
-	/// Iter to all pairs in key order
-	pub fn iter_pairs_in_order(&self) -> impl Iterator<Item=(Vec<u8>, Vec<u8>)> {
-		self.backend.pairs().iter()
-			.map(|&(ref k, ref v)| (k.to_vec(), Some(v.to_vec())))
-			.chain(self.overlay.committed.top.clone().into_iter().map(|(k, v)| (k, v.value)))
-			.chain(self.overlay.prospective.top.clone().into_iter().map(|(k, v)| (k, v.value)))
-			.collect::<BTreeMap<_, _>>()
-			.into_iter()
-			.filter_map(|(k, maybe_val)| maybe_val.map(|val| (k, val)))
-	}
```
thiolliere: at some point it would be better to modify it so it include all values including child storages ones.

But when I introduced it I think a few month ago and it doesn't seem to be used

core/state-machine/src/testing.rs
```diff
@@ -92,9 +88,29 @@ impl<H: Hasher, N: ChangesTrieBlockNumber> TestExternalities<H, N> {
 	pub fn changes_trie_storage(&mut self) -> &mut ChangesTrieInMemoryStorage<H, N> {
 		&mut self.changes_trie_storage
 	}
+
+	/// Return a new backend with all pending value.
+	pub fn commit_all(&self) -> InMemory<H> {
+		self.backend.update(
+			core::iter::empty()
```
bkchr: Why start with `empty` here?

core/state-machine/src/testing.rs
```diff
@@ -92,9 +88,29 @@ impl<H: Hasher, N: ChangesTrieBlockNumber> TestExternalities<H, N> {
...
+				.chain(self.overlay.committed.top.clone().into_iter()
+					.map(|(k, v)| (None, k, v.value)))
```
bkchr:
```suggestion
					.map(|(k, v)| (None, k, v.value))
				)
```

The rest should be formatted the same way.

core/state-machine/src/testing.rs
```diff
@@ -92,9 +88,29 @@ impl<H: Hasher, N: ChangesTrieBlockNumber> TestExternalities<H, N> {
...
+				.chain(self.overlay.prospective.top.clone().into_iter()
+					.map(|(k, v)| (None, k, v.value)))
+				.chain(self.overlay.committed.children.clone().into_iter()
+					.flat_map(|(keyspace, map)| map.1.into_iter()
+						.map(|(k, v)| (Some(keyspace.clone()), k, v))
```
bkchr: Here is the indentation even more confusing :D

core/state-machine/src/testing.rs
```diff
@@ -42,27 +44,32 @@ pub struct TestExternalities<H: Hasher, N: ChangesTrieBlockNumber> {
 
 impl<H: Hasher, N: ChangesTrieBlockNumber> TestExternalities<H, N> {
 	/// Create a new instance of `TestExternalities`
-	pub fn new(inner: HashMap<Vec<u8>, Vec<u8>>) -> Self {
-		Self::new_with_code(&[], inner)
+	pub fn new(storage: StorageTuple) -> Self {
```
bkchr: I think we should revert this change and add a `new_with_child` function.

thiolliere: hmm at the end I think user shouldn't differentiate top storage from child storages, also I feel like with children should be the default. I keep all the `From<top>` more for historical reason than because I want to. Also I can understand implementing your API

core/state-machine/src/testing.rs
thiolliere: for symetrie haha, ok removed and formatted

rphmeier:
@thiolliere It will need a corresponding update and PR in the Polkadot repo

thiolliere:
OK so I now included changes to basic to be able to handle children as well. I update PR description

thiolliere:
sorry for changing the goal of the PR, after some ppl reviewed I haven't seen Gav approval before comitting the changes. I know this is confusing and useless as the PR was fine

core/state-machine/src/testing.rs
bkchr: If I don't want to initialize a child storage, I will give `Default::default()` anyway and it seems that in most cases we don't want to initialize the child storage to something meaningful.

thiolliere: I made the change

core/state-machine/src/basic.rs
```diff
@@ -18,67 +18,64 @@
...
-		self.inner.insert(k, v)
+		self.top.insert(k, v)
 	}
-}
 
-impl ::std::fmt::Debug for BasicExternalities {
-	fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
-		write!(f, "{:?}", self.inner)
+	/// Get storages
+	pub fn storages(self) -> (
```
bkchr:
```suggestion
	pub fn into_storages(self) -> (
```

Or maybe `into_inner`.

core/state-machine/src/basic.rs
```diff
@@ -18,67 +18,64 @@
...
+		HashMap<Vec<u8>, HashMap<Vec<u8>, Vec<u8>>>,
+	) {
+		(self.top, self.children)
 	}
 }
 
 impl PartialEq for BasicExternalities {
 	fn eq(&self, other: &BasicExternalities) -> bool {
-		self.inner.eq(&other.inner)
+		self.top.eq(&other.top)
```
bkchr: Don't we also need to compare the child storage here?

core/state-machine/src/basic.rs
```diff
@@ -87,69 +84,79 @@ impl Default for BasicExternalities {
...
-		match key.as_ref() {
-			CODE => self.code = maybe_value,
-			_ => {
-				match maybe_value {
-					Some(value) => { self.inner.insert(key, value); }
-					None => { self.inner.remove(&key); }
-				}
-			}
+		if is_child_storage_key(&key) {
+			panic!("Refuse to directly set child storage key");
```
bkchr:
```suggestion
			panic!("Trying to set child storage key via main storage");
```

Or something similar? But we should improve the error message :)

core/state-machine/src/basic.rs
```diff
@@ -87,69 +84,79 @@ impl Default for BasicExternalities {
...
-	fn kill_child_storage(&mut self, _storage_key: ChildStorageKey<H>) { }
+	fn kill_child_storage(&mut self, storage_key: ChildStorageKey<H>) {
+		self.children.remove(storage_key.as_ref());
+	}
 
 	fn clear_prefix(&mut self, prefix: &[u8]) {
-		self.changes.clear_prefix(prefix);
-		self.inner.retain(|key, _| !key.starts_with(prefix));
+		if is_child_storage_key(&prefix) {
+			panic!("Refuse to directly set child storage key");
```
bkchr: Same.

bkchr: + make it clear that this is a different panic

core/state-machine/src/basic.rs
```diff
@@ -87,69 +87,79 @@ impl Default for BasicExternalities {
...
+			top: hashmap,
+			children: Default::default(),
 		}
 	}
 }
 
 impl<H: Hasher> Externalities<H> for BasicExternalities where H::Out: Ord {
 	fn storage(&self, key: &[u8]) -> Option<Vec<u8>> {
-		match key {
-			CODE => self.code.clone(),
```
bkchr: Did you test that this works?

core/state-machine/src/basic.rs
```diff
@@ -18,67 +18,67 @@
...
-		).expect("changes trie configuration is correct in test env; qed");
-
-		inner.insert(HEAP_PAGES.to_vec(), 8u64.encode());
+	/// Create a new instance of `BasicExternalities` with children
+	pub fn new_with_children(
+		mut top: HashMap<Vec<u8>, Vec<u8>>,
+		mut children: HashMap<Vec<u8>, HashMap<Vec<u8>, Vec<u8>>>,
+	) -> Self {
+		top.retain(|key, _| !is_child_storage_key(key));
+		children.retain(|key, _| is_child_storage_key(key));
```
thiolliere: those removals weren't done in the past. Hmm I'm not sure they are needed as those checks will be done when writing to the actual externality no ?

bkchr: I'm not happy with this and the one in "real externalities". This could get someone in trouble when the genesis storage fucks something up, but it will not be found, because we accepted everything.

thiolliere: "real externalities" is TestExternalities ?

maybe we can warn here as we do in `Ext` ?

bkchr: You called it `actual externality`, but I thought you speak about `Ext`.
I'm not sure a warning is enough. The question is, is this used in production code or just in tests and genesisconfig? If it is the later, I would maybe go with a panic to make the user aware that something is not correct.

core/state-machine/src/basic.rs
thiolliere: I improve a bit

core/state-machine/src/basic.rs
thiolliere: done, also I removed panic to a warning, as we shouldn't panic here

core/state-machine/src/basic.rs
thiolliere: oh yes `actual externality` was `Ext`.

no it is used in production, ppl have started writing with_storage in add_genesis_config::build, and this with_storage makes use of basicExternality.

I changed TestExternality to panic when given wrong inputs and BasicExternality not do anything on wrong input (as when storages of genesis config will be put into Ext it will actually warn there I suppose).

I think we might want TestExternality panics, BasicExternality warns.

core/state-machine/src/basic.rs
thiolliere: all CI works

core/sr-io/with_std.rs
```diff
@@ -418,9 +418,32 @@ pub type ChildrenStorageOverlay = HashMap<Vec<u8>, StorageOverlay>;
 pub fn with_storage<R, F: FnOnce() -> R>(storage: &mut StorageOverlay, f: F) -> R {
 	let mut alt_storage = Default::default();
 	rstd::mem::swap(&mut alt_storage, storage);
-	let mut ext: BasicExternalities = alt_storage.into();
+	let mut ext = BasicExternalities::new(alt_storage);
 	let r = ext::using(&mut ext, f);
-	*storage = ext.into();
+	*storage = ext.into_storages().0;
```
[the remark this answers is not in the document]
thiolliere: Yes, you can always do `with_storage_and_children(storage, Default::default())`

core/state-machine/src/basic.rs
bkchr: I think we also want `BasicExternality` to panic. If there is a serious problem in your genesisconfig, you probably want to be aware of it. A warning that is maybe not seen, does not help you.

PR #3010, opened 2019-07-03
Downgrade `memory-db` in `test-runtime` to fix WASM compilation

bkchr:


pepyakin:
I think there is no review (at least in conventional sense) required. However, I'd be curious to know what happened here

bkchr:
I would like to know the reason as well. It is really weird that it works for me, even without the patch :woman_shrugging:

Issue #2977, opened 2019-06-28
transaction factory generates blocks that can not be imported with import-blocks


pepyakin:
When I try to run `--import-blocks` it just seemingly stucks.

```
    Finished release [optimized] target(s) in 0.92s
     Running `target/release/substrate import-blocks blocks.blocks`
2019-06-28 18:18:28 Importing 1241 blocks
2019-06-28 18:18:28 #0 blocks were added to the queue
2019-06-28 18:18:28 #1000 blocks were added to the queue
```

There is no CPU activity coming from substrate while import so which suggests that there might be an issue with locking or dispatching the tasks.

tomaka:
Ah, my bad. We're using `tokio::run` instead of a method that exits immediately once the future passed as parameter has ended.

tomaka:
The fix is quite easy. Replace `tokio::run` with something like:

```rust
tokio::runtime::Runtime::new()
	.expect("Failed to initialize tokio")
	.block_on(fut)
	.expect("Error while importing blocks");
```

However the other part of the issue would be to add a CI test.

pepyakin:
Hey, thanks for the fix! However, unfortunately, it doesn't work. The symptoms are the same. There is no activity and no results.

arkpar:
Works for me in master. 
```
Initializing Genesis block/state (state: 0xea93…a60c, header-hash: 0x4830…833c)
2019-07-02 12:18:33 Loaded block-time = 4 seconds from genesis on first-launch
2019-07-02 12:18:33 Loading GRANDPA authority set from genesis on what appears to be first startup.
2019-07-02 12:18:33 Importing 1148 blocks
2019-07-02 12:18:33 #0 blocks were added to the queue
2019-07-02 12:18:33 #1000 blocks were added to the queue
2019-07-02 12:18:46 #1000 blocks were imported (#148 left)
2019-07-02 12:18:48 Imported 1147 blocks. Best: #1148
```

arkpar:
@pepyakin try importing to an empty DB

pepyakin:
Yes, I did the `purge-chain` before the import.

Just checked with the latest master:

```
./target/release/substrate purge-chain -y --dev  && time ./target/release/substrate factory --dev --mode MasterTo1 --num 500
./target/release/substrate export-blocks --dev blocks.blocks
./target/release/substrate import-blocks --dev blocks.blocks
```

the result is the same.

arkpar:
[...]

PR #2931, opened 2019-06-24
Ensure clean wasm instances

cmichi:
Addresses https://github.com/paritytech/substrate/issues/2051.

The runtime cache which substrate currently uses does reuse one runtime instance for every call.
In order to enable this the instance is cleaned up for every call. This leads to the problems detailed in 
https://github.com/paritytech/substrate/issues/2051.

Following up to private discussions with @pepyakin and @folsen, this PR tries out an alternative
approach: it replaces the runtime cache with a pool of pre-created runtime instances. For every call a "fresh" instance is taken out of the pool, a background thread monitors the pool and adds new instances.

In order to make this possible I had to implement threadsafe types for wasmi (https://github.com/paritytech/wasmi/pull/189, not yet merged).

I did a first (very very simple) benchmark by timing invocations of `call_in_wasm_module` and running `cargo run --release -- --dev` for 40 blocks:

```
AuraApi_authorities:
master:    169  µs (100%)
pool:      196  µs (115%)

AuraApi_slot_duration:
master:    2192 µs (100%)
pool:      96   µs (4%)

BlockBuilder_apply_extrinsic:
master:    1805 µs (100%)
pool:      2551 µs (141%)

BlockBuilder_finalize_block:
master:    6035 µs (100%)
pool:      4774 µs (79%)

BlockBuilder_inherent_extrinsics:
master:    996  µs (100%)
pool:      1416 µs (142%)

Core_execute_block:
master:    1335 µs (100%)
pool:      965  µs (72%)

Core_initialize_block:
master:    1731 µs (100%)
pool:      2143 µs (123%)

GrandpaApi_grandpa_authorities:
master:    2357 µs (100%)
pool:      150  µs (6%)

GrandpaApi_grandpa_forced_change:
master:    136  µs (100%)
pool:      155  µs (113%)

GrandpaApi_grandpa_pending_change:
master:    122  µs (100%)
pool:      121  µs (99%)

OffchainWorkerApi_offchain_worker:
master:    3    µs (100%)
pool:      11   µs (366%)
```

There is some room for optimization though (Jef had some comments on the wasmi PR).

At first we had intended for the runtime pool to do more complex actions when adding instances to the pool. Right now it just clones a runtime instance from a "template instance". I measured the clone operation in release mode on dev over 40 blocks and it takes on average about 1 µs on my machine. So it seems rather cheap and I'm wondering if we could maybe even do this synchronously without having the overhead of a separate thread?

[...]

PR #1345, opened 2019-01-03
State cache and other performance optimizations

arkpar:
Closes #257 

Major change introduced in this PR is canonical state cache. This works in the same way as in `parity-ethereum`. The cache caches a single state `C` that is considered "canonical" or "best". Any state `S` that is derived from `C` can use cashed value cache as long as the key is not overwritten in any of the blocks between `C` and `S`.

A few other minor changes:

* `changes` in`CallResult` and storage root calculation are not really required for each call. So the whole struct is dropped.
* Executor now queries the code hash from storage/cache instead of querying the code and hashing it every call.
* wasmi module instance is reused for each call. The instance is now cached per thread.

Issue #2022, opened 2019-03-16
InvalidTransaction on latest master (after first tx success, latest master)


jacogr:
After building current substrate master, there seems to be issues with the transaction handling. 

1. Node is started
2. Transfer is created and submitted
3. It succeeds, balances change, nonce increases
4. Another transfer is send (either the same account or a different one)
5. Transfer fails with `1010 InvalidTransaction`

To recover -

6. Stop the node
7. Restart the node
8. Submit hex value (for previous failed tx) exactly as-is again
9. Transaction succeeds, balances change, nonce increases
10. At this point any subsequent transfers gets it to point 5 again, with the same resurrection

gavofyork:
@tomusdrw any ideas? given a restart helps, could this be transaction queue?

tomusdrw:
Could be, I'll look into this, I think I experienced this locally on Friday as well.

tomusdrw:
Seems that the are two separate issues that manifest in the same way (they differ in the `data` part of the error message of `InvalidTransaction` - one is `0` (`BadSignature)` and second is `3` (`CantPay`):

1. The Authority account can't transfer funds after transferring (1 unit) them for 2-3 times. `validate_transaction` returns `CantPay` and wen debugging the runtime the free balance of the account is `0`
2. The newly created `sr25519` can't send any transactions and gets `BadSignature` from the runtime.

I'm currently tackling (2).

jacogr:
Tested against https://github.com/paritytech/substrate/pull/2028 - it seems to solve both the signature and fees issues (@tomusdrw would just appreciate another check from you)

PR #3011, opened 2019-07-03
Fair reusing of wasm runtime instances

### LINKED ISSUE #2051: Fix wasm module reuse
CC: @pepyakin 

Seems that the original cause for #2022 is somewhere in wasm module re-use.

```
diff --git a/core/executor/src/native_executor.rs b/core/executor/src/native_executor.rs
index 0944cbdb..f0793af3 100644
--- a/core/executor/src/native_executor.rs
+++ b/core/executor/src/native_executor.rs
@@ -58,6 +58,7 @@ fn fetch_cached_runtime_version<'a, E: Externalities<Blake2Hasher>>(
                Some(code_hash) => code_hash,
                None => return Err(ErrorKind::InvalidCode(vec![]).into()),
        };
+       cache.borrow_mut().remove(&code_hash[..]);
        let maybe_runtime_preproc = cache.borrow_mut().entry(code_hash.into())
                .or_insert_with(|| {
                        let code = match ext.storage(well_known_keys::CODE) {

```

This change causes the issue to disappear.

See branch https://github.com/paritytech/substrate/tree/td-debugging it's forked off from the #1800 that introduced `spin::Once`, but contains all the fixes that allow to reproduce currently using deployed version of https://polkadot.js.org/apps

Note that last commit is the patch above, so revert if you want to reproduce the issue:
1. Send one transaction from any account
2. Send 1-2 more transactions form the same account, they will fail with `InvalidTransaction`

### LINKED ISSUE #2967: Wasm stack pointer is not restored to its initial value
While I was pondering about #2051 I recalled that the current wasm executor doesn't restore the stack pointer global. So I am writing it here for making sure that we solve this problem. 

Basically, at the moment we are trying to reuse the same wasm instance for multiple invocations and trying to restore it after we done with each invocation. We are trying to restore the memory contents after execution, however, we are not restoring global variables and tables. The latter is OK since we don't perform any changes to it (at the moment of writing). 
However, some globals are actually changes: the stack pointer.

The stack pointer is usually used as something like this:

1. At the start of execution it is always* equals 0.
2. Call into the runtime is performed.
3. When throughout execution any function requires to place some data on stack, a stack frame is allocated, meaning the value of the stack pointer is decreased (stacks grow downwards in wasm). Before the function returns it increases the stack pointer to the last value.
4. When the last function returns, the function should* return the stack pointer to 0.

There is one big catch to the above though:

\* - this is the case only if the execution returned successfully.

In other words, if execution of the runtime trapped then the stack pointer won't get back to its initial value. This doesn't completely corrupt the wasm instance, but rather effectively decrease the amount of stack space available. This is also non-deterministic, since:

- if you restart the node, it will get back to normal
- the trap can happen while execution of RPC

The solution for this is to reset all global values to their initial states.

The test case I think would be:

1. Write a runtime function that allocates considerable amount of stack space, let's say a bit more than half of the available space (by default it is set to 1MB), and then traps if some input parameter equals to some specific value.
2. Call into this runtime function so that it traps. This should fail and consume a little bit more than half of the available stack space.
3. Call into the same function but it should not trap. If the bug is present this should fail as well, otherwise it should succeed.

pepyakin:
Closes #2051 
Closes #2967
Supersedes #2938

## Background

In #1345 an optimization for wasm runtime instantiation was implemented. The idea was that we would track the areas using some heuristic (`used_size` and `lowest_used`) that were used in the course of runtime execution and then zero them. However, it turned out that:

1. It assumed that the area where rustc usually puts the data section (i.e. static data, globals, etc) doesn't change. Introduction of mutable static variables in the runtime violated this assumption and ultimately caused #2051.
2. Upon a closer inspection of the heuristic, it cleared only the stack leaving the heap untouched. This can potentially (however, unlikely) cause undefined behavior because a compiler can assume and take advantage of the fact that WebAssembly guarantees the memory to be zeroed.

There were a couple attempts to fix it:

1. Initially, we thought that full instantiation in the background will do and tried to implement this approach in #2931, but it turned out that this would make us use `Arc` instead of `Rc` in wasmi which would considerably harm the execution performance. 
2. Then we assumed that maybe the initial assumption that @arkpar is not correct and instantiation doesn't take a lot of time. So we came up with #2938. Initial bencmark results showed potential in this approach, however, it turned out that the algorithm was flawed since it copied only up to `used_size` (the highest accessed memory location within the execution) and @arkpar was indeed right: there is a lot of unnecessary work being done. The current node-runtime allocates 1MB of stack space, has data section and on top of that we preallocate a lot of memory upfront, and copying and/or zeroing it takes a lot of time. The execution time on generating 1500 blocks almost doubled.

It turned out that it was hard to fix this issue with the given constraints (e.g. preserving determinism, allowing of execution of general wasm modules, etc) without regressing the performance. 

One option was to actually ban mutable static variables in the runtime. However, IMO, I think it is hard to ensure that we don't accidentally bring some sort of mutable static globals into the runtime and also it seems that they can be indeed useful.

## Solution

Because of the above I decided to take the following approach along with the following tradeoffs:

1. Implement the `mmap` backend for linear memory in wasmi https://github.com/paritytech/wasmi/pull/190 and introduce a function for quickly zeroing the memory instance. This allowed us to allocate arbitrary amounts of memory which is zeroed lazily on the first access. `erase` function allows to basically reallocate. The downside is that `mmap` is inherently platform dependent and works only on unix systems. E.g. Windows will suffer noticable slowdowns. I also tried to leverage the [`GlobalAlloc::alloc_zeroed`](https://doc.rust-lang.org/1.34.1/alloc/alloc/trait.GlobalAlloc.html#method.alloc_zeroed) function, but it turns out that it falls back to `bzero` (analogue of `memset` specialized for 0) on macOS which gives noticeble slowdowns (approx. 40 secs vs 230 secs for transaction factory of master on MasterTo1 with 1500 blocks). However,  we can do the similar trick on Windows since it has the similar APIs.
1. Extract the data segments (chunks of memory that are copied to the linear memory at the instantiation time) from the wasm module and cache them. At the time when the new wasm runtime instance is requested, the linear memory is zeroed and then the data segments are copied.
1. Scan the mutable global variables and restore their values when the new wasm runtime instance is requested. This solves the https://github.com/paritytech/substrate/issues/2967.
1. **Ban the `start` function** in wasm runtime modules. `start` function is run as the last step of instantiation and theoretically can do arbitrary changes to the wasm memory. This is OK for rustc produced binaries since they don't actually leverage the `start` function and IMO it is unlikely that it will use them. However, it might pose a problem for languages such as C++ which can have "pre-main" logic such as constructor initialization used in globals. This is not a fundamential problem since we can still run the `start` function and scan the memory for chunks of memory instead of relying on the static data segments. So this restriction can be lifted in the follow-up PRs.
1. **Require the `__heap_base` global variable**. It turned out that the allocator also depended on the `used_size` to detect the starting position of the heap which can be problematic. Luckily for us, wasm-ld (LLD) has a convention to expose the `__heap_base` in the produced binaries which points to the one byte past from the data section which we can use to seed the allocator. This solves potential problems that the `used_size` heuristic can miss, like BSS. I am not sure how this restriction can be lifted in the future though.

The initial benchmarks shows that there is no noticeble regression introduced in this PR compared to the heuristic approach we were using. Although unexpectedly now substrate performs a little bit better on macOS than on Linux.

I hope that these tradeoffs are reasonable. 

cc @arkpar @gavofyork @cmichi 

TODO:

- [x] test the performance on block import (was blocked #2977)
- [x] deal with the broken libc dependency (see the CI failure)


core/executor/src/native_executor.rs
```diff
@@ -140,18 +74,18 @@ pub struct NativeExecutor<D> {
...
 impl<D: NativeExecutionDispatch> NativeExecutor<D> {
 	/// Create new instance.
-	pub fn new(default_heap_pages: Option<u64>) -> Self {
+	pub fn new(initial_heap_pages: Option<u64>) -> Self {
 		NativeExecutor {
 			_dummy: Default::default(),
 			fallback: WasmExecutor::new(),
 			native_version: D::native_version(),
-			default_heap_pages,
+			initial_heap_pages: initial_heap_pages,
```
folsen: Do we have any consistent style on whether to use shortened form here or not? I would have just put `initial_heap_pages` and not repeated the line (like the one you removed). It's not in the styleguide so nothing is right or wrong.

pepyakin: ah, this is one thing that I forgot to fix actually! I asked @cmichi to come up with something different than `default_heap_pages` since we already have `DEFAULT_HEAP_PAGES` and proposed `initial_heap_pages` but now I think it is even worse

folsen:
What do we need to do as we add more execution environments in the future, the most relevant of course being lightbeam. Lets say we integrate lightbeam and start compiling the wasm to native. Can we make the above changes to execution environment in a reasonable way for any other compiler in the future?

core/executor/src/wasm_runtimes_cache.rs
```diff
@@ -0,0 +1,316 @@
...
+	/// Reset the runtime instance to the initial version by restoring
+	/// the preserved memory and globals.
+	///
+	/// Returns `Err` if applying the snapshot is failed.
+	fn apply(&self, instance: &WasmModuleInstanceRef) -> Result<(), CacheError> {
+		let memory = instance
+			.export_by_name("memory")
+			.ok_or(CacheError::ApplySnapshotFailed)?
+			.as_memory()
+			.cloned()
```
bkchr: COuld use `and_then` as well to have just one `ok_or`.

core/executor/src/wasm_runtimes_cache.rs
```diff
@@ -0,0 +1,316 @@
...
+				.set(*offset, contents)
+				.map_err(|_| CacheError::ApplySnapshotFailed)?;
+		}
+
+		// Second, restore the values of mutable globals.
+		for (global_ref, global_val) in instance
+			.globals()
+			.iter()
+			.filter(|g| g.is_mutable())
+			.zip(self.global_mut_values.iter())
```
bkchr: We could here just use a `try_for_each` and drop the last `Ok(())`.

core/executor/src/wasm_runtimes_cache.rs
```diff
@@ -0,0 +1,316 @@
...
+		let instance = WasmExecutor::instantiate_module::<E>(heap_pages as usize, &module)
+			.map_err(|_| CacheError::Instantiation)?;
+
+		// Take state snapshot before executing anything.
+		let state_snapshot = StateSnapshot::take(&instance, data_segments, heap_pages as u32)
+			.expect(
+				"`take` returns `Err` if the module is not valid;
+				we already loaded module above, thus the `Module` is proven to be valid at this point;
+				qed
+			",
```
bkchr:
```suggestion
				",
```

core/test-runtime/src/lib.rs
```diff
@@ -591,6 +608,39 @@ cfg_if! {
...
+					// Then make sure we actually write something to it.
+					//
+					// If:
+					// 1. the stack area is placed at the beginning of the linear memory space, and
+					// 2. the stack pointer points to out-of-bounds area, and
+					// 3. a write is performed around the current stack pointer.
+					//
+					// then a trap should happen.
+					//
+					for (i, v) in data.iter_mut().enumerate() {
```
bkchr: Are we sure that the compiler does not optimizes this away? Maybe also return the result of this function?

core/test-runtime/src/lib.rs
```diff
@@ -248,6 +248,8 @@ cfg_if! {
 				fn use_trie() -> u64;
 				fn benchmark_indirect_call() -> u64;
 				fn benchmark_direct_call() -> u64;
+				fn returns_mutable_static() -> u64;
+				fn allocates_stack_alot(trap: bool);
```
bkchr:
```suggestion
				fn allocates_huge_stack_array(trap: bool);
```

Or something similar.

bkchr:
A more general note, I would really like to see the cache being passed in instead of being a `thread_local!`, but that probably requires greater refactoring...

pepyakin:
@folsen 

Yeah, I was trying to keep that in mind. 

You might have noticed that I added `parity-wasm` as a dependency even though at the moment `wasmi` is built on top of `parity-wasm` but just tries to not expose it. At the cost of additional deserialization on the cold path we can get away without dependning on the execution engine providing the data segments.

If we were to integrate wasmtime (and by extension lightbeam) we would need to do similar changes. wasmtime already uses `mmap`-s for implementation of their linear memories. Globals seem to be a little bit more involved, but I think it would be possible to implement them in the engines as well.

@bkchr Yes! While working I noticed that the executor crate (and also state-machine) need some love. I was trying to hold myself from refactorings because I assumed that we need to land this PR quick and without big refactorings due to the upcoming audit and upcoming releases.

core/test-runtime/src/lib.rs
pepyakin: Good point! I actually tried to do that with `set_storage` but it legitamately runs out of stack space in that case. It would be great if we could use `data.as_ptr()` somewhere where optimizer couldn't get that it is being tricked! Do you have an idea how?

bkchr: Could we not just call `to_vec()` and return a `Vec`?

core/executor/src/wasm_runtimes_cache.rs
pepyakin: Haha, tried to eradicate `and_then`'s since I have (small) allergy on them. Put a thumbs up on this comment if you feel strongly about this.

bkchr:
As I need to have extendable externalities for Cumulus, this will probably be my job to refactor this stuff :sweat_smile:

core/executor/src/error.rs
```diff
@@ -58,6 +58,13 @@ pub enum Error {
 	/// Invalid memory reference.
 	#[display(fmt="Invalid memory reference")]
 	InvalidMemoryReference,
+	/// The runtime must provide a global named `__heap_base` of type i32 for specifying where the
+	/// allocator is allowed to place its data.
+	#[display(fmt="The runtime doesn't provide a global named __heap_base")]
+	HeapBaseNotFoundOrInvalid,
+	/// The runtime WebAssembly module is not allowed to have the `start` function.
+	#[display(fmt="The runtime has the \"start\" function")]
```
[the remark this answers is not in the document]
pepyakin: I thought for the console output it might make sense to use `"start"` instead of start with ``, put a thumbs up on this comment if you feel strongly about this and I will change it!

core/sr-api-macros/tests/runtime_calls.rs
```diff
@@ -191,3 +192,38 @@ fn record_proof_works() {
...
+	let block_id = BlockId::Number(client.info().chain.best_number);
+
+	// On the first invocation we allocate approx. 75% of stack and then trap.
+	let ret = runtime_api.allocates_stack_alot(&block_id, true);
+	assert!(ret.is_err());
+
+	// On the second invocation we allocate yet another 75% of stack 
+	let ret = runtime_api.allocates_stack_alot(&block_id, false);
+	assert!(ret.is_ok());
+}
```
pepyakin: This place is super convenient to place these tests. However, these tests test the runtime behavior and it feels weird to place them here. Do you have a better idea where to put them @bkchr ?

core/test-runtime/src/lib.rs
pepyakin: Let me try! Actually now I don't see what this could be a problem and if there is then it is most likely worth attention!

core/executor/src/wasm_runtimes_cache.rs
pepyakin: Ah, nice to know about `try_for_each`. However, I personally prefer good ol' for loops instead of combinator chains. The reason is one should be really careful using them when the chain of combinators becomes long and nested.

IMO combinators fail to convey the deep nesting compared to plain old control flow structures. In this particular case, they will also introduce additional indent which can harm the comments readability. 

That said, I don't feel strongly about this, if you do I can change it!

core/executor/src/error.rs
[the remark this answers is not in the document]
bkchr: Don't have a strong opinion on this, but have seen this a lot in the rust ecosystem, especially in `rustc`. But often I would like to have a terminal that understands markdown for this :man_shrugging: So, do what you think is the best :)

core/sr-api-macros/tests/runtime_calls.rs
bkchr: Yeah, that is correct. I used this here to test the actual behavior of `sr-api-macros`. 

Why not put the tests directly into `test-runtime`? I think that is the best place.

core/test-runtime/src/lib.rs
pepyakin: Found it! Turns out that the heap is only 1mb and can't fit the allocation of 786432 bytes

core/sr-api-macros/tests/runtime_calls.rs
pepyakin: But can we do that? These test use `TestClient` which is from `test-client` crate which depends on the `test-runtime`...

core/executor/src/error.rs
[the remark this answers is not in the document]
pepyakin: Ticks have an advantage that they don't require escaping and thus are a bit more readable

core/executor/src/native_executor.rs
```diff
@@ -140,18 +74,18 @@ pub struct NativeExecutor<D> {
 	fallback: WasmExecutor,
 	/// Native runtime version info.
 	native_version: NativeVersion,
-	/// The default number of 64KB pages to allocate for Wasm execution.
-	default_heap_pages: Option<u64>,
+	/// The number of 64KB pages to allocate for Wasm execution.
+	initial_heap_pages: Option<u64>,
```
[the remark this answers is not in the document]
pepyakin: No. AFAIR this comes from the command line and cannot be increased within the same invocation of the node.

However, this parameter is not the only source of the heap pages parameter for the runtime executor. For now it is:

1. The storage variable `:heappages`.
2. If it doesn't exist, take `initial_heap_pages`.
3. If it wasn't supplied, use the `DEFAULT_HEAP_PAGES` constant.

core/sr-api-macros/tests/runtime_calls.rs
bkchr: Yes, when you put this into `dev-dependencies`. We do this in multiple places in Substrate^^

pepyakin:
## Setup

The tests are conducted in transaction-factory.

In order to benchmark, I had to apply patches that looks like this:

```diff
diff --git a/core/executor/src/native_executor.rs b/core/executor/src/native_executor.rs
index 506cd46af..5a5ddebda 100644
--- a/core/executor/src/native_executor.rs
+++ b/core/executor/src/native_executor.rs
@@ -142,6 +142,10 @@ impl<D: NativeExecutionDispatch> CodeExecutor<Blake2Hasher> for NativeExecutor<D
                                Ok((module, onchain_version)) => (module, onchain_version),
                                Err(e) => return (Err(e), false),
                        };
+
+                       // TODO:
+                       let use_native = false;
+
                        match (
                                use_native,
                                onchain_version
diff --git a/node/cli/src/factory_impl.rs b/node/cli/src/factory_impl.rs
index 0d9461036..81998b219 100644
--- a/node/cli/src/factory_impl.rs
+++ b/node/cli/src/factory_impl.rs
@@ -157,7 +157,7 @@ impl RuntimeAdapter for FactoryState<Number> {

        fn minimum_balance() -> Self::Number {
                // TODO get correct amount via api. See #2587.
-               1337
+               500
        }

        fn master_account_id() -> Self::AccountId {

```

`use_native = false` makes the native-executor to use the wasm fallback unconditionally. `minimum_balance` change is needed for not running out of funds while execution transaction-factory.

Here is the script that I used for getting the master build and the build that comes from this PR:

```sh
#!/bin/bash

set -eux pipefail

export CARGO_INCREMENTAL=0
export RUSTFLAGS="-Ccodegen-units=1"

cd substrate
git reset --hard 42f564a6772cb3493342f3ae5fe722b44aa9c44c
git apply ../fixes-master.patch
cargo build --release
cp target/release/substrate ../substrate-master
cd ..

cd substrate-2
git reset --hard d053556bde319156a8a13b79988e79547e3b4223
git apply ../fixes.patch
cargo build --release
cp target/release/substrate ../substrate-cached
```

and here is how I run it:

```
#!/bin/bash

./substrate-master purge-chain -y --dev && time ./substrate-master factory --dev --mode MasterTo1 --num 1500 >& master.log
./substrate-cached purge-chain -y --dev && time ./substrate-cached factory --dev --mode MasterTo1 --num 1500 >& cached.log
```

## Results

|       |  real                     | sys          | 
|-------|----------------------------|---------------------------|
| master | 51.954s | 0.164s | 
| cached (this PR) | 53.994s  | 0.983s |

The increase in sys time is expected since now the OS clears pages (virtual memory pages, i.e. 4Kb, not wasm linear memory pages, i.e. 64Kb) lazily on the first access to them.

gavofyork:
So... overall this is a minor pessimisation? Remind me of the advantage to doing this? :)

burdges:
You might observe an improvement when swapping the wasm instance batch and forth between large and small runtimes, no?

DemiMarie-parity:
> Closes #2051
> Closes #2967
> Supersedes #2938
> 
> ## Background
> In #1345 an optimization for wasm runtime instantiation was implemented. The idea was that we would track the areas using some heuristic (`used_size` and `lowest_used`) that were used in the course of runtime execution and then zero them. However, it turned out that:
> 
> 1. It assumed that the area where rustc usually puts the data section (i.e. static data, globals, etc) doesn't change. Introduction of mutable static variables in the runtime violated this assumption and ultimately caused #2051.
> 2. Upon a closer inspection of the heuristic, it cleared only the stack leaving the heap untouched. This can potentially (however, unlikely) cause undefined behavior because a compiler can assume and take advantage of the fact that WebAssembly guarantees the memory to be zeroed.
> 
> There were a couple attempts to fix it:
> 
> 1. Initially, we thought that full instantiation in the background will do and tried to implement this approach in #2931, but it turned out that this would make us use `Arc` instead of `Rc` in wasmi which would considerably harm the execution performance.
> 2. Then we assumed that maybe the initial assumption that @arkpar is not correct and instantiation doesn't take a lot of time. So we came up with #2938. Initial bencmark results showed potential in this approach, however, it turned out that the algorithm was flawed since it copied only up to `used_size` (the highest accessed memory location within the execution) and @arkpar was indeed right: there is a lot of unnecessary work being done. The current node-runtime allocates 1MB of stack space, has data section and on top of that we preallocate a lot of memory upfront, and copying and/or zeroing it takes a lot of time. The execution time on generating 1500 blocks almost doubled.
> 
> It turned out that it was hard to fix this issue with the given constraints (e.g. preserving determinism, allowing of execution of general wasm modules, etc) without regressing the performance.
> 
> One option was to actually ban mutable static variables in the runtime. However, IMO, I think it is hard to ensure that we don't accidentally bring some sort of mutable static globals into the runtime and also it seems that they can be indeed useful.
> 
One solution would be to use `mprotect` to make these pages read-only.  If the wasm program tries to write to them, an exception (`SIGSEGV` or `STATUS_ACCESS_VIOLATION`) will occur, which we can catch.
> ## Solution
> Because of the above I decided to take the following approach along with the following tradeoffs:
> 
> 1. Implement the `mmap` backend for linear memory in wasmi [paritytech/wasmi#190](https://github.com/paritytech/wasmi/pull/190) and introduce a function for quickly zeroing the memory instance. This allowed us to allocate arbitrary amounts of memory which is zeroed lazily on the first access. `erase` function allows to basically reallocate. The downside is that `mmap` is inherently platform dependent and works only on unix systems. E.g. Windows will suffer noticable slowdowns. I also tried to leverage the [`GlobalAlloc::alloc_zeroed`](https://doc.rust-lang.org/1.34.1/alloc/alloc/trait.GlobalAlloc.html#method.alloc_zeroed) function, but it turns out that it falls back to `bzero` (analogue of `memset` specialized for 0) on macOS which gives noticeble slowdowns (approx. 40 secs vs 230 secs for transaction factory of master on MasterTo1 with 1500 blocks). However,  we can do the similar trick on Windows since it has the similar APIs.
> 2. Extract the data segments (chunks of memory that are copied to the linear memory at the instantiation time) from the wasm module and cache them. At the time when the new wasm runtime instance is requested, the linear memory is zeroed and then the data segments are copied.

Do we need to zero all of linear memory?  The kernel certainly knows which pages it had to fault in, and there might be a (non-portable) way to get that information from userspace.  There certainly is on Linux (userfaultfd).

pepyakin:
@gavofyork 

> So... overall this is a minor pessimisation? Remind me of the advantage to doing this? :)

Yep, that's right. 

Right now we are reusing a wasm instance for multiple calls into the runtime and doing a bare minimum amount of work to restore the wasm. Specifically, looking at the code, we are clearing the area that **starts** at the lowest accessed address **after** the execution and **ends** at the lowest accessed address **before** of the execution. This works well for clearing the stack space area for code that was linked LLD with the `-z stack-first` option which is used by the current versions of rustc. Because of the amount of work done it has such great performance which was hard to even get close to.

Unfortunately, this does not return the wasm linear memory to the initial state:

- Contents of the data section in the linear memory are changed and not restored. For instance, the second invocation of the runtime can observe changes to mutable global variables made from the first one. This actually caused #2022 / #2051 and can cause problems similar to #2967 for compilers that put the stack pointer into the linear memory .
- The heuristic relies on the fact that memory is written to and has some assumptions about it. For example, if a wasm runtime puts the stack not at the start of linear memory then the runtime will break.
- We actually don't restore the contents of the heap at all.
- Things like .bss sections might just break.
- memory corruptions from RPC calls can have effect on other because we share the single runtime

TL;DR:

The current solution seems to work (modulo #2051) for the current toolchain. However, it is hard to predict if it will work in the future for rustc+LLD and especially for other toolchains. It is also not clear if   
we are actually missing something again. The current solution leaves non-deterministic memory contents which can potentially lead to consensus issues. This solution also wasmi specific and would require some support on implementation of these features in the other engines we might want to use and most importantly it will impose additional rather high overhead (basically every store should be accompanied with checking minimum and and maximum address accessed).

On the other hand, the solution proposed in the current PR doesn't employ any heuristics and ensure that the wasm instance is returned to the initial state without cutting any corners and with IMO not essential tradeoffs. This solution also is similar what would do other wasm engines and would not require substantial modifications to them (if any at all).

@DemiMarie-parity 

> One solution would be to use mprotect to make these pages read-only. If the wasm program tries to write to them, an exception (SIGSEGV or STATUS_ACCESS_VIOLATION) will occur, which we can catch.

These pages are actually not read-only and can be mutated in normal circumstances, see #2051. Even if we agreed that we do not want to support modules that modify the data section (which IMO would be restrictive, e.g. because as I mentioned there at least might be compilers which put e.g. a stack pointer there, nevermind that there might be other reasons to use it), we would need to also require some ways to indicate the data section (this is rather a vague concept from the standpoint of a wasm binary. A wasm compiler can layout the memory in whatever way it deems necessary.). Moreover there are no guarantees that these sections will be placed at the page boundary, which means either dynamic checking of boundaries or imposing additional non-trivial restrictions on wasm modules.

TBH this sounds as no-go to me.

> Do we need to zero all of linear memory? The kernel certainly knows which pages it had to fault in, and there might be a (non-portable) way to get that information from userspace. There certainly is on Linux (userfaultfd).

Yes, we do. I am not sure why would we need that if `mmap` allocates virtual pages already and zeroes them on the first access. I am not sure what advantage of doing this in the userspace would give us.

Also, as another point, I am not sure if such deep changes are worth it at the moment. When I spoke with @folsen we agreed that we don't want to do any sophisticated changes. Also IMO it is not worth further optimizing wasmi and going far beyond relatively simple `mmap`/`VirtualAlloc` approach and especially modifying rustc/LLVM/LLD for adding support for wasm modules that obey to restrictions (like aligning stack to virtual page (non wasm linear memory page) boundary).

core/executor/src/wasm_runtimes_cache.rs
```diff
@@ -0,0 +1,328 @@
...
+
+/// Cache for the runtimes.
+///
+/// When an instance is requested for the first time it is added to this
+/// cache. Furthermore its initial memory and values of mutable globals are preserved here. Follow-up
+/// requests to fetch a runtime return this one instance with the memory
+/// reset to the initial memory. So, one runtime instance is reused for
+/// every fetch request.
+///
+/// For now the cache grows indefinitely, but that should be fine for now since runtimes can only be
```
DemiMarie-parity: I agree with this, but there should probably be an issue for this.  That said, I think that issue can be closed as WONTFIX unless it becomes an issue, and it almost certainly will not.

core/executor/src/wasm_runtimes_cache.rs
```diff
@@ -0,0 +1,328 @@
...
+				Ok(ref cached_runtime) => Ok(Rc::clone(cached_runtime)),
+			};
+
+		match self.instances.entry(code_hash.into()) {
+			Entry::Occupied(o) => handle_result(o.get()),
+			Entry::Vacant(v) => {
+				trace!(target: "runtimes_cache", "no instance found in cache, creating now.");
+				let result = Self::create_wasm_instance(wasm_executor, ext, default_heap_pages);
+				if let Err(ref err) = result {
+					warn!(target: "runtimes_cachce", "cannot create a runtime: {:?}", err);
```
DemiMarie-parity:
```suggestion
					warn!(target: "runtimes_cache", "cannot create a runtime: {:?}", err);
```

pepyakin: Thanks!

core/executor/src/wasm_executor.rs
```diff
@@ -1371,43 +1386,32 @@ impl WasmExecutor {
...
 			module,
 			&ImportsBuilder::new()
 			.with_resolver("env", FunctionExecutor::<E>::resolver())
 		)?;
 
-		// extract a reference to a linear memory, optional reference to a table
-		// and then initialize FunctionExecutor.
+		// extract a reference to a linear memory and extract the `__heap_base` start
 		let memory = Self::get_mem_instance(intermediate_instance.not_started_instance())?;
+		let _heap_base = Self::get_heap_base(intermediate_instance.not_started_instance())?;
```
jimpo: Is the purpose of this just to verify that the heap_base is set? If so the comment should reflect this -- the term "extract" implies to me that it will be used.

core/executor/src/wasm_runtimes_cache.rs
```diff
@@ -0,0 +1,328 @@
...
+}
+
+impl StateSnapshot {
+	// Returns `None` if instance is not valid.
+	fn take(
+		module_instance: &WasmModuleInstanceRef,
+		data_segments: Vec<DataSegment>,
+		heap_pages: u32,
+	) -> Option<Self> {
+		let mut prepared_segments = Vec::with_capacity(data_segments.len());
```
jimpo: nit: This can be created without mutability:

```rust
let prepared_segments = data_segments.into_iter()
    .map(...)
    .collect::<Option<_>>()?;
```

core/executor/src/wasm_runtimes_cache.rs
```diff
@@ -0,0 +1,328 @@
...
+
+		let heap_pages = ext
+			.storage(well_known_keys::HEAP_PAGES)
+			.and_then(|pages| u64::decode(&mut &pages[..]))
+			.or(default_heap_pages)
+			.unwrap_or(DEFAULT_HEAP_PAGES);
+
+		// Instantiate this module.
+		let instance = WasmExecutor::instantiate_module::<E>(heap_pages as usize, &module)
+			.map_err(|_| CacheError::Instantiation)?;
```
jimpo: This error shouldn't get silently dropped -- it made it harder to find the issue with `__heap_base`.

core/executor/src/wasm_runtimes_cache.rs
```diff
@@ -0,0 +1,328 @@
...
+		})
+	}
+
+	/// Reset the runtime instance to the initial version by restoring
+	/// the preserved memory and globals.
+	///
+	/// Returns `Err` if applying the snapshot is failed.
+	fn apply(&self, instance: &WasmModuleInstanceRef) -> Result<(), CacheError> {
+		let memory = instance
+			.export_by_name("memory")
```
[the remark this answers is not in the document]
pepyakin: Theoretically `export_by_name` is more proper since there is no way to ensure that the right memory will be at index 0. Practically, however, since at the moment at most one memory is allowed, we could use it (Although we would need to export `memory_by_index`, it is private right now).

However, IMO I think we should stick to names. I had some ideas about extracting a cached version of runtime instance where things like `MemoryRef` are preloaded. According to my profiling this should give noticable improvements.

core/executor/src/wasm_executor.rs
pepyakin: Ahh, you are right! Yes, we check that `__heap_base` exists (and return `Err` if it doesn't`). I was also thinking about leaving the comment but apparently forgot about it!

core/executor/src/wasm_runtimes_cache.rs
```diff
@@ -0,0 +1,331 @@
...
+/// reset to the initial memory. So, one runtime instance is reused for
+/// every fetch request.
+///
+/// For now the cache grows indefinitely, but that should be fine for now since runtimes can only be
+/// upgraded rarely and there are no other ways to make the node to execute some other runtime.
+pub struct RuntimesCache {
+	/// A cache of runtime instances along with metadata, ready to be reused.
+	///
+	/// Instances are keyed by the hash of their code.
+	instances: HashMap<[u8; 32], Result<Rc<CachedRuntime>, CacheError>>,
```
jimpo: Why keep around multiple cached runtimes instead of just the last one?

pepyakin:
Merging this since 20-25 us slowdown is tolerable.

core/executor/src/wasm_runtimes_cache.rs
bkchr: Because you can make RPC calls into old runtimes.

PR #3013, opened 2019-07-04
Adjust the RPC service for the WASM-browser use case

tomaka:
cc #2416 

Several changes:

- `StartRPC::start_rpc` now returns the RPC handlers instead of opaque RPC servers. The opaque RPC servers are created in `lib.rs`.
- I had to introduce platform-specific code in `rpc-servers` so that we don't depend on `mio`. It's debatable whether to do that or to instead directly depend on `rpc` from the service.
- Adds an `RpcSession` type that allows performing in-memory RPC queries.


core/service/src/lib.rs
```diff
@@ -698,22 +701,74 @@ impl<Components> Drop for Service<Components> where Components: components::Comp
...
+}
+
+impl RpcSession {
+	/// Creates an RPC session.
+	///
+	/// The `sender` is stored inside the `RpcSession` and is used to communicate spontaneous JSON
+	/// messages.
+	///
+	/// The `RpcSession` must be kept alive in order to receive messages on the sender.
+	pub fn new(sender: mpsc::Sender<String>) -> RpcSession {
```
tomusdrw: Maybe we could have the session return `(RpcSession, impl Stream<String>)` instead of expecting a channel?

core/service/src/lib.rs
```diff
@@ -698,22 +701,74 @@ impl<Components> Drop for Service<Components> where Components: components::Comp
...
+		RpcSession {
+			metadata: rpc::Metadata::new(sender)
```
tomusdrw:
```suggestion
			metadata: rpc::Metadata::new(sender),
```

core/rpc-servers/src/lib.rs
```diff
@@ -30,9 +30,12 @@ const MAX_PAYLOAD: usize = 15 * 1024 * 1024;
 /// Default maximum number of connections for WS RPC servers.
 const WS_MAX_CONNECTIONS: usize = 100;
 
-type Metadata = apis::metadata::Metadata;
-type RpcHandler = pubsub::PubSubHandler<Metadata>;
+pub type Metadata = apis::metadata::Metadata;
+pub type RpcHandler = pubsub::PubSubHandler<Metadata>;
+
+#[cfg(not(target_os = "unknown"))]
```
tomusdrw: Wouldn't it be better to have a module with only single `#[cfg]` and reexport everything from that module, i.e.:
```rust
mod inner {
  pub type Metadata = ...;
  ...
  
}

#[cfg(target_os = "unknown")]
mod inner {}

pub use self::inner::*;

File: core/service/src/lib.rs (hunk 1 above)
Comment (tomaka) [reply core/service/src/lib.rs]: What I initially wanted to do is have the `RpcSession` *be* the stream, so that it's also obvious that you have to keep it alive.

However the "problem" is that polling the `RpcSession` (or the `impl Stream` in your suggestion) won't give any result unless you poll the `Service` as well, which is confusing.

In the end I went for explicitly putting the `Sender` in the API so that people realize that the receiver side isn't going to produce anything, else another mechanism does so.

File: .gitlab-ci.yml [hunk 4]
```diff
@@ -191,6 +191,7 @@ check-web-wasm:
     - time cargo web build -p substrate-panic-handler
     - time cargo web build -p substrate-peerset
     - time cargo web build -p substrate-primitives
+    - time cargo build -p substrate-rpc-servers --target wasm32-unknown-unknown  # TODO: we can't use cargo web until https://github.com/paritytech/jsonrpc/pull/436 is deployed
```
bkchr: Why not `cargo web`?

core/rpc-servers/Cargo.toml
```diff
@@ -5,10 +5,12 @@ authors = ["Parity Technologies <admin@parity.io>"]
...
 [dependencies]
-http = { package = "jsonrpc-http-server", version = "12.0.0" }
 pubsub = { package = "jsonrpc-pubsub", version = "12.0.0" }
-ws = { package = "jsonrpc-ws-server", version = "12.0.0" }
 log = "0.4"
 serde = "1.0"
 substrate-rpc = { path = "../rpc" }
 sr-primitives = { path = "../sr-primitives" }
+
+[target.'cfg(not(target_os = "unknown"))'.dependencies]
```
bkchr: Maybe introduce a `std` feature and tag these dependencies as `optional`?

core/service/src/lib.rs
bkchr: Why not introduce a new type that wraps the `Receiver` and the `RpcSession` and polling this type polls both?

tomaka: No, it's the `Service` that would have to be called at the same time as the receiver or session.

.gitlab-ci.yml
rphmeier: (there is a comment explaining why)

tomaka: The `warning: pubsub methods are currently not supported in the generated client.` lines that get printed confuse `cargo web`.

core/rpc-servers/Cargo.toml
tomaka: That would be the ideal way to do it, if it wasn't for https://github.com/rust-lang/cargo/issues/5730.
In substrate-service, I would have to depend on `std` for non-WASM, and not-std for WASM. But because of that Cargo bug, the `std` would always be compiled.

.gitlab-ci.yml
pepyakin: Can we move the comment on the line above?

core/rpc-servers/Cargo.toml
bkchr: Ahh this issue again :)

.gitlab-ci.yml
bkchr: Ohh, did not see the comment on my mobile phone :sweat_smile:

Issue #2719, opened 2019-05-29
I'm online messages


cmichi:
This feature is intended to handle the case when validators go offline without malicious intent (i.e. without the intention of attacking the network). We don't want to slash them if there is no malicious intent.

### Sending `I'm online` messages

* Each validator gossips an `I'm online` heartbeat every ~n blocks~ session.
* The heartbeat is a signed transaction, which was signed using the session key and includes the recent best block **number** of the local validators chain as well as the multiaddr (physical node address). It is sent as an Unsigned Transaction (no account nonce/signature/sender).
  - The multiaddr provides a secondary mechanism to look nodes up, since the DHT is flaky.

### Interpreting `I'm online` messages

* There will be a points system for the token mechanics, which we can also use for tracking if validators are online.
  - A validator is considered online if a threshold number of points is reached.
  - If a validator doesn't hit the threshold it is considered offline.
  - By using a points threshold we determine if a validator is online by taking anything into account which a validator contributed to the protocol (and not just `I'm online` messages).
  - The threshold is probably going to be `1` (i.e. any points at all signify the authority is online).
* Create a new `srml-im-online` module which has a `node_online` hook.
* Introduce a `validate_transaction` trait in the srml-im-online module which interprets and validates the unsigned transaction.
* `srml-im-online` contains a `map babe_session_key => Option<multiaddr>`. Clear this map every time `on_session_change()` is called.
* Validators may only produce a single im-online transaction for each session in each chain. Further im-omline messages are ignored once a valid message has been received in the current session for that validator. No slash is made.

### Future stuff with the points system
* Slash when a node is considered online according to the points system, but that node hasn't authored any blocks (or significantly less than expected) in a long time during which it should have authored blocks.
  - This "long time" is calculated by the likelihood that the node should have authored a block by now. If this threshold is below x then slash them. The further we get below this threshold the more we slash.
* If a validator is considered offline it is ignored as a validator, but not slashed. Kinda like a protection mechanism.

### Open questions

* What of the above info is persisted on chain? **Just the one map.**
* How often is the heartbeat sent? Every block? **Once per session.**
* Where do we determine that a slash happens? In `srml-im-online`? **Don't bother**

rphmeier:
> Each validator gossips an I'm online heartbeat every n block

It should be triggered somehow, since it's not supposed to be arbitrary `n`, but rather all synchronized to the start of a BABE epoch ( == session)

rphmeier:
[...]

PR #3015, opened 2019-07-04
slashing interface

niklasad1:
## What does this PR:

It tries to introduce a few traits (Misconduct, OnEndEra, OnSlashing) to introduce a generic interface to slash misbehavior(s) 

Roughly those have the following functionality:
- Misconduct: Trait that defines a type that handles severity level
- OnEndEra: Misconduct that should be slashed in the end of era instead of `rolling`
- OnSlashing: A trait to handles balance and concrete slashing, i.e, need to both call `T::Currency::free_balance()` and `T::Currency::slash()` based on `severity_level` estimated by the 
   Assumed to implemented on top of a module that handles `Currency a.ka BalanceOf<T>`
Misconduct

## Example use case

```rust
pub struct StakingSlasher<T>(T);

impl<T: Trait> OnSlashing<T::AccountId, Exposure, Severity> for StakingSlasher<T>
where
	Severity: Into<BalanceOf<T>>,
	Exposure: Into<BalanceOf<T>>,
{....}

/// Misconduct type
pub struct MyMisconduct<T>(T);

impl<T: Trait> Misconduct for MyMisconduct<T> {
   type Severity = u64;

    fn as_misconduct_level(&self, _severity: Fraction<Self::Severity>) -> u8 {...}

    fn on_misconduct(&mut self, _misbehaved: &[Self::AccountId]) -> Fraction<Self::Severity> {..}
}

let misconduct = MyMisconduct(SomeModule);
let misbehaved = vec![FakeSlashRecipient::default()];
// will probably require type parameters..
let misconduct_level = rolling_data(&misbehaved, &mut misconduct);
// use severity_level to determine whether to kick out the validator or not
```

## TODO (before merge)
1. Use checked_operations on Fraction
2. Add documentation (will not do before we have some sort of consensus)
3. bump impl version
4. How to deal with https://github.com/paritytech/substrate/issues/2719

A possible solution to #2719 could be to change 
`pub fn end_of_era(end_of_era: &E) -> u8` 
// don't slash `ignored` based on some proof
`pub fn end_of_era(end_of_era: &E, ignore: &[AccountId]) -> u8`

## Follow-up
Implement the concrete misconduct types
 
## References
https://research.web3.foundation/en/latest/polkadot/slashing/


srml/slashing/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
```
rphmeier: docs?

niklasad1:
Closing, has issues

PR #2962, opened 2019-06-26
Merge Protocol and ProtocolBehaviour

tomaka:
Merges `Protocol` and `ProtocolBehaviour`.
The `Protocol` struct now implements `NetworkBehaviour`.

The PR is ready, except for the fact that it totally breaks the network tests in an unrecoverable way.
Right now the network tests infrastructure assumes that we have a single channel of communication with the remote, and that `Protocol` exposes this in its API. This PR makes this an implementation detail of `Protocol`. Because of that, the tests, by design, can't be "just fixed".

I'm going to start working on rewriting the `TestNet` struct, after which we will be able to merge this PR on top. I'm opening it now, so that people are aware of upcoming changes.

Can be reviewed commit by commit.

PR #3016, opened 2019-07-04
Rewrite the network TestNet

tomaka:
The `TestNet` struct in `substrate-network` allows simulating a network. It is used in the tests of the network, of Aura, of Babe, and of GrandPa.

At the moment, it uses the internals of the network crate (more specifically: `Protocol`) and simulates a network on top of that. This PR changes that so that it uses the API surface of the network crate, therefore including discovery, pinging, and all that the networking actually does.

In order to do so, it is now possible to put the network in "memory only" mode. In this mode, the node addresses are in the form `/memory/N` where `N` is a randomly-generated u64. Nodes "connect" to each other using channels.

In the process, I've made all the tests asynchronous. No background thread is being spawned anymore.

Since we no longer decide which node is connected to which, "disconnecting" is now done by destroying nodes altogether. Other than this detail, I have tried to not change anything in the logic of the tests.

This will make it possible to continue the internal refactoring of the network crate, as we can now move touch things that would break the tests before. I've already opened #2962 that should be merged on top of this PR.


core/finality-grandpa/src/tests.rs
```diff
@@ -569,28 +466,30 @@ fn finalize_3_voters_no_observers() {
...
 fn finalize_3_voters_1_full_observer() {
+	let mut runtime = current_thread::Runtime::new().unwrap();
+
 	let peers = &[AuthorityKeyring::Alice, AuthorityKeyring::Bob, AuthorityKeyring::Charlie];
 	let voters = make_ids(peers);
 
 	let mut net = GrandpaTestNet::new(TestApi::new(voters), 4);
 	net.peer(0).push_blocks(20, false);
-	net.sync();
+	runtime.block_on(futures::future::poll_fn::<(), (), _>(|| Ok(net.poll_until_sync()))).unwrap();
```
arkpar: Would rather see . `poll_until_sync' accept runtime parameter and calling `block_on` internally, to avoid repeating all the `poll_fn` ugliness. A more generic `poll_until` that accepts a function would also be useful.

arkpar:
Looks good, except for a minor nitpick

PR #3017, opened 2019-07-04
Update trie dependencies.

cheme:
This PR update trie dependencies. Trie crate update include update of parity-util-mem to version 0.2.0 which do not include anymore a problematic dependency to clear_on_drop (clear_on_drop remains used by dalek crate and schnorkel but it is probably correctly managed for no_std).

I added patch version to all trie dependency so it becomes easier to sed update next time.


bkchr:
TY!

Issue #2999, opened 2019-07-02
Networking Primitives to get ready for Sentry Nodes


rphmeier:
Sentry nodes are special nodes set up by a validator operator that are intended to be the points-of-contact for the validator node, which is shielded from public connections.

Typically, sentry nodes will function as normal nodes. But for consensus networking in particular, they will have to behave as though they _are_ the validator node. That means that it needs to become aware of what knowledge the validator node has. 

This will typically be handled on the level of particular networking protocols (knowing how to proxy certain requests to the underlying validator).

---

### Basics

What needs to be made available to the networking protocol:
  - Whether the local node _has_ sentry nodes and what their `PeerId`s are.
  - Whether the local node _is_ a sentry node and what its `PeerId` is.

This information needs to be made available to the core-level networking, any `NetworkSpecialization` and any `consensus_gossip::GossipValidator`

Then we'll also need CLI parameters for configuring and setting up sentry nodes.

rphmeier:
cc @twittner : do you have some input on how best to integrate into the substrate networking to attach info onto `PeerId`s when connecting?

twittner:
> - Whether the local node *is* a sentry node and what its `PeerId` is.

What the `PeerId` of its validator node is?

> do you have some input on how best to integrate into the substrate networking to attach info onto `PeerId`s when connecting?

I am not sure what you mean. I think information about `PeerId`s would go into `network::config::NetworkConfiguration`.

rphmeier:
> What the PeerId of its validator node is?

yes, sorry.

tomaka:
Done as part of https://github.com/paritytech/substrate/pull/5520 https://github.com/paritytech/substrate/pull/5665

PR #3018, opened 2019-07-04
Introduce flag to enable sentry nodes to participate in grandpa gossip

mxinden:
Given the following situation: A validator 'A' is not supposed to be
connected to the public internet to protect it from e.g. a DoS attack.
Instead it connects to a sentry node 'sentry-A' which connects to the
public internet. Validator 'B' can reach validator 'A' via sentry node
'sentry-A' and vice versa.

A sentry node needs to participate in the grandpa gossip without
becoming a validator itself in order to forward these message to its
validator. This commit adds a new command line flag (`--grandpa-voter`)
forcing a node to participate in the grandpa voting process even though
no `--key` was specified. Due to the fact that it does not have a key,
it does not become a validator in the network.

In order to simulate the above situation this commit also adds a Docker
Compose file (`scripts/sentry-node/docker-compose.yml`) with further
documentation.

Parent Github issue: https://github.com/paritytech/substrate/issues/2999


mxinden:
[clabot:check]

mxinden:
> Maybe we should instead add a --sentry flag

@rphmeier was voting against calling the flag `--sentry` as users might want to run a grandpa voter outside of the context of a sentry node. While maybe impacting sentry user experience it does make the binary more versatile.

@andresilva What do you think?

rphmeier:
@mxinden we can always add a `--sentry` flag which implies a number of other options later. But it's nice to have the flag for running a keyless GRANDPA voter.

mxinden:
> @mxinden we can always add a --sentry flag which implies a number of other options later. But it's nice to have the flag for running a keyless GRANDPA voter.

Sounds good to me @rphmeier. 

---

@andresilva and @rphmeier  any further comments other than the flag renaming?

rphmeier:
@mxinden is it still in-progress or ready for review?

mxinden:
@rphmeier ready for review.

rphmeier:
(then you should put the pleasereview label on, otherwise people will skip it :) )

scripts/sentry-node/docker-compose.yml
```diff
@@ -0,0 +1,138 @@
...
+      # Don't configure a key, as sentry-a is not a validator.
+      # - "--key"
+      # - "//Charlie"
+      - "--port"
+      - "30333"
+      # sentry-a is not a validator.
+      # - "--validator"
+      - "--name"
+      - "CharliesNode"
+      - "--bootnodes"
```
andresilva: Maybe use `--reserved-nodes` here instead? Same for the other nodes configs.

mxinden:
Guessing from the [contributing guidelines](https://github.com/paritytech/substrate/blob/master/CONTRIBUTING.adoc) I am the one squashing and merging this pull request.

> Generally PRs should sit with this label for 48 hours in order to garner feedback.

I will wait for a little longer and then merge if their are no objections.

bkchr:
Just merge it :)

scripts/sentry-node/docker-compose.yml
mxinden: Would you mind adding some more details @andresilva?

If I understand correctly, validator addresses can change at runtime. Thus if the address of *validator-b* changes, *sentry-a* would need to be restarted. I would suggest only interconnecting *validator-a* and *sentry-a* as reserved nodes and *sentry-a* and *validator-b* as bootstrap nodes, given that *a* and *b* are controlled by different entities.

What do you think?

On a side note, I am currently looking into how a user can tell a validator node to keep *x* amount of connections open to other validator nodes. As the validator address set can change at runtime, these long-lived connections would also be adjusted along the way.

mxinden:
> (I'm not sure how useful it is to be committed but won't grumble on that)

@andresilva I am also fine with removing it. No strong opinion. Just thought it might be helpful  documenting the use case as code.

scripts/sentry-node/docker-compose.yml
andresilva: In this config `A` connects to `B` (through bootnodes). The sentry connects to `A` and `B`, and `B` also tries to connect with `A`.

My understanding is that in a real scenario the validator nodes that are using a sentry will only be connected to the sentry (or multiple sentries). So I think in this config `Validator-A` should set `Sentry-A` as a reserved node (so that it only connects to this node), and the rest of the nodes (`Sentry-A` and `Validator-B`) can add everyone as bootnode (even though `Validator-B` won't be able to connect to `Validator-A`).

PR #3020, opened 2019-07-04
Make `wasm-builder` less spammy

bkchr:


core/utils/wasm-builder/src/lib.rs
```diff
@@ -156,8 +156,12 @@ fn create_out_file(file_name: &str, content: String) {
 
 /// Get a cargo command that compiles with nightly
 fn get_nightly_cargo() -> Command {
-	if Command::new("rustup").args(&["run", "nightly", "cargo"])
-		.status().map(|s| s.success()).unwrap_or(false)
+	if Command::new("rustup")
+		.args(&["run", "nightly", "cargo"])
+		.stdout(Stdio::null())
```
tomusdrw: Would it be worth to actually put all the logs to some temporary file instead?

core/utils/wasm-builder/src/wasm_project.rs
```diff
@@ -266,7 +266,8 @@ fn build_project(project: &Path) {
 
 	match build_cmd.status().map(|s| s.success()) {
 		Ok(true) => {},
-		_ => panic!("Failed to compile WASM binary"),
+		// Use `process.exit(1)` to have a clean error output.
+		_ => process::exit(1),
```
tomusdrw: Shall we print something to stderr, like: "WASM Build failed check logs at wasm_build.logs"?

core/utils/wasm-builder/src/lib.rs
bkchr: This output is 100% useless, so we don't need it at all :)

core/utils/wasm-builder/src/wasm_project.rs
bkchr: The error messages are printed to stdout/stderr and I would not like to pipe it to a file.

Issue #1458, opened 2019-01-16
Off-chain oracle modules


gavofyork:
Currently it's impossible for code on-chain to access data off-chain except via hard-coded `InherentData`. This would make it entirely programmable.

There are two components:
- A runtime module that keeps track of the blobs of wasm code that block authors are expected to run prior to block formation which populate the `InherentData` that will thereafter be fed into the runtime.
- The according piece of client logic that checks for the existence of this runtime module, and if extent, grabs the wasm blob, constructs an appropriate (non-deterministic) wasm environment and calls a given function (e.g. `fn get_inherent_data() -> HashMap<Vec<u8>, Vec<u8>>` - the function should be infallible since to fail would render the chain dead). The environment should include general internet access, secure random number generation, timing, secure enclave access and any other useful hardware externalities. (A later version should introduce a fine-grained permissioning system for these.) The result of the function should be amalgamated into the `InherentData` for the runtime to utilise.

An example module should be constructed that places the timestamp into an `InherentData` key.

burrrata:
I'd like to better understand how this would work and contribute if I can. @tomusdrw @gavofyork what is the best place to ask questions and contribute code or ideas?

tomusdrw:
Hi @burrrata, reach me out on riot: `@todr:matrix.parity.io` please

tomusdrw:
To be done: implementation of methods exposed in #2615:

- [x] `submit_transaction` - submit a transaction to the pool
- [ ] `random_seed` - return a random seed (nondeterministic; local)
- [ ] `sign` - sign a piece of data with pre-configured key (most likely authorities key)
- [ ] `local_storage_read/write` - save and read stuff between offchain workers runs
- [ ] HTTP apis - fetch stuff from the web

supercoder123456:
@tomusdrw, Is there any example about off-chain worker? Do you have riot id also?

tomusdrw:
@supercoder123456, not much examples yet, I want to make the offchain workers a bit more functional before writing some. 
The only docs can be found here:
https://github.com/paritytech/substrate/blob/7d7981c7fd472cb94d1dc5c727d15464b1bf1f74/srml/example/src/lib.rs#L472

I'm available on riot as well @todr:matrix.parity.io

lingqingmeng:
I can create an off-chain oracle (non-deterministic) for Substrate using BLS281, who can I talk to for more information on the specific requirements on this?

gavofyork:
All good.

PR #3023, opened 2019-07-04
[offchain] Support for sign & verify for crypto keys

tomusdrw:
Related #1458 

This PR enables offchain workers to:
1. Create crypto keys (ed25519 or sr25519) and use them to sign data and verify signatures without leaking the seed.
2. Use the preconfigured authority key to sign stuff and verify signatures. Note that the `CryptoKind` of the authority key needs to be known in the offchain worker context (related: #3008) 

The crypto keys created with (1) are stored in persistent database, so that we can use them even in case of a fork.

Some other introduced changes:
1. A `crypto::Protected` type that zeroizes the memory on `Drop` used to store key passwords.
2. Require `Public` and `Pair` to implement `TypedKey` as well to avoid repeating the bounds.
3. The underlying DB api allows to `compare_and_set` against an unset value. We might consider exposting this to offchain workers as well.
4. Use `key_types` for `CryptoKind` encoding.


core/offchain/src/api.rs
```diff
@@ -52,38 +69,122 @@ fn unavailable_yet<R: Default>(name: &str) -> R {
...
+			},
+			CryptoKind::Sr25519 => {
+				sr25519::Pair::generate_with_phrase(Some(self.keys_password.as_ref())).1
+			},
+		};
+
+		let (id, id_encoded) = loop {
+			let encoded = self.db.get(KEYS_PREFIX, NEXT_ID);
+			let encoded_slice = encoded.as_ref().map(|x| x.as_slice());
+			let new_id = encoded_slice.and_then(|mut x| u16::decode(&mut x)).unwrap_or_default()
```
bkchr: I think we can end here in an infinite loop. If `db.get` returns nothing or `decode` is failing, `unwrap_or_default()` is returning `0` and no other check will end the loop.

core/primitives/src/sr25519.rs
```diff
@@ -379,6 +379,7 @@ fn derive_hard_junction(secret: &SecretKey, cc: &[u8; CHAIN_CODE_LENGTH]) -> Sec
 	secret.hard_derive_mini_secret_key(Some(ChainCode(cc.clone())), b"").0.expand()
 }
 
+/// The raw `MiniSecretKey` bytes.
```
bkchr: I don't understnad why it is called in the docs `MinitSecretKey` and as type name `Seed`?

core/service/src/lib.rs
```diff
@@ -919,7 +947,7 @@ fn build_system_rpc_handler<Components: components::Components>(
 /// # use consensus_common::{BlockOrigin, ImportBlock, well_known_cache_keys::Id as CacheKeyId};
 /// # use node_runtime::{GenesisConfig, RuntimeApi};
 /// # use std::sync::Arc;
-/// # use node_primitives::Block;
+/// # use node_primitives::{Block};
```
bkchr:
```suggestion
/// # use node_primitives::Block;
```

core/sr-io/without_std.rs
```diff
@@ -882,41 +901,63 @@ impl OffchainApi for () {
...
 	}
 
-	fn decrypt(key: Option<offchain::CryptoKeyId>, data: &[u8]) -> Result<Vec<u8>, ()> {
+	fn decrypt(
+		key: Option<offchain::CryptoKeyId>,
+		kind: offchain::CryptoKind,
+		data: &[u8],
+	) -> Result<Vec<u8>, ()> {
 		let key = key.map(|x| x.0 as u32).unwrap_or(0);
+		let kind = kind as isize as u32;
```
bkchr: Why not implement `Into<u32>` for `CryptoKind`?

core/client/src/in_mem.rs
```diff
@@ -794,19 +794,23 @@ impl backend::OffchainStorage for OffchainStorage {
...
-		let mut set = false;
-		self.storage.entry(key).and_modify(|val| {
-			if val.as_slice() == old_value {
-				*val = new_value.to_vec();
-				set = true;
-			}
-		});
-		set
+		match self.storage.entry(key) {
+			Entry::Vacant(entry) => if old_value.is_none() {
```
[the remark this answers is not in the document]
tomusdrw: Yeah, unfortunatelly it's not possible cause if you use a guard you can't move value out :/ and `entry.insert` consumes `entry`.

core/offchain/src/api.rs
tomusdrw: Hmm, so `compare_and_set` should fail only if the underlying value was modified in the meantime. So even if the value is invalid (and `decode` failed) or if the value is not set (`unwrap_or_default`) the `compare_and_set` should still succeed and we should break the loop.

I can add an upper bound here to prevent infinite loop, but don't really have any idea how we could avoid it.

core/primitives/src/sr25519.rs
tomusdrw: I can actually rollback that change, cause at the end I ended up keeping `Seed` private and just used the recovery phrase to store it in the database.

But my understanding is that `Seed` is a thing that is used to actually create the `Pair` type (via `Pair::from_seed`).
And for `sr25519` this happens to be the `MiniSecretKey`:
```rust
// This is generated using schnorrkel's Mini-Secret-Keys.                          
///                                                                                 
/// A MiniSecretKey is literally what Ed25519 calls a SecretKey, which is just 32 random bytes.   
```
hence the comment.

core/offchain/src/api.rs
bkchr: Ahh yeah.

core/primitives/src/sr25519.rs
gavofyork: Prefer `Seed` or the seed phrase to `MiniSecretKey` or `SecretKey` - the latter two are not always available for effective access or are specific to particular crypto systems.

gavofyork: "Seed" is just a crypto-generic concept for some datatype that when filled with random bytes can uniquely generate a "secret key". Some cryptos have restrictions on their secret keys, some crypto libraries don't let you access the secret key. So this is a useful abstraction.

PR #2981, opened 2019-06-30
Update Kademlia API. Fix DiscoveryBehaviour delays.

romanb:
This is a draft that adapts the Kademlia code to upcoming changes from https://github.com/libp2p/rust-libp2p/pull/1174 and https://github.com/libp2p/rust-libp2p/pull/1189. Also contains a fix for the artificial discovery delays, which will be backported to the current master in a separate PR, since this PR only becomes relevant with the next libp2p release at the earliest.

PR #3024, opened 2019-07-04
Some documentation for discovery.rs

tomaka:


core/network/src/discovery.rs
```diff
@@ -26,6 +26,38 @@ use std::{cmp, num::NonZeroU8, time::Duration};
 use tokio_io::{AsyncRead, AsyncWrite};
 use tokio_timer::{Delay, clock::Clock};
 
+//! Discovery mechanisms of Substrate.
+//!
+//! The `DiscoveryBehaviour` struct implements the `NetworkBehaviour` trait of libp2p and is
+//! responsible for discovering the nodes that are part of the network.
```
romanb:
```suggestion
//! responsible for discovering other nodes that are part of the network.
```
Just to not allow the confusion that _the nodes_ can be read as _all nodes_. Apart from very small networks, a node is very unlikely to ever discover all other nodes, especially not if "discovered" actually means "is in my DHT routing table", which is how discovered is defined by the implementation of `DiscoveryBehaviour`.

core/network/src/discovery.rs
```diff
@@ -26,6 +26,38 @@ use std::{cmp, num::NonZeroU8, time::Duration};
...
+//! order for nodes to propagate to us their view of the network. This is performed automatically
+//! by the `DiscoveryBehaviour`.
+//!
+//! Additionally, the `DiscoveryBehaviour` is also capable of storing and loading value in the
+//! network-wide DHT.
+//!
+//! ## Usage
+//!
+//! The `DiscoveryBehaviour` generates events of type `DiscoveryOut`, most notably
+//! `DiscoveryOut::Discovered` that is generated whenever we discover the identity of a node.
```
romanb: What do you have in mind here when you say "discover the identity of a node"? `Discovered` is actually currently generated whenever a peer is added to (or updated in) the Kademlia routing table.

core/network/src/discovery.rs
```diff
@@ -26,6 +26,38 @@ use std::{cmp, num::NonZeroU8, time::Duration};
...
+//! Only the identity of the node is returned. The node's addresses are stored within the
+//! `DiscoveryBehaviour` and can be queried through the `NetworkBehaviour` trait.
+//!
+//! **Important**: In order for the discovery mechanism to work properly, there needs to be an
+//! active mechanism that asks nodes for the addresses they are listening on. Whenever we learn
+//! of a node's address, you must call `add_self_reported_address`.
```
romanb: As a heads-up, in https://github.com/libp2p/rust-libp2p/pull/1189 I made sure that only "routable" peers are added to the routing table and the Kademlia behaviour emits a dedicated event whenever a peer connects for whom no listen address is known, namely `KademliaEvent::UnroutablePeer` (instead of adding a peer with no addresses to the routing table). This makes the requirement more explicit, i.e. documents it in the public API. See also https://github.com/paritytech/substrate/pull/2981/files where this event is then propagated by the `DiscoveryBehaviour` including some related updated documentation.

PR #2854, opened 2019-06-12
Transaction Fee Multiplier

4meta5:
Implementation of the [slow fee mechanism](http://research.web3.foundation/en/latest/polkadot/Token%20Economics/#2-slow-adjusting-mechanism) -- required some changes after merging https://github.com/paritytech/substrate/pull/2799 
> addresses https://github.com/paritytech/substrate/issues/2430

TODO:
- [x] clean up and finish `make_payment` in `srml/balances`
- [x] test that the block weight limit is enforced (like [here](https://github.com/paritytech/substrate/blob/master/srml/executive/src/lib.rs#L553))
- [x] test correct fee calculation for a few transactions

cc @Kianenigma

PR #3026, opened 2019-07-04
Fixed64 type for runtime.

Kianenigma:
Mainly because I need it here #2854. 

self-explanatory: 

```rust
pub struct Fixed64 {
	/// Sign of the number
	sign: Sign,
	/// The natural part.
	n: u64,
	/// The decimal part.
	decimal: Perbill,
}
```
(I am nor confident about the naming of the `n` and `decimal`. maybe `fractional` or `f` instead of it?)

For now, only supports `Add` and `Sub`. I added some standard methods but this might change based on how I use it after being merged.

Also adds proper implementations of 
- `Add`, `Sub`, `Mul`
- `Saturating`
- `CheckedAdd`, `CheckedSub`

to `Perbill` and `Permill`.


core/sr-primitives/src/lib.rs
```diff
@@ -237,6 +237,88 @@ where
 	}
 }
 
+/// These implementation will never panic and will always return a _per-thing_ which is in the range
+/// `[0, 1] (inclusive)`.
+impl Saturating for Permill {
+	fn saturating_add(self, rhs: Self) -> Self {
+		// defensive inner saturation:
+		// both u32 values are below a million. sum won't overflow in u32.
+		Self::from_parts((self.0 + rhs.0).min(1_000_000))
```
thiolliere: the min is redundant with inner min in from_parts but I'm fine to keep it as well

core/sr-primitives/src/lib.rs
```diff
@@ -344,6 +426,87 @@ where
 	}
 }
 
+/// These implementation will never panic and will always return a _per-thing_ which is in the range
+/// `[0, 1] (inclusive)`.
+impl Saturating for Perbill {
+	fn saturating_add(self, rhs: Self) -> Self {
+		// defensive inner saturation:
+		// both u32 values are below a billion. sum won't overflow for u32.
+		Self::from_parts((self.0 + rhs.0).min(1_000_000_000))
```
thiolliere: same min is redundant

core/sr-primitives/src/lib.rs
```diff
@@ -374,6 +537,127 @@ impl From<codec::Compact<Perbill>> for Perbill {
...
+#[derive(PartialEq, Eq, PartialOrd, Ord)]
+pub enum Sign {
+	/// Negative
+	N,
+	/// Positive
+	P,
+}
+
+/// A fixed point number in the range of `[-(u64::max+1), u64::max+1]`. Composed of one `u64`
+///  _natural_ part and a `Perbill` _decimal_ part.
```
thiolliere: yes I'm mixed with decimal because it kind of means x/10 no ? but at the same time the mention of perbill makes it clear

core/sr-primitives/src/lib.rs
```diff
@@ -374,6 +537,127 @@ impl From<codec::Compact<Perbill>> for Perbill {
...
+///
+/// Note that a positive instance is always greater than a negative instance, even if both have `0`
+///  as their natural and decimal part.
```
thiolliere: hmm should we prefer it to be derived by hand so positive 0 == negative 0 ?

core/sr-primitives/src/lib.rs
```diff
@@ -374,6 +537,127 @@ impl From<codec::Compact<Perbill>> for Perbill {
...
+///
+/// Note that internal Perbill supports an inclusive range of of [0, 1], hence the number actually
```
thiolliere:
```suggestion
/// Note that internal Perbill supports an inclusive range of [0, 1], hence the number actually
```

core/sr-primitives/src/lib.rs
```diff
@@ -374,6 +537,127 @@ impl From<codec::Compact<Perbill>> for Perbill {
...
+impl ops::Neg for Fixed64 {
+	type Output = Self;
+
+    fn neg(self) -> Self::Output {
+		let mut r = self;
+        match r.sign {
+            Sign::P => r.sign = Sign::N,
+            Sign::N => r.sign = Sign::P,
+        };
+		r
```
thiolliere: identations

core/sr-primitives/src/lib.rs
```diff
@@ -374,6 +537,127 @@ impl From<codec::Compact<Perbill>> for Perbill {
...
+	fn add(self, rhs: Self) -> Self {
+		if self.sign == rhs.sign {
+			// sign will not change. Simply add them ignoring the sign
+			let new_decimal = match self.decimal.checked_add(&rhs.decimal) {
+				Some(d) => d,
+				None => Perbill::from_parts((self.decimal.0 + rhs.decimal.0) % 1_000_000_000)
+			};
+			Self {
+				sign: self.sign,
+				n: self.n.saturating_add(rhs.n), /// TODO
```
thiolliere: yes todo: add fractional part

core/sr-primitives/src/lib.rs
```diff
@@ -374,6 +537,127 @@ impl From<codec::Compact<Perbill>> for Perbill {
...
+				decimal: new_decimal
```
thiolliere: I think this could be written more quickly with

n = sef.n saturating add rhs.n + (self.decimal.0 + rhs.decimal.0).rem(billion)
deciman = (self.decimal.0 +otherdecimal) % billion

core/sr-primitives/src/lib.rs
```diff
@@ -374,6 +537,127 @@ impl From<codec::Compact<Perbill>> for Perbill {
...
+					None => {
+						r += 1;
+						Perbill::from_parts(self.decimal.0 + 1_000_000_000 - rhs.decimal.0)
+					}
+				};
+				Self {
+					sign: self.sign,
+					n: self.n - rhs.n - r,
+					decimal: new_decimal
+				}
```
thiolliere: this can be written:
```
n = self.n - rhs.n - ((self.decimal + billion - rhs.decimal) >= billion) as u64
decimal = (self.decimal + billion - rhs.decimal )% billion
```
because
> The bool represents a value, which could only be either true or false. If you cast a bool into an integer, true will be 1 and false will be 0.

core/sr-primitives/src/lib.rs
bkchr: Yeah remove it.

core/sr-primitives/src/lib.rs
```diff
@@ -237,6 +237,88 @@ where
...
+		// defensive inner saturation:
+		// both u32 values are below a million. sum won't overflow in u32.
+		Self::from_parts((self.0 + rhs.0).min(1_000_000))
+	}
+
+	fn saturating_sub(self, rhs: Self) -> Self {
+		Self::from_parts(self.0.checked_sub(rhs.0).unwrap_or(0))
+	}
+
+	/// A the multiplication of two numbers in [0, 1] is always within that range
```
bkchr:
```suggestion
	/// A multiplication of two numbers in [0, 1] is always within that range
```

core/sr-primitives/src/lib.rs
```diff
@@ -237,6 +237,88 @@ where
...
+}
+
+/// Standard subtraction operation.
+///
+/// Panics if `self` is smaller than `rhs`.
+impl ops::Sub for Permill {
+	type Output = Self;
+
+	fn sub(self, rhs: Self) -> Self {
+		Self::from_parts(&self.0 - rhs.0)
```
bkchr: This does not panic, this creates an underflow for `self < rhs`.

core/sr-primitives/src/lib.rs
```diff
@@ -237,6 +237,88 @@ where
...
+///
+/// 10^18 will fit in u64, hence it is safe to use it as the intermediary type.
+impl ops::Mul for Permill {
+	type Output = Self;
+
+	fn mul(self, rhs: Self) -> Self {
+		let p = self.0 as u64;
+		let q = rhs.0 as u64;
+
+		// This will never overflow (given that p and q are valid perbill)
```
bkchr:
```suggestion
		// This will never overflow (given that p and q are valid permill)
```

core/sr-primitives/src/lib.rs
```diff
@@ -237,6 +237,88 @@ where
...
+		let n = p * q;
+		// billion always fits in u32.
+		Self::from_parts((n / 1_000_000).min(1_000_000) as u32)
+	}
+}
+
+impl CheckedAdd for Permill {
+	fn checked_add(&self, rhs: &Self) -> Option<Self> {
+		if self.0 + rhs.0 <= 1_000_000 {
```
bkchr:
```suggestion
		let res = self.0 + rhs.0;
		if res <= 1_000_000 {
```

core/sr-primitives/src/lib.rs
```diff
@@ -237,6 +237,88 @@ where
...
+			return Some(Self::from_parts(self.0 + rhs.0));
```
bkchr:
```suggestion
			Some(Self::from_parts(res))
```

core/sr-primitives/src/lib.rs
```diff
@@ -237,6 +237,88 @@ where
...
+		}
```
bkchr:
```suggestion
		} else {
```

core/sr-primitives/src/lib.rs
```diff
@@ -237,6 +237,88 @@ where
...
+		None
```
bkchr:
```suggestion
			None
		}
```

core/sr-primitives/src/lib.rs
```diff
@@ -237,6 +237,88 @@ where
...
+	}
+}
+
+impl CheckedSub for Permill {
+	fn checked_sub(&self, rhs: &Self) -> Option<Self> {
+		if let Some(v) = self.0.checked_sub(rhs.0) {
```
bkchr: All this can be written as:
`Self.0.checked_sub(rhs.0).map(Self::from_parts)`

core/sr-primitives/src/lib.rs
```diff
@@ -344,6 +426,87 @@ where
...
+}
+
+/// Standard subtraction operation.
+///
+/// Panics if `self` is smaller than `rhs`.
+impl ops::Sub for Perbill {
+	type Output = Self;
+
+	fn sub(self, rhs: Self) -> Self {
+		Self::from_parts(&self.0 - rhs.0)
```
bkchr: Also does not panic.

core/sr-primitives/src/lib.rs
```diff
@@ -344,6 +426,87 @@ where
...
+		let n = p * q;
+		// billion always fits in u32.
+		Self::from_parts((n / 1_000_000_000).min(1_000_000_000) as u32)
+	}
+}
+
+impl CheckedAdd for Perbill {
+	fn checked_add(&self, rhs: &Self) -> Option<Self> {
+		if self.0 + rhs.0 <= 1_000_000_000 {
+			return Some(Self::from_parts(self.0 + rhs.0));
```
bkchr: Same as above.

core/sr-primitives/src/lib.rs
```diff
@@ -344,6 +426,87 @@ where
...
+		}
+		None
+	}
+}
+
+impl CheckedSub for Perbill {
+	fn checked_sub(&self, rhs: &Self) -> Option<Self> {
+		if let Some(v) = self.0.checked_sub(rhs.0) {
+			return Some(Self::from_parts(v));
```
bkchr: Same.

core/sr-primitives/src/lib.rs
bkchr: I think he has taken it from floats, there are `-0` != `+0` AFAIK

core/sr-primitives/src/lib.rs
```diff
@@ -374,6 +537,127 @@ impl From<codec::Compact<Perbill>> for Perbill {
...
+}
+
+/// A fixed point number in the range of `[-(u64::max+1), u64::max+1]`. Composed of one `u64`
+///  _natural_ part and a `Perbill` _decimal_ part.
+///
+/// Note that a positive instance is always greater than a negative instance, even if both have `0`
+///  as their natural and decimal part.
+///
+/// Note that internal Perbill supports an inclusive range of of [0, 1], hence the number actually
+/// be `n` + `perbill::one()`, which is technically `n + 1`.
```
bkchr:
```suggestion
/// be `n` + `Perbill::one()`, which is technically `n + 1`.
```

core/sr-primitives/src/lib.rs
```diff
@@ -374,6 +537,127 @@ impl From<codec::Compact<Perbill>> for Perbill {
...
+	fn new(sign: Sign, n: u64, decimal: Perbill) -> Self {
+		Self { sign, decimal, n }
+	}
+
+	fn abs(&self) -> Self {
+		Self { sign: Sign::P, n: self.n, decimal: self.decimal }
+	}
+
+	fn from_perbill(decimal: Perbill) -> Self {
+		Self { decimal, ..Default::default() }
```
bkchr: You don't implement `Default` for `Sign`.

core/sr-primitives/src/lib.rs
```diff
@@ -374,6 +537,127 @@ impl From<codec::Compact<Perbill>> for Perbill {
...
+	}
+}
+
+impl ops::Neg for Fixed64 {
+	type Output = Self;
+
+    fn neg(self) -> Self::Output {
```
bkchr: Spaces :scream:

core/sr-primitives/src/lib.rs
thiolliere: hmm not sure [play.rustlang](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=fn%20main()%7B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20-0.0%20%3C%200.0)%3B%0A%7D)

but I can agree on this actually we can think Positive0 is between 0 and 0+one_per_billion and negative 0 is between 0 and 0-one_per_billion

core/sr-primitives/src/lib.rs
Kianenigma: agree the name might be misleading. at least with `n` I tried to deliver the message that it is a natural number which should clarify it.

core/sr-primitives/src/lib.rs
bkchr: I think @thiolliere is right and we should provide a manual implementation for `PartialOrd` and `Eq` to follow the standard behavior of a float.

core/sr-primitives/src/lib.rs
Kianenigma: I was also referring to the type of the `add` here; I was thinking to change this to normal, panic-ing add since this is the standard `Add` operation.

core/sr-primitives/src/lib.rs
thiolliere: maybe _fractional_ is a good word 🤔

https://en.wikipedia.org/wiki/Fraction_(mathematics)#Mixed_numbers

gavofyork:
This wasn't what I was thinking of at all. Let's KISS and just have:

```
struct Fixed64(i64);

const DIV: i64 = 1_000_000_000;

impl Fixed64 {
  fn from_percent(percent: i32) -> Self { Self(percent * DIV / 100) }
  fn from_natural(int: i64) { Self(int * DIV) }
  fn from_rational(n: i64, d: u64) { Self(n * DIV / d) }
```

It will need to saturate properly &c. but you get the idea.

Kianenigma:
oh ok, this is much simpler. I will add this directly to my other PR.

PR #3028, opened 2019-07-05
BABE Epochs

### LINKED ISSUE #2435: BABE randomness
This comes after #2084.

### LINKED ISSUE #2436: Use the actual genesis block hash in BABE
instead of an empty slice.

DemiMarie-parity:
This adds support for epochs to BABE.  Most of the plumbing is done, but the code is still a WIP.

This is a draft, but I would appreciate review on what I have done so far, so as to know that I am on the right track.

Closes #2435
Closes #2436


core/consensus/babe/src/lib.rs
```diff
@@ -1051,6 +1056,7 @@ mod tests {
 				inherent_data_providers,
 				force_authoring: false,
 			    time_source: Default::default(),
```
bkchr:
```suggestion
				time_source: Default::default(),
```

core/consensus/slots/src/slots.rs
```diff
@@ -167,6 +173,7 @@ impl<SC: SlotCompatible> Stream for Slots<SC> {
 					timestamp,
 					ends_at,
 					inherent_data,
+					epoch: slot_num % u64::from(self.slots_per_epoch),
```
bkchr:
```suggestion
					epoch: slot_num % self.slots_per_epoch.get(),
```

srml/babe/src/lib.rs
```diff
@@ -151,11 +155,15 @@ decl_module! {
 				.iter()
 				.filter_map(|s| s.as_pre_runtime())
 				.filter_map(|(id, mut data)| if id == BABE_ENGINE_ID {
-					<[u8; VRF_OUTPUT_LENGTH]>::decode(&mut data)
+					<(u64, [u8; VRF_OUTPUT_LENGTH])>::decode(&mut data)
```
bkchr: Can we not put this somewhere in a type? That is used in the node and the runtime? This feels like something that can break very fast in a refactoring.

srml/babe/src/lib.rs
```diff
@@ -216,17 +233,26 @@ impl<T: Trait> OnTimestampSet<T::Moment> for Module<T> {
 	fn on_timestamp_set(_moment: T::Moment) { }
 }
 
+impl<T: Trait> session::ShouldEndSession<u64> for Module<T> {
+	fn should_end_session(_now: u64) -> bool {
+		LastSlotInEpoch::get()
```
bkchr:
```suggestion
		LastSlotInEpoch::take()
```

I think this should be safe, as when we return `true` here, we probably see a `on_new_session` in the next block?

srml/babe/src/lib.rs
```diff
@@ -216,17 +233,26 @@ impl<T: Trait> OnTimestampSet<T::Moment> for Module<T> {
...
+
 impl<T: Trait> session::OneSessionHandler<T::AccountId> for Module<T> {
-	type Key = AuthorityId;
+	type Key = (AuthorityId, u64);
 	fn on_new_session<'a, I: 'a>(changed: bool, validators: I)
-		where I: Iterator<Item=(&'a T::AccountId, AuthorityId)>
+		where I: Iterator<Item=(&'a T::AccountId, (AuthorityId, u64))>
 	{
+		// Clear the “need new session” flat, so we do not get called over and over
+		LastSlotInEpoch::put(false);
```
bkchr: If you accept the above change, this can be removed.

core/consensus/babe/src/lib.rs
```diff
@@ -1051,6 +1056,7 @@ mod tests {
...
+				slots_per_epoch: NonZeroU64::new(20).expect("20 is not 0"),
```
bkchr:
```suggestion
				slots_per_epoch: NonZeroU64::new(20).expect("20 is not 0; qed"),
```

core/consensus/slots/src/slots.rs
```diff
@@ -96,12 +99,13 @@ impl SlotInfo {
...
 
 /// A stream that returns every time there is a new slot.
-pub struct Slots<SC> {
+pub(crate) struct Slots<SC> {
 	last_slot: u64,
 	slot_duration: u64,
 	inner_delay: Option<Delay>,
 	inherent_data_providers: InherentDataProviders,
 	timestamp_extractor: SC,
+	slots_per_epoch: NonZeroU64,
```
rphmeier: this isn't really helpful for Aura. can it be done in BABE only?

rphmeier: This is requiring that any slot-based consensus have a notion of epoch demarcated by slot

srml/babe/src/lib.rs
```diff
@@ -151,11 +155,15 @@ decl_module! {
...
 				.filter_map(|s| s.as_pre_runtime())
 				.filter_map(|(id, mut data)| if id == BABE_ENGINE_ID {
-					<[u8; VRF_OUTPUT_LENGTH]>::decode(&mut data)
+					<(u64, [u8; VRF_OUTPUT_LENGTH])>::decode(&mut data)
 				} else {
 					None
 				}) {
-				Self::deposit_vrf_output(&i);
+				if i.0 != EpochIndex::get() {
+					// New epoch reported by validator
```
rphmeier: can't it be derived from slot number? the `epoch` parameter seems redundant

srml/babe/src/lib.rs
rphmeier: @bkchr that's right

srml/babe/src/lib.rs
DemiMarie-parity: That would require getting the chain configuration into the runtime, though.  Not sure how to do that.

core/consensus/slots/src/slots.rs
DemiMarie-parity: > This is requiring that any slot-based consensus have a notion of epoch demarcated by slot

I agree.  That said, this is enough to get BABE working.  Is it okay if I get BABE complete first, since that is what the product needs?

rphmeier: Sure, let's file something for later.

srml/babe/src/lib.rs
rphmeier: `GenesisConfig`, no?
The node should be informed by the chain which is informed by the genesis config. At the moment, the node will draw information about configuration only at genesis. Later on we'll have it draw information at other points.

core/consensus/babe/primitives/src/lib.rs
```diff
@@ -49,15 +49,27 @@ pub type SlotNumber = u64;
 /// The weight of an authority.
 pub type Weight = u64;
 
+/// BABE epoch information
+#[derive(Decode, Encode, Default, PartialEq, Eq, Clone)]
+#[cfg_attr(any(feature = "std", test), derive(Debug))]
+pub struct Epoch {
+	/// The slot number this block will start at
```
andresilva:
```suggestion
	/// The slot number this epoch will start at
```

core/consensus/babe/primitives/src/lib.rs
```diff
@@ -116,7 +134,7 @@ decl_runtime_apis! {
 		/// Dynamic configuration may be supported in the future.
 		fn startup_data() -> BabeConfiguration;
 
-		/// Get the current authorites for Babe.
-		fn authorities() -> Vec<AuthorityId>;
+		/// Get the current epoch data for Babe.
+		fn authorities() -> Epoch;
```
andresilva: Can't we rename the method to `epoch()`?

core/consensus/babe/src/lib.rs
```diff
@@ -357,13 +372,13 @@ impl<Hash, H, B, C, E, I, Error, SO> SlotWorker<B> for BabeWorker<C, E, I, SO> w
...
 
 			let (header, body) = b.deconstruct();
-			let pre_digest: Result<BabePreDigest, String> = find_pre_digest::<B>(&header);
-			if let Err(e) = pre_digest {
-				error!(target: "babe", "FATAL ERROR: Invalid pre-digest: {}!", e);
-				return
-			} else {
-				trace!(target: "babe", "Got correct number of seals.  Good!")
-			};
+			// let pre_digest = find_pre_digest::<B>(&header);
```
andresilva: Needs updating.

core/consensus/babe/src/lib.rs
```diff
@@ -464,12 +486,13 @@ fn check_header<B: Block + Sized, C: AuxStore>(
 		babe_err!("Header {:?} has a bad seal", hash)
 	})?;
 
-	let pre_digest = find_pre_digest::<B>(&header)?;
-	let BabePreDigest { slot_num, index, ref proof, ref vrf_output } = pre_digest;
+	let (_epoch, pre_digest) = find_pre_digest::<B>(&header)?;
```
andresilva: Shouldn't we be checking wether this is the expected epoch?

core/consensus/slots/src/slots.rs
```diff
@@ -81,6 +82,8 @@ pub struct SlotInfo {
 	pub inherent_data: InherentData,
 	/// Slot duration.
 	pub duration: u64,
+	/// Epoch.
+	pub epoch: u64,
```
andresilva: Do we need to track this here? Why not have just BABE track this based on slot number?

run.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash --
```
andresilva: Should this go in `scripts/`?

srml/babe/src/lib.rs
```diff
@@ -133,11 +139,24 @@ decl_storage! {
...
-		NextEpochRandomness: [u8; VRF_OUTPUT_LENGTH];
+		/// The data for the next epoch
+		NextEpoch get(next_epoch): Epoch;
 
 		/// The current epoch
 		EpochIndex get(epoch_index): u64;
+
+		/// The number of slots per epoch
+		SlotsPerEpoch get(slots_per_epoch) build(|config: &GenesisConfig| {
+			config.slots_per_epoch
```
andresilva: I think you can just do `SlotsPerEpoch get(slots_per_epoch) config(): u64;` instead.

srml/babe/src/lib.rs
```diff
@@ -151,16 +170,38 @@ decl_module! {
...
+					<(u64, [u8; VRF_OUTPUT_LENGTH])>::decode(&mut data)
 				} else {
 					None
 				}) {
-				Self::deposit_vrf_output(&i);
+				let slot_index = i.0
+					.checked_add(1)
+					.expect("slot numbers will never reach 2^64 in our lifetimes; qed");
+				let epoch_change_slot = slot_index
+					.checked_add(Self::slots_per_epoch())
```
andresilva: Isn't slot_index the current slot? How does this work? Shouldn't this be `slot_index % slots_per_epoch == 0` instead?

srml/babe/src/lib.rs
```diff
@@ -216,42 +262,53 @@ impl<T: Trait> OnTimestampSet<T::Moment> for Module<T> {
...
-impl<T: Trait> session::OneSessionHandler<T::AccountId> for Module<T> {
-	type Key = AuthorityId;
+impl<T: Trait> session::ShouldEndSession<u64> for Module<T> {
+	fn should_end_session(_now: u64) -> bool {
+		LastSlotInEpoch::take()
+	}
+}
 
-	fn on_new_session<'a, I: 'a>(changed: bool, validators: I)
+impl<T: Trait + staking::Trait> session::OneSessionHandler<T::AccountId> for Module<T> {
```
andresilva: Not sure if this coupling with staking module is the best approach, but I'm not very familiar with it.

core/consensus/babe/primitives/src/lib.rs
DemiMarie-parity: We can, and should.  Will fix.

core/consensus/babe/src/lib.rs
DemiMarie-parity: I will just delete this code.  It was strictly for debugging.

core/consensus/babe/src/lib.rs
DemiMarie-parity: We should.  Good catch!

core/consensus/slots/src/slots.rs
DemiMarie-parity: I do not think we do.  If I recall correctly, this was done for convenience, but I can fix this.

run.sh
DemiMarie-parity: Yes, it should.  Is this script one that is worth keeping?

srml/babe/src/lib.rs
DemiMarie-parity: I was not aware of that.  Thanks!

srml/babe/src/lib.rs
```diff
@@ -133,11 +139,24 @@ decl_storage! {
 		/// The randomness under construction
 		UnderConstruction: [u8; VRF_OUTPUT_LENGTH];
 
-		/// The randomness for the next epoch
-		NextEpochRandomness: [u8; VRF_OUTPUT_LENGTH];
+		/// The data for the next epoch
+		NextEpoch get(next_epoch): Epoch;
```
DemiMarie-parity: This needs to be initialized properly at genesis.

srml/babe/src/lib.rs
DemiMarie-parity: I will switch to `BabePreDigest` here, since I think that the performance penalty is worth it.

srml/babe/src/lib.rs
DemiMarie-parity: `slot_index` is indeed the current slot, and so this makes `epoch_change_slot` the slot that the next epoch change will happen at.  That said, your version is cleaner.

srml/babe/src/lib.rs
DemiMarie-parity: It was the only way I could figure out for getting weights into BABE.

core/consensus/babe/src/lib.rs
```diff
@@ -418,19 +425,24 @@ macro_rules! babe_err {
...
-	let mut pre_digest: Option<_> = None;
+	let (mut pre_digest, mut epoch): (Option<_>, Option<_>) = (None, None);
 	for log in header.digest().logs() {
 		trace!(target: "babe", "Checking log {:?}", log);
 		match (log.as_babe_pre_digest(), pre_digest.is_some()) {
 			(Some(_), true) => Err(babe_err!("Multiple BABE pre-runtime headers, rejecting!"))?,
 			(None, _) => trace!(target: "babe", "Ignoring digest not meant for us"),
 			(s, false) => pre_digest = s,
 		}
+		match (log.as_babe_epoch(), epoch.is_some()) {
```
bkchr: As we are replicating the same code as above, this could be extracted into its own function/closure :woman_shrugging:

core/consensus/babe/src/lib.rs
```diff
@@ -463,12 +477,13 @@ fn check_header<B: Block + Sized, C: AuxStore>(
...
-	let pre_digest = find_pre_digest::<B>(&header)?;
-	let BabePreDigest { slot_num, index, ref proof, ref vrf_output } = pre_digest;
+	let (_epoch, pre_digest) = find_pre_digest::<B>(&header)?;
+
+	let BabePreDigest { slot_num, index, ref proof, ref vrf_output, epoch: _ } = pre_digest;
 
 	if slot_num > slot_now {
 		header.digest_mut().push(seal);
-		Ok(CheckedHeader::Deferred(header, slot_num))
+		Ok(CheckedHeader::Deferred(([0; 32], header), slot_num))
```
bkchr: Do we maybe want to change `CheckedHeader` so accept a different type for `Deferred` and `Checked`?
Or at least return `Option<[u8;32}>` here?

core/consensus/common/src/lib.rs
```diff
@@ -123,4 +123,7 @@ pub mod well_known_cache_keys {
 
 	/// A list of authorities.
 	pub const AUTHORITIES: Id = *b"auth";
+
+	/// On-chain randomness.
+	pub const RANDOMNESS: Id = *b"rand";
```
bkchr: This is not used?

scripts/run.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh --
```
bkchr: Should not be added.

srml/babe/src/lib.rs
```diff
@@ -116,7 +116,13 @@ decl_storage! {
 		LastTimestamp get(last): T::Moment;
 
 		/// The current authorities set.
-		Authorities get(authorities): Vec<AuthorityId>;
+		Authorities get(authorities) build(|config: &GenesisConfig| {
```
bkchr: Here you can do the same as @andresilva has written below.

srml/babe/src/lib.rs
bkchr: Which performance penality? Does `BabePreDigest` contains more data?

srml/babe/src/lib.rs
```diff
@@ -151,16 +170,38 @@ decl_module! {
 				.iter()
 				.filter_map(|s| s.as_pre_runtime())
 				.filter_map(|(id, mut data)| if id == BABE_ENGINE_ID {
-					<[u8; VRF_OUTPUT_LENGTH]>::decode(&mut data)
+					<(u64, [u8; VRF_OUTPUT_LENGTH])>::decode(&mut data)
 				} else {
 					None
 				}) {
-				Self::deposit_vrf_output(&i);
+				let slot_index = i.0
```
bkchr: Is that the index and not the `slot_num`?

srml/babe/src/lib.rs
bkchr: I think @andresilva is right and your current code just does not work, how should both values be ever equal, when the second is the first + some value?

srml/babe/src/lib.rs
bkchr: Is this value variable? Aka being changed per epoch?

srml/babe/src/lib.rs
```diff
@@ -151,16 +170,38 @@ decl_module! {
...
+				return Self::deposit_vrf_output(&i.1)
 			}
 		}
 	}
 }
 
+pub fn epoch<T: Trait>() -> Epoch {
+	let authorities = Authorities::get();
+	if false && authorities.is_empty() {
+		panic!("No authorities ― network is deadlocked")
```
bkchr: Never panic in the runtime. Please remove.

srml/babe/src/lib.rs
```diff
@@ -151,16 +170,38 @@ decl_module! {
...
+	}
+	Epoch {
+		randomness: EpochRandomness::get(),
+		authorities: authorities,
+		slot_number: SlotsPerEpoch::get().wrapping_add(EpochIndex::get()),
```
bkchr: Is this correct? For one epoch this will always return the same slot number?

srml/babe/src/lib.rs
```diff
@@ -216,42 +262,53 @@ impl<T: Trait> OnTimestampSet<T::Moment> for Module<T> {
...
 		let mut s = [0; 72];
 		s[..32].copy_from_slice(&last_epoch_randomness);
 		s[32..40].copy_from_slice(&epoch_index.to_le_bytes());
 		s[40..].copy_from_slice(&rho);
-		NextEpochRandomness::put(runtime_io::blake2_256(&s))
+		let randomness = runtime_io::blake2_256(&s);
+		Self::change_authorities(Epoch {
+			randomness,
+			authorities: next_authorities,
+			slot_number: slot_number
```
bkchr: And here you calculate it in a different way?

srml/babe/src/lib.rs
bkchr: You mean `CurrencyToVote`? You could just add this to your trait as well.

srml/babe/src/lib.rs
DemiMarie-parity: Fixed (in a different way).

srml/babe/src/lib.rs
DemiMarie-parity: Fixed by deleting the `slot_number` field.

srml/babe/src/lib.rs
DemiMarie-parity: Fixed by deleting the `slot_number` field.

srml/babe/src/lib.rs
DemiMarie-parity: Thank you.

core/consensus/babe/src/lib.rs
DemiMarie-parity: This will be caught by the runtime (which executes on each full node).

srml/babe/src/lib.rs
DemiMarie-parity: @bkchr Yes, it does.

srml/babe/src/lib.rs
rphmeier: It doesn't make sense to have this variable without the `EpochChange` digest containing the `next_slot_number` (removed in this PR).

DemiMarie-parity: @bkchr @rphmeier it is indeed not variable, as keeping it constant simplifies the code significantly.  I can make it variable later, but I would like to get this code merged first.

core/consensus/babe/primitives/src/lib.rs
```diff
@@ -15,15 +15,22 @@
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
 //! Primitives for BABE.
-#![deny(warnings, unsafe_code, missing_docs)]
+//#![deny(warnings)]
+//#![forbid(unsafe_code, missing_docs, unused_variables, unused_imports)]
```
rphmeier: remove commented code

core/consensus/babe/src/lib.rs
```diff
@@ -281,44 +292,40 @@ impl<Hash, H, B, C, E, I, Error, SO> SlotWorker<B> for BabeWorker<C, E, I, SO> w
...
-		// https://github.com/paritytech/substrate/issues/2436
-		let proposal_work = if let Some(((inout, proof, _batchable_proof), index)) = claim_slot(
-			&[0u8; 0],
+		let proposal_work = if let Some(((inout, vrf_proof, _batchable_proof), authority_index)) = claim_slot(
+			&randomness,
 			slot_info.number,
-			&[0u8; 0],
-			0,
-			&authorities,
+			slot_info.number / self.slots_per_epoch,
```
rphmeier: I am fairly certain I have warned against this bug already in a couple places, but in case it wasn't clear: this calculation is wrong in the case where there have been more than `slots_per_epoch` slots since the parent block.

In this part of the code we should always draw the epoch number and randomness from the state of the best block (but note that we are then _committing_ to build on top of this block _before_ deciding whether we will build). This is safe because the head of the chain has always been imported and can be assumed to have available state. Authoring would be impossible otherwise.

core/consensus/babe/src/lib.rs
```diff
@@ -463,27 +468,27 @@ fn check_header<B: Block + Sized, C: AuxStore>(
...
 		if sr25519::Pair::verify(&sig, pre_hash, author.clone()) {
 			let (inout, _batchable_proof) = {
 				let transcript = make_transcript(
-					Default::default(),
-					slot_num,
-					Default::default(),
-					0,
+					&randomness,
+					slot_number,
+					slot_number / slots_per_epoch,
```
rphmeier: ditto: wrong.

core/consensus/babe/src/lib.rs
```diff
@@ -617,9 +622,10 @@ impl<B: Block, C> Verifier<B> for BabeVerifier<C> where
 			.map_err(|e| format!("Could not extract timestamp and slot: {:?}", e))?;
 		let hash = header.hash();
 		let parent_hash = *header.parent_hash();
-		let authorities = authorities(self.client.as_ref(), &BlockId::Hash(parent_hash))
-			.map_err(|e| format!("Could not fetch authorities at {:?}: {:?}", parent_hash, e))?;
-
+		let Epoch { authorities, randomness } =
+			epoch(self.client.as_ref(), &BlockId::Hash(parent_hash))
```
rphmeier: I'd like a comment here explaining that this rests on an assumption that the parent is already executed and state is available.

core/consensus/babe/src/lib.rs
```diff
@@ -617,9 +622,10 @@ impl<B: Block, C> Verifier<B> for BabeVerifier<C> where
...
+			.map_err(|e| format!("Could not fetch epoch at {:?}: {:?}", parent_hash, e))?;
+		let authorities: Vec<_> = authorities.into_iter().map(|(s, _)| s).collect();
```
rphmeier: why ignore the weights? easiest thing for the moment is for the runtime to set all weights to `1`, but the consensus-side should still adjust the VRF calculation correctly

core/consensus/babe/primitives/src/lib.rs
```diff
@@ -49,15 +56,25 @@ pub type SlotNumber = u64;
...
-#[derive(Decode, Encode)]
+#[derive(Decode, Encode, Clone, PartialEq, Eq)]
 pub enum ConsensusLog {
 	/// The epoch has changed. This provides information about the
 	/// epoch _after_ next: what slot number it will start at, who are the authorities (and their weights)
 	/// and the next epoch randomness. The information for the _next_ epoch should already
 	/// be available.
 	#[codec(index = "1")]
-	NextEpochData(SlotNumber, Vec<(AuthorityId, Weight)>, [u8; VRF_OUTPUT_LENGTH]),
+	NextEpochData(Epoch),
```
rphmeier: The removal of `SlotNumber` here means you can't change the epoch length without a hard fork. And yet, the SRML code doesn't account for a constant `slots_per_epoch`. This is probably a bad idea. I would favor a dynamic `SlotNumber` in the `NextEpochData` digest which is the earliest slot number we can start the epoch after the one started by the digest.

DemiMarie-parity: @rphmeier is making `SlotNumber` part of `Epoch` sufficient?

core/consensus/slots/src/lib.rs
```diff
@@ -76,6 +78,7 @@ pub fn start_slot_worker<B, C, W, T, SO, SC>(
 	sync_oracle: SO,
 	inherent_data_providers: InherentDataProviders,
 	timestamp_extractor: SC,
+	slots_per_epoch: NonZeroU64,
```
andresilva: I think this is unused because we never look at the epoch from `SlotInfo` (and we shouldn't given @rphmeier's comments above).

core/consensus/babe/primitives/src/lib.rs
```diff
@@ -49,15 +56,25 @@ pub type SlotNumber = u64;
 /// The weight of an authority.
 pub type Weight = u64;
 
+/// BABE epoch information
+#[derive(Decode, Encode, Default, PartialEq, Eq, Clone)]
+#[cfg_attr(any(feature = "std", test), derive(Debug))]
+pub struct Epoch {
```
andresilva: Maybe add the index here? This way when we're verifying or authoring a block we fetch the current epoch from runtime state instead of computing it ourselves (leading to potential inconsistencies when using `current_slots / slots_per_epoch`).

core/consensus/babe/primitives/src/lib.rs
rphmeier: as long as it informs the client, yes.

core/consensus/babe/primitives/src/lib.rs
rphmeier: yes

DemiMarie-parity: @andresilva I had already done that before I saw your comment 😃

core/consensus/slots/src/lib.rs
DemiMarie-parity: Removed.

core/consensus/babe/src/lib.rs
DemiMarie-parity: Fixed, by moving epoch calculation entirely to the runtime.

DemiMarie-parity:
@rphmeier The biggest difficulty here, which I am having serious problems with, is that the epoch duration needs to be known at the start of each epoch.

core/consensus/babe/src/aux_schema.rs
```diff
@@ -0,0 +1,93 @@
...
+			epoch_under_test.epoch_index,
+		)))
+	}
+
+	let expected_slot_number =
+		this_epoch.duration.checked_add(last_epoch_change).unwrap_or_else(|| {
+			error!("BABE slot number overflow ― storage is corrupt!");
+			u64::MAX
+		});
+	if slot_now != expected_slot_number {
```
andresilva: Is this correct? It's possible that no one proposes at the given slot, should we check for `slot_now >= expected_slot_number` instead?

gavofyork:
@rphmeier final review?

DemiMarie-parity:
@gavofyork I need to add some more tests, but it is up to you if that should block Kusama

core/client/Cargo.toml
```diff
@@ -9,7 +9,7 @@ derive_more = { version = "0.14.0", optional = true }
 fnv = { version = "1.0", optional = true }
 log = { version = "0.4", optional = true }
 parking_lot = { version = "0.8.0", optional = true }
-hex = { package = "hex-literal", version = "0.2", optional = true }
+hex = { package = "hex-literal", version = "0.2.0", optional = true }
```
gavofyork: why?

srml/babe/src/lib.rs
```diff
@@ -131,19 +129,20 @@ decl_storage! {
...
-		EpochIndex get(epoch_index): u64;
+		/// Current epoch index
+		EpochIndex: u64;
 	}
 }
 
+pub fn authorities<T: Trait>() -> Vec<(AuthorityId, u64)> {
+	<Module<T> as Store>::Authorities::get()
+}
+pub fn random<T: Trait>() -> [u8; 32] {
```
gavofyork: isn't this already done with `get(random)` in the storage item declaration?

srml/babe/src/lib.rs
```diff
@@ -131,19 +129,20 @@ decl_storage! {
...
 		/// used where a number is needed that cannot have been chosen by an
 		/// adversary, for purposes such as public-coin zero-knowledge proofs.
-		EpochRandomness get(epoch_randomness): [u8; VRF_OUTPUT_LENGTH];
-
-		/// The randomness under construction
-		UnderConstruction: [u8; VRF_OUTPUT_LENGTH];
-
-		/// The randomness for the next epoch
-		NextEpochRandomness: [u8; VRF_OUTPUT_LENGTH];
+		pub Randomness get(random) config(): [u8; 32];
```
gavofyork: why would you want to configure the initial value of the random number?

srml/babe/src/lib.rs
```diff
@@ -207,65 +201,94 @@ impl<T: Trait> Module<T> {
...
+		Self::deposit_consensus(ConsensusLog::OnDisabled(i as u64))
 	}
 }
 
+fn compute_randomness(
+	last_epoch_randomness: [u8; 32],
+	epoch_index: u64,
+	rho: [u8; VRF_OUTPUT_LENGTH],
+) -> [u8; 32] {
+	let mut s = [0; 40 + VRF_OUTPUT_LENGTH];
```
gavofyork: this function is much more easily written `T::Hash::hash_of((last_epoch_randomness, epoch_index, rho))`

gavofyork:
still waiting for another pair of eyes here...

gavofyork:
@DemiMarie-parity needs resolving.

.gitignore
```diff
@@ -21,3 +21,4 @@ rls*.log
 *.orig
 *.rej
 **/wip/*.stderr
+[0-9][0-9][0-9][0-9]-*.patch
```
andresilva: imo these should go into your own global gitignore.

core/consensus/babe/src/lib.rs
```diff
@@ -820,285 +843,298 @@ fn initialize_authorities_cache<B, C>(client: &C) -> Result<(), ConsensusError>
...
+	B: Backend<Block, Blake2Hasher> + 'static,
+	E: CallExecutor<Block, Blake2Hasher> + 'static + Clone + Send + Sync,
+	RA: Send + Sync,
+{
+	type Error = ConsensusError;
+
+	fn import_block(
+		&mut self,
+		mut block: BlockImportParams<Block>,
+		new_cache: HashMap<well_known_cache_keys::Id, Vec<u8>>,
```
rphmeier: what is this populated with when the blocks don't come from the authorship task?

andresilva: https://github.com/paritytech/substrate/pull/3028/files#diff-671e8a8d7d8649656f0988fafca31301R689 same thing. the import queue then just create an hashmap and adds those keys in.

PR #3030, opened 2019-07-05
End build when `wasm-builder` does not find a prerequisite

bkchr:
Using `compile_error!` was a stupid idea by me, as rust would not re-execute the build.rs.

### DIFF
--- a/core/utils/wasm-builder/src/lib.rs
+++ b/core/utils/wasm-builder/src/lib.rs
@@ -78,7 +78,7 @@
 //! - wasm-gc
 //!
 
-use std::{env, fs, path::PathBuf, process::{Command, Stdio}};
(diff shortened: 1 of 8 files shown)


core/utils/wasm-builder/src/lib.rs
```diff
@@ -107,24 +107,16 @@ pub fn build_project(file_name: &str, cargo_manifest: &str) {
 	let cargo_manifest = PathBuf::from(cargo_manifest);
 
 	if !cargo_manifest.exists() {
-		create_out_file(
-			file_name,
-			format!("compile_error!(\"'{}' does not exists!\")", cargo_manifest.display())
-		);
-		return
+		panic!("'{}' does not exists!", cargo_manifest.display());
```
andresilva:
```suggestion
		panic!("'{}' does not exist!", cargo_manifest.display());
```

Issue #2536, opened 2019-05-10
Centralize the threading strategy in substrate-service for improved testing and debuggability


tomaka:
*This is a long-term issue for a general refactor of the threading strategy of Substrate.*

Let's have a briefly overview of how Substrate works at the moment:

- We have the client, which, in the case of a light client, holds an Arc to the network (through the on_demand service) and sends messages to the network.
- We have the import queue, which holds an Arc to the client and queries it.
- We have the network, that spawns a background thread, and holds an Arc to the client, the import queue, the transaction pool, and sends messages to them.
- We have grandpa, which holds an Arc to the network, sends messages to it, and receives back gossiping messages through a channel.
- We have the informant, that spawns a background thread, holds an Arc to the client, the network, and the transaction pool, and grabs data from them.
- We have the RPC service, which spawns a background thread and holds Arcs to the client and the network and grabs data from them.
- We have the telemetry, which spawns a background thread, and holds closures to be executed on connection, which includes Arcs to sub-objects.
- And finally we have the service, which initializes everything, and does the additional plugging in that all the other services don't perform, such as notifying the network when the import queue receives something.

This architecture surely is decentralized, but it has quite a few issues:

- Since there are Arcs and threads everywhere, there are also mutexes everywhere. It is quite hard to debug lock contention issues. And in terms of metrics, we generally don't know what's taking the most time.
- There are at least two cyclic dependencies (A contains an Arc to B, and B contains an Arc to A). This leads to some awkward `stop` methods.
- All the APIs that use channels are concurrent, and it's arguably harder and more error-prone to use a concurrent API compared to a basic synchronous API. It is extremely valuable to have APIs that return an error in case of bad input, but if you send your message over a channel you cannot have that feedback.
- It is extremely hard to discover bugs caused by services being out of sync. For example, service A may think that we're connected to a peer, while service B may think that we're not. Considering that everything is asynchronous, we can't really differentiate between a normal situation (the two services are not in sync *yet*, because one of them hasn't processed all its incoming messages) and an abnormal situation (the two services are not in sync because of a bug).
- Because of the previous point (can't differentiate between valid and invalid input), there's a lot of code that just does nothing in case of invalid input, instead of returning an error or printing a log message that could be helpful for detecting bugs.
- This design clearly doesn't scale in terms of complexity. If we continue adding services and inter-dependencies, we're going to increase the chances of lock contention and the complexity of the whole thing.
[...]

Issue #1693, opened 2019-02-05
Allow querying the owner of a key on the networking layer


tomaka:
Add a DHT whose entries are keys, and whose associated values are a list of PeerId and listening addresses of nodes which own the key. The entries are signed.

This should be done by modifying `network-libp2p`. We can add a method to `Service` to start a request for a specific key. On success, we should automatically connect to the node which owns the key. The `OpenCustomProtocol` that reports the successful connection should indicate that this corresponds to a fulfilled request. Also add a new variant in `ServiceEvent` for requests that time out or fail.

tomaka:
The Kademlia variant that libp2p uses has `GET_PROVIDERS` and `ADD_PROVIDER` messages that are specifically made for that. However they don't allow signing their record. We will probably need our own variant of Kademlia unfortunately.

DemiMarie-parity:
DHT = Distributed Hash Table?

tomaka:
Yes

mxinden:
Current plan / status:

1. **Making itself discoverable**
    1. Retrieve its external addresses
    2. Adds its network peer id to the addresses
    3. Sign the above with its authority key (for now consensus)
    4. Put the signature and the addresses on the libp2p Kademlia DHT

2. **Discovering other validators**
    1. Retrieve the current set of authorities (currently only consensus, not finality)
    2. Start DHT queries for the ids of the authorities
    3. Validate the signatures of the retrieved key value pairs
    4. Add the retrieved external addresses as priority nodes to the peerset

A work in progress version of the above is in https://github.com/paritytech/substrate/pull/3247.

xlc:
I am wondering how this plays with sentry node setup? i.e. validator node that is protected and only accessible via a sentry node
https://github.com/paritytech/polkadot/issues/264

mxinden:
The idea would be the following:

An operator would configure its validator with a set of multi-addresses of its sentry nodes, e.g. via  a command line flag or a configuration file. Instead of putting its own external address (which would not be routable on the public internet) on the dht, it would put the external addresses of its sentry nodes on the dht.

With the above, validators trying to connect to other validators can find the multi-addresses of a validator or the multi-addresses of a validator's sentry nodes on the dht.

Does that answer your question @xlc?

xlc:
Yes take makes total sense.

mxinden:
[...]

Issue #1063, opened 2018-11-02
stop generating keys from seeds in service and keystore


azban:
as per this [comment](https://github.com/paritytech/substrate/blob/d7bb960c9d2a240a17ea18d45be57434b5ba02ba/core/keystore/src/lib.rs#L156), we should find some way to prefill the keystore with keys that are generated from seed that is not in the standard execution path, because it is meant to be used for testing only.

chevdor:
FYI: There is a very similar piece of code outside the keystore:
https://github.com/paritytech/substrate/blob/master/subkey/src/main.rs#L44

webmaster128:
> that is not in the standard execution path

what do you mean by "standard execution path"?

There seems to be a lot of naming confution here. `generate_from_seed` generates an Ed25519 seed from a 0-32 character string.

azban:
> what do you mean by "standard execution path"?
 it's just for testing

> There seems to be a lot of naming confution here
the name seems fine with some amount of context.. do you have a suggestion?

webmaster128:
> do you have a suggestion?

What about this derication chain: passphrase -> seed -> keypair

The good thing about the term "passphrase" is that it implies to be choosen by user and it has the potential of being low entropy.

webmaster128:
There are two types of seeds here in as input of `generate_from_seed`:

1. an arbitrary string
2. a 256bit Ed25519 seed

Do you want to get rid of both? Or only get rid of the 1. -> 2. derivation?

dvc94ch:
Is it ok to remove this code? After updating the keystore to handle lru, this is inconvenient since we'd have to store testing and session_keys separately to avoid the testing key being auto removed.

gavofyork:
This is sorted with the new RPCs.

PR #3034, opened 2019-07-05
Implement session key handover

dvc94ch:
TODO:
- [x] use session store in babe
- [x] use session store in grandpa
- [x] lru to remove old keys
- [x] work with offchain workers api
- [ ] move some cli args to node
- [ ] update im-online
- [ ] clean up mess
cc @joepetrowski


rphmeier:
This suffers from the problem I mentioned earlier, which is that users are never going to set their key _exactly_ on the end of a session. it will always be a bit before or after. If all nodes set their new keys a bit before the end of the session, no nodes will author blocks, and the new session will never start until someone intervenes manually.

So we probably need it to author on both keys for some time.

gavofyork:
So the way this should work:

- User calls RPC at some point *before* session with generate new key. Key is stored and indexed in local DB.
- User registers the new key for their next session (actually, with session key buffering, it'll kick in one session later, but that doesn't really matter here).
- Whenever a new block is to be authored or finalised or whatever, the key to be used *is not fixed*. Rather it just checks for all current session keys (i.e. keys that would be valid to sign with in this context) and cross-references them with the indexed list of keys *that we control*. If and only if there is one, then that key is used. If there isn't, then it's assumed that we're no longer validating and we don't bother signing anything.
- Keys should be kept in the key database along with an LRU (last recently used) timestamp and a boolean that get flagged when the key is replaced/a new key is generated. Keys that have been unused for over 24 hours and which have been replaced should be removed from the index.

This should be trivial for Babe and Aura. It might require a little help from @rphmeier for Grandpa.

dvc94ch:
> in BABE we probably need to start a new authoring future for every key that owns a slot, because slot ownership is associated w/ a VRF proof

Can a validator be assigned more than one slot? In the session module only one key can be active per validator. I'm just taking the first slot that we have a key for.

rphmeier:
@dvc94ch yep, a validator can have more than one slot. What you should do is rather take the first key that owns the slot, if any. So do the vrf-proof on all keys in the current validator set and _if there's only one_ do we author on the slot with that key

core/consensus/babe/src/lib.rs
```diff
@@ -773,11 +774,15 @@ fn claim_slot(
...
-) -> Option<((VRFInOut, VRFProof, VRFProofBatchable), usize)> {
-	let public = &key.public();
-	let index = authorities.iter().position(|s| s == public)?;
+) -> Option<((VRFInOut, VRFProof, VRFProofBatchable), usize, AuthorityPair)> {
+	let (index, key) = authorities.iter()
+		.enumerate()
+		.filter_map(|(index, public)| {
+			session_store.get_key(public).map(|key| (index, key))
+		})
+		.nth(0)?;
```
rphmeier:
```suggestion
		.next()?;
```

rphmeier: this also doesn't ensure that there is only 1 key owned by us which can author at this slot.

core/finality-grandpa/src/environment.rs
```diff
@@ -497,14 +498,18 @@ where
 		let prevote_timer = Delay::new(now + self.config.gossip_duration * 2);
 		let precommit_timer = Delay::new(now + self.config.gossip_duration * 4);
 
-		let local_key = self.config.local_key.as_ref()
-			.filter(|pair| self.voters.contains_key(&pair.public().into()));
+		let local_key = self.session_store.keys()
+			.iter()
+			.filter(|pair| self.voters.contains_key(&pair.public().into()))
+			.nth(0)
```
rphmeier:
```suggestion
			.next()
```

core/finality-grandpa/src/environment.rs
```diff
@@ -529,9 +534,11 @@ where
 	}
 
 	fn proposed(&self, _round: u64, propose: PrimaryPropose<Block>) -> Result<(), Self::Error> {
-		let local_id = self.config.local_key.as_ref()
+		let local_id = self.session_store.keys()
```
rphmeier: I don't think this or the logic in`prevoted`, etc. is right.

`round_data` chooses which key to sign messages with (i.e. which key is _actually_ voting). But this may return a completely different identity in the case that the session store's underlying key set has changed in the meantime. And then we'd store the vote to disk as if _that_ identity signed the vote.

We should probably select a key upon construction of an `Environment`. This is done every time the validator set changes anyway.

dvc94ch:
yeah, the main goal was to get it to build. I'm studying the aura paper you sent me a link to and expect to get through the babe and grandpa papers till tomorrow.

rphmeier:
@dvc94ch OK, great. In the meantime I've checked the logic in Aura and BABE and it seems right.

core/consensus/babe/src/lib.rs
dvc94ch: the aura paper states that there is only one elected leader for any slot. so either we are the leader or not. I don't understand why this doesn't work. I've not looked at the babe paper yet...

dvc94ch:
So updated babe and grandpa according to your comments. I think babe is correct now and grandpa selects a key before creating an environment. I'm not sure if the process to select the key is correct. It takes the first key that is in the `VoterSet` if there is one.

core/consensus/babe/src/lib.rs
rphmeier: (we discussed offline, it's because BABE may have multiple leaders for 1 slot)

core/finality-grandpa/src/lib.rs
```diff
@@ -556,9 +553,13 @@ pub fn run_grandpa_voter<B, E, Block: BlockT<Hash=H256>, N, RA, SC, X>(
 	};
 
 	let voters = authority_set.current_authorities();
+	let local_key = session_store
+		.get_keys(|public| voters.contains_key(public))
+		.into_iter().next();
```
rphmeier:
```suggestion
		.into_iter()
		.next();
```

core/finality-grandpa/src/lib.rs
```diff
@@ -683,12 +684,16 @@ pub fn run_grandpa_voter<B, E, Block: BlockT<Hash=H256>, N, RA, SC, X>(
 					aux_schema::write_voter_set_state(&**client.backend(), &set_state)?;
 
 					let set_state: SharedVoterSetState<_> = set_state.into();
-
+					let voters: VoterSet<_> = new.authorities.into_iter().collect();
+					let local_key = session_store
+						.get_keys(|public| env.voters.contains_key(public))
+						.into_iter().next();
```
rphmeier:
```suggestion
						.into_iter()
						.next();
```

core/finality-grandpa/src/lib.rs
```diff
@@ -683,12 +684,16 @@ pub fn run_grandpa_voter<B, E, Block: BlockT<Hash=H256>, N, RA, SC, X>(
 					aux_schema::write_voter_set_state(&**client.backend(), &set_state)?;
 
 					let set_state: SharedVoterSetState<_> = set_state.into();
-
+					let voters: VoterSet<_> = new.authorities.into_iter().collect();
+					let local_key = session_store
+						.get_keys(|public| env.voters.contains_key(public))
```
rphmeier: should rather be `new.authorities.contains_key`, right? this is setting the key based off of the _previous_ authority set where it should be based on the new authority set.

core/keystore/src/lib.rs
```diff
@@ -151,22 +145,123 @@ impl Store {
...
+}
+
+/// A local session key store supporting a single crypto algorithm.
+#[derive(Clone)]
+pub struct SessionStore<TPair: Pair> {
+	keys: Arc<Mutex<HashMap<TPair::Public, (TPair, Instant)>>>,
+}
+
+impl<TPair> SessionStore<TPair>
+where
```
rphmeier: this is not typically the style used in the codebase. we'd usually put the `where` on the preceding line.

core/primitives/src/crypto.rs
```diff
@@ -452,6 +452,41 @@ pub mod key_types {
 	pub const SR25519: KeyTypeId = 20;
 }
 
+/// An enum of known key types.
+#[cfg(feature = "std")]
+pub enum KeyType {
```
rphmeier: the reason I didn't set this out in the first place is that we don't want to claim to define _all_ key types. It's better to do the typedef because it gives users freedom to define their own crypto.

core/rpc/src/author/mod.rs
```diff
@@ -176,4 +188,24 @@ impl<B, E, P, RA> AuthorApi<ExHash<P>, BlockHash<P>> for Author<B, E, P, RA> whe
 	fn unwatch_extrinsic(&self, _metadata: Option<Self::Metadata>, id: SubscriptionId) -> Result<bool> {
 		Ok(self.subscriptions.cancel(id))
 	}
+
+	fn set_key(&self, key_type_id: KeyTypeId, key: Vec<u8>) -> Result<()> {
```
rphmeier: doing the keygen in this function is almost definitely wrong. The `key` bytes should be the key itself, not a seed. then you wouldn't need to try and do an exhaustive match on the key-type-id.

core/keystore/src/lib.rs
```diff
@@ -151,22 +145,123 @@ impl Store {
...
+		&mut self,
+		handler: THandler,
+	) {
+		self.handlers.push((TKey::KEY_TYPE, Box::new(handler)));
+	}
+}
+
+/// A local session key store supporting a single crypto algorithm.
+#[derive(Clone)]
+pub struct SessionStore<TPair: Pair> {
```
rphmeier: why not a store supporting _all_ crypto algorithms?

dvc94ch: the idea is that aura/babe and grandpa get their own session store and only subscribe to the crypto they want. and if libp2p needs their secio key it also would get it's own session store. Do you disagree with the design?

core/rpc/src/author/mod.rs
dvc94ch: How would this avoid doing an exhaustive match? We need to match on KeyTypeId to go from Vec<u8> to impl Pair.

core/keystore/src/lib.rs
dvc94ch: this is the official rust style. do you feel strongly about this?

core/service/src/lib.rs
```diff
@@ -185,10 +185,6 @@ impl<Components: components::Components> Service<Components> {
 		// This is meant to be for testing only
 		// FIXME #1063 remove this
 		if let Some(keystore) = keystore.as_mut() {
-			for seed in &config.keys {
```
dvc94ch: I removed this without a replacement, since FIXME #1063 suggests removing it. I can add it back to the store if it's still required for something...

core/keystore/src/lib.rs
dvc94ch: The disadvantage of supporting multiple cryptos concurrently is that we have to store them as bytes and then decode/encode every time.

core/keystore/src/lib.rs
```diff
@@ -151,22 +145,126 @@ impl Store {
...
 		let mut buf = self.path.clone();
 		let bytes: [u8; 4] = TPair::KEY_TYPE.to_le_bytes();
 		let key_type = hex::encode(bytes);
 		let key = hex::encode(public.as_slice());
 		buf.push(key_type + key.as_str());
 		buf
 	}
+
+	/// Register a `StoreHandler`.
+	fn register_handler<TKey: TypedKey, THandler: StoreHandler + 'static>(
```
montekki: Can this be `pub`? It could be useful to add `NetworkService` here as a handler so changes to session keys can get propagated to the network DHT right away.

dvc94ch: I'd rather not pass byte vecs around, this is private to avoid that.

node/cli/src/service.rs
```diff
@@ -110,21 +112,20 @@ construct_service_factory! {
...
 
 				let config = grandpa::Config {
-					local_key: grandpa_key.map(Arc::new),
 					// FIXME #1578 make this available through chainspec
 					gossip_duration: Duration::from_millis(333),
 					justification_period: 4096,
 					name: Some(service.config.name.clone())
 				};
 
-				match config.local_key {
```
mxinden: Previously when a user specified no key, Substrate would not start a grandpa voter service. Do I understand correctly that if a user does not specify a key and also does not set the `disable-grandpa` flag, substrate is going to start a grandpa voter now?

node/cli/src/service.rs
```diff
@@ -110,21 +112,20 @@ construct_service_factory! {
...
+				match grandpa_session_store {
```
mxinden: `grandpa_session_store` is only `None` when `service.config.disable_grandpa` is true. In addition `grandpa_session_store` is only used within the `Some` branch of this `match`. How about doing `match service.config.disable_grandpa` here and constructing `grandpa_session_store` inside the `Some` branch?

node/cli/src/service.rs
dvc94ch: Yes. If you do not want to be a grandpa voter, which is the default, you have to disable it.

dvc94ch: But the grandpa voter can handle the case where there is no key...

node/cli/src/service.rs
dvc94ch: this is because the keystore is optional, we can only construct a session_store if we have a keystore.

dvc94ch: @tomaka made it optional recently to compile substrate to wasm

mxinden: Ah, right `service.get_session_store::<grandpa_primitives::AuthorityPair>()` also returns an `Option` which could be `None`. Thanks.

core/service/src/lib.rs
```diff
@@ -81,7 +81,7 @@ pub struct Service<Components: components::Components> {
 		NetworkStatus<ComponentBlock<Components>>, NetworkState
 	)>>>>,
 	transaction_pool: Arc<TransactionPool<Components::TransactionPoolApi>>,
-	keystore: Option<Arc<Mutex<Store>>>,
+	pub keystore: Arc<Mutex<Store>>,
```
tomaka: Why `pub`?

dvc94ch: To avoid adding all store methods to service. They are all used (exclusively) in the AuthoritySetup part of node/cli/src/service.rs

node/cli/src/service.rs
dvc94ch: @mxinden do we still need the --grandpa-voter flag after this PR?

core/service/src/lib.rs
tomaka: This is extremely bad, but not easy to do properly without refactoring `components.rs`
Please at least add a `lock_keystore()` method on the service that returns a lock to the keystore, instead of having public fields.

core/service/src/lib.rs
```diff
@@ -869,33 +827,12 @@ impl<C: Components> network::TransactionPool<ComponentExHash<C>, ComponentBlock<
...
-				.map(|keys| keys.get(0)
-					.map(|k| keystore.load(k, self.password.as_ref()))
-				);
-
-		if let Ok(Some(Ok(key))) = loaded_key {
-			Some(key)
-		} else {
-			None
-		}
+		None
```
dvc94ch: @tomusdrw: I'm not sure how to implement this since it doesn't mingle well with the design of this PR.

- session key transitions are the responsability of aura/babe/grandpa
- the keystore broadcast new keys to local keystores
- after a session key transition aura/babe/grandpa remove the old session key from their local keystore

core/service/src/lib.rs
```diff
@@ -261,20 +229,17 @@ impl<Components: components::Components> Service<Components> {
...
 		let offchain_storage = client.backend().offchain_storage();
 		let offchain_workers = match (config.offchain_worker, offchain_storage) {
 			(true, Some(db)) => {
 				Some(Arc::new(offchain::OffchainWorkers::new(
 					client.clone(),
 					db,
-					keystore_authority_key.clone(),
-					config.password.clone(),
+					keystore_authority_key,
+					"".to_string().into(),
```
tomusdrw: I think it would be better to introduce a CLI flag to pass the password for offchain worker keys to prevent keeping them in raw form in the DB.
I'm fine if it''s not part of this PR, but please create an issue and add a TODO in the code.

core/service/src/lib.rs
tomusdrw: Offchain workers should basically get the same session key that is used for everything else. I suppose it will require querying the runtime (latest block) to do an intersection between local keys and the ones set there.

core/service/src/lib.rs
dvc94ch: do offchain workers require a password? In this PR I'm proposing not ever persisting a key to the filesystem as for session keys we don't want that. Currently session keys are the only ones stored in the keystore

tomusdrw: Offchain workers can create (and persist) their own crypto keys. That ends up in the local database, and was additionally protected by the same password that user passed to unlock authority keys.
Now that I think of it, I should have really introduced a separate CLI flag right from the start.

tomaka:
Sharing the keystore with the RPC goes in the opposite direction of what I'm doing with #2536, but it's fine for now.

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
-	pub fn load<TPair: Pair>(&self, public: &TPair::Public, password: &str) -> Result<TPair> {
-		if let Some(pair) = self.get_pair(public)? {
-			return Ok(pair)
-		}
+impl<TPair> LocalStore<TPair>
+where
+	TPair: Pair + Send + Clone,
+	TPair::Public: Send + Clone,
+{
+	/// Creates a new `LocalSessionStore`.
```
mxinden:
```suggestion
	/// Creates a new `LocalStore`.
```

mxinden:
Related question:

Goal is to be able to have validators keep long running connections to x amount of other validators. To do so a validator would need to know the `Vec<MultiAddr>` of other validators (multiple multi addresses, as a validator might be accessible through different subnets or through different sentry nodes).

We can gossip this information through a DHT (https://github.com/paritytech/substrate/issues/1693) as a mapping of `SessionKey -> Vec<MultiAddr>`.

With the above we need to update our DHT entry once our own session key changes. **Is this the right place to hook into the session key change event?** If so, would it be possible to have a public function to add oneself to `keystore::Store.handlers`?

Relates to @montekki earlier comment: https://github.com/paritytech/substrate/pull/3034#discussion_r301294608

dvc94ch:
I think this is similar to the offchain workers problem. @rphmeier might have some input.

When you say SessionKey -> Vec<Multiaddr>, which session key do you mean? We have multiple. Also is it relevant for networking? You have a secio key which serves as an identity, in theory multiple validators could be running in the same process and share a networking identity.

mxinden:
@dvc94ch where can I read up on the different session keys? As far as I can tell I am talking about [this kind of session key](https://substrate.dev/docs/en/overview/session-keys#docsNav).

Would it make more sense to hook into the *session module* via the [`SessionHandler`](https://github.com/paritytech/substrate/blob/master/srml/session/src/lib.rs#L187) trait?

> Also is it relevant for networking?

Yes. I need the session key to sign and validate the multi addresses of a validator, inserted and retrieved in the above mentioned DHT, in order to keep long living libp2p connections to them.

node/cli/src/service.rs
mxinden: @dvc94ch do I understand correctly, that the only way of only running a grandpa *observer* not a *voter* is through the `disable-grandpa` flag? In that case, no, the `--grandpa-voter` flag is obsolete. Would you mind removing it within this pull request and adjusting the `docker-compose.yml` so the master branch stays consistent? Let me know if you want me to push a commit.

```diff
diff --git a/scripts/sentry-node/docker-compose.yml b/scripts/sentry-node/docker-compose.yml
index 78f8ba36b..661ab9433 100644
--- a/scripts/sentry-node/docker-compose.yml
+++ b/scripts/sentry-node/docker-compose.yml
@@ -93,9 +93,6 @@ services:
       # Not only bind to localhost.
       - "--ws-external"
       - "--rpc-external"
-      # Make sure sentry-a still participates as a grandpa voter to forward
-      # grandpa finality gossip messages.
-      - "--grandpa-voter"
```

Also it might be worth extending the `--help` output a bit:

```diff
diff --git a/core/cli/src/params.rs b/core/cli/src/params.rs
index b22d25a9d..64bdfc692 100644
--- a/core/cli/src/params.rs
+++ b/core/cli/src/params.rs
@@ -297,7 +297,7 @@ pub struct RunCmd {
        #[structopt(long = "validator")]
        pub validator: bool,

-       /// Disable GRANDPA voter
+       /// Disable GRANDPA voter only running a GRANDPA observer
        #[structopt(long = "no-grandpa")]
        pub no_grandpa: bool,
```

node/cli/src/service.rs
```diff
@@ -83,8 +82,12 @@ construct_service_factory! {
 				let (block_import, link_half) = service.config.custom.grandpa_import_setup.take()
 					.expect("Link Half and Block Import are present for Full Services or setup failed before. qed");
 
-				if let Some(aura_key) = service.authority_key::<AuraPair>() {
-					info!("Using aura key {}", aura_key.public());
+				if service.config.roles == Roles::AUTHORITY {
```
andresilva: This is not correct, you might want to run the grandpa voter as a full node (in fact in master we just added a flag for this).

node/cli/src/service.rs
```diff
@@ -110,44 +113,40 @@ construct_service_factory! {
...
-							config: config,
-							link: link_half,
-							network: service.network(),
-							inherent_data_providers: service.config.custom.inherent_data_providers.clone(),
-							on_exit: service.on_exit(),
-							telemetry_on_connect: Some(telemetry_on_connect),
-						};
-						service.spawn_task(Box::new(grandpa::run_grandpa_voter(grandpa_config)?));
-					},
+				if service.config.roles == Roles::AUTHORITY && !service.config.disable_grandpa {
```
andresilva: This is not correct, you might want to run the grandpa voter as a full node (in fact in master we just added a flag for this).

dvc94ch: Is there a reason why you can't just use --validator? @mxinden thought that would also work for sentry, but I can add it back if you need a separate flag.

dvc94ch: If aura/babe are not staked as validators they won't author a block, since they don't get a slot, it's harmless to run the aura task, are you concerned about performance?

dvc94ch: ok I think the best and clearest solution is having an `--aura` `--grandpa-voter` and `--validator` option where `--validator` is just a shortcut for --aura --grandpa-voter. Then we can support all use cases and the semantics are clear (opt-in)

dvc94ch:
just to confirm I got it right this time: running grandpa voter is the default, if role == Authority we add a key and if no-grandpa we just run the observer. so we don't need a --grandpa-voter flag after all...

andresilva:
Currently the node role makes no difference regarding GRANDPA, if `--validator` is passed the node role is set as Authority (but maybe we should make sure that we only start the voter with a key if the role is Authority). 

If no key is provided we start the observer (this is the default), if a key is provided we start the voter. If no key is provided but the flag `--grandpa-voter` is enabled, we start the voter with no key. If some key is provided but we provide the flag `--no-grandpa`, then we start the observer (presumably the key would only be used for running Aura/Babe worker).

dvc94ch:
but you can set a key now via rpc. it's completely reasonable to start a node in validator mode without a key and then set the keys via rpc. I don't think the logic should depend on if there is a key passed via cli.

andresilva:
I need to review this PR more in-depth, I was just pointed to that change by @bkchr. IMO the observer should be the default if the node isn't started with a key. If you want to set a key through the RPC at a later time then you start the node with `--grandpa-voter` which makes sure the voter is indeed running (and not the observer).

dvc94ch:
so I think the last version I had with a --aura --grandpa-voter and --validator being a shortcut for --aura and --grandpa-voter would be the way to go. that handles all the use cases no?

tomaka:
Maybe we could just go with `Box<AuthorityKeyProvider>` for now instead of adding 200 lines of code to the service factory?

dvc94ch:
@tomaka it's not that bad... sorry for screwing with you refactoring...

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
-	pub fn generate_from_seed<TPair: Pair>(&mut self, seed: &str) -> Result<TPair> {
-		let pair = TPair::from_string(seed, None)
-			.ok().ok_or(Error::InvalidSeed)?;
-		self.insert_pair(&pair);
-		Ok(pair)
-	}
+/// A local session store supporting a single crypto algorithm.
+#[derive(Clone)]
+pub struct LocalStore<TPair: Pair> {
+	keys: Arc<Mutex<HashMap<TPair::Public, (TPair, Instant)>>>,
```
tomusdrw: I'd rather move `Arc` out of the struct. Cloning an `Arc<LocalStore>` shows the intention better imho - you can also get rid of a `Box` that way.

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
+	pub fn get_keys(&self, f: impl Fn(&TPair::Public) -> bool) -> Vec<TPair> {
+		let filtered_keys = {
+			let mut keys = self.keys.lock();
+			let filtered_keys = keys.iter()
+				.filter(|(public, _)| f(public))
+				.map(|(_, (pair, _))| pair.clone())
+				.collect::<Vec<_>>();
+			for key in &filtered_keys {
+				keys.get_mut(&key.public())
+					.expect("We hold a lock; qed").1 = Instant::now();
```
tomusdrw: Well the proof should rather be:
```suggestion
					.expect("All filtered_keys are known to exist in `keys`; qed").1 = Instant::now();
```

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
-			.ok().ok_or(Error::InvalidPhrase)?;
-		if &pair.public() != public {
-			return Err(Error::InvalidPassword);
-		}
-		Ok(pair)
+	/// Find keys matching a predicate.
+	pub fn get_keys(&self, f: impl Fn(&TPair::Public) -> bool) -> Vec<TPair> {
+		let filtered_keys = {
+			let mut keys = self.keys.lock();
+			let filtered_keys = keys.iter()
```
tomusdrw: You can actually use `iter_mut` to avoid traversing the list twice and having a need to provide `expect` proof.

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
-			if let Some(name) = path.file_name().and_then(|n| n.to_str()) {
-				match hex::decode(name) {
-					Ok(ref hex) => {
-						if hex[0..4] != key_type { continue	}
-						let public = TPublic::from_slice(&hex[4..]);
-						public_keys.push(public);
-					}
-					_ => continue,
-				}
+	/// Cleanup session store.
```
tomusdrw: Would be good to describe why single key will always be kept.

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
+	fn clean(&self) {
+		let mut keys = self.keys.lock();
+		if keys.len() > 1 {
+			let duration = Duration::new(INACTIVITY_PERIOD, 0);
+			let new_keys = keys.iter().filter(|(_, (_, instant))| {
```
tomusdrw: just use `retain`?

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
+				Instant::now().duration_since(instant.clone()) < duration
+			}).map(|(k, v)| (k.clone(), v.clone())).collect::<HashMap<_, _>>();
+			if new_keys.len() > 0 {
```
tomusdrw: Why do we keep the old keys?

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
-			return Ok(pair)
-		}
+impl<TPair> LocalStore<TPair>
+where
+	TPair: Pair + Send + Clone,
+	TPair::Public: Send + Clone,
+{
+	/// Creates a new `LocalStore`.
+	pub fn new() -> Self {
+		let session_store = Self {
```
tomusdrw: Just return `Self { keys: Default::default() }` right away no need to introduce the variable.

Or actually best just derive `Default`.

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
-impl std::error::Error for Error {
-	fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-		match self {
-			Error::Io(ref err) => Some(err),
-			Error::Json(ref err) => Some(err),
-			_ => None,
-		}
-	}
+/// Handler for `Store` events.
+trait StoreHandler: Send {
```
tomusdrw: Can we have a `const` value of keytype here and have `StoreHandler` `pair` arguement be properly typed?

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
-		buf.push(key_type + key.as_str());
-		buf
+impl<TPair> StoreHandler for LocalStore<TPair>
+where
+	TPair: Pair + Send,
+	TPair::Public: Send,
+{
+	fn on_set_session_key(&self, pair: &[u8]) {
+		let pair = TPair::from_seed_slice(pair)
+			.expect("store contains valid bytes");
```
tomusdrw: Imho the proof is insufficient. It will only give us a correct `pair` if the handler is filtered correctly (and hence registered correctly) by it's key type. I'd rather look for a solution that passes a properly typed `Pair` here.

core/rpc/src/author/mod.rs
```diff
@@ -72,6 +74,10 @@ pub trait AuthorApi<Hash, BlockHash> {
 	/// Unsubscribe from extrinsic watching.
 	#[pubsub(subscription = "author_extrinsicUpdate", unsubscribe, name = "author_unwatchExtrinsic")]
 	fn unwatch_extrinsic(&self, metadata: Option<Self::Metadata>, id: SubscriptionId) -> Result<bool>;
+
+	/// Add session key.
```
tomusdrw: Does it add a key or `set` it? Setting implies the previous one is being replaced, might be worth to rename the method?

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
 	}
 
-	fn get_pair<TPair: Pair>(&self, public: &TPair::Public) -> Result<Option<TPair>> {
-		let key = (TPair::KEY_TYPE, public.to_raw_vec());
-		if let Some(bytes) = self.additional.get(&key) {
-			let pair = TPair::from_seed_slice(bytes)
-				.map_err(|_| Error::InvalidSeed)?;
-			return Ok(Some(pair));
+	/// Adds a session key pair.
+	pub fn set_key<TPair: Pair>(&self, pair: &TPair) {
```
tomusdrw: Should it rather be called `add_key`?

core/service/src/components.rs
```diff
@@ -390,6 +404,8 @@ pub trait Components: Sized + 'static {
 	type ImportQueue: ImportQueue<FactoryBlock<Self::Factory>> + 'static;
 	/// The Fork Choice Strategy for the chain
 	type SelectChain: SelectChain<FactoryBlock<Self::Factory>>;
+	/// The authority key provider.
+	type AuthorityKeyProvider: offchain::AuthorityKeyProvider + Send + Sync;
```
tomusdrw: Can you elaborate why it needs to be part of `Service`?

node-template/src/service.rs
```diff
@@ -92,6 +95,9 @@ construct_service_factory! {
 				Ok(service)
 			}
 		},
+		AuthorityKeyProvider = () {
+			|_, _| Ok(())
```
tomusdrw: Not convinced about this. The `node-template` should really have proper configuration for offchain workers.

node/cli/src/service.rs
```diff
@@ -63,6 +66,60 @@ impl<F> Default for NodeConfig<F> where F: substrate_service::ServiceFactory {
...
+			client: self.client.clone(),
+			aura_store: self.aura_store.clone(),
+			grandpa_store: self.grandpa_store.clone(),
+		}
+	}
+}
+
+impl offchain::AuthorityKeyProvider for KeyProvider {
+	fn authority_key<TPair: Pair>(&self) -> Option<TPair> {
+		if TPair::KEY_TYPE == <AuraPair as TypedKey>::KEY_TYPE {
```
tomusdrw: What if `AuraPair::KEY_TYPE == GrandpaPair::KEY_TYPE`? We will use a wrong `runtime_api` it seems.
If that's a case we should probably call both apis and just find the first key that works?

core/cli/src/params.rs
```diff
@@ -305,25 +305,20 @@ pub struct ExecutionStrategies {
...
-	pub no_grandpa: bool,
+	/// Enable AURA.
+	#[structopt(long = "aura")]
+	pub aura: bool,
 
-	/// Run GRANDPA voter even when no additional key seed via `--key` is specified. This can for example be of interest
-	/// when running a sentry node in front of a validator, thus needing to forward GRANDPA gossip messages.
-	#[structopt(long = "grandpa-voter")]
+	/// Enable GRANDPA voter.
+	#[structopt(long = "grandpa_voter")]
```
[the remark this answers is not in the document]
dvc94ch: typo

dvc94ch:
```suggestion
	#[structopt(long = "grandpa-voter")]
```

node/cli/src/service.rs
dvc94ch: if the key isn't found in the aura authorities then we move on to the grandpa_authorities.

core/service/src/components.rs
dvc94ch: We need to know the AuraPair and GrandpaPair which aren't known yet. We could add those as types instead.

tomusdrw: Yeah, I think adding the key types to components/service would be better than allowing to customize the entire key provider.

node/cli/src/service.rs
tomusdrw: Oh, that's good then, I didn't notice. I like your idea splitting the API into two separate methods (one for grandpa, one for authority_key), but we can address that in a separate PR.

core/service/src/lib.rs
```diff
@@ -457,9 +425,9 @@ impl<Components: components::Components> Service<Components> {
 						"implementation" => impl_name.clone(),
 						"version" => version.clone(),
 						"config" => "",
+						"aura" => aura,
```
tomaka: That's the data that is sent to the telemetry server. Modifying it means that the backend needs to be updated as well. Additionally, I'd really prefer if what was in the `aura` and `grandpa_voter` variables had a well-defined structure, so as to not break the telemetry.

dvc94ch:
Can you elaborate which changes you mean? To the Factory/Components?

core/keystore/src/lib.rs
```diff
@@ -14,185 +14,180 @@
...
+		let value = pair.to_raw_vec();
+		let handlers = self.handlers.lock();
+		for (key_type, handler) in handlers.iter() {
+			if *key_type == TPair::KEY_TYPE {
+				handler.on_set_session_key(&value[..]);
```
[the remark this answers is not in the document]
dvc94ch: we know how long it will take because it's private, so the only implementation is in this file...

core/keystore/src/lib.rs
dvc94ch: retain would only work if we don't want to keep old keys

core/keystore/src/lib.rs
dvc94ch: to prevent a validator from being unstaked but later restaked. this could lead to the validator getting slashed if he no longer has his session key

tomaka:
> Can you elaborate which changes you mean? To the Factory/Components?

The entire thing, including the keystore.
Why are these "local store" and "store handler" systems necessary? Why do we have to add these undocumented complicated abstraction here, as opposed to passing an `Arc<Keystore>` everywhere?

I'm not saying it's a bad PR, but:

- The reason for this design isn't explained.
- The introduced code isn't really documented.
- I feel like it could be done in a way simpler way, but I can't be sure because of the first two reasons.

dvc94ch:
It's a sink/drain pattern where the rpc has the sink and consensus/finality
has the drain. Consensus and finality have different session key transition
mechanisms so aura can drop a key while grandpa will uses it.

dvc94ch:
since the AuthorityKeyProvider requires the aura api and the grandpa api it has to be implemented in the node. Otherwise we introduce dependencies on aura/grandpa to the service.

core/finality-grandpa/src/lib.rs
```diff
@@ -189,6 +191,58 @@ type CommunicationOutH<Block, H> = grandpa::voter::CommunicationOut<
...
 
+/// Aura key provider for offchain workers.
+pub struct GrandpaKeyProvider<Block, Client> {
+	_marker: PhantomData<Block>,
+	client: Arc<Client>,
+	keystore: Arc<LocalStore<fg_primitives::AuthorityPair>>,
+}
+
+impl<Block, Client> GrandpaKeyProvider<Block, Client>
+where
```
gavofyork: move to end of line above

gavofyork: and same on further instances of the style

core/keystore/src/lib.rs
gavofyork: you mean a `type KeyType;` here to allow `fn on_add_session_key(&mut self, pair: KeyType)`?

core/keystore/src/lib.rs
gavofyork: Indeed. This proof is wholly insufficient. A proof should be a logical tautology and not contain assumptions about prior state.

core/offchain/src/api.rs
```diff
@@ -52,128 +51,50 @@ enum ExtMessage {
...
 
 impl StoredKey {
-	fn generate_with_phrase(kind: CryptoKind, password: Option<&str>) -> Self {
+	fn generate_with_phrase(
+		kind: KeyTypeId,
+		password: Option<&str>,
+	) -> Result<Self, ()> {
 		match kind {
-			CryptoKind::Ed25519 => {
+			key_types::ED25519 => {
```
gavofyork: bad capitalisation.

core/primitives/src/crypto.rs
```diff
@@ -625,6 +625,31 @@ pub trait TypedKey {
...
+///
+/// 0-1024 are reserved.
+pub type KeyProviderId = u32;
+
+/// Constant key proviers.
+pub mod key_providers {
+	use super::KeyProviderId;
+
+	/// Aura key provider.
+	pub const AURA: KeyProviderId = 10;
```
gavofyork: Er... no.

core shouldn't be hard-coding consensus-algorithm-related stuff in here.

burdges:
@rphmeier :
> yep, a validator can have more than one slot. What you should do is rather take the first key that owns the slot, if any. So do the vrf-proof on all keys in the current validator set and _if there's only one_ do we author on the slot with that key

Actually this sounds confusing and/or harmful.  In any slot, there should only be _one_ babe or grandpa key from which the network accepts signatures.  

I think gav's description sounds correct https://github.com/paritytech/substrate/pull/3034#issuecomment-508822046 with the caveat that actually only one current babe or grandpa key is valid.  At least BABE need the transaction that registers session key changes finalized in an epoch before the epoch switch. 

We should likely include a counter in the session key, which I missed before.  If I register a session certificate with a lower counter value than currently set, then the transaction is dropped.  If I register a session certificate with a higher counter value, then the transaction sets the session certificate for the appropriate future slot and epoch, with babe requiring some waiting time, and grandpa maybe not so much.  

We're happy if registering a new session certificate makes block production with babe impossible for a couple epochs, in which case only one current session key exists.

I want this workflow for validator opeators:  Joe our validator operator wishes to migrate validator hardware.  First, Joe brings up his new validator, which generates a new session keypair.  Second, Joe updates his session certificate by signing the new session public key with his controller key and posting the change session certificate transaction.  And this transaction finalizes establishing the exact epoch and slot when Joe's old session key becomes invalid and Joe's new session key becomes valid, for babe and grandpa respectively.  Third, Joe's old validator stops signing harmlessly once it discovers it does not control any valid session key.  Fourth, Joe's new validator begins running grandpa and then babe with its new session key.  

In this workflow, there are no two machines that ever know the same session private key, which makes equivocation impossible, and even permits restricting those keys to SGX enclaves.  We thus avoid all the needless slashing done by Cosmos so far.  ;)

core/finality-grandpa/src/lib.rs
```diff
@@ -189,6 +191,58 @@ type CommunicationOutH<Block, H> = grandpa::voter::CommunicationOut<
...
+	Client: ProvideRuntimeApi + Send + Sync,
+	Client::Api: GrandpaApi<Block>,
+{
+	fn authority_key(
+		&self,
+		at: &BlockId<Block>,
+	) -> Option<Box<dyn offchain::OffchainKey>> {
+		let authorities = self.client
+			.runtime_api()
+			.grandpa_authorities(at)
```
andresilva: This won't give us the active authority set because GRANDPA authority set changes are only enacted on finality. What we need to do is figure out what the highest finalized ancestor of `at` is (or `at` itself), and query the runtime state at that block.

core/cli/src/params.rs
```diff
@@ -421,9 +416,9 @@ pub struct RunCmd {
 	#[structopt(long = "force-authoring")]
 	pub force_authoring: bool,
 
-	/// Interactive password for validator key.
-	#[structopt(short = "i")]
-	pub interactive_password: bool,
+	/// Interactive password for offchain worker keys
+	#[structopt(long = "offchain-worker-password")]
+	pub offchain_worker_password: bool,
```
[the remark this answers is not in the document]
dvc94ch: This is not the actual password, it's a flag

core/cli/src/params.rs
```diff
@@ -305,24 +305,19 @@ pub struct ExecutionStrategies {
...
-	/// Specify custom keystore path
-	#[structopt(long = "keystore-path", value_name = "PATH", parse(from_os_str))]
-	pub keystore_path: Option<PathBuf>,
-
 	/// Specify additional key seed
 	#[structopt(long = "key", value_name = "STRING")]
 	pub key: Option<String>,
 
-	/// Enable validator mode
+	/// Shortcut for `--aura` and `--grandpa-voter`.
```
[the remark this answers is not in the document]
dvc94ch: These flags should me moved to the node and not be in core

core/consensus/babe/src/lib.rs
```diff
@@ -772,11 +773,21 @@ fn claim_slot(
...
-	let public = &key.public();
-	let index = authorities.iter().position(|s| s == public)?;
+) -> Option<((VRFInOut, VRFProof, VRFProofBatchable), usize, AuthorityPair)> {
+	let mut keys = authorities.iter()
+		.enumerate()
+		.filter_map(|(index, public)| {
+			session_store.get_key(public).map(|key| (index, key))
+		})
+		.collect::<Vec<_>>();
+	let (index, key) = if keys.len() == 1 {
```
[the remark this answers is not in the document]
dvc94ch: @rphmeier said that can be handled in a follow up, maybe by someone more familiar with babe...

rphmeier:
@burdges 

> In any slot, there should only be one babe or grandpa key from which the network accepts signatures.

BABE slots can definitely be won by more than one key, just due to the probabilistic nature of the VRF. How likely this is depends on parameterization. I should have clarified: 

"Do the vrf-proof on all keys in the current validator set _that you control_ and if there's only one do we author on the slot with that key."

This is literally the same as what Gavin described above and is orthogonal to the migration scheme you mentioned.

PR #3041, opened 2019-07-05
Technical Committee

gavofyork:
Split up council crate into two crates: one from the `council::seats` module which becomes the `elections` crate and from from the collective-decision-making logic `council::motions` which becomes the `collective` crate. The latter becomes instanceable in order that we can have a secondary collective body "Technical Committee", in addition to the council.

A facility is introduced to `democracy` module in order to facilitate the Technical Committee in creating immediately-scheduled referenda; this facility is configured into the substrate-node's runtime.

This PR also introduces `insert_ref` and `put_ref` into storage items in order to allow equivalent reference-style types to be placed without copy (e.g. a storage item of `Vec<AccountId>` can now be written from a `&[AccountId]`).


pepyakin:
The offending version of wasmi is yanked. Should build fine now.

Issue #3001, opened 2019-07-02
Minimize Macro Magic


shawntabrizi:
I wanted to formally share some feedback I have seen regarding new developers building on Substrate, and some of the ways we can improve the experience by reducing the "magic" done by our Substrate macros.

I think it obvious that the Substrate macros are both a blessing and a curse. It helps users write clean and concise code, and reduces the amount of boilerplate code needed to get things working. However, it is also the place where users get the most complicated errors to solve, where it is hardest to gain visibility about what is happening in the background, and where it is hard to get help to resolve issues without talking to our team directly (versus talking to the Rust community or Rust references for example).

I suggest that we update some of the "features" in our various Substrate macros to increase consistancy and clarity on certain aspects, while having a minimal impact on the amount of code written.

Here are a list of "features" which I think should be removed, plus a small argument as to why I think this:

* [x] `decl_module`: Omitting `origin` as the first parameter makes a function require root origin.

    * This feature allows a user to omit `origin` as the first parameter of a dispatchable function. When this happens, the macro automatically adds origin back as the first parameter,  and also adds `ensure_root(origin)?` as the first line of the function. To begin, this decreases clarity of the "rules" when writing dispatchable functions. It is very easy to say "all dispatchable functions require origin as the first parameter to determine where the call is coming from, and the first check made in that function is to verify the origin is as expected." However, with this "feature", users who look at the Substrate code base will notice that we do _not_ include `origin` with all of our functions, and it is not clear at all in the program that the function is "privileged" in some way. Furthermore, in terms of auditing high privileged calls, our current syntax makes it impossible to do things like search for `ensure_root` to get a list of all the function which require ROOT origin. By removing this feature, we introduce at most ~25 characters to each privileged function, but increase the clarity of these functions and our rules immensely.

* [ ] `decl_module`: Omitting when functions return `Result` and automatically adding `Ok(())` at the end.

    * This feature allows a user to specify a dispatchable function to not return a `Result` and to not have `Ok(())` returned at the end. When the macro detects this, both of these items are added back to the function automatically. Again, this causes confusion to new users who are told that all dispatchable functions should return a `Result`, and then they look at our code and see that this is not the case. It will increase the number of characters by ~15 in functions which use the feature, but again drive clarity and simplicity where the alternative really is not needed.

* [ ] `construct_runtime`: the use of "no types" or `default` when defining your modules.

[...]

PR #3040, opened 2019-07-05
Fix ui tests (E0308 error)

DarkEld3r:

PR #3045, opened 2019-07-06
Make priviledged functions explicity use `origin`

shawntabrizi:
The `decl_module!` macro supports the declaration of privileged functions without an `origin` parameter. When `origin` is omitted, the `decl_module!` macro would add origin as the first parameter of a dispatchable function, and also include `ensure_root(origin)?;` as the first line of the function.

This PR removes this functionality to enhance clarity around the rules and behaviors of the `decl_module!` macro.

Changes:
* When `decl_module!` macro detects that the first parameter of a dispatchable function is _not_ `origin`, it returns the compile time error:

```
compile_error!(
	"Implicit conversion to privileged function has been removed. \
	First parameter of dispatch should be marked `origin`. \
	For root-matching dispatch, also add `ensure_root(origin)?`."
);
```

* The following modules have had their privileged functions updated to use `origin` and `ensure_root`:
    * Balances
    * Contracts
    * Council
    * Democracy
    * Example
    * Staking
    * System

* Tests have been updated in various modules
    * Calls to privileged functions in tests now use `Origin::ROOT` as the first input parameter
    * `srml-support` tests have been updated to include the missing `origin`

* Documentation has been updated
   * `srml-support` documentation for `decl_module` has been updated to reflect latest changes
   * `srml-example` documentation is updated to reflect the need for `origin` in a privileged function
   * Other error messages hinting at creating privileged functions without `origin` has been removed

Addresses one of the points in #3001


shawntabrizi:
I am not sure why the current test is failing, it seems to work on my local machine.

shawntabrizi:
Ah, error is not just me: https://github.com/paritytech/substrate/pull/3040

srml/council/src/seats.rs
```diff
@@ -487,15 +487,17 @@ decl_module! {
 		/// Set the desired member count; if lower than the current count, then seats will not be up
 		/// election when they expire. If more, then a new vote will be started if one is not
 		/// already in progress.
-		fn set_desired_seats(#[compact] count: u32) {
+		fn set_desired_seats(origin, #[compact] count: u32) {
```
thiolliere: one can still write `set_desired_seats(root, #[compact] ...`
the generated function won't have the first argument then.

Then the test call line 2325 doesn't have to be changed.

But my comment is more about this mecanism of writing origin as root is not defined in the doc, is it ? if no it might have been introduced for priviledge function actually. is something to be thought here ?

bkchr:
Did you speak about this with @gavofyork?

shawntabrizi:
> Did you speak about this with @gavofyork?

I got him to reply to the main issue. Seems that this change he is okay with :)

bkchr:
Can we go with what @thiolliere proposed and use `root` as first parameter for the functions? Or is this still to much "magic"?

shawntabrizi:
@bkchr my opinion is we should be driving for consistency, and thus it does not make sense to me to use `root`. I have removed those parts of the macro in the last commit.

If you feel strongly against these decisions, please do make that opinion known :)

srml/council/src/seats.rs
shawntabrizi: I have since removed this functionality as I believe it goes against the whole point of this PR.

bkchr:
I'm not happy with having to call this functions with `Origin::Root`, but it is probably the best we can do currently.

PR #3047, opened 2019-07-07
Add ExtCurrentBlockNumber in SRML::Contract

satellitex:
# What does it do?
We want to call current block number from contracts. So, I implement `ext_current_block` methods in srml::contract.
`ext_current_block` means that gets base chains current block number.


joepetrowski:
Is the block number the appropriate information to get?

Is the "current" block number:

- The latest finalized block?
- The current chain tip (from whose view)?
- The block that the transaction will execute in? This does not seem possible since you don't know which block a transaction will be included in. That's more to do with transaction _queue_ logic than the transaction logic.

Normally, if you want to reference a block, you should do so by its hash, as that is absolutely tied to a particular block, while block height could reference multiple blocks in the case of not-yet-finalized (or reverted) chains. The header would allow you to trace back the block hash to genesis and derive the block height if you need it, but I don't know that it should be included in the contract logic.

I don't have much experience in the Contracts module so maybe I am off base here, just what came to mind.

satellitex:
@joepetrowski Thanks for answering.
"current" block number means the latest finalized block.

And, I think that contract can get the latest block finalized time by `ext_now` method so as well I can get the latest block.

srml/contracts/src/wasm/mod.rs
```diff
@@ -1342,4 +1344,70 @@ mod tests {
 			Err("during execution"),
 		);
 	}
+
+	/// calls `ext_current_block`, loads the current block number from the scratch buffer and
+	/// compares it with the constant &121.
```
pepyakin:
```suggestion
	/// compares it with the constant 121.
```

srml/contracts/src/wasm/mod.rs
```diff
@@ -1342,4 +1344,70 @@ mod tests {
...
+		(block $ok
+			(br_if $ok
+				(get_local 0)
+			)
+			(unreachable)
+		)
+	)
+
+	(func (export "call")
+		;; fill the scratch buffer with the self address.
```
pepyakin: This comment seems to be irrelevant.

srml/contracts/src/wasm/mod.rs
```diff
@@ -1342,4 +1344,70 @@ mod tests {
...
+	fn current_block() {
+		let mut mock_ext = MockExt::default();
+		execute(
+			CODE_CURRENT_BLOCK,
+			&[],
+			&mut Vec::new(),
+			&mut mock_ext,
+			&mut GasMeter::with_limit(50_000, 1),
+		)
+			.unwrap();
```
pepyakin:
```suggestion
		.unwrap();
```

pepyakin:
I am not sure how would you implement the latest finalized block since it depends on the PoV of a node in question AFAIU so it might be problematic to allow access to this info in substrate.

However, the getting the block number in which the transaction executes is actually possible. In fact Ethereum's `BLOCKNUMBER` instruction does exactly that. 

I can see that the PR uses `System::block_number` which does return the current block number in which the transaction executes in and that should work.

satellitex:
@pepyakin Thanks for answering!
I understand you said and that I mistook the meaning of "finalized".
I modify soon.

srml/contracts/src/wasm/mod.rs
satellitex: It was fixed.

srml/contracts/src/exec.rs
```diff
@@ -120,6 +120,9 @@ pub trait Ext {
 
 	/// Rent allowance of the contract
 	fn rent_allowance(&self) -> BalanceOf<Self::T>;
+
+	// Returns a reference to the current block number.
+	fn current_block(&self) -> &<Self::T as system::Trait>::BlockNumber;
```
andresilva: nitpick but why not call this `block_number`?

srml/contracts/src/exec.rs
```diff
@@ -673,6 +679,8 @@ where
 		self.ctx.overlay.get_rent_allowance(&self.ctx.self_account)
 			.unwrap_or(<BalanceOf<T>>::max_value()) // Must never be triggered actually
 	}
+
+	fn current_block(&self) -> &T::BlockNumber { &self.current_block }
```
andresilva:
```suggestion
	fn current_block(&self) -> T::BlockNumber { self.current_block }
```

andresilva: `T::BlockNumber` is `Copy` so this should work and there's probably no need to return a reference.

srml/contracts/src/exec.rs
satellitex: Does this question mean about naming of function?

andresilva: Yep, I'm suggesting replacing this with `fn block_number(&self) -> BlockNumberOf<Self::T>;`

satellitex:
@andresilva Thanks for approving!!

I think that "current" block is better naming because it is called by contracts to know execution time block number. (the execution time is "current" for the contract.)
if you have better naming, I will follow your opinion.

pepyakin:
I am not sure if I agree with your logic. Note that you call the runtime function called block_number which is also the number of the current block.

I’d vote for consistency with the SRML

satellitex:
@pepyakin Ok! I agree. I change method names now.

srml/contracts/src/exec.rs
```diff
@@ -120,6 +121,9 @@ pub trait Ext {
 
 	/// Rent allowance of the contract
 	fn rent_allowance(&self) -> BalanceOf<Self::T>;
+
+	// Returns the current block number.
```
pepyakin:
```suggestion
	/// Returns the current block number.
```

satellitex: @pepyakin Thank you for suggesting! It fixed.

pepyakin:
Thanks @satellitex !

PR #3048, opened 2019-07-07
srml-staking: Force minimum bond amount + proper cleanup. 

### LINKED ISSUE #2805: Inconsistent dust removal between bond, bondExtra and unbond
`staking.unbond` checks for minimumBalance but not `bonExtra` nor `bond`.

This means that I can have only 1 unit bonded if I call `bond` or `bondExtra` with an original amount of 0 unit bonded.
However I will not end up with a 1 unit bonded if I call `unbond(9 units)` with an original amount of 10 units because the `unbond` function will automatically add the remaining 1 Unit.

This makes any interface with these functions feel inconsistent  

relevant part: https://github.com/paritytech/substrate/blob/bfabed51cf2c4ead5aca08634ddcaf4b3ba837a9/srml/staking/src/lib.rs#L644-L648

it was noted  here: https://github.com/paritytech/substrate/pull/2694/files#diff-b5a024b4f4ae96b0bfc0681acae56709R567

Kianenigma:
Fixes #2805.

- Don't allow `bond` with values less than `existential_deposit`. I think this must exist because `unbond()` checks the minimum bonded amount while `bond` itself doesn't and one can essentially get around it. See the issue above.
- Once an account `unbond`s everything (or almost everything when the remaining part falls below `existential_deposit`), and it calls `withdraw_unbonded()` for the last time (no `unlocking_chunks` left), remove the account entirely from the staking system. This might be debatable as there is a tradeoff between keeping old storage and doing some computation + checks to remove them. We might prefer to actually keep the storage items forever and simply not care about them?

CC @Tbaut.


srml/staking/src/lib.rs
```diff
@@ -617,6 +617,8 @@ decl_module! {
 		/// Take the origin account as a stash and lock up `value` of its balance. `controller` will
 		/// be the  account that controls it.
 		///
+		/// `value` must be more than the `existential_deposit` defined in balances module.
```
shawntabrizi:
```suggestion
		/// `value` must be more than the `existential_deposit` defined in the Balances module.
```
nit

shawntabrizi:
@Kianenigma you need to bump the `spec_version` right? Since this change will break consensus.

Kianenigma:
I believe test failure is due to https://github.com/paritytech/substrate/pull/3040 + remarks on the second point (cleanup) are still appreciated.

srml/staking/src/lib.rs
```diff
@@ -648,6 +650,11 @@ decl_module! {
 				return Err("controller already paired")
 			}
 
+			// reject a bond which is considered to be _dust_.
+			if  value < T::Currency::minimum_balance() {
```
gavofyork: double-space
```suggestion
			if value < T::Currency::minimum_balance() {
```

srml/staking/src/lib.rs
```diff
@@ -745,17 +754,30 @@ decl_module! {
...
 		/// - Contains a limited number of reads, yet the size of which could be large based on `ledger`.
 		/// - Writes are limited to the `origin` account key.
 		/// # </weight>
 		fn withdraw_unbonded(origin) {
 			let controller = ensure_signed(origin)?;
 			let ledger = Self::ledger(&controller).ok_or("not a controller")?;
 			let ledger = ledger.consolidate_unlocked(Self::current_era());
-			Self::update_ledger(&controller, &ledger);
+
+			if ledger.unlocking.is_empty() && ledger.active.is_zero() {
```
gavofyork: Not sure what this is for. If the (active) balance ever falls below ED, then these consequences will happen automatically.

Kianenigma: I think they are separate: an account can have 100 units and bond 50 of them while ED is 10, meaning that regardless of how/when they unbond, their total balance will remain above ED and won't be cleaned by the `OnFreeBalanceZero`. But, we still keep all of their relevant info in the staking module even if they `unbond` and `withdraw_unbonded` all of it out. This check essentially treats this situation.

Kianenigma:
```
/// NOTE: Two of the storage writes (`Self::bonded`, `Self::payee`) are _never_ cleaned unless
/// the `origin` falls below _existential deposit_ and gets removed as dust.
```

I think forcing minimum ED for the `bond()` (almost) solves ^^ but what I added (cleaning after withdraw) is a further improvement on top of it to _not store the information of anyone who ever bonded and withdrawn_ __but has some balance in their account left__. if theu have unbonded and withdrawn, I think it is clear that those accounts might still live (for long) but have nothing more to do with the Staking system.

gavofyork:
need resolving.

gavofyork:
Icing until #3041 is in.

PR #3050, opened 2019-07-07
Sensible locking for staking and council.

### LINKED ISSUE #2950: Allow fee withdraw on locked balance.
srml-council (and potentially other use cases) want to lockup the entire balance of an account for security reasons. This should not prevent them from submitting other transactions, namely to undo the lock. 

> it should be locked except, perhaps, for paying transaction fees.
(that's a specific feature of locking - you can provide a bit mask over what kinds of withdrawals are allowed)

Kianenigma:
closes #2950


gavofyork:
@Kianenigma doesn't build? or just needs merge with master for the ui fix?

gavofyork:
Icing until #3041 is in.

gavofyork:
@Kianenigma will need resolving

Kianenigma:
Done

PR #3057, opened 2019-07-08
Make `RUSTFLAGS` configurable in `wasm-builder`

bkchr:


core/utils/wasm-builder/src/wasm_project.rs
```diff
@@ -265,9 +260,15 @@ fn is_release_build() -> bool {
 fn build_project(project: &Path) {
 	let manifest_path = project.join("Cargo.toml");
 	let mut build_cmd = crate::get_nightly_cargo();
+
+	let rustflags = format!(
+		"-C link-arg=--export-table {}",
```
pepyakin: I am not entirely sure, is this going to work if the `WASM_BUILD_RUSTFLAGS` has spaces in it?

bkchr: Yeah. You can have spaces in your env variables.

PR #3061, opened 2019-07-08
Update jsonrpc-* dependencies

ascjones:
Should also fix the annoying pubsub warning


ascjones:
Closing because just saw duplicate #3060

PR #3062, opened 2019-07-08
srml: system: prune block hash mapping

andresilva:
Currently the block number to block hash mapping in the system module grows unbound. This PR adds a module configuration that determines the maximum amount of mappings to keep. On finalize the oldest mapping is pruned. Default is to keep the last 250 blocks. The genesis hash is kept indefinitely.

### DIFF
--- a/srml/system/src/lib.rs
+++ b/srml/system/src/lib.rs
@@ -362,7 +362,11 @@ decl_storage! {
 		#[serde(with = "substrate_primitives::bytes")]
 		config(code): Vec<u8>;
 
-		build(|storage: &mut primitives::StorageOverlay, _: &mut primitives::ChildrenStorageOverlay, config: &GenesisConfig<T>| {
+		build(|
+			storage: &mut primitives::StorageOverlay,
+			_: &mut primitives::ChildrenStorageOverlay,
+			config: &GenesisConfig<T>,
+		| {
 			use parity_codec::Encode;
 
 			storage.insert(well_known_keys::CODE.to_vec(), config.code.clone());
--- a/srml/system/src/lib.rs
+++ b/srml/system/src/lib.rs
@@ -89,7 +89,6 @@ use srml_support::{
 };
 use safe_mix::TripletMix;
 use parity_codec::{Encode, Decode};
-use crate::{self as system};
 
 #[cfg(any(feature = "std", test))]
 use runtime_io::{twox_128, TestExternalities, Blake2Hasher};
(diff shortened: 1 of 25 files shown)


andresilva:
Should we special case the genesis block and always keep its hash in the mapping?

srml/system/src/lib.rs
```diff
@@ -317,6 +316,8 @@ decl_storage! {
 		AllExtrinsicsWeight: Option<u32>;
 		/// Map of block numbers to block hashes.
 		pub BlockHash get(block_hash) build(|_| vec![(T::BlockNumber::zero(), hash69())]): map T::BlockNumber => T::Hash;
+		/// Maximum number of block number to block hash mappings to keep (oldest pruned first).
+		BlockHashCount get(block_hash_count) config(): T::BlockNumber = 250.into();
```
bkchr: Sounds like something that should be `parameter_type!`.

srml/system/src/lib.rs
```diff
@@ -580,6 +585,12 @@ impl<T: Trait> Module<T> {
 			digest.push(item);
 		}
 
+		// move block hash pruning window by one block
+		let block_hash_count = <BlockHashCount<T>>::get();
+		if number > block_hash_count {
+			<BlockHash<T>>::remove(number - block_hash_count - One::one());
```
```suggestion
			<BlockHash<T>>::remove(number - block_hash_count);
```

srml/system/src/lib.rs
```diff
@@ -580,6 +585,12 @@ impl<T: Trait> Module<T> {
 			digest.push(item);
 		}
 
+		// move block hash pruning window by one block
+		let block_hash_count = <BlockHashCount<T>>::get();
+		if number > block_hash_count {
```
```suggestion
		if number >+ block_hash_count {
```

Sir, I'm unsure if I'm to tired or not, but I think it should be this way.

Given we have number `101` and `100` as block hash count. This means we already produced `102` blocks.

srml/system/src/lib.rs
```diff
@@ -906,4 +917,45 @@ mod tests {
...
+			// first 5 block hashes are pruned
+			for n in 0..5 {
+				assert_eq!(
+					System::block_hash(n),
+					H256::zero(),
+				);
+			}
+
+			// the remaining 10 are kept
+			for n in 5..15 {
```
```suggestion
			for n in 5..=15 {
```

srml/system/src/lib.rs
andresilva: I don't know what `parameter_type!` is, but won't this always be defined in `T::BlockNumber`?

srml/system/src/lib.rs
andresilva: If number is `102` then it means that we have just inserted the hash for its parent `101`. And so we will remove `102 - 100 - 1` i.e. block `1`. So our block hash mapping will have the range `[2, 101]` (100 blocks).

srml/system/src/lib.rs
andresilva: If we insert until block 15 it means we have only mapped hashes until block 14.

srml/system/src/lib.rs
bkchr: I should read the other source code before commenting :D

srml/system/src/lib.rs
bkchr: Just look into `runtime` and search for it. You require to always look into the storage for a value that probably never changes. Using `parameter_types!`, the value is hard coded and compiled into the runtime.

Issue #2181, opened 2019-04-02
Utility for consensus engines to dispatch transactions to the network


rphmeier:
Since cryptography used in consensus is often generic or in some cases (e.g. PoW) doesn't exist, it will tend to be different from the account cryptography used in a runtime.

We will need to amend the various consensus Runtime API traits to get methods such as 

```rust
trait SomeConsensusApi {
    fn construct_report_call(evidence: ReportSpecificEvidence) -> Vec<u8>; // corresponds to some unsigned extrinsic or `Call` type
}
```

and then consensus engines should have access to some kind of trait or context that has a function

```rust
// signs and dispatches an encoded call with some pre-configured key
// perhaps via a `--reporting-key` CLI.
fn sign_and_dispatch_encoded_call(Vec<u8>) -> Result<(), Error>;
```

The implementation of this `sign_and_dispatch_encoded_call` function will vary from runtime to runtime.

marcio-diaz:
Closing in favor of better approach, see https://github.com/paritytech/substrate/pull/2342#issuecomment-486176371

PR #3063, opened 2019-07-08
API for constructing transactions from Calls

marcio-diaz:
Initial proposal for an API for constructing transactions from Calls.
This will be used, for instance, for constructing misbehavior report transactions.

Context: #2181

### Example usage
```rust
let signing_payload = runtime_api().signing_payload(block_id, report_call, public);
let signature = Pair::from_public(public).sign(signing_payload);
let tx = runtime_api().build_transaction(block_id, signing_payload, signature);
transaction_pool.submit_one(&block_id, tx);
```

### Example definitions
Signing Payload
```rust
fn signing_payload(encoded_call: &[u8], account_id: &[u8]) -> Vec<u8> {
    let nonce = get_account_nonce(public);
    let call = Decode(encoded_call);
    let genesis_hash = ...; // From where?
    (nonce, call, Era::immortal(), genesis_hash).encode()
}
```
Build Transaction
```rust
fn build_transaction(signing_payload: &[u8], signature: AnySignature) -> Vec<u8> {
    let (nonce, call, era, ...) = Decode(signing_payload);
    let public = ...; // From where?
    UncheckedExtrinsic::new_signed(nonce, call, public, signature, era).encode()
}
```


core/client/src/transaction_builder/transaction_builder.rs
```diff
@@ -0,0 +1,31 @@
...
+//! The runtime api for building transactions.
+
+use runtime_primitives::AnySignature;
+use sr_api_macros::decl_runtime_apis;
+
+decl_runtime_apis! {
+	/// The `TransactionBuilder` trait that provides required functions
+	/// for building a transaction from a Call for the runtime.
+	pub trait TransactionBuilder {
+		/// Construct the payload.
```
andresilva: Could improve the docs a bit more? How is this supposed to be used?

core/client/src/transaction_builder/transaction_builder.rs
```diff
@@ -0,0 +1,31 @@
...
+		fn signing_payload(encoded_call: &[u8], account_id: &[u8]) -> Vec<u8>;
+		/// Build the transaction.
+		fn build_transaction(signing_payload: &[u8], signature: AnySignature) -> Vec<u8>;
+	}
+}
```
andresilva:
```suggestion
}
```

core/client/src/transaction_builder/transaction_builder.rs
marcio-diaz: Good question, honestly I don't have it clear. I updated the description according to the way I think it should be used but I feel we may be missing parameters in the functions.

core/client/src/transaction_builder/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Utility struct to build a transaction.
+
+#[cfg(feature = "std")]
+mod transaction_builder;
+#[cfg(feature = "std")]
+pub use transaction_builder::*;
```
rphmeier: could just put a `#![cfg(feature = "std")]` at the top of the file.

rphmeier: (but that's probably not the desired behavior -- don't we want the `TransactionBuilder` API to be exposed in `no_std`? )

core/client/src/transaction_builder/transaction_builder.rs
```diff
@@ -0,0 +1,31 @@
...
+use sr_api_macros::decl_runtime_apis;
+
+decl_runtime_apis! {
+	/// The `TransactionBuilder` trait that provides required functions
+	/// for building a transaction from a Call for the runtime.
+	pub trait TransactionBuilder {
+		/// Construct the payload.
+		fn signing_payload(encoded_call: &[u8], account_id: &[u8]) -> Vec<u8>;
+		/// Build the transaction.
+		fn build_transaction(signing_payload: &[u8], signature: AnySignature) -> Vec<u8>;
```
rphmeier: Please don't use `AnySignature` -- the name is really misleading. it should be a byte-slice.

rphmeier: to make the interface more general, we may want a `fn possible_crypto() -> Vec<KeyTypeId>`. 

This should fit in pretty nicely with the work @dvc94ch has done already. I can imagine the client code asking the runtime which crypto is supported, then selecting a key, and then submitting a report. for now the client code should just be checking that `possible_crypto().contains(SOME_KNOWN_KEY_TYPE)` and looking for a registered signing key of that type.

Kianenigma:
fyi: the signature payload is changing in #3102.

core/client/src/transaction_builder/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Utility struct to build a transaction.
+
+mod transaction_builder;
```
bkchr: Why is this in client and not somewhere else?

And why is the trait not declared in this file directly? I don't see any reason for a folder + an extra file.

marcio-diaz: Right, I fixed it.

I wasn't sure where to put it, maybe just under `/core/`?

bkchr:
Some general questions, why do we need two methods? Don't you normally just want create a signed transaction in one call?

Next question would be, do we really need this? If we take the current design, we probably need 3 calls into the runtime for creating the transaction. Can we not directly return a transaction from the misbehavior construction call? As you always need to call into the runtime to get the correct call, we could just return the signed transaction and be done.

marcio-diaz:
> Some general questions, why do we need two methods? Don't you normally just want create a signed transaction in one call?
> 
> Next question would be, do we really need this? If we take the current design, we probably need 3 calls into the runtime for creating the transaction. Can we not directly return a transaction from the misbehavior construction call? As you always need to call into the runtime to get the correct call, we could just return the signed transaction and be done.

I think the problem is that we can't sign a transaction in `no-std` right now (although it should be possible, right? but can we assume it?). Then you need to make it as general as possible as @rphmeier commented above.

core/client/src/transaction_builder.rs
```diff
@@ -0,0 +1,38 @@
...
+decl_runtime_apis! {
+	/// The `TransactionBuilder` trait that provides required functions
+	/// for building a transaction from a Call for the runtime.
+	pub trait TransactionBuilder {
+		/// Construct the payload.
+		fn signing_payload(encoded_call: Vec<u8>, encoded_account_id: Vec<u8>) -> Vec<u8>;
+		/// Build the transaction.
+		fn build_transaction(
+			signing_payload: Vec<u8>,
+			encoded_account_id: Vec<u8>, // TODO: rename to extra_payload?
```
DemiMarie-parity: TODOs need to have corresponding issues.

core/client/src/transaction_builder.rs
```diff
@@ -0,0 +1,38 @@
...
+			signature: Vec<u8>,
+		) -> Vec<u8>;
+		/// Get list of supported crypto types.
+		fn possible_crypto() -> Vec<KeyTypeId>;
+	}
+}
```
DemiMarie-parity: Nit: missing newline at EOF.

marcio-diaz:
> How do we fill `encoded_call` parameter? Isn't that what we want to abstract over? Otherwise the consensus engine still needs to find a way to figure out which method to call (which is specific to the runtime). If you note on #2181, @rphmeier mentioned that consensus runtime APIs should get a method like `construct_report_call` which would return the encoded call.

Yes, `encoded_call` is the output of the `construct_report_call` for each consensus API, is it ok?

rphmeier:
> Don't you normally just want create a signed transaction in one call?

Yeah, that would be nice. But the runtime controls transaction format, not consensus code. And the runtime shouldn't have users' private keys in storage...

bkchr:
> > Don't you normally just want create a signed transaction in one call?
> 
> Yeah, that would be nice. But the runtime controls transaction & signing format, not consensus code. I don't see an obvious way to get the runtime to sign transactions either.

By "one call", I mean on runtime call. I know that we need to call into the runtime anway, to at least get the correct `Call`. If we are already in the runtime, we can also create the transaction and sign it. Everything in one call. This would not even require an extra api for transaction building, as each consensus just can adds the function to its own runtime api interface.

rphmeier:
> If we are already in the runtime, we can also create the transaction and sign it. Everything in one call.

Why would consensus keys be available within the runtime?

bkchr:
> Why would consensus keys be available within the runtime? That would require some offchain magic.

With this pr: https://github.com/paritytech/substrate/pull/3213 We will be able to sign from the runtime.

marcio-diaz:
> > Why would consensus keys be available within the runtime? That would require some offchain magic.
> 
> With this pr: #3213 We will be able to sign from the runtime.

Ok, then this PR is not needed and we can close, right? 
I'll prepare the PR for adding the `construct_calls` to the consensus APIs.

gavofyork:
Closing for now. Can reopen if turns out to be needed.

Issue #2777, opened 2019-06-04
Disable GRANDPA finality gadget from runtime


nczhu:
_Wanted to put up an issue for this as it’s important for Kusama, pls CRUD as needed._

@andresilva

PR #3068, opened 2019-07-08
srml: grandpa: Pause/resume current authority set

andresilva:
Runtime part of #2777.

Add methods to the GRANDPA srml module to signal pause/resume of the current authority set. They work similarly to authority set changes, they are enacted with a (potentially 0) delay and a log is emitted when these are signaled (which will be used by the node to track state changes). We keep the state of the authority set in an enum `Live, PendingPause, Paused, PendingResume` and validate correct state transitions.

The node side handling of these signals will be implemented in a follow-up PR. The node will enact `Pause` changes on finality and `Resume` changes on block depth. When voting we will vote for the lowest change possible (either resume/pause or authority set change). It is also possible to signal authority set changes while the authority set is paused, these should be applied in-order when the current authority set is resumed. For forced changes we currently already pause the authority set, so on the node side we will need to distinguish whether the pause was explicit or caused by a force change. If it was explicit after a forced change is applied we don't start the voter automatically. It must be resumed explicitly afterwards.


andresilva:
I wrote this before the method to disable the authority set of a given session was added and then just rebased on top of it. I'm not sure how/if these will interact.

core/finality-grandpa/primitives/src/lib.rs
```diff
@@ -93,22 +93,44 @@ pub enum ConsensusLog<N: Codec> {
 	/// Note that the authority with given index is disabled until the next change.
 	#[codec(index = "3")]
 	OnDisabled(AuthorityIndex),
+	/// A signal to pause the current authority set after the given delay.
+	/// After finalizing the block at _delay_ the authorities should stop voting.
+	Pause(N),
```
gavofyork:
```suggestion
	#[codec(index = "4")]
	Pause(N),
```

core/finality-grandpa/primitives/src/lib.rs
```diff
@@ -93,22 +93,44 @@ pub enum ConsensusLog<N: Codec> {
...
+	/// A signal to resume the current authority set after the given delay.
+	/// After authoring the block at _delay_ the authorities should resume voting.
+	Resume(N),
```
gavofyork:
```suggestion
	#[codec(index = "5")]
	Resume(N),
```

core/finality-grandpa/primitives/src/lib.rs
gavofyork: Why is this needed as a digest item? Surely digestitems are just for the benefit of light clients which don't vote anyway. An event is sufficient, no?

PR #3071, opened 2019-07-09
Use optimized append and len storage methods in SRML.

### LINKED ISSUE #2794: Use storage append when possible.
When making multiple RPC calls to `democracy::propose()` the execution appears to slow down exponentially.

eg:

```javascript
api.tx.democracy
    .propose(api.tx.balances.setBalance(random_address, 5000, 123),500)
    .signAndSend(sender, { nonce });
```

By the time we have sent approximately ~4k tx, transaction processing is stalling, maybe 2-4tps.

kianenigma:
closes #2794 

As shown in ^^, the getter function of storage items that store potentially long vectors can 1- easily become the bottleneck of the execution time 2- become exponentially worse over time.

The runtime storage already had a `::append()` for types that actually have a length (`Vec<_>`, `map T => Vec<_>`). Now, all types that can be appended to also have a method to read the length, without decoding the entire value from storage.

- replace `Storage::Foo::mutate(|x| x.push(y))` with `::append()`.
- replace occurrences where we need the length but not the value itself with `::len()`.

TODO:
- [x] release the new `parity_codec` and remove `parity_codec_edge`.

Thoughts:
- I had to add some `expect`s to handle results. I think append can remain as it is (return result) but now I think maybe `len()` should simply always return usize and collapse to zero in case of error. I think this is nice since the storage already returns the default value of a vec if it does not exists. length should behave the same and the length of the default vec is 0.

cc @mattrutherford


Kianenigma:
(I see that the edge branch in codec is gone and CI can't process anything. assume tests are passing.)

srml/council/src/motions.rs
```diff
@@ -139,8 +139,8 @@ decl_module! {
 				Self::deposit_event(RawEvent::Executed(proposal_hash, ok));
 			} else {
 				let index = Self::proposal_count();
+				<Proposals<T>>::append(&[proposal_hash]).expect("vec will always have append; qed");
```
bkchr: The error returned is not about having `append` or not, it is more about "could I decode the value".

srml/council/src/seats.rs
```diff
@@ -680,15 +680,15 @@ impl<T: Trait> Module<T> {
 	///
 	/// The voter index must be provided as explained in [`voter_at`] function.
 	fn do_set_approvals(who: T::AccountId, votes: Vec<bool>, index: VoteIndex, hint: SetIndex) -> Result {
-		let candidates = Self::candidates();
+		let candidates_len = <Self as Store>::Candidates::len().expect("vec always has a len; qed");
```
bkchr: As with the other decoding errors, we should probably hide it behind the scenes and just panic there.

srml/support/src/storage/hashed/generator.rs
```diff
@@ -205,6 +205,20 @@ pub trait StorageValue<T: codec::Codec> {
...
+
+	/// Read the length of the value in a fast way, without decoding the entire value.
+	///
+	/// `T` is required to implement `Codec::DecodeLength`.
+	fn len<S: HashedStorage<Twox128>>(storage: &mut S)
+		-> Result<usize, &'static str> where T: parity_codec_edge::DecodeLength
+	{
+		if let Some(k) = storage.get_raw(Self::key()) {
+			<T as parity_codec_edge::DecodeLength>::len(&k)
+				.ok_or_else(|| "Could not read len (value)")
```
bkchr: Maybe we should panic here directly.

srml/council/src/seats.rs
gavofyork: decoding should not panic in the runtime; it should return `None` behind the scenes, which gets coerced into `Default::default()` for items that are not `Option`s.

srml/council/src/seats.rs
```diff
@@ -721,10 +721,10 @@ impl<T: Trait> Module<T> {
 				CellStatus::Head | CellStatus::Occupied => {
 					// Either occupied or out-of-range.
 					let next = Self::next_nonfull_voter_set();
-					let mut set = Self::voters(next);
+					let set_len = <Voters<T>>::len(next).expect("map to vec always has len; qed");
```
gavofyork: no panickers in runtime, please.

srml/council/src/motions.rs
gavofyork: indeed. there should be no panicking in the runtime. in the case of a corrupted storage, try to bail out of the call.

srml/democracy/src/lib.rs
```diff
@@ -757,7 +756,7 @@ impl<T: Trait> Module<T> {
 	fn do_vote(who: T::AccountId, ref_index: ReferendumIndex, vote: Vote) -> Result {
 		ensure!(Self::is_active_referendum(ref_index), "vote given for invalid referendum.");
 		if !<VoteOf<T>>::exists(&(ref_index, who.clone())) {
-			<VotersFor<T>>::mutate(ref_index, |voters| voters.push(who.clone()));
+			<VotersFor<T>>::append(ref_index, &[who.clone()]).expect("vec can always append; qed");
```
gavofyork: no panickers in runtime!

srml/democracy/src/lib.rs
```diff
@@ -368,9 +368,8 @@ decl_module! {
 			PublicPropCount::put(index + 1);
 			<DepositOf<T>>::insert(index, (value, vec![who.clone()]));
 
-			let mut props = Self::public_props();
-			props.push((index, (*proposal).clone(), who));
-			<PublicProps<T>>::put(props);
+			<PublicProps<T>>::append(&[(index, (*proposal).clone(), who)])
+				.expect("vec can always append; qed");
```
gavofyork: no panickers in runtime!

srml/support/src/storage/hashed/generator.rs
gavofyork: Absolutely not!

No user input (for these purposes, storage counts as user input) should ever lead to panicking.

Kianenigma:
@bkchr so I started looking into this again and to be able to really know if a value is an option or not:

> decoding should not panic in the runtime; __it should return None behind the scenes, which gets coerced into Default::default() for items that are not Options__.

I have to add an implementation in impls.rs:

```rust
impl<#impl_trait> #scrate::storage::hashed::generator::StorageValue<#typ>
		for #name<#trait_and_instance> #where_clause
	{
                 ...
```

which can be problematic since: 
- unlike the trait the impl is trait is type-aware and it will fail for storage types that don't have DecodeLength.
- I tried filtering by type and only implementing this function for types that for example have the token `Vec`  in their value but that is
  - Dirty
  - Won't work (we have some nested Vecs somewhere which can't fulfill the trait requirement)

 
Furthermore, the current implementation has this issue (aside from panics): any implementation will have the flaw that: if a Vec in storage has a default value defined (say, `vec![1]`), anytime that it is uninitialized, `::get()` will return `[1]` but `::len()` will return `0` or `None`, which is wrong. 

I am not sure if even with the extra flexibility of implementing it in `impls.rs` we can fix the above.

-- 

How about letting the user define if they are this implementation or not? Similar to how `config()` enables genesis stuff, a `len()` can add the implementation of `::len()` for a type.

Kianenigma:
I've pushed som test code to demonstrate:
https://github.com/paritytech/substrate/pull/3071/files#diff-3b34e4ffe10982e59ac0cae46c46be9eR854

srml/democracy/src/lib.rs
```diff
@@ -803,7 +802,7 @@ impl<T: Trait> Module<T> {
 	fn do_vote(who: T::AccountId, ref_index: ReferendumIndex, vote: Vote) -> Result {
 		ensure!(Self::is_active_referendum(ref_index), "vote given for invalid referendum.");
 		if !<VoteOf<T>>::exists(&(ref_index, who.clone())) {
-			<VotersFor<T>>::mutate(ref_index, |voters| voters.push(who.clone()));
+			safe_append_map(VotersFor(PhantomData::<(T,)>), ref_index, who.clone());
```
kianenigma: @bkchr this was my question. I assumed something like below would work: 
```suggestion
			safe_append_map(VotersFor, ref_index, who.clone());
```

srml/support/procedural/src/storage/mod.rs
```diff
@@ -132,6 +134,12 @@ struct DeclStorageConfig {
 	pub expr: ext::Parens<Option<syn::Ident>>,
 }
 
+#[derive(Parse, ToTokens, Debug)]
+struct DeclStorageDecodeLength {
+	pub len_keyword: keyword::decode_len,
+	pub expr: ext::Parens<Option<syn::Ident>>,
```
kianenigma: This works but is wrong: I want to say an always empty pair of parenthesis.

srml/support/src/storage/hashed/generator.rs
```diff
@@ -212,6 +212,13 @@ pub trait StorageValue<T: codec::Codec> {
 		storage.put_raw(Self::key(), &new_val);
 		Ok(())
 	}
+
+	/// Read the length of the value in a fast way, without decoding the entire value.
+	///
+	/// `T` is required to implement `Codec::DecodeLength`.
+	fn decode_len<S: HashedStorage<Twox128>>(_: &mut S)
```
kianenigma: changed to name to emphasis even more that this is no ordinary length function.

kianenigma:
> How about letting the user define if they are this implementation or not? Similar to how config() enables genesis stuff, a len() can add the implementation of ::len() for a type.

Too late.

bkchr:
Hmmmmmm, so let's talk through all this:
1. I would not extend `decl_storage!` for this. I don't really see the benefits.
2. `DecodeLength`, I would just return the `Error` and okay. The programmer will be aware that this can fail and can write `unwrap_or(0)` on it's own. Then, he also is aware that `get()` could return the default value.
3. `safe_append`. Could be added, but like `append`, just add it to the traits. It would look like:
```
fn append(values) where V: EncodeAppend {
    Self::append(values).unwrap_or_else(|_| Self::put(values));
}
```

But the function needs a good documentation saying exactly what is happening!

kianenigma:
> Hmmmmmm, so let's talk through all this:
> 
> 1. I would not extend `decl_storage!` for this. I don't really see the benefits.
> 2. `DecodeLength`, I would just return the `Error` and okay. The programmer will be aware that this can fail and can write `unwrap_or(0)` on it's own. Then, he also is aware that `get()` could return the default value.
> 3. `safe_append`. Could be added, but like `append`, just add it to the traits. It would look like:
> 
> ```
> fn append(values) where V: EncodeAppend {
>     Self::append(values).unwrap_or_else(|_| Self::put(values));
> }
> ```
> 
> But the function needs a good documentation saying exactly what is happening!

I agree to a high degree. Frankly, I knew that adding it to decl_storage was a bit of an over-engineering but just wanted to give it a try anyhow.

srml/democracy/src/lib.rs
```diff
@@ -982,6 +982,24 @@ impl<T: Trait> OnFreeBalanceZero<T::AccountId> for Module<T> {
...
+}
+
+fn safe_append_map<K, V, S>(_: S, key: K, value: V)
+where
+	K: Encode + Decode + Copy,
+	V: Encode + Decode + Clone,
+	S: AppendableStorageMap<K, Vec<V>, Query=Vec<V>>,
+{
+	S::append(key, &[value.clone()])
+			.unwrap_or_else(|_| S::mutate(key, |v| v.push(value)));
```
thiolliere: I'm not sure I see the interest of this safe method here; append only fails if value is wrongly encoded.

Also it seems append function returns an error but get function actually panics if value is wrongly encoded. I don't know what we should prefer but this *safe* is actually just calling the other behavior which would panic instead of returning an error.

srml/support/src/storage/mod.rs
```diff
@@ -267,6 +354,27 @@ impl<K: Codec, V: Codec, U> AppendableStorageMap<K, V> for U
...
+/// A storage map with a decodable length.
+pub trait DecodeLengthStorageMap<K: Codec, V: Codec>: StorageMap<K, V> {
+	/// Read the length of the value in a fast way, without decoding the entire value.
+	///
+	/// `T` is required to implement `Codec::DecodeLength`.
+	///
+	/// Has the same logic as [`StorageValue`](trait.StorageValue.html).
+	fn decode_len<KeyArg: Borrow<K>>(key: KeyArg) -> Option<usize>
+		where V: codec::DecodeLength;
+}
```
thiolliere: what is the reason for it to be implemented on a where function on StorageValue but on an additional trait on StorageMap ? (maybe you just got inspiration from append)

thiolliere: hmm it is probably because of linked_map which has a different internal representation. ok make sense

srml/democracy/src/lib.rs
kianenigma: Your point holds, and I am not fully aware of when/how storage actually panics tbh. I've added a new version which is more or less the same but abstracted in the storage layer. But I personally documented it in a way to encourage the users to not use it and `unwrap_or_blah_blah()` themselves. 

But given your note: if `::get` and hence `::mutate()` can _only_ panic in the above code, for the same reason that the original `::append()` failed (corrupted encoding), then this whole `_safe()` is moot. Whenever `::append()` fails, the following `::mutate()` (or `::set()` in the new version) will also fails. Is this true @bkchr ?

srml/support/src/storage/mod.rs
kianenigma: I think because specialized `Map` behaviors (appendable, enumerable etc.) are already in different traits while for the `Value` we have just one and we prefer to keep it that way? tbh I saw that `append()` was added to the the main `StorageValue` trait and just did the same for `decode_len()`.

srml/democracy/src/lib.rs
kianenigma: if the above is correct then it should be called `fn append_or_set(items)` to not give the misunderstanding that it is actually _safer_. It seems to be not.

thiolliere: at the same time append already does set if empty.

srml/support/src/storage/mod.rs
thiolliere: Yes because append doesn't work for linked_map even for linked_map u32 => Vec<u32>.
because at the key is stored not only the value but also an additional structure that links elements

srml/democracy/src/lib.rs
bkchr: We are currently panicking, but this is not correct. We need to remove this panic. Otherwise you could end up with a corrupt state, when you have something that changed and you forgot to upgrade it in `on_initialize`.
As I said in my comment somewhere here, you should use `Self::put` when `append` fails. 

Hmm, but you probably can not use `put` as you would need to construct the actual type out of the items...

kianenigma: Well, I do `::put()` now (already pushed) with the extra where clause of `T: From<&'a[I]>` (e.g. `Vec<T> from &[T]`). Since I think 99% of the use cases of this method is for arrays and vec and they have this `From<&[_]>`  impl out of the box, it should be fine. Otherwise, the use has to implement this somewhere to be able to call `safe_append()`

srml/support/src/storage/storage_items.rs
```diff
@@ -832,6 +839,53 @@ mod test_map_vec_append {
...
+	fn len_works_for_uninitialized() {
+		with_externalities(&mut TestExternalities::default(), || {
+			assert_eq!(JustVec::get(), vec![]);
+			assert_noop!(JustVec::decode_len(), "could not find item to decode length");
+
+			assert_eq!(OptionVec::get(), None);
+			assert_noop!(OptionVec::decode_len(), "could not find item to decode length");
+
+			// NOTE: this must be WARNED and documented to the upmost extent.
+			assert_eq!(JustVecWithDefault::get(), vec![6u32, 9]);
```
bkchr: I just got an idea for this. We could introduce a marker trait. `NoDefault` or a similar name.

This trait is implemented by `decl_storage!` if there is no default given to `decl_storage!`. Then we can just add this trait bound for the `decode_len()` function.

thiolliere: hmm good question actually I kind of think that `JustVecWithDefault::decode_len()` should return 2.
Also for append actually I kind of think that if the storage is empty it should append to default value instead of empty value.

bkchr: The traits are not aware of the default values.

srml/elections/src/lib.rs
```diff
@@ -779,8 +779,10 @@ impl<T: Trait> Module<T> {
 						// NOTE: this is safe since the `withdraw()` will check this.
 						locked_balance -= T::VotingFee::get();
 					}
-					Self::checked_push_voter(&mut set, who.clone(), next);
-					<Voters<T>>::insert(next, set);
+					if set_len + 1 == VOTER_SET_SIZE {
+						NextVoterSet::put(next + 1);
+					}
+					<Voters<T>>::safe_append(next, &[Some(who.clone())])
```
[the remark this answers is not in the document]
kianenigma: incrementing the index (`next`) of the chuncked vector for the next append basically.

thiolliere:
hmm after reflection decode length can be implemented for linked_map as well thus it doesn't make sense to isolate it from StorageMap trait.

instead we should have same as for StorageValue a fn decode_len where DecodeLen... and then an automatic implementation in generate::StorageMap and a rewrite of the extension for linked_map (as it already does for some other function)

all in all I think it doesn't worth combining linkedmap and storagemap into a single trait, but this could be made in another PR.

srml/support/src/storage/storage_items.rs
thiolliere: is this default feature still used ? at some point if it is not we could get rid of it and if a user really want to use it then it has to declare its own variable and implement trait Default on it with the value he wants.

kianenigma:
Looks like this is stable again (hoping the CI will be green) as we speak. Some loose ends to decide on hopefully asap:

- [x] `safe_append()` uses `::put()` which is quite a strange behavior but a) is actually _safe_ b) the documentation clearly explains and warns the user. 
- [ ] `NoDefault` trait could be a good idea to make the `decode_length` more intuitive but I recommend for now merging this as-is and coming back to it via another issue since Kusama and Polkadot runtime also has some places where it can use decode length for better speed.

kianenigma:
> Looks like this is stable again (hoping the CI will be green) as we speak. Some loose ends to decide on hopefully asap:
> 
> * [x]  `safe_append()` uses `::put()` which is quite a strange behavior but a) is actually _safe_ b) the documentation clearly explains and warns the user.
> * [x]  `NoDefault` trait could be a good idea to make the `decode_length` more intuitive but I recommend for now merging this as-is and coming back to it via another issue since Kusama and Polkadot runtime also has some places where it can use decode length for better speed.

For the last time: _too late_ and did the second point: This should go for a final review now.

srml/support/src/traits.rs
```diff
@@ -26,6 +26,9 @@ use crate::sr_primitives::ConsensusEngineId;
 
 use super::for_each_tuple;
 
+/// A marker trait that should only be implemented for storage types to do not have a default value.
```
```suggestion
/// A marker trait that should only be implemented for storage types that do not have a default value.
```

srml/elections/src/lib.rs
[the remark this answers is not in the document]
kianenigma: Can't do sir. we need to still use it afterwards and can't move.

srml/support/src/storage/hashed/generator.rs
```diff
@@ -202,16 +203,54 @@ pub trait StorageValue<T: codec::Codec> {
...
 	/// `T` is required to implement `codec::EncodeAppend`.
+	///
+	/// This function only exists for storage values without default value..
 	fn append<S: HashedStorage<Twox128>, I: codec::Encode>(
-		items: &[I], storage: &mut S
-	) -> Result<(), &'static str> where T: codec::EncodeAppend<Item=I> {
+		items: &[I],
+		storage: &mut S,
+	) -> Result<(), &'static str>
+		where T: codec::EncodeAppend<Item=I>, Self: crate::traits::NoDefault,
```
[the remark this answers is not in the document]
kianenigma: To avoid the confusion of
```
pub foo: Vec<u32> = vec![1]; 
assert__eq!(foo::append(&[2]), vec![1,2]); 
assert__eq!(foo::append(&[2]), vec![2])
```

kianenigma:
Updated @bkchr 

Haven't done this yet but TODO:
- [ ] _ It would be nice if you could adapt this to the latest EncodeAppend interface changes. I tried to make it accepting a broader range of values ;)_. 

Prefer the rest of the changes to be reviewed again first.

srml/support/src/storage/hashed/generator.rs
```diff
@@ -203,15 +208,45 @@ pub trait StorageValue<T: codec::Codec> {
...
 	fn append<S: HashedStorage<Twox128>, I: codec::Encode>(
-		items: &[I], storage: &mut S
-	) -> Result<(), &'static str> where T: codec::EncodeAppend<Item=I> {
+		items: &[I],
+		storage: &mut S,
+	) -> Result<(), &'static str>
+		where T: codec::EncodeAppend<Item=I>
+	{
 		let new_val = <T as codec::EncodeAppend>::append(
 			storage.get_raw(Self::key()).unwrap_or_default(),
```
thiolliere: maybe the trait default should be used now here instead of type default.

let's say you have `MyStorage: Vec<u32> = vec![3]` and you do append on it, it will append from an empty vec.

srml/support/src/storage/hashed/generator.rs
```diff
@@ -203,15 +208,45 @@ pub trait StorageValue<T: codec::Codec> {
...
 			items,
 		).map_err(|_| "Could not append given item")?;
 		storage.put_raw(Self::key(), &new_val);
 		Ok(())
 	}
+
+	/// Safely append the given items to the value in the storage.
+	///
+	/// `T` is required to implement `codec::EncodeAppend`.
+	fn safe_append<'a, S: HashedStorage<Twox128>, I>(
```
thiolliere: at some point if we want to make api coherent I think in the future we could introduce safe_get and get the same as append. safe_get would return default on codec error, get would return err

srml/democracy/src/lib.rs
```diff
@@ -933,9 +931,9 @@ impl<T: Trait> Module<T> {
 			if info.delay.is_zero() {
 				Self::enact_proposal(info.proposal, index);
 			} else {
-				<DispatchQueue<T>>::mutate(
+				<DispatchQueue<T>>::safe_append(
```
gavofyork: this looks like `extend` (or `extend_from_slice`). also, is there an `unsafe_append`?

srml/support/src/storage/mod.rs
```diff
@@ -138,10 +141,31 @@ pub trait StorageValue<T: Codec> {
...
+	/// `T` is required to implement `Codec::EncodeAppend`.
+	///
+	/// Upon any failure, it replaces `items` as the new value (assuming that the previous stored
+	/// data is simply corrupt and no longer usable).
+	///
+	/// ### WARNING
+	///
+	/// use with care; if your use-case is not _exactly_ as what this function is doing,
+	/// you should use append and sensibly handle failure within the runtime code if it happens.
+	fn safe_append<'a, I: 'a + Encode + Clone>(items: &'a[I])
```
gavofyork: i would probably rename this to either `append_or_put` or `force_append`.

gavofyork:
looks reasonable aside from naming

srml/support/src/storage/hashed/generator.rs
```diff
@@ -203,15 +208,45 @@ pub trait StorageValue<T: codec::Codec> {
...
+	///
+	/// `T` is required to implement `Codec::DecodeLength`.
+	fn decode_len<S: HashedStorage<Twox128>>(storage: &mut S) -> Result<usize, &'static str>
+		where T: codec::DecodeLength, T: Len
+	{
+		// attempt to get the length directly.
+		if let Some(k) = storage.get_raw(Self::key()) {
+			<T as codec::DecodeLength>::len(&k).map_err(|e| e.what())
+		} else {
+			Ok(Self::Default::default().len())
```
thiolliere: note: this Len trait could have been avoided using: `DecodeLength::len(Self::Default::default().encode())`

but it's probably better this way

srml/democracy/src/lib.rs
kianenigma: the normal `append()` is the unsafe one. 

Though I wouldn't call it actually __unsafe__; if the data is corrupt and non-decodable, it will return an error -- I don't see this as an __unsafe behaviopr__. The other one (currently named `safe_append`) will simply `put()` if the data is corrupt. 

I will apply your recommendation of `append()` and `append_or_put()` since it is more clear.

srml/support/src/storage/hashed/generator.rs
kianenigma: That a great point, thanks! 

Done and added a bunch of tests for it.

srml/support/procedural/src/storage/impls.rs
```diff
@@ -283,6 +301,23 @@ impl<'a, I: Iterator<Item=syn::Meta>> Impls<'a, I> {
...
+			{
+				fn default() -> Option<#typ> {
+					#default_delegator_return
+				}
+			}
+
+			impl<#impl_trait> #scrate::storage::hashed::generator::DecodeLengthStorageMap<#kty, #typ>
+				for #name<#trait_and_instance> #where_clause
+			{
+				type Default = #default_delegator_ident<#trait_and_instance>;
```
thiolliere: (as a note: instead of having an associated type you could make it a function and get rid of StorageDefault trait.)

srml/support/src/storage/hashed/generator.rs
```diff
@@ -307,4 +347,41 @@ pub trait AppendableStorageMap<K: codec::Codec, V: codec::Codec>: StorageMap<K,
...
+	///
+	/// `T` is required to implement `Codec::DecodeLength`.
+	fn decode_len<S: HashedStorage<Self::Hasher>>(key: &K, storage: &mut S) -> Result<usize, &'static str>
+		where V: codec::DecodeLength, V: Len
+	{
+		let k = Self::key_for(key);
+		if let Some(v) = storage.get_raw(&k[..]) {
+			<V as codec::DecodeLength>::len(&v).map_err(|e| e.what())
+		} else {
+			let default = Self::Default::default().ok_or("could not use default as fallback")?;
```
thiolliere: I kind of feel weird returning an error in case of: `MyStorage: u32 => Option<u32>` `MyStorage::decode_len(not_existing_key) == Err("could not use default as fallback")`

At some point for me the best would be to have a different signature for storage with option and storage without:
decode_len would return `Result<Option<usize>, _>` for storage with option and `Result<usize, _>` for a storage without.

But this would need a new associated type in storagemap. I'm not sure what we want for this PR.

srml/support/src/storage/hashed/generator.rs
kianenigma: Not sure which one is more efficient here and makes the most sense. One problem with using an iterator to infer the size that I circumvented but still exists is that Options, including `None`, have size 1 when `into_iter()`-ed which I find weird and can lead to confusion.

https://github.com/paritytech/substrate/blob/346be4baad9286cf2c4f5ae835f6895664383737/srml/support/src/storage/storage_items.rs#L901-L928

Ideally, I would personally do neither and have the codec manually implement an auxiliary `.default_len()` through a trait for all the types that also have `DecodeLength`.

srml/support/src/storage/hashed/generator.rs
kianenigma: I have to check via a test but I think if you have `MyStorage::decode_len` you get another compile-time error complaining that u32 does not implement `DecodeLength`.

srml/support/src/storage/hashed/generator.rs
thiolliere: I don't understand which iterator you are talking about, I was saying you could encode default value and call DecodeLength on it.

thiolliere: hmm also after all I'm not sure I see the issue with using iterator to infer the size: DecodeLength is not implemented on Option so there is no issue there and if it was that would be fine as well to me actually.

srml/support/src/storage/hashed/generator.rs
thiolliere: EDIT: `MyStorage: u32 => Option<Vec<u32>>`

kianenigma: I see. Yeah, the error message could have been improved maybe. 

Basically, you are referring to these tests

https://github.com/paritytech/substrate/blob/346be4baad9286cf2c4f5ae835f6895664383737/srml/support/src/storage/storage_items.rs#L915-L925

and the flow that I see in it is:
- you want to read length
- key does not exist
- it is an option (we don't enforce `Default::default()` to be implemented for options hence I cannot assume it exists)
- the type does not have a default value given via our internal `type Default`. 
- Then you see this error.

I should note that at some point the test was:

```
assert_eq!(OptionMapVec::get(0), None);
assert_eq!(OptionMapVec::decode_len(0), Ok(0)); 
```

But I think now it is better and it is just a matter of returning a better error string. Maybe `Err("Could not read length through given default value")`?

thiolliere: maybe "Value does not exist and could not use default as fallback"
but yeah ok

srml/support/procedural/src/storage/impls.rs
```diff
@@ -571,8 +604,11 @@ impl<'a, I: Iterator<Item=syn::Meta>> Impls<'a, I> {
 				for #name<#trait_and_instance> #where_clause
 			{
 				type Query = #value_type;
-
 				type Hasher = #scrate::#hasher;
+				// NOTE: this must not be used as (). If a method wants to use this, A proper value
+				// (similar to normal map) should be passed in to get consistent behavior with other
+				// Defaults.
+				type Default = ();
```
thiolliere: hmm yes ok, maybe it worth implementing it as well even if we indeed don't use it.

kianenigma: Fair point: having a bit of extra code is better than having ourselves confused in 6 months time not knowing what is wrong.

srml/support/src/storage/hashed/generator.rs
```diff
@@ -202,16 +203,54 @@ pub trait StorageValue<T: codec::Codec> {
 	/// Append the given items to the value in the storage.
 	///
 	/// `T` is required to implement `codec::EncodeAppend`.
+	///
+	/// This function only exists for storage values without default value..
 	fn append<S: HashedStorage<Twox128>, I: codec::Encode>(
-		items: &[I], storage: &mut S
-	) -> Result<(), &'static str> where T: codec::EncodeAppend<Item=I> {
+		items: &[I],
```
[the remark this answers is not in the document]
bkchr: Can we please not resolve stuff, that is not resolved?

srml/support/src/storage/hashed/generator.rs
```diff
@@ -203,23 +208,64 @@ pub trait StorageValue<T: codec::Codec> {
...
+	) -> Result<(), &'static str>
+		where T: codec::EncodeAppend<Item=I>
+	{
 		let new_val = <T as codec::EncodeAppend>::append(
-			storage.get_raw(Self::key()).unwrap_or_default(),
+			// if the key exists, directly append to it.
+			storage.get_raw(Self::key()).unwrap_or_else(|| {
+				// otherwise, try and read a proper __provided__ default.
+				Self::Default::default().map(|v| v.encode())
+					// or just use the default value.
```
bkchr: This comment is misleading. In the context you first think that you talk about the `Self::Default` stuff. Actually we just use `Vec::default()`.

srml/support/src/storage/hashed/generator.rs
```diff
@@ -203,23 +208,64 @@ pub trait StorageValue<T: codec::Codec> {
...
+	///
+	/// `T` is required to implement `Codec::DecodeLength`.
+	fn decode_len<S: HashedStorage<Twox128>>(storage: &mut S) -> Result<usize, &'static str>
+		where T: codec::DecodeLength, T: Len
+	{
+		// attempt to get the length directly.
+		if let Some(k) = storage.get_raw(Self::key()) {
+			<T as codec::DecodeLength>::len(&k).map_err(|e| e.what())
+		} else {
+			let default = Self::Default::default()
```
bkchr: The else branch should be
```
Ok(Self::Default::default().map(|v| v.len()).unwrap_or(0))
```

When we don't have default value, then the length is always `0`.

srml/support/src/storage/hashed/generator.rs
```diff
@@ -295,16 +341,61 @@ pub trait EnumerableStorageMap<K: codec::Codec, V: codec::Codec>: StorageMap<K,
...
+	///
+	/// `T` is required to implement `Codec::DecodeLength`.
+	fn decode_len<S: HashedStorage<Self::Hasher>>(key: &K, storage: &mut S) -> Result<usize, &'static str>
+		where V: codec::DecodeLength, V: Len
+	{
+		let k = Self::key_for(key);
+		if let Some(v) = storage.get_raw(&k[..]) {
+			<V as codec::DecodeLength>::len(&v).map_err(|e| e.what())
+		} else {
+			let default = Self::Default::default()
```
bkchr: Same.

srml/support/src/traits.rs
```diff
@@ -26,6 +26,28 @@ use crate::sr_primitives::ConsensusEngineId;
...
+impl<T: Default> StorageDefault<T> for () { fn default() -> Option<T> { Some(Default::default()) } }
+
+/// Anything that can have a `::len()` method.
+pub trait Len {
+	/// Return the length of data type.
+    fn len(&self) -> usize;
+}
+
+impl<T: IntoIterator + Clone,> Len for T where <T as IntoIterator>::IntoIter: ExactSizeIterator {
+    fn len(&self) -> usize {
```
bkchr: Spaces.

srml/support/src/traits.rs
```diff
@@ -26,6 +26,28 @@ use crate::sr_primitives::ConsensusEngineId;
...
+	fn default() -> Option<V>;
+}
+
+// FIXME #1466 This is needed for `storage_items!`. Should be removed once it is deprecated.
+impl<T: Default> StorageDefault<T> for () { fn default() -> Option<T> { Some(Default::default()) } }
+
+/// Anything that can have a `::len()` method.
+pub trait Len {
+	/// Return the length of data type.
+    fn len(&self) -> usize;
```
bkchr: Spaces.

srml/support/procedural/src/storage/transformation.rs
```diff
@@ -753,10 +753,26 @@ fn decl_storage_items(
...
+			.map(|d| quote!( #d ))
+			.unwrap_or_else(|| quote!{ Default::default() });
 		let kind = type_infos.kind.clone();
 		// Propagate doc attributes.
 		let attrs = attrs.inner.iter().filter_map(|a| a.parse_meta().ok()).filter(|m| m.name() == "doc");
 
+		// create default value delegator
+		let default_delegator_ident = Ident::new(
+			&format!("{}{}", name.to_string(), "DefaultDelegator"),
+			proc_macro2::Span::call_site()
```
bkchr:
```suggestion
			proc_macro2::Span::call_site(),
```

srml/support/src/storage/hashed/generator.rs
```diff
@@ -295,16 +341,61 @@ pub trait EnumerableStorageMap<K: codec::Codec, V: codec::Codec>: StorageMap<K,
...
 		).map_err(|_| "Could not append given item")?;
 		storage.put_raw(&k[..], &new_val);
 		Ok(())
 	}
+
+	/// Safely append the given items to the value in the storage. If a codec error occurs, then the
+	/// old (presumably corrupt) value is replaced with the given `items`.
+	///
+	/// `T` is required to implement `codec::EncodeAppend`.
+	fn append_or_put<'a, S, I>(
```
bkchr:
```suggestion
	fn append_or_insert<'a, S, I>(
```

I would stick to the terminology of the map.

srml/support/src/storage/mod.rs
```diff
@@ -262,6 +299,19 @@ pub trait AppendableStorageMap<K: Codec, V: Codec>: StorageMap<K, V> {
...
+	/// Append the given items to the value in the storage.
+	///
+	/// `T` is required to implement `codec::EncodeAppend`.
+	///
+	/// Upon any failure, it replaces `items` as the new value (assuming that the previous stored
+	/// data is simply corrupt and no longer usable).
+	///
+	/// WARNING: use with care; if your use-case is not _exactly_ as what this function is doing,
+	/// you should use append and sensibly handle failure within the runtime code if it happens.
+	fn append_or_put<'a, KeyArg: Borrow<K>, I: Encode + Clone>(key: KeyArg, items: &'a[I])
```
bkchr:
```suggestion
	fn append_or_insert<'a, KeyArg: Borrow<K>, I: Encode + Clone>(key: KeyArg, items: &'a[I])
```

srml/support/src/storage/hashed/generator.rs
thiolliere: I'm ok for that as well, my point is still that the perfect implementation would be that decode_len on an `Option<Vec>` returns an `Option<usize>` but I'm ok with that as well.

srml/support/src/storage/hashed/generator.rs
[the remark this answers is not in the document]
kianenigma: Yes my apologies, I added it as a TODO somewhere else and resolved this. Done and will be pushed.

srml/support/procedural/src/storage/impls.rs
kianenigma: Linked map now has the same decode_length 👍

srml/support/src/storage/hashed/generator.rs
bkchr: I don't have a strong opinion on this. What would be the advantage to get a `None` as size?

srml/support/src/storage/hashed/generator.rs
kianenigma: the goal was to de-mistify this by explicitly mentioning in the line above: `**provided** default value` so that once knows by `default value` I mean rust's `Default`. made explicit now 👍

srml/support/src/storage/hashed/generator.rs
kianenigma: If both you don't have a strong opinion I would keep it as-is because of something like:

```rust
assert_eq(OptionVec::get(), None);
assert_eq(OptionVec::decode_len(), 0);
assert_eq(OptionVec::decode_len(), None);
```

bkchr: What? Your current solution does not reflect what you are saying in the last comment.

And `OptionVec::get() == None` implies `OptionVec::decode_len() ==0`.

I said that I don't have a strong opinion on if we want to follow @thiolliere approach or mine, but I don't want to follow the current implementation. Options having a default value does not makes any sense at all. I will come up with a PR that forbids that.

kianenigma: > What? Your current solution does not reflect what you are saying in the last comment.

Does with the exception that it returns `Err(...)` instead of `None`. I see it aligned with _`perfect implementation would be that decode_len on an Option<Vec> returns an Option<usize>`_. 

But I see your bigger point here esp. with 

> options having a default value does not make any sense at all. I will come up with a PR that forbids that.

(fyi. I have some tests now that assign defaults to options. I assume they will be removed once they are banned. For now, they are good examples)

So will change as you recommended. thanks for the description 👍

thiolliere: > I don't have a strong opinion on this. What would be the advantage to get a None as size?

It is just for logic for Some(vec) decode len returns Some(len) and for None decode len returns None.
Maybe sometimes someone want to know the len of 

> If both you don't have a strong opinion I would keep it as-is

it is more like there is three option:
```
assert_eq(OptionVec::get(), None);

// 1- Basti proposal:
assert_eq(OptionVec::decode_len(), Ok(0));
// 2- Current state:
assert_eq(OptionVec::decode_len(), Err(".."));
// 3- Thiolliere proposal but changing the signature might be cumbersome
assert_eq(OptionVec::decode_len(), Ok(None));
```

I'm ok with all solution actually, though yes 1- seems ambiguous to me and I prefer 2 or 3

kianenigma:
https://images.app.goo.gl/xQr6GZk1w35v9iuT8

@bkchr me being like ^

PR #3076, opened 2019-07-09
libp2p-next

romanb:
This PR supersedes https://github.com/paritytech/substrate/pull/2981 and incorporates all necessary changes for the next libp2p release, which currently include:

  * Updates to the Kademlia APIs.
  * Updated imports due to the extracted libp2p-swarm crate from libp2p-core.

At least the following libp2p PRs are likely to be still pending inclusion in the next libp2p release:

  * https://github.com/libp2p/rust-libp2p/pull/1189
  * https://github.com/libp2p/rust-libp2p/pull/1191
  * https://github.com/libp2p/rust-libp2p/pull/1194


core/network/src/discovery.rs
```diff
@@ -94,21 +97,30 @@ impl<TSubstream> DiscoveryBehaviour<TSubstream> {
 
 	/// Get a record from the DHT.
 	pub fn get_value(&mut self, key: &Multihash) {
-		self.kademlia.get_value(key, NonZeroU8::new(10)
-								.expect("Casting 10 to NonZeroU8 should succeed; qed"));
+		self.kademlia.get_record(key, Quorum::N(
+			NonZeroUsize::new(10).expect("10 > 0; qed")));
```
romanb: I would like to understand the reasoning behind the choice of `10` here, especially since, in the previous implementation, this meant "at least 1, at most 10" for a success. The new semantics of `Quorum` are "at least 10" for a success. I didn't find any code that actually processes these records, maybe someone can point me to it? @montekki ?

montekki: Just a randomly chosen value at that point.

For substrate the records need to be binary blobs with signatures, and nodes have to throw away any blobs

1. that can not be properly decoded.
2. have wrong signatures

that was the reason for the addition of `store_mut` API. 

So in any case we have to do our own substrate-specific record checking logic on top of the way kademlia checks things. Now, we have to make sure that the quoruming system is not exploitable: suppose you are looking for key `ABC`. you receive 2 votes from honest nodes that `ABC` is `FOO` and 3 votes from malicious nodes that `ABC` is `BAR`. I believe that the specs for `libp2p-kad` tell us that in this case `BAR` is the correct answer since it's just the majority of the votes.

core/network/src/discovery.rs
```diff
@@ -94,21 +97,30 @@ impl<TSubstream> DiscoveryBehaviour<TSubstream> {
...
 	}
 
 	/// Put a record into the DHT.
 	pub fn put_value(&mut self, key: Multihash, value: Vec<u8>) {
```
montekki: I think that the name of the api here also has to change for convenience

core/network/src/discovery.rs
romanb: > [..] that was the reason for the addition of store_mut API.

It may be better to not store the record in the first place. By wrapping the memory store in your own implementation, you could validate records before they are stored in `RecordStore::put`. Granted, you may want some new error variant for the error result, which I'm considering on adding later.

Let me just try to explain again what a "quorum" means for the Kademlia DHT as implemented in `libp2p-kad`, in case it is still unclear: It specifies the _minimum number of results from distinct nodes w.r.t. the configured replication factor that must be obtained in order for the query to be considered successful_. It is in the nature of the DHT that you neither know how many nodes participate, nor on how many nodes a record is actually stored. Therefore, the only use for a quorum > 1 that I can see is a scenario where you want to make a probabilistic argument that a `GET_VALUE` will see the "newest" or "best" value (for some definitions of "newest" or "best" determined by inspection of the value) in a non-malicious setting where the value associated with a key may either change over time (by the record being re-published with a new value) or the initial creation of the record is subject to non-determinism. In particular, I don't see how the DHT could help you with byzantine fault tolerance for some protocol built on top, if that is what you are referring to when you talk about "malicious nodes".

core/network/src/discovery.rs
romanb: Yes, but I'd like to leave that for a follow-up PR, to keep this once focused on minimal changes necessary for compatibility with the next libp2p release.

core/network/src/discovery.rs
montekki: so in that case we should just be using `quorum == 1` it looks like.

core/network/src/discovery.rs
```diff
@@ -365,6 +393,6 @@ mod tests {
 			}
 		});
 
-		tokio::runtime::Runtime::new().unwrap().block_on(fut).unwrap();
+		tokio::run(fut)
```
tomaka: `tokio::run` is equivalent to the code that was previously there, except that it silently ignores errors.

Let's put this back:

```suggestion
		tokio::runtime::Runtime::new().unwrap().block_on(fut).unwrap();
```

core/network/src/discovery.rs
```diff
@@ -224,50 +237,58 @@ where
...
+						return Async::Ready(NetworkBehaviourAction::GenerateEvent(ev));
+					}
+					KademliaEvent::RoutingUpdated { peer, .. } => {
+						let ev = DiscoveryOut::Discovered(peer);
+						return Async::Ready(NetworkBehaviourAction::GenerateEvent(ev));
+					}
+					KademliaEvent::GetClosestPeersResult(res) => {
+						match res {
+							Err(e) => {
+								trace!(target: "sub-libp2p", "Kademlia query for closest peers failed: {:?}", e);
```
tomaka: The logging scheme is to mention what informed us.

```suggestion
								trace!(target: "sub-libp2p", "Libp2p => Kademlia query for closest peers failed: {:?}", e);
```

Also, in which situations can this happen? Isn't it better to use `error!`?

core/network/src/discovery.rs
```diff
@@ -224,50 +237,58 @@ where
...
+							Ok(ok) => DiscoveryOut::ValuePut(ok.key),
+							Err(e) => {
+								DiscoveryOut::ValuePutFailed(e.into_key())
+							}
+						};
+						return Async::Ready(NetworkBehaviourAction::GenerateEvent(ev));
+					}
+					// We never start any other type of query.
+					e => {
+						trace!(target: "sub-libp2p", "Ignored Kademlia event: {:?}", e)
```
tomaka: Let's make this more verbose:

```suggestion
						warn!(target: "sub-libp2p", "Libp2p => Unexpected Kademlia event: {:?}", e)
```

tomaka:
Libp2p 0.11.0 is now published.

core/network/src/discovery.rs
romanb: I put it back because I don't mind, though I don't follow what you mean by "silently ignores errors". It seems to me that `tokio::run` takes a `Future<Item=(),Error=()>`, so it only compiles if the future does not produce any meaningful errors, which is the case here and if the error type would change, it would no longer compile. I thus don't see how errors would be unintentionally silenced by using `tokio::run`. :man_shrugging:

core/network/src/discovery.rs
romanb: > Also, in which situations can this happen? Isn't it better to use `error!`?

Right now, the only possible error for this query is a timeout. So what logging is desirable depends on the desired configuration / preferences. With the default timeout of 60 seconds, I would definitely not suggest logging these as errors, especially since these are "background queries". If the query timeout would be set large enough, I suppose a timeout could be considered an error. I'm not sure that is preferable though. Indeed, the timeout error also communicates the key and the peers found within the time constraints, so I suggest [this](https://github.com/paritytech/substrate/pull/3076/commits/f5259e24ee42c6c4b72510fe1ee51b11a713cae1#diff-7528f75916abf8807d5e5129430b9e19R316). What do you think?

core/network/src/discovery.rs
romanb: That implies some other changes though, since not all events that are currently ignored are unexpected. See the changes I made in: https://github.com/paritytech/substrate/pull/3076/commits/f5259e24ee42c6c4b72510fe1ee51b11a713cae1.

core/network/src/discovery.rs
tomaka: I'm thinking of runtime errors.
If the `Future` panics, then `tokio::run` will still succeed and the test will succeed.

mxinden:
@romanb friendly ping: Is there something else missing for this pull request to be merged? It would be great to use https://github.com/libp2p/rust-libp2p/commit/5861474d99b843141ae74672b7b21fb09a63b143 within Substrate.

Issue #2334, opened 2019-04-20
Substrate Runtime Interface


pepyakin:
The current substrate runtime API is not the easiest to work with:

- We have a couple of different runtime environments (`with_std`/`without_std` - or rather, staticly linked native code and wasm runtime), however interface should be the same. 
- Error handling might differ between environments.
- Implementation and interface are located in different places but they are really tightly coupled and I mean it: say, a resulting pointer from runtime API must have a certain alignment or we depend on that in unsafe code. TBH I won't be surprised if there is broken unsafe code out there since it is so easy to break it, so hard to check, and we don't really treat this code as unsafe (e.g. there are no proofs on safety).
- These unsafe invariants not usually stated anywhere, they are just in code. And there is also lack of documentation and the reason for that might be it is just not clear where this doc belongs.
- There is this weird behavior of `impl_function_executor`, which allows you declare parameters such as `usize` which, however, will have `u32` type, which might be baffling for new people working with this code. [There was a PR](https://github.com/paritytech/wasmi/pull/164) that introduces an honest Rust interface, but unfortunately it has a downside: there is a desire for signatures of host functions to be trivially copyable without any changes to reduce error proness.
- There are some differences and pitfalls to keep in mind when designing Substrate Runtime APIs. For example, multiple returns are not supported and thus if there is need to return multiple values it should be workarounded somehow. And yeah there are some inconstencies at the moment in the current implementation.

Basically, this is very error-prone and very boilerplaty code. This sounds as a good use-case for a code-generation. Here is my strawman proposal for such code generator:

We could introduce a special AST that describes an interface between the substrate runtime and the substrate host. It would support rather high-level types, e.g.: `bytes_vec`/`bytes`, `bool`, `(T, J)` (a tuple), `*T` (a raw pointer), `[T; N]` for parameters and return values. We could also declare if a function traps, maybe with a type `Result` or a special annotation.

Here is an example how it could look like:

```
# Not a rust file

# usize is not supported deliberatly, use u32.
fn malloc(size: u32) -> *const u8;
fn free(ptr: *const u8);

# `bytes` is converted to `Vec<u8>` on the substrate host side, and `&[u8]` on the rust side.
fn storage_exists(key: bytes) -> bool;

# `bytes?` - notation for optional value (for denoting absence of a value under the given key)
# note that we don't deal with problems such as how to return a bytes vector (which is two components `ptr` and `len`) via wasm: all such ABI details will be handled by codegeneration.
[...]

PR #3079, opened 2019-07-10
Introduce srml/im-online

cmichi:
Addresses https://github.com/paritytech/substrate/issues/2719.

There are not yet any tests, I'll add those in a follow-up PR. The PR is mostly here to quickly get this into the review pipeline and master.

I currently keep only the `Vec<u8>` version of `Multiaddr` and `PeerId.to_string()` in the runtime, since `parity-multiaddr` doesn't compile for wasm currently and `libp2p::core::PeerId` doesn't support encoding/serializing yet.


srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+				}
+			}
+
+			let gossip_at = <GossipAt<T>>::get();
+			let now = <system::Module<T>>::block_number();
+			if gossip_at == now {
+				match gossip::<T>(block_number) {
+					Ok(_) => {},
+					Err(_) => {
+						print("Error in offchain worker!");
```
cmichi: I'm not sure about the proper handling here. Maybe it makes sense to only print something out for more serious things?

These are the possible errors:
```
enum OffchainErr {
	DecodeAuthorityId,
	ExtrinsicCreation,
	FailedSigning,
	LocalNetworkState,
	SubmitTransaction,
}
```

core/service/Cargo.toml
```diff
@@ -32,6 +32,7 @@ transaction_pool = { package = "substrate-transaction-pool", path = "../../core/
 rpc = { package = "substrate-rpc-servers", path = "../../core/rpc-servers" }
 tel = { package = "substrate-telemetry", path = "../../core/telemetry" }
 offchain = { package = "substrate-offchain", path = "../../core/offchain" }
+parity-multiaddr = { package = "parity-multiaddr", version = "0.5.0" }
```
tomaka: I don't think you're using this dependency, and if you are, `multiaddr` is reexported by `network`.

core/offchain/src/api.rs
```diff
@@ -158,6 +170,32 @@ impl<Storage, KeyProvider> OffchainExt for Api<Storage, KeyProvider> where
...
+		}
+	}
+
+	fn local_network_state(&self) -> Result<Vec<u8>, ()> {
+		let network_state = &self.network_state;
+		let state = LocalNetworkState {
+			external_addresses: network_state.external_addresses(),
+
+			// TODO derive `Serialize` for `libp2p::core::PeerId`
+			peer_id: network_state.peer_id().to_string(),
```
tomaka:
```suggestion
			peer_id: network_state.peer_id().to_base58(),
```

`PeerId` is in this weird spot where the serialization option is not obvious, so you have to explicitly call `to_base58()` if you want base58, or `as_bytes()` if you want bytes.

core/offchain/Cargo.toml
```diff
@@ -16,6 +16,10 @@ parking_lot = "0.8.0"
 primitives = { package = "substrate-primitives", path = "../../core/primitives" }
 runtime_primitives = { package = "sr-primitives", path = "../../core/sr-primitives" }
 transaction_pool = { package = "substrate-transaction-pool", path = "../../core/transaction-pool" }
+network = { package = "substrate-network", path = "../../core/network" }
+serde = { version = "1.0", features = ["derive"] }
+bincode = "1"
+parity_multiaddr = { package = "parity-multiaddr", version = "0.5.0" }
```
tomaka: `multiaddr` can be accessed through `network::multiaddr`.

```suggestion
```

core/offchain/src/api.rs
```diff
@@ -33,6 +33,10 @@ use runtime_primitives::{
 	traits::{self, Extrinsic},
 };
 use transaction_pool::txpool::{Pool, ChainApi};
+use network::NetworkStateInfo;
+use parity_multiaddr::Multiaddr;
```
tomaka:
```suggestion
use network::multiaddr::Multiaddr;
```

tomaka:
Also, `parity-multiaddr` should normally compile fine for WASM. It's used in my WASM browser demo.

core/sr-primitives/src/testing.rs
```diff
@@ -44,6 +44,12 @@ impl TypedKey for UintAuthorityId {
 	const KEY_TYPE: KeyTypeId = UINT_DUMMY_KEY;
 }
 
+impl AsRef<[u8]> for UintAuthorityId {
+	fn as_ref(&self) -> &[u8] {
+		unsafe { &std::mem::transmute::<_, &[u8; 4]>(&self.0)[..] }
```
tomaka: Let's be **extremely** careful when using transmute. This code definitely isn't correct.

bkchr: Besides `u64` being `[u8;8]`, I don't see any problem here?

bkchr: And casting to `[u8;4]` is valid.

bkchr: Yeah, my fault. We should use `slice::from_raw_parts` here to get a correct fat pointer.

cmichi:
@tomaka 

> Also, parity-multiaddr should normally compile fine for WASM. It's used in my WASM browser demo.

Ok, thanks. I'll keep the encoded version in wasm for now, since we will probably never need the info in WASM, but only "outside". So we can save the dependency until there is a need to have the encoded object in WASM.

I addressed your comments, can you please take a look again?

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+				// verify that the incoming (unverified) pubkey is actually an authority id
+				let is_authority = <T as Trait>::is_valid_authority_id(&heartbeat.authority_id);
+				if !is_authority {
+					return TransactionValidity::Invalid(ApplyError::BadSignature as i8);
+				}
+
+				if signature.len() != 64 {
+					return TransactionValidity::Invalid(ApplyError::BadSignature as i8);
+				}
+				let mut array = [0; 64];
```
tomusdrw: Maybe something like this to have a proper name instead of `array`?
```rust
let signature = {
  let mut array = [0; 64]
  array.copy_from_slice(&signature); // we already know the len is ok, since we have a check earlier
  array
};
```

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+					return TransactionValidity::Invalid(ApplyError::BadSignature as i8);
+				}
+
+				// check if session index from heartbeat is recent
+				let current_session = <session::Module<T>>::current_index();
+				if heartbeat.session_index < current_session {
+					return TransactionValidity::Invalid(ApplyError::BadSignature as i8);
+				}
+
+				let transaction_tag = heartbeat.encode();
```
tomusdrw: You `encode` it for the second time here - please make sure we do it only once.

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+		// ignore
+	}
+}
+
+impl<T: Trait> srml_support::unsigned::ValidateUnsigned for Module<T> {
+	type Call = Call<T>;
+
+	fn validate_unsigned(call: &Self::Call) -> srml_support::unsigned::TransactionValidity {
+		match call {
+			Call::heartbeat(heartbeat, signature) => {
```
tomusdrw: We need additional check to make sure the `hearbeat` was not already sent by that `authority_id`. Most likely using `is_online_in_current_era`

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+	fn on_disabled(_i: usize) {
+		// ignore
+	}
+}
+
+impl<T: Trait> srml_support::unsigned::ValidateUnsigned for Module<T> {
+	type Call = Call<T>;
+
+	fn validate_unsigned(call: &Self::Call) -> srml_support::unsigned::TransactionValidity {
+		match call {
```
tomusdrw: maybe just do `if let Call::hearbeat(hearbeat, signature) = call {` since you only want one value. This will also lower the indentation level.

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+					Err(_) => {
+						// we only run when a local authority key is configured
+						Ok(())
+					}
+				}
+			}
+
+			let gossip_at = <GossipAt<T>>::get();
+			let now = <system::Module<T>>::block_number();
+			if gossip_at == now {
```
tomusdrw: We don't really have guarantees that offchain worker will run for that particular block (You might actually be offline).

I think it's better to check that `gossip_at < now` and have additional Offchain DB entry that says that we already submitted the statement for current session.

srml/im-online/src/lib.rs
tomusdrw: Would be best to just debug-print the entire error. I think we will need some logging from within offchain workers anyhow. Can you implement `sr_io::Printable` for the error and just do `print(err)` here?

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+					},
+					Err(_) => {
+						// we only run when a local authority key is configured
+						Ok(())
+					}
+				}
+			}
+
+			let gossip_at = <GossipAt<T>>::get();
+			let now = <system::Module<T>>::block_number();
```
tomusdrw: We already have `block_number`, I don't think we need to query `system`. Alternative we can get rid of this parameter for offchain workers.

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+			if gossip_at == now {
+				match gossip::<T>(block_number) {
```
tomusdrw: should be `gossip_at` after we change the conditions.

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+						}
+					},
+					Err(_) => {
+						// we only run when a local authority key is configured
+						Ok(())
+					}
+				}
+			}
+
+			let gossip_at = <GossipAt<T>>::get();
```
mxinden: If I understand correctly `gossip_at` is set only once in the initial chain specification. Thus the code below would only ever run once. Shouldn't this run multiple times each session? Am I missing something?

core/sr-io/without_std.rs
```diff
@@ -385,6 +385,20 @@ pub mod ext {
 		/// - nonzero otherwise.
 		fn ext_submit_transaction(data: *const u8, len: u32) -> u32;
 
+		/// Returns information about the local node's network state.
+		///
+		/// # Returns
+		///
+		/// - The encoded `Result<offchain::LocalNetworkState, ()>`.
```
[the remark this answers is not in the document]
pepyakin: That's a good question and I cannot answer it properly because I wasn't the most involved person in designing and adding the externalities.

About the single syscall-like entrypoint: I actually had some thoughts about this kind of design for srml-contracts wasm environment. This kind of build up would add some inflexibility and overhead into the substrate host interface. This doesn't just harms the compilation efficiency, but also introduces more high-level problems. For instance, if your runtime wants to call a function that takes two buffers and you have them at hand you cannot pass them directly and would have to encode them probably (re)allocating more than once.

Also, there is another special thing about this design: it has properties of dynamic dispatch. The default way of interaction between a wasm instance and the host is by resolving **all** imports at the instantiation time and if the instance imports something that doesn't exist the instance cannot be instantiated. On the other hand, with such a dynamic dispatch it would instantiate successfully (cause it imports only the syscall entrypoint) but when it would fail only at the runtime when it tries to call a non-supported function. In general this can be a feature or an issue, but I am inclined to think that in the case of substrate it is more an issue since I doubt that there is a lot to do with conditional logic (the runtime should behave the same regardless the set of available functions).

But yeah, it would definitely make the interface simpler.
As a reference, there is https://github.com/paritytech/substrate/issues/2334 to achieve the same (but arguably more complex than what you propose)

core/sr-io/without_std.rs
```diff
@@ -385,6 +385,20 @@ pub mod ext {
...
+		fn ext_local_network_state(msg_len: *mut u32) -> *mut u8;
+
+		/// Returns the locally configured authority public key, if available.
+		///
+		/// # Returns
+		///
+		/// - The encoded `Result<PublicKey encoded to Vec<u8>, ()>`.
```
pepyakin: Why is the rationaly not using `Option`?

core/sr-io/without_std.rs
pepyakin: I think it is worth noting that the ownership of the returned buffer is transferred to the runtime code and thus the runtime is reponsible for freeing it. Also it would useful to note that this is always a properly allocated pointer (and e.g. cannot be NULL), so the runtime code can always rely on it.

Can we have a description on what is `msg_len`, what does the message mean in this context, is the ownership of this pointer is transferred or just lended, etc.

core/sr-io/without_std.rs
pepyakin: - Can we have a description what does `crypto` mean here?
- The same comments as for `ext_local_network_state` regarding `msg_len` and semantics of the return value

core/executor/src/wasm_executor.rs
```diff
@@ -767,6 +768,42 @@ impl_function_executor!(this: FunctionExecutor<'e, E>,
...
+			.map_err(|_| "Invalid attempt to write msg_len in ext_local_network_state")?;
+
+		Ok(offset)
+	},
+	ext_local_authority_pubkey(
+		kind: u32,
+		written_out: *mut u32
+	) -> *mut u8 => {
+		let kind = offchain::CryptoKind::try_from(kind)
+			.map_err(|_| "crypto kind OOB while ext_decrypt: wasm")?;
```
pepyakin:
```suggestion
			.map_err(|_| "crypto kind OOB while ext_local_authority_pubkey: wasm")?;
```

pepyakin: Wait, is it acually OOB?

core/executor/src/wasm_executor.rs
```diff
@@ -767,6 +768,42 @@ impl_function_executor!(this: FunctionExecutor<'e, E>,
...
+
+		let res = this.ext.offchain()
+			.map(|api| api.local_authority_pubkey(kind))
+			.ok_or_else(|| "Calling unavailable API ext_decrypt: wasm")?;
```
pepyakin:
```suggestion
			.ok_or_else(|| "Calling unavailable API ext_local_authority_pubkey: wasm")?;
```

core/executor/src/wasm_executor.rs
```diff
@@ -767,6 +768,42 @@ impl_function_executor!(this: FunctionExecutor<'e, E>,
...
+
+		let encoded = res.encode();
+		let len = encoded.len() as u32;
+		let offset = this.heap.allocate(len)? as u32;
+		this.memory.set(offset, &encoded)
+			.map_err(|_| "Invalid attempt to set memory in ext_http_response_headers")?;
```
pepyakin:
```suggestion
			.map_err(|_| "Invalid attempt to set memory in ext_local_authority_pubkey")?;
```

core/executor/src/wasm_executor.rs
```diff
@@ -767,6 +768,42 @@ impl_function_executor!(this: FunctionExecutor<'e, E>,
...
+		this.memory.write_primitive(written_out, len)
+			.map_err(|_| "Invalid attempt to write written_out in ext_http_response_headers")?;
```
pepyakin:
```suggestion
			.map_err(|_| "Invalid attempt to write written_out in ext_local_authority_pubkey")?;
```

srml/im-online/src/lib.rs
cmichi: If you look at the implementation of `fn new_session()` (the hook which will be called by `srml-session` once the session changes):
```
let now = <system::Module<T>>::block_number();
<GossipAt<T>>::put(now);
```
So the value is updated each time the session changes, resulting in the code running for each session change.

mxinden: Thanks @cmichi.

core/executor/src/wasm_executor.rs
```diff
@@ -767,6 +768,42 @@ impl_function_executor!(this: FunctionExecutor<'e, E>,
 
 		Ok(offset)
 	},
+	ext_local_network_state(msg_len: *mut u32) -> *mut u8 => {
```
pepyakin: Why `msg_len`?

core/executor/src/wasm_executor.rs
```diff
@@ -767,6 +768,42 @@ impl_function_executor!(this: FunctionExecutor<'e, E>,
...
+			.map(|api| api.local_network_state())
+			.ok_or_else(|| "Calling unavailable API ext_local_network_state: wasm")?;
+
+		let encoded = res.encode();
+		let len = encoded.len() as u32;
+		let offset = this.heap.allocate(len)? as u32;
+		this.memory.set(offset, &encoded)
+			.map_err(|_| "Invalid attempt to set memory in ext_local_network_state")?;
+
+		this.memory.write_primitive(msg_len, len)
```
pepyakin: I think it would be better if we change the name of `msg_len` to something difference, since I find this line quite confusing. `written_out`, as used in the sibiling function, works better IMO.

core/network/src/service.rs
```diff
@@ -249,6 +256,12 @@ impl<B: BlockT + 'static, S: NetworkSpecialization<B>, H: ExHashT> NetworkWorker
 		self.network_service.user_protocol().num_connected_peers()
 	}
 
+	/// Returns the local external addresses.
+	pub fn external_addresses(&self) -> HashSet<Multiaddr> {
```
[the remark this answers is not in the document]
cmichi: I've introduced a `NetworkStateInfo` and it seems that it's currently in stable Rust not possible to return `impl Trait` in a trait definition (https://stackoverflow.com/a/39490692). So I have left it out for now.

core/sr-primitives/src/testing.rs
```diff
@@ -44,6 +44,13 @@ impl TypedKey for UintAuthorityId {
 	const KEY_TYPE: KeyTypeId = UINT_DUMMY_KEY;
 }
 
+impl AsRef<[u8]> for UintAuthorityId {
+	fn as_ref(&self) -> &[u8] {
+		let ptr = self.0 as *const _;
+		unsafe { std::slice::from_raw_parts(ptr, 8) }
```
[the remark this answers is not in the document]
cmichi: Yep, `transmute` would work as well and actually I had `unsafe { &std::mem::transmute::<_, &[u8; 8]>(&self.0)[..] }` in a first version, but one other reviewer had complained about transmute. Hence I changed it.

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+//!
+//! TODO: will be done in follow-up once the API is set.
+//!
+//! ## Dependencies
+//!
+//! This module depends on the [Session module](../srml_session/index.html).
+
+// Ensure we're `no_std` when compiling for Wasm.
+#![cfg_attr(not(feature = "std"), no_std)]
+
```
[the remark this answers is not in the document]
cmichi: I added the missing docs in the meantime, so it's obsolete.

core/executor/src/wasm_executor.rs
cmichi: Yeah, I checked again. Also I just copied this line from the analog methods.

core/sr-io/without_std.rs
cmichi: I just used the same return type as the other existing functions in the API and would leave it like this for consistency.

core/offchain/Cargo.toml
```diff
@@ -16,6 +16,9 @@ parking_lot = "0.8.0"
 primitives = { package = "substrate-primitives", path = "../../core/primitives" }
 runtime_primitives = { package = "sr-primitives", path = "../../core/sr-primitives" }
 transaction_pool = { package = "substrate-transaction-pool", path = "../../core/transaction-pool" }
+network = { package = "substrate-network", path = "../../core/network" }
+serde = { version = "1.0", features = ["derive"] }
```
tomusdrw: Is `serde` and `bincode` still needed for anything?

core/offchain/src/api.rs
```diff
@@ -158,6 +162,33 @@ impl<Storage, KeyProvider> OffchainExt for Api<Storage, KeyProvider> where
 		Ok(CryptoKeyId(id))
 	}
 
+	fn authority_pubkey(&self, kind: CryptoKind) -> Result<Vec<u8>, ()> {
+		match self.read_key(None, kind) {
```
tomusdrw:
```suggestion
		let key = self.read_key(None, kind)?;
		match key {
		 ...
		}
```

core/offchain/src/api.rs
```diff
@@ -158,6 +162,33 @@ impl<Storage, KeyProvider> OffchainExt for Api<Storage, KeyProvider> where
...
+				Ok(public)
+			},
+			Err(e) => Err(e),
+		}
+	}
+
+	fn network_state(&self) -> Result<OpaqueNetworkState, ()> {
+		let external_addresses = self.network_state
+			.external_addresses()
+			.into_iter()
```
tomusdrw: why do we `into_iter()` and `collect()`? Maybe better return `Vec` from `NetworkStateInfo` or have associated type there?

core/offchain/src/api.rs
```diff
@@ -285,6 +316,72 @@ impl<Storage, KeyProvider> OffchainExt for Api<Storage, KeyProvider> where
...
+		NetworkState {
+			peer_id,
+			external_addresses,
+		}
+	}
+}
+
+impl From<NetworkState> for OpaqueNetworkState {
+	fn from(state: NetworkState) -> OpaqueNetworkState {
+		let enc = Encode::encode(&state.peer_id.as_bytes().to_vec());
```
tomusdrw: Are you sure we need both `as_bytes()` and `to_vec()`? This seems like unneeded allocation.

core/offchain/src/api.rs
```diff
@@ -285,6 +316,72 @@ impl<Storage, KeyProvider> OffchainExt for Api<Storage, KeyProvider> where
...
+			external_addresses,
+		}
+	}
+}
+
+impl From<OpaqueNetworkState> for NetworkState {
+	fn from(state: OpaqueNetworkState) -> NetworkState {
+		let inner_vec = state.peer_id.0;
+
+		let bytes: Vec<u8> = <Vec<u8>>::decode(&mut &inner_vec[..])
```
tomusdrw:
```suggestion
		let bytes: Vec<u8> = Decode::decode(&mut &inner_vec[..])
```

or 

`let bytes = <<Vec<u8>>::decode(&mut &inner_vec[..])`

core/offchain/src/api.rs
```diff
@@ -285,6 +316,72 @@ impl<Storage, KeyProvider> OffchainExt for Api<Storage, KeyProvider> where
...
+			.expect("It's always a valid Vec<u8> which is encoded; qed");
+		let peer_id = PeerId::from_bytes(bytes)
+			.expect("These bytes were exported from a valid PeerId; qed");
```
tomusdrw: I really think we should never panic here. Let's implement `TryFrom` instead and handle the error.
There are multiple ways how you can end up with `OpaqueNetworkState` that does not represent a valid `NetworkState` at all.

core/sr-io/src/lib.rs
```diff
@@ -239,6 +240,13 @@ export_api! {
 		/// The transaction will end up in the pool.
 		fn submit_transaction<T: codec::Encode>(data: &T) -> Result<(), ()>;
 
+		/// Returns information about the local node's network state.
+		fn network_state() -> Result<OpaqueNetworkState, ()>;
+
+		/// Returns the currently configured authority public key, if available.
+		// TODO [ToDr] change into crypto_pubkey(&self, key: Option<CryptoKeyId>, kind: CryptoKind)
```
tomusdrw: We should have an issue for this one too.

core/offchain/src/api.rs
cmichi: It's a `HashSet` in `core/network`, but I transform to a `Vec`, since `Encode/Decode` is only implemented for `Vec` currently and the support for `Vec` in WASM is really good. I liked your suggestion of returning `impl Iterator<Item = Multiaddr>`, but as written in the reply for that suggestion that's sadly not possible with Rust in traits currently.

Do you see a shorter way of getting around this allocation? The only alternative I see would be to implement `Encode/Decode` for `HashSet` as well and see if/how well it's supported in WASM.

tomusdrw: Well you can always have an associated type and collect to that, but it might be a bit cumbersome to handle.

But maybe we can just change `HashMap` to `Vec`, since it seems we already get an iterator from `Swarm` so you'd just collect into a `Vec` on that level. I guess collecting to a `HashSet` allows you to deduplicate, but is that needed?

cmichi: Yes, good point. I talked to Pierre and the related code just uses a `HashSet` so that it gets serialized into a JavaScript object as opposed to a Map. So we can collect into a `Vec`.

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,442 @@
...
+			let next_gossip = <GossipAt<T>>::get();
+			let not_yet_gossipped = match was_not_yet_gossipped::<T>(now, next_gossip) {
+				Ok(v) => v,
+				Err(err) => {
+					print(err);
+					return;
+				},
+			};
+			if next_gossip < now && not_yet_gossipped {
+				set_worker_status::<T>(now, false);
```
tomusdrw: This is subject to race conditions. I think we should use `compare_and_set` from offchain workers API.

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,442 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```
tomusdrw: Do we have tests for this module? Or is it coming in a follow up PR?

cmichi: Will be a follow-up PR, I'm already working on it.

srml/im-online/src/lib.rs
cmichi: Created a follow-up (https://github.com/paritytech/substrate/issues/3143) and assigned to myself after private conversation.

core/primitives/src/offchain.rs
```diff
@@ -241,6 +276,12 @@ pub trait Externalities {
 	/// The transaction will end up in the pool and be propagated to others.
 	fn submit_transaction(&mut self, extrinsic: Vec<u8>) -> Result<(), ()>;
 
+	/// Returns information about the local node's network state.
+	fn network_state(&self) -> Result<OpaqueNetworkState, ()>;
+
+	/// Returns the locally configured authority public key, if available.
```
gavofyork: What is the "authority public key"? There are several keys in use for authorities now.

cmichi: There exists [a trait](https://github.com/paritytech/substrate/blob/master/core/offchain/src/lib.rs#L61-L65) for that with [this implementation](https://github.com/paritytech/substrate/blob/master/core/service/src/lib.rs#L885-L908), which I call to get the current authority public key. This whole part is currently getting reworked in https://github.com/paritytech/substrate/pull/3034, so I thought it's fine in master?

node/runtime/src/lib.rs
```diff
@@ -379,6 +379,20 @@ impl sudo::Trait for Runtime {
...
 }
 
+impl im_online::Trait for Runtime {
+	type AuthorityId = AuraId;
+	type Call = Call;
+	type Event = Event;
+	type SessionsPerEra = SessionsPerEra;
+	type UncheckedExtrinsic = UncheckedExtrinsic;
+
+	fn is_valid_authority_id(authority_id: &<Self as im_online::Trait>::AuthorityId) -> bool {
```
gavofyork: this is irregular. instead, create a new `trait IsMember<AuthorityId>`, have `Aura` and `Babe` implement it and have the `im_online::Trait` contain an extra item `type IsValidAuthorityId: IsMember<Self::AuthorityId>;`. here there should just be a `type IsValidAuthorityId = Aura;`.

core/offchain/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 description = "Substrate offchain workers"
 name = "substrate-offchain"
-version = "2.0.0"
+version = "2.1.0"
```
gavofyork: why? version should stay as blanket 2.0.0 until 2.0.0 is ready for release.

srml/im-online/src/lib.rs
```diff
@@ -0,0 +1,442 @@
...
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		/// Number of sessions per era.
+		const SessionsPerEra: session::SessionIndex = T::SessionsPerEra::get();
+
+		fn deposit_event<T>() = default;
+
+		fn heartbeat(
+			origin,
+			heartbeat: Heartbeat<T::BlockNumber, T::AuthorityId>,
+			_signature: Vec<u8>
```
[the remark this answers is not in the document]
cmichi: `_signature: Vec<u8>,` results in a number of errors, `_signature: Vec<u8>` compiles fine. I suspect that there might be an issue with the macro, are you aware of any issue?

core/network/src/service.rs
```diff
@@ -251,6 +256,12 @@ impl<B: BlockT + 'static, S: NetworkSpecialization<B>, H: ExHashT> NetworkWorker
 		self.network_service.user_protocol().num_connected_peers()
 	}
 
+	/// Returns the local external addresses.
+	pub fn external_addresses(&self) -> Vec<Multiaddr> {
+		let swarm = &self.network_service;
```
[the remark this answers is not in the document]
cmichi: To provide a stable getter for this property. There is an existing `NetworkState` struct, but it's API is highly unstable and it shouldn't be used.

srml/im-online/src/lib.rs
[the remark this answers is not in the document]
cmichi: Regarding checking: `validate_unsigned(…)` verifies it.

core/network/src/service.rs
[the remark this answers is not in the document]
cmichi: Good spotting, it was a survivor from an earlier version of the PR. I removed it now.

core/network/src/service.rs
```diff
@@ -520,6 +562,8 @@ enum ServerToWorkerMsg<B: BlockT, S: NetworkSpecialization<B>> {
 /// You are encouraged to poll this in a separate background thread or task.
 #[must_use = "The NetworkWorker must be polled in order for the network to work"]
 pub struct NetworkWorker<B: BlockT + 'static, S: NetworkSpecialization<B>, H: ExHashT> {
+	/// Updated by the `NetworkWorker` and loaded by the `NetworkService`.
```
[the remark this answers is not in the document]
cmichi: Here: https://github.com/paritytech/substrate/pull/3079/files#diff-a045635f3eec17c06b3c7ac7944acceaR500, do you have concerns regarding the name "loading" being associated with the method `load()`?

core/network/src/service.rs
```diff
@@ -86,6 +87,8 @@ impl ReportHandle {
 pub struct NetworkService<B: BlockT + 'static, S: NetworkSpecialization<B>, H: ExHashT> {
 	/// Number of peers we're connected to.
 	num_connected: Arc<AtomicUsize>,
+	/// The local external addresses.
+	external_addresses: Arc<Mutex<Vec<Multiaddr>>>,
```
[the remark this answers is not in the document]
cmichi: Thx, fixed.

core/primitives/src/offchain.rs
gavofyork: ok

dvc94ch: I started refactoring the offchain workers api (discussed with @tomusdrw) while @tomaka is refactoring the service factory/components. I'm on vacation till next week. The plan is to rebase #3034 on those PRs

Issue #3064, opened 2019-07-08
Aura block production stuck


pepyakin:
This commit:
https://github.com/paritytech/substrate/pull/3011/commits/d053556bde319156a8a13b79988e79547e3b4223 which is basically master + wasm executor stuff (which I feel confident about).

I tried to create a bunch of blocks with the following changes to increase the speed of block producing:

```diff
diff --git a/node/cli/src/chain_spec.rs b/node/cli/src/chain_spec.rs
index 78f74a6e2..2705d6819 100644
--- a/node/cli/src/chain_spec.rs
+++ b/node/cli/src/chain_spec.rs
@@ -266,7 +266,7 @@ pub fn testnet_genesis(
                        desired_seats: council_desired_seats,
                }),
                timestamp: Some(TimestampConfig {
-                       minimum_period: 2,                    // 2*2=4 second block time.
+                       minimum_period: 1,                    // 2*2=4 second block time.
                }),
                contracts: Some(ContractsConfig {
                        current_schedule: contracts::Schedule {
diff --git a/srml/aura/src/lib.rs b/srml/aura/src/lib.rs
index 56485b04c..722502f55 100644
--- a/srml/aura/src/lib.rs
+++ b/srml/aura/src/lib.rs
@@ -238,7 +238,7 @@ impl<T: Trait> Module<T> {
        pub fn slot_duration() -> T::Moment {
                // we double the minimum block-period so each author can always propose within
                // the majority of its slot.
-               <timestamp::Module<T>>::minimum_period().saturating_mul(2.into())
+               <timestamp::Module<T>>::minimum_period()
        }

        fn on_timestamp_set<H: HandleReport>(now: T::Moment, slot_duration: T::Moment) {
```

and it stuck two times. I am not sure if it reproduces with the normal settings.

First run stoped at 2099 and the second run at 699.

After a restart the block production doesn't continue. Here is the log:

```
^Cparity@qa-sync-1:~/pepyakin/substrate-sketch$ RUST_LOG=aura=debug ./target/release/substrate --dev
2019-07-08 17:11:36.891 main WARN substrate_cli  Running in --dev mode, RPC CORS has been disabled.
2019-07-08 17:11:36.891 main INFO node_cli  Substrate Node
2019-07-08 17:11:36.891 main INFO node_cli    version 2.0.0-d053556bd-x86_64-linux-gnu
2019-07-08 17:11:36.891 main INFO node_cli    by Parity Technologies, 2017-2019
[...]

PR #3083, opened 2019-07-10
Check that validator set is not empty at genesis

bkchr:


bkchr:
Fixes: https://github.com/paritytech/substrate/issues/3064

srml/session/src/lib.rs
```diff
@@ -316,6 +316,8 @@ decl_storage! {
 				let initial_validators = T::SelectInitialValidators::select_initial_validators()
 					.unwrap_or_else(|| config.keys.iter().map(|(ref v, _)| v.clone()).collect());
 
+				assert!(!initial_validators.is_empty(), "Empty validator set in genesis block!");
```
gavofyork: No panickers allowed in the runtime :-)

andresilva:
I don't think there's any need to bump `spec_version` or `impl_version`.

srml/session/src/lib.rs
andresilva: This panics but I think it's fine since it's in `add_extra_genesis` (i.e. it's not part of block execution in runtime)

PR #3093, opened 2019-07-10
 Convert unnecessary storage item to static.

gavofyork:
There was no need to keep minimum block time as a storage item. This just switches it to a static Getter.

### DIFF
--- a/srml/babe/src/lib.rs
+++ b/srml/babe/src/lib.rs
@@ -21,7 +21,7 @@
 pub use timestamp;
 
 use rstd::{result, prelude::*};
-use srml_support::{decl_storage, decl_module, StorageValue, traits::FindAuthor};
+use srml_support::{decl_storage, decl_module, StorageValue, traits::FindAuthor, traits::Get};
 use timestamp::{OnTimestampSet, Trait};
 use primitives::{generic::DigestItem, traits::{SaturatedConversion, Saturating, RandomnessBeacon}};
 use primitives::ConsensusEngineId;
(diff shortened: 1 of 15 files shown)


srml/babe/src/lib.rs
```diff
@@ -193,7 +193,7 @@ impl<T: Trait> Module<T> {
 	pub fn slot_duration() -> T::Moment {
 		// we double the minimum block-period so each author can always propose within
 		// the majority of their slot.
-		<timestamp::Module<T>>::minimum_period().saturating_mul(2.into())
+		<T as timestamp::Trait>::MinimumPeriod::get().saturating_mul(2.into())
```
andresilva: Missing import of `srml_support::traits::Get` in this file I think

shawntabrizi: I patched this real quick

srml/timestamp/src/lib.rs
```diff
@@ -208,23 +210,30 @@ pub trait Trait: system::Trait {
 
 	/// Something which can be notified when the timestamp is set. Set this to `()` if not needed.
 	type OnTimestampSet: OnTimestampSet<Self::Moment>;
+
+	/// The minimum period between blocks. Beware that this is different to the *expected* period
+	/// that the block production apparatus provides. Your chosen consensus system will generally
+	/// work with this to determine a sensible block time. e.g. For Aura, it will be double this
+	/// period on default settings.
+	type MinimumPeriod: Get<Self::Moment>;
```
bkchr: We still need to export this value in `decl_module!`.

gavofyork:
looks good.

andresilva:
Integration test failed due to some unrelated docker issue, restarted.

PR #3094, opened 2019-07-10
Introduce vesting offsets

### LINKED ISSUE #3090: Allow for partially vested balances at genesis
According to my understanding of the vesting genesis configuration in the Balance module: 
https://github.com/paritytech/substrate/blob/4e23e922eb56ef0fc4837e8d9b970ddf8a023d5b/srml/balances/src/lib.rs#L317-L334

It is not possible for an account to be _partially_ vested at genesis as this would require the `begin` parameter to be a negative integer. There are some use cases where an address may be partially vested at genesis, with only a portion of its genesis balance being released according to the vesting schedule.

One solution would be to allow to specify a separate amount as the `vesting_balance` which can be up to the entire genesis balance of that account. The other solution would be as I mention to allow the `begin` parameter start at a non-existent block in past so that at 0 it has already "vested" some of its balance.

gavofyork:
Closes #3090


lsaether:
The change solves the issue I reported in #3090 to allow for partially vested balances at genesis, but still has the complication reported above ^^

The most minimal fix would be to allow `begin` to be an `i64` which would allow the `offset` to be lower than the `balance` at genesis by allowing `begin` to be negative. This might be less safe.

Another change would involve changing the `VestingSchedule` struct and the logic for `locked_at`
```rust
pub struct VestingSchedule<Balance, BlockNumber>  {
    pub vested_amount: Balance,
    pub begin: BlockNumber,
    pub length: BlockNumber,
}
```
which would then just calculate the slope on the fly whenever `vested_balance` is called for an account.

gavofyork:
The issue is simply that there's too many degrees of freedom to encoding in a straight y=mx+c curve.

I'm not planning on making the curve any more complex and this feature is non-negotiable, so it's just up to the user to be aware of how the two parameters play together in the genesis config.

I could maybe require that only one is used, I guess.

shawntabrizi:
@gavofyork @lsaether I have introduced a small-ish rewrite the logic around vesting. The genesis config API is unchanged, but the `VestingSchedule` which is generated now looks like:

```
pub struct VestingSchedule<Balance, BlockNumber> {
	/// Locked amount at genesis.
	pub locked: Balance,
	/// Amount that gets unlocked every block from genesis.
	pub per_block: Balance,
	/// Starting block for vesting.
	pub starting_block: BlockNumber,
}
```

So rather than calculating an `offset`, we keep track of the `locked` amount and the block when vesting starts.

This means that a user can will have access to any amount of funds which is greater than the `locked` amount. Funds from `locked` start to vest only after `starting_block`, with `per_block` being unlocked per block.

This means a user can have some liquid funds on genesis and also have a vesting schedule that starts after block 0. Furthermore, any user can accumulate funds in their account before vesting starts, and spend those funds, whereas before a user might not be able to spend additional funds they receive because offset was "extra negative".

I will write tests for some of these extra scenarios, but wanted you to look at the design now.

shawntabrizi:
This is ready for review

gavofyork:
Someone else should review. @keorn ?

Issue #2636, opened 2019-05-20
CLI flag to force all native execution mode


xlc:
If will be good if there is a flag to do something similar to `--block-construction-execution=native --importing-execution=native --other-execution=native --syncing-execution=native` but also force always native mode regardless the runtime version.

This is intent for debugging runtime. Breakpoints are not getting hit due to many reasons and one of them is the wasm executor is been used instead of the native executor.

I don't care if the native version is compatible or not (maybe print some warning if not compatible), I just want to debug it.

It is already hard enough to set breakpoints on external crate due to potentially mismatched git version (IDE present me one version and cargo use another one and hence breakpoint not working). It will be good if I can disable wasm executor to make debugging easier.

I now know how to do it properly, but I cannot teach everyone the all the approaches required to diagnostic why breakpoint is not working.

fzzr-:
I'm interested this feature too.

PR #3098, opened 2019-07-11
Cleanup CLI and introduce `execution-all`

bkchr:
`--execution-all` sets the execution strategies for all available
execution contexts.

Fixes: https://github.com/paritytech/substrate/issues/2636

### DIFF
--- a/core/cli/src/params.rs
+++ b/core/cli/src/params.rs
@@ -33,13 +33,12 @@ macro_rules! impl_get_log_filter {
 
 arg_enum! {
 	/// How to execute blocks
-	#[derive(Debug, Clone)]
+	#[derive(Debug, Clone, Copy)]
 	pub enum ExecutionStrategy {
 		Native,
 		Wasm,
 		Both,
 		NativeElseWasm,
-		NativeWhenPossible,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


core/cli/src/params.rs
```diff
@@ -227,63 +225,81 @@ pub struct TransactionPoolParams {
...
 			default_value = r#""NativeElseWasm""#
 		)
 	)]
-	pub syncing_execution: ExecutionStrategy,
+	pub execution_syncing: ExecutionStrategy,
 
 	/// The means of execution used when calling into the runtime while importing blocks.
 	#[structopt(
-		long = "importing-execution",
+		long = "execution-import-block",
```
andresilva: imo it's a bit weird to have the order like this in the flag `--execution-block-import`, can we keep the originals?

core/cli/src/params.rs
```diff
@@ -227,63 +225,81 @@ pub struct TransactionPoolParams {
...
 			case_insensitive = "true",
-			default_value = r#""Wasm""#
+			default_value = r#""Native""#
+		)
+	)]
+	pub execution_other: ExecutionStrategy,
+
+	/// The execution strategy that should be used by all execution contexts.
+	#[structopt(
+		long = "execution-all",
```
andresilva: maybe just call this one `--execution` since it conflicts with all the other ones.

core/cli/src/params.rs
bkchr: First I thought that these `long` names would determine the order in the cli, after that I realized that actually the field name determines the order. So, I could revert this. However, I think with this order they look more uniform and the `execution` is more like a namespace.

Kianenigma:
cc @mattrutherford you probbaly can use this.

Issue #3099, opened 2019-07-11
Switch to stable Futures


tomaka:
Instead of using `futures 0.1`, let's use futures from the Rust standard library!

tomaka:
Note that after async/await get stable, the general intention is that it should be possible to use non-'static futures.

In other words, we could write something like this:

```rust
impl Foo {
    // the returned `Future` borrows `self`
    fn next_item(&mut self) -> impl Future { ... }
}
```

This will considerably reduce the need for the `Stream` trait, which isn't stable, to point where we might be able to eliminate its use.

tomaka:
Stables futures don't have any equivalent to `loop_fn`, which grandpa unfortunately uses.

bkchr:
Why not just implement a custom `Future`? Will make the code more readable.

kigawas:
And consider using [tokio 0.2](https://docs.rs/tokio/0.2.0-alpha.6/tokio/index.html) when it's ready

expenses:
Still waiting on:

* hyper
* libp2p
* jsonrpc (or maybe a switch to jsonrpsee)

PR #3100, opened 2019-07-11
Switch the telemetry to new futures

tomaka:
cc #3099 

The majority of the switch to new futures is straight-forward, but for the telemetry it is pretty much not straight-forward.
Since libp2p is using futures v0.1, and the telemetry has a wrapper that guarantees that one telemetry write equals one WebSocket frame, we need to have a compatibility shim.

After this PR, the API of `substrate-telemetry` uses new futures.

### DIFF
--- a/core/telemetry/src/worker.rs
+++ b/core/telemetry/src/worker.rs
@@ -27,12 +27,11 @@
 //!
 
 use bytes::BytesMut;
-use futures::prelude::*;
+use futures::compat::Compat01As03Sink;
 use libp2p::{core::transport::OptionalTransport, core::ConnectedPoint, Multiaddr, Transport, wasm_ext};
 use log::{trace, warn, error};
 use slog::Drain;
-use std::{io, time};
-use tokio_io::AsyncWrite;
+use std::{io, pin::Pin, task::Context, task::Poll, time};
 
 mod node;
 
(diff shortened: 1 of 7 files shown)


core/telemetry/src/worker/node.rs
```diff
@@ -169,36 +168,47 @@ where TTrans: Clone, TTrans::Output: Sink<SinkItem = BytesMut, SinkError = TSink
...
-			} else if self.need_flush && self.sink.poll_complete()?.is_ready() {
-				self.need_flush = false;
+				let item_len = item.len();
+				if let Err(err) = Sink::start_send(Pin::new(&mut self.sink), item) {
+					return Poll::Ready(Err(err))
+				}
+				trace!(target: "telemetry", "Successfully sent {:?} bytes message to {}",
+					item_len, my_addr);
+				self.need_flush = true;
+
```
bkchr:
```suggestion
```

core/telemetry/src/worker/node.rs
```diff
@@ -169,36 +168,47 @@ where TTrans: Clone, TTrans::Output: Sink<SinkItem = BytesMut, SinkError = TSink
...
 				}
 
-			} else if self.need_flush && self.sink.poll_complete()?.is_ready() {
-				self.need_flush = false;
+				let item_len = item.len();
+				if let Err(err) = Sink::start_send(Pin::new(&mut self.sink), item) {
+					return Poll::Ready(Err(err))
+				}
+				trace!(target: "telemetry", "Successfully sent {:?} bytes message to {}",
+					item_len, my_addr);
```
bkchr:
```suggestion
					item_len, my_addr
				);
```

core/telemetry/src/worker/node.rs
```diff
@@ -169,36 +168,47 @@ where TTrans: Clone, TTrans::Output: Sink<SinkItem = BytesMut, SinkError = TSink
...
+					return Poll::Pending
 				}
 
-			} else if self.need_flush && self.sink.poll_complete()?.is_ready() {
-				self.need_flush = false;
+				let item_len = item.len();
+				if let Err(err) = Sink::start_send(Pin::new(&mut self.sink), item) {
+					return Poll::Ready(Err(err))
+				}
+				trace!(target: "telemetry", "Successfully sent {:?} bytes message to {}",
```
bkchr:
```suggestion
				trace!(
					target: "telemetry", "Successfully sent {:?} bytes message to {}",
```

core/telemetry/src/worker.rs
```diff
@@ -58,19 +57,29 @@ pub struct TelemetryWorker {
 /// The pile of libp2p transports.
 #[cfg(not(target_os = "unknown"))]
 type WsTrans = libp2p::core::transport::timeout::TransportTimeout<
-	libp2p::core::transport::OrTransport<
-		libp2p::core::transport::map::Map<
-			OptionalTransport<wasm_ext::ExtTransport>,
-			fn(wasm_ext::Connection, ConnectedPoint) -> StreamSink<wasm_ext::Connection>
+	libp2p::core::transport::map::Map<
```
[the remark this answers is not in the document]
tomaka: The problem here is that I'm using libp2p in a way it wasn't really designed for, but theoretically I could add some utility that allows using a `Box` here instead.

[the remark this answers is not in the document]
bkchr: It is okay, probably more rust faults of not having `impl Something`.

PR #3102, opened 2019-07-11
Extensible transactions (and tips)

gavofyork:
This removes a lot of logic from the executive into the modules, which is good because it should never have been there in the first place. I also added tips to the payment mechanism, which is trivial now.

I removed the historical variants of `UncheckedExtrinsic` and `MakePayment`.

This changes the transaction format; for now it is:
- Address
- Signature
- Era
- Nonce
- Tip (Balance)
- Call

The signature payload is:
- Call
- Era
- Nonce
- Tip (Balance)
- Checkpoint hash (or genesis hash for immortal transactions)

TODO
- [x] Make sure the nonce and tip is compact encoded.
- [x] Make tests work again. @Kianenigma #3138 
- [x] Move the block-weight-check into an extensibility trait. @Kianenigma #3115 
- [x] Bring _weight-to-fee_ stuff back. @Kianenigma #2854 
- [x] Fix extrinsic priority based on tip + type of the transaction. @Kianenigma #3138 

Maybe a different PR, maybe not.
- [x] Charge transaction size independently from weight. (done in #3138; needs #2854)
- [x] Limit block size separately from block weight. (done in #3138)
- [ ] Add more tests for `subkey` and `factory` to make sure they are always creating txs with correct, decodable and verifiable signatures. 

Maybe do, maybe don't:
- [x] Consider moving all the signing stuff into an extensibility trait (won't do - signing is too intrinsic to a transaction).
- [x] Consider moving the era stuff into an extensibility trait (done).
- [x] Consider moving the validate-unsigned stuff into an extensibility trait (won't do - validate-unsigned is a per-module thing, which doesn't work with an attribute-system like `SignedExtension`).


gavofyork:
CC @Kianenigma

srml/system/src/lib.rs
```diff
@@ -774,6 +774,8 @@ impl<T: Trait> rstd::fmt::Debug for CheckNonce<T> {
 
 impl<T: Trait> SignedExtension for CheckNonce<T> {
 	type AccountId = T::AccountId;
+	type AdditionalSigned = ();
+	fn additional_signed(&self) -> Result<(), &'static str> { Ok(()) }
```
Kianenigma: probably easier to put this default impl into the trait definition and not even code it here.

Kianenigma:
```diff
- type AdditionalSigned: Encode;
+ type AdditionalSigned: Encode + Default;

+ fn additional_signed(&self) -> Result<Self::AdditionalSigned, &'static str> { Ok(Default::default()) }
``

Comment (gavofyork) [reply srml/system/src/lib.rs]: This is no reason to force `AdditionalSigned` to be `Default`.

File: core/sr-primitives/src/traits.rs [hunk 2]
```diff
@@ -749,6 +750,184 @@ impl<T: BlindCheckable, Context> Checkable<Context> for T {
...
+		[$($direct:ident)+] [$first:ident $($rest:ident)*]
+		;
+		[$($index:tt,)+] [$first_index:tt, $($rest_index:tt,)*]
+	) => {
+		tuple_impl_indexed!([$($direct)+] ; [$($index,)+]);
+		tuple_impl_indexed!([$($direct)+ $first] [$($rest)*] ; [$($index,)+ $first_index,] [$($rest_index,)*]);
+	};
+}
+
+// TODO: merge this into `tuple_impl` once codec supports `trait Codec` for longer tuple lengths.
```
Kianenigma: have to make an issue for this

node/cli/src/factory_impl.rs
```diff
@@ -132,16 +140,14 @@ impl RuntimeAdapter for FactoryState<Number> {
...
 				BalancesCall::transfer(
-					indices::address::Address::Id(
-						destination.clone().into()
-					),
+					indices::address::Address::Id(destination.clone().into()),
 					(*amount).into()
 				)
 			)
-		}, key, &prior_block_hash, phase)
+		}, key, (prior_block_hash.clone(), (), (), ()))
```
Kianenigma: This is likely to be wrong. Double-check and add a test.

node/executor/src/lib.rs
```diff
@@ -932,97 +945,28 @@ mod tests {
...
-	fn weight_multiplier_decreases_on_small_block() {
-		let mut t = new_test_ext(COMPACT_CODE, false);
-
-		let mut prev_multiplier = WeightMultiplier::default();
-
-		runtime_io::with_externalities(&mut t, || {
-			assert_eq!(System::next_weight_multiplier(), prev_multiplier);
+		// NOTE: This assumes that system::remark has the default.
+		let num_to_exhaust = multiplier_ideal() * 2 / SimpleDispatchInfo::default().weigh_data(());
+		println!("++ Generating {} transactions to fill {} weight units", num_to_exhaust, multiplier_ideal() * 2);
```
Kianenigma: These are not leftover. Since these tests are quite hacky it is good to have them there in case of failure.

node/runtime/src/lib.rs
```diff
@@ -116,6 +117,8 @@ pub const DAYS: Moment = HOURS * 24;
 
 parameter_types! {
 	pub const BlockHashCount: BlockNumber = 250;
+	pub const MaximumBlockWeight: Weight = 4 * 1024;
```
Kianenigma: probably increase these further to be more realistic.

srml/support/Cargo.toml
```diff
@@ -20,7 +20,7 @@ bitmask = { version = "0.5", default-features = false }
 
 [dev-dependencies]
 pretty_assertions = "0.6.1"
-srml-system = { path = "../system", default-features = false }
```
Kianenigma: I don't know why default was disabled for test build + it was causing errors. Hence, gone.

srml/support/src/dispatch.rs
```diff
@@ -1117,14 +1110,38 @@ macro_rules! decl_module {
...
+						);
+						return $crate::dispatch::DispatchInfo { weight, class };
+					}
+					if let $call_type::__PhantomItem(_, _) = self { unreachable!("__PhantomItem should never be used.") }
+				)*
+				// Defensive only: this function must have already returned at this point.
+				// all dispatchable function will have a weight which has the `::default`
+				// implementation of `SimpleDispatchInfo`. Nonetheless, we create one if it does
+				// not exist.
+				let weight = <dyn $crate::dispatch::WeighData<_>>::weigh_data(
```
Kianenigma: @bkchr is this sane and what you recommended?

srml/system/src/lib.rs
```diff
@@ -76,16 +76,20 @@ use serde::Serialize;
...
+use primitives::transaction_validity::{ValidTransaction, TransactionPriority, TransactionLongevity};
+use primitives::traits::{self, CheckEqual, SimpleArithmetic, Zero, SignedExtension, Convert,
 	SimpleBitOps, Hash, Member, MaybeDisplay, EnsureOrigin, CurrentHeight, BlockNumberToHash,
-	MaybeSerializeDebugButNotDeserialize, MaybeSerializeDebug, StaticLookup, One, Bounded, Lookup,
-	Zero, Convert,
-}};
+	MaybeSerializeDebugButNotDeserialize, MaybeSerializeDebug, StaticLookup, One, Bounded,
+	Lookup, DispatchError, SaturatedConversion,
+};
+use primitives::Perbill;
```
Kianenigma: This currently unused because: I want to get rid of all the `max_weight / 4` and make it a `Perbill` associated const in system.

Kianenigma:
I put some leftover tasks/todos/thoughts there myself. Except for the last one about removing `x / 4` the rest looks okay.

srml/system/src/lib.rs
Kianenigma: `type IdealWeightRatio = Get<Perbill>`

and wrap the multiplication in a public function.

gavofyork:
@Kianenigma don't forget the issue for the TODO. then once it's reviewed by a third person is can be merged.

srml/system/src/lib.rs
Kianenigma: Will be done in another PR. first, want to check if it is the correct way to go since it requires a daunting refactor.

core/sr-primitives/src/traits.rs
Kianenigma: https://github.com/paritytech/substrate/issues/3152

Kianenigma:
Merging by the end of today if no one else reacts.

PR #3103, opened 2019-07-11
Switch the client to new futures

tomaka:
cc #3099 

Switches `substrate-client` to new futures. The two changes are the notification channels and the fetcher.

### DIFF
--- a/core/finality-grandpa/src/until_imported.rs
+++ b/core/finality-grandpa/src/until_imported.rs
@@ -92,9 +92,10 @@ impl<Block: BlockT, Status, I: Stream, M> UntilImported<Block, Status, I, M>
 
 		let check_pending = Interval::new(now + CHECK_PENDING_INTERVAL, CHECK_PENDING_INTERVAL);
 		UntilImported {
-			import_notifications: (Box::new(
-				import_notifications.map::<_, fn(_) -> _>(|v| Ok::<_, ()>(v)).compat()
-			) as Box<dyn Stream<Item = _, Error = _> + Send>).fuse(),
+			import_notifications: {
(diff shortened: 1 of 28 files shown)


core/finality-grandpa/src/until_imported.rs
```diff
@@ -64,7 +65,7 @@ pub(crate) trait BlockUntilImported<Block: BlockT>: Sized {
 
 /// Buffering imported messages until blocks with given hashes are imported.
 pub(crate) struct UntilImported<Block: BlockT, Status, I, M: BlockUntilImported<Block>> {
-	import_notifications: Fuse<ImportNotifications<Block>>,
+	import_notifications: Fuse<Box<dyn Stream<Item = BlockImportNotification<Block>, Error = ()> + Send>>,
```
bkchr: Can you not just put the `Fuse` into the `Box` and prevent the huge number of type annotations below?

core/finality-grandpa/src/until_imported.rs
```diff
@@ -91,7 +92,9 @@ impl<Block: BlockT, Status, I: Stream, M> UntilImported<Block, Status, I, M>
 
 		let check_pending = Interval::new(now + CHECK_PENDING_INTERVAL, CHECK_PENDING_INTERVAL);
 		UntilImported {
-			import_notifications: import_notifications.fuse(),
+			import_notifications: (Box::new(
+				import_notifications.map::<_, fn(_) -> _>(|v| Ok::<_, ()>(v)).compat()
```
bkchr: If above does not work, please format this better^^

core/finality-grandpa/src/until_imported.rs
tomaka: No, because we're calling `is_done()` somewhere else.

PR #3105, opened 2019-07-11
 Remove `sync::Context` trait.

twittner:
Context: https://github.com/paritytech/cumulus/issues/7

As part of an effort to add support for `BlockAnnounce`ment validations which include candidate messages, it was suggested to move the block (announcement) validation logic to consensus-common. As a first step this PR attempts to make `sync.rs` more re-usable by removing the `Context` trait object parameter which is currently passed to nearly every method of `ChainSync`, which introduces side-effecting I/O actions everywhere. With this PR `ChainSync` only contains state which is updated by invoking the various callback methods (`on_*`) and actionable items are returned as regular results from method calls, often iterators yielding requests that should be issued to peers. It is up to the caller to handle these in an appropriate way and currently `Protocol` will send those as messages over the same channel which it previously provided to `ChainSync`. The methods `ChainSync::maintain_sync` and `ChainSync::tick` are gone. Progress is maintained by retrieving the pending requests as part of `Protocol::poll`.

If this is an acceptable approach and the PR is merged, as a second step it needs to be seen, which parts should be moved to consensus-common. (CC @bkchr).


core/network/src/protocol/sync.rs
```diff
@@ -24,154 +24,149 @@
...
-	pub recently_announced: VecDeque<B::Hash>,
+	/// A queue of blocks that this peer has announced to us, should only
+	/// contain `ANNOUNCE_HISTORY_SIZE` entries.
+	pub recently_announced: VecDeque<B::Hash>
 }
 
 /// The sync status of a peer we are trying to sync with
 #[derive(Debug)]
-pub(crate) struct PeerInfo<B: BlockT> {
+pub struct PeerInfo<B: Block> {
```
tomaka: The entire code base uses `BlockT`, so let's keep `BlockT`?

arkpar:
Looks good.

PR #3106, opened 2019-07-11
Operational Transaction.

### LINKED ISSUE #3092: Operational Transactions
These transactions will be 
- Signed.
- Must be charged fees as a normal tx.
- Must make sure that they have precedence in inclusion
- Must make sure that the block weight limit __NOT__ will cause them to be rejected.

Needs:
- `Weighable` trait to return the priority. `fn priority<B>(&self, encodec_len: usize, tip: B)`
- `Weighable` to define a function which indicates block fullness. Then we simply return false if we observe an operational tx. 

Not much work. Best to do it after https://github.com/paritytech/substrate/pull/2930 and https://github.com/paritytech/substrate/pull/2854 are merged.

Kianenigma:
closes #3092 

TODOs:

- Now that the `Weighable` is actually doing more than merely communicating weight info up to the rest of the stack, it should probably be renamed to sth more general.  Likewise the `#[weight]` thingy.

By definition, it can be applied to anything which lives in `decl_module`, it is _static information_ (stateless except for input parameters it receives) and it _describes the dispatch function_.`#[meta =$x]` and `pub trait DispatchMetaInfo`?

### DIFF
--- a/core/sr-primitives/src/generic/checked_extrinsic.rs
+++ b/core/sr-primitives/src/generic/checked_extrinsic.rs
@@ -19,6 +19,7 @@
 
 use crate::traits::{self, Member, SimpleArithmetic, MaybeDisplay};
 use crate::weights::{Weighable, Weight};
+use crate::transaction_validity::TransactionPriority;
 
 /// Definition of something that the external world might want to say; its
 /// existence implies that it has been checked and is good, particularly with
@@ -62,5 +63,11 @@ where
 {
 	fn weight(&self, len: usize) -> Weight {
 		self.function.weight(len)
+	}
+	fn priority(&self, len: usize) -> TransactionPriority {
+		self.function.priority(len)
+	}
+	fn is_block_full(&self, block_weight: Weight, len: usize) -> bool {
+		self.function.is_block_full(block_weight, len)
 	}
 }
--- a/core/sr-primitives/src/weights.rs
+++ b/core/sr-primitives/src/weights.rs
@@ -59,8 +59,7 @@ pub enum TransactionWeight {
 	/// Basic weight (base, byte).
 	/// The values contained are the base weight and byte weight respectively.
 	///
-	/// The priority of this transaction will be proportional to its computed weight, w.r.t. to the
-	/// total available weight.
+	/// The priority of this transaction will be proportional to its computed weight.
 	Basic(Weight, Weight),
 	/// Operational transaction. These are typically root transactions for operational updates,
 	/// runtime code changes or consensus reports through inherents. These transactions are still
--- a/core/sr-primitives/src/weights.rs
+++ b/core/sr-primitives/src/weights.rs
@@ -24,9 +24,14 @@
 //! Note that the decl_module macro _cannot_ enforce this and will simply fail
 //! if an invalid struct is passed in.
 
+use crate::transaction_validity::TransactionPriority;
+
 /// The final type that each `#[weight = $x:expr]`'s
 /// expression must evaluate to.
 pub type Weight = u32;
+
+/// The maximum possible weight of a block.
+pub const MAX_TRANSACTIONS_WEIGHT: u32 = 4 * 1024 * 1024;
 
 /// A `Call` enum (aka transaction) that can be weighted using the custom weight attribute of
 /// its dispatchable functions. Is implemented by default in the `decl_module!`.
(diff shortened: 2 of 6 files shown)


core/sr-primitives/src/generic/checked_extrinsic.rs
```diff
@@ -63,4 +64,10 @@ where
 	fn weight(&self, len: usize) -> Weight {
 		self.function.weight(len)
 	}
+	fn priority(&self, len: usize) -> TransactionPriority {
+		self.function.priority(len)
+	}
+	fn is_block_full(&self, block_weight: Weight, len: usize) -> bool {
```
Kianenigma: okay this is not a great name since it is misleading. Should be something along the lines of `xt.will_cause_full_block() -> bool`

srml/support/src/dispatch.rs
```diff
@@ -1126,6 +1127,22 @@ macro_rules! decl_module {
 					$call_type::__PhantomItem(_, _) => { unreachable!("__PhantomItem should never be used.") },
 				}
 			}
+			fn priority(&self, _len: usize) -> $crate::dispatch::TransactionPriority {
+				match self {
+					$( $call_type::$fn_name(..) => $crate::dispatch::Weighable::priority(&$weight, _len), )*
+					$call_type::__PhantomItem(_, _) => { unreachable!("__PhantomItem should never be used.") },
```
DemiMarie-parity: This should be ensured at compile-time.  For example, `__PhantomItem` could contain `Infallible`, and then the `unreachable!` could be replaced by a pattern-match against it.

srml/support/src/dispatch.rs
```diff
@@ -1126,6 +1127,22 @@ macro_rules! decl_module {
...
+				}
+			}
+			fn is_block_full(&self, _block_weight: $crate::dispatch::Weight, _len: usize) -> bool {
+				match self {
+					$( $call_type::$fn_name(..) => $crate::dispatch::Weighable::is_block_full(
+						&$weight,
+						_block_weight,
+						_len,
+					), )*
+					$call_type::__PhantomItem(_, _) => { unreachable!("__PhantomItem should never be used.") },
```
DemiMarie-parity: Same

core/sr-primitives/src/weights.rs
```diff
@@ -46,22 +58,35 @@ pub trait Weighable {
...
+	fn priority(&self, len: usize) -> TransactionPriority {
+		match self {
+			TransactionWeight::Basic(_, _) => self.weight(len) as TransactionPriority,
+			TransactionWeight::Operational(_, _) => TransactionPriority::max_value(),
+		}
+	}
+
+	fn is_block_full(&self, block_weight: Weight, len: usize) -> bool {
+		match self {
+			TransactionWeight::Basic(_, _) => self.weight(len) + block_weight > MAX_TRANSACTIONS_WEIGHT,
```
DemiMarie-parity:
```suggestion
			TransactionWeight::Basic(_, _) => self.weight(len).checked_add(block_weight).map(|weight| weight > MAX_TRANSACTIONS_WEIGHT).unwrap_or(true),
```

node/runtime/src/lib.rs
```diff
@@ -70,7 +70,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 109,
```
DemiMarie-parity:
```suggestion
	spec_version: 111,
```

srml/support/src/dispatch.rs
```diff
@@ -1754,10 +1783,15 @@ mod tests {
 
 	#[test]
 	fn weight_should_attach_to_call_enum() {
-		// max weight. not dependent on input.
-		assert_eq!(Call::<TraitImpl>::weighted().weight(100), 3 * 1024 * 1024);
+		// Operational tx.
+		assert_eq!(Call::<TraitImpl>::security_tx().weight(100), 0);
+		assert_eq!(Call::<TraitImpl>::security_tx().priority(100), u64::max_value());
+		// one simply does not return false of the transaction type is operational!.
```
DemiMarie-parity:
```suggestion
		// Operational transactions are never denied no matter how full
		// the block is.
```

srml/support/src/dispatch.rs
Kianenigma: @bkchr this pattern can be replaced probably everywhere in storage macros? I borrowed the `unreachable!()` from other places in the code.

srml/support/src/dispatch.rs
Kianenigma: fair enough. I am personally pretty much okay with having some humor/informal statements as long as it is in test code :p

PR #3108, opened 2019-07-11
Switch the peerset to new futures

tomaka:
cc #3103 

Not much to say. 1:1 translation.

### DIFF
--- a/core/peerset/src/lib.rs
+++ b/core/peerset/src/lib.rs
@@ -20,10 +20,11 @@
 mod peersstate;
 
 use std::{collections::{HashSet, HashMap}, collections::VecDeque, time::Instant};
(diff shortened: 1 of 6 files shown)


core/peerset/src/lib.rs
```diff
@@ -497,10 +508,8 @@ mod tests {
 		peerset
 	}
 
-	fn next_message(peerset: Peerset) -> Result<(Message, Peerset), ()> {
-		let (next, peerset) = peerset.into_future()
-			.wait()
-			.map_err(|_| ())?;
+	fn next_message(mut peerset: Peerset) -> Result<(Message, Peerset), ()> {
```
DemiMarie-parity: Hopefully we can get rid of this at some point, but that is not an issue here.

Issue #3033, opened 2019-07-05
Cannot add custom rpc without hacking substrate-servife crate


guoxbin:
1. cannot replace FullComponents with a custom one because ServiceFactory traits hold a FullService type bounded with ServiceTrait<FullComponents<Self>>

2. cannot add extra handler because rpc::rpc_handler receive 4 handlers(state, chain, author, system) which are not unextendible

bkchr:
Yeah, this is currently not supported.
@tomusdrw is that anything that will come in the near future?

tomusdrw:
I'll have a look today, cause it's long overdue. Same issues were raised in #2263, so let's continue there.

PR #3109, opened 2019-07-12
Custom RPC implementation for `node`.

### LINKED ISSUE #1096: Provide example of adding an API and using it in the node
_Originally posted by @gavofyork in https://github.com/paritytech/substrate/pull/1094#issuecomment-437364119_

### LINKED ISSUE #2263: RPC to query nonce value includes one in the tx pool
It is possible to query the nonce value for an account but it does not consider the transactions in the transaction pool. This means the client side have to implement nonce method to take account of sent but unconfirmed transactions.

It will be better if there is a RPC method give highest nonce value including the one in the transaction queue.

Something similar to `web3.eth.getTransactionCount("0xf82e...", "pending");`

tomusdrw:
Closes #2263 #3033
Closes #1096 

This PR introduces a way to customize and extend existing set of core RPCs with some additional methods that might be more runtime-specific.
It relies on not-yet-merged https://github.com/paritytech/jsonrpc/pull/447/ feature in JSON-RPC repo.

Extension is done inside `ServiceFactory`, and this breaks backward compatibility, since now you need to at least define an empty extension:
`type RpcExtensions = () {}`

As an example of custom RPC in `node` I've implemented `account_nextNonce` that takes into account nonces of transactions in the pending transaction pool (#2263).
Since current state nonce is only stored in the `system` module, I've added a custom RuntimeApi to expose that to rust code (#1096).

Additionality, the PR introduces `node-testing` crate that contains a `test-client` running on top of `node-runtime`.

Submitting as draft, as we still need to wait for new release of `jsonrpc`.


dvc94ch:
If you rebase the test should pass... Denis fixed this in the gitlab.yml file.

Tresdin-commander:
mark！

gavofyork:
one file left.

PR #3110, opened 2019-07-12
Logs will be saved to the job artifacts only in case of failure.

TriplEight:
Logs for `test-linux-stable-int` CI job are too massive (161 Mb now). 
Now they are saved to the job artifacts only in case of failure. 
The failure is also listed to the CI's UI after the job is finished.

An example of the failed job: https://gitlab.parity.io/parity/substrate/-/jobs/183680

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -154,13 +154,20 @@ test-linux-stable-int:
   <<:                              *test-linux
   except:
     refs:
-      - /^v[0-9]+\.[0-9]+.*$/        # i.e. v1.0, v2.1rc1
+      - /^v[0-9]+\.[0-9]+.*$/      # i.e. v1.0, v2.1rc1
     variables:
       - $DEPLOY_TAG
   script:
-    - time RUST_LOG=sync=trace,consensus=trace,client=trace,state-db=trace,db=trace,forks=trace,state_db=trace,storage_cache=trace
-        cargo test -p node-cli --release --verbose --locked -- --ignored --test-threads=1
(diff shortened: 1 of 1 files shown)


bkchr:
Does this also works when we encounter a time out?

TriplEight:
@bkchr should work it will be considered as failure by GitLab, never tried though.

.gitlab-ci.yml
```diff
@@ -154,13 +154,24 @@ test-linux-stable-int:
...
-    - time RUST_LOG=sync=trace,consensus=trace,client=trace,state-db=trace,db=trace,forks=trace,state_db=trace,storage_cache=trace
-        cargo test -p node-cli --release --verbose --locked -- --ignored --test-threads=1
+    - echo "___Logs will be partly shown at the end in case of failure.___"
+    - echo "___Full log will be saved to the job artifacts only in case of failure.___"
+    - RUST_LOG=sync=trace,consensus=trace,client=trace,state-db=trace,db=trace,forks=trace,state_db=trace,storage_cache=trace
+        time cargo test -p node-cli --release --verbose --locked -- --ignored --test-threads=1
+        &> ${CI_COMMIT_SHORT_SHA}_int_failure.log
     - sccache -s
+  after_script:
+    - awk '/FAILED/,0' ${CI_COMMIT_SHORT_SHA}_int_failure.log
```
bkchr: Why do we need to add the commit hash to the name? What happens when we have 2 prs with the same hash?

TriplEight: nothing happens, it's just to separate those files when downloaded to your PC.

PR #3112, opened 2019-07-12
Make parameter types implementation more flexible

bkchr:


bkchr:
@DemiMarie-parity this pr did not alter any logic or fixed any bugs in the runtime. Given this fact, we don't require any update of the `impl_version`.

PR #3114, opened 2019-07-14
srml-contracts: Update complexity.md

pepyakin:
Update COMPLEXITY.md after #3047. Also add a note on initalization of the execution context.
(Plus a tiny refactoring)


srml/contracts/src/exec.rs
```diff
@@ -366,7 +366,7 @@ where
 							ctx: &mut nested,
 							caller: self.self_account.clone(),
 							value_transferred: value,
-							timestamp: timestamp::Module::<T>::now(),
+							timestamp: <timestamp::Module<T>>::now(),
```
jimpo: Hmm, is there a reason `timestamp` and `block_number` aren't fields on `ExecutionContext` so that they don't need to be looked up on every `CallContext` initialization?

jimpo: Something like https://github.com/jimpo/substrate/commit/bb2ab7c79097890c6b452cbe7950bbb4da977f68

pepyakin: I believe there is no actually! Here is the documentation of the current state of affairs. The PR (with appropriate change of COMPLEXITY.md : ) ) would be useful though!

jimpo: Would you rather cherry pick that into this and update the docs or should I open a separate PR? I feel like cherry-picking and doing it all at once is easy enough.

pepyakin: I agree that it should be easy enough. However, I prefer more focused PRs and this PR is already looks good to merge, so there won't be a clear win changing it here.

jimpo:
Need to bump `impl_version` to get CI passing.

PR #3115, opened 2019-07-14
Weight Check signed extension.

Kianenigma:
To be mixed with #3102.

### DIFF
--- a/core/sr-primitives/src/generic/unchecked_extrinsic.rs
+++ b/core/sr-primitives/src/generic/unchecked_extrinsic.rs
@@ -199,7 +199,6 @@ mod tests {
 	use super::*;
 	use runtime_io::blake2_256;
 	use crate::codec::{Encode, Decode};
-	use crate::generic::Era;
 	use crate::traits::{SignedExtension, BlockNumberToHash, Lookup, CurrentHeight};
 	use serde::{Serialize, Deserialize};
 
@@ -239,6 +238,7 @@ mod tests {
 	impl SignedExtension for TestExtra {
 		type AccountId = u64;
 		type AdditionalSigned = ();
+		fn additional_signed(&self) -> rstd::result::Result<(), &'static str> { Ok(()) }
 	}
 
 	type Ex = UncheckedExtrinsic<TestAccountId, TestCall, TestSig, TestExtra>;
--- a/node-template/runtime/src/lib.rs
+++ b/node-template/runtime/src/lib.rs
@@ -231,11 +231,11 @@ pub type Block = generic::Block<Header, UncheckedExtrinsic>;
 /// BlockId type as expected by this runtime.
 pub type BlockId = generic::BlockId<Block>;
 /// The SignedExtension to the basic transaction logic.
-pub type SignedExtra<R> = (system::CheckNonce<R>, system::CheckWeight<R>, balances::TakeFees<R>);
+pub type SignedExtra = (system::CheckNonce<Runtime>, system::CheckWeight<Runtime>, balances::TakeFees<Runtime>);
 /// Unchecked extrinsic type as expected by this runtime.
-pub type UncheckedExtrinsic = generic::UncheckedExtrinsic<Address, Call, AccountSignature, SignedExtra<Runtime>>;
+pub type UncheckedExtrinsic = generic::UncheckedExtrinsic<Address, Call, AccountSignature, SignedExtra>;
 /// Extrinsic type that has already been checked.
-pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, Call, SignedExtra<Runtime>>;
+pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, Call, SignedExtra>;
 /// Executive: handles dispatch to the various modules.
 pub type Executive = executive::Executive<Runtime, Block, Context, Runtime, AllModules>;
 
--- a/node-template/runtime/src/lib.rs
+++ b/node-template/runtime/src/lib.rs
@@ -17,7 +17,7 @@ use primitives::bytes;
 use primitives::{ed25519, sr25519, OpaqueMetadata};
 use runtime_primitives::{
 	ApplyResult, transaction_validity::TransactionValidity, generic, create_runtime_str,
-	traits::{self, NumberFor, BlakeTwo256, Block as BlockT, StaticLookup, Verify}
+	traits::{self, NumberFor, BlakeTwo256, Block as BlockT, StaticLookup, Verify}, weights::Weight,
 };
 use client::{
 	block_builder::api::{CheckInherentsResult, InherentData, self as block_builder_api},
@@ -113,6 +113,7 @@ pub fn native_version() -> NativeVersion {
 
 parameter_types! {
 	pub const BlockHashCount: BlockNumber = 250;
+	pub const MaximumBlockWeight: Weight = 4 * 1024 * 1024;
 }
 
 impl system::Trait for Runtime {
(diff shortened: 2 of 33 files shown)


node/runtime/src/lib.rs
```diff
@@ -436,11 +439,11 @@ pub type SignedBlock = generic::SignedBlock<Block>;
 /// BlockId type as expected by this runtime.
 pub type BlockId = generic::BlockId<Block>;
 /// The SignedExtension to the basic transaction logic.
-pub type SignedExtra = (system::CheckNonce<Runtime>, balances::TakeFees<Runtime>);
+pub type SignedExtra<R> = (system::CheckNonce<R>, system::CheckWeight<R>, balances::TakeFees<R>);
```
gavofyork: Why?

srml/system/src/lib.rs
```diff
@@ -753,6 +755,42 @@ impl<T: Trait> Module<T> {
 	}
 }
 
+/// Weight limit check and increment.
+#[derive(Encode, Decode, Clone, Eq, PartialEq)]
+pub struct CheckWeight<T: Trait + Send + Sync>(PhantomData<T>);
+
+impl<T: Trait + Send + Sync> SignedExtension for CheckWeight<T> {
+	type AccountId = T::AccountId;
+	// TODO TODO: make sure that this has to be here and not in validate.
```
gavofyork: it should be in both; `validate` happens on the way from the txq into the authored block, `pre_dispatch` is when verifying an (already authored) block.

node/runtime/src/lib.rs
Kianenigma: Because in different places (`executive`, `executor` etc.), we also want to know the _real runtime's transaction extension format_ and based on the environment, if, in `TestExternalities`, we can't import the `SignedExtra` without the generic, and I don't want to duplicate the code there.

Kianenigma: https://github.com/paritytech/substrate/pull/3115/files#diff-9be02fd5e4de6c77f22d6327fdaa09afL438

srml/system/src/lib.rs
Kianenigma: will add it once the operational stuff is merged in.

core/sr-primitives/src/generic/unchecked_extrinsic.rs
```diff
@@ -312,63 +310,6 @@ mod tests {
 		);
 	}
 
-	#[test]
-	fn mortal_signed_check_should_work() {
```
Kianenigma: without the mortality logic here these tests are pointless.

node-template/runtime/src/lib.rs
```diff
@@ -228,11 +231,11 @@ pub type Block = generic::Block<Header, UncheckedExtrinsic>;
 /// BlockId type as expected by this runtime.
 pub type BlockId = generic::BlockId<Block>;
 /// The SignedExtension to the basic transaction logic.
-pub type SignedExtra = (system::CheckNonce<Runtime>, balances::TakeFees<Runtime>);
+pub type SignedExtra<R> = (system::CheckNonce<R>, system::CheckWeight<R>, balances::TakeFees<R>);
```
Kianenigma: needs updating once all signed extra components are added.

node/runtime/src/lib.rs
gavofyork: that sentence isn't english! please reformulate.

srml/executive/src/lib.rs
```diff
@@ -435,12 +429,17 @@ mod tests {
 		}
 	}
 
-	type SignedExtra = (system::CheckNonce<Runtime>, balances::TakeFees<Runtime>);
+	type SignedExtra = node_runtime::SignedExtra<Runtime>;
```
gavofyork: this shouldn't be pulling in random types/configuration from `node_runtime`. it should formulate its own `SignedExtra` type as needed, just as it formulates its own config trait implementations.

node/runtime/src/lib.rs
gavofyork: if what you're getting is what i think it is, then it's an incorrect assertion anyway. the node runtime's `SignedExtra` type is only meaningful when it's bound to the node runtime's `Runtime` type. the example you gave binds it to some other `Runtime` type, taking it out of context. in that case, the text runtime should have its own `SignedExtra` (which may look fairly similar, just as many of the config traits do, but isn't semantically deduplicatable).

core/sr-primitives/src/generic/unchecked_extrinsic.rs
gavofyork: then they should be replaced with tests for the `CheckEra` type directly.

node-template/runtime/src/lib.rs
gavofyork: this is incorrect. please revert.

node/runtime/src/lib.rs
Kianenigma: > in that case, the test runtime should have its own SignedExtra (which may look fairly similar, just as many of the config traits do, but isn't semantically deduplicatable).

Ok this makes sense. I prioritized not having duplicated code as much as possible but I see that it was not correct. The rest of your comments all point back to this matter, which I will revert. 👍

core/sr-primitives/src/generic/unchecked_extrinsic.rs
Kianenigma: Yeah will add some unit tests to `system` for it. Also did the same for all other `SignedExtensions`: each module is now responsible to test it properly.

node/cli/src/factory_impl.rs
```diff
@@ -51,6 +51,16 @@ pub struct FactoryState<N> {
 
 type Number = <<node_primitives::Block as BlockT>::Header as HeaderT>::Number;
 
+impl<Number> FactoryState<Number> {
+	fn build_extra(index: node_primitives::Index, phase: u64) -> node_runtime::SignedExtra<Runtime> {
```
Kianenigma: @gavofyork how about this? The factory seems to be an attachment that only works with substrate node and it should be fine to borrow this type.

core/sr-primitives/src/generic/unchecked_extrinsic.rs
Kianenigma: added

node/cli/src/factory_impl.rs
```diff
@@ -51,6 +51,16 @@ pub struct FactoryState<N> {
 
 type Number = <<node_primitives::Block as BlockT>::Header as HeaderT>::Number;
 
+impl<Number> FactoryState<Number> {
+	fn build_extra(index: node_primitives::Index, phase: u64) -> node_runtime::SignedExtra {
+		(
+			system::CheckEra::from(Era::mortal(256, phase)),
+			system::CheckNonce::from(index),
+			system::CheckWeight::from(),
+			balances::TakeFees::from(0))
```
gavofyork:
```suggestion
			balances::TakeFees::from(0)
		)
```

node/executor/src/lib.rs
```diff
@@ -81,8 +81,10 @@ mod tests {
 	type TestExternalities<H> = CoreTestExternalities<H, u64>;
 
 	// TODO: fix for being charged based on weight now.
-	fn transfer_fee(bytes: Balance) -> Balance {
-		<TransactionBaseFee as Get<Balance>>::get() + <TransactionByteFee as Get<Balance>>::get() * bytes
+	fn transfer_fee<E: Encode>(extrinsic: &E) -> Balance {
+		<TransactionBaseFee as Get<Balance>>::get() +
+		<TransactionByteFee as Get<Balance>>::get() *
```
gavofyork:
```suggestion
			<TransactionByteFee as Get<Balance>>::get() *
```

node/executor/src/lib.rs
```diff
@@ -81,8 +81,10 @@ mod tests {
...
+		(extrinsic.encode().len() as Balance)
```
gavofyork:
```suggestion
			(extrinsic.encode().len() as Balance)
```

PR #3116, opened 2019-07-15
Make mDNS part of DiscoveryBehaviour

tomaka:
Moves the mDNS discovery mechanism to the `DiscoveryBehaviour`, where it makes more sense.


core/network/src/discovery.rs
```diff
@@ -260,6 +287,34 @@ where
 			}
 		}
 
+		// Poll mDNS.
```
romanb: Just as a matter of preference, I would poll mDNS either before or after all Kademlia-related polling, not in-between, i.e. either before polling the random query timer or after polling Kademlia.

What's your general stance on the order of polling multiple internal sources? If neither can go into a "busy loop" where it rarely emits `NotReady`, and whatever does the polling always polls until `NotReady`, it shouldn't make much of a difference, of course, but in general the order does seem to imply a certain preference towards which "behaviour" has priority. Just wondering.

tomaka: I did the change!
There's no stance right now. mDNS and Kademlia are unrelated, so the ordering indeed doesn't matter.
I know that `Future::select` for example alternates between polling the first future first and the second future first.

PR #3120, opened 2019-07-15
node: fix hardcoded block time in integration tests

andresilva:
The chain spec used for the integration tests used to have a 4 second block time which was changed in #3093. In the integration tests we generate blocks manually which means that the timestamp we set on the block must be consistent with the slot duration of the block proposal mechanism. The tests were failing because the seal (i.e. slot number) and timestamp were not consistent with each other.

### DIFF
--- a/node/cli/src/service.rs
+++ b/node/cli/src/service.rs
@@ -216,7 +216,7 @@ mod tests {
 	use aura::CompatibleDigestItem;
 	use consensus_common::{Environment, Proposer, ImportBlock, BlockOrigin, ForkChoiceStrategy};
 	use node_primitives::DigestItem;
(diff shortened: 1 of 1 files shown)


node/cli/src/service.rs
```diff
@@ -312,7 +312,7 @@ mod tests {
 			});
 
 			let mut digest = Digest::<H256>::default();
-			digest.push(<DigestItem as CompatibleDigestItem<Pair>>::aura_pre_digest(slot_num * 10 / 2));
+			digest.push(<DigestItem as CompatibleDigestItem<Pair>>::aura_pre_digest(slot_num));
```
andresilva: The `aura_pre_digest` takes the slot number as parameter.

PR #2209, opened 2019-04-04
avoid key collision on child trie and proof on child trie

cheme:
* **Initial** **issue**

initial issue is that child trie key/value stored in rocksdb are all build the same way for all child trie, 
therefore two identical child trie will get all their key duplicated and the pruning will break one of the two tries (only child trie with guaranty of not having same key/value content can currently be use with pruning).

* **Keyspace change**

This PR solves the initial issue by prepending a unique id to the keyvalue db keys. It uses a HashDB implementation `KeySpaceDB` to prepend a unique trieid.

! with current implementation it is only prepended if `PrefixedMemoryDB` is use.
Long term design should move that keyspace information to the db layer (have keyvalue db that manage two level of collections: the heavy one and the light one).

Those prefix should also allow more efficient operation on child trie.
 

 - accessing all child trie keyvalue db key from rocksdb, allowing to skip what would otherwhise be a query of every nodes of the child trie for every of its states (blocks).

  - related to this access we can have efficient deletion.
  - related to this access we can export/import trie between chain (the prefix will need to be rewritten as it is only unique for a chain context).

* **Api change**

This PR also contains a change of child trie api: do not do operation depending on `storage_key` but do operation depending on `child_trie` state.

This can allow efficient child trie usage (no need to query parent trie on every operation).

Recently I tried to split this PR in two (keyspace first, api second), but this api change is needed when a child trie does not exists (having the child trie in parameter makes things easier).

PR #2865, opened 2019-06-14
Fix false positives in changes tries

svyatonik:
Long story short: there are sometimes false positives in changes tries and this PR fixes this. I.e. now when `key` at the beginning of the block has `original_value` AND block changes it to the same `original_value` => this change will appear in changes trie of this block. And after applying this fix - it won't.

But this PR is a bit more than just fixing a bug - I'd like to check that we actually need to fix this. It is a bug, for sure, when we talk about it in context of original specification. But we could, probably, benefit from it and convert this to feature. I'm currently talking about #2491 where I've lied (sorry for that) to @pepyakin , when he was asking whether changes tries will include notifications about changes like that.

Also I believe this closes #2826 - I've tested this RPC locally && found that it actually issues 1 entry per change, not per block (at least when changes tries are used). The only case is with these 'false changes' - then entry issued on every block that tries to change that value (though it changes it from `original_value` to `original_value`) && it seems redundant.

So here are pros of fixing the issue:
- it is a right thing to-do, because light (and RPC) implementations will have 100% guarantee that the value has actually been changed;
- changes tries are smaller => small performance boost, especially for digest tries.

And here are cons:
- fix requires additional `Backend::storage()` call for every key that has been touched by `Ext::set_storage()`;
- we need workarounds (like suggested in #2491) when we want light clients to be notified about false changes.

closes #2826 (I believe, or @pepyakin could you please clarify the case?)

Issue #2832, opened 2019-06-07
Re-rooting the state of a block


rphmeier:
Substrate chains may occasionally need to do state rollbacks. 
In Polkadot, that's the case when we have a fully-malicious validator committee whose misbehavior is discovered.

We introduce `ext_reroot(ancestor: BlockNumber) -> Hash` for this purpose.
This will set the current storage to be equal to the post-state of the ancestor of the current block with the given block number.

Some pseudocode for an implementation:

```rust
// substrate-executor
fn ext_reroot(ancestor: BlockNumber) -> Hash {
    ext.reroot(ancestor);
}

// substrate-state-machine

struct Ext<'a, B: 'a + Backend> {
     backend: Cow<'a, B>,
     prospective: &'a mut OverlayedChanges,
     // ...
}

impl Ext {
    fn reroot(number: BlockNumber) {
        // prospective is an overlay of all storage items changed so far within the runtime execution
        // of the block.
        prospective.clear();
        if current_number <= ancestor { trap() }
        self.backend = Cow::Owned(self.backend.reroot(number));
    }
}

struct TrieBackend {
    fn reroot(&self, number) -> TrieBackend {
        let db_overlay = MemoryDB::new();   

        // reverse, inserting all deletions of ancestors and deleting all insertions.
        for (num, hash) in ancestry_iter(parent_hash) {
            if num == ancestor { break }
            let pruning_record = db.pruning_record(hash);
            for (inserted_key, _val) in pruning_record.insertions {
                db_overlay.remove(&inserted_key); // undo an insertion.
            }

            // values might not actually be available in the current implementation of pruning.
            // but the value should still be available within the DB because we are assuming this block hasn't been canonicalized yet (so deletions haven't been applied yet).
            for (_deleted_key, val) in pruning_record.deletions {
                db_overlay.insert(val); // undo a deletion.
            }
        };

        
        // we used a memorydb for `db_overlay` because it's very likely that many of the storage operations cancel out.
        // now, we have to make sure that the `update` produced by `Backend::root` and passed to `client::BlockImportOperation::update_storage` is based on top of the `db_overlay`, but still includes any further deltas.
    }
}
```

[...]

PR #3122, opened 2019-07-15
Implement change trie for child trie.

### LINKED ISSUE #2622: Handle child trie key in change trie
Change tries as described in  #628 store extrinsic index for a key change.
In the case of a child trie child key change, the root parent trie key will be include in change trie but not the key in the  child trie, making the information a bit less relevant.

I wonder if child trie key could benefit from the same mechanism. I see two options:
- put all in change_trie (change trie will replace key by (option<parent_key>, key).
- create a change_trie per child trie and manage multiple change trie

the second option looks the most promising to me but I am not very familiar with change trie, @svyatonik maybe you have a good idea.

 Also it seems  to me that this should be optional per child trie (iirc the change trie is optional).

There is also the question of the proof for child change trie.

cheme:
This PR is an attempt to fix #2622.

It makes changes trie handle child trie content.

It makes a few choice, so it is only a proposal and things can be change:
- a third change trie input node type: a child trie node containing the root of the child change trie. 
An alternate way of doing things would be to directly put child changes content into the top change trie with another new node (index build over encoding of block, encoded child trie key, child trie content key).
- those new change trie child nodes are not used in digest (it would be very straightforward to implement), digest nodes are still only build with former existing nodes. 
- child trie node change of root do not see its extrinsics registered as previously (the information being the sum of the change child trie extrinsics) : this can be restored from former code (a hashset keeping trace of change globally in overerlaychange), or build at the time child trie root is calculated.

It also will also need some changes after #2209, main question being do we refer to child trie keyspace or parent address, in this case parent address seems to be right addressing.

I also wonder if removing block number info from change trie keys could be a good idea?
cc/ @svyatonik 
It would requires prefixing memorydb key with encoded block number, and pruning could be done by directly removing all keys starting with this block number prefix (no need for trie parsing).
Similarly change trie child content could be prefixed by a unique id such as its storage path to be able to isolate its key values without trie parsing (but I got no direct use case except for being able to export change child trie without trie parsing).


core/state-machine/src/changes_trie/build.rs
```diff
@@ -110,75 +180,129 @@ fn prepare_extrinsics_input<'a, B, H, Number>(
...
+						if let Some(InputKey::DigestIndex::<Number>(trie_key)) = Decode::decode(&mut &key[..]) {
+							insert_to_map(&mut map, trie_key.key);
+						});
+
+
+					let child_index = ChildIndex::<Number> {
+						block: block.clone(),
+						storage_key,
+					};
+					child_map.insert(child_index, map.into_iter().map(|(_, (k, v))| InputPair::DigestIndex(k, v)));
```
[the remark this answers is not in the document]
cheme: will need some intermediate map (test cases were wrong to).

cheme:
> 1. I'm still not sure - why separate tries are better than one trie for both top + all children tries. Do you have an opinion on this?

Flattening child trie in the parent trie seems doable (putting optional child storage key in the input keys). This will lead to trie containing encoded child storage key as child trie prefix and instead of having a child trie root in a value node pointing to another child trie, we will get directly the child trie top.
So indead it will skip one hash per child trie and will certainly be more efficient for construction (plus code will be simpler I believe).

For what it is worth there may be some possible quicker access for multiple child trie content (the parent trie query/proof can be shared). But this can also be achieved with flattened child trie by basing the triedb on an middle branch node corresponding to the common encoded key bytes of all child trie nodes (plus possible partial for key).

In my opinion, what could justify the use of child trie is only wether we want to split change trie or not.

If we want to be able to revert some child trie to a previous state, we could use the former child trie to, but I do not think this would really make sense (pushing some special info in change trie seems more relevant).
So splitting change trie could only make sense if we want to move child trie state between chains, then the accompanying child trie part can be move, but here again it does not make much sense unless if we want to rebuild some child trie history (we need to keep header and a child trie root inclusion proof) with related change trie.

This only works cleanly if child trie get some prefix (similar to #2209) to fetch their key in the collection (currently no prefix are use on child trie).

> 2. if there are no strong requirements for using separate tries, I'd prefer to do some benches before merging this PR. Like - given the same set of changes how faster/slower is maintaining multiple tries vs big single 

Hypothetical, the usage of child trie in digest may lead to skip some trie access (but the common child trie path would be cached in case of a big single trie). I don't have any idea if it can make for the added indirection level on creation.

svyatonik:
OK - you must be talking mostly about #2832, right (I mean reverting changes trie to previous state)? So the only advantage is potential boost of a revert-to-block performance. But imo it doesn't make sense. Like if you are going to revert to block#500 when you're at the block#1000, then changes trie for block#500 isn't the changes trie you want to duplicate for block#1000. Changes trie contains state difference between current and previous blocks. Example:
- at block#499 there's only one key in storage: [`key1` => `Some(value1)`];
- at block#500 another key is inserted (`key2` => `Some(value2)`) => **changes trie for block#500 will contain these keys: [`key2`]**;
- at block#600 another key is inserted (`key3` => `Some(value3)`);
- when at block#1000 you want to revert to state of block#500, then **the changes trie for block#1000 must have keys: [`key3`]**, not the [`key2`] as changes trie for block#500. That's because only `key3` has been changed in 500...1000, and `key1` + `key2` were staying the same.

Or have I misunderstood something? So imo we must build changes trie from the scratch as we normally do in the case of revert-to-block operation.

I'd also summon @gavofyork here - maybe he has a strong opinion on whether we need one-changes-trie-per-child-trie, or not.

cheme:
About reverting, I ended up with the same conclusion as you (does not make sense for change trie).
When considering extracting a child trie globally from a chain, I did see a possible interesting thing (keeping change trie history, but for similar reason it may not be that useful).

cheme:
@svyatonik , if you wish to bench a bit, I did a quick implementation of flattened child trie here:
https://github.com/paritytech/substrate/compare/master...cheme:ch-ch-trie2?expand=1
(tested only over substrate-state-machine and with low code quality).

svyatonik:
Ah, I've also started that :) Nvm - will use your version, thanks! :)

svyatonik:
Okay, I've got some bench results. And actually this implementation works faster than implementation from `ch-ch-trie2`. [Here's](https://gist.github.com/svyatonik/a2a159dd4588ffd60a53dc2640b67e1b) the test that  I've used and results. Measured time is execution time of `build_changes_trie()` call, though I've also measured total time of `trie.insert()`-s and again it is better for this PR. Since performance has been my only concern, I think better to stick with separate changes tries, as you've suggested. Thanks for your help!

cheme:
That is not really what I expected, it comes probably from the trie reading being split. I would say `trie.insert` can be improve a bit by using algo such as 'iter_build' in https://github.com/paritytech/trie/pull/11 , but for `build_changes_trie` it would change nothing.

svyatonik:
@cheme Is this still "A3-inprogress"? :)

cheme:
Oh, I forgot to switch the tag, thanks.

core/state-machine/src/changes_trie/mod.rs
```diff
@@ -14,24 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
-//! Changes trie related structures and functions.
```
svyatonik: Seems like these docs should stay?

core/state-machine/src/changes_trie/build.rs
```diff
@@ -38,45 +39,114 @@ pub fn prepare_input<'a, B, H, Number>(
...
 					// ignore temporary values (values that have null value at the end of operation
 					// AND are not in storage at the beginning of operation
-					if !changes.storage(k).map(|v| v.is_some()).unwrap_or_default() {
+					let existing = if let Some(sk) = storage_key.as_ref() {
+						changes.child_storage(sk, k)
+					} else {
+						changes.storage(k)
+					};
+					if !existing.map(|v| v.is_some()).unwrap_or_default() {
 						if !backend.exists_storage(k).map_err(|e| format!("{}", e))? {
```
svyatonik: If `storage_key.is_some()`, there should be a call to `backend.exists_child_storage()`

cheme: pretty bad one. I remember when looking at this part of code that this query should be expensive, I guess there may be something doable to have the info in the transaction without querying the backend (would probably require additional query of backend during execution in some case).

svyatonik: We have had this discussion in #2865 - it was an (unsuccessful) attempt to extend this check to non-temporary values. IMO that's what state cache actually handles - if value has been read from trie during execution, then it'll be read from in-memory cache for the second time. And if it wasn't, then there's no other way, than to perform this read.

cheme: I also guess that running this backend check during block execution, in case of invalid block, is also an unnecessary operation.
Maybe if overlay was aware of accessed data, it could conditionally store this in the transaction, but that would be similar as putting some access cache in the overlayed_change (which may be interesting to avoid some map access (we already need to check change for get_storage) but would be some substantial design change).

Issue #2966, opened 2019-06-27
srml-contracts: charge for the size of loading storage


pepyakin:
We allow for storing arbitrary sized values in storage for contracts. Obviously, we allow loading arbitrary sized values. This, however, might pose an issue for gas metering since with current available API we don't have a means of getting the size of the value before loading it.

> ℹ️Note that this assumes the size of the value contributes a non-negligable amount.

Here are some ideas:

1. srml-contracts can augument every value with another cell storing its size. This will make every load to perform 2 trie accesses and thus not desirable.
2. gas can be taken after the loading is complete. However this opens up a possibility that with minimal amount of gas a malicious actor can perform loads of max possible sizes. Coupled with a `ext_call` to another contract with capped gas this might be an issue.
3. a gas deposit can be taken upfront up to maximal size. Then, after loading and determining the size, we can refund the unused gas. (Note, that we were to add a user specified cap, it will be equivalent to the pt.2)

thiolliere:
> a gas deposit can be taken upfront up to maximal size

This implies that a user can't provide the exact amount of gas necessary for computation, but have to provide: `gas that will be used + gas needed to load maximum size value`

I feel like this property if accepted could be use for various similar mecanism: if we don't know the specific price in advance, charge the maximum and refund the actual price when you know it.

But then the amount of gas necessary to execute a call is `amount of gas used + max(all mecanism that take gas and refund when knowing value)` or more if those mecanism can operate in parallel in the code.

jimpo:
I did some benchmarking on how storage value size and trie structure affect storage load times. To fetch a value from storage, it requires a trie traversal currently*. I ran some benchmarks against a Trie implemented over RocksDB on my laptop.

The idea is to vary both the byte size of the value and the number of trie nodes that need to be traversed to access it. There is a "branch nodes" parameter which is a divisor of 256 (since contract storage uses 256-bit keys) and a "value size" parameter. A trie database is initialized with values placed at keys with 32 \x00 bytes and 1 bit set. Essentially, this should create branch nodes along the left-most branch of the trie. Then a value of the specified byte size is written to the all-\x00 key. During the benchmark, this value is read repeatedly from storage without busting the RocksDB cache. The RocksDB cache size is set at the default, which is as it is in Substrate, though maybe this should also be varied in further experiments. You can see the code [here](https://github.com/jimpo/substrate/tree/storage-access-benchmarking-2).

The results are shown in these graphs:

[...]

PR #3126, opened 2019-07-16
Contract storage limit

### LINKED ISSUE #2192: Limit the size of a stored value in the storage
A contract can store a value of arbitrary size in its KV storage. Theoretically, it is possible to store a value that won't be possible to load, assuming the possibility of decreasing the block gas limit.

We might want to introduce a limit for the size of a strorage value. For that, we can add a field that holds the size limit. On each call to `set_storage`, it will check the length of the storage value and if it exceeds the limit a trap will be generated.

jimpo:
Limits contract storage values to 16 KiB. Based on analysis [here](https://github.com/paritytech/substrate/issues/2966#issuecomment-510087309), this is a threshold where reads/writes to/from LevelDB are minimally impacted by the size of the value stored. Above this threshold, the IO required to read/write the value begins to have a noticeable impact.

Fixes #2192.

### DIFF
--- a/srml/contracts/src/exec.rs
+++ b/srml/contracts/src/exec.rs
@@ -61,10 +61,9 @@ pub trait Ext {
 	/// was deleted.
 	fn get_storage(&self, key: &StorageKey) -> Option<Vec<u8>>;
 
-	/// Sets the storage entry by the given key to the specified value.
-	///
-	/// If `value` is `None` then the storage entry is deleted.
-	fn set_storage(&mut self, key: StorageKey, value: Option<Vec<u8>>);
+	/// Sets the storage entry by the given key to the specified value. If `value` is `None` then
+	/// the storage entry is deleted. Returns an Err if the value size is too large.
+	fn set_storage(&mut self, key: StorageKey, value: Option<Vec<u8>>) -> Result<(), &'static str>;
 
 	/// Instantiate a contract from the given code.
 	///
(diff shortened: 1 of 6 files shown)


srml/contracts/src/exec.rs
```diff
@@ -124,6 +126,9 @@ pub trait Ext {
 
 	/// Returns the current block number.
 	fn block_number(&self) -> BlockNumberOf<Self::T>;
+
+	/// Returns whether an item of the given size may be placed in storage.
+	fn is_storage_len_allowed(&self, len: u32) -> bool;
```
pepyakin: Why not publish the value itself? It could be used by the contracts

srml/contracts/src/wasm/runtime.rs
```diff
@@ -263,6 +263,9 @@ define_env!(Env, <E: Ext>,
 	//   where the value to set is placed. If `value_non_null` is set to 0, then this parameter is ignored.
 	// - value_len: the length of the value. If `value_non_null` is set to 0, then this parameter is ignored.
 	ext_set_storage(ctx, key_ptr: u32, value_non_null: u32, value_ptr: u32, value_len: u32) => {
+		if value_non_null != 0 && !ctx.ext.is_storage_len_allowed(value_len) {
```
pepyakin: I think this is noteworthy for the documentation

srml/contracts/src/exec.rs
```diff
@@ -61,9 +61,11 @@ pub trait Ext {
 	/// was deleted.
 	fn get_storage(&self, key: &StorageKey) -> Option<Vec<u8>>;
 
-	/// Sets the storage entry by the given key to the specified value.
+	/// Sets the storage entry by the given key to the specified value. If `value` is `None` then
+	/// the storage entry is deleted.
 	///
-	/// If `value` is `None` then the storage entry is deleted.
+	/// Pre-conditions: The caller must check before that the value size is allowed using
+	/// `is_storage_len_allowed`, otherwise this method will panic.
```
pepyakin: I think a better way is to return `Result` and then at the call site proof that there can't be `Err`.

PR #3128, opened 2019-07-16
Operational Transaction signed extension.

Kianenigma:
built on top of #3115 and should be updated with it. Should only be marge after #3115 is merged.

- Renames the `#[weight]` to a more generalized `#[meta] ` tag which can carry static information about both weight, priority and basic functions on top of them. (`pub type CallDescriptor = (Weight, TransactionPriority);`-- okay name?)
- Consequently refactors some occurrences of `weight` in the code to the more general structure.

Gist of the change being: 

```rust
fn validate(
	&self,
	_who: &Self::AccountId,
	info: CallDescriptor,
	let (weight, priority) = info;
	Self::internal_check_weight(weight)?;
	Ok(ValidTransaction { priority, ..Default::default() })
}
```

where setting `#[meta = WeightedTransaction::Operational()]` simply boosts to your priority to infinity and drops your weight to `Zero` etc.

TODO:
 - [ ] allow `CallMetadata ` to define `will_cause_full_block()` to guarantee operational txs will go through. 

I will mark this as in-progress now since reviewing it is moot with the additional commits of #3115.

Also closes #3106


gavofyork:
> and drops your weight to Zero etc.

why would it do this? wouldn't that make the transaction free? what stops a dos/sybil attack?

Kianenigma:
This needs to be started over again.

PR #3133, opened 2019-07-16
Change dependencies on hex-literal 0.1 to 0.1.3 to fix build.

JoshOrndorff:
July 16th, `substrate-node-new` stopped correctly building new node templates. It failed compiling hex-literal 0.1.4

This PR changes all Cargo.toml entries
```diff
- hex-literal = "0.1"
+ hex-literal = "0.1.3"
```

It leaves dependencies on hex-literal that were already pinned to more specific versions untouched.

This PR is against v1.0 instead of master because there is no build problem on master, and the dependency has already been upgraded to 2.0 in most places on master.
I learned later that `cargo update -p hex-literal` may have obviated removing the Cargo.lock files. Suggestions how to proceed on that front are welcome.


bkchr:
That was fixed by https://github.com/paritytech/substrate/pull/3141

PR #3135, opened 2019-07-16
fix and test default instance for collective

thiolliere:
Some structs were lacking default instance, this is now tested and fixed.

cc @JoshOrndorff


bkchr:
I thought that `DefaultInstance` is not required?

thiolliere:
indeed it must not be required.

but Default is already put on Trait so we should be coherent if we want this module to have a default instance we need to put it on event and origin as well.

bkchr:
If we add the `DefaultInstance` can we not drop the PhantomData?

thiolliere:
the PhantomData is not needed, even without this pr change it was not needed but I'm actually surprised because the variants only capture `system::Trait + I`.

So now phantomdata is removed as it is not necessary

PR #3137, opened 2019-07-17
slashing aggregator v2

niklasad1:
Hey,

This PR is the next attempt to introduce the slashing aggregator "partly"
## What does this 
It enables the functionality of what to do each severity level instead of specifying strictly based on misbehavior.

It introducing the following:
- Slashing module for keeping track of history and slash.
- AfterSlashing implementation for Staking 
- Generic interface ReportSlash, DoSlash and DoReward
- RollingWindow Module

By introducing the traits ReportSlash and DoSlash for a generic interface to slash misbehavior(s) with support of a `RollingWindow` which will keep the last `n` sessions of misbehaviors. The number of sessions is determined by the misconduct type.

Additionally, the `basic Misconducts kinds` are pre-defined in the runtime with a window length as input which needs to be initialized by the concrete misconduct such as `BabeEquivocation`. So, for example, `GrandpaEquivocation` may need a different window than `BabeEquivocation` and so on.

However, it is assumed that the `window_length` and `base_severity` is static for the concrete misconduct. I have two different macros for them but they are basically the same with different function names because I think it will be easier to use and you might not need both,

See the example use-case below which is probably easier to understand

## What is missing

- Misbehavior reporting
- Misbehavior implementations (such as BabeEquivocation, Unresponsiveness and similar)
- macro for `ReportSlash` (currently very boilerplate-isch)

## Example use-case
See test implementation in `substrate/srml/staking/src/slash.rs`


srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+			last_seen_session = session;
+		}
+
+		unique
+	}
+}
+
+/// Macro for implement static `base_severity` which may be used for misconducts implementations
+#[macro_export]
+macro_rules! impl_base_severity {
```
niklasad1: these macros are basically the same,

srml/staking/src/lib.rs
```diff
@@ -1380,3 +1380,17 @@ impl<T: Trait> SelectInitialValidators<T::AccountId> for Module<T> {
 		<Module<T>>::select_validators().1
 	}
 }
+
+
+/// StakingSlasher
+pub struct StakingSlasher<T>(rstd::marker::PhantomData<T>);
+
+impl<T: Trait> DoSlash<Perbill, (T::AccountId, Exposure<T::AccountId, BalanceOf<T>>)> for StakingSlasher<T> {
+	fn do_slash(severity: Perbill, (who, exposure): (T::AccountId, Exposure<T::AccountId, BalanceOf<T>>)) {
```
niklasad1: note, perbill is a concrete type here

srml/support/src/traits.rs
```diff
@@ -649,3 +649,19 @@ pub trait ChangeMembers<AccountId> {
 impl<T> ChangeMembers<T> for () {
 	fn change_members(_incoming: &[T], _outgoing: &[T], _new_set: &[T]) {}
 }
+
+/// A generic trait for reporting slashing violations.
+pub trait ReportSlash<Identification, Hash> {
+    fn slash(identfication: Identification, footprint: Hash);
```
niklasad1: this may need to return the misconduct level (1..=4)

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+
+decl_module! {
+	/// Rolling Window module
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {}
+}
+
+impl<T: Trait> Module<T> {
+	/// Remove items that doesn't fit into the rolling window
+	/// Must be called when a session ends
+	pub fn on_session_end() {
```
rphmeier: it's a bit strange to refer to all the session machinery without actually linking to `session`. Why not do that?

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+		let session = SessionIndex::get().wrapping_add(1);
+
+		for (kind, _) in <MisconductReports<T>>::enumerate() {
+			let window_length = kind.window_length();
+			<MisconductReports<T>>::mutate(kind, |reports| {
+				// it is guaranteed that `reported_session` happened before `session`
+				reports.retain(|(reported_session, _)| {
```
rphmeier: this will continue iterating after passing all items to be pruned away. probably can be done more efficiently with a `position` and `drain` combo. but we can file that for later

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+					diff < *window_length
+				});
+			});
+		}
+
+		SessionIndex::put(session);
+	}
+
+	/// Report misbehavior for a kind
+	pub fn report_misbehavior(kind: T::Kind, footprint: T::Hash) {
```
rphmeier: is there a trait that this should be in?

rphmeier:
In the example, it should be 

`type FullId<T, K> = <<T as Trait>::KeyOwner as KeyOwnerProofSystem<(KeyTypeId, K)>>::FullIdentification;`

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+	/// Return number of misbehavior's in the current window which
+	/// may include duplicated misbehaviour's
+	pub fn get_misbehaved(kind: T::Kind) -> u64 {
+		<MisconductReports<T>>::get(kind).len() as u64
+	}
+
+
+	/// Return number of misbehavior's in the current window which
+	/// ignores duplicated misbehavior's in each session
+	pub fn get_misbehaved_unique(kind: T::Kind) -> u64 {
```
rphmeier: this function is missing the mark on what we want from uniqueness.

1. uniqueness should be detected on insertion, not on query. attempting to insert a non-unique misbehavior should lead to failure (which is why `report_misbehavior` should return a `Result`).
2. uniqueness should be enforced for the entire bonding period, not just a rolling window.

----

With that in mind, I think enforcing uniqueness should be done like this:
1. Create a new entry in storage mapping `Hash => Session`
2. When reporting, insert if and only if there is no entry there. The session is the current session index.
3. We don't prune for the moment, so it will grow unbounded. but storing the session index will let us do pruning easily later (we will have to think of how to reward it -- probably by re-minting coins to someone who issues a transaction to prune it when it's old)

srml/staking/src/lib.rs
rphmeier: we also will want to reward those who reported the slash using some of the funds from the slash.

the simplest but probably least-flexible way is to have `do_slash` take another argument `who: Vec<(AccountId, Perbill)>` where the Perbill is a "priority". For a single reporter, this would be 100%. If we have many reporters (e.g. fishermen), we would distribute the reward among them unevenly based on who was first.

srml/rolling-window/src/lib.rs
niklasad1: Ok, I implemented `srml_session::OnSessionEnding` for it. Is that what you meant?

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+		/// Misbehavior reports
+		///
+		/// It stores every unique misbehavior of a kind
+		// TODO(niklasad1): optimize how to shrink the window when sessions expire
+		MisconductReports get(kind): linked_map T::Kind => Vec<SessionIndex>;
+
+		/// Bonding guard
+		///
+		/// Keeps track of unique reported misconducts
+		BondingGuard get(uniq): linked_map (T::Kind, T::Hash) => ();
```
niklasad1: okay, this still doesn't work

we need to insert `EraIndex` or something similar because we need it `live` in one BondingDuration I guess

srml/rolling-window/src/lib.rs
niklasad1: I'm not sure how to do that because we not really interested in `session_index` instead `session_expire (session_index + window_length)` is relevant. Subsequent items can have very different `session_expire`.

However, alternatively we would store `session_expire` instead but then we couldn't detect whether it `wrapped around or not` because we wouldn't know if `current_session` happened before or after `session_expire`. 

I will file an issue let's revisit this then (it is an TODO atm)

srml/rolling-window/src/lib.rs
niklasad1: > uniqueness should be enforced for the entire bonding period,

Well, just to verify that I understand this correctly

bonding  period == bond(stash, controller) ------> unbond(controller) -------> BondingDuration

I'm not sure how to keep track of the `bonding period`. Is it sufficient to just have a event handler `on_bonding_period_end()` to be called and maybe reset it accordingly?

> We don't prune for the moment, so it will grow unbounded. but storing the session index will let us do pruning easily later (we will have to think of how to reward it -- probably by re-minting coins to someone who issues a transaction to prune it when it's old)

I don't understand what we are not `pruning` can you explain?

rphmeier: See the second list of 3 items. Just implement that

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+		MisconductReports get(kind): linked_map T::Kind => Vec<SessionIndex>;
+
+		/// EraData which have mapping from `EraIndex` to a unique Hash
+		///
+		EraData get(idx): linked_map EraIndex => Vec<T::Hash>;
+
+		/// Bonding guard
+		///
+		/// Keeps track of uniquely reported misconducts in the entire bonding duration
+		/// It will be unbounded
```
niklasad1: remove

gavofyork:
> `let footprint = (b"babe_equivocation", &proof.author, equivocated_slot).using_encoded(|s| T::Hashing::hash_of(s));`

This will give you the hash of the encoding of the byte slice/Vec<u8> that is the encoding of `s`. i.e. it'll be double encoded because `hash_of` already encodes implicitly.

Instead it should just be:
```
let footprint = T::Hashing::hash_of((b"babe_equivocation", &proof.author, equivocated_slot));
```

srml/rolling-window/src/lib.rs
gavofyork: @niklasad1 i think @rphmeier means that for now, any given account id must never report the same misbehaviour more than once, regardless of how far apart they are. so you should just store the hash of the report as the key of a storage item (session number at which it was reported would be a sensible value) and quickly look it up to make sure it's empty when a misbehaviour is reported. if it's not empty then it should return an `Err`. 

at some later stage we can create logic that knows about how long the bonding period is in order to allow these storage items to be removed.

niklasad1: thanks Gav, @rphmeier clarified it for me yesterday.

I have implemented logic to remove `misbehaviors` older than one bonding period where the `BondingDuration` is an associated type (not really coupled to staking). 

So, if Staking::BondingDuration != Self::BondingDuration it could we could get unexpected behavior. 

I guess we could require `RollingWindow` to inherit the `Staking trait` otherwise I don't think it is possible to fetch but I would like to avoid that.

Thoughts?

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+#[cfg(test)]
+mod mock;
+
+use srml_support::{
+	StorageMap, EnumerableStorageMap, decl_module, decl_storage,
+	traits::{Get, WindowLength}
+};
+use parity_codec::Codec;
+use sr_primitives::traits::MaybeSerializeDebug;
+use srml_session::SessionIndex;
```
niklasad1: could be an associated type instead but because we `impl` srml_session::OnSessionEnding it is probably fine

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+use srml_staking::EraIndex;
```
niklasad1: Could be an associated and remove dependency to `srml_staking`

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+	/// Rolling Window module
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {}
+}
+
+impl<T: Trait> Module<T> {
+	/// Report misbehavior for a kind
+	///
+	/// If the misbehavior is not unique `Err` is returned otherwise the number of misbehaviors for the kind
+	/// is returned
+	pub fn report_misbehavior(
```
gavofyork: this almost certainly belongs in a trait so that this module can be referenced by other modules from the aggregated runtime without the need to actually depend on it explicitly.

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+	/// Kind to report with window length
+	type Kind: Copy + Clone + Codec + MaybeSerializeDebug + WindowLength<u32>;
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as RollingWindow {
+		/// Misbehavior reports
+		///
+		/// It stores every unique misbehavior of a kind
+		// TODO(niklasad1): optimize how to shrink the window when sessions expire
```
gavofyork: needs an issue number for tracking this. (make sure you assign yourself)

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+		MisconductReports get(kind): linked_map T::Kind => Vec<SessionIndex>;
+
```
gavofyork: double line break.

gavofyork:
@rphmeier this looking ok to you?

srml/support/src/traits.rs
```diff
@@ -648,3 +648,24 @@ pub trait ChangeMembers<AccountId> {
 impl<T> ChangeMembers<T> for () {
 	fn change_members(_incoming: &[T], _outgoing: &[T], _new_set: &[T]) {}
 }
+
+/// A generic trait for reporting slashing violations.
+pub trait ReportSlash<Misbehaved, Reporter, Hash> {
+	/// Reports slashing for a misconduct where `to_slash` is the misbehaved entities and
```
marcio-diaz: I'm not native english speaker, but it sounds better to me `to_slash *are* the misbehaved entities`

srml/support/src/traits.rs
```diff
@@ -648,3 +648,24 @@ pub trait ChangeMembers<AccountId> {
...
+	/// `to_reward` is the entities that detected and reported the misbehavior
+	///
+	/// Returns `Ok` if the misconduct was unique otherwise `Err`
+	fn slash(to_slash: Misbehaved, to_reward: Reporter, footprint: Hash) -> Result<(), ()>;
```
marcio-diaz: `Reporter` -> `Reporters`. Maybe we should use `Err` in case the function fail (but the misbehavior was unique) and return `false` if the misbehavior is duplicated.

srml/support/src/traits.rs
niklasad1: yeah, it is a weird sentence but I think regarded the `to_slash` as a singular (single variable) thus `is` but I can change it to `are`

srml/rolling-window/src/mock.rs
```diff
@@ -0,0 +1,313 @@
...
+		let mut ext = t.into();
+		runtime_io::with_externalities(&mut ext, || {
+			let validators = Session::validators();
+			SESSION.with(|x|
+				*x.borrow_mut() = (validators.clone(), HashSet::new())
+			);
+		});
+		ext
+	}
+}
```
marcio-diaz: I wonder if we really need to write all of this, looks like too much work.

rphmeier: it's definitely longer than other SRML mocks, not sure why that is

srml/staking/src/lib.rs
```diff
@@ -1380,3 +1380,25 @@ impl<T: Trait> SelectInitialValidators<T::AccountId> for Module<T> {
 		<Module<T>>::select_validators().1
 	}
 }
+
+/// StakingSlasher
```
rphmeier: This documentation is absolutely useless

srml/staking/src/lib.rs
```diff
@@ -1380,3 +1380,25 @@ impl<T: Trait> SelectInitialValidators<T::AccountId> for Module<T> {
...
+pub struct StakingSlasher<T>(rstd::marker::PhantomData<T>);
+
+impl<T: Trait, Reporter>
```
rphmeier: Why would the staking slasher be able to reward any kind of reporter? That seems overly polymorphic. Try to think a little further ahead. It's probably not possible to actually write rewarding code for reporter types that are not `T::AccountId`. At the very least it's non-obvious and we shouldn't set traps for future maintainers.

rphmeier: Also, what if there are multiple reporters?

niklasad1: it was of pure laziness because it is not used yet, shall probably be something like `IntoIterator<Item=T::AccountId>`

will fix

srml/rolling-window/src/mock.rs
niklasad1: copy-pasted from `srml-staking/src/mock`

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+	type Kind: Copy + Clone + Codec + MaybeSerializeDebug + WindowLength<u32>;
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as RollingWindow {
+		/// Misbehavior reports
+		///
+		/// It stores every unique misbehavior of a kind
+		// TODO [#3149]: optimize how to shrink the window when sessions expire
+		MisconductReports get(kind): linked_map T::Kind => Vec<SessionIndex>;
```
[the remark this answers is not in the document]
niklasad1: > Why not mapping kind => hash instead? I think we are losing information that may be useful.

Because we are using `SessionIndex` to remove/filter out old misbehaviors and all hashes are stored in `BondingUniqueness` so we shouldn't be losing information. Ok?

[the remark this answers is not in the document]
marcio-diaz: I was thinking in case we want to get the misbehaviors (hashes) of a kind, looks not possible. For filtering performance could be `Kind -> (SessionIndex, Hash)`. But yeah, I don't know if it's worth.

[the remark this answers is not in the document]
niklasad1: Right I see, it more or less defeats the purpose of the rolling window but it might be useful. Let's keep it out of this PR and revisit again if we would need it, ok?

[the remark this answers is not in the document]
marcio-diaz: Sure.

srml/rolling-window/src/mock.rs
thiolliere: I don't think subtelties of config of staking should copy paste here, make the config of staking mock an ok default for your tests

also I'm not even sure you need staking for your test do you ?

niklasad1: yepp, it's probably not needed

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+			Err(())
+		} else {
+			<BondingUniqueness<T>>::insert(footprint, current_session);
+			<MisbehaviorReports<T>>::mutate(kind, |entry| entry.push(current_session));
+			Ok(())
+		}
+	}
+}
+
+impl<T: Trait> srml_session::OnSessionEnding<T::AccountId> for Module<T> {
```
gavofyork: shouldn't this be a `impl SessionHandler`, for the function `on_new_session`?

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+	#[test]
+	fn it_works() {
+		with_externalities(&mut new_test_ext(), || {
+			let zero = H256::zero();
+			let one: H256 = [1_u8; 32].into();
+			let two: H256 = [2_u8; 32].into();
+
+			let mut current_session = 0;
+
+			assert!(RollingWindow::report_misbehavior(Kind::Two, zero, current_session).is_ok());
```
gavofyork:
```suggestion
			assert_ok!(RollingWindow::report_misbehavior(Kind::Two, zero, current_session));
```

this is better since it'll nicely print the error text in tests if it fails.

gavofyork: should be replicated in rest of tests

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+			assert_eq!(RollingWindow::get_misbehaviors(Kind::Two), 1);
+			assert!(RollingWindow::report_misbehavior(Kind::Two, zero, current_session).is_err());
```
gavofyork:
```suggestion
			assert_noop!(RollingWindow::report_misbehavior(Kind::Two, zero, current_session), "whatever the error string is");
```

gavofyork: should be replicated in rest of tests.

gavofyork:
this code doesn't seem to do the slashing; presumably that must be added in a later PR?

also, i don't see any way that you'd be able to ensure that different (though compounding) slashes over a window end up being all for the same amount. is it assumed that the slashes would only happen at the end of a session? where would the disabling happen? don't we need a central slashing aggregator to extensibly have these sort of events? indeed, this PR seems pretty light on logic overall.

node/runtime/src/lib.rs
```diff
@@ -533,3 +536,38 @@ impl_runtime_apis! {
 		}
 	}
 }
+
+/// Misbehavior type which takes window length as input
+/// Each variant and its data is a seperate kind
+#[derive(Copy, Clone, Eq, Hash, PartialEq, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug))]
+pub enum Misbehavior {
```
Kianenigma: Could be added to `node/runtime/impls.rs` to keep the top-level file concise.

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+	}
+}
+
+decl_module! {
+	/// Rolling Window module
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {}
+}
+
+/// Trait for getting the number of misbehaviors in the current window
+pub trait GetMisbehaviors<Kind> {
```
Kianenigma: I wouldn't create a trait for something that is used just once + in the same file. could simply go to `impl Module<T>`. Though I might have to full picture but searching this PRs shows that this trait is nerve used elsewhere.

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+	fn report_misbehavior(
+		kind: T::Kind,
+		footprint: T::Hash,
+		current_session: SessionIndex,
+	) -> Result<(), ()> {
+		if <BondingUniqueness<T>>::exists(footprint) {
+			Err(())
+		} else {
+			<BondingUniqueness<T>>::insert(footprint, current_session);
+			<MisbehaviorReports<T>>::mutate(kind, |entry| entry.push(current_session));
```
Kianenigma: I am not sure using it currently okay or not, given that it had a panic issue, but this is a typical use case of how `::append()` is significantly faster than reading -> pushing -> storing.

(we are using it to add events and there is a PR for it #3071. Linking so that it can be refactored later.)

srml/staking/src/lib.rs
```diff
@@ -1380,3 +1380,28 @@ impl<T: Trait> SelectInitialValidators<T::AccountId> for Module<T> {
 		<Module<T>>::select_validators().1
 	}
 }
+
+/// A type for performing slashing based on severity of misbehavior
+pub struct StakingSlasher<T>(rstd::marker::PhantomData<T>);
```
Kianenigma: Just to understand things better, where is this used exactly?

srml/rolling-window/src/lib.rs
Kianenigma: Probably due to not knowing the big picture but since I don't find another reference to `get_misbehaviors`, and  `report_misbehavior `, I don't get why they need a Trait rather than being functions for `Module<T>`.

srml/rolling-window/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+				// it is guaranteed that `reported_session` happened in the same session or before `ending`
+				reports.retain(|reported_session| {
+					let diff = ending.wrapping_sub(*reported_session);
+					diff < *window_length
+				});
+			});
+		}
+
+		// don't provide a new validator set
+		None
```
Kianenigma: Same as the above comment, I think you can get rid of this, which is kinda weird, using `on_new_session` instead.

gavofyork:
yeah - this looks rather incomplete for its title and description.

srml/staking/src/lib.rs
niklasad1: it is not used yet but it would be used as follows:

1. Consensus Module (reports violation)
2. After that the violation is proofed, calls `ReportSlash:slash` (this will differ between different classes of slashing)
3. ReportSlash::slash calls `StakingSlasher::slash` and this performs the actual slashing and rewarding

srml/rolling-window/src/lib.rs
niklasad1: Ok, the only thing we need from this is the `session_number` in order to remove old misbehaviors and I don't see how `SessionHandler` could provide that?

I guess it could work by keeping a separate `SessionIndex` in this module?!

niklasad1:
> this code doesn't seem to do the slashing; presumably that must be added in a later PR?

slashing is performed by `srml_staking::StakingSlasher` in this PR but it is not really used because we are missing `Reporting (Consensus mod)` and `ReportSlash (actual misconduct implmentations)` 

> also, i don't see any way that you'd be able to ensure that different (though compounding) slashes over a window end up being all for the same amount. 

The assumption is that severity is calculated based on misbehaviors in the current window and each one gets slashed according to the severity of their exposure. So it may be a different amount for the punished ones. Is that a problem or am I misunderstanding something?

> is it assumed that the slashes would only happen at the end of a session?
Not necessarily but I guess it would common because we want to reward `misbehavior reporters` and that is done at the end of a session AFAIU 

> where would the disabling happen? don't we need a central slashing aggregator to extensibly have these sort of events? indeed, this PR seems pretty light on logic overall.

Yes, basically based on the slashing we need to estimate misconduct level and apply appropriate action such as disable the validator, kick out the validator from the next NPoS election and so on. I wrote a bit about it in the PR description `to follow up`

node/runtime/src/lib.rs
niklasad1: The reason why I didn't put there is because of `//! Some configurable implementations as associated type for the substrate runtime.`

srml/rolling-window/src/lib.rs
gavofyork: you could just add it as a param of `fn on_new_session`

gavofyork:
> So it may be a different amount for the punished ones. Is that a problem or am I misunderstanding something?

yeah this is an issue, i think. the order of reports coming in shouldn't result in different punishments. each additional slash should raise the slash of the previously slashed guys to the same level as the newly slashed guy.

AlistairStewart:
Yes, ideally for things we want to slash more with correlation, later reports against other validators should cause earlier reported validators who committed the same fault to be slashed more.

Ideally for multiple slashable faults occuring at the same time, a validator should be slashed the maximum amount, not the sum, and this should hold whenever the reports arrive.

If either of these is hard to implement, we can relax them a bit, especially for historical slashing. But for the first, we really don't want validators to be able to frontrun their own misbehaviour reports to get off lightly.

But @AlfonsoCev , who designed the latest spec might have other opinions.

niklasad1:
> Yes, ideally for things we want to slash more with correlation, later reports against other validators should cause earlier reported validators who committed the same fault to be slashed more.

Ok, makes sense. I think that doesn't need to be handled by the `rolling window`.  Instead, the `slashing or reporting` need to have a list with the previously slashed validators for a certain kind which have to be iterated over and apply slash iff the severity was lower than the current. (might be a poor abstraction if the slasher needs to know about the kind)

> Ideally for multiple slashable faults occuring at the same time, a validator should be slashed the maximum amount, not the sum, and this should hold whenever the reports arrive.

Ok, multiple faults may be different kinds right? I mean we will have different reporters for different kinds and might need some coordination then.

srml/rolling-window/src/lib.rs
niklasad1: Ok, I did it by modifying the function in the `traits`. I hope that is what you meant?

srml/rolling-window/src/lib.rs
niklasad1: #3071 will probably be merged for this one because it has issues. Will, wait

srml/support/src/traits.rs
```diff
@@ -640,3 +640,30 @@ pub trait ChangeMembers<AccountId> {
...
+	/// `to_reward` are the entities that detected and reported the misbehavior
+	///
+	/// Returns `Ok` if the misconduct was unique otherwise `Err`
+	fn slash(to_slash: Misbehaved, to_reward: Reporters, footprint: Hash) -> Result<(), ()>;
+}
+
+/// A generic trait for enacting slashes.
+pub trait DoSlash<Misbehaved, Reporters, Severity, Kind> {
+	/// Performs the actual slashing and rewarding based on severity
+	fn do_slash(to_slash: Misbehaved, to_reward: Reporters, severity: Severity, kind: Kind) -> Result<(), ()>;
```
niklasad1: yes, this is not an elegant abstraction but because we need to keep track on severity for a specific kind, I don't see another option when we need to take the history into account.

marcio-diaz:
I see that you added quite a few more things. Is it ready for another round of reviews? something missing?

niklasad1:
@marcio-diaz it is close, will probably be reviewable tomorrow

node/runtime/src/impls.rs
```diff
@@ -128,6 +132,47 @@ impl Convert<(Weight, WeightMultiplier), WeightMultiplier> for WeightMultiplierU
 	}
 }
 
+/// Misbehavior type which takes window length as input
+/// Each variant and its data is a seperate kind
+#[derive(Copy, Clone, Eq, Hash, PartialEq, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug))]
+pub enum Misbehavior {
```
gavofyork: This type shouldn't exist.

node/runtime/src/impls.rs
```diff
@@ -128,6 +132,47 @@ impl Convert<(Weight, WeightMultiplier), WeightMultiplier> for WeightMultiplierU
...
+	/// Rejecting set of votes
+	RejectSetVotes(u32),
+	/// Equivocation
+	Equivocation(u32),
+	/// Invalid Vote
+	InvalidVote(u32),
+	/// Invalid block
+	InvalidBlock(u32),
+	/// Parachain Invalid validity statement
+	ParachainInvalidity(u32),
```
gavofyork: Substrate Node has no notion of parachains.

srml/session/src/lib.rs
```diff
@@ -200,22 +201,33 @@ pub trait OneSessionHandler<ValidatorId> {
 	/// The key type expected.
 	type Key: Decode + Default + TypedKey;
 
-	fn on_new_session<'a, I: 'a>(changed: bool, validators: I, queued_validators: I)
-		where I: Iterator<Item=(&'a ValidatorId, Self::Key)>, ValidatorId: 'a;
+	fn on_new_session<'a, I: 'a>(
+		session_index: (SessionIndex, SessionIndex),
```
gavofyork: 👍

srml/support/src/traits.rs
```diff
@@ -640,3 +640,46 @@ pub trait ChangeMembers<AccountId> {
...
+
+	/// Performs the actual slashing based on severity
+	///
+	/// Return the slashes entities which may not be the same as `to_slash`
+	/// because history slashes of the same kind may be included
+	fn do_slash(to_slash: Misbehaved, severity: Severity, kind: Kind, timestamp: Timestamp)
+		-> Result<(Self::SlashedEntries, Self::SlashedAmount), ()>;
+}
+
+/// A generic trait for paying out rewards
```
gavofyork: it's only for paying out rewards to reporters. not for paying block rewards, finalisation rewards, parachain rewards, ...

gavofyork:
it's looking better, however there's still a structural issue with that fat `enum` which expects you to manually enumerate all possible offences. instead, just restrict offences to types that implement a trait to an associated type of the configuration trait. the offence trait can include an ID and window size as well as a function to determine how much the slash should be.

niklasad1:
Thanks Gav

However, I'm not sure that `fn slash_amount(offender_count: u32, balance: u32) -> Self::Balance` is what we want because it would require us to have an API to retrieve the entire slashing history for a certain kind (currently handled by slash mod) and ReportSlash would need the concrete type of the `slashing victim` and its exposure.

Thus, `ReportSlash` impl would be more complex and I have tried to keep it as simple as possible so far(i.e, just pass data and call some functions)

Well, I can try to implement it and so it how goes...

PR #3138, opened 2019-07-17
Operational Transaction; second attempt

Kianenigma:
to be merged with #3102 

As mentioned: Instead of passing down merely a weight down the line: 

```rust 

/// A bundle of static meta information collected from the `#[weight = $x]` meta tags.
#[derive(Clone, Copy, Default)]
pub struct TransactionInfo {
	/// Weight of this transaction.
	pub weight: Weight,
	/// Class of this transaction.
	pub class: DispatchClass,
}
```

If all looks good I should finalize it by:

```
// TODO: rename this to sth that says: this trait returns a bunch of static meta-information about
// the tx, including but NOT ONLY weight. Also rename #[weight] to #[meta]? + rename weight file to dispatch_info.rs
pub trait Weigh {
	fn weigh(&self) -> TransactionInfo;
}
```


srml/support/src/dispatch.rs
```diff
@@ -1111,12 +1113,18 @@ macro_rules! decl_module {
 		impl<$trait_instance: $trait_name $(<I>, $instance: $instantiable)?> $crate::dispatch::Weigh
 			for $call_type<$trait_instance $(, $instance)?> where $( $other_where_bounds )*
 		{
-			fn weigh(&self) -> $crate::dispatch::Weight {
-				match self {
+			fn weigh(&self) -> $crate::dispatch::TransactionInfo {
+				let weight = match self {
```
Kianenigma: @bkchr do you know why  I get such warnings here: 
```
warning: unreachable statement
    --> srml/support/src/dispatch.rs:1122:5
     |
1122 |                   let class = match self {
     |  _________________^
1123 | |                     $( $call_type::$fn_name($( ref $param_name ),*) =>
1124 | |                         <dyn $crate::dispatch::ClassifyDispatch<( $( & $param, )* )>>::class(&$weight, ($( $param_name, )*)), )*
1125 | |                     $call_type::__PhantomItem(_, _) => { unreachable!("__PhantomItem should never be used.") },
1126 | |                 };
     | |__________________^
```

or 

```
warning: unused variable: `weight`
    --> srml/support/src/dispatch.rs:1117:9
     |
1117 |                   let weight = match self {
     |                       ^^^^^^ help: consider prefixing with an underscore: `_weight`
```

Kianenigma: looks like it is coming from any empty module: 

```
decl_module! {
	pub struct Module<T: Trait> for enum Call where origin: T::Origin { }
}
```

core/sr-primitives/src/weights.rs
```diff
@@ -16,26 +16,61 @@
 
 //! Primitives for transaction weighting.
 //!
-//! Each dispatch function within `decl_module!` can now have an optional
-//! `#[weight = $x]` attribute. $x can be any object that implements the
-//! `Weigh` trait. By default, All transactions are annotated by
-//! `#[weight = TransactionWeight::default()]`.
+//! Each dispatch function within `decl_module!` can have an optional `#[weight = $x]` attribute.
```
DemiMarie-parity: Do we have a line length limit for comments? Rust standard seems to be 80.

srml/balances/src/mock.rs
```diff
@@ -187,3 +189,7 @@ impl ExtBuilder {
 
 pub type System = system::Module<Runtime>;
 pub type Balances = Module<Runtime>;
+
+pub fn info_from_weight(w: Weight) -> TransactionInfo {
+	TransactionInfo { weight: w, ..Default::default() }
+}
```
DemiMarie-parity: Nit: missing newline at EOF

srml/example/src/lib.rs
```diff
@@ -396,19 +396,18 @@ decl_module! {
...
-		// A basic weight is a tuple of `(base_weight, byte_weight)`. Upon including each transaction
-		// in a block, the final weight is calculated as `base_weight + byte_weight * tx_size`.
-		// If this value, added to the weight of all included transactions, exceeds `MAX_TRANSACTION_WEIGHT`,
-		// the transaction is not included. If no weight attribute is provided, the `::default()`
-		// implementation of `TransactionWeight` is used.
-		//
-		// The example below showcases a transaction which is relatively costly, but less dependent on
-		// the input, hence `byte_weight` is configured smaller.
-		#[weight = TransactionWeight::Basic(100_000, 10)]
+		// the _right-hand-side_ value of the `#[weight]` attribute can be any type that implements
```
DemiMarie-parity:
```suggestion
		// The _right-hand-side_ value of the `#[weight]` attribute can be any type that implements
```

srml/system/src/lib.rs
```diff
@@ -755,19 +766,40 @@ impl<T: Trait> Module<T> {
...
-	fn internal_check_weight(weight: Weight) -> Result<(), DispatchError> {
+	fn internal_check_weight(info: TransactionInfo) -> Result<Weight, DispatchError> {
 		let current_weight = Module::<T>::all_extrinsics_weight();
-		let next_weight = current_weight.saturating_add(weight);
-		if next_weight > T::MaximumBlockWeight::get() {
-			return Err(DispatchError::Payment)
+		let added_weight = info.weight;
+		let next_weight = current_weight.saturating_add(added_weight);
+		let limit = match info.class {
+			DispatchClass::Operational => T::MaximumBlockWeight::get(),
```
DemiMarie-parity: Should there be any limit at all for operational transactions?

core/sr-primitives/src/weights.rs
Kianenigma: I usually do the same as code: 100. Is it wrong? Maybe module docs have a shorter limit.

srml/system/src/lib.rs
Kianenigma: There is: the idea is that there is a maximum block weight limit. Normal transaction are only allowed to fill a quarter of that (or any ratio). The operational ones can fill the whole limit. In essence, the limit for operational ones is now 4 times more.

srml/support/src/dispatch.rs
Kianenigma: solved.

core/sr-primitives/src/traits.rs
```diff
@@ -869,33 +871,33 @@ macro_rules! tuple_impl_indexed {
...
+				info: TransactionInfo,
 				len: usize,
 			) -> Result<ValidTransaction, DispatchError> {
-				let aggregator = vec![$($direct::validate_unsigned(weight, len)?),+];
+				let aggregator = vec![$($direct::validate_unsigned(info, len)?),+];
 				Ok(aggregator.into_iter().fold(ValidTransaction::default(), |acc, a| acc.combine_with(a)))
 			}
 			fn pre_dispatch_unsigned(
-				weight: crate::weights::Weight,
+				info: TransactionInfo,
```
gavofyork: `TransactionInfo` -> `DispatchInfo`

core/sr-primitives/src/generic/checked_extrinsic.rs
```diff
@@ -75,25 +75,25 @@ where
...
-			Extra::pre_dispatch_unsigned(weight, len)?;
+			Extra::pre_dispatch_unsigned(info, len)?;
 			None
 		};
 		Ok(self.function.dispatch(Origin::from(maybe_who)))
 	}
 }
 
-impl<AccountId, Call, Extra> Weigh for CheckedExtrinsic<AccountId, Call, Extra>
+impl<AccountId, Call, Extra> DispatchInfo for CheckedExtrinsic<AccountId, Call, Extra>
```
gavofyork: probably need to rename `DispatchInfo` to `GetDispatchInfo` and `dispatch_info` to `get_dispatch_info` so that `TransactionInfo` (which is misnamed, since it's not necessarily about a transaction) can be `DispatchInfo`.

core/sr-primitives/src/weights.rs
gavofyork: 100 is fine, i think.

core/sr-primitives/src/weights.rs
```diff
@@ -44,38 +79,44 @@ pub trait WeighData<T> {
...
 ///
-/// A user may pass in any other type that implements [`Weigh`]. If not, the `Default`
-/// implementation of [`TransactionWeight`] is used.
-pub enum TransactionWeight {
-	/// Basic weight (base, byte).
-	/// The values contained are the base weight and byte weight respectively.
+/// A user may pass in any other type that implements the correct traits. If not, the `Default`
+/// implementation of [`WeightedTransaction`] is used.
+#[derive(Clone, Copy)]
+pub enum WeightedTransaction {
```
gavofyork: Rename `SimpleDispatchInfo`

gavofyork: Rename `SimpleDispatchInfo`

core/sr-primitives/src/weights.rs
```diff
@@ -44,38 +79,44 @@ pub trait WeighData<T> {
...
+	/// A fixed-weight transaction. No dependency on state or input.
 	Fixed(Weight),
```
gavofyork: Rename `FixedNormal`

core/sr-primitives/src/weights.rs
```diff
@@ -44,38 +79,44 @@ pub trait WeighData<T> {
...
-	/// Maximum fee. This implies that this transaction _might_ get included but
-	/// no more transaction can be added. This can be done by setting the
-	/// implementation to _maximum block weight_.
-	Max,
-	/// Free. The transaction does not increase the total weight
-	/// (i.e. is not included in weight calculation).
-	Free,
```
gavofyork: Reintroduce as `FreeNormal`

core/sr-primitives/src/weights.rs
```diff
@@ -44,38 +79,44 @@ pub trait WeighData<T> {
...
+/// A user may pass in any other type that implements the correct traits. If not, the `Default`
+/// implementation of [`WeightedTransaction`] is used.
+#[derive(Clone, Copy)]
+pub enum WeightedTransaction {
+	/// A fixed-weight transaction. No dependency on state or input.
 	Fixed(Weight),
-	/// Maximum fee. This implies that this transaction _might_ get included but
-	/// no more transaction can be added. This can be done by setting the
-	/// implementation to _maximum block weight_.
-	Max,
```
gavofyork: Reintroduce as `MaxNormal`

core/sr-primitives/src/weights.rs
```diff
@@ -44,38 +79,44 @@ pub trait WeighData<T> {
...
-	/// Free. The transaction does not increase the total weight
-	/// (i.e. is not included in weight calculation).
-	Free,
+	/// An operational transaction. Still incurs a weight-fee but typically has a higher priority.
+	Operational(Weight),
```
gavofyork: Change to `FixedOperational`, and introduce `MaxOperational` and `FreeOperational`.

node/executor/src/lib.rs
```diff
@@ -80,11 +80,10 @@ mod tests {
 
 	type TestExternalities<H> = CoreTestExternalities<H, u64>;
 
-	// TODO: fix for being charged based on weight now.
 	fn transfer_fee<E: Encode>(extrinsic: &E) -> Balance {
 		<TransactionBaseFee as Get<Balance>>::get() +
-			<TransactionByteFee as Get<Balance>>::get() *
-			(extrinsic.encode().len() as Balance)
+		<TransactionByteFee as Get<Balance>>::get() *
```
gavofyork: you reverted my whitespace fix. these two lines are run-on and need to be indented.

srml/system/src/lib.rs
gavofyork: @AlistairStewart is this right? ^^^

Kianenigma: I think @lsaether @keorn must verify this. But based on my discussion with them it is certainly correct that:

> If a block has X weight units available, in normal conditions, we only allow 1/4 to be filled.

What's new is that:
  - What we added as operational txs that can spike to the maximum. This was concluded based on our discussion @gavofyork and not in the specs. But I think it is a reasonable solution for the problem of _always allowing important reports to get in_.
  - We should also allow tipped transactions to go beyond the limit (specs clearly say that). But since we don't have realistic weight numbers yet it is hard to know exactly how. Although we can also simply add a fixed number of extra weight units to the threshold of tipped transactions and let them compete for it.

srml/balances/src/lib.rs
```diff
@@ -1150,12 +1153,33 @@ where
...
+	///   - _length-fee_: This is the amount paid merely to pay for size of the transaction.
+	///   - _weight-fee_: This amount is computer based on the weight of the transaction. Unlike
+	///      size-fee, this is not input dependent and reflects the _complexity_ of the execution
+	///      and the time it consumes.
+	///   - (optional) _tip_: if included in the transaction, it will be added on top. Only signed
+	///      transactions can have a tip.
+	fn compute_fee(len: usize, weight: Weight, tip: T::Balance) -> T::Balance {
+		// length fee
+		let len = T::Balance::from(len as u32);
+		let len_fee = T::TransactionBaseFee::get()
```
gavofyork: should be mutated based on `DispatchInfo`. i.e. introduce `DispatchInfo::get_base_fee(len: usize)`. for `Free` dispatch classes, this could then be reduced to zero.

Kianenigma: I thought of such a design a while ago but the problem for that would be that all of the parameters need to be passed in, since the `DispatchInfo` does not have access to them. It will be: `DispatchInfo::get_base_fee<B>(len: usize, base_fee: B, byte_fee: B)`

srml/balances/src/lib.rs
```diff
@@ -1165,34 +1189,49 @@ impl<T: Trait<I>, I: Instance> rstd::fmt::Debug for TakeFees<T, I> {
...
-use primitives::traits::{DispatchError, SaturatedConversion};
-use primitives::transaction_validity::ValidTransaction;
-use primitives::weights::Weight;
-
 impl<T: Trait<I>, I: Instance + Clone + Eq> SignedExtension for TakeFees<T, I> {
 	type AccountId = T::AccountId;
 	type AdditionalSigned = ();
 	fn additional_signed(&self) -> rstd::result::Result<(), &'static str> { Ok(()) }
 
+	fn pre_dispatch(
```
gavofyork: why introduce this? by default it just calls `validate`.

srml/balances/src/lib.rs
Kianenigma: (unless if we want to move fee-related associated constants stuff into to `core`, which I don't see good)

srml/balances/src/lib.rs
Kianenigma: Because I am proposing that maybe the logic in them is slightly different: In validate:
- Can just call `ensure_can_withdraw`.
- Omit the imbalance stuff.

If not reasonable, then it should be reverted and only use `validate()`.

srml/balances/src/lib.rs
```diff
@@ -1150,12 +1153,33 @@ where
...
+	#[cfg(feature = "std")]
 	pub fn from(fee: T::Balance) -> Self {
 		Self(fee)
 	}
+
+	/// Compute the final fee value for a particular transaction.
+	///
+	/// The final fee is composed of:
+	///   - _length-fee_: This is the amount paid merely to pay for size of the transaction.
+	///   - _weight-fee_: This amount is computer based on the weight of the transaction. Unlike
```
Kianenigma:
```suggestion
	///   - _weight-fee_: This amount is computed based on the weight of the transaction. Unlike
```

srml/system/src/lib.rs
```diff
@@ -755,19 +766,54 @@ impl<T: Trait> Module<T> {
...
+	}
+
+	/// Checks if the current extrinsic can fit into the block with respect to block length limits.
+	///
+	/// Upon successes, it returns the new block length as a `Result`.
+	fn check_block_length(_info: DispatchInfo, len: usize) -> Result<u32, DispatchError> {
+		let current_len = Module::<T>::all_extrinsics_len();
+		let added_len = len as u32;
+		let next_len = current_len.saturating_add(added_len);
+		if next_len > T::MaximumBlockSize::get() {
```
gavofyork: probably want to do the same as with weight and have 3/4 reserved for operational transactions.

srml/balances/src/lib.rs
Kianenigma: I think the best is to set the design as follows: 

Signed Extension only sees `DispatchClass` and won't interact with `SimpleDispatchInfo` directly (since it actually can't see it). If you want _truly zero-cost transactions_: This should be reflected in a new `DispatchClass::UserFree , DispatchClass::OperationalFree`. A bit verbose but keeps things where they belong.

btw. I am not sure if we want such transactions actually. You can have a `zero-weight-fee`, the decision to bring back the traditional `base + byte` is what is preventing us from having it. It might still be sensible to say: _you might get away with the weight fee, but you always pay byte fee (it should be smaller anyhow)_.

srml/system/src/lib.rs
Kianenigma: Sure 👍

srml/balances/src/lib.rs
gavofyork: fair enough, but then maybe there should be an auxilliary to deduplicate the code.

srml/system/src/lib.rs
AlistairStewart: This wasn't my understanding at all. We were thinking of havin reserved space or even no limit for important transactions. But the 1/4 was the target fullness that we adjusted transaction fees for, nit a hard limit. Tips just prioitize transactions in case the maximum has been reached. But maybe @keorn or @AlfonsoCev have a later design.

PR #3146, opened 2019-07-19
Switch consensus crates to new futures

tomaka:
I had this PR almost ready for a long time, was just waiting for #3117 to be merged.

Switches `consensus/*` to new futures.
Most notably, however, `start_aura` and `start_babe` still return an old future (by building a new future and putting the compatibility layer around), so that we avoid breaking everything.

cc #3099


core/consensus/common/src/lib.rs
```diff
@@ -75,7 +75,7 @@ pub trait Proposer<B: BlockT> {
 	/// Error type which can occur when proposing or evaluating.
 	type Error: From<Error> + ::std::fmt::Debug + 'static;
 	/// Future that resolves to a committed proposal.
-	type Create: IntoFuture<Item=B, Error=Self::Error>;
+	type Create: Future<Output = Result<B, Self::Error>>;
```
tomaka: Note: this change is because the `IntoFuture` trait no longer exists.

core/consensus/common/src/import_queue/basic_queue.rs
```diff
@@ -65,11 +62,27 @@ impl<B: BlockT> BasicQueue<B> {
...
+			.name_prefix("import-queue-worker-")
+			.pool_size(1)
+			.create()
+			.ok();
+
+		let manual_poll;
+		if let Some(pool) = &mut pool {
+			// TODO: this expect() can be removed once
+			// https://github.com/rust-lang-nursery/futures-rs/pull/1750 is merged and deployed
+			pool.spawn(future).expect("ThreadPool can never fail to spawn tasks; QED");
```
tomaka: Note that it is unfortunately no longer possible to properly recover from task spawning errors.
I went for creating a local `ThreadPool` here so that we are guaranteed to never have any error (as in: if creating the pool succeeds, then spawning the task will also succeed).
In the future, we should probably pass some task spawning utility by parameter, but the idiomatic way to do that remains to be discovered.

rphmeier:
I'd prefer to merge the BABE PRs first, if that's OK.

PR #3147, opened 2019-07-19
SS58 versioning (Network IDs)

gavofyork:
This introduces the ability to customise which version code is used for SS58 via a network configuration.

It will still default to 42 (Substrate generic version) within Substrate and Node, but Polkadot can now switch it to use Polkadot (0) or Kusama (2). Subkey has been changed to allow it to be configured on the CLI with the `--network` option.


subkey/src/main.rs
```diff
@@ -18,15 +18,18 @@
...
-use substrate_primitives::{ed25519, sr25519, hexdisplay::HexDisplay, Pair, Public, crypto::Ss58Codec, blake2_256};
+use substrate_primitives::{
+	ed25519, sr25519, hexdisplay::HexDisplay, Pair, Public,
+	crypto::{Ss58Codec, set_default_ss58_version, Ss58AddressFormat}, blake2_256
+};
 use parity_codec::{Encode, Decode, Compact};
 use sr_primitives::generic::Era;
 use node_primitives::{Balance, Index, Hash};
-use node_runtime::{Call, UncheckedExtrinsic, BalancesCall};
+use node_runtime::{Call, UncheckedExtrinsic, /*CheckNonce, TakeFees, */BalancesCall};
```
shawntabrizi: Commented code

subkey/src/main.rs
```diff
@@ -120,7 +130,7 @@ fn execute<C: Crypto>(matches: clap::ArgMatches) where
 			let sig = pair.sign(&message);
 			println!("{}", hex::encode(&sig));
 		}
-		("transfer", Some(matches)) => {
+		/*("transfer", Some(matches)) => {
```
shawntabrizi: Commented code

subkey/src/main.rs
gavofyork: yeah - it's there as a convenience measure - it'll be immediately uncommented once #3102 is in.

subkey/src/main.rs
gavofyork: same as above.

gavofyork:
doesn't need to be merged with the commented code, but should be reviewed and okayed on the proviso that it'll be merged with the commented code uncommented or otherwise removed.

PR #3151, opened 2019-07-20
docs: add srml readme, clarify comment in staking

4meta5:
We might as well move the relevant part of [the high-level Substrate Runtime docs](https://substrate.dev/docs/en/runtime/substrate-runtime-module-library) to the [SRML readme](https://github.com/paritytech/substrate/blob/master/srml/README.adoc). cc @shawntabrizi 

Also, I noticed that the `srml/staking/lib.rs` requires a `SessionInterface` because it claims that `Staking` sets the `ValidatorId` of the `session::Trait`. `Staking` seems to set the `ValidatorIdOf` for `session::Trait` in [`node/runtime`](https://github.com/paritytech/substrate/blob/cf98501f92a98ab7133829a746128be98aea0117/node/runtime/src/lib.rs#L218);`ValidatorId = AccountId` in `node/runtime`. I changed a comment in `staking` to reflect this, but, if there is something I'm missing here, this change can easily be reverted.

### DIFF
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -75,7 +75,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 115,
-	impl_version: 116,
+	impl_version: 115,
 	apis: RUNTIME_API_VERSIONS,
 };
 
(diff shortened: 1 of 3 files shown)


node/runtime/src/lib.rs
```diff
@@ -75,7 +75,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 115,
-	impl_version: 115,
+	impl_version: 116,
```
shawntabrizi: You don't need to do any change here since you do not touch any runtime logic.

shawntabrizi: Just ignore the failed test

Kianenigma:
@shawntabrizi is the content ok?

node/runtime/src/lib.rs
Kianenigma:
```suggestion
	impl_version: 115,
```

PR #3157, opened 2019-07-21
Weight annotation.

### LINKED ISSUE #2431: Determine transaction (fee) weights by transaction types
This issue is related to relay-chain transaction fees [research here](http://research.web3.foundation/en/latest/polkadot/Token%20Economics/#relay-chain-transaction-fees). 

**Motivation**
Different types of transactions will have different fee levels. This fee differentiation is used to reflect the different costs in resources incurred by transactions, and to encourage/discourage certain types of transactions. Thus, we need to analyze the resource usage of each type of transaction, to adjust the fees (to be done).

**Recommended Steps**

To start, we need some way to benchmark transaction execution in runtime. This will help to estimate the costs of Polkadot runtime transactions for users. 

0. Set up some benchmarking tools/primitives

1. Go through every module and evaluate all the transaction types and their complexity. e.g. writes to storage, complex reads, loops, etc. Identify transactions with bad or unbounded complexity (memory, storage, compute). 

2. Refactor code to improve the bounds ^

cc: @keorn @Kianenigma @mattrutherford

Kianenigma:
TLDR; Closes #2431 

This is potentially the final phase of a bunch of PRs regarding weight -> fee -> updates over time. Especially during the final phase, we changed a thing or two during implementation that might not be 100% in line with the specs. Furthermore, the spec itself is not sophisticated and does not cover all the details. First, quoting the important parts of the [specs](https://research.web3.foundation/en/latest/polkadot/Token%20Economics/#relay-chain-transaction-fees):

### Reflection on research

> There will be several types of transactions, with different fee levels. This fee differentiation is used to reflect the different costs in resources incurred by transactions, and to encourage/discourage certain types of transactions.

 - [x] We currently have different variations of `Normal` and `Operational` dispatches. Normal ones have less quota, pay the full fee and have a priority proportional to their weight. Operational ones can use the full block weight limit (extra 25% atm but can be changed), don't pay the `size-fee` (again, can be changed) and have _maximum_ priority.

_Note that more priority -> more inclusion **probability**, not a guarantee_.

Within each level, different fee levels, as the spec requires, can be defined by simply tuning the weight value in the `Fixed` variant.

> There will be an additional space in each block that is reserved only for crucial or urgent transactions, so that they can be included even if the block is full. Fishermen txs that report misbehaviours would be an example of crucial transaction.

- [x] Simply `Operational` transaction.

> A transaction fee includes a base fee and a variable fee that is proportional to the number of bytes to be put on chain.

- [x] Weight-wise, we dropped this half-way through the development and only have fixed weights. It is quite hard to know how much data a transaction will _put on chain_ (e.g. if a key does not exist a storage `::insert` is actually putting something while if it already exists, it is not adding any data to the state). Nonetheless, we can add this along the way since the weight interface has access to the input parameters of each dispatch and can examine them. I imagine that we would, at some point, add something like `InputOrderOfNormal(X)` as opposed to `FixedNormal(Y)` weight types where the weight of the former is calculated as `size_of_some_input_parameter ^ X` or `size_of_some_input_parameter * X`.

Also, we brought back the old substrate fee system which works next to the weight-fee system and takes the length of the transaction into account (`base + size * num_bytes`). This is promising since now that weights are _input-independent_, we have some mechanism to protect (and charge money) for processing large transactions. To sum, the final fee is something like: `len_fee + weight_fee [+ tip]`.

>  Slow adjusting mechanism.

- [x] Works well. The only question is where to configure the target fullness level (25% -- currently hardcoded as a constant). Also, note that this target fullness level is entirely independent of the quota distinction of `Normal` and `Operational`. It simply looks at the full block weight and tweaks the weight fee based on being above or below a threshold.

```                                                                    
                                           Block                                         
                              +-------------------------------+                          
                              |                               |                          
                              |                25%            |                          
                              |    Reserved for Operational   |                          
                              --------------------------------|                          
                              |                               |                          
                              |                               |                          
                              |                               |                          
                              |                               |                          
                              |                               |                          
                              |                               |                          
                              |                               |    25%                   
                              |                               | ----------               
                              |                               |weight update target         
                              |                               |                          
                              |                               |                          
                              +-------------------------------+                          
```

> The transaction fee is considered a base price. There will be a different field in the transaction called tip, and a user is free to put any amount of tokens in it or leave it at zero. Block producers can charge both the fee and tip, __so they have an incentive to include transactions with large tips__.

- [x] we do this by adding the fee to the priority which, as mentioned, increases the chance of inclusion. Yet, I think this has to be researched more. We are trivially adding two numbers (`Balance`, `Priority`) which may or may not be of the same scale. 

> ~~Not in specs but I remember from some conversations that we might want to allow tips to also allow transactions to go above their quota limit~~

- [x] This was a wrong statement. We don't have this and don't want it. Tips only give you a better chance of inclusion. Limits will stay fixed.

### Technical challenges left

As mentioned above, there are (at least) two related parameters (`system`'s 3/4 for Normal dispatches and `WeightMultiplierUpdate`'s target) that need to be configured somewhere and I wasn't certain where they belong to. for the former, it makes sense for it to be an associated constant in the system module. But, I want approval on it first since it is a daunting refactor of all tests that use the system module.

### Weight values.

The spec only defines the notion of weights but does not indicate how to compute them. We used a minimal methodology to do so with one major goal in mind: be as conservative as possible and take the worst-case scenario of everything into account.

The simple methodology for obtaining weight values is as follows: 

- We benchmarked as many of the dispatchable functions in substrate/Polkadot runtime as possible.
  - Benchmarks are using the native execution environment. 
  - Column: `#1 Native Runtime Exec Time`

- We assume the target block time to be `2` seconds.

- We assume that we want each block to be at most full by 25% (same as the weight update target mentioned before) to be conservative. `target_block_fullness = 0.25`.

- Due to 1- errors in benchmarks 2- the fact that benchmarks are not really measuring the worst-case scenario 3- caching 4- non-runtime code which is not measured (and evidence showed that it was quite time consuming), we immediately take `1` second apart for any unexpected behavior or delays. Hence, the _available time for dispatchable functions to execute_ is assumed to be `1` second per block.

- Since weights are (or better to say: should be) a representation of the time it takes to execute a function, we then apply the `target_block_fullness` to this time, yielding `250_000_000ns` as the `available_block_execution_time`.

- We define the total weight units available per block to be `1_000_000_000`. Applying the same `target_block_fullness `:
  - `total_block_weight = 1_000_000_000`.
  - `available_block_weight = 250_000_000`.

- We then multiply all execution times values by a _wasm-factor_, an estimate of how slower web-assembly can be. Currently, this value is `5`.
  - Column: `Ratio of available block time` 

- We divide the benchmarked execution time of each function by `available_block_execution_time` to obtain what _ratio_ of the block time it consumes.
  - Column: `Wasm Runtime Exec Time` 

- We multiply each obtained _wasm-execution-time ratio_ per function, by the obtained `available_block_weight`. This yields the actual weight of each dispatch.
  - Column: `Basic Weight` 
 
- We then round + scale-up these numbers by multiplying them to a value between [1,10] manually, in order to:
  - Take the actual implementation into account. Some transactions get a bigger bump here in weight since we _know_ from the code that they are expensive and the benchmarks might not showcase that.
  - Be a bit more pessimistic and round __up__ everything. It is better to be safe than sorry.
  - Column: `Basic weight - manual`

- We choose a very small number to be the fee for the absolute cheapest (but not free) transactions: `0.1 Cent` given that `1 Unit` is `100 Dollars`. Based on this, the weights should be scaled one last time to obtain this final fee. This number happens to be `1000`.
  - This is not yet implemented in the code yet.

- Finally, note that:
  - In the code the weights are factored by `1000` for simplicity. As mentioned, there should be a factor of `1000` added as well to correct the fee values.
  - 
 
The above is the outcome of collaboration with @keorn and @mattrutherford and mostly their work. The final results are here: 
https://docs.google.com/spreadsheets/d/1h0RqncdqiWI4KgxO0z9JIpZEJESXjX_ZCK6LFX6veDo/edit?usp=sharing

Screenshot 2019-07-21 at 12 42 20

The `# To Fill Target Block`, `# To Exhaust Block` and `Weight Fee in Cents` are interesting.

Of course, this is merely something to start with and I strongly believe that this matter should be dedicatedly researched in more detail.

I will rebase this PR to master if #3102 gets merged in.


node/runtime/src/impls.rs
```diff
@@ -44,27 +44,38 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
...
+// CRITICAL NOTE: The system module maintains two constants: a _maximum_ block weight and a
+// _ratio_ of it yielding the portion which is accessible to normal transactions (reserving the rest
+// for operational ones). `TARGET_BLOCK_FULLNESS` is entirely independent and the system module is
+// not aware of if, nor should it care about it. This constant simply denotes on which ratio of the
+// _maximum_ block weight we tweak the fees. It does NOT care about the type of the dispatch.
+//
+// For the system to be configured in a sane way, `TARGET_BLOCK_FULLNESS` should always be less than
+// the ratio that `system` module uses to find normal transaction quota.
+/// The block saturation level. Fees will be updates based on this value.
+pub const TARGET_BLOCK_FULLNESS: Perbill = Perbill(1_000_000_000 / 4);
```
Kianenigma: This is one of the items that we might want to place elsewhere. I think it actually looks okay to be in this crate `node/runtime` but preferably in a separate `consts.rs` file? We will probably have other configurable constants in the future as well.

node/runtime/src/impls.rs
```diff
@@ -44,27 +44,38 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
...
 impl Convert<(Weight, WeightMultiplier), WeightMultiplier> for WeightMultiplierUpdateHandler {
 	fn convert(previous_state: (Weight, WeightMultiplier)) -> WeightMultiplier {
 		let (block_weight, multiplier) = previous_state;
-		// CRITICAL NOTE: what the system module interprets as maximum block weight, and a portion
-		// of it (1/4 usually) as ideal weight demonstrate the gap in block weights for operational
-		// transactions. What this weight multiplier interprets as the maximum, is actually the
-		// maximum that is available to normal transactions. Hence,
-		let max_weight = <MaximumBlockWeight as Get<u32>>::get() / 4;
-		let ideal_weight = (max_weight / 4) as u128;
+		let max_weight = <MaximumBlockWeight as Get<u32>>::get();
```
Kianenigma: Note that here we are simply getting all the weights, not just the `Normal` portion.

node/runtime/src/impls.rs
```diff
@@ -214,7 +224,9 @@ mod tests {
 				WeightMultiplierUpdateHandler::convert((i, WeightMultiplier::default()));
 			});
 
-		vec![10 * mb, Weight::max_value() / 2, Weight::max_value()]
+		// Some values that are all above the target and will cause an increase.
+		let target_weight = TARGET_BLOCK_FULLNESS * <MaximumBlockWeight as Get<u32>>::get();
```
Kianenigma:
```suggestion
		let target_weight = target()
```

node/runtime/src/lib.rs
```diff
@@ -117,8 +117,8 @@ pub const DAYS: Moment = HOURS * 24;
 
 parameter_types! {
 	pub const BlockHashCount: BlockNumber = 250;
-	pub const MaximumBlockWeight: Weight = 4 * 1024 * 1024;
-	pub const MaximumBlockLength: u32 = 4 * 1024 * 1024;
+	pub const MaximumBlockWeight: Weight = 1_000_000_000;
+	pub const MaximumBlockLength: u32 = 5 * 1024 * 1024;
```
Kianenigma: Also note that the normal size limit of the node is now 5mb.

srml/session/src/lib.rs
```diff
@@ -355,6 +356,7 @@ decl_module! {
 		/// - O(log n) in number of accounts.
 		/// - One extra DB entry.
 		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedOperational(150)] // TODO: this is operational?
```
Kianenigma: is it?

srml/sudo/src/lib.rs
```diff
@@ -116,6 +117,7 @@ decl_module! {
 		/// - Limited storage reads.
 		/// - No DB writes.
 		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedOperational(1_000)]
```
Kianenigma: Interestingly, I got this number by averaging all other functions that were benchmarked.

srml/system/src/lib.rs
```diff
@@ -218,24 +219,35 @@ decl_module! {
 			Self::deposit_event_indexed(&[], event);
 		}
 
+		/// A big dispatch that will disallow any other transaction to be included.
+		// TODO: this must be preferable available for testing really (not possible at the moment).
+		#[weight = SimpleDispatchInfo::MaxOperational]
+		fn fill_block(origin) {
```
Kianenigma: I merely added this for a test (which is not that important actually). It should probably be removed unless if someone can find another use case for it OR claim that as a root function it is safe to exist?

srml/system/src/lib.rs
```diff
@@ -792,26 +805,28 @@ pub struct CheckWeight<T: Trait + Send + Sync>(PhantomData<T>);
...
-	/// a quarter.
-	fn get_dispatch_limit_divisor(class: DispatchClass) -> Weight {
+	/// a portion.
+	fn get_dispatch_limit_ratio(class: DispatchClass) -> Perbill {
 		match class {
-			DispatchClass::Operational => 1,
-			DispatchClass::Normal => 4,
+			DispatchClass::Operational => Perbill::one(),
+			// TODO: this must be some sort of a constant.
+			DispatchClass::Normal => Perbill::from_percent(75),
```
Kianenigma: This for sure has to be placed elsewhere.

core/sr-primitives/src/lib.rs
```diff
@@ -170,7 +170,7 @@ pub type ConsensusEngineId = [u8; 4];
 /// Permill is parts-per-million (i.e. after multiplying by this, divide by 1000000).
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug, Ord, PartialOrd))]
 #[derive(Encode, Decode, Default, Copy, Clone, PartialEq, Eq)]
-pub struct Permill(u32);
+pub struct Permill(pub u32);
```
bkchr: Why do you make this public?

I would rather add a method to get the value. (Otherwise someone may screws up the internal state)

Kianenigma: I forgot to put a comment about the why of it myself: to use it as a const value.

srml/system/src/lib.rs
Kianenigma: placed in `system::Trait::AvailableBlockRatio`.

node/runtime/src/impls.rs
Kianenigma: did as explained for now. Moved some constants to a new file.

gavofyork: language seems to flip between 1/4 of block is reserved for operation transactions and 3/4 of block is reserved. here, for example `TARGET_BLOCK_FULLNESS` is 25%, indicating 3/4 reserved for operational transactions. however, in the diagram at the top of this page, it says:

>  25% ... Reserved for Operational 

Please consult research team and get this consistent and correct.

gavofyork:
NOTE: weights have nothing to do with storage volume. There need to be separate, manual, economic (or other) safeguards against state bloat (which are already in place in existing modules, but this needs to be audited). Transaction (length-dependent) fees and size-limits act against transaction data bloat. This leaves weights to handle just computation and i/o, which together amount to *execution time*. We could try to separate them into CPU and i/o, but given that they can differ so much between nodes even individually, I don't think it makes all that much sense.

srml/system/src/lib.rs
gavofyork: can't we feature-gate it on tests?

gavofyork:
Target block *execution* time is likely to be around 5 seconds; there will be a safeguard of an additional 5 seconds (in the case of a block that the majority believes takes longer to execute than 10s then it won't be imported and the validator behind it is slashed).

gavofyork:
Tips should *not* have any effect on getting around limits.

srml/system/src/lib.rs
Kianenigma: I never tried it but I recall that apparently, we don't allow most `#[cfg]`s in `decl_module!`. As the comment says: 

> // TODO: this must be preferable available for testing really (__not possible at the moment__).


will try.

node/runtime/src/impls.rs
Kianenigma: This is exactly what I addressed in the comment above this line:

```
// CRITICAL NOTE: The system module maintains two constants: a _maximum_ block weight and a
// _ratio_ of it yielding the portion which is accessible to normal transactions (reserving the rest
// for operational ones). `TARGET_BLOCK_FULLNESS` is entirely independent and the system module is
// not aware of if, nor should it care about it. This constant simply denotes on which ratio of the
// _maximum_ block weight we tweak the fees. It does NOT care about the type of the dispatch.
```

To rephrase: They are entirely separate. 

System module knows:
  - Each block can be a __total__ of `X weight` units. 
  - It will keep `25%` of this only for operational transactions and the rest for normal.

Weight Multiplier:
  - Asks the system module for the __total__ weight units per block. 
  - Assumes that if we are `25%` of this limit then we are _relatively congested_ and should slightly increase the weight fees. 

The only adjustment I see is that the Weight multiplier should query the system module for the __available-to-normal-txs__ portion of weights and set its goal for 25% of that. But I still think it makes more sense like this. Weight multiplier should not care about `Operational` (heck, it even doesn't know they exist). Only note will be, as commented in the code: 

```
// For the system to be configured in a sane way, `TARGET_BLOCK_FULLNESS` should always be less than
// the ratio that `system` module uses to find normal transaction quota.
```

Agreeing that the research team should verify the numbers.

core/sr-primitives/src/lib.rs
```diff
@@ -411,13 +421,17 @@ impl Fixed64 {
 
 	/// Performs a saturated multiply and accumulate.
 	///
-	/// Returns `n + (self * n)`.
-	pub fn saturated_multiply_accumulate(&self, int: u32) -> u32 {
```
Kianenigma: I should try and generalize this into accepting any type.

core/sr-primitives/src/lib.rs
```diff
@@ -411,11 +421,12 @@ impl Fixed64 {
 
 	/// Performs a saturated multiply and accumulate.
 	///
-	/// Returns `n + (self * n)`.
+	/// Returns a saturated `n + (self * n)`.
+	/// TODO: generalize this to any weight type.
```
Kianenigma: It is also okay as-is since it is very unlikely for us to decide to change the weight type but still a good +

node-template/runtime/src/lib.rs
```diff
@@ -102,8 +96,9 @@ pub fn native_version() -> NativeVersion {
 
 parameter_types! {
 	pub const BlockHashCount: BlockNumber = 250;
-	pub const MaximumBlockWeight: Weight = 4 * 1024 * 1024;
-	pub const MaximumBlockLength: u32 = 4 * 1024 * 1024;
+	pub const MaximumBlockWeight: Weight = 1_000_000;
```
Kianenigma: As updated in the sheet, weight scale is now `[10, 1_000_00]` and the weight-to-fee factor `10^6`.

node/executor/src/lib.rs
```diff
@@ -1019,6 +1045,83 @@ mod tests {
...
+			let weight = default_transfer_call().get_dispatch_info().weight;
+			let weight_fee = WeightToFee::convert(weight);
+
+			// we know that weight to fee multiplier is effect-less in block 1.
+			assert_eq!(weight_fee as Balance, MILLICENTS);
+			balance_alice -= weight_fee;
+
+			balance_alice -= tip;
+
+			// TODO: why again??? creation fee should not be deducted here.
```
Kianenigma: @gavofyork do you know why this keeps happening? even though Bob has some initial balance, the creation fee is deducted from alice upon sending the transfer. It is also the same for some other tests.

At some point I put a log here and this was `false`. no clue what's going on.
https://github.com/paritytech/substrate/blob/master/srml/balances/src/lib.rs#L870

node/runtime/src/constants.rs
```diff
@@ -0,0 +1,52 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
```
Kianenigma: Not sure if it is a good idea to have this file. I think it is neater.

node/runtime/src/impls.rs
```diff
@@ -20,8 +20,16 @@ use node_primitives::Balance;
 use runtime_primitives::weights::{Weight, WeightMultiplier};
 use runtime_primitives::traits::{Convert, Saturating};
 use runtime_primitives::Fixed64;
-use support::traits::Get;
-use crate::{Balances, MaximumBlockWeight};
+use support::traits::{Get, OnUnbalanced, Currency};
+use crate::{Balances, Authorship, MaximumBlockWeight, NegativeImbalance};
+use crate::constants::fee::TARGET_BLOCK_FULLNESS;
+
+pub struct Author;
```
Kianenigma: As promised, I moved some other structs that implement stuff from the top level `lib.rs` here. Same rationale as `constants.rs`: trying to keep the `lib.rs` file concise and tidy.

srml/balances/src/lib.rs
```diff
@@ -1171,19 +1182,23 @@ impl<T: Trait<I>, I: Instance> TakeFees<T, I> {
...
 			Zero::zero()
 		};
 
-		// weight fee
-		let weight = info.weight;
-		let weight_fee: T::Balance = <system::Module<T>>::next_weight_multiplier().apply_to(weight).into();
+		let weight_fee = {
+			// cap the weight to the maximum defined in runtime, otherwise it will be the `Bounded`
+			// maximum of its data type, which is not desired.
+			let capped_weight = info.weight.min(<T as system::Trait>::MaximumBlockWeight::get());
```
Kianenigma: this kinda sucks that every signed extension using weight must cap it itself if they want to. This is because in the details of the `WeighData` we don't know the `MaximumBlockWeight`. Might be able to improve it by further abstracting `type Weight: u32` into some other custom type and give it a custom implementation of `Bounded` by the runtime.

Kianenigma: I also do the same in system `CheckWeight`. Might be debatable. 

I had the assumption that the weight fo one individual tx cannot go beyond the weight of the entire block. Even if thee annotationa sets the weight to infinity the runtime checks interpret it as whatever `MaximumBlockWeight` of the node is. 

If this does not exist, then `Max` variants of weight are actually a footgun, using it will cause a much higher fee and won't bring any benefit.

srml/system/src/lib.rs
Kianenigma: Looks like: no, we can't.

Kianenigma:
> Target block execution time is likely to be around 5 seconds; there will be a safeguard of an additional 5 seconds (in the case of a block that the majority believes takes longer to execute than 10s then it won't be imported and the validator behind it is slashed).

Given that the current assumption was 2 seconds, I can either re-calculate them based on 5 or we actually settle for the current number to _really_ be safe with respect to anything that might go wrong.

node/runtime/src/impls.rs
gavofyork: ok cool

core/sr-primitives/src/lib.rs
gavofyork: will need an issue number, then.

gavofyork:
> > Target block execution time is likely to be around 5 seconds; there will be a safeguard of an additional 5 seconds (in the case of a block that the majority believes takes longer to execute than 10s then it won't be imported and the validator behind it is slashed).
> 
> Given that the current assumption was 2 seconds, I can either re-calculate them based on 5 or we actually settle for the current number to _really_ be safe with respect to anything that might go wrong.

yeah fine by me, for now.

node-template/runtime/src/lib.rs
gavofyork: 1_000_000 seems small for a block. why not make it 1_000_000_000 per block to allow more fine-grained transaction weighting?

node/executor/src/lib.rs
gavofyork: more prints needed, then. shouldn't be hard to track down.

srml/session/src/lib.rs
gavofyork: don't think so.

node-template/runtime/src/lib.rs
Kianenigma: Note that this also came hand in hand with the fact that a fixed multiplier of 1000 was removed from weights. So the limit hasn't actually changed at all.

Could easily be reverted, yet, my arguments being: 

- We already have a redundant factor of 10 (smallest weight is 10, changes are within a few hundred) as backup and _being more fine-grained_.
- A bit more readable to skim the weight annotations a and see `weight = 10, weight = 50, weight = 100` rather than a stack of zeros attached to all of them.

srml/support/src/lib.rs
```diff
@@ -220,6 +220,37 @@ macro_rules! __assert_eq_uvec {
 	}
 }
 
+/// Checks that `$x` is equal to `$y` with an error rate of `$error`.
+///
+/// # Example
+///
+/// ```rust
+/// #[macro_use]
+/// # extern crate srml_support;
```
bkchr: Can just be written as:
```
srml_support::assert_eq_error_rate!(10, 10, 0);
srml_support::assert_eq_error_rate!(10, 11, 1);
srml_support::assert_eq_error_rate!(12, 10, 2);
```

srml/support/src/lib.rs
```diff
@@ -220,6 +220,37 @@ macro_rules! __assert_eq_uvec {
...
+/// assert_eq_error_rate!(12, 10, 2);
+/// # }
+/// ```
+///
+/// ```rust,should_panic
+/// #[macro_use]
+/// # extern crate srml_support;
+/// # use srml_support::{assert_eq_error_rate};
+/// # fn main() {
+/// assert_eq_error_rate!(12, 10, 1);
```
bkchr: Same as above.

srml/system/src/lib.rs
bkchr: Can you not use the `test-runtime`?

srml/system/src/lib.rs
```diff
@@ -1069,6 +1092,14 @@ mod tests {
 		GenesisConfig::default().build_storage::<Test>().unwrap().0.into()
 	}
 
+	fn normal_weight_limit() -> Weight {
+		<Test as Trait>::AvailableBlockRatio::get() * <Test as Trait>::MaximumBlockWeight::get()
```
bkchr: Pr.

node/runtime/src/impls.rs
```diff
@@ -130,81 +163,120 @@ mod tests {
...
+		let block_weight = 1000;
+		let mut wm = WeightMultiplier::default();
+		let mut iterations: u64 = 0;
+		loop {
+			let next = WeightMultiplierUpdateHandler::convert((block_weight, wm));
+			wm = next;
+			if wm == WeightMultiplier::from_rational(-1, 1) { break; }
+			iterations += 1;
+		}
+		println!("iteration {}, new wm = {:?}. Weight fee is now zero", iterations, wm);
```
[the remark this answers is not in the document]
Kianenigma: yeah, that's the purpose. And I am adding more. These are not benchmarks, but rather sort of a report. I couldn't think of putting an assertion into it since it is really not checking anything but definitely helpful to have them as reports. 

Though I can add assertions but then it would be a bit based on hardcoded values of `substrate-node` + it will be a damn slow test.

For now I am happy to have them as `ignore` tests that one can run manually to just check things.

Kianenigma:
@bkchr about: 

> fn fill_block(origin) 

Preferably not since these are tests for the exact `substrate-node` runtime (the real runtime). Maybe in the future, some can be duplicated and use `test-runtime` but for now I really prioritize testing the real runtime. 

If the function is not dangerous, it is a help to the test. If not, both the function and the test using it can be removed.

core/sr-primitives/src/lib.rs
Kianenigma: issue number added. case closed for now.

node/executor/src/lib.rs
```diff
@@ -979,12 +1003,14 @@ mod tests {
...
 				},
 				CheckedExtrinsic {
-					signed: Some((charlie(), signed_extra(num_to_exhaust.into(), 0))),
+					signed: Some((charlie(), signed_extra(1, 0))),
 					function: Call::System(system::Call::remark(vec![0; 1])),
 				}
 			]
 		);
 
+		println!("++ Block 1 size: {} / Block 2 size {}", block1.0.encode().len(), block2.0.encode().len());
```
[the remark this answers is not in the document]
Kianenigma: same reason as other prints, if that's acceptable.

node/runtime/src/impls.rs
```diff
@@ -39,32 +47,53 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
...
 impl Convert<(Weight, WeightMultiplier), WeightMultiplier> for WeightMultiplierUpdateHandler {
 	fn convert(previous_state: (Weight, WeightMultiplier)) -> WeightMultiplier {
 		let (block_weight, multiplier) = previous_state;
-		// CRITICAL NOTE: what the system module interprets as maximum block weight, and a portion
-		// of it (1/4 usually) as ideal weight demonstrate the gap in block weights for operational
-		// transactions. What this weight multiplier interprets as the maximum, is actually the
-		// maximum that is available to normal transactions. Hence,
-		let max_weight = <MaximumBlockWeight as Get<u32>>::get() / 4;
-		let ideal_weight = (max_weight / 4) as u128;
+		let max_weight = <MaximumBlockWeight as Get<Weight>>::get();
```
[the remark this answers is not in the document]
Kianenigma: sure 👍

node-template/runtime/src/lib.rs
DemiMarie-parity: @Kianenigma what about using powers of 2 instead?

node/runtime/src/constants.rs
```diff
@@ -0,0 +1,52 @@
...
+//
+// For the system to be configured in a sane way, `TARGET_BLOCK_FULLNESS` should always be less than
+// the ratio that `system` module uses to find normal transaction quota.
+/// Fee-related.
+pub mod fee {
+    pub use runtime_primitives::Perbill;
+
+    /// The block saturation level. Fees will be updates based on this value.
+    pub const TARGET_BLOCK_FULLNESS: Perbill = Perbill::from_const_parts(1_000_000_000 / 4);
+}
```
DemiMarie-parity: Missing newline at EOF

node/runtime/src/impls.rs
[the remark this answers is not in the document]
DemiMarie-parity: @Kianenigma what about using `debug!` from the `log` crate?
```suggestion
		debug!("iteration {}, new wm = {:?}. Weight fee is now zero", iterations, wm);
```

node/executor/src/lib.rs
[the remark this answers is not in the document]
DemiMarie-parity:
```suggestion
		debug!("++ Block 1 size: {} / Block 2 size {}", block1.0.encode().len(), block2.0.encode().len());
```

node/runtime/src/impls.rs
```diff
@@ -130,81 +163,120 @@ mod tests {
...
+		let tx_weight = 1000;
+		let mut wm = WeightMultiplier::default();
+		let mut iterations: u64 = 0;
+		loop {
+			let next = WeightMultiplierUpdateHandler::convert((block_weight, wm));
+			if wm == next { break; }
+			wm = next;
+			iterations += 1;
+			let fee = <Runtime as balances::Trait>::WeightToFee::convert(wm.apply_to(tx_weight));
+			println!(
```
DemiMarie-parity:
```suggestion
			debug!(
```

Kianenigma:
@bkchr only a few of your nits are still to be resolved:
- Having the `fill_block()`.
- Tests that have some prints and are sort of a _simulation_. I feature-gated them in a `stress-test` feature.

Rest seems fine.

subkey/src/main.rs
```diff
@@ -225,7 +225,7 @@ fn execute<C: Crypto>(matches: clap::ArgMatches) where
 			);
 
 			println!("0x{}", hex::encode(&extrinsic.encode()));
```
DemiMarie-parity:
```suggestion
			debug!("0x{}", hex::encode(&extrinsic.encode()));
```

srml/system/src/lib.rs
```diff
@@ -218,24 +225,35 @@ decl_module! {
 			Self::deposit_event_indexed(&[], event);
 		}
 
+		/// A big dispatch that will disallow any other transaction to be included.
+		// TODO: this must be preferable available for testing really (not possible at the moment).
+		#[weight = SimpleDispatchInfo::MaxOperational]
+		fn fill_block(origin) {
```
DemiMarie-parity:
```suggestion
		fn fill_block(origin) {
			#![cfg(test)]
```

See [this Playground example](//play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c375a624822567d2e25002ffeae624b1).

srml/system/src/lib.rs
```diff
@@ -872,8 +893,8 @@ impl<T: Trait + Send + Sync> SignedExtension for CheckWeight<T> {
 		len: usize,
 	) -> Result<ValidTransaction, DispatchError> {
 		// There is no point in writing to storage here since changes are discarded. This basically
-		// discards any transaction which is bigger than the length or weight limit alone, which is
-		// a guarantee that it will fail in the pre-dispatch phase.
+		// discards any transaction which is bigger than the length or weight limit __alone__,which
```
DemiMarie-parity:
```suggestion
		// discards any transaction which is bigger than the length or weight limit **alone**,which
```

core/sr-primitives/src/lib.rs
```diff
@@ -469,8 +480,8 @@ impl ops::Add for Fixed64 {
 	}
 }
 
-/// Note that this is a standard, _potentially-panicking_, implementation. Use `Saturating` trait for
-/// safe subtraction.
+/// Note that this is a standard, _potentially-panicking_, implementation. Use `Saturating` trait
+/// for safe subtraction.
```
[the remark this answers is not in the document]
Kianenigma: The priority was to be consistent with how the std operations work. This implementation might panic. There are saturating/checked ones that don't.

node/runtime/src/constants.rs
Kianenigma: Both are readable. Will consider.

node/runtime/src/impls.rs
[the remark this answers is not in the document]
Kianenigma: I don't really see a reason for it tbh, Logging is great when you want sorting, structure, levels, filtering, or if you have a very high volume etc. This is neither of them. In some tests, something is printed _literally just once_. I have no problem in changing it but I think  it is a bit of an overdue.

srml/system/src/lib.rs
Kianenigma: I don't get the suggestion here; The problem is that:
- `decl_module!{}` does not accept such compiler flags.
- anything that even wraps the internals of the function is useless since the function itself exists and can be called and has a super high weight.
  - The function is not doing anything anyhow.

node/runtime/src/constants.rs
Kianenigma: regarding receiving the nit of missing newline from you a couple of times. Now I will do `"files.insertFinalNewline": true` :p

Kianenigma:
@DemiMarie-parity I replied in one place about the logging matter. Rest either fixed or argued why not. Since your comments were mostly (except for if it is possible to use `cfg[test]`) not related to the _logic but rather style_, would be nice if you can re-approve since I am planning I am merging this first thing tmr morning and it is a blocker for v2-k. thanks.

(I don't know why there was a push here but I can recover from the minor build issues arisen from it :p)

node-template/runtime/src/lib.rs
gavofyork: > We already have a redundant factor of 10 

Not enough.

> A bit more readable

Not a good enough reason, since this is actually important logic.

node/runtime/src/impls.rs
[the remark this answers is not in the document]
gavofyork: agree. logging in tests is a royal pain since you have to link the log module and research that magical incantation to put at the beginning of the test in order to make the log do anything at all. `println`s are fine.

node-template/runtime/src/lib.rs
gavofyork: > We already have a redundant factor of 10 

Not enough.

> A bit more readable

Not a good enough reason, since this is actually important logic.

gavofyork: also, Po2 not a good idea here. weights are meant to be vaguely human-comprehensible, not things that are stored in memory or otherwise benefit from Po2 sizes.

srml/session/src/lib.rs
Kianenigma: yeah good point. Reverted

node/executor/src/lib.rs
Kianenigma: false alarm. It is just `transfer_fee`. `let fee = if would_create { T::CreationFee::get() } else { T::TransferFee::get() };`.

node-template/runtime/src/lib.rs
Kianenigma: Reverted to a scale of `1_000_000_000`

PR #3161, opened 2019-07-22
Report to the PSM when a severe protocol error happens

tomaka:
Right now we have a lot of network activity, and most of it is caused by trying to connect to "bad" nodes (nodes from a different chain, or IPFS nodes for example).

A "severe protocol error" happens when we detect a "bad" node. This PR makes it so that severe protocol errors now decrease the node's reputation by 2^31, hence lowering the chances we try this node again in the short term.

### DIFF
--- a/core/network/src/custom_proto/behaviour.rs
+++ b/core/network/src/custom_proto/behaviour.rs
@@ -930,6 +930,7 @@ where
 			CustomProtoHandlerOut::ProtocolError { error, .. } => {
 				debug!(target: "sub-libp2p", "Handler({:?}) => Severe protocol error: {:?}",
 					source, error);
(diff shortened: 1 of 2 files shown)


core/network/src/custom_proto/behaviour.rs
```diff
@@ -930,6 +930,7 @@ where
 			CustomProtoHandlerOut::ProtocolError { error, .. } => {
 				debug!(target: "sub-libp2p", "Handler({:?}) => Severe protocol error: {:?}",
 					source, error);
+				self.peerset.report_peer(source.clone(), i32::min_value());
```
andresilva: Maybe add a comment here explaining the common situation for `ProtocolError`. What you have in the PR description is good enough.

PR #3164, opened 2019-07-22
sleep_until off-chain API

gautamdhameja:
Implements the `sleep_until` API for off-chain workers.

### DIFF
--- a/core/offchain/src/api.rs
+++ b/core/offchain/src/api.rs
@@ -18,7 +18,8 @@ use std::{
 	str::FromStr, 
 	sync::Arc, 
 	convert::{TryFrom, TryInto}, 
-	time::SystemTime
+	time::SystemTime,
+	thread::sleep,
 };
 use client::backend::OffchainStorage;
 use crate::AuthorityKeyProvider;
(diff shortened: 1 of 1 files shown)


core/offchain/src/api.rs
```diff
@@ -329,8 +331,13 @@ where
 		}
 	}
 
-	fn sleep_until(&mut self, _deadline: Timestamp) {
-		unavailable_yet::<()>("sleep_until")
+	fn sleep_until(&mut self, deadline: Timestamp) {
+		// Get current timestamp.
+		let now = self.timestamp();
+		// Calculate the diff with the deadline.
+		let diff = deadline.sub(Duration::from_millis(now.unix_millis()));
```
tomusdrw: There is actually `Timestamp::diff` method that you should use instead converting to millis:
```
let diff = deadline.diff(now);
```
It returns a `Duration` that just needs to be converted to `std::time::Duration`.

core/offchain/src/api.rs
```diff
@@ -608,6 +615,23 @@ mod tests {
...
+		let now = api.timestamp();
+		let delta = Duration::from_millis(100);
+		let t1 = now.add(delta);
+
+		// Act.
+		api.sleep_until(t1);
+		let t2 = api.timestamp();
+		
+		// Assert.
+		assert_eq!(t2.unix_millis() - now.unix_millis(), 100);
```
tomusdrw: I'd rely on being so specific here. It's fine if the different is at least 100 millis, but it might be longer.

gautamdhameja: Yeah, after reading the docs for `thread::sleep` it seems the diff would be at least the sleep duration but could be more than that (because of sys calls).

Asserting it now accordingly:
```
new_now.unix_millis() - 100 >= now.unix_millis()
```

PR #3165, opened 2019-07-22
Make Verifier::verify mutable

tomaka:
Sits on top of #3146 ; will probably be easier to review once #3146 is merged.
Conflicts with #3159, which I'd prefer be merged first.

There's no reason for the `Verifier` to be an `Arc` and to take `&self` instead of `&mut self`.
Let's make it mutable.

If there's a need for a shared verifier, one can always implement the `Verify` trait on `Arc<T>` or `&T` instead of `T`.


tomaka:
Oh well, I forgot to fix the networking tests, but this would conflict even more. I'll do it after other PRs are merged.

PR #3169, opened 2019-07-22
Run each offchain worker in a separate thread.

tomusdrw:


tomusdrw:
@tomaka Offchain workers are regular runtime call, imho it's currently not feasible to rewrite the entire runtime (and runtime APIs) using futures.
I don't expect browser nodes will ever be validators nor will they run any heavy offchain workers. I've changed the code to avoid spawning a new thread and run the task inline. Does it sound good to you?

BTW Do you know of a way to translate the code to use web workers instead? I believe that would allow us to have the level of parallelism we need.

core/offchain/src/lib.rs
```diff
@@ -157,16 +157,54 @@ impl<Client, Storage, KeyProvider, Block> OffchainWorkers<
...
+				let run = runtime.offchain_worker_with_context(
+					&at,
+					ExecutionContext::OffchainWorker(api),
+					number
+				);
+				if let Err(e) =	run {
+					log::error!("Error running offchain workers at {:?}: {:?}", at, e);
+				}
+			});
 			futures::future::Either::A(runner.process())
```
bkchr: What is process() returning?

tomaka:
> Does it sound good to you?

That sounds worse, actually! I'd be ok with `thread::sleep` and assuming that browsers won't have offchain workers, rather than having platform-specific code.

tomusdrw:
@tomaka Fair enough, removed platform specific code then.

core/offchain/src/lib.rs
tomusdrw: A future in which we process APIs that has to be invoked asynchronously - it's `submit_transaction` currently, since we can't recursively call into the runtime.

PR #3171, opened 2019-07-22
Use BABE instead of AuRa in node

andresilva:
Depends on #3028.

This PR updates the node to use BABE instead of AuRa for block authoring. Additionally, I made some changes to the `srml-babe` module, notably it now tracks the current slot and the epoch start slot. This is used to implement `ShouldEndSession` and also for the node to track the expected last epoch slot accurately.


andresilva:
@jacogr Renamed the weight type to `BabeWeight` to avoid the issue you mentioned.

andresilva:
I believe I have ironed out all the kinks, seems to be working fine. @rphmeier mentioned that some clarification is needed regarding the threshold calculation from the research team.

srml/babe/src/lib.rs
```diff
@@ -131,16 +140,13 @@ decl_storage! {
 		/// (like everything else on-chain) it is public. For example, it can be
 		/// used where a number is needed that cannot have been chosen by an
 		/// adversary, for purposes such as public-coin zero-knowledge proofs.
-		EpochRandomness get(epoch_randomness): [u8; VRF_OUTPUT_LENGTH];
-
-		/// The randomness under construction
-		UnderConstruction: [u8; VRF_OUTPUT_LENGTH];
+		pub Randomness get(randomness): [u8; RANDOMNESS_LENGTH];
```
jacogr: This (and the 2 below), appear in the metadata as `[u8; RANDOMNESS_LENGTH]` as opposed to `[u8; 32]`. The macro seems to extract the information as-is instead of interpreting it. (The last bit is even more black magic). Effectively this yields less than useful info for the middleware parsers.

Just indicating the issue, not sure what the best long-term solution here would be -

- make the code above more magic-y (not using the constants)
- making the macros even more magic-y ()

cc @bkchr

core/consensus/babe/src/tests.rs
```diff
@@ -232,7 +237,10 @@ fn run_one_test() {
 	// wait for all finalized on each.
 	let wait_for = futures::future::join_all(import_notifications);
 
-	let drive_to_completion = futures::future::poll_fn(|| { net.lock().poll(); Ok(Async::NotReady) });
+	let drive_to_completion = futures::future::poll_fn(|| {
```
tomaka: As a heads up, this will conflict with #3146 that revamps that code

srml/babe/src/lib.rs
```diff
@@ -262,11 +285,17 @@ impl<T: Trait + staking::Trait + Duration> session::OneSessionHandler<T::Account
 
 		// What was the next epoch is now the current epoch
 		let randomness = Self::randomness_change_epoch(epoch_index);
+
+		EpochStartSlot::mutate(|previous| *previous += <T::EpochDuration>::get());
```
DemiMarie-parity: This should saturate.

node/runtime/src/lib.rs
```diff
@@ -524,12 +525,23 @@ impl_runtime_apis! {
...
 
-	impl consensus_aura::AuraApi<Block, AuraId> for Runtime {
-		fn slot_duration() -> u64 {
-			Aura::slot_duration()
+	impl babe_primitives::BabeApi<Block> for Runtime {
+		fn startup_data() -> babe_primitives::BabeConfiguration {
+			babe_primitives::BabeConfiguration {
+				median_required_blocks: 1000,
+				slot_duration: Babe::slot_duration(),
+				threshold: core::u64::MAX,
```
rphmeier: probably what we should really be parameterizing is the `c` value from `phi_c`, since the threshold is per-validator.

srml/babe/src/lib.rs
rphmeier: Also, should account for skipping of at least one whole epoch as discussed offline. So add `T::EpochDuration::get()` in a loop until `now - previous < T::EpochDuration::get()`

srml/babe/src/lib.rs
andresilva: For now I will just hardcode the values but keep the constants and add a comment to use them once this works. @bkchr mentioned this needs type reflection.

jacogr: That is great - that is in active dev, so not tomorrow, but on the way. It also means we don't spend effort on something that will be replaced anyway.

core/consensus/babe/src/lib.rs
```diff
@@ -686,7 +715,7 @@ impl<B: BlockT, C> Verifier<B> for BabeVerifier<C> where
 				// `Consensus` is the Babe-specific authorities change log.
 				// It's an encoded `Epoch`, the same format as is stored in the
 				// cache, so no need to decode/re-encode.
-				let maybe_keys = find_epoch_digest::<B>(&pre_header)
+				let maybe_keys = find_epoch_digest::<B>(&pre_header)?
 					.map(|epoch| vec![(well_known_cache_keys::AUTHORITIES, epoch.encode())]);
```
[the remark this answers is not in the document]
andresilva: This one isn't a style fix, `find_epoch_digest` now returns a `Result` (in case we see multiple epoch change digests).

andresilva:
I think all the issues we have discussed offline have been fixed. For the threshold calculation, the exponentation is done in `f64` which is then converted into a `num::BigRational`. Then we use `num::BigUint` to multiply by the previous ratio (multiply by `1 << 128`, where 128 is the number of bits from the VRF we'll use for comparing). For the constant `c` I set it to `0.3` but I have no basis for this (W3F research team should get back to us on this).

The light client is currently not working, I will file some issues for follow-up PRs to fix it. I updated the sync integration test to generate valid BABE predigests but disabled the light clients for now.

PR #3175, opened 2019-07-23
Make Environment, Proposer, OnSlot and SyncOracle mut

tomaka:
Based on top of #3146

Makes the `Environment`, `Proposer`, `OnSlot` and `SyncOracle` traits accept `&mut self` instead of `&self`.
Additionally removes a few unnecessary `Arc`s.


core/consensus/babe/src/tests.rs
```diff
@@ -236,7 +236,7 @@ fn run_one_test() {
 		net.lock().poll();
 		Ok::<_, ()>(futures01::Async::NotReady::<()>)
 	}));
-
+
```
DemiMarie-parity: Whitespace

tomaka:
```suggestion

```

tomaka:
I guess this got merged as part of #3206 :man_shrugging:

PR #3177, opened 2019-07-23
Handle local storage race conditions better

### LINKED ISSUE #3143: Enable Offchain API to handle local storage race conditions better
Follow-up from https://github.com/paritytech/substrate/pull/3079/.

This is how the race condition can appear:

Worker A finds no value in storage (`local_storage_get(…)` returns `None`). It can not issue `local_storage_compare_and_set(…)` with `old_value == None`, since that function requires a `&Vec<u8>` for `old_value`. Thus it must use a `local_storage_set(…)`.

Concurrently another worker does the same thing, resulting in a race condition.

- [ ] Change `local_storage_compare_and_set(…)` to take `Option<&Vec<u8>>`.
- [ ] Adapt the way `srml/im-online` sets local storage to use this new `local_storage_compare_and_set(…)` API then.

cmichi:
Closes #3143.


srml/im-online/src/lib.rs
```diff
@@ -233,27 +233,57 @@ decl_module! {
...
+			fn compare_and_set_worker_status<T: Trait>(
+				gossipping_at: T::BlockNumber,
+				done: bool,
+				curr_worker_status: Option<Vec<u8>>,
+			) -> bool {
+				let enc = WorkerStatus {
+					done,
+					gossipping_at,
+				};
+				match curr_worker_status {
```
tomusdrw: Is it necessary to use `match`? Can't we do `curr_worker_status.map(Vec::as_slice)`?

cmichi: Great idea, I updated the PR.

PR #3181, opened 2019-07-23
Fix mDNS not working

tomaka:
I accidentally broke it in #3116 
It could on principle be a good idea to write a test for mDNS, but the test unfortunately can't be isolated from the network, so I don't think it's a good idea in practice.

cc @andresilva


andresilva:
This would be caught by the black box tests I mentioned earlier (e.g. spawning 3 nodes they should be able to find each other).

bkchr:
Yeah, can we not just run two nodes, but don't provide the addresses to each other?

tomaka:
> Yeah, can we not just run two nodes, but don't provide the addresses to each other?

I feel like this would "pollute" the local network.

PR #3183, opened 2019-07-23
WIP: Subkey docker image

TriplEight:
Makes a docker release for the subkey.
Splits `./artifacts/` to internal `/substrate/` and `/subkey/`
S3 releases will be in the same place where substrate used to be, but now there also will be subkey binary.


TriplEight:
@gabreal please check my logic before I removed the testing harness.
Especially:
- substrate files now go to `./artifacts/substrate/`
- subkey files now go to  `./artifacts/subkey/`
- they are now published the same way (in subdirs)
- s3 sync works a bit different 
Example pipeline with details: https://gitlab.parity.io/parity/substrate/pipelines/44721

PR #3185, opened 2019-07-24
Child storage tests and genesis fix.

cheme:
This PR derives from branch https://github.com/paritytech/substrate/compare/master...cheme:child_module_storage where module storage got switch to child storage (branch is stale behind the need of change child trie).

It contains changes needed for child storage tests and for genesis initialization.

It contains additional changes needed to use genesis child for runtimes.
It continues #3009 effort by removing some variant of function (eg function using only a top structure) that can lead to error.
It also fix merging logic of child storage layer (no direct 'extend' on child map).
It also adapt testing and basic storage_root function to include child in calculation.

This breaks compatibility as it changes chainspec storage description (see flaming-fir json sample).


core/state-machine/src/lib.rs
```diff
@@ -968,6 +989,26 @@ where
 	}
 }
 
+/// Reads child storage value from overlay or from the backend.
+fn try_child_read_overlay_value<H, B>(
```
thiolliere: this is never used ?

srml/treasury/src/lib.rs
```diff
@@ -431,11 +433,11 @@ mod tests {
...
 
 	fn new_test_ext() -> runtime_io::TestExternalities<Blake2Hasher> {
-		let mut t = system::GenesisConfig::default().build_storage::<Test>().unwrap().0;
-		t.extend(balances::GenesisConfig::<Test>{
+		let mut t = system::GenesisConfig::default().build_storage::<Test>().unwrap();
+		extend_storage_overlays(&mut t, balances::GenesisConfig::<Test>{
 			balances: vec![(0, 100), (1, 99), (2, 1)],
 			vesting: vec![],
-		}.build_storage().unwrap().0);
+		}.build_storage().unwrap());
```
thiolliere: ok this can be done this way but I think it is preferable if we actually always do as the genesis config is build which means module storage are assimilated in order.

Thus I would prefer that on test even if it is not needed ppl use this method so first is build and other are assimilated (or all are assimilated from empty ones)

srml/treasury/src/lib.rs
```diff
@@ -357,6 +357,8 @@ impl<T: Trait> OnDilution<BalanceOf<T>> for Module<T> {
 mod tests {
 	use super::*;
 
+
```
thiolliere:
```suggestion
```

srml/session/src/historical.rs
```diff
@@ -326,13 +327,13 @@ mod tests {
...
-		let mut t = system::GenesisConfig::default().build_storage::<Test>().unwrap().0;
-		let (storage, _child_storage) = crate::GenesisConfig::<Test> {
+		let mut t = system::GenesisConfig::default().build_storage::<Test>().unwrap();
+		let storage = crate::GenesisConfig::<Test> {
 			keys: NEXT_VALIDATORS.with(|l|
 				l.borrow().iter().cloned().map(|i| (i, UintAuthorityId(i))).collect()
 			),
 		}.build_storage().unwrap();
-		t.extend(storage);
+		extend_storage_overlays(&mut t, storage);
```
thiolliere: better to use assimilate

srml/indices/src/mock.rs
```diff
@@ -98,10 +99,10 @@ pub fn new_test_ext() -> runtime_io::TestExternalities<Blake2Hasher> {
 		for i in 1..5 { h.insert(i); }
 	}
 
-	let mut t = system::GenesisConfig::default().build_storage::<Runtime>().unwrap().0;
-	t.extend(GenesisConfig::<Runtime> {
+	let mut t = system::GenesisConfig::default().build_storage::<Runtime>().unwrap();
+	extend_storage_overlays(&mut t, GenesisConfig::<Runtime> {
```
thiolliere: same

srml/grandpa/src/mock.rs
```diff
@@ -77,10 +78,10 @@ pub fn to_authorities(vec: Vec<(u64, u64)>) -> Vec<(AuthorityId, u64)> {
 }
 
 pub fn new_test_ext(authorities: Vec<(u64, u64)>) -> runtime_io::TestExternalities<Blake2Hasher> {
-	let mut t = system::GenesisConfig::default().build_storage::<Test>().unwrap().0;
-	t.extend(GenesisConfig {
+	let mut t = system::GenesisConfig::default().build_storage::<Test>().unwrap();
+	extend_storage_overlays(&mut t, GenesisConfig {
```
thiolliere: same

srml/generic-asset/src/mock.rs
```diff
@@ -114,9 +115,9 @@ impl ExtBuilder {
 
 	// builds genesis config
 	pub fn build(self) -> runtime_io::TestExternalities<Blake2Hasher> {
-		let mut t = system::GenesisConfig::default().build_storage::<Test>().unwrap().0;
+		let mut t = system::GenesisConfig::default().build_storage::<Test>().unwrap();
 
-		t.extend(
+		extend_storage_overlays(&mut t,
```
thiolliere: same

core/trie/src/lib.rs
```diff
@@ -145,7 +145,7 @@ where
 /// `child_trie_root` and `child_delta_trie_root` can panic if invalid value is provided to them.
 pub fn is_child_trie_key_valid<H: Hasher>(storage_key: &[u8]) -> bool {
 	use substrate_primitives::storage::well_known_keys;
-	let has_right_prefix = storage_key.starts_with(b":child_storage:default:");
+	let has_right_prefix = storage_key.starts_with(b":child_storage:");
```
thiolliere: is that true ? default is not mandatory anymore ?

core/test-runtime/src/system.rs
```diff
@@ -113,6 +113,12 @@ fn execute_block_with_state_root_handler(
 
 	let o_new_authorities = <NewAuthorities>::take();
 
+	// symetric removal of finalize for roots to match not putting parent hash as it
+	// seems to touch the state (required key value).
+	<Number>::kill();
+	<StorageDigest>::kill();
+	<NewAuthorities>::kill();
```
thiolliere: why is this now needed ?

cheme: I think I make a mistake in my fix for child (I needed those when I put them in a child trie), there is definitely something going on here that I do not like, but those lines does not make much sense (it seems there is some where root is calculated only from call to this method so initialize may not be call here and thing should be fine as long as those value are not use), I am removing it.

core/trie/src/lib.rs
cheme: Oups, I am not sure on wether default is mandatory or not, from my perspective it seems to me that 'default' was only needed for contract.
(currently this method is use on all child trie through sr-io ChildStorageKey so I switch it to the well known key only).
May not be the proper change (if it is not I would suggest changing well_know_keys child storage prefix to use 'default').

srml/treasury/src/lib.rs
cheme: I am not 100% sure about order for my assimilates (I think I reversed initial order here but it seems logic to me that system is loaded before balance)

core/state-machine/src/lib.rs
cheme: I am removing it (was use in may branch with child trie usage but is only an annoying warning here).

core/state-machine/src/basic.rs
```diff
@@ -35,13 +35,9 @@ pub struct BasicExternalities {
 }
 
 impl BasicExternalities {
-	/// Create a new instance of `BasicExternalities`
-	pub fn new(top: HashMap<Vec<u8>, Vec<u8>>) -> Self {
-		Self::new_with_children(top, Default::default())
-	}
 
-	/// Create a new instance of `BasicExternalities` with children
-	pub fn new_with_children(
```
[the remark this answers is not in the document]
cheme: Yes I remember, but it was the easier way for me to track down incomplete child trie usage (by moving all module to child trie it was needed everywhere). Do you think we can keep it that way (I expect we can loose more time trying to find where the child trie are not use than transitioning)?

core/sr-primitives/src/lib.rs
```diff
@@ -159,11 +144,33 @@ impl BuildStorage for (StorageOverlay, ChildrenStorageOverlay) {
 		child_storage: &mut ChildrenStorageOverlay
```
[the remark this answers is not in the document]
cheme: I strongly agree, I will take a look.

core/sr-io/without_std.rs
```diff
@@ -212,6 +212,12 @@ pub mod ext {
 		fn ext_exists_storage(key_data: *const u8, key_len: u32) -> u32;
 		/// Remove storage entries which key starts with given prefix.
 		fn ext_clear_prefix(prefix_data: *const u8, prefix_len: u32);
+		fn ext_clear_child_prefix(
```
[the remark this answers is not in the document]
cheme: Oups, sorry.

core/sr-primitives/src/lib.rs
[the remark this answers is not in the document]
thiolliere: it may be put in another PR though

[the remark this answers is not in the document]
cheme: Last commit goes toward this direction, it changes 'decl_storage' macro a bit (the genesis build fn takes two parameters now).

srml/treasury/src/lib.rs
```diff
@@ -431,11 +432,11 @@ mod tests {
...
 
 	fn new_test_ext() -> runtime_io::TestExternalities<Blake2Hasher> {
-		let mut t = system::GenesisConfig::default().build_storage::<Test>().unwrap().0;
-		t.extend(balances::GenesisConfig::<Test>{
+		let mut t = system::GenesisConfig::default().build_storage::<Test>().unwrap();
+		balances::GenesisConfig::<Test>{
 			balances: vec![(0, 100), (1, 99), (2, 1)],
 			vesting: vec![],
-		}.build_storage().unwrap().0);
+		}.build_storage().unwrap().assimilate_storage(&mut t).unwrap();
```
thiolliere:
```suggestion
		}.assimilate_storage(&mut t).unwrap();
```
the thing is that during its build balance can use what system has put into storage

core/test-runtime/client/Cargo.toml
```diff
@@ -9,6 +9,7 @@ generic-test-client = { package = "substrate-test-client", path = "../../test-cl
 primitives = { package = "substrate-primitives", path = "../../primitives" }
 runtime = { package = "substrate-test-runtime", path = "../../test-runtime", default-features = false }
 runtime_primitives = { package = "sr-primitives", path = "../../sr-primitives" }
+parity-codec = { version = "4.1.1", default-features = false, features = ["derive"] }
```
thiolliere: I'm surprise the `default-features = false` here doesn't require adding "parity-codec/std" in features::std (after "runtime/std" below)

thiolliere:
So this PR changes decl_storage API, now add_genesis build closure has only 2 params and first param is a tuple of top_storage and child_storage.

core/test-runtime/client/Cargo.toml
cheme: that is an interesting omission from my part.

cheme:
yes, it also breaks existing things by changing the chainspec json format.

core/state-machine/src/backend.rs
```diff
@@ -247,6 +247,25 @@ impl<H: Hasher> From<HashMap<Option<Vec<u8>>, HashMap<Vec<u8>, Vec<u8>>>> for In
 	}
 }
 
+impl<H: Hasher> From<(
+		HashMap<Vec<u8>, Vec<u8>>,
+		HashMap<Vec<u8>, HashMap<Vec<u8>, Vec<u8>>>,
+	)> for InMemory<H> {
```
cheme:
```suggestion
)> for InMemory<H> {
```

cheme:
I need to check again, but I think that changing the api of decl_storage was needed to avoid some awkward things (pointer to a pair object cannot be really build form two pointer).

gavofyork:
needs resolving, and will probably need resolving again once #3296 is in.

cheme:
During rebase there was few additional change to the new module 'membership' tests, mainly the same change as for other modules (assimilate instead of extend and include children).

PR #3186, opened 2019-07-24
Network: stop being generic over the type of message

tomaka:
Considering that the `network-libp2p` and `network` crates have now been merged, let's make the protocol not generic over the type of the network message anymore. This PR removes the generic template parameter of everything in `custom_proto` to be hardcoded to `Message<B>` (which is the enum of all the possible Substrate-specific network messages).

To me, the code in `custom_proto` is pretty much "legacy". It has several issues that are not caused by the code but by the protocol level itself, and that we cannot fix without breaking the entire network.

We can expand the `Message` enum, but we shouldn't expand the protocol itself using `custom_proto`. In other words, we shouldn't create a new `Foo` enum and start using `CustomProto<Foo>` in parallel with `CustomProto<Message<B>>`. Hence why hardcoding `Message` isn't a bad idea.


tomaka:
Why was this marked onice?

bkchr:
I think this was a mistake.

PR #3188, opened 2019-07-24
Expose instancing in metadata

bkchr:
- Introduces metadata v7
- If a module is using instancing, the storage exposes the instance in
metadata
- Metadata module name is now the upper case one given to
`construct_runtime!`

### DIFF
--- a/srml/metadata/src/lib.rs
+++ b/srml/metadata/src/lib.rs
@@ -292,9 +292,6 @@ pub enum StorageEntryModifier {
 pub struct StorageMetadata {
 	/// The common prefix used by all storage entries.
 	pub prefix: DecodeDifferent<&'static str, StringBuf>,
-	/// The instance identifier. This identifier is appended to each storage entry prefix to make it
-	/// unique per instance.
-	pub instance: Option<DecodeDifferent<&'static str, StringBuf>>,
 	pub entries: DecodeDifferent<&'static [StorageEntryMetadata], Vec<StorageEntryMetadata>>,
 }
 
--- a/srml/metadata/src/lib.rs
+++ b/srml/metadata/src/lib.rs
@@ -286,6 +286,18 @@ pub enum StorageEntryModifier {
 	Default,
 }
 
+/// All metadata of the storage.
+#[derive(Clone, PartialEq, Eq, Encode)]
+#[cfg_attr(feature = "std", derive(Decode, Debug, Serialize))]
+pub struct StorageMetadata {
+	/// The common prefix used by all storage entries.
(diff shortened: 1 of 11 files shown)


thiolliere:
this PR changes in the metadata:
* names of module are now the first name given in construct_runtime and not the crate name (as imported) anymore
* storage metadata structure is now changed.

thiolliere:
Hmm I was a bit mixed by having the instance appearing in the metadata as for me it shouldn't be any difference with a regular module I would just have added the instance inside the prefix directly. I don't why a user of the metadata would want to differentiate instantiated module and others

srml/metadata/src/lib.rs
```diff
@@ -286,6 +286,18 @@ pub enum StorageEntryModifier {
...
 
+/// All metadata of the storage.
+#[derive(Clone, PartialEq, Eq, Encode)]
+#[cfg_attr(feature = "std", derive(Decode, Debug, Serialize))]
+pub struct StorageMetadata {
+	/// The common prefix used by all storage entries.
+	pub prefix: DecodeDifferent<&'static str, StringBuf>,
+	/// The instance identifier. This identifier is appended to each storage entry prefix to make it
+	/// unique per instance.
+	pub instance: Option<DecodeDifferent<&'static str, StringBuf>>,
```
thiolliere: should we just add this into the prefix ? I think both are fine actually

bkchr: That will not work. `prefix` is for example `Staking`.
But the final prefix for a storage entry will be calculated like:
`PREFIX_FOR_Staking_{ValueName}{Instance}`

thiolliere: yes what I mean is that prefix should be the final one.

srml/support/src/metadata.rs
```diff
@@ -163,7 +194,7 @@ macro_rules! __runtime_modules_to_metadata_calls_storagename {
 	) => {
```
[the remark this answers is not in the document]
bkchr: Good point

srml/metadata/src/lib.rs
thiolliere: an instantiated module shouldn't differ in any way to a regular module from metadata point of view

bkchr: I don't wanted to break storage, but yeah. You are right. Better do it now than later.

bkchr: (Removed)

bkchr:
Ahhh.... :( 

Fixed.

PR #3192, opened 2019-07-24
Merge node, node-cli, node-executor, node-rpc-client into one

### LINKED ISSUE #2934: Merge the multiple `node` subdirectories into one crate
For historical reasons, the Substrate `node` is split into several smaller crates.
We should merge everything together, except for the `runtime`.

tomaka:
Close #2934 

Merges as much as possible in the Substrate node. The objective is to simplify and clean up a bit the architecture.

Moves the code of `node-cli` and `node-executor` in `node/src`.
I moved the main executable in a `src/bin` subdirectory instead of leaving it in `src`.
I initially wanted to `node-rpc-client` in `src/bin` as well, but if you do that then `cargo run` requires being passed which binary to run.


Cargo.toml
```diff
@@ -10,13 +14,70 @@ build = "build.rs"
 edition = "2018"
 
 [dependencies]
-cli = { package = "node-cli", path = "node/cli" }
+log = "0.4"
```
bkchr: Hmm, don't we maybe want to clean this up and just keep the members in the root `Cargo.toml`.

The following should help for `cargo run`:
https://github.com/rust-lang/cargo/issues/7032

(I'm not sure if this already included in stable or not.

tomaka: It's not in stable.

bkchr: Can we just wait and postpone until this is in stable?

tomaka: You mean around 2022

bkchr: This is already stabilized, so I would assume that it is released with the next rust release? https://github.com/rust-lang/cargo/pull/7056

node/src/executor.rs
```diff
@@ -47,9 +46,9 @@ mod tests {
 		NativeOrEncoded
 	};
 	use node_primitives::{Hash, BlockNumber, AccountId, Balance, Index};
-	use runtime_primitives::traits::{Header as HeaderT, Hash as HashT};
-	use runtime_primitives::{generic::Era, ApplyOutcome, ApplyError, ApplyResult, Perbill};
-	use runtime_primitives::weights::{WeightMultiplier, SimpleDispatchInfo, WeighData};
+	use sr_primitives::traits::{Header as HeaderT, Hash as HashT};
```
Kianenigma: I am quite happy to see some else also changing these --not wrong but confusing-- imports; it is quite annoying that we have `core/sr-primitives` imported into different crates with different names; makes it even more confusing that we have an actual `runtime-primitives` in `node`.

I have some thoughts on _codifying_ imports a bit and adding it to our style guide maybe. Will make an issue for it.

PR #3194, opened 2019-07-24
Informant cleanup

### LINKED ISSUE #2652: Don't use `saturated_into` in informant
We should require the `BlockNumber` to implement `Display + Serialize` for the informant as a local where bound.

https://github.com/paritytech/substrate/pull/2602_

tomaka:
Fix #2652 

Splits the informant in two files, one that does the `Future` stuff and one that only does the displaying.
Also fixes #2652 by trying `try_into::<u128>` and falling back to pure-integer mathematics if it fails.

For what it's worth, I locally tested the pure-integer code path, and it doesn't seem to cause any abnormal CPU usage, even in debug.

### DIFF
--- a/core/cli/src/informant/display.rs
+++ b/core/cli/src/informant/display.rs
@@ -15,12 +15,12 @@
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
 use ansi_term::Colour;
-use std::{fmt, marker::PhantomData, time};
-use service::NetworkStatus;
-use network::SyncState;
 use client::ClientInfo;
 use log::info;
+use network::SyncState;
 use runtime_primitives::traits::{Block as BlockT, SaturatedConversion};
+use service::NetworkStatus;
+use std::{fmt, marker::PhantomData, time};
 
 /// State of the informant display system.
 pub struct InformantDisplay<B: BlockT> {
--- a/core/cli/src/informant/display.rs
+++ b/core/cli/src/informant/display.rs
@@ -18,20 +18,17 @@ use ansi_term::Colour;
 use client::ClientInfo;
 use log::info;
 use network::SyncState;
-use runtime_primitives::traits::{Block as BlockT, SaturatedConversion};
+use runtime_primitives::traits::{Block as BlockT, SaturatedConversion, CheckedDiv, NumberFor, One, Zero, Saturating};
 use service::NetworkStatus;
-use std::{fmt, marker::PhantomData, time};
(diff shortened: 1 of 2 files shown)


core/cli/src/informant/display.rs
```diff
@@ -0,0 +1,149 @@
...
+		self.last_update = time::Instant::now();
+		self.last_number = Some(best_number);
+
+		let (status, target) = match (net_status.sync_state, net_status.best_seen_block) {
+			(SyncState::Idle, _) => ("Idle".into(), "".into()),
+			(SyncState::Downloading, None) => (format!("Syncing{}", speed), "".into()),
+			(SyncState::Downloading, Some(n)) => (format!("Syncing{}", speed), format!(", target=#{}", n)),
+		};
+
+		let finalized_number: u64 = info.chain.finalized_number.saturated_into::<u64>();
```
tomaka: Oh, I forgot to change that one!

core/cli/src/informant/display.rs
```diff
@@ -0,0 +1,149 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```
bkchr: 2019

core/cli/src/informant/display.rs
```diff
@@ -0,0 +1,149 @@
...
+		let speed = diff.saturating_mul(10_000).checked_div(u128::from(elapsed_ms))
+			.map_or(0.0, |s| s as f64) / 10.0;
+		format!(" {:4.1} bps", speed)
+
+	} else {
+		// If the number of blocks can't be converted to a regular integer, then we need a more
+		// algebraic approach and we stay within the realm of integers.
+		let ten = (0..10)
+			.fold(<NumberFor<B> as Zero>::zero(), |a, _| a.saturating_add(One::one()));
+		let one_thousand = ten * ten * ten;
```
bkchr: BlockNumber needs to implement From<u32>, just call 1000.into().

core/cli/src/informant/display.rs
```diff
@@ -0,0 +1,149 @@
...
+		let elapsed = (0..elapsed_ms)
+			.fold(<NumberFor<B> as Zero>::zero(), |a, _| a.saturating_add(One::one()));
```
bkchr: I would expect that elapsed always fits into u32, so you should be able to also from.

core/cli/src/informant/display.rs
tomaka: Is that true even if it's just extracted code?

PR #3195, opened 2019-07-24
wasm-builder: allow default cargo nightly

ascjones:
I [added](https://github.com/paritytech/ink/pull/124) a dev-dependency on substrate from `ink!`, and the build failed with https://github.com/paritytech/substrate/blob/33609c707a3bf7b1a520d639b7b1199dc77c0cff/core/utils/wasm-builder/src/prerequisites.rs#L40

It's because ink has the `rust-toolchain` file https://github.com/paritytech/ink/blob/master/rust-toolchain, which means that the vanilla `cargo` defaults to that version. It's a fixed version of nightly, and the wasm toolchain is installed for that version and not the latest nightly.

This PR fixes `get_nightly_cargo` to return the default `cargo` command if it is a nightly version.

### DIFF
--- a/core/utils/wasm-builder/src/lib.rs
+++ b/core/utils/wasm-builder/src/lib.rs
@@ -152,17 +152,11 @@ fn create_out_file(file_name: &str, content: String) {
 
 /// Get a cargo command that compiles with nightly
 fn get_nightly_cargo() -> Command {
-	let version = Command::new("cargo")
-		.arg("--version")
-		.output()
-		.map_err(|_| ())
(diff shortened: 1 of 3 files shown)


shawntabrizi:
@ascjones I talked to @Robbepop and we agreed to remove the toolchain file from ink!.

Rather we will have ink! use the latest nightly just like Substrate does.

In this case, I don't think this change is needed right?

ascjones:
It wouldn't be as urgent, but would still be needed for any project which has a toolchain file and  depends on substrate.

PR #3204, opened 2019-07-25
Make sure that the era length is always the same

bkchr:
(Besides something forced a new era)


srml/staking/src/lib.rs
```diff
@@ -1419,16 +1420,16 @@ impl<T: Trait> Module<T> {
 }
 
 impl<T: Trait> session::OnSessionEnding<T::AccountId> for Module<T> {
-	fn on_session_ending(_ending: SessionIndex, start_session: SessionIndex) -> Option<Vec<T::AccountId>> {
-		Self::new_session(start_session - 1).map(|(new, _old)| new)
+	fn on_session_ending(ending: SessionIndex, _: SessionIndex) -> Option<Vec<T::AccountId>> {
```
andresilva: This interface makes it seem like the next session index may not always be `current + 1`.

srml/staking/src/lib.rs
```diff
@@ -1419,16 +1420,16 @@ impl<T: Trait> Module<T> {
...
+		Self::new_session(ending + 1).map(|(new, _old)| new)
```
andresilva: Shouldn't this actually just be `start_session`?

Kianenigma: I am not even sure why the function has two arguments: it seems like the latter should always be former + 1?

andresilva: Yeah, if that's an invariant of the session module then we should remove the second parameter.

bkchr: Documentation says:
```
	/// The returned validator set, if any, will not be applied until `next_index`.
	/// `next_index` is guaranteed to be at least `ending_index + 1`, since session indices don't
```

It is at least `ending_index + 1

srml/staking/src/lib.rs
```diff
@@ -1419,16 +1420,16 @@ impl<T: Trait> Module<T> {
...
 	}
 }
 
 impl<T: Trait> OnSessionEnding<T::AccountId, Exposure<T::AccountId, BalanceOf<T>>> for Module<T> {
```
Kianenigma:
```suggestion
impl<T: Trait> historical::OnSessionEnding<T::AccountId, Exposure<T::AccountId, BalanceOf<T>>> for Module<T> {
```

Less confusing. + no need to be imported individually up top.

srml/staking/src/tests.rs
```diff
@@ -2100,3 +2102,25 @@ fn reward_from_authorship_event_handler_works() {
...
+			let session = Session::current_index();
+			ForceNewEra::put(true);
+			start_next_session();
+			assert_eq!(Staking::current_era(), 3);
+			assert_eq!(Staking::current_era_start_session_index(), session + 1);
+
+			start_era(4);
+			assert_eq!(Staking::current_era_start_session_index(), session + SessionsPerEra::get() + 1);
+		});
+}
```
Kianenigma:
```suggestion
}

```

srml/staking/src/lib.rs
andresilva: Ok so the invariant is that it increases but no guarantee by how much. So we should just use the second parameter here, without the `- 1` as it was originally (I think that was wrong).

bkchr: Why should we use the second?

srml/staking/src/lib.rs
```diff
@@ -1419,16 +1420,18 @@ impl<T: Trait> Module<T> {
...
-impl<T: Trait> OnSessionEnding<T::AccountId, Exposure<T::AccountId, BalanceOf<T>>> for Module<T> {
-	fn on_session_ending(_ending: SessionIndex, start_session: SessionIndex)
+impl<T: Trait> session::historical::OnSessionEnding<T::AccountId, Exposure<T::AccountId, BalanceOf<T>>>
+	for Module<T>
+{
+	fn on_session_ending(ending: SessionIndex, _: SessionIndex)
 		-> Option<(Vec<T::AccountId>, Vec<(T::AccountId, Exposure<T::AccountId, BalanceOf<T>>)>)>
 	{
-		Self::new_session(start_session - 1)
+		Self::new_session(ending + 1)
```
DemiMarie-parity: This should either be `start_session` or `ending`, but not `ending + 1` or `start_session - 1`.

bkchr: Did you read the documentation?

srml/staking/src/lib.rs
gavofyork:
```suggestion
	fn on_session_ending(_: SessionIndex, will_apply_at: SessionIndex) -> Option<Vec<T::AccountId>> {
```

srml/staking/src/lib.rs
```diff
@@ -1419,16 +1420,18 @@ impl<T: Trait> Module<T> {
 }
 
 impl<T: Trait> session::OnSessionEnding<T::AccountId> for Module<T> {
-	fn on_session_ending(_ending: SessionIndex, start_session: SessionIndex) -> Option<Vec<T::AccountId>> {
-		Self::new_session(start_session - 1).map(|(new, _old)| new)
+	fn on_session_ending(ending: SessionIndex, _: SessionIndex) -> Option<Vec<T::AccountId>> {
+		Self::new_session(ending + 1).map(|(new, _old)| new)
```
gavofyork:
```suggestion
		Self::new_session(will_apply_at).map(|(new, _old)| new)
```

srml/staking/src/lib.rs
```diff
@@ -1117,7 +1117,8 @@ impl<T: Trait> Module<T> {
 	fn new_session(session_index: SessionIndex)
```
gavofyork: the `session_index` here is the session index that the the return value (i.e. the potentially new validator set) will be applied at, and *NOT* the current (ending) session index. really, the name of the parameter and ideally the docs should be changed to reflect this.

srml/staking/src/lib.rs
gavofyork:
```suggestion
		Self::new_session(will_apply_at)
```

srml/staking/src/lib.rs
```diff
@@ -1419,16 +1420,18 @@ impl<T: Trait> Module<T> {
...
+		Self::new_session(ending + 1).map(|(new, _old)| new)
 	}
 }
 
-impl<T: Trait> OnSessionEnding<T::AccountId, Exposure<T::AccountId, BalanceOf<T>>> for Module<T> {
-	fn on_session_ending(_ending: SessionIndex, start_session: SessionIndex)
+impl<T: Trait> session::historical::OnSessionEnding<T::AccountId, Exposure<T::AccountId, BalanceOf<T>>>
+	for Module<T>
+{
+	fn on_session_ending(ending: SessionIndex, _: SessionIndex)
```
gavofyork:
```suggestion
	fn on_session_ending(_: SessionIndex, will_apply_at: SessionIndex)
```

gavofyork:
only concern here is that the `Exposure` returned is indeed correct for this future session, but i *think* it should be.

PR #3207, opened 2019-07-25
core/consensus,node/runtime: Declare and implement authorities endpoint

mxinden:
The goal of the commit is to be able to retrieve the current set of authorities
without needing to know the concrete consensus mechanism in place.

In order to achieve the above this commit introduces the
`core/consensus/common/primitives` crate, declaring the `ConsensusApi` runtime
API. In addition it implements the above mentioned trait definition in
`node/runtime` by returning the current authorities of the BABE consensus
mechanism.

>  You bumped the runtime version if there are breaking changes in the runtime.

Does adding an API require a runtime version bump? 

This is part of the effort of validators discovering and connecting other
validators on the network layer (See https://github.com/paritytech/substrate/issues/1693). I will be proposing small
pull requests instead of a single big one for better reviewability. Let me know
if you prefer a single one.

Next to directly connecting to consensus authorities, we should probably also
directly connect to finality authorities.

### DIFF
--- a/core/consensus/common/primitives/Cargo.toml
+++ b/core/consensus/common/primitives/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "substrate-consensus-common-primitives"
+version = "2.0.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+description = "Primitives for session"
+edition = "2018"
+
+[dependencies]
+parity-codec = { version = "4.1.1", default-features = false }
+client = { package = "substrate-client", path = "../../../client", default-features = false }
+runtime_primitives = { package = "sr-primitives", path = "../../../sr-primitives", default-features = false }
+rstd = { package = "sr-std", path = "../../../sr-std", default-features = false }
+
+[features]
+default = ["std"]
+std = [
+    "rstd/std",
+    "client/std",
+    "parity-codec/std",
+    "runtime_primitives/std"
+]
--- a/core/consensus/common/primitives/Cargo.toml
+++ b/core/consensus/common/primitives/Cargo.toml
@@ -2,13 +2,13 @@
 name = "substrate-consensus-common-primitives"
 version = "2.0.0"
 authors = ["Parity Technologies <admin@parity.io>"]
-description = "Primitives for session"
+description = "Common consensus primitives"
 edition = "2018"
 
 [dependencies]
 parity-codec = { version = "4.1.1", default-features = false }
 client = { package = "substrate-client", path = "../../../client", default-features = false }
-runtime_primitives = { package = "sr-primitives", path = "../../../sr-primitives", default-features = false }
+sr-primitives = { package = "sr-primitives", path = "../../../sr-primitives", default-features = false }
 rstd = { package = "sr-std", path = "../../../sr-std", default-features = false }
 
 [features]
@@ -17,5 +17,5 @@ std = [
     "rstd/std",
     "client/std",
     "parity-codec/std",
-    "runtime_primitives/std"
+    "sr-primitives/std"
 ]
--- a/core/consensus/common/primitives/src/lib.rs
+++ b/core/consensus/common/primitives/src/lib.rs
@@ -18,7 +18,7 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use parity_codec::{Codec};
+use parity_codec::Codec;
 use client::decl_runtime_apis;
 use rstd::vec::Vec;
 
--- a/core/consensus/common/primitives/src/lib.rs
+++ b/core/consensus/common/primitives/src/lib.rs
@@ -1,0 +1,31 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
+// This file is part of Substrate.
+
+// Substrate is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Common consensus primitives.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use parity_codec::{Codec};
+use client::decl_runtime_apis;
+use rstd::vec::Vec;
+
+decl_runtime_apis! {
+	/// Common consensus runtime api.
+	pub trait ConsensusApi<AuthorityId: Codec> {
+		/// Returns the set of authorities of the currently active consensus mechanism.
+		fn authorities() -> Vec<AuthorityId>;
+	}
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2382,6 +2382,7 @@ dependencies = [
  "srml-treasury 2.0.0",
  "substrate-client 2.0.0",
  "substrate-consensus-babe-primitives 2.0.0",
+ "substrate-consensus-common-primitives 2.0.0",
  "substrate-keyring 2.0.0",
  "substrate-offchain-primitives 2.0.0",
  "substrate-primitives 2.0.0",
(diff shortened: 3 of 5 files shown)


Kianenigma:
I'm a bit confused in terms of terminology here: 

Afaik Substrate is layered like: `Network (fixed) -> Consensus (fixed) -> Runtime (variable)`. 

So my understanding was that the runtime is actually unaware to some extent of what consensus is wrapping it. In other worse, the runtime just assumes that there exists some consensus in the client code with some unified Api. Based in this code though, it looks like _core/client is unaware of consensus_ (defining e.g. the returrn type of the Api as a generic which is encodable) and the _runtime being **aware**_  (replacing the generic with concrete values based on knowing that `babe` is being used). Can you clarify this a bit in terms of who knows what by default and which information gap are we filling here? 

Furthermore, how will this look while keeping in mind that some users might still want to use Aura? 

(I think I answered myself by recalling that babe/aura is not the consensus itself but just slot allocation, but still interested in your response)

(this is merely a question; not a review. As a review I'd recommend checking newline at EOF :p)

mxinden:
I am definitely the wrong person to answer the above in depth. Thanks for looking into this @Kianenigma!

### Why am I adding a runtime api?

The current set of authorities is known by the used substrate-runtime consensus module (aura or babe) within the runtime. Thereby to retrieve the set of authorities one has to interfere with the runtime. Thereby this pull request is adding a new runtime API.

### Why is this api generic instead of consensus specific?

If I understood the runtime abstraction correctly, it should be possible to hot-swap the consensus mechanism at runtime, by replacing the current logic within the web assembly abstraction by new logic.

See https://www.parity.io/substrate-has-arrived/:

> Down the road, consensus will be hot-swappable, meaning you can start your blockchain with one consensus mechanism and switch to another down the line without a hard fork.

With the above in mind code inside `core/client` or `core/service` can not assume the currently used consensus mechanism. To get around this, it would be using the generic `authorities` runtime API added with this pull request.

### What if a user still wants to use aura?

> Furthermore, how will this look while keeping in mind that some users might still want to use Aura?

If a user wants to use Aura, they will have to touch the code within `substrate/node/runtime` to set Aura as the consensus mechanism. Along the way they would also need to do the following change within `substrate/node/runtime`:

```diff
diff --git a/node/runtime/src/lib.rs b/node/runtime/src/lib.rs
index 547d2e1b1..786ff8541 100644
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -546,9 +546,9 @@ impl_runtime_apis! {
                }
        }
 
-       impl consensus_primitives::ConsensusApi<Block, babe_primitives::AuthorityId> for Runtime {
-               fn authorities() -> Vec<babe_primitives::AuthorityId> {
-                       Babe::authorities().into_iter().map(|(a, _)| a).collect()
+       impl consensus_primitives::ConsensusApi<Block, aura_primitives::AuthorityId> for Runtime {
+               fn authorities() -> Vec<aura_primitives::AuthorityId> {
+                       Aura::authorities().into_iter().map(|(a, _)| a).collect()
                }
        }
 }
```
---

@Kianenigma does the above make sense? Am I missing something?

mxinden:
> As a review I'd recommend checking newline at EOF

@Kianenigma which file are you referring to? As far as I can tell, they follow the same end-of-file convention like other files do:

```bash
$ find core/consensus/common -path ./target -prune -o  -type f | xargs file -k                                                                                                                                                                          
core/consensus/common/primitives/src/lib.rs:             ASCII text\012-                                                                                                                                                                                                                    
core/consensus/common/primitives/Cargo.toml:             ASCII text\012-                                                                                                                                                                                                                    
core/consensus/common/src/lib.rs:                        ASCII text\012-                                                                                                                                                                                                                    
core/consensus/common/src/offline_tracker.rs:            C source, ASCII text\012-                                                                                                                                                                                                          
core/consensus/common/src/select_chain.rs:               ASCII text\012-                                                                                                                                                                                                                    
core/consensus/common/src/import_queue.rs:               ASCII text\012-                                                                                                                                                                                                                    
core/consensus/common/src/block_import.rs:               ASCII text\012-                                                                                                                                                                                                                    
core/consensus/common/src/error.rs:                      ASCII text\012-                                                                                                                                                                                                                    
core/consensus/common/src/evaluation.rs:                 ASCII text\012-                                                                                                                                                                                                                    
core/consensus/common/src/import_queue/basic_queue.rs:   C source, ASCII text\012-                                                                                                                                                                                                          
core/consensus/common/src/import_queue/buffered_link.rs: ASCII text\012-                                                                                                                                                                                                                    
core/consensus/common/Cargo.toml:                        ASCII text\012-
```

Most of the files in substrate seem to use `CR`. But `.editorconfig` mentions `lf`. Which standard should I follow?

core/consensus/common/primitives/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Common consensus primitives.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use parity_codec::{Codec};
```
niklasad1:
```suggestion
use parity_codec::Codec;
```

dvc94ch:
This assumes that all consensus algorithms are proof of authority. Pow doesn't have the concept of authorities...

core/consensus/common/primitives/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+name = "substrate-consensus-common-primitives"
+version = "2.0.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+description = "Primitives for session"
+edition = "2018"
+
+[dependencies]
+parity-codec = { version = "4.1.1", default-features = false }
+client = { package = "substrate-client", path = "../../../client", default-features = false }
+runtime_primitives = { package = "sr-primitives", path = "../../../sr-primitives", default-features = false }
```
bkchr: Should be sr-primitives.

core/consensus/common/primitives/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "substrate-consensus-common-primitives"
+version = "2.0.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+description = "Primitives for session"
```
bkchr: Copy pasta?

bkchr:
When your chain uses pow, you just skip implementing the api.

bkchr:
> I'm a bit confused in terms of terminology here:
> 
> Afaik Substrate is layered like: `Network (fixed) -> Consensus (fixed) -> Runtime (variable)`.
> 
> So my understanding was that the runtime is actually unaware to some extent of what consensus is wrapping it. In other worse, the runtime just assumes that there exists some consensus in the client code with some unified Api. Based in this code though, it looks like _core/client is unaware of consensus_ (defining e.g. the returrn type of the Api as a generic which is encodable) and the _runtime being **aware**_ (replacing the generic with concrete values based on knowing that `babe` is being used). Can you clarify this a bit in terms of who knows what by default and which information gap are we filling here?
> 
> Furthermore, how will this look while keeping in mind that some users might still want to use Aura?
> 
> (I think I answered myself by recalling that babe/aura is not the consensus itself but just slot allocation, but still interested in your response)
> 
> (this is merely a question; not a review. As a review I'd recommend checking newline at EOF :p)

AFAIK the runtime will always be aware of the consensus being used. First you need to have the consensus related APIs implemented and secondly with pluggable consensus the runtime will decide when to switch to the new consensus by signaling it to the client.
Given the fact that we currently don't have pluggable consensus and yet don't really know how it will look like interface wise, the current api seems reasonable.
We could probably switch to a full opaque implementation, but I'm not sure if that is currently required.

core/consensus/common/primitives/Cargo.toml
mxinden: Good catch. I was implementing this as a runtime api against the session module first.

mxinden:
Thank you for all the comments. Can someone have another (final) look?

core/consensus/common/primitives/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+name = "substrate-consensus-common-primitives"
+version = "2.0.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+description = "Common consensus primitives"
+edition = "2018"
+
+[dependencies]
+parity-codec = { version = "4.1.1", default-features = false }
+client = { package = "substrate-client", path = "../../../client", default-features = false }
+sr-primitives = { package = "sr-primitives", path = "../../../sr-primitives", default-features = false }
```
bkchr:
```suggestion
sr-primitives = { path = "../../../sr-primitives", default-features = false }
```

DemiMarie-parity: @bkchr I just merged this, sorry!  This is a good reminder to not rush in the future, though.

Kianenigma: I can do a last round of checks for the pattern:

`sr-primitives = { package = "sr-primitives", ...` in #3214

Kianenigma: too late xD

mxinden: Fixed within https://github.com/paritytech/substrate/pull/3236

PR #3208, opened 2019-07-25
[WIP] Runtime telemetry for profiling

mattrutherford:
Opening this to discuss: profiling dispatchable functions in the runtime.

Allows for remote monitoring of this information which could be useful for many purposes. This would also be helpful for ongoing benchmarking, as it would negate the need to maintain forks of Substrate and Polkadot.

Is sent in the following format (prettified):

```json
{  
   "fn":"transfer",
   "is_ok":true,
   "level":"INFO",
   "mod":"srml_balances",
   "msg":"profiling.runtime",
   "ns":"41735",
   "ts":"2019-08-13T18:16:41.640896+02:00"
}
```

Would we also want to profile `on_initialize` and `on_finalize`? Or also more granular that that?

### DIFF
--- a/core/executor/src/allocator.rs
+++ b/core/executor/src/allocator.rs
@@ -22,6 +22,8 @@ use log::trace;
 use wasmi::MemoryRef;
 use wasmi::memory_units::Bytes;
 
+use substrate_telemetry::{telemetry, PROFILING};
+
 // The pointers need to be aligned to 8 bytes.
 const ALIGNMENT: u32 = 8;
 
@@ -40,6 +42,7 @@ pub struct FreeingBumpHeapAllocator {
 	max_heap_size: u32,
 	ptr_offset: u32,
 	total_size: u32,
+	historic_max: u32,
 }
 
 /// Create an allocator error.
@@ -74,6 +77,7 @@ impl FreeingBumpHeapAllocator {
 			max_heap_size: heap_size,
 			ptr_offset: ptr_offset,
 			total_size: 0,
+			historic_max: 0,
 		}
 	}
 
@@ -108,6 +112,12 @@ impl FreeingBumpHeapAllocator {
 
 		self.total_size = self.total_size + item_size + 8;
 		trace!(target: "wasm-heap", "Heap size is {} bytes after allocation", self.total_size);
+		if self.total_size > self.historic_max {
+			self.historic_max = self.total_size;
+			telemetry!(PROFILING; "profiling.wasm_max_mem";
+					"bytes" => self.historic_max
+				);
+		}
 
 		Ok(self.ptr_offset + ptr)
 	}
--- a/srml/support/src/dispatch.rs
+++ b/srml/support/src/dispatch.rs
@@ -60,6 +60,15 @@ pub trait Parameter: Codec + Clone + Eq {}
 #[cfg(not(feature = "std"))]
 impl<T> Parameter for T where T: Codec + Clone + Eq {}
 
+#[cfg(feature = "std")]
+thread_local! {
+	#[doc(hidden)]
+	pub static PROFILING: bool = match std::env::var("SUBSTRATE_PROFILING") {
+		Ok(ref e) if e == "1" => true,
+		_ => false,
+	};
+}
+
 /// Declares a `Module` struct and a `Call` enum, which implements the dispatch logic.
 ///
 /// ## Declaration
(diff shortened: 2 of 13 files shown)


Kianenigma:
I had a general thought a while back: 

This code is currently printing and timing entirely inside the runtime. Would also be interesting to add such an approach to the entry point of the runtime. This could be done in either of the `executor` codes. There are actually some benchmarks for the executor and they are much much slower than the numbers that this type of benchmarking gives.

mattrutherford:
Ready for someone to take a look, I think

srml/support/src/dispatch.rs
```diff
@@ -867,9 +876,31 @@ macro_rules! decl_module {
 		$vis fn $name(
 			$origin: $origin_ty $(, $param: $param_ty )*
 		) -> $crate::dispatch::Result {
-			{ $( $impl )* }
-			// May be unreachable.
-			Ok(())
+			#[cfg(feature = "std")]
```
kianenigma: why are we not gating this in another feature like `#[cfg(feature = "profile")]`?

srml/support/src/dispatch.rs
```diff
@@ -885,7 +916,29 @@ macro_rules! decl_module {
 	) => {
 		$(#[doc = $doc_attr])*
 		$vis fn $name($origin: $origin_ty $(, $param: $param_ty )* ) -> $result {
-			$( $impl )*
+	        #[cfg(feature = "std")]
+			return $crate::dispatch::PROFILING.with(|profiling| {
+				if *profiling {
+					let before = std::time::Instant::now();
+                    let result: $result = (move || { $( $impl )* })();
```
kianenigma: a single space misplaced.

core/executor/src/allocator.rs
```diff
@@ -108,6 +112,12 @@ impl FreeingBumpHeapAllocator {
 
 		self.total_size = self.total_size + item_size + 8;
 		trace!(target: "wasm-heap", "Heap size is {} bytes after allocation", self.total_size);
+		if self.total_size > self.historic_max {
+			self.historic_max = self.total_size;
+			telemetry!(PROFILING; "profiling.wasm_max_mem";
+					"bytes" => self.historic_max
+				);
```
kianenigma: bad indent.

core/executor/src/allocator.rs
```diff
@@ -108,6 +112,12 @@ impl FreeingBumpHeapAllocator {
 
 		self.total_size = self.total_size + item_size + 8;
 		trace!(target: "wasm-heap", "Heap size is {} bytes after allocation", self.total_size);
+		if self.total_size > self.historic_max {
```
kianenigma: I am not sure if I can follow the big-picture here but from the looks of it: you are calling telemetry whenever you are finding a new mex heap usage. depending the call-site of this `if self.total_size > self.historic_max {}`, how frequently will it be triggered? 

Alternatively, why not keep the condition, safe the max if we find a new one, and only send at the very end once? 

(again, I have no clue what/where this code is and am just reviewing based on seeing 10 lines so I might be totally out of cotext here 😀)

kianenigma:
> Would we also want to profile on_initialize and on_finalize? Or also more granular that that?

If possible, yes. Question is, if we easily can?

srml/support/src/dispatch.rs
kianenigma: as of now, I think these extra timing logics are exected all the time on a client with native runtime and sending a lot of data to telemetry; do we want this? (maybe we do and I don't know)

bkchr:
Why do you want to send this via telemetry? When you are doing some sort of profiling, you probably will do it locally and not with your node running somewhere. Then, we always collect this data. In my opinion we should never collect profiling information when we don't need them. 
I did not yet look into possible solutions, but is this really the best way we could go?

srml/support/src/dispatch.rs
mattrutherford: wrt feature - good question, I thought to do it this way, but it's possible also to use feature gate. Will answer q about telemetry below...

mattrutherford:
I think there's huge value in being able to continuously monitor this data. We've already seen from benchmarking that runtime performance changes as the state changes. I don't see ultimately how we can account for every possible scenario when doing isolated benchmarking (although I still think it has value). But there are things in production environments that are not easy to replicate in isolated benchmarks, like differing hardware and things that may change over time (in hard to predict ways), because of things like DB caching. If it's possible to just have one node running that's continuously sending this data to telemetry, where it can be stored in perpetuity, that seems like incredibly valuable information to have and can be easily cross-referenced to other data points. This data would allow insight to determine alert thresholds, including where rate of change accelerates compared to historic trends. I think telemetry is the best way to do this because it's already used for similar metrics and it allows easy opt-in and collection across different network boundaries.

pepyakin:
I think the I am very up for the general idea of continuous monitoring (Note, I'd avoid the word profiling since it has different meaning). To add to what @mattrutherford said, it would allow us to detect anomalies and regressions faster or even just detect at all.

bkchr:
The problem I see with continuous monitoring is that we more likely spend more time because we are doing the monitoring. 
I understand the need, but I would more likely switch to a more future proof approach. Something like: https://github.com/tokio-rs/tracing
This tracing library probably also supports somehow that you can pipe all the stuff into telemetry. And this library also supports `no_std`.

But your current approach is not really reusable. And I really don't want to have some hacked code in the runtime. I also still don't know if the telemetry will handle all this stuff, I remember reading about some problems with collecting the huge amount of data.

mattrutherford:
Will look into that @bkchr , thanks. Appreciate the point you made about re-usability - I had also considered that. 

AFAIK, the problems you mention with telemetry are fixed now, although it's obviously good to be mindful about how much we send.

bkchr:
Ty @mattrutherford. I'm more interested in a clean and reusable approach. 

Where you store or send it in the end, is not that important to me. When you say the telemetry can handle this and we want this, than I don't see any reason why we shouldn't do this :)

gavofyork:
what's the status of this, @mattrutherford ?

mattrutherford:
Closing in favour of forthcoming PR with a different approach, to address concerns stated here.

PR #3210, opened 2019-07-25
Use milliseconds for timestamp resolution

andresilva:
The srml timestamp module itself is unit dependent, as long as we are consistent with the data that we feed it. I only needed to update the `InherentDataProvider` to extract milliseconds from clock time, and the slot worker to use millisecond resolution for slot duration. This change is needed because BABE needs to be able to define slot durations at the millisecond level to account for the security parameters.

### DIFF
--- a/node/runtime/src/constants.rs
+++ b/node/runtime/src/constants.rs
@@ -18,21 +18,21 @@
 
 /// Money matters.
 pub mod currency {
-    use node_primitives::Balance;
+	use node_primitives::Balance;
 
-    pub const MILLICENTS: Balance = 1_000_000_000;
-    pub const CENTS: Balance = 1_000 * MILLICENTS;    // assume this is worth about a cent.
-    pub const DOLLARS: Balance = 100 * CENTS;
+	pub const MILLICENTS: Balance = 1_000_000_000;
+	pub const CENTS: Balance = 1_000 * MILLICENTS;    // assume this is worth about a cent.
+	pub const DOLLARS: Balance = 100 * CENTS;
 }
 
 /// Time.
 pub mod time {
-    use node_primitives::Moment;
+	use node_primitives::Moment;
 
-    pub const SECS_PER_BLOCK: Moment = 6;
-    pub const MINUTES: Moment = 60 / SECS_PER_BLOCK;
-    pub const HOURS: Moment = MINUTES * 60;
-    pub const DAYS: Moment = HOURS * 24;
+	pub const SECS_PER_BLOCK: Moment = 6;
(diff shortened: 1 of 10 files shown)


node/runtime/src/constants.rs
```diff
@@ -18,21 +18,25 @@
...
-    use node_primitives::Moment;
+	use node_primitives::Moment;
 
-    pub const SECS_PER_BLOCK: Moment = 6;
-    pub const MINUTES: Moment = 60 / SECS_PER_BLOCK;
-    pub const HOURS: Moment = MINUTES * 60;
-    pub const DAYS: Moment = HOURS * 24;
+	pub const MILLISECS_PER_BLOCK: Moment = 6000;
+	pub const SECS_PER_BLOCK: Moment = MILLISECS_PER_BLOCK / 1000;
+	pub const SLOT_DURATION: Moment = 6000;
```
Kianenigma: I assume this will not always be same as block time right? Given that, a short doc for at the least the two and how they are different would be helpful.

PS. Thanks for the _tab-ification_.

andresilva: Yes, you are right. If you don't mind I'll add some docs in the follow-up PR (#3211) where I set this correctly for BABE.

node/runtime/src/lib.rs
```diff
@@ -166,7 +166,7 @@ impl balances::Trait for Runtime {
 }
 
 parameter_types! {
-	pub const MinimumPeriod: u64 = SECS_PER_BLOCK / 2;
+	pub const MinimumPeriod: u64 = SLOT_DURATION / 2;
```
Kianenigma: not your change but historically wrong: this should be
```suggestion
	pub const MinimumPeriod: Moment = SLOT_DURATION / 2;
```

```
pub trait Trait: system::Trait {
	/// Type used for expressing timestamp.
	type Moment: Parameter + Default + SimpleArithmetic + ... 
...
	type MinimumPeriod: Get<Self::Moment>;
}
```

andresilva:
@tomusdrw Is this change relevant for offchain workers? Do I need to update anything there?

tomaka:
> I recall there was a docstring somewhere explaining why u64 as time representation, with seconds as resolution, is safe and sound, since it can cover X years.

That's not hard to calculate: 2^64 / 1000 / 3600 / 24 / 365.25 = 584542046 years

xlc:
Does this means it is possible to have less than 2s block time? very low block time are useful for development or testing purpose.

andresilva:
Yes, although I think Aura would be more useful for that.

PR #3211, opened 2019-07-25
node: Update BABE protocol parameters

andresilva:
~Depends on #3210. Only the last commit in this PR is relevant, you should review only that change until #3210 is merged.~

This PR updates the slot duration, epoch duration (to define it in slots), and `c` parameter for the BABE protocol. These parameters were chosen in accordance to the results in https://research.web3.foundation/en/latest/polkadot/BABE/Babe/#6-practical-results. Specifically, we are targetting a 6 second block time with resistance of maximum network delay up to 2 seconds. 

The epoch duration should be way longer in order to collect enough randomness, but for development purposes it is useful to have it relatively low. For Kusama we will use a much longer period.

cc @burdges

### DIFF
--- a/node/cli/src/service.rs
+++ b/node/cli/src/service.rs
@@ -371,7 +371,7 @@ mod tests {
 					&parent_id,
 					slot_num,
 					&alice,
-					(3, 10),
+					(278, 1000),
 				) {
 					break babe_pre_digest;
 				}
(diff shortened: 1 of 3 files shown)


andresilva:
With 4 authorities, after running for 250 blocks this is what I got (in seconds): 

Average block time - 5.317
Max block time - 33
Min block time - 1
99% Percentile - 20
95% Percentile - 15
90% Percentile - 11

andresilva:
Labeled inprogress so we don't merge before #3210.

PR #3193, opened 2019-07-24
Integrate srml/im-online with slashing

cmichi:
First draft based on https://github.com/paritytech/substrate/pull/3137.

Closes https://github.com/paritytech/substrate/issues/2719.

PR #3213, opened 2019-07-26
Refactor key management

gavofyork:
This PR is turning into a major refactoring over the crypto and key management system, which seems both long overdue and somewhat inescapable.

I'll want to integrate the general changes in (but not merge) #3034 and possibly #3193 together with #3137.

The main issues that this PR aims to solve are:
- Independent subsystems that happened to use the same crypto had to share the same session keys (and key-store division).
- Assumption that the twin consensus systems of block production (Aura/Babe) and finality (Grandpa) each required keys (which dirtied a lot of Substrate, placing APIs like `fn fg_authority_key()` all the way through the codebase). Though they do now, they won't forever and it shouldn't be an assumption that we bake in to internal APIs.
- Keys passed in through CLI are sticky; their existence enabled the consensus system workers and force them to use these keys alone. Rather whatever key in the key store that happens to be an authority should be used at any given time.

TODO:
- Refactor `ImOnline`:
  - [ ] Should get its own key and not piggy-back on other session keys.

- Dynamic key selection:
  - [ ] Enable Babe and Grandpa to be able to determine current authority key (if there is one).
  - [ ] Dynamically select the correct authoring key to use in each subsystem.
  - [ ] Use `--validator` to enable active validation process; remove any validator key-management stuff from the CLI (`--key`). (A second `--worker` arg can enable off-chain workers.)
  - [ ] Remove all static-key selection code (`--key` all the way down to key-choice).
  - [ ] Remove `authority_key`/`fg_authority_key` APIs completely.
  - [ ] Make `--password` work (should accept a manually-entered password and use it for everything in the keystore).
  - [ ] Introduce RPCs for session rotation, using code from #3034 if it makes sense. There should be a single RPC `author_rotateKeys` which creates a new set of session keys, saves them to the key-store and returns their public keys, so that a rotation-transaction can be constructed. There should be no pollution of the APIs with session keys: the rotation API should be wholly independent of each of the individual keys. This may mean new APIs in the `OpaqueKeys` trait.

- Refactor `Client`:
  - [ ] Create a basic `Client` trait that provides all the basic APIs and has `Block` as an associated item.

I would also like to solve the fact that service construction is highly monolithic and polluted by consensus-module-specific code. Really it should just be a case of specifying the block production and finality systems and letting it do the rest. Code for set-up, take-down, RPC and key-store integration should be provided through traits that the consensus modules (Babe, Aura, Grandpa and eventually Rho and PoW) implement *within their module's scope*. There shouldn't be any consensus-module-specific logic dirtying up `service.rs`. That might be a different PR, though.

- Refactor `service`:
  - [ ] Move out any consensus-module-specific code, and hardwired assumptions that they each have "authority keys", from `service.rs` into the respective consensus modules. Provide a single API endpoint (i.e. a type that implements some `ConsensusSystem` trait and have that as the only resource that service needs know of. 

This also adds:
- Add Call type to extensible transactions.
- Add `swap` to allow two storage items in the same map to be swapped.


core/sr-primitives/src/traits.rs
```diff
@@ -788,8 +788,8 @@ pub enum DispatchError {
 	/// General error to do with the transaction's proofs (e.g. signature).
 	BadProof,
 
-/*	/// General error to do with actually executing the dispatched logic.
-	User(&'static str),*/
+	/// General error to do with a limit being reached that may be reset on the next block.
+	BlockExhausted,
```
Kianenigma: Then maybe get rid of the `Resource` that I added above? they seem to have an overlap (at least when I added resource sth like _block exhausted_ was also a candidate for the same purpose).

srml/executive/src/lib.rs
```diff
@@ -114,6 +114,9 @@ mod internal {
 				DispatchError::Stale => ApplyError::Stale,
 				DispatchError::Future => ApplyError::Future,
 				DispatchError::BadProof => ApplyError::BadSignature(""),
+				DispatchError::BlockExhausted => ApplyError::Future,
```
Kianenigma: The same question: what is the difference between `::Resource` and `::BlockExhausted`.

I couldn't find any logic based on the error type except for the `basic_authorship.rs` retrying a few times, when it receives a `FullBlock` back, before quitting. `::Future` doesn't seem to re-schedule.

Kianenigma:
This could potentially deprecate the `info`. Both the inner and outer `Call`s implement `GetDispatchInfo` which can be used to get the same `struct info` being passed down from executive to signed extension.

Just needs some test-fixing as well, with which I could help if needed.

gavofyork:
yeah that'd be good. can also merge `Resource` and `BlockExhausted`

Kianenigma:
Signed-Extension-related tests should work fine now; I have the logic needed to throw out `info` from extension API (as mentioned, totally redundant with the new `call` parameter) packed in one commit but not sure if you want me to add it into this PR or not (given the big changes planned)?

core/primitives/src/crypto.rs
```diff
@@ -603,28 +670,440 @@ pub trait Pair: TypedKey + Sized + Clone + Send + Sync + 'static {
 	fn to_raw_vec(&self) -> Vec<u8>;
 }
 
+/// A type of supported crypto.
+#[derive(Clone, Copy, PartialEq, Eq, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(Debug))]
+#[repr(u32)]
+pub enum Kind {
```
rphmeier: The reason to avoid using an enum is that we will never be able to exhaustively define all crypto types. This change makes it impossible to introduce custom crypto without requiring an upstream pull request

rphmeier:
It would also be helpful to ensure that the model makes this possible. https://github.com/paritytech/substrate/pull/3063#discussion_r303354979

One reason that I've wanted `KeyTypeId` to exist is that it allows the runtime to make pretty much all crypto-based decisions. If we have a general keystore keyed by `KeyTypeId`, the runtime can tell us what kind of crypto to use and we will simply search our keystore for it. This enables Substrate to become extensible with user-crypto, as long as the user code adds support for registering a new kind of key.

gavofyork:
If it can be applied here, then I don't see why not.

core/primitives/src/crypto.rs
gavofyork: I'm not convinced it would be otherwise trivial to add new kinds of crypto. Realistically, that crypto would need extern support which would necessitate some alterations to core. However, if it were possible, then it would not be needed any time soon so can go in a separate PR.

core/primitives/src/crypto.rs
```diff
@@ -603,28 +672,440 @@ pub trait Pair: TypedKey + Sized + Clone + Send + Sync + 'static {
...
+	/// Key type for generic Ed25519 key.
+	pub const ED25519: KeyTypeId = *b"ed25";
+	/// Key type for Babe module, build-in.
+	pub const BABE: KeyTypeId = *b"babe";
+	/// Key type for Grandpa module, build-in.
+	pub const GRANDPA: KeyTypeId = *b"gran";
+	/// Key type for controlling an account in a Substrate runtime, built-in.
+	pub const ACCOUNT: KeyTypeId = *b"acco";
+	/// Key type for Aura module, built-in.
+	pub const AURA: KeyTypeId = *b"acco";
```
dvc94ch: Typo

dvc94ch:
So does the keystore have to persist keys to disk or not? I was under the impression that that was not the case, since the only keys in the keystore are session keys which should only be stored in memory according to a discussion on riot.

bkchr:
> So does the keystore have to persist keys to disk or not? I was under the impression that that was not the case, since the only keys in the keystore are session keys which should only be stored in memory according to a discussion on riot.

When you generate new session keys and only know the public key, what will you do when your node crashes? Let's say the node crashes and you already send the transaction to change the session key. You could not participate with this session key, as you have lost it. Sounds like we should persist them?

dvc94ch:
The previous thinking was that the client generates/persists the session keys and configures them via rpc.

bkchr:
> Introduce RPCs for session rotation, using code from #3034 if it makes sense. There should be a single RPC author_rotateKeys which creates a new set of session keys, saves them to the key-store and returns their public keys, so that a rotation-transaction can be constructed. There should be no pollution of the APIs with session keys: the rotation API should be wholly independent of each of the individual keys. This may mean new APIs in the OpaqueKeys trait.

Seems to be not wanted anymore.

gavofyork:
> Seems not to be wanted anymore.

Oh? Why?

bkchr:
Maybe I misunderstood @dvc94ch, but didn't he talked about sending the private key via RPC? Or is client the `substrate-client`? If yes, why should `substrate-client` persist the keys not using `keystore`?

As I understood your comment, you call `author_rotateKeys` and it generates new session key pairs. These pairs are persisted using the keystore and as answer of the RPC you get the public keys of the generated keys. So, the private will never leave the node.

dvc94ch:
I read the PR text, I was just wondering why the change. I believe @joe on riot and @gavofyork used to be of the opinion that offloading key management to the client is a good idea for security reasons, although I'm not really sure why that was.

bkchr:
I still don't know what you referring to when you talk about "client"^^

dvc94ch:
Wallet if you prefer...

tomaka:
> I would also like to solve the fact that service construction is highly monolithic and polluted by consensus-module-specific code. Really it should just be a case of specifying the block production and finality systems and letting it do the rest. Code for set-up, take-down, RPC and key-store integration should be provided through traits that the consensus modules (Babe, Aura, Grandpa and eventually Rho and PoW) implement within their module's scope. There shouldn't be any consensus-module-specific logic dirtying up service.rs. That might be a different PR, though.

I'm working on this and have a proof-of-concept of a better way to create the service, but considering the amount of changes I'm waiting for all PRs that touch the service to land before finishing the work.

gavofyork:
the idea is that both models are supported (user pushing private keys via RPC to the node and having the user pull public keys from node over RPC).

core/primitives/src/crypto.rs
rphmeier: Why should core need to know anything about any of the kinds of crypto?

gavofyork: Mainly for offchain worker stuff. Hopefully we can remove this.

core/consensus/babe/src/lib.rs
```diff
@@ -1195,6 +1197,117 @@ pub fn import_queue<B, E, Block: BlockT<Hash=H256>, I, RA, PRA>(
...
+	/// Instance of the source of sealing. Different consensus algorithms have different sealing
+	/// methods; for PoW it'll be a miner or mining instance. For PoA/PoS it'll be a key type.
+	type Sealer: SealingSource;
+
+	/// The Client type. Different services can make different constraints on this.
+	type Client: ProvideRuntimeApi + BlockOf + ProvideCache<<Self::Client as BlockOf>::Type>
+		+ HeaderBackend<<Self::Client as BlockOf>::Type>;
+
+	/// Initialize the consensus service.
+	fn initialize(client: &'a Self::Client, keystore: Arc<Store>) -> Self;
```
rphmeier: There doesn't seem to be a good reason why we should have service objects at all. 

The design of the pluggable consensus system was intended to avoid over-assumption about what resources a consensus system might need access to, and I'm afraid that's exactly what this trait function is -- it's omitting at a bare minimum network and event-loop access.

The primitive of "some thing that can be driven to completion" (i.e. a `Future`) is much less fragile.
I understand the motivation of wanting to avoid users having to write any consensus-specific initialization in their service files, but it should be 1 line per consensus engine at most, if we instead go the route of consensus crates exposing an `initialize` function (without fixed type signature!). 

That is much more anti-fragile compared to a bunch of trait machinery/macro magic trying to create an idealized declarative system that will almost certainly have to be ripped up later.

tomaka: Additionally, considering that this PR is pressing, let's please not introduce API changes that aren't strictly necessary, because we should discuss such API changes beforehand.

gavofyork: My concern is that it *actually* is "1 line per consensus engine at most".

Right now it's a load of incoherent boilerplate that chain creators can neither write nor understand.

gavofyork:
Closed in favour of #3296

Issue #1020, opened 2018-10-26
Reactivate RHD


gnunicorn:
Since 1.0 we are shipping with a pluggable consensus mechanism. In order to make this possible, we had to move out the previously installed `rhododendron`-PBFT, which has not been reactivated since. However, in order to proof that our consensus truly is pluggable we want to provide more than one algorithm to choose from. Let's reactivate the `substrate-consensus-rhd` crate and make rhd a possible choice for `node`. In order for that to be possible, one must:

- take a close look at the `substrate-consensus-rhd` crate and remove old, unnecessary cruft, adapt to latest API changes and make it compile and the tests pass again
- adapt the rhd-bindings to use the new messaging layer and system (introduced in #911)
- provide a rhd-Verifier to the ImportQueue-Mechanism
- change node (-template) to provide an example how to use rhd rather than Aura+Grandpa

PR #3216, opened 2019-07-26
Remove rhododendron

DemiMarie-parity:
It is unused and unmaintained, and has presumably regressed.  Remove it
for now.  If anyone wants to use it in the future, and has the resources
to maintain it, they are encouraged to bring it back from git history.


tomaka:
cc #1020

andresilva:
The code is already behind a feature flag, it is compartmentalized and there's really no maintenance burden as it is. I would prefer to keep it for discoverability, and also because at some point we want to get it working again (which shouldn't be too hard anyway). Substrate is meant to have pluggable consensus so having different consensus engines available is a plus, and Rhododendron provides instant finality so it could be preferred to BABE + GRANDPA for some users.

DemiMarie-parity:
Closing as per @andresilva’s review.

kigawas:
@andresilva 

Any specific plan on making it work again?

bkchr:
No, but if you need it. We are open to accept pull requests :)

tomaka:
I'd still be favour of removing rhododendron. What we have today is basically commented out code that doesn't compile.
Restoring the code later is as easy as doing `git checkout <commit> ./core/consensus/rhd`.

bkchr:
Cc @gnunicorn

andresilva:
I can make a PR to fix this after kusama, that is the proper way to fix this instead of just removing the code. Someone else can take it either if they are interested but I can't focus on this right now.

marcio-diaz:
> I can make a PR to fix this after kusama, that is the proper way to fix this instead of just removing the code. Someone else can take it either if they are interested but I can't focus on this right now.

I'm interested :raised_hands:

PR #3218, opened 2019-07-26
Minor cleanup in native_executor.rs

tomaka:
Remove a no-longer-true comment, and the `NativeExecutionDispatch::new` method.

### DIFF
--- a/core/client/src/genesis.rs
+++ b/core/client/src/genesis.rs
@@ -41,7 +41,7 @@ pub fn construct_genesis_block<
 mod tests {
 	use super::*;
 	use parity_codec::{Encode, Decode, Joiner};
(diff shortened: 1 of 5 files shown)


bkchr:
Why do you remove methods that are used?

tomaka:
I thought it was unused, as Substrate compiles just fine without it. It turns out that it is used, but only in tests.

It is a required trait method (so mandatory to implement). This trait is only ever implemented through this macro, so the user has no decision on that. And there's pretty much only way one you can implement this method anyway.

Because of these reasons, I strongly suspect this is some code that was mistakenly not removed during a past refactoring.

tomaka:
Oh, my guess is that it was created by someone who doesn't like the turbofish, so that you can write `Executor::new` instead of `NativeExecutor::<Executor>::new`

PR #3222, opened 2019-07-27
Update v1.0 runtime dependency for node-template, node and test-runtime

kaichaosun:
Fix https://github.com/substrate-developer-hub/substrate-package/issues/9


bkchr:
Does not build.

Issue #2986, opened 2019-07-01
Migrate Aura authorship-detection to Pre-runtime digests


rphmeier:
The inherent-based approach used before is now obsolete. See #2929 as an example of how to use `PreRuntime` digests in the runtime to compute authorship info.

This should be accompanied with a `FindAuthor<u32>` implementation.

DemiMarie-parity:
Are inherents obsolete at this point?

rphmeier:
@DemiMarie-parity for this use-case. Not for others.

thiolliere:
This is needed for staking to reward its validator, shouldn't it be Milestone 2.0-K ?

PR #3223, opened 2019-07-27
Add srml/im-online integrations

cmichi:
Follow-up to https://github.com/paritytech/substrate/issues/2719.

This PR tries to achieve two things:

* Once a session ends, all validators which according to `srml/im-online` were not online in the session which just ended should be disabled.
* Validators should automatically be marked online once they author a block.

Tests will follow in a follow-up PR, once the behavior/API is set.

### ToDos
- [x] Once a session ends, disable a validator if it wasn't online in the session which just ended.
- [x] Automatically mark validators as online once they author a block.
- [x] The two fields from the `ImOnlineConfig` don't need to be configurable, they should both just be `0` by default.
- [x] Update the `srml/im-online` version in Polkadot (here's the PR for the `master` version of this module, https://github.com/paritytech/polkadot/pull/342, needs to be updated once this PR is merged).

### Current Status
I checked two of the ToDos, since a review hasn't been done yet this is to be taken with reservation though.

The automatically-mark-online feature is implemented in this PR, but the `note_author()` hooks will only work properly after https://github.com/paritytech/substrate/issues/2986 has implemented `authorship::Trait` for the runtime (same goes for e.g. the `note_author()` hooks which slashing has implemented).

The tests run besides this one, which is deadlocking:
```
test tests::authoring_blocks ... test tests::authoring_blocks has been running for over 60 second
```
I suspect because of something to do with authorship.


srml/im-online/src/lib.rs
```diff
@@ -336,25 +355,56 @@ impl<T: Trait> Module<T> {
...
+		let curr = <session::Module<T>>::current_index();
+		let got_heartbeat = <ReceivedHeartbeats<T>>::exists(&curr, authority_id);
+		let was_author = <BlockAuthors<T>>::exists(&curr, authority_id);
+		got_heartbeat || was_author
+	}
+
+	/// Returns `true` if a heartbeat has been received for `AuthorityId`
+	/// during the previous session. Otherwise `false`.
+	pub fn is_online_in_previous_session(authority_id: &T::AuthorityId) -> bool {
+		let index = <session::Module<T>>::current_index() - 1;
```
niklasad1: what if `index` == 0 here? then it may wrap to u32::max_value() is that a problem?

srml/im-online/src/lib.rs
```diff
@@ -336,25 +355,56 @@ impl<T: Trait> Module<T> {
 			Some(start) => {
 				// iterate over every session
 				for index in start..curr  {
-					if <ReceivedHeartbeats<T>>::exists(&index, authority_id) {
+					let got_heartbeat = <ReceivedHeartbeats<T>>::exists(&index, authority_id);
+					let was_author = <BlockAuthors<T>>::exists(&index, authority_id);
+					if  got_heartbeat || was_author {
```
niklasad1:
```suggestion
					if got_heartbeat || was_author {
```

gavofyork:
This should all be sorted in #3296

PR #3225, opened 2019-07-27
Add transaction pool to Aura and Babe import queue

marcio-diaz:
We need to submit report transactions when we discover an equivocation while importing [(see here for definition of equivocation)](https://research.web3.foundation/en/latest/polkadot/slashing/).

As a starting point, this PR adds the transaction pool to Aura and Babe import queue (and `check_header`).

### DIFF
--- a/core/consensus/babe/src/lib.rs
+++ b/core/consensus/babe/src/lib.rs
@@ -464,7 +464,7 @@ fn find_next_epoch_digest<B: BlockT>(header: &B::Header) -> Result<Option<Epoch>
 ///
 /// This digest item will always return `Some` when used with `as_babe_pre_digest`.
 // FIXME #1018 needs misbehavior types
-fn check_header<B: BlockT + Sized, C: AuxStore>(
+fn check_header<B: BlockT + Sized, C: AuxStore, T>(
 	client: &C,
 	slot_now: u64,
 	mut header: B::Header,
@@ -473,8 +473,10 @@ fn check_header<B: BlockT + Sized, C: AuxStore>(
 	randomness: [u8; 32],
 	epoch_index: u64,
 	c: (u64, u64),
-) -> Result<CheckedHeader<B::Header, (DigestItemFor<B>, DigestItemFor<B>)>, String>
-	where DigestItemFor<B>: CompatibleDigestItem,
+	_transaction_pool: &T,
+) -> Result<CheckedHeader<B::Header, (DigestItemFor<B>, DigestItemFor<B>)>, String> where
+	DigestItemFor<B>: CompatibleDigestItem,
+	T: Send + Sync + 'static,
 {
 	trace!(target: "babe", "Checking header");
 	let seal = match header.digest_mut().pop() {
@@ -548,14 +550,15 @@ fn check_header<B: BlockT + Sized, C: AuxStore>(
 pub struct BabeLink(Arc<Mutex<(Option<Duration>, Vec<(Instant, u64)>)>>);
 
 /// A verifier for Babe blocks.
-pub struct BabeVerifier<C> {
+pub struct BabeVerifier<C, T> {
 	api: Arc<C>,
 	inherent_data_providers: inherents::InherentDataProviders,
 	config: Config,
 	time_source: BabeLink,
+	transaction_pool: Option<Arc<T>>,
 }
 
-impl<C> BabeVerifier<C> {
+impl<C, T> BabeVerifier<C, T> {
 	fn check_inherents<B: BlockT>(
 		&self,
 		block: B,
@@ -625,9 +628,10 @@ fn median_algorithm(
 	}
 }
 
-impl<B: BlockT, C> Verifier<B> for BabeVerifier<C> where
+impl<B: BlockT, C, T> Verifier<B> for BabeVerifier<C, T> where
 	C: ProvideRuntimeApi + Send + Sync + AuxStore + ProvideCache<B>,
 	C::Api: BlockBuilderApi<B> + BabeApi<B>,
+	T: Send + Sync + 'static,
 {
 	fn verify(
 		&mut self,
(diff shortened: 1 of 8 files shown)


tomaka:
Please add a comment in the Babe code explaining why we're passing the transaction pool.

We merge too many PRs that adds code "for later", and we end up not using that code, and then nobody remembers why it's there nor dares to remove it.

bkchr:
I would go even further as @tomaka and say, why not build directly the full functionality? This is no big change, which can be reviewed together with the actual functionality.

marcio-diaz:
> I would go even further as @tomaka and say, why not build directly the full functionality? This is no big change, which can be reviewed together with the actual functionality.

The full functionality may be too much for a PR. But sure, I will push more code to give a bigger picture to the reviewers.

marcio-diaz:
I added some more code, but looks like I can't go much further. The full functionality depends on #3063 and  #3137 at least.

core/consensus/babe/primitives/src/equivocation.rs
```diff
@@ -0,0 +1,124 @@
...
+	/// Get the first signature involved in the equivocation.
+	fn first_signature(&self) -> &S {
+		&self.first_signature
+	}
+
+	/// Get the second signature involved in the equivocation.
+	fn second_signature(&self) -> &S {
+		&self.second_signature
+	}
+}
```
DemiMarie-parity: Nit: missing newline at EOF

core/finality-grandpa/Cargo.toml
```diff
@@ -25,10 +25,10 @@ network = { package = "substrate-network", path = "../network" }
 service = { package = "substrate-service", path = "../service", optional = true }
 srml-finality-tracker = { path = "../../srml/finality-tracker" }
 fg_primitives = { package = "substrate-finality-grandpa-primitives", path = "primitives" }
-grandpa = { package = "finality-grandpa", version = "0.8.1", features = ["derive-codec"] }
+grandpa = { package = "finality-grandpa", path = "/home/marcio/repos/finality-grandpa/", features = ["derive-codec"] }
```
DemiMarie-parity: This won’t work for anyone but you.  This path should be relative.

core/finality-grandpa/Cargo.toml
```diff
@@ -25,10 +25,10 @@ network = { package = "substrate-network", path = "../network" }
...
 
 [dev-dependencies]
-grandpa = { package = "finality-grandpa", version = "0.8.1", features = ["derive-codec", "test-helpers"] }
+grandpa = { package = "finality-grandpa", path = "/home/marcio/repos/finality-grandpa/", features = ["derive-codec", "test-helpers"] }
```
DemiMarie-parity: Ditto.

marcio-diaz:
> This is not reviewable. You bundled a bunch of changes from other PRs into a single commit ([599c8b1](https://github.com/paritytech/substrate/commit/599c8b1f69ebdf8fa3ab2310e136a63cd71f2efb)). If you depend on other PRs the correct way to do it is to create a branch that descends from the one you depend on. FWIW I'm fine with merging just the changes to introduce the transaction pool to Babe (and Aura!) without depending on any pending PRs, provided that we comment it like @tomaka suggested.

I have not intention to merge the other PR, just to use it as a guide.

I'll follow the suggestions of @tomaka  and @andresilva  and just introduce the changes for the tx pool in Aura and Babe.

core/finality-grandpa/Cargo.toml
marcio-diaz: Yeah, it was depending on (now merged) changes in `Grandpa` repository (and I think `branch` was not working).

marcio-diaz:
This is ready for review.

core/consensus/babe/src/lib.rs
```diff
@@ -463,8 +463,9 @@ fn find_next_epoch_digest<B: BlockT>(header: &B::Header) -> Result<Option<Epoch>
 /// unsigned.  This is required for security and must not be changed.
 ///
 /// This digest item will always return `Some` when used with `as_babe_pre_digest`.
-// FIXME #1018 needs misbehavior types
-fn check_header<B: BlockT + Sized, C: AuxStore>(
+// FIXME #1018 needs misbehavior types. The `transaction_pool` parameter will be 
+// used to submit such misbehavior reports.
+fn check_header<B: BlockT + Sized, C: AuxStore, T>(
```
andresilva: Shouldn't we restrict the `T` type here, add some bounds to make sure it is some transaction pool type? Maybe actually receive a `Option<&T>` here since it seems like the transaction pool will be optional, and we just won't send reports if it is not available.

marcio-diaz: Yes, we should restrict `T` with a trait that still needs to be defined (IMO this also depends on  #3063). So I prefer to leave it open for now. 

I will change it to `Option<&T>`.

andresilva: https://github.com/paritytech/substrate/blob/master/core/network/src/service.rs#L65

marcio-diaz: We also need to create the tx here, right?

I'm not sure it works, will try to check.

marcio-diaz:
> As discussed offline this is not usable as is, since the `T` type is not bound by anything. @marcio-diaz did some changes to fix this on another branch not sure if we wants to get them here or not.

Now, that I read the comment again, it feels like an arbitrary decision.
`T` is useless with or without bound, and adding it doesn't look perfect either.
IMO, the point of splitting the other big PR was to facilitate review and merging.
If anyways we are going to merge this in a few days, then why not merging it now?

andresilva:
IMO my comment is justified, this PR is titled "Add transaction pool", and it adds a type `T` which does not relate at all to a `TransactionPool`, so as a logical unit of work it doesn't accomplish what it says it does. You added a trait `SubmitExtrinsic` here (https://github.com/paritytech/substrate/pull/3314) which solves this exact problem, my suggestion was to add something like that here in this PR. If that is done then in this PR we have a small set of changes and something that is logically correct (even though it is unused). I am OK with merging this anyway and just doing those changes in the other PR to avoid more work on this, but it has conflicts now.

marcio-diaz:
@andresilva Thanks, conflicts solved.

gavofyork:
@andresilva if you're happy with it, then please merge.

PR #3228, opened 2019-07-29
Create .rustfmt.toml

kigawas:
Fix `rustfmt` problem
Ref: https://github.com/rust-lang/rls-vscode/issues/466


bkchr:
We don't use `rustfmt`. So, we don't need this file

PR #3229, opened 2019-07-29
add post_dispatch

### LINKED ISSUE #3227: post_dispatch hook for SignedExtension
Currently there is a `pre_dispatch` hook to do additional work before transaction is dispatched. It will be useful to have a `post_dispatch` hook to do additional work after transaction is successfully dispatched. 
My use case is to implement a fees module that emit `fees.Charged(Amount)` event to notify how much transaction fee is been charged in this transaction. This have to be in `post_dispatch` because modules may charge additional fee (balances create account) or refund fee (contracts refund gas fee).

xlc:
Closes #3227


Kianenigma:
Looks good but maybe wait until some last changes in #3213 are included and integrate them in? Namely: 

- `call: &Self::Call` is added as an argument to the Api.
- Consequently, `info` is not really needed anymore since `Call` can carry it around.

core/sr-primitives/src/generic/checked_extrinsic.rs
```diff
@@ -85,7 +85,9 @@ where
 			Extra::pre_dispatch_unsigned(info, len)?;
 			None
 		};
-		Ok(self.function.dispatch(Origin::from(maybe_who)))
+		let res = self.function.dispatch(Origin::from(maybe_who));
+		Extra::post_dispatch(info, len);
```
xlc: I tried to rebase on #3296 and figure out that I can't just add `call` to `post_dispatch` because it is been moved in the dispatch call above.
I would like to avoid the clone of `self.function` because otherwise it won't be zero overhead.
I don't have a use case to include `call` in `post_dispatch` so I don't think we need to add it now.

gavofyork:
looks good; will need to define `type Pre = ();` for every impl of `SignedExtension`.

PR #3232, opened 2019-07-29
Update to parity-scale-codec

thiolliere:
Update parity-codec to parity-scale-codec.

To do so it also updates the following dependencies:
* [x] release scale-codec: https://github.com/paritytech/parity-scale-codec/pull/144
* [x] primitive-types: https://github.com/paritytech/parity-common/pull/192
* [x] grandpa: https://github.com/paritytech/finality-grandpa/pull/80
* [x] rhododendron: https://github.com/paritytech/rhododendron/pull/23
* [x] trie (trie-bench)

Also naming has been normalized: parity-scale-codec is imported as codec for every crates.

in sr-primitive there was a use of an infinite input which consisted of a wrapped input followed by infinite zeros. This is now converted to finite input with given number of zero. the number of zero choosen is 32*1024. I didn't implemented u32::max_value number of zero to avoid maybe some issue of having a not constrained decode. (because pre-allocation is constrained by input len).


thiolliere:
hmm I currently have this issue: 

```
thread 'tests::voter_catches_up_to_latest_round_when_behind' panicked at 'assertion failed: set_state.read().last_completed_round().number >= 4', core/finality-grandpa/src/tests.rs:1501:5
```

srml/grandpa/src/lib.rs
```diff
@@ -78,11 +78,11 @@ pub struct StoredPendingChange<N> {
 }
 
 impl<N: Decode> Decode for StoredPendingChange<N> {
-	fn decode<I: codec::Input>(value: &mut I) -> Option<Self> {
+	fn decode<I: codec::Input>(value: &mut I) -> core::result::Result<Self, Error> {
 		let old = OldStoredPendingChange::decode(value)?;
-		let forced = <Option<N>>::decode(value).unwrap_or(None);
+		let forced = <Option<N>>::decode(value)?;
```
thiolliere: this was introduce to support old format.

But a failing decode of a value result in weird behavior on input: the some data could have been read and removed from input, thus the decode of `Vec<StoredPendingChange<N>>` with stored_pengind change in old format (without the option) will result in having the byte after first element removed and so a two element vec in old format will fail.

core/sr-primitives/src/traits.rs
```diff
@@ -1055,21 +1055,31 @@ pub trait OpaqueKeys: Clone {
...
 	/// Verify a proof of ownership for the keys.
 	fn ownership_proof_is_valid(&self, _proof: &[u8]) -> bool { true }
 }
 
 struct TrailingZeroInput<'a>(&'a [u8]);
 impl<'a> codec::Input for TrailingZeroInput<'a> {
-	fn read(&mut self, into: &mut [u8]) -> usize {
+	fn remaining_len(&mut self) -> Result<usize, codec::Error> {
+		// TODO TODO: this might not be super correct: remaining_len change during execution
+		// Also this gives attack possibility no ?
```
thiolliere: still to be thought: this TrailingZeroInput is adding infinite number of 0 after input. but input length is now restricted to be known and is used to constrain allocation. thus I'm not the how we should implement it for now

srml/grandpa/src/lib.rs
thiolliere: I removed it this should be done in another PR if so

core/service/src/lib.rs
```diff
@@ -877,28 +877,31 @@ impl<C: Components> network::TransactionPool<ComponentExHash<C>, ComponentBlock<
...
-					},
-					Ok(e) => {
-						debug!("Error adding transaction to the pool: {:?}", e);
-						None
-					},
-					Err(e) => {
-						debug!("Error converting pool error: {:?}", e);
-						None
-					},
+		match Decode::decode(&mut &encoded[..]) {
```
DemiMarie-parity: Should this still be an `if let`?

core/sr-primitives/Cargo.toml
```diff
@@ -17,7 +17,7 @@ paste = { version = "0.1"}
 
 [dev-dependencies]
 serde_json = "1.0"
-primitive-types = "0.4"
+primitive-types = { version = "0.5.0" }
```
DemiMarie-parity:
```suggestion
primitive-types = "0.5.0"
```

core/trie/src/error.rs
```diff
@@ -12,22 +12,36 @@ use std::fmt;
 use std::error::Error as StdError;
 
 #[derive(Debug, PartialEq, Eq, Clone)]
-/// Error concerning the Parity-Codec based decoder.
+/// Error for trie node decoding.
 pub enum Error {
 	/// Bad format.
 	BadFormat,
```
DemiMarie-parity: Is this still used?

core/service/src/lib.rs
thiolliere: instead of if let ... else we now do a match Ok Err in order to have the error msg

DemiMarie-parity: `.map_err(|s| something)?` would be a good choice too, but the current code is good enough as is.

srml/metadata/src/lib.rs
```diff
@@ -333,9 +339,11 @@ impl Encode for RuntimeMetadataDeprecated {
 	fn encode_to<W: Output>(&self, _dest: &mut W) {}
 }
 
+impl codec::EncodeLike for RuntimeMetadataDeprecated {}
+
 #[cfg(feature = "std")]
 impl Decode for RuntimeMetadataDeprecated {
-	fn decode<I: Input>(_input: &mut I) -> Option<Self> {
+	fn decode<I: Input>(_input: &mut I) -> Result<Self, Error> {
 		unimplemented!()
```
bkchr:
```suggestion
		Err("Decoding is not supported".into())
```

core/sr-primitives/src/traits.rs
```diff
@@ -1029,21 +1029,35 @@ pub trait OpaqueKeys: Clone {
...
+	fn read(&mut self, into: &mut [u8]) -> Result<(), codec::Error> {
+		let len_from_input = into.len().min(self.input.len());
+		into[..len_from_input].copy_from_slice(&self.input[..len_from_input]);
+		for i in &mut into[len_from_input..] {
 			*i = 0;
 		}
-		self.0 = &self.0[len..];
-		into.len()
+		self.input = &self.input[len_from_input..];
+		self.trailing_zeros = self.trailing_zeros.checked_sub(into.len() - len_from_input)
```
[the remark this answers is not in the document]
thiolliere: I don't think so this check is for when we reach the end of the trailings zero. but whatever as we changed

core/trie/src/error.rs
thiolliere: yes  core/trie/src/node_codec.rs  lines 73 for instance

core/consensus/babe/primitives/src/digest.rs
```diff
@@ -72,21 +74,26 @@ impl Encode for BabePreDigest {
...
 
 		// Verify (at compile time) that the sizes in babe_primitives are correct
 		let _: [u8; super::VRF_OUTPUT_LENGTH] = vrf_output;
 		let _: [u8; super::VRF_PROOF_LENGTH] = vrf_proof;
-		Some(BabePreDigest {
-			vrf_proof: VRFProof::from_bytes(&vrf_proof).ok()?,
-			vrf_output: VRFOutput::from_bytes(&vrf_output).ok()?,
+		Ok(BabePreDigest {
+			vrf_proof: VRFProof::from_bytes(&vrf_proof)
+				.map_err(|_| Error::from("Invalid VRFProof"))?,
```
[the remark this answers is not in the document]
thiolliere: yes improved, conversion is from signature error is not full though if we want to put more information we need to have a Cow instead of a static str. but I feel like it is ok.

thiolliere:
so comment adressed and also I put version to 1.0.0 for scale-codec as it is fine with this version

thiolliere:
why is CI not answering though ?

PR #3233, opened 2019-07-29
Allow creating a ChainSpec from a Vec<u8>

tomaka:
For #2416 I'd like to be able to pass the chain-spec JSON from the JavaScript API.

Right now, you can create a `ChainSpec` from only three ways: from a `&'static [u8]`, from a file (through a `PathBuf`), or manually.

This PR adds the possibility to pass a `Vec<u8>`.

### DIFF
--- a/core/service/src/chain_spec.rs
+++ b/core/service/src/chain_spec.rs
@@ -161,11 +161,6 @@ impl<G: RuntimeGenesis> ChainSpec<G> {
 	}
 
 	/// Parse json content into a `ChainSpec`
-	pub fn from_embedded(json: &'static [u8]) -> Result<Self, String> {
-		Self::from_json_bytes(json)
-	}
-
-	/// Parse json content into a `ChainSpec`
 	pub fn from_json_bytes(json: impl Into<Cow<'static, [u8]>>) -> Result<Self, String> {
 		let json = json.into();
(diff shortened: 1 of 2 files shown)


core/service/src/chain_spec.rs
```diff
@@ -159,7 +160,13 @@ impl<G: RuntimeGenesis> ChainSpec<G> {
 
 	/// Parse json content into a `ChainSpec`
 	pub fn from_embedded(json: &'static [u8]) -> Result<Self, String> {
```
bkchr: Why not drop this function?

core/service/src/chain_spec.rs
```diff
@@ -29,15 +30,15 @@ use tel::TelemetryEndpoints;
 
 enum GenesisSource<G> {
 	File(PathBuf),
-	Embedded(&'static [u8]),
+	Embedded(Cow<'static, [u8]>),
```
bkchr:
```suggestion
	Binary(Cow<'static, [u8]>),
```

Or something different, but embedded sounds not that good.

core/service/src/chain_spec.rs
tomaka: Backwards compatibility

bkchr: We don't have any released version, so just remove it.

PR #3234, opened 2019-07-29
srml-contracts: Forbid declaration of float types in contracts

jimpo:
Currently, floating point *operations* are forbidden in function bodies of contract modules by the gas metering injection routine. However, it is still possible to declare global variables and unused local variables as `f32` or `f64`. For the sake of consistency, this scans the module and rejects any usage of floating point types.


node/runtime/src/lib.rs
```diff
@@ -80,7 +80,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 122,
-	impl_version: 123,
+	impl_version: 124,
```
pepyakin: It looks like the runtime behavior is changed by this PR and thus according to the comment above we need to bump `spec_version`.

srml/contracts/src/wasm/prepare.rs
```diff
@@ -93,6 +93,50 @@ impl<'a> ContractModule<'a> {
...
 
+	/// Ensures that no floating point types are in use.
+	fn ensure_no_floating_types(&self) -> Result<(), &'static str> {
+		if let Some(global_section) = self.module.global_section() {
+			for global in global_section.entries() {
+				match global.global_type().content_type() {
+					ValueType::F32 | ValueType::F64 =>
+						return Err("use of floating point type in globals is forbidden"),
+					_ => {}
+				}
```
pepyakin: Nit: It looks to me that `if let` would serve the purpose better. Also I am ok with using a macro/function using a less informative error message.

PR #3214, opened 2019-07-26
Refactor sr_primitives.

Kianenigma:
- `core/sr-primitives` is now always `sr-primitives`.
- `core/primitives` (`package = "substrate-primitives"`) is now always `primitives`.

didn't check `node-primitives` but should be clear which one is which now.

PR #3236, opened 2019-07-29
core/network: Fix bad import %s/runtime_primitives/sr_primitives/

mxinden:
Relates to https://github.com/paritytech/substrate/pull/3214.


mxinden:
> Isn't this also the thing that broke master temporarily?

Not quite sure I get the question. This is the thing I was referring to in Riot in regards to the current master failures.

PR #3238, opened 2019-07-29
Service various cleanups

tomaka:
I am currently in the process of refactoring the service creation API.
The refactoring itself is waiting for other changes to land first, but these are various small cleanups that I performed in the work-in-progress.
As a summary, this basically removes the dependencies on the `Components` trait.

Can be reviewed commit by commit.

There are two changes that are probably breaking:

- Call `create_config_with_db_path<F, _>(...)` should be replaced with just `create_config_with_db_path(...)`.
- `service.config` is no longer a public field (that's because I'd like to base `Service` on top of something else), but I added `config()` and `config_mut()` methods to replace that.


core/service/src/lib.rs
```diff
@@ -502,6 +495,27 @@ impl<Components: components::Components> Service<Components> {
...
+	pub fn config(&self) -> &FactoryFullConfiguration<Components::Factory> {
+		&self.config
+	}
+
+	/// Returns a reference to the config passed at initialization.
+	///
+	/// > **Note**: This method is currently necessary because we extract some elements from the
+	/// >			configuration at the end of the service initialization. It is intended to be
+	/// >			removed.
+	pub fn config_mut(&mut self) -> &mut FactoryFullConfiguration<Components::Factory> {
```
DemiMarie-parity: Should this be `#[deprecated]`?

tomaka: There's no way to replace it at the moment.

The configuration at the moment has a field that acts as some sort of "generic storage for variables you need to pass between the multiple stages of building a service", which is an extremely crappy way of dealing with the fact that we're having this weird `Factory` trait.

PR #3240, opened 2019-07-29
Properly stop block import on error

tomaka:
cc @jimpo 

Calling `import_blocks` returns a `Future` that imports the blocks passed as parameter into the database. However, at the moment, if an error happens in the import queue, that `Future` will just wait forever and never stop. This PR fixes this.

Additionally, improves the error being displayed on error.

### DIFF
--- a/core/service/src/chain_ops.rs
+++ b/core/service/src/chain_ops.rs
@@ -117,13 +117,17 @@ impl<B: Block> Link<B> for WaitLink {
 	fn blocks_processed(
 		&mut self,
 		imported: usize,
-		count: usize,
+		_count: usize,
 		results: Vec<(Result<BlockImportResult<NumberFor<B>>, BlockImportError>, B::Hash)>
 	) {
 		self.imported_blocks += imported as u64;
-		if results.iter().any(|(r, _)| r.is_err()) {
-			warn!("There was an error importing {} blocks", count);
-			self.has_error = true;
+
+		for result in results {
+			if let (Err(err), hash) = result {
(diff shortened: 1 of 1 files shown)


core/service/src/chain_ops.rs
```diff
@@ -198,6 +205,13 @@ pub fn import_blocks<F, E, R>(
 			queue.poll_actions(cx, &mut link);
 			std::task::Poll::Pending::<Result<(), ()>>
 		}).compat().poll();
+		if link.has_error {
+			info!(
+				"Stopping after #{} blocks because of an error",
+				link.imported_blocks,
+			);
+			return Ok(Async::Ready(()));
```
DemiMarie-parity: Should it return `Err` instead?

tomaka: That's debatable, depending on what we want the semantics of the function to be.
Considering that the function handles printing out what happens, I went with returning `Ok` as a way to say "we're done with this operation". Once we switch to stable futures you'd just return a `()` instead of a `Result`.

tomaka: To expand, if instead we were to return an `Err`, I would make it so that the `Ok` or the `Err` contains a report of what happens, and it's the caller that would print that out.

andresilva: This is a future that will be spawned and run to completion in one place, so I'm fine with keeping it as it is. It's not composable anyway as is because it prints all the stuff.

jimpo:
Confirmed this fix locally.

PR #3241, opened 2019-07-29
srml: babe: add expected block time and epoch duration constants

andresilva:
Since BABE is probabilistic it is not trivial to figure out what the expected average block time should be based on slot duration and the security parameter `c`. Still this is useful for the dashboard to be able to show what the expected block time is (and maybe compare it with what we're actually getting), so this PR adds a constant field with this value to the BABE runtime module.

Additionally I also exposed the `EpochDuration` field since it is needed for the dashboard to calculate progress in current session.

cc @jacogr

### DIFF
--- a/srml/babe/src/lib.rs
+++ b/srml/babe/src/lib.rs
@@ -164,7 +164,8 @@ decl_module! {
 		/// The expected average block time at which BABE should be creating
 		/// blocks. Since BABE is probabilistic it is not trivial to figure out
 		/// what the expected average block time should be based on the slot
-		/// duration and the security parameter `c`.
+		/// duration and the security parameter `c` (where `1 - c` represents
+		/// the probability of a slot being empty).
 		const ExpectedBlockTime: T::Moment = T::ExpectedBlockTime::get();
 
 		/// Initialization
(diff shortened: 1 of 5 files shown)


srml/babe/src/lib.rs
```diff
@@ -156,6 +157,16 @@ decl_storage! {
...
 	/// The BABE SRML module
 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		/// The number of **slots** that an epoch takes. We couple sessions to
+		/// epochs, i.e. we start a new session once the new epoch begins.
+		const EpochDuration: u64 = T::EpochDuration::get();
+
+		/// The expected average block time at which BABE should be creating
+		/// blocks. Since BABE is probabilistic it is not trivial to figure out
+		/// what the expected average block time should be based on the slot
+		/// duration and the security parameter `c`.
```
marcio-diaz: I would like the code to be as self contained as possible. For instance, we can mention somewhere that `c` denotes the probability of non-empty slot (as far I understood).

marcio-diaz: Also, maybe add a test before merging ;)

andresilva:
Expanded docs to explain what `c` is. I can add tests but I'm not sure what to test, all I'm doing in this PR is add two constants, there's no logic to test here.

PR #3246, opened 2019-07-30
Synchronize state cache on finalization

arkpar:
The cache is not currently updated to match the best state when re-org is caused by setting finality, and not by importing a block. 
Also fixed informant reporting false re-orgs.

### DIFF
--- a/core/client/db/src/storage_cache.rs
+++ b/core/client/db/src/storage_cache.rs
@@ -155,7 +155,7 @@ impl<B: BlockT, H: Hasher> Cache<B, H> {
 	/// Synchronize the shared cache with the best block state.
 	/// This function updates the shared cache by removing entries
 	/// that are invalidated by chain reorganization. It should be
-	/// be called when chain reorg rappens without importing a new block.
+	/// be called when chain reorg happens without importing a new block.
 	pub fn sync(&mut self, enacted: &[B::Hash], retracted: &[B::Hash]) {
 		trace!("Syncing shared cache, enacted = {:?}, retracted = {:?}", enacted, retracted);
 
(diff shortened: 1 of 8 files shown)


core/client/src/client.rs
```diff
@@ -2690,4 +2690,93 @@ pub(crate) mod tests {
...
+
+	#[test]
+	fn state_reverted_on_set_head() {
+		let _ = env_logger::try_init();
+		let client = test_client::new();
+
+		let alice = blake2_256(&runtime::system::balance_of_key(AccountKeyring::Alice.into())).to_vec();
+		let current_balance = ||
+			u64::decode(
+				&mut client.storage(&BlockId::number(client.current_height()), &StorageKey(alice.clone()))
```
bkchr: Instead of accessing the raw storage, you use the `TestApi`.

`client.runtime_api().balance_of(&BlockId::number(client.current_height()), AccountKeyring::Alice.into())`

core/client/db/src/storage_cache.rs
```diff
@@ -151,6 +151,65 @@ impl<B: BlockT, H: Hasher> Cache<B, H> {
 			+ self.lru_child_storage.used_size()
 			//  ignore small hashes storage and self.lru_hashes.used_size()
 	}
+
+	/// Synchronize the shared cache with the best block state.
+	/// This function updates the shared cache by removing entries
+	/// that are invalidated by chain reorganization. It should be
+	/// be called when chain reorg rappens without importing a new block.
```
andresilva:
```suggestion
	/// be called when chain reorg happens without importing a new block.
```

PR #3247, opened 2019-07-30
[WIP] Enable validators to find and connect to other validators

mxinden:
The goal of this pull request is to enable Substrate validators to directly connect to other validators. With this pull request a validator does two things:

1. **Making itself discoverable**
    1. Retrieve its external addresses
    2. Adds its network peer id to the addresses
    3. Sign the above
    4. Put the signature and the addresses on the libp2p Kademlia DHT

2. **Discovering other validators**
    1. Retrieve the current set of authorities (currently only consensus, not finality)
    2. Start DHT queries for the ids of the authorities
    3. Validate the signatures of the retrieved key value pairs
    4. Add the retrieved external addresses as ~reserved~ priority nodes to the peerset

The first point *Making itself discoverable* fixes https://github.com/paritytech/substrate/pull/3247.

### Pull request status:
**Work in progress**

To do:

- [ ] Code style, especially function complexity through indentation
- [ ] Error handling / logging
- [ ] Authority key retrieval abstraction (Supporting hot-swappable crypto-primitive changes)
- [ ] Redo commit history (before removing *[WIP]* tag)


arkpar:
This should use `set_priority_group` peerset API

mxinden:
Status: Waiting for https://github.com/paritytech/substrate/pull/3296 as this pull request requires retrieving the current authority key.

core/service/src/components.rs
```diff
@@ -268,12 +280,257 @@ impl<C: Components> OffchainWorker<Self> for C where
...
+					num_active_peers: network.num_active_peers(),
+					average_download_per_sec: network.average_download_per_sec(),
+					average_upload_per_sec: network.average_upload_per_sec(),
+				};
+				let state = network.network_state();
+
+				status_sinks.lock().retain(|sink| sink.unbounded_send((status.clone(), state.clone())).is_ok());
+			}
+
+			let authorities = client.runtime_api().authorities(&BlockId::hash(client.info().chain.best_hash));
```
andresilva: For GRANDPA we need to check the active authority set from the node state and not the runtime, it is tracked [here](https://github.com/paritytech/substrate/blob/master/core/finality-grandpa/src/environment.rs#L375). Maybe we can just pass a `Fn() -> Vec<Authority>` into this function and then build up the authorities externally (babe authorities + grandpa authorities).

mxinden:
Replaced by https://github.com/paritytech/substrate/pull/3452.

PR #3249, opened 2019-07-30
Documentation and small cleanup in panic-handler

tomaka:
A few low-effort changes to `substrate-panic-handler`:

- Renames `AbortGuard::new(false)` to `AbortGuard:force_unwind` and `AbortGuard::new(true)` to `AbortGuard::force_abort` (breaking).
- Makes the implementation details (`set_abort` and `ABORT`) private.
- Adds documentation to the crate.

### DIFF
--- a/core/panic-handler/src/lib.rs
+++ b/core/panic-handler/src/lib.rs
@@ -76,7 +76,7 @@ pub struct AbortGuard {
 
 impl AbortGuard {
 	/// Create a new guard. While the guard is alive, panics that happen in the current thread will
-	/// unwind the stack (unless another guard is is created afterwards).
+	/// unwind the stack (unless another guard is created afterwards).
 	pub fn force_unwind() -> AbortGuard {
 		AbortGuard {
 			previous_val: set_abort(false),
@@ -85,7 +85,7 @@ impl AbortGuard {
 	}
 
 	/// Create a new guard. While the guard is alive, panics that happen in the current thread will
-	/// abort the process (unless another guard is is created afterwards).
+	/// abort the process (unless another guard is created afterwards).
 	pub fn force_abort() -> AbortGuard {
 		AbortGuard {
 			previous_val: set_abort(true),
--- a/core/panic-handler/src/lib.rs
+++ b/core/panic-handler/src/lib.rs
@@ -15,18 +15,31 @@
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
 //! Custom panic hook with bug report link
+//!
+//! This crate provides the [`set`] function, which wraps around [`std::panic::set_hook`] and
+//! sets up a panic hook that prints a backtrace and invites the user to open an issue to the
+//! given URL.
+//!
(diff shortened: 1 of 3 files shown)


core/panic-handler/src/lib.rs
```diff
@@ -39,30 +52,55 @@ This is a bug. Please report it at:
...
+	/// Marker so that `AbortGuard` doesn't implement `Send`.
+	_not_send: PhantomData<std::rc::Rc<()>>
+}
 
 impl AbortGuard {
-	/// Create a new guard and set abort flag to specified value.
-	pub fn new(enable: bool) -> AbortGuard {
-		AbortGuard(set_abort(enable))
+	/// Create a new guard. While the guard is alive, panics that happen in the current thread will
+	/// unwind the stack (unless another guard is is created afterwards).
```
pepyakin:
```suggestion
	/// unwind the stack (unless another guard is created afterwards).
```

core/panic-handler/src/lib.rs
```diff
@@ -39,30 +52,55 @@ This is a bug. Please report it at:
...
+	pub fn force_unwind() -> AbortGuard {
+		AbortGuard {
+			previous_val: set_abort(false),
+			_not_send: PhantomData
+		}
+	}
+
+	/// Create a new guard. While the guard is alive, panics that happen in the current thread will
+	/// abort the process (unless another guard is is created afterwards).
```
pepyakin:
```suggestion
	/// abort the process (unless another guard is is created afterwards).
```
```suggestion
	/// abort the process (unless another guard is created afterwards).
```

core/panic-handler/src/lib.rs
```diff
@@ -39,30 +52,55 @@ This is a bug. Please report it at:
...
+/// Sets a thread-local abort flag on construction and reverts to the previous setting when dropped.
+/// Does not implement `Send` on purpose.
+///
+/// > **Note**: Because we restore the previous value when dropped, you are encouraged to leave
+/// > the `AbortGuard` on the stack and let it destroy itself naturally.
+pub struct AbortGuard {
+	/// Value that was in `ABORT` before we created this guard.
+	previous_val: bool,
+	/// Marker so that `AbortGuard` doesn't implement `Send`.
+	_not_send: PhantomData<std::rc::Rc<()>>
```
pepyakin: If you were looking for a simpler alternative, `*const ()` should [do](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ae4e590a3c4674504a1eef97561747db).

tomaka: I was specifically looking for something that clearly doesn't implement `Send`

pepyakin: > There's unfortunately no way AFAIK to write a test to make sure that it's not Send

Not to say that we should use it here, but:

https://docs.rs/static_assertions/0.3.4/static_assertions/macro.assert_not_impl_all.html

PR #3255, opened 2019-07-30
srml-contracts: Deferred actions

### LINKED ISSUE #2705: Contract ext_dispatch_call refactor
Restoration mechanism has been implemented as an extrinsic call that contract can call using `ext_dispatch_call`. Even though that this extrinsic is technically available to be called by any user, it can only be meaningfully called by a contract (i.e. it bails out if called by an account without code).

We could refactor into the following architecture:

```
enum Delayed {
    DispatchCall(call)
    SendEvent(event)
    RestoreTo(...)
}
```

and merge events and calls fields of ExecutionContext into a delayed field.

CC @pepyakin

pepyakin:
Closes #2705 

This change introduce a concept of deferred action which replaces vectors of events and dispatched calls.

This PR has changes in behavior, specifically:

1. The events are no longer deposited before refund.
1. All deferred actions are interleaved.
1. `restore_to` is no longer a dispatchable call but a function provided by the host (srml-contracts).

The part worth giving attention is that dispatchable calls are interleaved with other actions like `restore_to` and depositing events. However, I couldn't find a problem with this.


srml/contracts/src/wasm/runtime.rs
```diff
@@ -586,6 +586,86 @@ define_env!(Env, <E: Ext>,
...
+	// Attempt to restore the specified contract.
+	//
+	// This will compute a tombstone hash from the caller's storage and the given code hash and
+	// if the hash matches the hash found in the tombstone at the specified address - kill the
+	// caller contract and restore the destination contract and set the specified `rent_allowance`.
+	// All caller's funds are transfered to the destination.
+	//
+	// This function doesn't perform restoration right away but defers it to the end of the
+	// transaction. If there is no tombstone in the destination address or if the hashes don't match
+	// then restoration is cancelled and no changes are made.
```
jimpo: Perhaps make it explicit that the transaction/extrinsic is still applied if the restoration in cancelled. "no changes are made" makes it sound otherwise.

srml/contracts/src/wasm/runtime.rs
```diff
@@ -586,6 +586,86 @@ define_env!(Env, <E: Ext>,
 		Ok(())
 	},
 
+	// Attempt to restore the specified contract.
+	//
+	// This will compute a tombstone hash from the caller's storage and the given code hash and
```
jimpo: Please explain in this comment what "delta" is.

srml/contracts/src/tests.rs
```diff
@@ -1256,17 +1256,24 @@ fn default_rent_allowance_on_create() {
...
+			;; Pointer and length of the encoded code hash buffer
+			(i32.const 264)
+			(i32.const 32)
+			;; Pointer and length of the encoded rent_allowance buffer
+			(i32.const 296)
+			(i32.const 8)
+			;; Pointer and number of items in the delta buffer.
+			;; This buffer specifies multiple keys for removal before restoration.
+			(i32.const 100)
+			(i32.const 1)
```
thiolliere: so you remove the test for if the same key is given two times, but fine.

pepyakin: Yeah, I realized that it won't be so nice : )

srml/contracts/src/wasm/runtime.rs
```diff
@@ -586,6 +586,86 @@ define_env!(Env, <E: Ext>,
...
+			let rent_allowance_buf = read_sandbox_memory(
+				ctx,
+				rent_allowance_ptr,
+				rent_allowance_len
+			)?;
+			BalanceOf::<<E as Ext>::T>::decode(&mut &rent_allowance_buf[..])
+				.ok_or_else(|| sandbox::HostError)?
+		};
+		let delta = {
+			// We don't use `with_capacity` here to not eagerly allocate the user specified amount
```
jimpo: Hmm, why not? Couldn't you just do an up-front bounds check on `delta_ptr` and `delta_ptr + KEY_SIZE * delta_count` then reserve the correct amount of memory?

srml/contracts/src/wasm/runtime.rs
```diff
@@ -586,6 +586,86 @@ define_env!(Env, <E: Ext>,
...
+			// of memory.
+			let mut delta = Vec::new();
+			let mut key_ptr = delta_ptr;
+
+			for _ in 0..delta_count {
+				const KEY_SIZE: usize = 32;
+
+				// Read the delta into the provided buffer and collect it into the buffer.
+				let mut delta_key: StorageKey = [0; KEY_SIZE];
+				read_sandbox_memory_into_buf(ctx, key_ptr, &mut delta_key)?;
```
jimpo: Maybe not a good idea for the sake of API clarity, but if `delta` in `Ext::note_restore_to` was `Vec<[&u8]>` instead of `Vec<StorageKey>` you could read the entire vector in one call to `read_sandbox_memory_into_buf` which might be more efficient.

srml/contracts/src/wasm/runtime.rs
pepyakin: I was more thinking about if the user specifies the count `(usize)-1` the runtime will happily try to allocate this amount of memory and then abort due to OOM. Here we do this incrementally with paying as we go.

Changing this would require an explicit gas charge and and corresponding alterations. The optimization makes sense to me, but this is just something that I don't want to do just yet.

srml/contracts/src/wasm/runtime.rs
pepyakin: `Vec<[&u8]>`? 

But yeah, the same as in https://github.com/paritytech/substrate/pull/3255#discussion_r308773284. I'd like to avoid optimizations like this for the moment preferring more clear API.

srml/contracts/src/exec.rs
```diff
@@ -254,13 +254,28 @@ impl<T: Trait> Token<T> for ExecFeeToken {
 	}
 }
 
+#[cfg_attr(any(feature = "std", test), derive(Debug, PartialEq, Eq, Clone))]
```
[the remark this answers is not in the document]
pepyakin: Hehe, grep for `26925` in the codebase.

Here the bound for derived traits will be (like, `T: Debug`) which doesn't make sense for configuration traits, because what really matters is that used types have these bounds (like, `T::AccountId: Debug`)

srml/contracts/src/wasm/runtime.rs
pepyakin: I guess I know where the confusion comes from. The docs imply that all of this is performed right away whereas all of this is performed at the finalization stage, i.e. after all changes are commited.

srml/contracts/src/lib.rs
```diff
@@ -752,11 +675,92 @@ impl<T: Trait> Module<T> {
...
 					Self::deposit_event(RawEvent::Dispatched(origin, result.is_ok()));
 				}
+				RestoreTo {
+					donor,
+					dest,
+					code_hash,
+					rent_allowance,
+					delta,
+				} => {
+					let _result = Self::restore_to(donor, dest, code_hash, rent_allowance, delta);
```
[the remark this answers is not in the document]
pepyakin: Right... I am really reluctant to add events right away (and TBH I'd like remove some of the current ones). I'd rather not add an event and add it later when we find out that it is needed than add an event which is unnecessary.

jimpo:
Just realized this requires an update to COMPLEXITY.md.

PR #3251, opened 2019-07-30
babe: Disable unused median calculation

andresilva:
The median calculation that is done to figure out at what time the current slot should start (i.e. offset) is currently not properly implemented. Since it is currently unused I have removed the call from verify to it to avoid any issues. I have also fixed an issue that caused it to panic.

Fixes #3248.

PR #3258, opened 2019-07-30
node: update flaming fir chain spec

andresilva:
Updates the flaming fir chain spec so that we can restart the testnet using latest master. The authority ids for Babe were updated since they now refer to sr25519 keys (previously using ed25519 in Aura).

Merge only after #3246 and #3251.


andresilva:
cc @ddorgan

tomaka:
We could also use this opportunity to change the network keys from secp256k1 to ed25519.

andresilva:
@tomaka Apparently we are already using ed25519 network keys. Probably did it last time because `--node-key-type` default changed.

andresilva:
Don't merge this afterwards, I want to rebase this over master to update chain spec (to make sure code storage entry is actually for the latest runtime that will end up in master at this commit).

PR #3259, opened 2019-07-30
Setup the telemetry in substrate-cli instead of substrate-service

tomaka:
The situation right now is:

- We create the `Service`, passing to it a list of telemetry endpoints.
- The `Service` initializes the telemetry, and creates a task that processes said telemetry. This also registers the telemetry as the `slog` global logger. The `Telemetry` object acts as an RAII guard for said registration, and is kept inside the `Service`.
- The node extracts a copy of the telemetry guard from the service and keeps it alive during the service's destruction, so that the logger doesn't get unregistered towards `slog`. If otherwise the telemetry is unregistered, we will get a panic if we send out a telemetry message, which can happen during the service cleanup.
- Then finally the node destroys the telemetry at the very end.

After this PR:

- We initialize the telemetry from `substrate-cli`, before creating the service. The telemetry processing happens in a background thread. You might say: "but Pierre, aren't you trying to remove background threads for your browser thingy". To which I answer: "we won't use substrate-cli from within the browser". Also, theoretically, we would process the telemetry task in the same executor as the service, but doing so requires refactoring half of the code of substrate-cli. This will be the subject of further refactoring.
- We pass to the service a `Receiver`. Whenever a message is sent to it, the service sends out a `"system.connected"` message. I wrote a big chunk of documentation about this `Receiver` in the `Configuration` struct if you want details.
- The `Telemetry` object is kept alive by the background thread in `substrate-cli`.

In details, this PR shuffles a bit the telemetry-related fields of `Configuration`, and removes the `Service::telemetry()` function.

### DIFF
--- a/core/cli/src/lib.rs
+++ b/core/cli/src/lib.rs
@@ -38,7 +38,7 @@ use primitives::H256;
 
 use std::{
 	io::{Write, Read, stdin, stdout, ErrorKind}, iter, fs::{self, File}, net::{Ipv4Addr, SocketAddr},
-	path::{Path, PathBuf}, str::FromStr,
+	path::{Path, PathBuf}, thread, str::FromStr,
 };
 
 use names::{Generator, Name};
@@ -57,7 +57,8 @@ use app_dirs::{AppInfo, AppDataType};
 use log::info;
 use lazy_static::lazy_static;
 
-use futures::Future;
+use futures::{Future, Stream as _, sync::mpsc};
+use futures03::{StreamExt as _, TryStreamExt as _};
 use substrate_telemetry::TelemetryEndpoints;
 
 /// The maximum number of characters for a node name.
@@ -498,13 +499,6 @@ where
 		])
 	}).into();
 
-	// Override telemetry
-	if cli.no_telemetry {
-		config.telemetry_endpoints = None;
-	} else if !cli.telemetry_endpoints.is_empty() {
-		config.telemetry_endpoints = Some(TelemetryEndpoints::new(cli.telemetry_endpoints));
-	}
-
 	// Imply forced authoring on --dev
 	config.force_authoring = cli.shared_params.dev || cli.force_authoring;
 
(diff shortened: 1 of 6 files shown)


tomaka:
I forgot to mention it, but the motivation for this PR is to remove this chunk of code: https://github.com/paritytech/substrate/blob/9db4bd96153f6a0aff65b336d1c4874f3a68f08d/node/cli/src/lib.rs#L237-L239

bkchr:
Hmm, I'm not sure. In my opinion the cli should not start any stuff like telemetry. That should be the task of the `Service`. 
The current `Service` setup is shit, but I hope that after a refactoring we can improve this. Later, people may don't want that their blockchain x sends telemetry data and don't require the telemetry stuff at all. With a modular `Service`, this should be do-able, but not by splitting up the stuff into the CLI.

tomaka:
I can only see two possible directions:

- Either the telemetry uses `log` or `slog`, and then we have to set it up *around* the service.
- Or the telemetry gets refactored to not use `log`/`slog`, and instead gets passed around as a parameter.

I'm find with either, but the current situation is too bad. Everything in substrate right now just assumes that there's a global variable available containing the telemetry.

arkpar:
> The node extracts a copy of the telemetry guard from the service and keeps it alive during the service's destruction, so that the logger doesn't get unregistered towards slog. If otherwise the telemetry is unregistered, we will get a panic if we send out a telemetry message, which can happen during the service cleanup.

Why not just make sure telemetry is destroyed last in the process of service cleanup?
Also, why not make `Service` to manage slog internally, rather than expect the user to set it up?

tomaka:
> Why not just make sure telemetry is destroyed last in the process of service cleanup?

I actually tried that, and it still panics. I'm not sure why.
I think that we still spawn tasks using the tokio runtime (although I'm not sure *where*), and this tokio runtime is owned by the user. Tasks will continue to run for a little bit after the service has cleaned up.

> Also, why not make Service to manage slog internally, rather than expect the user to set it up?

Just like the `log` logger, it's a global variable.
I don't think we should forbid people from using the `slog` library and substrate at the same time.

tomaka:
The telemetry situation really is the same as `log`, as the panic hook, and as the file descriptors limit, which we set up in the CLI:

- https://github.com/paritytech/substrate/blob/2a3b4ccccaf166389d4cc3ac88c00b0f032a01f7/core/cli/src/lib.rs#L201
- https://github.com/paritytech/substrate/blob/2a3b4ccccaf166389d4cc3ac88c00b0f032a01f7/core/cli/src/lib.rs#L219-L220

bkchr:
What about doing it this way:
```
diff --git a/core/telemetry/src/lib.rs b/core/telemetry/src/lib.rs
index 88d515e53..f07234357 100644
--- a/core/telemetry/src/lib.rs
+++ b/core/telemetry/src/lib.rs
@@ -205,6 +205,13 @@ impl Stream for Telemetry {
        }
 }
 
+impl Drop for Telemetry {
+       fn drop(&mut self) {
+               let root = slog::Logger::root(slog::Drain::fuse(slog::Discard), slog::o!());
+               slog_scope::set_global_logger(root).cancel_reset();
+       }
+}
+
 impl slog::Drain for TelemetryDrain {
        type Ok = ();
        type Err = ();
diff --git a/node/cli/src/lib.rs b/node/cli/src/lib.rs
index adf9d81fa..9d3edeaf1 100644
--- a/node/cli/src/lib.rs
+++ b/node/cli/src/lib.rs
@@ -234,10 +234,6 @@ fn run_until_exit<T, C, E>(
        let informant = cli::informant::build(&service);
        runtime.executor().spawn(exit.until(informant).map(|_| ()));
 
-       // we eagerly drop the service so that the internal exit future is fired,
-       // but we need to keep holding a reference to the global telemetry guard
-       let _telemetry = service.telemetry();
-
        let _ = runtime.block_on(service.select(e.into_exit()));
        exit_send.fire();
```

tomaka:
What I'm trying to address is that it's ideologically wrong to have a library set up the logger handling, panic hook, or anything globally accessible.

gavofyork:
> The telemetry situation really is the same as log, as the panic hook,

Unconvinced. Logging and panicking are standard system services, typically expected to be in the global state by general (i.e. non-substrate) crates. For these, there is no alternative but to have them be set up and provided by the application-level code.

Telemetry however, *as it stands with its present adoption*, is a niche utility that the Substrate framework and only the Substrate framework uses/needs. It's reasonable for Substrate library code to manage its lifetime.

If Telemetry ever gets so popular that there is library or application code - *unrelated to Substrate* - that makes use of it, then we can rethink this, but I suspect that is a very long way off.

bkchr:
As this pr seems to be dead, I will close it now.

Issue #3243, opened 2019-07-29
Upgrade schnorrkel


DemiMarie-parity:
It is currently several major versions old.

burdges:
There was a reason for skipping those version in that they were breaking wire format.  We're now done with the audit, and anything else I thought up, so the current wire format is final in 0.7.0.  

An update would break any testnets running, unless someone writes a shim for the verify routines that accepts both.

kianenigma:
related to https://github.com/paritytech/substrate/issues/3245

PR #3261, opened 2019-07-30
Update schnorrkel to `0.7.0`

bkchr:
Closes: https://github.com/paritytech/substrate/issues/3243 https://github.com/paritytech/substrate/issues/3244

# ATTENTION
This breaks the addresses being generated by subkey for sr25519 keys. So, we need people to regenerate their addresses for Kusama (if they used sr25519, which is the default).

CC: @burdges


core/primitives/src/sr25519.rs
```diff
@@ -475,15 +475,20 @@ impl TraitPair for Pair {
...
 			Ok(some_signature) => some_signature,
 			Err(_) => return false
 		};
 		match PublicKey::from_bytes(pubkey.as_ref().as_slice()) {
 			Ok(pk) => pk.verify(
-				signing_context(SIGNING_CTX).bytes(message.as_ref()), &signature
-			),
+				signing_context(SIGNING_CTX).bytes(message.as_ref()),
+				&signature,
+			).map(|_| true).unwrap_or(false),
```
DemiMarie-parity:
```suggestion
			).is_ok(),
```

core/primitives/src/sr25519.rs
```diff
@@ -496,8 +501,9 @@ impl TraitPair for Pair {
 		};
 		match PublicKey::from_bytes(pubkey.as_ref()) {
 			Ok(pk) => pk.verify(
-				signing_context(SIGNING_CTX).bytes(message.as_ref()), &signature
-			),
+				signing_context(SIGNING_CTX).bytes(message.as_ref()),
+				&signature,
+			).map(|_| true).unwrap_or(false),
```
DemiMarie-parity:
```suggestion
			).is_ok(),
```

andresilva:
This PR was eagerly merged. As a rule of thumb if the author asks for a review from a given person and that person has not reviewed yet (and it wasn't too long ago to timeout) then please don't merge it. The changes in this PR are good but the consequences of it should be addressed. As mentioned in the description this breaks the addresses that were gathered for Kusama as well as the ones from people that participated in the claims process. The JS API also doesn't support this, so any transactions from that side will fail.

burdges:
I'm unsure about the best upgrade path, but yes all changes here looks correct.

burdges:
As I said elsewhere, one could employ a shim that verified, which suffices for transitioning BABE probably, but accounts create a much harder problem because we cannot know when people upgrade their components.

DemiMarie-parity:
@andresilva That makes sense.  I was not aware of that rule.

PR #3262, opened 2019-07-30
srml-contracts: Minor fixes to docs.

pepyakin:
Just minor documentation fixes.


srml/contracts/src/wasm/runtime.rs
```diff
@@ -372,17 +371,18 @@ define_env!(Env, <E: Ext>,
 		}
 	},
 
-	// Instantiate a contract with code returned by the specified initializer code.
+	// Instantiate a contract with the specified code hash.
 	//
-	// This function creates an account and executes initializer code. After the execution,
-	// the returned buffer is saved as the code of the created account.
+	// This function creates an account and executes the constructor defined in the code specified
+	// by the code hash. After the execution, the returned buffer is saved as the code of the
```
Robbepop: What is "the returned buffer"? Where is it saved? (unclear to me)

pepyakin: Oh, this is not correct at all! Good spot, removed

PR #3266, opened 2019-07-31
Replace the telemetry Mutex with a channel

### LINKED ISSUE #2823: Telemetry: don't serialize the message before passing it to the background task
At the moment (after #2812), calling `telemetry!` locks a mutex, then serializes the message, and adds it to a `VecDeque`. The telemetry task is then waken up and drains the content of the `VecDeque`.

A better way to do this would be to send the `Record` to a channel, and then make the background task read from that channel, serialize the message, and add it to the `VecDeque`s.

This would remove the `Mutex` in favour of a channel, and remove the need for the `AtomicTask`.

The challenge in this is that the `log` method of `slog::Drain` passes the `Record` by reference, and there's no way to clone it in order to send it to the channel. The `slog_async` crate (which sends records on a channel but in a way that's no suitable for us) actually clones `Record`, but hides this in its internal logic. A good approach to this issue would be to open a PR in either `slog`or `slog_async` to expose the record cloning feature.

tomaka:
Fix #2823 

The context is explained in #2823 
This PR replaces a `Mutex` with a channel. The `Mutex` still exists because of the hack that #3259 fixes, but it is only ever expected to be locked once at a time. We use `try_lock` and print a warning if the mutex happens to already locked.

I went with forking `slog-async` and doing some minor changes while waiting for https://github.com/slog-rs/slog/issues/221.
The changes compared to upstream `slog-async` can be seen here: https://github.com/slog-rs/async/compare/master...paritytech:master

cc @andresilva

### DIFF
--- a/core/telemetry/Cargo.toml
+++ b/core/telemetry/Cargo.toml
@@ -16,6 +16,8 @@ log = "0.4"
 rand = "0.6"
 serde = { version = "1.0.81", features = ["derive"] }
 slog = { version = "^2", features = ["nested-values"] }
+# TODO: we're using slog-async just to be able to clone records; https://github.com/slog-rs/slog/issues/221
+slog-async = { git = "https://github.com/paritytech/slog-async", features = ["nested-values"] }
 slog-json = { version = "^2", features = ["nested-values"] }
 slog-scope = "^4"
 tokio-io = "0.1"
--- a/core/telemetry/Cargo.toml
+++ b/core/telemetry/Cargo.toml
@@ -16,7 +16,8 @@ log = "0.4"
 rand = "0.6"
 serde = { version = "1.0.81", features = ["derive"] }
 slog = { version = "^2", features = ["nested-values"] }
-# TODO: we're using slog-async just to be able to clone records; https://github.com/slog-rs/slog/issues/221
+# TODO: we're using slog-async just to be able to clone records; See https://github.com/slog-rs/slog/issues/221,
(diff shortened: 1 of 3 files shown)


core/telemetry/src/lib.rs
```diff
@@ -210,17 +232,24 @@ impl slog::Drain for TelemetryDrain {
...
+		let before = Instant::now();
+
+		let serialized = slog_async::AsyncRecord::from(record, values);
+		// Note: interestingly, `try_send` requires a `&mut` because it modifies some internal value, while `clone()`
+		// is lock-free.
+		let result = match self.sender.clone().try_send(serialized) {
+			Ok(()) => Ok(()),
+			Err(err) => {
+				warn!(target: "telemetry", "Ignored telemetry message because of error on channel: {:?}", err);
+				Err(())
```
bkchr: I'm not sure we want to do this. `slog` will panic when it sees an error as we use `Drain::fuse`.

An error could be that the channel is full or closed.

core/telemetry/Cargo.toml
```diff
@@ -16,6 +16,8 @@ log = "0.4"
 rand = "0.6"
 serde = { version = "1.0.81", features = ["derive"] }
 slog = { version = "^2", features = ["nested-values"] }
+# TODO: we're using slog-async just to be able to clone records; https://github.com/slog-rs/slog/issues/221
```
bkchr: Maybe add one of your issue numbers for telemetry here as well?

PR #3267, opened 2019-07-31
Update to schnorrkel `0.8.0`

bkchr:
CC @burdges


burdges:
It looks good!  Just one new signature test vector, but that's fine for a new network like kusama.  

We'll need to add more test vectors into schnorkel itself of course.  https://github.com/w3f/schnorrkel/issues/36

Just fyi, https://github.com/w3f/schnorrkel/issues/9 contains my internal debate about using ed25519 style secret key derivation, never a clear decision.

jacogr:
Derives are different, i.e

```
$ cargo run --release -- inspect "//Alice"
    Finished release [optimized] target(s) in 3.14s
     Running `.../target/release/subkey inspect //Alice`
Secret Key URI `//Alice` is account:
  Public key (hex): 0xe2d5259549e73bffe82a6132d0bb4b6ca9fa8d78ef063c4b58d28da7ec319767
  Address (SS58): 5HC7zmvyyESMevkCTJi7YQnPwujJc1d5K5Bya5kJAtxgSQn7

```

core/primitives/src/sr25519.rs
```diff
@@ -618,6 +613,20 @@ mod test {
...
 
+	#[test]
+	fn derive_soft_known_pair_should_work() {
+		let pair = Pair::from_string(&format!("{}/Alice", DEV_PHRASE), None).unwrap();
+		let expected = hex!("d6c71059dbbe9ad2b0ed3f289738b800836eb425544ce694825285b958ca755e");
+		assert_eq!(pair.public().to_raw_vec(), expected);
+	}
+
+	#[test]
+	fn derive_hard_known_pair_should_work() {
```
jacogr: One thing to be aware of (which is biting me on the JS-side). When storing a keypair from 0.1.1 (i.e. the secretKey/publicKey or just the combined 96-bytes) and trying to import that via the new schnorrkel, well, it is not that happy.

For any minisecret it would seem the pair derivation yields the same publicKey, but the secretKey part of the generated bundle is not the same. As an example for `//Alice`

0.1.1 -

```
sk: 28b0ae221c6bb06856b287f60d7ea0d98552ea5a16db16956849aa371db3eb51fd190cce74df356432b410bd64682309d6dedb27c76845daf388557cbac3ca34
pk: 46ebddef8cd9bb167dc30878d7113b7e168e6f0646beffd77d69d39bad76b47a
```

0.8.3 -

```
sk: 05d65584630d16cd4af6d0bec10f34bb504a5dcb62dba2122d49f5a663763d0afd190cce74df356432b410bd64682309d6dedb27c76845daf388557cbac3ca34
pk: 46ebddef8cd9bb167dc30878d7113b7e168e6f0646beffd77d69d39bad76b47a
```

jacogr: There is a way around this which is employed in the polkadot-js/wasm to remain compatible with pairs created/stored on exiting wallets. (Basically using the Keypair from_half_ed25519 and to_half_ed25519_bytes)

If required here, it should be available in 0.8.4 when that is published.

gavofyork:
@burdges @jacogr do we have test vectors for both hard and soft derivation? i don't want people with derived kusama keys (like me :-) ) to end up with different addresses.

jacogr:
@gavofyork I added those since @bkchr suggested them (they were massaged into shape by @kianenigma). The only difference between them and the existing derivation tests are that they actually check the resulting public keys against known, which previously was missing.

If the pair serialization format is important, we should certainly add tests for checking against known for those as well, those are different here to the current 0.1.1 formats. (It is important for wallets, hence the WASM has extended tests for those in addition to the public key parts catered for here)

For the w3f/schnorrkel repo itself, similar tests are now the code as comments, however the test vectors there are WIP.

burdges:
If you use `ExpandionMode::Ed25519` then all derivations should now agree with 0.1.1, and I think @jacogr old test vectors confirm this.  I'd suggest running more tests with a version of subkey using 0.8.4 of course.  ;)

We've a very minor change in the serialization format for expanded secret keys, but the from_*_ed25519_bytes methods do the old method.  A deserializer can test if the public key still agrees to figure out if you're deserializating the old or new format, but any code compiled with 0.1.1 cannot deserialize the new keys. 

I've one change https://github.com/w3f/schnorrkel/commit/38f726eea0145aa55d660d726ec6dc0f361e6e5a in derive.rs since 0.8.4 but it only impacts the nonce of the derived key, and thus cannot impact key or signature validity.  In fact, it makes it possible to write a full test vector including the nonce seed using `attach_rng`.  :)

PR #3272, opened 2019-07-31
Fix telemetry not responding to pings

### LINKED ISSUE #3190: Telemetry websocket connection does not respond to PING
Previously when using the `ws` library, the telemetry connection would respond to ping ( https://tools.ietf.org/html/rfc6455#section-5.5.2 ). Would be great if we can reinstate this functionality.

tomaka:
Fix #3190 

We now read from the telemetry websocket, because the underlying library relies on that to respond to pings.

The change is untested. If someone has an idea how to test if we're responding to pings, please try.

### DIFF
--- a/core/telemetry/src/worker.rs
+++ b/core/telemetry/src/worker.rs
@@ -187,6 +187,25 @@ impl TelemetryWorker {
 /// For some context, we put this object around the `wasm_ext::ExtTransport` in order to make sure
 /// that each telemetry message maps to one single call to `write` in the WASM FFI.
 struct StreamSink<T>(T);
+
+impl<T: tokio_io::AsyncRead> futures01::Stream for StreamSink<T> {
+	type Item = BytesMut;
+	type Error = io::Error;
+
+	fn poll(&mut self) -> futures01::Poll<Option<Self::Item>, Self::Error> {
+		let mut buf = [0; 128];
+		Ok(self.0.poll_read(&mut buf)?
+			.map(|n|
+				if n == 0 {
+					None
+				} else {
+					let buf: BytesMut = buf[..n].into();
+					Some(buf)
+				}
+			))
+	}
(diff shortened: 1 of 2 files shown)


twittner:
> The change is untested. If someone has an idea how to test if we're responding to pings, please try.

The easiest is to clone https://github.com/paritytech/substrate-telemetry/ and issue a `cargo run` within its `backend` folder which starts the telementry backend on 127.0.0.1:8080. Using `ws://127.0.0.1:8080/submit` as a custom telemetry URL will make substrate talk to this backend which periodically sends PINGs. If the remote does not answer the backend will print an error. Also starting substrate with `env RUST_LOG=soketto` will show the frame headers, and should show a `Ping` every now an then. I was able to successfully test your change using this approach.

core/telemetry/src/worker.rs
```diff
@@ -187,6 +187,25 @@ impl TelemetryWorker {
...
+	type Error = io::Error;
+
+	fn poll(&mut self) -> futures01::Poll<Option<Self::Item>, Self::Error> {
+		let mut buf = [0; 128];
+		Ok(self.0.poll_read(&mut buf)?
+			.map(|n|
+				if n == 0 {
+					None
+				} else {
+					let buf: BytesMut = buf[..n].into();
```
twittner: Since `StreamSink` is a private type and we ignore data read from telemetry, maybe just returning `BytesMut::new()` is an option in order to avoid the heap allocation?

tomaka: Well, this codepath should never even happen in the first place, so I don't think it matters.
I get the idea, but I kind of dislike the surprise factor of polling empty buffers from the `Stream`.

twittner: Yeah, I agree.

PR #3277, opened 2019-07-31
Fix panic if a logger has already been initted

Sushisource:
If a user desires instantiating and registering their own logger with
the `log` crate before loading substrate, the substrate cli crate
will panic when attempting to initialize the logger. This commit fixes
that.

I targeted the 1.0 branch since that's what I'm using, and this seems
really easy to forward-port. Lemme know if there's an issue with that.

Thanks for your hard work on Substrate!

### DIFF
--- a/core/cli/src/lib.rs
+++ b/core/cli/src/lib.rs
@@ -57,7 +57,7 @@ pub use params::{NoCustom, CoreParams};
 pub use traits::{GetLogFilter, AugmentClap};
 use app_dirs::{AppInfo, AppDataType};
 use error_chain::bail;
-use log::info;
+use log::{info, SetLoggerError};
 use lazy_static::lazy_static;
 
 use futures::Future;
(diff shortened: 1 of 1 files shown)


Sushisource:
[clabot:check]

tomaka:
Oh I didn't realize that this was against the 1.0 branch. Did you do that on purpose, @Sushisource? We're not really maintaining v1.0 anymore.

Sushisource:
> Oh I didn't realize that this was against the 1.0 branch. Did you do that on purpose, @Sushisource? We're not really maintaining v1.0 anymore.

Yeah, I did that on purpose (mentioned in OP). Is there a reasonably stable branch I should be basing on? There's not much around that anywhere I can find in the docs or website. Guidance there I'm sure would be very welcome by more than just me :smile:

gavofyork:
Yeah - a forward port to master would be useful, thanks!

Issue #3036, opened 2019-07-05
Remove usage of AssertUnwindSafe


tomaka:
`AssertUnwindSafe` basically means "we accept that this object might be in a corrupted state". It's quite a bad idea.

We should either things isolate more, so that the object can be thrown away entirely when it's bad, or make the objects *actually* unwind-safe.

tomaka:
After discussion, this should be solved by removing catching panics altogether. Panicking just shouldn't happen.

expenses:
The only remaining usage of `AssertUnwindSafe` is in `TelemetryDrain`, and I don't believe we can remove that unless `futures::channel::mspc::Sender` becomes `UnwindSafe`.

bkchr:
> After discussion, this should be solved by removing catching panics altogether. Panicking just shouldn't happen.

Hmm, not sure if that will be possible. This would require that every runtime developed with Substrate never panics. I think that is something we can never achieve. Nevertheless, I support the initial request of make the object unwind-safe.

tomaka:
> Panicking just shouldn't happen.

This was originally referring to some `catch_unwind`s that were present in Aura and Babe. They have then been removed in #3281 

Runtime code is of course different.

bkchr:
Ahh okay, yeah that makes sense.

ggwpez:
It us used in a few spots. GitHub search dos not show it up.

PR #3281, opened 2019-08-01
Slots should not try to catch panics.

montekki:
The loop that tries to recover after panics and errors in `on_slot` is not correct since it tries to call `poll` on `TryForEeach` multiple times which leads to a panic bc some `Select` inside futures is polled twice.

After some discussions with @tomaka it was decided to remove the catching logic at all.


tomaka:
cc'ing #3036 for bookkeeping

PR #3284, opened 2019-08-01
Fix staking

gavofyork:
`total` (which includes the amounts that are being unstaked) is wrong when evaluating stakes of validators/nominators for Phragmen. Rather should be `active`.


gavofyork:
@kianenigma could you add a regression test here?

PR #3293, opened 2019-08-02
node: use BABE predigest data to find authorship

### LINKED ISSUE #2986: Migrate Aura authorship-detection to Pre-runtime digests

andresilva:
We already had all the infrastructure in place to collect authorship data from BABE predigests but it wasn't wired in on the node runtime.

Fixes #2986.

### DIFF
--- a/core/consensus/babe/primitives/src/lib.rs
+++ b/core/consensus/babe/primitives/src/lib.rs
@@ -53,7 +53,7 @@ pub const VRF_PROOF_LENGTH: usize = 64;
 pub const PUBLIC_KEY_LENGTH: usize = 32;
 
 /// The index of an authority.
-pub type AuthorityIndex = u64;
+pub type AuthorityIndex = u32;
 
 /// A slot number.
 pub type SlotNumber = u64;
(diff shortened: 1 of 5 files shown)


andresilva:
I changed the `FindAuthor` session adapter to use a `u64` for authority index because that's what BABE uses. But maybe it's more reasonable to instead have BABE use a `u32` instead.

rphmeier:
> I changed the `FindAuthor` session adapter to use a `u64` for authority index because that's what BABE uses. But maybe it's more reasonable to instead have BABE use a `u32` instead.

We've bounded the length of authority lists by u32 elsewhere, so I don't think this would be controversial.

andresilva:
Updated BABE to use `u32` for authority index, this breaks sync with FF but there are more breaking changes on the pipeline and we will reset FF shortly anyway.

PR #3296, opened 2019-08-03
Refactor key management

gavofyork:
Supersedes #3213 

This PR is turning into a major refactoring over the crypto and key management system, which seems both long overdue and somewhat inescapable.

I'll want to integrate the general changes in (but not merge) #3034 and possibly #3193 together with #3137.

The main issues that this PR aims to solve are:
- Independent subsystems that happened to use the same crypto had to share the same session keys (and key-store division).
- Assumption that the twin consensus systems of block production (Aura/Babe) and finality (Grandpa) each required keys (which dirtied a lot of Substrate, placing APIs like `fn fg_authority_key()` all the way through the codebase). Though they do now, they won't forever and it shouldn't be an assumption that we bake in to internal APIs.
- Keys passed in through CLI are sticky; their existence enabled the consensus system workers and force them to use these keys alone. Rather whatever key in the key store that happens to be an authority should be used at any given time.

TODO:
- Refactor `ImOnline`:
  - [x] Should get its own key and not piggy-back on other session keys.

- Dynamic key selection:
  - [x] Enable Babe and Grandpa to be able to determine current authority key (if there is one).
  - [ ] Dynamically select the correct authoring key to use in each subsystem.
  - [ ] Use `--validator` to enable active validation process; remove any validator key-management stuff from the CLI (`--key`). (A second `--worker` arg can enable off-chain workers.)
  - [ ] Remove all static-key selection code (`--key` all the way down to key-choice).
  - [ ] Remove `authority_key`/`fg_authority_key` APIs completely.
  - [ ] Make `--password` work (should accept a manually-entered password and use it for everything in the keystore).
  - [ ] Introduce RPCs for session rotation, using code from #3034 if it makes sense. There should be a single RPC `author_rotateKeys` which creates a new set of session keys, saves them to the key-store and returns their public keys, so that a rotation-transaction can be constructed. There should be no pollution of the APIs with session keys: the rotation API should be wholly independent of each of the individual keys. This may mean new APIs in the `OpaqueKeys` trait.

- Refactor `Client`:
  - Create a basic `Client` trait that provides all the basic APIs and has `Block` as an associated item. *Probably a separate PR.*

I would also like to solve the fact that service construction is highly monolithic and polluted by consensus-module-specific code. Really it should just be a case of specifying the block production and finality systems and letting it do the rest. Code for set-up, take-down, RPC and key-store integration should be provided through traits that the consensus modules (Babe, Aura, Grandpa and eventually Rho and PoW) implement *within their module's scope*. There shouldn't be any consensus-module-specific logic dirtying up `service.rs`.

- Refactor `service`:
  - Move out any consensus-module-specific code, and hardwired assumptions that they each have "authority keys", from `service.rs` into the respective consensus modules. Provide a single API endpoint (i.e. a type that implements some `ConsensusSystem` trait and have that as the only resource that service needs know of.  *Probably a separate PR.*

This also adds:
- Add Call type to extensible transactions.
- Add `swap` to allow two storage items in the same map to be swapped.


rphmeier:
> I would also like to solve the fact that service construction is highly monolithic and polluted by consensus-module-specific code. Really it should just be a case of specifying the block production and finality systems and letting it do the rest. Code for set-up, take-down, RPC and key-store integration should be provided through traits that the consensus modules (Babe, Aura, Grandpa and eventually Rho and PoW) implement within their module's scope. There shouldn't be any consensus-module-specific logic dirtying up service.rs.

Let's defer this to another PR - this isn't an API decision that should be rushed.
My ideal solution would just be to encapsulate everything within a `fn initialize` exposed by every crate by convention. If a trait is really necessary,


```rust
trait Initialize {
    type Args;
    type Background: Future<Item = ()> + Send + Sync + 'static;

    fn initialize(args: Args) -> Self::Background;
}
```

gavofyork:
@rphmeier Yup I'm happy with this model.

core/client/src/client.rs
```diff
@@ -260,6 +253,7 @@ impl<H> PrePostHeader<H> {
 pub fn new_in_mem<E, Block, S, RA>(
 	executor: E,
 	genesis_storage: S,
+	keystore: Option<primitives::traits::KeyStorePtr>,
```
rphmeier: is there a strong case for putting this in the client? 
a god-object client gave us a lot of trouble in parity-ethereum; we have an opportunity here to avoid making the same mistake.

rphmeier: for instance, why can the keystore not be encapsulated as part of the `E` parameter?

core/consensus/babe/src/lib.rs
```diff
@@ -1190,6 +1192,117 @@ pub fn import_queue<B, E, Block: BlockT<Hash=H256>, I, RA, PRA>(
...
+	const SEALER_TYPE: &'static str = "Key";
+	fn format(&self) -> String { use primitives::crypto::Ss58Codec; self.to_ss58check() }
+}
+
+/// A runtime service for a consensus worker.
+///
+/// This is administered by the client service to help it spin up everything necessary. It can
+/// provide API hooks for handling user-level/RPC events. It can return information regarding
+/// the status of the service.
+pub trait Service<'a>: 'a {
```
rphmeier: this and related stubs should be removed before merging - an issue with the pseudocode to go over options can contain it instead

core/executor/src/wasm_executor.rs
```diff
@@ -657,7 +661,69 @@ impl_function_executor!(this: FunctionExecutor<'e, E>,
...
+		let key_type = KeyTypeId(id);
+
+		let mut pubkey = [0u8; 32];
+		this.memory.get_into(pubkey_data, &mut pubkey[..])
+			.map_err(|_| "Invalid attempt to get pubkey in ext_ed25519_sign")?;
+
+		let msg = this.memory.get(msg_data, msg_len as usize)
+			.map_err(|_| "Invalid attempt to get message in ext_ed25519_sign")?;
+
+		let signature = runtime_io::ed25519_sign(key_type, &pubkey, &msg);
```
rphmeier: this seems as though it could easily break consensus

core/finality-grandpa/src/lib.rs
```diff
@@ -777,3 +781,41 @@ pub fn run_grandpa<B, E, Block: BlockT<Hash=H256>, N, RA, SC, X>(
...
+/// Provide the authority key, if any, that is controlled by this node as of the given block.
+fn authority<C>(client: &C, keystore: Arc<Store>) -> Option<AuthorityPair> where
+	C: ProvideRuntimeApi + BlockOf + ProvideCache<<C as BlockOf>::Type>
+		+ HeaderBackend<<C as BlockOf>::Type>,
+	C::Api: GrandpaApi<<C as BlockOf>::Type>
+{
+	let owned = keystore.public_keys::<AuthorityId>().ok()?;
+	let at = BlockId::Number(client.info().finalized_number);
+	// The list of authority keys that is current. By default this will just use the state of
+	// the best block, but you might want it to use some other block's state instead if it's
```
rphmeier: copypasta?

core/application-crypto/src/ed25519.rs
```diff
@@ -0,0 +1,75 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Sr25519 crypto types.
```
gavofyork:
```suggestion
//! Ed25519 crypto types.
```

core/application-crypto/src/ed25519.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use crate::{RuntimePublic, KeyTypeId};
+
+pub use primitives::ed25519::*;
+
+mod app {
+	use crate::key_types::ED25519;
+	crate::app_crypto!(super, ED25519);
```
gavofyork: i'l like to remove this entirely in the not too distant future. is anything using it?

core/application-crypto/src/sr25519.rs
```diff
@@ -0,0 +1,75 @@
...
+
+//! Sr25519 crypto types.
+
+use crate::{RuntimePublic, KeyTypeId};
+
+pub use primitives::sr25519::*;
+
+mod app {
+	use crate::key_types::SR25519;
+	crate::app_crypto!(super, SR25519);
```
gavofyork: Also this file.

core/application-crypto/src/traits.rs
```diff
@@ -0,0 +1,114 @@
...
+	///
+	/// Returns the signature or `None` if the private key could not be found or some other error
+	/// occurred.
+	fn sign<M: AsRef<[u8]>>(&self, key_type: KeyTypeId, msg: &M) -> Option<Self::Signature>;
+
+	/// Verify that the given signature matches the given message using this public key.
+	fn verify<M: AsRef<[u8]>>(&self, msg: &M, signature: &Self::Signature) -> bool;
+}
+
+/// A runtime interface for a application's public key.
```
gavofyork:
```suggestion
/// A runtime interface for an application's public key.
```

core/consensus/babe/src/lib.rs
```diff
@@ -1190,6 +1191,24 @@ pub fn import_queue<B, E, Block: BlockT<Hash=H256>, I, RA, PRA>(
 	Ok((queue, timestamp_core, block_import, pruning_task))
 }
 
+/// Type of source for block sealing. Different consensus algorithms have different sealing
```
gavofyork: TODO: move this to `consensus_common`.

core/application-crypto/src/ed25519.rs
bkchr: I use it in some tests. But yeah, I could probably do this in the test itself.

core/client/src/client.rs
bkchr: It is not part of the client.

core/executor/src/wasm_executor.rs
gavofyork: i think it can only be called from the offchain worker. if not, then it needs to change.

core/network/src/config.rs
```diff
@@ -103,6 +103,11 @@ impl Roles {
 		self.intersects(Roles::FULL | Roles::AUTHORITY)
 	}
 
+	/// Does this role represents a client that does not participates in the consensus?
```
gavofyork: Avoid questions in docs. But also, what does this mean? There's now no longer any reason that a full node with valid session keys for an authority should not be a validator. So if there is to be a CLI argument for this it should probably be `--force-disable-validating` to reflect the fact the full nodes should, by default, try to validate when they can.

core/offchain/src/lib.rs
```diff
@@ -250,11 +186,11 @@ mod tests {
 		let client = Arc::new(test_client::new());
 		let pool = Arc::new(Pool::new(Default::default(), ::transaction_pool::ChainApi::new(client.clone())));
 		let db = client_db::offchain::LocalStorage::new_test();
-		let mock = Arc::new(MockNetworkStateInfo());
-
+		let network_state = Arc::new(MockNetworkStateInfo());
+		// TODO Test keystore
```
gavofyork: TODO in code.

core/primitives/src/ed25519.rs
```diff
@@ -488,17 +479,22 @@ impl Pair {
 	}
 }
 
-impl TypedKey for Public {
-	const KEY_TYPE: KeyTypeId = key_types::ED25519;
+impl CryptoType for Public {
+	const KIND: Kind = Kind::Ed25519;
```
gavofyork: would have been nice to get rid of `KIND` altogether.

bkchr: I think we can do this now.

core/sr-io/src/lib.rs
```diff
@@ -245,32 +275,28 @@ export_api! {
...
 
-		/// Returns the currently configured authority public key, if available.
-		fn pubkey(key: CryptoKey) -> Result<Vec<u8>, ()>;
-
-		/// Create new key(pair) for signing/encryption/decryption.
+		/// Create new key pair for signing/encryption/decryption.
 		///
 		/// Returns an error if given crypto kind is not supported.
-		fn new_crypto_key(crypto: CryptoKind) -> Result<CryptoKey, ()>;
+		fn new_key(crypto: CryptoKind, app_id: KeyTypeId) -> Result<CryptoKey, ()>;
```
gavofyork: why is this still here?

now that we have `sr25519_generate` and `ed25519_generate`, it shouldn't be needed, right?

bkchr: Yes, the offchain stuff still needs a little bit love by me.

core/sr-io/src/lib.rs
```diff
@@ -200,11 +200,41 @@ export_api! {
 
 export_api! {
 	pub(crate) trait CryptoApi {
-		/// Verify a ed25519 signature.
-		fn ed25519_verify<P: AsRef<[u8]>>(sig: &[u8; 64], msg: &[u8], pubkey: P) -> bool;
+		/// Generate an ed22519 key for the given key type and store it in the keystore.
+		///
+		/// Returns the raw public key.
+		fn ed25519_generate(id: KeyTypeId, seed: Option<&str>) -> [u8; 32];
```
gavofyork: this is non-deterministic so it should certainly not be in the standard API. it should be offchain only.

bkchr: That is correct, we just do not set the keystore and then these functions are not usable from the runtime.

core/executor/src/wasm_executor.rs
tomusdrw: We have at least 3 different "capabilities" for runtime calls that affect the set of callable APIs:
1. Offchain extensions available (offchain workers)
2. Keystore available (offchain workers, rotate session, equivocation report generation)
3. Regular runtime calls

So regular, consensus-critical execution path will neither have keystore nor offchain extensions set so those APIs will be disabled (or will just always return an error).

The offchain and keystore CryptoApis are now a bit duplicated, but afaik @bkchr is working on unification (@bkchr let me know if you need any help).

bkchr: Yeah, the `keystore` argument is optional and will be `None` for block production/block importing. Thus, these apis will not work.

After fixing the tests again, I will start to clean this stuff up.

core/client/src/client.rs
rphmeier: yes it is! it goes into the `LocalCallExecutor` which is part of the client.

srml/membership/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! # Membership Module
+//!
+//! Allows control of membership of a set of `AccountId`s, useful for managing membership of of a
```
cmichi:
```suggestion
//! Allows control of membership of a set of `AccountId`s, useful for
```

srml/membership/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+//! collective.
```
cmichi:
```suggestion
//! managing membership of a collective.
```

core/consensus/aura/src/lib.rs
```diff
@@ -657,6 +660,68 @@ fn authorities<A, B, C>(client: &C, at: &BlockId<B>) -> Result<Vec<A>, Consensus
 		.ok_or_else(|| consensus_common::Error::InvalidAuthoritiesSet.into())
 }
 
+/// Means of determining a list of authority keys, generally used to determine if we are an
+/// authority by cross-referencing this list against the keys in the store and finding one that
+/// appears in both.
+///
+/// This is implemented directly on the `AppPair` key type.
+pub trait AuthorityProvider: AppPair {
```
gavofyork: Do we actually use this anywhere? Will we? (I guess not as long as we support being a multi-authority in Aura like in Babe, Grandpa and ImOnline?)

core/finality-grandpa/src/environment.rs
```diff
@@ -500,15 +500,25 @@ where
 		let prevote_timer = Delay::new(now + self.config.gossip_duration * 2);
 		let precommit_timer = Delay::new(now + self.config.gossip_duration * 4);
 
-		let local_key = self.config.local_key.as_ref()
-			.filter(|pair| self.voters.contains_key(&pair.public().into()));
+		let local_key = crate::is_voter(&self.voters, &self.config.keystore);
```
gavofyork: should return an iterator, so that we can vote with multiple keys if needed. @folsen if you're at a loose end, maybe you can do this (in a new branch :-))?

core/finality-grandpa/src/lib.rs
```diff
@@ -777,3 +750,17 @@ pub fn run_grandpa<B, E, Block: BlockT<Hash=H256>, N, RA, SC, X>(
 {
 	run_grandpa_voter(grandpa_params)
 }
+
+/// Checks if this node is a voter in the given voter set.
+///
+/// Returns the key pair of the node that is being used in the current voter set or `None`.
+fn is_voter(
```
gavofyork: should return an iterator of all keys that we own and that are authorities, rather than just the first one. cc @folsen

core/consensus/aura/src/lib.rs
bkchr: No, but I also did not fixed `Aura` yet. Will come in a later pr.

core/session/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+	client::Client<B, E, Block, RA>: ProvideRuntimeApi,
+	<client::Client<B, E, Block, RA> as ProvideRuntimeApi>::Api: SessionKeys<Block>,
+{
+	let info = client.info().chain;
+	let runtime_api = client.runtime_api();
+
+	for seed in seeds {
+		let seed = seed.as_bytes();
+
+		// We need to generate the keys for the best block + the last finalized block.
```
gavofyork: so we have two sets of keys? what's the point in this? surely we will only want to generate a single set of keys?

node/cli/src/service.rs
```diff
@@ -89,8 +85,6 @@ impl<F> Default for NodeConfig<F> where F: substrate_service::ServiceFactory {
 construct_service_factory! {
 	struct Factory {
 		Block = Block,
-		ConsensusPair = BabePair,
```
gavofyork: 🌈☀️🚀🎉

node/cli/src/service.rs
```diff
@@ -119,74 +113,59 @@ construct_service_factory! {
...
-							config: config,
-							link: link_half,
-							network: service.network(),
-							inherent_data_providers: service.config.custom.inherent_data_providers.clone(),
-							on_exit: service.on_exit(),
-							telemetry_on_connect: Some(telemetry_on_connect),
-						};
-						service.spawn_task(Box::new(grandpa::run_grandpa_voter(grandpa_config)?));
-					},
+				if service.config.roles.is_authority() {
```
gavofyork: this shouldn't be needed any more, right?

gavofyork:
Reviewed it all (though much of it is my code anyway). Looks good except for the comments.

core/consensus/aura/src/lib.rs
gavofyork: is there an issue for the rest of the PRs?

core/finality-grandpa/src/lib.rs
folsen: Filing as issue here from looking at it: https://github.com/paritytech/substrate/issues/3324

PR #3300, opened 2019-08-05
Staking optimisation

thiolliere:
DONE:
* rename add_reward_points_to_validator -> add_reward_points
* modify add_reward_points to get a vec as parameter for optimisation
* add add_reward_points_using_index which does the same but with the validator index instead of validator id.

For validity statements and even for uncle inclusion we want to reward multiple potentially duplicated set of validator. This function handle it better as we don't decode the elector set for each reward now.


thiolliere:
Hmm I think I should extend with a method that only takes index instead of full id.

srml/staking/src/lib.rs
```diff
@@ -325,6 +325,17 @@ pub struct EraRewards {
 	rewards: Vec<u32>,
 }
 
+impl EraRewards {
+	/// add the reward to the validator at index given. Index must valid.
```
kianenigma:
```suggestion
	/// add the reward to the validator at the given valid `index`.
	///
	/// Invalid index will be ignored.
```

srml/staking/src/lib.rs
```diff
@@ -325,6 +325,17 @@ pub struct EraRewards {
...
+	fn add_points_to_index(&mut self, index: u32, points: u32) {
+		if let Some(new_total) = self.total.checked_add(points) {
+			self.total = new_total;
+			self.rewards.resize((index as usize + 1).max(self.rewards.len()), 0);
```
kianenigma: From your doc above I assumed (without looking at the code) that invalid indexes will be ignored. Looks like that is not the case and actually, the function __assumes that `index` is always valid__.

I haven't reviewed the caller part of this function but looking at it from the callee site, it is a bit strange. I assumed an out-of-bound index to return error or ignore the call rather than just expanding the vactor of rewards.

srml/staking/src/lib.rs
```diff
@@ -1399,22 +1410,42 @@ impl<T: Trait> Module<T> {
 		}
 	}
 
-	/// Add reward points to validator.
+	/// Add reward points to validators. Using their stash account ID.
```
kianenigma:
```suggestion
	/// Add reward points to validators using their stash account ID.
```

srml/staking/src/lib.rs
```diff
@@ -1399,22 +1410,42 @@ impl<T: Trait> Module<T> {
...
+			}
+		});
+	}
+
+	/// Add reward points to validators. Using their validator index.
+	///
+	/// For each element in the iterator the given number of points in u32 is added to the
+	/// validator, thus duplicates are handled.
+	pub fn add_reward_points_using_index(validators_points: Vec<(u32, u32)>) {
+		// TODO: This can be optimised once #3302 is implemented.
```
kianenigma: indeed. Just note my comment in the issue.

srml/staking/src/tests.rs
```diff
@@ -357,12 +357,12 @@ fn rewards_should_work() {
 		Session::on_initialize(System::block_number());
 		assert_eq!(Staking::current_era(), 0);
 		assert_eq!(Session::current_index(), 1);
-		<Module<Test>>::add_reward_points_to_validator(11, 50);
-		<Module<Test>>::add_reward_points_to_validator(11, 50);
+		<Module<Test>>::add_reward_points(vec![(11, 50)]);
```
kianenigma: can't we replace this (and such cases where a test needs to reward everyone) with your `add_reward_points_to_all_elected `?

srml/staking/src/lib.rs
```diff
@@ -1399,22 +1410,42 @@ impl<T: Trait> Module<T> {
...
-			.map(|index| {
-				CurrentEraRewards::mutate(|rewards| {
-					if let Some(new_total) = rewards.total.checked_add(points) {
-						rewards.total = new_total;
-						rewards.rewards.resize((index + 1).max(rewards.rewards.len()), 0);
-						rewards.rewards[index] += points; // Addition is less than total
+	pub fn add_reward_points(validators_points: Vec<(T::AccountId, u32)>) {
+		CurrentEraRewards::mutate(|rewards| {
+			for (index, elected) in <Module<T>>::current_elected().iter().enumerate() {
+				for (validator, points) in &validators_points {
```
kianenigma: Might be worth looking into in order to remove the inner loop:
- If most modules are storing the indexes nowadays, if authorship can tell you reward validator at index `I` (instead of validator with address `A`) that would be nice.

srml/staking/src/lib.rs
thiolliere: hmm yes and no you give an index that is beyond the size of elected vec then its points will be discarded but not distributed to other thus it is not a no-op. at some point making this a no-op can be done in new_era.

srml/staking/src/lib.rs
thiolliere: yes it is not ignored. the caller part take care of it by checking that elected_set is coherent with the given index. but at some point with decode_len the check could be made here. though that is less efficient.

Or we could do a guarantee like beyond a too big value it is discarded but otherwise it gets inserted and user of this struct should be careful

srml/staking/src/tests.rs
thiolliere: yes but this test test more than just that it test that you can reward two times the same and that a not elected one is no-op

srml/staking/src/lib.rs
thiolliere: also I thought that if current elected is sorted we could do some binary_search.

in polkadot when rewarding validity statement we have indices

thiolliere: oh I missunderstood, if you have index then use the other method add_reward_points_using_index

srml/staking/src/lib.rs
kianenigma: I see, then no need for `/// Invalid index will be ignored.`

srml/staking/src/lib.rs
kianenigma: fair enough; I assume it is more intuitive for the callee to check the conditions rather than assuming that the caller takes care of it, but it looks like you have some optimisation gain in doing so.

srml/staking/src/lib.rs
```diff
@@ -1399,22 +1411,42 @@ impl<T: Trait> Module<T> {
...
+	///
+	/// For each element in the iterator the given number of points in u32 is added to the
+	/// validator, thus duplicates are handled.
+	pub fn add_reward_points_using_index(validators_points: Vec<(u32, u32)>) {
+		// TODO: This can be optimised once #3302 is implemented.
+		let current_elected_len = <Module<T>>::current_elected().len() as u32;
+
+		CurrentEraRewards::mutate(|rewards| {
+			for (validator_index, points) in validators_points {
+				if validator_index < current_elected_len {
```
kianenigma: (I see the check here now).

srml/staking/src/lib.rs
```diff
@@ -1399,22 +1411,42 @@ impl<T: Trait> Module<T> {
...
-	pub fn add_reward_points_to_validator(validator: T::AccountId, points: u32) {
-		<Module<T>>::current_elected().iter()
-			.position(|elected| *elected == validator)
-			.map(|index| {
-				CurrentEraRewards::mutate(|rewards| {
-					if let Some(new_total) = rewards.total.checked_add(points) {
-						rewards.total = new_total;
-						rewards.rewards.resize((index + 1).max(rewards.rewards.len()), 0);
-						rewards.rewards[index] += points; // Addition is less than total
+	pub fn add_reward_points(validators_points: Vec<(T::AccountId, u32)>) {
```
rphmeier: why does it have to be a `Vec` and force allocation (expensive in runtime)? it should be a slice or better yet, an `IntoIterator`.

srml/staking/src/lib.rs
```diff
@@ -1399,22 +1411,42 @@ impl<T: Trait> Module<T> {
...
+				}
+			}
+		});
+	}
+
+	/// Add reward points to validators using their validator index.
+	///
+	/// For each element in the iterator the given number of points in u32 is added to the
+	/// validator, thus duplicates are handled.
+	pub fn add_reward_points_using_index(validators_points: Vec<(u32, u32)>) {
```
rphmeier: why vec?

rphmeier: could we rename to `reward_by_indices`?

srml/staking/src/lib.rs
rphmeier: rename to `reward_by_ids`?

srml/staking/src/lib.rs
thiolliere: renamed and put intoiterator

srml/staking/src/lib.rs
thiolliere: renamed and put intoiterator.

now we iterate on validators to reward first and inside this loop we iterate on current_elected.

thiolliere: Also if this method is used for rewarding a lots of ppl at the same time we should do a sort of current_elected and then a binary_search.

hmm I should state that in the doc

PR #3301, opened 2019-08-05
Fix light client synchronization on master

### LINKED ISSUE #3289: Light client fails to sync flaming fir
For BABE chains light client should have an access to the `Epoch` data to validate blocks. Given that genesis block is available on light client, we know the epoch#0 data => client verifies epoch#0 blocks successfully. Next epoch#1 (on current Flaming Fir) starts from block#55 => block#54 contains a digest that has a data for epoch#2 (that's the main diff with Aura, where we had a digests that were applied starting from the next block). But apparently we're missing data for epoch#1 here.

I've chatted with @andresilva (thanks for your help) && looks like there's no way to build epoch#1 data using genesis block (which is available locally on light) => the only way is to fetch it from remote node. Like we do with finality proofs. This, however, differs from finality proofs because there we're still able to import block && here import will fail. So the solution is: (1) either pause sync process until response is received or (2) allow import of block even though it can be invalid.

svyatonik:
depends on #3413 

closes #3289 (please ignore comments there, because it is fixed in other, more appropriate way)
**WARNING**: it won't sync with FF, because it's genesis block is build in the incompatible (with this implementation) way. If we need a testnet that supports light clients, let's restart FF.

So it turned out that validators we're using at session0 and session1 are always the same. And we could use the same set of BABE authorities in both epoch#0 and epoch#1 (thanks to @andresilva for mentioning that). So the fix is simply to use the same data (except for epoch index) for first two epochs. But then there have been several issues and controversial decisions (in my impl) that I wanted to mention:
1) in current implementation the set of authorities at epoch0 and epoch1 is the same, but order differs. I.e. at epoch0 it is `[Authority0, Authority1]`, but at epoch1 it is `[Authority1, Authority0]`. That is because session module mutates validators passed using config [here](https://github.com/paritytech/substrate/blob/9ede42b1d21e16265172d647eab5146e308e6436/srml/session/src/lib.rs#L335). Since this mutation could also mutate storage [like here](https://github.com/paritytech/substrate/blob/9ede42b1d21e16265172d647eab5146e308e6436/srml/staking/src/lib.rs#L1295), it is impossible to use it twice to reproduce the same mutation in the babe module (again - thanks to Andre for mentioning that). The decision is to use temporary in-memory storage as a way to pass mutated set of authorities from session module to babe module. This leads to three other issues that I've ran into (2, 3, 4);
2) in current implementation babe module is initialized before session module => we haven't yet computed mutated set when babe is initialized. The solution is to change order of initialization [here](https://github.com/paritytech/substrate/blob/9ede42b1d21e16265172d647eab5146e308e6436/node/runtime/src/lib.rs#L402). But then we have another issue (3);
3) the order of modules initialization also determines the order of their `on_initialize()` calls. And since session module from its `on_initialize()` calls BABE' `on_initialize()` AND session initialization now happens before BABE => BABE was incorrectly determined epoch boundaries. The solution was to explicitly call BABE' `on_initialize()` from `should_end_session()`. So now it could be called twice => I've added additional check so that it won't mutate the same data several times;
4) I had to add `OpaqueKeys` associated type to the babe module to be able to decode mutated set of session validators at the genesis block. The problem is that session doesn't know who's going to use that mutated set => it should encode all the available keys;
5) core babe code now holds new enum `MaybeSpanEpoch` instead of simply `Epoch` in the consensus cache. There's special `Genesis` entry in this enum that holds data for two epochs (epoch0 and epoch1). When we're dealing with this entry AND we have failed header verification using epoch0, then we retry with epoch1. This way we allow epoch0 headers verification on light clients;
6) I've also removed insertion to `::AUTHORITIES` cache from Babe, because it seems unnecessary. It was only used to inform Grandpa that it should emit/expect justification for imported block. But imo Babe epoch change isn't that different to Aura authorities change => Grandpa will consider if justification required for the block by checking if anything has been inserted into consensus cache (in Aura it'll be new authorities set, in Babe - new epoch data).


srml/babe/src/lib.rs
```diff
@@ -251,6 +291,34 @@ impl<T: Trait> Module<T> {
...
+				RawBabePreDigest::decode(&mut data)
+			} else {
+				None
+			})
+		{
+			// since do_initialize can be called twice (if session module is present)
+			// => let's ensure that we only modify the storage once per block by checking
+			// current slot number
+			let current_slot = CurrentSlot::get();
+			if current_slot != digest.slot_number {
```
svyatonik: Could you, please, verify that this is correct? If there could be two blocks with the same `digest.slot_number`, this condition should be fixed.

core/consensus/babe/src/lib.rs
```diff
@@ -1043,6 +1120,16 @@ impl<B, E, Block, I, RA, PRA> BlockImport<Block> for BabeBlockImport<B, E, Block
 		// this way we can revert it if there's any error
 		let mut old_epoch_changes = None;
 
+		if let Some(enacted_epoch) = enacted_epoch.as_ref() {
```
svyatonik: I was not sure - why cache has been updated only when both `next_epoch` and `enacted_epoch` are `Some()`. So reorganized code a bit

srml/babe/src/lib.rs
```diff
@@ -152,6 +158,52 @@ decl_storage! {
...
+		config(authorities): Vec<(AuthorityId, BabeWeight)>;
+		build(|
+			storage: &mut sr_primitives::StorageOverlay,
+			_: &mut sr_primitives::ChildrenStorageOverlay,
+			config: &GenesisConfig,
+		| {
+			runtime_io::with_storage(storage, || {
+				let mut authorities = config.authorities.clone();
+
+				// if session module is here AND it has mutated initial validators list, then we want to sync
```
svyatonik: This code is written in assumption that `config.babe.authorities` is the subset of `config.session.keys`. If that isn't always true, this should be fixed.

svyatonik:
(will reopen once tests are fixed)

srml/babe/src/lib.rs
svyatonik: This wasn't true for `node-executor` tests. I'm not sure - if it is a test problem, or my assumption is invalid. For now I've fixed tests.

srml/babe/src/lib.rs
```diff
@@ -152,6 +157,55 @@ decl_storage! {
...
+				// we expect that the validators set passed received from session will have the same set or subset
+				// of authorities passed to the BABE module
+
+				assert!(
+					session_validators_keys.len() <= authorities.len(),
+					"Trying to select unknown BABE authorities for genesis block",
+				);
+
+				let session_validators_keys_len = session_validators_keys.len();
+				for (idx, session_validator_keys) in session_validators_keys.into_iter().enumerate() {
```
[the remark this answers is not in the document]
svyatonik: Given two lists A and B, only leave elements in A that have corresponding elements in B.

srml/babe/src/lib.rs
```diff
@@ -162,6 +169,59 @@ decl_storage! {
...
+
+			runtime_io::with_storage(storage, || {
+				let mut authorities = config.authorities.clone();
+
+				// if session module is here AND it has mutated initial validators list, then we want to sync
+				// it with our authorities list
+				let session_validators_keys: Vec<(T::AccountId, T::Keys)> = srml_support::storage::unhashed::get_raw(
+					well_known_keys::MUTATED_SESSION_VALIDATORS_KEYS,
+				).and_then(|v| Decode::decode(&mut &v[..]))
+					.expect("unable to decode initial session validators keys");
```
svyatonik: This expect is incorrect && will panic if babe is used without sessions module or if validators aren't updated

svyatonik:
Apart of fixing the standalone BABE initialization issue, the last commit introduces temporary (in-memory) storage which is only used when genesis block is built. It is used to pass mutated validators list from session module to BABE => there are no extra unused-after-genesis value in storage.

srml/babe/src/lib.rs
```diff
@@ -318,6 +327,15 @@ impl<T: Trait> Module<T> {
 	}
 
 	fn do_initialize() {
+		// since do_initialize can be called twice (if session module is present)
+		// => let's ensure that we only modify the storage once per block by checking
```
DemiMarie-parity: Out-of-date comment

svyatonik: Why do you think so?

svyatonik: OK - I see that "by checking block number" is obsolete :)

srml/babe/src/lib.rs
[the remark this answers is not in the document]
DemiMarie-parity: Can you add a comment to this effect?  I suspect that there are faster algorithms than this one.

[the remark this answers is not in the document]
svyatonik: For sure there are :) But given number of session validators/babe authorities isn't that big (there are 4 authorities on FF right now), I believe it works fast enough. It makes no additional allocations + the time to build sorted set/vec of 4 elements and_then search in that set is comparable to the time of doing 4 searches.

Not to mention that this code is only executed once (when genesis is built).

core/finality-grandpa/src/import.rs
```diff
@@ -425,7 +425,7 @@ impl<B, E, Block: BlockT<Hash=H256>, RA, PRA, SC> BlockImport<Block>
 
 		// we don't want to finalize on `inner.import_block`
 		let mut justification = block.justification.take();
-		let enacts_consensus_change = new_cache.contains_key(&well_known_cache_keys::AUTHORITIES);
+		let enacts_consensus_change = !new_cache.is_empty();
```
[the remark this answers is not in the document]
svyatonik: It **will** take it into account - please see how `enacts_consensus_change` is processed: when it's true, it'll fire `self.consensus_changes.lock().note_change((number, hash))` and then this will be processed by the `environment/finalize_block.rs` - see how `justification_required` is built. So the justification will be emitted for such blocks.

But OK - if you do not consider epoch change as consensus change, then let's not emit justification in this case.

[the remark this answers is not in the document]
andresilva: Yes, but the problem is that the GRANDPA voters currently won't take that into account, and will not limit their votes to such blocks (like they do for authority changes), so we might not get a justification for it (we just finalize a further block for example).
@rphmeier Can you comment on whether epoch change should be considered a consensus change?

andresilva:
https://github.com/paritytech/substrate/blob/master/core/service/test/src/lib.rs#L356

Let's re-enable light clients in this test. :)

svyatonik:
I've tried to run local chain to test my merge, but it fails to produce blocks (this also true for clean master branch) :/ Both Alice and Bob are on genesis block forever, without any errors/warnings:
```
Alice: [2019-08-12 08:55:01] 2019-08-12 08:55:01 Idle (1 peers), best: #0 (0xa683…b97d), finalized #0 (0xa683…b97d), ⬇ 0.1kiB/s ⬆ 0.1kiB/s

  Bob: [2019-08-12 08:55:06] 2019-08-12 08:55:06 Idle (1 peers), best: #0 (0xa683…b97d), finalized #0 (0xa683…b97d), ⬇ 0 ⬆ 0
```

I'm running it with:
```
rm -rf db.alice && unbuffer ./substrate --execution=Native --chain=local --node-key 0000000000000000000000000000000000000000000000000000000000000002 --alice --port=30333 -d db.alice --rpc-port=11010 --ws-port=11011 --db-cache=1 2>&1 | unbuffer -p gawk '{ print strftime("Alice: [%Y-%m-%d %H:%M:%S]"), $0 }' | unbuffer -p tee start_alice.log
rm -rf db.bob && unbuffer ./substrate --execution=Native --chain=local --node-key 0000000000000000000000000000000000000000000000000000000000000001 --bob --port 30334 -d db.bob --rpc-port=11012 --ws-port=11013 --db-cache=1 2>&1 | unbuffer -p gawk '{ print strftime("  Bob: [%Y-%m-%d %H:%M:%S]"), $0 }' | unbuffer -p tee start_bob.log
```

If you know, whether the problem is with my args, or with master, please ping me. I'd like to ensure this PR works in the wild before merging it.

core/finality-grandpa/src/import.rs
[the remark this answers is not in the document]
rphmeier: The only downside I see is that we'll get more justifications. Although the point is kind of moot in general because the GRANDPA authorities will change 99% of the time on the same block as BABE or other ones do. It's unlikely that they'll be out of sync.

svyatonik:
There are some issues after last merge, apparently

svyatonik:
So with some debug changes (epoch lasts 30secs + session validators are changing at every epoch change), I'm now getting:
```
Light: [2019-08-14 09:30:15] 2019-08-14 09:30:15 Verification failed from peer: VRF verification of block by author Public(d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d (5GrwvaEF...)) failed: threshold 52390589295859840240752306010969866240 exceeded
```

Investigating...

svyatonik:
I'm still not sure - how the issue should be fixed, but just in case I've synced authorities that are used in next epoch with authorities from next epoch digest in [this commit](https://github.com/paritytech/substrate/commit/970ac46abc382182d1e58efd6ef11062b1c38bad). Looks like it works, but it needs approval before cherry-picking it here.

andresilva:
After discussing this elsewhere we decided to go with the following approach of changing the `Epoch` digest:

```rust
struct Epoch {
	/// ...

	/// The authorities for this Epoch
	pub authorities: Vec<AuthorityId>,
	/// The weights for authorities of previous Epoch (which should be starting when this one is announced).
	pub weights: Vec<BabeAuthorityWeight>,
}
```

core/utils/fork-tree/src/lib.rs
```diff
@@ -305,6 +307,79 @@ impl<H, N, V> ForkTree<H, N, V> where
 		}
 	}
 
+	/// Finalize a node in the tree and all its ancestors. The given function
+	/// `is_descendent_of` should return `true` if the second hash (target) is
+	// a descendent of the first hash (base).
+	pub fn finalize_with_ancestors<F, E>(
```
andresilva: I think this is very similar to what we do in `prune` method.

svyatonik:
So after another discussion, original patch (which stores authorities for the next session in the storage) has been applied.

core/utils/fork-tree/src/lib.rs
svyatonik: Hmm. If I call `prune()`, for example, for block N in [test storage](https://github.com/paritytech/substrate/pull/3301/files#diff-a7d85380647b5b4236cad6aee45caa96R657), it leaves M node. If called for H, the H node is left in the storage. This isn't what expected from `finalize_with_ancestors()`. Probably it could be refactored to support both cases (or we could call `prune()` from within `finalize_with_ancestors()`), but I haven't found a straightforward way to do that.

svyatonik:
I believe I cleaned up all redundant changes. The only suspicious change is about preventing double `on_intialize()` in session module, but it'll cleaned in #3380 anyway.

PR #3305, opened 2019-08-05
fix BABE randomness calculation

rphmeier:
Follow the spec: Use the concatenation of VRF outputs within the epoch as opposed to the XOR.

### DIFF
--- a/srml/babe/src/lib.rs
+++ b/srml/babe/src/lib.rs
@@ -113,7 +113,9 @@ pub trait Trait: timestamp::Trait {
 }
 
 /// The length of the BABE randomness
-pub const RANDOMNESS_LENGTH: usize = 32;
+pub const RANDOMNESS_LENGTH: u32 = 32;
+
+const UNDER_CONSTRUCTION_SEGMENT_LENGTH: usize = 256;
 
 decl_storage! {
 	trait Store for Module<T: Trait> as Babe {
@@ -150,7 +152,16 @@ decl_storage! {
 		NextRandomness: [u8; 32 /* RANDOMNESS_LENGTH */];
 
 		/// Randomness under construction.
-		UnderConstruction: [u8; 32 /* VRF_OUTPUT_LENGTH */];
+		///
+		/// We make a tradeoff between storage accesses and list length.
+		/// We store the under-construction randomness in segments of up to
+		/// `UNDER_CONSTRUCTION_SEGMENT_LENGTH`.
+		///
+		/// Once a segment reaches this length, we begin the next one.
+		/// We reset all segments and return to `0` at the beginning of every
+		/// epoch.
+		SegmentIndex build(|_| 0): u32;
+		UnderConstruction: map u32 => Vec<[u8; 32 /* VRF_OUTPUT_LENGTH */]>;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


srml/babe/src/lib.rs
```diff
@@ -248,19 +259,35 @@ impl<T: Trait> Module<T> {
...
 	}
 
 	/// Call this function exactly once when an epoch changes, to update the
 	/// randomness. Returns the new randomness.
 	fn randomness_change_epoch(next_epoch_index: u64) -> [u8; RANDOMNESS_LENGTH] {
 		let this_randomness = NextRandomness::get();
+		let segment_idx: u32 = <SegmentIndex>::mutate(|s| rstd::mem::replace(s, 0));
+
+		// overestimate to the segment being full.
+		let rho_size = segment_idx.saturating_add(1) as usize * UNDER_CONSTRUCTION_SEGMENT_LENGTH;
```
DemiMarie-parity:
```suggestion
		let rho_size = (segment_idx.saturating_add(1) as usize).saturating_mul(UNDER_CONSTRUCTION_SEGMENT_LENGTH);
```

srml/babe/src/lib.rs
```diff
@@ -345,15 +372,24 @@ impl<T: Trait + staking::Trait> session::OneSessionHandler<T::AccountId> for Mod
...
+// compute randomness for a new epoch. rho is the concatenation of all
+// VRF outputs in the prior epoch.
+//
+// an optional size hint as to how many VRF outputs there were may be provided.
 fn compute_randomness(
 	last_epoch_randomness: [u8; RANDOMNESS_LENGTH],
 	epoch_index: u64,
-	rho: [u8; VRF_OUTPUT_LENGTH],
+	rho: impl Iterator<Item=[u8; VRF_OUTPUT_LENGTH]>,
+	rho_size_hint: Option<usize>,
```
xlc: I thought `usize` is not allowed due to different size between wasm and native build?

bkchr: It is not allowed to be shared between `WASM` and native, but otherwise you can use it. The size of a vector can not be greater than `usize::max` anyway.

rphmeier: yes, that's right -- this isn't exposed anywhere and it's only used to do memory allocation (in which case it is the right unit)

burdges: Interesting, can wasm's usize differ when running on different platforms?  If so, we might advise only to run on 64 bit platforms, so no bad parachain code can leak `usize::MAX` somehow.

srml/babe/src/lib.rs
```diff
@@ -345,15 +372,24 @@ impl<T: Trait + staking::Trait> session::OneSessionHandler<T::AccountId> for Mod
...
-	s[40..].copy_from_slice(&rho);
+	let mut s = Vec::with_capacity(40 + rho_size_hint.unwrap_or(0) * VRF_OUTPUT_LENGTH);
+	s.extend_from_slice(&last_epoch_randomness);
+	s.extend_from_slice(&epoch_index.to_le_bytes());
+
+	for vrf_output in rho {
+		s.extend_from_slice(&vrf_output[..]);
+	}
+
 	runtime_io::blake2_256(&s)
```
burdges: I'm just curious:  We do not hash incrementally in situations like this because repeatedly crossing the wasm boundary costs more than an allocation?

srml/babe/src/lib.rs
bkchr: Wasm is currently always only 32bit. Rho can only store 2^32 elments at max.

burdges:
I like this much more than xor.  

In future, if it improves anything like fetches from storage, then we can always change the spec to include some `randomness_accumulator : [u8; 32]` in the block header, so each block does
```
randomness_accumulator = blake2s(randomness_accumulator ++ our_vrf_output)
```

rphmeier:
@burdges yeah, i thought of doing that as well. it is more hashes in the end, but that might be a good alternative.

burdges:
It's one extra hash and the block header being 32 bytes larger vs an extra fetch from storage.  I donno..  :)

PR #3309, opened 2019-08-06
Rewrite the GrandPa observer work future

tomaka:
Small refactoring.
In preparation for #3099, we get rid of the `loop_fn` function because it has been removed from the new futures library.
This also has the effect of cleaning up a bit the code.

Please either review carefully, or wait for us to have a testnet in order to test the observer.


core/finality-grandpa/src/observer.rs
```diff
@@ -210,85 +272,105 @@ pub fn run_grandpa_observer<B, E, Block: BlockT<Hash=H256>, N, RA, SC>(
...
+	NumberFor<B>: BlockNumberOps,
+	RA: 'static + Send + Sync,
+	E: CallExecutor<B, Blake2Hasher> + Send + Sync + 'static,
+	Bk: Backend<B, Blake2Hasher> + 'static,
+{
+	type Item = ();
+	type Error = Error;
+
+	fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+		match self.observer.poll() {
```
jimpo: Why is `observer` a field on `ObserverWork` rather than constructed on demand at the beginning of `poll()`?

tomaka: I didn't look at the code of the `observer`, but the previous code behaves in the same way.

jimpo: Maybe I don't understand. The old code looks like it creates a new `grandpa_observer` on each execution of the closure. And in the current code, `rebuild_observer` creates the observer from scratch and drops the old one, so it seems like all of the state is given in the args to `grandpa_observer`.

andresilva: It is actually hard to understand because the code relies on implicit side-effects. We only need to `rebuild_observer` when we handle a new command, because the data in https://github.com/paritytech/substrate/pull/3309/files#diff-2fb237d4f37ba90960eacc42fbc35f6aR244 will have changed (from either importing or finalizing a block in another task). Maybe we can add a comment there to explain what is going on here.

tomaka: > The old code looks like it creates a new grandpa_observer on each execution of the closure.

The closure is called at each iteration of the loop, which happens after either the observer or `voter_commands_rx` produce a value.

The old code is indeed extremely confusing, and that's partially why I think it's a good idea to rewrite it.

tomaka: > Maybe we can add a comment to explain what is going on here.

What do you think I should comment on?
I'm not going to explain that we don't recreate the observer every time we poll, because the obvious way of doing it is indeed to not recreate the observer every time we poll.

core/finality-grandpa/src/observer.rs
```diff
@@ -171,34 +171,96 @@ pub fn run_grandpa_observer<B, E, Block: BlockT<Hash=H256>, N, RA, SC>(
...
+		work.rebuild_observer();
+		work
+	}
 
-	let observer_work = future::loop_fn(initial_state, move |state| {
-		let (authority_set, consensus_changes, set_state, voter_commands_rx) = state;
-		let set_id = authority_set.set_id();
-		let voters = Arc::new(authority_set.current_authorities());
-		let client = client.clone();
+	/// Rebuilds the `self.observer` field using the other fields.
```
andresilva:
```suggestion
	/// Rebuilds the `self.observer` field using the current authority set
	/// state. This method should be called when we know that the authority set
	/// has changed (e.g. as signalled by a voter command).
```

PR #3312, opened 2019-08-06
Remove support for mplex

### LINKED ISSUE #3304: Remove support for mplex
Decision taken during the networking workshop.
Let's switch to yamux main.

tomaka:
Fix #3304

Yamux has been the default multiplexing layer for a long time and seems to work well. Mplex was still available as a backup option, but we have not needed it in the past six months. Let's remove support for it.

### DIFF
--- a/core/network/src/lib.rs
+++ b/core/network/src/lib.rs
@@ -89,8 +89,6 @@
 //!
 //! The following multiplexing protocols are supported:
 //!
-//! - [Mplex](https://github.com/libp2p/specs/tree/master/mplex). Support for mplex will likely
-//! be deprecated in the future.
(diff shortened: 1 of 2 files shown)


tomaka:
One "blocker" to this is that the code of PolkadotJS only supports mplex.
Since mplex doesn't really have any support cost, we might just delay this removal.

I'm leaving this PR open a bit in case there's a discussion around that, but please don't merge.

tomaka:
> Can we also remove the references to `mplex` in `network_simult.rs` and `network_dial_error.rs`?

That's in libp2p. We're not removing mplex from libp2p, only from Substrate.

PR #3314, opened 2019-08-06
Equivocations slashing

marcio-diaz:


core/consensus/babe/primitives/src/digest.rs
```diff
@@ -102,41 +121,40 @@ impl Decode for BabePreDigest {
 }
 
 /// A digest item which is usable with BABE consensus.
-#[cfg(feature = "std")]
+// #[cfg(feature = "std")]
 pub trait CompatibleDigestItem: Sized {
 	/// Construct a digest item which contains a BABE pre-digest.
-	fn babe_pre_digest(seal: BabePreDigest) -> Self;
+	fn babe_pre_digest<D: Codec>(seal: D) -> Self;
```
andresilva: Why is this change needed? The runtime should be using `RawBabePreDigest` anyway. This function signature doesn't make a lot of sense, it looks like you can give it anything and it will make a BABE seal which is not the case at all, since you need to give it some specific struct to make a proper BABE seal.

core/consensus/aura/src/lib.rs
```diff
@@ -447,13 +450,14 @@ fn check_header<C, B: BlockT, P: Pair>(
 }
 
 /// A verifier for Aura blocks.
-pub struct AuraVerifier<C, P> {
+pub struct AuraVerifier<C, P, T> {
 	client: Arc<C>,
 	phantom: PhantomData<P>,
 	inherent_data_providers: inherents::InherentDataProviders,
+	transaction_pool: Option<Arc<T>>,
```
andresilva: This should probably be replaced with a `F: Fn(Block::Extrinsic)` which will just push the extrinsic to the transaction pool As it stands you can't do anything with this `T`, and it's probably easier to not drag any complicated `TransactionPool` types here.

core/consensus/babe/primitives/src/lib.rs
```diff
@@ -133,6 +138,90 @@ impl slots::SlotData for BabeConfiguration {
 	const SLOT_KEY: &'static [u8] = b"babe_bootstrap_data";
 }
 
+/// Represents an Babe equivocation proof.
+#[derive(Debug, Clone, Encode, Decode, PartialEq)]
+pub struct BabeEquivocationProof<H> {
```
andresilva: I thought we had a type for `EquivocationProof` already (in slots). Why add another? I also don't know why you need the authority signatures here, if you pass the headers you can extract the author from it.

core/consensus/common/primitives/src/lib.rs
```diff
@@ -29,3 +31,45 @@ decl_runtime_apis! {
 		fn authorities() -> Vec<AuthorityId>;
 	}
 }
+
+pub trait AuthorshipEquivocationProof {
```
andresilva: IMO this is not needed, it's not abstracting over anything anyway.

core/consensus/slots/src/aux_schema.rs
```diff
@@ -52,67 +55,65 @@ pub struct EquivocationProof<H> {
...
 		// 1) signed by the same voter,
 		if prev_signer == signer {
 			// 2) with different hash
 			if header.hash() != prev_header.hash() {
-				return Ok(Some(EquivocationProof {
-					slot, // 3) and mentioning the same slot.
-					fst_header: prev_header.clone(),
-					snd_header: header.clone(),
-				}));
+				return Ok(Some(AuthorshipEquivocationProof::new(
```
andresilva: Why not just use the existing struct and just pass the remaining data as arguments to the construct report call?

marcio-diaz: Why not doing it in this way? Anyway the `proof` and the `SessionIndex` should be in the `Equivocation`, right?.

core/consensus/common/primitives/src/lib.rs
marcio-diaz: It's abstracting over Aura and Babe equivocation proof (I have the Aura implementation with low priority).

core/consensus/babe/primitives/src/lib.rs
marcio-diaz: Because we are taking the signatures out of the header (I think we talk about this yesterday). In your case I will need to push the signature again, and pop it again when checking...

core/consensus/common/primitives/src/lib.rs
andresilva: https://github.com/paritytech/substrate/pull/3314/files/1a88f9016ae3b16937afce5e3fb189749dd8e521#diff-1144b2e008f250f4ca22726b35ee22c7R65

How will you implement this for GRANDPA? Same goes for some other methods which are specific to block proposal equivocations. My point is there's too little in common between BABE and GRANDPA equivocations to abstract anything, so there's no point in abstracting. They'll be reported to different modules anyway so I don't know why you need a common type.

core/consensus/babe/primitives/src/digest.rs
marcio-diaz: Yeah, it should be a better way to do it. But it also using a non raw pre digest.

core/consensus/common/primitives/src/lib.rs
marcio-diaz: It's interfacing  Aura and Babe with the Slots module. If I remove this I need to duplicate code in Slots module, or there is a better way?

andresilva: Sorry I misread your comment earlier and thought this was supposed to abstract over GRANDPA and BABE.

core/consensus/aura/src/lib.rs
marcio-diaz: I'l add add a trait that implements `submit_one` to the transaction pool, as per offline discussion :).

core/consensus/babe/src/lib.rs
```diff
@@ -519,20 +511,49 @@ fn check_header<B: BlockT + Sized, C: AuxStore>(
...
 					"Slot author {:?} is equivocating at slot {} with headers {:?} and {:?}",
 					author,
 					slot_number,
-					equivocation_proof.fst_header().hash(),
-					equivocation_proof.snd_header().hash(),
+					equivocation_proof.first_header().hash(),
+					equivocation_proof.second_header().hash(),
 				);
+
+				let keystore = keystore.expect("FIXME");
```
tomusdrw: I don't think keystore should be needed for anything.

core/consensus/babe/primitives/src/lib.rs
marcio-diaz: I don't think that is possible unless we keep the signature inside. I think signature outside is more flexible.

core/consensus/babe/primitives/src/digest.rs
```diff
@@ -93,6 +88,30 @@ impl BabePreDigest {
 	}
 }
 
+/// Get the slot.
+pub fn get_slot<H: Header>(header: &H) -> Result<SlotNumber, &str>
```
rphmeier: The documentation and function naming needs real improvement

core/consensus/babe/primitives/src/digest.rs
```diff
@@ -93,6 +88,30 @@ impl BabePreDigest {
...
+	where DigestItemForHeader<H>: CompatibleDigestItem,
+{
+	find_pre_digest::<H, RawBabePreDigest>(header)
+		.map(|raw_pre_digest| raw_pre_digest.slot_number())
+}
+
+/// Extract the BABE pre digest from the given header. Pre-runtime digests are
+/// mandatory, the function will return `Err` if none is found.
+pub fn find_pre_digest<H: Header, D: Codec>(header: &H) -> Result<D, &str>
```
rphmeier: it's strange to return a `str` with the lifetime of the given header as opposed to a `&'static str`.

core/consensus/babe/src/lib.rs
```diff
@@ -537,20 +537,45 @@ fn check_header<B: BlockT + Sized, C: AuxStore, T>(
...
+			slot_number,
+			equivocation_proof.first_header.hash(),
+			equivocation_proof.second_header.hash(),
 		);
+
+		// Submit a transaction reporting the equivocation.
+		let block_id = BlockId::number(client.info().best_number);
+		let maybe_report_transaction = client
+			.runtime_api()
+			.construct_equivocation_transaction_with_context(
```
rphmeier: This is IMO not the right level of abstraction.

It's better to think of things in terms of making code testable: rather than a `SubmitExtrinsic` trait (which is itself better than passing a raw transaction queue), you should have a `SubmitEquivocationProof` trait, which deals with extrinsics internally.

This means that tests don't have to deal with extrinsics, which is way more nasty.

core/consensus/babe/src/lib.rs
```diff
@@ -537,20 +537,45 @@ fn check_header<B: BlockT + Sized, C: AuxStore, T>(
...
+
+		if let Ok(Some(report_transaction)) = maybe_report_transaction {
+			transaction_pool.as_ref().map(|txpool| {
+				let uxt = Decode::decode(&mut report_transaction.as_slice())
+					.expect("Encoded extrinsic is valid; qed");
+				txpool.submit_extrinsic(&block_id, uxt)
+			});
+			info!(target: "afg", "Babe equivocation report has been submitted")
+		} else {
+			error!(target: "afg", "Error constructing Babe equivocation report")
```
rphmeier: why afg?

core/consensus/babe/src/tests.rs
```diff
@@ -151,7 +165,8 @@ impl TestNetFactory for BabeTestNet {
 				inherent_data_providers,
 				config,
 				time_source: Default::default(),
-				transaction_pool : Default::default(),
+				// TODO [slashing] get the transaction pool from somewhere.
+				transaction_pool: Default::default(),
```
rphmeier: see -- you have trouble writing the test because you've tried to push too much onto the BABE logic.

core/consensus/common/primitives/src/lib.rs
```diff
@@ -18,10 +18,23 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use codec::Codec;
+use codec::{Encode, Decode, Codec};
 use client::decl_runtime_apis;
 use rstd::vec::Vec;
 
+/// Represents an Babe equivocation proof.
```
rphmeier: but the crate is consensus-common

core/consensus/common/primitives/src/lib.rs
```diff
@@ -18,10 +18,23 @@
...
+#[cfg_attr(feature = "std", derive(Debug))]
+#[derive(Clone, PartialEq, Eq, Decode, Encode)]
+pub struct EquivocationProof<H, P, S> {
+	pub reporter: P,
+	pub identity: P,
```
rphmeier: no docs.

core/consensus/common/primitives/src/lib.rs
```diff
@@ -18,10 +18,23 @@
...
-use codec::Codec;
+use codec::{Encode, Decode, Codec};
 use client::decl_runtime_apis;
 use rstd::vec::Vec;
 
+/// Represents an Babe equivocation proof.
+#[cfg_attr(feature = "std", derive(Debug))]
+#[derive(Clone, PartialEq, Eq, Decode, Encode)]
+pub struct EquivocationProof<H, P, S> {
+	pub reporter: P,
```
rphmeier: why should the proof care?

core/consensus/common/primitives/src/lib.rs
```diff
@@ -18,10 +18,23 @@
...
+	pub identity: P,
+	pub slot: u64,
```
rphmeier: should use a strong type alias as in the one from babe primitives.

I'm not sure why `consensus-common` needs to define this type. Trying to unify things like equivocation proofs across all consensus engines is going to fail.

core/consensus/slots/src/lib.rs
```diff
@@ -20,8 +20,8 @@
 //! time during which certain events can and/or must occur.  This crate
 //! provides generic functionality for slots.
 
-#![deny(warnings)]
-#![forbid(unsafe_code, missing_docs)]
+// #![deny(warnings)]
+// #![forbid(unsafe_code, missing_docs)]
```
rphmeier: don't comment out code

rphmeier: it should be either removed or left there

core/finality-grandpa/primitives/src/lib.rs
```diff
@@ -52,6 +54,35 @@ pub type AuthorityWeight = u64;
...
+pub fn localized_payload<E: Encode>(round: u64, set_id: u64, message: &E) -> Vec<u8> {
+	(message, round, set_id).encode()
+}
+
+// TODO [slashing] consider removing and use generic `EquivocationProof`.
+#[cfg_attr(feature = "std", derive(Debug))]
+#[derive(Clone, PartialEq, Eq, Encode, Decode)]
+pub struct GrandpaEquivocation<H, N> {
+	/// Reporter of the equivocation.
+	pub reporter: AuthorityId,
```
rphmeier: why does it matter here?

core/finality-grandpa/primitives/src/lib.rs
```diff
@@ -52,6 +54,35 @@ pub type AuthorityWeight = u64;
 /// The index of an authority.
 pub type AuthorityIndex = u64;
 
+pub fn localized_payload<E: Encode>(round: u64, set_id: u64, message: &E) -> Vec<u8> {
+	(message, round, set_id).encode()
+}
+
+// TODO [slashing] consider removing and use generic `EquivocationProof`.
```
rphmeier: "generic"

core/finality-grandpa/primitives/src/lib.rs
```diff
@@ -52,6 +54,35 @@ pub type AuthorityWeight = u64;
...
+	pub round_number: u64,
+	/// The identity of the equivocator.
+	pub identity: AuthorityId,
+	/// The first vote in the equivocation.
+	pub first: (Message<H, N>, AuthoritySignature),
+	/// The second vote in the equivocation.
+	pub second: (Message<H, N>, AuthoritySignature),
+}
+
+pub type GrandpaEquivocationFrom<Block> = GrandpaEquivocation<
```
rphmeier: docs? convention is `Of`, not `From`

core/finality-grandpa/src/environment.rs
```diff
@@ -377,9 +379,10 @@ pub(crate) struct Environment<B, E, Block: BlockT, N: Network<Block>, RA, SC> {
 	pub(crate) network: crate::communication::NetworkBridge<Block, N>,
 	pub(crate) set_id: u64,
 	pub(crate) voter_set_state: SharedVoterSetState<Block>,
+	pub(crate) transaction_pool: Arc<T>,
```
rphmeier: Again: punting extrinsic generation to consensus modules is beyond scope.

core/finality-grandpa/src/environment.rs
```diff
@@ -840,7 +844,45 @@ where
...
+		let second_signature = equivocation.second.1;
+
+		if let Some(local_key) = crate::is_voter(&self.voters, &self.config.keystore) {
+			let reporter = local_key.public();
+
+			let grandpa_equivocation = GrandpaEquivocationFrom::<Block> {
+				reporter,
+				round_number: equivocation.round_number,
+				// TODO [slashing] get proper session index.
+				session_index: SessionIndex::default(),
```
rphmeier: why do you want to do this in the node and not in the runtime?

node/runtime/src/lib.rs
```diff
@@ -136,6 +141,10 @@ parameter_types! {
 impl babe::Trait for Runtime {
 	type EpochDuration = EpochDuration;
 	type ExpectedBlockTime = ExpectedBlockTime;
+	type IdentificationTuple = historical::IdentificationTuple<Self>;
```
rphmeier: this is redundant because it can be accessed by the `KeyOwnerSystem`. (it should be called `KeyOwnerProofSystem`)

node/runtime/src/lib.rs
```diff
@@ -407,6 +416,9 @@ impl offences::Trait for Runtime {
 
 impl grandpa::Trait for Runtime {
 	type Event = Event;
+	type IdentificationTuple = historical::IdentificationTuple<Self>;
```
rphmeier: also redundant

srml/babe/src/lib.rs
```diff
@@ -191,14 +212,48 @@ decl_storage! {
...
+	if maybe_first_slot.is_err() || maybe_second_slot.is_err() {
+		return false
+	}
+	let first_slot = maybe_first_slot.expect("checked before; qed");
+	let second_slot = maybe_second_slot.expect("checked before; qed");
+
+	if equivocation.slot == first_slot && first_slot == second_slot {
+		let author = &equivocation.identity;
+
+		if !author.verify(&first_header.hash(), &equivocation.first_signature) {
```
rphmeier: this is a serious misunderstanding of BABE - the hash includes the seal includes the signature, so it will always fail. I'm not sure why the entire header needs to go into the proof at all.

srml/babe/src/lib.rs
```diff
@@ -223,6 +278,33 @@ decl_module! {
...
+			equivocation: EquivocationProof<T::Header, AuthorityId, AuthoritySignature>,
+			proof: T::Proof
+		) {
+			let who = ensure_signed(origin)?;
+
+			if !equivocation_is_valid::<T>(&equivocation) {
+				return Err("invalid equivocation")
+			}
+
+			let to_punish = <T as Trait>::KeyOwnerSystem::check_proof(
```
rphmeier: I would prefer if the proof was checked first as opposed to afterwards

srml/babe/src/lib.rs
```diff
@@ -344,6 +424,53 @@ impl<T: Trait> Module<T> {
...
+			let check_nonce = system::CheckNonce::<T>::from(Default::default());
+			let check_weight = system::CheckWeight::<T>::new();
+			let take_fees = balances::TakeFees::<T>::from(Default::default());
+			let extra = (check_genesis, check_era, check_nonce, check_weight, take_fees);
+
+			let genesis_hash = <system::Module<T>>::block_hash(T::BlockNumber::zero());
+			let raw_payload = (function, extra.clone(), genesis_hash, genesis_hash);
+
+			let maybe_signature: Option<app::Signature> = raw_payload.using_encoded(|payload| if payload.len() > 256 {
+				reporter.sign(&sr_io::blake2_256(payload))
```
rphmeier: have you tested that this actually works?

srml/grandpa/src/lib.rs
```diff
@@ -166,14 +178,73 @@ decl_storage! {
...
+) -> bool {
+	let first_vote = &equivocation.first.0;
+	let first_signature = &equivocation.first.1;
+
+	let second_vote = &equivocation.second.0;
+	let second_signature = &equivocation.second.1;
+
+	let identity = &equivocation.identity;
+
+	if first_vote == second_vote { // TODO: fix equality before merging
```
rphmeier: yeah..

tomusdrw:
The commit history of this PR grew too much and there is still a lot of work that needs changing. This PR is going to be superseded by a new branch where we will copy over the correct parts.

Closing it now to save the CI and reviewers cycles.

PR #3316, opened 2019-08-06
Fix decl_storage map and linked_map for insert_ref

thiolliere:
Default implementation of insert_ref was writting to using twox 128 hasher instead of choosen hasher.

Also linkedmap implementation was wrong because default implementation wasn't override.

### DIFF
--- a/srml/support/procedural/src/storage/impls.rs
+++ b/srml/support/procedural/src/storage/impls.rs
@@ -634,7 +634,6 @@ impl<'a, I: Iterator<Item=syn::Meta>> Impls<'a, I> {
 						None => Self::new_head_linkage(storage, key),
 					};
 					storage.put(key_for, &(val, linkage))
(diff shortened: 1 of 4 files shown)


srml/support/procedural/src/storage/impls.rs
```diff
@@ -615,6 +615,28 @@ impl<'a, I: Iterator<Item=syn::Meta>> Impls<'a, I> {
...
+
+					let key_for = &*#as_map::key_for(key);
+					let linkage = match Self::read_with_linkage(storage, key_for) {
+						// overwrite but reuse existing linkage
+						Some((_data, linkage)) => linkage,
+						// create new linkage
+						None => Self::new_head_linkage(storage, key),
+					};
+					storage.put(key_for, &(val, linkage))
+					// val.using_encoded(|b| storage.put_raw(&Self::key_for(key)[..], b))
```
bkchr:
```suggestion
```

PR #3317, opened 2019-08-06
Set uncles inherent

### LINKED ISSUE #2956: Actually construct the `Authorship::set_uncles` inherent
#2941 creates logic for a `set_uncles` inherent to set some uncles. But we'll want to actually create a piece of logic that can call `client.uncles()` and ignore any duplicates (i.e. those which have already been included). We can model off the code for the `FinalityTracker` inherent, which makes use of the client as well.

arkpar:
Closes #2956 

Inherent data is populated with all possible uncles for the best block. When authoring this list is additionally filtered to remove all uncles that the runtime considers invalid.
There a small issue with this implementation. Currently inherents are populated before the block to propose on is chosen. If canonical block changes after uncles are collected we might end up with no uncles included for this block.


srml/authorship/src/lib.rs
```diff
@@ -101,16 +156,16 @@ pub trait FilterUncle<Header, Author> {
 
 	/// Do additional filtering on a seal-checked uncle block, with the accumulated
 	/// filter.
-	fn filter_uncle(header: &Header, acc: Self::Accumulator)
-		-> Result<(Option<Author>, Self::Accumulator), &'static str>;
+	fn filter_uncle(header: &Header, acc: &mut Self::Accumulator)
```
arkpar: Changed `acc` to be a mutable reference because we also need it in case of error so that filtering might be able to skip invalid uncle and continue. Returning `acc` for both `Ok` and `Err` results is way less readable and more verbose code.

core/client/src/client.rs
```diff
@@ -1343,6 +1350,20 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
+impl<B, E, Block, RA> ProvideUncles<Block> for Client<B, E, Block, RA> where
+	B: backend::Backend<Block, Blake2Hasher>,
+	E: CallExecutor<Block, Blake2Hasher>,
+	Block: BlockT<Hash=H256>,
+{
+	fn uncles(&self, target_hash: Block::Hash, max_generation: NumberFor<Block>) -> error::Result<Vec<Block::Header>> {
+		Client::uncles(self, target_hash, max_generation)?
+			.into_iter()
+			.map(|hash| Client::header(self, &BlockId::Hash(hash))
+				.map(|maybe_header| maybe_header.expect("Each uncle exists in the database; qed")))
```
rphmeier: it would be much safer to `filter_map`. this expectation doesn't prove to me why the uncle will exist (in the future, it might race against block pruning or something similar)

core/consensus/babe/src/lib.rs
```diff
@@ -85,6 +85,9 @@ mod aux_schema;
 mod tests;
 pub use babe_primitives::AuthorityId;
 
+/// Maximum uncles generations we may provide to the runtime.
+const MAX_UNCLE_GENERATIONS: u32 = 8;
```
rphmeier: this should be drawn from the runtime instead

core/consensus/babe/src/lib.rs
```diff
@@ -769,6 +777,41 @@ fn register_babe_inherent_data_provider(
 	}
 }
 
+/// Register uncles inherent data provider, if not registered already.
+fn register_uncles_inherent_data_provider<B, C, SC>(
```
rphmeier: putting this in the BABE crate doesn't seem right - other block production systems will also want uncles

srml/authorship/src/lib.rs
```diff
@@ -267,55 +355,80 @@ impl<T: Trait> Module<T> {
...
 
-			let in_chain = <system::Module<T>>::block_hash(uncle.number()) == hash;
+impl<T: Trait> ProvideInherent for Module<T> {
+	type Call = Call<T>;
+	type Error = MakeFatalError<()>;
+	const INHERENT_IDENTIFIER: InherentIdentifier = INHERENT_IDENTIFIER;
 
-			if duplicate || in_chain { return Err("uncle already included") }
+	fn create_inherent(data: &InherentData) -> Option<Self::Call> {
+		let uncles = data.uncles().expect("Gets and decodes final number inherent data");
```
rphmeier: panicking here is wrong

arkpar: Panic happens when the data has invalid encoding, not when it is missing.

core/consensus/babe/src/lib.rs
arkpar: Runtime calls are expensive. And this will be called for each block verification. My reasoning is that it is better to pick a hard limit, pass a few extra headers and let the runtime sort it out later rather than make another call into WASM.

srml/authorship/src/lib.rs
rphmeier: yes. still wrong IMO. what proves that the data will have correct encoding? 

you could maybe make a case that the provider provides data that's correct, but it's a bit of a far-reaching assumption, since it can be broken by a change in provider implementation later.

core/consensus/babe/src/lib.rs
rphmeier: it will be called for every block creation only. but yeah, this is fine for now.

arkpar: From what I see `InherentData` is populated for verification as well.

PR #3319, opened 2019-08-07
Propagate cache when calling into runtime

arkpar:


arkpar:
This is a performance fix

Issue #3038, opened 2019-07-05
[RFC] srml-contracts: Remove ext_return


jimpo:
Right now there's an issue with gas metering and `ext_return`. In the following block of code:

```
block
get_local $x
call 4
i32.add
end
```

Currently, we will charge gas at the top of the block for all three instructions in the block. However, `call 4` might invoke a function which [invokes a function which]* calls `ext_return`. In this case the user is charged for the `i32.add` instruction which was never executed.

This is because `ext_return` is special in how it affects control flow. Handling it properly could lead to a lot more gas metering calls being injected after any control block or call to a function that might invoke `ext_return`.

So... I think we should discuss removing `ext_return` entirely. The main point of `ext_return` seems to be returning the result of the contract's execution to the caller. Instead, the `call` function on a contract could return an integer status code (0 = success, 1 = some time of error, 2 = another type of error, etc), and whatever output data remains in the scratch buffer when the function returns is its output data. This means that contracts would have to explicitly clear the scratch buffer before returning if they do not want to return anything, as there may be garbage left in it. If the function traps, the final scratch buffer state is ignored.

This makes sense to me because it maps pretty easily to a Rust function that returns `Result<Vec<u8>, ContractError>`, where the status code represents the variant type of the `Err(ContractError(..))` or 0 for `Ok(..)`. Then it is the job of ink! to properly handle the final state of the scratch buffer.

In short, removing `ext_return` simplifies the contract runtime by pushing some complexity into the contract DSL, ink!

I'd be interested to any thoughts. @Robbepop @pepyakin 

This is related to #2852.

pepyakin:
A couple of thoughts:

1. As we noticed, this is not only property of `ext_return` but also of the wasm instruction `unreachable` and traps in general (e.g. OOB access). However, if we were to fix that we would have to basically create a new gas metering block after each instruction that can trap which is not feasible.
1. Another reason for this problem to happen is that we refund the gas. If we weren't returning the gas after the trap we wouldn't have this problem. While it might seem that it is OK to burn the gas to incentivise non-trapping execution, it seems useful to refund gas in some cases when the trap is deliberate (reasons similar to #2852). There is also incentive mismatch that the people who pay for the execution of the code (i.e. transaction) are most often different from the people who write the code. And even if they are the same, in case of callbacks used in contracts the gap increases.
[...]

PR #3320, opened 2019-08-07
srml-contracts: Contract calls/instantiations to return exit statuses

jimpo:
This PR does two things

1. Refactor the exec logic so that a single allocated scratch buffer is used for all nested calls during a contract execution.
2. Allow the "call" and "deploy" functions in contracts to return an exit status and output data through the scratch buffer. This is an implementation of #3038.

The PR could be split into two if people prefer, but the status code changes somewhat motivate the other refactors.

### Global scratch buffer

The idea with the global scratch buffer is that a contract execution may invoke other nested executions through `ext_call` or `ext_create`. Each execution needs a scratch buffer. Notably, the scratch buffer is overwritten on both of these calls so it is possible to pass ownership of the same scratch buffer to the nested executions in order to reduce memory allocations. So this is ultimately a memory optimization.

I'm not sure how important it actually is given that each nested execution creates a new sandbox anyway which requires more memory, so that could just be priced into the imported function cost. But it does seem like kind of an obvious thing to do, and one that is partially implemented already using the `OutputBuffer`/`EmptyOutputBuffer`.

The downside is additional complexity as the buffer needs to be passed back even in the error case, requiring a custom error type. An alternative approach would be to pass around the scratch buffer in a `RefCell` instead of entirely through move semantics.

### Exit status

With this PR, the "call" and "deploy" functions exported from a contract can now return a 32-bit exit code. The low 8 bits of the exit code are the exit status, which is returned to the caller. The top 24 bits are currently unused and are reserved for interpretation by the contracts runtime. (See https://github.com/paritytech/substrate/issues/3038#issuecomment-516852022). If the exit status is 0, then the execution succeeds, otherwise the changes are reverted. If an exit status is returned, then the scratch buffer on exit is returned as output data, providing an alternative to `ext_return`.

This change is made in a backwards compatible way. `ext_return` still exists for now and if the "call" or "deploy" functions have an empty return type, then on exit they return a status of 0 with empty output data. Part of the motivation for backwards compatibility is so that this change does not need to be synchronized with `ink!`.  Going forward, allowing "call" and "deploy" to return unit might be desirable to keep around since it is a fairly common pattern, especially in the tests. But maybe we want to drop this and require them to return a status code.

### DIFF
--- a/srml/contracts/src/exec.rs
+++ b/srml/contracts/src/exec.rs
@@ -979,15 +979,49 @@ mod tests {
 			ctx.overlay.set_balance(&origin, 100);
 			ctx.overlay.set_balance(&dest, 0);
 
-			let result = ctx.call(
+			let output = ctx.call(
 				dest,
 				55,
 				&mut GasMeter::<Test>::with_limit(1000, 1),
 				vec![],
-			);
-			assert_matches!(result, Ok(_));
+			).unwrap();
+
+			assert!(output.is_success());
 			assert_eq!(ctx.overlay.get_balance(&origin), 45);
 			assert_eq!(ctx.overlay.get_balance(&dest), 55);
+		});
+	}
+
+	#[test]
+	fn changes_are_reverted_on_failing_call() {
+		// This test verifies that a contract is able to transfer
+		// some funds to another account.
+		let origin = ALICE;
+		let dest = BOB;
+
+		let vm = MockVm::new();
+		let mut loader = MockLoader::empty();
+		let return_ch = loader.insert(
+			|_| Ok(ExecReturnValue { status: 1, data: Vec::new() })
+		);
+
+		with_externalities(&mut ExtBuilder::default().build(), || {
+			let cfg = Config::preload();
+			let mut ctx = ExecutionContext::top_level(origin, &cfg, &vm, &loader);
+			ctx.overlay.create_contract(&BOB, return_ch).unwrap();
+			ctx.overlay.set_balance(&origin, 100);
+			ctx.overlay.set_balance(&dest, 0);
+
+			let output = ctx.call(
+				dest,
+				55,
+				&mut GasMeter::<Test>::with_limit(1000, 1),
+				vec![],
+			).unwrap();
+
+			assert!(!output.is_success());
+			assert_eq!(ctx.overlay.get_balance(&origin), 100);
+			assert_eq!(ctx.overlay.get_balance(&dest), 0);
 		});
 	}
 
@@ -1120,8 +1154,8 @@ mod tests {
 	}
 
 	#[test]
-	fn output_is_returned() {
-		// Verifies that if a contract returns data, this data
+	fn output_is_returned_on_success() {
+		// Verifies that if a contract returns data with a successful exit status, this data
 		// is returned from the execution context.
 		let origin = ALICE;
 		let dest = BOB;
@@ -1145,12 +1179,44 @@ mod tests {
 			);
 
 			let output = result.unwrap();
+			assert!(output.is_success());
 			assert_eq!(output.data, vec![1, 2, 3, 4]);
 		});
 	}
 
 	#[test]
-	fn input_data() {
+	fn output_is_returned_on_failure() {
+		// Verifies that if a contract returns data with a failing exit status, this data
+		// is returned from the execution context.
+		let origin = ALICE;
+		let dest = BOB;
+
+		let vm = MockVm::new();
+		let mut loader = MockLoader::empty();
+		let return_ch = loader.insert(
+			|_| Ok(ExecReturnValue { status: 1, data: vec![1, 2, 3, 4] })
+		);
+
+		with_externalities(&mut ExtBuilder::default().build(), || {
+			let cfg = Config::preload();
+			let mut ctx = ExecutionContext::top_level(origin, &cfg, &vm, &loader);
+			ctx.overlay.create_contract(&BOB, return_ch).unwrap();
+
+			let result = ctx.call(
+				dest,
+				0,
+				&mut GasMeter::<Test>::with_limit(1000, 1),
+				vec![],
+			);
+
+			let output = result.unwrap();
+			assert!(!output.is_success());
+			assert_eq!(output.data, vec![1, 2, 3, 4]);
+		});
+	}
+
+	#[test]
+	fn input_data_to_call() {
 		let vm = MockVm::new();
 		let mut loader = MockLoader::empty();
 		let input_data_ch = loader.insert(|ctx| {
@@ -1171,6 +1237,16 @@ mod tests {
 				vec![1, 2, 3, 4],
 			);
 			assert_matches!(result, Ok(_));
+		});
+	}
+
+	#[test]
+	fn input_data_to_instantiate() {
+		let vm = MockVm::new();
+		let mut loader = MockLoader::empty();
+		let input_data_ch = loader.insert(|ctx| {
+			assert_eq!(ctx.input_data, &[1, 2, 3, 4]);
+			exec_success()
 		});
 
 		// This one tests passing the input data into a contract via instantiate.
@@ -1348,11 +1424,13 @@ mod tests {
 	}
 
 	#[test]
-	fn instantiation() {
+	fn instantiation_work_with_success_output() {
 		let vm = MockVm::new();
 
 		let mut loader = MockLoader::empty();
-		let dummy_ch = loader.insert(|_| exec_success());
+		let dummy_ch = loader.insert(
+			|_| Ok(ExecReturnValue { status: STATUS_SUCCESS, data: vec![80, 65, 83, 83] })
+		);
 
 		with_externalities(
 			&mut ExtBuilder::default().existential_deposit(15).build(),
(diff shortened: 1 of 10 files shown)


srml/contracts/src/exec.rs
```diff
@@ -982,19 +981,52 @@ mod tests {
...
+			assert!(output.is_success());
 			assert_eq!(ctx.overlay.get_balance(&origin), 45);
 			assert_eq!(ctx.overlay.get_balance(&dest), 55);
 		});
 	}
 
+	#[test]
+	fn changes_are_reverted_on_failing_call() {
+		// This test verifies that a contract is able to transfer
+		// some funds to another account.
```
pepyakin: This comment doesn't seem to be relevant.

srml/contracts/src/tests.rs
```diff
@@ -1565,3 +1570,357 @@ fn storage_max_value_limit() {
...
+		(call $assert
+			(i64.eq (get_local $balance) (call $current_balance (get_local $sp)))
+		)
+	)
+
+	(data (i32.const 0) "\00\80")		;; The value to transfer on instantiation and calls.
+										;; Chosen to be greater than existential deposit.
+	(data (i32.const 8) "\00\11\22\33\44\55\66\77")		;; The input data to instantiations and calls.
+)
+"#;
```
pepyakin: ༼ ༎ຶ ෴ ༎ຶ༽

srml/contracts/src/wasm/runtime.rs
```diff
@@ -352,22 +394,25 @@ define_env!(Env, <E: Ext>,
...
-				ctx.scratch_buf = output_data;
-				Ok(0)
+			Ok(output) => {
+				ctx.scratch_buf = output.data;
+				Ok(output.status.into())
+			},
+			Err(buffer) => {
+				ctx.scratch_buf = buffer;
+				ctx.scratch_buf.clear();
+				Ok(0x0100)
```
pepyakin: What is this value? Can we extract and leave a note about it?

srml/contracts/src/wasm/prepare.rs
```diff
@@ -227,14 +227,16 @@ impl<'a> ContractModule<'a> {
 			};
 
 			// Then check the signature.
-			// Both "call" and "deploy" has a () -> () function type.
+			// Both "call" and "deploy" has a [] -> [] or [] -> [i32] function type.
```
pepyakin: Well, I am not sure that we need to do this. Although I don't see any problems with it right now, I think it might complicate things in future, while it doesn't provide much benefit (maybe I am missing something though).

Let's make it strictly `[] -> [i32]`

jimpo: Yeah, the only benefits are that 1) it's backwards compatible, which is maybe a good idea until ink! is updated, and 2) it makes the tests, which use hand-written Wasm a bit simpler.

My preference would be to keep in this PR and remove the `[] -> []` signature in a follow-up, but I could remove it in this one as well if you prefer.

srml/contracts/src/tests.rs
jimpo: Haha, yeah... I thought it would be annoying to split into multiple test functions, as there's a decent amount of boilerplate to just create an execute the contracts. OTOH, having one huge test makes it a slog to read through and more difficult to fix if it breaks for some reason. WDYT?

srml/contracts/src/wasm/prepare.rs
pepyakin: Ah, no. That works! Just add a comment.

srml/contracts/src/exec.rs
```diff
@@ -34,15 +34,54 @@ pub type BlockNumberOf<T> = <T as system::Trait>::BlockNumber;
...
+/// precondition of a system call, etc.) or errors with the orchestration (eg. out-of-gas errors, a
+/// non-existent destination contract, etc.).
 #[cfg_attr(test, derive(Debug))]
-pub struct CallReceipt {
-	/// Output data received as a result of a call.
-	pub output_data: Vec<u8>,
+pub struct ExecError {
+	pub reason: &'static str,
+	/// This is an allocated buffer that may be reused. The buffer must be cleared be explicitly
+	/// cleared before reuse.
```
thiolliere: typo `The buffer must be explicitly cleared before reuse.` ?

srml/contracts/src/exec.rs
```diff
@@ -374,111 +353,127 @@ where
...
 
 			// If code_hash is not none, then the destination account is a live contract, otherwise
 			// it is a regular account since tombstone accounts have already been rejected.
-			let output_data = match nested.overlay.get_code_hash(&dest) {
+			match nested.overlay.get_code_hash(&dest) {
 				Some(dest_code_hash) => {
-					let executable = nested.loader.load_main(&dest_code_hash)?;
+					let executable = try_or_exec_error!(
+						nested.loader.load_main(&dest_code_hash),
+						 input_data
```
thiolliere: is there an indentation issue here ?

srml/contracts/src/tests.rs
thiolliere: I haven't completely read it but I might like it, it is kind of a mock that allow to write lot of test, this seems useful to me.

srml/contracts/src/wasm/runtime.rs
```diff
@@ -186,12 +208,32 @@ fn read_sandbox_memory<E: Ext>(
...
+	ptr: u32,
+	len: u32,
+) -> Result<(), sandbox::HostError> {
+	charge_gas(ctx.gas_meter, ctx.schedule, RuntimeToken::ReadMemory(len))?;
 
-	Ok(buf)
+	let mut buf = mem::replace(&mut ctx.scratch_buf, Vec::new());
+	buf.resize(len as usize, 0);
+	ctx.memory().get(ptr, buf.as_mut_slice()).map_err(|_| sandbox::HostError)?;
+	let _  = mem::replace(&mut ctx.scratch_buf, buf);
```
thiolliere: The reason you use replace is to satisfy the property: `Resizes the scratch buffer to the specified length on success.` ?
Is this property very usefull ? because it seems to me we could avoid this Vec::new() here by resizing any way. do I miss something ?

srml/contracts/src/wasm/runtime.rs
```diff
@@ -344,12 +393,10 @@ define_env!(Env, <E: Ext>,
...
 			read_sandbox_memory_as(ctx, value_ptr, value_len)?;
-		let input_data = read_sandbox_memory(ctx, input_data_ptr, input_data_len)?;
 
-		// Grab the scratch buffer and put in its' place an empty one.
-		// We will use it for creating `EmptyOutputBuf` container for the call.
-		let scratch_buf = mem::replace(&mut ctx.scratch_buf, Vec::new());
-		let empty_output_buf = EmptyOutputBuf::from_spare_vec(scratch_buf);
+		// Read input data into the scratch buffer, then take ownership of it.
+		read_sandbox_memory_into_scratch(ctx, input_data_ptr, input_data_len)?;
+		let input_data = mem::replace(&mut ctx.scratch_buf, Vec::new());
```
thiolliere: the intereset of using scratch buffer here is to clean it at the same time ? why not read_sandbox_memory to input_data directly ?

srml/contracts/src/wasm/runtime.rs
```diff
@@ -454,33 +527,11 @@ define_env!(Env, <E: Ext>,
...
-				RuntimeToken::ReturnData(data_len)
-			)
-		{
-			GasMeterResult::Proceed => (),
-			GasMeterResult::OutOfGas => return Err(sandbox::HostError),
-		}
+		charge_gas(ctx.gas_meter, ctx.schedule, RuntimeToken::ReturnData(data_len))?;
+
+		read_sandbox_memory_into_scratch(ctx, data_ptr, data_len)?;
+		let output_buf = mem::replace(&mut ctx.scratch_buf, Vec::new());
```
thiolliere: same I don't really understand why it is using scratch buffer here

srml/contracts/src/wasm/runtime.rs
jimpo: The purpose is to reduce memory allocations. `read_sandbox_memory` creates a new `Vec`, but the `scratch_buffer` is already available and its contents will be thrown out anyway.

thiolliere: yes make sense

srml/contracts/src/wasm/runtime.rs
jimpo: The reason for the first replace is because of the borrow checker. We need a mutable reference to the scratch buffer and need to borrow `ctx` for `memory()`.

The second `replace` could be switched to `ctx.scratch_buf = buf;`, which is a bit more obvious.

thiolliere: can't you just access memory by the field directly ?

jimpo: Yeah, good point. Will change.

srml/contracts/src/tests.rs
pepyakin: I mean, we already have this code here which is certainly better than the status quo. Let's see how annoying it gets in the future. I think if we left some notes that would allow posterity to split it up it would greatly reduce amount of frustration

PR #3322, opened 2019-08-07
Offences reporting and slashing

tomusdrw:
The spec is available [here](https://research.web3.foundation/en/latest/polkadot/slashing/). 

This PR introduces infrastructure for tracking offences and enacting them, i.e. slashing. Actual reporting is implemented only for unresponsivness; detection, validation and filing the reports is out of scope for this PR and should be implemented in https://github.com/paritytech/substrate/pull/3314. There are some differences from the spec, namely:

1. Only reports for BABE equivocation, GRANDPA equivocation and unresponsivness are in place.
1. This doesn't have any notion of severity levels, and we always start a new era on if any of validators is slashed and we always disable a validator on an offence.
1. The reward ratio is a global scalar whereas in the spec it could depend on the actual misbehavior (e.g. for BABE equivocation the reporter [is rewarded](https://research.web3.foundation/en/latest/polkadot/slashing/#babe-equivocation) not from the total slashed amount but rather from an amount that would have been slashed with the offender count equals to 1).


core/sr-primitives/src/offence.rs
```diff
@@ -0,0 +1,128 @@
...
+		slash_fraction: &[Perbill],
+	);
+}
+
+/// A details about an offending authority for a particular kind of offence.
+#[derive(Clone, PartialEq, Eq, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub struct OffenceDetails<Reporter, Offender> {
+	/// The offending authority id
+	pub offender: Offender,
```
rphmeier: will this be the `IdentificationTuple` in practice?

core/sr-primitives/src/offence.rs
```diff
@@ -0,0 +1,128 @@
...
+	fn offenders(&self) -> Vec<Offender>;
+
+	/// The session index which is used for querying the validator set for the `slash_fraction`
+	/// function.
+	fn current_era_start_session_index(&self) -> u32; // TODO [slashing]: Should be a SessionIndex.
+
+	/// Return a validators count at the time when the offence took place.
+	fn validators_count(&self) -> u32;
+
+	/// A point in time when this offence happened.
```
rphmeier: how are offenses which happened long ago but were only caught recently handled?

srml/grandpa/src/lib.rs
```diff
@@ -376,3 +378,53 @@ impl<T: Trait> finality_tracker::OnFinalizationStalled<T::BlockNumber> for Modul
 		<Stalled<T>>::put((further_wait, median));
 	}
 }
+
+// TODO [slashing]: Integrate this.
+/// A grandpa equivocation offence report.
+#[allow(dead_code)]
+struct GrandpaEquivocationOffence<FullIdentification> {
+	/// A round in which the incident happened.
+	round: u64,
```
rphmeier: round resets to 0 every set. This should contain the Set ID too.

srml/grandpa/src/lib.rs
```diff
@@ -376,3 +378,53 @@ impl<T: Trait> finality_tracker::OnFinalizationStalled<T::BlockNumber> for Modul
...
+	) -> Perbill {
+		// the formula is min((3k / n)^2, 1)
+		let x = Perbill::from_rational_approximation(3 * offenders_count, validators_count);
+
+		// _ ^ 2
+		// For now, Perbill doesn't support taking the power of it. Until it does, do it manually.
+		// The conversion to u64 is performed to guarantee it fits.
+		// TODO: #3189 should fix this.
+		let x = x.into_parts();
+		let x = ((x as u64 * x as u64) / 1_000_000_000) as u32;
```
rphmeier: is it actually squaring the denominator correctly?

srml/staking/src/lib.rs
```diff
@@ -144,10 +145,10 @@
 //!
 //! decl_module! {
 //! 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
-//!			/// Report whoever calls this function as offline once.
-//! 		pub fn report_sender(origin) -> Result {
+//!			/// Slash a validator for an offence.
```
rphmeier:
```suggestion
//!		/// Slash a validator for an offence.
```

srml/grandpa/src/lib.rs
```diff
@@ -376,3 +378,53 @@ impl<T: Trait> finality_tracker::OnFinalizationStalled<T::BlockNumber> for Modul
...
 }
+
+// TODO [slashing]: Integrate this.
+/// A grandpa equivocation offence report.
+#[allow(dead_code)]
+struct GrandpaEquivocationOffence<FullIdentification> {
+	/// A round in which the incident happened.
+	round: u64,
+	/// The session index that starts an era in which the incident happened.
+	current_era_start_session_index: u32, // TODO [slashing]: Should be a SessionIndex.
```
rphmeier: it doesn't need to indicate the start of an era exactly; just the session where the offence occurred.

core/sr-primitives/src/offence.rs
tomusdrw: Yes.

core/sr-primitives/src/offence.rs
tomusdrw: If the node can produce the proof we will accept it, and afaik we will only produce proofs for not-pruned states initially (i.e. we will call into the runtime at specific block and generate the proof).
So we don't plan to store tries in any kind of DB initially.

srml/grandpa/src/lib.rs
tomusdrw: The offences module will use `(SessionIndex, Round)` as a key, do you think that's sufficient?

srml/grandpa/src/lib.rs
tomusdrw: Yeah, we've figured out that historical module will actually store the trie root for every session, so that should be renamed. CC @pepyakin

pepyakin: Fixed!

srml/grandpa/src/lib.rs
pepyakin: Good point, fixed.

core/sr-primitives/src/offence.rs
rphmeier: what I mean is about the point in time.

We may have a bonding period of 3 months but a rolling window is probably not going to be that long.

What happens when we catch an offense that happened almost 3 months ago?

srml/grandpa/src/lib.rs
rphmeier: Yeah, although I don't know how you'll go from set-id to session index

joepetrowski:
A nit, but a while ago we decided on US spellings. That would be _offense._

srml/babe/src/lib.rs
```diff
@@ -241,6 +248,56 @@ impl<T: Trait> session::ShouldEndSession<T::BlockNumber> for Module<T> {
 	}
 }
 
+// TODO [slashing]: @marcio use this, remove the dead_code annotation.
+/// A BABE equivocation offence report.
+#[allow(dead_code)]
+struct BabeEquivocationOffence<FullIdentification> {
```
joepetrowski: What does equivocation in BABE look like?
- Authoring blocks on a chain that conflicts with the finalized chain.
- Authoring two different blocks in the same slot.
- Authoring for a slot that you weren't slot leader.

pepyakin: According to the spec, an equivocation in BABE means that a validator produced more than one block in his slot.

gavofyork:
> A nit, but a while ago we decided on US spellings. That would be offense.

Actually, we decided on *Google* english.

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+	///
+	/// The list has no duplicates, so it is rather a set.
+	fn offenders(&self) -> Vec<Offender>;
+
+	/// The session index which is used for querying the validator set for the `slash_fraction`
+	/// function.
+	fn session_index(&self) -> SessionIndex;
+
+	/// Return a validators count at the time when the offence took place.
+	fn validators_count(&self) -> u32;
```
pepyakin: `validator_set_count` might be a better name for this to imply that this count has to do with the validator set and not with the report.

rphmeier: "Validators count" also isn't the right way to say that in english anyway.
"Validator count" would be fine

tomusdrw: Fixed.

srml/aura/src/lib.rs
```diff
@@ -265,42 +222,15 @@ impl<T: Trait> Module<T> {
...
 		let last_slot = last / slot_duration.clone();
-		let first_skipped = last_slot.clone() + One::one();
 		let cur_slot = now / slot_duration;
 
 		assert!(last_slot < cur_slot, "Only one block may be authored per slot.");
-		if cur_slot == first_skipped { return }
-
-		let skipped_slots = cur_slot - last_slot - One::one();
-
-		H::handle_report(AuraReport {
```
tomusdrw: Note that AuRa-specific reporting/slashing has now been removed. One can use `im-online` together with `AuRa` to slash offline validators.

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+	fn slash_fraction(
+		offenders_count: u32,
+		validator_set_count: u32,
+	) -> Perbill;
+}
+
+/// A trait for decoupling offence reporters from the actual handling of offence reports.
+pub trait ReportOffence<Reporter, Offender, O: Offence<Offender>> {
+	/// Report an `offence` and reward the `reporter`.
+	fn report_offence(reporter: Option<Reporter>, offence: O);
```
rphmeier: single reporter is a bit inflexible. i guess it should be possible to extend later

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/// A trait to take action on an offence.
+///
+/// Used to decouple the module that handles offences and
+/// the one that should punish for those offences.
+pub trait OnOffenceHandler<Reporter, Offender> {
+	/// A handler for offence of particular kind.
+	///
+	/// Note that this contains a list of all previous offenders
```
rphmeier: _all_?

rphmeier:
@gavofyork @joepetrowski 

image

Frankly, "google english" is useless because it is localized based on what they know about you :)

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+pub struct OffenceDetails<Reporter, Offender> {
+	/// The offending authority id
+	pub offender: Offender,
+	/// A number of times the authority was already reported for this offence.
+	///
+	/// Since we punish all authorities for that made the same offence in the past as well
+	/// we keep track the "age" of the report, so that the slashes can be lowered
+	/// in case the authority was already slashed in the past.
+	/// Note that we don't buffer slashes and instead use this approach.
+	pub count: OffenceCount,
```
rphmeier: is the intended purpose of this member to make it so an authority slashed for amount `f(a)` before, and now who would receive a slash of `f(b)` actually receives a slash of `f(b) - f(a)`?

i.e., not to over-step the slash amount function when someone is slashed multiple times

If so, wouldn't we also need to know how many other authorities had performed the same offence at the time of the prior slash?

srml/babe/src/lib.rs
```diff
@@ -241,6 +248,58 @@ impl<T: Trait> session::ShouldEndSession<T::BlockNumber> for Module<T> {
...
+	) -> Perbill {
+		// the formula is min((3k / n)^2, 1)
+		let x = Perbill::from_rational_approximation(3 * offenders_count, validator_set_count);
+
+		// _ ^ 2
+		// For now, Perbill doesn't support taking the power of it. Until it does, do it manually.
+		// The conversion to u64 is performed to guarantee it fits.
+		// TODO: #3189 should fix this.
+		let x = x.into_parts();
+		let x = ((x as u64 * x as u64) / (1_000_000_000 * 1_000_000_000)) as u32;
```
rphmeier: 1. this is likely to overflow
2. `1_000_000_000`^2 is only 1 order of magnitude away from `u64::max`. The granularity of fractions here is going to suck with integer division

srml/grandpa/src/lib.rs
```diff
@@ -376,3 +381,53 @@ impl<T: Trait> finality_tracker::OnFinalizationStalled<T::BlockNumber> for Modul
...
+	) -> Perbill {
+		// the formula is min((3k / n)^2, 1)
+		let x = Perbill::from_rational_approximation(3 * offenders_count, validator_set_count);
+
+		// _ ^ 2
+		// For now, Perbill doesn't support taking the power of it. Until it does, do it manually.
+		// The conversion to u64 is performed to guarantee it fits.
+		// TODO: #3189 should fix this.
+		let x = x.into_parts();
+		let x = ((x as u64 * x as u64) / (1_000_000_000 * 1_000_000_000)) as u32;
```
rphmeier: ditto.

srml/offences/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+		// Check if an offence is already reported for the offender authorities
+		// and otherwise stores that report.
+		let mut changed = false;
+		let all_offenders = <OffenceReports<T>>
+			::mutate(&O::ID, &(session, time_slot), |offending_authorities| {
+				for offender in offenders {
+					// TODO [slashing] This prevents slashing for multiple reports of the same kind at the same slot,
+					// note however that we might do that in the future if the reports are not exactly the same (dups).
+					// De-duplication of reports is tricky though, we need a canonical form of the report
+					// (for instance babe equivocation can have headers swapped).
```
rphmeier: well, this isn't actually difficult.

`footprint = babe_equivoation_footprint(who, slot_number)`.

srml/offences/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+	type OnOffenceHandler: OnOffenceHandler<Self::AccountId, Self::IdentificationTuple>;
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as Offences {
+		/// A mapping between unique `TimeSlots` within a particular session and the offence `Kind` into
+		/// a vector of offending authorities.
+		///
+		/// It's important that we store all authorities reported for an offence for any kind and
+		/// timeslot since the slashing will increase based on the length of this vec.
```
rphmeier: It's unlikely that we will have more than one offence at any given time-slot.
I understand that this PR isn't meant to have any rolling-window, but would you simply upgrade later by instead looking backwards + forwards (you probably need a notion of "now") from the time-slot of the actual offence?

srml/offences/src/lib.rs
tomusdrw: Does this function exist already? What if you issue 4 different headers in the same slot? I guess you should be slashed for all possible pairs, no?

core/sr-staking-primitives/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+//! approaches in general. Definitions related to sessions, slashing, etc go here.
+
+use rstd::vec::Vec;
+
+pub mod offence;
+
+/// Simple index type with which we can count sessions.
+pub type SessionIndex = u32;
+
+pub trait CurrentElectedSet<ValidatorId> {
```
gavofyork: missing docs

srml/offences/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+						details.count += 1;
+					}
+				}
+
+				offending_authorities.clone()
+			});
+
+		if !changed {
+			// The report contained only duplicates, so there is no need to slash again.
+			return
```
rphmeier: shouldn't fully-duplicate slashing reports return an error of some kind, so that the transaction-processing later on can reject the transaction?

in the event of an equivocation, we should have many many reports floating around the network. it would be useless to include all of them in blocks. We might want to include the first few to lower variance a bit.

there might be some possible use of the `TaggedTransactionQueue` to ensure that inclusion of one immediately invalidates others.

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+	/// This is used for looking up offences that happened at the "same time".
+	///
+	/// The timescale is abstract and doesn't have to be the same across different implementations
+	/// of this trait. The value doesn't represent absolute timescale though since it is interpreted
+	/// along with the `session_index`. Two offences considered to happen at the same time iff
+	/// both `session_index` and `time_slot` are equal.
+	///
+	/// As an example, for GRANDPA timescale could be a round number and for BABE it could be a slot
+	/// number. Note that for BABE the round number is reset each epoch.
+	fn time_slot(&self) -> TimeSlot;
```
gavofyork: why not `Self::TimeSlot`?

srml/staking/src/lib.rs
```diff
@@ -1564,3 +1470,80 @@ impl<T: Trait> SelectInitialValidators<T::AccountId> for Module<T> {
...
+		let mut remaining_imbalance = <NegativeImbalanceOf<T>>::zero();
+		let slash_reward_fraction = SlashRewardFraction::get();
+
+		for (details, slash_fraction) in offenders.iter().zip(slash_fraction) {
+			let stash = &details.offender.0;
+			let exposure = &details.offender.1;
+
+			// Early exit if validator is invulnerable.
+			if Self::invulnerables().contains(stash) {
+				return
```
rphmeier:
```suggestion
				continue
```

One offender being invulnerable shouldn't get the others off

core/sr-staking-primitives/src/offence.rs
gavofyork:
```suggestion
	type TimeSlot: Copy + Clone + Eq + PartialEq;
	fn time_slot(&self) -> Self::TimeSlot;
```

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+
+	/// A slash fraction of the total exposure that should be slashed for this
+	/// particular offence kind for the given parameters.
```
gavofyork:
```suggestion
	/// particular offence kind for the given parameters that happened at a singular point in time.
```

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+/// the one that should punish for those offences.
+pub trait OnOffenceHandler<Reporter, Offender> {
+	/// A handler for offence of particular kind.
+	///
+	/// Note that this contains a list of all previous offenders
+	/// as well. The implementer should cater for a case, where
+	/// the same authorities were reported for the same offence
+	/// in the past (see `OffenceCount`).
+	///
+	/// The vector of `slash_fraction` contains perbils
```
gavofyork:
```suggestion
	/// The vector of `slash_fraction` contains `Perbill`s
```

core/sr-staking-primitives/src/offence.rs
gavofyork: since offenders are reported one at a time, we know that previous slashes will account for `count - 1` slashes to all previous offenders, no?

srml/offences/src/lib.rs
rphmeier: @tomusdrw 

We've decided not to slash for every equivocation - only the first. The reason being that you can equivocate an infinite amount of times for relatively low practical cost, and then you would easily overwhelm the chain processing all the reports.

srml/aura/src/lib.rs
gavofyork: since this slashing code is strictly better than im_online at detecting "malicious" offlineness, then it should be reintegrated properly in a follow-up PR. maybe make an issue for that.

core/sr-staking-primitives/src/offence.rs
rphmeier: My understanding is that `srml-offences` requires all offences to use the same `TimeSlot` type, since it sorts things in storage by time-slot.

It could use an encoding of generic time-slot type instead.

srml/offences/src/lib.rs
gavofyork: > It's unlikely that we will have more than one offence at any given time-slot.

A timeslot is a flexible notion that should inherently quantise to naturally group together typical cases of co-offending.

e.g. for Grandpa, the timeslot would be concerning one specific vote. If there were multiple offenders for that same vote then we slash them each more. If the offences were spread between different votes, then the danger is far lessened and they'd each get away with a lighter punishment. This is the case even if the voting still within a limited window of time.

For Babe and ImOnline, then timeslot could perhaps be the session/epoch quantisation.

For Parachains, then it would naturally be parachain block number.

joepetrowski:
> @gavofyork @joepetrowski
> 
> Frankly, "google english" is useless because it is localized based on what they know about you :)
> I'm not sure "Alphabet Corporation's Gavin Wood Data Profile English" is sustainable or even immutable...
> 
> (not to say that we should necessarily change the naming in the PR - but this is something we should seek clarity on in the near future)

I often use [ngram](https://books.google.com/ngrams/graph?content=offense%2Coffence&case_insensitive=on&year_start=1800&year_end=2008&corpus=15&smoothing=3&share=&direct_url=t4%3B%2Coffense%3B%2Cc0%3B%2Cs0%3B%3Boffense%3B%2Cc0%3B%3BOffense%3B%2Cc0%3B.t4%3B%2Coffence%3B%2Cc0%3B%2Cs0%3B%3Boffence%3B%2Cc0%3B%3BOffence%3B%2Cc0) as the decider. In this case "offence" is used more but it's quite close.

srml/offences/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+		// and otherwise stores that report.
+		let mut changed = false;
+		let all_offenders = <OffenceReports<T>>
+			::mutate(&O::ID, &(session, time_slot), |offending_authorities| {
+				for offender in offenders {
+					// TODO [slashing] This prevents slashing for multiple reports of the same kind at the same slot,
+					// note however that we might do that in the future if the reports are not exactly the same (dups).
+					// De-duplication of reports is tricky though, we need a canonical form of the report
+					// (for instance babe equivocation can have headers swapped).
+					if !offending_authorities.iter().any(|details| details.offender == offender) {
```
gavofyork: this is O(N**2) - better to retain sorted lists of offenders.

srml/offences/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+			// The report contained only duplicates, so there is no need to slash again.
+			return
+		}
+
+		// The report is not a duplicate. Deposit an event.
+		Self::deposit_event(Event::Offence(O::ID, session, time_slot));
+
+		let offenders_count = all_offenders.len() as u32;
+		let expected_fraction = O::slash_fraction(offenders_count, validator_set_count);
+		let slash_perbil = all_offenders
```
gavofyork:
```suggestion
		let slash_perbill = all_offenders
```

gavofyork:
@rphmeier @joepetrowski

> I'm not sure "Alphabet Corporation's Gavin Wood Data Profile English" is sustainable or even immutable...

I was using a private browsing tab over a non-UK EU VPN.

To ensure I wasn't getting an auto-anglicised view though, I retried with Argentinian, Algerian and Vietnamese VPNs (for each of three non-English prevalent continents). I chose the specific countries each because of their lack of historical/imperial ties to both the US and the UK (e.g. I didn't select Guyana, Kenya and Singapore) and thus to try to get what Google believe is a vaguely impartial "international" spelling. In all three cases, it returned the same as my original, correcting "offense" to "offence".

I had to use a specifically US VPN before I was able to get your screenshot.

srml/babe/src/lib.rs
pepyakin: Done in https://github.com/paritytech/substrate/pull/3322/commits/0b57b0dd794ecc8947736c24f4c8514bb0352fb3

core/sr-staking-primitives/src/offence.rs
rphmeier: > A number of times the authority was already reported for this offence.

I understand this documentation line to mean only the `count` of _reports_ for the same authority `Self: OffenceDetails` is implicitly associated with.

If it is actually the `count` of slashes the authority has received (whether as a result of its own misbehavior or others'), then yes.

core/sr-staking-primitives/src/lib.rs
pepyakin: Updated!

srml/offences/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+					Perbill::from_parts(perbil)
+				} else {
+					expected_fraction.clone()
+				}
+			})
+			.collect::<Vec<_>>();
+
+		T::OnOffenceHandler::on_offence(
+			&all_offenders,
+			&slash_perbil
```
rphmeier:
```suggestion
			&slash_perbil,
```

We may want a separate function `reward_perbill` as the research document says not to adjust the proportion of the reward given to the rewarder (i.e. always treat `count` as 1) in some cases.

core/sr-staking-primitives/src/offence.rs
pepyakin: Yeah, this was our thinking at the moment.

core/sr-staking-primitives/src/offence.rs
rphmeier: (although, if that is the case, then the `count` parameter will be the same across all offenders in the same time-slot and should probably be refactored elsewhere)

srml/staking/src/lib.rs
pepyakin: Good point

srml/offences/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+		Self::deposit_event(Event::Offence(O::ID, session, time_slot));
+
+		let offenders_count = all_offenders.len() as u32;
+		let expected_fraction = O::slash_fraction(offenders_count, validator_set_count);
+		let slash_perbil = all_offenders
+			.iter()
+			.map(|details| {
+				if details.count > 1 {
+					let previous_fraction = O::slash_fraction(
+						offenders_count.saturating_sub(details.count - 1),
```
rphmeier: Is this actually the previous fraction of slashing?

`details.count` is initialized to `0` and then incremented to `1` the first time the authority is slashed. But the first slashing is based on `offenders_count`, not on `details.count`. This line seems to assume that the prior slash can be computed based on the `details.count`, but really what we want to store is the number of offenders the last time this offender was slashed.

srml/offences/src/lib.rs
tomusdrw: Yes, we should definitely use `TaggedTransactionQueue` to deduplicate `UnsignedTransaction`s containing the same reports. (afair we already do that for `imonline` reports).
Afaict, with babe/grandpa we will use regular `SignedTransactions` to report equivocation, so the callers need to pay for the execution anyway. As future optimization we should probably change the reporting code to avoid generating the report if we find that there is already some other report for this offence in the pool.

srml/staking/src/lib.rs
pepyakin: Fixed and add a test in https://github.com/paritytech/substrate/pull/3322/commits/53b2c6449b41ced1dd3cfc5330e59e004187af78

srml/offences/src/lib.rs
rphmeier: I don't think we'll be able to dis-incentivize front-running. The way we've structured things now, there's going to be a mad rush every time there's a misbehavior and variance over whether you actually get paid out over a report will be really high.

There's no point in including such wasted transactions in blocks, though.

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+
+/// The kind of an offence, is a byte string representing some kind identifier
+/// e.g. `b"im-online:offlin"`, `b"babe:equivocatio"`
+// TODO [slashing]: Is there something better we can have here that is more natural but still
+// flexible? as you see in examples, they get cut off with long names.
+pub type Kind = [u8; 16];
+
+/// Number of times the offence of this authority was already reported in the past.
+///
+/// Note we don't buffer offence reporting so everytime we see a new offence
```
joepetrowski:
```suggestion
/// Note we don't buffer offence reporting so every time we see a new offence
```

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+pub trait Offence<Offender> {
+	/// Identifier which is unique for this kind of an offence.
+	const ID: Kind;
+
+	/// The list of all offenders involved in this incident.
+	///
+	/// The list has no duplicates, so it is rather a set.
+	fn offenders(&self) -> Vec<Offender>;
+
+	/// The session index which is used for querying the validator set for the `slash_fraction`
```
joepetrowski:
```suggestion
	/// The session index that is used for querying the validator set for the `slash_fraction`
```

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+	/// Return a validator set count at the time when the offence took place.
+	fn validator_set_count(&self) -> u32;
+
+	/// A point in time when this offence happened.
+	///
+	/// This is used for looking up offences that happened at the "same time".
+	///
+	/// The timescale is abstract and doesn't have to be the same across different implementations
+	/// of this trait. The value doesn't represent absolute timescale though since it is interpreted
+	/// along with the `session_index`. Two offences considered to happen at the same time iff
```
joepetrowski:
```suggestion
	/// along with the `session_index`. Two offences are considered to happen at the same time iff
```

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+impl<Reporter, Offender, O: Offence<Offender>> ReportOffence<Reporter, Offender, O> for () {
+	fn report_offence(_reporter: Option<Reporter>, _offence: O) {}
+}
+
+/// A trait to take action on an offence.
+///
+/// Used to decouple the module that handles offences and
+/// the one that should punish for those offences.
+pub trait OnOffenceHandler<Reporter, Offender> {
+	/// A handler for offence of particular kind.
```
joepetrowski:
```suggestion
	/// A handler for an offence of a particular kind.
```

core/sr-staking-primitives/src/offence.rs
```diff
@@ -0,0 +1,147 @@
...
+	/// The offending authority id
+	pub offender: Offender,
+	/// A number of times the authority was already reported for this offence.
+	///
+	/// Since we punish all authorities for that made the same offence in the past as well
+	/// we keep track the "age" of the report, so that the slashes can be lowered
+	/// in case the authority was already slashed in the past.
+	/// Note that we don't buffer slashes and instead use this approach.
+	pub count: OffenceCount,
+	/// A list of reporters of offences of this authority id. Possibily empty where there is no
```
joepetrowski:
```suggestion
	/// A list of reporters of offences of this authority ID. Possibly empty where there are no
```

srml/babe/src/lib.rs
```diff
@@ -241,6 +248,52 @@ impl<T: Trait> session::ShouldEndSession<T::BlockNumber> for Module<T> {
...
+/// When a validator released two or more blocks at the same slot.
+#[allow(dead_code)]
+struct BabeEquivocationOffence<FullIdentification> {
+	/// A babe slot number in which this incident happened.
+	slot: u64,
+	/// The session index in which the incident happened.
+	session_index: SessionIndex,
+	/// The size of the validator set at the time of the offence.
+	validator_set_count: u32,
+	/// The authority which produced the equivocation.
```
joepetrowski:
```suggestion
	/// The authority that produced the equivocation.
```

srml/im-online/src/lib.rs
```diff
@@ -162,6 +167,17 @@ pub trait Trait: system::Trait + session::Trait {
 	/// A extrinsic right from the external world. This is unchecked and so
 	/// can contain a signature.
 	type UncheckedExtrinsic: ExtrinsicT<Call=<Self as Trait>::Call> + Encode + Decode;
+
+	/// A type that gives us ability to submit unresponsivness offence reports.
```
joepetrowski:
```suggestion
	/// A type that gives us the ability to submit unresponsiveness offence reports.
```

srml/im-online/src/lib.rs
```diff
@@ -422,3 +477,73 @@ impl<T: Trait> srml_support::unsigned::ValidateUnsigned for Module<T> {
 		TransactionValidity::Invalid(0)
 	}
 }
+
+/// An offense which is filed if a validator didn't send a heartbeat message.
+pub struct UnresponsivnessOffence<Offender> {
```
joepetrowski: This should change to "Unresponsiv**e**ness"

srml/im-online/src/lib.rs
```diff
@@ -422,3 +477,73 @@ impl<T: Trait> srml_support::unsigned::ValidateUnsigned for Module<T> {
...
+	/// The current session index in which we report the unresponsive validators.
+	///
+	/// It acts as a time measure for unresponsivness reports and effectively will always point
+	/// at the end of the session.
+	session_index: SessionIndex,
+	/// The size of the validator set in current session/era.
+	validator_set_count: u32,
+	/// Authorities which were unresponsive during the current era.
```
joepetrowski: "Authorities that didn't send a heartbeat at the beginning of the current era."? It's confusing to me to have past tense and then "during the current era", but I understand that the heartbeat should have been sent earlier.
```suggestion
	/// Authorities that were unresponsive during the current era.
```

srml/im-online/src/lib.rs
```diff
@@ -422,3 +477,73 @@ impl<T: Trait> srml_support::unsigned::ValidateUnsigned for Module<T> {
 		TransactionValidity::Invalid(0)
 	}
 }
+
+/// An offense which is filed if a validator didn't send a heartbeat message.
```
joepetrowski:
```suggestion
/// An offense that is filed if a validator didn't send a heartbeat message.
```

core/sr-staking-primitives/src/offence.rs
gavofyork: i don't think that's a great excuse for clobbering type information.

putting on an `Encode` bound is trivial, and in reality, it'll end up being much more compact and efficient by using the real types (which are exclusively 32-bit and 64-bit).

they should only be coerced to a heavy-weight anonymous type like `u128` when they actually need to be compared to each other, and in that case you'd still more likely want to require an `Into<TimeSlot>` bound on an associated type rather than forcing the caller to do an ugly conversion.

srml/im-online/src/lib.rs
```diff
@@ -422,3 +477,73 @@ impl<T: Trait> srml_support::unsigned::ValidateUnsigned for Module<T> {
 		TransactionValidity::Invalid(0)
 	}
 }
+
+/// An offense that is filed if a validator didn't send a heartbeat message.
```
lsaether:
```suggestion
/// An offence that is filed if a validator didn't send a heartbeat message.
```

srml/offences/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+		// The amount new offenders are slashed
+		let new_fraction = O::slash_fraction(offenders_count, validator_set_count);
+		// The amount previous offenders are slashed additionally.
+		//
+		// Since they were slashed in the past, we slash by:
+		// x = (new - prev) / (1 - prev)
+		// because:
+		// Y = X * (1 - prev)
+		// Z = Y * (1 - x)
+		// Z = X * (1 - new)
```
rphmeier: This logic is all correct, but our main concern was whether per-bill can handle it. 
To keep the decimal representation from going out of control, we could always round the `slash_fraction` to the nearest 1000th or 10,000th. 

We will then incur some rounding error but are at much lower risk of overflow.

srml/grandpa/src/lib.rs
```diff
@@ -411,8 +420,8 @@ impl<FullIdentification: Clone> Offence<FullIdentification> for GrandpaEquivocat
 		self.validator_set_count
 	}
 
-	fn time_slot(&self) -> TimeSlot {
-		self.round as TimeSlot
+	fn time_slot(&self) -> GrandpaTimeSlot {
```
tomusdrw: I rather prefer `Self::TimeSlot` - less alterations if we change the type or rename it.

srml/babe/src/lib.rs
```diff
@@ -279,8 +280,8 @@ impl<FullIdentification: Clone> Offence<FullIdentification> for BabeEquivocation
 		self.validator_set_count
 	}
 
-	fn time_slot(&self) -> TimeSlot {
-		self.slot as TimeSlot
+	fn time_slot(&self) -> u64 {
```
tomusdrw: `s/u64/Self::TimeSlot`

srml/im-online/src/lib.rs
```diff
@@ -506,8 +507,8 @@ impl<Offender: Clone> Offence<Offender> for UnresponsivenessOffence<Offender> {
 		self.validator_set_count
 	}
 
-	fn time_slot(&self) -> TimeSlot {
-		self.session_index as TimeSlot
+	fn time_slot(&self) -> SessionIndex {
```
tomusdrw: same

srml/offences/src/lib.rs
```diff
@@ -159,8 +161,111 @@ impl<T: Trait, O: Offence<T::IdentificationTuple>> ReportOffence<T::AccountId, T
...
+	fn report_id<O: Offence<T::IdentificationTuple>>(
+		time_slot: &O::TimeSlot,
+		offender: &T::IdentificationTuple,
+	) -> ReportIdOf<T> {
+		(O::ID, time_slot.encode(), offender).using_encoded(T::Hashing::hash)
+	}
+
+	/// Goes through all offenders in a report and record each unique offence updating indexes.
+	///
+	/// Returns `true` if there were unique reports.
```
tomusdrw: Seems it rather returns a set of new offenders?

srml/offences/src/lib.rs
```diff
@@ -159,8 +161,111 @@ impl<T: Trait, O: Offence<T::IdentificationTuple>> ReportOffence<T::AccountId, T
...
+	fn collect_unique_reports<O: Offence<T::IdentificationTuple>>(
```
tomusdrw: I think I'd rather rename it so something like: `store_new_offenders` or at least `collect_new/unique_offenders`

srml/offences/src/lib.rs
```diff
@@ -159,8 +161,111 @@ impl<T: Trait, O: Offence<T::IdentificationTuple>> ReportOffence<T::AccountId, T
...
+		}
+
+		new_offenders
+	}
+}
+
+/// An auxilary struct for dealing with caches of report indexes localized for a specific offence
+/// kind.
+///
+/// It is expected that access to the indexes will be only performed through this struct while it
```
tomusdrw: Should we just say that it's responsible for aggregating storage writes?

srml/offences/src/mock.rs
```diff
@@ -109,33 +113,43 @@ pub type System = system::Module<Runtime>;
 
 pub const KIND: [u8; 16] = *b"test_report_1234";
 
+/// Returns all offence details for the specific `kind` happened at the specific time slot.
+pub fn offence_reports(kind: Kind, time_slot: u128) -> Vec<OffenceDetails<u64, u64>> {
+	<crate::ConcurrentReportsIndex<Runtime>>::get(&kind, &time_slot.encode())
+		.into_iter()
+		.map(|report_id| <crate::Reports<Runtime>>::get(&report_id).unwrap())
```
tomusdrw: it's test code, but I think still `expect` could provide a better error message.

srml/offences/src/lib.rs
```diff
@@ -75,57 +89,45 @@ decl_module! {
...
-				// (for instance babe equivocation can have headers swapped).
-				if !offending_authorities.iter().any(|details| details.offender == offender) {
-					new_offenders.insert(offender.clone());
-					offending_authorities.push(OffenceDetails {
-						offender,
-						reporters: reporters.clone().into_iter().collect(),
-					});
-				}
-			}
+		let mut cache = ReportIndexCache::load(&time_slot);
```
tomusdrw: Maybe better to store `new_offenders` in the `ReportIndexCache` and just return it from `collect_*` function?

srml/offences/src/lib.rs
```diff
@@ -159,8 +161,111 @@ impl<T: Trait, O: Offence<T::IdentificationTuple>> ReportOffence<T::AccountId, T
...
+			.binary_search_by_key(&time_slot, |&(ref when, _)| when)
+		{
+			Ok(idx) => idx,
+			Err(idx) => idx,
+		};
+		self.same_kind_reports.insert(idx, (time_slot.clone(), report_id));
+	}
+
+	/// Dump the indexes to the storage.
+	fn save(self) {
```
tomusdrw: Since you really have to call `save` at the end, maybe let's add `must_use` annotation.

srml/offences/src/lib.rs
```diff
@@ -159,8 +161,111 @@ impl<T: Trait, O: Offence<T::IdentificationTuple>> ReportOffence<T::AccountId, T
...
+		new_offenders
+	}
+}
+
+/// An auxilary struct for dealing with caches of report indexes localized for a specific offence
+/// kind.
+///
+/// It is expected that access to the indexes will be only performed through this struct while it
+/// exists.
+struct ReportIndexCache<T: Trait, O: Offence<T::IdentificationTuple>> {
```
tomusdrw: I'm not convinced to call it a `cache`, the struct is strictly responsible for:
1. Processing new reports (new offenders)
2. Providing the list of new_offenders and all offenders of the same time slot (`concurrent_reports`)
3. Persisting the list at the end.

I'd merge this logic with `collect_*` thing, and just do all alterations here, while only keeping the offence reporting (event) and perbill slashing in the main part.

srml/offences/src/lib.rs
pepyakin: yeah, a merge brainfart!

PR #3325, opened 2019-08-07
Add a small ban when we get disconnected from a node

tomaka:
Adds a random 5 to 10 seconds ban when the remote disconnects us.
During this period, we won't connect to that node again even if the PSM requests a connection. It's only a ban for the outgoing side, and we will still allow incoming connections though.

Why is the duration random? Because if an accidental disconnect happens, both sides will ban each other, and we preferably want to avoid a situation where they dial each other simultaneously.


folsen:
Code LGTM, but why the ban at all?

arkpar:
Currently we try reconnecting immediately after other party explicitly disconnects us. This leads to a lot of connection attempts that end up in immediate disconnect, since the other party has banned us. 

BtW @tomaka shouldn't the other side just not accept the connection if our reputation with them is too low?

tomaka:
> BtW @tomaka shouldn't the other side just not accept the connection if our reputation with them is too low?

The PSM only controls the Substrate-specific protocols, and not Kademlia for example.

What happens (I think) in details is:

- The remote accepts our connection.
- We open a Substrate substream towards this remote. There is no way in the protocol for the remote to refuse this.
- The remote sees that our reputation is too low and closes it.
- In order to avoid a race condition when reopening the substream, we added a hack a while ago where closing the Substream substream also closes the entire connection.

Points 2 and 4 are why I'd like to switch to something else.

arkpar:
Apparently this breaks one of GRANDPA tests somehow.

andresilva:
Merging/rebasing master should fix the test now.

andresilva:
CI passed but I restarted the tests to be a little bit more sure.

PR #3326, opened 2019-08-07
Improve console output

### LINKED ISSUE #3269: Logging sync=trace is way too verbose
`sync=trace` is extremely spammy, because we print log messages in getters that are called every time the network future is getting polled.

This should be reworked a bit.

arkpar:
Fixes #3269 
* Cache idle state in sync so that requests are not collected with every network poll.
* Block import are printed in the console by default
* Removed some additional console spam.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4416,8 +4416,8 @@ dependencies = [
 name = "substrate-consensus-babe-primitives"
 version = "2.0.0"
 dependencies = [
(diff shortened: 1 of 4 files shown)


arkpar:
`info` is printed to stderr by default. We are already using it for important messages that user should see. I agree that `is_idle` is somewhat fragile. Having the request diagnostic is fairly important though as it already helped finding a few issues in the past.

PR #3332, opened 2019-08-07
Fix for network test failing randomly

arkpar:


core/network/src/protocol/sync.rs
```diff
@@ -198,7 +198,9 @@ pub struct Status<B: BlockT> {
 	/// Target sync block number.
 	pub best_seen_block: Option<NumberFor<B>>,
 	/// Number of peers participating in syncing.
-	pub num_peers: u32
+	pub num_peers: u32,
```
bkchr: Why are these values `u32` and not `usize`?

core/network/src/test/mod.rs
```diff
@@ -591,6 +591,9 @@ pub trait TestNetFactory: Sized {
 		// Return `NotReady` if there's a mismatch in the highest block number.
 		let mut highest = None;
 		for peer in self.peers().iter() {
+			if peer.is_major_syncing() || peer.network.num_queued_blocks() != 0 {
+				return Async::NotReady
```
bkchr: Is that really okay? Are we sure that something wakes up the task to continue?

arkpar: Queue events wake up the task I believe.

core/network/src/protocol/sync.rs
arkpar: Because they don't depend on machine pointer size. `u32` is more than enough.

PR #3334, opened 2019-08-07
Clean up CLI interface

bkchr:
- Removes `--validator` and `--grandpa-voter`
- Make `--alice` etc work without `--dev`

### DIFF
--- a/core/cli/src/lib.rs
+++ b/core/cli/src/lib.rs
@@ -471,23 +471,16 @@ where
 		is_dev,
 	)?;
 
-	fill_transaction_pool_configuration::<F>(
-		&mut config,
-		cli.pool_config,
-	)?;
+	fill_transaction_pool_configuration::<F>(&mut config, cli.pool_config)?;
 
-	config.dev_key_seed = cli.keyring.account
-		.map(|a| format!("//{}", a));
+	config.dev_key_seed = cli.keyring.account.map(|a| format!("//{}", a));
 
 	let rpc_interface: &str = if cli.rpc_external { "0.0.0.0" } else { "127.0.0.1" };
 	let ws_interface: &str = if cli.ws_external { "0.0.0.0" } else { "127.0.0.1" };
 
-	config.rpc_http = Some(
-		parse_address(&format!("{}:{}", rpc_interface, 9933), cli.rpc_port)?
-	);
-	config.rpc_ws = Some(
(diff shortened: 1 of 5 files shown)


xlc:
Does this means is no longer needed? Just need to pass `--key`? Readme and other docs will need change if that's the case.

bkchr:
`--key` doesn't exist anymore

xlc:
So all configured via RPC?

bkchr:
Yes that is correct

tomaka:
If I understand correctly, all non-light nodes are now validators?

bkchr:
In theory yes. You don't know if there is a key or of there will be a key for authoring in the keystore.

tomaka:
My only slight worry is that we're sending our `Roles` on the network, and pragmatically speaking we might send `AUTHORITY` without actually being an authority.
In practice, though, this is only used to differentiate light vs non-light at the moment.

arkpar:
Is there a message printed on startup if the key is there? Would be useful.

gavofyork:
on startup doesn't make sense - it should be whenever it newly uses a session key. without a richer consensus interface (like below) then we really have no way of interpreting what's going on in a coherent way outside of the consensus, so it would just be up to the individual consensus modules (and other SRML/offchain modules parachains and im_online) to report individually.

```
/// Type of source for block sealing. Different consensus algorithms have different sealing
/// methods; for PoW it'll be a miner or mining instance. For PoA/PoS it'll be a key type.
pub trait SealingSource {
	/// Human-readable description of this general class.
	///
	/// e.g. `"ProgPoW"`, `"S/R 25519 Key"`.
	const SEALER_TYPE: &'static str;
	/// Human-readable description of this specific instance.
	///
	/// e.g. `"OpenCL GPU Radeon 8970 @ slot 1"`, `"5Gav1mbbo348grBREHTYgevvf4Gfe5GFE4"`.
	fn format(&self) -> String;
}

impl<T: app_crypto::AppPublic> SealingSource for T {
	const SEALER_TYPE: &'static str = "Key";
	fn format(&self) -> String { use primitives::crypto::Ss58Codec; self.to_ss58check() }
}
```

bkchr:
> My only slight worry is that we're sending our `Roles` on the network, and pragmatically speaking we might send `AUTHORITY` without actually being an authority.
> In practice, though, this is only used to differentiate light vs non-light at the moment.

In the future we will have a `--no-validator` flag, but for the `Roles` we maybe need to reduce them.

PR #3336, opened 2019-08-08
Benchmark execution times of runtime calls using Wasm execution.

jimpo:
These benchmarks were useful in measuring the impact of #3011.


bkchr:
We already have such benchmarks here: https://github.com/paritytech/substrate/blob/master/core/sr-api-macros/benches/bench.rs

bkchr:
The runtime api uses the same api internally as you are using. I'm not aware of any full block import benchmark.

kianenigma:
I have a question: How close are these to a realistic execution? 

(I am pretty sure I asked this in some technical chat but have no memory of what was the answer): How often do we, in reality, spawn a new executor? 

About the code: we also directly benchmark executor in `node/executor`. If you really care about something within the executor, then it makes the most sense to actually do it from there imo.

bkchr:
As he wanted to measure the impact of the runtime caching, it is more like a high level benchmark anyway.

pepyakin:
As an aside, I wonder if `test-runtime` would be a better place than `sr-api-macros` for housing the benches?

jimpo:
Closing for now until I have time to rework it.

PR #3337, opened 2019-08-08
Adds `author_rotateKeys` RPC

bkchr:
`author_rotateKeys` will rotate all session keys, store them into the
keystore and returns their encoded public keys as result. These encoded
public keys can be used directly to send the transaction to the chain.

### DIFF
--- a/core/test-runtime/src/lib.rs
+++ b/core/test-runtime/src/lib.rs
@@ -436,7 +436,7 @@ fn code_using_trie() -> u64 {
 
 impl_opaque_keys! {
 	pub struct SessionKeys {
-		#[id(key_types::SR25519)]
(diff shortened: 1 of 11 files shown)


core/test-runtime/src/lib.rs
```diff
@@ -434,6 +434,15 @@ fn code_using_trie() -> u64 {
 	} else { 103 }
 }
 
+impl_opaque_keys! {
+	pub struct SessionKeys {
+		#[id(key_types::SR25519)]
+		pub ed25519: ed25519::AppPublic,
+		#[id(key_types::SR25519)]
```
rphmeier: both key types have the SR25519 ID

PR #3343, opened 2019-08-08
Introduce AppVerify

gavofyork:
This trait works just like `Verify` except on `AppKey` types. I'd like for it all just to be `Verify`, but Rust's trait rules concerning upstream changes mean it can't happen. This is a simple
workaround needed for modules using the new `app_crypto` types to use `Verify` automatically based upon their inner crypto.

### DIFF
--- a/core/sr-primitives/src/traits.rs
+++ b/core/sr-primitives/src/traits.rs
@@ -35,6 +35,7 @@ use rstd::ops::{
 	Add, Sub, Mul, Div, Rem, AddAssign, SubAssign, MulAssign, DivAssign,
 	RemAssign, Shl, Shr
 };
+use crate::AppKey;
 
 /// A lazy value.
 pub trait Lazy<T: ?Sized> {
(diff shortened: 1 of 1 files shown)


bkchr:
FWIW:
```
impl<T: RuntimeAppPublic> Verify for T::Signature {
    type Signer = T;
    fn verify<L: Lazy<[u8]>>(&self, mut msg: L, signer: &Self::Signer) -> bool {
        signer.verify(msg, self)
    }
}
```

And then we don't need any new trait and looks way simpler^^

gavofyork:
that won't work as it is. some sort of `RuntimeAppSignature` type would be needed, i think:

```
error[E0207]: the type parameter `T` is not constrained by the impl trait, self type, or predicates
  --> core/sr-primitives/src/traits.rs:75:6
   |
75 | impl<T: RuntimeAppPublic> Verify for T::Signature {
   |      ^ unconstrained type parameter

```

Issue #3341, opened 2019-08-08
StorageMap::swap for linked_map breaks enumeration


thiolliere:
because the next and previous are not sync.

```diff
diff --git a/srml/support/src/lib.rs b/srml/support/src/lib.rs
index abc9d7d21..befaaf61f 100644
--- a/srml/support/src/lib.rs
+++ b/srml/support/src/lib.rs
@@ -342,6 +342,17 @@ mod tests {
                });
        }
 
+       #[test]
+       fn linked_map_swap_works() {
+               with_externalities(&mut new_test_ext(), || {
+                       OptionLinkedMap::insert(1, 1);
+                       OptionLinkedMap::insert(2, 1);
+                       assert_eq!(OptionLinkedMap::enumerate().count(), 2);
+                       OptionLinkedMap::swap(1, 2);
+                       assert_eq!(OptionLinkedMap::enumerate().count(), 2);
+               });
+       }
+
        #[test]
        fn linked_map_basic_insert_remove_should_work() {
                with_externalities(&mut new_test_ext(), || {
```

thiolliere:
I think StorageMap and linkedStorageMap should be in two different trait but I'm working on such a PR because this intrication can result in lots of error. also for instance: decl_storage override default method of generate::StorageMap for its implementation for linked_map

bkchr:
Is this fixed now? I think so?

thiolliere:
yes

PR #3344, opened 2019-08-08
`srml_support` storage reorganize

thiolliere:
Done:
* removal of generators: previously ppl were writing into `StorageOverlay` in add_extra_genesis. But now we are mostly using `with_storage(&mut storage, || call to storagemap function directly)` so this ability to write to storage overlay is useless and new method should be prefered.

* add_extra_genesis::build have only one argument config instead of 3, with_storage is called by caller. in this closure you now only have to write using traditionnal method (either StorageMap traits or function available in sr_io).

* split between StorageMap and StorageLinkedMap: previously both were intricated and decl_storage macro was overriding default method to make linked_map possible. Also this leads to issue in regards to method such as append indeed these method might be only wanted on one and not the other. This intrication was also error prone as we can see for swap default implementation was not override in decl_storage https://github.com/paritytech/substrate/issues/3341

* Merge of EnumerableStorage and StorageLinkedMap: as there is no longer necessity to make 2 traits

* Merge of AppendableStorageMap and StorageMap: same.

* introduction of new generators: small trait implemented by decl_storage and storage_items. The new generators are put in storage/generator. And StorageMap, StorageValue etc.. are implemented for them. This allow to define most of the logic out of procedural macro using a readable auditable implementation.

* Modification of StorageValue::key and StorageMap::key_for: they now return the key in top storage, which make more sense. Previously it returned the key in storage but not hashed. At some point I think we should remove those as in the future a storageValue can actually work on child_storage so this function would not have that much meaning.

* Modification of Storage[Value,Map,DoubleMap]::append. previously calling append on a unexisting value was calling was appending on the type default value. But decl_storage allow to define your own default. For me it make sense that this defined default is whenever there is no value (the strict definition of this functionality should be: the default provided is used whenever there is no value in storage) thus. appending on no value should append on this default instead of type default.

* Swap is not implemented for linked_map.
  EDIT: finally it is implemented as it was solved on master and it can have some usage for performance reason.

Next work in another PR:

* maybe some more clean up in decl_storage.
* implement decodelen is straightforward and doesn't require any decl_storage modification now.
* maybe storage::hashed module should be removed entirely.
* make decl_storage allowing to work on child storage instead of top storage. (just a new associated type in the new generators)
* traits should make use of codec::EncodeLike now
* I want to improve documentation of storages, how it the final key computed.
* make use of span in procedural macro

CRITIC:
* separate storagemap and storagelinkedmap have one drawback: one can't build abstraction on top of storagemap and expect user to provide a storagemap or storagelinkedmap type.
  the separation was needed so each traits can be implemented on their respective generator.
  and generator was useful in order to have the implementation in plain rust instead of procedural macro.
* this PR can be divided into some more minimal ones and I will work on that:
  * put with_storage inside build_storage not inside closure.
  * removes the ability to work on hashmaps in srml-support as we will use with_storage everywhere now
  * key return hashed key and not unhashed key.
  * split of trait StorageMap into StorageLinkedMap and StorageMap, this might be controversial, and in the end we might keep StorageMap only but then I will argue for no default implementation (as this default implementation breaks linkedmap assumptions) and provide a way for it be written correctly in procedural macro by writing functions srml-support maybe, we'll see.


srml/support/src/storage/generator/double_map.rs
```diff
@@ -0,0 +1,170 @@
...
+	type Query;
+
+	/// Hasher for the first key.
+	type Hasher1: StorageHasher;
+
+	/// Hasher for the second key.
+	type Hasher2: StorageHasher;
+
+	/// Get the prefix key in storage.
+	fn key1_prefix() -> &'static [u8];
```
jimpo: nit: I think it's fine to call this `prefix` -- it prefixes both keys.

srml/support/src/storage/generator/linked_map.rs
```diff
@@ -0,0 +1,270 @@
...
+
+	fn remove<KeyArg: Borrow<K>>(key: KeyArg) {
+		G::take(key);
+	}
+
+	fn mutate<KeyArg: Borrow<K>, R, F: FnOnce(&mut Self::Query) -> R>(key: KeyArg, f: F) -> R {
+		let final_key = Self::storage_linked_map_final_key(key.borrow());
+
+		let (mut val, _linkage) = read_with_linkage::<_, _, G>(final_key.as_ref())
+			.map(|(data, linkage)| (G::from_optional_value_to_query(Some(data)), Some(linkage)))
```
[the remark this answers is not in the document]
thiolliere: for the following: `MyStorage: Option<u32>` `from_optional_value_to_query(Some(data))= Some(data)`

srml/support/src/storage/hashed.rs
```diff
@@ -16,12 +16,72 @@
...
+pub struct Twox256;
+impl StorageHasher for Twox256 {
+	type Output = [u8; 32];
+	fn hash(x: &[u8]) -> [u8; 32] {
+		twox_256(x)
+	}
+}
 
 /// Return the value of the item in storage under `key`, or `None` if there is no explicit entry.
 pub fn get<T, HashFn, R>(hash: &HashFn, key: &[u8]) -> Option<T>
```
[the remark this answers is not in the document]
thiolliere: they are used by StorageVec traits below.

StorageVec is also an abstraction over storage such as StorageMap but it is not handled by decl_storage (so no metadata).
At some point maybe we could remove it or also make the below StorageVec use unhashed directly.

I was thinking after this PR we could remove hashed module and just have child and top modules.

srml/support/src/storage/generator/double_map.rs
thiolliere: actually it only prefixes the first key as final key is:
hasher1(key1_prefix ++ key1) ++ hasher2(key2)

srml/support/src/storage/generator/mod.rs
```diff
@@ -0,0 +1,9 @@
+mod linked_map;
```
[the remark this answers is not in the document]
thiolliere: done

thiolliere:
> I still don't understand all of the code that was removed, but the new code looks good AFAICT.

Before generator were implementing get, insert, ... for a storage type in argument: `put(val, storage)`. This was used when building genesis so we can put into the genesis storage which is a tuple of hashmaps. But then ppl start using with_storage making this feature useless. So now we just call with_storage from inside the build_storage function and user doesn't have to care if it insert into genesis storage or final storage.

srml/support/procedural/src/lib.rs
```diff
@@ -126,7 +144,7 @@ use proc_macro::TokenStream;
 ///			config(genesis_field): GenesisFieldType;
 ///			config(genesis_field2): GenesisFieldType;
 ///			...
-///			build(|_: &mut StorageOverlay, _: &mut ChildrenStorageOverlay, _: &GenesisConfig<T>| {
+///			build(|_: &GenesisConfig<T>| {
```
bkchr:
```suggestion
///			build(|_: &Self| {
```

srml/support/src/storage/child.rs
```diff
@@ -0,0 +1,105 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! child storage NOTE could replace unhashed by having only one kind of storage (root being null storage
```
bkchr: Weird sentence.

srml/support/src/storage/generator/double_map.rs
```diff
@@ -0,0 +1,197 @@
...
+	G: StorageDoubleMap<K1, K2, V>,
+{
+	type Query = G::Query;
+
+	fn exists<KArg1, KArg2>(k1: &KArg1, k2: &KArg2) -> bool
+	where
+		K1: Borrow<KArg1>,
+		K2: Borrow<KArg2>,
+		KArg1: ?Sized + Encode,
+		KArg2: ?Sized + Encode
```
bkchr:
```suggestion
		KArg2: ?Sized + Encode,
```

srml/support/src/storage/generator/double_map.rs
```diff
@@ -0,0 +1,197 @@
...
+	{
+		unhashed::exists(&Self::storage_double_map_final_key(k1, k2))
+	}
+
+	fn get<KArg1, KArg2>(k1: &KArg1, k2: &KArg2) -> Self::Query
+	where
+		K1: Borrow<KArg1>,
+		K2: Borrow<KArg2>,
+		KArg1: ?Sized + Encode,
+		KArg2: ?Sized + Encode
```
bkchr:
```suggestion
		KArg2: ?Sized + Encode,
```

srml/support/src/storage/generator/double_map.rs
```diff
@@ -0,0 +1,197 @@
...
+	{
+		G::from_optional_value_to_query(unhashed::get(&Self::storage_double_map_final_key(k1, k2)))
+	}
+
+	fn take<KArg1, KArg2>(k1: &KArg1, k2: &KArg2) -> Self::Query
+	where
+		K1: Borrow<KArg1>,
+		K2: Borrow<KArg2>,
+		KArg1: ?Sized + Encode,
+		KArg2: ?Sized + Encode
```
bkchr:
```suggestion
		KArg2: ?Sized + Encode,
```

srml/support/src/storage/generator/double_map.rs
```diff
@@ -0,0 +1,197 @@
...
+	{
+		let final_key = Self::storage_double_map_final_key(k1, k2);
+
+		let value = unhashed::get(&final_key);
```
bkchr: Why not use `unhashed::take` here?

srml/support/src/storage/generator/double_map.rs
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	fn insert<KArg1, KArg2, VArg>(k1: &KArg1, k2: &KArg2, val: &VArg)
+	where
+		K1: Borrow<KArg1>,
+		K2: Borrow<KArg2>,
+		V: Borrow<VArg>,
+		KArg1: ?Sized + Encode,
+		KArg2: ?Sized + Encode,
+		VArg: ?Sized + Encode
```
bkchr:
```suggestion
		VArg: ?Sized + Encode,
```

srml/support/src/storage/generator/double_map.rs
```diff
@@ -0,0 +1,197 @@
...
+	{
+		unhashed::put(&Self::storage_double_map_final_key(k1, k2), &val.borrow())
+	}
+
+	fn remove<KArg1, KArg2>(k1: &KArg1, k2: &KArg2)
+	where
+		K1: Borrow<KArg1>,
+		K2: Borrow<KArg2>,
+		KArg1: ?Sized + Encode,
+		KArg2: ?Sized + Encode
```
bkchr:
```suggestion
		KArg2: ?Sized + Encode,
```

srml/support/src/storage/generator/double_map.rs
```diff
@@ -0,0 +1,197 @@
...
+		unhashed::kill_prefix(Self::storage_double_map_final_key1(k1).as_ref())
+	}
+
+	fn mutate<KArg1, KArg2, R, F>(k1: &KArg1, k2: &KArg2, f: F) -> R
+	where
+		K1: Borrow<KArg1>,
+		K2: Borrow<KArg2>,
+		KArg1: ?Sized + Encode,
+		KArg2: ?Sized + Encode,
+		F: FnOnce(&mut Self::Query) -> R
```
bkchr:
```suggestion
		F: FnOnce(&mut Self::Query) -> R,
```

srml/support/src/storage/generator/double_map.rs
```diff
@@ -0,0 +1,197 @@
...
+		k2: &KArg2,
+		items: &[I],
+	) -> Result<(), &'static str>
+	where
+		K1: Borrow<KArg1>,
+		K2: Borrow<KArg2>,
+		KArg1: ?Sized + Encode,
+		KArg2: ?Sized + Encode,
+		I: codec::Encode,
+		V: EncodeAppend<Item=I>
```
bkchr:
```suggestion
		V: EncodeAppend<Item=I>,
```

srml/support/src/storage/generator/linked_map.rs
```diff
@@ -0,0 +1,289 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use codec::{Codec, Encode, Decode};
+use crate::{
+	storage::{self, unhashed, hashed::StorageHasher},
+	rstd::{
```
bkchr: I would move this out of `crate::{}`.

srml/support/src/storage/generator/linked_map.rs
```diff
@@ -0,0 +1,289 @@
...
+
+		let value = full_value.map(|(data, linkage)| {
+			remove_linkage::<_, _, G>(linkage);
+			data
+		});
+
+		G::from_optional_value_to_query(value)
+	}
+
+	fn enumerate() -> Box<dyn Iterator<Item = (K, V)>>
```
bkchr: Can we not return `Enumerator` directly here? Without boxing?

srml/support/src/storage/generator/map.rs
```diff
@@ -0,0 +1,137 @@
...
+		match G::from_query_to_optional_value(val) {
+			Some(ref val) => G::insert(key, val),
+			None => G::remove(key),
+		}
+		ret
+	}
+
+	fn take<KeyArg: Borrow<K>>(key: KeyArg) -> Self::Query {
+		let key = Self::storage_map_final_key(key);
+		let value = unhashed::get(key.as_ref());
```
bkchr: unhashed::take?

srml/support/src/storage/mod.rs
```diff
@@ -19,97 +19,21 @@
...
+pub mod generator;
 
 /// A trait for working with macro-generated storage values under the substrate storage API.
 pub trait StorageValue<T: Codec> {
 	/// The type that get/take return.
 	type Query;
 
 	/// Get the storage key.
-	fn key() -> &'static [u8];
+	fn key() -> Vec<u8>;
```
bkchr: Why does this return a `Vec<u8>`?

srml/support/src/storage/storage_items.rs
```diff
@@ -50,109 +50,129 @@ pub use crate::rstd::marker::PhantomData;
...
 		storage_items!($($t)*);
 	};
 	($name:ident : $key:expr => default $ty:ty; $($t:tt)*) => {
-		$crate::__storage_items_internal!(() () (RAW_TYPE $ty) (get_or_default) (take_or_default) $name: $key => $ty);
+		$crate::__storage_items_internal!(() () (RAW_TYPE $ty) (unwrap_or_default) (wrap) $name: $key => $ty);
 		storage_items!($($t)*);
 	};
 	(pub $name:ident : $key:expr => default $ty:ty; $($t:tt)*) => {
-		$crate::__storage_items_internal!((pub) () (RAW_TYPE $ty) (get_or_default) (take_or_default) $name: $key => $ty);
+		$crate::__storage_items_internal!((pub) () (RAW_TYPE $ty) (unwrap_or_default) (wrap) $name: $key => $ty);
```
bkchr: Can you not just directly pass `Some` here instead of `wrap`?

srml/system/src/lib.rs
```diff
@@ -413,19 +413,17 @@ decl_storage! {
...
+			runtime_io::set_storage(well_known_keys::CODE, &config.code);
+			runtime_io::set_storage(well_known_keys::EXTRINSIC_INDEX, &0u32.encode());
 
 			if let Some(ref changes_trie_config) = config.changes_trie_config {
-				storage.0.insert(
-					well_known_keys::CHANGES_TRIE_CONFIG.to_vec(),
-					changes_trie_config.encode());
+				runtime_io::set_storage(
+					well_known_keys::CHANGES_TRIE_CONFIG,
+					&changes_trie_config.encode()
```
```suggestion
					&changes_trie_config.encode(),
```

srml/support/src/storage/mod.rs
thiolliere: the PR changes the key function to returned hashed key instead of raw keys. thus this is not static anymore.

Instead of Vec we could return `[u8; 32]` as we always hash StorageValue with twox_128.

srml/support/src/storage/generator/linked_map.rs
thiolliere: yeah but for that we need to make Enumerator as a generic parameter of the trait StorageLinkedMap<K, V, Enumerator>

because we can't have `fn enumerate() -> impl Iterator<Item = (K, V)> where K: 'static, V: 'static, Self: 'static;` in the trait definition

This can be done indeed.

srml/support/src/storage/child.rs
thiolliere: yes I rewrote

srml/support/src/storage/storage_items.rs
thiolliere: Actually we can probably do that with this import on latest rust stable
```rust
#[doc(hidden)]
use Option::Some;
```

srml/support/src/storage/generator/linked_map.rs
thiolliere: I mean I don't mind either implementation, if you say you prefer one I go for it.

srml/support/src/storage/mod.rs
bkchr: Sounds like a good idea.

thiolliere: done

srml/staking/src/lib.rs
```diff
@@ -628,37 +626,32 @@ decl_storage! {
...
+					balance,
+					RewardDestination::Staked
+				);
+				let _ = match status {
+					StakerStatus::Validator => {
+						<Module<T>>::validate(
+							T::Origin::from(Some(controller.clone()).into()),
+							Default::default()
+						)
+					}, StakerStatus::Nominator(votes) => {
```
bkchr:
```suggestion
					},
					StakerStatus::Nominator(votes) => {
```

srml/staking/src/lib.rs
```diff
@@ -628,37 +626,32 @@ decl_storage! {
...
+						<Module<T>>::nominate(
+							T::Origin::from(Some(controller.clone()).into()),
+							votes.iter().map(|l| T::Lookup::unlookup(l.clone())).collect()
```
bkchr:
```suggestion
							votes.iter().map(|l| T::Lookup::unlookup(l.clone())).collect(),
```

srml/staking/src/lib.rs
```diff
@@ -628,37 +626,32 @@ decl_storage! {
...
+					T::Origin::from(Some(stash.clone()).into()),
+					T::Lookup::unlookup(controller.clone()),
+					balance,
+					RewardDestination::Staked
+				);
+				let _ = match status {
+					StakerStatus::Validator => {
+						<Module<T>>::validate(
+							T::Origin::from(Some(controller.clone()).into()),
+							Default::default()
```
bkchr:
```suggestion
							Default::default(),
```

srml/staking/src/lib.rs
```diff
@@ -628,37 +626,32 @@ decl_storage! {
...
+			for &(ref stash, ref controller, balance, ref status) in &config.stakers {
+				assert!(
+					T::Currency::free_balance(&stash) >= balance,
+					"Stash does not have enough balance to bond."
+				);
+				let _ = <Module<T>>::bond(
+					T::Origin::from(Some(stash.clone()).into()),
+					T::Lookup::unlookup(controller.clone()),
+					balance,
+					RewardDestination::Staked
```
bkchr:
```suggestion
					RewardDestination::Staked,
```

srml/support/src/storage/generator/linked_map.rs
bkchr: YEah, we could just add it as associated type or as generic parameter. I would like to prevent to allocate :)

srml/support/src/storage/storage_items.rs
```diff
@@ -50,109 +50,127 @@ pub use crate::rstd::marker::PhantomData;
 #[doc(hidden)]
 pub use crate::rstd::boxed::Box;
 
+#[doc(hidden)]
+pub fn id<T>(t: T) -> T {
+	t
+}
+
+#[doc(hidden)]
+pub use Some;
```
bkchr: Isn't `Some` always in the prelude? Even in `no_std`?

srml/support/src/storage/generator/value.rs
```diff
@@ -0,0 +1,156 @@
...
+	fn unhashed_key() -> &'static [u8];
+
+	/// Convert an optional value retrieved from storage to the type queried.
+	fn from_optional_value_to_query(v: Option<T>) -> Self::Query;
+
+	/// Convert a query to an optional value into storage.
+	fn from_query_to_optional_value(v: Self::Query) -> Option<T>;
+
+	/// Generate the full key used in top storage.
+	fn storage_value_final_key() -> [u8;16] {
```
bkchr:
```suggestion
	fn storage_value_final_key() -> [u8; 16] {
```

srml/support/src/storage/storage_items.rs
thiolliere: yes true. it is only because we define the keyword as an ident and expand later in the macro with `$crate::storage::storage_items::$into_query(v)`

at some point I could actually change ident to path and put full path earlier

thiolliere:
> Modification of StorageValue::key and StorageMap::key_for: they now return the key in top storage, which make more sense. Previously it returned the key in storage but not hashed. At some point I think we should remove those as in the future a storageValue can actually work on child_storage so this function would not have that much meaning.

should we want to rename it to hashed_key instead so that it is safe for user (like the method won't change with the same name leading to confusing errors)

bkchr:
I don't know what `top storage` is, but if the key is hashed `hashed_key` sounds reasonable.

thiolliere:
top storage is just the regular storage, (as opposed to child storages).

yes currently all keys are hashed but at some point maybe in the future we want to have some storage abstraction that works without not hashed keys. Well we'll see later when doing this kind of stuff we could anyway switch back as it won't be confusing anymore.

srml/support/src/storage/generator/linked_map.rs
thiolliere: done :-)

thiolliere:
function renamed*

PR #3346, opened 2019-08-08
Add block announce validator.

twittner:
Adds a `BlockAnnounceValidator` trait which is used to check `BlockAnnounce`ments. It also adds an associated data field to `BlockAnnounce` which is used for things like candidate messages.


core/consensus/common/src/block_validation.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/// Type which checks incoming block announcements.
+pub trait BlockAnnounceValidator<B: Block> {
+	/// Validate the announced header and its associated data.
+	fn validate(&mut self, header: &B::Header, data: &[u8]) -> Result<Validation, Box<dyn Error + Send>>;
+}
+
+#[derive(Debug)]
+pub struct DefaultBlockAnnounceValidator<C> {
```
rphmeier: missing docs

core/consensus/common/src/block_validation.rs
```diff
@@ -0,0 +1,67 @@
...
+// (at your option) any later version.
+//
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
```
rphmeier: missing module docs

rphmeier:
@arkpar 

> To me it looks like the extra data should go into a separate message in the protocol extension.

To be honest, that is a decent approach as well. Perhaps we should rather have a method to use a custom block announcement message rather than bundle the extra data.

The desired use-case is for Cumulus: before a block is included on the relay chain but after voting on it has begun, the block should be synced by peers only when bundled with a submission attestation.

We would need two changes:
  - A way for substrate-network to ask some user code if it would like to use its own message for the announcement.
  - A way for user code to tell substrate-network there has been a block announcement.

arkpar:
How is it different from e.g block justifications, that come not as part of the announcement, but as block data?

rphmeier:
Justifications are meant to prove finality and be permanently valid.  It's exactly what you pinpointed - that this is for a use-case where the justification-like thing is temporary and doesn't need to be persisted, at least not for long.

bkchr:
So, I personally would go with an extra field for the block announcement message and move the default validation completely out of the sync code. I could imagine that people want to bring in their own validation of the block announcement, for whatever reasons. 

But yeah, for the sake of bringing Cumulus to a point that it works, we can also go the way of adding an extra message for our own block announcements. 

@arkpar @rphmeier which way to we want to go?

arkpar:
> So, I personally would go with an extra field for the block announcement message and move the default validation completely out of the sync code. 

I'm fine with this

twittner:
It would be helpful to know what you consider "default validation" from here: https://github.com/paritytech/substrate/blob/d2e7d660f8dbbb8f9753dfd231cef1c9b502c41c/core/network/src/protocol/sync.rs#L893-L1002? Most of it relates to the state of peers which would mean to replicate a lot of the peer state handling and also to keep it up to date w.r.t. what is currently being downloaded from where etc. I am not sure it is a good idea to replicate this which begs the question what is left to validate?

bkchr:
The peer management does not seems to be used after line 933? As the updating of the peer information is not validating, it should stay in the sync layer.

twittner:
> The peer management does not seems to be used after line 933?

It is, e.g. 
- `is_already_downloading` (line 941),
- `download_unknown_stale` (lines 955, 969),
- `select_new_blocks` (line 988).

bkchr:
Can we not make this available to the announce validator using a trait? 

The validator would probably return the following:
```
enum ValidationResult {
    Request(Request),
    Nothing,
}
```

Then in the `on_block_announce` we convert this into the `OnBlockAnnounce` by doing:
```
match res {
    ValidationResult::Request(req) => if requires_additional_data {
						return OnBlockAnnounce::Request(who, request)
					} else {
						return OnBlockAnnounce::ImportHeader
					}
     ValidationResult::Nothing => OnBlockAnnounce::Nothing
}
```

twittner:
Could you be more specific? Do you mean we should run `on_block_announce` in `sync.rs` as it is today and within this method collect some data (what exactly?) which we then pass to the validator which in turn constructs a `substrate_network::protocol::message::BlockRequest` that `on_block_announce` then sends over the network? If so I still do not understand what validation logic should be moved to the validator. Also, is this coupling not a bit tight?

twittner:
@bkchr?

rphmeier:
The goal of the PR is to give us the option to do additional, customized validation on whether a block announcement is valid based on possibly time-sensitive circumstances. We'd do so by bundling a bit of optional extra data with the announcement to be checked by some user code, along with the announcement.

Most of the checks that the `on_block_announce` function is doing are technically announcement validation, but are the kind of checks that you would always want to do. I don't think it's worthwhile to abstract to the point where these are extractable:
  - Checking if the peer has recently announced the block
  - Checking if the block's parent is ancient
  - Checking what protocol state the peer is in
  - Checking whether we know the block already or are downloading it

With that in mind I'd suggest that "default" verification could be an empty or nearly-empty trait implementation - this trait implementation could live outside of the sync code, although I'd say it's reasonable for it to live in the network crate.

rphmeier:
Then there's also the other side of the coin to discuss: how the block announcement should actually be bundled with this extra-data.

I'm not particularly opinionated on this, but ideally the two methods can fall within the same trait to prevent user-protocol logic from becoming spread out.

bkchr:
Yeah, sorry for the delay! 

I think what @rphmeier writes, sounds reasonable. Regarding the extra-data, as this is probably just some opaque field, we could probably check this in the same function we are giving the general "yes/no" to the block request.

twittner:
Ok, turning it into a non-WIP PR now. Please review when time permits it. Regarding

> Then there's also the other side of the coin to discuss: how the block announcement should actually be bundled with this extra-data.

This is not addressed in this PR. The various places where interfaces require associated data use empty vectors for now.

core/client/src/client.rs
```diff
@@ -1185,10 +1173,7 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 		Ok(())
 	}
 
-	fn notify_imported(
-		&self,
-		notify_import: ImportSummary<Block>,
-	) -> error::Result<()> {
+	fn notify_imported(&self, notify_import: ImportSummary<Block>) -> error::Result<()> {
 		if let Some(storage_changes) = notify_import.storage_changes {
 			// TODO [ToDr] How to handle re-orgs? Should we re-emit all storage changes?
```
kianenigma: @tomusdrw solved or needs an issue?

core/network/src/protocol.rs
```diff
@@ -360,25 +363,31 @@ impl<B: BlockT, S: NetworkSpecialization<B>, H: ExHashT> Protocol<B, S, H> {
...
+		block_announce_validator: Box<dyn BlockAnnounceValidator<B> + Send>
 	) -> error::Result<(Protocol<B, S, H>, peerset::PeersetHandle)> {
 		let info = chain.info();
-		let sync = ChainSync::new(config.roles, chain.clone(), &info, finality_proof_request_builder);
+		let sync = ChainSync::new(
+			config.roles,
+			chain.clone(),
+			&info,
+			finality_proof_request_builder,
+			block_announce_validator);
```
kianenigma:
```suggestion
			block_announce_validator
		);
```

kianenigma:
```suggestion
			block_announce_validator,
		);
```

core/network/src/protocol/sync.rs
```diff
@@ -28,11 +28,24 @@
...
+	BlockOrigin,
+	BlockStatus,
+	block_validation::{BlockAnnounceValidator, Validation},
+	import_queue::{IncomingBlock, BlockImportResult, BlockImportError}
+};
 use crate::{
 	config::{Roles, BoxFinalityProofRequestBuilder},
-	message::{self, generic::FinalityProofRequest, BlockAttributes, BlockRequest, BlockResponse, FinalityProofResponse},
+	message::{
+		BlockAnnounce,
```
kianenigma: better to wrap when the line limit exceeds.

twittner: Why?

kianenigma: consistency with the rest of substrate -- though this (afaik) is not strictly in our style guide. Usually what I do (and I think it is sort of the implicit styleguide):

```
use crate::{
     top_level_stuff_1, top_level_stuff_2, top_level_stuff_3, top_level_stuff_wrap_if_needed,
     inner_module_1::{stuff, bar, Bazinga},
     inner_module_2::{foo, kaboom, blah},
}
```

basically break once per sub-module but other than that only when needed.

core/consensus/common/src/block_validation.rs
```diff
@@ -0,0 +1,66 @@
...
+	fn block_status(&self, id: &BlockId<B>) -> Result<BlockStatus, Box<dyn Error + Send>> {
+		(&**self).block_status(id)
+	}
+}
+
+/// Result of `BlockAnnounceValidator::validate`.
+#[derive(Debug, PartialEq, Eq)]
+pub enum Validation {
+	/// Valid block announcement.
+	Success,
```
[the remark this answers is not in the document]
twittner: I think naming the type "Valid" can be misleading because "valid" is already biased towards "yes". Saying something valid is not valid feels a little inconsistent to me, so at a minimum I would like to see a neutral name. I agree that the "Result" prefix should better be avoided, especially since the type name will often be embedded in another `Result`. I would rather read `Result<Validation, ...>` instead of `Result<ValidationResult, ...>`.

core/client/src/client.rs
```diff
@@ -201,6 +186,8 @@ pub struct BlockImportNotification<Block: BlockT> {
 	pub is_new_best: bool,
 	/// List of retracted blocks ordered by block number.
 	pub retracted: Vec<Block::Hash>,
+	/// Any data associated with this block import.
+	pub associated_data: Vec<u8>,
```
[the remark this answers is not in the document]
twittner: If we remove this, how do we send block announcements with associated data to network peers?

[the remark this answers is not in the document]
bkchr: https://github.com/paritytech/substrate/pull/3346/files#diff-16b4c34d50bdfcbf732c09c7be1fdcf9R684 here do we send the block announcement. The import notification is just used to know that we need to send a block announcement. But as I already said, we will send the block announcements differently in Cumulus.

twittner:
> Just one last master merge.

Merging was actually quite involved. I think https://github.com/paritytech/substrate/pull/3602 has introduced some changes in the vicinity. Maybe @andresilva could check that things still make sense?

arkpar:
Indeed the protocol should be backwards compatible. It should not send any additional data to nodes running older version. Furthermore message encoding for the current version should match standard SCALE encoding.

core/network/src/protocol/message.rs
```diff
@@ -284,16 +286,21 @@ pub mod generic {
 			if let Some(state) = &self.state {
 				state.encode_to(dest);
 			}
+			if !self.data.is_empty() {
```
arkpar: This will interfere with encoding any additional fields. I.e. this new protocol version should always encode `data` when communicating with the same protocol version and do not encode when communicating with an older version. I'd suggest using `Option<Vec<u8>>` in `BlockAnnounce` for that. Later we can remove backwards compatibility and `Option` without changing the actual encoding for the current version.

twittner: I would normally have called `encode_to` unconditionally. Wrapping `data` in an `Option` is not bulletproof either, as one can easily just put it in a `Some`. Also, changing the type ripples through the whole system which is not great. As I mentioned [elsewhere](https://github.com/paritytech/substrate/pull/3452#issuecomment-525172177), we should reconsider our encoding, otherwise minor changes such as this will continue to hurt us.

arkpar: Sure, we should get proper versioning support at some point, but this is out of scope of this PR I believe.

core/network/src/protocol/sync.rs
```diff
@@ -932,6 +941,20 @@ impl<B: BlockT> ChainSync<B> {
 			return OnBlockAnnounce::Nothing
 		}
 
+		// Let external validator check the block announcement.
+		let assoc_data = announce.data.as_ref().map(|v| v.as_slice()).unwrap_or(&[]);
```
bkchr:
```suggestion
		let assoc_data = announce.data.as_ref().map_or(&[], |v| v.as_slice());
```

core/network/src/protocol/message.rs
```diff
@@ -273,6 +273,8 @@ pub mod generic {
 		pub header: H,
 		/// Block state. TODO: Remove `Option` and custom encoding when v4 becomes common.
 		pub state: Option<BlockState>,
+		/// Data associated with this block announcement, e.g. a candidate message.
```
bkchr: The `Option` should be removed here as well, when `4` becomes common.

core/network/src/protocol/message.rs
```diff
@@ -284,16 +286,21 @@ pub mod generic {
...
+				data.encode_to(dest)
+			}
 		}
 	}
 
 	impl<H: Decode> Decode for BlockAnnounce<H> {
 		fn decode<I: Input>(input: &mut I) -> Result<Self, codec::Error> {
 			let header = H::decode(input)?;
 			let state = BlockState::decode(input).ok();
+			let data = Vec::decode(input).ok();
```
bkchr: I'm not happy with this. Conceptually it should be `Option<(BlockState, Vec<u8>)>`. As both should be present with version 4. But yeah, for the sake of merging this and don't over-complicating it more, we probably can go with this.

PR #3351, opened 2019-08-08
node: Reintroduce `--validator` cli flag

andresilva:
Conflicts with #3348 but I think I prefer this one instead, because realistically no one will actually pass `--no-validator` and full nodes will be unnecessarily running validator code.


andresilva:
Sorry when I wrote this I didn't see you were already working on #3348 (I had no internet).

PR #3353, opened 2019-08-09
node: exit on GRANDPA voter or BABE authoring error

andresilva:
Currently if either GRANDPA or BABE fails the node continues running, this behavior might be hard to spot by the users. For example, if the GRANDPA voter fails, the node might keep producing blocks but not observing finality anymore (and potentially not participating in the protocol). Most likely validators will have some process monitoring in place which will just restart the node afterwards.

In this PR when either of these validator tasks fails we shut down the service and exit cleanly.

### DIFF
--- a/core/service/src/lib.rs
+++ b/core/service/src/lib.rs
@@ -88,9 +88,9 @@ pub struct Service<Components: components::Components> {
 	on_exit: exit_future::Exit,
 	/// A signal that makes the exit future above resolve, fired on service drop.
 	exit_signal: Option<Signal>,
-	/// When set to `true` the next time the service future is polled it should
-	/// be completed (i.e. returns `Async::Ready(())`).
-	should_exit: Arc<AtomicBool>,
+	/// Set to `true` when a spawned infallible task has failed. The next time
+	/// the service future is polled it should complete with an error.
+	infallible_failed: Arc<AtomicBool>,
 	/// Sender for futures that must be spawned as background tasks.
 	to_spawn_tx: mpsc::UnboundedSender<Box<dyn Future<Item = (), Error = ()> + Send>>,
 	/// Receiver for futures that must be spawned as background tasks.
@@ -447,7 +447,7 @@ impl<Components: components::Components> Service<Components> {
 			transaction_pool,
 			on_exit,
 			exit_signal: Some(exit_signal),
-			should_exit: Arc::new(AtomicBool::new(false)),
+			infallible_failed: Arc::new(AtomicBool::new(false)),
 			to_spawn_tx,
 			to_spawn_rx,
 			to_poll: Vec::new(),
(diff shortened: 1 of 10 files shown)


andresilva:
I would like to write a test for this but I have no idea how to actually create a `Service` in a test. Do I have to setup some test factory with `construct_service_factory!`?

core/service/src/lib.rs
```diff
@@ -549,6 +569,10 @@ impl<Components> Future for Service<Components> where Components: components::Co
 	type Error = ();
 
 	fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+		if self.should_exit.load(Ordering::Relaxed) {
+			return Ok(Async::Ready(()));
```
DemiMarie-parity: We should exit with a failure code here, so that service managers such as `systemd` or `daemontools` know that Substrate has failed and can respond accordingly.

tomaka:
What about making all tasks "infallible" instead?

core/service/src/lib.rs
```diff
@@ -549,6 +569,10 @@ impl<Components> Future for Service<Components> where Components: components::Co
 	type Error = ();
 
 	fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+		if self.should_exit.load(Ordering::Relaxed) {
```
jimpo: Is it possible to check `self.on_exit.is_live()` here instead of having a separate `should_exit` flag? Or do you not want to because the mutex that uses is more expensive than checking an atomic?

kianenigma:
probably needs to be ported to polkadot later as well.

core/service/src/lib.rs
```diff
@@ -82,8 +83,14 @@ pub struct Service<Components: components::Components> {
 		NetworkStatus<ComponentBlock<Components>>, NetworkState
 	)>>>>,
```
rphmeier: my brain assumed this was a bad merge at first

core/service/src/lib.rs
```diff
@@ -489,6 +497,18 @@ impl<Components: components::Components> Service<Components> {
 		let _ = self.to_spawn_tx.unbounded_send(Box::new(task));
 	}
 
+	/// Spawns a task in the background that runs the future passed as
+	/// parameter. The given task is considered infallible, i.e. if it errors we
+	/// trigger a service exit.
+	pub fn spawn_infallible_task(&self, task: impl Future<Item = (), Error = ()> + Send + 'static) {
```
rphmeier: bikeshed: i wouldn't say infallible is the right term here. that implies that the task itself cannot fail, where we rather want to convey that we can't tolerate the task failing.

alternate terms: "vital" or "imperative"

andresilva: Yeah! When talking to Pierre he called it "critical" task, which seems adequate.

DemiMarie-parity: What about the term “essential”?  That seems like the best one.

andresilva:
Updated PR to propagate error until main where the process should check the error and exit with `1`. @tomaka This touched even more stuff on service, not sure how annoying this is for your pending refactoring. I will revert some of the renamings of existing `Service` fields.

PR #3355, opened 2019-08-09
More robust state pinning

arkpar:
Currently, when pinned state is discarded on canonicalization state-db would refuse to canonicalize and delay it till next call to `canonicalize_block`. This may lead to state-db canonicalized block and blockchain metadata finalized block to go out of sync. 

This PR changes the implementation to apply canonicalization regardless of pinned state, keeping pinned values in memory until they are not needed anymore. 

Also fixed a race between `is_pruned` and `pin`

### DIFF
--- a/core/state-db/src/noncanonical.rs
+++ b/core/state-db/src/noncanonical.rs
@@ -347,7 +347,7 @@ impl<BlockHash: Hash, Key: Hash> NonCanonicalOverlay<BlockHash, Key> {
 						0,
 						&mut self.parents,
 						&mut self.pinned,
-						&overlay.hash
+						&overlay.hash,
 					);
 				}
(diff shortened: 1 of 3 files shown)


core/state-db/src/noncanonical.rs
```diff
@@ -339,9 +341,16 @@ impl<BlockHash: Hash, Key: Hash> NonCanonicalOverlay<BlockHash, Key> {
...
 				self.parents.remove(&overlay.hash);
 				if i != index {
-					discard_descendants(&mut self.levels, &mut self.values, 0, &mut self.parents, &overlay.hash);
+					discard_descendants(
+						&mut self.levels,
+						&mut self.values,
+						0,
+						&mut self.parents,
+						&mut self.pinned,
+						&overlay.hash
```
rphmeier:
```suggestion
						&overlay.hash,
```

PR #3357, opened 2019-08-09
 Scale trait and move to u32 blocknumbers

gavofyork:


node/runtime/src/lib.rs
```diff
@@ -179,7 +179,7 @@ impl timestamp::Trait for Runtime {
 }
 
 parameter_types! {
-	pub const UncleGenerations: u64 = 5;
+	pub const UncleGenerations: u32 = 5;
```
bkchr:
```suggestion
	pub const UncleGenerations: BlockNumber = 5;
```

andresilva:
> Looks good.
> 
> Why do we do this?

With 6 seconds block time `u32` should be good for about 800 years.

core/sr-primitives/src/traits.rs
```diff
@@ -303,6 +303,47 @@ pub trait CheckedConversion {
 }
 impl<T: Sized> CheckedConversion for T {}
 
+/// Multiply and devide by a number that isn't necessarily the same type. Basically just the same
```
shawntabrizi:
```suggestion
/// Multiply and divide by a number that isn't necessarily the same type. Basically just the same
```

shawntabrizi: It feels a little unfortunate we have a SCALE codec and a Scale trait, but that may just be me.

PR #3358, opened 2019-08-09
Refactor parse_and_execute

tomaka:
Deprecates `parse_and_execute` and replaces it with a new `parse_and_prepare` function whose control flow is in my opinion way easier to read.

I invite you to check out the third commit to see what the new code looks like.

The reason for this change is that I'd like to get rid of the `ServiceFactory` and `Components` traits, which `parse_and_execute` uses.
This PR instead moves the `ServiceFactory` trait usage to when the user calls `run()`, which makes it possible in the future to add functions other than `run()` that don't require this trait.
As an example, we could add a function to `ParseAndExecute::RevertChain` that simple requires passing a `Client` by parameter, instead of asking for a `ServiceFactory`.

### DIFF
--- a/core/cli/src/lib.rs
+++ b/core/cli/src/lib.rs
@@ -274,8 +274,8 @@ impl<'a, RP> ParseAndPrepareRun<'a, RP> {
 		C: Default,
 		G: RuntimeGenesis,
 		E: IntoExit,
-		RS: FnOnce(E, RunCmd, RP, Configuration<C, G>) -> Result<(), String> {
-
+		RS: FnOnce(E, RunCmd, RP, Configuration<C, G>) -> Result<(), String>
+	{
 		let config = create_run_node_config(self.params.left.clone(), spec_factory, self.impl_name, self.version)?;
 
 		run_service(exit, self.params.left, self.params.right, config).map_err(Into::into)
@@ -295,8 +295,8 @@ impl<'a> ParseAndPrepareBuildSpec<'a> {
 		spec_factory: S
 	) -> error::Result<()>
 	where S: FnOnce(&str) -> Result<Option<ChainSpec<G>>, String>,
-		G: RuntimeGenesis {
-		
+		G: RuntimeGenesis
+	{
 		info!("Building chain spec");
 		let raw_output = self.params.raw;
 		let mut spec = load_spec(&self.params.shared_params, spec_factory)?;
@@ -324,8 +324,8 @@ impl<'a> ParseAndPrepareExport<'a> {
 	) -> error::Result<()>
 	where S: FnOnce(&str) -> Result<Option<ChainSpec<FactoryGenesis<F>>>, String>,
 		F: ServiceFactory,
-		E: IntoExit {
-
+		E: IntoExit
+	{
 		let config = create_config_with_db_path(spec_factory, &self.params.shared_params, self.version)?;
 
 		info!("DB path: {}", config.database_path.display());
@@ -359,8 +359,8 @@ impl<'a> ParseAndPrepareImport<'a> {
 	) -> error::Result<()>
 	where S: FnOnce(&str) -> Result<Option<ChainSpec<FactoryGenesis<F>>>, String>,
 		F: ServiceFactory,
-		E: IntoExit {
-		
+		E: IntoExit
+	{
 		let mut config = create_config_with_db_path(spec_factory, &self.params.shared_params, self.version)?;
 		config.execution_strategies = ExecutionStrategies {
 			importing: self.params.execution.into(),
@@ -396,12 +396,12 @@ impl<'a> ParseAndPreparePurge<'a> {
 		spec_factory: S
 	) -> error::Result<()>
 	where S: FnOnce(&str) -> Result<Option<ChainSpec<G>>, String>,
-		G: RuntimeGenesis {
-		
+		G: RuntimeGenesis
+	{
 		let config = create_config_with_db_path::<(), _, _>(spec_factory, &self.params.shared_params, self.version)?;
 		let db_path = config.database_path;
 
-		if self.params.yes == false {
+		if !self.params.yes {
 			print!("Are you sure to remove {:?}? (y/n)", &db_path);
 			stdout().flush().expect("failed to flush stdout");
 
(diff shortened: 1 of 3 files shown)


tomaka:
Another possible change in the future would be to add new functions to `ParseAndPrepareRun`, `ParseAndPrepareImport` and `ParseAndPrepareExport` that return a `Future`, thus letting the user control the blocks import or export and removing the `exit` parameter.

core/cli/src/lib.rs
```diff
@@ -222,23 +213,294 @@ where
...
+		spec_factory: S,
+		exit: E,
+		run_service: RS,
+	) -> error::Result<()>
+	where S: FnOnce(&str) -> Result<Option<ChainSpec<G>>, String>,
+		RP: StructOpt + Clone,
+		C: Default,
+		G: RuntimeGenesis,
+		E: IntoExit,
+		RS: FnOnce(E, RunCmd, RP, Configuration<C, G>) -> Result<(), String> {
```
```suggestion
		RS: FnOnce(E, RunCmd, RP, Configuration<C, G>) -> Result<(), String>
	{
```

core/cli/src/lib.rs
```diff
@@ -222,23 +213,294 @@ where
...
+
```
```suggestion
```

core/cli/src/lib.rs
```diff
@@ -222,23 +213,294 @@ where
...
+}
+
+impl<'a> ParseAndPrepareBuildSpec<'a> {
+	/// Runs the command and build the chain specs.
+	pub fn run<G, S>(
+		self,
+		spec_factory: S
+	) -> error::Result<()>
+	where S: FnOnce(&str) -> Result<Option<ChainSpec<G>>, String>,
+		G: RuntimeGenesis {
```
```suggestion
		G: RuntimeGenesis
	{
```

core/cli/src/lib.rs
```diff
@@ -222,23 +213,294 @@ where
...
+
```
```suggestion
```

core/cli/src/lib.rs
```diff
@@ -222,23 +213,294 @@ where
...
+impl<'a> ParseAndPrepareExport<'a> {
+	/// Runs the command and exports from the chain.
+	pub fn run<F, S, E>(
+		self,
+		spec_factory: S,
+		exit: E,
+	) -> error::Result<()>
+	where S: FnOnce(&str) -> Result<Option<ChainSpec<FactoryGenesis<F>>>, String>,
+		F: ServiceFactory,
+		E: IntoExit {
```
```suggestion
		E: IntoExit
	{
```

core/cli/src/lib.rs
```diff
@@ -222,23 +213,294 @@ where
...
+
```
```suggestion
```

core/cli/src/lib.rs
```diff
@@ -222,23 +213,294 @@ where
...
+impl<'a> ParseAndPrepareImport<'a> {
+	/// Runs the command and imports to the chain.
+	pub fn run<F, S, E>(
+		self,
+		spec_factory: S,
+		exit: E,
+	) -> error::Result<()>
+	where S: FnOnce(&str) -> Result<Option<ChainSpec<FactoryGenesis<F>>>, String>,
+		F: ServiceFactory,
+		E: IntoExit {
```
```suggestion
		E: IntoExit
	{
```

core/cli/src/lib.rs
```diff
@@ -222,23 +213,294 @@ where
...
+
```
```suggestion
```

core/cli/src/lib.rs
```diff
@@ -222,23 +213,294 @@ where
...
+}
+
+impl<'a> ParseAndPreparePurge<'a> {
+	/// Runs the command and purges the chain.
+	pub fn run<G, S>(
+		self,
+		spec_factory: S
+	) -> error::Result<()>
+	where S: FnOnce(&str) -> Result<Option<ChainSpec<G>>, String>,
+		G: RuntimeGenesis {
```
```suggestion
		G: RuntimeGenesis
	{
```

core/cli/src/lib.rs
```diff
@@ -222,23 +213,294 @@ where
...
+
```
```suggestion
```

core/cli/src/lib.rs
```diff
@@ -222,23 +213,294 @@ where
...
+		self,
+		spec_factory: S
+	) -> error::Result<()>
+	where S: FnOnce(&str) -> Result<Option<ChainSpec<G>>, String>,
+		G: RuntimeGenesis {
+		
+		let config = create_config_with_db_path::<(), _, _>(spec_factory, &self.params.shared_params, self.version)?;
+		let db_path = config.database_path;
+
+		if self.params.yes == false {
```
```suggestion
		if !self.params.yes {
```

tomaka: That was already there, but yeah let's change that.

tomaka:
> I'm not sure we want to have this code duplication in every run function.

I feel like the control flow is so much clearer that it is totally worth the code duplication.

> However you change the Service, it will result in a breaking change anyway.

For now I'm not breaking anything. Updating Polkadot can be tricky, so if possible I'm trying to not break stuff.

bkchr:
I know that the current change does not break anything, but we will generate a new warning, that requires a new suppression or a switch to the new function.

I just want to make sure that we don't build any compatible intermediate state, just to not break interfaces. The current Service stuff is not well designed and I don't want that this bleeds into the new design in any way.

bkchr:
(Feel free to merge this)

tomaka:
> I know that the current change does not break anything, but we will generate a new warning, that requires a new suppression or a switch to the new function.

No, what I mean is that my work-in-progress service refactoring (which I haven't shown anyone yet) doesn't break anything.

What I'm doing is: I'm adding commits that introduce new stuff without breaking anything, then on top add commits that remove the old stuff. And then I'd open two PRs (one that adds this intermediary step that you don't like, and one that cleans up the stuff). This way we can upgrade Polkadot in a smoother way, and potentially revert the clean-up if it turns out that I forgot something.

tomaka:
> (Feel free to merge this)

(I'm not in a rush, so will leave this PR a bit open in case more people want to comment on it)

bkchr:
> > I know that the current change does not break anything, but we will generate a new warning, that requires a new suppression or a switch to the new function.
> 
> No, what I mean is that my work-in-progress service refactoring doesn't break anything.
> 
> What I'm doing is: I'm adding commits that introduce new stuff without breaking anything, then on top add commits that remove the old stuff. And then I'd open two PRs (one that adds this intermediary step that you don't like, and one that cleans up the stuff). This way we can upgrade Polkadot in a smoother way, and potentially revert the clean-up if it turns out that I forgot something.

If the intermediate state doesn't break anything, what is the reason to have it at all? The second pr will break stuff anyway and we will need to update Polkadot. 

I think having one concrete pr, will make it easier to iterate on the interface.

Issue #3330, opened 2019-08-07
Node template does not produce blocks in --dev mode


JoshOrndorff:
The substrate node template does not produce blocks when run in `--dev` mode. The full node still works as expected.

## Steps to reproduce
* Clone substrate master (c3b9d6c75ad6b9321bdcdaa93910f07faf918f85 at time of filing this issue)
* `cd node-template`
* `cargo build --release`
* `../target/release/node-template --dev`

## Observed behavior
```2019-08-07 15:48:52 Running in --dev mode, RPC CORS has been disabled.
2019-08-07 15:48:52 Substrate Node
2019-08-07 15:48:52   version 2.0.0-c3b9d6c75-x86_64-linux-gnu
2019-08-07 15:48:52   by Anonymous, 2017, 2018
2019-08-07 15:48:52 Chain specification: Development
2019-08-07 15:48:52 Node name: ajar-flesh-0088
2019-08-07 15:48:52 Roles: AUTHORITY
2019-08-07 15:48:52 Initializing Genesis block/state (state: 0x1ff1…aaa7, header-hash: 0xb680…d71e)
2019-08-07 15:48:52 Loaded block-time = 10000 seconds from genesis on first-launch
2019-08-07 15:48:52 Highest known block at #0
2019-08-07 15:48:52 Using default protocol ID "sup" because none is configured in the chain specs
2019-08-07 15:48:52 Local node identity is: QmZSAqo9qrKh7uStUZucgwoYsnnYQnDW9fQMzZ4gAxBPRf
2019-08-07 15:48:57 Idle (0 peers), best: #0 (0xb680…d71e), finalized #0 (0xb680…d71e), ⬇ 0 ⬆ 0
2019-08-07 15:49:02 Idle (0 peers), best: #0 (0xb680…d71e), finalized #0 (0xb680…d71e), ⬇ 0 ⬆ 0
2019-08-07 15:49:07 Idle (0 peers), best: #0 (0xb680…d71e), finalized #0 (0xb680…d71e), ⬇ 0 ⬆ 0
2019-08-07 15:49:12 Idle (0 peers), best: #0 (0xb680…d71e), finalized #0 (0xb680…d71e), ⬇ 0 ⬆ 0
2019-08-07 15:49:17 Idle (0 peers), best: #0 (0xb680…d71e), finalized #0 (0xb680…d71e), ⬇ 0 ⬆ 0
^C
```

## Expected behavior
[...]

PR #3359, opened 2019-08-09
Switch AuRa to dynamic keystore lookup

bkchr:
Closes: https://github.com/paritytech/substrate/issues/3330


JoshOrndorff:
This does fix the issue for me. node-template now produces blocks as expected.

Issue #3363, opened 2019-08-11
srml-election: Loser candidates are not properly handled.


kianenigma:
This guy, in particular, is never used anywhere, not to mention that node/runtime passes a `()` downwards.

https://github.com/paritytech/substrate/blob/e0f5dd9f394fd19238388054d76de070db6182f3/srml/elections/src/lib.rs#L169-L171

PR #3364, opened 2019-08-12
srml-module: Phragmen election

kianenigma:
https://github.com/paritytech/substrate/blob/88e81084528d7417bd5110f3d8473a324183cd9b/srml/elections-phragmen/src/lib.rs#L17-L63

#### Before merge, should consider: 

- Add the module to the `node/runtime` if needed.

#### Maybe in this PR, maybe later:

- [x] ability to report other voters.
- [ ] ability to remove candidacy.


srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,952 @@
...
+
+			// unreserve the bond and remove.
+			T::Currency::unreserve(&who, T::VotingBond::get());
+			T::Currency::remove_lock(MODULE_ID, &who);
+		}
+
+		/// Submit oneself for candidacy.
+		///
+		/// A candidate will either:
+		///   - Lose at the end of the term and gets their bond unreserved
```
kianenigma:
```suggestion
		///   - Lose at the end of the term and gets their bond unreserved.
```

And I suppose maybe this is too lenient and we should burn/keep the bond for loser candidates? we do the same for the normal approval based election (though it has an issue #3363,but at least the intention is to not return the bond)

srml/staking/src/lib.rs
```diff
@@ -1221,16 +1220,17 @@ impl<T: Trait> Module<T> {
...
 	fn select_validators() -> (BalanceOf<T>, Option<Vec<T::AccountId>>) {
-		let maybe_elected_set = elect::<T, _, _, _>(
-			Self::validator_count() as usize,
-			Self::minimum_validator_count().max(1) as usize,
-			<Validators<T>>::enumerate(),
-			<Nominators<T>>::enumerate(),
-			Self::slashable_balance_of,
-		);
+		let maybe_elected_set = elect::<T::AccountId, BalanceOf<T>, _, T::CurrencyToVote>(
+				Self::validator_count() as usize,
```
gavofyork: bad indentation (one too much).

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,952 @@
...
+				seats,
+				false,
+				0,
+				candidates.clone(),
+				voters_and_votes,
+				Self::locked_stake_of
+			);
+
+			let old_members = <Members<T>>::take();
+			if let Some((new_members_with_approval, _)) = maybe_new_members {
```
gavofyork: we should probably sort both of these to turn it from O(N**2) to O(NlogN).

gavofyork:
runtime version should be bumped.

srml/elections-phragmen/src/lib.rs
kianenigma: great point -- this + `is_candidate()` now operates on sorted lists, hence being _O(NlogN)_, hence making the `submit_candidate()` also _O(NlogN)_.

kianenigma:
cc @AlfonsoCev

srml/staking/src/equalize.rs
```diff
@@ -0,0 +1,148 @@
...
+
+use sr_primitives::traits::{Zero, Convert, Saturating};
+use sr_primitives::phragmen::{ExtendedBalance};
+
+/// Performs equalize post-processing to the output of the election algorithm
+/// This function mutates the input parameters, most noticeably it updates the exposure of
+/// the elected candidates.
+///
+/// No value is returned from the function and the `expo_map` parameter is updated.
+pub fn equalize<T: Trait + 'static>(
```
gavofyork: this should be in sr-prtimitives, too. when i said "phragmen", i meant the whole of the phragmen system, not just he initial phrase.

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+		WithdrawReasons, ChangeMembers
+	}
+};
+use system::{self, ensure_signed, ensure_root};
+
+const MODULE_ID: LockIdentifier = *b"phrelect";
+
+type BalanceOf<T> = <<T as Trait>::Currency as Currency<<T as system::Trait>::AccountId>>::Balance;
+
+pub const DEFAULT_CANDIDACY_BOND: u32 = 9;
```
gavofyork: wtf are these? units? docs?

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+	type CandidacyBond: Get<BalanceOf<Self>>;
+
+	/// How much should be locked up in order to be able to submit votes.
+	type VotingBond: Get<BalanceOf<Self>>;
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as PhragmenElection {
+		// ---- parameters
+		/// Number of seats to elect.
```
gavofyork: if this changes, does this take effect immediately? or does it only take effect at the next election? if so is there a new election immediately?

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+		pub DesiredSeats get(desired_seats) config(): u32;
+		/// The total number of vote rounds that have happened, exclusive of the upcoming one.
+		pub ElectionRounds get(election_rounds): u32 = Zero::zero();
```
gavofyork: surely this is `State`

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+		fn deposit_event<T>() = default;
+
+		const CandidacyBond: BalanceOf<T> = T::CandidacyBond::get();
+		const VotingBond: BalanceOf<T> = T::VotingBond::get();
+
+		/// Vote for a set of candidates for the upcoming round of election.
+		///
+		/// The `votes` should:
+		///   - not be empty.
+		///   - not be less than the number of candidates.
```
gavofyork: surely the `not` is wrong here.

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+		///
+		/// `who` cannot be an old voter who has not yet called [`remove_voter`].
```
gavofyork:
```suggestion
		/// `who` cannot be an existing voter
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+
+				// lock and reserve
+				T::Currency::reserve(&who, T::VotingBond::get())
+					.map_err(|_| "voter can not pay voting bond")?;
+				T::Currency::set_lock(
+					MODULE_ID,
+					&who,
+					locked_balance,
+					T::BlockNumber::max_value(),
+					WithdrawReasons::except(WithdrawReason::TransactionPayment),
```
gavofyork: this is fine for polkadot where 80% of tx fees go to the treasury, but if ever anyone did a chain where 100% transaction fees are given to the block author and that had arbitrary tipping in, then it would be a huge security flaw. you would simply lock your account, send a dumb transaction with a huge tip for a block that you authored, then collect the entirety of the "locked" (now unlocked) funds as the block author. the lock would remain in place, but the funds would not.

to make everything safer, you should provide a second argument `value: BalanceOf<T>` and that should be the amount that is locked and which is used for the purposes of voting. it should be interpreted as `value.min( T::Currency::total_balance(&who))` to avoid the need to faff around calculating how much is left in the account after tx fees if you just want to lock all of it.

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+			}
+
+			let now = Self::election_rounds();
+			<VotesOf<T>>::insert(&who, (votes, now));
+		}
+
+		/// Remove `origin` as a voter. One can use this to
+		///   1. _undo_ their votes before the election has started.
+		///   2. Remove their lock and unreserve the bond after an election round.
+		#[weight = SimpleDispatchInfo::FixedNormal(500_000)]
```
gavofyork: seems a bit steep. it's only doing four storage writes and a couple of reads.

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+			T::Currency::unreserve(&who, T::VotingBond::get());
+			T::Currency::remove_lock(MODULE_ID, &who);
+		}
+
+		/// Submit oneself for candidacy.
+		///
+		/// A candidate will either:
+		///   - Lose at the end of the term and gets their bond unreserved
+		///   - Win and become a member. Members will get their bond back at the end ot their term.
+		#[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
```
gavofyork: also seems very steep. where do these values come from?

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+				.map_err(|_| "candidate has not enough funds")?;
+
+			<Candidates<T>>::mutate(|c| c.insert(index, who));
+			CandidateCount::mutate(|c| *c += 1);
+		}
+
+		/// Set the desired member count; if lower than the current count, then seats will not be up
+		/// election when they expire. If more, then a new vote will be started if one is not
+		/// already in progress.
+		#[weight = SimpleDispatchInfo::FixedOperational(10_000)]
```
gavofyork: does one storage write and it's 10k. the one above does 3 mutations and costs 100x as much. wtf?

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+		/// Note: A tally should happen instantly (if not already in a presentation
+		/// period) to fill the seat if removal means that the desired members are not met.
+		#[weight = SimpleDispatchInfo::FixedOperational(10_000)]
+		fn remove_member(origin, who: <T::Lookup as StaticLookup>::Source) {
+			ensure_root(origin)?;
+			let who = T::Lookup::lookup(who)?;
+
+			if Self::is_candidate(&who).is_ok() {
+				let members = Self::members();
+				let new_members: Vec<T::AccountId> = members
```
gavofyork: just use `members.remove(&who)`

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+					.into_iter()
+					.filter(|i| *i != who)
+					.collect();
+				<Members<T>>::put(new_members.clone());
+				T::Currency::unreserve(&who, T::CandidacyBond::get());
+				T::ChangeMembers::change_members(&[], &[who], new_members);
+			}
```
gavofyork: this doesn't seem to automatically kick-off a new election. it should.

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+		}
+	}
+}
+
+decl_event!(
+	pub enum Event<T> where <T as system::Trait>::AccountId {
+		/// A new term with new members.
+		NewTerm(Vec<AccountId>),
+		/// No candidates were elected for this round.
+		EmptyCouncil(),
```
gavofyork:
```suggestion
		EmptyCouncil,
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+
+	/// Check if `who` was a voter who has not yet called `remove_voter`.
+	fn is_old_voter(who: &T::AccountId) -> bool {
+		let now = Self::election_rounds();
+		Self::votes_of(who).1 < now && <StakeOf<T>>::exists(who)
+	}
+
+	/// The locked stake of a voter.
+	fn locked_stake_of(who: &T::AccountId) -> BalanceOf<T> {
+		if Self::is_current_voter(who) {
```
gavofyork: This is super-expensive for a sanity check. Why not just use `stake_of` directly?

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,957 @@
...
+
+				// return bond to losers.
+				candidates.into_iter().for_each(|c| {
+					// if this candidate is a loser return the bond
+					if new_members.binary_search(&c).is_err() {
+						T::Currency::unreserve(&c, T::CandidacyBond::get());
+					}
+				});
+
+				T::ChangeMembers::change_members(
```
gavofyork: Incorrect. Use `T::ChangeMembers::set_members_sorted` if you don't know who the actual incoming and outgoing members are.

srml/staking/src/equalize.rs
kianenigma: It is a bit complicated to do that now. I am refactoring phragmen in a separate branch and if we don't want to equalize for council yet (there is no reason to need it since we are not using edge/weight data; only the list of winners) I prefer doing it later on after my refactor. There's a TODO for it at the top of the file. 

```
// TODO #3365: deprecate this file, decouple from staking and move to sr_primitives.
```

srml/elections-phragmen/src/lib.rs
kianenigma: idk these were added as configs to council modules some few weeks ago e.g.

https://github.com/paritytech/substrate/blob/a622361559b31f1eb9eb8ce222ad6f94bf4b79c4/srml/elections/src/lib.rs#L144-L151

kianenigma: but now I see they are unused in both places apparently. Will clean up.

srml/elections-phragmen/src/lib.rs
kianenigma: nope; the change will happen on next elect. Will add a doc for it.

srml/elections-phragmen/src/lib.rs
kianenigma: actually: it _can be an existing voter_ to _update_ their vote. 

It cannot be, exactly as phrased: _old voter who has not yet called `remove_voter`_

srml/elections-phragmen/src/lib.rs
kianenigma: Fair enough and great catch -- then this needs updating in both election versions. We also do the same in the other election module. 

I interpret the suggested approach simply as: you as the voter decide how much of your stake is locked and taken into account when weighing the votes in either approval voting or phragmen. You are also responsible to keep some of your balance unlocked to be able to pay for some future txs.

srml/elections-phragmen/src/lib.rs
kianenigma: about your 3 comments on the steepness of weights:

1. they are taken by approximately halving the weights that are decided for the other election (maybe a wrong decision, but that's not the main reason of your complaint) via benchmarking, which is _not through manual analysis_ but rather execution time measuring and _some manual juice added by me_. Some examples:

    1. submitting a candidate (as you recently discussed in some chats) is quite expensive with regards to computing the outcome of the election so I bumped it up a bit.
    2. removing a voter and adding one is more or less the same in terms of cpu/io but I manually reduced the weight of removing a bit for the same reason: it will potentially reduce the block time of the next election.


2. and more importantly this type of weight that you are analyzing, while I agree with it, is the `state-weight` which we, _could_, but at the moment have not yet introduced anywhere. Hence, I didn't apply it here either. The current weight numbers are merely a representation of computation time. Based on this, setting the desired seats could be arguably substantially less computationally intensive than submitting candidacy, hence the 100x difference.

srml/elections-phragmen/src/lib.rs
kianenigma: okay now I see the reason for the comment above: will start a new election on this call.

srml/elections-phragmen/src/lib.rs
kianenigma: yeah -- my msitake.

srml/elections-phragmen/src/lib.rs
gavofyork: not sure i understand point (2) could you rephrase?

in general, weight should not be used to encode indirect costs because it is used to determine raw computation complexity of a block and is only incidentally used for fees.

if it's one-off indirect cost, then an additional non-weight fee is reasonable, to be charged in a discrete instance as part of the call. if it's an on-going cost, then an accordingly large deposit should be held (`ReservableCurrrency` can be used for this).

srml/elections-phragmen/src/lib.rs
kianenigma: Okay but wait: 

I assume the purpose of this is to be able to 1) kick out one 2) but keep the rest 3) and find a replacement. This is a bit tricky to do with the current system because we don't store any old vote/candidacy info. 

Assume term durations is 5, [1,2] are elected at block 10. as mentioned, all voters and cadidates are removed from before block 10. at block 13 we kick 2. Now, based on what voting data and candidacy should we start a new election? 

I see two possible answers:
1- the ongoing votes and candidacy that happened in block 10 until 13. In this case, removing one member and starting a new election is moot and this should just be a function named `force_new_election(root)`.
2- elect _only one new member_ from ongoing votes and candidacies in block range [10, 13] (if any) and combine it with the remainder of the previous set.

srml/elections-phragmen/src/lib.rs
gavofyork: is "old voter" a technical term?

srml/elections-phragmen/src/lib.rs
gavofyork: right (or you can lock all of it and transfer some extra in when you need to transact)

srml/elections-phragmen/src/lib.rs
kianenigma: The full story is: 

The code does not clean the votes at all and relies on two incentivization that voters themselves will do it. As a voter, once your vote has been counted once, then you are an _old voter_  aka. retired if you want to name it like that. You should manually remove yourself via another tx because:

1. you want your reserved bond back.
2. you can't vote and participate anymore in rounds.

If you are fine with the logic I can formulate it better as _inactive voter_  is the module docs. I agree that the term _old voter_ is not clear.

srml/elections-phragmen/src/lib.rs
kianenigma: But I don't understand from reading the docs of `trait ChangeMemeber` why in if a single one is kicked out `T::ChangeMembers::change_members(&[], &[who], members);` is fine? Based on your comment, shouldn't all of them be `set_members_sorted`?

kianenigma:
Pushed patches to most review comments. 

The main blockers are: 

1. how to kick someone: https://github.com/paritytech/substrate/pull/3364#discussion_r314248503

2. you mentioned that the concept of runner ups should be added here. But I would note that runner ups (at least in my head right now) are merely a small logic change and I don't intend to make it as complicated as the main council election: 
  - currently, all losers are wiped and their bond is returned. This should change to:
  - grab the top K of the losers and make them a candidate for the next round. Burn the bond for the rest. 

So even no new storage item is needed to store them; they are just automatic appends to the candidate list of the next round. Is this aligned with what you have in mind or too simple? 

Optionally, we can store them and the logic of `remove_member` will be: 
1. remove 2. if a runner up exists 3. pop and use the best runner up as the replacement. 4. else do nothing.

srml/elections-phragmen/src/lib.rs
kianenigma: > not sure i understand point (2) could you rephrase?

What I mean is that judging these based on how many mutate/read/writes they have is _correct_ but not something that we yet do; the current weight in all modules is merely based on execution time. regardless of what they do internally. Totally black-boxed. analyzing read/write/mutate etc. will be one of the early next steps after Kusama.  

> ... perverse incentive for validators to do less work by choosing transactions like these that are over-priced because they're accounting for a future cost, to be born by Some Other Guy.

Noted and I can conclude what the little bits of my manual tweaks are wrong. Will reverse them + fix the rest of them in further weight updates.

srml/elections-phragmen/src/lib.rs
gavofyork: we should store all candidate information except that which is culled.

gavofyork:
That seems reasonable, but you also need to ensure that everyone who won is also retained in the next election (whether it's at the end of the period or before because of a member leaving).

kianenigma:
a bunch of requested features added. Best explained by the name of these test cases: 

```rust 
// All current members are appended to the candidate list for the next election
fn current_members_are_always_implicitly_next_candidate() {}

fn runner_ups_should_be_kept() {}

// IFF remove_member and no runner-ups, start a new phragmen, otherwise use the runner-ups
fn remove_members_triggers_election_if_no_runner_up() {}
fn unexpected_election_will_use_runner_ups_if_exists() {}

// instead of just giving it back
fn losers_will_lose_the_bond() {}

// anyone who wins will not lose any money
fn outgoing_will_get_the_bond_back() {}

```

Also updated the module doc at the top. Would be good to give it a read.

srml/elections-phragmen/src/lib.rs
gavofyork: `defunct_voter`

srml/elections-phragmen/src/lib.rs
gavofyork: they're just meant for different situations. if you naturally have the diff ahead of time (like elections module does), then you can avoid some computation and provide it directly. otherwise, you pass the pre-sorted set and have it figure out the diff for you.

srml/support/src/traits.rs
```diff
@@ -652,6 +652,16 @@ pub trait ChangeMembers<AccountId: Clone + Ord> {
 	/// Set the new members; they **must already be sorted**. This will compute the diff and use it to
 	/// call `change_members_sorted`.
 	fn set_members_sorted(new_members: &[AccountId], old_members: &[AccountId]) {
+		let (incoming, outgoing) = Self::compute_members_diff(new_members, old_members);
```
gavofyork: why?

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+			allowed_runner_ups = allowed_runner_ups.min(desired_runner_ups);
+			let runner_ups = new_members_filtered.into_iter()
+				.take(allowed_runner_ups)
+				.collect::<Vec<T::AccountId>>();
+			// should still be sorted by stake. Low to high.
+			<RunnerUps<T>>::put(runner_ups.clone());
+
+			// report. We don't know/compute the diff here, therefore we call `set_member`.
+			let (incoming, outgoing) = T::ChangeMembers::compute_members_diff(&new_members, &old_members);
+			T::ChangeMembers::change_members_sorted(
```
gavofyork: why unrolling the API? just use `set_members_sorted`.

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+
+//! # Phragmen Election Module.
+//!
+//! An election module based on sequential phragmen.
+//!
+//! ### Term and Round
+//!
+//! The election happens in _rounds_: every `N` blocks, all previous members are retired and a new
+//! set is elected (which may or may not have an intersection with the previous set). Each round
+//! lasts for some number of blocks defined by `TermDuration` storage item. The works _term_ and
```
gavofyork:
```suggestion
//! lasts for some number of blocks defined by `TermDuration` storage item. The words _term_ and
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+//! _round_ can be used interchangeably in this context.
+//!
+//! `TermDuration` might change during a round. This can shorten or extend the length of the round.
+//! The next election round's block number is never stored but rather always checked on the fly.
+//! Based on the current block height and `TermDuration`, `BlockNumber % TermDuration == 0` will
+//! always trigger a new election round.
+//!
+//! ### Voting
+//!
+//! Voters can vote for as many of the candidates by providing a list of account ids. Invalid votes
```
gavofyork:
```suggestion
//! Voters can vote for any set of the candidates by providing a list of account ids. Invalid votes
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+//! (voting for non-candidates) are ignored during election. Yet, a voter _might_ vote for a
+//! future candidate. Voters reserve a bond as they vote. Each vote defines a `value`. This amount
+//! is locked from the account of the voter and indicated the weight of the vote. Voters can update
```
gavofyork:
```suggestion
//! is locked from the account of the voter and indicates the weight of the vote. Voters can update
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+//! their votes at any time by calling `vote()` again. This keeps the bond untouched but can
+//! optionally change the locked `value`. After a round, votes are kept and might still be valid
+//! for further rounds. A voter is responsible to call `remove_voter` upon once they are done to
```
gavofyork:
```suggestion
//! for further rounds. A voter is responsible to call `remove_voter` once they are done to
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+//! have their bond back and remove the lock.
+//!
+//! ### Candidacy and Members
+//!
+//! Candidates also reserve a bond as they submit candidacy. A candidate cannot take their candidacy
+//! back. A candidate can end up in one of the below situations:
+//!   - **Winner**: A winner is kept as a _member_. They must still have a bond in reserve
+//!		and they are automatically counted as a candidate for the next election.
+//!   - **Runner-up**: Runner ups are the best candidates immediately after the winners. The number
```
gavofyork: how do runners-up work with phragmen?

kianenigma: yeah, it is a bit quirky and so far I only introduced it for the sake of having runner-ups, as requested during reviews. 

I do this to compute it without a lot of overhead in a way that doesn't seem horribly wrong: 

- if you want `m` members and `r` runnerups, call into phragmen to give you a result set of size `m + r`.
- Sort them based on the backed stake and take the best `m` as winners and the rest as runner ups. 

Though, as mentioned, I'm actually more leaned toward having no runner ups in this election. It would make it substantially simpler and as you pointed out the concept doesn't really exist in Phragmen by default.

srml/elections-phragmen/src/lib.rs
kianenigma: Because I use the diff which is computed within `set_member_sorted` somewhere a bit further down and --call me selfish for changing `ChangeMember` according to my uses case-- I think it is better than duplicating the code that computes the diff of incoming/outgoing set.

srml/elections-phragmen/src/lib.rs
gavofyork: right; let's just go with the no-runners-up case. not sure what i was thinking during the reviews.

srml/elections-phragmen/src/lib.rs
kianenigma: This is already updated and we handle the loser bonds via an imbalance.

kianenigma:
Aside from the runner ups being gone, I went through this A-Z one more time and some final notes about it:

- Both the number of voters (we iterate through all of them to pass them to phragmen) and the edges of each vote are potential points of spamming and exploit. The former requires a sensible value for `VoterBond` and the latter is capped for now by:
```
/// The maximum votes allowed per voter.
pub const MAXIMUM_VOTE: usize = 16;
```

- Since we
  - Keep the old voters and count them in.
  - Assume current members are always candidates for the next election round, 
  the overall election round has a property that: if a set is elected, given that there will be no further input and change to the system, the same set keeps getting re-elected all the time. 


If no further critical requirements are needed + this is not going to be added to a runtime asap, would be nice to merge this initial branch in since I could really use the changes that are made to phragmen in #3456. If we're going to ice or stale this, then I might move these changes over.

srml/staking/src/equalize.rs
kianenigma: this is actually done with the new rebase.

kianenigma:
added the ability to report other voters; this note popped up while doing it:

```
// TODO: we don't hold the voters accountable in any sense; a voter can remove themselves (unbond)
// right after an election (even if they voted for some winners) and they can be reported right
// after an election. We should probably add some accountability here. Simplest would be that voters
// cannot remove/report as long as they have a voted target who is in the member list.
```

Also rebased with new phragmen which is now entirely in `core`. Could rest now until the reviewer is back.

srml/support/src/traits.rs
gavofyork: ???

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1405 @@
...
+	/// Handler for the unbalanced reduction when a member has been kicked.
+	type KickedMember: OnUnbalanced<NegativeImbalanceOf<Self>>;
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as PhragmenElection {
+		// ---- parameters
+		/// Number of members to elect.
+		pub DesiredMembers get(desired_members) config(): u32;
+		/// How long each seat is kept. This defined the next block number at which an election
```
gavofyork:
```suggestion
		/// How long each seat is kept. This defines the next block number at which an election
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1405 @@
...
+		/// round will happen.
+		pub TermDuration get(term_duration) config(): T::BlockNumber;
+
+		// ---- State
+		/// The current elected membership. Sorted based on account id.
+		pub Members get(members) config(): Vec<T::AccountId>;
+		/// The total number of vote rounds that have happened, exclusive of the upcoming one.
```
gavofyork:
```suggestion
		/// The total number of vote rounds that have happened, excluding the upcoming one.
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1405 @@
...
+		/// # <weight>
+		/// O(1)
+		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedNormal(500_000)]
+		fn vote(origin, votes: Vec<T::AccountId>, value: BalanceOf<T>) {
+			let who = ensure_signed(origin)?;
+
+			// TODO: use decode_len #3071 and remove candidate_count.
+			let candidates_count = Self::candidate_count();
+			// addition is valid: candidates adn members never overlap.
```
gavofyork:
```suggestion
			// addition is valid: candidates and members never overlap.
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1405 @@
...
+
+			Self::do_remove_voter(&who, true);
+		}
+
+		/// Report `target` for being an defunct voter. In case of a valid report, the reporter is
+		/// rewarded by the bond amount of `target`. Otherwise, the reporter itself is removed and
+		/// their bond is slashed.
+		///
+		/// A defunct voter is defined to be:
+		///   - a voter who's current submitted votes are all invalid. i.e. all of them are no
```
gavofyork:
```suggestion
		///   - a voter whose current submitted votes are all invalid. i.e. all of them are no
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1405 @@
...
+				Self::do_remove_voter(&reporter, false);
+			}
+			Self::deposit_event(RawEvent::VoterReported(target, reporter, valid));
+		}
+
+
+		/// Submit oneself for candidacy.
+		///
+		/// A candidate will either:
+		///   - Lose at the end of the term and get slashed.
```
gavofyork: third option: be a runner up for the next vote.

gavofyork:
```suggestion
		///   - Lose at the end of the term and forfeit their deposit.
```

kicking everyone except the winners is no good. we really want to retain a number of well-backed runners-up while kicking the others. two options spring to mind:

- re-run phragmen with all stakes on the losers to determine N runners-up.
- re-run phragmen with all stakes to determine seats + N winners + runners-up, then remove the winners from the set to give runners-up.

kianenigma: Option B is what i was doing before (in 1 run, not twice) and we sorta decided to remove it as it not really a runner up set. 

Nonetheless it might be the only way forward. I also think it could, in practice, always yield the same results as option A. Either way I will have a chat with Al and the research team before taking this forward.

gavofyork: yeah but before was no good as we corrupted the winning set in order to find runners-up. option B is different in that we run phragmen twice with a different size of winning set, thereby avoiding such conflicts.

kianenigma: brought back the runner ups.

gavofyork: (runners up) :-)

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1571 @@
...
+//!
+//! Candidates also reserve a bond as they submit candidacy. A candidate cannot take their candidacy
+//! back. A candidate can end up in one of the below situations:
+//!   - **Winner**: A winner is kept as a _member_. They must still have a bond in reserve and they
+//!     are automatically counted as a candidate for the next election.
+//!   - **Loser**: Any of the candidate who are not a winner are left as losers. A loser might be an
+//!     _outgoing member_, meaning that they are an active member who failed to keep their spot. In
+//!     this case, the outgoing member will get their bond back. Otherwise, the bond is slashed from
+//!     the loser candidate.
+//!   - **Runner-up**: Runner ups are the best candidates immediately after the winners. The number
```
gavofyork:
```suggestion
//!   - **Runner-up**: Runners-up are the best candidates immediately after the winners. The number
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1571 @@
...
+//!     of runner-ups to keep is configurable. Runner-ups serve one major role: if a member is
```
gavofyork:
```suggestion
//!     of runner-ups to keep is configurable. Runners-up serve one major role: if a member is
```

The main point of runners-up is not to provide replacements for removed candidates (indeed, this might be a security risk). Rather it's to ensure that the number of candidates is kept fairly low by periodically culling the crappy candidates (and burning their deposit to disincentivise frivolous usage).

srml/elections-phragmen/src/lib.rs
gavofyork: ^^^ Is this the case yet @kianenigma ?

srml/elections-phragmen/src/lib.rs
kianenigma: My writing here (everything after: _Runner-ups serve one __major__ role:..._ should be gone) is too constrained, I agree, but I don't also quite get what you mean by:

> Rather it's to ensure that the number of __candidates__ is kept fairly low by periodically culling the crappy candidates

I don't see how the runner-ups limit the number of candidates; candidates always strat from an empty list at the beginning of a term and are always thrown away (and forfeit bond) after election happens.

srml/elections-phragmen/src/lib.rs
```diff
@@ -0,0 +1,1606 @@
...
+	/// Reads: O(C + V*E) where C = candidates, V voters and E votes per voter exits.
+	/// Writes: O(M + R) with M desired members and R runners_up.
+	/// # </weight>
+	fn do_phragmen() {
+		let desired_seats = Self::desired_members() as usize;
+		let desired_runners_up = Self::desired_runners_up() as usize;
+		let num_to_elect = desired_runners_up + desired_seats;
+
+		let mut candidates = Self::candidates();
+		// candidates who exactly called `submit_candidacy`. Only these folks are at the risk of
```
gavofyork:
```suggestion
		// candidates who explicitly called `submit_candidacy`. Only these folks are at the risk of
```

gavofyork: (?)

kianenigma: The terminology might be strange once seen isolated, but here's the full story: 

As requested and discussed, current members and runners up are kept as candidates for future rounds. To do this, we don't duplicate them into `Candidate<T>` but (the way I named it) _implicitely_ add them as candidates. As opposed to this, some ppl will always be candidates who made it in by `submit_candidacy()`. I named this _explicit candidates_. Both namings can work.

PR #3366, opened 2019-08-12
document that session key generation stores keys

rphmeier:


core/service/src/components.rs
```diff
@@ -145,10 +145,11 @@ pub type PoolApi<C> = <C as Components>::TransactionPoolApi;
 pub trait RuntimeGenesis: Serialize + DeserializeOwned + BuildStorage {}
 impl<T: Serialize + DeserializeOwned + BuildStorage> RuntimeGenesis for T {}
 
-/// Something that can create initial session keys from given seeds.
+/// Something that can create and store initial session keys from given seeds.
 pub trait InitialSessionKeys<C: Components> {
-	/// Generate the initial session keys for the given seeds.
-	fn generate_intial_session_keys(
+	/// Generate the initial session keys for the given seeds and store them in
+	/// an internal keystore.
```
bkchr:
```suggestion
	/// the keystore.
```

Otherwise it sounds like just "some" keystore to me.

core/session/src/lib.rs
```diff
@@ -37,7 +39,8 @@ client::decl_runtime_apis! {
 	}
 }
 
-/// Generate the initial session keys with the given seeds.
+/// Generate the initial session keys with the given seeds and store them in
+/// the client's keystore.
```
bkchr:
```suggestion
/// the node's keystore.
```

core/session/src/lib.rs
```diff
@@ -29,6 +29,8 @@ client::decl_runtime_apis! {
 	/// Session keys runtime api.
 	pub trait SessionKeys {
 		/// Generate a set of session keys with optionally using the given seed.
+		/// The keys should be stored within the keystore exposed via runtime
```
bkchr: It is correct that we do it this way, but this is not a requirement of this api.

core/service/src/components.rs
rphmeier: what is "the" keystore?
From the perspective of the trait, which is an _abstract interface_, it should be "some" keystore -- I'm glad it came across correctly.

core/session/src/lib.rs
rphmeier: It's the client's keystore, since that's the parameter to the function.

core/session/src/lib.rs
rphmeier: This documentation makes it a requirement.

PR #3370, opened 2019-08-12
Implement Deserialize for SignedBlock and Header.

dvc94ch:
Required for sharing `SignedBlock` and `Header` with users so they don't have to redefine them.


bkchr:
The `::` before extern crates it not required anymore with rust 2018

PR #3371, opened 2019-08-12
Add code owners file

pepyakin:
Add CODEOWNERS file and list myself as owners for `srml-contracts` and `substrate-executor` crates. After this PR, I will be requested for a review for each PR that modifies code in owned crates. 

Note that this doesn't imply that my approve became required for merging code for these crates.


bkchr:
Can we do this with github handles as well? Or does this require an email?

pepyakin:
We can do with github handles if required. I thought using email would be a bit more appropriate but I am not strong on this.

PR #3374, opened 2019-08-12
Era change can be forced on and forced off

gavofyork:
Before it was only possible to force an era. Now we can force eras not to happen, useful for the PoA period of Kusama.

### DIFF
--- a/srml/staking/src/lib.rs
+++ b/srml/staking/src/lib.rs
@@ -555,6 +555,22 @@ pub trait Trait: system::Trait {
 	type SessionInterface: self::SessionInterface<Self::AccountId>;
 }
 
+/// Mode of era-forcing.
+#[derive(Copy, Clone, PartialEq, Eq, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(Debug, Serialize, Deserialize))]
+pub enum Forcing {
+	/// Not forcing anything - just let whatever happen.
+	NotForcing,
+	/// Force a new era, then reset to `NotForcing` as soon as it is done.
+	ForceNew,
+	/// Avoid a new era indefinitely.
+	ForceNone,
+}
+
+impl Default for Forcing {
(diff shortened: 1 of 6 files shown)


srml/staking/src/lib.rs
```diff
@@ -555,6 +555,22 @@ pub trait Trait: system::Trait {
...
+/// Mode of era-forcing.
+#[derive(Copy, Clone, PartialEq, Eq, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(Debug, Serialize, Deserialize))]
+pub enum Forcing {
+	/// Not forcing anything - just let whatever happen.
+	NotForcing,
+	/// Force a new era, then reset to `NotForcing` as soon as it is done.
+	ForceNew,
+	/// Avoid a new era indefinitely.
+	ForceNone,
```
rphmeier: more idiomatic would be to have `Forcing::New` and `Forcing::Avoid`, with storage holding an `Option<Forcing>`

rphmeier: (typical Rust style would call this `Force` as opposed to `Forcing`)

gavofyork: yeah, didn't manage to make that work with the module JSON config stuff. let me try again though - might have been some other issue.

gavofyork: yeah, genesis configs require `Option` items to be `Some` value.

gavofyork: See branch `gav-disable-eras-alt`

rphmeier: OK, that's a bit unfortunate. but no reason to block this PR on fixing the macro

rphmeier: Regardless, the repetition of `Force` in the variant names is redundant and it would be a bit cleaner without

PR #3375, opened 2019-08-12
subkey network default works

gavofyork:
Old behaviour didn't take account of `--network` flag when inspecting a derived address.


gavofyork:
test fail is unrelated. @tomaka ?

PR #3376, opened 2019-08-12
core/network: Surface peerset's `set_priority_group` in NetworkService

mxinden:
`PeerSetHandle.set_priority_group` allows modifying a priority group by
group identifier. With this commit the function can be accessed through
`NetworkService`.

This is need in order for a validator to connect to as many other
validators as configured without reserving a specific connection slot
for them.

This is needed for #3247, see https://github.com/paritytech/substrate/pull/3247#issuecomment-516351141 for details.

### DIFF
--- a/core/network/src/config.rs
+++ b/core/network/src/config.rs
@@ -163,7 +163,8 @@ impl ProtocolId {
 	}
 }
 
-/// Parses a string address and returns the component, if valid.
+/// Parses a string address and splits it into Multiaddress and PeerId, if
+/// valid.
 ///
 /// # Example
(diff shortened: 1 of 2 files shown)


arkpar:
Just a note: `add_to_priority_group` and `remove_from_priority` group might be useful as well.

mxinden:
> Just a note: add_to_priority_group and remove_from_priority group might be useful as well.

As of right now I don't need those functions for https://github.com/paritytech/substrate/pull/3247. Do you still want me to add them here @arkpar?

Issue #2689, opened 2019-05-25
Merge common code in AuRa and BABE


DemiMarie-parity:
AuRa and BABE have a massive amount of code that is duplicated between them.  Most of the changes I have made to one have also had to be made to the other.  As such, it is prime time to merge the common code.

The merged code should go in `core/consensus/slots`, as much of it is tightly coupled to slot-based algorithms.

bowd:
Was this closed by #3377 or is there more to it? I'm looking around something to work on as my first issue 🚀

PR #3377, opened 2019-08-13
consensus: refactor aura and babe proposer

andresilva:
Introduce `SimpleSlotWorker` trait which implements `on_slot` with the current logic shared between Aura and Babe.


rphmeier:
I guess this closes https://github.com/paritytech/substrate/issues/2689 ?

andresilva:
There's still some code duplication in the `Verifier` and related to inherent data. The `Verifier` I'm not sure it's worth it to abstract though.

PR #3378, opened 2019-08-13
Make template module instantiable

JoshOrndorff:
Make the module template at `node-template/runtime.src/template.rs` an instantiable module.

This PR draft is primarily to discuss whether making the template instantiable is a good idea. It always helps to have something concrete to refer to when making such decisions.

**Pros**
* Most (all?) modules may as well be instantiable
* Developers will not need to learn instantiable modules as a "new feature", it's just how modules are
* Code complexity increase is minimal
* With `DefaultInstance` including the module in a runtime is identical to non-instantiable.

**Cons**
* Code complexity increase is non-zero
* Probably increases compile time for each module (I'll do some preliminary benchmarking)


JoshOrndorff:
Thanks for confirming that the code is fine. I'm also interested in getting feedback about whether this is a change we _want_ to include. I've explained above why I think it is, but before I merge, would appreciate some more feedback. I've requested a few reviewers.

TL;DR Do we want the template module to be instantiable by default?

shawntabrizi:
I am of the opinion this is not a change we want.

The default "template" for users should not be that their module is instantiable. Furthermore, explaining to a user what an instantiable module is, and having them add extra generic types (which is already one of the more confusing parts of Rust), I dont think is helpful.

I suggest you create your own repository, either a `substrate-node-template` or a `substrate-module-template` where you provide this instantiable module template, and then if users ask for it, you can point them there and get a feel for how much people actually want this.

My guess is that the need for this is low.

gavofyork:
Agree with @shawntabrizi ; until the syntax for instantiatable modules improves sufficiently to remove non-instantiatable modules altogether, then we'll want to keep the non-instantiatable as the default.

PR #3380, opened 2019-08-13
babe: Introduce secondary slots (Aurababeous)

andresilva:
Depends on #3377.

This PR extends the BABE consensus engine into Aurababeous. There are now two kinds of slot assignments: Primary (VRF-based) and Secondary (Slot-number based). Primary slots take precedence over secondary slots, when authoring the node starts by trying to claim a primary slot and falls back to a secondary slot claim attempt. The fork choice rule is weight-based, where weight equals the number of primary blocks in the chain. We will pick the heaviest chain and will go with the longest one in case of a tie.

The primary block author is chosen using the BABE VRF function threshold calculation. The secondary block author is chosen using: `blake2_256(epoch_randomness ++ slot_number) % authorities_len`. 

I decided to keep the same name for the engine instead of renaming it or duplicating it. There are currently open PRs for BABE (#3301) that we want to include for Kusama and the rename would create unnecessary conflicts. I also decided not to create a new standalone engine because re-using the runtime module code is not trivial, and if we disable secondary slot assignment the engine should behave exactly like straight BABE. Additionally, I think we should keep the name BABE and just extend the spec (the slot assignment rule for secondary proposals isn't related to Aura at this point anyway).

@bkchr Suggested naming it backup slot instead. What do you think? I actually prefer primary/secondary.

### TODO

- [x] - Fix tests
- [x] - Documentation
- [x] - Allow disabling secondary slot assignment from runtime
- [x] - Change secondary slot assignment rule


burdges:
I agree with keeping the same name, but..

Any way we can support the aura-like blocks running on a much slower bock time than the babe slots?  We should keep a testnet running with babe's erratic block times because anything that breaks indicates deeper problems.

rphmeier:
> I actually prefer primary/secondary.

👍 - the term "primary" has some pedigree for this purpose in the literature as well. extending to "secondary" makes sense.

core/consensus/babe/primitives/src/digest.rs
```diff
@@ -36,46 +36,136 @@ use schnorrkel::{
...
+		/// VRF proof
+		vrf_proof: VRFProof,
+		/// Authority index
+		authority_index: super::AuthorityIndex,
+		/// Slot number
+		slot_number: SlotNumber,
+		/// Chain weight (measured in number of Primary blocks)
+		weight: BabeBlockWeight,
+	},
+	/// A secondary deterministic slot assignmnet.
```
joepetrowski:
```suggestion
	/// A secondary deterministic slot assignment.
```

core/consensus/babe/primitives/src/digest.rs
```diff
@@ -36,46 +36,136 @@ use schnorrkel::{
...
+		/// Slot number
+		slot_number: SlotNumber,
+		/// Chain weight (measured in number of Primary blocks)
+		weight: BabeBlockWeight,
+		/// VRF output
+		vrf_output: [u8; VRF_OUTPUT_LENGTH],
+		/// VRF proof
+		vrf_proof: [u8; VRF_PROOF_LENGTH],
+	},
+	/// A secondary deterministic slot assignmnet.
```
joepetrowski:
```suggestion
	/// A secondary deterministic slot assignment.
```

core/consensus/babe/src/lib.rs
```diff
@@ -416,67 +469,184 @@ fn check_header<B: BlockT + Sized, C: AuxStore, T>(
...
+	}
+
+	let pre_digest = CompatibleDigestItem::babe_pre_digest(pre_digest);
+	Ok(CheckedHeader::Checked(header, (pre_digest, seal)))
+}
+
+/// Check a primary slot proposal header. We validate that the given header is
+/// properly signed by the expected authority, and that the contained VRF proof
+/// is valid. Additionally, the weight of this block must increase compared to
+/// it's parent since it is a primary block.
```
joepetrowski:
```suggestion
/// its parent since it is a primary block.
```

core/consensus/babe/src/lib.rs
```diff
@@ -416,67 +469,184 @@ fn check_header<B: BlockT + Sized, C: AuxStore, T>(
...
+		Ok(())
+	} else {
+		Err(babe_err!("Bad signature on {:?}", pre_hash))
+	}
+}
+
+/// Check a secondary slot proposal header. We validate that the given header is
+/// properly signed by the expected authority, which we have a deterministic way
+/// of computing. Additionally, the weight of this block must stay the same
+/// compared to it's parent since it is a secondary block.
```
joepetrowski:
```suggestion
/// compared to its parent since it is a secondary block.
```

core/consensus/babe/src/lib.rs
```diff
@@ -758,34 +972,155 @@ fn calculate_threshold(
...
+	slot_number: u64,
+	authorities: &[(AuthorityId, BabeAuthorityWeight)],
+	randomness: [u8; 32],
+) -> Option<&AuthorityId> {
+	if authorities.is_empty() {
+		return None;
+	}
+
+	let rand = {
+		let mut data = [0u8; 40];
```
gavofyork: this block can be rewritten `U256::from((randomness, slot_number).using_encoded(blake2_256))`

core/consensus/babe/src/lib.rs
```diff
@@ -758,34 +972,155 @@ fn calculate_threshold(
...
+fn secondary_slot_author(
+	slot_number: u64,
+	authorities: &[(AuthorityId, BabeAuthorityWeight)],
+	randomness: [u8; 32],
+) -> Option<&AuthorityId> {
+	if authorities.is_empty() {
+		return None;
+	}
+
+	let rand = {
```
gavofyork:
```suggestion
	let rand = U256::from((randomness, slot_number).using_encoded(blake2_256));
```

gavofyork:
@andresilva just needs resolving then good to merge.

PR #3381, opened 2019-08-13
Introduce srml/scored-pool

cmichi:
Introduces a scored membership module, ajar to `srml/membership`.


srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+				&new_members[..]
+			);
+		}
+	}
+
+	/// Sorts the `Pool` by score in an ascending order. Entities which
+	/// have a score of `None` are sorted to the beginning of the vec.
+	fn sort_pool(pool: &mut Vec<(T::AccountId, Option<T::Score>)>) {
+		pool.sort_by_key(|(_who, maybe_score)|
+			Option::unwrap_or_default(maybe_score.clone())
```
cmichi: I'm not happy with the `.clone()` here, but haven't found a way around it yet. So happy to hear better suggestions.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+		// remove from pool
+		let mut pool = <Pool<T, I>>::get();
+		let location = Self::find_in_pool(&remove)?;
+		pool.remove(location);
+		<Pool<T, I>>::put(&pool);
+
+		// remove from set, if it was in there
+		let members = <Members<T, I>>::get();
+		let maybe_location = members.binary_search(&remove);
+		if let Ok(_location) = maybe_location {
```
bkchr:
```suggestion
		if members.binary_search(&remove).is_ok() {
```

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+	fn remove_member(remove: T::AccountId) -> Result<(), &'static str> {
+		// remove from pool
+		let mut pool = <Pool<T, I>>::get();
+		let location = Self::find_in_pool(&remove)?;
+		pool.remove(location);
+		<Pool<T, I>>::put(&pool);
+
+		// remove from set, if it was in there
+		let members = <Members<T, I>>::get();
+		let maybe_location = members.binary_search(&remove);
```
bkchr:
```suggestion
```

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+	pub struct Module<T: Trait<I>, I: Instance=DefaultInstance>
+		for enum Call
+		where origin: T::Origin
+	{
+		fn deposit_event<T, I>() = default;
+
+		/// Every `Period` blocks the `Members` set is refreshed from the
+		/// highest scoring members in the pool.
+		fn on_initialize(n: T::BlockNumber) {
+			if n % T::Period::get() == T::BlockNumber::zero() {
```
bkchr:
```suggestion
			if n % T::Period::get() == Zero::zero() {
```

core/sr-std/src/lib.rs
```diff
@@ -68,7 +68,7 @@ include!("../without_std.rs");
 pub mod prelude {
 	pub use crate::vec::Vec;
 	pub use crate::boxed::Box;
-	pub use crate::cmp::{Eq, PartialEq};
+	pub use crate::cmp::{Eq, PartialEq, Ordering};
```
bkchr: Why do we need this? (rust std also does not have it in prelude.)

cmichi: Good spotting, was a relict from a prior version. I removed it now.

cmichi:
> Please adapt to this: #3379 after it has been merged.

Done.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+			Option::unwrap_or_default(maybe_score.clone())
+		);
+	}
+
+	/// Find an entity in the pool.
+	///
+	/// Returns its position in the `Pool` vec, if found.
+	fn find_in_pool(seek: &T::AccountId) -> Result<usize, &'static str> {
+		let pool = <Pool<T, I>>::get();
+		// we can't use binary search here since the pool is sorted by score
```
kianenigma: but what about https://doc.rust-lang.org/std/vec/struct.Vec.html#method.binary_search_by_key ? Should make all of your `foo.iter().find()`s faster.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+		pool
```
kianenigma:
```suggestion
		Self::pool()
```

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+		pool.sort_by_key(|(_who, maybe_score)|
+			Option::unwrap_or_default(maybe_score.clone())
+		);
+	}
+
+	/// Find an entity in the pool.
+	///
+	/// Returns its position in the `Pool` vec, if found.
+	fn find_in_pool(seek: &T::AccountId) -> Result<usize, &'static str> {
+		let pool = <Pool<T, I>>::get();
```
kianenigma:
```suggestion

```

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+			if n % T::Period::get() == Zero::zero() {
+				<Module<T, I>>::refresh_members(true);
+			}
+		}
+
+		/// Add `origin` to the pool of candidates.
+		pub fn issue_candidacy(origin) {
+			let who = ensure_signed(origin)?;
+
+			let _ = Self::find_in_pool(&who)
```
kianenigma: Why not just: 

```suggestion
			let _ = Self::find_in_pool(&who)? 
```

The function, I think, has the same return type as the dispatch.

But I'd actually recommend this because it is more _srml-idiomatic_:

```suggestion
			ensure!(Self::find_in_pool(&who).is_err(), "already a member");
```

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+			Self::remove_member(who)?;
+			Self::deposit_event(RawEvent::CandidateWithdrew);
+		}
+
+		/// Kick a member `who` from the set.
+		///
+		/// May only be called from `KickOrigin` or root.
+		pub fn kick(origin, who: T::AccountId) {
+			T::KickOrigin::try_origin(origin)
+				.map(|_| ())
```
kianenigma: I am not sure but I think these can be `map_or_else`

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+		///
+		/// May only be called from `ScoreOrigin` or root.
+		pub fn score(origin, who: T::AccountId, score: T::Score) {
+			T::ScoreOrigin::try_origin(origin)
+				.map(|_| ())
+				.or_else(ensure_root)
+				.map_err(|_| "bad origin")?;
+
+			let mut pool = <Pool<T, I>>::get();
+			let location = Self::find_in_pool(&who)?;
```
kianenigma: Since this pool guy is sorted: 
- you do: search (`O(LogN)`) remove and naively push (`O(1)`) and sort (`O(NlogN)`) => `O(NlogN)`
- could do: search and remove and **search again for the new spot** and `insert` => `O(2LogN)`

(didn't think it through thoroughly but seems fair to me)

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+
+		let pool = <Pool<T, I>>::get();
+		let mut new_members: Vec<T::AccountId> = pool
+			.into_iter()
+			.filter(|(_, score)| score.is_some())
+			.rev()
+			.take(count as usize)
+			.map(|(account_id, _)| account_id)
+			.collect();
+		new_members.sort();
```
kianenigma: hmmm can't we prove that it already is since it is the`.rev().take(x)` of sth which was already sorted?

srml/scored-pool/src/lib.rs
kianenigma: I think evitable unless `Score: Copy`, which I find a safe bound btw.

srml/scored-pool/src/lib.rs
cmichi: I then run into `use of unstable library feature 'result_map_or_else'`. I had taken this way of checking from the related srml module `membership`, so I'll leave it in place for now.

srml/scored-pool/src/lib.rs
cmichi: Unfortunately not because of this:

`Pool = Vec(T::AccountId, Option<T::Score>>` is sorted by score, not by account id.
`new_members` is only `Vec<T::AccountId>` and needs to be sorted by account_id since `ChangeMembers::change_members_sorted()` requires the input to be sorted.

Also the `binary_search` functions which follow in the next lines (used to derive incoming and outgoing elements) require a sorted vec.

srml/scored-pool/src/lib.rs
cmichi: Hey Kian, great spotting! I adapted the code.

srml/scored-pool/src/lib.rs
cmichi: Unfortunately the `.clone()` is still needed, when I add `Score: Copy` as a trait bound.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+	{
+		/// The given member was removed; see the transaction for who.
+		MemberRemoved,
+		/// An entity has issued a candidacy.
+		CandidateAdded,
+		/// An entity withdrew candidacy.
+		CandidateWithdrew,
+		/// The candidacy was forcefully removed for an entity.
+		CandidateKicked,
+		/// Phantom member, never used.
```
gavofyork: Probably want `CandidateScored` in here , too.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+		/// Add `origin` to the pool of candidates.
+		pub fn issue_candidacy(origin) {
+			let who = ensure_signed(origin)?;
+			ensure!(Self::find_in_pool(&who).is_err(), "already a member");
+
+			T::Currency::reserve(&who, T::CandidateDeposit::get())
+				.map_err(|_| "balance too low")?;
+
+			let mut pool = <Pool<T, I>>::get();
+			pool.push((who.clone(), None));
```
gavofyork: this is incredibly inefficient... your data-structures lecturer would be ashamed

you already did a search for this candidate above with `find_in_pool`! a `binary_search` would provide the index where it should be inserted. and then you can just use `insert` and your algorithm went from O(N.logN) to O(N).

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+			let who = ensure_signed(origin)?;
+
+			Self::remove_member(who)?;
+			Self::deposit_event(RawEvent::CandidateWithdrew);
+		}
+
+		/// Kick a member `who` from the set.
+		///
+		/// May only be called from `KickOrigin` or root.
+		pub fn kick(origin, who: T::AccountId) {
```
gavofyork: should use `T::Lookup::Source` to pass address, not `T::AccountId`. see balances module for an example.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+				.map_err(|_| "bad origin")?;
+
+			Self::remove_member(who)?;
+			Self::deposit_event(RawEvent::CandidateKicked);
+		}
+
+		/// Score a member `who` with `score`.
+		///
+		/// May only be called from `ScoreOrigin` or root.
+		pub fn score(origin, who: T::AccountId, score: T::Score) {
```
gavofyork: same

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+				.map_err(|_| "bad origin")?;
+
+			let mut pool = <Pool<T, I>>::get();
+			let location = Self::find_in_pool(&who)?;
+			pool.remove(location);
+
+			// we binary search the pool (which is sorted by score).
+			// if there is already an element with `score`, we insert
+			// right before that. if not, the search returns a location
+			// where we can insert while maintaining order.
```
gavofyork: 👍

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+	/// If `notify` is set to true `T::MembershipChanged::change_members_sorted`
+	/// will be invoked at the end of the method.
+	fn refresh_members(notify: bool) {
+		let count = <MemberCount<I>>::get();
+
+		let pool = <Pool<T, I>>::get();
+		let mut new_members: Vec<T::AccountId> = pool
+			.into_iter()
+			.filter(|(_, score)| score.is_some())
+			.rev()
```
gavofyork: consider storing in reverse sorted order to avoid this.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+			.into_iter()
+			.filter(|old| new_members.binary_search(&old).is_err())
+			.collect();
+		let incoming: Vec<T::AccountId> = new_members.clone()
+			.into_iter()
+			.filter(|new| old_members.binary_search(&new).is_err())
+			.collect();
+
+		if notify {
+			T::MembershipChanged::change_members_sorted(
```
gavofyork: if you can't compute incoming & outgoing any momre efficiently than this, just use `T::MembershipChanged:: set_members_sorted` and it will do the work for you.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+				&incoming[..],
+				&outgoing[..],
+				&new_members[..]
+			);
+		}
+	}
+
+	/// Sorts the `Pool` by score in an ascending order. Entities which
+	/// have a score of `None` are sorted to the beginning of the vec.
+	fn sort_pool(pool: &mut Vec<(T::AccountId, Option<T::Score>)>) {
```
gavofyork: this should never be needed.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+		pool.sort_by_key(|(_, maybe_score)| maybe_score.unwrap_or_default());
```
gavofyork: this function should never be needed.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	/// Find an entity in the pool.
+	///
+	/// Returns its position in the `Pool` vec, if found.
+	fn find_in_pool(seek: &T::AccountId) -> Result<usize, &'static str> {
```
gavofyork: rename to `ensure_in_pool` to make its meaning clear

srml/scored-pool/src/lib.rs
cmichi: I've now added a commit which changes the sort order of `Pool` so that `None` scores are always at the end (and `Some(_)` scores are sorted descending).

This made it possible to get rid of (among other things) the sorting stuff here and just always append unscored entities to the end.

srml/scored-pool/src/lib.rs
cmichi: It's only needed for building the storage from the config during pre-genesis. There would be problems if somebody passes in an unsorted vec. But yeah it's unnecessary to have an extra function for it and I moved the function content to the `build(…)`.

srml/scored-pool/src/lib.rs
cmichi: There is a `ensure!(Self::find_in_pool(&who).is_err(), "already a member");` which takes use of this function, but besides that there is another use of the function which actually needs the `usize` location in the vec. So I think the name still fits and would leave it as is.

cmichi:
All comments have been addressed and the PR updated.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+		/// Every `Period` blocks the `Members` set is refreshed from the
+		/// highest scoring members in the pool.
+		fn on_initialize(n: T::BlockNumber) {
+			if n % T::Period::get() == Zero::zero() {
+				<Module<T, I>>::refresh_members(true);
+			}
+		}
+
+		/// Add `origin` to the pool of candidates.
+		pub fn issue_candidacy(origin) {
```
gavofyork:
```suggestion
		pub fn submit_candidacy(origin) {
```

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+			let who = ensure_signed(origin)?;
+			ensure!(Self::find_in_pool(&who).is_err(), "already a member");
+
+			T::Currency::reserve(&who, T::CandidateDeposit::get())
+				.map_err(|_| "balance too low")?;
+
+			let mut pool = <Pool<T, I>>::get();
```
gavofyork:
```suggestion
			<Pool<T, I>>::mutate(|pool| pool.push((who.clone(), None)));
```

and delete next two lines.

srml/scored-pool/src/lib.rs
gavofyork: actually, `Vec::remove` is O(N), so it's O(2LogN + N) =~ O(N). still better than O(N.logN) though, but timsort is fiendishly quick, so there's probably not much in it.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+}
+
+impl<T: Trait<I>, I: Instance> Module<T, I> {
+
+	/// Fetches the `MemberCount` highest scoring members from
+	/// `Pool` and puts them into `Members`.
+	///
+	/// If `notify` is set to true `T::MembershipChanged::change_members_sorted`
+	/// will be invoked at the end of the method.
+	fn refresh_members(notify: bool) {
```
gavofyork: why would notify ever be `false`?

srml/scored-pool/src/lib.rs
gavofyork: in that case avoid returning a string error.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+			);
+		}
+	}
+
+	/// Find an entity in the pool.
+	///
+	/// Returns its position in the `Pool` vec, if found.
+	fn find_in_pool(seek: &T::AccountId) -> Result<usize, &'static str> {
+		// we can't use binary search here since the pool is sorted by score
+		Self::pool()
```
gavofyork: you shouldn't be doing linear searches in the runtime, and certainly not from user functions into sets that can grow indefinitely large with little cost.

instead, require the transactor to provide the index themselves and bail out early if it cannot be found.

srml/scored-pool/src/lib.rs
gavofyork: actually, just remove this function altogether. it's way too dangerous.

srml/scored-pool/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "srml-scored-pool"
+version = "1.0.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false, features = ["derive"] }
+parity-codec = { version = "4.1.1", default-features = false }
```
bkchr:
```suggestion
```

srml/scored-pool/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+system = { package = "srml-system", path = "../system", default-features = false }
+
+[dev-dependencies]
+balances = { package = "srml-balances", path = "../balances" }
+primitives = { package = "substrate-primitives",  path = "../../core/primitives" }
+
+[features]
+default = ["std"]
+std = [
+	"parity-codec/std",
```
bkchr:
```suggestion
	"codec/std",
```

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+			sr_io::with_storage(storage, || {
+				let mut pool = config.pool.clone();
+
+				// reserve balance for each candidate in the pool.
+				// panicking here is ok, since this just happens one time, pre-genesis.
+				pool
+					.iter()
+					.for_each(|(who, _)| {
+						T::Currency::reserve(&who, T::CandidateDeposit::get())
+							.expect("balance too low");
```
bkchr: Could we please get a better error message?

"Balance too low to submit candidacy" or something else.

srml/scored-pool/src/lib.rs
bkchr: You can just use `<Pool<T, I>>::append(&[(who.clone(), None)]);`

srml/scored-pool/src/tests.rs
```diff
@@ -0,0 +1,226 @@
...
+		let score = 99;
+		assert_ok!(ScoredPool::issue_candidacy(Origin::signed(who)));
+
+		// when
+		assert_ok!(ScoredPool::score(Origin::signed(ScoreOrigin::get()), who, score));
+
+		// then
+		assert_eq!(fetch_from_pool(who), Some((who, Some(score))));
+		assert_eq!(ScoredPool::find_in_pool(&who),
+				   Ok(0)); // must be first element, since highest scored
```
bkchr: Weird formatting.

srml/scored-pool/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+				&new_members[..],
+				&old_members[..],
+			);
+		}
+	}
+
+	/// Find an entity in the pool.
+	///
+	/// Returns its position in the `Pool` vec, if found.
+	fn find_in_pool(seek: &T::AccountId) -> Result<usize, &'static str> {
```
bkchr: You always request the full pool before or after calling this function. You should be able to refactor this to only always have one access to the pool in storage.

srml/scored-pool/src/lib.rs
cmichi: This was used to differentiate between calling `T::MembershipChanged` vs. `T::MembershipInitialized`. I've refactored the code now to make this clearer.

cmichi:
All comments have been addressed and the PR updated.

PR #3384, opened 2019-08-13
Service factory refactor, part 1

tomaka:
As described in #3382, this is the first part that doesn't break anything and that we should merge before #3382.

### DIFF
--- a/core/cli/src/informant.rs
+++ b/core/cli/src/informant.rs
@@ -21,22 +21,19 @@ use futures::{Future, Stream};
 use futures03::{StreamExt as _, TryStreamExt as _};
 use log::{info, warn};
 use sr_primitives::{generic::BlockId, traits::Header};
-use service::{Service, Components};
+use service::AbstractService;
 use tokio::runtime::TaskExecutor;
 
 mod display;
 
 /// Spawn informant on the event loop
 #[deprecated(note = "Please use informant::build instead, and then create the task manually")]
-pub fn start<C>(service: &Service<C>, exit: ::exit_future::Exit, handle: TaskExecutor) where
-	C: Components,
-{
+pub fn start(service: &impl AbstractService, exit: ::exit_future::Exit, handle: TaskExecutor) {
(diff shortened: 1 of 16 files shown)


arkpar:
Any good reason to turn `export_blocks` and other chain op functions into a macro?

tomaka:
> Any good reason to turn export_blocks and other chain op functions into a macro?

I did that in order to share the same code between the old and new APIs, and to make solving conflicts while rebasing easier.

(to be honest I would have gone a different route if I had had the guarantee to not encounter any conflict)

bkchr:
By conflict you mean merge conflict? Or something in polkadot?

tomaka:
> By conflict you mean merge conflict? Or something in polkadot?

Yes, merge conflicts.

tomaka:
There's a lot of dirt remaining after the refactor (#3382), and where to stop adding commits and when to submit the PR is a bit arbitrary.

For example there are lots of items to move around, we can remove the `custom` field of the config, we can move passing the `config` to `build` so that we don't need to pass it as parameter to the closures, and so on. But I thought that this would be for later.

bkchr:
I still don't understand why we not merge everything at once? If you are afraid of porting the stuff to polkadot, why not start porting directly while we review the new interfaces? And when we merge the other one, you can directly open a pr in polkadot?

tomaka:
> If you are afraid of porting the stuff to polkadot, why not start porting directly while we review the new interfaces?

I actually tried doing that, but Polkadot never seems to be up to date with Substrate master, and I don't really want to fix the breaking changes in the SRML stuff myself.

tomaka:
I looked a bit more at Polkadot's code, and I think it should be fine be port.

What I was worried about are the fact that I removed `grandpa::BlockImportForService` from Substrate, and that there's this `PolkadotService` trait.
However I understand more now, and they shouldn't pose any issue.

Feel free to merge only the entire refactor if you prefer.

bkchr:
Ty.

PR #3385, opened 2019-08-13
srml/authority-discovery: Introduce srml module to sign and verify

mxinden:
In order to have authorities (validators) discover each other, they need
to publish their public addresses (ip addresses) on the Kademlia
Dht indexed by their public key. This payload needs to be signed by a
key identifying them as a valid authority.

Code inside `/core` does not know the current set of authorities nor
can it assume what kind of cryptography primitives are currently in use.
Instead it can retrieve its public key and the current set of
authorities from the runtime and have it sign and verify Dht payloads.

This commit enables code in `/core` to do so by introducing a srml
module and runtime api to:

1. Retrieve own public key.

2. Retrieve public keys of current authority set.

3. Sign a Dht payload.

4. Verify a Dht payload.

This commit makes the logic from the previous commit
(`core/consensus/common/primitives.ConsensusApi`)
444bf719ebdaa733415af288332556542e7b3ac6 obsolete and thus removes it.

### Status

Open questions:

- Should this logic live in its own srml module?

- Should this logic piggyback on the *im-online* session crypto?

This pull request is part of https://github.com/paritytech/substrate/issues/1693. The full implementation of the feature can be found here: https://github.com/mxinden/substrate/tree/authority-discovery

---

@rphmeier does this look similar to the way you imagined the API? @bkchr would you prefer this to just live in each node implementation separately?


burdges:
We should be careful about DHT level attacks on validators, like nodes censoring validators by either choosing their DHT position to flooding DHT buckets.  @FatemeShirazi

cmichi:
`im-online` configures the module with some initial keys on startup: https://github.com/paritytech/substrate/blob/master/node/cli/src/chain_spec.rs#L171-L173. Does it make sense to pre-configure the same ones for this module? Otherwise there are no initial keys for the first session.

Besides that the code lgtm.

core/authority-discovery/primitives/Cargo.toml
```diff
@@ -1,15 +1,15 @@
 [package]
-name = "substrate-consensus-common-primitives"
+name = "substrate-authority-discovery-primitives"
```
rphmeier: this file has been moved, not copied to a new dir

core/authority-discovery/primitives/src/lib.rs
```diff
@@ -14,18 +14,19 @@
...
+use codec::Codec;
 use rstd::vec::Vec;
 
 decl_runtime_apis! {
-	/// Common consensus runtime api.
-	pub trait ConsensusApi<AuthorityId: Codec> {
-		/// Returns the set of authorities of the currently active consensus mechanism.
-		fn authorities() -> Vec<AuthorityId>;
-	}
+      pub trait AuthorityDiscoveryApi<AuthorityId: Codec> {
```
rphmeier: is it not redundant to call it `Api`?

rphmeier: and where are the docs?

core/authority-discovery/primitives/src/lib.rs
```diff
@@ -14,18 +14,19 @@
...
+            fn public_key() -> Option<AuthorityId>;
+            fn authorities() -> Vec<AuthorityId>;
+            fn sign(payload: Vec<u8>) -> Option<Vec<u8>>;
```
rphmeier: Note that the calls to `public_key` and `sign` can race against a local session key changing. I'd recommend making it a single atomic operation because of this. We probably need the runtime to construct the payload because of that.

srml/authority-discovery/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use app_crypto::RuntimeAppPublic;
+use codec::{Decode, Encode};
+use rstd::prelude::*;
+use srml_support::{decl_module, decl_storage, StorageValue};
+
+pub trait Trait: system::Trait + session::Trait {}
+
+decl_storage! {
+    trait Store for Module<T: Trait> as ImOnline {
```
rphmeier: typo?

rphmeier:
We have a documentation rule in the codebase that all public items must be documented.

core/authority-discovery/primitives/Cargo.toml
mxinden: @rphmeier I don't understand this comment. Can you add some more details? 

This pull request introduces `substrate-authority-discovery-primitives` and removes `substrate-consensus-common-primitives`. Instead of showing the two operations separately, Git interprets them as a *move*.

core/authority-discovery/primitives/src/lib.rs
mxinden: There are two possible races, depending on whether we first retrieve the public key or whether we first sign the payload.

### First signing then retrieving public key

In that case the *key*, *signature* validation by other nodes would fail, but given that we put our external addresses on the Dht periodically this should not lead to any failures.

### First retrieving public key then signing

Given that the public key is changed, I am assuming that the public key is no longer needed, thereby other nodes wouldn't query the Dht for the public key anymore. While this results in an invalid dht entry, that entry would never be read by anyone.

---

I am guessing that making this an atomic operation does not only involve being atomic within the runtime api but also all the way down to the key store. 

How often do we estimate a session key to change? With the above failure scenarios in mind, does it need to block this feature?

core/authority-discovery/primitives/src/lib.rs
mxinden: > is it not redundant to call it Api?

I thought this was a convention for a runtime Api e.g. like `OffChainApi`, `GrandpaApi` and `AuraApi`. Also while `AuthorityDiscovery` is the entire process of making our self discoverable and discovering others, the `Api` is only one part enabling logic in `/core` to interoperate with logic in `/srml`.

mxinden:
> im-online configures the module with some initial keys on startup: https://github.com/paritytech/substrate/blob/master/node/cli/src/chain_spec.rs#L171-L173. Does it make sense to pre-configure the same ones for this module? Otherwise there are no initial keys for the first session.

Thanks for the hint @cmichi, can you review https://github.com/paritytech/substrate/pull/3385/commits/05a69efce963fc878117e55c878736d96f510add (part of this pull request now).

core/authority-discovery/primitives/Cargo.toml
rphmeier: yeah. why did you remove that?

mxinden: Discussion happened offline:

`core/consensus/common/primitives` was introduces in https://github.com/paritytech/substrate/pull/3207. It only contained the `ConsensusApi`. `ConsensusApi` is obsolete with this pull request and is called nowhere. Thereby this pull request removes `core/consensus/common/primitives` entirely.

core/authority-discovery/primitives/src/lib.rs
mxinden: As discussed offline, note added in https://github.com/paritytech/substrate/pull/3385/commits/e8de9c65cdbb491431847bff345ae772f182cc82.

mxinden:
https://github.com/paritytech/substrate/pull/3385/commits/9bb3f8f0a90e915d86a7810b1770d6c95b22826f adds basic unit tests to the module.

srml/authority-discovery/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        intersect.pop()
+    }
+
+    pub fn authorities() -> Vec<im_online::AuthorityId> {
```
rphmeier: docs lines on these public functions should be present as well - it's much nicer for the generated rustdoc

mxinden: Ah, sorry for missing that. Thanks @rphmeier. Fixed in https://github.com/paritytech/substrate/pull/3385/commits/fc709189b609a292de975edae7153db24cd1341f

mxinden:
Ping @cmichi @rphmeier. I addressed all the comments. Can you give this another look?

mxinden:
678ff150b248691ddc83af9c309159c5d5f95026 adjusts this pull request to the changes in https://github.com/paritytech/substrate/pull/3413. @andresilva given that you authored https://github.com/paritytech/substrate/pull/3413 can you give this change set a look?

core/authority-discovery/primitives/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+	/// This api is used by the `core/authority-discovery` module to retrieve our
+	/// own authority identifier, to retrieve identifiers of the current authority
+	/// set, as well as sign and verify Kademlia Dht external address payloads
+	/// from and to other authorities.
+	///
+	/// # Note
+	///
+	/// The underlying cryptography key can change at any moment in time. This can
+	/// for example be problematic when a key change happens in between calling
+	/// `public_key` and `sign`, resulting in a signature that is not signed by
```
bkchr: This is not correct. If you use the same blockid for both calls, nothing will change.

node/runtime/src/lib.rs
```diff
@@ -407,6 +407,9 @@ impl offences::Trait for Runtime {
 	type OnOffenceHandler = Staking;
 }
 
+impl authority_discovery::Trait for Runtime {
+}
```
bkchr: Move this up.

srml/authority-discovery/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+rstd = { package = "sr-std", path = "../../core/sr-std", default-features = false }
+serde = { version = "1.0", optional = true }
+session = { package = "srml-session", path = "../session", default-features = false }
+im-online = { package = "srml-im-online", path = "../im-online", default-features = false }
+srml-support = { path = "../support", default-features = false }
+sr-io = { package = "sr-io", path = "../../core/sr-io", default-features = false }
+system = { package = "srml-system", path = "../system", default-features = false }
+
+[features]
+default = ["std", "session/historical"]
```
bkchr: Don't you want to have the historical feature in wasm as well?

core/authority-discovery/primitives/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+	/// the key corresponding to the public key retrieved in the first call. To
+	/// prevent races like the one above, one can verify that the retrieved
+	/// signature corresponds to the public key by calling `verify` as a third
```
DemiMarie-parity: This sounds like a better API is needed.

srml/authority-discovery/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+	/// softened in the future in case a consumer needs to learn own authority
+	/// identifier in any case.
+	pub fn public_key() -> Option<im_online::AuthorityId> {
+		let authorities = Keys::get();
+
+		let local_keys = im_online::AuthorityId::all();
+
+		let mut intersect: Vec<im_online::AuthorityId> = authorities
+			.into_iter()
+			.filter_map(|authority| {
```
bkchr: Change to find_map, drop collect and the last line.

srml/authority-discovery/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	/// Verify the given signature for the given payload with the given
+	/// authority identifier.
+	pub fn verify(
+		payload: Vec<u8>,
+		signature: Vec<u8>,
+		public_key: im_online::AuthorityId,
+	) -> bool {
+		let sig: Result<im_online::AuthoritySignature, _> = Decode::decode(&mut &signature[..]);
```
bkchr: Use AuthoritySignature::decode, then you can drop the Result<... stuff.

srml/authority-discovery/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+
+		match sig {
```
bkchr: sig.map(|s| public.verify()).unwrap_or(false)

node/runtime/src/lib.rs
mxinden: I am guessing you want the `xxx::Trait` implementations to be alphabetical @bkchr?

core/authority-discovery/primitives/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+		fn public_key() -> Option<AuthorityId>;
+
+		/// Retrieve authority identifiers of the current authority set.
+		fn authorities() -> Vec<AuthorityId>;
+
+		/// Sign the given payload with one of our authority keys. This key will
+		/// correspond to the public key returned by `public_key` unless the
+		/// underlying key is changed in between calls. If own authority key is not
+		/// part of the current set of authorities, this function returns None.
+		fn sign(payload: Vec<u8>) -> Option<Vec<u8>>;
```
DemiMarie-parity: This should take an `AuthorityId` parameter.

srml/authority-discovery/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+	}
+}
+
+impl<T: Trait> Module<T> {
+	/// Returns own authority identifier iff it is part of the current authority
+	/// set, otherwise this function returns None. The restriction might be
+	/// softened in the future in case a consumer needs to learn own authority
+	/// identifier in any case.
+	pub fn public_key() -> Option<im_online::AuthorityId> {
```
DemiMarie-parity: This is not scalable (`O(n)`) if we have a large authority set.

node/runtime/src/lib.rs
bkchr: No just the `}`. I was reviewing on mobile, so I could not provide a suggestion.

srml/authority-discovery/src/lib.rs
mxinden: @DemiMarie-parity can you add some details to your comment? Are you referring to generating the intersection of two authority id sets?

I have written a small benchmark with the assumption of 1000 current authorities (*n*) and 10 local authority keys (*m*). In the worst case we have to iterate the entire set *n* and compare it to each key in *m* resulting in an `O(n*m)`. Running this worst case scenario on a fairly new laptop compiled to machine code (not web assembly) takes ~13 us.

Is that within a range that we should worry about?

Source code can be found here: https://github.com/mxinden/benchmark-authority-id-intersection/blob/master/benches/intersection.rs#L13

DemiMarie-parity: @mxinden Yes, I am.  My worry is that this could wind up being called by wasm code in a tight loop, which would be slow.

bkchr: The code is not running on-chain. So, even if we iterate 1mio times, it should not be any bottleneck.

core/authority-discovery/primitives/src/lib.rs
mxinden: Good point. Also with having `sign` take an `authority_id` (see https://github.com/paritytech/substrate/pull/3385#discussion_r315278623) this race is removed.

srml/authority-discovery/Cargo.toml
mxinden: Actually this was a left over from some testing. None of `authority-discovery` depends on `session/historical` thus I removed the reference.

srml/authority-discovery/src/lib.rs
mxinden: Thanks!

core/authority-discovery/primitives/src/lib.rs
mxinden: By having `sign` take the `authority id` this race is prevented, thus I removed the comment.

core/authority-discovery/primitives/src/lib.rs
mxinden: Good idea. Done.

mxinden:
@bkchr @DemiMarie-parity thank you for the feedback. https://github.com/paritytech/substrate/pull/3385/commits/d0979e954b97cb942a4ca7d34c401f3afc9edf52 addresses them. Could you take another look?

core/authority-discovery/primitives/src/lib.rs
bkchr: Hmm. Where is the authority_id stored between the calls?

mxinden: My process looks like this:

1. Retrieve authority id.

2. Prepare payload (possibly including the authority id) and serialize it.

3. Sign the payload with the private key corresponding to the authority id retrieved in *1*.

4. Put the *(key, value)* pair *(authority id, payload)* on the Dht.

> Hmm. Where is the authority_id stored between the calls?

It is not stored long term within the *authority-discovery* process.

bkchr:
Needs resolving and then it is good to be merged.

srml/authority-discovery/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+
+//! # Authority discovery module.
+//!
+//! This module is used by the `core/authority-discovery` to retrieve the
+//! current set of authorities, learn its own authority id as well as sign and
+//! verify messages to and from other authorities.
+//!
+//! ## Dependencies
+//!
+//! This module depends on the [Im online module](../srml_im_online/index.html)
```
DemiMarie-parity:
```suggestion
//! This module depends on the [I’m online module](../srml_im_online/index.html)
```

or

```suggestion
//! This module depends on the [I'm online module](../srml_im_online/index.html)
```

srml/authority-discovery/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+	}
+}
+
+impl<T: Trait> Module<T> {
+	/// Returns own authority identifier iff it is part of the current authority
+	/// set, otherwise this function returns None. The restriction might be
+	/// softened in the future in case a consumer needs to learn own authority
+	/// identifier in any case.
```
DemiMarie-parity:
```suggestion
	/// identifier.
```

core/authority-discovery/primitives/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+	///
+	/// This api is used by the `core/authority-discovery` module to retrieve our
+	/// own authority identifier, to retrieve identifiers of the current authority
+	/// set, as well as sign and verify Kademlia Dht external address payloads
+	/// from and to other authorities.
+	pub trait AuthorityDiscoveryApi<AuthorityId: Codec> {
+		/// Returns own authority identifier iff it is part of the current authority
+		/// set, otherwise this function returns None. The restriction might be
+		/// softened in the future in case a consumer needs to learn own authority
+		/// identifier in any case.
```
DemiMarie-parity:
```suggestion
		/// identifier.
```

mxinden:
Thanks for the help everyone.

DemiMarie-parity:
@mxinden You’re welcome!

andresilva:
Too late to review this, but LGTM.

PR #3388, opened 2019-08-13
Replace config field with default value (srml/im-online)

cmichi:
The behavior we want is for gossiping to occur automatically right at startup.

It's not necessary to make this configurable. Also it's possibly dangerous to make this configurable, since if misconfigured a heartbeat would only be sent out at the configured block ‒ possibly resulting in slashing if this block is greater than the session length.

### DIFF
--- a/srml/im-online/src/lib.rs
+++ b/srml/im-online/src/lib.rs
@@ -70,7 +70,7 @@
 use primitives::offchain::{OpaqueNetworkState, StorageKind};
 use codec::{Encode, Decode};
 use sr_primitives::{
-	ApplyError, traits::{Extrinsic as ExtrinsicT, Zero},
+	ApplyError, traits::Extrinsic as ExtrinsicT,
 	transaction_validity::{TransactionValidity, TransactionLongevity, ValidTransaction},
 };
 use rstd::prelude::*;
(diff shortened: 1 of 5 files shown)


srml/im-online/src/lib.rs
```diff
@@ -70,7 +70,7 @@
 use primitives::offchain::{OpaqueNetworkState, StorageKind};
 use codec::{Encode, Decode};
 use sr_primitives::{
-	ApplyError, traits::{Extrinsic as ExtrinsicT},
+	ApplyError, traits::{Extrinsic as ExtrinsicT, Zero},
```
bkchr:
```suggestion
	ApplyError, traits::Extrinsic as ExtrinsicT,
```

srml/im-online/src/lib.rs
```diff
@@ -174,7 +174,7 @@ decl_event!(
 decl_storage! {
 	trait Store for Module<T: Trait> as ImOnline {
 		/// The block number when we should gossip.
-		GossipAt get(gossip_at) config(): T::BlockNumber;
+		GossipAt get(gossip_at) build(|_| Zero::zero()): T::BlockNumber;
```
bkchr:
```suggestion
		GossipAt get(gossip_at) : T::BlockNumber;
```

Zero is the default value for numeric types.

PR #3395, opened 2019-08-14
Blockchain cache pruning strategy

svyatonik:
Extracted part of #3201 . This PR:
1) moves `well_known_cache_keys` from consensus to client;
2) introduces cache pruning strategy. There are two now - `ByDepth`, which is default && it is strategy which has been used before. The other one is `NeverPrune`, which will be used (after #3201) for changes tries configurations cache


core/client/src/blockchain.rs
```diff
@@ -258,3 +257,18 @@ pub fn tree_route<Block: BlockT, Backend: HeaderBackend<Block>>(
 		pivot,
 	})
 }
+
+/// A list of all well known keys in the blockchain cache.
+pub mod well_known_cache_keys {
```
bkchr: I think the `client` is the wrong place. We should not try to put more and more stuff into client. I think `primitives` is the correct place for these constants + types.

svyatonik: Could do, but the cache itself is declared in client. And not in primitives. So I thought it is better to keep them together. What do you think?

bkchr: Hmm yeah, that is maybe something that can be done when the cache is extracted from the client crate.

core/client/db/src/cache/mod.rs
```diff
@@ -335,3 +334,17 @@ impl<Block: BlockT> BlockchainCache<Block> for DbCacheSync<Block> {
...
 
+/// Get pruning strategy for given cache.
+fn cache_pruning_strategy<N: From<u32>>(cache: CacheKeyId) -> PruningStrategy<N> {
+	// the cache is mostly used to store data from consensus engines
+	// this kind of data is only required for non-finalized blocks
+	// => by default we prune finalized cached entries
+
+	match cache {
+		// we need to keep changes tries configurations forever (or at least until changes tries,
+		// that were built using this configuration, are not pruned) to make it possible to refer
```
DemiMarie-parity:
```suggestion
		// that were built using this configuration, are pruned) to make it possible to refer
```

core/client/db/src/cache/mod.rs
```diff
@@ -335,3 +334,17 @@ impl<Block: BlockT> BlockchainCache<Block> for DbCacheSync<Block> {
 	}
 }
 
+/// Get pruning strategy for given cache.
+fn cache_pruning_strategy<N: From<u32>>(cache: CacheKeyId) -> PruningStrategy<N> {
+	// the cache is mostly used to store data from consensus engines
+	// this kind of data is only required for non-finalized blocks
```
rphmeier: i'm not sure if this would hold true forever. for instance, if we put BABE epoch data in the cache we would care about something from well before the most recently finalized block.

rphmeier: however, we don't keep that in the cache so I suppose it's a moot point.

rphmeier:
Can merge when checkmarks are green.

PR #3400, opened 2019-08-14
Make inflation mod public to access inflation functions

drewstone:
This PR makes the inflation module public so that one can access the functions inside.

I am trying to access the inflation functions/rate in order to mint tokens for the Treasury. We would like to do so in different time intervals. My intuition is that if I provide a smaller interval I will still get roughly the same inflation rate as is created over an era.

### DIFF
--- a/srml/staking/src/lib.rs
+++ b/srml/staking/src/lib.rs
@@ -277,7 +277,7 @@ mod mock;
 mod tests;
 
 mod phragmen;
-mod inflation;
+pub mod inflation;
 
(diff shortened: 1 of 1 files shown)


kianenigma:
fail enough -- @thiolliere I recently moved phragmen out of SRML to be re-used by other modules if needed. Do you think it ever applies to this as well?

thiolliere:
yes maybe indeed, but maybe the doc is to be improved then

PR #3401, opened 2019-08-14
Support block revert operation in blockchain cache

svyatonik:
Extracted part of #3201 . This PR add support (under `#[cfg(test)]` for now) for block revert operation in cache - it truncates and (optionally) destroys unfinalized cache forks. This is required, because after #3201 blockchain cache will be used to store consensus-critical information on full nodes (changes tries configuration, which are used to build changes tries => block header is affected). Having any inconsistencies in the cache will lead to fatal errors.


core/client/db/src/cache/list_cache.rs
```diff
@@ -512,6 +566,44 @@ impl<Block: BlockT, T: CacheItemT> Fork<Block, T> {
 			best_finalized_block,
 		)
 	}
+
+	/// Truncate fork by deleting all entries that are descendants of given block.
+	#[cfg(test)]
```
tomaka: What's wrong with having this available all the time?

I'd be in favour of having a function that you're not supposed to call rather than having `#[cfg(test)]` everywhere.

```suggestion
```

svyatonik: Initially I was agree with your suggestion, but as I wrote the code, I have had some strange feelings surrounding me. So I would prefer to revert this commit && leave `#[cfg(test)]` instead. But if you still have a strong feeling that last commit is OK, then let it be

core/client/db/src/cache/list_cache.rs
```diff
@@ -317,6 +320,44 @@ impl<Block: BlockT, T: CacheItemT, S: Storage<Block, T>> ListCache<Block, T, S>
...
+		reverted_block: &ComplexBlockId<Block>,
+	) -> ClientResult<CommitOperation<Block, T>> {
+		// can't revert finalized blocks
+		debug_assert!(self.best_finalized_block.number < reverted_block.number);
+
+		// iterate all unfinalized forks and truncate/destroy if required
+		let mut updated = BTreeMap::new();
+		for (index, fork) in self.unfinalized.iter().enumerate() {
+			// we only need to truncate fork if its head is connected to truncated block
+			if !chain::is_connected_to_block(&self.storage, reverted_block, &fork.head.valid_from)? {
```
cheme: This is probably a costy operation, but I think there is already on progress work to cache headers of blocks.

cheme: still then next condition would be better before this one.

core/client/db/src/cache/list_cache.rs
```diff
@@ -317,6 +320,44 @@ impl<Block: BlockT, T: CacheItemT, S: Storage<Block, T>> ListCache<Block, T, S>
...
+				continue;
+			}
+
+			let updated_fork = fork.truncate(
+				&self.storage,
+				tx,
+				reverted_block.number,
+				self.best_finalized_block.number,
+			)?;
+			updated.insert(index, updated_fork);
```
cheme: storing forks index means that the unfinalized map would not change before commit, I believe it is the case (if we combine multiple commitset pb could hapen (wonder if we can have a safer api by ensuring only one commitset is used but that is probably not very needed).

core/client/db/src/cache/list_cache.rs
```diff
@@ -350,6 +391,14 @@ impl<Block: BlockT, T: CacheItemT, S: Storage<Block, T>> ListCache<Block, T, S>
 					self.unfinalized.remove(*fork_index);
 				}
 			},
+			CommitOperation::BlockReverted(forks) => {
+				for (fork_index, updated_fork) in forks.into_iter().rev() {
+					match updated_fork {
+						Some(updated_fork) => self.unfinalized[fork_index] = updated_fork,
+						None => { self.unfinalized.remove(fork_index); },
```
cheme: need ordered content in blockreverted (btreemap ), so it is correct.

core/client/db/src/cache/list_cache.rs
```diff
@@ -512,6 +561,43 @@ impl<Block: BlockT, T: CacheItemT> Fork<Block, T> {
...
+
+	/// Truncate fork by deleting all entries that are descendants of given block.
+	pub fn truncate<S: Storage<Block, T>, Tx: StorageTransaction<Block, T>>(
+		&self,
+		storage: &S,
+		tx: &mut Tx,
+		reverting_block: NumberFor<Block>,
+		best_finalized_block: NumberFor<Block>,
+	) -> ClientResult<Option<Fork<Block, T>>> {
+		let mut current = self.head.valid_from.clone();
```
cheme: I would have see 'self.best_block' here, I understand truncate as a reverse iteration on fork range of blocks deleting in tx all entry until finalization or the intersection of the fork and its parent branch which is the reverting block number).
Maybe I got confuse between 'self.best_block' and 'self.head.valid_from', for me it defines a range of block number for the branch/fork.

svyatonik: The `best_block` of the `Fork` struct is an `Option`al field that is used just for optimization - i.e. when new block is inserted into database, we normally are not reading anything from database at all. Instead we are checking if new block is appended to some known fork by checking if `Some(block.parent_hash()) == fork.best_block`. If that's true => then we have found a fork where we going to append a new/same value => no reads from DB are performed. We only access database when some new fork is created - i.e. new block parent isn't the best block of some known fork.

`head.valid_from` points to the block where value (like CT configuration, or authorities, ...) has been updated last time. I.e. it is a normal situation when value has been updated at 100 and the best block (of this fork) is 150. So this loop is exactly what you have said - a reverse iteration. But the iteration elements aren't blocks, but instead nodes of the list-cache.

core/client/db/src/cache/list_cache.rs
svyatonik: I agree that the API could be better here. But since the cache is internal to the `substrate-client-db` and currently db crate only supports one operation per transaction (like import single block, revert blocks) let's leave this as is - it is faster than than anything else.

core/client/db/src/cache/list_cache.rs
```diff
@@ -80,6 +80,9 @@ pub enum CommitOperation<Block: BlockT, T: CacheItemT> {
 	/// - new entry is finalized AND/OR
 	/// - some forks are destroyed
 	BlockFinalized(ComplexBlockId<Block>, Option<Entry<Block, T>>, BTreeSet<usize>),
+	/// When best block is reverted - contains the forks that have to be updated
```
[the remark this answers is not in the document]
svyatonik: I'm not sure about this change :) Actually, we can only update `best block` of the blockchain (i.e. there's 'best block' in the blockchain and 'best block' of forks in the cache). And we're reverting the best block of the blockchain here.

Also - what's the '(the fork itself and the descendant of its block).' is about? :)

core/client/db/src/cache/list_cache.rs
cheme: Ok, guess I confuse 'valid_from' with 'prev_valid_from' of 'StorageEntry', I will need to read that again :)

svyatonik: The conversion from `StorageEntry` to `Entry` is quite simple. There's a list of all active forks saved in the cache' meta. Every fork is described by the `ComplexBlockId` which points to the `head` of the fork. We then read the value from the head && when converting to `Entry` simply ignore the `prev_valid_from` field of the `StorageEntry`. Instead, value and the block of the `head` is used. wE only use `prev_valid_from` when traversing the list - when truncating or searching for the value that has been actual at the given block, ...

core/client/db/src/cache/list_cache.rs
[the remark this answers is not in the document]
cheme: Yes I was assuming wrongly that we could revert at any given block. so the second part  of the comment was about saying that not only the block is reverted but also unfinalized branch content that was attached to it (same idea as comment from line 340).

cheme:
Things get a bit clearer after a reread, I am rather interested in this list entry cache code because I realize I write some code with similar purpose. Do you know what are the current uses for it? A quick search, return my mostly access to authorities from consensus and light client.
For the pr in itself is seems ok to me, maybe a review from someone more familiar with the code may be needed, but it seems good.

svyatonik:
atm the cache only exists on light clients && used to save data of consensus algorithm (authorities in Aura && era in BABE). I'm planning to use it to store CT configurations on light/full nodes as well.

cheme:
So it is not strictly speaking a cache, seems like I started some work on something rather close to be able to store some child trie keyspace for #3648 (the dev intends to be extensible to other use cases but it seems like this 'cache' could have been a good way to get things done quickier).

svyatonik:
Right, it isn't used as a cache anymore. It used to be in the very beginning, but now it is merely a storage (though it still can be used as a cache).

You can use it, of course. But please keep in mind that it is optimized for values that aren't changing often (or better - not changed at every block). If you need something that have different value at every block, then probably it is better to use simple map + separate db column for that.

gavofyork:
one test still failing..

PR #3403, opened 2019-08-15
Value lifetime is returned from blockchain cache

svyatonik:
Extracted part of #3201 . `Cache::get_at()` will now return value with its lifetime (it is the block where it has been inserted into the cache + optionally block where next value, that supersedes it, has been inserted). This allows traversing changes tries configurations in #3201


gavofyork:
@svyatonik needs resolving...

svyatonik:
The issue with firs commit is some panic in libp2p, which is unrelated. The second issue is about nightly - should be ok when we'll revert to prev/next nightly?

PR #3405, opened 2019-08-15
add some docs for consensus-gossip

rphmeier:
Previously, some terms such as `topic` and the role of the consensus engine ID were not discussed.

### DIFF
--- a/core/network/src/protocol/consensus_gossip.rs
+++ b/core/network/src/protocol/consensus_gossip.rs
@@ -18,24 +18,29 @@
 //! Handles chain-specific and standard BFT messages.
 //!
 //! Gossip messages are separated by two categories: "topics" and consensus engine ID.
+//! The consensus engine ID is sent over the wire with the message, while the topic is not,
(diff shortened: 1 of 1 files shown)


core/network/src/protocol/consensus_gossip.rs
```diff
@@ -14,8 +14,34 @@
...
+//! The consensus engine ID is sent over the wire with the message, while the topic is not,
+//! with the expectation that the topic can be derived implicitly from the content of the
+//! message, assuming it is valid.
+//!
+//! Topics are a single 32-byte tag associated with a message, used to group those messages
+//! in an opaque way. Consensus code can invoke `broadcast_topic` to attempt to send all messages
+//! under a single topic to all peers who don't have them yet, and `send_topic` to
+//! send all messages under a single topic to a specific peer.
+//!
+//! Each consensus engine ID can have an associated,
```
andresilva:
```suggestion
//! Each consensus engine ID must have an associated,
```

PR #3406, opened 2019-08-15
prevent dust stash balance from bonding

rphmeier:
the check was previously only done before comparing the intended bonding value with the stash balance.

### DIFF
--- a/srml/staking/src/lib.rs
+++ b/srml/staking/src/lib.rs
@@ -746,8 +746,8 @@ decl_module! {
 				return Err("controller already paired")
 			}
 
-			let value = value.min(stash_balance);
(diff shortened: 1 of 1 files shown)


srml/staking/src/lib.rs
```diff
@@ -746,6 +746,9 @@ decl_module! {
 				return Err("controller already paired")
 			}
 
+			let value = value.min(stash_balance);
```
shawntabrizi: Why did you flip the order?

`stash_balance` is not defined until the line after now.

rphmeier: yeah - a little too quick :)

gavofyork:
needs a regression test that fails without this change. and i'll bet you a beer that you can't provide one.

PR #3409, opened 2019-08-15
Bad spelling

gavofyork:


srml/support/src/dispatch.rs
```diff
@@ -422,7 +422,7 @@ macro_rules! decl_module {
 		{ $( $constants:tt )* }
 		[ $( $dispatchables:tt )* ]
 		$(#[doc = $doc_attr:tt])*
-		fn on_initialise($($param_name:ident : $param:ty),* ) { $( $impl:tt )* }
+		fn on_initialize($($param_name:ident : $param:ty),* ) { $( $impl:tt )* }
```
bkchr: `on_initialise` was by intention as you see below, we would generate a `compile_error!` to tell the user that is was renamed.

pepyakin: Should we revert this change then?

PR #3411, opened 2019-08-15
Allow slashing only on current era

rphmeier:
Offences from a prior era are discarded, and we track how much of the current era's exposure we've already slashed to avoid over-slashing.

### DIFF
--- a/srml/staking/src/lib.rs
+++ b/srml/staking/src/lib.rs
@@ -447,7 +447,7 @@ pub struct Exposure<AccountId, Balance: HasCompact> {
 pub struct SlashJournalEntry<AccountId, Balance: HasCompact> {
 	who: AccountId,
 	amount: Balance,
-	own_slash: Balance,
+	own_slash: Balance, // the amount of `who`'s own exposure that was slashed
 }
 
 pub type BalanceOf<T> =
--- a/srml/staking/src/lib.rs
+++ b/srml/staking/src/lib.rs
@@ -672,6 +672,9 @@ decl_event!(
 		Reward(Balance),
 		/// One validator (and its nominators) has been slashed by the given amount.
 		Slash(AccountId, Balance),
+		/// An old slashing report from a prior era was discarded because it could
+		/// not be processed.
+		OldSlashingReportDiscarded(SessionIndex),
 	}
 );
 
@@ -1595,8 +1598,13 @@ impl<T, Reporter, Offender, R, O> ReportOffence<Reporter, Offender, O>
 {
 	fn report_offence(reporters: Vec<Reporter>, offence: O) {
 		// disallow any slashing from before the current era.
-		if offence.session_index() >= <Module<T>>::current_era_start_session_index() {
+		let offence_session = offence.session_index();
+		if offence_session >= <Module<T>>::current_era_start_session_index() {
 			R::report_offence(reporters, offence)
+		} else {
+			<Module<T>>::deposit_event(
+				RawEvent::OldSlashingReportDiscarded(offence_session).into()
+			)
 		}
 	}
 }
--- a/srml/staking/src/lib.rs
+++ b/srml/staking/src/lib.rs
@@ -1213,6 +1213,10 @@ impl<T: Trait> Module<T> {
 
 		// Increment current era.
 		let current_era = CurrentEra::mutate(|s| { *s += 1; *s });
+
+		// prune journal for last era.
+		<EraSlashJournal<T>>::remove(current_era - 1);
+
 		CurrentEraStartSessionIndex::mutate(|v| {
 			*v = start_session_index;
 		});
--- a/srml/staking/src/lib.rs
+++ b/srml/staking/src/lib.rs
@@ -280,7 +280,7 @@ use sr_primitives::traits::{
 };
 use sr_staking_primitives::{
 	SessionIndex, CurrentElectedSet,
-	offence::{OnOffenceHandler, OffenceDetails},
+	offence::{OnOffenceHandler, OffenceDetails, Offence, ReportOffence},
 };
 #[cfg(feature = "std")]
 use sr_primitives::{Serialize, Deserialize};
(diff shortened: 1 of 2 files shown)


srml/staking/src/lib.rs
```diff
@@ -441,6 +441,15 @@ pub struct Exposure<AccountId, Balance: HasCompact> {
 	pub others: Vec<IndividualExposure<AccountId, Balance>>,
 }
 
+/// A slashing event occurred, slashing a validator for a given amount of balance.
```
pepyakin: What is `own_slash`?

srml/staking/src/lib.rs
```diff
@@ -1533,12 +1575,32 @@ impl <T: Trait> OnOffenceHandler<T::AccountId, session::historical::Identificati
...
 		}
+		<EraSlashJournal<T>>::insert(era_now, journal);
 
 		// Handle the rest of imbalances
 		T::Slash::on_unbalanced(remaining_imbalance);
 	}
 }
 
+/// Filter historical offences out and only allow those from the current era.
+pub struct FilterHistoricalOffences<T, R> {
```
pepyakin: Why not hook it up in node-runtime right away?

srml/staking/src/lib.rs
```diff
@@ -1030,17 +1043,33 @@ impl<T: Trait> Module<T> {
...
+			.map(|entry| entry.own_slash)
+			.fold(<BalanceOf<T>>::zero(), |a, c| a.saturating_add(c));
+
+		let own_remaining = exposure.own.saturating_sub(already_slashed_own);
+
 		// The amount we'll slash from the validator's stash directly.
-		let own_slash = exposure.own.min(slash);
+		let own_slash = own_remaining.min(slash);
 		let (mut imbalance, missing) = T::Currency::slash(stash, own_slash);
 		let own_slash = own_slash - missing;
```
pepyakin: Dumb question: Under what circumstances `missing` can be greater than zero?

srml/staking/src/lib.rs
rphmeier: the amount of slashing we've done on our own exposure

srml/staking/src/lib.rs
rphmeier: I wanted to, but I didn't see that `ReportOffence` was used anywhere

srml/staking/src/lib.rs
rphmeier: to be honest, in this case it should always be zero; because we should never slash more than 100% all-in-all.

srml/staking/src/lib.rs
pepyakin: yep, do you think it’s worthwhile to put it in the rustdoc

rphmeier: well, it's not a public member, so no

srml/staking/src/lib.rs
```diff
@@ -1030,17 +1043,33 @@ impl<T: Trait> Module<T> {
...
 		let slash = slash.min(exposure.total);
+
+		// limit what we'll slash of the stash's own to only what's in
+		// the exposure.
+		//
+		// note: this is fine only because we limit reports of the current era.
+		// otherwise, these funds may have already been slashed due to something
+		// reported from a prior era.
+		let already_slashed_own = journal.iter()
+			.filter(|entry| &entry.who == stash)
```
tomusdrw: I suppose `stash` does not change, right? You can't move to another `stash` account without unbonding/bonding?

srml/staking/src/lib.rs
gavofyork: I don't see much point in documenting functions or fields *not* in rustdoc.

srml/staking/src/lib.rs
```diff
@@ -1533,12 +1575,32 @@ impl <T: Trait> OnOffenceHandler<T::AccountId, session::historical::Identificati
...
+	for FilterHistoricalOffences<Module<T>, R> where
+	T: Trait,
+	R: ReportOffence<Reporter, Offender, O>,
+	O: Offence<Offender>,
+{
+	fn report_offence(reporters: Vec<Reporter>, offence: O) {
+		// disallow any slashing from before the current era.
+		if offence.session_index() >= <Module<T>>::current_era_start_session_index() {
+			R::report_offence(reporters, offence)
+		}
```
tomusdrw: Shall we at least print something or emit an event to monitor old reports?

srml/staking/src/lib.rs
```diff
@@ -1178,6 +1213,10 @@ impl<T: Trait> Module<T> {
 
 		// Increment current era.
 		let current_era = CurrentEra::mutate(|s| { *s += 1; *s });
+
+		// prune journal for last era.
+		<EraSlashJournal<T>>::remove(current_era - 1);
```
tomusdrw: Couldn't we prune era's with some offset (`X`) and accept reports for last `X` eras?

srml/staking/src/lib.rs
gavofyork: correct

gavofyork:
@rphmeier just waiting on answers to the final queries before i can merge this.

srml/staking/src/lib.rs
rphmeier: (I added a comment about it last night - but it won't go in rustdoc)

srml/staking/src/lib.rs
rphmeier: We could, but it wouldn't be of any use with this runtime. Historical slashing just isn't doable without more work beyond this milestone.

srml/staking/src/lib.rs
rphmeier: done: an event is emitted

srml/staking/src/lib.rs
tomusdrw: It could still be a `///` comment, without going to html docs - I don't really see any issue with that, it just helps with readability, by having consistency in where to look for description of an item.

pepyakin: Ah, FWIW, I thought about mentioning it in the rustdoc of the struct in the same way `who` and `amount` was mentioned.

While I agree with @tomusdrw, I am good with status quo as well since it is improvement and we already spent enough time on this.

PR #3413, opened 2019-08-15
session: add handler for genesis session

andresilva:
The `SessionHandler` is extended with a method for informing about the genesis validator set. This avoids having to configure all of the keys in the chain spec as they will all be initialized from the session module. I still kept the authorities configurable in the modules that interact with session (grandpa, babe, i'm online) because it is useful when using those modules without session.


andresilva:
I have no clue why the build is failing to compile. Seems to be working locally.

gavofyork:
should really be bumping runtime version.

Issue #3254, opened 2019-07-30
Explicit selfdestruct


pepyakin:
_Originally posted by @pepyakin in https://github.com/paritytech/substrate/issues/2705#issuecomment-499888672_

After #3255, we could implement another deferred action - inevitabile deliberate self destruction. It would not have a problem that the current "drain-to-death" mechanism has, i.e. it won't be possible to send funds to a contract to prevent its self destruction.

DemiMarie-parity:
I recommend against this, because of the multi-sig wallet bug.

pepyakin:
I believe this particular bug doesn't mean that we should reject the idea of selfdestruct. Especially considering that the problem is not caused by selfdestruct exclusively but rather because of combination of factors like using a (selfdestructable) contract as a dynamic library.

I think selfdestruct is useful on its own since there are cases when you really done with a contract and there is nothing better to do that just remove it. Think of an agreement-like contract, when the contract performed the conditional logic it was created for.

srml-contracts, unlike Ethereum, is not a homogenious single piece but is a moudle and a part of SRML and it has to integrate into it. One conern that it has to respect is the way that the outer system handles accounting. In SRML if an account's balance falls below certain threshold the account is removed. Because a contract is also an account it is removed when it ends up with its balance below such treshold. We kind of already have a way to selfdestruct, dubbed, "drain-to-death".

> This is not to mention state rent which purpose is to kill accounts which can't pay for their storage.

However, "drain-to-death" has an issue - after the contract sent all of its funds it is possible that another contract sends funds to that contract until the end of transaction effectively blocking the self destruction. Adding an explicit request for termination can fix that.

Btw, srml-contracts separates code from a contract paving the way for the first-class dynamic libraries (This is not to say that it is completely eliminates the risk that somebody decide to delegate some critical logic to a selfdestructable contract, but rather that this feature would make that less probable)

andresilva:
What happens if we selfdestruct a contract with funds? Would this be an effective way for burning funds? Otherwise if we don't want to allow that, we can make the selfdestruct only work when balance is below existential. This should still prevent the current problem of sending funds to contract to defer destruction.

jimpo:
This wouldn't be a problem if contract destruction is not deferred. I propose the following:

- If a contract is called and returns from a call with a balance below the existential deposit, the contract is immediately removed and its balance is set to 0. This is visible to any contracts calling it higher in the execution stack.
[...]

PR #3417, opened 2019-08-16
srml-contracts: Apply contract removals immediately

jimpo:
This implements a solution to #3254, in particular the one described in [this comment](https://github.com/paritytech/substrate/issues/3254#issuecomment-521630165).

- If a contract is called and returns from a call with a balance below the existential deposit, the contract is immediately removed and its balance is set to 0. This is visible to any contracts calling it higher in the execution stack.
  - There is an exception if the contract with insufficient balance is being recursively called from itself. If a contract returns with balance below existential deposit and is also being executed higher up in the execution stack, then the nested call traps immediately. The caller sees the trap and may be able to recover.
- If a contract is instantiated and returns from the constructor with a balance below the existential deposit, then this traps immediately. The caller sees the trap and may be able to recover.
- If a contract self-destructs, then the caller sends it a balance exceeding the existential deposit, then this creates a new account at that address with no contract code.
- If a contract self-destructs, then a caller may still deploy the same contract with the same code hash and input data such that the same address is assigned. This replaces the old contract with the newly instantiated contract. (Exactly what would happen if this occurred in a subsequent extrinsic).

The benefits of this approach is that it ensures a contract cannot be destroyed with a remaining balance and it seems more like expected behavior to me that a contract removal is applied immediately.

### DIFF
--- a/srml/contracts/src/account_db.rs
+++ b/srml/contracts/src/account_db.rs
@@ -34,11 +34,57 @@ use system;
 // the trie_id in the overlay, thus we provide an overlay on the fields
 // specifically.
 pub struct ChangeEntry<T: Trait> {
+	/// If Some(_), then the account balance is modified to the value. If None and `reset` is false,
+	/// the balance unmodified. If None and `reset` is true, the balance is reset to 0.
 	balance: Option<BalanceOf<T>>,
-	/// If None, the code_hash remains untouched.
+	/// If Some(_), then a contract is created with the code hash. If None and `reset` is false,
+	/// then the contract code is unmodified. If None and `reset` is true, the contract is deleted.
 	code_hash: Option<CodeHash<T>>,
+	/// If Some(_), then the rent allowance is set to the value. If None and `reset` is false, then
+	/// the rent allowance is unmodified. If None and `reset` is true, the contract is deleted.
 	rent_allowance: Option<BalanceOf<T>>,
 	storage: BTreeMap<StorageKey, Option<Vec<u8>>>,
+	/// If true, indicates that the existing contract and all its storage entries should be removed
+	/// and replaced with the fields on this change entry. Otherwise, the fields on this change
+	/// entry are updates merged into the existing contract info and storage.
+	reset: bool,
+}
+
+impl<T: Trait> ChangeEntry<T> {
+	fn balance(&self) -> Option<BalanceOf<T>> {
+		self.balance.or_else(|| {
+			if self.reset {
+				Some(<BalanceOf<T>>::zero())
+			} else {
+				None
+			}
+		})
+	}
+
+	fn code_hash(&self) -> Option<Option<CodeHash<T>>> {
+		if self.reset {
+			Some(self.code_hash)
+		} else {
+			self.code_hash.map(Some)
+		}
+	}
+
+	fn rent_allowance(&self) -> Option<Option<BalanceOf<T>>> {
+		if self.reset {
+			Some(self.rent_allowance)
+		} else {
+			self.rent_allowance.map(Some)
+		}
+	}
+
+	fn storage(&self, location: &StorageKey) -> Option<Option<Vec<u8>>> {
+		let value = self.storage.get(location).cloned();
+		if self.reset {
+			Some(value.unwrap_or(None))
+		} else {
+			value
+		}
+	}
 }
 
 // Cannot derive(Default) since it erroneously bounds T by Default.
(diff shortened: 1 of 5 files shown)


srml/contracts/src/exec.rs
```diff
@@ -507,6 +534,12 @@ where
 
 		Ok(output)
 	}
+
+	/// Returns whether a contract, identified by address, is currently being executed.
```
pepyakin: I believe more accurate would be to say that the contract identified by address is present on the call stack. Wdyt?

srml/contracts/src/exec.rs
```diff
@@ -385,13 +388,29 @@ where
...
 							&executable,
 							nested.new_call_context(caller, value),
 							input_data,
 							gas_meter,
-						)
+						)?;
+
+					// Destroy contract if insufficient remaining balance.
+					if nested.overlay.get_balance(&dest) < nested.config.existential_deposit {
+						let parent = nested.parent
```
pepyakin: I feel a bit weird about this parent construction. Passing `self` through `with_nested_context(|parent, nested|)` seems to me cleaner, did you consider going this way? If so, what are the downsides?

srml/contracts/src/account_db.rs
```diff
@@ -34,11 +34,57 @@ use system;
...
+	/// then the contract code is unmodified. If None and `reset` is true, the contract is deleted.
 	code_hash: Option<CodeHash<T>>,
+	/// If Some(_), then the rent allowance is set to the value. If None and `reset` is false, then
+	/// the rent allowance is unmodified. If None and `reset` is true, the contract is deleted.
 	rent_allowance: Option<BalanceOf<T>>,
 	storage: BTreeMap<StorageKey, Option<Vec<u8>>>,
+	/// If true, indicates that the existing contract and all its storage entries should be removed
+	/// and replaced with the fields on this change entry. Otherwise, the fields on this change
+	/// entry are updates merged into the existing contract info and storage.
+	reset: bool,
```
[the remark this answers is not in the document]
jimpo: Hmm, I prefer `reset` personally. To me, `overwrite` implies there *will* be new data written into the same place, whereas the most common scenario is that the contract is simply deleted.

srml/contracts/src/account_db.rs
```diff
@@ -34,11 +34,57 @@ use system;
...
+				None
+			}
+		})
+	}
+
+	fn code_hash(&self) -> Option<Option<CodeHash<T>>> {
+		if self.reset {
+			Some(self.code_hash)
+		} else {
+			self.code_hash.map(Into::into)
```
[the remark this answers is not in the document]
jimpo: Should just be replaced with `Some`, which wraps the value into an Option and is much easier to read. I'll change that.

srml/contracts/src/exec.rs
jimpo: I'll add that to the comment as well to clarify, but I don't think it's inaccurate to say that the contract is still being executed since it just made a call to another contract.

srml/contracts/src/exec.rs
jimpo: Well, so the parent linking is needed to implement `is_live` because we need to check the whole execution stack. An alternative would be for the `ExecutionContext` to keep a set of all of the "live" contracts (that are being executed). But given how rarely its checked, I think linearly traversing the stack is simpler.

So given that each ExecutionContext already has a reference to its parent for that purpose, I don't see the benefit of passing `parent` into the closure as a separate param since it can be accessed from the child context already.

pepyakin: Ah, yeah, I forgot about `is_live`

srml/contracts/src/tests.rs
```diff
@@ -1926,3 +1926,467 @@ fn deploy_and_call_other_contract() {
...
+		)
+
+		;; Self-destruct by sending full balance to the 0 address.
+		(call $assert
+			(i32.eq
+				(call $ext_call
+					(i32.const 0)	;; Pointer to destination address
+					(i32.const 8)	;; Length of destination address
+					(i64.const 0)	;; How much gas to devote for the execution. 0 = all.
+					(i32.const 8)	;; Pointer to the buffer with value to transfer
```
pepyakin: Wouldn't it be better from the readability standpoint to explicitly declare these data regions and not reuse them for different purposes?

srml/contracts/src/exec.rs
pepyakin: Yep, sure, it wasn't inaccurate it just could be more accurate )

srml/contracts/src/tests.rs
jimpo: Yeah... I was lazy.

Issue #3091, opened 2019-07-10
Unsigned transactions not validated during `Executive::apply_extrinsic`


tomusdrw:
It seems that while you can't propagate incorrect unsigned transaction cause it won't be accepted to the pool you can still craft a block with bullshit and it's going to be processed just fine.

We should add a call to `validate_unsigned` in case we detect the extrinsic is unsigned transaction in `apply_extrinsic` and check that it returns `TransactionValidity::Valid`.

Alternatively the `UnsignedValidator` should have two different methods, where one is called during `validate_transaction` and second during `apply_extrinsic` phase. For convenience we can have a default implementation for the latter to just check for `TransactionValidity::Valid`.

Kianenigma:
Not sure how this can be _confirmed_ but:
- I was for some time under the impression that every transaction (local or propagated) will go through `validate_transaction` anyhow and that is enough. I think you recently told me that it is not the case and only propagated txs that a node receives from the outer world go through this. (or the other way around; can't remember)

either way: if a transaction can avoid going through `validate_transaction` then: `apply_extrinsic` and `validate_transaction` should literally do the same checks and do the same steps (almost) except one is stateless and does not touch the state; hence the issue is valid.

tomusdrw:
> I was for some time under the impression that every transaction (local or propagated) will go through validate_transaction

Every transaction that goes into the pool indeed goes through `validate_transaction`. But when we execute a block received over the network, it's not really the case - we just call `apply_extrinsic` and in current implementation of `Executive::apply_extrinsic` you can find some duplication (signature checking) with `validate_transaction`.

Kianenigma:
Okay then re-thinking out-load here:

Every inherent need to be submitted (for the first time) to __some node__ and that node will __for sure run it through__ `validate_transaction`. Then any other node, as the receiver of this block containing the inherent, can simply trust that the at least one of the nodes has run `validate_transaction()` on it and I can trust that. Ok But then you realize that you actually can't trust that since a node can be malicious hence yeah back to _every `apply_extrinsic()` should also check the inherent.

Unless if since we are explicitly talking about inherent then it is not needed? I don't know the detailed definition of an inherent and what level of security/trust it should carry.

Kianenigma:
I keep coming back to this question: what should go through `validate_ext` and what should not. (everything, including things that come from the pool, goes through `apply_ext` so I think it is safe to assume that _everything_ goes through `apply_extrinsic` at least once). So far I see in the code: 

[...]

PR #3418, opened 2019-08-16
Verify signature and session index during apply phase of im-online.

tomusdrw:
Closes: #3091

We can't rely on checks done in `validate_unsigned`, since they are only done for the transaction pool and during regular execution (i.e. `apply_extrinsic`) we really need to perform them again, to prevent a malicious actor creating a block with such invalid extrinsic.

Since the testing boilerplate for `im-online` is non-existent in `master` I'll add a test for this after #3322 is merged.


andresilva:
Bump `impl_version` and merge it after CI passes :)

tomusdrw:
I've added a bunch of additional docs to `SignedExtension` since it can be used to validate unsigned transactions as well.

PR #3424, opened 2019-08-16
node: update flaming fir

andresilva:


andresilva:
This is required because of the schnorrkel, the generated addresses are different (for the existing seed). I did not remember about this before deploying flaming fir and seeing it fail.

PR #3428, opened 2019-08-17
Fix staking tests

kianenigma:
Apparently, the separate CI job to test staking without the default features (which is needed) was skipped.

### DIFF
--- a/srml/staking/src/tests.rs
+++ b/srml/staking/src/tests.rs
@@ -2076,7 +2076,7 @@ fn reporters_receive_their_slice() {
 			&[Perbill::from_percent(50)],
 		);
 
-		// 1250 x 50% (slash fraction) x 10% (rewards slice)
+		// initial_balance x 50% (slash fraction) x 10% (rewards slice)
 		let reward = initial_balance / 20 / 2;
 		assert_eq!(Balances::free_balance(&1), 10 + reward);
 		assert_eq!(Balances::free_balance(&2), 20 + reward);
--- a/srml/staking/src/tests.rs
+++ b/srml/staking/src/tests.rs
@@ -2058,7 +2058,12 @@ fn reporters_receive_their_slice() {
 	// amount.
 	with_externalities(&mut ExtBuilder::default().build(), || {
 		// The reporters' reward is calculated from the total exposure.
-		assert_eq!(Staking::stakers(&11).total, 1250);
+		#[cfg(feature = "equalize")]
+		let initial_balance = 1250;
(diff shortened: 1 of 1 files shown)


srml/staking/src/tests.rs
```diff
@@ -2072,8 +2077,9 @@ fn reporters_receive_their_slice() {
 		);
 
 		// 1250 x 50% (slash fraction) x 10% (rewards slice)
```
kianenigma:
```suggestion
		// initial_balance x 50% (slash fraction) x 10% (rewards slice)
```

srml/staking/src/tests.rs
```diff
@@ -2103,7 +2114,8 @@ fn invulnerables_are_not_slashed() {
 
 			// The validator 11 hasn't been slashed, but 21 has been.
 			assert_eq!(Balances::free_balance(&11), 1000);
-			assert_eq!(Balances::free_balance(&21), 1750); // 2000 - (0.2 * 1250)
+			// 2000 - (0.2 * 1250)
```
kianenigma:
```suggestion
			// 2000 - (0.2 * initial_balance)
```

kianenigma:
@gabreal is the staking test mandatory to merge? it should be imo.

gabreal:
right now only the `gitlab-test-linux-stable` is mandatory. I'm happy to change if needed.

bkchr:
Why isn't everything that has not `allowed-to-fail=true`(or whatever this is called), mandatory?

bkchr:
Plus, the `staking` test does not detect changes in `srml/staking` correctly. As you see in this pr, the test was not compiled/running.

kianenigma:
yeah as @bkchr said ideally iy __must be mandatory__ for any PR that touches `srml/staking/*`__ from the ci file I assumed this is already the case.

gabreal:
okay so I've additionally made `gitlab-cargo-check-benches`, `gitlab-cargo-check-subkey`, `gitlab-check-web-wasm`, `gitlab-test-linux-stable-int` required for merging anything into the `master` branch. the situation about `srml-staking` is unfortunately not so trivial as it only runs when files have changed in a certain subdirectory and there is no obvious way to only require that then for github. @TriplEight maybe do you have an idea?

kianenigma:
@gabreal Thanks + good that the situation with the staking is at least clear. Regardless of being mandatory or not (imo it should be) it is not being watched properly.

bkchr:
We should maybe just drop the watch property and build it every time like all the other tests.

PR #3430, opened 2019-08-18
Signed transactions now have affinity with runtime versions

gavofyork:
This basically just ensures that transactions (or rather their signature) are linked to the runtime version for which they were intended. It will mean that signed transactions submitted before a runtime upgrade will need to be resubmitted if they don't get included prior to the runtime upgrade going through.

This is important since a runtime upgrade can alter the meaning of a transaction and we don't want a transaction signed for one purpose doing something else when finally executed.

It could have been done by adding a generic param `Get<RuntimeVersion>` to the `CheckVersion` struct, rather than as an associated type of the `system::Trait`, however I figure having the runtime version available to the system trait will probably be a useful thing to have in the future.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4125,6 +4125,7 @@ dependencies = [
  "sr-io 2.0.0",
  "sr-primitives 2.0.0",
  "sr-std 2.0.0",
+ "sr-version 2.0.0",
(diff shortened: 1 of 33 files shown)


gavofyork:
CC @jacogr

PR #3433, opened 2019-08-18
Custom runtime module errors

bkchr:
This pr enables runtime modules to declare a custom error type by using `decl_error!`.

Each module error can be converted to a `DispatchError`, this error carries the module index, the error index and an optional error message. Using the module index and the error index, we should be able to support much nicer error messages in the UI and as a RPC result.

Later, we could expose the `decl_error!` declaration via the metadata to show the actual error message.

CC @xlc @jacogr


core/sr-primitives/src/transaction_validity.rs
```diff
@@ -30,23 +29,240 @@ pub type TransactionLongevity = u64;
...
+		}
+	}
+}
+
+/// An unknown transaction validity.
+#[derive(Clone, PartialEq, Eq, Encode, Decode, Copy)]
+#[cfg_attr(feature = "std", derive(Debug, serde::Serialize))]
+pub enum UnknownTransactionValidity {
+	/// An invalid/unknown account index
+	InvalidIndex,
```
gavofyork: `CannotLookup` would be a better name; indexes are not a first-class concept.

node/executor/src/lib.rs
```diff
@@ -241,7 +247,7 @@ mod tests {
 			None,
 		).0.unwrap();
 		let r = ApplyResult::decode(&mut &v.as_encoded()[..]).unwrap();
-		assert_eq!(r, Err(ApplyError::CantPay));
+		assert_eq!(r, Err(InvalidTransactionValidity::Payment.into()));
```
gavofyork: This terminology is pretty opaque.

By "invalid transaction validity", do you mean that the *validity* of this transaction is invalid? Or the transaction itself? (In which case, it should read "invalid transaction".)

In any case, it replaces `ApplyError::CantPay`, which is

srml/council/src/lib.rs
```diff
@@ -159,11 +163,11 @@ mod tests {
 		type EmergencyVotingPeriod = VotingPeriod;
 		type VotingPeriod = VotingPeriod;
 		type MinimumDeposit = MinimumDeposit;
-		type ExternalOrigin = motions::EnsureProportionAtLeast<_1, _2, u64>;
-		type ExternalMajorityOrigin = motions::EnsureProportionAtLeast<_2, _3, u64>;
-		type EmergencyOrigin = motions::EnsureProportionAtLeast<_1, _1, u64>;
-		type CancellationOrigin = motions::EnsureProportionAtLeast<_2, _3, u64>;
-		type VetoOrigin = motions::EnsureMember<u64>;
+		type ExternalOrigin = motions::EnsureProportionAtLeast<_1, _2, u64, &'static str>;
```
gavofyork: do we really have to have the extra generic param here?

srml/elections/src/lib.rs
```diff
@@ -375,7 +375,7 @@ decl_module! {
 			#[compact] assumed_vote_index: VoteIndex
 		) {
 			let reporter = ensure_signed(origin)?;
-			let who = T::Lookup::lookup(who)?;
+			let who = T::Lookup::lookup(who).ok_or("Could not find `who` account")?;
```
gavofyork: why?

core/sr-primitives/src/lib.rs
```diff
@@ -636,53 +623,127 @@ impl From<ed25519::Signature> for AnySignature {
...
-	/// Successful application (extrinsic reported no issue).
-	Success = 0,
-	/// Failed application (extrinsic was probably a no-op other than fees).
-	Fail = 1,
+/// Reason why an extrinsic couldn't be applied (i.e. invalid extrinsic).
+pub enum ApplyError {
+	/// General error to do with the permissions of the sender.
+	NoPermission,
+
+	/// General error to do with the state of the system in general.
```
pepyakin: Can you elaborate on this? "Bad state" sounds a bit ambigious... 

For example:
1. is a corrupted storage value counts as example of `BadState`? 
1. What if an extrinsic tries to transfer balance from an account that doesn't have enough funds (although it has funds to cover fees), should it also be `BadState`?

srml/contracts/src/gas.rs
```diff
@@ -205,7 +204,7 @@ pub fn buy_gas<T: Trait>(
 	let gas_available = T::BlockGasLimit::get() - <Module<T>>::gas_spent();
 	if gas_limit > gas_available {
 		// gas limit reached, revert the transaction and retry again in the future
-		return Err(BLOCK_FULL);
+		return Err("gas block limit reached");
```
pepyakin: This `BLOCK_FULL` acted as a special value to ultimately signal the block production logic that the block is full. As far as I understand, this was used because there is no other way to convey that signal.

Looking at the code, this message no longer has any semantics. Is there a way to restore previous behavior here? Is it worth it given that we would want to extend and tap onto the weight mechanism?

cc @tomusdrw

core/sr-primitives/src/lib.rs
gavofyork: yeah that's the sort of thing.

srml/contracts/src/gas.rs
gavofyork: indeed, this needs to be `TemporarilyExhausted` or something similar.

core/sr-primitives/src/lib.rs
```diff
@@ -636,53 +623,127 @@ impl From<ed25519::Signature> for AnySignature {
...
+	/// Any error to do with the transaction validity.
+	Validity(transaction_validity::TransactionValidityError),
 }
 
-impl codec::Encode for ApplyOutcome {
-	fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
-		f(&[*self as u8])
+impl ApplyError {
+	/// Returns if the reason for the error was block resource exhaustion.
+	pub fn exhaust_resources(&self) -> bool {
```
tomusdrw:
```suggestion
	pub fn exhausted_resources(&self) -> bool {
```
?

core/sr-primitives/src/lib.rs
```diff
@@ -636,53 +623,127 @@ impl From<ed25519::Signature> for AnySignature {
...
+		match self {
+			Self::Validity(e) => e.exhaust_resources(),
+			_ => false,
+		}
 	}
 }
 
-impl codec::EncodeLike for ApplyOutcome {}
+impl Into<&'static str> for ApplyError {
```
tomusdrw:
```suggestion
impl From<ApplyError> for &'static str {
```
?

core/sr-primitives/src/transaction_validity.rs
```diff
@@ -30,23 +29,240 @@ pub type TransactionLongevity = u64;
...
+	/// General error to do with the transaction not yet being valid (e.g. nonce too high).
+	Future,
+	/// General error to do with the transaction being outdated (e.g. nonce too low).
+	Stale,
+	/// General error to do with the transaction's proofs (e.g. signature).
+	BadProof,
+	/// The transaction birth block is ancient.
+	AncientBirthBlock,
+	/// The transaction **alone** would exhaust the resources of a block.
+	ExhaustResources,
```
tomusdrw:
```suggestion
	ExhaustsResources,
```
? unless there is a non-verb use of `exhaust` too that I'm not aware of.

srml/contracts/src/gas.rs
tomusdrw: Yes!

gavofyork:
needs resolving and some corrections

srml/elections/src/lib.rs
bkchr: By "why" you mean, "why this custom error message?". I changed `Lookup` to return an `Option` instead of a `Result`. Otherwise we would require a `type Error` in the trait and would need to make sure that this is convertible into the module error. Given the fact that a lookup fails in 99% of the cases, because it could not find the item it should lookup, I think we can go with the `Option` as return value.

srml/council/src/lib.rs
bkchr: I don't know why this parameter was there, it is now removed.

srml/contracts/src/gas.rs
bkchr: I added a `CheckBlockGasLimit` signed extension. That needs to be added to the transaction and makes sure that the gas limit is not hit.

srml/contracts/src/lib.rs
```diff
@@ -937,3 +952,62 @@ impl Default for Schedule {
...
+					// gas limit reached, revert the transaction and retry again in the future
+					InvalidTransaction::ExhaustsResources.into()
+				} else {
+					ValidTransaction::default().into()
+				}
+			},
+			Call::__PhantomItem(_, _)  => unreachable!("Variant is never constructed"),
+		}
+	}
+}
```
pepyakin:
```suggestion

}
```

core/sr-primitives/src/transaction_validity.rs
```diff
@@ -30,23 +29,240 @@ pub type TransactionLongevity = u64;
...
-	Unknown(i8),
+	Unknown(UnknownTransaction),
 }
 
-impl From<Result<ValidTransaction, DispatchError>> for TransactionValidity {
-	fn from(r: Result<ValidTransaction, DispatchError>) -> Self {
-		match r {
-			Ok(v) => TransactionValidity::Valid(v),
-			Err(e) => TransactionValidity::Invalid(e.into()),
+impl Into<TransactionValidity> for TransactionValidityError {
```
tomusdrw: Do we have `derive_more` around? We could get rid of all those `Into/From` stuff.

core/sr-primitives/src/transaction_validity.rs
```diff
@@ -30,23 +29,240 @@ pub type TransactionLongevity = u64;
...
+impl Into<crate::ApplyError> for UnknownTransaction {
+	fn into(self) -> crate::ApplyError {
+		TransactionValidityError::from(self).into()
+	}
+}
+
 /// Information on a transaction's validity and, if valid, on how it relates to other transactions.
 #[derive(Clone, PartialEq, Eq, Encode)]
 #[cfg_attr(feature = "std", derive(Debug))]
 pub enum TransactionValidity {
```
tomusdrw: Is it still worth to support this enum? Why not use `Result<ValidatTransaction, TransactionValidityError>` everywhere?

core/sr-primitives/src/transaction_validity.rs
```diff
@@ -30,23 +29,240 @@ pub type TransactionLongevity = u64;
...
+		TransactionValidity::Valid(self)
+	}
+}
+
+impl TransactionValidity {
+	/// Combine two `TransactionValidity`s.
+	///
+	/// If both are valid, they are combined.
+	///
+	/// If one of them is not valid, the non-valid one is returned. If both are not valid, `self` is
```
tomusdrw: Should we prefer `InvalidTransaction` over `UnknownTransaction`? Especially in the context of `exhausted_resources()`. The former indicates the the transaction might still be valid in the future.

tomusdrw: That said, in the aggregator it probably makes sense to propagate the first error that occurred so maybe it's alright.

srml/contracts/src/lib.rs
```diff
@@ -937,3 +952,62 @@ impl Default for Schedule {
 		}
 	}
 }
+
+/// `SignedExtension` that checks if a transaction would exhausts the block gas limit.
+#[derive(Encode, Decode, Clone, Eq, PartialEq)]
+pub struct CheckBlockGasLimit<T: Trait + Send + Sync>(PhantomData<T>);
+
+#[cfg(feature = "std")]
```
tomusdrw:
```suggestion
```
Please remove, we are going to need it if we want to create transaction in the runtime.

srml/contracts/src/lib.rs
```diff
@@ -937,3 +952,62 @@ impl Default for Schedule {
...
+	fn additional_signed(&self) -> rstd::result::Result<(), TransactionValidityError> { Ok(()) }
+
+	fn validate(
+		&self,
+		_: &Self::AccountId,
+		call: &Self::Call,
+		_: DispatchInfo,
+		_: usize,
+	) -> TransactionValidity {
+		let call = match call.is_sub_type() {
```
tomusdrw: I think this might become a common pattern if we use `SignedExtensions` for specific srmls. Might be worth abstracting in the future.

srml/system/src/lib.rs
```diff
@@ -1184,7 +1217,7 @@ mod tests {
 		fn from(e: Event) -> u16 {
 			match e {
 				Event::ExtrinsicSuccess => 100,
-				Event::ExtrinsicFailed => 101,
+				Event::ExtrinsicFailed(err) => Encode::using_encoded(&err, |s| (s[0] as u16) | ((s[1] as u16) << 8)),
```
tomusdrw: What's the reasoning behind this encoding?

core/sr-primitives/src/lib.rs
```diff
@@ -636,53 +623,127 @@ impl From<ed25519::Signature> for AnySignature {
...
+			module,
+			error,
+			message,
+		}
+	}
+}
+
+impl runtime_io::Printable for DispatchError {
+	fn print(&self) {
+		"DispatchError".print();
```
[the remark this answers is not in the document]
bkchr: Yep, but we print numbers, strings etc and we don't support formatting from the runtime. So we only can do this call by call.

core/sr-primitives/src/transaction_validity.rs
bkchr: Before we also always returned the first error. But as we now don't return a Result, I needed to do this manually.

srml/contracts/src/lib.rs
bkchr: Hmm maybe, the problem is that some `SignedExtensions` want to work on the full `Call` enum and some just on the module specific `Call` enum. I'm not sure we can develop any generic solution with the current rust generic type system.

srml/system/src/lib.rs
bkchr: That is some leftover code from the initial version of @xlc. Removed it.

xlc: this is just a way to convert `DispatchError` to `u16`.  DispatchError should serialize to a two byte value (according to the initial discussion, which I am not sure is still the case now)

bkchr: It is now 3 bytes, but the manual conversion isn't required anymore.

core/sr-primitives/src/lib.rs
```diff
@@ -636,53 +623,127 @@ impl From<ed25519::Signature> for AnySignature {
...
+		}
 	}
 }
 
-impl codec::EncodeLike for ApplyOutcome {}
+impl From<ApplyError> for &'static str {
+	fn from(err: ApplyError) -> &'static str {
+		match err {
+			ApplyError::NoPermission => "Transaction does not have required permissions",
+			ApplyError::BadState => "System in bad state",
```
gavofyork:
```suggestion
			ApplyError::BadState => "System state currently prevents this transaction",
```

core/sr-primitives/src/lib.rs
```diff
@@ -636,53 +623,127 @@ impl From<ed25519::Signature> for AnySignature {
...
-	BadSignature = 0,
-	/// Nonce too low.
-	Stale = 1,
-	/// Nonce too high.
-	Future = 2,
-	/// Sending account had too low a balance.
-	CantPay = 3,
-	/// Block is full, no more extrinsics can be applied.
-	FullBlock = 255,
+pub enum ApplyOutcome {
```
gavofyork: Why is this not just `Result<(), DispatchError>`?

core/sr-primitives/src/traits.rs
```diff
@@ -793,61 +804,21 @@ pub trait BlindCheckable: Sized {
...
-	fn check(self, _c: &Context) -> Result<Self::Checked, &'static str> {
-		BlindCheckable::check(self)
-	}
-}
-
-/// An abstract error concerning an attempt to verify, check or dispatch the transaction. This
-/// cannot be more concrete because it's designed to work reasonably well over a broad range of
-/// possible transaction types.
-#[cfg_attr(feature = "std", derive(PartialEq, Debug))]
-pub enum DispatchError {
```
gavofyork: Some of these (`Future`, `BadState`, `Exhausted` and `Payment`) have quite particular meanings that help the transaction queue know that the transaction, though currently invalid, might not be invalid in the future. How is this meaning promulgated now?

gavofyork: NM - I see it was just moved.

srml/contracts/src/lib.rs
```diff
@@ -937,3 +952,61 @@ impl Default for Schedule {
...
+			Some(call) => call,
+			None => return Ok(ValidTransaction::default()),
+		};
+
+		match call {
+			Call::claim_surcharge(_, _) | Call::update_schedule(_) =>
+				Ok(ValidTransaction::default()),
+			Call::put_code(gas_limit, _)
+				| Call::call(_, _, gas_limit, _)
+				| Call::create(_, gas_limit, _, _) => {
```
gavofyork:
```suggestion
				| Call::create(_, gas_limit, _, _) =>
			{
```

gavofyork: The rule is that sequential lines with the same indentation level *must* be syntactically equivalent entities (i.e. in terms of parsing, they should be interchangeable).

srml/elections/src/lib.rs
gavofyork: Could there not be a `LookupError` value common to all modules?

core/sr-primitives/src/testing.rs
```diff
@@ -304,18 +305,21 @@ impl<Origin, Call, Extra> Applyable for TestXt<Call, Extra> where
...
 			Extra::pre_dispatch(extra, &who, &self.1, info, len)?;
 			Some(who)
 		} else {
 			Extra::pre_dispatch_unsigned(&self.1, info, len)?;
 			None
 		};
-		Ok(self.1.dispatch(maybe_who.into()))
+
+		let res = self.1.dispatch(maybe_who.into());
+		Ok(res.map_err(|e| DispatchError::from(e.into()).into()).into())
```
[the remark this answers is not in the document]
bkchr: Applied some magic to make it look better :)

bkchr:
@gavofyork applied your feedback.

core/client/src/block_builder/block_builder.rs
```diff
@@ -104,7 +103,7 @@ where
 				ExecutionContext::BlockConstruction,
 				xt.clone()
 			)? {
-				Ok(ApplyOutcome::Success) | Ok(ApplyOutcome::Fail) => {
+				Ok(Ok(())) | Ok(Err(_)) => {
```
gavofyork:
```suggestion
				Ok(_) => {
```

node/executor/src/lib.rs
```diff
@@ -784,8 +784,10 @@ mod tests {
 		assert!(r.is_ok());
 		let r = WasmExecutor::new()
 			.call(&mut t, 8, COMPACT_CODE, "BlockBuilder_apply_extrinsic", &vec![].and(&xt())).unwrap();
-		let r = ApplyResult::decode(&mut &r[..]).unwrap();
-		assert_eq!(r, Ok(ApplyOutcome::Success));
+		ApplyResult::decode(&mut &r[..])
+			.unwrap()
+			.expect("Extrensic could be applied")
```
gavofyork:
```suggestion
			.expect("Extrinsic could be applied")
```

node/executor/src/lib.rs
```diff
@@ -784,8 +784,10 @@ mod tests {
...
+			.expect("Extrensic did not fail");
```
gavofyork:
```suggestion
			.expect("Extrinsic did not fail");
```

srml/contracts/src/lib.rs
```diff
@@ -937,3 +952,62 @@ impl Default for Schedule {
...
+				// This cannot underflow since `gas_spent` is never greater than `T::BlockGasLimit`.
+				let gas_available = T::BlockGasLimit::get() - <Module<T>>::gas_spent();
+				if *gas_limit > gas_available {
+					// gas limit reached, revert the transaction and retry again in the future
+					InvalidTransaction::ExhaustsResources.into()
+				} else {
+					Ok(ValidTransaction::default())
+				}
+			},
+			Call::__PhantomItem(_, _)  => unreachable!("Variant is never constructed"),
```
DemiMarie-parity: This should be enforced at compile-time.  We can add an empty enum to this variant, for example.

DemiMarie-parity: This should be caught at compile-time.

core/sr-primitives/src/lib.rs
```diff
@@ -967,6 +1013,26 @@ mod tests {
...
+		};
+		let encoded = error.encode();
+		let decoded = DispatchError::decode(&mut &encoded[..]).unwrap();
+		assert_eq!(encoded, vec![1, 1, 2]);
+		assert_eq!(
+			decoded,
+			DispatchError {
+				module: Some(1),
+				error: 2,
+				message: None,
```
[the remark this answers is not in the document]
bkchr: Because it would require to much space in the storage.

core/sr-primitives/src/generic/checked_extrinsic.rs
```diff
@@ -55,28 +53,24 @@ where
 		self.signed.as_ref().map(|x| &x.0)
 	}
 
-	fn validate<U: ValidateUnsigned<Call=Self::Call>>(&self,
+	fn validate<U: ValidateUnsigned<Call = Self::Call>>(
```
kianenigma: are we still using `U` anywhere?

shawntabrizi:
Epic!

PR #3436, opened 2019-08-19
Do not call externalities without `Ext` being set

bkchr:
Closes https://github.com/paritytech/polkadot/issues/385


gavofyork:
tests run ok locally.

Issue #2580, opened 2019-05-14
OnDilution should update total issuance


joepetrowski:
[`on_dilution`](https://github.com/paritytech/substrate/blob/d9c36a02d02adf9b35619b127293e7ae43b7c9a9/srml/treasury/src/lib.rs#L291) mints new funds for the Treasury pot, but does not increase `total_issuance`.

This only gets called from [Staking](https://github.com/paritytech/substrate/blob/d9c36a02d02adf9b35619b127293e7ae43b7c9a9/srml/staking/src/lib.rs#L895) in `new_era`. `reward_validators` correctly updates the `total_issuance` each time it rewards a validator, but `on_dilution` does not.

joepetrowski:
Looking at this a bit further, the pot's funds are added to `total_issuance` when they get spent. So, they do not count as part of total issuance when they are in the pot.

What was the motivation behind this? Maybe I am missing something, but IMO, those funds are in the system and should be counted in issuance when they are added to the pot, not when they are spent.

```
	fn spend_funds() {
		let mut budget_remaining = Self::pot();
		Self::deposit_event(RawEvent::Spending(budget_remaining));

		let mut missed_any = false;
		let mut imbalance = <PositiveImbalanceOf<T>>::zero();
		<Approvals<T>>::mutate(|v| {
			v.retain(|&index| {
				// Should always be true, but shouldn't panic if false or we're screwed.
				if let Some(p) = Self::proposals(index) {
					if p.value <= budget_remaining {
						budget_remaining -= p.value;
						<Proposals<T>>::remove(index);

						// return their deposit.
						let _ = T::Currency::unreserve(&p.proposer, p.bond);

						// provide the allocation.
						imbalance.subsume(T::Currency::deposit_creating(&p.beneficiary, p.value));

						Self::deposit_event(RawEvent::Awarded(index, p.value, p.beneficiary));
						false
					} else {
						missed_any = true;
						true
					}
				} else {
					false
				}
			});
		});

		T::MintedForSpending::on_unbalanced(imbalance);

		if !missed_any {
			// burn some proportion of the remaining budget if we run a surplus.
			let burn = (Self::burn() * budget_remaining).min(budget_remaining);
			budget_remaining -= burn;
			Self::deposit_event(RawEvent::Burnt(burn))
		}

		Self::deposit_event(RawEvent::Rollover(budget_remaining));

		<Pot<T>>::put(budget_remaining);
	}
```

thiolliere:
[...]

PR #3443, opened 2019-08-20
Fix srml-balance burn implementation

thiolliere:
Burn doesn't update total issuance.


thiolliere:
related to https://github.com/paritytech/substrate/issues/2580

PR #3447, opened 2019-08-20
Implement HTTP request in offchain workers

tomaka:
Implements the currently-unimplemented API.

**I'm opening this PR as a draft because I'm going to be away for a bit, and I encourage someone to finish it if they want**.

The only thing that's not implemented is a small TODO to fix: we unwrap if we fail to initialize the `TlsConnector`. How is it even possible to fail to initialize TLS? I have absolutely no idea, and I think it's a platform-specific "just in case" thingy. But since the word `unwrap` is explicitly written in the code, I guess we have to handle this situation.
I'd like to fix that by having some `enum { HttpConnector, HttpsConnector }` where we fall back to the `HttpConnector` if the HTTPS one fails to initialize. We could also just return the error and print a big error saying "failed to initialize offchain workers". But considering that this is the only reason why offchain workers can fail to initialize, it's a bit annoying.

In order to implement this, I had to properly document/define how the API is to be used.

From the point of view of the offchain worker, the order in which functions must be called on an HTTP request is: initialization (`start`), adding headers, sending non-empty body, sending empty body, waiting for answer, reading headers, reading body until EOF.

Most steps can be skipped (for example, reading the body implicitly waits for the answer), but changing the order of the steps is generally not allowed. The exceptions are: you are still allowed to wait for answer or read response headers after having started to read the body, as long as EOF hasn't been returned.

As side changes, I also:
- Moved out some utility functions into a new `timestamp` module.
- Renamed `HttpRequestStatus::Unknown` to `Invalid` and `Timeout` to `IoError` (so that we also handle the situation where the remote force-closes the socket for example).

### DIFF
--- a/core/offchain/src/api/http.rs
+++ b/core/offchain/src/api/http.rs
@@ -1,0 +1,980 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
+// This file is part of Substrate.
+
+// Substrate is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! This module is composed of two structs: [`HttpApi`] and [`HttpWorker`]. Calling the [`http`]
+//! function returns a pair of [`HttpApi`] and [`HttpWorker`] that share some state.
+//!
+//! The [`HttpApi`] is (indirectly) passed to the runtime when calling an offchain worker, while
+//! the [`HttpWorker`] must be processed in the background. The [`HttpApi`] mimicks the API of the
+//! HTTP-related methods available to offchain workers.
+//!
+//! The reason for this design is driven by the fact that HTTP requests should continue running
+//! (i.e.: the socket should continue being processed) in the background even if the runtime isn't
+//! actively calling any function.
+
+use crate::api::timestamp;
+use bytes::Buf as _;
+use fnv::FnvHashMap;
+use futures::{prelude::*, channel::mpsc, compat::Compat01As03};
+use log::error;
+use primitives::offchain::{HttpRequestId, Timestamp, HttpRequestStatus, HttpError};
+use std::{fmt, io::Read as _, mem, pin::Pin, task::Context, task::Poll};
+
+/// Creates a pair of [`HttpApi`] and [`HttpWorker`].
+pub fn http() -> (HttpApi, HttpWorker) {
+	let (to_worker, from_api) = mpsc::unbounded();
+	let (to_api, from_worker) = mpsc::unbounded();
+
+	let api = HttpApi {
+		to_worker,
+		from_worker: from_worker.fuse(),
+		// We start with a random ID for the first HTTP request, to prevent mischievous people from
+		// writing runtime code with hardcoded IDs.
+		next_id: HttpRequestId(rand::random::<u16>() % 2000),
+		requests: FnvHashMap::default(),
+	};
+
+	let engine = HttpWorker {
+		to_api,
+		from_api,
+		// TODO: don't unwrap; we should fall back to the HttpConnector if we fail to create the
+		// Https one; there doesn't seem to be any built-in way to do this
+		http_client: hyper::Client::builder().build(hyper_tls::HttpsConnector::new(1).unwrap()),
+		requests: Vec::new(),
+	};
+
+	(api, engine)
+}
+
+/// Provides HTTP capabilities.
+///
+/// Since this struct is a helper for offchain workers, its API is mimicking the API provided
+/// to offchain workers.
+pub struct HttpApi {
+	/// Used to sends messages to the worker.
+	to_worker: mpsc::UnboundedSender<ApiToWorker>,
+	/// Used to receive messages from the worker.
+	/// We use a `Fuse` in order to have an extra protection against panicking.
+	from_worker: stream::Fuse<mpsc::UnboundedReceiver<WorkerToApi>>,
+	/// Id to assign to the next HTTP request that is started.
+	next_id: HttpRequestId,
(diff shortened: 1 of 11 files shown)


tomaka:
Ah, the CI failure is because obviously we can't compile `hyper` for WASM. I didn't think of that.

bkchr:
Why not use `surf`, they claim to support wasm as a target?

tomaka:
> Why not use surf, they claim to support wasm as a target?

They seem to require async/await.

bkchr:
Ahh okay.
Another question, how do we handle certificates? Or do we just ignore validating them?

tomaka:
> Another question, how do we handle certificates? Or do we just ignore validating them?

This is done by the `hyper_tls` library. We use the regular TLS process, with certificate authorities and all.

core/offchain/src/api/http.rs
```diff
@@ -0,0 +1,980 @@
...
+				Some(HttpApiRequest::Response(HttpApiRequestRp { sending_body, .. })) => {
+					let _ = sending_body.take();
+					continue
+				}
+				_ => continue
+			};
+
+			let (request, _sender) = match self.requests.remove(id) {
+				Some(HttpApiRequest::NotDispatched(rq, s)) => (rq, s),
+				_ => unreachable!("we checked for NotDispatched above; qed")
```
tomusdrw: Wouldn't it be clearer to `remove()` the request earlier and just insert it back if we `continue`? This doesn't seem super future-proof.

core/offchain/src/api/http.rs
```diff
@@ -0,0 +1,980 @@
...
+				id: *id,
+				request
+			});
+
+			// We also destroy the sender in order to forbid writing more data.
+			self.requests.insert(*id, HttpApiRequest::Dispatched(None));
+		}
+
+		// Convert the deadline into a `Future` that resolves when the deadline is reached.
+		let mut deadline = future::maybe_done(match deadline {
```
tomusdrw: Should we extract this to a function? Seems repeated in other methods.

core/offchain/src/api/http.rs
```diff
@@ -0,0 +1,980 @@
...
+					Err(err) => {
+						// This code should never be reached unless there's a logic error somewhere.
+						error!("Failed to read from current read chunk: {:?}", err);
+						return Err(HttpError::IoError)
+					}
+				}
+			}
+
+			// If we reach here, that means the `current_read_chunk` is empty and needs to be
+			// filled with a new chunk from `body`. We block on either the next body of the
```
tomusdrw:
```suggestion
			// filled with a new chunk from `body`. We block on either the next body or the
```

core/offchain/src/api/http.rs
```diff
@@ -0,0 +1,980 @@
...
+			// deadline.
+			let mut next_body = future::maybe_done(response.body.next());
+			futures::executor::block_on(future::select(&mut next_body, &mut deadline));
+
+			if let future::MaybeDone::Done(next_body) = next_body {
```
tomusdrw: Just to make sure: even if deadline is reached we are not going to discard the next body chunk, right? The fact that you've started polling the `response.body.next()` doesn't cause any issues?

core/offchain/src/api/http.rs
```diff
@@ -0,0 +1,980 @@
...
+		/// the next item.
+		body: mpsc::Receiver<Result<hyper::Chunk, hyper::Error>>,
+	},
+	/// A request has failed because of an error.
+	Fail {
+		/// The ID that was passed to the worker.
+		id: HttpRequestId,
+		/// Error that happened.
+		error: hyper::Error,
+	}
```
tomusdrw:
```suggestion
	},
```

core/offchain/src/api/http.rs
```diff
@@ -0,0 +1,980 @@
...
+							me.requests.push((id, HttpWorkerRequest::ReadBody { body, tx }));
+							continue
+						}
+					}
+
+					match Stream::poll_next(Pin::new(&mut body), cx) {
+						Poll::Ready(Some(Ok(chunk))) => {
+							let _ = tx.start_send(Ok(chunk));
+							me.requests.push((id, HttpWorkerRequest::ReadBody { body, tx }));
+							cx.waker().wake_by_ref();	// notify in order to poll again
```
tomusdrw: I'd either add `continue` here for consistency, or remove it from other branches.

mxinden:
> I'd like to fix that by having some enum { HttpConnector, HttpsConnector } where we fall back to the HttpConnector if the HTTPS one fails to initialize. We could also just return the error and print a big error saying "failed to initialize offchain workers". But considering that this is the only reason why offchain workers can fail to initialize, it's a bit annoying.

Falling back to http when https fails should be exploitable by a downgrade attack, right? *Hope I am not missing something.*

tomusdrw:
@mxinden well, so the thing is that no `https` call should work in `HttpConnector`, so it's just a fallback if for some reason:
1. One is using `http://` URL
2. `HttpsConnector/TlsConnector` failed to initialise (as Pierre mentioned in the description).

I think it's a pretty niche case, but might be worth handling anyway.

burdges:
Yeah, I'd avoid the fall back unless it's specifically requested by the caller, probably for testing situations.

tomusdrw:
We are not falling back from `https` to `http` on a requested URL, we are falling back on supporting `http://` requests if we can't initialize `HttpsConnector` (that supports `http://` and `https://` requests).
I don't see any security issues here.

gavofyork:
so, the only instance where an `http:` request will happen is if the caller explicitly requests an `http` URL (and this will work even if the TLS stuff couldn't be initialised)?

tomusdrw:
> so, the only instance where an http: request will happen is if the caller explicitly requests an http URL (and this will work even if the TLS stuff couldn't be initialised)?

Indeed. That's exactly the case we want to handle.

tomusdrw:
Superseded by #3447 (I couldn't push to Pierre's fork).

core/offchain/src/api/http.rs
tomaka: We could put that in the `timestamp.rs` module indeed.

core/offchain/src/api/http.rs
tomaka: Normally no. Calling `next()` wraps around a `&mut Body`, which is a stream. If the `Body` produces an item, `next` is going to be ready and we'll detect it despite the timeout. Destroying the wrapper around the `&mut Body` has no consequence.

core/offchain/src/api/http.rs
tomusdrw: Ok, will do in #3461

PR #3448, opened 2019-08-21
Update soketto to version 0.2.3

twittner:
This version fixes an issue with lifetime elision which causes compilation to fail on recent rustc versions (e.g. 1.39.0-nightly (bea0372a1 2019-08-20)).

See https://github.com/paritytech/soketto/pull/1 for more information.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1959,7 +1959,7 @@ dependencies = [
  "libp2p-core 0.12.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "rw-stream-sink 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -5724,7 +5724,7 @@ name = "twox-hash"
 version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "rand 0.6.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rand 0.3.23 (registry+https://github.com/rust-lang/crates.io-index)",
```
andresilva: Not sure why this changed.

bkchr: Really weird downgrade.

twittner:
About the `rand` downgrade for the `twox-hash` dependency -- I do not know what is causing this either. It is what `cargo update -p soketto` produced. But we can manually edit `Cargo.lock` to retain `rand 0.6.5` for `twox-hash` if that is preferred. FWIW it does compile on my machine.

PR #3454, opened 2019-08-21
Introduce capabilities filtering for off-chain runtime calls.

tomusdrw:
If we call into the runtime, we can pass various `ExecutionContext`s to identify where is it coming from.
Initially this was only affecting how we run the call (a.k.a execution strategy = native / wasm), with introduction of `OffchainWorkers` this also controls the APIs that are available for that part of the runtime.

Recently in #3296 we've introduced additional keystore APIs that must not be available for consensus-critical code, but are very useful for off-chain calls, like rotating the keys - they are only valid in  `ExecutionContext::Other`.

This PR introduces a possibility to further control what APIs (`Capabilties`) are available for particular off-chain call and tries to make it more explicit. This will come handy to implement equivocation reporting for Grandpa/Babe, since we will be able to call into the runtime, produce, sign and submit transaction to the pool in one call. In the future we even want the off-chain call to be able to read (And possibly) write the offchain worker DB, for instance to retrieve the merkle proofs for session membership.

### DIFF
--- a/core/client/src/client.rs
+++ b/core/client/src/client.rs
@@ -27,8 +27,9 @@ use codec::{Encode, Decode};
 use hash_db::{Hasher, Prefix};
 use primitives::{
 	Blake2Hasher, H256, ChangesTrieConfiguration, convert_hash,
-	NeverNativeValue, ExecutionContext,
-	storage::{StorageKey, StorageData, well_known_keys}, NativeOrEncoded
+	NeverNativeValue, ExecutionContext, NativeOrEncoded,
+	storage::{StorageKey, StorageData, well_known_keys},
+	offchain,
 };
 use substrate_telemetry::{telemetry, SUBSTRATE_INFO};
 use sr_primitives::{
@@ -1471,7 +1472,7 @@ impl<B, E, Block, RA> CallRuntimeAt<Block> for Client<B, E, Block, RA> where
 		context: ExecutionContext,
 		recorder: &Option<Rc<RefCell<ProofRecorder<Block>>>>,
 	) -> error::Result<NativeOrEncoded<R>> {
-		let enable_keystore = context.enable_keystore();
+		let capabilities = context.capabilities();
 
 		let manager = match context {
 			ExecutionContext::BlockConstruction =>
@@ -1480,16 +1481,19 @@ impl<B, E, Block, RA> CallRuntimeAt<Block> for Client<B, E, Block, RA> where
 				self.execution_strategies.syncing.get_manager(),
 			ExecutionContext::Importing =>
 				self.execution_strategies.importing.get_manager(),
+			ExecutionContext::OffchainCall =>
+				self.execution_strategies.other.get_manager(),
+			ExecutionContext::RichOffchainCall(_) =>
+				self.execution_strategies.other.get_manager(),
 			ExecutionContext::OffchainWorker(_) =>
 				self.execution_strategies.offchain_worker.get_manager(),
-			ExecutionContext::Other =>
-				self.execution_strategies.other.get_manager(),
 		};
 
 		let mut offchain_extensions = match context {
-			ExecutionContext::OffchainWorker(ext) => Some(ext),
+			ExecutionContext::OffchainWorker(ext)
+			| ExecutionContext::RichOffchainCall(ext) => Some(ext),
 			_ => None,
-		};
+		}.map(|ext| offchain::LimitedExternalities::new(capabilities, ext));
 
 		self.executor.contextual_call::<_, _, fn(_,_) -> _,_,_>(
 			|| core_api.initialize_block(at, &self.prepare_environment_block(at)?),
@@ -1502,7 +1506,7 @@ impl<B, E, Block, RA> CallRuntimeAt<Block> for Client<B, E, Block, RA> where
 			native_call,
 			offchain_extensions.as_mut(),
 			recorder,
-			enable_keystore,
+			capabilities.has(offchain::Capability::Keystore),
 		)
 	}
 
--- a/core/client/src/client.rs
+++ b/core/client/src/client.rs
@@ -1472,8 +1472,6 @@ impl<B, E, Block, RA> CallRuntimeAt<Block> for Client<B, E, Block, RA> where
 		context: ExecutionContext,
 		recorder: &Option<Rc<RefCell<ProofRecorder<Block>>>>,
 	) -> error::Result<NativeOrEncoded<R>> {
-		let capabilities = context.capabilities();
-
 		let manager = match context {
 			ExecutionContext::BlockConstruction =>
 				self.execution_strategies.block_construction.get_manager(),
@@ -1481,17 +1479,15 @@ impl<B, E, Block, RA> CallRuntimeAt<Block> for Client<B, E, Block, RA> where
 				self.execution_strategies.syncing.get_manager(),
 			ExecutionContext::Importing =>
 				self.execution_strategies.importing.get_manager(),
-			ExecutionContext::OffchainCall =>
-				self.execution_strategies.other.get_manager(),
-			ExecutionContext::RichOffchainCall(_) =>
-				self.execution_strategies.other.get_manager(),
-			ExecutionContext::OffchainWorker(_) =>
+			ExecutionContext::OffchainCall(Some((_, capabilities))) if capabilities.has_all() =>
 				self.execution_strategies.offchain_worker.get_manager(),
+			ExecutionContext::OffchainCall(_) =>
+				self.execution_strategies.other.get_manager(),
 		};
 
+		let capabilities = context.capabilities();
 		let mut offchain_extensions = match context {
-			ExecutionContext::OffchainWorker(ext)
-			| ExecutionContext::RichOffchainCall(ext) => Some(ext),
+			ExecutionContext::OffchainCall(ext) => ext.map(|x| x.0),
 			_ => None,
 		}.map(|ext| offchain::LimitedExternalities::new(capabilities, ext));
 
--- a/core/primitives/src/lib.rs
+++ b/core/primitives/src/lib.rs
@@ -100,8 +100,10 @@ impl ExecutionContext {
 		use ExecutionContext::*;
 
 		match self {
-			Importing | Syncing | BlockConstruction | OffchainCall(None) =>
+			Importing | Syncing | BlockConstruction =>
 				offchain::Capabilities::none(),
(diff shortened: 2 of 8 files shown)


core/primitives/src/offchain.rs
```diff
@@ -430,6 +475,147 @@ impl<T: Externalities + ?Sized> Externalities for Box<T> {
...
+			self.externalities.sleep_until(deadline)
+		}
+	}
+
+	fn random_seed(&mut self) -> [u8; 32] {
+		if self.check(Capability::Randomness).is_ok() {
+			self.externalities.random_seed()
+		} else {
+			[0_u8; 32]
+		}
```
marcio-diaz: I think this should return a `Result`.

core/primitives/src/lib.rs
```diff
@@ -88,19 +88,36 @@ pub enum ExecutionContext {
...
+	/// Context used for rich offchain calls.
+	///
+	/// This includes access to offchain worker db (read / write)
+	/// and transaction pool.
+	RichOffchainCall(Box<dyn offchain::Externalities>),
+	/// A context used for offchain workers.
+	///
+	/// This should include all capabilities of `offchain::Externalities`.
 	OffchainWorker(Box<dyn offchain::Externalities>),
-	/// Context used for other calls.
```
marcio-diaz: Are all of these exclusive?

core/primitives/src/offchain.rs
tomusdrw: Yeah, I was thinking about it. But the thing is that:
1. If you are calling it properly it will never fail.
2. If you are calling it in wrong context it will always fail.

So going into `else` branch indicates some serious issue with the runtime code - it means that someone is using these functions in places where they shouldn't. I could also consider just panicking here, but I feel this would be ugly and may potentially cause consensus issues in the future - so now I'm just printing a warning.

core/primitives/src/lib.rs
tomusdrw: Yes.

core/sr-api-macros/src/decl_runtime_apis.rs
```diff
@@ -552,7 +552,7 @@ impl<'a> ToClientSideDecl<'a> {
 	fn fold_trait_item_method(&mut self, method: TraitItemMethod)
 		-> (TraitItemMethod, Option<TraitItemMethod>, TraitItemMethod) {
 		let crate_ = self.crate_;
-		let context_other = quote!( #crate_::runtime_api::ExecutionContext::Other );
+		let context_other = quote!( #crate_::runtime_api::ExecutionContext::OffchainCall );
 		let fn_impl = self.create_method_runtime_api_impl(method.clone());
```
marcio-diaz: Maybe rename `context_other`.

core/client/src/client.rs
```diff
@@ -1480,16 +1481,19 @@ impl<B, E, Block, RA> CallRuntimeAt<Block> for Client<B, E, Block, RA> where
 				self.execution_strategies.syncing.get_manager(),
 			ExecutionContext::Importing =>
 				self.execution_strategies.importing.get_manager(),
+			ExecutionContext::OffchainCall =>
+				self.execution_strategies.other.get_manager(),
+			ExecutionContext::RichOffchainCall(_) =>
+				self.execution_strategies.other.get_manager(),
```
marcio-diaz: Should we modify the `CLI`?

core/primitives/src/offchain.rs
```diff
@@ -430,6 +475,147 @@ impl<T: Externalities + ?Sized> Externalities for Box<T> {
...
+			#[cfg(feature = "log")]
+			log::warn!("Accessing a forbidden API. No: {:?} capability.", capability);
+			Err(())
+		}
+	}
+}
+
+impl<T: Externalities> Externalities for LimitedExternalities<T> {
+	fn is_validator(&self) -> bool {
+		if self.check(Capability::Keystore).is_ok() {
```
bkchr: `is_validator` does not accesses the keystore. Maybe we just need a `General` capability or something similar.

core/primitives/src/lib.rs
```diff
@@ -88,19 +88,36 @@ pub enum ExecutionContext {
...
-	/// Offchain worker context.
+	/// Context used for offchain calls.
+	///
+	/// Includes keystore capabilities.
+	OffchainCall,
+	/// Context used for rich offchain calls.
+	///
+	/// This includes access to offchain worker db (read / write)
+	/// and transaction pool.
+	RichOffchainCall(Box<dyn offchain::Externalities>),
```
bkchr: As we talked on the retreat, I really would like to merge these two calls. (`RichOffchainCall` and `OffchainCall`)

I don't know if you want to do this in this pr? Could also be done by me/you in a later pr.

core/client/src/client.rs
tomusdrw: `Other` still encapsulates previous kinds of calls, I don't there is any reason to be so specific about when we call WASM and when not. Both `OffchainCall` and `RichOffchainCall` are only going to be used when calling into runtime from node and I think it's fine for them to have the same strategy.

core/primitives/src/offchain.rs
tomusdrw: To be honest I question the existence of this API at all:
1. The only reason I can see to check if someone is a validator is to avoid running the code that would fail at signing stuff (i.e. requires keystore anyway). I'd rather test if keystore public keys are on the session list instead - the fact that someone is configured as a validator doesn't really mean that they are validating.
2. Offchain workers are disabled for non-validators by default, if someone enables them from CLI this check would unnecessarily prevent them from doing so. Note this check is by no means a security measure, since everyone can just take the offchain worker code and run it without the check anyway.

I'd rather have a way to filter out offchain worker modules one wants to run from CLI than introducing such checks.

core/primitives/src/lib.rs
tomusdrw: Well, the whole point of this PR is to actually split the two calls. I don't want to allow people to make HTTP calls from the runtime everytime the node calls into some function.
After discussions with Andre and Rob we'd figured out that `RichOffchainCall` is going to be specifically needed for some cases (like equivocation report generation) and I think we should really limit the APIs that are available in such case (i.e. we only want the ones listed on the capabilities list). I think I'll even split read and writes to offchain worker db, so that `RichOffchainCall` is only allowed to read.

tomusdrw:
> Otherwise I see us at some point debugging a problem that will have its root cause in such a function.

That's why we print a warning if such function is being used. I'm really scared of `panicking` in the code that:
1. Is very mixed up with regular, consensus-critical runtime code (it's just `RuntimeApi`)
2. Is being called very frequently by the node
3. Most likely will not run in wasm sandbox, but as `native` code.

Any mistake in a code like this will cause the entire network to brick forever.

core/primitives/src/offchain.rs
bkchr: We introduced the code, so that nodes with a key that are not sending im-online messages. If only validators are running offchain-code. Then there is no reason at all for this code. So, we should remove the function completely. Maybe keep a stub to not break kusama, but always return true and remove the usage in im-online.

bkchr:
For that exists testing, if you deploy code that panics it is your own fault. You should not use these functions from within the chain.
If you print a warning, no one will see it or just ignore it. We panic anyway when no offchain context is set, so I don't see the difference.

core/primitives/src/lib.rs
bkchr: I like the idea, I would just make `OffchainCall` take the capabilities and make them by default allow nothing or so. I don't see why we want to maintain two versions. But we can do this later, when I refactor the keystore usage.

core/primitives/src/offchain.rs
tomusdrw: If we want to remove it we should do it in a separate PR - it's not related. I'm fine with leaving it, since it's already here, but don't think there is enough reasons to introduce a separate capability for it - it would be a super-set of `Keystore` anyway.

tomusdrw:
> If you print a warning, no one will see it or just ignore it. We panic anyway when no offchain context is set, so I don't see the difference.
Offchain workers are much better separate from other parts of the runtime code IMHO (it's one special function that is being called). Regular `RuntimeApi`s are usually a code that is shared with consensus-crticial code.

> For that exists testing, if you deploy code that panics it is your own fault. You should not use these functions from within the chain.

IMHO we've experienced multiple times that testing is never enough, and just saying: "It's your fault to brick a $1M network" is a lame excuse to introduce security measures to prevent that.

That said, I'm fine to bend to pressure as I'm also torn apart between having a clean code / easy debugging experience and catering for possible future (possibly unlikely) mistakes. Will update the code shortly.

bkchr:
> 
> IMHO we've experienced multiple times that testing is never enough, and just saying: "It's your fault to brick a $1M network" is a lame excuse to introduce security measures to prevent that.
> 

That is correct, but the code will fail anyway when you try to call it, as the `offchain` worker is not set.

bkchr:
We can also return a `Result` and fail in `sr-io`, as we do it for the other stuff as well.

core/primitives/src/offchain.rs
bkchr: Yeah

core/primitives/src/lib.rs
bkchr: As discussed "offchain", we will revisit this later.

core/primitives/src/offchain.rs
tomusdrw: @DemiMarie-parity That sounds good, but is also way more involved than simple filtering. The main reason is that runtime is a single blob that tries to import a bunch of external calls, we would need to split the runtime into multiple separate modules that import different set of external calls depending on the context (i.e. offchain workers, RuntimeApis, etc). That sounds like a serious undertake, with no huge advantages - the call would never run vs failing when run.

core/primitives/src/offchain.rs
```diff
@@ -230,6 +230,51 @@ impl Timestamp {
...
+
+	/// Check if particular capability is enabled.
+	pub fn has(&self, capability: Capability) -> bool {
+		self.0 & capability as u8 != 0
+	}
+}
+
+impl<'a> From<&'a [Capability]> for Capabilities {
+	fn from(list: &'a [Capability]) -> Self {
+		Capabilities(list.iter().fold(0_u8, |a, b| a | *b as u8))
```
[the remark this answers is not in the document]
tomusdrw: I find the previous notation more readable, and we have a bunch of usages like that already in the code.

PR #3455, opened 2019-08-21
typo

thiolliere:


thiolliere:
> CI fails with:
> 
> > wasm source files changed but not the spec/impl version and the runtime
> > binary blob. If changes made do not alter logic, just bump 'impl_version'.
> > If they do change logic, bump 'spec_version' and rebuild wasm.
> 
> As far as I know you would need to bump the _impl_ version here: https://github.com/paritytech/substrate/blob/master/node/runtime/src/lib.rs#L84.

This test is not required

PR #3456, opened 2019-08-21
Refactor: fixed point arithmetic for SRML.

### LINKED ISSUE #2908: Make use of Perbill in staking/phragmen for readibility
We still want to keep PerU128 as it is, but we can get read of ACCURACY by replacing operation using ACCURACY by a Perbill implementation.

Note that using Perbill implementation is less precise than the current const. If we really want that precision we can create a PerU32 in the same way of Perbill, or even a PerU64 is actually implementable (because we need double space and u128 is available).

This would improve readibility of the code.

follows https://github.com/paritytech/substrate/pull/2889,

see this comment https://github.com/paritytech/substrate/pull/2889#issuecomment-503477814

### LINKED ISSUE #3189: Aggregate + Clean runtime fixed point arithmetic types
### Currently we have: 

- Perbill
- Permill
- PerU128
- Fixed64

###  Need to add: 

- [ ] [`RationalU32 `](https://github.com/paritytech/substrate/pull/2882#discussion_r306745580)
- [ ] Clean up PerU128
- [ ] Make the Api and exposed functions of _all_ of them consistent. All should expose `normal`, `checked`, `saturating` etc.

### Maybe

- [ ] Move to a new sr-arithmetic crate or arithmetic.rs in `sr-primitives`.
- [ ] Extensive and unified tests for all of them.
- [ ] Should probably close ~https://github.com/paritytech/substrate/issues/2558~ https://github.com/paritytech/substrate/issues/2908 along the way.
- [ ] There are some usages of multiplication and power of 2 in https://github.com/paritytech/substrate/pull/3322

CC @thiolliere

kianenigma:
closes #3189 and resolves #2908 

- [x] A macro generates a unified code for all per-things. 
   - [x] Does not work for types that have 64-bit native size while it should. e.g. I want to add:

```
implement_per_thing!(Perquintill, test_perquintill, 10.pow(18), u64, u128, "_Parts per Quintillion_");
```
- [x] u128 is gone and a generic `Rational128` is added with some basic operations.
- [x] Phragmen is refactored and uses the `Rational128`
- [x] Adding mul and power impl to per-things (still needed @pepyakin?) 
- [x] Has still a bunch of TODOs in the code. 
- [x] Much much much more tests.

cc @thiolliere


kianenigma:
I won't be fully available till mid next week but still could use a round of reviews now. Note the two checklists up top: could still use some docs/test and has some TODOs.

As anticipated it also touches phragmen and uses a `Rational128` to store very large and very accurate fractions. tbh it was a bit harder to do actually than the PerU128 but I believe it is a nicer approach. Basically I ended up using a small niche portion of the functionality that a `Rational128` can provide in phragmen because I wanted to have the denominator _almost always_ fixed to the largest possible number.

marcio-diaz:
Did you consider using a [fixed](https://docs.rs/fixed/0.4.5/fixed/) and [num-rational](https://docs.rs/num-rational/0.2.2/num_rational/) crates? I think we may need to implement a few more types to get a better performance/precision tradeoff.

kianenigma:
> Did you consider using [fixed](https://docs.rs/fixed/0.4.5/fixed/) and [num-rational](https://docs.rs/num-rational/0.2.2/num_rational/) crates? I think we may need to implement a few more types to get a better performance/precision tradeoff.

In the past we dismissed such options because we didn't nee the entire functionality and adding a library to srml as dependency might not be preferable. Maybe in the future. This PR is to mostly refactor what we already have.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,974 @@
...
+
+			/// Approximate the fraction `p/q` into a per-thing fraction. This will never overflow.
+			///
+			/// The computation of this approximation is performed in the generic type `N`. Given
+			/// `M` as the data type that can hold the maximum value of this per-thing (e.g. u32 for
+			/// perbill), this can only work if `N == M` or `N: From<M> + TryInto<M>`.
+			pub fn from_rational_approximation<N>(p: N, q: N) -> Self
+				where N: Clone + Ord + From<$type> + TryInto<$type> + ops::Div<N, Output=N>
+			{
+			// TODO: double check with @thiolliere the new bound. If your fraction type has a max
```
kianenigma: @thiolliere I can really use a double-check on how I changed things here.

thiolliere: oh I missed the transition from draft to reviewable I'll take a look for sure

kianenigma: thanks 👍

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1013 @@
...
+					* ($max as $upper_type)
+					/ q_reduce as $upper_type;
+
+				$name(part as $type)
+			}
+		}
+
+		impl Saturating for $name {
+			fn saturating_add(self, rhs: Self) -> Self {
+				Self::from_parts(self.0 + rhs.0)
```
thiolliere: this is also guaranteed not to overflow because of tests no ? like imagine we want to implement PerTenquintill then this operation should have been done with saturating_add.

Actually maybe we can do saturating_add here just so its sure anyway. (even if it is already sure for our types)

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1013 @@
...
+	"_Percent_",
+);
+
+/// An unsigned fixed point number. Can hold any value in the range [-9_223_372_036, 9_223_372_036]
+/// with fixed point accuracy of one billion.
+#[cfg_attr(feature = "std", derive(Debug))]
+#[derive(Encode, Decode, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Fixed64(i64);
+
+/// The maximum value of the `Fixed64` type
```
thiolliere: what does that mean doesn't Fixed64 can be more than a billion actually ?

core/phragmen/src/lib.rs
```diff
@@ -291,48 +282,65 @@ pub fn elect<AccountId, Balance, FS, C>(
...
 
 			if diff_per_vote > 0 {
-				for i in 0..l {
-					assignment.1[i%l].1 =
-						assignment.1[i%l].1
-						.saturating_add(diff_per_vote);
+				for i in 0..len {
+					let current_ratio = assignment.1[i%len].1;
+					let next_ratio = current_ratio.saturating_add(Perbill::from_parts(diff_per_vote));
+					assignment.1[i%len].1 = next_ratio;
```
gavofyork:
```suggestion
					assignment.1[i % len].1 = next_ratio;
```

core/phragmen/src/lib.rs
```diff
@@ -291,48 +282,65 @@ pub fn elect<AccountId, Balance, FS, C>(
...
+			let diff = accuracy.checked_sub(sum).unwrap_or(0);
+			let diff_per_vote = (diff / vote_count).min(accuracy);
 
 			if diff_per_vote > 0 {
-				for i in 0..l {
-					assignment.1[i%l].1 =
-						assignment.1[i%l].1
-						.saturating_add(diff_per_vote);
+				for i in 0..len {
+					let current_ratio = assignment.1[i%len].1;
```
gavofyork:
```suggestion
					let current_ratio = assignment.1[i % len].1;
```

core/phragmen/src/lib.rs
```diff
@@ -291,48 +282,65 @@ pub fn elect<AccountId, Balance, FS, C>(
...
 
-			// `remainder` is set to be less than maximum votes of a voter (currently 16).
+			// `remainder` is set to be less than maximum votes of a nominator (currently 16).
 			// safe to cast it to usize.
 			let remainder = diff - diff_per_vote * vote_count;
 			for i in 0..remainder as usize {
-				assignment.1[i%l].1 =
-					assignment.1[i%l].1
-					.saturating_add(1);
+				let current_ratio = assignment.1[i%len].1;
```
gavofyork:
```suggestion
				let current_ratio = assignment.1[i % len].1;
```

core/phragmen/src/lib.rs
```diff
@@ -291,48 +282,65 @@ pub fn elect<AccountId, Balance, FS, C>(
...
+				let next_ratio = current_ratio.saturating_add(Perbill::from_parts(1));
+				assignment.1[i%len].1 = next_ratio;
```
gavofyork:
```suggestion
				assignment.1[i % len].1 = next_ratio;
```

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1013 @@
...
+	}
+
+	/// Performs [`multiply_by_rational`]. In case of failure, it greedily tries to shift the
+	/// numerator (`b`) and denominator (`c`) (looking at the whole thing is `a * b / c`) until the
+	/// multiplication fits. This is guaranteed to work if `b < c`.
+	///
+	/// In case `b > c` and overflow happens, `a` is returned.
+	///
+	/// c must be greater than or equal to 1.
+	pub fn safe_multiply_by_rational(a: u128, b: u128, c: u128) -> u128 {
```
thiolliere: not sure safe_ is a good name maybe multiply_by_rational_or_$shortNameOfTheAlgo

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1013 @@
...
+		128 - n.leading_zeros()
+	}
+
+	/// Tries to compute `a * b / c`. The approach is:
+	///   - Simply try a * b / c.
+	///   - Swap the operations in case the former multiplication overflows. Divide first. This
+	///     might collapse to zero.
+	///
+	/// If none worked then return Error.
+	pub fn multiply_by_rational(a: u128, b: u128, c: u128) -> Result<u128, &'static str> {
```
thiolliere: isn't the second approach a bit too much unprecise in some case like this one 
`assert_eq!(multiply_by_rational(2*MAX64 - 1, MAX64, MAX64).unwrap(), 2*MAX64 - 1);`

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1013 @@
...
+			// using `SimpleArithmetic` here was okay as long as we don't add a new fraction type
+			// which is embedded in a u64. With perquintill we have to change this.
+				// q cannot be zero.
+				let q = q.max((1 as $type).into());
+				// p should not be bigger than q.
+				let p = p.min(q.clone());
+
+				let factor = (q.clone() / $max.into()).max((1 as $type).into());
+
+				// q cannot overflow: (q / (q/$max)) < $max. p < q hence p also cannot overflow.
```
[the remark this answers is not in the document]
kianenigma: you are right that my comment is wrong but it can still not overflow. Look at this example: 

```rust
let max = 1_000_000_000_u32;
Perbill::from_rational_approximation(10u32, max + max - 1);
// p 10 / q 1999999999 / factor 1
// pr 10 / qr 1999999999
Perbill::from_rational_approximation(10u32, max + max);
// p 10 / q 2000000000 / factor 2
// pr 5 / qr 1000000000
Perbill::from_rational_approximation(10u32, max + max + 1);
// p 10 / q 2000000001 / factor 2
// pr 5 / qr 1000000000
```

q_reduce can exceed $max, but it can never exceed 2 * $max. Nonetheless a great catch and I should  update this assertion:

https://github.com/paritytech/substrate/blob/14fe8ac16d142e5c70aafcc1cd990b04e7374755/core/sr-primitives/src/sr_arithmetic.rs#L205

to `2 * $max`

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: Now with the new assertion that `$max < $type::max_value()` this is safe right? `self.0` and `rhs.0` are both less than `$max` and will still fit into `$type`, and the `from_parts` coerces it to `$max` if it is more than `$max`. 

will make it saturating anyhow.

thiolliere: yes ok I'm ok as is actually

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1013 @@
...
+			}
+		}
+
+		/// Overflow-prune multiplication.
+		///
+		/// tailored to be used with a balance type.
+		impl<N> ops::Mul<N> for $name
+		where
+			N: Clone + From<$type> + UniqueSaturatedInto<$type> + ops::Rem<N, Output=N>
+				+ ops::Div<N, Output=N> + ops::Mul<N, Output=N> + ops::Add<N, Output=N>,
```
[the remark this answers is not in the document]
kianenigma: Yeah, changing `From<u32> + UniqueSaturatedInto<u32>` to `From<$type> + UniqueSaturatedInto<$type>`  actually only makes a difference for that. But I think it is mandatiry: 

correct me if I am wrong but afair you added this implementation with the mindset that:
2- it operates in type `N` which the user provides.
1- it does not work for all `N` types but if does, then it does not overflow. 

Before, both perthings were smaller than u32 and `From<u32>` was enough. Perquintill needs u64 and it is natural that if you want to, for example, multiply `10` by `Perquintill::from_percent(50)`, you need to pass `10` as `10u64` (or `10u128` in this case), hence delcaring `N` as `u64` which makes all operations safe.


> if $max.try_into<N>() is err that means that N fits into $type so we could just make use of upper type no ?
> maybe we should keep as it is though and user will just require a new bound on its type he wants to multiply with perquintill

But I can't get what you mean here, let me know if I am missing something or my argument above is incorrect.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1013 @@
...
+			fn from(x: codec::Compact<$name>) -> Self {
+				x.0
+			}
+		}
+
+		#[cfg(test)]
+		mod $test_mod {
+			use codec::{Encode, Decode};
+			use super::{$name, Saturating};
+			use srml_support::assert_eq_error_rate;
```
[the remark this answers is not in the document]
kianenigma: true, I added it there just to have it next to out other custom assertions 👍

core/sr-primitives/src/sr_arithmetic.rs
```suggestion
/// The accuracy of the `Fixed64` type.
```

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: well, that's one of my problems here. I don't know the name of the algorithm and I cam up with it. I think it is simple and sound, but a proof check would be great. 

As for the name, `safe_` implies that it always returns, but it actually can return an inaccurate value. I also like the name `multiply_by_rational_or_greedy()`?  The initial version actually attempted to greedily shift to find the spot where the multiplication works. Recently I changed it to something a bit smarter.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1013 @@
...
+			((n as i128).saturating_mul(DIV as i128) / (d as i128).max(1))
+			.try_into()
+			.unwrap_or(Bounded::max_value())
+		)
+	}
+
+	/// Performs a saturated multiply and accumulate by unsigned number.
+	///
+	/// Returns a saturated `n + (self * n)`.
+	pub fn saturated_multiply_accumulate<N>(&self, int: N) -> N
```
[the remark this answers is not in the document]
kianenigma: maybe the api is a bit vague (the doc is vague, it should be _ Returns a saturated `int + (self * int)`._) but your example should return 10 and it is correct. It reads like: 

```
10 + ( -1/1_000_000_000  *  10) ==> 10
```

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1013 @@
...
+			/// This can be called at compile time.
+			pub const fn from_parts(parts: $type) -> Self {
+				Self([parts, $max][(parts > $max) as usize])
+			}
+
+			/// Converts from a percent. Equal to `x / 100`.
+			///
+			/// This can be created at compile time.
+			pub const fn from_percent(x: $type) -> Self {
+				Self([x, 100][(x > 100) as usize] * ($max / 100))
```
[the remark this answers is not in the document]
kianenigma: No that could very easily overflow for sth like Perbill with a percent of 4 or more. As long as $max is bigger than 100 for all per_things, we are safe. Alternatively, we can compute the multiplication first in the $upper_type but in favor of speed I prefer, for now, to just disallow the macro to be expanded for $max < 100, adding

```rust 
fn macro_expanded_correctly() {
	// needed for the `from_percent` to work.
	assert!($max >= 100);
        ....
```

core/sr-primitives/src/sr_arithmetic.rs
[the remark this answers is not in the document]
thiolliere: no everything you say is correct just maybe we can change a bit the behavior so that PerQuintill works with SimpleArithmetic.

>> if $max.try_into() is err that means that N fits into $type so we could just make use of upper type no ?

This means: for PerQuintill the change can be made this way: use the bound `TryInto<u64>` instead of `From<u64>` which is not on SimpleArithmetic.
```rust
				let maximum: N = $max.try_into();
```
Here if maximum is err that means that N is less u64 thus `b: N` can be converted to u64 without any issue. if maximum is not err then fine same as now.

>> maybe we should keep as it is though and user will just require a new bound on its type he wants to multiply with perquintill

I think my above proposal is too complex and probably doesn't worth it. the user can add an additional bound on its type if he wants to.

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: Yeah, frankly, this is horrible. I will rethink it.

Though since we don't really use it directly. If we couldn't come up with some other approach, the function can/should be removed altogether. Then `multiply_by_rational_greedy()` just does the first part of this function and falls back to the greedy approach if it fails.

core/sr-primitives/src/sr_arithmetic.rs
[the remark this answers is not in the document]
thiolliere: oh I missread, doc is fine

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: bad doc. updated.

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: This is equal to computing:

```
// ideally 
31 * 15  / 16 ~= 30
// but gives 
31 * 15  / 16 ~= 15
```

in an 8-bit number scale and I don't see any other trivial way to do it. You multiply first, you overflow. you divide first, either you will collapse to zero (15 / 16) or you will lose a lot of accuracies (31 / 16)

Swapping is useful when `a` is many orders of magnitude bigger than `b` and `c`. But I couldn't really formulate how it can be used safely.. hence: 
- remove the swap
- if `b < c` then we build a perQuintill and use that. 
- otherwise, we don't do anything and return an error. 

This makes some tests that were particularly easy and accurate with the swapping method to become a bit inaccurate, but less horrible

```rust
assert_eq!(
	multiply_by_rational_greedy(MAX128, 2, 3),
	226854911280625642 082061493673886498661,
        // with swap this would accurately be computed to: MAX128 / 3 * 2
        // which is
     // 226854911280625642 308916404954512140970
);
``` 

which is still inaccurate but better I suppose.

cc @dvdplm

core/sr-primitives/src/sr_arithmetic.rs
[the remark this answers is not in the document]
thiolliere: yes this work perfectly for now as our max are always a multiple of 100, if we start introducing PerMaxU32 (which part represent part/u32::MAX) then this conversion lose information.
do the following assert then `assert!($max%100 == 0)` and add a comment on top of this division saying $max is multiple of 100 so no information is lost.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1097 @@
...
+	where
+		N: TryFrom<u64> + From<u32> + UniqueSaturatedInto<u32> + Bounded + Clone + Saturating +
+		ops::Rem<N, Output=N> + ops::Div<N, Output=N> + ops::Mul<N, Output=N> +
+		ops::Add<N, Output=N>,
+	{
+		let parts = self.0;
+		let positive = parts > 0;
+
+		// will always fit.
+		let natural_parts: u64 = (parts / DIV) as u64;
```
thiolliere: if parts is negative and inferior to -DIV then this result in a big integer number.
as we compute int + int*self. self being a small less than -1 or a very big less than -1 number doesn't change the result but maybe we should add a comment or restructure the logic so that negative case is handle more understandably.

thiolliere: also (but this is an issue in whole substrate maybe) the traits bounded to N doesn't force N to be a positive arithmetic type. In the case N is i128 this function is wrong.

kianenigma: I see. Honestly, I entirely missed this aspect and was only thinking about positive numbers when writing this. The `self` can, of course, be negative. 

But at least in this, a fix is simple and when we save `let positive = parts > 0;` we basically kill the sign of the `self` and can just use the `abs()` of `parts` which will fit into a u64 when divided by DIV.

I agree with the general comment about substrate as well, having some `Trait Unsigned` would be good but I am not sure how to make it in a generic way.

core/sr-primitives/src/sr_arithmetic.rs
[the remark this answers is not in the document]
kianenigma: Assertion added.

core/sr-primitives/src/sr_arithmetic.rs
thiolliere: about Unsigned trait we can do the same as num-traits, they have just https://docs.rs/num-traits/0.2.8/num_traits/sign/trait.Unsigned.html which is an empty trait

kianenigma:
@thiolliere I _think_ we're all good about all the comment.

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: I went as far as adding that (something almost the same) to `traits` but do we actually want that? In practice, someone can have a `struct CustomType` and implement all the requirements of `SimpleArithmetic` for an and use it in the balances module. Doing `pub trait Unsigned: Num {}` would actually block this and only allow these type to be used for balance etc.

srml/staking/src/lib.rs
```diff
@@ -1187,7 +1188,7 @@ impl<T: Trait> Module<T> {
 
 			for (v, p) in validators.iter().zip(points.individual.into_iter()) {
 				if p != 0 {
-					let reward = multiply_by_rational(total_payout, p, points.total);
+					let reward = Perbill::from_rational_approximation(p, points.total) * total_payout;
```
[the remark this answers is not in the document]
thiolliere: I'm not sure introducing Per-u32 would solve it what I had in mind was this:
```
p = 1, points.total = 3, total_payout = 3000).
with Per-thing it will do
3000*333_333_333/1_000_000_000 which will returns 999.
```

possibilities:
* maybe introducing rational(u32, u32) would solve this rounding error.
* maybe we can actually say that this is fine
* also we really care about this rounding error we could do rounding in the multiplication of per-thing instead of just a cut operation with something like this:
```diff
diff --git a/core/sr-primitives/src/lib.rs b/core/sr-primitives/src/lib.rs
index f6784c8f2..90019188e 100644
--- a/core/sr-primitives/src/lib.rs
+++ b/core/sr-primitives/src/lib.rs
@@ -213,7 +213,11 @@ where
 
                        // `rem_multiplied_u64` is less than 10^12 therefore divided by a million it fits into
                        // u32
-                       let rem_multiplied_divided_u32 = (rem_multiplied_u64 / 1_000_000) as u32;
+                       let mut rem_multiplied_divided_u32 = (rem_multiplied_u64 / 1_000_000) as u32;
+
+                       if rem_multiplied_u64 % 1_000_000 >= 1_000_000/2 {
+                               rem_multiplied_divided_u32 += 1;
+                       }
 
                        // `rem_multiplied_divided` is inferior to b, thus it can be converted back to N type
                        rem_multiplied_divided_u32.into()
```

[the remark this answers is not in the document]
kianenigma: I think in this case since the difference of 1billion and 4 billion is not that much, we can opt to `maybe we can actually say that this is fine`

I will also add your suggestion. sounds reasonable.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1103 @@
...
+				// q cannot be zero.
+				let q = q.max((1 as $type).into());
+				// p should not be bigger than q.
+				let p = p.min(q.clone());
+
+				let factor = (q.clone() / $max.into()).max((1 as $type).into());
+
+				// q cannot overflow: (q / (q/$max)) < 2 * $max. p < q hence p also cannot overflow.
+				// this implies that $type must be able to fit 2 * $max.
+				let q_reduce: $type = (q / factor.clone())
```
thiolliere: if max is u8::max_value(), type is u8 and upper type is u16 then factor is 39 and q/factor == 256 and doesn't fit into u8.

Maybe when playing with u8::max_value and edge stuff approximation can make inequalities wrong :-/

kianenigma: yes indeed better to saturate but I think maybe we can also just not permit it. Tbh I don't see any value in the current `Percent` which I implemented just for demonstration. For the first round, we can enforce `$type: From<u32>` (based on wasm being 32bit it also doesn't really bring us much in terms of memory or performance if we don arithmetic in 16 bit) and later refactor and harden if we actually need sth like `Percent`.

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: regarding use case: no, it is also used in the inner implementation of `Rational128` for addition etc. 

regarding providing something minimal that works rather than something flaky: I agree. I've changed the way that both functions work together now and they actually cover all the edge cases that we discussed here. I think it is good to be used for now. Regardless, what we were using before was horribly simple (naively trying to swap the operations and that was it), so this PR is a step forward anyhow. If we really need the multi-limb arithmetic, that is a bigger task and we can work on it in a follow-up PR.

srml/staking/src/lib.rs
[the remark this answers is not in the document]
kianenigma: applied

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: I just checked and:

- Things like percent can be actually created. Just had to change some tests. 
- The example you gave will not be accepted since we already discussed that for any per-thing, `$type` must be able to fit `2 * $max`. What you described is a `per_u8` with `$type = u8` which does not fit into this description.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1155 @@
...
+					// $max^2 and fits into $upper_type. This is assured by a test.
+					let rem_multiplied_upper = rem_sized as $upper_type * self.0 as $upper_type;
+
+					// `rem_multiplied_upper` is less than $max^2 therefore divided by $max it fits
+					// in $type. remember that $type always fits $max.
+					let mut rem_multiplied_divided_sized =
+						(rem_multiplied_upper / upper_max) as $type;
+					// fix a tiny rounding error
+					if rem_multiplied_upper % upper_max > upper_max / 2 {
+							rem_multiplied_divided_sized += 1;
```
thiolliere: indentation issue

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1155 @@
...
+	pub fn saturated_multiply_accumulate<N>(&self, int: N) -> N
+	where
+		N: TryFrom<u64> + From<u32> + UniqueSaturatedInto<u32> + Bounded + Clone + Saturating +
+		ops::Rem<N, Output=N> + ops::Div<N, Output=N> + ops::Mul<N, Output=N> +
+		ops::Add<N, Output=N>,
+	{
+		let div = DIV as u64;
+		let positive = self.0 > 0;
+		// safe to convert as absolute value.
+		let parts = self.0.checked_abs().unwrap_or(Bounded::max_value()) as u64;
```
thiolliere: if self.0 is i64::min_value then it will result in `9223372036854775807` instead of `9223372036854775808`
```suggestion
		let parts = self.0.checked_abs().map(|v| v as u64).unwrap_or(i64::max_value() as u64 + 1);
```

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1155 @@
...
+		if let Some(x) = a.checked_mul(b) {
+			// This is the safest way to go. Try it.
+			Ok(x / c)
+		} else if a > c && b <= c {
+			// if it can be safely swapped and it is a fraction, then swap. The operations in this
+			// block cannot overflow since b <= c.
+			let q = a / c;
+			let r = a % c;
+			let r_additional = (r * b) / c; // todo check overflow
+			Ok(q * b + r_additional) // todo check overflow
```
thiolliere: TODO ?

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: true, only can happen in `i64::MIN`. Your scrutiny is impressive 👍

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: no I just forgot to remove them. The comment above defines why it cannot overflow. Nonetheless, I've made the non-obvious ones saturating. the `(q * b)` is for sure safe because `q = a / c` and `b < c`.

This is actually the main difference now to compensate for the accuracy loss of this swapping method. Should be interesting for you to also see the test cases for this. The edge-case you reported earlier is now covered. (like `(multiply_by_rational(2*MAX64 - 1, MAX64, MAX64).unwrap()`)

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1155 @@
...
+
+		if let Some(x) = a.checked_mul(b) {
+			// This is the safest way to go. Try it.
+			Ok(x / c)
+		} else if a > c && b <= c {
+			// if it can be safely swapped and it is a fraction, then swap. The operations in this
+			// block cannot overflow since b <= c.
+			let q = a / c;
+			let r = a % c;
+			let r_additional = (r.saturating_mul(b)) / c;
```
thiolliere: I think this saturation can be an issue for instance for this call: `let res = multiply_by_rational(2u128.pow(66) - 1, 2u128.pow(65) - 1, 2u128.pow(65)).unwrap();`

see the error is about 50% of the result
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=87c04c91fbeed659c17d0f40abfe242d

kianenigma: yeah -- the statement `r * b / c;` is basically a recursive situation of the same function: the result is for sure less than `r` but we cannot compute in size... darn. 

Though, using this works perfectly. Potentially lossy and not fast but accurate. 

```suggestion
			let r_additional = multiply_by_rational_greedy(r, b, c);
```

note: the must_be in the gist you send is `73786976294838206461`.

kianenigma:
@thiolliere I've added some fuzzing code which clearly shows we can still be inaccurate (which is expected) when we are forced to collapse to `Perquintillion` as the ratio. I really don't think we can stretch this further with just u128 tricks in a good way and now my feeling is that it is better to get a multi-limb structure working, asap. Does not have to be super performant, just accurate. We worked on the mul and it should be fine. Getting a div working should be enough to cover the current use cases of the entire SRML.

In the meantime, as mentioned, I am fine with leaving this PR as a sort of _stall_ because it is was not a priority task and started as just a refactor. But, given the number of bugs that we found here, and the fact that current master phragmen is based on the buggy _swap `b` and `c`_ method, it is really good to get this in, with or without the limb stuff (having an inaccurate arithmetic is much better than an utterly wrong one), for Kusama thawning as the staking election will be unleashed there /cc @gavofyork

core/sr-primitives/src/sr_arithmetic.rs
thiolliere: oh yes.
Yes ok multiply_by_rational can be greedy now maybe we should only implement greedy version, is there a use for not greedy one ?

kianenigma: Only in `lcm()` and `to_den()` where we don't have the guarantee that `b < c` and we might overflow actually. in other phragmen use cases, we know that `b < c`. This is because the greedy one just returns `a` if the result is bigger than u128 and we never want this when computing lcm and such,

kianenigma: Other than that the non-greedy one is just a nice way to put some code away. The greedy always tries the normal one and if it fails starts the loop of accuracy reduction.

thiolliere:
multi-limb structure for div looks like https://github.com/rust-num/num-bigint/blob/089dcc0c1d22a51b8f3a031199e3b9f0c4134f39/src/algorithms.rs#L602

I'm happy with the current PR, maybe we can improve on rational even more.
Also we could think about introducing a real fuzzer like the one that is in parity-scale-codec.

kianenigma:
> multi-limb structure for div looks like https://github.com/rust-num/num-bigint/blob/089dcc0c1d22a51b8f3a031199e3b9f0c4134f39/src/algorithms.rs#L602
> 
> I'm happy with the current PR, maybe we can improve on rational even more.
> Also we could think about introducing a real fuzzer like the one that is in parity-scale-codec.

Agree with both. Could be done here or separate PR.

kianenigma:
Based on the looks that CC2 might be suddenly released as soon as BABE PRs are merged, I would vote for doing ^^ in a separate PR.

core/sr-primitives/src/sr_arithmetic.rs
thiolliere: well I think I'm still against having the non greedy calling greedy in one of its branch.

also maybe we could just remove this branch and use perquintill approximation.

kianenigma: > also maybe we could just remove this branch and use perquintill approximation.

As you pointed out in some examples, this is poorly inaccurate in many cases. The general approach should be to use the divide and multiply (swap) and if that fails, try perquintillion. Basically the second branch (swap) is much more accurate and preferable. For example

```rust
assert_eq!(
	Perquintill::from_rational_approximation(2u128, 3) * MAX128,
	multiply_by_rational(MAX128, 2u128, 3).unwrap(),
	"should have been = {}", mul_div(MAX128, 2, 3),
);
```

will fail with 

```bash
thread 'sr_arithmetic::test_rational128::multiply_by_rational_greedy_works' panicked at 'assertion failed: `(left == right)`
  left: `226854911280625642082061493673886498661`,
 right: `226854911280625642308916404954512140970`: should have been = 226854911280625642308916404954512140970'
```

kianenigma:
> well I think I'm still against having the non greedy calling greedy in one of its branch.

For the time being, the two functions are now separated in a more sensible manner:

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1304 @@
...
+	match ((a, b), (a & 1, b & 1)) {
+		((x, y), _) if x == y => y,
+		((0, x), _) | ((x, 0), _) => x,
+		((x, y), (0, 1)) | ((y, x), (1, 0)) => gcd(x >> 1, y),
+		((x, y), (0, 0)) => gcd(x >> 1, y >> 1) << 1,
+		((x, y), (1, 1)) => {
+			let (x, y) = (min(x, y), max(x, y));
+			gcd((y - x) >> 1, x)
+		},
+		_ => unreachable!(),
```
thiolliere: as a note converting a&1 to boolean would avoid this unreachable branch

node/runtime/src/lib.rs
```diff
@@ -83,7 +83,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 158,
```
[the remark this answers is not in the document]
kianenigma: Yeah I usually do this only after approval of each PR :)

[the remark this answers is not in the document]
thiolliere: this thing is that you only bump impl_version, but spec should be bump as well imo

core/sr-primitives/Cargo.toml
```diff
@@ -21,6 +21,8 @@ impl-trait-for-tuples = "0.1.1"
 [dev-dependencies]
 serde_json = "1.0"
 primitive-types = "0.5.0"
+srml_support = { package = "srml-support", path = "../../srml/support" }
```
thiolliere:
```suggestion
```

srml/staking/src/lib.rs
```diff
@@ -1190,7 +1191,7 @@ impl<T: Trait> Module<T> {
...
```
thiolliere: Oh yes so we have concluded that this approximation is fine.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -0,0 +1,1373 @@
...
+					match multiply_by_rational(a, shifted_b, shifted_c) {
+						Ok(r) => break r,
+						Err(_) => {
+							shift = shift + 1;
+							// by the time we reach here, b >= 1 && c >= 1. Before panic, they have
+							// to be zero which is prevented to happen by the break.
+							shifted_b = b.checked_shr(shift)
+								.expect(
+									"b >= 1 && c >= 1; break prevents them from ever being zero; \
+									panic can only happen after either us zero; qed"
```
thiolliere:
```suggestion
									panic can only happen after either is zero; qed"
```

core/phragmen/src/lib.rs
```diff
@@ -296,48 +294,64 @@ pub fn elect<AccountId, Balance, FS, C>(
...
-			// To ensure an assertion indicating: no stake from the voter going to waste, we add
-			//  a minimal post-processing to equally assign all of the leftover stake ratios.
-			let vote_count = assignment.1.len() as ExtendedBalance;
-			let l = assignment.1.len();
-			let sum = assignment.1.iter().map(|a| a.1).sum::<ExtendedBalance>();
-			let diff = ACCURACY.checked_sub(sum).unwrap_or(0);
-			let diff_per_vote= diff / vote_count;
+			// To ensure an assertion indicating: no stake from the nominator going to waste,
+			// we add a minimal post-processing to equally assign all of the leftover stake ratios.
+			let vote_count = assignment.1.len() as u32;
```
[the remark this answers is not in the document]
HCastano: I haven't been able to find much online about the implications of destructing, and then reconstructing a tuple. I want to say that the readability will beat the potential performance loss, but I have no good basis for that right now.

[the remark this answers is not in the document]
HCastano: From here: https://old.reddit.com/r/rust/comments/79ry4s/tuple_performance/

> For those who care, I stumbled upon the flawed benchmark, because I wondered whether using pattern matching for the function argument, i.e. (a, b): (Vec<T>, Vec<T>), is slower than binding the whole tuple and indexing into it, i.e. using x: (Vec<T>, Vec<T>) as the argument and x.0 and X.1 in the body. We do perform an extra copy in the pattern matching case, which seems useless in this example.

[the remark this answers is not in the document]
kianenigma: quite an interesting read. Yer, let's keep it for an optimisation PR since it needs verification + this code can use a LOT more optimisation here and there.

PR #3459, opened 2019-08-22
Clean runtime constants

kianenigma:
I realized ([thorough a copypasta](https://github.com/paritytech/substrate/pull/3364#discussion_r313800241)) that they are sitting on top of a few runtime modules with no purpose whatsoever. My guess is that the original author intended to do something like they have done in `finality-tracker`: declaring a few `pub const DEFAULT_XXX` in each module and aggregating in `node/runtime` if needed. Either way, removed since not used. We also have `const.rs` for this.

### DIFF
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -80,7 +80,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 149,
-	impl_version: 149,
+	impl_version: 150,
 	apis: RUNTIME_API_VERSIONS,
 };
 
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -48,7 +48,6 @@ use version::NativeVersion;
 use primitives::OpaqueMetadata;
 use grandpa::{AuthorityId as GrandpaId, AuthorityWeight as GrandpaWeight};
 use im_online::{AuthorityId as ImOnlineId};
-use finality_tracker::{DEFAULT_REPORT_LATENCY, DEFAULT_WINDOW_SIZE};
 
 #[cfg(any(feature = "std", test))]
 pub use sr_primitives::BuildStorage;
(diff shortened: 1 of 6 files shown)


kianenigma:
well, it looks like the beast is taking a nap 👀 .

bkchr:
Hmm? But these are default constants per module and `const.rs` is per runtime. How should any one with its own runtime use const.rs? (without copy-pasting)
We could actually improve the default values by making them parameter types directly in the module.

kianenigma:
> Hmm? But these are default constants per module and const.rs is per runtime. How should anyone with its own runtime use const.rs? (without copy-pasting)

Yeah, I see your point. can also be done that way. I would aggregate all in one place but I suppose it is a matter of preference? 

> We could actually improve the default values by making them parameter types directly in the module.

Indeed, but whatever the approach is, it is probably a bit more sophisticated than having `pub const DEFAULT_XXX` in each module. As you said, directly `paramtere_type!`-ing could be a good way. 

But all in all: 
I do not propose what _could be done to improve it_ in this PR. I merely removed what is _for sure wrong and redundant_ :p, hence pretty insubstantial.

bkchr:
When we have a constant in the module and the same value again in the runtime somewhere. Then the value from the runtime is redundant. Before everything was declared in the GenesisConfig and had it defaults value given in decl_storage. With parameter types that does not work anymore. So, having these constants in the module is the way to expose it to the peopel and make it accessible. When you want to build a chain with substrate, you probably do not care for these values at the beginning, you just want something running. So, having these sensitive default values in the module, makes the development easier from the outside.

kianenigma:
One problem with moving `parameter_types!` to modules would be that many of them --nay, almost all of them-- depend on types defined in the runtime's primitive file/module (most notably system types like Block number).

The only way I see forward is to say: each module defines one _default value_ __also__ with some _default type_  and a runtime file is free to either choose it or not. Doing so will more or less throw the `node/primitives` out of the window. Cool to do so?

kianenigma:
Discussed further with @bkchr and agreed to disagree.  Would either close it down or wait to see if someone else has an opinion?

gavofyork:
In reply to @bkchr 's comment on defaults, honestly, the old model of providing defaults was plain wrong and simply an artefact of storing what should be essentially fixed chain parameters in a mutable storage.

There should not be these hapless const values floating around as not-quite-code-not-quite-docs "defaults". If a developer is new to a module, then they should damn well read the documentation before deploying it to figure out sensible values. Template runtimes can provide integrations of multiple modules with working parameters that, for rapid development setups, will essentially serve as defaults. In reality these things often have no sensible, non-opinionated "default" that will automatically work sensibly regardless of how it is integrated.

bkchr:
Okay :)

PR #3461, opened 2019-08-22
Implement HTTP request in offchain workers

tomusdrw:
Supersedes #3447 
Since I couldn't push to @tomaka's branch, I'm reopening the PR with the last TODO fixed.
Repasting the description here as well.

Implements the currently-unimplemented API.

**I'm opening this PR as a draft because I'm going to be away for a bit, and I encourage someone to finish it if they want**.

The only thing that's not implemented is a small TODO to fix: we unwrap if we fail to initialize the `TlsConnector`. How is it even possible to fail to initialize TLS? I have absolutely no idea, and I think it's a platform-specific "just in case" thingy. But since the word `unwrap` is explicitly written in the code, I guess we have to handle this situation.
I'd like to fix that by having some `enum { HttpConnector, HttpsConnector }` where we fall back to the `HttpConnector` if the HTTPS one fails to initialize. We could also just return the error and print a big error saying "failed to initialize offchain workers". But considering that this is the only reason why offchain workers can fail to initialize, it's a bit annoying.

In order to implement this, I had to properly document/define how the API is to be used.

From the point of view of the offchain worker, the order in which functions must be called on an HTTP request is: initialization (`start`), adding headers, sending non-empty body, sending empty body, waiting for answer, reading headers, reading body until EOF.

Most steps can be skipped (for example, reading the body implicitly waits for the answer), but changing the order of the steps is generally not allowed. The exceptions are: you are still allowed to wait for answer or read response headers after having started to read the body, as long as EOF hasn't been returned.

As side changes, I also:
- Moved out some utility functions into a new `timestamp` module.
- Renamed `HttpRequestStatus::Unknown` to `Invalid` and `Timeout` to `IoError` (so that we also handle the situation where the remote force-closes the socket for example).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -202,7 +202,6 @@ check-web-wasm:
     - time cargo web build -p substrate-keystore
     - time cargo web build -p substrate-executor
     - time cargo web build -p substrate-network
-    - time cargo web build -p substrate-offchain
     - time cargo web build -p substrate-panic-handler
     - time cargo web build -p substrate-peerset
     - time cargo web build -p substrate-primitives
@@ -336,7 +335,7 @@ check_warnings:
     - docker push $CONTAINER_IMAGE:$VERSION
     - docker push $CONTAINER_IMAGE:latest
 
-publish-docker-substrate:          
+publish-docker-substrate:
   stage:                           publish
   <<:                              *publish-docker-release
   # collect VERSION artifact here to pass it on to kubernetes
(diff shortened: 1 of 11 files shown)


tomusdrw:
I've tested the behaviour with the new client, and `https` URLs are simply rejected by `HttpConnector` - so there is no way to run an `http` server on `:443` to still handle the request if `Tls` couldn't been initialized.

As a side note, I've realised that the requests are not following redirects currently. I think this is fine, but if we think that following redirects would be desirable, we can work on this in a follow up PR.

tomusdrw:
CC: @gautamdhameja

PR #3464, opened 2019-08-23
Remove MakePayment and introduce SignedExtension docs/examples

### LINKED ISSUE #3299: Outdated references to `MakePayment`
`MakePayment` trait is been removed but it still exists in a few comments.

https://github.com/paritytech/substrate/blob/89a56adaf8512ab55aabe4f1d9725e3a1baaba11/srml/contracts/src/lib.rs#L428

https://github.com/paritytech/substrate/blob/1fcd1286d87dc7acb8c2b0ee3ff6fe2199536ea0/srml/balances/src/lib.rs#L75

kianenigma:
Closes #3299 

Also adds some documentation about the replacement of `MakePayment`: `SignedExtensions` to the example module.

### DIFF
--- a/srml/balances/src/lib.rs
+++ b/srml/balances/src/lib.rs
@@ -72,8 +72,6 @@
 //! - [`Imbalance`](../srml_support/traits/trait.Imbalance.html): Functions for handling
 //! imbalances between total issuance in the system and account balances. Must be used when a function
 //! creates new funds (e.g. a reward) or destroys some funds (e.g. a system fee).
-//! - [`MakePayment`](../srml_support/traits/trait.MakePayment.html): Simple trait designed
-//! for hooking into a transaction payment.
 //! - [`IsDeadAccount`](../srml_system/trait.IsDeadAccount.html): Determiner to say whether a
 //! given account is unused.
 //!
@@ -87,6 +85,17 @@
 //! ### Public Functions
 //!
 //! - `vesting_balance` - Get the amount that is currently being vested and cannot be transferred out of this account.
+//!
+//! ### Signed Extensions
+//!
+//! The balances module defines the following extensions:
+//!
+//!   - [`TakeFees`]: Consumes fees proportional to the length and weight of the transaction.
+//!     Additionally, it can contain a single encoded payload as a `tip`. The inclusion priority
+//!     is increased proportional to the tip.
+//!
+//! Lookup the runtime aggregator file (e.g. `node/runtime`) to see the full list of signed
+//! extensions included in a chain.
 //!
 //! ## Usage
 //!
--- a/srml/example/src/lib.rs
+++ b/srml/example/src/lib.rs
@@ -258,7 +258,7 @@ use srml_support::{StorageValue, dispatch::Result, decl_module, decl_storage, de
 use system::{ensure_signed, ensure_root};
 use codec::{Encode, Decode};
 use sr_primitives::{
-	traits::{SignedExtension, DispatchError, Bounded },
+	traits::{SignedExtension, DispatchError, Bounded},
 	transaction_validity::ValidTransaction,
 	weights::{SimpleDispatchInfo, DispatchInfo},
 };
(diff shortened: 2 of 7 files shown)


kianenigma:
[clabot:check]

srml/balances/src/lib.rs
```diff
@@ -88,6 +86,17 @@
...
 //!
+//! ### Signed Extensions
+//!
+//! The balances module defines the following extensions:
+//!
+//!   - [`TakeFees`]: Consumes fees proportional to the length and weight of the transaction.
+//!     Additionally, it can contain a single encoded payload as a `tip`. The inclusion priority
+//!     is increased proportional to the tip.
+//!
+//! Lookup the runtime aggregator file (e.g. `node/runtime`) to see the full list of signed
```
bkchr: I don't see any reason to write this sentence?

srml/system/src/lib.rs
```diff
@@ -42,6 +42,23 @@
...
+//!     exceed the limits.
+//!   - ['CheckNonce']: Checks the nonce of the transaction. Contains a single payload of type
+//!     `T::Index`.
+//!   - [`CheckEra`]: Checks the era of the transaction. Contains a single payload of type `Era`.
+//!   - [`CheckGenesis`]: Checks the provided genesis hash of the transaction. Must be a part of the
+//!     signed payload of the transaction.
+//!   - [`CheckVersion`]: Checks that the runtime version is the same as the one encoded in the
+//!     transaction.
+//!
+//! Lookup the runtime aggregator file (e.g. `node/runtime`) to see the full list of signed
```
bkchr: Again

srml/example/src/lib.rs
```diff
@@ -500,15 +506,105 @@ impl<T: Trait> Module<T> {
...
+// Using the extension, you can add some hooks to the lifecycle of each transaction. Note that by
+// default, an extension is applied to all `Call` functions (i.e. all transactions). the `Call` enum
+// variant is given to each function of `SignedExtension`. Hence, you can filter based on module or
+// a particular call if needed.
+//
+// Some extra information, such as encoded length, some static dispatch info like weight and the
+// sender of the transaction (if signed) are also provided.
+//
+// The main hooks that you can add are:
+//   - `pre_dispatch`.
```
bkchr: Just link the trait, otherwise any function you describe here could be already non existent.

srml/example/src/lib.rs
```diff
@@ -500,15 +506,105 @@ impl<T: Trait> Module<T> {
...
+//   - `pre_dispatch_unsigned`.
+//   - `validate_unsigned`.
+//   - `post_dispatch`.
+//
+// The signed extensions are aggregated in the runtime file of a substrate chain. All extensions
+// should be aggregated in a tuple and passed to the `CheckedExtrinsic` and `UncheckedExtrinsic`
+// types defined in the runtime. Lookup `pub type SignedExtra = (...)` in `node/runtime` and
+// `node-template` for an example of this.
+//
+// > `SignedExtension` is a very recent addition to substrate at the time of this writing and might
```
bkchr: This could be written for any part of substrate. So, I think we don't need to mention it here.

kianenigma: fair enough 👍

srml/balances/src/lib.rs
kianenigma: I have a strong incentive to mention it as much as I can: so that the user would not think that just defining this struct and implementing the trait is enough and we, like it is magic, take care of the rest. In reality, doing ^^ does exactly _nothing_. Once included in the runtime's `UncheckedExtrinsic` type then it starts doing stuff :D

Similar to modules that get aggregated somewhere, extensions also get combined at some point in the runtime file and I find it an important note to point out. Hence, let's keep such sentences.

bkchr: Why not write this? You write someone should look into the runtime to see all signed extensions. You should write that the user needs to add them to the extrinsics type to be used.

PR #3465, opened 2019-08-23
decl_storage metadata: use spin instead of once_cell and make use of it in no_std

thiolliere:
previously the metadata was cached into a Once structure only on std by the use of the crate once_cell (which doesn't work on no_std).

Now it uses spin crate and implements the caching for both std and no_std.

Spin crate is used by lazy_static for no_std. (note: lazy_static doesn't use spin for std).

EDIT: Maybe there was a reason not to cache in no_std, but I'm not aware of it

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4178,11 +4178,11 @@ name = "srml-support"
 version = "2.0.0"
 dependencies = [
  "bitmask 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


bkchr:
If we cache in wasm, it will probably be thrown away by the next execution. The memory instance is reset before each call into wasm.
I think that we just should integrate a cache for the Metadata in the node. So, that we don't even need to call into the runtime to get the value.

PR #3468, opened 2019-08-23
Fix linkemap swap

thiolliere:
### Context:

As for now default implementation of swap in srml-support/storage/hashed/generator breaks linked_map implementation of its linked list.

I'm not sure what are the most idiomatic rust but it feels to me in the future we should remove this default implementation of storage map as it is confusing with other implementation (such as linkedmap) which break assumptions of this default.

for me either we divide into two different trait or we make no default implementation.

But before this happens here is a fix in the current design.

### Done

default swap method in storagemap is override in the implementation for linkedmap made in decl_storage.

### Comments

* The changes or remains of order of the linked_list for a swap is not specified in the API. No guarantee must be made by user. maybe this should deserve a comment but at some point nothing has been stated in the whole API in this regards so I feel fine with this.

### DIFF
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -81,7 +81,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 149,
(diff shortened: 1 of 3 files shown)


thiolliere:
Fix https://github.com/paritytech/substrate/issues/3341

PR #3469, opened 2019-08-23
Fixed digest in execution proof check

svyatonik:
At some point  we started to use parent header' digest in `check_execution_proof()`, instead of empty digest (that is used when proof is built). This PR fixes that + modifies existing test to catch this type of errors.


svyatonik:
Integration tests error is unrelated - panic in libp2p - search by "panic" [here](https://gitlab.parity.io/parity/substrate/-/jobs/216625/raw).

PR #3472, opened 2019-08-23
GRANDPA links set IDs to sessions.

rphmeier:
This clarifies the invariants around the `changed` parameter passed to session handlers, such that it is _only_ and _always_ true when at least one of the session keys has changed _or_ the economic conditions behind it have shifted (i.e. staking: weights, and thus slashable balance, have changed).

For already-deployed networks, this may need to be coupled with a governance call to update the storage of `grandpa::CurrentSetId` to match reality. Although, since grandpa previously didn't dispatch any set changes for the same keys as before (even when underlying economic changes had occurred), it most likely won't be an issue for Kusama.


srml/grandpa/src/lib.rs
```diff
@@ -151,6 +152,12 @@ decl_storage! {
 
 		/// `true` if we are currently stalled.
 		Stalled get(stalled): Option<(T::BlockNumber, T::BlockNumber)>;
+
+		/// The current set-id.
+		CurrentSetId get(current_set_id) build(|_| fg_primitives::genesis_set_id()): SetId;
+
+		/// A mapping from grandpa set ID to session index.
```
gavofyork:
```suggestion
		/// A mapping from grandpa set ID to the index of the *most recent* session for which its members were responsible.
```

srml/grandpa/src/lib.rs
```diff
@@ -151,6 +152,12 @@ decl_storage! {
 
 		/// `true` if we are currently stalled.
 		Stalled get(stalled): Option<(T::BlockNumber, T::BlockNumber)>;
+
+		/// The current set-id.
```
gavofyork: Useless docs.

srml/grandpa/src/lib.rs
```diff
@@ -151,6 +152,12 @@ decl_storage! {
...
+		CurrentSetId get(current_set_id) build(|_| fg_primitives::genesis_set_id()): SetId;
```
gavofyork: Why placing a function which returns 0? Won't it always be zero? Under what conditions would this not be initialised to zero?

core/finality-grandpa/primitives/src/lib.rs
```diff
@@ -52,6 +52,15 @@ pub type AuthorityWeight = u64;
 /// The index of an authority.
 pub type AuthorityIndex = u64;
 
+/// The identifier of a GRANDPA set.
+pub type SetId = u64;
+
+/// The genesis set ID.
+pub fn genesis_set_id() -> SetId { 0 }
```
gavofyork: If it really can't just be `SetId::default()`, then consider using a constant.

srml/grandpa/src/lib.rs
rphmeier: It's always going to be 0 for the moment (although later on, chains might base it off of the genesis hash / fork ID). Regardless, I'd prefer not to scatter the knowledge and justification that it's 0 too far and wide.

srml/grandpa/src/lib.rs
```diff
@@ -380,17 +394,24 @@ impl<T: Trait> session::OneSessionHandler<T::AccountId> for Module<T> {
...
+			} else {
+				let _ = Self::schedule_change(next_authorities, Zero::zero(), None);
 			}
+			let new_set_id = CurrentSetId::mutate(|s| { *s += 1; *s });
+			SetIdSession::insert(new_set_id, &session_index);
+		} else {
+			// nothing's changed, neither economic conditions nor session keys. update the pointer
+			// of the current set.
+			let set_id = Self::current_set_id();
+			SetIdSession::insert(set_id, &session_index);
```
marcio-diaz: Maybe you can get rid of some `if else` here, the code looks symmetric.

srml/grandpa/src/lib.rs
```diff
@@ -151,6 +152,13 @@ decl_storage! {
 
 		/// `true` if we are currently stalled.
 		Stalled get(stalled): Option<(T::BlockNumber, T::BlockNumber)>;
+
+		/// The number of changes (both in terms of keys and underlying economic responsibilities)
+		/// in the "set" of Grandpa validators from genesis.
+		CurrentSetId get(current_set_id) build(|_| fg_primitives::SetId::default()): SetId;
```
gavofyork:
```suggestion
		CurrentSetId get(current_set_id): SetId;
```

^^^ this is strictly equivalent.

srml/grandpa/src/lib.rs
```diff
@@ -380,18 +395,27 @@ impl<T: Trait> session::OneSessionHandler<T::AccountId> for Module<T> {
...
-				} else {
-					let _ = Self::schedule_change(next_authorities, Zero::zero(), None);
-				}
+			if let Some((further_wait, median)) = <Stalled<T>>::take() {
+				let _ = Self::schedule_change(next_authorities, further_wait, Some(median));
+			} else {
+				let _ = Self::schedule_change(next_authorities, Zero::zero(), None);
 			}
-		}
+			CurrentSetId::mutate(|s| { *s += 1; *s })
```
andresilva: I think should only be done when we actually apply the change. But in practice it only matters for forced changes as normal changes will be applied immediately.

PR #3473, opened 2019-08-23
Consensus: Proof of Work

sorpaas:
This PR implements PoW consensus engine in Substrate. We currently have a PoW verifier, and a simple thread-based miner. The PoW algorithm and difficulty adjustment algorithm is generic over `PowAlgorithm` struct. The caller is expected to implement one of this that suits its need.

For mining, we currently just use a simple runtime function `PowAlgorithm::mine`. It takes a header pre-hash, and is expected to loop until it finds a seal that satisfies the current difficulty.

Things that can be improved, but probably not in this PR:

- RPC methods for `getWork` and `submitWork`.


core/consensus/pow/primitives/src/lib.rs
```diff
@@ -0,0 +1,29 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
rphmeier: missing license header

core/consensus/pow/primitives/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use substrate_client::decl_runtime_apis;
+use rstd::vec::Vec;
+use sr_primitives::ConsensusEngineId;
+use primitives::H256;
+
+/// The `ConsensusEngineId` of PoW.
+pub const POW_ENGINE_ID: ConsensusEngineId = [b'p', b'o', b'w', b'_'];
```
rphmeier: PoW is a class of algorithms. Consider rename to the actual PoW algorithm name.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,302 @@
+use std::sync::{Mutex, Arc};
```
rphmeier: missing license header

core/consensus/pow/primitives/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+pub type Seal = Vec<u8>;
+
+decl_runtime_apis! {
+	/// API necessary for block authorship with Proof of Work.
+	pub trait PowApi {
+		/// Verify a given proof of work against the current difficulty.
+		/// Note that `pre_hash` is always a hash of a direct child.
+		///
+		/// Returns the current difficulty.
+		fn verify(pre_hash: &H256, seal: &Seal) -> Option<Difficulty>;
```
rphmeier: I see - instead of using a particular algorithm we are delegating all of it to the runtime. Is it a pure function? This will definitely complicate things for light clients (and parallel verification queues, which are a kind of light client), if the expectation is that it will be called at the parent's runtime. If that's not the expectation, then how do you do algorithm changes?

core/consensus/pow/primitives/src/lib.rs
sorpaas: The current implementation is designed to be generic. We only define the absolute minimal things needed to carry out PoW fork choice in consensus engine. Using this, the runtime should be able to implement any class of PoW algorithms.

In other words, there's nothing we can rename to other than `GenericPow`. :)

rphmeier:
You can use a pre-runtime digest for the coinbase. A runtime API to fill it out using the runtime's keystore API would work perfectly for that.

core/consensus/pow/primitives/src/lib.rs
sorpaas: > This will definitely complicate things for light clients (and parallel verification queues, which are a kind of light client), if the expectation is that it will be called at the parent's runtime.

Yes. The expectation is that it will be called at the parent's runtime. This is a "pure function" in the sense that it will not modify anything in the parent runtime. However, usually it needs to take in parent storage information for it's difficulty adjustment algorithm (parent block's difficulty).

rphmeier: TBH, I think it isn't usable for anything except the most basic circumstances to take such an approach. It makes light clients and warp sync much more difficult and less space-efficient.

sorpaas: TBH I don't think we can create a light client implementation that fits all consensus engines, so I just assumed that we temporarily don't consider light client here.

rphmeier: > I don't think we can create one light client implementation that fits all consensus engines

I agree - pluggable consensus is not designed to do that. But you'll have a hard time writing a PoW light client with this design. That's unfortunate, since PoW should be one of the easiest things to write a light client for (as it is not intertwined with the state)

sorpaas: > But you'll have a hard time writing a PoW light client with this design. That's unfortunate, since PoW should be one of the easiest things to write a light client for (as consensus state is not intertwined with the state machine)

Consider, if we design a light client for this PoW, can you explain more why you think it would be hard? If we just add a digest item of the hash of the current runtime code, the light client can then make assumptions about the PoW algorithm. This then is effectively the same as defining separate consensus engines for PoW and use markers to change them on the fly.

sorpaas: The particular use case I have in mind is that a lot of PoW blockchains like to "tweak" their algorithm once in a while. If we define separate engines for each PoW algorithm, then there'll never be an end that we can support all of them -- it will just be too many. Besides, any tweak in the algorithm will require "hard fork" (in the sense that one must update the Substrate version running). This kind of goes against the goals of Substrate's "online upgrade".

rphmeier: Yes, I understand the goal.

> If we just add a digest item of the hash of the current runtime code, the light client can then make assumptions about the PoW algorithm

Right - either this, or having the runtime include some other WASM code for PoW verification (where light clients can download this module when it changes, and the full node could execute it using the sandbox utilities). We also need to write some code for full nodes to ensure that the hash actually matches the runtime state at all times.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+
+	Ok(BasicQueue::new(
+		verifier,
+		block_import,
+		None,
+		None
+	))
+}
+
+pub fn start_mine<B: BlockT<Hash=H256>, I, C, E>(
```
tomaka: That should return a `Future`. Almost all of Substrate uses futures and not background threads.
You can do that by doing something like `futures_timer::Interval::new().for_each(|_| mine_one())`.

sorpaas: The issue is, mine_one is not a yieldable operation at all. Any thread running it basically causes that thread to spin at 100% for all time, and have no chance to process anything else. Putting this under a future, I think, will pollute the future pool. It may *appear* fine in a threadpool future runtime, but if the future runtime is single threaded, it breaks all other running futures.

I do think we can put the initial block builder under a future, but when doing the actual mining, I still think we need to use a background thread.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+		header,
+		justification: None,
+		post_digests: vec![DigestItem::Seal(POW_ENGINE_ID, seal)],
+		body: Some(body),
+		finalized: false,
+		auxiliary: vec![(key, Some(aux.encode()))],
+		fork_choice: ForkChoiceStrategy::Custom(true),
+	};
+
+	if let Err(err) = block_import.lock().expect("Lock failed").import_block(import_block, Default::default()) {
```
pepyakin: This should pass a real proof or handle the failure somehow.

core/consensus/pow/primitives/src/lib.rs
pepyakin: Can we use this API for implementing something like ethash? If so, then I guess the type of difficulty should be able to fit `H256`, I guess, right?

core/consensus/pow/primitives/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+/// Type of seal.
+pub type Seal = Vec<u8>;
+
+decl_runtime_apis! {
+	/// API necessary for block authorship with Proof of Work.
+	pub trait PowApi {
+		/// Verify a given proof of work against the current difficulty.
+		/// Note that `pre_hash` is always a hash of a direct child.
+		///
+		/// Returns the current difficulty.
```
pepyakin: What if the value returned is `None`?

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+		{
+			Some(bytes) => PowAux::decode(&mut &bytes[..]).map_err(|e| format!("{:?}", e))?,
+			None => Default::default(),
+		};
+		let mut aux = match self.client.get_aux(&parent_aux_key)
+			.map_err(|e| format!("{:?}", e))?
+		{
+			Some(bytes) => PowAux::decode(&mut &bytes[..]).map_err(|e| format!("{:?}", e))?,
+			None => Default::default(),
+		};
```
pepyakin: Can the code querying `PowAux` be extracted to an auxilary (ba-dum-tss.jpg) function? Or code for generating the key?

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+
+		let (checked_header, difficulty, seal) = self.check_header::<B>(
+			header,
+			BlockId::Hash(parent_hash),
+		)?;
+		aux.total_difficulty += difficulty;
```
pepyakin: I'd propose to turn this into a checked addition. This might require to do more work for proofing or documenting, since this is possible to overflow in certain (even though rare) circumstances.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+		justification: None,
+		post_digests: vec![DigestItem::Seal(POW_ENGINE_ID, seal)],
+		body: Some(body),
+		finalized: false,
+		auxiliary: vec![(key, Some(aux.encode()))],
+		fork_choice: ForkChoiceStrategy::Custom(true),
+	};
+
+	if let Err(err) = block_import.lock().expect("Lock failed").import_block(import_block, Default::default()) {
+		warn!("Error with block built on {:?}: {:?}",
```
pepyakin: I am not sure, but shouldn't this be an `error!` level? The fact that we cannot import a block built by ourselves seems to me quite severe.

core/consensus/pow/src/lib.rs
pepyakin: If we decided to leave this as is, we should leave a note in the comment on saying that this is rather an exception, so people won't take this as an example way of doing this (or at least they understood the tradeoffs)

tomaka: If runtime calls ever become asynchronous, we can make `mine_one` yieldable. But right now, you're correct that a separate thread might be a good idea.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+				Err(e) => warn!("Mining block failed: {:?}", e),
+			}
+
+			std::thread::sleep(std::time::Duration::new(1, 0));
+		}
+	});
+}
+
+pub fn mine_one<B: BlockT<Hash=H256>, I, C, E>(
+	block_import: &Arc<Mutex<I>>,
```
tomaka: Related to the threading question: I'm reaaally not a fan of having `Mutex`es, and I'm generally trying to remove `Mutexes` from the codebase entirely (cc #2536).

Wouldn't it be preferable for `mine_one` to return some sort of "sealed block" object, and do the importing somewhere else?
I'm also generally trying to un-spaghettify the general design of Substrate, and passing the import queue by parameter to a background thread is really not great IMO.

core/consensus/pow/src/lib.rs
sorpaas: We no longer needed the Mutex around `BlockImport`. So I think this is resolved!

core/consensus/pow/primitives/src/lib.rs
sorpaas: The current difficulty type is `u128`, which I *think* should be more than enough. We should be able to use this to fit nearly all mining algorithms (including ethash). The current difficulty on Ethereum is around 1000x less than `u64::max_value()`. The current total difficulty of Ethereum is more than `u64::max_value()`, but it's way less than `u128::max_value()`, and I think it will never have a chance to exceed it in any reasonable timespan.

Practically using either `u128` or `U256` should be fine. But m concern for changing this to `U256` is that it's less nice compared with `u128`, which is a native type so easier to use.

core/consensus/pow/primitives/src/lib.rs
sorpaas: If the value returned is `None`, it means the seal is invalid. I'll add docs.

core/consensus/pow/src/lib.rs
sorpaas: I changed this to use `saturating_add`. I think this is slightly nicer than `checked_add` as we make sure the chain continue to function whenever it could (for example, if it has a secondary finality method).

core/consensus/pow/src/lib.rs
sorpaas: Changed this to error out in `mine_loop`. The current behavior is that if the loop fails, it prints a message, wait one second before mining again.

core/consensus/pow/src/lib.rs
sorpaas: I figured out that this `Mutex` is not necessary at all! We can use `BoxBlockImport` because `Client` already handles locking for its `BlockImport` implementation.

core/consensus/pow/primitives/src/lib.rs
pepyakin: The argument about `u64` w.r.t to ethash is quite convicing for me and I of course deter to your discretion about other PoW algos. 

But it would be great if you added this rationale to the docs!

rphmeier: > The particular use case I have in mind is that a lot of PoW blockchains like to "tweak" their algorithm once in a while

Thought about this a bit more. How do you reason about fork-choice? What happens when a chain switches from longest-chain PoW to GHOST-based or vice-versa? What about GHOST-based to GHOST-based? What about something 

> If we define separate engines for each PoW algorithm, then there'll never be an end that we can support all of them -- it will just be too many. 

It makes some sense to define a skeleton consensus crate that is instantiatable by a `T: PoWAlgorithm`.

sorpaas: > Thought about this a bit more. How do you reason about fork-choice? What happens when a chain switches from longest-chain PoW to GHOST-based or vice-versa? What about GHOST-based to GHOST-based? What about something

Then they will *have to* switch to a different consensus engine. Here's assumptions we make for the current engine:

* Definition of "difficulty" and "timestamp".
* Each block has an associated seal and difficulty.
* Chain selection chooses highest total difficulty.
* We make no assumption how runtime defines difficulty or validates its seal.

> It makes some sense to define a skeleton consensus crate that is instantiatable by a T: PoWAlgorithm.

Anyway, I think we can just support both -- light client "friendly" version where algorithms are defined outside of runtime, and light client "unfriendly" version where they are defined within runtime by using this `T: PowAlgorithm` generalization. I think that'll just resolve this once and for all. xD

rphmeier: How do you envision light-clients working, where the PoW algorithm can change on any block? Since there isn't any delay in the PoW algorithm changing, all a bad miner needs to do is give light clients 1 block and then they can be tricked into following that chain forever (by giving it a difficulty that grows faster than the legitimate chain).

I don't think it's a good idea to give people the option to deploy chains that will _never be safe for light clients_ without a strong disclaimer. Like naming the crate `full_clients_only_somewhat_dynamic_pow`.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+use inherents::{InherentDataProviders, InherentData};
+use consensus_common::{
+	BlockImportParams, BlockOrigin, ForkChoiceStrategy,
+	well_known_cache_keys::Id as CacheKeyId, Environment, Proposer,
+};
+use consensus_common::import_queue::{BoxBlockImport, BasicQueue, Verifier};
+use codec::{Encode, Decode};
+use log::*;
+
+/// Auxiliary prefix for PoW engine.
```
pepyakin: aux storage, right?

```suggestion
/// Auxiliary storage prefix for PoW engine.
```

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+		self.client.runtime_api().difficulty(parent)
+			.map_err(|e| format!("Fetch difficulty failed: {:?}", e))
+	}
+
+	fn verify(
+		&self,
+		parent: &BlockId<B>,
+		pre_hash: &H256,
+		seal: &Seal,
+		difficulty: Difficulty
```
pepyakin:
```suggestion
		difficulty: Difficulty,
```

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+			.map_err(|e| format!("Verify PoW seal failed: {:?}", e))
+	}
+
+	fn mine(
+		&self,
+		parent: &BlockId<B>,
+		pre_hash: &H256,
+		seed: &H256,
+		difficulty: Difficulty,
+		round: u32
```
pepyakin:
```suggestion
		round: u32,
```

srml/pow/src/algos/blake2_256.rs
```diff
@@ -0,0 +1,78 @@
...
+	pre_hash: &H256,
+	seed: &H256,
+	difficulty: Difficulty,
+	round: u32,
+	proportion: Difficulty,
+) -> Option<RawSeal> {
+	for i in 0..round {
+		let nonce = {
+			let mut ret = H256::default();
+			(U256::from(&seed[..]) + U256::from(i)).to_big_endian(&mut ret[..]);
```
pepyakin: Can we switch this to little endian, since substrate generally leans towards LE?

srml/pow/src/diffs/average_span.rs
```diff
@@ -0,0 +1,71 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use srml_support::{StorageValue, decl_storage, decl_module};
+use srml_support::traits::Get;
+use pow_primitives::Difficulty;
+use sr_primitives::traits::{UniqueSaturatedInto, Zero};
+
+pub trait Trait: timestamp::Trait {
+	type Span: Get<Self::BlockNumber>;
```
pepyakin: Can we leave a comment that this must be > 0?

srml/pow/src/diffs/average_span.rs
```diff
@@ -0,0 +1,71 @@
...
+	type TargetPeriod: Get<Self::Moment>;
+	type InitialDifficulty: Get<Difficulty>;
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as AverageSpanDifficultyAdjustment {
+		LastTimestamp get(last_timestamp): Option<T::Moment>;
```
pepyakin: I guess this is more the timestamp of the last span? If so, can we either leave a comment or rename appropriately?

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+	client: Arc<C>,
+	algorithm: Algorithm,
+	inherent_data_providers: inherents::InherentDataProviders,
+}
+
+impl<C, Algorithm> PowVerifier<C, Algorithm> {
+	fn check_header<B: BlockT<Hash=H256>>(
+		&self,
+		mut header: B::Header,
+		block_id: BlockId<B>,
```
pepyakin: is this actually the parent block id?

core/consensus/pow/primitives/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+	/// API necessary for block authorship with Proof of Work.
+	///
+	/// This API allows online upgrades of PoW algorithms as well as the difficulty
+	/// adjustment algorithm. Note that we always validate difficulty and seals
+	/// against the parent block. This means that in the case of code upgrade, the
+	/// first block that upgraded the code will still use the old algorithm. The block
+	/// after that will switch to the new algorithm.
+	///
+	/// All functions here should not modify block state. Instead, state modifications
+	/// related to PoW should use `on_initialize` and `on_finalize` routine.
```
pepyakin: 👍 !

core/consensus/pow/primitives/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+			seal: &Seal,
+			difficulty: Difficulty
+		) -> bool;
+
+		/// Mine a seal that satisfy the given difficulty.
+		fn mine(
+			pre_hash: &H256,
+			seed: &H256,
+			difficulty: Difficulty,
+			round: u32
```
pepyakin: What is `round` and `seed`?

pepyakin:
```suggestion
			round: u32,
```

core/consensus/pow/primitives/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+	pub trait PowApi {
+		/// Return the difficulty that should be applied for the next block.
+		fn difficulty() -> Difficulty;
+
+		/// Verify a given proof of work against the given difficulty.
+		/// Note that `pre_hash` is always a hash of a direct child.
+		fn verify(
+			pre_hash: &H256,
+			seal: &Seal,
+			difficulty: Difficulty
```
pepyakin:
```suggestion
			difficulty: Difficulty,
```

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+	well_known_cache_keys::Id as CacheKeyId, Environment, Proposer,
+};
+use consensus_common::import_queue::{BoxBlockImport, BasicQueue, Verifier};
+use codec::{Encode, Decode};
+use log::*;
+
+/// Auxiliary prefix for PoW engine.
+pub const POW_AUX_PREFIX: [u8; 4] = *b"PoW:";
+
+/// Auxiliary data for PoW.
```
pepyakin: Again does it refer to auxilary storage? If so, I'd like to directly mention this in the docs since the term "auxilary" is ambiguous and "auxilary storage" is not?

core/consensus/pow/primitives/src/lib.rs
rphmeier: Actually, you can extend this to full nodes too. Let's say you could have 2 forks which each pass a different runtime upgrade via a governance process. Runtime upgrade A does a minor tweak of the process, and upgrade B does a big change, that causes difficulty to skyrocket afterwards. Fork B is done in secret by a cartel of malicious miners, who push through a unanimous coin-vote to pass upgrade B. They then mine on top of it a bit more and release fork B to the network. Now fork B will be considered higher total difficulty than fork A.

They could also bundle any other changes in the runtime upgrade that they wanted.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+pub trait PowAlgorithm<B: BlockT> {
+	/// Get the next block's difficulty.
+	fn difficulty(&self, parent: &BlockId<B>) -> Result<Difficulty, String>;
+	/// Verify proof of work against the given difficulty.
+	fn verify(
+		&self,
+		parent: &BlockId<B>,
+		pre_hash: &H256,
+		seal: &Seal,
+		difficulty: Difficulty
```
pepyakin:
```suggestion
		difficulty: Difficulty,
```

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+	) -> Result<bool, String>;
+	/// Mine a seal that satisfy the given difficulty.
+	fn mine(
+		&self,
+		parent: &BlockId<B>,
+		pre_hash: &H256,
+		seed: &H256,
+		difficulty: Difficulty,
+		round: u32
```
pepyakin:
```suggestion
		round: u32,
```

core/consensus/pow/src/lib.rs
sorpaas: Fixed!

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+	C: HeaderBackend<B> + AuxStore,
+	Algorithm: PowAlgorithm<B>,
+	E: Environment<B>,
+	E::Error: core::fmt::Debug,
+{
+	'outer: loop {
+		let best_hash = client.info().best_hash;
+		let best_header = client.header(BlockId::Hash(best_hash))
+			.map_err(|e| format!("Best header does not exist: {:?}", e))?
+			.ok_or("Fetching best header failed")?;
```
pepyakin: Shouldn't it be the other way around? Like, `map_err`'s message should say that fetching the best header failed and unwrapping option with `ok_or` should say that `best header doesn't exist`?

core/consensus/pow/primitives/src/lib.rs
rphmeier: (you don't even need fork A to do a runtime upgrade; it's irrelevant. all that matters it that a small minority of miners can change things in a secret fork and then overwhelm the honest fork. that's not the security model of PoW)

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+	mut env: E,
+	preruntime: Vec<u8>,
+	round: u32,
+	inherent_data_providers: inherents::InherentDataProviders,
+) where
+	C: HeaderBackend<B> + AuxStore + ProvideRuntimeApi + 'static,
+	C::Api: PowApi<B>,
+	Algorithm: PowAlgorithm<B> + Send + Sync + 'static,
+	E: Environment<B> + Send + Sync + 'static,
+	E::Error: core::fmt::Debug,
```
pepyakin: Why not std?

```suggestion
	E::Error: std::fmt::Debug,
```

sorpaas:
I'm pulling this out of draft. There may still be many documentations or testing to be added, but in general I think it's ready for review (and merge).

There's a new abstraction added `PowAlgorithm`. One can use `PowRuntimeAlgorithm` to delegate the algorithm to runtime, or define a standalone implementation out of runtime. Hopefully this addresses some of the concerns of possible light client integration. For `PowRuntimeAlgorithm`, I still think we *should be able to* create a light client implementation for it. I'll try to draft a design for that and create an issue, but that'll be out of scope of this PR.

`PowApi` interface is slightly changed, to decouple `difficulty` and `verify`. The reason for this is that I find it simplify (a little bit) when using `srml-pow`.

srml/pow/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+/// Find author routine for PoW engine.
+pub struct FindAuthor<V: Codec>(PhantomData<V>);
+
+impl<V: Codec> srml_support::traits::FindAuthor<V> for FindAuthor<V> {
+	fn find_author<'a, I>(digests: I) -> Option<V> where
+		I: 'a + IntoIterator<Item=(ConsensusEngineId, &'a [u8])>
+	{
+		for (engine_id, mut value) in digests {
+			if engine_id == POW_ENGINE_ID {
+				return V::decode(&mut value).ok()
```
pepyakin: Can you elaborate a bit how would it work? Where exactly the author is/would be put into a seal?

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+	algorithm: &Algorithm,
+	env: &mut E,
+	preruntime: &[u8],
+	round: u32,
+	inherent_data_providers: &inherents::InherentDataProviders,
+) -> Result<(), String> where
+	C: HeaderBackend<B> + AuxStore,
+	Algorithm: PowAlgorithm<B>,
+	E: Environment<B>,
+	E::Error: core::fmt::Debug,
```
pepyakin: std here as well?

```suggestion
	E::Error: std::fmt::Debug,
```

core/consensus/pow/primitives/src/lib.rs
sorpaas: > Runtime upgrade A does a minor tweak of the process, and upgrade B does a big change, that causes difficulty to skyrocket afterwards. Fork B is done in secret by a cartel of malicious miners, who push through a unanimous coin-vote to pass upgrade B. They then mine on top of it a bit more and release fork B to the network. Now fork B will be considered higher total difficulty than fork A.

I don't think this is any worse than PoS process. To actually do this, the attacker would need to trick the governance process so that its fork B somehow passed. This is rather a vulnerability in onchain goverance, than PoW engine.

Anyway, I wouldn't mind if we don't support `PowRuntimeAlgorithm` and always stick to fixed PoW algorithm for a chain. Maybe we just remove it?

srml/pow/src/lib.rs
rphmeier: Although my knee-jerk reaction was to  say "use a pre-runtime digest for the coinbase", I actually think that there's a strong case for only putting it in the body instead via an inherent as opposed to paying for the extra size in the header-chain. Who mined the block is not relevant to whether PoW was done correctly.

So we should do it with a standard `Author` inherent -- if inherent generation has access to the runtime keystore facilities, it should be straightforward to set it correctly. I don't know if those APIs are available there, though.

srml/pow/src/algos/blake2_256.rs
sorpaas: I used this because `U256::from(&seed[..])` reads the value as big endian, so that I think it maybe better to also just convert it back in BE.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+	fn check_header<B: BlockT<Hash=H256>>(
+		&self,
+		mut header: B::Header,
+		block_id: BlockId<B>,
+	) -> Result<(B::Header, Difficulty, DigestItem<H256>), String> where
+		Algorithm: PowAlgorithm<B>,
+	{
+		let hash = header.hash();
+
+		let (seal, inner_seal) = match header.digest_mut().pop() {
```
pepyakin: Dumb question: what guarantees that the pow seal is the last item?

core/consensus/pow/primitives/src/lib.rs
sorpaas: `seed` is a random 256-bit value passed to initiate mining. `round` defines how many calculations we try before giving up, and return `None`.

srml/pow/src/algos/blake2_256.rs
pepyakin: Ah, yeah, I think it is even better since no byte shuffling is taking place, right?

core/consensus/pow/primitives/src/lib.rs
rphmeier: > Anyway, I wouldn't mind if we don't support PowRuntimeAlgorithm and always stick to fixed PoW algorithm for a chain. Maybe we just remove it?

I recommend it as the safest option for now

I'm happy to revisit the runtime-based algorithm later on, but it appears to need more research about unification of fork-choice rules.

> I don't think this is any worse than PoS process. To actually do this, the attacker would need to trick the governance process so that its fork B somehow passed. This is rather a vulnerability in onchain goverance, than PoW engine.

That argument can be made, but it is in fact the case with SRML governance (you can pass anything with unanimity in `srml-democracy`, and a secret fork can contain only "for" transactions.

It's not the same situation as BABE/GRANDPA since you have finality that prevents reorganization, although there might be a similar vector on BABE that I'd like to confirm before posting here.

sorpaas: Hmm, I’m actually having a feeling that the issue may be worse. Even with finality, if last finalised block is old enough than current chain head, this attack can still be carried out, right? Because a minority fork can arbitrarily change the validator set (and thus fork choice) in a short period of time (if epoch is less than voting period).

sorpaas: I removed `PowRuntimeApi` in `745d604b02b19ef98aa89ca488941d12b299bd53` temporarily. I think the actual problem lies deeper than in the runtime PoW engine, and we may want to explore to see whether we can do something about it. From my understanding, it looks like the governance/democracy module may be *unsafe* for any consensus engine that requires runtime data. Two potential ways I see moving forward:

* Keep it strictly that consensus engine cannot access runtime data. This is the same way as Serenity -- change consensus using hard fork, change runtime execution using WebAssembly.
* Find some ways to make governance module safe for those consensus engines.

sorpaas: I created a new issue discussing swappable PoW (https://github.com/paritytech/substrate/issues/3517). As we mostly reached conclusion that we're just not touching that subject on this PR, going to mark this as resolved.

core/consensus/pow/src/lib.rs
sorpaas: @DemiMarie-parity the algorithm is defined as a trait and you can implement whatever you want, be it CPU-friendly, GPU-friendly or ASIC-friendly.

srml/pow/src/algos/blake2_256.rs
sorpaas: `U256` still has a few byte shuffling actually -- we need to convert from `u8` to `u64`, and those values are stored as little-endian. I think that's an issue we might want to address in `U256` maybe some day.

sorpaas: Realized that this code no longer exists so I'm going to mark it as resolved!

core/consensus/pow/src/lib.rs
sorpaas: You're right. Fixed!

core/consensus/pow/src/lib.rs
sorpaas: Fixed!

core/consensus/pow/src/lib.rs
sorpaas: Fixed!

srml/pow/src/lib.rs
sorpaas: We can't really have a srml module for pow currently. Maybe this should instead belong to `srml-authorship`?

core/consensus/pow/src/lib.rs
sorpaas: I think we actually guarantee that it's the last item for pow engine. We only pushes seal after all the rest of the block finished building. It also must be the last block (and if not we should reject it as invalid), because otherwise, the prehash we got will be incorrect.

DemiMarie-parity: @sorpass That is correct.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+					return Err(format!("Header uses the wrong engine {:?}", id))
+				}
+			},
+			_ => return Err(format!("Header {:?} is unsealed", hash)),
+		};
+
+		let pre_hash = header.hash();
+		let difficulty = self.algorithm.difficulty(&parent_block_id)?;
+
+		if !self.algorithm.verify(
```
DemiMarie-parity: Where are the signatures checked?  This is not obvious from the code, and it should be.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+		let inherent_res = self.client.runtime_api().check_inherents(
+			&block_id,
+			block,
+			inherent_data,
+		).map_err(|e| format!("{:?}", e))?;
+
+		if !inherent_res.ok() {
+			inherent_res
+				.into_errors()
+				.try_for_each(|(i, e)| match TIError::try_from(&i, &e) {
```
DemiMarie-parity: This should be merged with the corresponding code in AuRa and BABE.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+		verifier,
+		block_import,
+		None,
+		None
+	))
+}
+
+/// Start the background mining thread for PoW. Note that because PoW mining
+/// is CPU-intensive, it is not possible to use async future to define this.
+/// However, it's not recommended to use background thread in the rest of the
```
DemiMarie-parity:
```suggestion
/// However, it's not recommended to use background threads in the rest of the
```

That said, most PoW mining is offloaded to dedicated hardware, either GPUs or ASICs.  Substrate will interact with such miners using asynchronous RPC.

core/consensus/pow/src/lib.rs
sorpaas: There’s no signature check in PoW engine.

burdges:
As an aside, we should make doing parachains with proof-of-stake block production relatively easy and discourage parachains from actually using proof-of-work.  It's fine for testing of course but maybe bad optics if we have real parachains doing this.

core/consensus/pow/src/lib.rs
sorpaas: @DemiMarie-parity I checked and unfortunately there's not an easy way to merge this code. While the code structure is similar, the function is actually quite different:

* In Aura, when we find a block with timestamp too far in the future, we suspend import and wait till that block becomes valid. This seems fine because all blocks are signed by semi-trusted authorities.
* In PoW, however, we must not suspend import, but return the appropriate error, because blocks are only sealed, not signed, and they cannot be trusted until it's verified.
* In Babe, there's no check related to timestamp.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+			Some(bytes) => PowAux::decode(&mut &bytes[..]).map_err(|e| format!("{:?}", e)),
+			None => Ok(Default::default()),
+		}
+	}
+}
+
+/// Algorithm used for proof of work.
+pub trait PowAlgorithm<B: BlockT> {
+	/// Get the next block's difficulty.
+	fn difficulty(&self, parent: &BlockId<B>) -> Result<Difficulty, String>;
```
cloudhead: How would this work for eg. Nakamoto PoW where block difficulty depends on the last *2016* blocks?

sorpaas: The struct that implements `PowAlgorithm` is expected to take a reference to client `Arc<C>`. That will give it access to all past blocks, where it can find out their difficulty through auxiliary store. This way we can implement Nakamoto PoW. :)

cloudhead: Great, thanks!

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+	Ok(BasicQueue::new(
+		verifier,
+		block_import,
+		None,
+		None
+	))
+}
+
+/// Start the background mining thread for PoW. Note that because PoW mining
+/// is CPU-intensive, it is not possible to use async future to define this.
```
DemiMarie-parity:
```suggestion
/// is CPU-intensive, it is not possible to use an async future to define this.
```

core/consensus/pow/src/lib.rs
0x7CFE: But what about light clients? Is aux store available for them? I thought that they have access to headers only.

sorpaas: As far as I understand light client should have access to aux store. Even if they don't, because we only need to store tiny amount of metadata in aux store, it's also quite easy to move that elsewhere.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Proof of work consensus for Substrate.
+//!
+//! To use this engine, you can either implement a standalone `PowAlgorithm`
+//! or use the runtime PoW algorithm engine. The runtime engine requires you
+//! to implement the `PowApi` defined in `substrate-consensus-pow-primitives`.
```
0x7CFE: I believe this is no longer relevant.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+	well_known_cache_keys::Id as CacheKeyId, Environment, Proposer,
+};
+use consensus_common::import_queue::{BoxBlockImport, BasicQueue, Verifier};
+use codec::{Encode, Decode};
+use log::*;
+
+/// Auxiliary storage prefix for PoW engine.
+pub const POW_AUX_PREFIX: [u8; 4] = *b"PoW:";
+
+fn aux_key(hash: &H256) -> Vec<u8> {
```
0x7CFE: Better have a comment explaining the purpose of this function.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+
+/// A verifier for PoW blocks.
+pub struct PowVerifier<C, Algorithm> {
+	client: Arc<C>,
+	algorithm: Algorithm,
+	inherent_data_providers: inherents::InherentDataProviders,
+}
+
+impl<C, Algorithm> PowVerifier<C, Algorithm> {
+	fn check_header<B: BlockT<Hash=H256>>(
```
0x7CFE: Here we pinning a `Hash` to be `H256`. What if someone wants something else?

core/consensus/pow/src/lib.rs
0x7CFE: I think this info should be in a comment above the `let` binding.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+			block,
+			inherent_data,
+		).map_err(|e| format!("{:?}", e))?;
+
+		if !inherent_res.ok() {
+			inherent_res
+				.into_errors()
+				.try_for_each(|(i, e)| match TIError::try_from(&i, &e) {
+					Some(TIError::ValidAtTimestamp(timestamp)) => {
+						if timestamp > timestamp_now + MAX_TIMESTAMP_DRIFT_SECS {
```
0x7CFE: Could this overflow?

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+					e
+				),
+			}
+
+			std::thread::sleep(std::time::Duration::new(1, 0));
+		}
+	});
+}
+
+fn mine_loop<B: BlockT<Hash=H256>, C, Algorithm, E>(
```
0x7CFE: I do not see how cancellation is handled in this design. We have several cases where we want to cancel mining early:
- A new best block had been received and we should restart mining on top of it
- The system is shutting down

core/consensus/pow/src/lib.rs
sorpaas: Fixed!

core/consensus/pow/src/lib.rs
sorpaas: It's not possible. We pin `Hash` to be `H256` in many places of Substrate. I think this is a valid tradeoff for code complexity.

core/consensus/pow/src/lib.rs
sorpaas: The `timestamp_now` passed is always created from fresh timestamp inherent data providers, and thus should always be valid. Thus I think it's not possible to overflow.

core/consensus/pow/src/lib.rs
sorpaas: Cancellation is handled like this: for `PowAlgorithm::mine`, it takes a parameter called `round`. The CPU miner is then expected to only mine that particular "rounds" as defined by the algorithm. When calling `start_mine`, it is expected that the caller passes an initial `round` that results in subsecond execution of `PowAlgorithm::mine`. If seal is found after rounds, then we import the block. If no seal is found, then `mine` returns `None`. The mining function then loops, checks whether best block has changed. If not, directly call `PowAlgorithm::mine` again. Otherwise, build a new block and mining on that instead.

TBH I'm hoping to keep `Algorithm` as the generic type name. `B`, `C`, `E` all has standard meanings in other consensus engines, but `Algorithm` is one specific to PoW engine. Keeping it like this just makes things more descriptive, and it doesn't actually result in long code anyway.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+			header,
+			justification: None,
+			post_digests: vec![DigestItem::Seal(POW_ENGINE_ID, seal)],
+			body: Some(body),
+			finalized: false,
+			auxiliary: vec![(key, Some(aux.encode()))],
+			fork_choice: ForkChoiceStrategy::Custom(true),
+		};
+
+		block_import.import_block(import_block, Default::default())
```
0x7CFE: Instead of `Default` it's better to write actual type that impls this trait. That way, the reader will be able to follow the logic without having to look at `import_block` prototype.

core/consensus/pow/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+/// is CPU-intensive, it is not possible to use an async future to define this.
+/// However, it's not recommended to use background threads in the rest of the
+/// codebase.
+pub fn start_mine<B: BlockT<Hash=H256>, C, Algorithm, E>(
+	mut block_import: BoxBlockImport<B>,
+	client: Arc<C>,
+	algorithm: Algorithm,
+	mut env: E,
+	preruntime: Vec<u8>,
+	round: u32,
```
0x7CFE: It's not very clear for the reader what `preruntime` and `round` fields do. Better comment things out.

shawntabrizi:
@sorpaas Do you have time to work with someone from the docs team to help write some information about this?

How to use it, how it works, etc...

Would be looking to you to answer questions for whomever writes the docs.

sorpaas:
@shawntabrizi Yeah for sure!

JoshOrndorff:
@sorpaas I've taken the lead on consensus on the docs team. Could we set up a call this week? I'll do my homework first. I'm @joshyorndorff:matrix.parity.io on riot. Tuesday or Wednesday would be fine. I'm UTC-8.

IPSE-TEAM:
> This PR implements PoW consensus engine in Substrate. We currently have a PoW verifier, and a simple thread-based miner. The PoW algorithm and difficulty adjustment algorithm is generic over `PowAlgorithm` struct. The caller is expected to implement one of this that suits its need.
> 
> For mining, we currently just use a simple runtime function `PowAlgorithm::mine`. It takes a header pre-hash, and is expected to loop until it finds a seal that satisfies the current difficulty.
> 
> Things that can be improved, but probably not in this PR:
> 
> * RPC methods for `getWork` and `submitWork`.

come on, How to implement this RPC?

PR #3476, opened 2019-08-25
add Copy to Moment type

xlc:
Saves some typing and I don't see way it can't have `Copy`


bkchr:
Looks good so far, but could you also clean up the module? (remove calls to clone)

PR #3477, opened 2019-08-25
add worksapce to node-template Cargo.toml

xlc:
so that `cargo check --tests --all` will actually check everything


xlc:
this seems causing more issues. maybe should be done as part of node-template-release script?

PR #3481, opened 2019-08-26
Fix typo in readme

joepetrowski:
Missed a typo ("SEEED") in a review on Saturday, fixing here.


joepetrowski:
> SEEED - German hip hop band - https://en.wikipedia.org/wiki/Seeed :D

Feature request: `SEEED` RPC that returns a word that rhymes :)

PR #3483, opened 2019-08-26
Re-connect when telemetry is unresponsive during dialing

### LINKED ISSUE #3276: Telemetry worker doesn't detect if server isn't responsive
If the telemetry server closes the connection (ie. sends a FIN) then Substrate will try to reconnect as it should.

However if the connection simply isn't responsive (no ACK is sent back) then Substrate doesn't care.

gterzian:
FIX #3276

### DIFF
--- a/core/telemetry/src/worker/node.rs
+++ b/core/telemetry/src/worker/node.rs
@@ -127,7 +127,7 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
 							break NodeSocket::Connected(conn)
 						},
 						Poll::Pending => {
-							if is_writing_data {
+							if is_writing_data && self.connection_timeout.is_none() {
 								self.connection_timeout = Some(Delay::new(Duration::from_millis(500)));
 							}
 							break NodeSocket::Connected(conn)
@@ -184,10 +184,12 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
 				Poll::Pending => {},
 				Poll::Ready(Err(err)) => {
 					warn!(target: "telemetry", "Connection timeout error for {} {:?}", self.addr, err);
+					self.connection_timeout = None;
 				}
 				Poll::Ready(Ok(_)) => {
 					println!("Reconnecting");
 					warn!(target: "telemetry", "Server unresponsive from {}", self.addr);
+					self.connection_timeout = None;
 					let timeout = gen_rand_reconnect_delay();
 					self.socket = NodeSocket::WaitingReconnect(timeout);
 				}
--- a/core/telemetry/src/worker/node.rs
+++ b/core/telemetry/src/worker/node.rs
@@ -116,7 +116,16 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
 		let mut socket = mem::replace(&mut self.socket, NodeSocket::Poisoned);
 		self.socket = loop {
 			match socket {
-				NodeSocket::Connected(mut conn) =>
+				NodeSocket::Connected(mut conn) => {
+					let mut connection_timeout = Delay::new(Duration::from_millis(500));
+					match Future::poll(Pin::new(&mut connection_timeout), cx) {
+						Poll::Pending => {},
+						Poll::Ready(Err(_)) | Poll::Ready(Ok(_)) => {
+							warn!(target: "telemetry", "Server unresponsive from {}", self.addr);
+							let timeout = gen_rand_reconnect_delay();
+							break NodeSocket::WaitingReconnect(timeout)
+						}
+					}
 					match NodeSocketConnected::poll(Pin::new(&mut conn), cx, &self.addr) {
 						Poll::Ready(Ok(v)) => match v {}
 						Poll::Pending => break NodeSocket::Connected(conn),
@@ -127,6 +136,7 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
 							return Poll::Ready(NodeEvent::Disconnected(err))
 						}
 					}
+				}
 				NodeSocket::Dialing(mut s) => match Future::poll(Pin::new(&mut s), cx) {
 					Poll::Ready(Ok(sink)) => {
 						debug!(target: "telemetry", "Connected to {}", self.addr);
--- a/core/telemetry/src/worker/node.rs
+++ b/core/telemetry/src/worker/node.rs
@@ -120,15 +120,19 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
 		self.socket = loop {
 			match socket {
 				NodeSocket::Connected(mut conn) => {
-					let mut is_writing_data = false;
+					let mut is_writing_data = self.connection_timeout.is_some();
 					match NodeSocketConnected::poll(Pin::new(&mut conn), cx, &self.addr, &mut is_writing_data) {
 						Poll::Ready(Ok(_)) => {
 							self.connection_timeout = None;
 							break NodeSocket::Connected(conn)
 						},
 						Poll::Pending => {
-							if is_writing_data && self.connection_timeout.is_none() {
-								self.connection_timeout = Some(Delay::new(Duration::from_millis(500)));
+							if is_writing_data {
+								if self.connection_timeout.is_none() {
+									self.connection_timeout = Some(Delay::new(Duration::from_millis(5000)));
+								}
+							} else {
+								self.connection_timeout = None;
 							}
 							break NodeSocket::Connected(conn)
 						},
@@ -187,7 +191,6 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
 					self.connection_timeout = None;
 				}
 				Poll::Ready(Ok(_)) => {
-					println!("Reconnecting");
 					warn!(target: "telemetry", "Server unresponsive from {}", self.addr);
 					self.connection_timeout = None;
 					let timeout = gen_rand_reconnect_delay();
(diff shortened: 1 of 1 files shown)


gterzian:
Not sure if this is what you had in mind, so please take a look @tomaka 

What would be the correct way to test this by the way?

tomaka:
> Not sure if this is what you had in mind, so please take a look @tomaka

No, we already have a timeout for connecting:

https://github.com/paritytech/substrate/blob/c63ee7ad8fef5aba30930e56b82e282e2f14d1d5/core/telemetry/src/worker.rs#L112

What I mean is: once a connection is open, we don't detect if the telemetry server doesn't send back ACKs.

The way we should do that is adding a timeout when sending data, so:
- Here: https://github.com/paritytech/substrate/blob/c63ee7ad8fef5aba30930e56b82e282e2f14d1d5/core/telemetry/src/worker/node.rs#L212
- And here: https://github.com/paritytech/substrate/blob/c63ee7ad8fef5aba30930e56b82e282e2f14d1d5/core/telemetry/src/worker/node.rs#L196

Once async/await are stable, it will be very easy to add timeouts. In the meanwhile, thought, I don't know how clean it's possible to do that.
Maybe using [`MaybeDone`](https://docs.rs/futures-preview/0.3.0-alpha.17/futures/future/enum.MaybeDone.html) can make it easier to implement, I'm not sure.

bkchr:
Can we not just add a timer that is reset each time we receive an ack? If the ack is not received in the expected time frame, we close the connection? (This assumes we send regular ping messages)

tomaka:
We can't detect ACKs and we don't ping the connection. A timeout when writing is an indirect way of doing that.

bkchr:
Yeah okay, then a timer should still be do-able. Just reset the timer each time we can send something. If we don't send data, we will timeout.

gterzian:
@tomoka what do you want to do with pending messages when we timeout?

bkchr:
Just throw them away. This are just telemetry messages. But as we don't further poll the channel, we drop the messages automatically.

gterzian:
Ok here is another try. Any ideas on how to test this?

core/telemetry/src/worker/node.rs
```diff
@@ -114,19 +118,35 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
...
 		self.socket = loop {
 			match socket {
-				NodeSocket::Connected(mut conn) =>
+				NodeSocket::Connected(mut conn) => {
+					self.connection_start = Some(Instant::now());
 					match NodeSocketConnected::poll(Pin::new(&mut conn), cx, &self.addr) {
 						Poll::Ready(Ok(v)) => match v {}
-						Poll::Pending => break NodeSocket::Connected(conn),
+						Poll::Pending => {
+							let timed_out = match self.connection_start {
```
bkchr: You should use https://docs.rs/futures-timer/0.3.0/futures_timer/struct.Delay.html

THis will wake up the task when we hit the timeout.

gterzian: Thanks

gterzian:
Ok, this may be more like it.

tomaka:
This will never resolve since you're recreating a new `Delay` every time we poll.
There should be an `Option<Delay>` stored in the struct, that gets set to something (if necessary) whenever we start sending, and reset whenever we successfully send.

gterzian:
@tomaka Ok, this might be better, testing locally...

tomaka:
I think this will trigger a timeout if we don't send any message to the telemetry for more than 500ms.
Really what we need is the timeout when sending data and flushing.

gterzian:
@tomaka Ok I think this might be better. One thing I noticed is that I get a `Poll::Ready(Ok(()))` from the call to `Sink::poll_flush` even if I turn off the wifi. Any ideas on how to test this?

tomaka:
Flushing doesn't necessarily do anything. If everything has already been sent out, I expect that flushing simply returns `Ok()` even if in reality you're not connected anymore.

gterzian:
@tomaka Ok, ready for review then.

gavofyork:
@maciejhirsz is this code useful?

gavofyork:
@tomaka the original issue is yours, so it's your to review :-P

core/telemetry/src/worker/node.rs
```diff
@@ -36,6 +36,8 @@ pub struct Node<TTrans: Transport> {
 	socket: NodeSocket<TTrans>,
 	/// Transport used to establish new connections.
 	transport: TTrans,
+    /// Timeout for a connection to the node that is writing data.
+    connection_timeout: Option<Delay>,
```
tomaka: Should be indented with tabs.

core/telemetry/src/worker/node.rs
```diff
@@ -189,8 +223,9 @@ where TTrans::Output: Sink<BytesMut, Error = TSinkErr>
 	fn poll(
 		mut self: Pin<&mut Self>,
 		cx: &mut Context,
-		my_addr: &Multiaddr
-	) -> Poll<Result<futures::never::Never, TSinkErr>> {
+		my_addr: &Multiaddr,
+		is_writing_data: &mut bool,
+	) -> Poll<Result<(), TSinkErr>> {
```
tomaka: Why remove the `futures::never::Never`?

core/telemetry/src/worker/node.rs
```diff
@@ -76,6 +78,7 @@ impl<TTrans: Transport> Node<TTrans> {
 			addr,
 			socket: NodeSocket::ReconnectNow,
 			transport,
+			connection_timeout: None,
```
tomaka: Why is this in `NodeSocket` and not in `NodeSocketConnected`? It is only ever user or relevant when we're connected.

tomaka:
The behaviour looks good to me, but I'd like `connection_timeout` to be in `NodeSocketConnected`.

gterzian:
@tomaka A question in order to move the logic to inside `NodeSocketConnected`: when the timeout fires, and the socket is still writing data, presumably the right thing to do would be to return an error, so that the outer loop would reconnect. How can I manually return the right error type for `TSinkErr`?

All the other code just forward errors otherwise, and I can't quite make out how to manually create such an error.

tomaka:
> the right thing to do would be to return an error, so that the outer loop would reconnect.

Yes!

> How can I manually return the right error type for TSinkErr?

The right thing to do would be to create a custom enum, of either "timeout error" or "sink error", and use it as the return type instead of `TSinkErr`, similar to this:

https://github.com/paritytech/jsonrpsee/blob/076122d9d6e053c33bcdb24b785cb0ba79c854cc/ws/src/client.rs#L79-L108

core/telemetry/src/worker/node.rs
```diff
@@ -199,8 +225,23 @@ where TTrans::Output: Sink<BytesMut, Error = TSinkErr>
 				}
 
 				let item_len = item.len();
+				self.is_writing_data = true;
+				let mut timeout = Delay::new(Duration::from_millis(5000));
```
tomaka: I don't know if this PR is ready for review, but the `timeout` has to be in the struct, otherwise you're creating and destroying it immediately after.
`is_writing_data: bool` can probably be replaced with `timeout: Option<Delay>`.

gterzian: Thanks, it's ready for review now. I'm still wondering how to properly test this, as turning off the wifi doesn't seem to hit the timeout, for example.

core/telemetry/src/worker/node.rs
```diff
@@ -116,21 +127,36 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
 		let mut socket = mem::replace(&mut self.socket, NodeSocket::Poisoned);
 		self.socket = loop {
 			match socket {
-				NodeSocket::Connected(mut conn) =>
+				NodeSocket::Connected(mut conn) => {
 					match NodeSocketConnected::poll(Pin::new(&mut conn), cx, &self.addr) {
-						Poll::Ready(Ok(v)) => match v {}
-						Poll::Pending => break NodeSocket::Connected(conn),
-						Poll::Ready(Err(err)) => {
+						Poll::Ready(Ok(_)) => {
```
bkchr:
```suggestion
						Poll::Ready(Ok(_)) | Poll::Pending => {
```

core/telemetry/src/worker/node.rs
```diff
@@ -116,21 +127,36 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
...
+							break NodeSocket::Connected(conn)
+						},
+						Poll::Pending => {
```
bkchr:
```suggestion
```

core/telemetry/src/worker/node.rs
```diff
@@ -116,21 +127,36 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
...
+							break NodeSocket::Connected(conn)
```
bkchr:
```suggestion
```

core/telemetry/src/worker/node.rs
```diff
@@ -116,21 +127,36 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
...
+						},
```
bkchr:
```suggestion
```

core/telemetry/src/worker/node.rs
```diff
@@ -199,8 +225,22 @@ where TTrans::Output: Sink<BytesMut, Error = TSinkErr>
...
 				let item_len = item.len();
+				self.timeout = Some(Delay::new(Duration::from_millis(5000)));
+				if let Some(mut timeout) = self.timeout.as_mut() {
+					match Future::poll(Pin::new(&mut timeout), cx) {
+						Poll::Pending => {},
+						Poll::Ready(Err(err)) => {
+							warn!(target: "telemetry", "Connection timeout error for {} {:?}", my_addr, err);
+						}
+						Poll::Ready(Ok(_)) => {
+							return Poll::Ready(Err(ConnectionError::Timeout))
```
bkchr: This will never happen. You always create a new timeout before this `if`.

core/telemetry/src/worker/node.rs
```diff
@@ -189,18 +213,35 @@ where TTrans::Output: Sink<BytesMut, Error = TSinkErr>
...
+				}
+			}
+		}
 		loop {
 			if let Some(item) = self.pending.pop_front() {
 				if let Poll::Pending = Sink::poll_ready(Pin::new(&mut self.sink), cx) {
 					self.pending.push_front(item);
+					self.timeout = Some(Delay::new(Duration::from_millis(5000)));
+					if let Some(mut timeout) = self.timeout.as_mut() {
+						let _ = Future::poll(Pin::new(&mut timeout), cx);
```
gterzian: What about polling here, after setting it to `Some(_)`, and then again at the top of `poll` if it is still `Some(_)`?

core/telemetry/src/worker/node.rs
tomaka: Why this suggestion? We extract the socket then put it back. If you remove this `break` we stay in a `Poisoned` state.

core/telemetry/src/worker/node.rs
```diff
@@ -116,27 +127,40 @@ where TTrans: Clone + Unpin, TTrans::Dial: Unpin,
 		let mut socket = mem::replace(&mut self.socket, NodeSocket::Poisoned);
 		self.socket = loop {
 			match socket {
-				NodeSocket::Connected(mut conn) =>
+				NodeSocket::Connected(mut conn) => {
 					match NodeSocketConnected::poll(Pin::new(&mut conn), cx, &self.addr) {
-						Poll::Ready(Ok(v)) => match v {}
-						Poll::Pending => break NodeSocket::Connected(conn),
-						Poll::Ready(Err(err)) => {
+						Poll::Ready(Ok(_)) | Poll::Pending => {
```
tomaka:
```suggestion
						Poll::Ready(Ok(v)) => match v {}
						Poll::Pending => {
```

core/telemetry/src/worker/node.rs
bkchr: I did not looked at the original code close enough.

I wanted to have this `Poll::Ready(Ok(_)) | Poll::Pending => { break Connected() }`.

bkchr: (You can not have multiline suggestions :( )

gavofyork:
@tomaka worth continuing to review this?

tomaka:
It's still quite far away from being perfect in terms of logic.
To be fair, this issue is way more complex than it looks.

I opened #3916 instead.

PR #3487, opened 2019-08-26
Make sure that `on_before_session_ending` is called

bkchr:


srml/session/src/lib.rs
```diff
@@ -490,6 +490,9 @@ impl<T: Trait> Module<T> {
 
 		let applied_at = session_index + 2;
 
+		// Inform the session handlers that a session is going to end.
+		T::SessionHandler::on_before_session_ending();
```
tomusdrw: Should we call this before `Validators<T>` is populated with new validators? I'd assume that the state still describes the previous session.

PR #3491, opened 2019-08-27
Fixed uncle pruning

arkpar:
I would also suggest uncle pruning to be moved from `on_initalize` to `set_uncles`

### DIFF
--- a/srml/authorship/src/lib.rs
+++ b/srml/authorship/src/lib.rs
@@ -27,7 +27,7 @@ use srml_support::traits::{FindAuthor, VerifySeal, Get};
 use srml_support::dispatch::Result as DispatchResult;
 use codec::{Encode, Decode};
 use system::ensure_none;
-use sr_primitives::traits::{SimpleArithmetic, Header as HeaderT, One, Zero};
+use sr_primitives::traits::{Header as HeaderT, One, Zero};
 use sr_primitives::weights::SimpleDispatchInfo;
 use inherents::{
 	RuntimeString, InherentIdentifier, ProvideInherent,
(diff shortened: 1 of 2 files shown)


srml/authorship/src/lib.rs
```diff
@@ -387,6 +372,18 @@ impl<T: Trait> Module<T> {
 		// check uncle validity.
 		T::FilterUncle::filter_uncle(&uncle, accumulator)
 	}
+
+	fn prune_old_uncles(minimum_height: T::BlockNumber) {
+		let mut uncles = <Self as Store>::Uncles::get();
```
tomusdrw: We don't have a way to get an iterator without allocating the entire thing, right?

bkchr: No

kianenigma: Shouldn't we though (in the ling future probably)? 

sounds like a good enhancement: sth that only decodes elements of collection one by one in an iterator-like fashion and yields them.

bkchr: What is the ling future? Some Chinese future? :D

My brain already started working on this task as @tomusdrw mentioned it :P

kianenigma: > What is the ling future? Some Chinese future? :D

it comes from the same origin as _fail enough_ if you know what I am talking about 😁

> My brain already started working on this task as @tomusdrw mentioned it :P

Sweet.

rphmeier: Theoretically, the storage APIs could let you choose which position to read from and you could read into a buffer. But at the end of the day you have to load the entire trie leaf, and that'll be a big one. Automatically splitting a vector over many storage entries and having convenient APIs for access would be useful

PR #3494, opened 2019-08-27
Eradicate native_equivalent

pepyakin:
This PR consist of 3 changes separated by commits:

1. In substrate-test-client we can now add arbitrary storage entries (only main storage though),
2. Use that functionality for creating custom storage entry for testing the RPC storage behavior,
3. Since that leaves no usages of it in the codebase, retire `native_equivalent`.

### DIFF
--- a/core/executor/src/native_executor.rs
+++ b/core/executor/src/native_executor.rs
@@ -49,9 +49,6 @@ pub fn with_native_environment<F, U>(ext: &mut dyn Externalities<Blake2Hasher>, 
 
 /// Delegate for dispatching a CodeExecutor call to native code.
 pub trait NativeExecutionDispatch: Send + Sync {
-	/// Get the wasm code that the native dispatch will be equivalent to.
-	fn native_equivalent() -> &'static [u8];
-
 	/// Dispatch a method and input data to be executed natively.
 	fn dispatch(ext: &mut dyn Externalities<Blake2Hasher>, method: &str, data: &[u8]) -> Result<Vec<u8>>;
 
@@ -211,19 +208,13 @@ impl<D: NativeExecutionDispatch> CodeExecutor<Blake2Hasher> for NativeExecutor<D
 /// Implements a `NativeExecutionDispatch` for provided parameters.
 #[macro_export]
 macro_rules! native_executor_instance {
-	( $pub:vis $name:ident, $dispatcher:path, $version:path, $code:expr) => {
+	( $pub:vis $name:ident, $dispatcher:path, $version:path) => {
 		/// A unit struct which implements `NativeExecutionDispatch` feeding in the hard-coded runtime.
 		$pub struct $name;
-		$crate::native_executor_instance!(IMPL $name, $dispatcher, $version, $code);
+		$crate::native_executor_instance!(IMPL $name, $dispatcher, $version);
 	};
-	(IMPL $name:ident, $dispatcher:path, $version:path, $code:expr) => {
+	(IMPL $name:ident, $dispatcher:path, $version:path) => {
 		impl $crate::NativeExecutionDispatch for $name {
-			fn native_equivalent() -> &'static [u8] {
-				// WARNING!!! This assumes that the runtime was built *before* the main project. Until we
-				// get a proper build script, this must be strictly adhered to or things will go wrong.
-				$code
-			}
-
 			fn dispatch(
 				ext: &mut $crate::Externalities<$crate::Blake2Hasher>,
 				method: &str,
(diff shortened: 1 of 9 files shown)


core/test-runtime/src/genesismap.rs
```diff
@@ -29,6 +29,10 @@ pub struct GenesisConfig {
 	pub authorities: Vec<AuthorityId>,
 	pub balances: Vec<(AccountId, u64)>,
 	pub heap_pages_override: Option<u64>,
+	/// Additional storage key pairs that will be added to the genesis map.
+	pub extra_storage: Vec<(Vec<u8>, Vec<u8>)>,
+	/// Prevent direct construction of this object.
+	_use_ctor: std::marker::PhantomData<()>,
```
tomusdrw: Why do we need this? Can't we just make the fields private or crate-public?
If not then I think `_use_ctor:()` would work equally well :)

pepyakin: I think crate-public won't get the effect I wanted. But I didn't even try to use private fields because I didn't want to wait for a compile cycle. Regarding `PhantomData` and `()` yeah that was just an unconscious brainfart.

pepyakin: yep, it seems that there were no reason to keep them in non-private!

core/executor/src/native_executor.rs
```diff
@@ -211,19 +208,13 @@ impl<D: NativeExecutionDispatch> CodeExecutor<Blake2Hasher> for NativeExecutor<D
 /// Implements a `NativeExecutionDispatch` for provided parameters.
 #[macro_export]
 macro_rules! native_executor_instance {
-	( $pub:vis $name:ident, $dispatcher:path, $version:path, $code:expr) => {
+	( $pub:vis $name:ident, $dispatcher:path, $version:path) => {
```
bkchr:
```suggestion
	( $pub:vis $name:ident, $dispatcher:path, $version:path $(,)?) => {
```

core/test-runtime/client/src/lib.rs
```diff
@@ -97,12 +96,33 @@ pub type LightExecutor = client::light::call_executor::RemoteOrLocalCallExecutor
...
+
+impl GenesisParameters {
+	fn genesis_config(&self) -> GenesisConfig {
+		GenesisConfig::new(
+			self.support_changes_trie,
+			vec![
+				sr25519::Public::from(Sr25519Keyring::Alice).into(),
+				sr25519::Public::from(Sr25519Keyring::Bob).into(),
+				sr25519::Public::from(Sr25519Keyring::Charlie).into(),
+			], vec![
```
bkchr:
```suggestion
			],
			vec![
```

core/test-runtime/client/src/lib.rs
```diff
@@ -157,6 +177,13 @@ pub trait TestClientBuilderExt<B>: Sized {
 	/// Override the default value for Wasm heap pages.
 	fn set_heap_pages(self, heap_pages: u64) -> Self;
 
+	/// Add an extra value into the genesis storage.
+	///
+	/// # Panics
+	///
+	/// Panics if the key is empty.
+	fn add_extra_storage(self, key: Vec<u8>, value: Vec<u8>) -> Self;
```
bkchr:
```suggestion
	fn add_extra_storage<K: Into<Vec<u8>, V: Into<Vec<u8>>(self, key: K, value: V) -> Self;
```

So the api is more ergonomic.

core/test-runtime/client/src/lib.rs
```diff
@@ -182,28 +209,17 @@ impl<B> TestClientBuilderExt<B> for TestClientBuilder<
 		self
 	}
 
+	fn add_extra_storage(mut self, key: Vec<u8>, value: Vec<u8>) -> Self {
```
bkchr:
```suggestion
	fn add_extra_storage<K: Into<Vec<u8>, V: Into<Vec<u8>>(self, key: K, value: V) -> Self {
```

core/test-runtime/client/src/lib.rs
```diff
@@ -182,28 +209,17 @@ impl<B> TestClientBuilderExt<B> for TestClientBuilder<
...
+		assert!(!key.is_empty());
```
bkchr:
```suggestion
		let key = key.into();
		assert!(!key.is_empty());
```

core/test-runtime/client/src/lib.rs
```diff
@@ -182,28 +209,17 @@ impl<B> TestClientBuilderExt<B> for TestClientBuilder<
...
+		self.genesis_init_mut().extra_storage.push((key, value));
```
bkchr:
```suggestion
		self.genesis_init_mut().extra_storage.push((key, value.into()));
```

core/test-runtime/client/src/lib.rs
pepyakin: Hah, had this change in some other stashed PR 👍

core/test-runtime/client/src/lib.rs
pepyakin: Agreed!

PR #3497, opened 2019-08-27
babe: filter secondary slot uncles

andresilva:
Filter babe uncles to ensure that only primary slot blocks can be included. Secondary slots have deterministic allocation therefore they should not clash with other secondary slots. They might clash with primary slots, but in this case we don't want the secondary slot block to be uncled since it was only a fallback for the primary slot block.

### DIFF
--- a/srml/babe/src/lib.rs
+++ b/srml/babe/src/lib.rs
@@ -29,7 +29,7 @@ use rstd::{result, prelude::*};
 use srml_support::{decl_storage, decl_module, StorageValue, StorageMap, traits::FindAuthor, traits::Get};
 use timestamp::{OnTimestampSet};
 use sr_primitives::{generic::DigestItem, ConsensusEngineId, Perbill};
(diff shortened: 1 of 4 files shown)


srml/babe/src/lib.rs
```diff
@@ -276,6 +276,46 @@ impl<T: Trait> IsMember<AuthorityId> for Module<T> {
...
+			.iter()
+			.filter_map(|s| s.as_pre_runtime())
+		{
+			if id == BABE_ENGINE_ID {
+				let pre_digest = RawBabePreDigest::decode(&mut data)
+					.map_err(|_| "Failed to decode raw babe pre-digest.")?;
+
+				match pre_digest {
+					RawBabePreDigest::Primary { .. } =>
+						return Ok(None),
```
rphmeier: Should this do some seal verification or extract the author?

andresilva:
There are some things that are wrong and need to be reworked.

andresilva:
Closing this for now.

PR #3498, opened 2019-08-27
Decouple Phragmen from Staking. 

kianenigma:
For the sake of it being used in two modules now (Staking and new council) I want to both clean and optimize the phragmen code to some extend. 

So what happened is that:

- I partially moved the main chunk of it to sr-primitives in #3364 .
- I also started some improvements to it #3456.

At this point, I don't want to take #3356 any further because I want to rebase everything into a phragmen implementation which is _100% decoupled from staking module_. I assumed #3364 would get merged easily but I don't see that happening anytime soon because @gavofyork is offline this week. Hence, this PR is to override this. Also renames some terms and types to make it independent of Staking.

As soon as it is done, both of the mentioned PRs above should be rebased with these changes. From that point onward #3364 would stay unchanged and just needs a merge and #3456 can proceed without the problem of a merge-hell later on

An enhancement would be to move the benchmarks and tests of phragmen out of staking but I will not do that in this PR to minimize the changes.


bkchr:
Why not move it into an own crate? This is an algorithm and sounds like something that can happily live in its own space. In the future, I would even expect that this crate live in its own repository like Grandpa.

kianenigma:
I am totally in for that but once the tests + benchmarked + reference impl with f64 is also provided. If you agree with my roadmap above and why I really want this decoupling to push my other two PRs forward, for now, we can start reviewing this (it is almost insubstantial in terms of logic change) and migrate to a crate soonish.

kianenigma:
Well, it is possible but moving to its own crate needs a bit more work since we are using a few traits/type from substrate primitives. Not sure how valuable will the migration be if we still depend on substrate?

bkchr:
In a first step I meant just its own crate in this repository.

srml/staking/src/lib.rs
```diff
@@ -1339,11 +1317,24 @@ impl<T: Trait> Module<T> {
...
-				if e.total < slot_stake {
-					slot_stake = e.total;
+			for (c, s) in supports.into_iter() {
+				// build `struct exposure` from `support`
+				let exposure = Exposure {
+					own: to_balance(s.own),
+					// This might reasonably saturate and we cannot do much about it. The sum of
+					// someone's stake might exceed the balance type if they have the maximum amount
+					// of balance and receive some support. This is super unlikely to happen, yet
+					// we simulate it it in some tests.
```
kianenigma:
```suggestion
					// we simulate it in some tests.
```

core/phragmen/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "phragmen"
```
bkchr:
```suggestion
name = "substrate-phragmen"
```

srml/staking/src/benches.rs
```diff
@@ -71,67 +72,55 @@ fn do_phragmen(
...
-					a.into_iter().map(|(acc, r)| (
-						acc.clone(),
-						r,
-						to_balance(ratio_of(Staking::slashable_balance_of(&n), r)),
-					))
-					.collect::<Vec<Assignment<Test>>>()
-				)).collect::<Vec<(AccountId, Balance, Vec<Assignment<Test>>)>>();
-
-				let mut exposures = <ExpoMap<Test>>::new();
+					<<Test as Trait>::CurrencyToVote as Convert<Balance, AccountId>>::convert(b) as ExtendedBalance;
```
bkchr:
```suggestion
				<<Test as Trait>::CurrencyToVote as Convert<Balance, AccountId>>::convert(b) as ExtendedBalance;
```

core/phragmen/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+	candidate_count: usize,
+	minimum_candidate_count: usize,
+	initial_candidates: Vec<AccountId>,
+	initial_voters: Vec<(AccountId, Vec<AccountId>)>,
+	stake_of: FS,
+	self_vote: bool,
+) -> Option<PhragmenResult<AccountId>> where
+	AccountId: Default + Ord + Member,
+	Balance: Default + Copy + SimpleArithmetic,
+	for <'r> FS: Fn(&'r AccountId) -> Balance,
```
bkchr:
```suggestion
	for<'r> FS: Fn(&'r AccountId) -> Balance,
```

core/phragmen/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+	let mut voters: Vec<Voter<AccountId>> = Vec::with_capacity(num_voters);
+
+	// collect candidates. self vote or filter might apply
+	let mut candidates = if self_vote {
+		// self vote. filter.
+		initial_candidates.into_iter().map(|who| {
+			let stake = stake_of(&who);
+			Candidate { who, approval_stake: to_votes(stake), ..Default::default() }
+		})
+		.filter_map(|c| {
```
bkchr: You want to have a: `filter(|c| !c.approval_stake.is_zero())`

core/phragmen/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+
+		for (voter, assignment) in assignments.iter_mut() {
+			let voter_budget = stake_of(&voter);
+
+			let diff = do_equalize::<_, _, C>(
+				voter,
+				voter_budget,
+				assignment,
+				supports,
+				tolerance
```
bkchr:
```suggestion
				tolerance,
```

PR #3499, opened 2019-08-27
grandpa: only send catch-up requests to authorities

andresilva:
Observers update their view with neighbor packets but since they don't run the full GRANDPA protocol they don't keep track of the vote data necessary to reply to catch-up requests. We need to make sure that we only issue catch up requests to authorities.

### DIFF
--- a/core/finality-grandpa/src/communication/gossip.rs
+++ b/core/finality-grandpa/src/communication/gossip.rs
@@ -1835,9 +1835,8 @@ mod tests {
 		// importing a neighbor message from a peer in the same set in a later
 		// round should lead to a catch up request but since the node is not an
 		// authority we should get `None`.
-		match import_neighbor_message(peer_full) {
-			None => {},
-			_ => panic!("expected no catch up message"),
+		if import_neighbor_message(peer_full).is_some() {
+			panic!("expected no catch up message");
 		}
 
 		// importing the same neighbor message from a peer who is an authority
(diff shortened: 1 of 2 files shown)


core/finality-grandpa/src/communication/gossip.rs
```diff
@@ -1794,4 +1799,55 @@ mod tests {
...
+				},
+			);
+
+			catch_up_request
+		};
+
+		// importing a neighbor message from a peer in the same set in a later
+		// round should lead to a catch up request but since the node is not an
+		// authority we should get `None`.
+		match import_neighbor_message(peer_full) {
```
bkchr:
```suggestion
		if import_neighbor_message(peer_full).is_some() {
```

core/finality-grandpa/src/communication/gossip.rs
```diff
@@ -1794,4 +1799,55 @@ mod tests {
...
+			None => {},
```
bkchr:
```suggestion
```

core/finality-grandpa/src/communication/gossip.rs
```diff
@@ -1794,4 +1799,55 @@ mod tests {
...
+			_ => panic!("expected no catch up message"),
```
bkchr:
```suggestion
			panic!("expected no catch up message");
```

rphmeier:
Couldn't we ask from anyone advertising that they are a full GRANDPA node watching for votes? I suppose that the main barrier is that we don't have a good way to see that right now.

andresilva:
Yeah, right now we just ask passively when we see a neighbor packet. But in practice we should see a neighbor packet from an authority at some point at which point we'll ask for a catch up.

PR #3500, opened 2019-08-27
im-online: use generic crypto

andresilva:
The i'm online module is hardcoded to use sr25519 keys but we have no specific requirement on crypto type. The crypto is now generic and we support ed25519 and sr25519 for the module. Additionally, I also had to change the `authority-discovery` module to depend on `im_online` and use it's generic public key type.

cc @drewstone

### DIFF
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -47,7 +47,7 @@ use elections::VoteIndex;
 use version::NativeVersion;
 use primitives::OpaqueMetadata;
 use grandpa::{AuthorityId as GrandpaId, AuthorityWeight as GrandpaWeight};
-use im_online::{AuthorityId as ImOnlineId};
+use im_online::sr25519::{AuthorityId as ImOnlineId};
 
 #[cfg(any(feature = "std", test))]
 pub use sr_primitives::BuildStorage;
@@ -393,6 +393,7 @@ impl sudo::Trait for Runtime {
 }
 
 impl im_online::Trait for Runtime {
+	type AuthorityId = ImOnlineId;
 	type Call = Call;
 	type Event = Event;
 	type UncheckedExtrinsic = UncheckedExtrinsic;
@@ -447,8 +448,8 @@ construct_runtime!(
 		Treasury: treasury::{Module, Call, Storage, Event<T>},
 		Contracts: contracts,
 		Sudo: sudo,
-		ImOnline: im_online::{Module, Call, Storage, Event, ValidateUnsigned, Config},
-		AuthorityDiscovery: authority_discovery::{Module, Call, Config},
+		ImOnline: im_online::{Module, Call, Storage, Event<T>, ValidateUnsigned, Config<T>},
+		AuthorityDiscovery: authority_discovery::{Module, Call, Config<T>},
 		Offences: offences::{Module, Call, Storage, Event},
 	}
 );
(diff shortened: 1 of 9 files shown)


srml/im-online/src/lib.rs
```diff
@@ -67,44 +67,60 @@
...
+	pub type AuthorityPair = super::app_sr25519::Pair;
+
+	/// An i'm online signature using sr25519 as its crypto.
+	pub type AuthoritySignature = super::app_sr25519::Signature;
+
+	/// An i'm online identifier using sr25519 as its crypto.
+	pub type AuthorityId = super::app_sr25519::Public;
+}
+
+mod app_ed25519 {
```
kianenigma: much apologies for the mundane question just wondering why this module requires two different pairs?

Also, is this how we want to standardize it internally? a `mod app_xxx` and a `pub mod xxx`?

srml/im-online/src/lib.rs
```diff
@@ -158,8 +174,11 @@ pub struct Heartbeat<BlockNumber>
...
 
 pub trait Trait: system::Trait + session::historical::Trait {
+	/// The identifier type for an authority.
+	type AuthorityId: Member + Parameter + AppPublic + RuntimeAppPublic + Default;
+
 	/// The overarching event type.
-	type Event: From<Event> + Into<<Self as system::Trait>::Event>;
+	type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
 
 	/// The function call.
```
kianenigma:
```suggestion
	/// A dispatchable call type.
```

srml/im-online/src/lib.rs
```diff
@@ -232,15 +253,15 @@ decl_module! {
 				&current_session,
 				&heartbeat.authority_index
 			);
-			let keys = Keys::get();
+			let keys = Keys::<T>::get();
 			let public = keys.get(heartbeat.authority_index as usize);
```
kianenigma: @bkchr while you are thinking of a way for returning an encoded iterator you might also think of a way to return n'th element of a vector without decoding the whole thing :p

srml/im-online/src/lib.rs
```diff
@@ -67,44 +67,60 @@
...
 use srml_support::{
-	StorageValue, decl_module, decl_event, decl_storage, StorageDoubleMap, print, ensure
+	decl_module, decl_event, decl_storage, print, ensure,
+	Parameter, StorageValue, StorageDoubleMap,
 };
 use system::ensure_none;
 
-mod app {
-	pub use app_crypto::sr25519 as crypto;
+mod app_sr25519 {
```
bkchr: I would move the app modules into the respective crypto modules.

srml/im-online/src/lib.rs
bkchr: It does not require two different pairs. It can just support both. There is no reason to just support `sr25519`.

node/runtime/src/lib.rs
```diff
@@ -393,6 +393,7 @@ impl sudo::Trait for Runtime {
 }
 
 impl im_online::Trait for Runtime {
+	type AuthorityId = ImOnlineId;
```
kianenigma: Just to be clear, we are importing sth from the im-online module and feeding it again to the module's trait, right? is this like a sort of a default?

srml/im-online/src/lib.rs
bkchr: That does not work, as we don't know where an element starts in the encoded data.

kianenigma: ahh really? I thought we could simply find the offset in a collection that has a unified type. 
But now I see your point: even for simplest types like numbers the encoding of the same type with different values could have different lengths so nvmnd.

node/runtime/src/lib.rs
andresilva: No, here we are fixing the crypto to be sr25519. If you check `ImOnlineId` is imported from the `sr25519` package, we could as well import and use `ed25519` instead now.

andresilva:
All grumbles have been addressed.

PR #3504, opened 2019-08-28
Never panic during execution proof check

### LINKED ISSUE #3470: Light client may panic if invalid execution proof is provided
Steps to reproduce:
1) revert changes from #3469;
2) run Alice + Bob + light client locally;
3) open https://polkadot.js.org/apps/#/ page that connects to the light node.

Result: light client panics with "Externalities not allowed to fail within runtime: ...".
We need special mode for `Ext` that disables calls to `panic_handler::AbortGuard::force_abort()` when untrusted code is executed.

svyatonik:
closes #3470 

Description of problem: remote execution proof only contains parts of state (trie) that were touched during execution on remote node. Like (simplifying) if full state is `{ K1=V1, K2=V2, K3=V3 }`, then remote proof could be `{ K2=V2 }`. And if, when executed locally, code will try to access `K3`, it'll *probably* panic, because as we all know `Externalities not allowed to fail within runtime`. So this assumption isn't valid for this case.

When I was thinking about possible solution, I assumed that:
1) we do not want to make `Externalities` methods  return `Result` - this'll lead either to changing the whole runtime(s) to return results everywhere, or infinite number of expects in runtime code, or the same number of expects in some glue code (`WasmExecutor`, runtime support, ...);
2) we only want to capture unwindings of panics from within `Externalities` (that's what happens now).

So the solution I suggest is to:
1) during execution proof check, mark state backend as 'untrusted';
2) if state backend is untrusted, then we're turning on special panic handler mode (`NeverAbort`), which will ignore all further changes to `Abort` until we drop the guard;
3) in addition to `Externalities` there's now `StorageExternalities`. The main difference is that all methods of all implementations of `StorageExternalities` should only modify its internal state (like `OverlayChanges`), which must be discarded if execution fails;
4) in wasm executor there are now two methods that provide access to `Externalities`: `with_ext` - will work as before. `with_storage_ext` - will provide access only to `StorageExternalities` part AND will capture all panics, turning them into `Error::Runtime`;
5) so, when working with 'untrusted state backend' if panic occurs within `StorageExternalities` when wasm code is executing, it will be captured and transformed into execution error, allowing us to handle this properly. As before, if 'trusted state backend' is used (it is, in all cases, but the remote execution proof check), all externalities panics will lead to process abort.


core/state-machine/src/lib.rs
```diff
@@ -145,8 +145,14 @@ impl fmt::Display for ExecutionError {
...
 
-/// Externalities: pinned to specific active address.
-pub trait Externalities<H: Hasher> {
+/// Storage externalities API: pinned to specific active address.
+///
+/// Implementations of methods of this trait guarantee that there are no side effects,
+/// other than modifications to some **internal state** which could easily be
+/// discarded by throwing this instance of this Externalities away.
+///
+/// Every method may panic if storage access fails.
```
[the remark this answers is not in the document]
svyatonik: I think may is better - it is the method' logic - whether to handle errors, or panic.

core/panic-handler/src/lib.rs
```diff
@@ -31,7 +31,18 @@ use std::cell::Cell;
...
+
+/// Panic action.
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum OnPanic {
+	/// Abort when panic occurs.
+	Abort,
+	/// Unwind when panic occurs.
+	Unwind,
+	/// Always unwind even if someone changes strategy to Abort afterwards.
+	NeverAbort,
```
pepyakin: Oh no, this is worrysome because if I see `force_abort` it doesn't mean that it will abort. The name `force_abort` doesn't add clarity as well...

core/executor/src/wasm_executor.rs
```diff
@@ -50,18 +50,52 @@ struct FunctionExecutor<'e, E: Externalities<Blake2Hasher> + 'e> {
...
+	/// Execute closure that has access to storage externalities part.
+	///
+	/// All panics that happen within closure are captured and transformed into
+	/// runtime error. This requires special panic handler mode to be enabled
+	/// during the call (see `panic_handler::AbortGuard::never_abort`).
+	/// If this mode isn't enabled, then all panics within externalities are
+	/// leading to process abort.
+	pub fn with_storage_ext<T, F>(&mut self, f: F) -> Result<T>
+		where
+			F: ::std::panic::UnwindSafe + FnOnce(&mut dyn StorageExternalities<Blake2Hasher>) -> Result<T>
```
pepyakin: This line exceed 100 chars which is acceptable but we can still improve it:

1. I believe that we don't need the leading `::` 
2. Since there are a few `std::` doesn't it warrant us to have `use std::panic`?

core/state-machine/src/lib.rs
```diff
@@ -245,7 +253,13 @@ pub trait Externalities<H: Hasher> {
 
 	/// Get the change trie root of the current storage overlay at a block with given parent.
 	fn storage_changes_root(&mut self, parent: H::Out) -> Result<Option<H::Out>, ()> where H::Out: Ord;
+}
 
+/// Extra externalities API: pinned to specific active address.
```
pepyakin: What active address? I bet that was from super old times when there were contracts in substrate/polkadot. I.e. totally irrelevant

core/state-machine/src/lib.rs
```diff
@@ -245,7 +253,13 @@ pub trait Externalities<H: Hasher> {
...
+///
+/// Methods of this trait may have side-effects.  Therefore, it isn't safe
+/// to catch unwinding of any panics that occur during calls to them.
```
pepyakin: Can you elaborate what connection do panics and side-effect have here?

core/state-machine/src/lib.rs
```diff
@@ -146,8 +146,14 @@ impl fmt::Display for ExecutionError {
 
 type CallResult<R, E> = Result<NativeOrEncoded<R>, E>;
 
-/// Externalities: pinned to specific active address.
-pub trait Externalities<H: Hasher> {
+/// Storage externalities API: pinned to specific active address.
```
pepyakin: See the other comment below regarding the active address (unfortunately no link right now, since it is not possible to get a link while drafting a review)

core/state-machine/src/lib.rs
```diff
@@ -397,13 +413,25 @@ type DefaultHandler<R, E> = fn(
 	CallResult<R, E>,
 ) -> CallResult<R, E>;
 
+/// Storage backend trust level.
+#[derive(Debug, Clone)]
+pub enum BackendTrustLevel {
+	/// Panics from trusted backend are justified and never caught.
```
DemiMarie-parity:
```suggestion
	/// Panics from trusted backends are considered justified, and never caught.
```

core/state-machine/src/lib.rs
```diff
@@ -397,13 +413,25 @@ type DefaultHandler<R, E> = fn(
...
+	Trusted,
+	/// Panics from untrusted backend are caught and interpreted as runtime error.
```
DemiMarie-parity:
```suggestion
	/// Panics from untrusted backend are caught and interpreted as runtime errors.
```

core/state-machine/src/lib.rs
```diff
@@ -397,13 +413,25 @@ type DefaultHandler<R, E> = fn(
...
+	/// Untrusted backend may miss some parts of trie => panics are not considered
```
DemiMarie-parity:
```suggestion
	/// Untrusted backend may be missing some parts of the trie, so panics are not considered
```

core/state-machine/src/lib.rs
svyatonik: My thoughts were:
1) we do not update storage directly from `StorageExternalities` methods - instead we're buffering changes && only update storage when block is actually inserted into db. You may think of this as an implementation details (I mean `Ext` + `OverlayChanges`), but imo there's no other way to organize block import - just because if you're writing directly to storage, then any panic will lead to storage corruption. That's why I consider catching panics from within storage-related methods to be `AssertUnwindSafe`;
2) keystore/offchain are singletons, which may have their own global state (I could say for sure that keystore stores some keys in memory). If keystore/offchain method will panic and we will catch this panic, this could leave these objects in some inconsistent state. That's why we can't catch panics from these methods.

core/panic-handler/src/lib.rs
svyatonik: Yes - I agree :) I could rename to `unwind()`, `abort()` and `force_unwind()` if you think it is better. Or you could suggest some other naming

svyatonik:
I'm not super-happy with this solution too, but also was unable to find better solution for 2 days. Tried many options (could share some on request), but this one looked 'mostly OK' to me.

And let me be a devil's advocate here - expects in `Ext` implementation do not fit proof-or-remove policy, for sure. And I would prefer them not to be there. Nevertheless, there were so many issues that we have found + **not ignored simply because process has aborted** + fixed because they're panics.

pepyakin:
To be clear, I am not proposing removing panics there and I also share the opinion that aborting the process is better than swallowing/coercing these panics to runtime execution errors.

Ok, I think a better solution might be to centralize panic handling. 

I.e. instead of directly specifying code scopes where either to abort, to unwind or to forceful abort, we could instead do something like this:


```
           ┌──────────────────────────────┐   
           │  > Panics are caught here <  │   
           │Handling depends on source and│   
           │          exec mode           │   
        ┌──┴──────────────────────────────┴──┐
        │                                    │
        │                                    │
        │  ▲                                 │
           │                                  
           │                                  
           │    ┌─────────────────┐           
           │    │                 │           
           │    │                 │           
  Panics   │    │     Runtime     │           
 bubble up │    │                 │           
 in stack  │    │                 │           
           │    │                 │           
           │    └─────────────────┘           
           │                                  
           │    ━━━━━━━━━━━━━━━━━━━           
           │                                  
                   Externalities              
```

We track the source of a panic to know where a specific panic came from: runtime or externalities. All panics unwind regardless where they came from. We catch them before entering to the runtime code. We handle these panics depending on the mode (trusted/not trusted) and source. E.g. we would return a runtime error if it came from the runtime, if panic came from externalities, we would return a special backend error in case we are in untrusted backend mode or straight abort if the backend is trusted.

svyatonik:
I'm not sure I got the idea - your suggestion (at least how I understood it) is to `throw ExternalitiesError()` from all `ext.rs` panics. But there's no way to associate some data with panic in rust, right? So we must store the 'type' of panic somewhere else. Some global var? So basically - either we will add something like
```rust
match storage.get(hash) {
  Ok(value) => Ok(value),
  Err(error) => {
     i_m_going_to_panic_inside_externalities();
     panic!("Externalities not allowed to fail within runtime");
  }
}
```
, or
```
let ext_result = catch_unwind(|| { ext.storage(hash); })
if let Err(error) = ext_result {
     i_m_going_to_panic_inside_externalities();
     resume_unwind(error);
} 
```

If that's what you suggesting, I do not see - why it is better than scoped approach :/

pepyakin:
Yeah, basically you can think of it as "throwing an exception" and yes, you are right, there is no way to directly associate a custom payload with a panic. However, we already doing precisely that with [panic-handler](https://github.com/paritytech/substrate/blob/6f4d992b2a381908f8bc159b6b8e1e88c042ec4c/core/panic-handler/src/lib.rs#L34).

So code-wise my idea is something like this.

Externalities could almost the same from outside.

```rust
	fn original_storage(&self, key: &[u8]) -> Option<Vec<u8>> {
		let _guard = ExtGuard::ext_guard();
		self.backend.storage(key).expect(EXT_NOT_ALLOWED_TO_FAIL)
	}
```

However, they use an `ext_guard` or something. `ExtGuard` can also answer the `is_in_ext() -> bool` query. Note that it is similar to `AbortGuard` in that it also has a `thread_local!` flag. However, in contrast to `AbortGuard`, `ExtGuard` doesn't change any panic semantics - it is just a dumb flag.

The consequence of this change is that all panics both from the (native) runtime and from externalities always unwind.

Then, somewhere where we call into a runtime, we put a single centralized panic handler:

```rust
match catch_unwind(|| call_into_runtime(..)) {
  Ok(result) => Ok(result),
  Err(_) if untrusted_backend => Err(InvalidProof),
  err @ Err(_) => resume_unwind(err),
}
```

Advantage of this approach is that you'd have a single place where you decide what to do in case of panic, instead of overriding logic in specific places.

svyatonik:
OK, I still do not think it is any better than proposed implementation. That's why:
1) `ExtGuard` won't work that way, simply because `ExtGuard::drop()` will be called on panic, which will set `is_in_ext` to false again. So we'll either need additional `catch_unwind`s, or something like:
```rust
fn fn_from_ext() {
	we_are_in_ext();
	// do what required
	we_are_not_in_ext();
}
```
in every method of `Ext` (and all other non-test `Externalities` implementations that will ever be created);
2) actually, some kind of `ExtGuard` will only be helpful where we call `catch_unwind` - we may want clear that flag. And, tbh, we **must** clear it, because it has very strange semantics - 'last panic has happened inside Externalities'. This semantics is only defined within runtime call. Compare this to general panic handler strategy, which is always valid;
3) if we'll use `resume_unwind`, that would mean that the process won't abort if panic happens in non-main thread. I think we agreed that this isn't good;
4) if we'll `abort` instead of `resume_unwind`, then we're losing location+backtrace of original panic. We can try to save this in the same global variable, but this makes the whole construction even worse. Like infamous `GetLastError` function from WinApi;
5) and last, but not the least, we'll have to mark closure used in `catch_unwind` as `AssertUnwindSafe`. And though the handler you're proposing should be fine with that, I still do not like the fact - objects that will be passed to that closure aren't actually safe for unwinding. That's why I've separated `Externalities` and `StorageExternalities` traits.

pepyakin:
I see really good points there! However, it seems to me that they mostly address the particular code snippet. Of course, I am not even sure if it compiles and there might be other issues: I haven't delved into the code to check everything. IMO they should be solvable (e.g. pt.1 is solved by using a closure) if we put some thought into it (OTOH maybe some local hacks would be required.).

It seems to be that the main issue is with 

> This semantics is only defined within runtime call. Compare this to general panic handler strategy, which is always valid;

I'd argue that this is perflectly fine and reflects the real state of things. Let's remember ourselves that the runtime, or rather it's native version, is really special. The runtime is written with an assumption that is has some sort of ambient state around it. In wasm runtime we have wasm_executor. For the native API we have [this](https://github.com/paritytech/substrate/blob/6f4d992b2a381908f8bc159b6b8e1e88c042ec4c/core/sr-io/with_std.rs#L32): note all of this `ext::with` and also the [`with_externalities`](https://github.com/paritytech/substrate/blob/6f4d992b2a381908f8bc159b6b8e1e88c042ec4c/core/sr-io/with_std.rs#L453). This implies that the runtime is inherently scoped.

> Compare this to general panic handler strategy, which is always valid;

Right, although it doesn't seem to me as an advantage: it allows code that overrides overriden modes which would be perfectly valid and there is no way to check that. OTOH, the explicitly scoped approach would allow you to actually check if the closure was called outside the runtime call scope to catch this.

If you bought on the advantages of the approach we can discuss the details : )

DemiMarie-parity:
I think the real problem here is that we are using panics for exception handling, which they are’t designed for.  That brings in all of exception handling’s problems.

A better idea, in my opinion, is to wrap a `catch_unwind` around exactly the blocks of code where we _do_ expect to panic, and no others.

svyatonik:
@pepyakin Nope - my comment is not about the snippet, it is mostly about number of hacks we'll need to introduce to 'centralize' panic handling. IMO instead this makes panic handling spread all over the codebase (by introducing additional strange global vars, by preserving panic' artifacts for processing elsewhere) + introducing additional error-prone assumptions (like marking everything `AssertUnwindSafe`) + by changing implementations. Instead of aborting as soon, as we see the panic, we're trying to introduce (`setjmp`+`longjmp`+`static`)-type exception handling just to...abort. 

That said, I don't think that 'centralized' panic handling worth implementing that structure you've suggested + I'm not seeing how this makes panic handling actually centralized. I do not want to continue arguing, though - I think we both have had enough arguments here. I can fix the issue once the 'centralized handling' is ready, but I'm not going to sign it off, or be a co-author. Feel free to close this PR once there's issue/other PR that suggests/introduces that new panic handling mechanism.

pepyakin:
While I still think that the centralized approach is worth to explore, I am not 100% sure that it will be fruitful. Failing to convience you in advantages of this didn't give me more confidence.

Since I am completely agree that we spent on this enough time, I wouldn't block merging this PR and not going to work towards exploring my idea unless someone else comes in and either supports it or has a better idea. (I also think that this hasn't had enough attention from the rest of the team, but oh well).

Anyway, the code looks good in that it implements the stated approach properly.

core/state-machine/src/lib.rs
```diff
@@ -397,13 +413,25 @@ type DefaultHandler<R, E> = fn(
...
+
 /// Like `ExecutionStrategy` only it also stores a handler in case of consensus failure.
 #[derive(Clone)]
 pub enum ExecutionManager<F> {
 	/// Execute with the native equivalent if it is compatible with the given wasm module; otherwise fall back to the wasm.
 	NativeWhenPossible,
-	/// Use the given wasm module.
-	AlwaysWasm,
+	/// Use the given wasm module. The backend on which code is executed code could be
+	/// trusted or untrusted.
```
gavofyork:
```suggestion
	/// trusted to provide all storage or not (i.e. the light client cannot be trusted to provide
	/// for all storage queries since the storage entries it has come from an external node).
```

core/state-machine/src/testing.rs
```diff
@@ -278,7 +278,14 @@ impl<H, N> Externalities<H> for TestExternalities<H, N>
 			parent,
 		)?.map(|(_, root)| root))
 	}
+}
 
+impl<H, N> Externalities<H> for TestExternalities<H, N>
+	where
```
gavofyork: move `where` to line before and reduce indent of next few lines.

core/state-machine/src/testing.rs
```diff
@@ -278,7 +278,14 @@ impl<H, N> Externalities<H> for TestExternalities<H, N>
...
+		H: Hasher,
+		N: ChangesTrieBlockNumber,
+		H::Out: Ord + 'static
```
gavofyork:
```suggestion
		H::Out: Ord + 'static,
```

PR #3506, opened 2019-08-28
Simplify deposit_event declaration in decl_module

thiolliere:
## Context

in decl_module to declare a default deposit_event previously user had to put into brackets the generic type used by deposit_event (like `deposit_event<T, I>()`, this wasn't super easy to guess so now it is not necessary anymore just having `fn deposit_event() = default;` will add a default which works.

## Changes:

* Removes the possibility to declare a custom deposit_event in decl_module, (if you want to declare custom functions on your module just do so outside this function).
* Change the way to declare deposit_event from `deposit_event <$event_generics>() = default;` to `deposit_event() = default;`


srml/support/src/dispatch.rs
```diff
@@ -317,23 +317,11 @@ macro_rules! decl_module {
...
-			{ $( $other_where_bounds )* }
-			{ $vis fn deposit_event $(<$dpeg $(, $dpeg_instance)?>)* ($( $param_name: $param ),* ) { $( $impl )* } }
-			{ $( $on_initialize )* }
-			{ $( $on_finalize )* }
-			{ $( $offchain )* }
-			{ $( $constants )* }
-			[ $( $dispatchables )* ]
-			$($rest)*
+		compile_error!(
+			"`deposit_event` function is reserved and must follow the syntax: `$vis:vis fn deposit_event() = default;`"
```
bkchr: I think we should delegate this to the `__check_reserved_fn_name` macro.

shawntabrizi:
Dumb question, will this effect adding `Event` or `Event<T>` to the `construct_runtime`?

srml/support/src/dispatch.rs
thiolliere: actually check_reserved_fn_name check if some dispatchable function have the name of a reserved function. Like someone can declare a dispatchable function on_finalize just by not following on_finalize accepting signature and this is confusing.

Thus if we remove this branch and I add an old syntax `deposit_event<T>() =default;` somewhere the error is actually:

```
error: no rules expected the token `normalize`
   --> srml/balances/src/lib.rs:399:1
    |
399 | / decl_module! {
400 | |     pub struct Module<T: Trait<I>, I: Instance = DefaultInstance> for enum Call where origin: T::Origin {
401 | |         /// The minimum amount required to keep an account open.
402 | |         const ExistentialDeposit: T::Balance = T::ExistentialDeposit::get();
...   |
506 | |     }
507 | | }
    | |_^ no rules expected this token in macro call
    |

```

note that those parsing message error should be easier to solve with procedural macro and I'm currently working on porting it to this.

bkchr: Yeah okay, makes sense.

thiolliere:
yes nothing as changed for construct_runtime perspective nor for decl_event types are all same just deposit_event signature and declaration has changed

JoshOrndorff:
Dumb question. Is it possible or desirable to eliminate the need to write `fn deposit_event() = ...` entirely? Like could `decl_module!` detect whether the runtime declares any events and act appropriately?

thiolliere:
Not in the current state: deposit event could be used outside the
decl_module only, and also user could provide its own implementation.

But maybe this function could be different and instead of being implemented
on module it could be implemented on event itself 🤔

PR #3508, opened 2019-08-28
node: update flaming fir

andresilva:
Redeploy flaming fir with latest master. Reduced the block time to 3s and the primary probability is now set to 25%.


andresilva:
cc @jacogr @ddorgan 

I tested a runtime upgrade locally and it worked with 3s blocks.

PR #3510, opened 2019-08-28
Make it easier to just rebuild the WASM files

bkchr:
- Adds a new env variable `WASM_TARGET_DIRECTORY` that controls the target directory for copying the builded WASM binary
- Adds script `build-only-wasm.sh` to only build the WASM binary
- Further improvements to `wasm-builder` to reduce the number of required rebuilds

### DIFF
--- a/core/utils/wasm-builder/src/lib.rs
+++ b/core/utils/wasm-builder/src/lib.rs
@@ -130,7 +130,9 @@ pub fn build_project(file_name: &str, cargo_manifest: &str) {
 		process::exit(1);
 	}
 
-	let (wasm_binary, bloaty) = wasm_project::create_and_compile(&cargo_manifest);
+	let (wasm_binary, bloaty) = wasm_project::create_and_compile(
+		&cargo_manifest,
+	);
 
 	create_out_file(
 		file_name,
--- a/core/utils/wasm-builder/src/lib.rs
+++ b/core/utils/wasm-builder/src/lib.rs
@@ -66,6 +66,8 @@
 //!                          needs to change. As WASM builder instructs `cargo` to watch for file changes
 //!                          this environment variable should only be required in certain circumstances.
 //! - `WASM_BUILD_RUSTFLAGS` - Extend `RUSTFLAGS` given to `cargo build` while building the WASM binary.
+//! - `WASM_TARGET_DIRECTORY` - Will copy any build WASM binary to the given directory. The path needs
+//!                            to be absolute.
 //!
 //! Each project can be skipped individually by using the environment variable `SKIP_PROJECT_NAME_WASM_BUILD`.
 //! Where `PROJECT_NAME` needs to be replaced by the name of the cargo project, e.g. `node-runtime` will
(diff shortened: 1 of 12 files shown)


core/utils/wasm-builder/src/wasm_project.rs
```diff
@@ -191,6 +203,8 @@ fn create_wasm_workspace_project(wasm_workspace: &Path) {
 	).expect("WASM workspace `Cargo.toml` writing can not fail; qed");
 }
 
+
```
andresilva:
```suggestion
```

core/utils/wasm-builder/src/wasm_project.rs
```diff
@@ -191,6 +203,8 @@ fn create_wasm_workspace_project(wasm_workspace: &Path) {
...
+
```
andresilva:
```suggestion
```

core/utils/wasm-builder/src/wasm_project.rs
```diff
@@ -349,4 +364,29 @@ fn generate_rerun_if_changed_instructions(
...
+		);
+	}
+
+	fs::create_dir_all(&target_dir).expect("Creates `WASM_TARGET_DIRECTORY`.");
+
+	fs::copy(
+		wasm_binary.wasm_binary_path(),
+		target_dir.join(format!("{}.wasm", get_wasm_binary_name(cargo_manifest))),
+	).expect("Copies WASM binary to `WASM_TARGET_DIRECTORY`.");
+}
```
andresilva:
```suggestion
}
```

scripts/build-only-wasm.sh
```diff
@@ -0,0 +1,30 @@
...
+
+if [ -d $WASM_BUILDER_RUNNER ]; then
+  export DEBUG=false
+  export OUT_DIR="$PROJECT_ROOT/target/release/build"
+  cargo run --release --manifest-path="$WASM_BUILDER_RUNNER/Cargo.toml" \
+    | grep -v "cargo:rerun-if-"                                         \
+    | grep -v "Executing build command"
+else
+  cargo build --release -p $1
+fi
```
andresilva:
```suggestion
fi
```

andresilva:
```
error: failed to select a version for the requirement `substrate-wasm-builder = "^1.0.5"`
  candidate versions found which didn't match: 1.0.4, 1.0.3, 1.0.2, ...
  location searched: crates.io index
required by package `wasm-build-runner-impl v1.0.0 (/builds/parity/substrate/target/release/build/node-template-runtime-223c277736ad335e/out/wasm_build_runner)`
```

I think you need to publish `substrate-wasm-builder`.

scripts/build-only-wasm.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env sh
+
```
xlc: should node-template/scripts have this script as well?

core/utils/wasm-builder/src/lib.rs
```diff
@@ -143,11 +169,11 @@ fn check_skip_build() -> bool {
...
 
-fn create_out_file(file_name: &str, content: String) {
-	fs::write(
-		file_name,
-		content
-	).expect("Creating and writing can not fail; qed");
+/// Write to the given `file` if the `content` is different.
+fn write_file_if_changed(file: PathBuf, content: String) {
+	if fs::read_to_string(&file).ok().as_ref() != Some(&content) {
+		fs::write(&file, content).expect(&format!("Writing `{}` can not fail!", file.display()));
```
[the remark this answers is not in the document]
bkchr: Cargo holds a lock for us and the file content is everytime the same anyway.

scripts/build-only-wasm.sh
bkchr: Always use `/usr/bin/env`. Posix even recommends that. While Nixos also has `/bin/sh` to be Posix compliant and `/usr/bin/env` to support scripts, `/usr/bin/env sh` is the cleaneast way.

https://news.ycombinator.com/item?id=17069408

PR #3512, opened 2019-08-29
Make public key mandatory in `insert_key`

bkchr:


gavofyork:
yey. now things are more painful for users.

PR #3514, opened 2019-08-29
Submit (& sign) extrinsics from the runtime.

tomusdrw:
This PR introduces a helper types to construct, sign and submit transactions from within the runtime.

I've refactored the `im-online` code to use the new types and also created a way to submit signed transactions. That's going to be needed for instance for equivocation reporting.
Also all places where signatures where constructed (and hashed) are now replaced with `SignedPayload`. Proper typing on `Extra` also detected a bug in `subkey` (where the transfers did not contain the correct `SignedExtensions` payload).

Note that to actually use this API, the call needs to be invoked in the specific `OffchainCall` context.


core/sr-primitives/src/generic/unchecked_extrinsic.rs
```diff
@@ -123,6 +130,67 @@ where
 	}
 }
 
+/// A payload that has been signed for a unchecked extrinsics.
```
bkchr:
```suggestion
/// A payload that has been signed for an unchecked extrinsics.
```

core/sr-primitives/src/generic/unchecked_extrinsic.rs
```diff
@@ -123,6 +130,67 @@ where
...
+		})
+	}
+
+	/// Deconstruct the payload into it's components.
+	pub fn deconstruct(self) -> (Call, Extra, Extra::AdditionalSigned) {
+		self.raw_payload
+	}
+}
+
+impl<Call, Extra> Encode for SignedPayload<Call, Extra> where
```
bkchr: Why is this implementation required?

core/sr-primitives/src/generic/unchecked_extrinsic.rs
```diff
@@ -123,6 +130,67 @@ where
...
+	pub fn from_raw(call: Call, extra: Extra, additional_signed: Extra::AdditionalSigned) -> Self {
+		Self {
+			raw_payload: (call, extra, additional_signed),
+		}
+	}
+
+	/// Get an encoded version of this payload.
+	///
+	/// Payloads longer than 256 bytes are going to be `blake2_256`-hashed.
+	pub fn using_encoded<O>(&self, f: impl FnOnce(&[u8]) -> O) -> O {
```
bkchr: Maybe we should give this a different name?

core/sr-primitives/src/generic/unchecked_extrinsic.rs
```diff
@@ -123,6 +130,67 @@ where
 	}
 }
 
+/// A payload that has been signed for a unchecked extrinsics.
+///
+/// Note that the payload that we sign to produce unchecked extrinsic signature
+/// is going to be different than the `SignaturePayload` - so the thing the extrinsic
+/// actually contains.
+pub struct SignedPayload<Call, Extra: SignedExtension> {
+	raw_payload: (
```
kianenigma: I could not find a reason but I just wonder if you also argued/tought of making this a flat struct rather than one with a single element of a tuple? or is there a strong reason?

node/runtime/src/lib.rs
```diff
@@ -424,6 +425,33 @@ impl finality_tracker::Trait for Runtime {
 	type ReportLatency = ReportLatency;
 }
 
+impl system::offchain::CreateTransaction<Runtime, UncheckedExtrinsic> for Runtime {
```
kianenigma: where is this being used?

srml/system/src/offchain.rs
```diff
@@ -0,0 +1,136 @@
...
+pub trait Signer<Account, Signature> {
+	/// Sign any encodable payload with given account and produce a signature.
+	///
+	/// Returns `Some` if signing succeeded and `None` in case the `account` couldn't be used.
+	fn sign<Payload: Encode>(account: Account, payload: &Payload) -> Option<Signature>;
+}
+
+impl<Account, Signature, AppPublic> Signer<Account, Signature> for AppPublic where
+	AppPublic: RuntimeAppPublic + From<Account>,
+	AppPublic::Signature: Into<Signature>,
```
kianenigma: I am a bit unclear on these new crypto types: what do the bounds here translate to? Who will end up having this impl?

core/sr-primitives/src/generic/unchecked_extrinsic.rs
tomusdrw: The generic `Signer` is working on anything that is `Encode`. I'll rather remove the `using_encoded` function, I've only added it to be able to document the behavior.

node/runtime/src/lib.rs
tomusdrw: We pass this as `SubmitTransaction` to `im-online`:
```
type SubmitTransaction = TransactionSubmitter<ImOnlineId, Runtime, UncheckedExtrinsic>;
```
(see `Runtime` as the second parameter)

core/sr-primitives/src/generic/unchecked_extrinsic.rs
tomusdrw: Yeah, makes sense. I guess, arguably `self.raw_payload.using_encoded` is more readable than `self.0`, but indeed it feels simpler to avoid naming it.

subkey/src/main.rs
```diff
@@ -182,22 +182,21 @@ fn execute<C: Crypto>(matches: clap::ArgMatches) where
 
 			println!("Using a genesis hash of {}", HexDisplay::from(&genesis_hash.as_ref()));
 
-			let raw_payload = (
+			let raw_payload = SignedPayload::from_raw(
```
kianenigma: great that this wasidentified! I exactly had in on a personal todo list to: check subkey with signed extension. If not here, at some point I would like to add some tests to it to make sure subkey is always generating stuff that can be verified by `node` (if that's a correct assumption).

srml/system/src/offchain.rs
tomusdrw: My understanding:

So we have a low-level crypto types (like `sr25519` and `ed25519`) that are commonly used within the "core" runtime part (as `AccountId`, etc).

We also have application-specific crypto types, which are basically wrappers for the low-level types with additional information about the module (i.e. application) they are coming from.

So for instance, `im-online` is (or may be) using `sr25519` as the crypto type, but instead of `sr25119::Public` you will rather use `im_online::app::Public`, as this type has additional `APP_ID` identification to know it's coming from `im_online` module.
The `APP_ID` is very useful especially when querying the `Keystore` cause we might have many sr25519 keys, but only one (or some) of them will be the one used by `im_online` module.


So the bounds here basically say:
> Please let me know the application-specific crypto type that you want to use for a particular `Account`.

Now that I think of it, in the future it might actually be better to use `AppPublic` instead of `Account` right from the start (i.e. when we initiate the signing), rather then convert in the middle of the process.
I'll refactor this if I find it more convenient to use that way, when equivocations reporting will be implemented.

core/sr-primitives/src/generic/unchecked_extrinsic.rs
bkchr: You can also put documentation on top of a trait implementation. It will be shown in rustdocs as welll.

tomusdrw:
@andresilva yes. There is a special `OffchainCall` context (see #3454) and you can specify the desired capabilities (for this you'd need `Keystore` and `TransactionPool`).

PR #3519, opened 2019-08-30
Publish minimum balance from within srml-contracts API

pepyakin:
There are two changes separated by commits:

1. Introduction of a new srml-contracts API function `ext_minimum_balance`. Now contracts don't know the minimum amount of funds to create a contract. Contract writers can hardcode it, but then it is going to be problematic if the balance is increased. `ext_minimum_balance` returns the exact amount which is required for creation of an account.
2. Renaming `existential_deposit`to `minimum_balance`. The reason for this change is that `existential_balance` is the term of the `balances` module. However, we don't really interact with it directly and query `T::Currency::minimum_balance()`. So I guess `minimum_balance` is a better name.

### DIFF
--- a/srml/contracts/src/exec.rs
+++ b/srml/contracts/src/exec.rs
@@ -154,6 +154,9 @@ pub trait Ext {
 	/// Returns a reference to the timestamp of the current block
 	fn now(&self) -> &MomentOf<Self::T>;
 
+	/// Returns the minimum balance that is required for creating an account.
+	fn minimum_balance(&self) -> BalanceOf<Self::T>;
+
 	/// Returns a random number for the current block with the given subject.
 	fn random(&self, subject: &[u8]) -> SeedOf<Self::T>;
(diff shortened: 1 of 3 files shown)


srml/contracts/src/exec.rs
```diff
@@ -154,6 +154,9 @@ pub trait Ext {
 	/// Returns a reference to the timestamp of the current block
 	fn now(&self) -> &MomentOf<Self::T>;
 
+	/// Returns the minimum balance that is required for creating an account.
+	fn minimum_balance(&self) -> BalanceOf<Self::T>;
```
bkchr: Why not call it `existential_deposit` as it is called everywhere else in Substrate?

Robbepop: Tbh I think `minimum_balance` is the better name here. Name is explained above in the PR comment.

PR #3520, opened 2019-08-30
network: Use "one shot" protocol handler.

twittner:
Add two new `NetworkBehaviour`s, one handling remote block requests and another one to handle light client requests (both local and from remote). The change is motivated by the desire to use multiple substreams of a single connection for different protocols. To achieve this, libp2p's `OneShotHandler` is used as a protocol handler in each behaviour. It will open a fresh substream for the duration of the request and close it afterwards. For block requests, we currently only handle incoming requests from remote and tests are missing. For light client handling we support incoming requests from remote and also ported a substantial amount of functionality over from `light_dispatch.rs` (including several tests). However the result lacks in at least two aspects:

1. We require external updates w.r.t. the best block per peer and currently nothing updates this information.
2. We carry a lot of peer-related state around.

Both aspects could be simplified by externalising peer selection and just requiring a specific peer ID where the request should be sent to. We still have to maintain some peer related state due to the way libp2p's swarm and network behaviour work (e.g. we must make sure to always issue `NetworkBehaviourAction::SendEvent`s to peers we are connected to, otherwise the actions die a silent death).

Another change implemented here is the use of protocol buffers as the encoding for network messages. Certain individual fields of messages are still SCALE encoded. There has been some discussion about this in another PR (https://github.com/paritytech/substrate/pull/3452), so far without resolution.


arkpar:
Is there any round-trip involved for opening a substream?
Does it increase network traffic?

twittner:
> Is there any round-trip involved for opening a substream?

Opening a new substream involves the use of [multistream-select](https://docs.rs/multistream-select/0.5.1/multistream_select/) to negotiate the protocol to use with the stream. I think this involves at least 1 roundtrip currently. Version 2 of `multistream-select` will eventually improve on this.

> Does it increase network traffic?

Depending on the multiplexer there might be an increase in control data. For example with Yamux there are `WindowUpdate` frames to implement flow control and those are sent per stream.

arkpar:
> The change is motivated by the desire to use multiple substreams of a single connection for different protocols.

So what motivates this desire then? Extra roundtrips for each request does not sound like a good idea to me.

tomaka:
> So what motivates this desire then? Extra roundtrips for each request does not sound like a good idea to me.

The main motivation is de-coupling the gossiping from block requests, so that:

- Light clients and nodes in general can make requests without occupying a gossiping slot (which is the case right now). In other words, you wouldn't occupy a gossiping slot anymore while major syncing for example.
- Nodes can close their gossiping channel without interrupting all on-going requests.
- We can open multiple gossiping channels (for example one only within validators, one for the relay chain, and one for each parachain) without having the problem of which channel to send the request to.

tomaka:
Note that, theoretically, if multistream-select 2 happens, and a better multiplexing such as QUIC happens, then that new system will be very slightly more efficient than what we currently have, as we would no longer transmit the type of message nor the request ID.

arkpar:
I don't see how adding extra network latency can be more efficient.

What exactly is "gossiping slot"? 
The current protocol was designed in such a way that parallel requests are possible.

In general I'm OK with using a separate substream for gossiping, but not with opening a new substream for each request/response. One of our major goals is to minimize network latency and data propagation time.

tomaka:
> What exactly is "gossiping slot"?

The slots whose number is controlled with the `--in-peers` and `--out-peers` options. What the peerset manager decides.

> I don't see how adding extra network latency can be more efficient.

Theoretically, there shouldn't be any additional round-trip compared to what we have right now.
Opening a substream is a very cheap operation. A substream is basically equivalent to a request ID.

The problem is that, due to the way the multistream-select protocol is designed, we have to wait for the remote to acknowledge that it supports the protocol we want (eg. "requesting a block").

There is an optimization in place in the code of multistream-select that allows skipping this round-trip, but it unfortunately can't apply here.

As a work-around, we could add a few hacks to the code of multistream-select to remove this round-trip.

core/network/src/behaviour.rs
```diff
@@ -109,6 +117,11 @@ impl<B: BlockT, S: NetworkSpecialization<B>, H: ExHashT> Behaviour<B, S, H> {
 	pub fn put_value(&mut self, key: record::Key, value: Vec<u8>) {
 		self.discovery.put_value(key, value);
 	}
+
+	/// Get unique access to the light client handler.
+	pub fn light_client_handler(&mut self) -> &mut protocol::LightClientHandler<Substream<StreamMuxerBox>, B> {
```
tomaka: I don't think this is used anywhere, so I'd prefer to remove it, otherwise someone is going to use this method. The `LightClientHandler` is supposed to be an implementation detail, so not accessible directly.

core/network/src/protocol/block_requests.rs
```diff
@@ -0,0 +1,336 @@
...
+	/// - max. block data in response = 128
+	/// - max. request size = 1 MiB
+	/// - inactivity timeout = 15s
+	/// - protocol string = b"/polkadot/sync/1"
+	pub fn new() -> Self {
+		Config {
+			max_block_data_response: 128,
+			max_request_len: 1024 * 1024,
+			inactivity_timeout: Duration::from_secs(15),
+			protocol: b"/polkadot/sync/1",
```
tomaka: I don't think there should be a default value for that.
The value should be `format!("/{}/sync/1", protocol_id)` (where `protocol_id` is the `protocol_id` of type `ProtocolId` that is passed around during the network's initialization), and that must be passed by parameter.

core/network/src/protocol/light_client_handler.rs
```diff
@@ -0,0 +1,1667 @@
...
+	/// - inactivity timeout = 15s
+	/// - request timeout = 15s
+	/// - protocol string = b"/polkadot/light/1"
+	pub fn new() -> Self {
+		Config {
+			max_data_size: 1024 * 1024,
+			max_pending_requests: 128,
+			inactivity_timeout: Duration::from_secs(15),
+			request_timeout: Duration::from_secs(15),
+			protocol: b"/polkadot/light/1",
```
tomaka: Same remark.

core/network/src/protocol/light_client_handler.rs
```diff
@@ -0,0 +1,1667 @@
...
+					if let Some(info) = self.peers.get_mut(&peer) {
+						if info.status != PeerStatus::BusyWith(id) {
+							// If we get here, something is wrong with our internal handling of peer
+							// status information. At any time, a single peer processes at most one
+							// request from us and its status should contain the request ID we are
+							// expecting a response for. If a peer would send us a response with a
+							// random ID, we should not have an entry for it with this peer ID in
+							// our `outstanding` map, so a malicious peer should not be able to get
+							// us here. It is our own fault and must be fixed!
+							panic!("unexpected peer status {:?} for {}", info.status, peer);
```
tomaka: I'd go for calling `log::error!` and returning instead.

core/network/src/protocol/light_client_handler.rs
```diff
@@ -0,0 +1,1667 @@
...
+								}
+							}
+						}
+					} else {
+						// If we get here, something is wrong with our internal handling of peers.
+						// We apparently have an entry in our `outstanding` map and the peer is the one we
+						// expected. So, if we can not find an entry for it in our peer information table,
+						// then these two collections are out of sync which must not happen and is a clear
+						// programmer error that must be fixed!
+						panic!("missing peer information for {}; response {}", peer, id);
```
tomaka: Same: `error!`

core/network/src/behaviour.rs
twittner: How do you want to expose the request functionality of the light client handler? Do you want to reproduce its `request` method here and forward the call?

core/network/src/protocol/light_client_handler.rs
twittner: I do not think it makes sense to continue past that point as the internal state is inconsistent. As mentioned in the comment, this should never be reached. If it is however, the programme as written is wrong and needs to be fixed.

It would be nice if the `NetworkBehaviour` trait provides a way to signal an error condition to the swarm which would then stop polling it.

core/network/src/behaviour.rs
tomaka: > Do you want to reproduce its request method here and forward the call?

Yes!

core/network/src/behaviour.rs
```diff
@@ -109,6 +117,11 @@ impl<B: BlockT, S: NetworkSpecialization<B>, H: ExHashT> Behaviour<B, S, H> {
 	pub fn put_value(&mut self, key: record::Key, value: Vec<u8>) {
 		self.discovery.put_value(key, value);
 	}
+
+	/// Issue a light client request.
+	pub fn request(&mut self, r: light_client_handler::Request<B>) -> Result<(), light_client_handler::Error> {
```
tomaka:
```suggestion
	pub fn light_client_request(&mut self, r: light_client_handler::Request<B>) -> Result<(), light_client_handler::Error> {
```

gavofyork:
sadly, i think we will need to retain a compatibility model since there are at least three other languages (JS, Go, C++) creating a polkadot implementation and thus who we will need to talk to.

romanb:
> sadly, i think we will need to retain a compatibility mode since there are at least three other languages (JS, Go, C++) creating a polkadot implementation and thus who we will need to talk to.

I see. It may still be an option though to convince all these parties to adopt some additional assumptions around the use of multistream-select v1 that are necessary to make the use of it safe with "lazy" negotiation and thus 0-RTT negotiation on substreams, because I don't think it strictly requires any changes to the protocol itself (on the wire), just some additional assumptions on how it is implemented and knowledge of the (fixed set of) protocols being used. See https://github.com/libp2p/rust-libp2p/pull/1212 and https://github.com/libp2p/rust-libp2p/pull/1245 for further detail.

core/network/src/protocol/light_client_handler.rs
rphmeier: @twittner our codebase policy with panickers is to prove that they won't happen or remove them - is it not possible to recover again to a valid protocol state? mild programmer error should not take down a decentralized network

rphmeier:
> The choice would have to be static, of course, since 0-RTT is only really possible when the "dialer" only supports a single protocol at each level of the protocol stack (i.e. for each upgrade, including the initial multistream protocol)

This I don't quite understand. Is there no way to do a few round-trips in the beginning of a connection to negotiate on what kinds of substreams you could do 0-RTT negotiation for?

core/network/src/protocol/light_client_handler.rs
twittner: If you prefer to keep the behaviour going no matter what, we could maybe attempt to reset all internal state.

rphmeier: Is there a way to kill the node gracefully then?

twittner: No that I am aware of. The `NetworkBehaviour` trait offers no such mechanism at the moment AFAIK. This would need to be addressed on libp2p-side first. CC @tomaka.

twittner: I have opened https://github.com/libp2p/rust-libp2p/issues/1256 to discuss this further.

tomaka: It doesn't make sense to have in libp2p a generic mechanism to say "hey there's a bug in the code, please do something".
The behaviour can however emit an event that can correspond to anything we want.

tomaka:
@twittner Could you please update to the master branch? We might merge this soon.

bkchr:
Any update on this?

twittner:
I do not know if there is agreement in principle that the approach taken here would be merged. Other than that there are quite a few conflicts which need to be resolved once more.

tomaka:
I'm still in favour of merging this. Again, this code path isn't used in production yet and we will upgrade over time to see if it works first.
If our network protocol is to eventually make sense, we need either to merge this PR or to rewrite a similar PR later.

twittner:
Updated and rebased. Ready for review again.

client/network/src/protocol/block_requests.rs
```diff
@@ -0,0 +1,338 @@
+//! `NetworkBehaviour` implementation which handles incoming block requests.
```
tomaka: Missing license header

client/network/src/protocol/light_client_handler.rs
```diff
@@ -0,0 +1,1777 @@
+//! [`NetworkBehaviour`] implementation which handles light client requests.
```
tomaka: Missing license header

tomaka:
For the sake of not leaving this PR open for a couple more months, I'm going to merge it.
As discussed above, these code paths aren't used by default yet.

PR #3522, opened 2019-08-30
Add missing child trie rpc on_custom_message handler

cheme:
@twittner find that on_custom_message for child trie read was missing. This PR adds it (see on_remote_read_child_request).

There is also additional test for child trie (requires a slight change to test genesis builder in order
to be able to init some children values).

Possible suggestion/change:
- merge child test with non child test (and use a store value different than heap for this one).
- use a single primitive and storage when using add_extra_data (with optional storagekey and none for top storage).

### DIFF
--- a/core/test-runtime/client/src/lib.rs
+++ b/core/test-runtime/client/src/lib.rs
@@ -23,7 +23,7 @@ pub mod trait_tests;
 mod block_builder_ext;
 
 use std::sync::Arc;
-use std::collections::BTreeMap;
+use std::collections::HashMap;
 pub use block_builder_ext::BlockBuilderExt;
 pub use generic_test_client::*;
 pub use runtime;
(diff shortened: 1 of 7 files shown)


core/test-runtime/client/src/lib.rs
```diff
@@ -98,6 +99,7 @@ pub struct GenesisParameters {
 	support_changes_trie: bool,
 	heap_pages_override: Option<u64>,
 	extra_storage: Vec<(Vec<u8>, Vec<u8>)>,
+	child_extra_storage: BTreeMap<Vec<u8>, Vec<(Vec<u8>, Vec<u8>)>>,
```
bkchr: Why not use a `HashMap<Vec<u8>, Vec<u8>>` directly instead of `Vec<(Vec<u8>, Vec<u8>)>`?

cheme: I am keeping it the same way as existing code, I guess original code is this way because vec is more close to the iterator for appending, using hashmap would make thing simpler (no need to think if it is fine if keys are defined twice, it is), so switching both extra_storage and child_extra_storage could be make sense. I do not think performance really matter here (a few clone can be remove to).
So should I switch?

bkchr: I would switch, making it simpler sounds good :) And yeah, performance does not matter that much here.

PR #3524, opened 2019-08-31
Cleanup of the `state-machine` crate

bkchr:
- Moves `Externalities` and `CodeExecutor` trait so `substrate-primitives`
- Creates a separate file for the state-machine error and implements the trait automatically for all matching types
- Move crate function `new` to the `StateMachine` type itself. 
- `ChildStorageKey` and `NeverOffchainExt` is moved to `substrate-primitives`

I started with this for another pr, but then realized that I probably not need it directly and it is easier to review in a separate pr.

In a future pr, the externalities implementations could probably be moved to its own crate as well.


core/state-machine/src/lib.rs
```diff
@@ -639,49 +330,57 @@ impl<'a, H, N, B, T, O, Exec> StateMachine<'a, H, N, B, T, O, Exec> where
...
-					let (result, _, storage_delta, changes_delta) = self.execute_aux(compute_tx, false, native_call);
-					(result, storage_delta, changes_delta)
+					let res = self.execute_aux(compute_tx, false, native_call);
+					(res.0, res.2, res.3)
 				},
 				ExecutionManager::NativeWhenPossible => {
-					let (result, _was_native, storage_delta, changes_delta) = self.execute_aux(compute_tx, true, native_call);
-					(result, storage_delta, changes_delta)
+					let res = self.execute_aux(compute_tx, true, native_call);
+					(res.0, res.2, res.3)
```
pepyakin: Doesn't look like improvement to me. Shouldn't we just introduce a nominal type for this?

pepyakin:
Can you give a bit more background why, in your opinion, `core/primitives` is a better place for `Externalities`?

bkchr:
1. I'm not quite happy with it in `primitives` as well, but I think it is way better than having it in `state-machine`.
2. `Externalities` is a trait that is used in a lot of places in Substrate and that is the reason why I "promoted" it into `primitives`.
3. As already written in the pr description, I would like to move the implementations we provide for `Externalities` into its own crate. This crate could probably also provide a home for the trait itself :)

core/state-machine/src/error.rs
```diff
@@ -0,0 +1,45 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use std::fmt;
+
+/// State Machine Error bound.
+///
+/// This should reflect WASM error type bound for future compatibility.
```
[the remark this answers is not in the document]
bkchr: IDK :D I just copied it ^^

core/state-machine/src/error.rs
```diff
@@ -0,0 +1,45 @@
...
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
```
gnunicorn: Missing file wide docs.
```suggestion

/// State Machine Errors

```

core/state-machine/src/error.rs
```diff
@@ -0,0 +1,45 @@
...
+pub trait Error: 'static + fmt::Debug + fmt::Display + Send {}
+
+impl<T: 'static + fmt::Debug + fmt::Display + Send> Error for T {}
+
+/// Externalities Error.
+///
+/// Externalities are not really allowed to have errors, since it's assumed that dependent code
+/// would not be executed unless externalities were available. This is included for completeness,
+/// and as a transition away from the pre-existing framework.
+#[derive(Debug, Eq, PartialEq)]
```
gnunicorn: From the comment I feel this should be marked as soon-to-be-deprecated?

core/state-machine/src/lib.rs
gnunicorn: Not sure this needs to be done in this PR. We can also put this into a nice starter-ticket.

core/state-machine/src/error.rs
bkchr: I did not add this comment, I just copied it^^ I'm not sure anyone is working on this.

core/state-machine/src/lib.rs
pepyakin: Good idea1

PR #3528, opened 2019-09-02
Drive by docs fix re minimum_balance

pepyakin:
We use `minimum_balance` unless it refers directly to balances module which uses `existential_deposit`.


bkchr:
> I am just doubtful if we need a version update?

Would not be needed, but yeah, also doesn't hurt anyone.

PR #3533, opened 2019-09-02
Force new era only if 1/3 validators is disabled.

tomusdrw:
Currently validators that are slashed are disabled till the end of the era (i.e. we ignore anything they do).  We also trigger a new era right after the end of current session, which seems completely unnecessary.

This PR forces a new era only if we disable at least one third of the validators to prevent stalling block production.

Marking as WiP as I'm unsure if this is actually needed either or if we should rather never force a new era.


burdges:
We're still missing an implementation for the finality fall back from GRANDPA to BABE.  We therefore cannot actually finalize anything if 1/3 get disabled and thus cannot switch, so this should probably happen lower like 1/5 or 1/6.  cc @AlistairStewart

srml/session/src/lib.rs
```diff
@@ -550,13 +560,33 @@ impl<T: Trait> Module<T> {
...
 	/// Disable the validator of index `i`.
-	pub fn disable_index(i: usize) {
-		T::SessionHandler::on_disabled(i);
+	///
+	/// Returns `true` if this causes a third of validators to be already disabled.
+	pub fn disable_index(i: usize) -> bool {
+		let (fire_event, third_reached) = DisabledValidators::mutate(|disabled| {
+			let i = i as u32;
+			if let Err(index) = disabled.binary_search(&i) {
+				let count = <Validators<T>>::get().len();
```
bkchr:
```suggestion
				let count = <Validators<T>>::decode_len().unwrap_or(0);
```

srml/staking/src/lib.rs
```diff
@@ -1538,10 +1542,11 @@ impl <T: Trait> OnOffenceHandler<T::AccountId, session::historical::Identificati
 				continue;
 			}
 
-			// make sure to disable validator in next sessions
-			let _ = T::SessionInterface::disable_validator(stash);
-			// force a new era, to select a new validator set
-			ForceEra::put(Forcing::ForceNew);
+			// make sure to disable validator till the end of this session
+			if let Ok(true) = T::SessionInterface::disable_validator(stash) {
```
bkchr:
```suggestion
			if T::SessionInterface::disable_validator(stash).unwrap_or(false) {
```

srml/session/src/lib.rs
```diff
@@ -550,13 +560,33 @@ impl<T: Trait> Module<T> {
...
-		T::SessionHandler::on_disabled(i);
+	///
+	/// Returns `true` if this causes a third of validators to be already disabled.
+	pub fn disable_index(i: usize) -> bool {
+		let (fire_event, third_reached) = DisabledValidators::mutate(|disabled| {
+			let i = i as u32;
+			if let Err(index) = disabled.binary_search(&i) {
+				let count = <Validators<T>>::decode_len().unwrap_or(0);
+				disabled.insert(index, i);
+				(true, disabled.len() >= (count + 2) / 3)
```
DemiMarie-parity: What if this overflows?

srml/session/src/lib.rs
```diff
@@ -550,13 +560,33 @@ impl<T: Trait> Module<T> {
...
 	/// Disable the validator of index `i`.
-	pub fn disable_index(i: usize) {
-		T::SessionHandler::on_disabled(i);
+	///
+	/// Returns `true` if this causes a third of validators to be already disabled.
+	pub fn disable_index(i: usize) -> bool {
+		let (fire_event, third_reached) = DisabledValidators::mutate(|disabled| {
+			let i = i as u32;
+			if let Err(index) = disabled.binary_search(&i) {
+				let count = <Validators<T>>::decode_len().unwrap_or(0);
```
kianenigma: nice xD

srml/session/src/lib.rs
kianenigma: I think in many places in the runtime we assume _count of stuff < u32_. Some examples: 
- number of validators
- number of councilors.
- number of election rounds in council. 

Some strong reasons exist to claim all of them will not exceed u32 in the foreseeable future. But if we want to change them (apply your recommendation), I would suggest doing that for all of them and basically sweeping the runtime for any `as` of normal `ops::{*}` operations that could overflow (I don't think it is necessary tbh)

kianenigma: Might be worthwhile to store the _emergency threshold_ of the consensus as a config here? Given that I might --in the future-- want to use this module with a different consensus.

rphmeier:
> we are screwed only if 1/3 are disabled seem to be Grandpa-specific. 

See the FLP impossibility result - 1/3 offline is the boundary for asynchronous liveness. But yes, could be made configurable. Many protocols dealing with a common coin to achieve async liveness can only tolerate 1/5 or 1/7 byzantine.

srml/session/src/lib.rs
tomusdrw: I don't think we should expect a >2G `ValidatorSet`, even if we do I think we will run into other issues much earlier (we deserialize & allocate the entire vec in other parts of the code).

tomusdrw:
Ready for review. Added parametrisation and set to `1/5` for `node-runtime`.

srml/authority-discovery/src/lib.rs
```diff
@@ -166,6 +166,10 @@ mod tests {
 		}
 	}
 
+	parameter_types! {
+		pub const DisabledValidatorsThreshold: Perbill = Perbill::from_percent(33);
```
gavofyork: why 33 here?

gavofyork:
I'm not convinced this is necessary and it's strictly less safe than before.

It makes us more open to liveness attacks where the attacker cannot control the exact timing of each individual node that they take down or where they can't take down all nodes at once.

Basically, by further delaying the replacement of offline or even outright Byzantine validators by upto 24 hours, we reduce our liveness tolerance. In the simplest case with 4 validators, then as soon as as one validator goes offline (accidentally or otherwise), then a Byzantine party has up to 24 hours in which to mount an attack to stall the network. In reality, the sooner we replace known faulty nodes the safer.

CC @alistairstewart

gavofyork:
on ice until I have proof from the research team that this is not strictly less safe.

srml/authority-discovery/src/lib.rs
tomusdrw: I'ts only a test, I don't think it matters at all.

tomusdrw:
@gavofyork afaict, there is nothing currently in code preventing the same validator to be reelected in the next era, so forcing it does not guarantee that the offline/byzantine validators are replaced at all.
Actually before we do elections they are at least disabled, if they are elected again they are considered back online afaict.

burdges:
@gavofyork If we rerunning elections due to slashing, then small shashes can influence the randomness, maybe influencing parachain assignments, elections, smart contract results, etc., so this improves security for those functions.  We cannot do 1/3rd here like I said above, so the issue is balancing safer randomness with any power attackers gain by luck or imprecise attacks.  

I gather:  We currently oversupply validator candidates in Phragmen, which helps ensure all elected validators actually start when an era starts somehow?  We cannot however reuse those extras when one gets kicked because we cannot expect a validator who fails to be elected to remain online?  We must therefore rerun Phragmen to begin a new era, right?  

We could allow extras to make blocks and do secondary validity checks, so that they'd stay online for the profits, and so that inserting them cannot influence the randomness. 

We'll discuss the options some this week.

@tomusdrw I discussed that issues in the new [Slashing with NPoS](https://research.web3.foundation/en/latest/polkadot/slashing/npos/) ([md](https://github.com/w3f/research/blob/master/docs/polkadot/slashing/npos.md)) document, thanks to @rphmeier noticing our slashing issues.  Anytime a nominator N gets slashed due to actions by a validator V then we should: 

> First, we post a slashing transaction to the chain, which drops V from the active validator list by invalidating their session keys, which makes everyone ignore V for the remainder of the era, and also invalidates any future blocks that do not ignore V. We also remove all nomination approval votes by any nominator for V, even those who currently allocate V zero stake.

As a result, V should stay out until nominators manually reinstate them.

> Second, we remove all N's nomination approval votes for future eras. We do not remove N's current nominations for the current era or reduce the stake currently backing other validators. Also we permit N to add new nomination approval votes for future eras during the current era. We also notify N that V cause them to be slashed.

It'd remain possible for nominators to reinstate V in the next era with manual effort though.  We could permit reinstating in the next session even without an era change, but not without manual approval since this approval switches the slashing computation from a max to a sum.

gavofyork:
> We currently oversupply validator candidates in Phragmen

incorrect. you might be thinking of our usage of phragmen for council elections.

> If we rerun elections due to slashing

we don't rerun them immediately; only at the end of the session. it has no effect on the rotation of babe authorities - the next epoch's validator set will be the same. all it means is that we rotate them off a bit faster than we otherwise would. for polkadot, assuming a validator wanted to grief the network in this way, they would essentially rotate themselves off/on every epoch, thereby causing 6x more elections than normal (there are only 6 epochs/sessions in an era). this isn't such a big deal.

> Actually before we do elections they are at least disabled

this currently has no effect.

--

the only question in my mind is: is a 6x increase in elections (that can be caused by only 1 bad validator) sufficiently annoying to bring in this logic that would essentially increase the requirement from only 1 validator to some minimum number of validators.

@AlistairStewart @burdges

burdges:
Anytime we rerun elections we should wait at least one full sessions so that nobody can bias their voting based on the expected randomness.  I gather this is already baked into the code pretty deep, so actually not problem there. 

I doubt a the factor of 6 matters here.  I'll push Al to weigh in.

node/runtime/src/lib.rs
```diff
@@ -211,6 +211,9 @@ impl_opaque_keys! {
 // `SessionKeys`.
 // TODO: Introduce some structure to tie these together to make it a bit less of a footgun. This
 // should be easy, since OneSessionHandler trait provides the `Key` as an associated type. #2858
+parameter_types! {
+	pub const DisabledValidatorsThreshold: Perbill = Perbill::from_percent(20);
```
gavofyork:
```suggestion
	pub const DisabledValidatorsThreshold: Perbill = Perbill::from_percent(17);
```

gavofyork:
Updated to 1/6 of validators as per @AlistairStewart 's suggestion.

Issue #1860, opened 2019-02-25
Improve blockchain import/export


cmichi:
* Improve the status output for the import. At the moment the status output for importing many blocks looks like this:
```
$ cargo run -- import-blocks --dev ./5k-blocks.bin                                                                                                            
    Finished dev [unoptimized + debuginfo] target(s) in 0.82s
     Running `target/debug/substrate import-blocks --dev ./5k-blocks.bin`
2019-02-25 11:19:53 Importing 5000 blocks
2019-02-25 11:19:53 #0
2019-02-25 11:19:53 #1000
2019-02-25 11:19:53 #2000
2019-02-25 11:19:53 #3000
2019-02-25 11:19:53 #4000
```
It seems as if nearly all blocks were immediately imported, but what actually happened is that they were merely added to the import queue and the program will then hang for a long time whilst processing them all and giving no new output. The output should make clear how many were imported and how many are still processed.
* Rate-limit loading from the import file based on backpressure from the import queue, so that the case above is handled better. This was escalated from https://github.com/paritytech/substrate/pull/1807, where @rphmeier mentioned that we do it this way in parity-ethereum. Maybe some inspiration can be taken from there.
* Make it clear in the `--export-blocks` flag documentation that JSON cannot be re-imported.

rphmeier:
@cmichi we should see if it's possible to print output using the informant (maybe modifying it so it has an `Option<NetworkHandle>` or whatever. This uses import notifications internally so it will give the behavior that we want.

tomaka:
The import block part of the issue is fixed.

Remains:

> Make it clear in the --export-blocks flag documentation that JSON cannot be re-imported.

Issue #2587, opened 2019-05-15
Fetch hardcoded transaction factory constants dynamically from runtime


cmichi:
The transaction factory (https://github.com/paritytech/substrate/pull/2481) contains hardcoded constants for minimum balance, the minimum period between blocks, etc. (see `TODO` comments).

These values should should be fetched dynamically from the runtime instead.

PR #3534, opened 2019-09-03
Fixes for allocator + factory + misc improvements

cmichi:
The first commit closes https://github.com/paritytech/substrate/issues/1860.

The bug in the allocator was spotted by @jimpo, great job!


core/executor/src/allocator.rs
```diff
@@ -196,7 +245,21 @@ mod tests {
...
+	}
+
+	#[test]
+	fn should_use_minimum_allocation_size_for_zero_allocations() {
+		// given
+		let mem = MemoryInstance::alloc(Pages(1), None).unwrap();
+		let mut heap = FreeingBumpHeapAllocator::new(mem, 0);
+
+		// when
+		let ptr = heap.allocate(0).unwrap();
```
bkchr: Why do we return a valid pointer for a zero size allocation?

I think we should return a null pointer.

https://stackoverflow.com/questions/1073157/zero-size-malloc

https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html#safety-1

Or some sentinel address, but the caller is already doing something incorrect, when he wants to allocate `0` bytes.

core/executor/src/allocator.rs
```diff
@@ -382,7 +445,43 @@ mod tests {
...
+
+		// when
+		// the `bumper` value is equal to `max_heap_size` here and any
+		// further allocation which would increment the bumper must fail.
+		// we try to allocate 8 bytes here, which will increment the
+		// bumper since no 8 byte item has been allocated+freed before.
+		let ptr = heap.allocate(8);
+
+		// then
+		assert_eq!(ptr.is_err(), true);
```
bkchr:
```suggestion
```

core/executor/src/allocator.rs
```diff
@@ -382,7 +445,43 @@ mod tests {
...
+		if let Err(err) = ptr {
+			match err {
```
bkchr:
```suggestion
			match ptr.unwrap_err() {
```

core/executor/src/allocator.rs
```diff
@@ -382,7 +445,43 @@ mod tests {
...
+		// when
+		// the `bumper` value is equal to `max_heap_size` here and any
+		// further allocation which would increment the bumper must fail.
+		// we try to allocate 8 bytes here, which will increment the
+		// bumper since no 8 byte item has been allocated+freed before.
+		let ptr = heap.allocate(8);
+
+		// then
+		assert_eq!(ptr.is_err(), true);
+		if let Err(err) = ptr {
```
bkchr:
```suggestion
```

core/executor/src/allocator.rs
```diff
@@ -382,7 +445,43 @@ mod tests {
...
+		heap.deallocate(ptr2).expect("failed freeing ptr2");
+		assert_eq!(heap.total_size, 0);
+		assert_eq!(heap.bumper, 64);
+
+		// when
+		// the `bumper` value is equal to `max_heap_size` here and any
+		// further allocation which would increment the bumper must fail.
+		// we try to allocate 8 bytes here, which will increment the
+		// bumper since no 8 byte item has been allocated+freed before.
+		let ptr = heap.allocate(8);
```
bkchr: Wait? We don't reuse any of the old allocations, just because they don't have the correct size? Why?

core/executor/src/allocator.rs
pepyakin: Zero-sized `malloc`ations are legal and they can return any value they wish. The only requirement is that this value should be accepted by `free`. The stackoverflow answer mentions that.

As to `GlobalAlloc`, returning 0 from `alloc` is treated as an out-of-memory error by its callers. So typically the implementations of `GlobalAlloc` that forward calls to some sort of `malloc` they would have to explicitly check the result of malloc for null or for the size of allocation.

In our case, we don't do it, because we rely on the property that `ext_malloc` returns non zero addresses.

>  but the caller is already doing something incorrect, when he wants to allocate 0 bytes.

In my experience I have never seen calls to `alloc` with zero size in Rust. However, I've seen some "weird" code, like `memcpy` copying something from invalid address, which should be generally UB, however, in that case the length of the copy was zero ¯\_(ツ)_/¯ . So I wouldn't be surprised if I see zero sized allocations in Rust.

pepyakin: I think this warrants some docs actually.

bkchr: The rust documentation itself says:
> This function is unsafe because undefined behavior can result if the caller does not ensure that layout has non-zero size.

So, I would not expect any valid rust code to allocate zero bytes.

And further the rust documentation says:
> Returning a null pointer indicates that either memory is exhausted or layout does not meet allocator's size or alignment constraints.

Whatever we do, I would not recommend to actually allocate 8 bytes and return this memory space. We should go for returning some sentinel address, like the address of our base address or something similar. This could be checked in `free` as well.

pepyakin: Ah, good point regarding non-zero layout requirement. But yeah, I'd agree that actual allocation doesn't make a lot of sense. I am not sure if we should fix it within this PR though.

core/executor/src/allocator.rs
```diff
@@ -78,35 +115,45 @@ impl FreeingBumpHeapAllocator {
...
 			return Err(Error::AllocatorOutOfSpace);
 		}
 
 		let list_index = (item_size.trailing_zeros() - 3) as usize;
 		let ptr: u32 = if self.heads[list_index] != 0 {
 			// Something from the free list
 			let item = self.heads[list_index];
+			let ptr = item + PREFIX_SIZE;
+			if ptr + item_size > self.max_heap_size {
+				return Err(Error::AllocatorOutOfSpace);
```
jimpo: This should not be possible given that the free list entry is looked up by the item size. Perhaps this should be an assertion rather than returning an error?

core/executor/src/allocator.rs
jimpo: Right, the current algorithm only reuses an old allocation if it's exactly the right size (rounded up to the next power of two). There's definitely room for improvement by splitting/coalescing free regions.

core/executor/src/allocator.rs
jimpo: nit: `assert!(ptr.is_err())`

Or just match directly on `ptr` and panic on `Ok`.

core/executor/src/allocator.rs
```diff
@@ -15,14 +15,46 @@
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
 //! This module implements a freeing-bump allocator.
-//! See more details at https://github.com/paritytech/substrate/issues/1615.
+//!
+//! The algorithm is as follows:
```
jimpo: Nice docs

core/executor/src/allocator.rs
cmichi: Yeah good point, I had also thought about an assertion when writing the code, but thought it might be bad to panic here. After talking to Basti I've now changed it to an assertion.

core/executor/src/allocator.rs
cmichi: I removed the test again and created a follow-up ticket where further discussion can take place: https://github.com/paritytech/substrate/issues/3561.

cmichi:
All comments have been addressed and the PR updated.

node/cli/src/factory_impl.rs
```diff
@@ -162,7 +162,6 @@ impl RuntimeAdapter for FactoryState<Number> {
 	}
 
 	fn minimum_balance() -> Self::Balance {
-		// TODO get correct amount via api. See #2587.
```
kianenigma: what happened to this? 😀

kianenigma: (I actually wanted to check it out at some point over a weekend or sth)

core/executor/src/allocator.rs
```diff
@@ -337,20 +385,17 @@ mod tests {
...
 
 		// then
 		// there is no room for another half page incl. its 8 byte prefix
-		assert_eq!(ptr2.is_err(), true);
-		if let Err(err) = ptr2 {
-			match err {
-				Error::AllocatorOutOfSpace => {},
-				_ => panic!("Expected out of space error"),
-			}
+		match ptr2.unwrap_err() {
```
kianenigma: why don't we test equality?

kianenigma: I think I see that it is because `error::Error` cannot have `Eq`.

node/cli/src/factory_impl.rs
cmichi: My understanding is that the `ExistentialDeposit::get()` in the line below does already fetch the value from the runtime?

There was previously a hard-coded constant, but this was changed as part of https://github.com/paritytech/substrate/pull/3239. There are some more hard-coded constants though, if you search for `#2587` you'll find more mentions :-).

core/executor/src/allocator.rs
```diff
@@ -15,14 +15,46 @@
...
+//! back the allocation into the linked list from the head.
 
 use crate::error::{Error, Result};
 use log::trace;
 use wasmi::MemoryRef;
 use wasmi::memory_units::Bytes;
 
-// The pointers need to be aligned to 8 bytes.
+// The pointers need to be aligned to 8 bytes. This is because the
+// maximum value type handled by wasm32 is u64; hence we need to make
```
gavofyork: the text following the `;` is a near verbatim copy of the first sentence and can be removed.

core/executor/src/allocator.rs
```diff
@@ -78,35 +115,46 @@ impl FreeingBumpHeapAllocator {
...
 			self.heads[list_index] = FreeingBumpHeapAllocator::le_bytes_to_u32(four_bytes);
-			item + 8
+			ptr
 		} else {
 			// Nothing to be freed. Bump.
-			self.bump(item_size + 8) + 8
+			if self.bumper + PREFIX_SIZE + item_size > self.max_heap_size {
+				return Err(Error::AllocatorOutOfSpace);
+			}
+			self.bump(item_size + PREFIX_SIZE) + PREFIX_SIZE
```
gavofyork: cleaner if `bump` returns a `Result`, then we don't access `self.bumper` from two abstraction levels.

```suggestion
			self.bump(item_size + PREFIX_SIZE)? + PREFIX_SIZE
```

cmichi:
All comments have been addressed and the PR updated.

core/executor/src/allocator.rs
```diff
@@ -78,35 +114,43 @@ impl FreeingBumpHeapAllocator {
...
 		}
 
 		let list_index = (item_size.trailing_zeros() - 3) as usize;
 		let ptr: u32 = if self.heads[list_index] != 0 {
 			// Something from the free list
 			let item = self.heads[list_index];
+			let ptr = item + PREFIX_SIZE;
+			assert!(
+				ptr + item_size <= self.max_heap_size,
+				"Pointer and requested allocation size outside of heap!"
```
pepyakin: Since this is a potential panicker it should have a proof that it doesn't panic.

core/executor/src/allocator.rs
```diff
@@ -78,35 +114,43 @@ impl FreeingBumpHeapAllocator {
 	}
 
 	/// Gets requested number of bytes to allocate and returns a pointer.
+	/// The minimum size which can be allocated is 8 bytes.
```
pepyakin: This sounds like I cannot allocate like 4 or 0 bytes. But this is not the case, right?

cmichi: You can call `allocate(4)` or `allocate(0)`, but there will be a minimum bucket of 8 bytes allocated. So yeah I guess it's phrased ambiguously and removed the comment.

cmichi:
All comments addressed and PR has been updated.

core/executor/src/allocator.rs
pepyakin: That's better, but the best would be to actually write the unambigious comment.

cmichi:
@gavofyork Could you take another look please? There are changes requested by you which should be done.

PR #3539, opened 2019-09-03
BABE: always fetch epoch from runtime

rphmeier:
This is in preparation for a PR that changes the workings of BABE to change epoch based purely on time-slot, not on block height. That would make the current block height caching of epoch invalid - so we strip it out as a first step.

Kusama has been tested to sync with this PR staged.

### DIFF
--- a/core/consensus/babe/src/lib.rs
+++ b/core/consensus/babe/src/lib.rs
@@ -96,7 +96,7 @@ use srml_babe::{
 	BabeInherentData,
 	timestamp::{TimestampInherentData, InherentType as TimestampInherent}
 };
-use consensus_common::{SelectChain, well_known_cache_keys};
+use consensus_common::SelectChain;
 use consensus_common::import_queue::{Verifier, BasicQueue};
 use client::{
 	block_builder::api::BlockBuilder as BlockBuilderApi,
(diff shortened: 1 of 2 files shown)


rphmeier:
Can't reproduce failure locally :\

core/consensus/babe/src/lib.rs
```diff
@@ -929,29 +930,13 @@ impl MaybeSpanEpoch {
 	}
 }
 
-/// Extract current epoch data from cache and fallback to querying the runtime
-/// if the cache isn't populated.
-fn epoch<B, C>(client: &C, at: &BlockId<B>) -> Result<MaybeSpanEpoch, ConsensusError> where
+/// Extract current epoch data from the runtime.
+fn epoch<B, C>(client: &C, at: &BlockId<B>) -> Option<MaybeSpanEpoch> where
```
DemiMarie-parity: I would prefer for such a short function, which is only used once, to be inlined into its call site.

core/consensus/babe/src/lib.rs
```diff
@@ -929,29 +930,13 @@ impl MaybeSpanEpoch {
...
-	B: BlockT,
-	C: ProvideCache<B>,
-{
-	// the epoch that is BABE-valid at the block is not the epoch that is cache-valid at the block
-	// we need to go back for maximum two steps
-	client.cache()
-		.and_then(|cache| cache
-			.get_at(&well_known_cache_keys::EPOCH, at)
-			.and_then(|(_, _, v)| Decode::decode(&mut &v[..]).ok()))
+	epoch_from_runtime(client, at).map(MaybeSpanEpoch::Regular)
```
DemiMarie-parity: This will break light clients.  If this is intentional, that is fine, but it must be documented, both in the rustdoc and in the `--help` output and/or man pages of the final binary.  Additionally, I believe that there should be an issue filed for light client support.

If light client support was already broken, then this should be documented too.

rphmeier: Light clients are tagged as `Experimental` - they aren't guaranteed to work or to continue to work.

core/consensus/babe/src/lib.rs
rphmeier: it's used in 2 places in this file, as well as in the tests module

DemiMarie-parity:
Tests still broken :(

rphmeier:
(I'm gonna close this PR - there are a number of changes we need to make to BABE and I'd prefer to make them as separate PRs to one in-draft PR)

PR #3540, opened 2019-09-03
grandpa: don't require justification of consensus changes on full node

andresilva:
Currently `GrandpaBlockImport` tracks authority set changes in `AuthoritySet`, we require a justification for blocks that enact an authority set change. Additionally, the GRANDPA voter will restrict its votes to blocks that enact an authority set change. This guarantees that said blocks, if finalized, will always have a justification. Additionally, `GrandpaBlockImport` keeps track of consensus changes, this is any kind of changes to the client cache (e.g. a BABE epoch change), and we also require justification for those blocks (added in #3301).

The problem with the current logic is that the sync code will always make the justification requests in-order, therefore if we cannot guarantee that a justification exists for a given block, we might get stuck requesting something endlessly. This is the case with BABE epoch changes since the GRANDPA voter won't limit its votes to these blocks.

This PR changes `GrandpaBlockImport` to only require a justification from sync for authority set changes (not consensus changes). I started by removing all of consensus changes from `GrandpaBlockImport` but this is still required so that we persist a justification for blocks that enact a consensus change (so that we can later on serve it to light clients).

I also changed the logic for finalizing pending requests in sync back to what it was before #3301. Since we only request justifications that **must** be satisfiable then we can also make sure that we finalize the requests in-order (i.e. `finalize_with_ancestors -> finalize`).

### DIFF
--- a/core/network/src/protocol/sync.rs
+++ b/core/network/src/protocol/sync.rs
@@ -815,12 +815,7 @@ impl<B: BlockT> ChainSync<B> {
 	/// with or without errors.
 	pub fn on_justification_import(&mut self, hash: B::Hash, number: NumberFor<B>, success: bool) {
 		let finalization_result = if success { Ok((hash, number)) } else { Err(()) };
-		if !self.extra_justifications.try_finalize_root((hash, number), finalization_result, true) {
-			debug!(target: "sync", "Got justification import result for unknown justification {:?} {:?} request.",
-				hash,
-				number,
-			)
-		}
+		self.extra_justifications.try_finalize_root((hash, number), finalization_result, true);
 		self.is_idle = false;
 	}
 
--- a/core/network/src/protocol/sync/extra_requests.rs
+++ b/core/network/src/protocol/sync/extra_requests.rs
@@ -138,7 +138,7 @@ impl<B: BlockT> ExtraRequests<B> {
 		}
 
 		if best_finalized_number > self.best_seen_finalized_number {
-			match self.tree.finalize_with_ancestors(
+			match self.tree.finalize(
 				best_finalized_hash,
 				best_finalized_number,
 				&is_descendent_of,
--- a/core/finality-grandpa/src/import.rs
+++ b/core/finality-grandpa/src/import.rs
@@ -497,6 +497,8 @@ impl<B, E, Block: BlockT<Hash=H256>, RA, PRA, SC> BlockImport<Block>
 						"Imported unjustified block #{} that enacts authority set change, waiting for finality for enactment.",
 						number,
 					);
+
+					imported_aux.needs_justification = true;
 				}
 
 				// we have imported block with consensus data changes, but without justification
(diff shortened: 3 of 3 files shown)


core/network/src/protocol/sync.rs
```diff
@@ -815,12 +815,7 @@ impl<B: BlockT> ChainSync<B> {
 	/// with or without errors.
 	pub fn on_justification_import(&mut self, hash: B::Hash, number: NumberFor<B>, success: bool) {
 		let finalization_result = if success { Ok((hash, number)) } else { Err(()) };
-		if !self.extra_justifications.try_finalize_root((hash, number), finalization_result, true) {
-			debug!(target: "sync", "Got justification import result for unknown justification {:?} {:?} request.",
-				hash,
-				number,
-			)
-		}
+		self.extra_justifications.try_finalize_root((hash, number), finalization_result, true);
```
andresilva: This can happen if we process the notification for `on_block_finalized` before we actually process the import result. Either way it is fine because we also call `try_finalize_root` on `on_block_finalized` so any necessary bookkeeping is done correctly.

DemiMarie-parity: @andresilva can you add this to the code?

svyatonik: @andresilva I thought [this check](https://github.com/paritytech/substrate/blob/master/core/network/src/protocol/sync/extra_requests.rs#L176) prevents it from happening, no?

core/network/src/protocol/sync/extra_requests.rs
```diff
@@ -138,7 +138,7 @@ impl<B: BlockT> ExtraRequests<B> {
 		}
 
 		if best_finalized_number > self.best_seen_finalized_number {
-			match self.tree.finalize_with_ancestors(
+			match self.tree.finalize(
```
svyatonik: Could you, please, change it back to `finalize_with_ancestors()` for finality proofs? It isn't guaranteed that they're coming in-order. If that PR needs to be merged ASAP, just ping me - I'll create a follow-up PR.

gavofyork: @andresilva @svyatonik i guess this is on-ice until this change is made?

andresilva: @svyatonik Doesn't it have the same guarantees as for justifications? The code that makes the requests is shared with justifications and it will request things in-order as they are "popped" from the tree. I understand that the finality proof for `B` might have a justification for block `B+n`, but for checking in the tree we will be using in `B` anyway right?

svyatonik: @andresilva You're right, `on_block_finalized` is called for every block in finalization path => they'll be in-order anyway. But then probably we should reconsider [this limitation](https://github.com/paritytech/substrate/blob/9b08e7ff938a45dbec7fcdb854063202e2b0cb48/core/client/src/client.rs#L1149) - it could happen that we'll only receive notification for `B+n`, but no notification for `B` => we could end up with the same `UnfinalizedAncestor` error, right?

andresilva: Indeed, I did not remember that. Let's revert back to `finalize_with_ancestors()` then.

PR #3543, opened 2019-09-03
Normalize Crate Names in SRML/Node Template

shawntabrizi:
This PR normalizes some of the imports used across the SRML and the Substrate Node Template.

Ultimately, we want a user who looks at the SRML to be able to "copy paste" as much as possible into the Substrate Node Template and have it work.

* `sr-io` is always imported as `runtime-io`
* `srml-*` is always imported as `*` (ex: `srml-support` is imported as `support`)

I found a few cases of packages which were imported in Cargo.toml, but never used, so I removed them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4210,7 +4210,6 @@ dependencies = [
  "sr-primitives 2.0.0",
  "sr-std 2.0.0",
  "srml-support 2.0.0",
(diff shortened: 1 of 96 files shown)


shawntabrizi:
Is there some app/tool which will clean up toml files of unused imports, format the toml, etc?

PR #3548, opened 2019-09-04
client: only report block import to telemetry if new best

andresilva:
Fixes https://github.com/paritytech/substrate-telemetry/issues/175.

Right now we always report to telemetry when we import a new block. Telemetry assumes that this is the new best block which may not be true (we sync all forks we see). This leads the telemetry UI to show more "re-orgs" than actually exist, and also to always show the highest block number for a peer (instead of best). After this PR telemetry UI should always show the latest block notified by the peer (regardless of number), since that it will always be the best block.

### DIFF
--- a/core/client/src/client.rs
+++ b/core/client/src/client.rs
@@ -879,11 +879,15 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 			fork_choice,
 		);
 
-		telemetry!(SUBSTRATE_INFO; "block.import";
-			"height" => height,
-			"best" => ?hash,
-			"origin" => ?origin
-		);
+		if let Ok(ImportResult::Imported(ref aux)) = result {
+			if aux.is_new_best {
+				telemetry!(SUBSTRATE_INFO; "block.import";
+					"height" => height,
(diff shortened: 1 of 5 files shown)


andresilva:
cc @maciejhirsz

bkchr:
So we will only see finalized blocks? And will not be able to see when the finalization is stalled?

bkchr:
> LGTM as well.
> 
> @bkchr telemetry UI has both best block and finalized block, you might have to enable the relevant columns in the settings if you want to see both numbers / hashes for all nodes.

Yeah I know, but I meant something different. However, I finally looked into the code and found the answer to my question.

andresilva:
@bkchr No, this means that we'll only report to telemetry when we import something that is a new best block (regardless of finality). E.g. with longest chain fork choice rule we will always report to telemetry every time we import a block whose number is higher than the previous best, but e.g. if we import two distinct blocks #10 only one of those will be considered to be a "best" block and therefore we won't report the other one to telemetry. It also means that whenever we import blocks from a different fork, we only report to telemetry if we actually re-org (which makes sense to me).

PR #3549, opened 2019-09-04
Implement `TryInto` for outer events

bkchr:


srml/support/src/event.rs
```diff
@@ -452,31 +452,45 @@ macro_rules! impl_outer_event {
...
-			$(
-				[< $module_name $(_ $generic_instance )? >](
-					$module_name::Event < $( $generic_param )? $(, $module_name::$generic_instance )? >
-				),
-			)*
-		}
-		impl From<$system::Event> for $name {
-			fn from(x: $system::Event) -> Self {
-				$name::system(x)
+			// Workaround for https://github.com/rust-lang/rust/issues/26925 . Remove when sorted.
```
tomusdrw: I know you didn't really add this code, but is this comment actually valid? What exactly is the workaround here?

bkchr: You are right, I removed it :)

bkchr: The comment is so "common" in the Substrate code, that I did not pay attention anymore to it xD

alexsednz:
@bkchr I would have suggested to implement TryFrom and leave TryInto to rstd library. Was there a specific need that you preferred TryInto?

bkchr:
This does not work here.

PR #3550, opened 2019-09-04
srml-contracts: Fix values used for state rent

pepyakin:
While working on integration tests it was discovered that the units that are used within srml-contracts module are disproportional with the other units used in other places and was disproportionally small.


thiolliere:
value seems ok, also not all values for state rent are fixed

pepyakin:
Ah, yeah, right. Only the balance values are needed to be fixed.

PR #3555, opened 2019-09-05
WIP: new nightly srml int test job

TriplEight:
https://github.com/paritytech/srml-contracts-waterfall
Integration tests for srml-contracts.
So the idea is that they would be coded in JS and they would speak with a running node in background
The node would be a default node.
The rest is simple actually. Some rust (nightly most likely also some stable) code, potentially some AssemblyScript
So more precisely something like this:
- submodule update
- building all ink contracts (requires nightly rustc)
- doing yarn (= some npm/js magic)
- downloading a substrate-node
- spinning it up in background exposing RPC interface locally
```
./target/release/substrate purge-chain --dev -y && ./target/release/substrate --dev
```
- then running yarn start, which uses the local RPC

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -355,6 +355,11 @@ test-srml-contracts-waterfall:
     # spinning it up in background exposing RPC interface locally
     # then running yarn start, which uses the local RPC
   script:
+    - echo 'deb http://deb.debian.org/debian/ buster main' > /etc/apt/sources.list.d/wabt.list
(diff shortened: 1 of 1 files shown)


bkchr:
@TriplEight can this be closed?

bkchr:
?

TriplEight:
> ?

no, it's still WIP. @pepyakin asked me to do this.

pepyakin:
oh, interesting. I assumed that this would be implemented as CI for https://github.com/paritytech/srml-contracts-waterfall, not in substrate.

PR #3562, opened 2019-09-06
Add transactions from retracted blocks back to the pool

### LINKED ISSUE #3462: Re-import transactions from retracted blocks to the pool.
Currently the transactions from blocks that become non-canonical are simply ignored, since with current consensus algorithms we expect the number of forks to be close to 0.

With aurababeus forks are going to be much frequent and we should make sure that all transactions end up being included as soon as possible.

arkpar:
Closes #3462

### DIFF
--- a/core/service/src/builder.rs
+++ b/core/service/src/builder.rs
@@ -832,7 +832,7 @@ ServiceBuilder<
 					rpc_extensions
 				))
 			},
-			|h, c, tx| maintain_transaction_pool(h, c, tx),
+			|h, c, tx, r| maintain_transaction_pool(h, c, tx, r),
 			|n, o, p, ns, v| offchain_workers(n, o, p, ns, v),
 			|c, ssb, si, te, tp, ext, ks| start_rpc(&rpc_builder, c, ssb, si, te, tp, ext, ks),
 		)
@@ -885,6 +885,7 @@ pub(crate) fn maintain_transaction_pool<Api, Backend, Block, Executor, PoolApi>(
 	id: &BlockId<Block>,
 	client: &Client<Backend, Executor, Block, Api>,
 	transaction_pool: &TransactionPool<PoolApi>,
+	retracted: &[Block::Hash],
 ) -> error::Result<()> where
 	Block: BlockT<Hash = <Blake2Hasher as primitives::Hasher>::Out>,
 	Backend: client::backend::Backend<Block, Blake2Hasher>,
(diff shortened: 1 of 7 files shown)


core/service/src/builder.rs
```diff
@@ -893,6 +894,16 @@ pub(crate) fn maintain_transaction_pool<Api, Backend, Block, Executor, PoolApi>(
 	Executor: client::CallExecutor<Block, Blake2Hasher>,
 	PoolApi: txpool::ChainApi<Hash = Block::Hash, Block = Block>,
 {
+	// Put transactions from retracted blocks back into the pool.
+	for r in retracted {
+		if let Some(block) = client.block(&BlockId::hash(*r))? {
+			let extrinsics = block.block.extrinsics();
+			if let Err(e) = transaction_pool.submit_at(id, extrinsics.iter().cloned(), true) {
+				warn!("Error re-submitting transactions: {:?}", e);
```
tomusdrw: This might be triggered quite often in case of re-orgs. I would either lower this to `debug` or figure out how to detect transactions that were already included in another chain and print a warning only for other transactions.

core/service/src/builder.rs
```diff
@@ -893,6 +894,16 @@ pub(crate) fn maintain_transaction_pool<Api, Backend, Block, Executor, PoolApi>(
 	Executor: client::CallExecutor<Block, Blake2Hasher>,
 	PoolApi: txpool::ChainApi<Hash = Block::Hash, Block = Block>,
 {
+	// Put transactions from retracted blocks back into the pool.
+	for r in retracted {
```
svyatonik: Does it makes sense to change to `retracted.into_iter().rev()`, so that depending transactions are imported in-order and not moving to/from future/ready queues? I mean if block #5 had a TX1 that provides OUT1 and block #6 had a TX2 that requires OUT1, then if we first import TX2, then it'll be inserted to future queue (because we miss OUT1). And after we import TX1, it'll be moved to ready queue (because TX1 provides OUT1). If we import TX1 first, then both transactions will be inserted in ready queue immediately.

*(and hashes in `retracted` are in reverse order, afaik)*

core/service/src/builder.rs
arkpar: `submit_at` pushes a list of transactions. I believe it only fails when there's something fundamentally wrong with the pool. Otherwise it returns a list of individual results that we ignore here.

core/service/src/builder.rs
arkpar: Changes `retracted` in notification to be in ascending order, as documented.

PR #3563, opened 2019-09-06
Add an RPC method for calling a contract.

pepyakin:
This PR adds a new custom RPC to node that allows users to call a contract with the current state and receive the output buffer. This is unlike to the executive, where it is not possible to return a result of contract execution.

This feature will help us in writing srml-contracts [integration tests](https://github.com/pepyakin/srml-contracts-waterfall), right now we are checking the result of execution by querying the hardcoded storage entries which is not ideal, calling getters would be more convenient.

I believe that the runtime behavior didn't change, hence incrementing only `impl_version`.

TODO / Follow-ups:

- [ ] Tests. This is tricky one, because the setup might get quite complicated.
- [ ] A similar method for instantiation of a contract.
- [ ] Probably another for `put_code`. That would allow to check if a contract is deployable locally.


node/rpc/src/contracts.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+impl<C> Contracts<C> {
+	/// Create new `Contracts` with the given reference to the client.
+	pub fn new(client: Arc<C>) -> Self {
+		Contracts { client }
+	}
+}
+
+impl<C> ContractsApi for Contracts<C>
```
tomusdrw: The implementation for all `RuntimeApi`s will pretty much look exactly the same, I'll have a look on ways of automatically deriving this.

node/rpc/src/contracts.rs
```diff
@@ -0,0 +1,90 @@
...
+		&self,
+		origin: AccountId,
+		dest: AccountId,
+		value: Balance,
+		gas_limit: u64,
+		input_data: Vec<u8>,
+	) -> Result<ContractExecResult> {
+		let api = self.client.runtime_api();
+		let best = self.client.info().best_hash;
+		let at = BlockId::hash(best);
```
tomusdrw: I wonder if it makes sense to call the contract using particular block hash to ensure "transactionality" (i.e. you make a bunch of contract calls and want them to be executed exactly on the same `block_hash`).

I'd be in favor of adding an `Option<Block::Hash>` parameter to the RPC, and changing all current positional parameters to a struct.

pepyakin: Yeah, of course it makes! 

I thought to add it afterwards, but can add it here as well.

core/rpc-primitives/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
```
tomusdrw: Maybe let's move this crate to `core/rpc/primitives` instead of `core/rpc-primitives`?

node/rpc/src/contracts.rs
```diff
@@ -0,0 +1,114 @@
...
+	AccountId, Balance, Block, BlockId, ContractExecResult, ContractsApi as ContractsRuntimeApi,
+};
+use sr_primitives::traits::{
+	self,
+	Block as BlockT,
+};
+use rpc_primitives::number;
+
+/// A struct that encodes RPC parameters required for a call to a smart-contract.
+#[derive(Serialize, Deserialize)]
```
tomusdrw:
```suggestion
#[derive(Serialize, Deserialize)]
#[serde(rename_all="camelCase")]
#[serde(deny_unknown_fields)]
```

core/rpc-primitives/src/lib.rs
pepyakin: keeping the same crate name? k!

PR #3565, opened 2019-09-06
Better testing for lock types in staking/democracy.

kianenigma:


srml/executive/src/lib.rs
```diff
@@ -595,10 +604,13 @@ mod tests {
...
 					&1,
 					110,
-					10,
+					Bounded::max_value(),
 					lock,
 				);
-				let xt = sr_primitives::testing::TestXt(sign_extra(1, 0, 0), Call::transfer(2, 10));
+				let xt = sr_primitives::testing::TestXt(
+					sign_extra(1, 0, 0),
+					Call::System(SystemCall::remark(vec![1u8]))
```
bkchr:
```suggestion
					Call::System(SystemCall::remark(vec![1u8])),
```

srml/executive/src/lib.rs
```diff
@@ -609,15 +621,8 @@ mod tests {
...
-					assert_eq!(
-						Executive::apply_extrinsic(xt).unwrap(),
-						Err(DispatchError {
-							module: None,
-							error: 0,
-							message: Some("account liquidity restrictions prevent withdrawal"),
-						}),
-					);
-					// but tx fee has been deducted. the transaction failed on transfer, not on fee.
+					assert!(Executive::apply_extrinsic(xt).is_ok());
```
bkchr:
```suggestion
					assert!(Executive::apply_extrinsic(xt).map(Result::is_ok).unwrap_or(false));
```

Before, you checked if the transaction itself returned an error. If you want to know if the transaction could be executed successfully, you need to check that the result is `Ok(Ok(()))`.

PR #3567, opened 2019-09-06
First step for generating host externals out of the function definition in `sr-io`

bkchr:
This pr marks the first step in supporting of automatic host externals generation out of the definition in `sr-io`. To accomplish this, we begin in this PR to clean up the interface of the `WasmExecutor`. 

1. We introduce a new crate `substrate-wasm-interface`. This crates adds types and traits for communicating between the host and the wasm instance. As a bonus point, this is already interpreter/compiler agnostic, to support different wasm engines in the future.

2. Rewrite of the wasm utils macros to use the new facilities of `substrate-wasm-interface`. The macros will be replaced by proc-macros in some of the next prs.

3. We now distinguish between substrate and inherent externals. The inherent externals are stuff like sandbox, malloc, free, that are always being provided by the wasm executor. This split up was required, as I don't wanted to expose all the stuff required for sandboxing in the `ExternalsContext` trait. For `malloc` and `free`, I choose them as inherent, as they are first not provided by `sr-io` and writing the implementation of this two functions is not that complicated at all.

4. Based on feedback of @jimpo, this also introduces `WordSize` as the type that represents the wasm word size (usize)  and `Pointer`. The pointer is an abstraction over a wasm pointer, to make working with wasm pointers more safe in the host runtime and less error prone.


core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+	pub fn new(ptr: u32) -> Self {
+		Self {
+			ptr,
+			_marker: Default::default(),
+		}
+	}
+
+	/// Calculate the offset from this pointer.
+	///
+	/// `offset`is in units of `T`. So, `3` means `3 * mem::size_of::<T>()` as offset to the pointer.
```
pepyakin:
```suggestion
	/// `offset` is in units of `T`. So, `3` means `3 * mem::size_of::<T>()` as offset to the pointer.
```

core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+	fn resolve_function(name: &str, signature: &Signature) -> Result<Option<FunctionRef>, String>;
+
+	/// Returns the number of host functions this type provides.
+	fn function_count() -> usize;
+
+	/// Execute the function at the given index.
+	///
+	/// - `index` - Is equal to the index given to `FunctionRef`.
+	/// - `args` - The arguments given to the function.
+	/// - `context` - Provides access to the allocator and memory of the memory wasm instance.
```
pepyakin:
```suggestion
```

core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+	fn deallocate_memory(&mut self, ptr: Pointer<u8>) -> Result<(), String>;
+}
+
+/// Something that provides implementations for host functions.
+pub trait Externals {
+	/// Try to resolve the function with the given name and signature.
+	fn resolve_function(name: &str, signature: &Signature) -> Result<Option<FunctionRef>, String>;
+
+	/// Returns the number of host functions this type provides.
+	fn function_count() -> usize;
```
pepyakin: Why would we need the count?

core/executor/src/wasm_executor.rs
```diff
@@ -1677,6 +1635,8 @@ mod tests {
 		calldata.extend_from_slice(key.public().as_ref());
 		calldata.extend_from_slice(sig.as_ref());
 
+		println!("SIG: {:?}", sig.0.as_ref());
```
pepyakin: stray println

core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+		index: usize,
+		args: A,
+		context: &mut dyn ExternalsContext,
+	) -> Result<Option<Value>, String>;
+}
+
+/// Something that provides implementations for inherent host functions.
+///
+/// Will be used interanlly by Substrate to provide functions like `malloc`, `free` etc.
+pub trait InherentExternals {
```
jimpo: I don't like the division between `Externals` and `InherentExternals`.

My preference would be that there be two (or more) implementations of the same interface, `Externals`, and `FunctionExecutor` keeps a `Vec<Box<dyn Externals>>` of externals. Then the `ModuleImportResolver` impl of `FunctionExecutor` would scan through all of them for the function it wants.

core/executor/src/lib.rs
```diff
@@ -47,6 +47,8 @@ pub use runtime_version::{RuntimeVersion, NativeVersion};
 pub use codec::Codec;
 #[doc(hidden)]
 pub use primitives::Blake2Hasher;
+#[doc(hidden)]
+pub use wasm_interface;
```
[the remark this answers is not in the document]
bkchr: This is default behavior when you want to use something from a macro, but don't want to show the outside that you have exported it. This will be removed in one of the followings prs anyway.

core/wasm-interface/src/lib.rs
bkchr: Currently we only have these 2 externals. When we come to the point that we have `N` externals. It will be used to call into the correct `execute_function`.

core/sr-std/without_std.rs
```diff
@@ -19,7 +19,7 @@
 pub extern crate alloc;
 
 extern "C" {
-	fn ext_malloc(size: usize) -> *mut u8;
+	fn ext_malloc(size: u32) -> *mut u8;
```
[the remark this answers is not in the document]
bkchr: I don't want to see any non-fixed sized types in an interface. You are right that this is the same, but I better see this being concrete.

core/executor/src/wasm_utils.rs
```diff
@@ -16,112 +16,47 @@
...
-	const VALUE_TYPE: ValueType = ValueType::I32;
-	fn to_runtime_value(self) -> RuntimeValue { RuntimeValue::I32(self as isize as i32) }
-}
-
 /// Converts arguments into respective WASM types.
 #[macro_export]
 macro_rules! convert_args {
 	() => ([]);
-	( $( $t:ty ),* ) => ( [ $( { use $crate::wasm_utils::ConvertibleToWasm; <$t>::VALUE_TYPE }, )* ] );
+	( $( $t:ty ),* ) => ( [ $( <$t as $crate::wasm_interface::IntoValue>::VALUE_TYPE, )* ] );
```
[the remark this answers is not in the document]
bkchr: You do this in macros, to make sure they are usable from the outside. Currently it is not really required, but is good macro writing :)

core/executor/src/wasm_executor.rs
```diff
@@ -16,99 +16,115 @@
...
 
-impl WritePrimitive<u32> for MemoryInstance {
-	fn write_primitive(&self, offset: u32, t: u32) -> Result<()> {
+impl WritePrimitive<u32> for &mut dyn ExternalsContext {
+	fn write_primitive(&mut self, ptr: Pointer<u32>, t: u32) -> std::result::Result<(), String> {
 		use byteorder::{LittleEndian, ByteOrder};
 		let mut r = [0u8; 4];
 		LittleEndian::write_u32(&mut r, t);
-		self.set(offset, &r).map_err(Into::into)
+		self.write_memory(ptr.cast(), &r)
```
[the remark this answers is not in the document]
bkchr: Basically the same.

core/executor/src/wasm_executor.rs
```diff
@@ -16,99 +16,115 @@
...
-	fn read_memory(&self, ptr: u32, len: u32) -> Result<Vec<u8>> {
-		self.memory.get(ptr, len as usize).map_err(Into::into)
+	fn read_memory(&self, ptr: Pointer<u8>, len: WordSize) -> Result<Vec<u8>> {
+		self.memory.get(ptr.into(), len as usize).map_err(Into::into)
+	}
+}
+
+impl ExternalsContext for FunctionExecutor {
+	fn read_memory_into(&self, address: Pointer<u8>, dest: &mut [u8]) -> std::result::Result<(), String> {
+		self.memory.get_into(address.into(), dest).map_err(|e| format!("{:?}", e))
```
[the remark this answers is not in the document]
bkchr: I would assume that the memory gives a more appropriate error?

core/executor/src/wasm_executor.rs
```diff
@@ -121,1201 +137,1136 @@ fn deadline_to_timestamp(deadline: u64) -> Option<offchain::Timestamp> {
...
+			kind: u32,
+			key: Pointer<u8>,
+			key_len: WordSize,
+			old_value: Pointer<u8>,
+			old_value_len: WordSize,
+			new_value: Pointer<u8>,
+			new_value_len: WordSize,
+		) -> u32 {
+			let kind = offchain::StorageKind::try_from(kind)
+					.map_err(|_| "storage kind OOB while ext_local_storage_compare_and_set: wasm")?;
```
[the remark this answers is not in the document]
bkchr: What kind of indentation do you expect?

core/executor/src/wasm_executor.rs
```diff
@@ -121,1201 +137,1136 @@ fn deadline_to_timestamp(deadline: u64) -> Option<offchain::Timestamp> {
...
+		}
+
+		ext_local_storage_get(
+			kind: u32,
+			key: Pointer<u8>,
+			key_len: WordSize,
+			value_len: Pointer<u32>,
+		) -> Pointer<u8> {
+			let kind = offchain::StorageKind::try_from(kind)
+					.map_err(|_| "storage kind OOB while ext_local_storage_get: wasm")?;
```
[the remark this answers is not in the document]
bkchr: What kind of indentation do you expect?

core/wasm-interface/src/lib.rs
bkchr: If we move more stuff into the externals, we could probably remove the `InherentExternals` in the future, but I would really like to leave this for the future :) 

The general design is already suited for multiple externals, we will just add a generic Parameter `E: Externals` that can hold a tuple of externals. When you then call one function of this generic parameter, it will automatically cascade down to all the tuple elements. (Comparable to `OnInitialize` in the runtime)

core/executor/src/wasm_executor.rs
```diff
@@ -121,1201 +137,1136 @@ fn deadline_to_timestamp(deadline: u64) -> Option<offchain::Timestamp> {
...
-				primitives::hexdisplay::ascii_format(&key),
-				HexDisplay::from(&value),
-				HexDisplay::from(&key)
-			);
+	}
+}
+
+impl_wasm_host_interface! {
+	#[inherent_externals]
+	impl FunctionExecutor where this {
```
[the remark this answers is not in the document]
bkchr: I wanted to do the same, but it would have required a much more complicated declarative macro. I want to change the macro to a proc-macro anyway and then it will be better :)

bkchr:
> In general, I really like the stuff where it is going. I gave a brief Here are some notes:
> # Terminlogy
> 
> Let's avoid the term externals. Using it in wasmi was a mistake since that was a common term already used in few different places. I think we should use th "host functions" term.
> 
> I also not super fan of the term `inherent` since it is already used for `inherent extrinsics` and there are also `inherent methods` coming from Rust. This use of inherent seems to be not connected with the former two, or at least I cannot see the connection.

Okay, renamed to `HostFunctions`.

I named it `inherent`, as these are functions that are inherently present when using the `WasmExecutor`, but I go with you that this should be changed. I don't wanted to touch the `Externalities` in this pr as well. I will move the sandboxing to the `Externalities` in a future pr and remove this extra distinction between inherent and "normal" host functions.  

> # Inherent externals
> 
> TBH, I am not sure if that's division is worthwhile. You rightly noted that `malloc`/`free` are not provided by `sr-io` and they are provided by `sr-std`. However, I think this is just happened by overlook and in my opinion they should be merged with the other host function definitions.
> 
> I would also want to hear as for why you want to extract sandboxing capabilities in a separate group of host functions? I thought maybe it could be related to Cumulus and Polkadot validation function environment, but this shouldn't be the reason since polkadot validation function should provide the sandboxing capabilities (or otherwise wasm smart-contract parachains won't be a thing, AFAIU).

Hopefully is already being answered above :) 

> # Wasm Interface (the most important point)
> 
> > As a bonus point, this is already interpreter/compiler agnostic, to support different wasm engines in the future.
> 
> I'd argue with this: so far this crate looks really biased towards wasmi. It would be suboptimal for compiler-based engines to use this interface.
> 
> For example in wasmtime the most likely situation would be that externals (or more accurate, host functions) are bound through pointers to `extern "C"` functions. So all arguments are passed in the right places already and won't require any more fiddling with them and there is no need in dispatching switch that calls a specific function by `index`.
> 
> So there is no need in modelling `Value` and conversion machinery. `FunctionRef` assumes that the functions are referred by index, however this is wasmi specific, and an untyped function pointer is a possibility as well.
> 
> We could workaround this, by producing proper definitions would be produced by the high-level macro, i.e. `impl_wasm_host_interface`, but then why would we devote an entire crate with such a general name for wasmi specific definitions? This doesn't seem the cleanest approch and I'd reengineer the bindings to suit to (at least) the both engines.

So, I looked in the WIP code of @jimpo for integrating wasmtime and the interfacing between the host and wasm looks comparable to wasmi? Looking at the docs: https://docs.rs/wasmtime-api/0.3.0/wasmtime_api/ it also looks very familiar?

core/executor/src/wasm_executor.rs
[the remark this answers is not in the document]
jimpo: One more level of indentation than the line above. There is one extra tab.

core/wasm-interface/src/lib.rs
jimpo: What else needs to be moved into externals? The `HostFunctions` trait is nearly identical to the `InherentHostFunctions` trait, except that one takes a context parameter and in the other one, the context is `self`. (And the passed-in context turns out to be `self` anyway`).

I really don't understand the reason for two separate traits and code paths.

jimpo:
@bkchr No, the wasmtime host function interface is quite different. Currently, [this](https://github.com/jimpo/substrate/blob/wasmtime-bench/core/executor/src/wasmtime_utils.rs#L80) is the macro it is using to make host calls (where it defines an extern "C" function and registers it) and I'm likely going to change to using [Extern](https://github.com/CraneStation/wasmtime/blob/master/wasmtime-api/src/externals.rs#L14).

core/executor/src/wasm_executor.rs
```diff
@@ -19,99 +19,115 @@
...
+	}
+}
+
+impl HostFunctionsContext for FunctionExecutor {
+	fn read_memory_into(&self, address: Pointer<u8>, dest: &mut [u8]) -> std::result::Result<(), String> {
+		self.memory.get_into(address.into(), dest).map_err(|e| format!("{:?}", e))
+	}
+
+	fn write_memory(&mut self, address: Pointer<u8>, data: &[u8]) -> std::result::Result<(), String> {
+		self.memory.set(address.into(), data).map_err(|e| format!("{:?}", e))
```
tomusdrw: maybe:
```
fn debug<D: Debug>(obj: D) -> String { format!("{:?}", obj) }
```
and then just
```
.map_err(debug)`
```
?

core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+		)*
+	}
+}
+
+impl_into_and_from_value! {
+	u32, I32,
+	i32, I32,
+	u64, I64,
+	i64, I64,
+	*const T, <T>, I32,
```
jimpo: Is this and the below still necessary now that we have `Pointer<T>`?

core/executor/src/wasm_executor.rs
```diff
@@ -124,1201 +140,1135 @@ fn deadline_to_timestamp(deadline: u64) -> Option<offchain::Timestamp> {
...
+		ext_sandbox_instantiate(
+			dispatch_thunk_idx: u32,
+			wasm_ptr: Pointer<u8>,
+			wasm_len: WordSize,
+			imports_ptr: Pointer<u8>,
+			imports_len: WordSize,
+			state: u32,
+		) -> u32 {
+			let wasm = this.read_memory(wasm_ptr, wasm_len)
+				.map_err(|_| "OOB while ext_sandbox_instantiate: wasm")?;
```
tomusdrw: Could we add the function name to the error message inside the macro? Repeating it now is really tedious. Maybe an idea for the `proc-macro` you've mentioned earlier.

tomusdrw: Also can the functions just return `Result<_, impl Debug>`? To simplify error handling?

core/sr-io/src/lib.rs
```diff
@@ -279,7 +279,7 @@ export_api! {
 		/// Submit transaction to the pool.
 		///
 		/// The transaction will end up in the pool.
-		fn submit_transaction<T: codec::Encode>(data: &T) -> Result<(), ()>;
+		fn submit_transaction(data: Vec<u8>) -> Result<(), ()>;
```
tomusdrw: This is a breaking change, does it also mean we won't be able to do generic functions in the future on the `sr-io` level?
I was actually thinking of making it properly typed (i.e. we expect `Block::Extrinsic` here).

bkchr:
> @bkchr No, the wasmtime host function interface is quite different. Currently, [this](https://github.com/jimpo/substrate/blob/wasmtime-bench/core/executor/src/wasmtime_utils.rs#L80) is the macro it is using to make host calls (where it defines an extern "C" function and registers it) and I'm likely going to change to using [Extern](https://github.com/CraneStation/wasmtime/blob/master/wasmtime-api/src/externals.rs#L14).

Looking at the `Extern` enum, I don't see why the current structure does not maps? The `Func` expects the function arguments as `ValType` and that looks very familiar to the one in wasmi. The only difference is the `callable` object, but that should be genertable from the information we have.
I will change the `HostFunctions` trait to return all available functions. This should make it compatible between wasmtime and wasmi.

core/wasm-interface/src/lib.rs
bkchr: Good point.

core/wasm-interface/src/lib.rs
bkchr: As written in the comments to the pr, we/I can remove this in a future pr. I just don't wanted to touch the `Externalities` trait in this pr. When we move the `sandbox` stuff into the `Externalities`, we can remove the `InherentHostFunctions` trait.

core/sr-io/src/lib.rs
bkchr: That is no breaking change (in the sense that it does not breaks any chain). We don't change the `ext_submit_transaction` in this case.

The problem is that you can not generate c functions that take generics. In the next pr, I will start to clean up the usage of `generics` in the `sr-io` interface functions. We really should move this into the structs. 

We have for example `ordered_trie_root` and expose the function `ext_blake2_256_enumerated_trie_root`. The original name should directly be `blake2_256_ordered_trie_root` and then we have a trait in `sr-primitives` that provides the ordered trie root function and the implementation for `Blake2Hasher` calls the correct function.

`submit_transaction` should follow the same pattern and the trait should be implemented directly for the extrinsic types.

core/executor/src/wasm_executor.rs
bkchr: Yeah, when all this stuff is auto generated, I will also attach the function names to the error messages automatically.

core/executor/src/wasm_executor.rs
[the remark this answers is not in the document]
jimpo: But we could get rid of the `byteorder` dependency from this crate most likely.

core/executor/src/lib.rs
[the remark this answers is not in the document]
pepyakin: Ah, yeah, would be better with a comment but if you are sure that it would be removed then it is fine!

core/executor/src/wasm_executor.rs
```diff
@@ -19,100 +19,243 @@
...
+	fn write_memory(&mut self, ptr: Pointer<u8>, data: &[u8]) -> Result<()> {
+		self.memory.set(ptr.into(), data).map_err(Into::into)
+	}
+	fn read_memory(&self, ptr: Pointer<u8>, len: WordSize) -> Result<Vec<u8>> {
+		self.memory.get(ptr.into(), len as usize).map_err(Into::into)
+	}
+}
+
+impl FunctionContext for FunctionExecutor {
+	fn read_memory_into(&self, address: Pointer<u8>, dest: &mut [u8]) -> std::result::Result<(), String> {
```
pepyakin: Nit: There are way more `std::result::Result<_, _>` than just `Result`. I'd prefer to properly alias one or both results.

core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+
+mod wasmi_impl;
+
+/// Value types supported by Substrate on the boundary between host/Wasm.
+#[derive(Copy, Clone, PartialEq, Debug, Eq)]
+pub enum ValueType {
+	/// An `i32` value type.
+	I32,
+	/// An `i64` value type.
+	I64,
```
pepyakin: I assume that it was actually decided that we do not longer want to support host functions that return or receive float values.

core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+		Self {
+			ptr,
+			_marker: Default::default(),
+		}
+	}
+
+	/// Calculate the offset from this pointer.
+	///
+	/// `offset` is in units of `T`. So, `3` means `3 * mem::size_of::<T>()` as offset to the pointer.
+	pub fn offset(self, offset: u32) -> Self {
```
pepyakin: This function seems to be a bit of a footgun: `size_of` is the size of of a type on the host machine while this type is supposed to be used in the wasm context.

core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+		Self {
+			ptr: self.ptr + offset * mem::size_of::<T>() as u32,
```
pepyakin: This could overflow and potentially panic.

core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+pub enum Value {
+	/// An `i32` value.
+	I32(i32),
+	/// An `i64` value.
+	I64(i64),
+}
+
+/// Type to represent a pointer in wasm at the host.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub struct Pointer<T> {
```
pepyakin: TBH, I am not sure about this `T`. Why won't we just assume that it is a `void *` pointer? Esp. in the light of the flaw of `offset`?

pepyakin: On another note, the derives have the bounds on `T` even though they are not required.

core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+		args: &mut dyn Iterator<Item=Value>,
+	) -> Result<Option<Value>, String>;
+}
+
+/// Context used by `Function` to interact with the allocator and the memory of the wasm instance.
+pub trait FunctionContext {
+	/// Read memory from `address` into a vector.
+	fn read_memory(&self, address: Pointer<u8>, size: WordSize) -> Result<Vec<u8>, String> {
+		let mut vec = Vec::with_capacity(size as usize);
+		unsafe { vec.set_len(size as usize); }
```
pepyakin: This lacks the safety proof and it would be problematic to guarantee *here* that this is safe since this depends on the implementation of `read_memory_into`, namely:

1. that it doesn't read the provided buffer,
2. that it writes `size` bytes.

It's UB if the implementer of `FunctionContext` doesn't ensure these invariants. That means in turn that `read_memory_into` should be marked `unsafe`.

Are you sure that `unsafe` block is required here? I think chances are that the zero initialization will be actually elided.

core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+/// Sandbox memory identifier.
+pub type MemoryId = u32;
+
+/// Something that provides access to the sandbox.
+pub trait Sandbox {
+	/// Get sandbox memory from the `memory_id` instance at `offset` into the given buffer.
+	fn memory_get(
+		&self,
+		memory_id: MemoryId,
+		offset: WordSize,
```
pepyakin: I think this should be actually `Pointer` if followed the logic of the other usages of it

core/wasm-interface/src/lib.rs
bkchr: Before we have done this manually. Working with pointers is always a footgun :D We could enforce that the user just can give a `ValueType`.

core/wasm-interface/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+		&mut self,
+		memory_id: MemoryId,
+		offset: WordSize,
+		val_ptr: Pointer<u8>,
+		val_len: WordSize,
+	) -> Result<u32, String>;
+	/// Delete a memory instance.
+	fn memory_teardown(&mut self, memory_id: MemoryId) -> Result<(), String>;
+	/// Create a new memory instance with the given `initial` size and the `maximum` size.
+	fn memory_new(&mut self, initial: WordSize, maximum: WordSize) -> Result<MemoryId, String>;
```
pepyakin: Making those `WordSize` might be a bit confusing: they actually represent a number of pages.

core/wasm-interface/src/wasmi_impl.rs
```diff
@@ -0,0 +1,73 @@
...
+		}
+	}
+}
+
+impl From<wasmi::RuntimeValue> for Value {
+	fn from(value: wasmi::RuntimeValue) -> Self {
+		match value {
+			wasmi::RuntimeValue::I32(val) => Self::I32(val),
+			wasmi::RuntimeValue::I64(val) => Self::I64(val),
+			o => panic!("`wasmi::Value` `{:?}` not supported by Substrate.", o)
```
pepyakin: This should be a proof but I am not sure how to provide this...

core/wasm-interface/src/wasmi_impl.rs
```diff
@@ -0,0 +1,73 @@
...
+		}
+	}
+}
+
+impl From<wasmi::ValueType> for ValueType {
+	fn from(value: wasmi::ValueType) -> Self {
+		match value {
+			wasmi::ValueType::I32 => Self::I32,
+			wasmi::ValueType::I64 => Self::I64,
+			o => panic!("`wasmi::ValueType` `{:?}` not supported by Substrate.", o)
```
pepyakin: See this https://github.com/paritytech/substrate/pull/3567/files#r322414210

core/wasm-interface/src/wasmi_impl.rs
```diff
@@ -0,0 +1,73 @@
...
+		let args = sig.args.iter().map(|a| (*a).into()).collect::<Vec<_>>();
+		wasmi::Signature::new(args, sig.return_value.map(Into::into))
+	}
+}
+
+impl From<&wasmi::Signature> for Signature {
+	fn from(sig: &wasmi::Signature) -> Self {
+		Signature::new(
+			sig.params().into_iter().copied().map(Into::into).collect::<Vec<_>>(),
+			sig.return_type().map(Into::into),
```
pepyakin: Doesn't it mean that if the runtime declares a signature with referring to a float type this is going to panic?

core/wasm-interface/src/lib.rs
bkchr: Isn't that just the offset into the data? Offsets are never given as pointers.

core/wasm-interface/src/wasmi_impl.rs
bkchr: Fixed.

core/wasm-interface/src/lib.rs
pepyakin: well, all wasm pointers are offset from the start of a linear memory : ) This is one of those.

core/executor/src/wasm_executor.rs
```diff
@@ -124,1201 +270,1066 @@ fn deadline_to_timestamp(deadline: u64) -> Option<offchain::Timestamp> {
...
-		};
 
-		debug_trace!(
-			target: "wasm-trace", "*** Getting child storage: {} -> {} == {}   [k={}]",
-			primitives::hexdisplay::ascii_format(&storage_key),
-			if let Some(_preimage) = this.hash_lookup.get(&key) {
-				format!("%{}", ::primitives::hexdisplay::ascii_format(&_preimage))
+		ext_storage_changes_root(
+			parent_hash_data: Pointer<u8>,
+			_len: WordSize,
```
pepyakin: Why don't we check for the length anymore?

core/executor/src/wasm_executor.rs
```diff
@@ -124,1201 +270,1066 @@ fn deadline_to_timestamp(deadline: u64) -> Option<offchain::Timestamp> {
...
+		}
+
+		ext_chain_id() -> u64 {
+			Ok(runtime_io::chain_id())
+		}
+
+		ext_twox_64(data: Pointer<u8>, len: WordSize, out: Pointer<u8>) {
+			let result: [u8; 8] = if len == 0 {
+				let hashed = twox_64(&[0u8; 0]);
+				hashed
```
pepyakin: So this effectively removes caching logic. Did you find out why it was there and whether we can remove it? If so, I'd be curious to hear

core/executor/src/wasm_executor.rs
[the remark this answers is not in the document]
bkchr: Done.

core/executor/src/wasm_executor.rs
bkchr: We could readd the check, but for all the other places where we take fixed size data types we don't have this check. If you look into the `sr-io` api that calls this function, it only takes an array of 32 elements.

core/executor/src/wasm_executor.rs
bkchr: I did not understand why this logic is required? The native api also does not have any caching logic :woman_shrugging: 

When we want to move to generating all this interface stuff, we can not have all these edge cases anyway.

pepyakin: > I did not understand why this logic is required? The native api also does not have any caching logic

Yeah, I am confused by this as well and that is the reason why I am curious. I also cannot come up why it was there in the first place.

And at the same time, I am a bit of unsure that it was there for nothing. If you want I can try to find it out myself

core/executor/src/wasm_executor.rs
pepyakin: Fine by me! Although in this case we should remove `len` param.

core/executor/src/wasm_executor.rs
jimpo: Do you mean the `hash_lookup` stuff? I think that was *just* for `debug_trace` readability. I agree with removing it.

pepyakin: I just did some git archeology, and it seems that `hash_lookup` was introduced [for debugging](https://github.com/paritytech/substrate/commit/15f34526bc31161da6f27ef2a03fb00482993f86). I am good with removing it

bkchr: Yeah, as @jimpo highlighted it was mostly used in debug traces. All tests are passing, so I really hope nobody added this without any test.

core/executor/src/wasm_executor.rs
bkchr: Yeah, but this would break Kusama as we could not import the wasm blob anymore.

pepyakin: Let's leave a TODO then?

Issue #3566, opened 2019-09-06
Prevent daisy-chained stash/controller accounts


joepetrowski:
When `staking::bond` is called, it only checks that the stash is not already a stash and the controller is not already a controller. This allows you to make the controller a stash account with another downstream controller.

Account roles should be exclusionary.

burdges:
We should discourage making them the same, but it could be done correctly, so not sure if we're better off forbidding it or fixing any surrounding issue, but I'd default to fix anything this breaks, and then maybe forbid it if we think that improves usability or something.

kianenigma:
Given the typical situation that a chain launches but has no balance transfers enabled yet, then a singular account as both stash and controller is basically the only way. 

All in all and given this, I am in for closing this as it is fine. A UI can prevent you from locking yourself out (bonding everything and thus your controller cannot pay for fees) and show you warnings.

burdges:
Cool.  A stash selecting itself as controller is different form a daisy chain of course, so good UX might be giving that a special exception, but I've no strong opinions.

PR #3568, opened 2019-09-06
Make Stash and Controller Roles Exclusive

joepetrowski:
Closes https://github.com/paritytech/substrate/issues/3566

Prevents bonded accounts from becoming controllers. Likewise prevents controllers from bonding.

### DIFF
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -80,8 +80,8 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// and set impl_version to equal spec_version. If only runtime
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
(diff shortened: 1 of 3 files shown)


gavofyork:
Not clear that we actually want to forbid this at the chain level, and certainly not during the Kusama run-in.

kianenigma:
Looks good but, as mentioned, iced for now. If and when this is in the pipeline to be merged, it would be nice to use this PR to push in also:
- let's make helper functions for `is_stash` and `is_controller` and `is_bonded` (which is basically the `&&` of the previous two) as they are quite frequently used in the code and always manually checked by reading `::exists()`. 
- The test can probably be re-written using the already-funded accounts in mock.

bkchr:
Close for now as discussed with @joepetrowski

PR #3572, opened 2019-09-08
Allow signing hex messages from stdin

raykyri:
subkey can't sign or verify hex-encoded messages right now (using `-h`) because hex decoding chokes on the trailing newline. This PR strips the trailing newline only from hex messages; string messages will still have a trailing newline appended to them when signed.

Test with a burner key: 
```
echo "aaaa" | target/debug/subkey sign -h "absent caution helmet bar pelican draw require blood liar defense about sad"`

echo "aaaa" | target/debug/subkey verify -h 002b8513d3f1556017f7a41f8e9f9c848238dc7e0810aa3ad36b1948ad4e5d73073d9825c5bde558f15355ac6c5225d039f15f95dc706d0f2d2538c4a3b8738f "absent caution helmet bar pelican draw require blood liar defense about sad"
```

### DIFF
--- a/subkey/src/main.rs
+++ b/subkey/src/main.rs
@@ -146,6 +146,11 @@ fn execute<C: Crypto>(matches: clap::ArgMatches) where
 			let mut message = vec![];
 			stdin().lock().read_to_end(&mut message).expect("Error reading from stdin");
 			if matches.is_present("hex") {
+				if let Some(last_char) = message.last() {
+					if *last_char == b'\n' {
+						message.pop();
+					}
+				}
 				message = hex::decode(&message).expect("Invalid hex in message");
 			}
 			let sig = pair.sign(&message);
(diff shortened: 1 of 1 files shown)


raykyri:
[clabot:check]

marcio-diaz:
Try with `-n`:
```
echo -n 'aaaa' | ../target/debug/subkey sign -h "absent caution helmet bar pelican draw require blood liar defense about sad"
7892e0d9beaacea10b6e80e0d615481a35d20fe5909f3583427e353fa3719b0c431b8257ec87ac6447de18747a62856429c7194c5888efdf6d4ed2764c90208c
```
Technically speaking the message of the example is not hex-encoded. I'm not sure we should modify the message depending on what we expect.

raykyri:
-n works!

PR #3578, opened 2019-09-09
Don't initialize block for offchain workers.

tomusdrw:
Offchain workers should be run in the context of block `X`, at the same state as it was after the block is imported. This allows us to inspect `system` module and for instance get events that were deposited by extrinsics in a block.

Having no `skip_initialize_block`, actually re-intializes the state with header at `X+1` so this useful data was being erased.

### DIFF
--- a/core/offchain/primitives/src/lib.rs
+++ b/core/offchain/primitives/src/lib.rs
@@ -26,6 +26,7 @@ decl_runtime_apis! {
 	/// The offchain worker api.
 	pub trait OffchainWorkerApi {
 		/// Starts the off-chain task for given block number.
+		#[skip_initialize_block]
 		fn offchain_worker(number: NumberFor<Block>);
 	}
 }


bkchr:
I'm not sure if this is the correct approach. I understand why you want to do this, but this is probably not the correct way. 
Maybe we should call `initialize` in `offchain_workers` and add a new parameter `kill_events`?

gnunicorn:
Hmmm...

the kill_events feels to specific IMHO. We should rather think hard about what state we do expect the offchain worker to run in. I don't think that I would expect the offchain worker to run in state prior to `finalize`, but after and before any future `initialize` was called (setting up a new block). I think that is a pretty natural expectation: it is run in the context of the last block untouched. I would not expect any other code to run after `finalize` before `offchain_worker` is called. Therefore running `initialize` before calling `offchain_worker` is IMHO a bug, this PR addressed appropriately.

//cc @bkchr

bkchr:
Yeah, I understand, but look [here](https://github.com/paritytech/substrate/blob/master/srml/system/src/lib.rs#L667).
We remove `BlockNumber`, `ParentHash` and some other stuff from the storage. When we can live without this, fine. I just wanted to highlight it :)

tomusdrw:
@bkchr this other stuff is totally fine - at least for now, in the future we can also overcome this by passing an import-route as a parameter to offchain worker call, which might also be interesting for some cases.
We actually are about all things mentioned here:
https://github.com/paritytech/substrate/blob/master/srml/system/src/lib.rs#L699

And it makes sense for me to have the same data that is available for "client inspection" to be available for offchain workers.

bkchr:
Okay, as long as we inform the people that some stuff is not available in the offchain, it is fine to me.

What you mean by import route? What would this do?

PR #3583, opened 2019-09-10
Fixing BABE epochs to change between blocks

### LINKED ISSUE #3586: BABE epochs should change based on time-slot, not block number
[The BABE Spec](https://research.web3.foundation/en/latest/polkadot/BABE/1-cycle/)

Time is divided into discrete `slot`s. In the spec, with each epoch consisting of _exactly_ `epoch_len` slots. In the current implementation, epochs consist of _at minimum_ `epoch_len` slots.

However, the current implementation changes the epoch at the first block `>= epoch_len` slots after the start slot of the epoch. This is bad for two reasons:
  - it's not to-spec
  - block authors can temporarily delay the time for the new epoch to start, which forces the old randomness to be used

It also complicates equivocation slashing, since different blocks on small forks may have the same slot number, but be on different epochs (i.e. sessions). If the same validator has 2 different session keys on 2 chains, he can equivocate without being punished.

-------

Required Changes:
 
**Account for Time-Slot when Authoring/Verifying**
(core)

When determining which authority-set/randomness to verify under, we should account for time-slot.

We will change the `fn epoch` in BABE to follow this procedure:
  1. find the epoch information of the parent block
  2. compare that epoch's end-slot to the current slot number
    2a. if less, return that epoch
    2b. if greater than or equal to the end, return the successor epoch

**Account for chains that did not start with these changes**
(core)

unsure. there are a couple different options

1. hard-fork, which introduces special-cased validation logic for such chains (as part of chain spec or BABE config)
2. remove support for such chains

**Compute Randomness without the VRF Output of the Current Block**
(SRML)

After making the above core changes, the epoch can be considered to have changed just _before_ the epoch-changing block, as opposed to during/after. This means that the responsibility of authoring the epoch-changing block moves from the prior set to the new set. Thus the next-epoch randomness, as computed during the epoch change logic, should not include the VRF output from the pre-runtime digest of the block.

rphmeier:
Closes #3586 -- detailed description within that issue.

This PR is rewriting a good portion of BABE to have well-defined behavior at bootstrap time and at epoch-change time.

This is an accumulation PR, with PRs made against it for individual review. More information can be found in those PR descriptions:
  - SRML changes: #3611 
  - Core changes: #3620 

Commits from `c15a2c3` were made directly to this branch and should bear additional review.

Remaining TODOs:
  - [x] SRML tests (#3652)
  - [x] Some more core tests (#3653)


core/consensus/babe/src/lib.rs
```diff
@@ -929,29 +930,13 @@ impl MaybeSpanEpoch {
 	}
 }
 
-/// Extract current epoch data from cache and fallback to querying the runtime
-/// if the cache isn't populated.
-fn epoch<B, C>(client: &C, at: &BlockId<B>) -> Result<MaybeSpanEpoch, ConsensusError> where
+/// Extract current epoch data from the runtime.
```
DemiMarie-parity: There should be a TODO here for light clients.

node/cli/src/chain_spec.rs
```diff
@@ -398,7 +399,11 @@ pub(crate) mod tests {
 		service_test::connectivity(
 			integration_test_config_with_two_authorities(),
 			|config| new_full(config),
-			|config| new_light(config),
+			|mut config| {
+				// light nodes are unsupported
```
DemiMarie-parity: This should also be a TODO.  If this were not a test case, I would prefer to bail out and exit with an error here.

node/cli/src/service.rs
```diff
@@ -361,7 +361,11 @@ mod tests {
 		service_test::sync(
 			chain_spec::integration_test_config(),
 			|config| new_full(config),
-			|config| new_light(config),
+			|mut config| {
+				// light nodes are unsupported
+				config.roles = Roles::FULL;
```
DemiMarie-parity: ditto

node/cli/src/service.rs
```diff
@@ -389,7 +393,11 @@ mod tests {
 		service_test::sync(
 			chain_spec,
 			|config| new_full!(config),
-			|config| new_light(config),
+			|mut config| {
+				// light nodes are unsupported
+				config.roles = Roles::FULL;
```
DemiMarie-parity: ditto

node/cli/src/service.rs
```diff
@@ -500,7 +508,11 @@ mod tests {
 		service_test::consensus(
 			crate::chain_spec::tests::integration_test_config_with_two_authorities(),
 			|config| new_full(config),
-			|config| new_light(config),
+			|mut config| {
+				// light nodes are unsupported
+				config.roles = Roles::FULL;
```
DemiMarie-parity: ditto

core/consensus/babe/src/lib.rs
rphmeier: light client support doesn't work now. not changing the status quo

node/cli/src/chain_spec.rs
rphmeier: disagree - we haven't added light client support as a feature yet. An issue would be fine; I don't see the need for a TODO

srml/babe/src/lib.rs
```diff
@@ -378,19 +380,39 @@ impl<T: Trait> Module<T> {
...
+		let maybe_vrf = maybe_pre_digest.and_then(|digest| {
+			// on the first non-zero block (i.e. block #1)
+			// this is where the first epoch (epoch #0) actually starts.
+			// we need to adjust internal storage accordingly.
+			if GenesisSlot::get() == 0 {
+				GenesisSlot::put(digest.slot_number());
+				debug_assert_ne!(GenesisSlot::get(), 0);
+
+				// deposit a log because this is the first block in epoch #0
+				// we use the same values as genesis because
```
andresilva:
```suggestion
				// we use the same values as genesis because we haven't collected any 
				// randomness yet.
```

node/cli/src/service.rs
```diff
@@ -359,13 +359,18 @@ mod tests {
...
+				new_full(config)
+			},
 			block_factory,
 			extrinsic_factory,
 		);
 	}
 
 	#[test]
+	#[cfg(any())]
```
andresilva: Why is this needed?

srml/babe/src/mock.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+impl_opaque_keys! {
+	pub struct MockSessionKeys {
+		#[id(DUMMY)]
+		pub dummy: UintAuthorityId,
+	}
+}
+
+impl session::Trait for Test {
```
andresilva:
```
79 | impl session::Trait for Test {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `DisabledValidatorsThreshold` in implementation
   |
   = note: `DisabledValidatorsThreshold` from trait: `type DisabledValidatorsThreshold;`
```

node/cli/src/service.rs
rphmeier: dunno, must have come in with one of the merges. will remove

PR #3584, opened 2019-09-10
Make NPoS code less brittle

gavofyork:


drewstone:
Hey Gav, wanted to give clarity on how this has affected us. For some reason, the rewards on our chain are extremely large for a small amount staked. In some tests, I have run with the following
- 2 minute eras
- 60 total tokens staked
- 5 billion total tokens issued

With these details, I see era rewards of over 450,000 tokens to 6 validators. This doesn't seem entirely correct given this would lead to a very high inflation rate if the network continued on this schedule. Do you have an idea of if this would fix that or how I might tune things to properly reward validators? We're looking to fix this, this week if possible.

thiolliere:
I think value is correct:
there is almost no token at stake thus inflation is I_0 = 0.025.
the total_payout for 2 minutes is `5*10^27 * 0.025 * 120 / (3600*24*365.25)=~4.7e20`

thiolliere:
ah but indeed the total payout is wrong as gav lead me to this fact. it expect duration to be expressed in seconds but it is expressed in milliseconds so there is a 1000 time factor.

I'm doing a PR now

gavofyork:
closing for now; can be rebased/reopened once the TODOs are complete @thiolliere

PR #3589, opened 2019-09-10
read_storage panics

pepyakin:
There is a problem with `read_storage`, if an offset passed that is out of bounds of the length of the value the implementation of an extrinsic will panic. This PR solves it.

There is another problem with the API itself: there is no way to indicate how many bytes were read. So even if we fix the panic, the API wouldn't be universally usable since the API cannot indicate how many bytes it wrote to the buffer.

The API is not used within Substrate, except for implementing `exists` (e.g. [this](https://github.com/paritytech/substrate/blob/b21d76db6edd4540ee9bd5ed2cbaef2b395a21f8/srml/support/src/storage/child.rs#L85)).

### DIFF
--- a/core/executor/src/wasm_executor.rs
+++ b/core/executor/src/wasm_executor.rs
@@ -629,9 +629,9 @@ impl_wasm_host_interface! {
 			)?;
 
 			if let Some(value) = maybe_value {
-				let value = &value[value.len().min(value_offset as usize)..];
-				let written = std::cmp::min(value_len as usize, value.len());
-				context.write_memory(value_data, &value[..written])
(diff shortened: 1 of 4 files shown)


gavofyork:
alternative here: https://github.com/paritytech/substrate/pull/3597 which i prefer since it just forces a cap on the value rather than introducing a full conditional.

the point of the call isn't to be "universal" - if you want true universality, then you either use the allocating `storage` function or use two calls to `read_storage`, one to find the length and the second to actually read some or all of it into a pre-allocated buffer.

gavofyork:
should do the trick.

Issue #3575, opened 2019-09-09
Exponentially growing validator rewards


drewstone:
I'm having issues with respect to the title. What runtime parameters should be changed to curb this?

Here's a link to the runtime and the chainspec:
- https://github.com/hicommonwealth/edgeware-node/blob/develop/node/runtime/src/lib.rs
- https://github.com/hicommonwealth/edgeware-node/blob/develop/node/service/src/chain_spec.rs

drewstone:
My observations since they may help, I was running a local chain with alice, bob, charlie, dave, eve, and ferdie as validators. We have a per block treasury reward being issued.

- At first treasury earns roughly 4 tokens each block
- After some point (60 blocks I believe) the staking rewards skyrocket and treasury begins earning 89 Tera token units
- Validator balances also jump from 10 tokens to around 260,000, and after 120 blocks or so they grow to giga values.

Eventually, the node crashes in an attempt to multiple into an overflow since these values become so large. The nodes stall at block 179 since 180 would issue new rewards.

thiolliere:
I take a look,

for information inflation model is written here https://github.com/paritytech/substrate/blob/9b08e7ff938a45dbec7fcdb854063202e2b0cb48/srml/staking/src/inflation.rs#L17-L54
for the spec https://research.web3.foundation/en/latest/polkadot/Token%20Economics/#inflation-model

drewstone:
@thiolliere does this have issues when one changes the number of decimals in the token units? We use 18 but most chains I see being deployed on your end are 15. Could this have adverse effects?

drewstone:
I may have found an issue (or I'm using something wrong) with respect to the OnRewardMinted hook of the staking module and the corresponding on_dilution of the Treasury module.
This call here: https://github.com/hicommonwealth/substrate/blob/develop/srml/staking/src/lib.rs#L1196
Combined with: https://github.com/hicommonwealth/edgeware-node/blob/develop/node/runtime/src/lib.rs#L260
Makes for an extremely large dilution issuance
I added an event in our copy of substrate and after the first era, am getting over 50 trillion new tokens minted to the Treasury
I added an event here: https://github.com/hicommonwealth/substrate/blob/develop/srml/treasury/src/lib.rs#L350
You can also replicate it by running with the following two validators on develop branch of edgeware-node

drewstone:
I can confirm, removing the Treasury from the `OnRewardMinted` fixes these problems. @thiolliere

PR #3590, opened 2019-09-10
Fix reward to validator

thiolliere:
This fix the issue in regards rewards to validator. But this breaks consensus isn't it ? this might need additionnal modification then.

Fix https://github.com/paritytech/substrate/issues/3575
Related https://github.com/paritytech/substrate/pull/3584


gavofyork:
(Subsumed into my more general NPoS PR)

PR #3591, opened 2019-09-10
Use `--release` for running a Substrate dev node

shawntabrizi:
Users should be told to use the `--release` flag when running a node, otherwise users may run into block production time problems.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -277,9 +277,9 @@ Or just run the tests of a specific package (i.e. `cargo test -p srml-assets`)
 You can start a development chain with:
 
 [source, shell]
-cargo run \-- --dev
+cargo run --release -- --dev
 
-Detailed logs may be shown by running the node with the following environment variables set: `RUST_LOG=debug RUST_BACKTRACE=1 cargo run \-- --dev`.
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -277,9 +277,9 @@ Or just run the tests of a specific package (i.e. `cargo test -p srml-assets`)
 You can start a development chain with:
 
 [source, shell]
-cargo run \-- --dev
+cargo run --release -- --dev
 
-Detailed logs may be shown by running the node with the following environment variables set: `RUST_LOG=debug RUST_BACKTRACE=1 cargo run \-- --dev`.
+Detailed logs may be shown by running the node with the following environment variables set: `RUST_LOG=debug RUST_BACKTRACE=1 cargo run --release -- --dev`.
```
bkchr:
```suggestion
Detailed logs may be shown by running the node with the following environment variables set: `RUST_LOG=debug RUST_BACKTRACE=1 cargo run --release \-- --dev`.
```

shawntabrizi: what is the `\` for?

bkchr: That it is rendered as `--` and not as one `-`.

PR #3596, opened 2019-09-11
Fix tracking validator set in ImOnline

### LINKED ISSUE #3553: Time gap for validators between srml/staking and srml/session will result in direct slash for new validators.
Hi, I noticed there were time gap between srml/staking and srml/session for validators query. 
1. at first, my chain has 4 initial validators.
2. then two new validators added
3. at the next era(era5), staking/CurrentElected showed the two new nodes elected and there were 6 validators at present as below:
image. However, session/Validators still stayed as total 4 validators
4. sometime later, at era6, session/Validators changed to 6 validators as below:
image

So there were time gap between above 3 and 4. That will result in direct slashing at the very beginning for the two new validators.
1. since the current elected count changed to 6, but the session validator_set/previous im-online count was 4
2. the the new two nodes were reported offence at the very beginning
3. slash happened

See logs below:
2019-09-05 10:35:36 //////on_before_session_ending function current elected length 6, keys length 4
2019-09-05 10:35:36 //////report_offence function offenders count 2, validator set 4
2019-09-05 10:35:36 /////on_offence function current era 6
2019-09-05 10:35:36 /////on offence slash_exposure9999999999327072
2019-09-05 10:35:36 /////slash validator total slash 374999999974765
2019-09-05 10:35:36 /////slash validator own_slash 374999999974765
2019-09-05 10:35:36 /////on offence slash_exposure9999999999327072
2019-09-05 10:35:36 /////slash validator total slash 374999999974765
2019-09-05 10:35:36 /////slash validator own_slash 374999999974765

I think it's unreasonable since the two new validators got slashing at the very beginning, but they did nothing wrong. 
Is the time gap for validator count between staking and session module as expected?
Is the using previous im-online record for the new round of era at which new nodes may  join reasonable?

Thanks.

tomusdrw:
Resolves #3553 

Currently `im-online` is using (indirectly) `staking::current_elected_set` to get the validators and report those who didn't send a `Heartbeat`.
That's incorrect, since `session` is buffering the keys (and validator set changes). In this PR we use `Session::validators` instead, which allows us to get rid of the extra associated type.

- Add some simple tests to `im-online` module.
- Remove dependency on `staking`
- Relax crypto requirements (don't require `AppKey`).
- Make `UintAuthorityId` more usable in tests.

Polkadot PR: #429


core/sr-primitives/src/testing.rs
```diff
@@ -33,6 +33,24 @@ use crate::transaction_validity::{TransactionValidity, TransactionValidityError}
 #[derive(Default, PartialEq, Eq, Clone, Encode, Decode, Debug, Hash, Serialize, Deserialize)]
 pub struct UintAuthorityId(pub u64);
 
+impl From<u64> for UintAuthorityId {
+	fn from(id: u64) -> Self {
+		UintAuthorityId(id)
+	}
+}
+
+impl PartialOrd for UintAuthorityId {
```
bkchr: WHy not just auto derive it?

core/sr-primitives/src/testing.rs
```diff
@@ -33,6 +33,24 @@ use crate::transaction_validity::{TransactionValidity, TransactionValidityError}
...
+	fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+		self.0.partial_cmp(&other.0)
+	}
+}
+
+impl Ord for UintAuthorityId {
```
bkchr: Same

core/sr-primitives/src/testing.rs
```diff
@@ -47,34 +65,44 @@ impl CryptoType for UintAuthorityId {
...
-	#[cfg(not(feature = "std"))]
-	fn generate_pair(_: Option<&str>) -> Self {
-		unimplemented!("`generate_pair` not implemented for `UIntAuthorityId` on `no_std`.")
-	}
-
 	fn sign<M: AsRef<[u8]>>(&self, msg: &M) -> Option<Self::Signature> {
 		let mut signature = [0u8; 8];
 		msg.as_ref().iter()
-			.chain(rstd::iter::repeat(&42u8))
+			.chain(std::iter::repeat(&42u8))
```
bkchr: Why? This is also compiled to wasm and will not be happy about `std`^^

core/sr-primitives/src/testing.rs
```diff
@@ -85,7 +113,7 @@ impl app_crypto::RuntimeAppPublic for UintAuthorityId {
 	fn verify<M: AsRef<[u8]>>(&self, msg: &M, signature: &Self::Signature) -> bool {
 		let mut msg_signature = [0u8; 8];
 		msg.as_ref().iter()
-			.chain(rstd::iter::repeat(&42))
+			.chain(std::iter::repeat(&42))
```
bkchr: Same.

srml/im-online/src/tests.rs
```diff
@@ -0,0 +1,187 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2019 Parity Technologies (UK) Ltd.
```

core/sr-primitives/src/testing.rs
tomusdrw: It's not. The entire module is compiled only for `std`.

srml/im-online/src/tests.rs
```diff
@@ -0,0 +1,187 @@
...
+		network_state: OpaqueNetworkState {
+			peer_id: OpaquePeerId(vec![1]),
+			external_addresses: vec![],
+		},
+		session_index,
+		authority_index,
+	};
+	let signature = id.sign(&heartbeat.encode()).unwrap();
+
+	// when
```
kianenigma:
```suggestion

```

srml/im-online/src/tests.rs
```diff
@@ -0,0 +1,187 @@
...
+		System::set_block_number(block);
+		// buffer new validators
+		Session::rotate_session();
+		// enact the change and buffer another one
+		VALIDATORS.with(|l| *l.borrow_mut() = Some(vec![1, 2, 3, 4, 5, 6]));
+		Session::rotate_session();
+		assert!(!ImOnline::is_online_in_current_session(0));
+
+		// when
+		heartbeat(block, 2, 0, 1.into());
```
kianenigma: I don't get how in block 1 session_index = 2

srml/im-online/src/tests.rs
```diff
@@ -0,0 +1,219 @@
...
+		advance_session();
+		// given
+		VALIDATORS.with(|l| *l.borrow_mut() = Some(vec![1, 2, 3, 4, 5, 6]));
+		assert_eq!(Session::validators(), Vec::<u64>::new());
+		// enact the change and buffer another one
+		advance_session();
+
+		assert_eq!(Session::current_index(), 2);
+		// TODO: hmmm...
+		assert_eq!(Session::validators(), vec![1, 2, 3]);
```
kianenigma: why? I was expecting all 6 of them to get buffered.

srml/im-online/src/tests.rs
kianenigma: (I changed the code a bit but my question still holds)
(I am aware of the buffering and that we need one extra rotate_session to write the buffered validators)

tomusdrw: Session rotations are independent of blocks. In regular `node-runtime` that would never happen though.

srml/im-online/src/tests.rs
tomusdrw: The first time you call `advance_session` we switch to:
`current(vec![]), queued(vec![1, 2, 3])`
second time you call `advance_session` we switch to:
`current(vec![1, 2, 3]), queued(vec![1,...6]))`

Initially both current and queued are empty.

core/sr-primitives/src/testing.rs
bkchr: Yeah, sorry for the noise. As reviewing it, I looked into the wrong `primitives` crate :( And after the review I saw my mistake and marked them as resolved directly..^^

srml/im-online/src/tests.rs
```diff
@@ -0,0 +1,219 @@
...
+	with_externalities(&mut new_test_ext(), || {
+		advance_session();
+		// given
+		VALIDATORS.with(|l| *l.borrow_mut() = Some(vec![1, 2, 3, 4, 5, 6]));
+		assert_eq!(Session::validators(), Vec::<u64>::new());
+		// enact the change and buffer another one
+		advance_session();
+
+		assert_eq!(Session::current_index(), 2);
+		// TODO: hmmm...
```
cmichi: Was this ToDo forgotten here? If not a follow-up ticket or a better explanation would be nice.

tomusdrw:
Everything fixed now, should be good to go.

PR #3597, opened 2019-09-11
Alternative

gavofyork:


core/sr-io/with_std.rs
```diff
@@ -87,11 +85,9 @@ impl StorageApi for () {
 			let storage_key = child_storage_key_or_panic(storage_key);
 			ext.child_storage(storage_key, key)
 				.map(|value| {
-					if value_offset < value.len() {
-						let value = &value[value_offset..];
-						let written = std::cmp::min(value.len(), value_out.len());
-						value_out[..written].copy_from_slice(&value[..written]);
-					}
+					let value = &value[value_offset.min(value.len())..];
+					let written = std::cmp::min(value.len(), value_out.len());
```
pepyakin: nit: can we use a single way of calling `min`?

gavofyork: i don't think so...

PR #3602, opened 2019-09-11
grandpa: reannounce voted blocks

andresilva:
In order for GRANDPA to keep progressing it is imperative that nodes sync to all forks that are being voted on, since we cannot process a vote until we import the block it refers to (and it ancestors).

This PR adds a background worker that keeps track of the block hashes of our latest votes and re-announces them through sync when GRANDPA stops progressing. It just keeps a bounded `VecDeque` of latest hashes and announces all of them when stalled, without tracking any context of round or set id.

### DIFF
--- a/core/finality-grandpa/src/communication/periodic.rs
+++ b/core/finality-grandpa/src/communication/periodic.rs
@@ -16,19 +16,26 @@
 
 //! Periodic rebroadcast of neighbor packets.
 
-use super::{gossip::{NeighborPacket, GossipMessage}, Network};
+use std::collections::VecDeque;
+use std::time::{Instant, Duration};
+
+use codec::Encode;
 use futures::prelude::*;
 use futures::sync::mpsc;
-use sr_primitives::traits::{NumberFor, Block as BlockT};
-use network::PeerId;
-use tokio_timer::Delay;
 use log::{debug, warn};
-use codec::Encode;
+use tokio_timer::Delay;
 
-use std::time::{Instant, Duration};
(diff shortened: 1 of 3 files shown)


core/finality-grandpa/src/communication/periodic.rs
```diff
@@ -106,3 +114,141 @@ pub(super) fn neighbor_packet_worker<B, N>(net: N) -> (
...
+		if !self.latest_voted_blocks.contains(&block) {
+			if self.latest_voted_blocks.len() >= LATEST_VOTED_BLOCKS_TO_ANNOUNCE {
+				self.latest_voted_blocks.pop_front();
+			}
+
+			self.latest_voted_blocks.push_back(block);
+
+			return true;
+		}
+
```
DemiMarie-parity:
```suggestion
	else {
		false
	}
```

core/finality-grandpa/src/communication/periodic.rs
```diff
@@ -106,3 +114,141 @@ pub(super) fn neighbor_packet_worker<B, N>(net: N) -> (
...
+						self.net.announce(block);
+						self.reset_delay();
+					}
+				},
+				Async::NotReady => break,
+			}
+		}
+
+		// after the delay fires announce all blocks that we have stored. note
+		// that this only happens if we don't receive any new blocks above for
```
DemiMarie-parity: This comment is confusing.  The code seems correct, but we will hit this loop even if the timer has not expired.

PR #3607, opened 2019-09-12
Send block status with announcement

### LINKED ISSUE #3545: Use fork choice rule when syncing
Currently sync algo assumes that the longest chain is the best when syncing. This need to be changed to account for consensus fork choice rule.

arkpar:
Adds block status to `BlockAnnounce` message. This allows for better tracking of perceived best block for peers and removes assumption that their longest chain is the best.

Fixes #3545.

### DIFF
--- a/core/network/src/protocol/sync.rs
+++ b/core/network/src/protocol/sync.rs
@@ -917,7 +917,7 @@ impl<B: BlockT> ChainSync<B> {
 		if let PeerSyncState::AncestorSearch(_, _) = peer.state {
 			return OnBlockAnnounce::Nothing
 		}
-		// If the announced block is the latest they have seen, our common number
+		// If the announced block is the best they have seen, our common number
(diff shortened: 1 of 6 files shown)


core/network/src/protocol/sync.rs
```diff
@@ -907,17 +909,17 @@ impl<B: BlockT> ChainSync<B> {
...
+		if is_best && number > peer.best_number {
 			// update their best block
 			peer.best_number = number;
 			peer.best_hash = hash;
 		}
 		if let PeerSyncState::AncestorSearch(_, _) = peer.state {
 			return OnBlockAnnounce::Nothing
 		}
-		// We assume that the announced block is the latest they have seen, and so our common number
+		// If the announced block is the latest they have seen, our common number
```
andresilva:
```suggestion
		// If the announced block is the best they have seen, our common number
```

PR #3609, opened 2019-09-13
Clean up sr-io

bkchr:
As preparation for auto generating the host extern interface based on the sr-io interface, we need to remove any usage of generics.


core/sr-primitives/src/traits.rs
```diff
@@ -456,18 +457,11 @@ pub trait Hash: 'static + MaybeSerializeDebug + Clone + Eq + PartialEq {	// Stup
 		Encode::using_encoded(s, Self::hash)
 	}
 
-	/// Iterator-based version of `ordered_trie_root`.
-	fn ordered_trie_root<
-		I: IntoIterator<Item = A>,
-		A: AsRef<[u8]>
-	>(input: I) -> Self::Output;
+	/// The ordered Patricia tree root of the given `input`.
+	fn ordered_trie_root(input: Vec<Vec<u8>>) -> Self::Output;
```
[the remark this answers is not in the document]
bkchr: If you look into the changes of actual runtime code, you see that this should even be faster, as it requires less allocations :)

core/client/src/light/fetcher.rs
```diff
@@ -436,7 +436,9 @@ impl<E, Block, H, S> FetchChecker<Block> for LightDataChecker<E, H, Block, S>
 		body: Vec<Block::Extrinsic>
 	) -> ClientResult<Vec<Block::Extrinsic>> {
 		// TODO: #2621
-		let	extrinsics_root = HashFor::<Block>::ordered_trie_root(body.iter().map(Encode::encode));
+		let	extrinsics_root = HashFor::<Block>::ordered_trie_root(
+			body.iter().map(Encode::encode).collect(),
```
gavofyork: :-(

core/sr-io/src/lib.rs
```diff
@@ -160,26 +143,10 @@ export_api! {
...
-		where
-			I: IntoIterator<Item = (A, B)>,
-			A: AsRef<[u8]>,
-			A: Ord,
-			B: AsRef<[u8]>,
-			H: Hasher,
-			H: self::imp::HasherBounds,
-			H::Out: Ord
-		;
+		fn blake2_256_trie_root(input: Vec<(Vec<u8>, Vec<u8>)>) -> H256;
```
gavofyork: not really great for memory usage...

gavofyork:
i guess the iteration was only possible on native, where memory usage/copying isn't such a big issue.

gavofyork:
will need a runtime bump.

bkchr:
> i guess the iteration was only possible on native, where memory usage/copying isn't such a big issue.

Yep, this was only possible on native. In the runtime we even loose some extra allocations that were required to call the api. 

Only this little piece of code in the light client fetcher requires an extra allocation in "production" code.

PR #3610, opened 2019-09-13
Decouple node-executor structures from wasmi

jimpo:
These are some refactors to ease the integration of a compiled Wasm executor. This decouples the memory allocator and sandbox from wasmi. In the case of the sandbox, sandboxed environments will still execute with wasmi in the near future, so the decoupling is just on the supervisor side. (The supervisor is the substrate runtime, which handles external calls from the sandboxed runtime).

With the allocator, it no longer relies on dynamic borrows (which the `wasmi::MemoryRef` does under the hood), and instead gets a reference to the runtime memory on each `allocate` and `deallocate` call.

### DIFF
--- a/core/executor/src/allocator.rs
+++ b/core/executor/src/allocator.rs
@@ -48,8 +48,8 @@
 
 use crate::error::{Error, Result};
 use log::trace;
-use std::convert::TryInto;
-use wasmi::{MemoryRef, memory_units::Bytes};
+use std::convert::{TryFrom, TryInto};
+use std::ops::Range;
 use wasm_interface::{Pointer, WordSize};
 
 // The pointers need to be aligned to 8 bytes. This is because the
@@ -72,8 +72,6 @@ const PREFIX_SIZE: u32 = 8;
 pub struct FreeingBumpHeapAllocator {
 	bumper: u32,
 	heads: [u32; N],
-	heap: MemoryRef,
-	max_heap_size: u32,
 	ptr_offset: u32,
 	total_size: u32,
 }
@@ -89,26 +87,15 @@ impl FreeingBumpHeapAllocator {
 	///
 	/// # Arguments
 	///
-	/// - `mem` - reference to the linear memory instance on which this allocator operates.
 	/// - `heap_base` - the offset from the beginning of the linear memory where the heap starts.
-	pub fn new(mem: MemoryRef, heap_base: u32) -> Self {
-		let current_size: Bytes = mem.current_size().into();
-		let current_size = current_size.0 as u32;
-
-		let mut ptr_offset = heap_base;
-		let padding = ptr_offset % ALIGNMENT;
-		if padding != 0 {
-			ptr_offset += ALIGNMENT - padding;
-		}
-
-		let heap_size = current_size - ptr_offset;
+	pub fn new(heap_base: u32) -> Self {
+		// ptr_offset is the next alignment boundary on or after heap_base.
+		let ptr_offset = (heap_base + ALIGNMENT - 1) / ALIGNMENT * ALIGNMENT;
 
 		FreeingBumpHeapAllocator {
 			bumper: 0,
 			heads: [0; N],
-			heap: mem,
-			max_heap_size: heap_size,
-			ptr_offset: ptr_offset,
+			ptr_offset,
 			total_size: 0,
 		}
 	}
@@ -118,14 +105,23 @@ impl FreeingBumpHeapAllocator {
 	/// There is no minimum size, but whatever size is passed into
 	/// this function is rounded to the next power of two. If the requested
 	/// size is below 8 bytes it will be rounded up to 8 bytes.
-	pub fn allocate(&mut self, size: WordSize) -> Result<Pointer<u8>> {
+	///
+	/// # Arguments
+	///
+	/// - `mem` - a slice representing the linear memory on which this allocator operates.
+	/// - `size` - size in bytes of the allocation request
+	pub fn allocate(&mut self, mem: &mut [u8], size: WordSize) -> Result<Pointer<u8>> {
+		let mem_size = u32::try_from(mem.len())
+			.expect("size of Wasm linear memory is <2^32");
+		let max_heap_size = mem_size - self.ptr_offset;
+
 		if size > MAX_POSSIBLE_ALLOCATION {
 			return Err(Error::RequestedAllocationTooLarge);
 		}
 
 		let size = size.max(MIN_POSSIBLE_ALLOCATION);
 		let item_size = size.next_power_of_two();
-		if item_size + PREFIX_SIZE + self.total_size > self.max_heap_size {
+		if item_size + PREFIX_SIZE + self.total_size > max_heap_size {
 			return Err(Error::AllocatorOutOfSpace);
 		}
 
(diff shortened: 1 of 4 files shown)


core/executor/src/allocator.rs
```diff
@@ -200,45 +195,56 @@ impl FreeingBumpHeapAllocator {
...
+	//
+	// u64 is encoded in memory in little-endian form with all bits flipped. The bit flip serves to
+	// make the high order bits 1 instead of 0, minimizing the chances of reading a corrupted by
+	// valid-looking value.
+	fn get_heap_u64(&self, heap: &[u8], offset: u32) -> Result<u64> {
+		let range = self.heap_range(offset, 8, heap.len())
+			.ok_or_else(|| error("read out of heap bounds"))?;
+		let bytes = heap[range].try_into()
+			.expect("[u8] slice of length 8 must be convertible to [u8; 8]");
+		Ok(!u64::from_le_bytes(bytes))
```
pepyakin: I've read the explaination but still don't get this mechanism, can you ELI5?

core/executor/src/allocator.rs
```diff
@@ -118,14 +106,20 @@ impl FreeingBumpHeapAllocator {
 	/// There is no minimum size, but whatever size is passed into
 	/// this function is rounded to the next power of two. If the requested
 	/// size is below 8 bytes it will be rounded up to 8 bytes.
-	pub fn allocate(&mut self, size: WordSize) -> Result<Pointer<u8>> {
+	///
+	/// # Arguments
+	///
+	/// - `mem` - reference to the linear memory instance on which this allocator operates.
```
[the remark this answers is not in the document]
jimpo: Oops, copy and pasted the comment from somewhere else.

core/executor/src/allocator.rs
jimpo: I'm glad you asked because I would be perfectly happy getting rid of this. I'd really like @cmichi 's opinion here as well.

The intent is to sort of mimic the behavior of [this assertion](https://github.com/paritytech/substrate/pull/3610/files#diff-5da990ee099f29391d7bf93c20a0b25bL171) while being simpler to implement. In the current version of the code, when you write the list index, which should fit in 8 bits, you set the other 64 - 8 bits to all 1s. Then when you read the list index, it asserts they are all 1s. My assumption is that this is because stretches of 7 consecutive bytes of 1 bits are rare in the heap normally, so it acts as some extra detection mechanism against heap corruption. We should probably decide whether that is necessary. BUT assuming we want to keep something like that, I think the easier way to do it is to just write a 64-bit int, read the whole thing as a u64, then check that it's in the expected range. If any of the higher-order bits got flipped, we could detect that the list index is out of range and there is some sort of corruption. On the other hand, bits in the heap are probably more likely to be 0 than 1 simply because the linear memory is 0-initialized and 0 is a more common default. So we would rather these detection bits be 1. So instead of just writing the u64 in LE as one normally would, we can do a bitwise flip as part of the encoding. Then if you read it back and and any of the high order bits are 0, then they will be 1 after flipping and the list index will be out of range, which is detectable.

cmichi: > In the current version of the code, when you write the list index, which should fit in 8 bits, you set the other 64 - 8 bits to all 1s.

AFAIR the intention was to make sure pointers are always aligned to 8 bytes, hence the long 8*8 bytes prefix where only the first 8 bits are used. The bitflip-detection was no intention originally, thoug it doesn't sound bad to me. You suggestion of using a u64 sounds good to me, the other refactorings also make the code better readable :+1: .

pepyakin: TBH, I think that some other pattern other than all 1s, would work better here.
If we are going to leave this approach, I'd love to see this explaination in the code.

core/executor/src/sandbox.rs
```diff
@@ -175,15 +184,37 @@ pub trait SandboxCapabilities {
...
+	/// Returns `Err` if the dispatch_thunk function has an incorrect signature or traps during
+	/// execution.
+	fn invoke(
+		&mut self,
+		dispatch_thunk: &Self::SupervisorFuncRef,
+		invoke_args_ptr: Pointer<u8>,
+		invoke_args_len: WordSize,
+		state: u32,
+		func_idx: SupervisorFuncIndex,
+	) -> Result<i64>;
```
pepyakin: This is pity that we are exposing this, since I'd be happy to get rid of the dispatch thunk shenanigans, and operate on function references directly.

jimpo: Hmm, that sounds like a pretty big change. Would rather just keep it like this for now.

pepyakin: Yep, it is : ) To be clear, I am **not** asking for you to do this, esp. in this PR. I was just iterating on new sandbox v2 designs when I tried to do this.

core/executor/src/allocator.rs
jimpo: I clarified the comment a bit.

core/executor/src/allocator.rs
```diff
@@ -200,45 +194,57 @@ impl FreeingBumpHeapAllocator {
...
-	fn get_heap_4bytes(&mut self, ptr: u32) -> Result<[u8; 4]> {
-		let mut arr = [0u8; 4];
-		self.heap.get_into(self.ptr_offset + ptr, &mut arr)?;
-		Ok(arr)
+	// Read a u64 from the heap. Used to read heap allocation prefixes.
+	//
+	// u64 is encoded in memory in little-endian form with all bits flipped. Since the heap
+	// allocation prefixes are integers in a low range (eg. they fit in a u8 or u32), the bit flip
+	// serves to make the high order bits 1 instead of 0. This acts as a simple error detection
+	// mechanism, as random bits in the heap space are assumed more likely to be 0 than 1.
```
bkchr: Hmm, I don't see any advantage of all this bit flipping. Where would we use this error detection and how? And to assume that random bits are rather 0 than 1, I don't think that this is correct either. If you want to make the values really unique, we could use something like `0xDEADBEEF` or any other recognizable hex value.

pepyakin: FWIW, see the prior discussion [here](https://github.com/paritytech/substrate/pull/3610#discussion_r324264261).

core/executor/src/allocator.rs
```diff
@@ -118,14 +105,20 @@ impl FreeingBumpHeapAllocator {
...
+	/// - `size` - size in bytes of the allocation request
+	pub fn allocate(&mut self, mem: &mut [u8], size: WordSize) -> Result<Pointer<u8>> {
 		if size > MAX_POSSIBLE_ALLOCATION {
 			return Err(Error::RequestedAllocationTooLarge);
 		}
 
 		let size = size.max(MIN_POSSIBLE_ALLOCATION);
 		let item_size = size.next_power_of_two();
-		if item_size + PREFIX_SIZE + self.total_size > self.max_heap_size {
+		let max_heap_size = mem.len().try_into().unwrap_or(u32::max_value()) - self.ptr_offset;
```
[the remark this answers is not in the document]
jimpo: Why? This is the only place `mem.len()` is called.

core/executor/src/allocator.rs
jimpo: Removed.

core/executor/src/allocator.rs
jimpo: Removed the bit flipping because its confusing and questionably effective.

core/executor/src/allocator.rs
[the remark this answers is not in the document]
bkchr: If the memory is greater than 4GB, it is invalid. Wasm could not handle that much of memory anyway.

jimpo:
@pepyakin I'm sure sure making everything generic over `SupervisorFuncRef` is a good idea. Another option would be to just store the dispatch thunk table index as a `u32` and do the lookup again inside invoke. It's a bit of wasted effort, but it should have a negligible performance impact and wouldn't mess with the type signatures on a lot of structs.

core/executor/src/allocator.rs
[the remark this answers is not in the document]
gavofyork: (`grumble` until this is resolved)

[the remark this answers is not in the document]
jimpo: Fixed

PR #3611, opened 2019-09-13
Fix the BABE epoch logic in SRML

rphmeier:
This is a PR against #3583 to add the SRML functionality. It is not intended that BABE will build with this PR (hence no green checkmark). It's only for review of the SRML changes. I've checked that SRML-babe and node-runtime build with these changes.

Notable changes:
  - Epoch #0 is not considered to have started until the first block in the chain.
  - More information at configuration time, less redundant information at epoch change
  - The VRF output in the first block of an epoch does not end up in the on-chain's impression of the _last_ epoch's randomness beacons


core/consensus/babe/primitives/src/digest.rs
```diff
@@ -52,17 +52,13 @@ pub enum BabePreDigest {
 		authority_index: super::AuthorityIndex,
 		/// Slot number
 		slot_number: SlotNumber,
-		/// Chain weight (measured in number of Primary blocks)
-		weight: BabeBlockWeight,
```
andresilva: Where should this go then?

srml/babe/src/lib.rs
```diff
@@ -135,22 +135,13 @@ decl_storage! {
...
-		/// signalled later than this.
-		pub EpochStartSlot get(epoch_start_slot): u64;
+		/// The slot at which the first epoch actually started. This is 0
+		/// until the first block of the chain.
+		pub GenesisSlot get(genesis_slot): u64;
 
 		/// Current slot number.
 		pub CurrentSlot get(current_slot): u64;
 
-		/// Whether secondary slots are enabled in case the VRF-based slot is
```
andresilva: I am OK with removing support for dynamically enabling/disabling secondary slots since we can't change the slot duration anyway, so in practice this is not so useful.

srml/babe/src/lib.rs
```diff
@@ -212,20 +203,13 @@ decl_module! {
...
-			match change {
-				Some(change) =>	PendingSecondarySlotsChange::put(change),
-				None => {
-					PendingSecondarySlotsChange::take();
-				},
+			// at the end of the block, we can safely include the new VRF output
+			// from this block into the under-construction randomness. If we've determined
+			// that this block was the first in a new epoch, the changeover logic has
+			// already occurred at this point, so the under-construction randomness
+			// will only contain outputs from the right epoch.
```
andresilva: :+1:

srml/babe/src/lib.rs
```diff
@@ -269,15 +253,25 @@ impl<T: Trait> IsMember<AuthorityId> for Module<T> {
...
-		let diff = CurrentSlot::get().saturating_sub(EpochStartSlot::get());
-		diff >= T::EpochDuration::get()
+		// The session has technically ended during the passage of time
+		// between this block and the last, but we have to "end" the session now,
+		// since there is no earlier possible block we could have done it.
+		//
+		// The exception is for block 1: the genesis has slot 0, so we treat
+		// epoch 0 as having started at the slot of block 1. We want to use
+		// the same randomness and validator set as signalled in the genesis,
+		// so we don't rotate the session.
```
andresilva: :+1:

srml/babe/src/lib.rs
```diff
@@ -363,12 +364,20 @@ impl<T: Trait> Module<T> {
...
+		#[derive(Default)]
+		struct Guard { maybe_vrf: MaybeVrf };
+		impl Drop for Guard {
+			fn drop(&mut self) {
+				Initialized::put(self.maybe_vrf.take())
+			}
+		}
+
+		let mut guard = Guard::default();
 		for digest in Self::get_inherent_digests()
```
andresilva: I know this code was already written like this but in practice this loop should only have one iteration so maybe we could rewrite this to make that more explicit? Using `find` instead of `filter_map` or just doing a `take(1)` at the end should work. We can probably avoid using this `Guard` if it's explicitly just getting one value from the iterator.

core/consensus/babe/primitives/src/digest.rs
rphmeier: it'll be maintained off-chain; it's a bit redundant to put it in the block header.

srml/babe/src/lib.rs
rphmeier: Right, that was my rationale. You'd never want to enable/disable secondary slots without tweaking the duration or the `c` parameter

core/consensus/babe/primitives/src/lib.rs
```diff
@@ -59,6 +59,11 @@ pub const VRF_PROOF_LENGTH: usize = 64;
 /// The length of the public key
 pub const PUBLIC_KEY_LENGTH: usize = 32;
 
+/// How many blocks to wait before running the median algorithm for relative time
+/// This will not vary from chain to chain as it is not dependent on slot duration
+/// or epoch length.
+pub const MEDIAN_ALGORITHM_CARDINALITY: usize = 1200; // arbitrary suggestion by w3f-research.
```
burdges: It'll become some number of slots maybe, yes?

core/consensus/babe/primitives/src/lib.rs
```diff
@@ -87,57 +92,55 @@ pub struct Epoch {
...
-	pub median_required_blocks: u64,
+	/// The authorities for the genesis epoch.
+	pub genesis_authorities: Vec<(AuthorityId, BabeAuthorityWeight)>,
+
+	/// The randomness for this epoch.
+	pub randomness: [u8; VRF_OUTPUT_LENGTH],
+
+	/// Whether this chain should run with secondary slots, which are assigned
+	/// in round-robin manner.
+	pub secondary_slots: bool,
```
burdges: Another option might be `pub secondary_slots: u16,` or `Option<..>` that expresses how many VRF slots should elapse between secondary slots, with `u16::MAX` or `None` denoting not to bother with secondary slots.

PR #3612, opened 2019-09-13
Fetching multiple remote values using single message

svyatonik:
This PR changes existing messages format => breaks compatibility between existing full nodes and post-PR light nodes. I'm not sure, whether I need to bump protocol version, or mark PR as `breakeverything` in this case - light is unstable anyway.

The PR is a must for:
1) implementing client-side storage polling on light client (using `query_storage` for the latest block);
2) properly fixing storage subscriptions (`subscribe_storage`) on light client (I'll open PR for that later today).


core/network/src/protocol.rs
```diff
@@ -1301,16 +1316,29 @@ impl<B: BlockT, S: NetworkSpecialization<B>, H: ExHashT> Protocol<B, S, H> {
 		who: PeerId,
 		request: message::RemoteReadChildRequest<B::Hash>,
 	) {
+		let keys_str = match request.keys.len() {
```
tomusdrw: I guess it's not a performance bottleneck, but we could do that lazily, only when logging is enabled. I guess it's better to do it twice when tracing than to do it even if we don't use it at all.

core/rpc/src/state/state_light.rs
```diff
@@ -153,9 +153,16 @@ impl<Block, F, B, E, RA> StateBackend<B, E, Block, RA> for LightState<Block, F,
 				Ok(header) => Either::Left(fetcher.remote_read(RemoteReadRequest {
 					block: header.hash(),
 					header,
-					key: key.0,
+					keys: vec![key.0.clone()],
```
tomusdrw: Should we change the RPC to support fetching multiple keys at once? I guess that would be way more efficient with light client in the backend.

svyatonik: Yes - I've mentioned that in description :) I was working on fixing `state_subscribeRuntimeVersion` and `state_subscribeStorage` for light client and they're working really slow without this fix. Hope it'll make these subscriptions on light slightly faster (though still ineffective imo).

I'll create another PR for RPC just because I consider this PR as a must-have and the other as an optional. I'm still not sure - if we need subscription on light client. They could be implemented by polling from client code with the same efficiency. But I believe, at least now, we still need subscriptions-on-light, because I find it very convenient to test light client using https://polkadot.js.org/apps/#/ and it doesn't work without subscriptions.

PR #3613, opened 2019-09-13
RPC call to get all RPC methods

tomusdrw:
This PR introduces a meta call to list all available RPC methods. This will allow the UI to display only the calls that are actually available.
While this wasn't really an issue before, now that we allow for runtime-specific RPCs (#3563, #3109), the supported set may vary.

CC @jacogr


tomusdrw:
After speaking with @jacogr we've decided to update the format to have at least some simplistic versioning, so that we can easily update it in the future.
One of the ideas would be to support versioned RPC methods, so then the meta call would return method versions as well.

PR #3617, opened 2019-09-14
Cumulative fixes to make working with consensus-pow easier

sorpaas:
This PR contains several cumulative fixes to make working with consensus-pow easier:

* Added difficulty data to `PowAux`. The reason for this is that nearly all difficulty adjustment algorithms need this information.
* Added `TimestampApi` in `pow-primitives` so that consensus engine can query timestamp data.
* Added an implementation of `FinalityProofProvider` for `()`, which always return `None`. This seems to be needed because we don't currently have a "dummy" finality proof provider for those engines that do not need it.
* Generalize `Difficulty`, removing assumption that it is `u128`.


core/network/src/chain.rs
```diff
@@ -78,6 +78,12 @@ pub trait FinalityProofProvider<Block: BlockT>: Send + Sync {
 	fn prove_finality(&self, for_block: Block::Hash, request: &[u8]) -> Result<Option<Vec<u8>>, Error>;
 }
 
+impl<Block: BlockT> FinalityProofProvider<Block> for () {
```
andresilva: I think the finality proof provider is optional when creating the service.

sorpaas: I'll try to provide a reproducible test case, but I think what I had trouble with is that if I remove `.with_finality_proof_provider` for the service builder, the whole type resolution will fail. But maybe we should fix that instead.

core/consensus/pow/src/lib.rs
```diff
@@ -63,6 +63,8 @@ fn aux_key(hash: &H256) -> Vec<u8> {
 /// Auxiliary storage data for PoW.
 #[derive(Encode, Decode, Clone, Debug, Default)]
 pub struct PowAux {
+	/// Difficulty.
+	pub difficulty: Difficulty,
```
pepyakin: What difficulty is this? The current difficulty?

core/network/src/chain.rs
andresilva: Yeah I see, it will fail to infer the generic type of the optional finality proof provider.

sorpaas:
I made `Difficulty` a generic and removed the assumption that it is `u128`. After trying out some PoW implementations, I realized that we have a variety of cases -- we have difficulty that are always really small (Bitcoin -- `u32`) and cases where difficulty can be potentially really big (Monero -- `u128`). For the former, fixed size is a waste of space, and for the latter, it risks overflow. So making it generic may be a good thing to do.

core/consensus/pow/src/lib.rs
sorpaas: Yeah, it's the current block's difficulty.

core/consensus/pow/src/lib.rs
```diff
@@ -205,11 +226,12 @@ impl<B: BlockT<Hash=H256>, C, Algorithm> Verifier<B> for PowVerifier<C, Algorith
...
 		let mut aux = PowAux::read(self.client.as_ref(), &parent_hash)?;
 
-		let (checked_header, difficulty, seal) = self.check_header::<B>(
+		let (checked_header, difficulty, seal) = self.check_header(
 			header,
 			BlockId::Hash(parent_hash),
 		)?;
-		aux.total_difficulty = aux.total_difficulty.saturating_add(difficulty);
+		aux.difficulty = difficulty;
+		aux.total_difficulty.add(difficulty);
```
HCastano: Why was this changed from a `saturating_add()`?

core/consensus/pow/src/lib.rs
pepyakin: Can we reflect this in the comment?

sorpaas: Fixed!

core/consensus/pow/src/lib.rs
sorpaas: This is the `add` function from `TotalDifficulty` trait. The idea here is that some `Difficulty` definitions may want to use its own customized method for calculating the total difficulty from current difficulty. Most of the time `TotalDifficulty::add` will still be implemented as `saturating_add`.

core/consensus/pow/src/lib.rs
```diff
@@ -205,11 +225,12 @@ impl<B: BlockT<Hash=H256>, C, Algorithm> Verifier<B> for PowVerifier<C, Algorith
...
```
pepyakin: A small nit: my brain treats `add` as a function which takes `&self` and returns `Self`. For this case, it seems `increase(&mut self, other: Self) → ()` would be more appropriate, don't you think so?

core/consensus/pow/src/lib.rs
```diff
@@ -157,6 +173,10 @@ impl<C, Algorithm> PowVerifier<C, Algorithm> {
 	{
 		const MAX_TIMESTAMP_DRIFT_SECS: u64 = 60;
 
+		if *block.header().number() < self.check_inherents_after {
```
andresilva: Why did we need to add this? Do we have to deal with any existing deployment?

core/consensus/pow/primitives/src/lib.rs
```diff
@@ -20,17 +20,46 @@
...
+
+impl TotalDifficulty for primitives::U256 {
+	fn add(&mut self, other: Self) {
+		let ret = self.saturating_add(other);
+		*self = ret;
+	}
+}
+
+impl TotalDifficulty for u128 {
+	fn add(&mut self, other: Self) {
```
andresilva: `fn increment`?

core/consensus/pow/src/lib.rs
```diff
@@ -331,27 +359,35 @@ pub fn start_mine<B: BlockT<Hash=H256>, C, Algorithm, E>(
...
 {
 	'outer: loop {
+		if sync_oracle.is_major_syncing() {
+			debug!(target: "pow", "Skipping proposal due to sync.");
+			std::thread::sleep(std::time::Duration::new(1, 0));
+			continue 'outer
+		}
+
 		let best_hash = client.info().best_hash;
 		let best_header = client.header(BlockId::Hash(best_hash))
```
andresilva: We should be using a `SelectChain` here with `LongestChain`. There are some edge cases where the best block as reported by the client may be a little outdated.

core/consensus/pow/src/lib.rs
```diff
@@ -397,10 +431,22 @@ fn mine_loop<B: BlockT<Hash=H256>, C, Algorithm, E>(
...
+			let mut header = header.clone();
+			header.digest_mut().push(DigestItem::Seal(POW_ENGINE_ID, seal.clone()));
+			header.hash()
+		};
 
 		let key = aux_key(&hash);
+		let best_hash = client.info().best_hash;
+		let best_aux = PowAux::<Algorithm::Difficulty>::read(client, &best_hash)?;
+
+		if best_aux.total_difficulty > aux.total_difficulty {
```
andresilva:
```suggestion
		// if the best block has changed in the meantime drop our proposal
		if best_aux.total_difficulty > aux.total_difficulty {
```

core/consensus/pow/primitives/src/lib.rs
pepyakin: Hehe, see [this](https://github.com/paritytech/substrate/pull/3617#discussion_r330491441). My suggestion was `increase` but I think `increment` should be also good

core/consensus/pow/src/lib.rs
sorpaas: I added this parameter because runtime upgrades may break older inherent data check. For example, when the inherent data format is changed completely or an old inherent module is removed. In those cases, some chains might want to completely skip the check, because the older the block is, the more "finalized" it is because of PoW built on top of it.

This branch can be disabled by simply setting `check_inherents_after` to 0.

PR #3619, opened 2019-09-15
Enable nice wasm panic messages by default

bkchr:
By accident it was already enabled in master for quite some time. To
make sure that we don't blow up the wasm binary size, I compiled the
binary with the feature enabled and disabled. With nice panic messages
enabled, the binary size increases by 908 bytes. Given the value that
this feature brings, I think it is okay to have these panic messages
enabled by default.


andresilva:
Needs to bump `impl_version`.

PR #3620, opened 2019-09-16
Fix BABE epochs-transition logic on the core side

rphmeier:
This is the counterpart to the initial type-definition work done in #3611 on the core side.

Rough overview of changeset:
  - Extract `SharedEpochChanges` logic out to its own module in a way that can be more cleanly tested
  - Track chain total-weight offchain and handle it in the block-import worker
  - Define epoch 0 as starting at block 1. There might be multiple block 1s, so they might all kick off slightly competing epochs.
  - Allow epochs to change between blocks and grab epoch based on start slot as well as ancestry.
  - Detect that the first block in an epoch is such by comparing the epoch start slot to the parent block's slot. The first block in an epoch must issue a `NextEpochDescriptor` about the following epoch.
  - Does not define pruning of the epoch fork-tree at this point. That can come in a follow-up PR, since the logic is complex.
  - Fix BABE tests

### DIFF
--- a/core/consensus/babe/src/epoch_changes.rs
+++ b/core/consensus/babe/src/epoch_changes.rs
@@ -96,10 +96,10 @@ impl<Block: BlockT> EpochChanges<Block> {
 		let is_descendent_of = descendent_of_builder
 			.build_is_descendent_of(None);
 
-		// prune everything except for the first node _above_ the just-finalized block
-		// because it may represent the epoch that descendents of the finalized block
-		// are in.
-		unimplemented!()
+		// prune any epochs which could not be _live_ as of the children of the
+		// finalized block.
+		// i.e. re-root the fork tree to the earliest ancestor of (hash, number)
+		// where epoch.start_slot + epoch.duration >= slot(hash)
 	}
 
 	/// Finds the epoch for a child of the given block, assuming the given slot number.
--- a/core/consensus/babe/src/epoch_changes.rs
+++ b/core/consensus/babe/src/epoch_changes.rs
@@ -90,16 +90,19 @@ impl<Block: BlockT> EpochChanges<Block> {
 	pub fn prune_finalized<D: IsDescendentOfBuilder<Block>>(
 		&mut self,
 		descendent_of_builder: D,
-		hash: &Block::Hash,
-		number: NumberFor<Block>,
+		_hash: &Block::Hash,
+		_number: NumberFor<Block>,
 	) -> Result<(), fork_tree::Error<D::Error>> {
-		let is_descendent_of = descendent_of_builder
+		let _is_descendent_of = descendent_of_builder
 			.build_is_descendent_of(None);
 
+		// TODO:
 		// prune any epochs which could not be _live_ as of the children of the
 		// finalized block.
 		// i.e. re-root the fork tree to the earliest ancestor of (hash, number)
 		// where epoch.start_slot + epoch.duration >= slot(hash)
+
+		Ok(())
 	}
 
 	/// Finds the epoch for a child of the given block, assuming the given slot number.
--- a/core/consensus/babe/src/epoch_changes.rs
+++ b/core/consensus/babe/src/epoch_changes.rs
@@ -80,6 +80,16 @@ impl<Block: BlockT> EpochChanges<Block> {
 			.and_then(|n| n)
 			.map(|n| n.data.clone())
 	}
+
+	/// Import a new epoch-change, signalled at the given block.
+	pub fn import(
+		&mut self,
+		hash: Block::Hash,
+		number: NumberFor<Block>,
+		epoch: Epoch,
+	) {
+		unimplemented!()
+	}
 }
 
 /// A shared epoch changes tree.
--- a/core/consensus/babe/src/epoch_changes.rs
+++ b/core/consensus/babe/src/epoch_changes.rs
@@ -31,12 +31,12 @@ use client::blockchain::HeaderBackend;
 use primitives::H256;
 
 /// A builder for `is_descendent_of` functions.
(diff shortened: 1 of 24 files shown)


core/consensus/babe/src/epoch_changes.rs
```diff
@@ -0,0 +1,224 @@
...
+		_number: NumberFor<Block>,
+	) -> Result<(), fork_tree::Error<D::Error>> {
+		let _is_descendent_of = descendent_of_builder
+			.build_is_descendent_of(None);
+
+		// TODO:
+		// prune any epochs which could not be _live_ as of the children of the
+		// finalized block.
+		// i.e. re-root the fork tree to the earliest ancestor of (hash, number)
+		// where epoch.start_slot + epoch.duration >= slot(hash)
```
andresilva: `self.inner.prune(hash, number, is_descendent_of)` should work, but I'll verify again before merging this PR.

core/consensus/babe/src/epoch_changes.rs
```diff
@@ -0,0 +1,224 @@
...
+		SharedEpochChanges {
+			inner: Arc::new(Mutex::new(epoch_changes))
+		}
+	}
+}
+
+// TODO: write tests within the scope of this module.
+//
+// 1. mock up an `IsDescendentOfBuilder`. Don't use the `Client` - these tests should be
+// minimal and focused.
```
andresilva: https://github.com/paritytech/substrate/blob/master/core/utils/fork-tree/src/lib.rs#L649

We can mock an arbitrary tree like this.

core/consensus/babe/src/lib.rs
```diff
@@ -197,80 +211,104 @@ pub struct BabeParams<C, E, I, SO, SC> {
...
+	register_babe_inherent_data_provider(&inherent_data_providers, config.slot_duration())?;
 	uncles::register_uncles_inherent_data_provider(
 		client.clone(),
 		select_chain.clone(),
 		&inherent_data_providers,
 	)?;
-	Ok(slots::start_slot_worker(
+
+	let epoch_changes = babe_link.epoch_changes.clone();
+	let pruning_task = client.finality_notification_stream()
```
andresilva: This will only be started by validators this way and it must be run by all full nodes. That's why it was returned with the import queue.

core/consensus/babe/src/lib.rs
```diff
@@ -300,22 +345,23 @@ impl<H, B, C, E, I, Error, SO> slots::SimpleSlotWorker<B> for BabeWorker<C, E, I
...
+		slot_number: SlotNumber,
+		epoch_data: &Epoch,
 	) -> Option<Self::Claim> {
-		let parent_weight = {
-			let pre_digest = find_pre_digest::<B>(&header).ok()?;
-			pre_digest.weight()
-		};
-
-		claim_slot(
+		babe_info!("Attempting to claim slot {}", slot_number);
```
andresilva: Probably downgrade this one and below to `debug` to avoid a lot of noise.

core/consensus/babe/src/lib.rs
```diff
@@ -336,18 +382,18 @@ impl<H, B, C, E, I, Error, SO> slots::SimpleSlotWorker<B> for BabeWorker<C, E, I
...
 				post_digests: vec![signature_digest_item],
 				body: Some(body),
 				finalized: false,
-				auxiliary: Vec::new(),
-				fork_choice: ForkChoiceStrategy::Custom(true),
+				auxiliary: Vec::new(), // block-weight is written in block import.
+				// TODO: block-import handles fork choice and this shouldn't even have the
+				// option to specify one.
+				// https://github.com/paritytech/substrate/issues/3623
+				fork_choice: ForkChoiceStrategy::LongestChain,
```
andresilva: ~~When we are authoring (before import), i.e. by this exact point, do we guarantee that nothing else gets imported in the meantime? I'm worried that we can author a primary block and won't reorg to it because in the meantime we might have imported a secondary block (with longest chain we would keep the secondary at the same height). The previous code was also susceptible to this kind of race condition if it exists.~~

In the meantime I read the code below and figured out that this logic was moved to `BabeBlockImport` which also takes care of the race condition issue.

core/consensus/babe/src/lib.rs
```diff
@@ -1330,37 +1271,38 @@ impl<B, E, Block, I, RA, PRA> BlockImport<Block> for BabeBlockImport<B, E, Block
...
-						next_epoch.epoch_index,
-					)));
-				}
-			}
+		if let Some(next_epoch_descriptor) = next_epoch_digest {
+			let next_epoch = epoch.increment(next_epoch_descriptor);
 
 			old_epoch_changes = Some(epoch_changes.clone());
 
+			babe_info!("New epoch {} launching at block {} (block slot {} >= start slot {}).",
```
andresilva: As above, should these be downgraded?

core/consensus/babe/src/epoch_changes.rs
rphmeier: Would love to share that code!

core/consensus/babe/src/lib.rs
rphmeier: Yeah, I realized that earlier today, which a bit too late. Still, I don't think that forcing the user to spawn a background task is friendly at all. 

I'd rather have us prune implicitly every time we add an epoch, based on the finalized block.

core/consensus/babe/src/lib.rs
rphmeier: I'd prefer these to go in the info logs since we won't get them that often (although due to forks we might get a couple ~duplicates at a time).

PR #3622, opened 2019-09-16
Fix missing `contracts` RPC on a full node.

tomusdrw:
The `ContractsApi` was only created for light-client `rpc_extension`.
I've moved the instantiation to `node/rpc` crate, to avoid duplication.

CC @jacogr

### DIFF
--- a/node/rpc/src/lib.rs
+++ b/node/rpc/src/lib.rs
@@ -29,6 +29,12 @@
 
 #![warn(missing_docs)]
 
+use std::sync::Arc;
+
+use node_primitives::{Block, AccountNonceApi, ContractsApi};
+use sr_primitives::traits::ProvideRuntimeApi;
+use transaction_pool::txpool::{ChainApi, Pool};
+
 pub mod accounts;
 pub mod contracts;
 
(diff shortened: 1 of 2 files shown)


node/rpc/src/lib.rs
```diff
@@ -38,3 +44,27 @@ mod constants {
 	/// This typically means that the runtime trapped.
 	pub const RUNTIME_ERROR: i64 = 1;
 }
+
+/// Instantiate all RPC extensions.
+pub fn create<C, P, M>(client: Arc<C>, pool: Arc<Pool<P>>) -> jsonrpc_core::IoHandler<M> where
```
pepyakin: If we expose only `create`, should we then unmark `self::{accounts, countracts}` as `pub`?

tomusdrw: I suppose `account` RPCs are still valuable on runtimes with no contracts support and vice-versa, right? That will allow people to import `node-rpc` and just instantiate one.

pepyakin: Ah, yeah, good point. Although I didn't realize that `node` is a part of public API.

PR #3626, opened 2019-09-17
Fix state RPC subscriptions on light node

svyatonik:
This PR introduces light-specific implementation of `state_subscribeRuntimeVersion`/`state_unsubscribeRuntimeVersion` and `state_subscribeStorage`/`state_unsubscribeStorage`. Implementation for full node is moved from `mod.rs` to `state_full.rs` as-is.

Short overview of light-specific implementation: for every imported block we send only one remote-request message. The exception is the initial value request (for the current best block) - we're issuing it for every new subscription. After response is received, every listener is notified. It then filters (if required) the response and passes it to the subscriber.

So here are two major improvements (in addition to fix itself): (1) previously, we were requesting every key with its own message and (2) previously we were requesting values of all keys of all subscriptions with separate messages. And now there's only one message for all subscriptions, for all keys.


core/rpc/src/state/state_light.rs
```diff
@@ -267,31 +316,469 @@ impl<Block, F, B, E, RA> StateBackend<B, E, Block, RA> for LightState<Block, F,
...
 		_meta: Option<crate::metadata::Metadata>,
-		_id: SubscriptionId,
+		id: SubscriptionId,
 	) -> RpcResult<bool> {
-		Ok(false)
+		if !self.subscriptions.cancel(id.clone()) {
+			return Ok(false);
+		}
+
+		// forget subscription keys
```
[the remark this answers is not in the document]
tomusdrw: Yes, the `unsubscribe` method is always called - both when the client calls it explicitly or just disconnects.

core/rpc/src/state/state_light.rs
```diff
@@ -147,26 +207,15 @@ impl<Block, F, B, E, RA> StateBackend<B, E, Block, RA> for LightState<Block, F,
...
-
-		Box::new(storage.boxed().compat())
+		Box::new(storage(
+			&*self.remote_blockchain,
+			self.fetcher.clone(),
+			self.block_or_best(block),
+			vec![key.0.clone()],
+		).boxed().compat().map(move |mut values| values
+			.remove(&key)
+			.expect("successful request has entries for all requested keys; qed")
```
tomusdrw: Is that really verified somewhere? I wonder if a malicious full node sends us an empty response and we panic here.

svyatonik: Yes - there's so called `FetchChecker` which checks every response from remote node on light client. For remote read it tries to read every requested value from the response && fails if there's something wrong with the response. Otherwise - value is inserted into `HashMap`. This happens [here](https://github.com/paritytech/substrate/blob/master/core/state-machine/src/lib.rs#L624). So if everything goes that way, this expect should be valid

PR #3627, opened 2019-09-17
Suggest `--release` compilation

shawntabrizi:
If the user sees a warning about `block production took too long` and we detect the user has compiled their node with `debug` mode, suggest to them to instead compile with `--release` mode for computational optimizations which could remove problems with block production taking too long.

```
2019-09-17 13:02:46 Discarding proposal for slot 522906053; block production took too long
2019-09-17 13:02:46 Recompile your node in `--release` mode to mitigate this problem.
2019-09-17 13:02:47 Starting consensus session on top of parent 0x059e1fd59e4e4829eaa33e61696da472b79109056845767965afd14ccaa95f2a
```

### DIFF
--- a/core/consensus/slots/src/lib.rs
+++ b/core/consensus/slots/src/lib.rs
@@ -208,6 +208,9 @@ pub trait SimpleSlotWorker<B: BlockT> {
 			let slot_after_building = SignedDuration::default().slot_now(slot_duration);
 			if slot_after_building != slot_number {
 				info!("Discarding proposal for slot {}; block production took too long", slot_number);
+				// If the node was compiled with debug, tell the user to use release optimizations.
+				#[cfg(debug_assertions)]
+				info!("Recompile your node in `--release` mode to mitigate this problem.");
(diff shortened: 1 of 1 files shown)


kigawas:
what is the root cause of "block production taking too long"?

pepyakin:
That's a good question. I am unaware of any attempts to dig deep into this issue.

My hypothesis though is that there are a few really tight loops (such as the wasm interpreter dispatch loop or hashing) and inefficiencies in the way they are compiled provide enormous slowdowns so that most of machines cannot execute them in the currently set time slot.

kianenigma:
that being said, running release is somewhat  analogous to purging chain when there is a state error. Very frequently it is the reason. Yet, it would have been better if the log would make it clear that running --release _could_ solve this, not for sure.

xlc:
previously (1.0), runtime wasm are always built with release mode. but now by default runtime wasm is built with debug mod with `cargo run -- --dev` which I believe contributes a lot of the slowness. `WASM_BUILD_TYPE=release cargo run -- --dev` mimetic the old behaviour. another part of the slowness could be introduced by BABE VRF.

PR #3630, opened 2019-09-17
Properly detect debug build in slots

bkchr:
`debug-assertions` can also be enabled in for release builds. This
introduces a new build-script that extracts the build type from the
`PROFILE` env variable and sets the `build_type` cfg.


andresilva:
> @pepyakin That said, I guess it is the best thing we have.

Half an hour later, I guess not!

pepyakin:
Glad to be proven wrong!

PR #3631, opened 2019-09-17
Update parity-wasm to 0.40

### LINKED ISSUE #2517: Update dependency on parity-wasm
Blocked on:

- https://github.com/paritytech/wasm-utils/issues/119
- https://github.com/paritytech/wasmi/issues/168

expenses:
I updated the direct `parity-wasm` dependencies from `0.31` to `0.40`. This resolves #2517 (I think). There are still some indirect dependencies, see the output of `cargo tree -p parity-wasm:0.31.3 -i`.

I'm getting a few failed tests where the gas left is up by 1, but I'm not sure if these are really errors or not.

```
---- wasm::tests::contract_transfer stdout ----
thread 'wasm::tests::contract_transfer' panicked at 'assertion failed: `(left == right)`
  left: `[TransferEntry { to: 9, value: 6, data: [1, 2, 3, 4], gas_left: 49971 }]`,
 right: `[TransferEntry { to: 9, value: 6, data: [1, 2, 3, 4], gas_left: 49970 }]`', srml/contracts/src/wasm/mod.rs:468:3
```

### DIFF
--- a/core/executor/src/wasm_runtimes_cache.rs
+++ b/core/executor/src/wasm_runtimes_cache.rs
@@ -99,7 +99,12 @@ impl StateSnapshot {
 				// anyway.
 				let contents = mem::replace(segment.value_mut(), vec![]);
 
-				let init_expr = segment.offset().code();
+				let init_expr = match segment.offset() {
+					Some(offset) => offset.code(),
+					// Return if the segment is passive
+					None => return None
+				};
+
 				// [op, End]
 				if init_expr.len() != 2 {
 					return None;
--- a/srml/contracts/src/wasm/mod.rs
+++ b/srml/contracts/src/wasm/mod.rs
@@ -471,7 +471,7 @@ mod tests {
 				to: 9,
 				value: 6,
 				data: vec![1, 2, 3, 4],
-				gas_left: 49970,
+				gas_left: 49971,
 			}]
 		);
 	}
@@ -533,7 +533,7 @@ mod tests {
 				code_hash: [0x11; 32].into(),
 				endowment: 3,
 				data: vec![1, 2, 3, 4],
-				gas_left: 49946,
+				gas_left: 49947,
 			}]
 		);
 	}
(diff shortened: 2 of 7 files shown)


core/executor/src/wasm_runtimes_cache.rs
```diff
@@ -99,7 +99,12 @@ impl StateSnapshot {
 				// anyway.
 				let contents = mem::replace(segment.value_mut(), vec![]);
 
-				let init_expr = segment.offset().code();
+				let init_expr = match segment.offset() {
+					Some(offset) => offset.code(),
+					// Return if the segment is passive
+					None => return None
```
expenses: I'm not sure what the correct behaviour here is.

expenses:
[clabot:check]

core/executor/src/wasm_runtimes_cache.rs
pepyakin: Yeah, I think this should do. Passive segments are only relevant to either bulk-memory ops or threading and they are both related to non-MVP features which we don't support yet.

pepyakin:
My initial guess is that it should be OK, reasoning for that the gas metering algorithm was altered in the latest version so that it doesn't insert some superfluous metering calls.

I'd like to check that though.

pepyakin:
It looks OK, so you can just update the expected values. Apart from that the check `continuous-integration/gitlab-check-runtime` fails and in order to please it you should update the runtime version [here](https://github.com/paritytech/substrate/blob/80a80953311aad10719a471c5ccebdf5406a2e37/node/runtime/src/lib.rs#L81-L86)

srml/contracts/src/wasm/mod.rs
```diff
@@ -1359,11 +1359,7 @@ mod tests {
 			vec![0x00, 0x01, 0x2a, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xe5, 0x14, 0x00])
 		]);
 
-		assert_eq!(gas_meter.gas_left(), 50_000
-			- 6            // Explicit
-			- 13 - 1 - 1   // Deposit event
-			- (13 + 33)    // read memory
-		);
+		assert_eq!(gas_meter.gas_left(), 49934);
```
expenses: The other tests just used raw numbers instead of calculating what the remaining gas should be, so I think this should be ok.

PR #3633, opened 2019-09-17
Explicit sync API for downloading important, possibly orphaned, forks

arkpar:
Sync services provides an API to download specified block chains explicitly.


core/network/src/protocol/sync.rs
```diff
@@ -441,6 +441,53 @@ impl<B: BlockT> ChainSync<B> {
 		})
 	}
 
+	/// Request syncing for the given block from given set of peers.
+	// This is similar to on_block_announce with unknown parent hash.
+	pub fn sync_fork(&mut self, peers: Vec<PeerId>, hash: &B::Hash, number: NumberFor<B>) ->
+		Vec<(PeerId, BlockRequest<B>)>
+	{
+		debug!(target: "sync", "Explicit sync request for block {:?} with {:?}", hash, peers);
+		if self.is_known(&hash) || self.is_already_downloading(&hash) {
```
rphmeier: I'd say we need something a bit more explicit. The fork-sync requests should be kept around _forever_ until removed or the block is actually known, and the `is_already_downloading` check should be done periodically. We can imagine a case where `is_already_downloading` is true at the moment, but then for some reason the node stops downloading that block. Then our requested block still wouldn't be synced.

core/network/src/protocol/sync.rs
```diff
@@ -441,6 +441,53 @@ impl<B: BlockT> ChainSync<B> {
...
+			debug!(target: "sync", "Refusing to sync known hash {:?}", hash);
+			return Vec::default();
+		}
+
+		let block_status = self.client.block_status(&BlockId::Number(number - One::one()))
+			.unwrap_or(BlockStatus::Unknown);
+		if block_status == BlockStatus::InChainPruned {
```
rphmeier: As a follow-up, there should be a special mode where we just download the headers for this chain if requested, but we don't have the state to process them.

rphmeier:
It's also missing the APIs for higher-level code to inform the sync service about which peers it knows has the block.

arkpar:
> It's also missing the APIs for higher-level code to inform the sync service about which peers it knows has the block. 

But `sync_fork` does exactly that. 
What's the case for remove? Peer forgets a block?
Sync tries to download all forks that other peers have and we don't. It does not matter if GRANDPA is interested in the block or not. Do you suggest changing that?

rphmeier:
The set may change over time, is what I mean - are multiple calls guaranteed to alter the existing "download session" so that it will sync the fork from the union of the given peer-sets?

arkpar:
No, but I'm going to implement that along with request persistence.

rphmeier:
> It does not matter if GRANDPA is interested in the block or not. Do you suggest changing that?

In general, I suggest sync being a little bit smarter about figuring out what "all forks" are based on what GRANDPA is interested in. Even if the fork is no longer active it should constantly be trying to sync it, even if attempts fail.

core/network/src/protocol/sync.rs
```diff
@@ -441,6 +441,53 @@ impl<B: BlockT> ChainSync<B> {
...
+				debug!(target: "sync", "Called sync_fork with a bad peer ID");
+				continue;
+			};
+			if let PeerSyncState::AncestorSearch(_, _) = peer.state {
+				continue;
+			}
+
+			if number > peer.best_number {
+				peer.best_number = number;
+				peer.best_hash = hash.clone();
```
rphmeier: why set that to the peer's best block?

arkpar: `best_number` is used to track global sync target. `best_hash` is not used for anything really, other than reporting generic peer info.

core/network/src/protocol/sync.rs
rphmeier: https://github.com/paritytech/substrate/issues/3639 for a more detailed writeup of ^

arkpar:
Fork download requests are now persistent. Sync will continue trying even if peer reconnects.

core/network/src/protocol.rs
```diff
@@ -728,7 +728,7 @@ impl<B: BlockT, S: NetworkSpecialization<B>, H: ExHashT> Protocol<B, S, H> {
 			match request.direction {
 				message::Direction::Ascending => id = BlockId::Number(number + One::one()),
 				message::Direction::Descending => {
-					if number.is_zero() {
+					if number.is_one() {
```
rphmeier: if it is zero, will that cause an error further down the line when `parent_hash` gives `None` for the header/block?

core/network/src/protocol/sync.rs
```diff
@@ -443,49 +452,41 @@ impl<B: BlockT> ChainSync<B> {
...
-			return Vec::default();
+			return;
 		}
 
 		let block_status = self.client.block_status(&BlockId::Number(number - One::one()))
 			.unwrap_or(BlockStatus::Unknown);
 		if block_status == BlockStatus::InChainPruned {
 			trace!(target: "sync", "Refusing to sync ancient block {:?}", hash);
-			return Vec::default();
+			return;
```
rphmeier: this is also not the only kind of ancient fork we can encounter

core/network/src/protocol/sync.rs
```diff
@@ -443,49 +452,41 @@ impl<B: BlockT> ChainSync<B> {
...
+			if let Some(peer) = self.peers.get_mut(peer_id) {
+				if let PeerSyncState::AncestorSearch(_, _) = peer.state {
+					continue;
+				}
 
-			if number <= self.best_queued_number {
-				if let Some(request) = self.download_unknown_stale(&peer_id, &hash) {
-					requests.push((peer_id, request));
+				if number > peer.best_number {
+					peer.best_number = number;
```
rphmeier: What is the invariant that sync is meant to uphold? 

Is this always supposed to be the peer's highest block, or is it the peer's best chain? We do have some non longest-chain fork choice rules (1. follow finality, 2. aurababous), so they are not always the same thing.

core/network/src/protocol/sync.rs
```diff
@@ -1258,3 +1270,32 @@ fn peer_block_request<B: BlockT>(
...
 	}
 }
+
+/// Get pending explicit sync request for a peer.
+fn explicit_sync_request<B: BlockT>(
+	id: &PeerId,
+	requests: &HashMap<B::Hash, SyncRequest<B>>,
+	best_num: NumberFor<B>,
+	finalized: &B::Hash,
+	attributs: &message::BlockAttributes,
```
rphmeier: typo: `attributes`?

core/network/src/protocol/sync.rs
```diff
@@ -1258,3 +1270,32 @@ fn peer_block_request<B: BlockT>(
...
+) -> Option<(B::Hash, BlockRequest<B>)>
+{
+	for (hash, r) in requests {
+		if !r.peers.contains(id) {
+			continue
+		}
+		if r.number <= best_num {
```
rphmeier: questionable for the same reason as before - the orphan chain may be longer than the peer's "best" chain because of the fork-choice rule.

rphmeier:
Looks good but needs accounting for fork-choice rules

core/network/src/protocol/sync.rs
arkpar: Sync attempts to download the chain up to the last finalized block, not just one block. There's currently a limit of how many blocks these forks thare downloaded up the chain can be. We can't be allowed to download infinite chains of headers/blocks without validating them.

core/network/src/protocol/sync.rs
arkpar: This is just the highest block currently. The sync tracks the longest fork a bit differently from the shorter forks. E.g. it supports parallel block fetch for the longest chain only. It still fetches all forks though so it should not matter which one is actually "best".  There will be a refactoring to get rid of the "main" chain notion, but this is out of scope of this PR.

core/network/src/protocol/sync.rs
arkpar: The other case (`r.number > best_num`) is handled in a different code path in `peer_block_request`.

core/network/src/protocol.rs
arkpar: `parent_hash` is not an `Option`. for the genesis block it is a zero hash.

arkpar: That change is still wrong though. We should allow downloading the genesis block in principle.

core/network/src/protocol/sync.rs
rphmeier: Well, if it's supposed to be the highest block, then the code here is wrong and I wouldn't want to merge it

core/network/src/protocol/sync.rs
rphmeier: > Sync attempts to download the chain up to the last finalized block, not just one block

I don't understand what you mean. Yes, there has to be a practical cutoff somewhere, and I think refusing to download forks beyond `finalized_height + constant_allowance` would be reasonable (with `constant_allowance` around 1000). 

But the most recent finalized block of the fork is going to be the common ancestor (otherwise it wouldn't be a fork). So we always have to download beyond that.

core/network/src/protocol/sync.rs
rphmeier: This is confusing. Why is the logic so scattered?

rphmeier: What I mean is that there's a lot of implicit "but actually" in the interface of this function. 

It's an explicit sync request...but actually only returns explicit requests for chains shorter than the best.

So the peer doesn't download that chain...but actually it does, it's covered in another function in another file.

rphmeier: I'd recommend that the function return a special type when there is implicit context that needs to be passed from the execution of this function into the execution of another, where the reader of the code can clearly see that the handling of some case has been deferred beyond the execution of the function.

arkpar:
After a chat with @rphmeier we've agreed to address all remaining issues in the following PRs

core/network/src/protocol/sync.rs
arkpar: This is fundamentally caused by the same issue described here: https://github.com/paritytech/substrate/pull/3633#discussion_r326888400
To be refactored.

core/network/src/service.rs
```diff
@@ -496,6 +496,18 @@ impl<B: BlockT + 'static, S: NetworkSpecialization<B>, H: ExHashT> NetworkServic
 		Ok(())
 	}
 
+	/// Adds a `PeerId` and its address as reserved.
```
mxinden:
```suggestion
```
This is a copy/paste error from above, right?

core/network/src/protocol/sync.rs
```diff
@@ -449,6 +458,51 @@ impl<B: BlockT> ChainSync<B> {
 		})
 	}
 
+	/// Request syncing for the given block from given set of peers.
+	// This is similar to on_block_announce with unknown parent hash.
```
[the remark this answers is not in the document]
arkpar: This comment is about the implementation details, not the interface.

core/network/src/protocol/sync.rs
```diff
@@ -449,6 +458,51 @@ impl<B: BlockT> ChainSync<B> {
...
+	pub fn sync_fork(&mut self, peers: Vec<PeerId>, hash: &B::Hash, number: NumberFor<B>) {
+		if peers.is_empty() {
```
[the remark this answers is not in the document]
arkpar: renamed to `set_fork_sync_request`

Issue #3615, opened 2019-09-14
Fail to building Substrate on ARM64


akru:
**Error reproduction**

1. Get tiny ARM64 server on Scaleway with preinstalled Ubuntu 18.04. 
2. Using `rustup` install toolchain and wasm target.
3. Install `wasm-gc` using cargo.
4. Starting to build: `cargo +nightly build --release`
5. Get `Rust WASM toolchain not installed, please install it!`

**Explanation**

New `wasm-builder` module detects WASM toolchain by compiling simple library:
https://github.com/paritytech/substrate/blob/master/core/utils/wasm-builder/src/prerequisites.rs#L51

When repeat this step locally I get this problem:
```
   Compiling wasm-test v1.0.0 (/tmp/testproj)
error: linker `rust-lld` not found
  |
  = note: No such file or directory (os error 2)
```

> It's means that `aarch64-unknown-linux-gnu` target from `rustup` distribution unfortunately doesn't provide linker (linker provided for Tier 1 platforms only).

I was try to use LLVM8 from Ubuntu distribution linker:
```
apt install lld-8

RUSTFLAGS="-Clinker=lld-8" cargo +nightly build --release --target=wasm32-unknown-unknown
```

But get problem with unknown section type in WASM binaries:
```
  = note: lld-8: error: /root/.rustup/toolchains/nightly-aarch64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-unknown/lib/libstd-4bce71b
eddcc0955.rlib(std-4bce71beddcc0955.std.bwpn9z31-cgu.0.rcgu.o): Bad section type
```

Finally,
- I build `lld` from scratch using this script: https://github.com/mcrosson/Plume/blob/dfa89e227a986d64c56b265490bf97e3597fc177/script/wasm-deps.sh
- Link it to /usr/bin/rust-lld
 and it **works well**.

akru:
I'm not sure that it's substrate problem issue,
I think its just a point for discussion about substrate 2.0 porting to ARM64.

PR #3635, opened 2019-09-17
Make `wasm-builder` check for `rust-lld`

bkchr:
Closes: https://github.com/paritytech/substrate/issues/3615

@akru could you check that the error message tells you to install `rust-lld` now? :)

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -195,7 +195,7 @@ sudo apt install cmake pkg-config libssl-dev git clang libclang-dev
 
 - Linux on ARM:
 `rust-lld` is required for linking wasm, but is missing on non Tier 1 platforms.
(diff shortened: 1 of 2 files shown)


README.adoc
```diff
@@ -193,6 +193,11 @@ You will also need to install the following packages:
 [source, shell]
 sudo apt install cmake pkg-config libssl-dev git clang libclang-dev
 
+- Linux on ARM:
+`rust-lld` is required for linking wasm, but is missing on non Tier 1 platforms.
+So, use this https://github.com/mcrosson/Plume/blob/dfa89e227a986d64c56b265490bf97e3597fc177/script/wasm-deps.sh[script]
```
riusricardo:
```suggestion
So, use this https://github.com/Plume-org/Plume/blob/master/script/wasm-deps.sh[script]
```
It is better if we link the original script.

PR #3636, opened 2019-09-17
use pdqselect for median_algorithm selection

### LINKED ISSUE #2926: Use a selection algorithm instead of a sorting algorithm in median algorithm
The current code uses Rust’s sort routines, which are O(n log n), as opposed to a proper selection algorithm, which is O(n).

This is very low priority, as I cannot see this being a bottleneck in any practical scenario.  I suspect that unless such an algorithm is added to the standard library, this issue is probably not worth fixing.

expenses:
This would resolve #2926, but `median_algorithm` isn't currently used. I haven't written a test yet, but I'd like to if it's important enough.

### DIFF
--- a/core/consensus/babe/src/lib.rs
+++ b/core/consensus/babe/src/lib.rs
@@ -110,7 +110,6 @@ use slots::{CheckedHeader, check_equivocation};
 use futures::prelude::*;
 use futures01::Stream as _;
 use log::{error, warn, debug, info, trace};
-use pdqselect::select;
 
 use slots::{SlotWorker, SlotData, SlotInfo, SlotCompatible};
 
--- a/core/consensus/babe/src/lib.rs
+++ b/core/consensus/babe/src/lib.rs
@@ -110,6 +110,7 @@ use slots::{CheckedHeader, check_equivocation};
 use futures::prelude::*;
 use futures01::Stream as _;
 use log::{error, warn, debug, info, trace};
+use pdqselect::select;
 
 use slots::{SlotWorker, SlotData, SlotInfo, SlotCompatible};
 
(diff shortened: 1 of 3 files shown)


gavofyork:
> This would resolve #2926 

would it?

https://docs.rs/pdqselect/0.1.0/pdqselect/fn.select.html

> "This sort is in-place, unstable, and O(n log n) worst-case"

expenses:
I ran a simple benchmark comparing getting the median with both `select` and `unsafe_sort` [here](https://gist.github.com/expenses/d2d84305aeaa512200676dc0198e9f81#file-bench-rs). Results:

```
test bench_select ... bench:       1,898 ns/iter (+/- 208)
test bench_sort   ... bench:       9,288 ns/iter (+/- 1,340)
```

rphmeier:
Thanks! Mind if I ice this until #3583 is in, since it contains a much larger change-set on BABE that this might conflict with?

expenses:
Of course!

core/consensus/babe/src/lib.rs
```diff
@@ -110,6 +110,7 @@ use slots::{CheckedHeader, check_equivocation};
 use futures::prelude::*;
 use futures01::Stream as _;
 use log::{error, warn, debug, info, trace};
+use pdqselect::select;
```
andresilva:
```suggestion
```

core/consensus/babe/src/lib.rs
```diff
@@ -717,8 +718,8 @@ fn median_algorithm(
 			t + Duration::new(secs, nanos)
 		}).collect();
 
-		// FIXME #2926: use a selection algorithm instead of a full sorting algorithm.
-		new_list.sort_unstable();
+		// Use a partial sort to move the median timestamp to the middle of the list
+		select(&mut new_list, num_timestamps / 2);
```
andresilva:
```suggestion
		pdqselect::select(&mut new_list, num_timestamps / 2);
```

andresilva: nitpick: maybe easier to understand what this is just by reading this line alone. feel free to ignore it though :)

rphmeier: yes, i find this style much better for free functions - it means that whenever you see a function call without a module prefix, that it's defined in the same file as you're reading right now.

rphmeier:
Thanks!

PR #3637, opened 2019-09-18
Remove the no_std feature from the WASM builder

tomaka:
Removes the fact that the WASM builder requires the `no_std` feature.

The motivation behind this change is the surprise effect. Why do I need a `no_std` feature in my runtime?
Additionally, semantically it's not a great idea to have a "no_std" feature. People might misuse it because they think it means "opposite of std feature", while it is not that at all. There is a history in the Rust ecosystem of misusing features named "no_std", and we're giving out a bad example.

In practice, Polkadot doesn't use that feature, only the Substrate runtime does.

The only consequence might be that we have to activate features on "std" that we wouldn't normally need, which will result in a very slightly larger binary size. I really don't think that's a problem.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2548,7 +2548,7 @@ dependencies = [
  "substrate-offchain-primitives 2.0.0",
  "substrate-primitives 2.0.0",
  "substrate-session 2.0.0",
- "substrate-wasm-builder-runner 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "substrate-wasm-builder-runner 1.0.3",
 ]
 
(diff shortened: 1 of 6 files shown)


tomaka:
So, this doesn't seem possible because we're using `wasm-validation 0.1` in srml/contracts instead of the newest 0.2.
I don't really want to dig down that rabbit hole.

pepyakin:
FWIW, the master is like this
https://github.com/paritytech/substrate/blob/5c9a84a12ffd96b3002d282190e808efdd89c1f8/srml/contracts/Cargo.toml#L12

PR #3640, opened 2019-09-18
Remove the no_std feature from the WASM builder

tomaka:
(GitHub won't let me reopen #3637 because I force-pushed, so I have to create another PR)

Removes the fact that the WASM builder requires the no_std feature.

The motivation behind this change is the surprise effect. Why do I need a no_std feature in my runtime?
Additionally, semantically it's not a great idea to have a "no_std" feature. People might misuse it because they think it means "opposite of std feature", while it is not that at all. There is a history in the Rust ecosystem of misusing features named "no_std", and we're giving out a bad example.

In practice, Polkadot doesn't use that feature, only the Substrate runtime does.

The only consequence might be that we have to activate features on "std" that we wouldn't normally need, which will result in a very slightly larger binary size. I really don't think that's a problem.

### DIFF
--- a/node-template/runtime/Cargo.toml
+++ b/node-template/runtime/Cargo.toml
@@ -28,7 +28,7 @@ client = { package = "substrate-client", path = "../../core/client", default_fea
 offchain-primitives = { package = "substrate-offchain-primitives", path = "../../core/offchain/primitives", default-features = false }
 
 [build-dependencies]
-wasm-builder-runner = { package = "substrate-wasm-builder-runner", version = "1.0.2", path = "../../core/utils/wasm-builder-runner" }
+wasm-builder-runner = { package = "substrate-wasm-builder-runner", version = "1.0.2" }
 
 [features]
 default = ["std"]
--- a/node-template/runtime/Cargo.toml
+++ b/node-template/runtime/Cargo.toml
@@ -28,7 +28,7 @@ client = { package = "substrate-client", path = "../../core/client", default_fea
 offchain-primitives = { package = "substrate-offchain-primitives", path = "../../core/offchain/primitives", default-features = false }
 
 [build-dependencies]
-wasm-builder-runner = { package = "substrate-wasm-builder-runner", version = "1.0.2" }
+wasm-builder-runner = { package = "substrate-wasm-builder-runner", version = "1.0.2", path = "../../core/utils/wasm-builder-runner" }
 
 [features]
 default = ["std"]
@@ -55,4 +55,3 @@ std = [
 	"offchain-primitives/std",
 	"substrate-session/std",
 ]
-no_std = []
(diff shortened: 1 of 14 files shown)


bkchr:
@tomaka needs some fixing.

node-template/runtime/build.rs
```diff
@@ -19,7 +19,11 @@ use wasm_builder_runner::{build_current_project_with_rustflags, WasmBuilderSourc
 fn main() {
 	build_current_project_with_rustflags(
 		"wasm_binary.rs",
-		WasmBuilderSource::Crates("1.0.6"),
+		// Note: replace with `WasmBuilderSource::Crates("1.0.6")` if you copy-paste this template.
+		WasmBuilderSource::CratesOrPath {
```
bkchr: Please up the version of wasm-builder to `1.0.7`, publish it to crates.io and then restore the old behavior here.

node-template/runtime/Cargo.toml
```diff
@@ -28,7 +28,7 @@ client = { package = "substrate-client", path = "../../core/client", default_fea
 offchain-primitives = { package = "substrate-offchain-primitives", path = "../../core/offchain/primitives", default-features = false }
 
 [build-dependencies]
-wasm-builder-runner = { package = "substrate-wasm-builder-runner", version = "1.0.2" }
+wasm-builder-runner = { package = "substrate-wasm-builder-runner", version = "1.0.2", path = "../../core/utils/wasm-builder-runner" }
```
bkchr:
```suggestion
wasm-builder-runner = { package = "substrate-wasm-builder-runner", version = "1.0.2" }
```

node/runtime/Cargo.toml
```diff
@@ -30,7 +30,7 @@ authorship = { package = "srml-authorship", path = "../../srml/authorship", defa
 babe = { package = "srml-babe", path = "../../srml/babe", default-features = false }
 balances = { package = "srml-balances", path = "../../srml/balances", default-features = false }
 collective = { package = "srml-collective", path = "../../srml/collective", default-features = false }
-contracts = { package = "srml-contracts", path = "../../srml/contracts", default-features = false }
+contracts = { package = "srml-contracts", path = "../../srml/contracts", default-features = false, features = ["core"] }
```
bkchr: Please remove this `core` feature from `srml-contracts`. There is no reason we have this feature anymore.

bkchr:
And please merge master, before releasing to crates.io

thiolliere:
does this breaks adder crate in polkadot ? https://github.com/paritytech/polkadot/blob/master/test-parachains/adder/Cargo.toml

bkchr:
@thiolliere yes, https://github.com/paritytech/polkadot/pull/441 this fixes it.

PR #3644, opened 2019-09-18
Make staking inflation curve configurable.

thiolliere:
# Context:

currently the curve used to compute the reward of the validators and nominators is fixed in srml-staking. This PR makes it configurable at the runtime (node-runtime) level.

# Done:

* Definition of the curve has changed to just a list of points: (Perbill, Perbill), this allows to define any function from [0, 1] to [0, 1] even the curve is very steep. (The limitation for `[0, 1] -> [0, 1]` is to avoid difficulties with overflows)
* definition of PiecewiseLinear is moved to sr-primitives, it uses its own function multiply_by_rational_saturating. Maybe we want to make this latter function available publicly.
* to declare a curve easily a proc-macro crate is introduced at srml-staking-reward-curve.
  this crate generate a test for the precision alongside the declaration.
* the generation of the curve has changed for efficiency reason because the time to compile is important.
  new algorithm just divide into variable segment with some maximum size.


gavofyork:
Awesome :)

Will need a polkadot update PR but looking very good otherwise.

srml/staking/reward-curve/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "srml-staking-reward-curve"
+version = "2.0.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+# srml-support-procedural-tools = { package = "srml-support-procedural-tools", path = "./tools" }
```
kianenigma:
```suggestion

```

core/sr-primitives/src/curve.rs
```diff
@@ -0,0 +1,164 @@
...
+		} else {
+			(prev.1 * d).saturating_sub(delta_y)
+		}
+	}
+}
+
+// Compute value * p / q.
+// This is guarantee not to overflow on whatever values nor loose precision.
+// `q` must be superior to zero.
+fn multiply_by_rational_saturating<N>(value: N, p: u32, q: u32) -> N
```
kianenigma: this stuff is okay for now but as me want to refactor more of the arithmatic stuff and move them into their own crate, I wonder: 

What is the accuracy requirement here? if p and q are always u32 and we want to be able to accurately compute it, then perquintillion is the option to go atm. 

On the other hand, that's a bit of on overkill and has some redundant bytes. If this pattern (`a * p / q` where `p` and `q` are `u32`) is common then actually making the `per-u32` might be worthwhile. It will also be a challenge to see if the macros in the sr-arithmatic refactor actually work or now 😀

core/sr-primitives/src/curve.rs
```diff
@@ -0,0 +1,164 @@
...
+			(prev.1 * d).saturating_add(delta_y)
+		// Otherwise result is negative
+		} else {
+			(prev.1 * d).saturating_sub(delta_y)
+		}
+	}
+}
+
+// Compute value * p / q.
+// This is guarantee not to overflow on whatever values nor loose precision.
```
kianenigma: Should we add this note: `// Mul can saturate if p > q`? meaning that this is 100% safe and accurate given `p < q`

srml/staking/reward-curve/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+///   Expressed in millionth, must be between 0 and 1_000_000.
+///
+/// - `max_piece_count`: The maximum number of pieces in the curve. A greater number uses more
+///   resources but results in higher accuracy.
+///   Must be between 2 and 1_000.
+///
+/// - `test_precision`: The maximum error allowed in the generated test.
+///   Expressed in millionth, must be between 0 and 1_000_000.
+///
+/// # Example
```
kianenigma: cc/ @drewstone ^^

srml/staking/reward-curve/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+///   of total issuance. Known as `I_0` in the literature.
+///   Expressed in millionth, must be between 0 and 1_000_000.
+///
+/// - `max_inflation`: the maximum amount to be rewarded between validators, expressed as a fraction
+///   of total issuance. This is attained only when `ideal_stake` is achieved.
+///   Expressed in millionth, must be between min_inflation and 1_000_000.
+///
+/// - `ideal_stake`: the fraction of total issued tokens that should be actively staked behind
+///   validators. Known as `x_ideal` in the literature.
+///   Expressed in millionth, must be between 0_100_000 and 0_900_000.
```
kianenigma: what do you mean by 

> must be between 0_100_000 and 0_900_000.
 
or is it a type?

srml/staking/reward-curve/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+			delta_y = delta_y.saturating_sub(1);
+			continue
+		}
+
+		if next_x >= 1_000_000 {
+			let prev = points.last().unwrap();
+			// Compute the y corresponding to x=1_000_000 using the this point and the previous one.
+			let y = next_y + (
+				(next_x - 1_000_000) as u128
+				* (prev.1 - next_y) as u128
```
kianenigma: won't u64 be enough for this?

srml/staking/reward-curve/src/log.rs
```diff
@@ -0,0 +1,67 @@
+/// Return Per-million value.
+pub fn log2(p: u32, q: u32) -> u32 {
```
kianenigma: this probably should be placed in sr-primitives?

kianenigma: (or now I see your assertions and comments below about how __this is not a general purpose log2__. Maybe let it be here and at some point move to sr-primitives if it became general purpose.)

core/sr-primitives/src/curve.rs
thiolliere: > What is the accuracy requirement here? if p and q are always u32 and we want to be able to accurately compute it, then perquintillion is the option to go atm.

well doesn't perquintillion::from_rational_approximation does loose a bit because like 1/3 is 33333.../quintillion but maybe we can afford this.
Or maybe now that I think of it we should go the other way around make perthings use this function 🤔

core/sr-primitives/src/curve.rs
thiolliere: At some point it is always 100% safe just when your result doesn't fit into N then it saturates with N::max_value

srml/staking/reward-curve/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+		if y == self.i_0 {
+			return u32::max_value();
+		}
+		let log = log2(self.i_ideal_times_x_ideal - self.i_0, y - self.i_0);
+
+		self.x_ideal + ((self.d as u64 * log as u64) / 1_000_000) as u32
+	}
+}
+
+fn compute_points(input: &INposInput) -> Vec<(u32, u32)> {
```
kianenigma: So I assume that if the arithmetic in such functions overflow (I know that a sensible curve should work fine, but assume the user makes a mistake) then it will be a compile-time error. Yet, we might have silent errors where `as` conversion is lossy (I haven't found a concrete example but I sense the potential --worth double-checking) is used. To also prevent such wrong curves to be compiled, I would recommend replacing a potentially wrong as with `try_into().unwrpa_or_else(|_| compile_error!(...))`

srml/staking/reward-curve/src/lib.rs
thiolliere: it is just a visual way to write permillionth.
I'm parsing a u32 anyway to one could write between 0.1 and 0.9
At some point I might be able to actually make a custom parser with tokens but I feel like just taking the permillions representation is ok

core/sr-primitives/src/curve.rs
kianenigma: I think what you mean is not what I meant by _accuracy_, but rather a _rounding error_ of at most `1` unit which, imo, we can afford. What I mean is: 

The most fine-grained approximation of this function is `p = 1, q = u32::MAX`. This will be roughly 1 divided by 4 billion. Naturally, per-1-billion is inherently incapable of representing this number while a per quintillion is. This is why to abstract this, something like per-billion is inaccurate, per quintillion is slightly wasteful while per-u32 is ideal.

kianenigma:
> Some refactor suggestions but overall seems to deliver what it intended to do --**make the curve configurable**-- perfectly :) LGTM

(just saw the label being in-progress -- ^^ given that it is actually done.)

core/sr-primitives/src/curve.rs
kianenigma: My misunderstanding was due to confusing overflow and precision. makes sense. Nonetheless: 

```suggestion
// This is guaranteed not to overflow on whatever values nor lose precision.
```

srml/staking/reward-curve/src/log.rs
thiolliere: for now it is not part of runtime just compile-time.
thus I didn't completely checked that loop cannot loop forever in any situation and there is assert inside.
What I check is that if it produce something is it accurate.

srml/staking/reward-curve/src/lib.rs
thiolliere: done

core/sr-primitives/src/curve.rs
thiolliere: no what I meant is that there is an inherent imprecision from converting from one fraction to one another in some cases: let's say a = 100billion, p=1, q=3 converting 1/3 into a perbillion is inaccurate and dependings on how much a is bigger than a billion the inaccuracy remains:

```rust
	let a = 100 * 1_000_000_000u64;
	let res = multiply_by_rational_saturating(a, 1, 3);
	assert_eq!(res, 33333333333);
	let res = super::Perbill::from_rational_approximation(1u32, 3u32) * a;
	assert_eq!(res, 33333333300);
```

Issue #2828, opened 2019-06-07
srml-contract: Finally rename create to instantiate


pepyakin:
At some point we decided to go for `instantiate` instead of `create`. We even changed the name in some places, however, in some other places the old name is still used.

expenses:
I'm fairly sure this can be closed now.

bkchr:
Ty

PR #3645, opened 2019-09-19
srml-contract: Rename Contract::create to instantiate

expenses:
See #2828. I'm not sure how much of the documentation needs updating.


xlc:
you have to bump spec_version as this is a spec breaking change

expenses:
@xlc Done 🙂

PR #3646, opened 2019-09-19
Send local GRANDPA authority id to telemetry

cmichi:
The authority id was previously sent to telemetry, but removed as part of the [key management refactoring](https://github.com/paritytech/substrate/commit/ed61b1fd98010b85d508bdd17867bf6bfb5cfee7#diff-16b4c34d50bdfcbf732c09c7be1fdcf9L468). It is needed for the consensus visualization.

The corresponding telemetry PR is https://github.com/paritytech/substrate-telemetry/pull/179.

### DIFF
--- a/core/finality-grandpa/src/lib.rs
+++ b/core/finality-grandpa/src/lib.rs
@@ -531,7 +531,7 @@ pub fn run_grandpa_voter<B, E, Block: BlockT<Hash=H256>, N, RA, SC, X>(
 				let maybe_authority_id = authority_id(&authorities.current_authorities(), &conf.keystore)
 					.unwrap_or(Default::default());
 				telemetry!(CONSENSUS_INFO; "afg.authority_set";
-					"authority_id" => format!("{}", maybe_authority_id),
+					"authority_id" => maybe_authority_id.to_string(),
 					"authority_set_id" => ?authorities.set_id(),
 					"authorities" => {
 						let curr = authorities.current_authorities();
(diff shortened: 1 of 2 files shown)


core/finality-grandpa/src/lib.rs
```diff
@@ -523,20 +523,24 @@ pub fn run_grandpa_voter<B, E, Block: BlockT<Hash=H256>, N, RA, SC, X>(
 
 	register_finality_tracker_inherent_data_provider(client.clone(), &inherent_data_providers)?;
 
+	let conf = config.clone();
 	let telemetry_task = if let Some(telemetry_on_connect) = telemetry_on_connect {
 		let authorities = persistent_data.authority_set.clone();
 		let events = telemetry_on_connect
 			.for_each(move |_| {
+				let maybe_authority_id = authority_id(&authorities.current_authorities(), &conf.keystore)
```
bkchr: Hmm, when we rotate the session and the key changes, this will not notify the telemetry without a disconnect.

core/finality-grandpa/src/lib.rs
```diff
@@ -523,20 +523,24 @@ pub fn run_grandpa_voter<B, E, Block: BlockT<Hash=H256>, N, RA, SC, X>(
...
+					.unwrap_or(Default::default());
 				telemetry!(CONSENSUS_INFO; "afg.authority_set";
-					 "authority_set_id" => ?authorities.set_id(),
-					 "authorities" => {
+					"authority_id" => format!("{}", maybe_authority_id),
```
bkchr:
```suggestion
					"authority_id" => maybe_authority_id.to_string(),
```

core/finality-grandpa/src/lib.rs
cmichi: Updated the PR to address this (after talking to @andresilva about a fitting place in the code).

cmichi:
Comment is addressed and PR is updated.

PR #3647, opened 2019-09-19
fn import_block -> fn block_import_params in SimpleSlotWorker

kigawas:
#3158 

Rename confusing `import_block` function to `block_import_params`


andresilva:
Thanks for the PR!

PR #3649, opened 2019-09-19
Finish off srml-contract create -> instantiate renamings

expenses:
Continuation of #3645. @pepyakin.


pepyakin:
Since this changes the runtime logic we need to bump the runtime version here
https://github.com/paritytech/substrate/blob/e8d8b0bcb8f497ab9c15e91cde56ed94cade815a/node/runtime/src/lib.rs#L81-L86

expenses:
@pepyakin ah, I'm still getting the hang of when to do that haha

expenses:
@pepyakin Where would those be?

pepyakin:
https://github.com/paritytech/substrate/blob/8d30872dec7b17e120c810024d7a71edc810ae4c/srml/contracts/src/account_db.rs#L281
(And appropriate docs, including COMPLEXITY.md)

https://github.com/paritytech/substrate/blob/8d30872dec7b17e120c810024d7a71edc810ae4c/srml/contracts/src/exec.rs#L545

https://github.com/paritytech/substrate/blob/8d30872dec7b17e120c810024d7a71edc810ae4c/srml/contracts/src/wasm/mod.rs#L176

And docs like this
https://github.com/paritytech/substrate/blob/8d30872dec7b17e120c810024d7a71edc810ae4c/srml/contracts/src/wasm/runtime.rs#L431
mentioning contract creation where it is actually instantiation

pepyakin:
Well, for everything that used to refer "to create a contract". Here is a counterexample though:

https://github.com/paritytech/substrate/blob/8d30872dec7b17e120c810024d7a71edc810ae4c/srml/contracts/src/exec.rs#L610

we still use "create" to say "to create an account"

PR #3650, opened 2019-09-19
Prepare for asynchronous transaction validation in tx pool

svyatonik:
This is required to restore txpool && txpool-related RPCs on light client after #3538 (light-related changes will be added in follow-up PRs). The main change in this PR is that `ChainApi::validate_transaction` now returns `Future<>` instead of `Result<>`. This affects `Pool::{submit_at, submit_one, submit_and_watch, prune, prune_tags}` methods. All other changes are simply fixes to support that change.

Planned light-related changes: (1) separate `ChainApi` implementation that will use `Fetcher` to validate transaction on remote node (that has been removed in #3538) and (2) separate `maintain_transaction_pool()` implementation that will (presumably) fetch proof-of-tx-inclusion from remote node and remove transactions that are included in the block.


core/network/src/protocol.rs
```diff
@@ -953,6 +953,14 @@ impl<B: BlockT, S: NetworkSpecialization<B>, H: ExHashT> Protocol<B, S, H> {
 		who: PeerId,
 		extrinsics: message::Transactions<B::Extrinsic>
 	) {
+		// sending extrinsic to light node is considered a bad behavior
+		if !self.config.roles.is_full() {
```
svyatonik: Ahhh, forgot about this change - could create a separate PR for it, if required. Extrinsics from other nodes are ignored by light tx pool anyway => this change only decreases traffic on light client.

core/transaction-pool/graph/src/pool.rs
```diff
@@ -115,109 +133,39 @@ pub struct Pool<B: ChainApi> {
...
 
 	/// Imports one unverified extrinsic to the pool
-	pub fn submit_one(&self, at: &BlockId<B::Block>, xt: ExtrinsicFor<B>) -> Result<ExHash<B>, B::Error> {
-		Ok(self.submit_at(at, ::std::iter::once(xt), false)?.pop().expect("One extrinsic passed; one result returned; qed")?)
+	pub fn submit_one(
+		&self,
+		at: &BlockId<B::Block>,
+		xt: ExtrinsicFor<B>,
+	) -> impl Future<Output=Result<ExHash<B>, B::Error>> {
+		match self.submit_at(at, ::std::iter::once(xt), false) {
```
tomusdrw:
```suggestion
		match self.submit_at(at, std::iter::once(xt), false) {
```

core/transaction-pool/graph/src/pool.rs
```diff
@@ -115,109 +133,39 @@ pub struct Pool<B: ChainApi> {
 impl<B: ChainApi> Pool<B> {
 	/// Imports a bunch of unverified extrinsics to the pool
 	pub fn submit_at<T>(&self, at: &BlockId<B::Block>, xts: T, force: bool)
-		-> Result<Vec<Result<ExHash<B>, B::Error>>, B::Error>
+		-> Result<impl Future<Output=Vec<Result<ExHash<B>, B::Error>>>, B::Error>
```
tomusdrw: Having `Result<Future, Error>` seems a bit weird, if we ever extend the range of async APIs further it will complicate the returned errors. Maybe let's just return a `Future<Output=Result<..>>`?

core/transaction-pool/graph/src/pool.rs
```diff
@@ -226,41 +174,51 @@ impl<B: ChainApi> Pool<B> {
...
 	/// into runtime too often we first lookup all extrinsics that are in the pool and get
 	/// their provided tags from there. Otherwise we query the runtime at the `parent` block.
-	pub fn prune(&self, at: &BlockId<B::Block>, parent: &BlockId<B::Block>, extrinsics: &[ExtrinsicFor<B>]) -> Result<(), B::Error> {
-		let mut tags = Vec::with_capacity(extrinsics.len());
+	pub fn prune(
+		&self,
+		at: &BlockId<B::Block>,
+		parent: &BlockId<B::Block>,
+		extrinsics: &[ExtrinsicFor<B>],
+	) -> impl Future<Output=Result<(), B::Error>> {
```
tomusdrw: Couldn't we relax the lifetime in the returned future and avoid cloning `self.data` (and keeping it in an `Arc`)?

core/transaction-pool/graph/src/pool.rs
```diff
@@ -286,13 +244,204 @@ impl<B: ChainApi> Pool<B> {
...
+		let validation_future = join_all(validation_futures);
+
+		// and then import all successfully validated extrinsics at once + enforce limits
+		let self_clone = self.clone();
+		let import_future = validation_future
+			.then(move |validation_results| ready({
+				let import_results = validation_results
+					.into_iter()
+					.map(|validation_result| {
+						let imported = self_clone.0.pool.write().import(validation_result?)?;
```
tomusdrw: This is locking inside the event loop :/ I think it might have very bad performance consequences.

svyatonik:
The separate-thread approach won't work for light-in-browser, I believe :/ By decoupling you mean passing pre-validated transactions to `submit_at()` / `submit_one()`? I've tried this initially, but when noticed that revalidation is required in `prune()`, switched to async-validate-approach. Do you think it still worth move out validation interface from `pool.rs`?

Also - could you please elaborate on "avoid locking inside the event pool" - not sure I got it :/ Did you mean locks like `Mutex::lock`, or blocking on futures in event loop threads? Afaik there's no explicit thread spawns (other than in offchain) now in substrate - we are always executing code from within `tokio_executor::DefaultExecutor` threads. So whenever we call something like `Mutex::lock`, it is locking within tokio threads.

tomusdrw:
> So whenever we call something like Mutex::lock, it is locking within tokio threads.

Yeah, that's exactly what I mean. I don't think it's healthy to do that.

> I've tried this initially, but when noticed that revalidation is required in prune(), switched to async-validate-approach. Do you think it still worth move out validation interface from pool.rs?

I still think it might be a bit cleaner to do so. `prune` could get passed a function `Fn(&[Hash]) -> impl Future<Output=Vec<Validity>>` to verify all needed transactions.

svyatonik:
OK, thanks :) I still not sure - what's wrong with locking in this PR - there are exactly same locks as before, locking exactly same (tokio) threads (at different moments, though) :) And after fix, they still be there. But decoupling, probably, makes sense.

core/service/src/builder.rs
```diff
@@ -914,39 +920,54 @@ where
 
 pub(crate) fn maintain_transaction_pool<Api, Backend, Block, Executor, PoolApi>(
 	id: &BlockId<Block>,
-	client: &Client<Backend, Executor, Block, Api>,
+	client: &Arc<Client<Backend, Executor, Block, Api>>,
 	transaction_pool: &TransactionPool<PoolApi>,
 	retracted: &[Block::Hash],
-) -> error::Result<()> where
+) -> error::Result<Box<dyn Future<Item = (), Error = ()> + Send>> where
```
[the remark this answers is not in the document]
svyatonik: We send it then through `mpsc::UnboundedSender<Box<dyn Future<Item = (), Error = ()> + Send>>` [here](https://github.com/paritytech/substrate/blob/13fc71c681cc9a3cc911c32c7890b52885092969/core/service/src/lib.rs#L87). So we may (at least I could try) leave it as `impl Future` here, but it needs to be wrapped in `Box` anyway

core/service/src/builder.rs
```diff
@@ -914,39 +920,54 @@ where
...
 {
 	// Put transactions from retracted blocks back into the pool.
-	for r in retracted {
-		if let Some(block) = client.block(&BlockId::hash(*r))? {
-			let extrinsics = block.block.extrinsics();
-			if let Err(e) = transaction_pool.submit_at(id, extrinsics.iter().cloned(), true) {
-				warn!("Error re-submitting transactions: {:?}", e);
-			}
+	let client_copy = client.clone();
+	let retracted_transactions = retracted.to_vec().into_iter()
```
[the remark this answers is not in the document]
svyatonik: This way I'm eliminating non-static references - returned future needs to be 'static.

svyatonik:
So the `Pool` is now `ValidatedPool` and `Pool`. Crate-internal `ValidatedPool` only deals with pre-verified transactions and has synchronous API. `Pool` has future-based API and internally uses `ValidatedPool` to store transactions after they're validated.

I've also found that in my first commit I've missed one `block_on()` call in production code - there's a [separate commit](https://github.com/paritytech/substrate/pull/3650/commits/0d4f8421f8cc59d617bf854091bc417ad4617f7b) that fixes it.

core/rpc/src/author/mod.rs
```diff
@@ -61,12 +67,14 @@ pub struct Author<B, E, P, RA> where P: PoolChainApi + Sync + Send + 'static {
 impl<B, E, P, RA> Author<B, E, P, RA> where P: PoolChainApi + Sync + Send + 'static {
 	/// Create new instance of Authoring API.
 	pub fn new(
+		executor: Arc<dyn Executor<Box<dyn Future<Item = (), Error = ()> + Send>> + Send + Sync>,
```
tomusdrw: Should we typedef this? It's used in `api/src/subscriptions.rs` as well and looks pretty nasty.

core/rpc/src/author/mod.rs
```diff
@@ -108,15 +116,20 @@ impl<B, E, P, RA> AuthorApi<ExHash<P>, BlockHash<P>> for Author<B, E, P, RA> whe
...
+	fn submit_extrinsic(&self, ext: Bytes) -> FutureResult<ExHash<P>> {
+		let xt = match Decode::decode(&mut &ext[..]) {
+			Ok(xt) => xt,
+			Err(err) => return Box::new(result(Err(err.into()))),
+		};
 		let best_block_hash = self.client.info().chain.best_hash;
-		self.pool
+		Box::new(self.pool
 			.submit_one(&generic::BlockId::hash(best_block_hash), xt)
+			.boxed()
```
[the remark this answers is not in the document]
tomusdrw: Bump

core/rpc/src/author/mod.rs
```diff
@@ -151,30 +164,40 @@ impl<B, E, P, RA> AuthorApi<ExHash<P>, BlockHash<P>> for Author<B, E, P, RA> whe
...
-			sink
-				.sink_map_err(|e| warn!("Error sending notifications: {:?}", e))
-				.send_all(Compat::new(watcher.into_stream().map(|v| Ok::<_, ()>(Ok(v)))))
-				.map(|_| ())
-		})
+		let subscriptions = self.subscriptions.clone();
+		let subscription_future = future_watcher
+			.map_err(|err| { warn!("Failed to submit extrinsic: {}", err); })
+			.map(move |watcher| subscriptions.add(subscriber, move |sink| {
+				sink
```
tomusdrw: Can't the `future_watcher` be moved here and combined with `sink` future? That would allow us to get rid of `executor` here.

core/service/src/builder.rs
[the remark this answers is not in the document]
tomusdrw: I see. I think defering `Box`ing is good in general, but I'm not strong on this here.

core/service/src/lib.rs
```diff
@@ -914,38 +923,40 @@ where
...
-						Err(e) => {
-							debug!("Error converting pool error: {:?}", e);
-							None
-						},
-					}
+				let import_future = self.pool.submit_one(&best_block_id, uxt);
+				let import_future = import_future
+					.then(move |import_result| {
+						match import_result {
+							Ok(_) => report_handle.report_peer(who, reputation_change),
```
tomusdrw: So this is now going to be async, I hope we do handle non-existing peers correctly.

core/transaction-pool/graph/src/pool.rs
```diff
@@ -286,200 +210,236 @@ impl<B: ChainApi> Pool<B> {
...
-		base::Imported::Future { ref hash } => {
-			listener.future(hash)
-		},
+impl<B: ChainApi> Clone for SharedValidatedPool<B> {
+	fn clone(&self) -> Self {
+		Self(self.0.clone())
+	}
+}
+
+impl<B: ChainApi> SharedValidatedPool<B> {
```
tomusdrw: Why do we implement the methods for `SharedValidatedPool` and not `Pool`?

core/rpc/src/author/mod.rs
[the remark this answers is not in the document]
svyatonik: Sorry - not sure why I have marked as resolved :/ Fixed now

core/transaction-pool/graph/src/pool.rs
svyatonik: Just double-checking all your comments. This has been fixed by switching from `Result<Future<Vec<Result>>>` to `Future<Result<Vec<Result>>>`

core/transaction-pool/graph/src/pool.rs
svyatonik: It needed to be `Box<>` .with `'static` lifetime to be spawned in executor' threads.

core/rpc/src/author/mod.rs
svyatonik: I've added a typedef, but then got rid of `executor` field in `Author` as you've suggested. Typedef is still there - if you thing it isn't required now, I can revert

core/service/src/lib.rs
svyatonik: Yes - reporting is done using `mpsc` channel anyway => it has to be safe.

core/rpc/src/author/mod.rs
svyatonik: Right, thanks! `Stream::flatten` has helped here.

core/transaction-pool/graph/src/pool.rs
svyatonik: That's an artefact of previous implementation - merged methods of both in single `Pool` struct.

PR #3653, opened 2019-09-19
Improve tests for BABE

rphmeier:


core/consensus/babe/primitives/src/lib.rs
```diff
@@ -105,6 +105,12 @@ impl Epoch {
 			randomness: descriptor.randomness,
 		}
 	}
+
+	/// Produce the "end slot" of the epoch. This is NOT inclusive to the epoch,
+	// i.e. the slots covered by the epoch are `self.start_slot .. self.end_slot()`.
+	pub fn end_slot(&self) -> SlotNumber {
+		self.start_slot + self.duration
```
DemiMarie-parity: Is there a proof that this cannot overflow?

Issue #1440, opened 2019-01-15
Split QueueInfo from ClientInfo and amalgate into single struct


gnunicorn:
Should be possible to amalgate ChainInfo into ClientInfo:

https://github.com/paritytech/substrate/blob/a2dfec4d3c467a481da70efb34f105e141a289fa/core/client/src/client.rs#L107-L114

tomusdrw:
What does it help with exactly? I actually like the fact that we have a nice separation between `Client` and `Blockchain` state. And the `info` call is just proxied (I'd say it would be even better to just use blockchain directly).

gnunicorn:
this was opened when removing inline todos and fixmes. has been fixed meanwhile. closing.

PR #3654, opened 2019-09-20
Copy blockchain Info fields into ClientInfo

expenses:
Related to #1440. This cleans a lot of code up by removing the redundant `.chain` accesses. Currently I'm just coping the fields off the `Info` struct. It might be cleaner for `HeaderBackend::info` to return a `ClientInfo` and just set `used_state_cache_size` to `None` until it can be set by the backend.

### DIFF
--- a/core/cli/src/informant.rs
+++ b/core/cli/src/informant.rs
@@ -40,7 +40,7 @@ pub fn build(service: &impl AbstractService) -> impl Future<Item = (), Error = (
 	let client = service.client();
 	let mut last_best = {
 		let info = client.info();
-		Some((info.chain.best_number, info.chain.best_hash))
+		Some((info.best_number, info.best_hash))
 	};
 
 	let display_block_import = client.import_notification_stream().map(|v| Ok::<_, ()>(v)).compat().for_each(move |n| {
--- a/core/cli/src/informant/display.rs
+++ b/core/cli/src/informant/display.rs
@@ -54,8 +54,8 @@ impl<B: BlockT> InformantDisplay<B> {
 
 	/// Displays the informant by calling `info!`.
 	pub fn display(&mut self, info: &ClientInfo<B>, net_status: NetworkStatus<B>) {
-		let best_number = info.chain.best_number;
-		let best_hash = info.chain.best_hash;
+		let best_number = info.best_number;
+		let best_hash = info.best_hash;
 		let speed = speed::<B>(best_number, self.last_number, self.last_update);
 		self.last_update = time::Instant::now();
 		self.last_number = Some(best_number);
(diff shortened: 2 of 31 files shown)


bkchr:
@gnunicorn as you closed the issue, we can close this pr as well?

gnunicorn:
@bkchr unclear. I understood the initial issue to be about removing two distinct info-items (from client and backend) and have just one. That has already happened.

But this PR doesn't intend to fix this anymore but instead clean up the substruct, which arguably has some merit as it removes an unnecessary `.chain` call and, if we did split upon the only differentiating factor, as @expenses suggested, might have some further clean up down the line, too. 

For that, however, I would go the opposite approach - in my on-going quest to remove direct dependency on `client` in the code-base: the specific `info`-field is only used for metrics at the moment, so let's see if we can remove that struct entirely by just have a single function call for that field and let the outer caller otherwise directly connect to `backend` and call `info` there. What do you think?

bkchr:
I don't have any strong opinion on it :) Maybe we should hide the implementation behind a trait, so we do not replace the dependency on `Client` with `Backend` everywhere :D

expenses:
@gnunicorn so you're suggesting just having functions like like `backend.best_hash()` etc?

gnunicorn:
`Backend` is a trait. But my mistake the info come from the `trait HeaderBackend`. `Backend` itself only exposes the other metric and both are - currently - put into the outer struct. Arguably not a really useful thing that warrants to have a dependency on client directly rather than expecting the traits to be given.

gnunicorn:
> @gnunicorn so you're suggesting just having functions like backend.best_hash() etc?

No, I am suggesting we remove the `client.info` function and attempt to see if we can't get direct access to the `traits` actually implementing what we want in these locations instead.

expenses:
I think that you could add on an `info` function to the `AbstractService` trait so that https://github.com/paritytech/substrate/blob/4117bb9ff5536ae632056e77219789151e0f6aed/core/cli/src/informant.rs#L40-L44

becomes:

```rust
let mut last_best = {
    let info = service.info();
    Some((info.best_number, info.best_hash))
}
```

gnunicorn:
Hmm...just moving the info function from `client` to `AbstractService` is very unsatisfying (and doesn't solve the issue as `AbstractService` would still have to go through `.client().info()` to get that info). The problem is that the informant currently binds to that rather than having direct access to `blockchain` to call `info` directly - which has all info it needs.

This would require quite a bit of a refactor and I wouldn't put that under this label anymore. Maybe we should open a new ticket about this?

expenses:
Yep, that sounds good.

PR #3659, opened 2019-09-20
Remove unused key types sr25 and ed25

JoshOrndorff:
The crypto `KeyTypeId`s are given four-character identifiers specific to their application such as `babe`, `gran`, `acco`, etc. This PR removes unused general purpose types `sr25` and `ed25`.

### DIFF
--- a/core/application-crypto/src/ed25519.rs
+++ b/core/application-crypto/src/ed25519.rs
@@ -59,7 +59,7 @@ mod tests {
 						ED25519,
 				},
 				crypto::Pair,
-				traits::BareCryptoStore as _
+				traits::BareCryptoStore as _,
 		};
 	use test_client::{
 		TestClientBuilder, DefaultTestClientBuilderExt, TestClientBuilderExt,
--- a/core/application-crypto/src/ed25519.rs
+++ b/core/application-crypto/src/ed25519.rs
@@ -21,8 +21,8 @@ use crate::{RuntimePublic, KeyTypeId};
 pub use primitives::ed25519::*;
 
 mod app {
-	use crate::key_types::ED25519;
-	crate::app_crypto!(super, ED25519);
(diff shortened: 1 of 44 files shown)


bkchr:
DOes not compile and as you see, it is used in some test code. So, we should move the `KeyType` and the app crypto declaration into testing.

bkchr:
Could you please move the both key types to `primitives/testing.rs` and the `app_crypto!` into `sr-primitives/testing.rs`?

core/application-crypto/src/ed25519.rs
```diff
@@ -53,7 +53,14 @@ impl RuntimePublic for Public {
...
 mod tests {
 	use sr_primitives::{generic::BlockId, traits::ProvideRuntimeApi};
-	use primitives::{testing::KeyStore, crypto::Pair, traits::BareCryptoStore as _};
+	use primitives::{
+				testing::{
+						KeyStore,
+						ED25519,
+				},
+				crypto::Pair,
+				traits::BareCryptoStore as _
```
gavofyork:
```suggestion
				traits::BareCryptoStore as _,
```

core/application-crypto/src/sr25519.rs
```diff
@@ -53,7 +53,14 @@ impl RuntimePublic for Public {
...
 mod tests {
 	use sr_primitives::{generic::BlockId, traits::ProvideRuntimeApi};
-	use primitives::{testing::KeyStore, crypto::Pair, traits::BareCryptoStore as _};
+	use primitives::{
+				testing::{
+						KeyStore,
+						SR25519,
+				},
+				crypto::Pair,
+				traits::BareCryptoStore as _
```
gavofyork:
```suggestion
				traits::BareCryptoStore as _,
```

Issue #2688, opened 2019-05-25
Proper error types


DemiMarie-parity:
The consensus engines (at least AuRa and BABE; haven’t checked GRANDPA) currently use `String` as the error type, which is not awesome.  This should be fixed.

I will probably implement this after #2689 and session key verification.  If someone wants to do this, I will be happy to mentor them.

expenses:
I'm looking into this. I think ideally you'd have each `ProvideInherentData` have an associated `Error` type that they parse the encoded error into. The problem is that `InherentDataProviders` holds a `Vec<Box<ProvideInherentData<Error=Something>>>` so you end up using types anyway. You could use a dyn trait like `Error=dyn std::error::Error` possibly but that seems messy. What do you think @DemiMarie-parity?

DemiMarie-parity:
@expenses I think that a dyn trait is a good idea here.  Errors are slow-path code, so the overhead in runtime performance should be more than compensated by the decrease in code size, to the extent that either matters.

burdges:
I suppose substrate might actually be complex enough to benefit form the [failure](https://docs.rs/crate/failure) crate and libbacktrace, although maybe the wasm boundary makes this problematic. And simpler solutions likely suffice too.

rphmeier:
Failure looks promising, but after error-chain we decided to stay away from all of the ecosystem error-handling stuff until it's very clearly stable

bkchr:
> @expenses I think that a dyn trait is a good idea here. Errors are slow-path code, so the overhead in runtime performance should be more than compensated by the decrease in code size, to the extent that either matters.

Replacing a String with a dynamic traits makes no sense at all. If you want proper errors, define an enum and this only make sense for the consensus crates. The inherent data providers are that generic, that you can not forsee all the combinations of errors that could occur.

expenses:
The problem here is really how you fit different error types into the `Vec<Box<dyn ProvideInherentData>>` of `InherentDataProviders`. You could have something like
```rust
enum InherentErrors {
    AuraError1,
    AuraError2,
    BABEError1,
    ...
}
```
but this of course would create problems for custom modules.  I wish that you have traits with concrete assocated types that you didn't have to specify when the trait is `dyn`, which would enable something like

```rust
trait Whatever {
    type Something: Display;
}

struct A;

impl Whatever for A {
    type Something = &'static str;
}

struct B;

impl Whatever for B {
    type Something = String;
}

struct C {
[...]

PR #3662, opened 2019-09-21
Add a InherentDataProviderError trait for inherents error handling

expenses:
See #2688. Currently, `ProvideInherentData` uses `String`s for errors, which is not ideal. This commit introduces a `InherentDataProviderError` trait for errors and `InherentError` as a type alias for `Box<dyn InherentDataProviderError>`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5039,6 +5039,7 @@ dependencies = [
  "substrate-client 2.0.0",
  "substrate-consensus-babe-primitives 2.0.0",
  "substrate-consensus-common 2.0.0",
+ "substrate-inherents 2.0.0",
  "substrate-keyring 2.0.0",
  "substrate-peerset 2.0.0",
  "substrate-primitives 2.0.0",
--- a/core/consensus/aura/src/lib.rs
+++ b/core/consensus/aura/src/lib.rs
@@ -48,7 +48,7 @@ use sr_primitives::{generic::{BlockId, OpaqueDigestItemId}, Justification};
 use sr_primitives::traits::{Block as BlockT, Header, DigestItemFor, ProvideRuntimeApi, Zero, Member};
 
 use primitives::crypto::Pair;
-use inherents::{InherentDataProviders, InherentData};
+use inherents::{InherentDataProviders, InherentData, InherentError};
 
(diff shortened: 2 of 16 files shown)


bkchr:
First, thank you for the pr. However, given the generic architecture of the inherent data providers, I don't see any reason for this pr. While I think that having better errors is a good idea, for the inherent data providers it makes no sense at all. Especially replacing a String with a Box. 

The linked issue is more about the consensus engines and their error types. For them it makes sense to define an enum, as you can forsee all possible error types. 

Given that, I'm against this pr.

expenses:
Example of a potential error enum after this change:
```rust
#[derive(derive_more::Display, Debug)]
enum AuraError<B: BlockT> {
    #[display(fmt = "Could not fetch authorities at {:?}: {:?}", _0, _1)]
    FetchAuthorities(B::Hash, consensus_common::Error),
    #[display(fmt = "Could not extract timestamp and slot: {:?}", _0)]
    ExtractTimestamp(consensus_common::Error),
    #[display(fmt = "{:?}", _0)]
    Runtime(RuntimeString),
    #[display(fmt = "{:?}", _0)]
    Client(ClientError)
}

impl<B: BlockT> InherentDataProviderError for AuraError<B> {}

impl<B: BlockT> From<AuraError<B>> for InherentError {
    fn from(error: AuraError<B>) -> Self {
        Box::new(error) as Self
    }
}
```

bkchr:
`Aura` could also see any other inherent error, if there occurs any error in the runtime while creating the inherent extrinsics. It does not only sees its own errors. So, the potential enum should look like:

```
#[derive(derive_more::Display, Debug)]
enum AuraError<B: BlockT> {
    #[display(fmt = "Could not fetch authorities at {:?}: {:?}", _0, _1)]
    FetchAuthorities(B::Hash, consensus_common::Error),
    #[display(fmt = "Could not create inherent extrinsics: {}", _0)]
    InherentExtrinsics(String),
    #[display(fmt = "{:?}", _0)]
    Runtime(RuntimeString),
    #[display(fmt = "{:?}", _0)]
    Client(ClientError)
}
```

expenses:
See #2688.

PR #1410, opened 2019-01-14
Sync block justifications

andresilva:
This PR extends the sync protocol to handle requesting and importing block justifications. `BlockImport` can now return `NeedsJustification` which will trigger the sync module to try and fetch a justification for the given block from peers, which is then passed to `BlockImport` as well for validation and finalizing the given block.

I will implement new justification announcements in a follow up PR.

PR #3663, opened 2019-09-21
Fix invalid debug assertion.

twittner:
Context: https://github.com/paritytech/polkadot/issues/440

The current `debug_assert_eq!` is incorrect as the code is prepared to handle zero response blocks. What should have been expressed is that we expect 0 or 1 response blocks.

The assertion was introduced as part of #3105 to reify the a comment by @andresilva here: https://github.com/paritytech/substrate/pull/1410/files#diff-0c805d3ef65d16a3738f8270c6726d38R484. Please let me know if this was a bad idea.

### DIFF
--- a/core/network/src/protocol/sync.rs
+++ b/core/network/src/protocol/sync.rs
@@ -659,7 +659,7 @@ impl<B: BlockT> ChainSync<B> {
 			peer.state = PeerSyncState::Available;
 
 			// We only request one justification at a time
-			debug_assert_eq!(1, response.blocks.len());
+			debug_assert!(response.blocks.len() < 2);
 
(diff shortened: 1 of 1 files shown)


arkpar:
There should be no such assertions in the first place. We can't control what remote side sends us. The code should handle all kinds of invalid responses by logging and reducing peer reputation, not by panicking in debug mode.

arkpar:
Misbehaviour would be replying with something other than requested or sending useless data in general. As for the actual value, I guess it should be similar to other offences of the same kind, which are set to `-(1 << 20)`

twittner:
I will open another PR that just removes the `debug_assert!` for now.

PR #3664, opened 2019-09-22
Rename more things related to contract instantiation

expenses:
Closes #3645

See #3645. I've tried to have 'instantiation' refer specifically to contracts while 'creation' refer to accounts. This gets a little tricky when you're talking about smart contract accounts, so I've kept lines like:

```
srml/contracts/src/lib.rs
392:	/// The fee required to create an account.
393:	type CreationFee: Get<BalanceOf<Self>>;
423:/// and the account id that requested the account creation.
491:		/// The fee required to create an account.
492:		const CreationFee: BalanceOf<T> = T::CreationFee::get();
568:		/// a regular account will be created and any value will be transferred.
589:		/// - The smart-contract account is created at the computed address.
590:		/// - The `ctor_code` is executed in the context of the newly-created account. Buffer returned
681:				// We don't have a spare buffer here in the first place, so create a new empty one.
803:		T::Currency::deposit_creating(&dest, origin_free_balance);
874:	pub account_create_fee: BalanceOf<T>,
886:			account_create_fee: T::CreationFee::get(),
```

Please don't merge until this issue is closed, thanks :^)

### DIFF
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -409,7 +409,7 @@ impl contracts::Trait for Runtime {
 	type TransactionByteFee = ContractTransactionByteFee;
 	type ContractFee = ContractFee;
 	type CallBaseFee = contracts::DefaultCallBaseFee;
-	type CreateBaseFee = contracts::DefaultCreateBaseFee;
+	type InstantiateBaseFee = contracts::DefaultInstantiateBaseFee;
 	type MaxDepth = contracts::DefaultMaxDepth;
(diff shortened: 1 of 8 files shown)


Robbepop:
I am not sure if this has already happened in this PR but it might also make sense to rename the expected Wasm contract exported function names from `create` to `instantiate`.

expenses:
Yep, that happened in #3649.

pepyakin:
Thanks @expenses !

PR #3665, opened 2019-09-22
Optimize tree route to sync faster

### LINKED ISSUE #3592: Optimize tree_route and related, avoid read_header
When syncing it seems that we waste a lot of time in `read_header`, function that reads a header from db, when actually we only need the hash of the parent header. Functions that should be optimized:

- `is_descendent_of`, this function is quite un-optimized but heavily used in `import_block`.
- `ancestry`.
- `tree_route`

Using a hash map from `(hash -> parent_hash)` I could apparently reduce syncing Kusama from 2:20 hours to 20 minutes on my laptop.

marcio-diaz:
Fixes #3592

### Summary

This PR removes some inefficiencies of `tree_route` and calling functions. Previously, `tree_route` was computing the path between two blocks by loading headers from database. Now, we cache the information needed in an LRU cache in-memory. 
Also, many times we were using `tree_route` just to check if a block is descendant of another one. Instead, this PR introduces a function `lowest_common_ancestor` that compute descendants in `O(1)` (most of the time).

As a side effect, this improves syncing speed for Kusama from ~2 hours to ~20 minutes (25 bps -> 35 bps in wasm, 80 bps -> 400 bps in native).

### Changes:
- Introduce a `CachedHeaderMetadata` with fields hash, number, parent and ancestor. The `ancestor` field is used to quickly jump through the tree. 
- We cache the last `20_000 HeaderMetadata` used, by means of a `LRU cache`. The optimal number depends on the difference between best and finalized block numbers.
- Use `CachedHeaderMetadata` to traverse the tree in the `tree_route` function. Previously we were loading the whole headers from DB to do this task.
- Implement a `lowest_common_ancestor` function and use it in `is_descendent_of` and other places, replacing `tree_route`. Right now, the query pattern of this function is as follows: `lca(best, final), lca(best+1, final), lca(best+2, final), ..., lca(best+5000, final)` . By using the `ancestor` field of `LightHeader`, the first query is `O(n)` and the followings `O(1)`.

### Notes:
- I also tried implementing `lowest_common_ancestor` by dividing the tree in sections and jumping between them, getting `O(sqrt(h))` for query, but it didn't work very well.

### DIFF
--- a/core/cli/src/informant.rs
+++ b/core/cli/src/informant.rs
@@ -17,10 +17,11 @@
 //! Console informant. Prints sync progress and block events. Runs on the calling thread.
 
 use client::BlockchainEvents;
+use header_metadata::TreeBackend;
 use futures::{Future, Stream};
 use futures03::{StreamExt as _, TryStreamExt as _};
 use log::{info, warn};
-use sr_primitives::{generic::BlockId, traits::Header};
+use sr_primitives::traits::Header;
 use service::AbstractService;
 
 mod display;
@@ -47,20 +48,17 @@ pub fn build(service: &impl AbstractService) -> impl Future<Item = (), Error = (
 		// detect and log reorganizations.
 		if let Some((ref last_num, ref last_hash)) = last_best {
 			if n.header.parent_hash() != last_hash && n.is_new_best  {
-				let maybe_ancestor = ::client::blockchain::lowest_common_ancestor(
-					|id| client.get_light_header(&id)?.ok_or_else(
-						|| client::error::Error::UnknownBlock(format!("{:?}", id))),
-					|data| client.set_light_header(data),
-					BlockId::Hash(last_hash.clone()),
-					BlockId::Hash(n.hash),
+				let maybe_ancestor = client.lowest_common_ancestor(
+					last_hash.clone(),
+					n.hash,
 				);
 
 				match maybe_ancestor {
-					Ok(ancestor) if ancestor.0 != *last_hash => info!(
+					Ok(ref ancestor) if ancestor.hash != *last_hash => info!(
 						"Reorg from #{},{} to #{},{}, common ancestor #{},{}",
 						last_num, last_hash,
 						n.header.number(), n.hash,
-						ancestor.1, ancestor.0,
+						ancestor.number, ancestor.hash,
 					),
 					Ok(_) => {},
 					Err(e) => warn!("Error computing tree route: {}", e),
--- a/core/cli/src/informant.rs
+++ b/core/cli/src/informant.rs
@@ -17,7 +17,7 @@
 //! Console informant. Prints sync progress and block events. Runs on the calling thread.
 
 use client::BlockchainEvents;
-use header_metadata::TreeBackend;
+use header_metadata::lowest_common_ancestor;
 use futures::{Future, Stream};
 use futures03::{StreamExt as _, TryStreamExt as _};
 use log::{info, warn};
@@ -48,7 +48,8 @@ pub fn build(service: &impl AbstractService) -> impl Future<Item = (), Error = (
 		// detect and log reorganizations.
 		if let Some((ref last_num, ref last_hash)) = last_best {
 			if n.header.parent_hash() != last_hash && n.is_new_best  {
-				let maybe_ancestor = client.lowest_common_ancestor(
+				let maybe_ancestor = lowest_common_ancestor(
+					&*client,
 					last_hash.clone(),
 					n.hash,
 				);
--- a/core/cli/src/informant.rs
+++ b/core/cli/src/informant.rs
@@ -17,7 +17,6 @@
 //! Console informant. Prints sync progress and block events. Runs on the calling thread.
 
 use client::BlockchainEvents;
-use header_metadata::lowest_common_ancestor;
 use futures::{Future, Stream};
 use futures03::{StreamExt as _, TryStreamExt as _};
 use log::{info, warn};
@@ -48,7 +47,7 @@ pub fn build(service: &impl AbstractService) -> impl Future<Item = (), Error = (
 		// detect and log reorganizations.
 		if let Some((ref last_num, ref last_hash)) = last_best {
 			if n.header.parent_hash() != last_hash && n.is_new_best  {
-				let maybe_ancestor = lowest_common_ancestor(
+				let maybe_ancestor = header_metadata::lowest_common_ancestor(
 					&*client,
 					last_hash.clone(),
 					n.hash,
--- a/core/cli/src/informant.rs
+++ b/core/cli/src/informant.rs
@@ -47,19 +47,20 @@ pub fn build(service: &impl AbstractService) -> impl Future<Item = (), Error = (
 		// detect and log reorganizations.
 		if let Some((ref last_num, ref last_hash)) = last_best {
 			if n.header.parent_hash() != last_hash && n.is_new_best  {
-				let tree_route = ::client::blockchain::tree_route(
-					|id| client.header(&id)?.ok_or_else(
+				let maybe_ancestor = ::client::blockchain::lowest_common_ancestor(
+					|id| client.get_light_header(&id)?.ok_or_else(
 						|| client::error::Error::UnknownBlock(format!("{:?}", id))),
+					|data| client.set_light_header(data),
 					BlockId::Hash(last_hash.clone()),
 					BlockId::Hash(n.hash),
 				);
 
-				match tree_route {
-					Ok(ref t) if !t.retracted().is_empty() => info!(
+				match maybe_ancestor {
+					Ok(ancestor) if ancestor.0 != *last_hash => info!(
 						"Reorg from #{},{} to #{},{}, common ancestor #{},{}",
 						last_num, last_hash,
 						n.header.number(), n.hash,
-						t.common_block().number, t.common_block().hash,
+						ancestor.1, ancestor.0,
 					),
 					Ok(_) => {},
 					Err(e) => warn!("Error computing tree route: {}", e),
--- a/core/client/db/src/lib.rs
+++ b/core/client/db/src/lib.rs
@@ -42,8 +42,10 @@ use client::backend::NewBlockState;
 use client::blockchain::{well_known_cache_keys, HeaderBackend};
 use client::{ForkBlocks, ExecutionStrategies};
 use client::backend::{StorageCollection, ChildStorageCollection};
+use client::error::Result as ClientResult;
 use codec::{Decode, Encode};
 use hash_db::{Hasher, Prefix};
+use lru::LruCache;
 use kvdb::{KeyValueDB, DBTransaction};
 use trie::{MemoryDB, PrefixedMemoryDB, prefixed_key};
 use parking_lot::{Mutex, RwLock};
@@ -61,7 +63,7 @@ use state_machine::{
 	DBValue, ChangesTrieTransaction, ChangesTrieCacheAction, ChangesTrieBuildCache,
 	backend::Backend as StateBackend,
 };
-use crate::utils::{Meta, db_err, meta_keys, read_db, block_id_to_lookup_key, read_meta};
+use crate::utils::{Meta, db_err, meta_keys, read_db, read_meta};
 use client::leaves::{LeafSet, FinalizationDisplaced};
 use client::children;
 use state_db::StateDb;
@@ -77,6 +79,10 @@ const MIN_BLOCKS_TO_KEEP_CHANGES_TRIES_FOR: u32 = 32768;
 
 /// Default value for storage cache child ratio.
 const DEFAULT_CHILD_RATIO: (usize, usize) = (1, 10);
+
+/// Size of LRU cache for light headers. It's set to the maximum expected
+/// difference between best and finalized block while syncing.
+const LIGHT_HEADER_CACHE_SIZE: usize = 20_000;
 
 /// DB-backed patricia trie state, transaction type is an overlay of changes to commit.
 pub type DbState = state_machine::TrieBackend<Arc<dyn state_machine::Storage<Blake2Hasher>>, Blake2Hasher>;
@@ -266,21 +272,46 @@ impl<'a> state_db::MetaDb for StateMetaDb<'a> {
 	}
 }
 
+struct HeaderCache<Block: BlockT> {
+	hash_to_data: LruCache<Block::Hash, LightHeader<Block>>,
+}
+
+impl<Block: BlockT> HeaderCache<Block> {
+	fn new(capacity: usize) -> Self {
+		HeaderCache {
+			hash_to_data: LruCache::new(capacity),
+		}
+	}
+
+	fn get_data(&mut self, id: BlockId<Block>) -> Option<LightHeader<Block>> {
+		match id {
+			BlockId::Hash(hash) => self.hash_to_data.get(&hash).cloned(),
+			BlockId::Number(_) => None,
+		}
+	}
+
+	fn put_data(&mut self, data: LightHeader<Block>) {
+		self.hash_to_data.put(data.hash, data);
+	}
+}
+
 /// Block database
 pub struct BlockchainDb<Block: BlockT> {
 	db: Arc<dyn KeyValueDB>,
 	meta: Arc<RwLock<Meta<NumberFor<Block>, Block::Hash>>>,
 	leaves: RwLock<LeafSet<Block::Hash, NumberFor<Block>>>,
+	header_cache: RwLock<HeaderCache<Block>>,
 }
 
 impl<Block: BlockT> BlockchainDb<Block> {
-	fn new(db: Arc<dyn KeyValueDB>) -> Result<Self, client::error::Error> {
+	fn new(db: Arc<dyn KeyValueDB>) -> ClientResult<Self> {
 		let meta = read_meta::<Block>(&*db, columns::META, columns::HEADER)?;
 		let leaves = LeafSet::read_from_db(&*db, columns::META, meta_keys::LEAF_PREFIX)?;
 		Ok(BlockchainDb {
 			db,
 			leaves: RwLock::new(leaves),
 			meta: Arc::new(RwLock::new(meta)),
+			header_cache: RwLock::new(HeaderCache::new(LIGHT_HEADER_CACHE_SIZE)),
 		})
 	}
 
@@ -310,8 +341,33 @@ impl<Block: BlockT> BlockchainDb<Block> {
 }
 
 impl<Block: BlockT> client::blockchain::HeaderBackend<Block> for BlockchainDb<Block> {
-	fn header(&self, id: BlockId<Block>) -> Result<Option<Block::Header>, client::error::Error> {
+	fn header(&self, id: BlockId<Block>) -> ClientResult<Option<Block::Header>> {
 		utils::read_header(&*self.db, columns::KEY_LOOKUP, columns::HEADER, id)
+	}
+
+	fn set_light_header(&self, data: LightHeader<Block>) {
+		self.header_cache.write().put_data(data)
+	}
+
+	fn get_light_header(&self, id: BlockId<Block>) -> ClientResult<Option<LightHeader<Block>>> {
+		let mut header_cache = self.header_cache.write();
+		if let Some(header_data) = header_cache.get_data(id) {
+			Ok(Some(header_data))
+		} else {
+			self.header(id).and_then(|maybe_header| match maybe_header {
+				Some(header) => {
+					let light_header = LightHeader {
+						hash: header.hash(),
+						number: *header.number(),
+						parent: *header.parent_hash(),
+						ancestor: *header.parent_hash(),
+					};
+					header_cache.put_data(light_header.clone());
+					Ok(Some(light_header))
+				},
+				None => Ok(None),
+			})
+		}
 	}
 
 	fn info(&self) -> client::blockchain::Info<Block> {
@@ -325,7 +381,7 @@ impl<Block: BlockT> client::blockchain::HeaderBackend<Block> for BlockchainDb<Bl
 		}
 	}
 
-	fn status(&self, id: BlockId<Block>) -> Result<client::blockchain::BlockStatus, client::error::Error> {
+	fn status(&self, id: BlockId<Block>) -> ClientResult<client::blockchain::BlockStatus> {
 		let exists = match id {
 			BlockId::Hash(_) => read_db(
 				&*self.db,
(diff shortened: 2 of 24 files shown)


core/client/src/blockchain.rs
```diff
@@ -37,6 +50,13 @@ pub trait HeaderBackend<Block: BlockT>: Send + Sync {
 	/// Get block hash by number. Returns `None` if the header is not in the chain.
 	fn hash(&self, number: NumberFor<Block>) -> Result<Option<Block::Hash>>;
 
+	/// Get light header, usually from an in-memory cache, without hitting database.
+	/// Returns `None` if the header is not in the chain.
+	fn get_light_header(&self, id: BlockId<Block>) -> Result<Option<LightHeader<Block>>>;
```
rphmeier: convention is not to call this `get`. also `light_header` is a confusing name

marcio-diaz: Would you prefer  `cached_header` or `header_data`?

rphmeier:
I'm curious as to what other approaches you have tried. It seems much more memory and computation light to cache the results of queries to `is_descendent_of` themselves. My thesis:
  - `is_descendent_of` queries will be common for short chains near the HEAD
  - queries deeper into the chain will tend to focus on some very specific blocks

The memory cache approach is clearly faster than doing nothing, but it requires keeping around a big LRU cache

core/client/src/blockchain.rs
rphmeier: `CachedHeaderMetadata` perhaps

marcio-diaz:
> I'm curious as to what other approaches you have tried. It seems much more memory and computation light to cache the results of queries to `is_descendent_of` themselves. My thesis:
> 
>     * `is_descendent_of` queries will be common for short chains near the HEAD
> 
>     * queries deeper into the chain will tend to focus on some very specific blocks
> 
> 
> The memory cache approach is clearly faster than doing nothing, but it requires keeping around a big LRU cache

It seems we are making many queries from `imported block` to `finalized block`, as I commented in the description, the cache is to keep this window. While syncing Kusama I have seen max differences of about 10k blocks. Probably we can reduce the cache size from 20k to 10k (or even more) and keep similar performance. I also thought about disabling it after main sync.

core/client/db/src/lib.rs
```diff
@@ -265,21 +271,46 @@ impl<'a> state_db::MetaDb for StateMetaDb<'a> {
 	}
 }
 
+struct HeaderCache<Block: BlockT> {
+	hash_to_data: LruCache<Block::Hash, LightHeader<Block>>,
```
rphmeier: architecturally, I don't follow why this would have to be in the `db` layer, when it is specifically a cache above the DB

marcio-diaz: I guess I can move the methods in `Header` trait a bit up to `client::blockchain::Backend`, where you have similar things like `leaves,() children(), cache()`. And still be able to use it from `is_descendent_of` and `tree_route`.

But I was trying to maximize usage, so people don't read full headers from the DB so often.

I agree that from the architectural viewpoint is not so nice.

rphmeier: Suggestion:
  1. introduce a separate trait for `HeaderMetadata`.
  2. implement it in a separate crate
  3. use this crate in `client-db` and do trait implementations + forwarding trait implementations for client, backends, etc.

core/client/src/blockchain.rs
```diff
@@ -24,6 +24,19 @@ use sr_primitives::Justification;
...
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct LightHeader<Block: BlockT> {
+	/// Hash of the header.
+	pub hash: Block::Hash,
+	/// Block number.
+	pub number: NumberFor<Block>,
+	/// Hash of parent header.
+	pub parent: Block::Hash,
+	/// Hash of an ancestor header. Used to jump through the tree.
+	pub ancestor: Block::Hash,
```
rphmeier: for the user of the light-header API, is this `ancestor` field actually useful? Seems like it will change all the time. Best to hide such implementation details

core/client/db/src/lib.rs
marcio-diaz: It seems like a good idea, but what would be the scope of such crate? what would be the methods to implement in the crate for `HeaderMetadata`? I guess more than just `set/get`, otherwise is too much overhead.

rphmeier: i don't understand what you mean. it's completely fine to have a trait with only 1 or 2 methods, and to extract that out to a separate crate.

arkpar:
> Right now, the query pattern of this function is as follows: lca(best, final), lca(best+1, final), lca(best+2, final), ..., lca(best+5000, final)

I wonder where is this pattern coming from. I imagine we only need to check tree paths or ancestry w.r.t the final block only when finalizing, and not when importing or checking each block. 

Does GRANDPA really need to check if each new block is descendent from the last finalized block?
@andresilva @rphmeier

core/client/src/blockchain.rs
```diff
@@ -195,56 +215,128 @@ impl<Block: BlockT> TreeRoute<Block> {
 	}
 }
 
+/// Get lowest common ancestor between two blocks in the tree.
+///
+/// Note: this implementation is efficient because of our current query pattern:
+/// lca(best, final) (O(h)), lca(best + 1, final) (O(1)), lca(best + 2, final) (O(1)), etc.
+pub fn lowest_common_ancestor<
```
arkpar: 1. Why not just make it a client method and get rid of functor arguments? It does require the client in all cases anyway.

2. Do I get it right that in order to function correctly this requires the arguments to follow ancestry order between calls? Otherwise the `ancestor` field may end up skipping the actual ancestor required for another call. Is there a fallback that work even if this condition is not upheld?

andresilva:
Some explanation of where `tree_routes` are being calculated regarding consensus:

- GRANDPA - we track pending authority set changes in a tree, we calculate `tree_routes` when making queries on this tree, either when importing a block if it includes a consensus change digest, or when finalizing a block to check if finalizing the given block finalizes anything in the tree.

- BABE - epoch changes are announced one epoch in advance and we also track them in a tree. Whenever we import a block that signals an epoch change we must import it into the tree, and we also query for epoch that just started (previously announced). Some changes introduced in #3583 changed some of this logic, but I think now it should cost less as we minimize the number of calls to `is_descendent_of` when querying the tree for pending epochs.

Nodes in the tree have an implicit ancestry relationship that is given by `is_descendent_of` (which uses `tree_route` internally).

I doubt that on Kusama the issue is being caused by GRANDPA since there are no authority changes happening. Most likely it was being caused by BABE (although that should have been improved by #3583).

Regardless of any improvements on BABE having a way to speed up common ancestry queries makes sense to me.

rphmeier:
In #5833 we'll do `is_descendent_of` calls for every epoch, since I removed the caching that it previously used. I think a really small LRU-cache to track which blocks we recently queried epoch for (so we can see if the new block would be in the same epoch as parent) would kill 95% of those queries.

core/client/src/blockchain.rs
marcio-diaz: 2. Yes, it's true, it has linear complexity in those cases (I'm missing something if you mean incorrect as with wrong result ^^). I thought about adding more links when you traverse a long path and about implementing a generic algorithm for LCA (maybe using something like https://www.topcoder.com/community/competitive-programming/tutorials/range-minimum-query-and-lowest-common-ancestor/), but I'm still not convinced is worth to lose the `O(1)` (in most cases) and simplicity. Do you have an idea?

marcio-diaz: Also, the paths are not so long, and if they are long we would need to save something to database probably.

marcio-diaz:
This is ready for another look, I introduce the crate and traits as  @rphmeier suggested (although I'm not sure about the location and name).

core/client/header-metadata/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+	/// Get a slice of enacted blocks (descendents of the common ancestor)
+	pub fn enacted(&self) -> &[HashAndNumber<Block>] {
+		&self.route[self.pivot + 1 ..]
+	}
+}
+
+/// Handles header metadata: hash, number, parent hash, etc.
+pub trait HeaderMetadata<Block: BlockT> {
+	type Metadata;
+	type Error;
```
rphmeier: missing docs on these items

core/client/header-metadata/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+	fn remove_header_metadata(&self, hash: Block::Hash);
+}
+
+/// Caches header metadata in an in-memory LRU cache.
+pub struct HeaderMetadataCache<Block: BlockT> {
+	cache: RwLock<LruCache<Block::Hash, CachedHeaderMetadata<Block>>>,
+}
+
+impl<Block: BlockT> HeaderMetadataCache<Block> {
+	pub fn new(capacity: usize) -> Self {
```
rphmeier: needs doc-comment

core/client/db/src/lib.rs
```diff
@@ -405,6 +404,36 @@ impl<Block: BlockT> client::blockchain::ProvideCache<Block> for BlockchainDb<Blo
...
 }
 
+impl<Block: BlockT> HeaderMetadata<Block> for BlockchainDb<Block> {
+	type Metadata = CachedHeaderMetadata<Block>;
+	type Error = client::error::Error;
+
+	fn header_metadata(&self, hash: Block::Hash) -> Result<Self::Metadata, Self::Error> {
+		self.header_metadata_cache.header_metadata(hash).or_else(|_| {
+			self.header(BlockId::hash(hash))?.map(|header| {
+					let header_metadata = CachedHeaderMetadata::from(&header);
```
andresilva: Extra indentation in this block?

core/client/src/in_mem.rs
```diff
@@ -320,6 +319,24 @@ impl<Block: BlockT> HeaderBackend<Block> for Blockchain<Block> {
 	}
 }
 
+impl<Block: BlockT> HeaderMetadata<Block> for Blockchain<Block> {
+	type Metadata = CachedHeaderMetadata<Block>;
+	type Error = error::Error;
+
+	fn header_metadata(&self, hash: Block::Hash) -> Result<Self::Metadata, Self::Error> {
+			self.header(BlockId::hash(hash))?.map(|header| CachedHeaderMetadata::from(&header))
```
andresilva: Extra indentation in this block?

core/client/src/light/blockchain.rs
```diff
@@ -281,6 +302,20 @@ pub mod tests {
 		}
 	}
 
+	impl HeaderMetadata<Block> for DummyStorage {
+		type Metadata = CachedHeaderMetadata<Block>;
+		type Error = ClientError;
+
+		fn header_metadata(&self, hash: Hash) -> Result<Self::Metadata, Self::Error> {
+				self.header(BlockId::hash(hash))?.map(|header| CachedHeaderMetadata::from(&header))
```
andresilva: Extra indentation.

core/client/src/blockchain.rs
andresilva: Even in the case where our ancestor is too old and we fallback to linear search it should be faster than what we had previously because this linear search will very likely all be done in-memory (all hashes will most likely be cached because we add them when importing a block).

marcio-diaz:
Nits addressed @rphmeier @andresilva

core/client/db/src/lib.rs
```diff
@@ -2134,18 +2177,58 @@ mod tests {
...
+			let tree_route = tree_route(&blockchain, block0, block1).unwrap();
 
 			assert_eq!(tree_route.common_block().hash, block0);
 			assert!(tree_route.retracted().is_empty());
 			assert_eq!(tree_route.enacted().iter().map(|r| r.hash).collect::<Vec<_>>(), vec![block1]);
 		}
 	}
 
+	#[test]
+	fn lowest_common_ancestor_works() {
```
andresilva: Is this test actually using ancestor links anywhere? Maybe we could add some more edge cases.

marcio-diaz: It's is used because they are executed before and initialized to parents. I added a couple of cases more.

core/cli/src/informant.rs
```diff
@@ -17,10 +17,11 @@
 //! Console informant. Prints sync progress and block events. Runs on the calling thread.
 
 use client::BlockchainEvents;
+use header_metadata::lowest_common_ancestor;
```
andresilva:
```suggestion
```

core/cli/src/informant.rs
```diff
@@ -47,19 +48,18 @@ pub fn build(service: &impl AbstractService) -> impl Future<Item = (), Error = (
 		// detect and log reorganizations.
 		if let Some((ref last_num, ref last_hash)) = last_best {
 			if n.header.parent_hash() != last_hash && n.is_new_best  {
-				let tree_route = ::client::blockchain::tree_route(
-					|id| client.header(&id)?.ok_or_else(
-						|| client::error::Error::UnknownBlock(format!("{:?}", id))),
-					BlockId::Hash(last_hash.clone()),
-					BlockId::Hash(n.hash),
+				let maybe_ancestor = lowest_common_ancestor(
```
andresilva:
```suggestion
				let maybe_ancestor = header_metadata::lowest_common_ancestor(
```

core/client/db/src/lib.rs
```diff
@@ -61,10 +62,11 @@ use state_machine::{
 	DBValue, ChangesTrieTransaction, ChangesTrieCacheAction, ChangesTrieBuildCache,
 	backend::Backend as StateBackend,
 };
-use crate::utils::{Meta, db_err, meta_keys, read_db, block_id_to_lookup_key, read_meta};
+use crate::utils::{Meta, db_err, meta_keys, read_db, read_meta};
 use client::leaves::{LeafSet, FinalizationDisplaced};
 use client::children;
 use state_db::StateDb;
+use header_metadata::{HeaderMetadata, HeaderMetadataCache, CachedHeaderMetadata, tree_route};
```
andresilva:
```suggestion
use header_metadata::{CachedHeaderMetadata, HeaderMetadata, HeaderMetadataCache};
```

core/client/db/src/lib.rs
```diff
@@ -406,6 +405,56 @@ impl<Block: BlockT> client::blockchain::ProvideCache<Block> for BlockchainDb<Blo
...
+	fn insert_header_metadata(&self, hash: Block::Hash, metadata: CachedHeaderMetadata<Block>) {
+		self.header_metadata_cache.insert_header_metadata(hash, metadata)
+	}
+
+	fn remove_header_metadata(&self, hash: Block::Hash) {
+		self.header_metadata_cache.remove_header_metadata(hash);
+	}
+}
+
+impl<Block: BlockT> HeaderMetadata<Block> for &BlockchainDb<Block> {
```
andresilva: Can't we just reuse the code from above?

core/client/db/src/lib.rs
```diff
@@ -406,6 +405,56 @@ impl<Block: BlockT> client::blockchain::ProvideCache<Block> for BlockchainDb<Blo
...
+	type Error = client::error::Error;
+
+	fn header_metadata(&self, hash: Block::Hash) -> Result<CachedHeaderMetadata<Block>, Self::Error> {
```
andresilva: `HeaderMetadata::header_metadata(*self)` or something like that?

core/client/db/src/lib.rs
```diff
@@ -904,21 +953,20 @@ impl<Block: BlockT<Hash=H256>> Backend<Block> {
...
 		// cannot find tree route with empty DB.
 		if meta.best_hash != Default::default() {
-			let tree_route = ::client::blockchain::tree_route(
-				|id| self.blockchain.header(id)?.ok_or_else(
-					|| client::error::Error::UnknownBlock(format!("{:?}", id))
-				),
-				BlockId::Hash(meta.best_hash),
-				BlockId::Hash(route_to),
-			)?;
+			let tree_route = tree_route(&self.blockchain, meta.best_hash, route_to)?;
```
andresilva:
```suggestion
			let tree_route = header_metadata::tree_route(&self.blockchain, meta.best_hash, route_to)?;
```

core/client/db/src/light.rs
```diff
@@ -35,6 +35,7 @@ use codec::{Decode, Encode};
 use primitives::Blake2Hasher;
 use sr_primitives::generic::{DigestItem, BlockId};
 use sr_primitives::traits::{Block as BlockT, Header as HeaderT, Zero, One, NumberFor};
+use header_metadata::{CachedHeaderMetadata, HeaderMetadata, HeaderMetadataCache, tree_route};
```
andresilva:
```suggestion
use header_metadata::{CachedHeaderMetadata, HeaderMetadata, HeaderMetadataCache};
```

core/client/db/src/light.rs
```diff
@@ -208,17 +237,18 @@ impl<Block: BlockT> LightStorage<Block> {
...
 
 		// handle reorg.
 		let meta = self.meta.read();
 		if meta.best_hash != Default::default() {
-			let tree_route = ::client::blockchain::tree_route(
-				|id| self.header(id)?.ok_or_else(|| client::error::Error::UnknownBlock(format!("{:?}", id))),
-				BlockId::Hash(meta.best_hash),
-				BlockId::Hash(route_to),
-			)?;
+			let tree_route = tree_route(self, meta.best_hash, route_to)?;
```
andresilva:
```suggestion
			let tree_route = header_metadata::tree_route(self, meta.best_hash, route_to)?;
```

core/client/src/client.rs
```diff
@@ -51,6 +51,7 @@ use consensus::{
 	ImportResult, BlockOrigin, ForkChoiceStrategy,
 	SelectChain, self,
 };
+use header_metadata::{HeaderMetadata, CachedHeaderMetadata, tree_route, lowest_common_ancestor};
```
andresilva:
```suggestion
use header_metadata::{CachedHeaderMetadata, HeaderMetadata};
```

core/client/src/client.rs
```diff
@@ -966,11 +967,7 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 		};
 
 		let retracted = if is_new_best {
-			let route_from_best = crate::blockchain::tree_route(
-				|id| self.header(&id)?.ok_or_else(|| Error::UnknownBlock(format!("{:?}", id))),
-				BlockId::Hash(info.best_hash),
-				BlockId::Hash(parent_hash),
-			)?;
+			let route_from_best = tree_route(self.backend.blockchain(), info.best_hash, parent_hash)?;
```
andresilva:
```suggestion
			let route_from_best = header_metadata::tree_route(self.backend.blockchain(), info.best_hash, parent_hash)?;
```

core/client/src/client.rs
```diff
@@ -1100,11 +1097,7 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 			return Ok(());
 		}
 
-		let route_from_finalized = crate::blockchain::tree_route(
-			|id| self.header(&id)?.ok_or_else(|| Error::UnknownBlock(format!("{:?}", id))),
-			BlockId::Hash(last_finalized),
-			BlockId::Hash(block),
-		)?;
+		let route_from_finalized = tree_route(self.backend.blockchain(), last_finalized, block)?;
```
andresilva:
```suggestion
		let route_from_finalized = header_metadata::tree_route(self.backend.blockchain(), last_finalized, block)?;
```

core/client/src/client.rs
```diff
@@ -1113,11 +1106,7 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 			return Err(error::Error::NotInFinalizedChain);
 		}
 
-		let route_from_best = crate::blockchain::tree_route(
-			|id| self.header(&id)?.ok_or_else(|| Error::UnknownBlock(format!("{:?}", id))),
-			BlockId::Hash(best_block),
-			BlockId::Hash(block),
-		)?;
+		let route_from_best = tree_route(self.backend.blockchain(), best_block, block)?;
```
andresilva:
```suggestion
		let route_from_best = header_metadata::tree_route(self.backend.blockchain(), best_block, block)?;
```

core/client/src/client.rs
```diff
@@ -1831,13 +1840,9 @@ pub mod utils {
 				}
 			}
 
-			let tree_route = blockchain::tree_route(
-				|id| client.header(id)?.ok_or_else(|| Error::UnknownBlock(format!("{:?}", id))),
-				BlockId::Hash(*hash),
-				BlockId::Hash(*base),
-			)?;
+			let ancestor = lowest_common_ancestor(client, *hash, *base)?;
```
andresilva:
```suggestion
			let ancestor = header_metadata::lowest_common_ancestor(client, *hash, *base)?;
```

core/client/src/in_mem.rs
```diff
@@ -27,12 +27,15 @@ use state_machine::backend::{Backend as StateBackend, InMemory};
 use state_machine::{self, InMemoryChangesTrieStorage, ChangesTrieAnchorBlockId, ChangesTrieTransaction};
 use hash_db::{Hasher, Prefix};
 use trie::MemoryDB;
+use header_metadata::{CachedHeaderMetadata, HeaderMetadata, tree_route};
```
andresilva:
```suggestion
use header_metadata::{CachedHeaderMetadata, HeaderMetadata};
```

core/client/src/in_mem.rs
```diff
@@ -221,11 +224,7 @@ impl<Block: BlockT> Blockchain<Block> {
 			if &best_hash == header.parent_hash() {
 				None
 			} else {
-				let route = crate::blockchain::tree_route(
-					|id| self.header(id)?.ok_or_else(|| error::Error::UnknownBlock(format!("{:?}", id))),
-					BlockId::Hash(best_hash),
-					BlockId::Hash(*header.parent_hash()),
-				)?;
+				let route = tree_route(self, best_hash, *header.parent_hash())?;
```
andresilva:
```suggestion
				let route = header_metadata::tree_route(self, best_hash, *header.parent_hash())?;
```

core/client/src/in_mem.rs
```diff
@@ -320,6 +319,21 @@ impl<Block: BlockT> HeaderBackend<Block> for Blockchain<Block> {
 	}
 }
 
+impl<Block: BlockT> HeaderMetadata<Block> for Blockchain<Block> {
+	type Error = error::Error;
+
+	fn header_metadata(&self, hash: Block::Hash) -> Result<CachedHeaderMetadata<Block>, Self::Error> {
+			self.header(BlockId::hash(hash))?.map(|header| CachedHeaderMetadata::from(&header))
```
andresilva: This is still indented wrong.

core/finality-grandpa/src/environment.rs
```diff
@@ -498,11 +499,7 @@ pub(crate) fn ancestry<B, Block: BlockT<Hash=H256>, E, RA>(
 {
 	if base == block { return Err(GrandpaError::NotDescendent) }
 
-	let tree_route_res = ::client::blockchain::tree_route(
-		|id| client.header(&id)?.ok_or(client::error::Error::UnknownBlock(format!("{:?}", id))),
-		BlockId::Hash(block),
-		BlockId::Hash(base),
-	);
+	let tree_route_res = tree_route(client, block, base);
```
andresilva:
```suggestion
	let tree_route_res = header_metadata::tree_route(client, block, base);
```

core/finality-grandpa/src/environment.rs
```diff
@@ -47,6 +47,7 @@ use crate::{
 };
 
 use consensus_common::SelectChain;
+use header_metadata::tree_route;
```
andresilva:
```suggestion
```

core/network/src/chain.rs
```diff
@@ -24,6 +24,7 @@ use sr_primitives::traits::{Block as BlockT, Header as HeaderT};
 use sr_primitives::generic::{BlockId};
 use sr_primitives::Justification;
 use primitives::{H256, Blake2Hasher, storage::StorageKey};
+use header_metadata::lowest_common_ancestor;
```
andresilva:
```suggestion
```

core/network/src/chain.rs
```diff
@@ -153,14 +154,8 @@ impl<B, E, Block, RA> Client<Block> for SubstrateClient<B, E, Block, RA> where
...
 		}
 
-		let tree_route = ::client::blockchain::tree_route(
-			|id| self.header(&id)?.ok_or_else(||
-				client::error::Error::UnknownBlock(format!("{:?}", id))
-			),
-			BlockId::Hash(*block),
-			BlockId::Hash(*base),
-		)?;
+		let ancestor = lowest_common_ancestor(self, *block, *base)?;
```
andresilva:
```suggestion
		let ancestor = header_metadata::lowest_common_ancestor(self, *block, *base)?;
```

PR #3669, opened 2019-09-23
fix #2177: port GRANDPA node-side code to use Consensus digests

### LINKED ISSUE #2177: Update GRANDPA to use new `Consensus` digests
Now that #2083 is done, we can remove the `Other`-style decoding of GRANDPA digests.

The old GrandpaApi trait functions for interpreting header digests should be deprecated but remain available for backwards compatibility.

rphmeier:
Removes the old runtime APIs for inspecting these digests.


core/finality-grandpa/src/import.rs
```diff
@@ -174,75 +168,69 @@ impl<'a, Block: 'a + BlockT> Drop for PendingSetChanges<'a, Block> {
...
+		ConsensusLog::ScheduledChange(change) => Some(change),
+		_ => None,
+	};
+
+	// find the first consensus digest with the right ID which converts to
+	// the right kind of consensus log.
+	header.digest().convert_first(|l| l.try_to(id).and_then(filter_log))
+}
+
+fn find_forced_change<B: BlockT>(header: &B::Header)
```
DemiMarie-parity: We can obtain a marginal speedup by fusing this function with `find_scheduled_change`, but I doubt it is worthwhile.  We do, however, need to check that there is at most one of each kind of consensus digest, as I did in BABE.  Alternatively, if multiple such digests are allowed, they need to be handled appropriately.

rphmeier: We give preference to the first digest; subsequent ones are allowed but ignored.

PR #2869, opened 2019-06-14
client: Reorg to new best when finalizing divergent branch

andresilva:
When we finalize a block on a divergent branch from the currently tracked best we need to figure out what the new best block is on the finalized branch and we need to re-org to it. This is done by having `finalize_block` optionally take a `SelectChain` which is the pluggable logic for best fork selection (currently only longest chain is implemented). If no `SelectChain` is provided and the finalized block is on a divergent branch then we set the finalized block as best (this is the case for the light client since it doesn't track leaves). When importing a block that is finalized the same thing can happen but in this case we don't need `SelectChain` since the given block that's being imported must be the new best.
I also renamed the `finality_target` in `SelectChain` since there's nothing relating to finality there (e.g. GRANDPA has further voting rules that restrict what we vote on), instead `SelectChain` is strictly concerned with defining the notion of a best block, regardless of what we do with it.

Fix #1442.

PR #3673, opened 2019-09-23
grandpa: pluggable voting rules

andresilva:
This PR adds support for plugging arbitrary voting rules to restrict GRANDPA votes.  A `VotingRule` trait is introduced that allows restricting a GRANDPA vote based on what was the initial best finality target and the current finality target.

The GRANDPA environment uses `SelectChain` to find an initial finality target candidate, additionally the GRANDPA environment restricts the vote to be on 3/4 of the unfinalized chain and furthermore the vote is limited by any pending authority set change. After these mandatory voting rules, the environment uses the given `VotingRule` to further restrict the vote if necessary. 

The node is updated to use a voting rule to always vote behind the best block. I also added a composite voting rule and builder that allows combining multiple voting rules. This might be useful in polkadot to implement voting rules that take into account data availability.


rphmeier:
Could the 3/4s rule be repurposed as a voting rule as well?

core/finality-grandpa/src/voting_rule.rs
```diff
@@ -0,0 +1,181 @@
...
+impl<B, Block> VotingRule<Block> for VotingRules<B, Block> where
+	B: HeaderBackend<Block>,
+	Block: BlockT<Hash = H256>,
+{
+	fn restrict_vote(
+		&self,
+		best_target: &Block::Header,
+		current_target: &Block::Header,
+	) -> Option<(Block::Hash, NumberFor<Block>)> {
+		let restricted_target = self.rules.iter().fold(
```
HCastano: nit: I think this section is easier to read after a little bit of Rustfmt-ing:

```Rust
        let restricted_target =
            self.rules
                .iter()
                .fold(current_target.clone(), |current_target, rule| {
                    rule.restrict_vote(best_target, &current_target)
                        .and_then(|(hash, _)| self.backend.header(BlockId::Hash(hash)).ok())
                        .and_then(std::convert::identity)
                        .unwrap_or(current_target)
                });
```

andresilva:
Going to rejig the PR to instead do this with `SelectChain` rather than creating a new `VotingRule` trait.

core/finality-grandpa/src/environment.rs
```diff
@@ -396,16 +398,17 @@ impl<B, E, Block: BlockT, N: Network<Block>, RA, SC> Environment<B, E, Block, N,
...
-for Environment<B, E, Block, N, RA, SC>
+for Environment<B, E, Block, N, RA, SC, VR>
 where
 	Block: 'static,
 	B: Backend<Block, Blake2Hasher> + 'static,
 	E: CallExecutor<Block, Blake2Hasher> + 'static,
 	N: Network<Block> + 'static,
 	N::In: 'static,
 	SC: SelectChain<Block> + 'static,
+	VR: VotingRule<Block>,
```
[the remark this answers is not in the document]
andresilva: We need to guarantee it is `Clone` somewhere else, with a trait object we can't enforce that bound.

core/finality-grandpa/src/voting_rule.rs
andresilva: I don't have any strong opinions about code formatting but this is not how our _undefined_ code style looks like.

andresilva:
I think that `SelectChain` is not appropriate for implementing voting rules, specifically it is not very amenable to composition. Still the separation seems appropriate to me: `SelectChain` is concerned with picking some "best" chain (longest, GHOST, etc.) which is used as an initial target for finality voting, and then on top of that we optionally restrict it with GRANDPA-specific `VotingRule`s.

I've addressed all the previous comments and moved the 3/4 unfinalized chain voting logic to a `VotingRule`. The only rule which is implicit/mandatory is limiting votes to authority set changes which seems sensible to me since it is fundamental for the security of the protocol itself.

core/finality-grandpa/src/environment.rs
```diff
@@ -462,7 +462,11 @@ where
 					// walk backwards until we find the target block
 					loop {
 						if *target_header.number() < target_number {
-							unreachable!();
+							unreachable!(
+								"we are traversing backwards from a know block; \
```
DemiMarie-parity: typo: `know` → `known`

andresilva: Weird thing, I had the exact same message elsewhere but without the typo. Does this mean I did not copy-paste it when writing the code? :thinking::smile:

core/finality-grandpa/src/voting_rule.rs
```diff
@@ -0,0 +1,270 @@
...
+	/// order to pick our target (e.g. last finalized block), and `best_target`
+	/// is the initial best vote target before any vote rules were applied. When
+	/// applying multiple `VotingRule`s both `base` and `best_target` should
+	/// remain unchanged.
+	///
+	/// The contract of this interface requires that when restricting a vote, the
+	/// returned value **must** be an ancestor of the given `current_target`,
+	/// this also means that a variant must be maintained throughout the
+	/// execution of voting rules wherein `current_target <= best_target`.
+	fn restrict_vote(
```
rphmeier: the generic parameter could also be on the function here

core/finality-grandpa/src/voting_rule.rs
```diff
@@ -0,0 +1,270 @@
...
+
+/// A trait for custom voting rules in GRANDPA.
+pub trait VotingRule<Block, B>: Send + Sync where
+	Block: BlockT,
+	B: HeaderBackend<Block>,
+{
+	/// Restrict the given `current_target` vote, returning the block hash and
+	/// number of the block to vote on, and `None` in case the vote should not
+	/// be restricted. `base` is the block that we're basing our votes on in
+	/// order to pick our target (e.g. last finalized block), and `best_target`
```
rphmeier: base should be the last-round-estimate, not last finalized?

core/finality-grandpa/src/environment.rs
```diff
@@ -429,7 +433,7 @@ where
 		debug!(target: "afg", "Finding best chain containing block {:?} with number limit {:?}", block, limit);
 
 		match self.select_chain.finality_target(block, None) {
-			Ok(Some(mut best_hash)) => {
+			Ok(Some(best_hash)) => {
```
rphmeier: I guess in this context I don't understand the purpose of `finality_target` anymore.

core/finality-grandpa/src/voting_rule.rs
andresilva: We use `VotingRule` as a trait object so we can't have a generic here.

core/finality-grandpa/src/environment.rs
andresilva: I agree, we already had a discussion about `SelectChain` on https://github.com/paritytech/substrate/pull/2869#issuecomment-502267671. Since we now have voting rules to implement the vote target logic what do you think about renaming `finality_target` in `SelectChain` to `best_containing`? In this case `SelectChain` is only responsible for implementing some custom notion of "best".

gavofyork:
@rphmeier still ok with this?

Issue #1876, opened 2019-02-26
Avoid some clone when creating keys to use in StorageMap methods


thiolliere:
When we create a StorageMap with a tuple for example `(T::AccountId, u64)` [here in substrate kitties](https://github.com/shawntabrizi/substrate-collectables-workshop/blob/eb2ebfdfc0f68dcbaa02e585e9dd0665dd59db1c/2/assets/2.4-finished-code.rs#L39) then to use it we often need to clone AccountId (it doesn't implements Copy) like [here](https://github.com/shawntabrizi/substrate-collectables-workshop/blob/eb2ebfdfc0f68dcbaa02e585e9dd0665dd59db1c/2/assets/2.4-finished-code.rs#L85)

In this example it is not an issue as AccountId is meant to be an integer such as u64 and clone it is probably same cost as referencing it.

But in other situation it could be an issue. I don't know if such situations exist though (long key that shouldn't be cloned, (caching there encoding, I don't know)).

A solution I have in mind for that is using a trait SameEncodingAs as such:
```rust
pub trait SameEncodeAs<T>: codec::Encode {}

impl<'a, 'b, A: codec::Encode, B: codec::Encode> SameEncodeAs<(A, B)> for (&'a A, &'b B) {}
impl<T: codec::Encode> SameEncodeAs<T> for T {}
```

a partial implement is then (based on ebe4c972bab088a760ca43e7662bf750ccf1e919):
```diff
diff --git a/srml/example/src/lib.rs b/srml/example/src/lib.rs
index 4f4e0b96..98b93d81 100644
--- a/srml/example/src/lib.rs
+++ b/srml/example/src/lib.rs
@@ -20,7 +20,7 @@
 // Ensure we're `no_std` when compiling for Wasm.
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use srml_support::{StorageValue, dispatch::Result, decl_module, decl_storage, decl_event};
+use srml_support::{StorageValue, dispatch::Result, decl_module, decl_storage, decl_event, storage::generator::StorageMap, storage::RuntimeStorage};
 use system::ensure_signed;
 
 /// Our module's configuration trait. All our types and consts go in here. If the
@@ -63,8 +63,10 @@ decl_storage! {
 		// `fn getter_name(key: KeyType) -> ValueType` for map items.
 		Dummy get(dummy) config(): Option<T::Balance>;
 
+		TestDouble: map (T::AccountId, T::AccountId) => u32;
+
[...]

PR #3676, opened 2019-09-24
Use EncodeLike for storages traits

thiolliere:
Fix https://github.com/paritytech/substrate/issues/1876

# Done:

* this makes storage map, linked map and else use encodelike instead of there direct argument.

# Critics:

* There is usage of Ref::from in map and linkedmap, with some refactor we can avoid those and derive the final key only once.
   I think we could do refactoring of this in a following PR
* ppl have to bound FullCodec or EncodeLike in their associated type of their module.


srml/session/src/lib.rs
```diff
@@ -628,27 +628,27 @@ impl<T: Trait> Module<T> {
...
 	}
 
 	fn put_key_owner(id: KeyTypeId, key_data: &[u8], v: &T::ValidatorId) {
-		<KeyOwner<T>>::insert(DEDUP_KEY_PREFIX, &(id, key_data.to_vec()), v)
+		<KeyOwner<T>>::insert(&DEDUP_KEY_PREFIX, &(id, key_data.to_vec()), v)
 	}
 
 	fn clear_key_owner(id: KeyTypeId, key_data: &[u8]) {
-		<KeyOwner<T>>::remove(DEDUP_KEY_PREFIX, &(id, key_data.to_vec()));
+		<KeyOwner<T>>::remove(&DEDUP_KEY_PREFIX, &(id, key_data.to_vec()));
```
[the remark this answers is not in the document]
thiolliere: done, and tried to catch up most of them

thiolliere:
also maybe I could check for compile time regression

srml/support/src/storage/mod.rs
```diff
@@ -61,11 +56,14 @@ pub trait StorageValue<T: Codec> {
 	/// Append the given item to the value in the storage.
 	///
 	/// `T` is required to implement `codec::EncodeAppend`.
-	fn append<'a, I, R>(items: R) -> Result<(), &'static str> where
-		I: 'a + Encode,
-		T: EncodeAppend<Item=I>,
-		R: IntoIterator<Item=&'a I>,
-		R::IntoIter: ExactSizeIterator;
+	///
```
bkchr:
```suggestion
```

srml/support/src/storage/storage_items.rs
```diff
@@ -839,13 +840,17 @@ mod test_append_and_len {
 	#[test]
 	fn append_or_put_works() {
 		with_externalities(&mut TestExternalities::default(), || {
-			let _ = MapVec::append_or_insert(1, [1, 2, 3].iter());
-			let _ = MapVec::append_or_insert(1, [4, 5].iter());
```
bkchr: Shouldn't this work as well?

srml/democracy/src/lib.rs
```diff
@@ -929,7 +929,7 @@ impl<T: Trait> Module<T> {
 			} else {
 				<DispatchQueue<T>>::append_or_insert(
 					now + info.delay,
-					[Some((info.proposal, index))].into_iter()
+					&[Some((info.proposal, index))][..]
```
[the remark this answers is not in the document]
thiolliere: no it will compile with something like:
```
the trait bound `Iter<'_, std::option::Option<(<T as Trait>::Proposal, u32)>>: EncodeLike<std::vec::Vec<std::option::Option<(<T as Trait>::Proposal, u32)>>>` is not satisfied
```

srml/elections/src/lib.rs
```diff
@@ -816,7 +816,7 @@ impl<T: Trait> Module<T> {
 					if set_len + 1 == VOTER_SET_SIZE {
 						NextVoterSet::put(next + 1);
 					}
-					<Voters<T>>::append_or_insert(next, [Some(who.clone())].into_iter())
+					<Voters<T>>::append_or_insert(next, &[Some(who.clone())][..])
```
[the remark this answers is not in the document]
thiolliere: yes because satisfying both interator constrained and `EncodeLike<Vec<T>>` is hard.

srml/democracy/src/lib.rs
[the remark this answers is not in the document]
thiolliere: hmm at some point we should derive EncodeLike for standard iterator structures.

srml/assets/src/lib.rs
```diff
@@ -159,7 +159,7 @@ decl_module! {
 			let id = Self::next_asset_id();
 			<NextAssetId<T>>::mutate(|id| *id += One::one());
 
-			<Balances<T>>::insert((id, origin.clone()), total);
+			<Balances<T>>::insert((id, &origin), total);
```
kianenigma: cc @shawntabrizi in v2 for kitties tutorial: the original issue of the PR references the v2 tutorial. If keys of type tuple are still around they can be refactored now

srml/democracy/src/lib.rs
```diff
@@ -377,10 +377,10 @@ decl_module! {
 
 			let index = Self::public_prop_count();
 			PublicPropCount::put(index + 1);
-			<DepositOf<T>>::insert(index, (value, vec![who.clone()]));
+			<DepositOf<T>>::insert(index, (value, &[&who][..]));
 
-			let new_prop = (index, (*proposal).clone(), who);
-			<PublicProps<T>>::append_or_put([new_prop].into_iter());
+			let new_prop = (index, proposal, who);
+			<PublicProps<T>>::append_or_put(&[Ref::from(&new_prop)][..]);
```
kianenigma: If can be refactored and you think it is worth it, would be good to have a small issue for it after the PR is merged 👍

srml/support/src/storage/generator/linked_map.rs
```diff
@@ -96,13 +93,29 @@ impl<Key> Default for Linkage<Key> {
...
 
+// Encode like a linkage.
+#[derive(Encode)]
+struct EncodeLikeLinkage<PKey: EncodeLike<Key>, NKey: EncodeLike<Key>, Key: Encode> {
+	// Previous element key in storage (None for the first element)
+	previous: Option<PKey>,
+	// Next element key in storage (None for the last element)
+	next: Option<NKey>,
+	// The key of the linkage this type encode to
+	phantom: core::marker::PhantomData<Key>,
```
kianenigma: import from rstd for consistency?

srml/democracy/src/lib.rs
thiolliere: Ref from inside srml-support/storage/generator like in the implementation of linkedmap and map.
here it is legit. no super beautiful indeed.

srml/assets/src/lib.rs
shawntabrizi: awesome!

PR #3677, opened 2019-09-24
Refactor NativeExecutor to support multiple Wasm execution methods

jimpo:
This is a further refactor of `substrate-executor` leading the way towards wasmtime integration. This PR should make clear how a compiled execution method would be added and somewhat decouples `NativeExecutor` and `RuntimesCache` from Wasmi execution through a new `WasmRuntime` trait.


core/executor/src/wasmi_runtimes_cache.rs
```diff
@@ -17,13 +17,14 @@
 //! Implements a cache for pre-created Wasm runtime module instances.
 
 use crate::error::Error;
-use crate::wasm_executor::WasmExecutor;
+use crate::native_executor::SubstrateRuntime;
```
bkchr: Can we not have a cache that works for every possible wasm executor?

jimpo: No, the cache is rather different between at least Wasmi and Wasmtime. In particular, the structs that are cached are completed different.

It's probably possible to make another trait for a SubstrateRuntime cache which returns different dynamically boxed SubstrateRuntimes depending on the implementation, but I haven't seen much benefit in that.

core/executor/src/native_executor.rs
```diff
@@ -59,28 +62,78 @@ pub trait NativeExecutionDispatch: Send + Sync {
...
 impl<D: NativeExecutionDispatch> NativeExecutor<D> {
 	/// Create new instance.
-	pub fn new(default_heap_pages: Option<u64>) -> Self {
+	///
+	/// # Parameters
+	///
+	/// `wasm_method` - Method used to execute fallback Wasm code.
+	///
+	/// `default_heap_pages` - Number of 64KB pages to allocate for Wasm execution.
+	/// Defaults to `DEFAULT_HEAP_PAGES` if `None` is provided.
```
bkchr: Could you please align with the upper line after the `-`

core/executor/src/native_executor.rs
```diff
@@ -59,28 +62,78 @@ pub trait NativeExecutionDispatch: Send + Sync {
...
+		f: impl for <'a> FnOnce(Box<dyn SubstrateRuntime + 'a>, &'a mut E) -> Result<R>
+	) -> Result<R>
+		where E: Externalities<Blake2Hasher>
+	{
+		match self.wasm_method {
+			WasmExecutionMethod::Interpreted =>
+				RUNTIMES_CACHE.with(|cache| {
+					let mut cache = cache.borrow_mut();
+					let runtime = cache.fetch_runtime(&WasmiExecutor, ext, self.default_heap_pages)?;
+					f(Box::new(runtime), ext)
```
bkchr: Why do you box the value here?

core/executor/src/wasmi_runtimes_cache.rs
bkchr: Yeah I know that the inner structs would be different, anything else would not make any sense at all.

But given that we can reuse the rest of the logic, I don't see why we should split the stuff. You already have your `SubstrateRuntime` that basically abstracts over the cached runtime. You should probably just rename `SubstrateRuntime` to `CachedRuntime` and use this trait to store the runtimes internally.

node/runtime/src/lib.rs
```diff
@@ -85,7 +85,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 162,
-	impl_version: 162,
+	impl_version: 163,
```
bkchr: Not required.

jimpo: I agree but the linter got mad b/c `core/sr-api-macros` changed.

In commit message: "Nothing at all actually changed, but core/sr-api-macros was updated to
pass a new arg to NativeExecutor and the linter things that requires an impl update."

core/executor/src/wasmi_runtimes_cache.rs
jimpo: For the wasmtime runtime (and the WasmiRuntime too, but to a lesser degree), you can't just return the cached object. You use the cached object to construct the actual runtime. In the case of `WasmiRuntime` it has a reference to an executor, which is not cached. This is avoidable, but not for wasmtime.

core/executor/src/native_executor.rs
jimpo: If `self.wasm_method` is different, it might be a different impl of SubstrateRuntime, so it needs to be a trait object. Would you prefer it be a `&'a dyn SubstrateRuntime`?

node/runtime/src/lib.rs
bkchr: Yeah, just ignore it.

core/executor/src/native_executor.rs
bkchr: Yes. Box requires allocating memory.

core/executor/src/native_executor.rs
```diff
@@ -59,28 +62,78 @@ pub trait NativeExecutionDispatch: Send + Sync {
 	fn native_version() -> NativeVersion;
 }
 
+/// The callable Substrate runtime. When called, a version of the Wasm runtime code is executed.
+pub trait SubstrateRuntime {
```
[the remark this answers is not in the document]
pepyakin: While I also don't find `SubstrateRuntime` a good term, I think `WasmRuntime` might be not much better. This is because wasm runtime can mean something that executes wasm. OTOH in the context of substrate it means the wasm version of substrate runtime.

node/executor/src/lib.rs
```diff
@@ -117,7 +117,7 @@ mod tests {
 	}
 
 	fn executor() -> ::substrate_executor::NativeExecutor<Executor> {
-		substrate_executor::NativeExecutor::new(None)
+		substrate_executor::NativeExecutor::new(Default::default(), None)
```
pepyakin: I think this grown unwieldy at this point. Should we provide an associated function on `NativeExecutor` to create an instance configured with defaults?

core/executor/src/native_executor.rs
```diff
@@ -59,28 +62,78 @@ pub trait NativeExecutionDispatch: Send + Sync {
...
 	/// Create new instance.
-	pub fn new(default_heap_pages: Option<u64>) -> Self {
+	///
+	/// # Parameters
+	///
+	/// `wasm_method` - Method used to execute fallback Wasm code.
+	///
+	/// `default_heap_pages` - Number of 64KB pages to allocate for Wasm execution.
+	/// Defaults to `DEFAULT_HEAP_PAGES` if `None` is provided.
+	pub fn new(wasm_method: WasmExecutionMethod, default_heap_pages: Option<u64>) -> Self {
```
pepyakin: I would argue that we should have not added this argument without actually threading it to the actual configuration. Otherwise, there is a chance that we miss some places while updating

node/executor/src/lib.rs
jimpo: We could implement `Default` for `NativeExecutor`?

core/executor/src/wasmi_runtimes_cache.rs
jimpo: OK, I made the RuntimesCache store all types of runtimes.

node/executor/src/lib.rs
pepyakin: Yeah you got me: I have a bit of allergy on it since it is being overused. You don't have to go far for an example, you can see it in the diff: `Default::default()`. Yes, it is easier to write and it doesn't need to be updated if the type of an argument changes, but I would argue that one should optimize for readability not convenience of writing.

I won't block merge a PR that implements `Default` for `NativeExecutor` though, I think at this moment it is just matter of my preference. So your call.

core/executor/src/native_executor.rs
jimpo: There's only one place that a `NativeExecutor` is constructed outside of tests, so I feel like that's not much of a risk. But I'm happy to add the configuration of the Wasm method in this PR as well.

jimpo: Added a commit `cli: CLI configuration for Wasm execution method.` which does this.

node/executor/src/lib.rs
jimpo: Decided to remove the `impl Default for WasmExecutionMethod` and pass it explicitly everywhere.

core/client/src/light/fetcher.rs
```diff
@@ -503,7 +503,7 @@ pub mod tests {
 	use parking_lot::Mutex;
 	use codec::Decode;
 	use crate::client::tests::prepare_client_with_key_changes;
-	use executor::{self, NativeExecutor};
+	use executor::{self, NativeExecutor, WasmExecutionMethod};
```
bkchr:
```suggestion
	use executor::{NativeExecutor, WasmExecutionMethod};
```

core/test-client/src/lib.rs
```diff
@@ -209,18 +209,20 @@ impl<E, Backend, G: GenesisInit> TestClientBuilder<
...
 		client::LongestChain<Backend, Block>,
 	) where
-		I: Into<Option<executor::NativeExecutor<E>>>,
+		I: Into<Option<NativeExecutor<E>>>,
 		E: executor::NativeExecutionDispatch,
 		Backend: client::backend::Backend<Block, Blake2Hasher>,
 		Block: BlockT<Hash=<Blake2Hasher as Hasher>::Out>,
 	{
-		let executor = executor.into().unwrap_or_else(|| executor::NativeExecutor::new(None));
+		let executor = executor.into().unwrap_or_else(|| {
```
bkchr:
```suggestion
		let executor = executor.into().unwrap_or_else(||
```

core/test-client/src/lib.rs
```diff
@@ -209,18 +209,20 @@ impl<E, Backend, G: GenesisInit> TestClientBuilder<
...
+			NativeExecutor::new(WasmExecutionMethod::Interpreted, None)
+		});
```
bkchr:
```suggestion
		);
```

core/executor/src/wasm_runtime.rs
```diff
@@ -0,0 +1,170 @@
...
+
+use crate::error::{Error, WasmError};
+use crate::wasmi_execution;
+use log::{trace, warn};
+use codec::Decode;
+use primitives::{storage::well_known_keys, Blake2Hasher, traits::Externalities};
+use runtime_version::RuntimeVersion;
+use std::{collections::hash_map::{Entry, HashMap}};
+
+/// The Wasm Substrate runtime.
```
bkchr:
```suggestion
/// The Substrate Wasm runtime.
```

core/executor/src/wasm_runtime.rs
```diff
@@ -0,0 +1,170 @@
...
+pub trait WasmRuntime {
+	/// Attempt to set the number of heap pages available during execution.
+	///
+	/// Returns false if the heap pages cannot be changed to the given value. Always returns true
```
bkchr:
```suggestion
	/// Returns `false` if the heap pages cannot be changed to the given value. Always returns `true`
```

bkchr: I also can not really follow. "Always returns true if the heap pages is set to the value already configured" and it returns `false` when it can not change the value. What does it return when it could change the value?

bkchr: Ahh, reading the implementation, I think that I understand the idea behind the function. I think it should be renamed. Maybe something like "heap_pages_equals" or so? As you just want to know that the heap pages are at the same value you expect.

core/executor/src/wasmi_execution.rs
```diff
@@ -0,0 +1,912 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Wasm interface module.
```
bkchr: Should be updated to reflect that this implements it using wasmi.

core/executor/src/wasmi_execution.rs
```diff
@@ -0,0 +1,912 @@
...
+	}
+}
+
+/// Call a given method in the given code.
+///
+/// Signature of this method needs to be `(I32, I32) -> I64`.
+///
+/// This should be used for tests only.
+#[cfg(test)]
+pub fn call(
```
bkchr: Can we not just switch to `create_instance().call` in tests and drop this function?

core/executor/src/wasm_runtime.rs
jimpo: Well, the reason for this is that some runtimes may be able to update the heap pages and some may not. Specifically, the heap pages is fixed with wasmi and updateable with wasmtime. But you could also consider that maybe it's possible for some engine to grow the heap pages but not shrink them.

So the point is that this attempts to set the heap pages to the new value and returns false if the update could not be performed. And it returns true if the heap pages would not change as a result of the update (this is guaranteed for all runtime implementations).

bkchr: Then something like `update_heap_pages`? Your description is much better than the current documentation of the function + name.

```
/// Try to update the heap pages.
///
/// Returns `true` if the update could be applied successfully or `false` otherwise.
fn update_heap_pages(heap_pages: usize);
```

Maybe something like this?

PR #3679, opened 2019-09-24
Document how to get head of linked map.

kianenigma:
and export `Linkage` which can very well used by someone writing an RPC client.

### DIFF
--- a/srml/support/procedural/src/lib.rs
+++ b/srml/support/procedural/src/lib.rs
@@ -85,7 +85,7 @@ use proc_macro::TokenStream;
 ///    previous and the next element.
 ///
 ///    Each linked map has a special key in the tire that can be queried and it can be particularly
-///    useful to query all items: `$hash(head of module_name ++ " " ++ storage_name`). This will
+///    useful to query all items: `"head of  " ++ module_name ++ " " ++ storage_name`). This will
 ///    return a key `K` that can be later on used with the aforementioned `$hash(module_name ++ " "
 ///    ++ storage_name ++ encoding(K))` to fetch the head. For consequent elements, the
 ///    [`Linkage`](../srml_support/storage/generator/struct.Linkage.html) can be used.
--- a/srml/support/procedural/src/lib.rs
+++ b/srml/support/procedural/src/lib.rs
@@ -84,7 +84,7 @@ use proc_macro::TokenStream;
 ///    [`Linkage`](../srml_support/storage/generator/struct.Linkage.html) is a pointer to the
 ///    previous and the next element.
 ///
-///    Each linked map has a special key in the tire that can be queried and it can be particularly
+///    Each linked map has a special key in the trie that can be queried and it can be particularly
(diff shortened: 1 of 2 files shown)


srml/support/procedural/src/lib.rs
```diff
@@ -79,6 +79,17 @@ use proc_macro::TokenStream;
 ///   If the keys are not trusted (e.g. can be set by a user), a cryptographic `hasher` such as
 ///   `blake2_256` must be used. Otherwise, other values in storage can be compromised.
 ///
+///    Once read from the trie (e.g. through an RPC call), a tuple of `(T, Linkage<K>)` is returned,
+///    where `T` is the value and `K` is the key type of the mapping.
+///    [`Linkage`](../srml_support/storage/generator/struct.Linkage.html) is a pointer to the
+///    previous and the next element.
+///
+///    Each linked map has a special key in the tire that can be queried and it can be particularly
```
thiolliere:
```suggestion
///    Each linked map has a special key in the trie that can be queried and it can be particularly
```

srml/support/procedural/src/lib.rs
```diff
@@ -79,6 +79,17 @@ use proc_macro::TokenStream;
...
+///    useful to query all items: `$hash(head of module_name ++ " " ++ storage_name`). This will
```
thiolliere: actually this is not hashed but maybe this is an issue ? like someone could target this if module_name nad storage_name are short ?
```suggestion
///    useful to query all items: `"head of  " ++ module_name ++ " " ++ storage_name`). This will
```

thiolliere:
Oh the fact that the head is not hashed has been introduced by my PR refactor this is an issue.

srml/support/procedural/src/lib.rs
```diff
@@ -79,6 +79,17 @@ use proc_macro::TokenStream;
 ///   If the keys are not trusted (e.g. can be set by a user), a cryptographic `hasher` such as
 ///   `blake2_256` must be used. Otherwise, other values in storage can be compromised.
 ///
+///    Once read from the trie (e.g. through an RPC call), a tuple of `(T, Linkage<K>)` is returned,
```
bkchr: First, I would not mention the trie at all, we are just talking about storage here.
Second, I'm not sure this is the right place. Here it should be documented what kind of syntax the macro supports and not how the data is encoded in the storage. This should probably be moved above the trait.

srml/support/procedural/src/lib.rs
kianenigma: According to the current code it is not hashed and I wrote it by mistake in the doc 👍

srml/support/procedural/src/lib.rs
kianenigma: The primary reason that I placed it here is that this page is --in some way-- the only aggregated, well document crate that we have about storage (you also sent it to me when I had some questions). I assumed anyone wanting to know this would have a high chance of finding it here. 

Otherwise, I agree that the context of the new paragraph is different.

PR #3680, opened 2019-09-24
Fix head not hashed

thiolliere:
in https://github.com/paritytech/substrate/pull/3344 I wrongly change the specification of linkedmap
before the key of the head of the linkedmap was stored at `$hash("head of " ++ module_name ++ " " ++ storoagename)` and after it was stored at `"head of " ++ module_name ++ " " ++ storoagename`.

Fix PR Fixes it.

### DIFF
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -84,7 +84,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// and set impl_version to equal spec_version. If only runtime
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
-	spec_version: 162,
+	spec_version: 163,
 	impl_version: 163,
 	apis: RUNTIME_API_VERSIONS,
 };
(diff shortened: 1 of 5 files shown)


thiolliere:
you can see the old code here https://github.com/paritytech/substrate/pull/3344/files#diff-33d847aa282ef513a046d0fa420fbaccL587-L589

thiolliere:
at least linked_map is used at:
```
./srml/elections-phragmen/src/lib.rs:151:		pub VotesOf get(votes_of): linked_map T::AccountId => Vec<T::AccountId>;
./srml/democracy/src/lib.rs:293:		pub Delegations get(delegations): linked_map T::AccountId => (T::AccountId, Conviction);
./srml/staking/src/lib.rs:574:		pub Validators get(validators): linked_map T::AccountId => ValidatorPrefs<BalanceOf<T>>;
./srml/staking/src/lib.rs:577:		pub Nominators get(nominators): linked_map T::AccountId => Vec<T::AccountId>;
```

but not in polkadot repo*

thiolliere:
in case we can't make it in this is the consequences:
* for each linkedmap the head of the linkedmap is stored at `"head of "++$module_name++" "++$storagename` this is with substrate modules of length > 28 byte which is probably ok
* making a transition might be doable but might be difficult

PR #3682, opened 2019-09-24
Change `decl_storage!` to forbid default value for `Option` types

### LINKED ISSUE #3671: `decl_storage!` should forbid default value for `Option` types
```
decl_storage! {
	trait Store for Module<T: Trait> as SomeModule {
		pub Value : Option<bool> = None;
	}
}
```

This should compile to an error saying that `= None` is not supported. Having a default value for optional types makes no sense, so we should just forbid it. The default value, if not set at genesis, should always be `None`.

Estimated time frame to complete: 3-5hours.

expenses:
Closes #3671. 

Something like:

```rust
decl_storage! {
	trait Store for Module<T: Trait> as SomeModule {
		pub Value : Option<bool> = None;
	}
}
```

Should result in an error like:

```
error: Default values for Option types are not supported
   --> test.rs:3:1
    |
3   |             pub Value : Option<bool> = None;
    |                                      ^^^^^^
```


xlc:
FYI you can use `Closes #3671` or `Fixes #3671` in the PR description to automatically close underlying issue when PR is merged.

expenses:
Closes #3671.

expenses:
@xlc thanks :^)

PR #3688, opened 2019-09-25
Add an Error type to Aura

expenses:
See #2688. Currently, Aura (and other consensus crates, I'll get round to those later) use `String`s for error handling. This PR replaces those `String`s with an `Error`. Currently, these errors have to be cast to `String`s for the `Verifier` trait: https://github.com/paritytech/substrate/blob/759ecfcaad312a8325b135761f7b4720dda96bc0/core/consensus/common/src/import_queue.rs#L72-L83

This is something that we should look into changing.

### DIFF
--- a/core/consensus/aura/src/lib.rs
+++ b/core/consensus/aura/src/lib.rs
@@ -48,7 +48,7 @@ use sr_primitives::{generic::{BlockId, OpaqueDigestItemId}, Justification};
 use sr_primitives::traits::{Block as BlockT, Header, DigestItemFor, ProvideRuntimeApi, Zero, Member};
 
 use primitives::crypto::Pair;
-use inherents::{InherentDataProviders, InherentData};
+use inherents::{InherentDataProviders, InherentData, RuntimeString};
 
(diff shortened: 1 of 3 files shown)


core/consensus/aura/src/lib.rs
```diff
@@ -308,16 +308,33 @@ impl<H, B: BlockT, C, E, I, P, Error, SO> SlotWorker<B> for AuraWorker<C, E, I,
...
-		}
-	};
+fn aura_err<B: BlockT>(error: Error<B>) -> Error<B> {
+	debug!(target: "aura", "{}", error);
+	error
+}
+
+#[derive(derive_more::Display)]
+enum Error<B: BlockT> {
+	#[display(fmt = "Multiple AuRa pre-runtime headers")]
```
kianenigma: nit but I would name it either `AuRa` OR `Aura` everywhere, not both. (no clue which one is correct tho' :p)

expenses: yeah, good idea!

PR #3690, opened 2019-09-25
Fix linkedmap instance head of

thiolliere:
Currently the head of a linked map for an instantiable module is put at: "Instance0head of MyModule MyStorage" instead of "head of Instance0MyModule MyStorage"

from the metadata user have the module name `Instance0MyModule` and the storage name `MyStorage` so he can't really get the head of the linkedmap.

I think this is a bug. fortunately there is no linked_map in instantiable module in substrate nor polkadot.

# Done

* fix the head key for instantiable module
* make test to test instantiable module with default instance and instance2

### DIFF
--- a/srml/support/procedural/src/storage/transformation.rs
+++ b/srml/support/procedural/src/storage/transformation.rs
@@ -600,15 +600,15 @@ fn create_and_impl_instance(
 	instance_prefix: &str,
 	ident: &Ident,
 	doc: &TokenStream2,
-	const_names: &[(Ident, String)],
+	const_names: &[(Ident, String, String)],
 	scrate: &TokenStream2,
 	instantiable: &Ident,
 	cratename: &Ident,
 ) -> TokenStream2 {
 	let mut const_impls = TokenStream2::new();
 
-	for (const_name, partial_const_value) in const_names {
-		let const_value = format!("{}{}", instance_prefix, partial_const_value);
+	for (const_name, const_value_prefix, const_value_suffix) in const_names {
+		let const_value = format!("{}{}{}", const_value_prefix, instance_prefix, const_value_suffix);
 		const_impls.extend(quote! {
 			const #const_name: &'static str = #const_value;
 		});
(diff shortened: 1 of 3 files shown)


srml/support/test/tests/final_keys.rs
```diff
@@ -14,87 +14,211 @@
...
-use codec::{Encode, Decode};
+use codec::Encode;
+use support::{StorageDoubleMap, StorageLinkedMap, StorageMap, StorageValue};
 
-pub trait Trait {
-	type Origin;
-	type BlockNumber: Encode + Decode + Default + Clone;
-}
+mod no_instance {
+	use codec::{Encode, Decode};
```
tomusdrw: The indentation here is a bit weird, should just be a regular tab?

srml/support/test/tests/final_keys.rs
```diff
@@ -14,87 +14,211 @@
...
+			config(phantom): core::marker::PhantomData<I>;
+		}
+	}
 }
 
-fn new_test_ext() -> runtime_io::TestExternalities<Blake2Hasher> {
-	GenesisConfig::<Test>::default().build_storage().unwrap().into()
+#[test]
+fn final_keys_no_instance() {
+	with_storage(&mut Default::default(), || {
```
tomusdrw: same here and in the `instance` module.

srml/support/test/tests/final_keys.rs
thiolliere: I don't get it, this is a regular tab

thiolliere: oh I think it is related to github way to render it. maybe they count the "+ " into the tab here and this end into this. weird indeed, but in the code this is a tab

tomusdrw: Ah, fair enough :)

srml/support/procedural/src/storage/transformation.rs
```diff
@@ -600,15 +600,15 @@ fn create_and_impl_instance(
...
+	const_names: &[(Ident, String, String)],
 	scrate: &TokenStream2,
 	instantiable: &Ident,
 	cratename: &Ident,
 ) -> TokenStream2 {
 	let mut const_impls = TokenStream2::new();
 
-	for (const_name, partial_const_value) in const_names {
-		let const_value = format!("{}{}", instance_prefix, partial_const_value);
+	for (const_name, const_value_prefix, const_value_suffix) in const_names {
```
bkchr: Conceptually this makes no difference or? 
Why did you make this change?

thiolliere: this makes a difference: for head of linkedmap the implementation is now `"Instance0head of MyModule MyStorage"`   instead of `"head of Instance0MyModule MyStorage"`

thiolliere: this test failed before this PR: https://github.com/paritytech/substrate/pull/3690/files#diff-c3e43d84ecf048969752108967ffa7c6L76

bkchr:
Needs resolving.

srml/support/procedural/src/storage/transformation.rs
bkchr: Ahh, just read your comment that we currently don't use it.

PR #3692, opened 2019-09-25
ChainSpec extensions

### LINKED ISSUE #3605: Chain-spec: non-state and consensus parameters
At the moment, the entire chain-spec is generated based on what information the genesis state needs in order to run. 

We should have the option to have key-value pairs in groups that are used by other parts of the runtime environment (in the case that genesis-state configuration was not enough and needs to be overridden somewhere by a hard-fork).

Also there should be a means of hardcoding certain block number and hash pairs into the chain-spec, so that the runtime environment will refuse to import any block with different hash at that number.

tomusdrw:
Resolves #3605

This PR introduces a way to customize `ChainSpec` (and chain spec files), with additional data that is used as configuration for other parts of Substrate node.

The parameters may (but don't have to) be "forkable", which means that the actual set of parameters depends on the current block number.

It introduces a bunch of changes to how `ChainSpec` json is being parsed to enforce that we validate the entire content and give meaningful error messages if something is missing. The extensions are also strictly typed, so there is no way to add misspelled or invalid parameter. For this to work I had to introduce a custom derive, which implements a mirror struct with all the fields being optional and additional custom derive which allows composition of strictly typed parameters be passed as configuration without adding too much generics in the code (see `Extension::get::<T>()`).

As an example, I've implemented the requested mechanism of rejecting blocks on specific height which do not match the hash provided in the spec.

I think we could now move parameters like `TelemetryEndpoints`, `ProtocolId` and `Properties` to be just extensions - can do that in this or follow up PR.

The whole scheme ended up being a bit more complicated than I hoped for so I'm also ok with rejecting this PR (or some parts of it) and going with more relaxed and loosly-typed solution (i.e. just having an arbitrary `json::Map<String, json::Value>` and letting the modules do the parsing.


core/chain-spec/src/extension.rs
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ChainSpecExtension)]
+	#[serde(deny_unknown_fields)]
+	pub struct Ext2 {
+		#[serde(flatten)]
+		ext1: Extension1,
+		#[forks]
+		forkable: Forks<u64, Extensions>,
+	}
```
thiolliere: I'm not sure I see the whole need for this derive macros, if it is only for constraining inputs this could be done with a simple function declared as such:
```rust
constrain! {
    non_forkable {
        ext1: Extension1,
    },
    forkable {
        ext1: Extension1,
        ext2: Extension2,
    },
}
```

With constrain declare as a proc macro or macro_rules, and it just check that input has one and only one field for each non_forkable and same for forkable, plus possible forks with at most one field for each forkable.

Do I miss some feature here ?

tomusdrw: `ChainSpecGroup` implements quite a lot of stuff, namely for struct:
```rust
struct Example {
 a: u64,
 b: u64,
}
```
it implements
```rust
struct ExampleFork {
 a: Option<u64>,
 b: Option<u64>,
}
```
and a bunch of traits to tie this too together. This is needed since you might want to hard fork only one of the parameters and leave previous one at previous values (see a test here: https://github.com/paritytech/substrate/blob/3171fc5063f9f32bb564da4076d42375fa2d03fb/core/chain-spec/src/extension.rs#L304)

The `ChainSpecExtension` is a bit different as it implements `Extension` trait, which has two methods:
- `get` which allows you to get a parameter of specific type (no mater how it's named in the struct)
- `forks` which allows to get `Forks` for specific type.
The `#[forks]` attribute is needed to figure out which field holds the forks.

But back to your proposal:
1. I think there is still a need to have `ChainSpecGroup` macro, to generate fork structs for `Extension1` and `Extension2`.
2. We still need to generate intermediate `struct Extension { ext1: Extension1, ext2: Extension2 }` to (de)serialize to JSON.
3. With current trait structure we still need to implement `Extension` trait for it.

So I find the derive macro a bit more readable than your proposal, unless we could get rid of some additional things (like some traits).

thiolliere: oh Fork are used for nested structure as well. then ok it make sense.
Example could be a field of another structure here.

core/consensus/common/src/block_import.rs
```diff
@@ -193,9 +194,10 @@ impl<B: BlockT> BlockImport<B> for crate::import_queue::BoxBlockImport<B> {
 	fn check_block(
 		&mut self,
 		hash: B::Hash,
+		number: NumberFor<B>,
```
DemiMarie-parity: This should probably be a struct, since it is being passed around so much.

gavofyork:
some tests failing and there's a fat line in there?

tomusdrw:
Updated.

xlc:
Are there any reasons to use `Properties` over typed ChainSpec extensions?

tomusdrw:
@xlc `Properties` are exposed over RPC (typed extensions currently are not). I imagine that `Properties` will still be used as hints for the UIs, while typed extensions will be used within the codebase.
Other than that, if you don't need strict typing and proper chain spec validation (i.e. doing some PoC), you might still be fine to just `Properties`.

JoshOrndorff:
Is there a reason that the `ChainSpecGroup` derive macro does not support Enums? I would like to use it on an enum like this:

```rust
/// A chain spec extension for a chain that starts with no real consensus rules (manual seal) but later adds a PoW requirement.
#[derive(Clone, Debug, Deserialize, Serialize, ChainSpecGroup)]
pub enum HotSwapConsensus {
	/// No real consensus rules enforced. This allows manual seal.
	None,
	/// Proof of Work seal required with block hash below given difficulty threshold.
	Pow{ difficulty: u64 },
	/// Proof of Authority seal required by on of the authorities specified.
	Poa{ authorities: Vec<Authority> },
}
```

The idea is that we can start a chain with manual seal, then hard fork to PoW, then hard fork more times to change the difficulty threshold, then finally fork to PoA.

JoshOrndorff:
I'm also struggling because the `ChainSpecGroup` trait is not implemented for basic types like `Vec<_>` or even tuples. Am I misunderstanding the intended use of this? Or is it just not widely used and nobody noticed these limitations?

bkchr:
AFAIK it was never used. Looking at the code I also don't really get how it should work.

```
/// A chain spec extension for a chain that starts with no real consensus rules (manual seal) 
/// but later adds a PoW requirement, and then later a Poa requirement.
#[derive(Clone, Debug, Deserialize, Serialize)]
pub enum HotSwapConsensus {
	/// No real consensus rules enforced. This allows manual seal.
	None,
	/// Proof of Work seal required with block hash below given difficulty threshold.
	Pow{ difficulty: u64 },
	/// Proof of Authority seal required by one of the authorities specified.
	Poa{ authorities: Vec<Authority> },
}

#[derive(Clone, Debug, Deserialize, Serialize, ChainSpecGroup)]
struct Consensus {
    consensus: HotSwapConsensus
}
```

Doing something like this should work. However, I would let the runtime signal the consensus change. Not sure why you want to decide this on the node side.

JoshOrndorff:
> Not sure why you want to decide this on the node side.

This is for educational purposes (specifically the academy). I want to demonstrate that the traditional/legacy way to upgrade a network is through hard forks and show how it is hard and messy to coordinate. I'm trying to motivate why forkless upgrades are good and what problem they solve and demonstrate that later. I agree using the runtime is ultimately better.

Anyway, thanks for taking a look.

JoshOrndorff:
The idea is that node operators can sit together in the classroom and discuss what block numbers and difficulties to fork to. And then just edit their spec file. Of course the node will need to support all the underlying consensuses in advance. The other win is that some "sects" will form that don't want to hard fork and then we end up with diverging chains with a common prefix like bitcoin cash or eth classic. And we could demonstrate all this without recompiling. But in the end recompiling isn't such a big deal, so I will just have a few releases prepared in advance and achieve the bitcoin cash effect by encouraging some students to not upgrade in time.

PR #3693, opened 2019-09-25
Fix Phragmen equalize loop.

kianenigma:
Apparently I assumed at the time return from closure is the same as break 🤦🏻‍♂️


bkchr:
@kianenigma needs runtime bump

bkchr:
Yep

kianenigma:
darn.

PR #3694, opened 2019-09-25
Random seed API for offchain workers

### LINKED ISSUE #3687: Random seed API for offchain workers
Related #1458

One of the few missing APIs of the initial set for offchain workers is `random_seed`: See the code here:
https://github.com/paritytech/substrate/blob/58da1c760ef00ff44875d6cada668c31313b3059/core/offchain/src/api.rs#L104

The 32-byte seed should be returned from a strong entropy random source (like `OsRng`) and returned to the caller.
The implementation obviously involves some test for that API.

expenses:
Closes #3687.

### DIFF
--- a/core/offchain/src/api.rs
+++ b/core/offchain/src/api.rs
@@ -32,7 +32,6 @@ use primitives::offchain::{
 };
 use sr_primitives::{generic::BlockId, traits::{self, Extrinsic}};
 use transaction_pool::txpool::{Pool, ChainApi};
-use rand::{Rng, rngs::OsRng};
 
 mod http;
 mod timestamp;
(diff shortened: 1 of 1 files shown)


core/offchain/src/api.rs
```diff
@@ -101,7 +103,7 @@ where
 	}
 
 	fn random_seed(&mut self) -> [u8; 32] {
-		unavailable_yet("random_seed")
+		self.rng.gen()
```
bkchr: I think [thread_rng](https://docs.rs/rand/0.7.2/rand/fn.thread_rng.html) should be enough.

PR #3698, opened 2019-09-26
srml/authority-discovery: Abstract session key type

mxinden:
Previously `srml/authority-discovery` depended on the `srml/im-online`
session key type directly. With this patch `srml/authority-discovery` is
generic over the session key type it is going to use, as long as it
implements the RuntimeAppPublic trait.

With this patch one can use the `srml/authority-discovery` module
without the `srml/im-online` module.

Next to the above, this patch configures `node/runtime` to use the babe
session keys for the authority discovery module.


srml/authority-discovery/Cargo.toml
```diff
@@ -5,16 +5,16 @@ authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
 [dependencies]
+app-crypto = { package = "substrate-application-crypto", path = "../../core/application-crypto", default-features = false }
+babe-primitives = { package = "substrate-consensus-babe-primitives", path = "../../core/consensus/babe/primitives", default-features = false }
```
bkchr: This needs to go into `dev-dependencies`. You don't need it for the normal build.

mxinden:
@bkchr thanks for the review! Can you take another look?

srml/authority-discovery/src/lib.rs
```diff
@@ -22,29 +22,28 @@
...
 use rstd::prelude::*;
 use support::{decl_module, decl_storage};
 
-pub trait Trait: system::Trait + session::Trait + im_online::Trait {}
-
-type AuthorityIdFor<T> = <T as im_online::Trait>::AuthorityId;
-type AuthoritySignatureFor<T> =
-	<<T as im_online::Trait>::AuthorityId as RuntimeAppPublic>::Signature;
+pub trait Trait: system::Trait + session::Trait {
+	type AuthorityId: RuntimeAppPublic + Default + Decode + Encode + PartialEq;
```
rphmeier: needs docs

node/runtime/src/lib.rs
```diff
@@ -85,7 +85,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 165,
```
andresilva: Shouldn't the spec version be increased (and not impl version)?

mxinden: @andresilva do I understand correctly that we increase the `spec_version` whenever we have breaking changes visible only outside of the runtime boundary?

This patch set does not touch the `core/authority-discovery/primitives::AuthorityDiscoveryApi`, thus as far as I can tell it is not visible outside of the runtime boundary. Am I missing something?

rphmeier:
can merge when passing

node/runtime/src/lib.rs
bkchr: Given that we switch to `Babe` as authority id, @andresilva is right. You change the behavior of runtime and this new runtime would not be compatible with the old runtime when it comes to authority discovery.

andresilva: My vague definition for when to increment `spec_version` is a bit stricter than that, since internal behavior of the module can change without any changes being introduced at the boundaries. Keeping the same `spec_version` means that a previous compiled blob of the runtime without these changes can be executed as is.

PR #3699, opened 2019-09-26
Split off System random functions into a new Randomness module

### LINKED ISSUE #1451: Split SafeMix and random_seed out of System module.
It's also an open question as to whether the `Core` API should require a randomness beacon as it's not required for all consensus algorithms.

expenses:
This closes #1451.


node/runtime/src/lib.rs
```diff
@@ -487,6 +487,7 @@ construct_runtime!(
 		NodeBlock = node_primitives::Block,
 		UncheckedExtrinsic = UncheckedExtrinsic
 	{
+		Randomness: randomness::{Module, Call, Storage},
```
xlc: new module should be added to end to avoid breaking the module index, which is used to construct transactions

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+	trait Store for Module<T: Trait> as System {
+		/// Series of block headers from the last 81 blocks that acts as random seed material. This is arranged as a
+		/// ring buffer with the `i8` prefix being the index into the `Vec` of the oldest hash.
+		RandomMaterial get(random_material): (i8, Vec<T::Hash>);
+	}
+}
+
+impl<T: Trait> Module<T> {
+	/// Initialize the random material for a particular block.
+	pub fn initialize(parent_hash: &T::Hash) {
```
xlc: this code is not called anywhere. you will need to implement `on_initialize` in `decl_module` macro (search it in code base for examples) and call this.

srml/randomness/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "srml-randomness"
+version = "1.0.0"
```
kianenigma:
```suggestion
version = "2.0.0"
```

(not sure but we bumped all up after substrate 0.1 release)

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,145 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```
kianenigma:
```suggestion
// Copyright 2019 Parity Technologies (UK) Ltd.
```

srml/system/src/lib.rs
```diff
@@ -77,7 +77,6 @@
 //! 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
 //! 		pub fn system_module_example(origin) -> Result {
 //! 			let _sender = ensure_signed(origin)?;
-//! 			let _random_seed = <system::Module<T>>::random_seed();
```
kianenigma: This example now makes no sense. It should be reworked. 

Either use something else from `System` to demonstrate something that it can do (easier) or `pub trait Trati: system::Trait + randomness::Tratit` and do the same.

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+
+use rstd::prelude::*;
+use sr_primitives::traits::Hash;
+use support::{decl_module, decl_storage};
+use safe_mix::TripletMix;
+use codec::Encode;
+use system::Trait;
+
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
```
kianenigma: not a question for you but a general note about the macros that: it would have been nice if we could drop the `where origin: T::Origin` when it is not being used anywhere.

bkchr: Not related to this pr.

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+		fn on_initialize() {
+			let parent_hash = <system::Module<T>>::parent_hash();
+
+			<RandomMaterial<T>>::mutate(|&mut(ref mut index, ref mut values)| if values.len() < 81 {
+				values.push(parent_hash)
+			} else {
+				values[*index as usize] = parent_hash;
+				*index = (*index + 1) % 81;
```
bkchr: I know that you just copied it, but isn't the `index == block_number % 81`?

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+	///   reduces the ability of a cabal of block producers from conspiring
+	///   against individuals.
+	///
+	/// WARNING: Hashing the result of this function will remove any
+	/// low-influence properties it has and mean that all bits of the resulting
+	/// value are entirely manipulatable by the author of the parent block, who
+	/// can determine the value of `parent_hash`.
+	pub fn random(subject: &[u8]) -> T::Hash {
+		let (index, hash_series) = <RandomMaterial<T>>::get();
+		if hash_series.len() > 0 {
```
bkchr:
```suggestion
		if !hash_series.is_empty() {
```

srml/randomness/src/lib.rs
expenses: Yep, I'll change that.

srml/randomness/src/lib.rs
kianenigma: > not a question for you, but ..

Yeah -- just a note.

srml/randomness/src/lib.rs
expenses: actually, it's just either to leave it as-is:
```
  --> srml/randomness/src/lib.rs:68:29
   |
68 |                 *index = block_number % 81;
   |                                         ^^ expected associated type, found integer
   |
   = note: expected type `<T as srml_system::Trait>::BlockNumber`
              found type `{integer}`
```

bkchr: `81.into()`

bkchr: And I meant to not store this index anymore.

expenses: Oh ok, that works.

expenses: Hmm, `T::BlockNumber` implementing `TryInto<usize>` but not `Into<usize>` is a bit of a problem if we're trying to avoid panics. I could use `unique_saturating_into` though which casts to `u128` and then cast that to `usize`. What do you suggest?

bkchr: Just use expect. You can prove it statically that `something % 81` always generates a number that is between `0..81` and that is way below `usize::max_value()`.

bkchr:
Could you please add some sort of tests?

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! # Randomness Module
+//!
+//! The Randomness module provides a [`random`](./struct.Module.html#method.random) function that
+//! generates low-influence random values based on the block hashes from the previous 81 blocks.
```
burdges: "Low-influence randomness can be useful when defending against a specific adversary of known strength."

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+	///
+	/// This randomness uses a low-influence function, drawing upon the block
+	/// hashes from the previous 81 blocks. Its result for any given subject
+	/// will be known in advance by the block producer of this block (and,
+	/// indeed, anyone who knows the block's `parent_hash`). However, it is
+	/// mostly impossible for the producer of this block *alone* to influence
+	/// the value of this hash. A sizable minority of dishonest and coordinating
+	/// block producers would be required in order to affect this value. If that
+	/// is an insufficient security guarantee then two things can be used to
+	/// improve this randomness:
```
burdges: ".. will be known far in advance by anyone observing the chain."

Replace "However, ..." with "Any block producer has significant influence over their block hashes bounded only by their computational resources.  Our low-influence function reduces the actual block producer's influence over the randomness, but increases the influence of small colluding groups of recent block producers."

"Some BABE blocks have VRF outputs where the block producer has exactly one bit of influence, either they make the block or they do not make the block and thus someone else makes the next block.  Yet, this randomness is not fresh in all BABE blocks."

expenses:
One check isn't passing:
```
Hash:
    given=3e51b47b6cc8449eece93eee4b01f03b00a0ca7981c0b6c0447b6e0d50ca886d,
    expected=a6378d7fdd31029d13718d54bdff10a370e75cc624aaf94a90e7e7d4a24e0bcc
thread 'tests::block_import_works' panicked at
'Storage root must match that calculated.', srml/executive/src/lib.rs:274:3
```

https://github.com/paritytech/substrate/blob/e84f6158b383880ba5fc86cb62c9fea10f5f3071/srml/executive/src/lib.rs#L252-L272

Not entirely sure what's going on here.

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+		});
+	}
+
+	#[test]
+	fn test_random() {
+		with_externalities(&mut new_test_ext(), || {
+			let mut parent_hash = System::parent_hash();
+
+			for i in 2..164 {
+				System::initialize(&i, &parent_hash, &Default::default(), &Default::default());
```
kianenigma: suggestion: maybe a utility function that does this repeated stuff in the loop?

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+				Randomness::on_initialize(i);
+
+				let header = System::finalize();
+				parent_hash = header.hash();
+				// Why is this needed?
+				System::set_block_number(*header.number());
```
kianenigma: because the last thing that you call is `finalize()` which `kill`s `system::Number<T>`

rphmeier:
bikeshed: i wouldn't call this `randomness`, since it's really only one mode of randomness. actually, most of the time we would use the BABE randomness wherever available.

maybe `randomness-triplet-mix` or something?

expenses:
both of those would do the trick, but _ideally_ I'd like to have something more concise.

srml/randomness/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+safe-mix = { version = "1.0", default-features = false }
+codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false, features = ["derive"] }
+sr-primitives = { path = "../../core/sr-primitives", default-features = false }
+support = { package = "srml-support", path = "../support", default-features = false }
+system = { package = "srml-system", path = "../system", default-features = false }
+rstd = { package = "sr-std", path = "../../core/sr-std", default-features = false }
+
+[dev-dependencies]
+primitives = { package = "substrate-primitives",  path = "../../core/primitives" }
+runtime-io ={ package = "sr-io", path = "../../core/sr-io" }
```
gavofyork:
```suggestion
runtime-io = { package = "sr-io", path = "../../core/sr-io" }
```

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use rstd::{prelude::*, convert::TryInto};
+use sr_primitives::traits::Hash;
+use support::{decl_module, decl_storage};
+use safe_mix::TripletMix;
+use codec::Encode;
+use system::Trait;
+
+
```
bkchr:
```suggestion
```

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+fn block_number_to_index<T: Trait>(block_number: T::BlockNumber) -> usize {
+	// '- 2' because on_initialize is called on the first block after genesis - block 2
+	let index = (block_number - 2.into()) % 81.into();
+	index.try_into().ok().expect("something % 81 should always be less than usize")
```
bkchr:
```suggestion
	index.try_into().ok().expect("something % 81 is always smaller than usize; qed")
```

srml/randomness/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+use sr_primitives::traits::Hash;
+use support::{decl_module, decl_storage};
+use safe_mix::TripletMix;
+use codec::Encode;
+use system::Trait;
+
+
+fn block_number_to_index<T: Trait>(block_number: T::BlockNumber) -> usize {
+	// '- 2' because on_initialize is called on the first block after genesis - block 2
+	let index = (block_number - 2.into()) % 81.into();
```
bkchr: Hmm, I think that the genesis number is `0`.

expenses: I was basing my tests off of: https://github.com/paritytech/substrate/blob/094cca7b7726ef407cd82e65efe4ab42fff33801/srml/finality-tracker/src/lib.rs#L350-L368

which made me assume that it started at 1. I'll change this.

rphmeier:
LGTM after a crate rename - it shouldn't be called `randomness` unless it's the only / forever guaranteed / safest way to do randomness. which it is none of

expenses:
@rphmeier I've renamed the crate. Should documentation such as
```rust
//! # Randomness Module
//!
//! The Randomness module provides a [`random`](./struct.Module.html#method.random) function that
//! generates low-influence random values based on the block hashes from the previous 81 blocks.
//! Low-influence randomness can be useful when defending against relatively weak adversaries.
```
be renamed too?

kianenigma:
> @rphmeier I've renamed the crate. Should documentation such as
> 
> ```rust
> //! # Randomness Module
> //!
> //! The Randomness module provides a [`random`](./struct.Module.html#method.random) function that
> //! generates low-influence random values based on the block hashes from the previous 81 blocks.
> //! Low-influence randomness can be useful when defending against relatively weak adversaries.
> ```
> 
> be renamed too?

None of by business but I assume yes. Make sure to also merge master before merging this to make sure all external usages of `randomness::randomness()` etc. are properly refactored

srml/randomness-collective-flip/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+use rstd::{prelude::*, convert::TryInto};
+use sr_primitives::traits::Hash;
+use support::{decl_module, decl_storage};
+use safe_mix::TripletMix;
+use codec::Encode;
+use system::Trait;
+
+fn block_number_to_index<T: Trait>(block_number: T::BlockNumber) -> usize {
+	// on_initialize is called on the first block after genesis
+	let index = (block_number - 1.into()) % 81.into();
```
bkchr: Could we please make the `81` a file local constant?

srml/randomness-collective-flip/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+//!
+//! ```
+//! use support::{decl_module, dispatch::Result};
+//!
+//! pub trait Trait: system::Trait {}
+//!
+//! decl_module! {
+//! 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+//! 		pub fn random_module_example(origin) -> Result {
+//! 			let _random_seed = <srml_randomness::Module<T>>::random_seed();
```
bkchr: `srml_randomness` is not correct anymore here.

srml/randomness-collective-flip/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+			} else {
+				let index = block_number_to_index::<T>(block_number);
+				values[index] = parent_hash;
+			});
+		}
+	}
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as System {
```
bkchr:
```suggestion
	trait Store for Module<T: Trait> as RandomnessCollectiveFlip {
```

srml/randomness-collective-flip/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! # Randomness Module
+//!
+//! The Randomness Collective Flip module provides a [`random`](./struct.Module.html#method.random)
+//! function that generates low-influence random values based on the block hashes from the previous
+//! 81 blocks. Low-influence randomness can be useful when defending against relatively weak
```
bkchr:
```suggestion
//! `81` blocks. Low-influence randomness can be useful when defending against relatively weak
```

shawntabrizi:
@expenses Nice!

PR #3701, opened 2019-09-26
create some vectors with initial capacities

bnjjj:
Just some improvements about Vec initialization found when reading your codebase :)


bnjjj:
[clabot:check]

core/sr-primitives/src/generic/mod.rs
```diff
@@ -52,7 +52,7 @@ fn encode_with_vec_prefix<T: Encode, F: Fn(&mut Vec<u8>)>(encoder: F) -> Vec<u8>
 
 	// need to prefix with the total length to ensure it's binary compatible with
 	// Vec<u8>.
-	let mut length: Vec<()> = Vec::new();
+	let mut length: Vec<()> = Vec::with_capacity(v.len() - reserve);
```
tomusdrw: That doesn't seem necessary, we call resize in the next line anyway.

tomusdrw:
Needs `impl_version` bump in `node/runtime/src/lib.rs`

core/sr-primitives/src/generic/mod.rs
bnjjj: I just follow the same pattern as few lines above https://github.com/paritytech/substrate/pull/3701/files/6eb3dda7ab5c1f505fe70de93548bce4358dec7d#diff-de8946ac7f8c2ef11110896a776c5f37L49 do you want that I change this one to `Vec::new` to have the same ?

tomusdrw: The other one is a bit different, since we change the capacity to `reserve + size`, but actually resize up to `reserve` only.
Here it's superfluous afaict.

bnjjj: Done, rollback to the previous version

core/sr-io/without_std.rs
```diff
@@ -737,8 +737,8 @@ impl StorageApi for () {
 	}
 
 	fn blake2_256_ordered_trie_root(input: Vec<Vec<u8>>) -> H256 {
-		let mut values = Vec::new();
-		let mut lengths = Vec::new();
+		let mut values = Vec::with_capacity(input.len());
```
bkchr: This is okay, but not correct.

PR #3702, opened 2019-09-26
Refactor srml-support/storage

thiolliere:
# Done:

* remove StorageDoubleMapWithHasher as now you can just use DoubleMap and specify hasher
* move hash structure to hash.rs module instead of storage/hashed.rs
* remove StoraveVec in storage/unhashed.rs because not used and not handled by decl_storage
* remove StorageVec in storage/hashed.rs because not used and not handled by decl_storage

This is easy refactor and must ease maintainance and further PR review


srml/support/src/hash.rs
```diff
@@ -48,3 +46,65 @@ impl<T: Codec> Hashable for T {
...
+	type Output = [u8; 32];
+	fn hash(x: &[u8]) -> [u8; 32] {
+		blake2_256(x)
+	}
+}
+
+/// Hash storage keys with twox 128
+pub struct Twox128;
+impl StorageHasher for Twox128 {
+	type Output = [u8; 16];
```
kianenigma: indentation seems smaller here or is it github rendering it like this?

srml/support/src/storage/storage_vec.rs
```diff
@@ -0,0 +1,140 @@
...
+	fn set_items<I, T>(items: I)
+		where
+			I: IntoIterator<Item=T>,
+			T: Borrow<Self::Item>,
+	{
+		let mut count: u32 = 0;
+
+		for i in items.into_iter() {
+			unhashed::put(&count.to_keyed_vec(Self::PREFIX), i.borrow());
+			count = count.checked_add(1).expect("exceeded runtime storage capacity");
```
kianenigma: a bit pedantic: I know that it has the same function, but isn't panic semantically correct here? I would use expect to be a proof that the panic cannot happen. something like: `expect("runtime storage limit cannot be reached")` (which is moot, hence panic is better IMO)

srml/support/src/storage/storage_vec_hashed.rs
```diff
@@ -0,0 +1,150 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
```
kianenigma: Can someone give a brief explanation of how we eneded up having two versions of these traits (hashed, unhashed)? I assume that it has a particular reason/bottleneck that we have this code duplication and I can't see the big picture now.

srml/support/src/hash.rs
bkchr: Is correct.

srml/support/src/hash.rs
```diff
@@ -48,3 +46,65 @@ impl<T: Codec> Hashable for T {
...
+		twox_64(x)
+			.into_iter()
+			.chain(x.into_iter())
+			.cloned()
+			.collect::<Vec<_>>()
+	}
+}
+
+#[test]
+fn test_twox_64_concat() {
```
bkchr: Please move to bottom in a `tests` module.

srml/support/src/storage/storage_vec.rs
```diff
@@ -0,0 +1,140 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Storage vec abstraction on top of runtime storage.
```
bkchr: WE don't  use this and the other one anyway. I think we should remove them. No reason to have them in the code base.

thiolliere: yes I agree I removed them

srml/support/src/storage/storage_vec_hashed.rs
thiolliere: I think they are here for historical reason

thiolliere:
so now storage vec are removed as they are not used and not part of decl_storage

Issue #1850, opened 2019-02-22
Bridge parachain and service


gavofyork:
This is essentially the "Parity Bridge" but for Substrate. The idea is to build a Substrate light-client SRML module that is able to track the Grandpa authority set of another Substrate chain (that runs Grandpa finality obviously). With this is can track headers. Each chain would keep a `DigestItem` (in the header) of the trie root of outgoing messages.

Validators on the "sub-chain" would track the "super-chain" and vice-versa. There would be a trie root (from the `DigestItem`s in the headers) for the outgoing message queue on each chain that is synced into the other chain along with the preimages (the bridge nodes would need to be full nodes in order to extract the preimages). This would allow both chains to send and receive messages (in the form of serialised `Call`s).

Since tracking a Grandpa set (and the associated headers) is computationally cheap, the super-chain would be able to handle a lot of bridged sub-chains.

wooparadog:
@gavofyork Hi Gavin. Is there a timeline this feature will be implemented? If not, could you give some tips how to write a sane one(between substrate chains)?

Thanks

burdges:
We noticed earlier this week that limiting the value transacted by a bridge improves security.  

We only have some limited number of validators v check each block, with μ of those being extra checks.  Among these, we have on average μ' of these extra checks occurring in the first tranche of the extra checks.  These expected μ' checks are completely unforeseeable by an adversary. 

I think the strongest rational attack goes roughly:  

An adversary who possesses over 1/3rd of the stake, makes a bad parachain block available to less than 1/3rd of the honest validators, and votes on it themselves, so that the block can reach finality without being available to 1/3rd of the honest validators.  In advance, the adversary has ensured that they control a full set of v checks early, so all parachain validators and a full set of μ in the first tranche of extra checks.  

In expectation, there are still μ'/3 honest extra checks in the in the first tranche, but the actual number k might be zero.  If k is non-zero, then the adversary must reveal their bad block in the time dictated by the validity scheme for extra checks.  If k is zero, then the adversary can reveal their bad block anytime they like.  In either case, they much eventually reveal they bad block and be slashed s*v where s is the stake per validator, but most of their nodes survive unscathed.  

An adversary who liked the k non-zero case could be slashed less than v, so bridges should not consider relay chain blocks final until they witness the "super-finality" of a full v extra checks.  We caution that "super-finality" is lost if an extra checker gets slashed, so our extra checks rules should add more extra checks whenever an extra checker gets slashed.

[...]

PR #3703, opened 2019-09-26
Implement Bridge Module

HCastano:
This PR is going to track the development of #1850. Development will be done on other branches and PR'd into this one. This is being done in order to not have half baked code in Substrate, as well as making it easier for others to review.

Some of the TODOs are:
- [ ] Have a module that can act as a light client for another chain (not necessarily do anything with that data though)
- [ ] Verify GRANDPA finality proofs on-chain
- [ ] Verify block ancestry proofs on-chain
- [ ] Have an RPC for actors to submit claims from other chains
- [ ] Message delivery and acknowledgement infrastructure
- [ ] Default message handler for the receiving end of the bridge

Additionally, a bridge relay node will need to be implemented, which will need to do the following:
- [ ] Relay finalized headers to bridged chains
- [ ] Relay outbound messages to bridged chains
- [ ] Relay message delivery acknowledgements to bridged chains

The bridge relay node issues could probably be tracked in a separate issue as they are fairly independent of the module itself.

cc @jimpo


srml/bridge/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+		pub Validators get(validators): Vec<T::ValidatorId>;
+	}
+}
+
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		// TODO: Figure out the proper type for these proofs
+		fn new(origin, block_header: T::Header, validator_set: Vec<T::ValidatorId>, validator_set_proof: Vec<u8>, storage_proof: Vec<u8>) {
+		//
+		// The macro can't handle arguments like this :(
```
andresilva: iirc you can break parameters on multiple lines but you cannot have the trailing comma on the last parameter.

kianenigma: exactly. And one should make an insubstantial PR for that

srml/bridge/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+		pub LastFinalizedBlock get(last_finalized_block): u64; // Block;
+
+		/// Get latest block number included in the chain
+		pub BlockNumber get(lastest_block_num): T::BlockNumber;
+
+		// What I want to have wrt to the Block are:
+		// 1. Block Number
+		// 2. Block Hash
+		// 3. State Root
+		// Can I get these from a `Block`, or do they need to
```
kianenigma: All of them should be given through system module. Question is, where do you want them and do you want them per block or just for the latest block?

kianenigma: - Block number is trivial and you can get it in `on_finalize()` or directly from `system::block_number`
- Block hashes are stored in a map for all blocks in system
- I am not sure about state root but it should be there somewhere :P System has a `type Header` this type has `traits::Header` which has a function that yields state root.

srml/bridge/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as Bridge {
+		/// Get the ID of the current bridge
+		pub BridgeId get(bridge_id): T::Hash;
+
+		/// Get the ID of the current bridge
+		// TODO: Figure out how to ge a Block here
+		pub LastFinalizedBlock get(last_finalized_block): u64; // Block;
```
kianenigma: is this any different from `system::block_number()`?

srml/bridge/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+		// The macro can't handle arguments like this :(
+		// fn new(
+		// 	origin,
+		// 	block_header: T::Header,
+		// 	validator_set: Vec<T::ValidatorId>,
+		// 	validator_set_proof: Vec<u8>,
+		// 	storage_proof: Vec<u8>,
+		// ) {
+
+			// NOTE: Should this be a root call?
```
kianenigma: not sure about sender but from the looks of it for sure it must be a singleton function; i.e. I think you don't want this to be called ever twice, right?

srml/bridge/src/lib.rs
kianenigma: https://github.com/paritytech/substrate/pull/3712

HCastano: Yep, you're right. It was the trailing comma that was breaking the macro.

srml/bridge/src/lib.rs
HCastano: The idea behind these is that the bridge will periodically, and not necessarily at a regular interval, update the info around the last finalized block it has seen.

Let's consider the case where we have chain A whose headers are being sent to chain B (bridge). Chain B would just want to keep track of the last finalized block it accepted from chain A so it can check the validity of any new blocks from A.

HCastano: Yeah, I'm still trying to find my way around where everything is located, haha. I'll take a look around the system module a little more closely.

srml/bridge/src/lib.rs
HCastano: It looks like that returns the current block number, which isn't necessarily what we want here. See my comment above.

srml/bridge/src/lib.rs
HCastano: Not necessarily. You can imagine a case where one chain wants to act as a bridge for multiple other chains, in which case you'd initialize multiple bridges.

As far as the sender goes, the idea behind this is that governance would be the one calling the shots about whether or not a bridge should be built. I assume that this is some type of privileged call, but I'm not sure how this would work in practice tbh.

srml/bridge/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "bridge"
```
jimpo: srml-bridge

srml/bridge/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+
+pub trait Trait: system::Trait + session::Trait {
+	// The identifier type for an authority.
+	// type AuthorityId: Member + Parameter + RuntimeAppPublic + Default;
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as Bridge {
+		/// Get the ID of the current bridge
+		pub BridgeId get(bridge_id): T::Hash;
```
jimpo: The module should store multiple bridges. We want a `Bridge` struct and store a `map BridgeId => Bridge`. A `BridgeId` can just be a sequentially issued `u64`.

srml/bridge/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+
+		/// Latest set of validators
+		pub Validators get(validators): Vec<T::ValidatorId>;
+	}
+}
+
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		// TODO: Figure out the proper type for these proofs
+		fn new(origin, block_header: T::Header, validator_set: Vec<T::ValidatorId>, validator_set_proof: Vec<u8>, storage_proof: Vec<u8>) {
```
jimpo: What is `new`? That's a very confusing runtime call name.

srml/bridge/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+		// ) {
+
+			// NOTE: Should this be a root call?
+			let _sender = ensure_signed(origin)?;
+
+			Self::check_storage_proof()?;
+			Self::check_validator_set_proof()?;
+
+			// TODO: Do something better than this
+			let bridge_id = <system::Module<T>>::random(b"this_is_not_a_good_source_of_randomness");
```
jimpo: Yeah, they don't need to be random. Sequentially incrementing IDs are fine.

srml/bridge/src/lib.rs
kianenigma: > Not necessarily. You can imagine a case where one chain wants to act as a bridge for multiple other chains, in which case you'd initialize multiple bridges.

Indeed, but in that case you'd want `BridgeId<T>` to hold multiple ids, right? I made this assumption that this function is a one-off because `BridgeId<T>` is a singular value, so I assume your module can be initialised once and only once unless if it has some means of being rested. 

> As far as the sender goes, the idea behind this is that governance would be the one calling the shots about whether or not a bridge should be built. I assume that this is some type of privileged call, but I'm not sure how this would work in practice tbh.

For now I think it does not matter for your development then. You can mock both and I think at the end you'd end up using root or a custom origin.

srml/bridge/src/lib.rs
kianenigma: I think I accidentally mentioned this above. +1

srml/bridge/src/lib.rs
kianenigma: ahh so it is the `LastFinalizedBlock ` of some other chain? Then I'd guess it could be named a bit better + store like a tuple of this finalize hash and some identifier of the chain if needed.

srml/bridge/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+			let block_number = block_header.number();
+			let block_hash = block_header.hash();
+			let state_root = block_header.state_root();
+			let bridge_info = BridgeInfo::new(block_number, &block_hash, state_root, validator_set);
+
+			let new_bridge_id = NumBridges::get() + 1;
+
+			// Hmm, can a call to `insert` fail?
+			// If this is an Option, why do we not need to wrap it in Some(T)?
+			<TrackedBridges<T>>::insert(new_bridge_id, bridge_info);
```
shawntabrizi: Insert cannot fail.

You don't wrap in `Some(T)` because the API handles that for you.

srml/bridge/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+	use sr_primitives::{
+		Perbill, traits::{Header as HeaderT, IdentityLookup}, testing::Header, generic::Digest,
+	};
+	use support::{assert_ok, impl_outer_origin, parameter_types};
+	use runtime_io::with_externalities;
+
+	// NOTE: What's this for?
+	impl_outer_origin! {
+		pub enum Origin for Test {}
+	}
```
shawntabrizi: You can define custom origins, so there is an outer origin which wraps the different origins you support.

HCastano:
Closing this in favour of paritytech/parity-bridges-common#18.

PR #3706, opened 2019-09-26
peerset: fix reserved nodes

andresilva:
The `PeersState` was not taking into account whether we were running `reserved_only` mode when deciding whether to accept a connection to/from a given peer. A flag `priority_only` was added to `PeersState` to control whether only connections to/from nodes in a priority group should be accepted.

I also added a cli parameter `--reserved-only` to control whether we should connect only to reserved/priority nodes. Currently this would be set to true if we specified any reserved nodes which is a bit restrictive. When enabling reserved nodes only we will in fact allow connections from any priority group (namely any authority discovered through the authority module). I think this behavior is reasonable and is documented on the cli parameter, but we can change it to be reserved-only/reserved-and-priority/any if you feel that's necessary.


core/peerset/src/peersstate.rs
```diff
@@ -315,6 +331,11 @@ impl PeersState {
 		self.priority_nodes.get(group_id).cloned()
 	}
 
+	/// Set whether to only allow connections to/from peers in a priority group.
```
tomaka: Must mention that this doesn't affect exiting connections.

tomaka:
Although this is off-topic, I don't really like what peersstate.rs has become.

It used to be 100% documented what happens in every single possible situation that could ever arise.
But since priority groups have been introduced, the number of possible corner cases has increased and some code paths are suddenly blurry and require reading the source code to understand what could happen.

andresilva:
@tomaka If you have any other specific case where documentation can be improved let me know (doesn't matter if it's related to this PR or not). Otherwise we can create a follow-up PR to improve it, I'm not very familiar with this code so I'll take your suggestions on it.

tomaka:
> @tomaka If you have any other specific case where documentation can be improved let me know (doesn't matter if it's related to this PR or not). Otherwise we can create a follow-up PR to improve it, I'm not very familiar with this code so I'll take your suggestions on it.

I haven't touched this code for a very long time, but I would personally look into removing reserved nodes entirely from `PeersState` and move it as a side thing.
This would not be a trivial change however (trivial changes are why the peerstate became what it is now), and probably not worth it.

PR #3708, opened 2019-09-26
Use method instead of Secret::new

c410-f3r:
By using a default method, each `NetworkConfiguration` instance now has
the same set of `node_key` keys as it previously couldn't be true for a cloned
`NodeKeyConfig::Ed25519/Secp256k1(Secret::New)` field

### DIFF
--- a/core/network/src/config.rs
+++ b/core/network/src/config.rs
@@ -272,7 +272,7 @@ impl Default for NetworkConfiguration {
 			listen_addresses: Vec::new(),
 			public_addresses: Vec::new(),
 			boot_nodes: Vec::new(),
-			node_key: NodeKeyConfig::Ed25519(Secret::New),
+			node_key: NodeKeyConfig::Ed25519(Secret::new_random_ed25519()),
 			in_peers: 25,
 			out_peers: 75,
 			reserved_nodes: Vec::new(),
@@ -388,8 +388,20 @@ pub enum Secret<K> {
 	///   * `secp256k1::SecretKey`: An unencoded 32 bytes Secp256k1 secret key.
 	///   * `ed25519::SecretKey`: An unencoded 32 bytes Ed25519 secret key.
 	File(PathBuf),
-	/// Always generate a new secret key `K`.
-	New
+}
+
+impl Secret<ed25519::SecretKey> {
+	/// Generates a random ed25519 secret key
+	pub fn new_random_ed25519() -> Self {
+		Self::Input(ed25519::SecretKey::generate())
+	}
+}
+
+impl Secret<secp256k1::SecretKey> {
+	/// Generates a random secp256k1 secret key
+	pub fn new_random_secp256k1() -> Self {
+		Self::Input(secp256k1::SecretKey::generate())
+	}
 }
 
 impl NodeKeyConfig {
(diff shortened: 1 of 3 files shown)


core/network/src/config.rs
```diff
@@ -388,8 +388,18 @@ pub enum Secret<K> {
 	///   * `secp256k1::SecretKey`: An unencoded 32 bytes Secp256k1 secret key.
 	///   * `ed25519::SecretKey`: An unencoded 32 bytes Ed25519 secret key.
 	File(PathBuf),
-	/// Always generate a new secret key `K`.
-	New
+}
+
+impl Default for Secret<ed25519::SecretKey> {
```
andresilva: Maybe `Default` is not the best trait here. I'd expect to always get the same value when calling `Default` which isn't the case here. Maybe just add a `random` or `generate` method to `Secret` to make it more explicit?

c410-f3r: Added a new method for each variant

romanb:
I would just like to mention that I think this is a misunderstanding of the current (intended, documented) semantics. Quoting from the documentation (emphasis mine):

`NodeKeyConfig`:

> /// The configuration of a node's secret key, describing the type of key
> /// and __how it is obtained__. A node's identity keypair is __the result of
> /// the evaluation of the node key configuration__.

`Secret`:

> /// The configuration options for __obtaining__ a secret key `K`.

`NodeKeyConfig::into_keypair`:

> /// Evaluate a `NodeKeyConfig` to obtain an identity `Keypair`:
> ///
> ///  * If the secret is configured as input, the corresponding keypair is returned.
> ///
> ///  * If the secret is configured as a file, __it is read from that file__, if it exists.
> ///    Otherwise a __new secret is generated__ and stored. In either case, the
> ///    keypair obtained from the secret is returned.
> ///
> ///  * If the secret is configured to be new, __it is generated__ and the corresponding
> ///    keypair is returned.

In other words, a `NodeKeyConfig` configures _how to obtain the key_, not _which key to use_, when turning the config into a keypair. These two notions only coincide if a `NodeKeyConfig` is constructed with `Secret::Input`.

If, for whatever reason, you want to change these semantics such that cloning a `NodeKeyConfig` guarantees that `NodeKeyConfig::into_keypair` always yields the same keypair on every clone, then `Secret::File` would have to be removed as well, since the file might be replaced (or change contents) between the first call of `NodeKeyConfig::into_keypair` and any additional call on a clone, or the keypair is newly generated if the file does not exist. Or the `NetworkConfiguration` would need to be changed to not contain a nested `NodeKeyConfig` but rather only take the resulting keypair as input. In any case, it does not make sense to me to just remove the `Secret::New` variant, especially without updating the documentation properly.

andresilva:
I wasn't very familiar with this code and I should probably have read the documentation a bit before reviewing. I agree with what you said @romanb.

c410-f3r:
Very well explained, @romanb.
Even so, the problem of a different keys still remain and might surprise users. I see 3 different approaches to tackle this.

* Document this intended behavior more clearly
* Change visibility of `node_key` from `pub` to `pub(crate)` (Breaking change)
* @romanb suggestion of using the resulting keypair

Please tell me which solution you guys think is best

romanb:
> Very well explained, @romanb.
> Even so, the problem of different keys still remain and might surprise users. I see 3 different approaches to tackle this.
> 
>     * Document this intended behavior more clearly
> 
>     * Change visibility of `node_key` from `pub` to `pub(crate)` (Breaking change)
> 
>     * @romanb suggestion of using the resulting keypair
> 
> 
> Please tell me which solution you guys think is best

I would suggest to just clarify the documentation where you think it needs clarifying, unless there are important use-cases that simply cannot be implemented correctly with the current API, but I didn't get that impression.

c410-f3r:
Thank you  @romanb. I will close this PR and open a new one with the improved documentation

Issue #3684, opened 2019-09-25
ProvideRuntimeApi is not implemented for AbstractService::client()


alleycat-at-git:
**Reproducing issue:**

Add these lines to node/cli/src/lib.rs, `fn run` (https://github.com/paritytech/substrate/blob/master/node/cli/src/lib.rs#L172)

```
use sr_primitives::traits::ProvideRuntimeApi;

let svc = service::new_full(config).map_err(|e| format!("{:?}", e))?;
let runtime_api = svc.client().runtime_api();
```

This errors with 
```
174 |             let rutime_api = svc.client().runtime_api();
    |                                           ^^^^^^^^^^^
    |
    = note: the method `runtime_api` exists but the following trait bounds were not satisfied:
            `substrate_client::client::Client<<impl substrate_service::AbstractService as substrate_service::AbstractService>::Backend, <impl substrate_service::AbstractService as substrate_service::AbstractService>::CallExecutor, <impl substrate_service::AbstractService as substrate_service::AbstractService>::Block, <impl substrate_service::AbstractService as substrate_service::AbstractService>::RuntimeApi> : sr_primitives::traits::ProvideRuntimeApi`
```

This makes it impossible to interact with the service's `runtime_api` and it looks like it should be implemented for a client https://github.com/paritytech/substrate/blob/master/core/client/src/client.rs#L1393

bkchr:
What do you want to achieve? Why do you need the runtime api at this point?

If we add some trait bounds, `runtime_api` works, but you would not be able to call any method in the runtime as there is some more Rust trait magic required.

alleycat-at-git:
> What do you want to achieve? Why do you need the runtime api at this point?
> 
> If we add some trait bounds, `runtime_api` works, but you would not be able to call any method in the runtime as there is some more Rust trait magic required.

So basically I want to instantiate a Substrate service and be able to query Runtime data like Balances or other modules using that instance. 

The concrete use case for me is I need to make a notification stream of imported blocks (I do that with `import_notification_stream`) and add to each block some data from the blockchain storage. I did that by creating `decl_runtime_api` and `impl_runtime_api` which worked fine but now the problem is I cannot access this API from the Substrate service instance at all.

alleycat-at-git:
@bkchr So I tried to do the same thing using the `new_full!` macro and it worked fine. To me it looks somewhat inconsistent that `runtime_api` is available with `new_full!` but is not available for `new_full` function.

[...]

PR #3709, opened 2019-09-26
Enable runtime_api for Substrate service

alleycat-at-git:
**Description**

Currently when you instantiate a service with `node_cli::service::new_full` there's no way to use `runtime_api()` method of this instance. I.e. 
```rust
use sr_primitives::traits::ProvideRuntimeApi;

let svc = service::new_full(config).map_err(|e| format!("{:?}", e))?;
let runtime_api = svc.client().runtime_api();
```

This code fails with error:
```
174 |             let rutime_api = svc.client().runtime_api();
    |                                           ^^^^^^^^^^^
    |
    = note: the method `runtime_api` exists but the following trait bounds were not satisfied:
            `substrate_client::client::Client<<impl substrate_service::AbstractService as substrate_service::AbstractService>::Backend, <impl substrate_service::AbstractService as substrate_service::AbstractService>::CallExecutor, <impl substrate_service::AbstractService as substrate_service::AbstractService>::Block, <impl substrate_service::AbstractService as substrate_service::AbstractService>::RuntimeApi> : sr_primitives::traits::ProvideRuntimeApi`
```

This is unnecessary limiting since `new_full!` macro (https://github.com/paritytech/substrate/blob/master/node/cli/src/service.rs#L101) actually returns an instance that implements `sr_primitives::traits::ProvideRuntimeApi`. And the `new_full` function is as simple as:
```rust
/// Builds a new service for a full client.
pub fn new_full<C: Send + Default + 'static>(config: Configuration<C, GenesisConfig>)
-> Result<impl AbstractService, ServiceError> {
	new_full!(config).map(|(service, _)| service)
}
```

But because it returns `AbsractService` which doesn't implement `sr_primitives::traits::ProvideRuntimeApi` there's no way to call `runtime_api()`.

Using `new_full!` macro from external modules is impractical because it requires bringing all the additional dependencies into the scope. So the solution I've come up with is to implement the `new_full_concrete` method that returns the exact types and therefore allowing calling `runtime_api()`.

**Alternative solution**

An alternative would be to implement `ProvideRuntimeApi` for `AbstractService` but looks like it's against some design decision as mentioned by @bkchr in the related issue

**Related issues**
#3684


alleycat-at-git:
[clabot:check]

node/cli/src/service.rs
```diff
@@ -33,6 +33,16 @@ use transaction_pool::{self, txpool::{Pool as TransactionPool}};
 use inherents::InherentDataProviders;
 use network::construct_simple_protocol;
 
+use substrate_service::{NewService, NetworkStatus};
+use client::{Client, LocalCallExecutor};
+use client_db::Backend;
+use sr_primitives::traits::{Block as BlockT};
```
bkchr:
```suggestion
use sr_primitives::traits::Block as BlockT;
```

node/cli/src/service.rs
```diff
@@ -221,6 +231,46 @@ pub fn new_full<C: Send + Default + 'static>(config: Configuration<C, GenesisCon
...
+		ConcreteBlock,
+		ConcreteClient,
+		LongestChain<ConcreteBackend, ConcreteBlock>,
+		NetworkStatus<ConcreteBlock>,
+		NetworkService<ConcreteBlock, crate::service::NodeProtocol, <ConcreteBlock as BlockT>::Hash>,
+		TransactionPool<ChainApi<ConcreteClient, ConcreteBlock>>,
+		OffchainWorkers<
+			ConcreteClient,
+			<ConcreteBackend as client::backend::Backend<Block, Blake2Hasher>>::OffchainStorage,
+			ConcreteBlock
```
bkchr:
```suggestion
			ConcreteBlock,
```

node/cli/src/service.rs
```diff
@@ -221,6 +231,46 @@ pub fn new_full<C: Send + Default + 'static>(config: Configuration<C, GenesisCon
...
+		>
+	>,
+	ServiceError
```
bkchr:
```suggestion
	ServiceError,
```

node/cli/src/service.rs
```diff
@@ -221,6 +231,46 @@ pub fn new_full<C: Send + Default + 'static>(config: Configuration<C, GenesisCon
...
+	>;
+#[allow(dead_code)]
+type ConcreteBackend = Backend<ConcreteBlock>;
+
+/// Builds a new service for a full client.
+/// This function returns a [NewService](substrate_service::NewService) rather than `AbstractService`.
+/// It might be handy when you need additional features from service that are not implemented by `AbrtractService`.
+/// Examples might include `service.client().runtime_api()` which is not accessible in `AbstractService` instance
+#[allow(dead_code)]
+pub fn new_full_concrete<C: Send + Default + 'static>(config: Configuration<C, GenesisConfig>)
```
bkchr: Please replace the `new_full`.

PR #3714, opened 2019-09-27
Comment the features used for SGX builds.

pepyakin:
Hey, @brenzi, does it look ok? Feel free to add something

### DIFF
--- a/core/sr-io/Cargo.toml
+++ b/core/sr-io/Cargo.toml
@@ -37,5 +37,10 @@ std = [
 ]
 nightly = []
 strict = []
+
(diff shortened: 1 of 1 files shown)


core/sr-io/Cargo.toml
```diff
@@ -37,5 +37,10 @@ std = [
 ]
 nightly = []
 strict = []
+
+# These two features are used for `no_std` builds for the environments which already provides
```
andresilva:
```suggestion
# These two features are used for `no_std` builds for the environments which already provide
```

andresilva:
I think we can safely ignore the failed check-runtime build.

PR #3717, opened 2019-09-28
refactor contracts to use Time trait

xlc:


kianenigma:
@shawntabrizi @4meta5 this PR is a very good, textbook style, example of how modules should _not directly depend on each other_ and instead try and communicate via associated typed bounded by some common trait. Deserves attention in our docs/recipes/best-practices

PR #3719, opened 2019-09-29
WIP: Node role RPC call

### LINKED ISSUE #1941: RPC to query node role and authority ID
Need a RPC to query if the node is running as a full node, or light client, or authority (with `--validator` cli argument), and the authority ID if the role is authority.

expenses:
(Sort-of) closes #1941. It currently doesn't get the authority ID (I might need a few pointers as for how to connect that up), and the way it figures out node roles is a little hacky.

### DIFF
--- a/core/service/src/lib.rs
+++ b/core/service/src/lib.rs
@@ -722,14 +722,11 @@ fn build_network_future<
 				rpc::system::Request::NodeRole(sender) => {
 					use rpc::system::NodeRole;
 
-					let role = if roles.is_authority() {
-						Some(NodeRole::Authority)
-					} else if roles.is_full() {
-						Some(NodeRole::Full)
-					} else if roles.is_light() {
-						Some(NodeRole::LightClient)
-					} else {
-						None
+					let role = match roles {
+						Roles::AUTHORITY => Some(NodeRole::Authority),
+						Roles::FULL => Some(NodeRole::Full),
+						Roles::LIGHT => Some(NodeRole::LightClient),
+						_ => None
 					};
 
 					if let Some(role) = role {
--- a/core/service/src/lib.rs
+++ b/core/service/src/lib.rs
@@ -663,7 +663,7 @@ fn build_network_future<
 	should_have_peers: bool,
 	dht_event_tx: Option<mpsc::Sender<DhtEvent>>,
 ) -> impl Future<Item = (), Error = ()> {
-	// Compatibility shim while we're transitionning to stable Futures.
+	// Compatibility shim while we're transitioning to stable Futures.
 	// See https://github.com/paritytech/substrate/issues/3099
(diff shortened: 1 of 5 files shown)


core/service/src/lib.rs
```diff
@@ -717,6 +719,20 @@ fn build_network_future<
 						let _ = sender.send(network_state);
 					}
 				}
+				rpc::system::Request::NodeRole(sender) => {
+					use rpc::system::NodeRole;
+
+					let role = match roles {
```
tomusdrw: That really won't work all that well. Usually the `roles` will be a combination of those. Why not return a vector in the RPC as well?

expenses: Can a light client be an authority? If not, I think it's best that we just return a single role, because authority should imply that it's also a full node. Maybe I'm just being confused by this code though (which might not be fully accurate):

https://github.com/paritytech/substrate/blob/00ac65fca22652807e578326ad9f2fe6b169390a/core/network/src/config.rs#L102-L117

I'll try and change the code to better check bitmasks though.

expenses:
ok, so I can clone a `keystore::KeyStorePtr` into `build_network_future`, but how do I use that to get the stash accounts?

bkchr:
Not sure if we want to return the authority ids as well. Currently the node can not know which authority ids are assigned to itself, as only the runtime knows that. We would need to call into the runtime, get all the different key types of the session keys, get all currently active public keys for each of these key types that are registered in the runtime and then we could iterate the keystore to see which of all these keys we can find locally.

I think that is too complicated and we should not do that.

core/service/src/lib.rs
```diff
@@ -717,6 +719,23 @@ fn build_network_future<
 						let _ = sender.send(network_state);
 					}
 				}
+				rpc::system::Request::NodeRole(sender) => {
+					use rpc::system::NodeRole;
+
+					let role = if roles.intersects(Roles::AUTHORITY) {
```
tomusdrw: Sorry for a late review, just noticed it. As I said earler I think it would be best to just return all the roles instead of a single one in the RPC.

The code like this is not future proof, and if we ever happen to add more roles we will need to change the format or return incomplete data.

Also simply returning all `roles` will not require any changes in the RPC if we ever add new ones. If we want to return something human readable, I'd rather have a code that looks like this:
1. Iterate over all bits of `roles`
2. If it's `1` and known return the name of the role it corresponds to.
3. If it's `1`, but not known, return `UnknownRole(<bitnumber>)`

expenses: Oh, right, that suggestion makes sense now, thanks!

tomusdrw:
Looks good! Can you take a look at the test failures?

expenses:
Oops! Is there a way to set things up so I get a notification when CI fails?

tomusdrw:
@expenses I don't know of any solution, but please let me know if you find something!

expenses:
@tomusdrw There might be a setting on gitlab.parity.io, but I don't have access to that.

PR #3729, opened 2019-09-30
srml: system: add kill_prefix

andresilva:
Add a `kill_prefix` extrinsic to system. This is useful to manually cleanup storage from double maps (needed as a result of #3724 but should be useful in general).

### DIFF
--- a/srml/system/src/lib.rs
+++ b/srml/system/src/lib.rs
@@ -278,6 +278,13 @@ decl_module! {
 				storage::unhashed::kill(&key);
 			}
 		}
+
+		/// Kill all storage items with a key that starts with the given prefix.
+		#[weight = SimpleDispatchInfo::FixedOperational(10_000)]
+		fn kill_prefix(origin, prefix: Key) {
+			ensure_root(origin)?;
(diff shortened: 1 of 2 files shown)


srml/system/src/lib.rs
```diff
@@ -278,6 +278,13 @@ decl_module! {
 				storage::unhashed::kill(&key);
 			}
 		}
+
+		/// Kill all storage items with a key that starts with the given prefix.
+		#[weight = SimpleDispatchInfo::FixedOperational(10_000)]
```
andresilva: What should we set this to?

rphmeier: cc @kianenigma

gavofyork: it's entirely dependent on state, so we'd need a specialised weight type. and calculating it would take about as long as actually doing it, so kind of pointless. in any case it's root-only, so doesn't actually matter.

kianenigma: https://github.com/paritytech/substrate/issues/3730

PR #3732, opened 2019-10-01
Let `construct_runtime!` generate `ALL_MODULES` as nested tuples

bkchr:
So, instead of:
`type ALL_MODULES = (Module1, Module2, Module3)`;
Generate it like this:
`type ALL_MODULES = ((Module1, Module2), Module3,);`

This should make `construct_runtime!` support an "unlimited" number of
modules.

### DIFF
--- a/srml/support/src/runtime.rs
+++ b/srml/support/src/runtime.rs
@@ -220,6 +220,7 @@ macro_rules! construct_runtime {
 			$runtime;
 			;
 			{};
+			{};
 			$(
 				$name: $module:: $( < $module_instance >:: )? { $( $modules ),* },
 			)*
@@ -434,6 +435,7 @@ macro_rules! __decl_all_modules {
 		$runtime:ident;
 		;
 		{ $( $parsed:tt )* };
+		{ $( $parsed_nested:tt )* };
 		System: $module:ident::{ Module $(, $modules:ident )* },
 		$( $rest:tt )*
 	) => {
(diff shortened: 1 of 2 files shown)


thiolliere:
actually it doesn't generate `type ALL_MODULES = ((Module1, Module2), Module3,);`

but `type ALL_MODULES = (((Module1, Module2), Module3));`

which means in case of unqiue module: `ALL_MODULES = Module1` and in case of multiple module `ALL_MODULES = ((Module1, Module2), Module3);`

srml/support/src/runtime.rs
```diff
@@ -434,20 +435,42 @@ macro_rules! __decl_all_modules {
...
+		$( $rest:tt )*
+	) => {
+		$crate::__decl_all_modules!(
+			$runtime;
+			$( $system )?;
+			{
+				$( $parsed )*
+				$module::$name $(<$module_instance>)?,
+			};
+			{ $name };
```
thiolliere: for single one there is no comma here

srml/support/src/runtime.rs
```diff
@@ -458,20 +481,23 @@ macro_rules! __decl_all_modules {
 				$( $parsed )*
 				$module::$name $(<$module_instance>)?,
 			};
+			{ ( $( $parsed_nested )*, $name, ) };
```
thiolliere: for multiple one I can't really tell why this last comma doesn't appear, maybe cargo expand does removes it because it is equivalent

PR #3734, opened 2019-10-01
grandpa: Use storage proofs for Grandpa authorities

jimpo:
Currently, the Grandpa protocol requires light clients to request and check execution proofs for the Substrate runtime in order to get the authority set, when in fact the executed runtime API is just a getter for a value from module storage. This is somewhat inefficient and particularly difficult to check inside of a Substrate module acting as a light client for a different chain (as is the case for bridged chains).

This PR migrates the authority set in the `srml-grandpa` module to a well-known storage key and replaces the execution proofs in the Grandpa finality mechanism with storage proofs. For backwards compatibility, an on-chain storage migration can be enabled in the runtime through a feature flag.

Relevant section of chat history with @rphmeier 

> **jimpo**: I've noticed that the Grandpa finality proofs that light clients check are actually execution proofs of the GrandpaApi_grandpa_authorities runtime call
> **rphmeier**: this is an optimization I'm fairly comfortable with, but we will want to standardize the storage location
> **rphmeier**: we'd also want to watch out for Kusama and other deployed networks; we can't just blindly make this change in the runtime.
> **rphmeier**: probably we'd have to feature-flag out the storage location change


core/finality-grandpa/primitives/src/lib.rs
```diff
@@ -172,6 +179,6 @@ decl_runtime_apis! {
 		/// When called at block B, it will return the set of authorities that should be
 		/// used to finalize descendants of this block (B+1, B+2, ...). The block B itself
 		/// is finalized by the authorities from block B-1.
-		fn grandpa_authorities() -> Vec<(AuthorityId, AuthorityWeight)>;
+		fn grandpa_authorities() -> AuthorityList;
```
svyatonik: Could this method be removed? There's dedicated well-known storage key, so why keep it?

srml/grandpa/src/lib.rs
```diff
@@ -175,6 +179,11 @@ decl_module! {
 			// FIXME: https://github.com/paritytech/substrate/issues/1112
 		}
 
+		fn on_initialize() {
+			#[cfg(feature = "migrate-authorities")]
+			Self::migrate_authorities();
```
svyatonik: This migration isn't good for light clients - updated clients may ask for proof-of-read for old blocks (where value under that new well-known key has been empty). So on chains that have been started before this PR, light clients will be unable to finalize (that is only true if there have been GRANDPA authorities changes before migration, though). Not sure if we need to fix this.

srml/grandpa/src/lib.rs
```diff
@@ -242,8 +251,13 @@ decl_module! {
 
 impl<T: Trait> Module<T> {
 	/// Get the current set of authorities, along with their respective weights.
-	pub fn grandpa_authorities() -> Vec<(AuthorityId, AuthorityWeight)> {
-		Authorities::get()
+	pub fn grandpa_authorities() -> AuthorityList {
+		storage::unhashed::get_or_default(GRANDPA_AUTHORITIES_KEY)
```
svyatonik: I'm possibly wrong, but let's say we have upgraded runtime on block B. Until next block (i.e. until `on_initialize` will be called), the value at well known key will be empty. Is it true?

svyatonik:
*(generally, I'm super glad that we're switching from proof-of-execution to proof-of-read here)*

srml/grandpa/src/lib.rs
jimpo: IMO, light clients should *never* check Grandpa justifications from validator sets older than the bonding duration to protect against long-range attacks. Ideally, light clients would be configured with a recent finalized block hash manually (either through software checkpoints or command-line input), then fetch the authority set from that recent block and continue.

However, this is not how Substrate works today, so maybe we do need to worry about it.

core/finality-grandpa/primitives/src/lib.rs
jimpo: Yes, I suspect so. Good point!

core/finality-grandpa/primitives/src/lib.rs
```diff
@@ -46,6 +46,10 @@ pub type AuthoritySignature = app::Signature;
 /// The `ConsensusEngineId` of GRANDPA.
 pub const GRANDPA_ENGINE_ID: ConsensusEngineId = *b"FRNK";
 
+/// The storage key for the current set of weighted Grandpa authorities.
+/// The value stored is an encoded AuthorityList.
+pub const GRANDPA_AUTHORITIES_KEY: &'static [u8] = b":grandpa_authorities";
```
rphmeier: One thing I'd keep an eye out for (that the runtime API approach made handling-of easier) is upgradeability.

We might want to upgrade GRANDPA to use BLS keys at some point. 

We could also have a storage item `:grandpa_protocol_version` which holds the protocol version key. We'd need 2 merkle branches, but it would allow us to change the key type later on without issues.

jimpo: I don't see how the runtime API would make this sort of upgrade any easier? Are keys already encoded with the key type? If so, there's no difference, I'd think, and if not, then the `GrandpaApi_grandpa_authorities` return type would have to change anyway?

rphmeier: the runtime API has an implicit version, that's why

srml/grandpa/Cargo.toml
```diff
@@ -35,3 +35,4 @@ std = [
 	"session/std",
 	"finality-tracker/std",
 ]
+migrate-authorities = []
```
rphmeier: should it not be default?

jimpo: I figured since new chains would want this off that it shouldn't be the default, but I could see it the other way as well.

rphmeier: well, we just have to be careful when backporting then. i'd add something to the PR notes for backporters

DemiMarie-parity: Nit: missing newline at EOF

srml/grandpa/src/lib.rs
DemiMarie-parity: @jimpo does that mean that light clients need to be updated at least every bonding duration?

jimpo: Only if they were not online at all during one bonding period. Basically, if a light client comes online and it's most recent epoch (or is it era? whenever it's most recent validator set was set) is more than one bonding period ago, then it shouldn't catch up without somehow being given a more recent block header by the user or by upgrading software or through some other trusted mechanism.

core/finality-grandpa/primitives/src/lib.rs
jimpo: Or we could store the "authorities version" alongside the authorities in the same key. So the value is an encoded `(u8, AuthorityList)`. That way it's still just one merkle proof.

rphmeier: sure. that will require a more adaptive `Decode` implementation later on, but it should be fine.

jimpo: Something like 8ee07038?

gavofyork:
worth getting this in for substrate 2.0.0 and kusama pos?

srml/grandpa/src/lib.rs
jimpo: Yes, I think that's right.

rphmeier:
For Substrate 2.0, yes, I'd say so. However, @jimpo is working on an alternate storage proof system for the trie (more compact format; omission of value where redundant), so I don't think we could call the GRANDPA light client stable until that's integrated as well.

jimpo:
@rphmeier Are you OK with the versioning scheme?

core/finality-grandpa/src/finality_proof.rs
```diff
@@ -72,33 +73,28 @@ impl<B, E, Block: BlockT<Hash=H256>, RA> AuthoritySetForFinalityProver<Block> fo
...
+		let storage_key = StorageKey(GRANDPA_AUTHORITIES_KEY.to_vec());
+		self.storage(block, &storage_key)?
+			.and_then(|encoded| VersionedAuthorityList::decode(&mut encoded.0.as_slice()).ok())
+			.map(|versioned| versioned.into())
+			.ok_or(ClientError::InvalidAuthoritiesSet)
 	}
 
 	fn prove_authorities(&self, block: &BlockId<Block>) -> ClientResult<Vec<Vec<u8>>> {
-		self.execution_proof(block, "GrandpaApi_grandpa_authorities",&[]).map(|(_, proof)| proof)
+		self.read_proof(block, iter::once(GRANDPA_AUTHORITIES_KEY))
```
rphmeier: nice.

srml/grandpa/src/lib.rs
rphmeier: late to the party, but agreed. that's part of the security model of light clients, which we haven't really launched/deployed yet anyway. breakage there is acceptable especially if it works with less bandwidth.

srml/grandpa/src/lib.rs
```diff
@@ -241,8 +247,16 @@ decl_module! {
 
 impl<T: Trait> Module<T> {
 	/// Get the current set of authorities, along with their respective weights.
-	pub fn grandpa_authorities() -> Vec<(AuthorityId, AuthorityWeight)> {
-		Authorities::get()
+	pub fn grandpa_authorities() -> AuthorityList {
```
rphmeier: nit: the name is a little outdated now.

core/finality-grandpa/primitives/src/lib.rs
```diff
@@ -154,24 +161,51 @@ pub const PENDING_CHANGE_CALL: &str = "grandpa_pending_change";
 /// WASM function call to get current GRANDPA authorities.
 pub const AUTHORITIES_CALL: &str = "grandpa_authorities";
 
-decl_runtime_apis! {
```
andresilva: We will need to re-add the runtime API, it's required for exposing equivocation reporting functionality (#3868). But I can re-add it in my PR after this is merged.

srml/grandpa/src/lib.rs
```diff
@@ -136,8 +133,12 @@ decl_event!(
...
 decl_storage! {
 	trait Store for Module<T: Trait> as GrandpaFinality {
-		/// The current authority set.
-		Authorities get(fn authorities): Vec<(AuthorityId, AuthorityWeight)>;
+		/// DEPRECATED
+		///
+		/// This used to store the current authority set, which has been migrated to the well-known
+		/// GRANDPA_AUTHORITES_KEY unhashed key.
+		#[cfg(feature = "migrate-authorities")]
+		pub(crate) Authorities get(authorities): AuthorityList;
```
andresilva:
```suggestion
		pub(crate) Authorities get(fn authorities): AuthorityList;
```

PR #3737, opened 2019-10-01
Remove unused imports

rockbmb:
Removes some unused imports that gave warnings when compiling `substrate`.

Plus there's a comment in `srml/support/src/storage/storage_items.rs` that refers to three data types being supported where it seems to be just two: `StorageValue` and `StorageMap`.

### DIFF
--- a/core/consensus/babe/src/tests.rs
+++ b/core/consensus/babe/src/tests.rs
@@ -30,7 +30,7 @@ use consensus_common::import_queue::{
 use network::test::*;
 use network::test::{Block as TestBlock, PeersClient};
 use network::config::BoxFinalityProofRequestBuilder;
-use sr_primitives::{generic::DigestItem, traits::{Block as BlockT, DigestFor, NumberFor}};
+use sr_primitives::{generic::DigestItem, traits::{Block as BlockT, DigestFor}};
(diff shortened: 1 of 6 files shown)


rockbmb:
@DemiMarie-parity thanks for merging, but there is something I would like to understand.

One of the CI checks failed because `scripts/gitlab/check_runtime.sh` did as well, because 

```
wasm source files changed but not the spec/impl version and the runtime
binary blob.
```

It then tells me

```
If changes made do not alter logic, just bump 'impl_version'.
If they do change logic, bump 'spec_version' and rebuild wasm.
```

I bumped `impl_version` in `node/runtime/src/lib.rs`, rebuilt locally, and reran the script, yet it fails anyway. Am I doing something improperly here?

DemiMarie-parity:
@rockbmb Honestly, I am not sure, but I suspect it is because they do not alter the compiled runtime at all.  I merged your PR despite the failing check because I considered it a false positive.

thiolliere:
yes your PR doesn't change any logic nor any API so this doesn't need to update spec version nor impl_version.

But if you bump the version and push into github PR the CI must success on this test. Maybe the script doesn't work on local that's all

PR #3739, opened 2019-10-02
Introduce batch call for batch root ops.

gavofyork:


srml/system/src/lib.rs
```diff
@@ -242,36 +243,45 @@ decl_module! {
...
-		#[weight = SimpleDispatchInfo::FixedOperational(10_000)]
+		#[weight = SimpleDispatchInfo::FreeOperational]
 		fn set_storage(origin, items: Vec<KeyValue>) {
 			ensure_root(origin)?;
 			for i in &items {
 				storage::unhashed::put_raw(&i.0, &i.1);
 			}
 		}
 
+		/// Set some items of storage.
```
lsaether:
```suggestion
		/// Send a batch of dispatch calls (only root).
```

PR #3740, opened 2019-10-02
Split off TypeId so as not to pull in sr-io

gavofyork:


gavofyork:
No logic added, so `insubstantial`

PR #3742, opened 2019-10-02
Imply `--base-path ./.local` when `--dev`

gnunicorn:
`--dev` is inherently bound to my currently local run node. This PR adds an implied `./.local` base path when using the `--dev` flag. This came up at a workshop yesterday, when we noticed people having troubles - again - because they had used an earlier version of the workshop. `purge-chain` quickly resolved the - otherwise mysterious - error message of failing authoring and we were questioning whether it really makes sense to store the dev-chain in the user-global directory.

So this PR tries to lighten that problem by implying `--base-path ./.local` (unless set otherwise) when running `--dev`. This prevents new clones from ever interfering with other local clones of either substrate-node or similarly renamed projects (e.g. for workshops).

Additionally this:

 - improves the documentation and description of `--chain` and `--dev` and `--base-path`
 - adds `.local` to the `.gitignore`

//cc @shawntabrizi , @riusricardo

### DIFF
--- a/core/cli/src/params.rs
+++ b/core/cli/src/params.rs
@@ -58,15 +58,28 @@ arg_enum! {
 /// Shared parameters used by all `CoreParams`.
 #[derive(Debug, StructOpt, Clone)]
 pub struct SharedParams {
-	/// Specify the chain specification (one of dev, local or staging).
+	/// Specify the chain specification
+	///
+	/// Can be any of dev, local or staging or a path to a local json definition
 	#[structopt(long = "chain", value_name = "CHAIN_SPEC")]
 	pub chain: Option<String>,
 
-	/// Specify the development chain.
+	/// Specify to run in development setup.
+	///
+	/// It allows the chain to author even if no peers are present.
+	/// Unless any set otherwise, it also implies:
+	///   --chain dev
+	///   --alice
+	///   --base-path ./.local
+	///   --validator
+	///   --no-mdns
+	///   --rpc-cors *
 	#[structopt(long = "dev")]
 	pub dev: bool,
 
-	/// Specify custom base path.
+	/// Specify the base path for the database, keystore and other state
+    ///
+	/// Defaults to the operating systems default location for apps.
 	#[structopt(long = "base-path", short = "d", value_name = "PATH", parse(from_os_str))]
 	pub base_path: Option<PathBuf>,
 
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ rls*.log
 *.orig
 *.rej
 **/wip/*.stderr
+.local/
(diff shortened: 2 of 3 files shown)


tomusdrw:
Wouldn't it be easier to check genesis hash and display a more meaningful error message that the chain doesn't seem to be matching? Maybe even with a proposal to purge the db?)

shawntabrizi:
@tomusdrw it totally makes sense to me to keep the `--dev` chains out of the global filesystem.

`--dev` is almost entirely for development, and as such, it makes sense to keep those artifacts right with the project.

bkchr:
The `node-template` and substrate node will not share the same chain folder.

This does not solve the underlying problem, it just mitigates it for some. Expecting that everyone always clones Substrate to a different path is not something we should expect.

Nevertheless, I would support moving to a "local" `.local` folder if we improve the situation:
1. Check the genesis hash.
2. If the hash is different, provide the user directly the "purge-chain" prompt.
3. Run the node

We could display along 2. that this needs to be done, otherwise it will lead to errors.

core/cli/src/params.rs
```diff
@@ -58,15 +58,28 @@ arg_enum! {
...
+	///   --base-path ./.local
+	///   --validator
+	///   --no-mdns
+	///   --rpc-cors *
 	#[structopt(long = "dev")]
 	pub dev: bool,
 
-	/// Specify custom base path.
+	/// Specify the base path for the database, keystore and other state
+    ///
```
bkchr: Spaces

core/cli/src/params.rs
```diff
@@ -58,15 +58,28 @@ arg_enum! {
...
+	/// Can be any of dev, local or staging or a path to a local json definition
 	#[structopt(long = "chain", value_name = "CHAIN_SPEC")]
 	pub chain: Option<String>,
 
-	/// Specify the development chain.
+	/// Specify to run in development setup.
+	///
+	/// It allows the chain to author even if no peers are present.
+	/// Unless any set otherwise, it also implies:
+	///   --chain dev
```
bkchr: I would put all of them into backticks.

tomusdrw:
@shawntabrizi Definitely that's an improvement as well, I'm not opposing it. Pretty much @bkchr has just summed up my thoughts and I think that's the way we should go too.

core/cli/src/params.rs
tomusdrw:
```suggestion
	///
```

gnunicorn:
I don't think the requested improvements are neither 1. as trivial, nor 2. in the scope of what this PR attempts to. So if this is only expected to be merged with those together, I'll postpone this for a bigger future PR.

Issue #3716, opened 2019-09-27
Arithmetic is incorrectly expected to be unsigned in lots of place in the code base.


thiolliere:
for instance:

https://github.com/paritytech/substrate/blob/35bbdc61a5a8bea545558c9873460e383d2b572e/core/sr-primitives/src/curve.rs#L89-L92

here rem can be negative and thus doesn't really fit into u32.

or here:

https://github.com/paritytech/substrate/blob/35bbdc61a5a8bea545558c9873460e383d2b572e/srml/balances/src/lib.rs#L313-L321

saturating to zero is only if the type is actually unsigned otherwise it simply goes into negative number.

# Strategy

* we should bound num_trait::sign::Unsigned in lots places, also maybe we could create type UnsignedSimpleArithmetic to simplify this use.
* we could also extend some of our tools to handle signed arithmetic, I'm mostly thinking that maybe perbill could be extended to work with signed arithmetic as well. (but this is only if we see a need to support them, otherwise we should just move everything to Unsigned. And this is currently fine).

thiolliere:
This is also the case for Percent (and other Per*) if you do `Percent::from_parts(50) * 11i32` it returns `127`

PR #3743, opened 2019-10-02
Multi-limb arithmetic for runtime

### LINKED ISSUE #3670: Accurately compute operations on u128
Add minimal structures to the runtime arithmetic primitives to support most importantly 

```
fn multiply_by_rational<N: Mul + Div>(a: N, b: N, c: N) -> Option<N>
```

The only tricky case will be u128. Rest is trivial.

This is needed to make all the computations in Staking and phragmen accurate for any possible balance type.

This will be a follow up of https://github.com/paritytech/substrate/pull/3456

kianenigma:
Closes #3670

This adds a new minimal type, `Number` to the arithmetic primitive types of the runtime that can be used to store potentially infinitely large unsigned (or not https://github.com/paritytech/substrate/issues/3716) integers and safely apply the basic operations on them. `Number` can be built from all numeric types and provide `try_into` all typical numeric types. 

The main objective was to support an always accurate 
`pub fn multiply_by_rational(a: u128, b: u128, c: u128) -> Result<u128, &'static str>` while this PR brings more. The main reason is that to have a multiply and division you need both the addition and subtraction for large numbers as well. And the abstraction added is the minimum needed to have a reasonably readable code. The only --potential-- downside is that maybe a super-ugly code that was specialised for only two limbs of u128 would have been slightly faster due to loop unrolling etc, but the algorithms would have been the same. 

before it can properly be reviewed: 

- [x] deal with the few remaining TODOs
- [x] actually use `Number` in multiply_by_rational.
- [x] decide if `Number` is even a good name 😕 

The code is already big I so I will probably not do it here but if this gets merged, there should be follow-up PRs to: 
- relocate all the arithmetic stuff into a separate crate.
- create better and more fuzzing code for all tests that bombard the code with random values.


core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,6 +669,628 @@ impl CheckedAdd for Fixed64 {
 	}
 }
 
+pub mod big_num {
+	use super::{Zero, One};
+	use rstd::{cell::RefCell, convert::TryFrom, cmp::Ordering};
+
+
+	// A sensible value for this would be half of the word size of the host machine. Since the
```
kianenigma: have to double check this

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,6 +669,628 @@ impl CheckedAdd for Fixed64 {
...
+	// runtime is compiled to 32bit webassembly, using 32 and 64 for single and double respectively
+	// should yield the most performance.
+	/// A representation of a single limb.
+	/// TODO
```
kianenigma: double check the statement here.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,6 +669,628 @@ impl CheckedAdd for Fixed64 {
...
+				fn from(a: $type) -> Self {
+					Self { digits: vec! [a.into()] }
+				}
+			})*
+		}
+	}
+	impl_from_for_smaller_than_word!(u8, u16, Single);
+
+	// TODO: allow from for all numeric types bigger than double in a nice way.
+	// TODO: add an implementation for `trait Zero`.
```
kianenigma: both are nice to have, but not strictly needed.

kianenigma:
Should be now ready for review.

cc @expenses

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +669,824 @@ impl CheckedAdd for Fixed64 {
...
+
+		/// Adds `self` with `other`. self and other do not have to have any particular size. Given
+		/// that the `n = max{size(self), size(other)}`, it will produce a number with `n + 1`
+		/// limbs.
+		///
+		/// This function does not strip the output and returns the original allocated `n + 1`
+		/// limbs. The caller may strip the output if desired.
+		///
+		/// Taken from "The Art of Computer Programming" by D.E. Knuth, vol 2, chapter 4.
+		pub fn add(mut self, mut other: Self) -> Self {
```
jimpo: Why does this consume `other` instead of taking `other: &Self`?

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +669,824 @@ impl CheckedAdd for Fixed64 {
...
+		///
+		/// Taken from "The Art of Computer Programming" by D.E. Knuth, vol 2, chapter 4.
+		pub fn mul(self, other: Self) -> Self {
+			let n = self.len();
+			let m = other.len();
+			let mut w = Self::with_capacity(m + n);
+
+			for j in 0..n {
+				if self.get(j) == 0 {
+					w.set(j + m, 0);
```
jimpo: Isn't this already 0?

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +669,824 @@ impl CheckedAdd for Fixed64 {
...
+				"Number {{ {:?} ({:?})}}",
+				self.digits,
+				u128::try_from(self.clone()).unwrap_or_else(|_| 0),
+			)
+		}
+	}
+
+	impl PartialEq for Number {
+		fn eq(&self, other: &Self) -> bool {
+			// sadly, we have to reallocate here as strip mutably uses self.
```
jimpo: Wouldn't it be faster to just do a custom comparison ignoring the leading zero limbs than to reallocate unnecessarily? Or find the index of the first non-zero limbs and do a slice comparison.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +669,824 @@ impl CheckedAdd for Fixed64 {
...
+								for _ in 0..b {
+									acc = acc.checked_mul(a)?;
+								}
+								Some(acc)
+							};
+							let mut acc: $type = Zero::zero();
+							for (i, d) in value.digits.iter().rev().cloned().enumerate() {
+								let d: $type = d.into();
+								let b: $type = B.into();
+								let m = checked_pow(b, i as u32).ok_or(error_message)?;
```
jimpo: Can't you just use left shifts here since B is a power of two? Also, I don't think the overflow checking is necessary since you already know `value.len() * SHIFT <= $len` with `$len` as a multiple of `SHIFT`.

kianenigma:
> There seems to be lots of room for memory efficiency improvements. I think switching to little-endian order (least significant limbs at the front of the vec) would make a lot of stuff easier.

Oh big time. but I intentionally did not (and rather not) include it in this initial PR in favour of brevity. But would like to hear more about your opinion on it.

xlc:
Will it better to move this into a standalone crate? So that other non-substrate projects (smart contracts) can use this as well

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +669,824 @@ impl CheckedAdd for Fixed64 {
...
+		}
+	}
+
+	impl PartialOrd for Number {
+		fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+			Some(self.cmp(other))
+		}
+	}
+
+	impl ops::Add for Number {
```
expenses: Do these traits need to be implemented? I think ideally you want to encourage adds, muls etc. that use references instead of taking ownership. Otherwise looks good.

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: I am generally more in favour of consuming when an arithmetic operation cannot fail. If it might fail, then it is more sensible to take a reference. 

But now seeing the big picture here, I suppose I could also take all the `rhs`s by reference.

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: it for sure is. I was pedantically just following the written steps from the book 😁 👍

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: Yeah we could have the same approach but instead of `lstrip()`-ing we take slices and instead of `resize()`-ing we check if one of the slices has a bigger size. In that case we can early-return. makes sense?

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: ahh right this is a dichotomy actually in the code: I started with `d` value being any to it being the more optimised `leading_zeros()` (defined at the top of the function) wich can then be used with shifts. POwer can go away now 👍

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: yeah I was being too biased here with my personal preference of taking ownership when an operation always returns. I will make all the rhs's reference.

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: Okay sorry my previous comment was confused and was about something else. 

Does B matter here? I can optimize the closure to shift if `i` is a power of two though

kianenigma:
> Will it better to move this into a standalone crate? So that other non-substrate projects (smart contracts) can use this as well

@xlc see above:

> The code is already big I so I will probably not do it here but if this gets merged, there should be follow-up PRs to:
> 
> relocate all the arithmetic stuff into a separate crate.
create better and more fuzzing code for all tests that bombard the code with random values.

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: I have to see, maybe we can also not do it here and leave it for some benchmarks to clearly show the difference and then refactor to reference in some future PR.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +669,824 @@ impl CheckedAdd for Fixed64 {
...
+			rhs.lstrip();
+			lhs.digits.eq(&rhs.digits)
+		}
+	}
+
+	impl Eq for Number {}
+
+	impl Ord for Number {
+		fn cmp(&self, other: &Self) -> Ordering {
+			// sadly, we have to reallocate here as strip mutably uses self.
```
thiolliere: same here no ? we can probably avoid cloning same as patialeq

core/sr-primitives/src/sr_arithmetic.rs
jimpo: It'd be preferable to choose the API in this PR and leave refactors in future PRs to just optimize implementation without changing API ideally.

core/sr-primitives/src/sr_arithmetic.rs
jimpo: You don't need to resize if `get` returns a default of 0 for out-of-range indices.

core/sr-primitives/src/sr_arithmetic.rs
jimpo: I don't think you even need the variable `i`. Just like

```rust
let mut acc: $type = Zero::zero();
for d in value.digits.iter().rev().cloned() {
    let d: $type = d.into();
    acc = (acc << SHIFT) + d;
}
```

kianenigma: implemented with a bit of change

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: This is has already changed and we no longer allocate.

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: resize cannot change (it must change the caller somehow) and get should be a raw getter preferably, similar to other std types. We can have a `checked_get() -> Option<_>` and `unwrap_or` over it. 

I am quite doubtful still regarding accepting arguments as ref or owning. 
For the reason mentioned above, I prefer owning. But since this type can own a large vector of numbers moving stuff around is for sure non-sensical. Moreover, this type is still a _number_ and the std implementation of those takes everything by value (thou' this is not a strong argument as those all have `Copy`).

Hence, if we are to value speed over consistency, I don't see it fair to only make `other` a ref and still consume self. Both will be ref. I am in favour of implementing this.

kianenigma:
> decide if Number is even a good name 😕

Anyone with an opinion on this^?

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +668,954 @@ impl CheckedAdd for Fixed64 {
...
+				// edge cases that should not happen. This basically means that one or both were
+				// zero.
+				(None, None) => Ordering::Equal,
+				(Some(_), None) => Ordering::Greater,
+				(None, Some(_)) => Ordering::Less,
+				(Some(lhs_idx), Some(rhs_idx)) => {
+					let lhs = &self.digits[lhs_idx..];
+					let rhs = &other.digits[rhs_idx..];
+					let len_cmp = lhs.len().cmp(&rhs.len());
+					match len_cmp {
```
thiolliere: matching on only one arm can be done with `if let`, but OK anyway

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +668,954 @@ impl CheckedAdd for Fixed64 {
...
+						Ordering::Equal => {
+							let mut r = Ordering::Equal;
+							for (i, p) in lhs.iter().enumerate() {
+								let q = rhs[i];
+								match p.cmp(&q) {
+									Ordering::Less => { r = Ordering::Less; break} ,
+									Ordering::Greater => { r = Ordering::Greater; break} ,
+									_ => continue,
+								}
+							}
```
thiolliere: can't you simply use implementation of Ord for `[T]`

thiolliere: can't you simply use implementation of Ord for `[T]`

kianenigma: oh I did not know such impl exists :p applied.

But it does not understand length so it says apparently `assertion failed: Number{digits: vec![1, 2, 1, 0],} > Number{digits: vec![1, 2, 3],}'` so I will have the inner match but for lengths are equal I use your suggestion.

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: If you are not super strict on it, I find match to be more readable. Surprisingly, on the current version match is actually 1 line of code shorter than if/else.

core/sr-primitives/src/sr_arithmetic.rs
thiolliere: yes but anyway this is not a big deal. Number implements Clone, changing the API doesn't really break code, user just have to make some more clone and borrow in part of its code that's all.

core/sr-primitives/src/sr_arithmetic.rs
thiolliere: oh yes indeed they say lexicographic order in the doc

thiolliere:
> > decide if Number is even a good name confused
> 
> Anyone with an opinion on this^?

We could also use BigNum, or BigUnsigned, or BigUint as in num-bigint crate 🤔

kianenigma:
> > > decide if Number is even a good name confused
> > 
> > 
> > Anyone with an opinion on this^?
> 
> We could also use BigNum, or BigUnsigned, or BigUint as in num-bigint crate 🤔

BigUint seems the most reasonable.

kianenigma:
API has been changed to `(self, other: &mut Self)` in all cases for now. I might still try and remove the mutability requirement by eliminating `resize`, but other than that I won't touch anything else. Lots of improvements to made listed all here https://github.com/paritytech/substrate/issues/3745

thiolliere:
taking mutable reference seems a bit weird to me, let's say you only have a immutable value you will have to clone it to add it to another one.

Maybe you could remove the resize by introducing a function iterating_resized which only takes reference to BigUInt and iterate on its digits as if it was resized.

kianenigma:
> taking mutable reference seems a bit weird to me, let's say you only have a immutable value you will have to clone it to add it to another one.

muts are gone. I see the current state of the PR as a good fit to get in.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +668,934 @@ impl CheckedAdd for Fixed64 {
...
+		///
+		/// All arguments are examined without being stripped for the above conditions. If any of
+		/// the above fails, `None` is returned.`
+		///
+		/// Taken from "The Art of Computer Programming" by D.E. Knuth, vol 2, chapter 4.
+		pub fn div(self, other: &Self, rem: bool) -> Option<(Self, Self)> {
+			if other.len() <= 1
+				|| other.msb() == 0
+				|| self.msb() == 0
+				|| self.len() <= other.len()
```
thiolliere: is this correct ? 400/300 returns None ?

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +668,934 @@ impl CheckedAdd for Fixed64 {
...
+		}
+
+		/// Divides an `n + m` limb self by a `n` limb `other`. The result is a `m + 1` limb
+		/// quotient and a `n` limb remainder, if enabled by passing `true` in `rem` argument, both
+		/// in the form of an option's `Ok`.
+		///
+		/// - requires `other` to be stripped and have no leading zeros.
+		/// - requires `self` to be stripped and have no leading zeros.
+		/// - requires `other` to have at least two limbs.
+		/// - requires `self` to have a greater or equal length compared to `other`.
```
thiolliere: I feel those constrains are quite strong maybe this could be reduced.

Actually I wonder if we want user to use this BigUInt directly maybe the most exposed API should be with less contrained as require and returning stripped output always ?

like we would rename this div_stripped_multiple_limbs and expose a div which only constrain is that other is different from 0 otherwise it returns 0

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +668,934 @@ impl CheckedAdd for Fixed64 {
...
+					r = self_norm;
+				}
+			}
+
+			Some((q, r))
+		}
+	}
+
+	#[cfg(feature = "std")]
+	impl rstd::fmt::Debug for BigUint{
```
thiolliere:
```suggestion
	impl rstd::fmt::Debug for BigUint {
```

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: yes -- this algorithm, up to my reading from it only works for when the divisor is strictly larger than dividend Of course, there might be some variations that I have not yet seen or thought of. At least in the book, it is exactly the same. For this, there is a `div_single` defined. One can easily write a wrapper and do something like `if lhs.len() > 1 { rhs.div(lhs) } else { rhs.div_single(lhs) }`. Maybe we can provide this. But I am already doing it manually in `multiply_by_rational`

thiolliere: then instead of 
```

		/// - requires `self` to have a greater or equal length compared to `other`.
```

you should write ?
```

		/// - requires `self` to have a greater compared to `other`.
```

core/sr-primitives/src/sr_arithmetic.rs
```suggestion
		/// - requires `self` to have a greater length compared to `other`.
```

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -719,69 +1649,30 @@ pub mod helpers_128bit {
...
-								);
-						}
-					}
-				}
-			}
-		})
+				// PROOF: both `ab` and `c` cannot have leading zero limbs; if length of `c` is 1,
+				// the previous branch would handle. The only case where `div` might return none is
+				// if dividend is smaller than divisor, in which case `Zero` is exactly what we
+				// want.
```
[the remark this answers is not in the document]
kianenigma: Yeah this needs a small rework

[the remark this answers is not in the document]
kianenigma: Okay gave it a better proof. Actually it seems to be imporssible for it to return None.

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +668,934 @@ impl CheckedAdd for Fixed64 {
...
+		pub fn get(&self, index: usize) -> Single {
+			self.digits[self.len() - 1 - index]
+		}
+
+		/// A naive getter for limb at `index`. Note that the order is lsb -> msb.
+		///
+		/// #### Panics
+		///
+		/// This panics if index is out of range.
+		pub fn checked_get(&self, index: usize) -> Option<Single> {
```
thiolliere: this doesn't panic

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -669,16 +668,934 @@ impl CheckedAdd for Fixed64 {
...
+		pub fn len(&self) -> usize { self.digits.len() }
+
+		/// A naive getter for limb at `index`. Note that the order is lsb -> msb.
+		///
+		/// #### Panics
+		///
+		/// This panics if index is out of range.
+		pub fn get(&self, index: usize) -> Single {
+			self.digits[self.len() - 1 - index]
+		}
```
thiolliere: maybe I could be in favor of silently returning zero instead of panic but I'm also fine with that.

kianenigma: Rational was to be act similar to a raw `[]` operator. I suppose with checked_get() it should be fine.

PR #3746, opened 2019-10-02
babe: ancient epoch tree pruning

### LINKED ISSUE #3651: BABE: prune ancient epochs from the fork-tree and prune on epoch input
Right now they grow indefinitely (but slowly).

Two refactorings to be done here:
  1. Flesh out the `babe::epoch_changes::EpochChanges::prune` function. We would want to discard all epochs where `epoch.start_slot + epoch.duration < finalized_header_slot_number`.
  2. Replace the background epoch pruning task with logic to do it whenever we import something new into the tree. At the moment, it's only written when the epoch is updated anyway, so it wouldn't change behavior.

andresilva:
Currently as we observe BABE epoch changes we track them in a fork tree which keeps growing indefinitely as the chain grows. Since we have finality, after a given block is finalized we know that the chain will not revert past that point and therefore we can prune the tree out of epochs that could not be live at the finalized block (i.e. epochs from other forks, and epochs on the canonical chain for which `epoch.end_slot() <= finalized_slot`). The implementation is rather simple since we just use `find_node_where` to find the deepest ancestor of the finalized block that passes the given predicate, and if found we re-root the tree to it.

I built polkadot with this patch using the database of one of the validators which was filled with ancient forks. The tree was pruned correctly and import time went back to normal afterwards (getting the current epoch from the tree taking ~3ms rather than a second). #3665 should also make this significantly faster.

Fixes #3651. @rphmeier suggested the following: 

> IMO we should not prune as eagerly as we possibly can, but instead bake in an allowance of a few epochs just in case we'd like to keep around data from those epochs.

I suggest that we implement this by subtracting a fixed delay to the finalized slot we pass in the predicate when searching the tree, i.e. `epoch.end_slot() <= (finalized_slot - N_EPOCHS_TO_KEEP * SLOTS_PER_EPOCH)`. My initial implementation was keeping the path from root to the current node when traversing the tree, which was then used to traverse back arbitrarily when the correct node was found. This made the implementation much more complicated and I'd also like to avoid introducing `parent` pointers in the tree for the same reason.

**TODO:**
- [x] Test babe pruning on block import

### DIFF
--- a/core/consensus/babe/src/epoch_changes.rs
+++ b/core/consensus/babe/src/epoch_changes.rs
@@ -190,7 +190,7 @@ impl<Hash, Number> EpochChanges<Hash, Number> where
 	}
 
 	/// Prune out finalized epochs, except for the ancestor of the finalized
-	/// block, the given slot should be the slot number at which the finalized
+	/// block. The given slot should be the slot number at which the finalized
 	/// block was authored.
 	pub fn prune_finalized<D: IsDescendentOfBuilder<Hash>>(
 		&mut self,
--- a/core/consensus/babe/src/epoch_changes.rs
+++ b/core/consensus/babe/src/epoch_changes.rs
@@ -189,23 +189,35 @@ impl<Hash, Number> EpochChanges<Hash, Number> where
 		EpochChanges { inner: ForkTree::new() }
 	}
 
-	/// Prune out finalized epochs, except for the ancestor of the finalized block.
+	/// Prune out finalized epochs, except for the ancestor of the finalized
+	/// block, the given slot should be the slot number at which the finalized
+	/// block was authored.
 	pub fn prune_finalized<D: IsDescendentOfBuilder<Hash>>(
 		&mut self,
 		descendent_of_builder: D,
-		_hash: &Hash,
-		_number: Number,
+		hash: &Hash,
+		number: Number,
+		slot: SlotNumber,
 	) -> Result<(), fork_tree::Error<D::Error>> {
-		let _is_descendent_of = descendent_of_builder
+		let is_descendent_of = descendent_of_builder
 			.build_is_descendent_of(None);
 
-		// TODO:
-		// https://github.com/paritytech/substrate/issues/3651
-		//
+		let predicate = |epoch: &PersistedEpoch| match *epoch {
+			PersistedEpoch::Genesis(_, ref epoch_1) =>
+				slot >= epoch_1.end_slot(),
+			PersistedEpoch::Regular(ref epoch_n) =>
+				slot >= epoch_n.end_slot(),
+		};
+
 		// prune any epochs which could not be _live_ as of the children of the
-		// finalized block.
-		// i.e. re-root the fork tree to the oldest ancestor of (hash, number)
-		// where epoch.end_slot() >= slot(hash)
+		// finalized block, i.e. re-root the fork tree to the oldest ancestor of
+		// (hash, number) where epoch.end_slot() >= finalized_slot
+		self.inner.prune(
+			hash,
+			&number,
+			&is_descendent_of,
+			&predicate,
+		)?;
 
 		Ok(())
 	}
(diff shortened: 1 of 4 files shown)


andresilva:
@rphmeier I didn't include any changes to traverse the tree according to max depth (sorting children), I think if we have fast `lca` and a reasonably small tree by pruning that shouldn't be necessary. I can add it in a separate PR if you think it's worth it.

core/utils/fork-tree/src/lib.rs
```diff
@@ -1203,18 +1184,36 @@ mod test {
...
 		assert_eq!(
 			tree.roots.iter().map(|node| node.hash).collect::<Vec<_>>(),
-			vec!["C"],
+			vec!["D"],
 		);
 
 		assert_eq!(
 			tree.iter().map(|(hash, _, _)| *hash).collect::<Vec<_>>(),
-			vec!["C", "D", "E"],
+			vec!["D", "E"],
```
DemiMarie-parity: `"D"` is an ancestor of `"E"`.  Why is it kept?

core/consensus/babe/src/epoch_changes.rs
```diff
@@ -189,23 +189,35 @@ impl<Hash, Number> EpochChanges<Hash, Number> where
...
 
-		// TODO:
-		// https://github.com/paritytech/substrate/issues/3651
-		//
+		let predicate = |epoch: &PersistedEpoch| match *epoch {
+			PersistedEpoch::Genesis(_, ref epoch_1) =>
+				slot >= epoch_1.end_slot(),
+			PersistedEpoch::Regular(ref epoch_n) =>
+				slot >= epoch_n.end_slot(),
+		};
```
marcio-diaz: Looks nicer in the same line ;). Maye forward `end_slot` to `PersistedEpoch`.

core/utils/fork-tree/src/lib.rs
marcio-diaz: It doesn't prune the parent.

core/consensus/babe/src/epoch_changes.rs
andresilva: The logic of which epoch to pick is specific to the context. Notice that here we're interested on `epoch_1` of `Genesis` epoch, since we're interested in the highest slot.

core/consensus/babe/src/lib.rs
```diff
@@ -1289,21 +1273,48 @@ impl<B, E, Block, I, RA, PRA> BlockImport<Block> for BabeBlockImport<B, E, Block
...
 
+				epoch_changes.prune_finalized(
+					descendent_query(&*self.client),
+					&info.finalized_hash,
+					info.finalized_number,
+					finalized_slot,
+				).map_err(|e| ConsensusError::ClientImport(format!("{:?}", e)))?;
+
+				epoch_changes.import(
+					descendent_query(&*self.client),
```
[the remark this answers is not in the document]
andresilva: Ownership of the builder is transferred to the function so it can't be reused.

andresilva:
Missing test added.

core/utils/fork-tree/src/lib.rs
```diff
@@ -111,7 +111,16 @@ impl<H, N, V> ForkTree<H, N, V> where
 		)?;
 
 		if let Some(root) = new_root {
-			self.roots = vec![root.clone()];
+			let mut root = root.clone();
+
+			// we found the deepest ancestor of the finalized block, so we prune
+			// out any children that don't include the finalized block.
+			root.children.retain(|node| {
```
rphmeier: there is only 1 child that can fulfill this precondition, right?

andresilva: Yes, do you think this should be more explicit?

rphmeier: it is just going to be a little more inefficient

rphmeier:
```rust
let children = mem::replace(&mut root.children, Vec::new());
root.children = children.into_iter().filter(|node| ...).take(1).collect()
```

PR #3752, opened 2019-10-03
expose module errors into metadata

### LINKED ISSUE #2954: Generate Metadata for module errors
Followup of #2880

- Generate metadata for module Error enum
- Update runtime metadata to include error information
- Bump metadata version
- Update SDK

xlc:
Closes #2954 
This introduces Metadata V8 cc @jacogr

### DIFF
--- a/srml/metadata/src/lib.rs
+++ b/srml/metadata/src/lib.rs
@@ -217,6 +217,22 @@ pub struct ErrorMetadata {
 	pub documentation: DecodeDifferentArray<&'static str, StringBuf>,
 }
 
+/// All the metadata about errors in a module.
+pub trait ModuleErrorMetadata {
+	fn metadata() -> &'static[ ErrorMetadata ];
+}
+
+impl ModuleErrorMetadata for &'static str {
+	fn metadata() -> &'static[ ErrorMetadata ] {
+		&[
+			ErrorMetadata {
+				name: DecodeDifferent::Encode("Other"),
+				documentation: DecodeDifferent::Encode(&["Other unspecified error"]),
+			},
+		]
+	}
+}
+
 /// A technical trait to store lazy initiated vec value as static dyn pointer.
 pub trait DefaultByte: Send + Sync {
 	fn default_byte(&self) -> Vec<u8>;
@@ -377,7 +393,7 @@ pub struct ModuleMetadata {
 	pub calls: ODFnA<FunctionMetadata>,
 	pub event: ODFnA<EventMetadata>,
 	pub constants: DFnA<ModuleConstantMetadata>,
-	pub errors: ODFnA<ErrorMetadata>,
+	pub errors: DFnA<ErrorMetadata>,
 }
 
 type ODFnA<T> = Option<DFnA<T>>;
--- a/srml/metadata/src/lib.rs
+++ b/srml/metadata/src/lib.rs
@@ -219,17 +219,12 @@ pub struct ErrorMetadata {
 
 /// All the metadata about errors in a module.
 pub trait ModuleErrorMetadata {
-	fn metadata() -> &'static[ ErrorMetadata ];
+	fn metadata() -> &'static[ErrorMetadata];
 }
 
 impl ModuleErrorMetadata for &'static str {
-	fn metadata() -> &'static[ ErrorMetadata ] {
-		&[
-			ErrorMetadata {
-				name: DecodeDifferent::Encode("Other"),
-				documentation: DecodeDifferent::Encode(&["Other unspecified error"]),
-			},
-		]
+	fn metadata() -> &'static[ErrorMetadata] {
+		&[]
 	}
 }
 
--- a/srml/metadata/src/lib.rs
+++ b/srml/metadata/src/lib.rs
@@ -219,11 +219,11 @@ pub struct ErrorMetadata {
 
 /// All the metadata about errors in a module.
 pub trait ModuleErrorMetadata {
-	fn metadata() -> &'static[ErrorMetadata];
+	fn metadata() -> &'static [ErrorMetadata];
 }
 
 impl ModuleErrorMetadata for &'static str {
-	fn metadata() -> &'static[ErrorMetadata] {
+	fn metadata() -> &'static [ErrorMetadata] {
 		&[]
 	}
 }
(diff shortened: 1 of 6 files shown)


srml/support/src/error.rs
```diff
@@ -115,6 +116,29 @@ macro_rules! decl_error {
 				$crate::dispatch::DispatchError::new(None, self.as_u8(), Some(self.as_str()))
 			}
 		}
+
+		impl $crate::error::ModuleErrorMetadata for $error {
+			fn metadata() -> &'static [ $crate::error::ErrorMetadata] {
+				&[
+					$crate::error::ErrorMetadata {
```
bkchr: I'm not sure we want to add these inherent error descriptions every time.

What do you think @jacogr?

If we want them, we should add them only once.

srml/support/src/error.rs
```diff
@@ -115,6 +116,29 @@ macro_rules! decl_error {
 				$crate::dispatch::DispatchError::new(None, self.as_u8(), Some(self.as_str()))
 			}
 		}
+
+		impl $crate::error::ModuleErrorMetadata for $error {
+			fn metadata() -> &'static [ $crate::error::ErrorMetadata] {
```
bkchr:
```suggestion
			fn metadata() -> &'static [$crate::error::ErrorMetadata] {
```

srml/metadata/src/lib.rs
```diff
@@ -209,6 +209,30 @@ pub struct ModuleConstantMetadata {
...
+/// All the metadata about a module error.
+#[derive(Clone, PartialEq, Eq, Encode)]
+#[cfg_attr(feature = "std", derive(Decode, Debug, Serialize))]
+pub struct ErrorMetadata {
+	pub name: DecodeDifferentStr,
+	pub documentation: DecodeDifferentArray<&'static str, StringBuf>,
+}
+
+/// All the metadata about errors in a module.
+pub trait ModuleErrorMetadata {
```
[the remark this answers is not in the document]
xlc: I should be able to remove this. It will just make the macro implementation bit more complicated.
Maybe I should also change `pub errors: DFnA<ErrorMetadata>,` to `pub errors: ODFnA<ErrorMetadata>,` so no error type will be None?

srml/support/src/error.rs
xlc: I am also not sure what is the best way to handle those. Have another field in `ModuleMetadata` can avoid the duplicates but just don't feel right to me. Or just omitting them and that means SDK have to hardcode those messages.
Also du-dupe those will mean every Error type must have `Other` and `CannotLookup` as the first two variants, which could be an acceptable trade-off.

jacogr: Not 100% sure on the scope, if it is just `Other` and `CannotLookup` and they have static indexes for all, I think it is quite acceptable to allow them to be coded externally. 

If we want to move these top-level, I'd rather revisit that we we do types as well (which needs top-level information).

The alternative (and bear in mind this is my first time looking), would probably be to just have all the errors messages top-level and each module errors just has an index to the message.

Can do either, but for V8, I would probably lean "to omitting, specified as defaults" instead of doing too much?

jacogr: PS: Extracting with the current branch, I only see the errors structure being non-empty for the system module, i.e. all the other modules are empty. (So don't see duplication atm)

xlc: right now only system and balances module are using decl_error but I will expect balances module also have errors field. will take a look later

srml/metadata/src/lib.rs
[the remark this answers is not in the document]
bkchr: I think having an array is okay, needs less space.

srml/support/src/error.rs
bkchr: Only system declares an error, so that is expected behavior. 

> The alternative (and bear in mind this is my first time looking), would probably be to just have all the errors messages top-level and each module errors just has an index to the message.

Hmm, would add too much complexity as we don't know error types of other modules in front of. I think the current approach is okay. Maybe at some point later when we need a 2 pass strategy anyway, we can change this as well.

`Other` and `CannotLookup` will be present for every error.

xlc: I have removed those inherent errors from metadata so SDK will need to know the indices starts from 2

srml/metadata/src/lib.rs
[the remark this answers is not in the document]
xlc: I tried to remove the trait but it makes the macro much more complicated, mainly to handle the `&'static str` case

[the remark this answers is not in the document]
xlc: One of the failed attempt https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=339b2cdbd6747104f373b152fb21914d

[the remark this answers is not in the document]
bkchr: Yeah the trait is okay. Sorry for the noise.

srml/metadata/src/lib.rs
```diff
@@ -209,6 +209,25 @@ pub struct ModuleConstantMetadata {
...
+	fn metadata() -> &'static[ErrorMetadata];
```
bkchr:
```suggestion
	fn metadata() -> &'static [ErrorMetadata];
```

srml/metadata/src/lib.rs
```diff
@@ -209,6 +209,25 @@ pub struct ModuleConstantMetadata {
...
+}
+
+impl ModuleErrorMetadata for &'static str {
+	fn metadata() -> &'static[ErrorMetadata] {
```
bkchr:
```suggestion
	fn metadata() -> &'static [ErrorMetadata] {
```

srml/support/src/error.rs
```diff
@@ -62,7 +63,7 @@ macro_rules! decl_error {
 			Other(&'static str),
 			CannotLookup,
 			$(
-				$(#[$variant_attr])*
+				$( #[doc = $doc_attr ] )*
```
bkchr:
```suggestion
				$( #[doc = $doc_attr] )*
```

jacogr:
@xlc A small request for the future - please do not force push to the branches. It makes review harder than it should be. (And in my case, the testing as well, since a simple `git pull` is not simple anymore)

Since we squash merge everything, there is no need to forcefully lose the history in the PR.

Doing final testing on the polkadot-js/api side now.

xlc:
@jacogr sorry for your troubles. I used to rebase over master instead of merge master to have a nicer branch commit history. will just do merge next time.

PR #3753, opened 2019-10-03
Example of how to inspect arguments in weight calculation

kianenigma:
as requested by @JoshOrndorff 

ps. I am not 100% if this is the easiest way to do it. I might have over complicated things and for that a peer review would be nice here. 

@shawntabrizi I noticed that there is a lot of content in `/srml/example` regarding weight and signed extensions. I can gather both as 1-page documents for v2 if they are needed and have a section.


srml/example/src/lib.rs
```diff
@@ -258,13 +258,53 @@ use support::{dispatch::Result, decl_module, decl_storage, decl_event};
...
+/// a particular call and fully extract the `T::Balance` argument out of it. This is
+/// quite some boilerplate to determine the weight of just one call. The reasons for that being:
+///   - This is tailored for a call that only accepts a `T::Balance`. If you want to examine the
+///     argument, there must be one struct per call, because these arguments' type are determined at
+///     compile time and it cannot be a _one size fits all_. In other words, the generic type `T` in
+///     `WeighData<T>` can only be one thing at a time. Of course, this depends on how deeply you
+///     want to examine your arguments. As an example, If you just want to interpret your arguments
+///     in an encoded manner, you can bound `T` to `T: Encode` and you will be able to encode `T`
+///     and examine it.
+///   - Note that you this is a non-runtime code and we cannot directly access a `Call` instance,
```
kianenigma:
```suggestion
///   - Note that you this is a __non-runtime__ code and we cannot directly access a `Call` instance,
```

This is probably nonsense

kianenigma:
Yeah so idk why I assume that we _must not use any knowledge of the runtime_ and the PR is based on that, and, well that's just like making things hard for yourself for no reason 🤷🏻‍♂️. Here's a mock of what would work as a replacement. I will update the PR itself tmr.

```rust 
type BalanceOf<T> = <T as balances::Trait>::Balance;

struct WeightForSetDummy<T>(PhantomData<T>);

impl<T: balances::Trait> WeighData<(&BalanceOf<T>,)> for WeightForSetDummy<T>
	// TODO: why is this needed? T is already defined as being balances::Trait, hence I don't see
	// why we have to say: `<T as balances::Trait>::Balance` is `SimpleArithmetic`
	where BalanceOf<T>: SimpleArithmetic
{
	fn weigh_data(&self, target: (&BalanceOf<T>,)) -> Weight {
		let multiplier = <BalanceOf<T>>::from(100u32);
		(*target.0 * multiplier).saturated_into()
	}
}

impl<T: balances::Trait> ClassifyDispatch<(&BalanceOf<T>,)> for WeightForSetDummy<T> {
	fn classify_dispatch(&self, _: (&BalanceOf<T>,)) -> DispatchClass {
		DispatchClass::default()
	}
}
```

shawntabrizi:
AFAIK Joshy was recently doing some experiements with TX Weight, so it may make sense for you two to tag team and help fill out this page:
https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/blob/source/docs/development/module/weight.md

To start, making a PR which edits or adds bullet points to the questions you expect answered by the doc would be great, and then eventually turning into a one-pager for developers is what we need.

srml/example/src/lib.rs
```diff
@@ -258,13 +258,53 @@ use support::{dispatch::Result, decl_module, decl_storage, decl_event};
...
+///     hence the arguments are examined as a generic `T`, which you can bound to some traits if
+///     that suits you.
```
shawntabrizi: This whole section is really hard for me to read and understand (at least as a first pass). Would want to know if others feel that way too.

srml/example/src/lib.rs
gavofyork: > Note that you this 

that's not english.

srml/example/src/lib.rs
gavofyork: Yeah, could do with rephrasing a little.

kianenigma: rewritten with a simpler implementation as well

srml/example/src/lib.rs
```diff
@@ -258,13 +258,52 @@ use support::{dispatch::Result, decl_module, decl_storage, decl_event};
...
+	traits::{SignedExtension, Bounded, SimpleArithmetic, SaturatedConversion},
+	weights::{SimpleDispatchInfo, DispatchInfo, DispatchClass, ClassifyDispatch, WeighData, Weight},
 	transaction_validity::{
 		ValidTransaction, TransactionValidityError, InvalidTransaction, TransactionValidity,
 	},
 };
 
-/// Our module's configuration trait. All our types and consts go in here. If the
+// A custom weight calculator tailored for the dispatch call `set_dummy()`. This actually examines
+// the arguments and makes a decision based upon them.
```
shawntabrizi: You assume here the reader knows what transaction weight is. I think you can elaborate on "decision" either talking about the max weight of a block or that weights trigger associated fees, or whatever is actually accurate :)

srml/example/src/lib.rs
```diff
@@ -258,13 +258,52 @@ use support::{dispatch::Result, decl_module, decl_storage, decl_event};
...
+//
+// The `WeightData<T>` trait has access to the arguments of the dispatch that it wants to assign a
+// weight to. Nonetheless, the trait itself can not make any assumptions about what that type
+// generic type of the arguments, `T`, is. Based on our needs, we could replace `T` with a more
+// concrete type while implementing the trait. The `decl_module!` expects whatever implements
+// `WeighData<T>` to replace `T` with a tuple of the dispatch arguments. This is exactly how we will
+// craft the implementation below.
```
shawntabrizi: Is this actually important data for the user to read? Sounds like an implementation detail?

What am I supposed to learn here which will help my development?

kianenigma: This (+ the rest of the text actually) all stems from the questions that Joshy had. This scenario would be: you want custom weights. You've heard that you can inspect the arguments whilst calculating weights. Due to some hidden macro code or generally lack of documentation, you don't quite know _how_. This explains exactly that, in detail.

srml/example/src/lib.rs
kianenigma: So I would keep that for a developer hub page about weights. For the sake of your concern, there is another full paragraph in the same page about _what transaction weight is_. I will keep this file simple and short as it is not really intended to be a root documentation. 

What I really want to achieve here is just to have one working example of a common scenario, as Joshy faced.

kianenigma:
> > there is another full paragraph in the same page about what transaction weight is
> 
> I missed this as it comes after your impl.
> 
> Maybe it makes sense to move this impl after `decl_module!` for this reason?
> 
> Otherwise, this looks like a net positive.

I do see the point but then the quirk would be that I will be talking about the internals of the statement `impl <T> WeighData<T> for WeightForSetDummy<_> { ... }` and this statement would be at the top of the file. 

I really think the real solution is a dedicated page in our tutorials. Hence _case closed_ here 😁

Issue #3629, opened 2019-09-17
GRANDPA: extend neighbor packet to ensure vote-blocks are picked up by sync


rphmeier:
In GRANDPA, voters vote on the chains that they perceive to be the best. In order to process a vote, you have to at least be aware of the header it was cast on and its ancestry. In Substrate, the best way to do that now is to import the header. Thus, to complete a GRANDPA round where voters are fragmented across different forks, we have to ensure that we can eventually sync all of those forks.

The sync service makes a best-effort to download all forks right now, but can't do it perfectly because it doesn't have the following two pieces of information:
  - Which blocks do we _really_ need right now?
  - Which peers actually have those blocks?

GRANDPA peers currently broadcast a neighbor packet to each other containing information about the protocol state. Most importantly `round_number` and `set_id`.

This issue proposes extending that neighbor packet with a field `known_vote_blocks: Vec<Block::Hash>`. This should contain the block hashes that votes (`Prepare` and `Precommit`) _in the advertised round of the neighbor packet_ have referenced, which the node has already synced. Essentially, this tells peers which blocks they have which might be necessary in order to make progress.

This list should be bounded by `2*num_voters`, but maybe `2*num_voters + 1` (if we include the primary-proposal message, although that's not necessary to make progress).

We should feed all of these block hashes, along with the `PeerId` of any peers which advertise them, into the sync service. The sync service should attempt to download these chains from the given peer set until told that those block hashes aren't needed anymore (i.e. when we progress onto the next round, we don't _need_ to import those votes anymore).

rphmeier:
The issue could reasonably be done in two parts:
  1. Adjust the sync service so that it can be informed about forks to sync and who to sync from.
  2. Adjust GRANDPA so it informs the sync service

arkpar:
I'll make a PR for the sync service API

arkpar:
#3633

rphmeier:
A couple more things to address:
  - downloading of very long stale forks
  - light client import for stale, needed forks

Also in the GRANDPA side of things, we should not use the `sync::sync_fork` API unless we have already detected a stall for some time. And when moving on beyond a stuck round, we should drop all the requested forks.

mxinden:
Thanks for doing *part one* @arkpar. To ensure we are not duplicating work: I am looking into *part two* right now.

mxinden:
Summary of discussions with André and Rob:

Instead of piggybacking on the *neighbor* messages, we can use the *prevote* and
[...]

PR #3755, opened 2019-10-03
core/finality-grandpa: Gossip synced blocks via neighbor packets

mxinden:
## Goal

In order to participate in the *Grandpa* gossip process a *Grandpa* voter needs
to know:

1. Which blocks are relevant for the current round.

2. Where to retrieve these relevant blocks from.

*Grandpa* voters already exchange *neighbor packets* with each other in an
 epidemic fashion. This pull request extends the *neighbor packet* by a set of
 block hashes which:

1. The node has imported.

2. Have been mentioned by *Grandpa* *prevote* or *precommit* messages.

3. Within the current *Grandpa* round.

Listening to these extended *neighbor packets* a node can pass the information
down to the *network sync* service enabled through
https://github.com/paritytech/substrate/pull/3633.

This is part two of https://github.com/paritytech/substrate/issues/3629.

## Possible (current) solution

In order for a component within `core/finality-grandpa/src` to be able to
achieve the above goal it would need to:

(1) Receive *prevote* and *precommit* messages to keep track of the mentioned
block hashes.

(2) Somehow know whether a given block is available locally (is synced).

(3) Have access to a sink to send out the neighbor packets.

This pull request uses
`core/finality-grandpa/src/communication/gossip::GossipValidator` as it has
access to (1) and (3) and is already the component sending out
`NeighborPacket`s. We pass down a `Client` (stripped down version via
`BlockStatus` trait) through `NetworkBridge` into the `GossipValidator` thus
also achieving (2).

## Pull request status

For now this pull request is only a **draft**. I am opening it up at this point
to ease architectural discussions and to ask questions marked with `TODO`.

What do people think of the approach of extending the `GossipValidator`? Does
this stretch the responsibilities of a *network validator* a bit too much given
that the new logic is proactive?

### DIFF
--- a/core/finality-grandpa/src/communication/mod.rs
+++ b/core/finality-grandpa/src/communication/mod.rs
@@ -43,8 +43,7 @@ use substrate_telemetry::{telemetry, CONSENSUS_DEBUG, CONSENSUS_INFO};
 use tokio_executor::Executor;
 
 use crate::{
-	CatchUp, Commit, CommunicationIn, CommunicationOut, CompactCommit, Error,
-	Message, SignedMessage, BlockImportedChecker,
+	BlockStatus, CatchUp, Commit, CommunicationIn, CommunicationOut, CompactCommit, Error, Message, SignedMessage,
 };
 use crate::environment::HasVoted;
 use gossip::{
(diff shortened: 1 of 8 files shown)


core/finality-grandpa/src/communication/mod.rs
```diff
@@ -147,6 +147,9 @@ impl<B, S, H> Network<B> for Arc<NetworkService<B, S, H>> where
 {
 	type In = NetworkStream;
 
+	// TODO: This is very complex. If I understood correctly we are passing a closure to the network in which we pass a
+	// stream back via a oneshot. The other end of the oneshot is embedded in the `NetworkStream`, which retrieves the
+	// actual stream from the oneshot and polls it from now on. Couldn't `gossip.messages_for` return a stream?
```
andresilva: TBH I don't remember of the top of my head why we do this.

mxinden:
Closing in favor of https://github.com/paritytech/substrate/pull/3800. I will follow up on the comments in future pull requests. Thanks for the help @andresilva.

PR #3757, opened 2019-10-04
Use header metadata in state full and refactor.

marcio-diaz:
Use the more efficient `header_metadata` instead of `header`, and refactor.


core/rpc/src/state/state_full.rs
```diff
@@ -83,58 +84,46 @@ impl<B, E, Block: BlockT, RA> FullState<B, E, Block, RA>
...
+
+		// check if we can get from `to` to `from` by going through parent_hashes.
+		let from_number = from.number;
+		let blocks = {
+			let mut blocks = vec![to.hash];
+			let mut last = to.clone();
+			while last.number > from_number {
+				let hdr = self.client
+					.header_metadata(last.parent)
+					.map_err(client_err)?;
```
tomusdrw: same here

core/rpc/src/state/state_full.rs
```diff
@@ -83,58 +84,46 @@ impl<B, E, Block: BlockT, RA> FullState<B, E, Block, RA>
...
-					hashes: blocks,
-					first_number: from_number,
-					unfiltered_range,
-					filtered_range,
-				})
-			},
-			(from, to) => Err(
-				invalid_block_range(from.as_ref(), to.as_ref(), "Invalid range or unknown block".into())
-			),
+		let from = self.client.header_metadata(from).map_err(client_err)?;
```
tomusdrw: Previously the error message in case a block is missing was much more friendly than this one. I guess right now you will get quite cryptic `ClientError`, where previously you were actually getting errors in consistent format. I really think we should get this behaviour back and perhaps even add a test for this case.

marcio-diaz: This is exactly the same as the original implementation, printing something like `header not found in db: {hash}`, just compare the lines. So, `much more friendly and consistent` sounds like an exaggeration  :-).

But I guess we can use `invalid_block_range` everywhere and print both hashes/number.

bkchr:
Status?

gavofyork:
@marcio-diaz when this is ready for review again, please tag `pleasereview` and ensure your reviewer is notified.

marcio-diaz:
I added tests @tomusdrw , please take another look.

client/rpc/src/state/state_full.rs
```diff
@@ -91,59 +93,49 @@ impl<B, E, Block: BlockT, RA> FullState<B, E, Block, RA>
...
-			(from, to) => Err(
-				invalid_block_range(from.as_ref(), to.as_ref(), "Invalid range or unknown block".into())
-			),
+		let to = self.block_or_best(to).map_err(|e| invalid_block::<Block>(from, to, e.to_string()))?;
+
+		let invalid_block_err = |e: ClientError| invalid_block::<Block>(from, Some(to), e.to_string());
+		let from_meta = self.client.header_metadata(from).map_err(invalid_block_err)?;
+		let to_meta = self.client.header_metadata(to).map_err(invalid_block_err)?;
+
+		if from_meta.number > to_meta.number {
```
tomusdrw: Does `from == to` make sense?

client/rpc/src/state/state_full.rs
```diff
@@ -91,59 +93,49 @@ impl<B, E, Block: BlockT, RA> FullState<B, E, Block, RA>
...
+			return Err(invalid_block_range(&from_meta, &to_meta, "from number > to number".to_owned()))
 		}
+
+		// check if we can get from `to` to `from` by going through parent_hashes.
+		let from_number = from_meta.number;
+		let blocks = {
```
tomusdrw:
```suggestion
		let hashes = {
```
Since we only store hashes, renaming the var makes more sense to me.

client/rpc/src/state/state_full.rs
```diff
@@ -91,59 +93,49 @@ impl<B, E, Block: BlockT, RA> FullState<B, E, Block, RA>
...
+			let mut last = to_meta.clone();
+			while last.number > from_number {
+				let hdr = self.client
+					.header_metadata(last.parent)
+					.map_err(|e| invalid_block_range::<Block>(&from_meta, &to_meta, e.to_string()))?;
+				blocks.push(hdr.hash);
+				last = hdr;
+			}
+			if last.hash != from_meta.hash {
+				return Err(invalid_block_range(&from_meta, &to_meta, "invalid block range".to_owned()))
```
tomusdrw:
```suggestion
				return Err(invalid_block_range(&from_meta, &to_meta, "from and to are on different forks".to_owned()))
```

client/rpc/src/state/state_full.rs
```diff
@@ -91,59 +93,49 @@ impl<B, E, Block: BlockT, RA> FullState<B, E, Block, RA>
...
+
+		// check if we can get from `to` to `from` by going through parent_hashes.
+		let from_number = from_meta.number;
+		let blocks = {
+			let mut blocks = vec![to_meta.hash];
+			let mut last = to_meta.clone();
+			while last.number > from_number {
+				let hdr = self.client
+					.header_metadata(last.parent)
+					.map_err(|e| invalid_block_range::<Block>(&from_meta, &to_meta, e.to_string()))?;
```
tomusdrw:
```suggestion
					.map_err(|e| invalid_block_range::<Block>(&last, &to_meta, e.to_string()))?;
```

client/rpc/src/state/state_full.rs
marcio-diaz: Added.

client/rpc/src/state/state_full.rs
marcio-diaz: yeah

PR #3759, opened 2019-10-04
Utility module for doing stuff like batch calls

gavofyork:


srml/utility/src/lib.rs
```diff
@@ -0,0 +1,175 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2019 Parity Technologies (UK) Ltd.
```

PR #3761, opened 2019-10-04
Alternative sysvar setup for Windows

Swader:
The command line setup did not work for me. This adds instructions and images on how to do it through the UI.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -253,6 +253,13 @@ If you are trying to set up Substrate on Windows, you should do the following:
 	[System.Environment]::SetEnvironmentVariable('OPENSSL_DIR', $env:OPENSSL_DIR, [System.EnvironmentVariableTarget]::User)
 	[System.Environment]::SetEnvironmentVariable('OPENSSL_STATIC', $env:OPENSSL_STATIC, [System.EnvironmentVariableTarget]::User)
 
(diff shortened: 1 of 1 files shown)


Swader:
[clabot:check]

kianenigma:
Thanks for the PR!

I wonder: 
- Why would the original approach not work? is there a common pitfall? 
- I am not against having this additional section, but if we do, the images should preferably be placed in one file or a way to render while taking less vertical space. I think all of them can fit in one image with _steps_.

shawntabrizi:
@Swader I think the better solution is to move these specific instructions into the Substrate Developer Hub?

I too am interested to know why these steps do not work. I wrote them having tested on my personal windows machine and actually got a few people to install on their windows machine too with the same instructions.

Swader:
@kianenigma I agree the rendering is... less than pretty. Wasn't sure how else to do it without breaking user flow (i.e. see this _yet another document_).

@shawntabrizi  They did not work for me. See image below. This is the error I get for every one of the lines in the instruction paragraph. @lsaether confirmed some others saw this problem too.

image

bkchr:
You need to use powershell

Swader:
The readme specifically references CMD, and in CMD (and Git Bash) all instructions up until this point work. So this should be specifically mentioned then.

shawntabrizi:
@Swader can you make this PR that change to specifically use PowerShell rather than adding the extra images?

Or do you think that is not sufficient?

Swader:
It should be sufficient. Let me test and if so, I'll make the PR update.

Swader:
Powershell works fine

Issue #2914, opened 2019-06-20
Refactor `srml-support-procedural`


bkchr:
The procedural code of `srml-support` is just a big mess that is almost not maintainable.

thiolliere:
It seems ok now

PR #3765, opened 2019-10-04
Refactor decl storage

thiolliere:
Related to https://github.com/paritytech/substrate/issues/2914

OK this is a step in the direction:

### Done:

* split implementation in multiple files:
  * transformation -> genesis_config/  getters.rs  instance_trait.rs  metadata.rs  mod.rs  store_trait.rs
  * mod.rs -> parser.rs
  * impl.rs -> storage_struct.rs
* parser is isolated into parse module, it could be improved as well but this can be done in another PR
* modules contains a defintion of decl_storage input which must be ok to work with.
* implementation change:
  * T: Trait might be more often bound to 'static (anyway we only use static one and it is needed for metadata current implementation).
  * GenesisConfig no longer requires its fields to be Clone (possible since to EncodeLike feature)
  * builder for map storages must return precise type Vec<(key, value)>


srml/support/procedural/src/storage/genesis_config.rs
```diff
@@ -0,0 +1,349 @@
...
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use srml_support_procedural_tools::syn_ext as ext;
+use proc_macro2::TokenStream as TokenStream2;
```
bkchr:
```suggestion
use proc_macro2::TokenStream;
```

srml/support/procedural/src/storage/genesis_config.rs
```diff
@@ -0,0 +1,349 @@
...
+
+use syn::{WhereClause, spanned::Spanned};
+use quote::{quote, quote_spanned};
+
+use super::instance_trait::DEFAULT_INSTANTIABLE_TRAIT_NAME;
+
+const DEFAULT_INSTANCE_NAME: &str = "__GeneratedInstance";
+
+pub fn impl_genesis_config(scrate: &TokenStream2, def: &super::DeclStorageDefExt) -> TokenStream2 {
```
bkchr: Refactor and introducing a 320 lines function, is no "refactoring" ;) No one is able to read/understand this.

srml/support/procedural/src/storage/getters.rs
```diff
@@ -0,0 +1,78 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use proc_macro2::TokenStream as TokenStream2;
```
bkchr:
```suggestion
use proc_macro2::TokenStream;
```

srml/support/procedural/src/storage/genesis_config.rs
thiolliere: Yes actually this function hasn't been really touched, just moved. but ok I can do it here as well.

thiolliere: Ok improved

srml/support/procedural/src/storage/mod.rs
```diff
@@ -14,218 +14,313 @@
...
+	optional_storage_runtime_bound_comma: Option<proc_macro2::TokenStream>,
+	optional_storage_where_clause: Option<proc_macro2::TokenStream>,
+	storage_trait: proc_macro2::TokenStream,
+	storage_generator_trait: proc_macro2::TokenStream,
+	is_generic: bool,
+	is_option: bool,
 }
 
+impl StorageLineDefExt {
+	fn from_def(storage_def: StorageLineDef, def: &DeclStorageDef) -> Self {
```
kianenigma: Maybe make it consistent with `impl From<DeclStorageDef> for DeclStorageDefExt` by implementing sth like 
`impl From<(StorageLineDef, &DeclStorageDef)> for StorageLineDef `

thiolliere: hmm I usually don't very like to use From trait for tuple types like this. this looks more like a function signature than a type by itself.

thiolliere:
the general structure hasn't been really changed. things have been moved around and modified by adding structures here and there.

OK I'll add doc.

Also parsing refactor can be done later with https://github.com/paritytech/substrate/issues/3788

thiolliere:
doc added

srml/support/procedural/src/storage/genesis_config/builder_def.rs
```diff
@@ -0,0 +1,134 @@
...
+	/// * build block for storage with build attribute.
+	/// * build block for storage with config attribute and no build attribute.
+	/// * build block for extra genesis build expression.
+	pub blocks: Vec<TokenStream>,
+	/// The build blocks requires generic traits.
+	pub is_generic: bool,
+}
+
+impl BuilderDef {
+	pub fn from_def(scrate: &TokenStream, def: &DeclStorageDefExt) -> Self {
```
bkchr: Most of the code in this function is still duplicated. Mostly it is just the code for the providing the data that should be inserted, that is different.

srml/support/procedural/src/storage/genesis_config/builder_def.rs
```diff
@@ -0,0 +1,134 @@
...
+
+//! Builder logic definition used to build genesis storage.
+
+use srml_support_procedural_tools::syn_ext as ext;
+use proc_macro2::TokenStream;
+use syn::spanned::Spanned;
+use quote::{quote, quote_spanned};
+use super::super::{DeclStorageDefExt, StorageLineTypeDef};
+
+pub struct BuilderDef {
```
bkchr: Some docs would be nice.

srml/support/procedural/src/storage/genesis_config/builder_def.rs
thiolliere: oh this was because builder data is `Vec<key, value>` whereas config data is `Vec<&key, &value>` but I can come with some factorisation I think

thiolliere: improved, and also now builder output must be a vec for maps not any iterator ? I feel that it improves error message like:
```
  --> srml/support/test/tests/genesisconfig.rs:28:36
   |
28 |         pub AppendableDM config(t) build(|config| { ()}): double_map u32, blake2_256(T::BlockNumber) => Vec<u32>;
   |                                          ^ expected struct `std::vec::Vec`, found ()
   |
   = note: expected type `&std::vec::Vec<(u32, <T as Trait>::BlockNumber, std::vec::Vec<u32>)>`
              found type `&()`
```

And if really one wants to make specific insert he can just use extra genesis.

srml/support/procedural/src/storage/genesis_config/builder_def.rs
thiolliere: done

PR #3769, opened 2019-10-05
Move module-template to sub-package of node-template

JoshOrndorff:
The Developer Experience team has sometimes been torn about where to recommend new runtime modules be built. The candidates are to build in a separate rust package like the SRML modules, or to build directly in [template.rs](https://github.com/paritytech/substrate/blob/062748f2b932a00947d1b9b2ae85f6dcbbce03e1/node-template/runtime/src/template.rs).

Pros for own package.
* Easier to use in 3rd-party runtimes
* Module can evolve separately from a particular runtime
* Emphasizes that runtimes are made by composing runtime modules
* Straightforward to create a new one by cloning the [module template](https://github.com/substrate-developer-hub/substrate-module-template/)

Pros for `template.rs`
* Beginners can start hacking a file that already exists in node template
* No need to learn cargo dependencies before building

This PR attempts to get the best of both worlds by leaving a simple module template directly in the node-template's runtime, but by structuring the module as a package. This approach is already precedented by the runtime living as a package inside the node and node-templates


shawntabrizi:
@JoshOrndorff the problem with this for a beginner is that, and as you have seen, there is a huge number of ways a user can start to mess up when doing development.

For example, your Cargo.toml file imports the bare minimum number of crates needed, and often when writing a runtime, users quickly start depending on other parts of Substrate core to be used in their runtime logic.

As you mentioned the crate system is something they will have to learn, but not only that, but also the feature system (`no_std`), internal knowledge of the crates available in Substrate, etc...

While I agree, generally creating modules as individual crates is the more practical thing to do if you plan to share your module, it is really not the "initial path" for most people. See for example the Polkadot runtime, which does not even do this kind of process: https://github.com/paritytech/polkadot/tree/master/runtime/src

Before you make a merge like this to the main Substrate repo, and change the development behavior of everyone building on Substrate, I think we externally should more heavily rely on the `substrate-module-template` and see how that affects a new users ability to onboard.

Furthermore, if we were to ultimately merge something like this, I think it would make sense to fill the Cargo.toml file with many more imports to try and at least reduce one more possible error situation for the user. (unless there is a reason not to do this)

4meta5:
I like this and am pro-merging. I appreciate how it separates module development from runtime configuration. This will make it more clear to developers how the modules are libraries and the runtime is an executable that configures the modules. As Joshy said, "Emphasizes that runtimes are made by composing modules"

> generally creating modules as individual crates is the more practical thing to do if you plan to share your module, it is really not the "initial path" for most people. See for example the Polkadot runtime, which does not even do this kind of process

I don't think the Polkadot runtime is the best example for the development path expected from our target audience...Joshy's approach is especially nice because it demonstrates how to create modules as individual crates so that you can share the modules (as Shawn said). In my opinion, this is a nice first step for building with Substrate before the modules are configured into a runtime.

JoshOrndorff:
Thanks for the feedback so far everyone. Let me address a few of the concerns.

> `Cargo.toml` file imports the bare minimum number of crates needed.

It imports enough to support the template module itself and also the [Proof of Existence tutorial](https://substrate.dev/docs/en/next/tutorials/creating-your-first-substrate-chain/). I will add others if doing so is thought wise. Do note that this which-dependencies-should-we-pre-import problem exists whether or not the module is its own package.

> As you mentioned the crate system is something they will have to learn

This PR allows users to develop their module in its own crate _without_ having to first learn cargo dependencies and features. It does the crate scaffolding for them. Beginners still just copy paste code into a single file. Now that file is `/node-template/runtime/module-template/src/lib.rs` instead of `node-template/runtime/src/template.rs`.

shawntabrizi:
But this is an incomplete thought imo.

The way the node-template is generated (as part of the Substrate Node Template) is to generate specific git dependencies on all the Substrate cargo items.

This means that node templates, and modules created this way will only support runtimes which have that exact commit number.

To think through this thought completely, I want to understand, with this change merged in, then what? Are users today able to gain something today from this more complicated process?

Looking at your Pros:

> Easier to use in 3rd-party runtimes

Not today, no it isnt for the commit dependency reason I described.

> Module can evolve separately from a particular runtime

I don't know what this means.

> Emphasizes that runtimes are made by composing runtime modules.

I don't see the change in emphasis here. You switched from a Rust module to a Rust library. From a mental model, they are the same.

> Straightforward to create a new one by cloning the module template

We already have the module template as a separate repo, which does not interfere with the easiest path for runtime development.

shawntabrizi:
I also want to re-emphasize my statement:

> Before you make a merge like this to the main Substrate repo, and change the development behavior of everyone building on Substrate, I think we externally should more heavily rely on the `substrate-module-template` and see how that affects a new users ability to onboard.

Why when you have an idea like this, can't we test this idea on customers using the package/templates we have set up on Developer Hub?

I am not saying to throw the change away, I am simply saying to _test_ with users how it affects their onboarding experience. I think that if you did such a test, you would quickly see how impractical these separate runtime modules as crates are with hard-coded dependencies to a specific `rev`.

JoshOrndorff:
I chose to make this PR to Substrate itself first because Substrate Developer Hub's [`substrate-node-template` repo](https://github.com/substrate-developer-hub/substrate-node-template) has always been  taken directly from Substrate itself. That seems like a fine idea and I expected that to continue. I've opened ` [PR against devhub's node template](https://github.com/substrate-developer-hub/substrate-node-template/pull/2) to illustrate how this change would flow downstream to devhub repos if it is merged.

I'm claiming that with this change merged, nothing gets harder or worse for beginners, and some things get better. Specifically these things get better:
* They start developing their module in its own crate (although not it's own repo) from day 1, so they experience how modules can each live in their own crates. This is without any overhead or messing with `Cargo.toml` files.
* If or when their module does become mature enough to live in it's own repo, that transition is less work than without this PR. It is merely a matter of moving the module-template directory to it's own repo, and changing a few lines in `Cargo.toml` files.

I would like some additional input. Maybe @gavofyork or @bkchr could give feedback. The first question is "Is it a good idea to move the module template in its own crate like I've shown here?" If so I'll happily address any specific concerns.

bkchr:
Why not just have both in the node template? One "internal-module" module and one "crate-module"?

shawntabrizi:
Basti, your idea is agreeable, but here is what I would do:

* When v2.0 branch is released, all parts of the package (node and module template) are set to use this branch (rather than a commit), which means for that period of time (the lifetime of 2.0), all nodes and modules created with the template are cross compatible.

* Module template is taught as a "medium level" runtime development practice. So anything where we expect the user to be beyond beginner (whatever that means), we teach them to create a module crate, to do custom crate imports, etc...

* In the Substrate main repository, we do not update the node-template to include a new separate crate. Instead we point users to look at the SRML folder (like the example module) to see how it can be done. Any of the existing modules like Balances, System, etc are already examples of how this can be done from within the Substrate repository.

I want to re-state my general concern:

* There is a non-zero increase in module development difficulty when creating runtime modules in this way. Even if it is better long term to teach a user to create a seperate module crate, this is in my opinion something only medium and higher developers would do, and value is only added doing it this way when original ideas are created on a runtime module. Having this be the default pathway for all new runtime developers does not make sense at the moment.

* When you include the module template in the node template like we did here, the script which creates the `substrate-node-template` generates hard coded git commit for the specific version of Substrate used at the time of generating the node-template. This means that the module generated in this way will also have a hard coded commit, which is not very useful to any other node which happens to use any commit which is not that specific one. The v2.0 branch commit will fix that, but we will need to do that manually in our own repo, since the script does not (and should not) work that way in Substrate

bkchr:
I think there was enough discussion here, when there is still something to be discussed, it should be done in an issue or in Riot. 

I will close this pr for now.

(mainly doing it for cleaning up open prs)

PR #3771, opened 2019-10-06
Use algorithm of path sections for lca

marcio-diaz:
This is better than current algorithm in worst case situations.

TODO: do the same for light and other implementations.

\cc @andresilva

### DIFF
--- a/core/client/db/src/lib.rs
+++ b/core/client/db/src/lib.rs
@@ -411,11 +411,29 @@ impl<Block: BlockT> HeaderMetadata<Block> for BlockchainDb<Block> {
 	fn header_metadata(&self, hash: Block::Hash) -> Result<CachedHeaderMetadata<Block>, Self::Error> {
 		self.header_metadata_cache.header_metadata(hash).or_else(|_| {
 			self.header(BlockId::hash(hash))?.map(|header| {
-				let header_metadata = CachedHeaderMetadata::from(&header);
+				let mut ancestor = header.parent_hash().clone();
+				let number = *header.number();
+				let parent = *header.parent_hash();
+				let section_size = 500;
+
+				if  number % section_size.into() != Zero::zero() && number > One::one() {
+					if let Ok(parent) = self.header_metadata(ancestor) {
+						ancestor = parent.ancestor;
+					}
+				}
+
+				let header_metadata = CachedHeaderMetadata {
+					hash,
+					number,
+					parent,
+					ancestor,
+				};
+
 				self.header_metadata_cache.insert_header_metadata(
 					header_metadata.hash,
 					header_metadata.clone(),
 				);
+
 				header_metadata
 			}).ok_or(client::error::Error::UnknownBlock("header not found in db".to_owned()))
 		})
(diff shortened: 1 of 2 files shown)


rphmeier:
> TODO: do the same for light and other implementations.

I'm surprised that this needs to be done at all. Shouldn't this code be in only one place?

marcio-diaz:
> > TODO: do the same for light and other implementations.
> 
> I'm surprised that this needs to be done at all. Shouldn't this code be in only one place?

Well, of course it can be done. I meant "make it work for everything" if people think this change is worth.

core/client/db/src/lib.rs
```diff
@@ -411,11 +411,29 @@ impl<Block: BlockT> HeaderMetadata<Block> for BlockchainDb<Block> {
 	fn header_metadata(&self, hash: Block::Hash) -> Result<CachedHeaderMetadata<Block>, Self::Error> {
 		self.header_metadata_cache.header_metadata(hash).or_else(|_| {
 			self.header(BlockId::hash(hash))?.map(|header| {
-				let header_metadata = CachedHeaderMetadata::from(&header);
+				let mut ancestor = header.parent_hash().clone();
+				let number = *header.number();
+				let parent = *header.parent_hash();
+				let section_size = 500;
```
rphmeier: should be `const`. and is there any justification for that value?

core/client/db/src/lib.rs
```diff
@@ -411,11 +411,29 @@ impl<Block: BlockT> HeaderMetadata<Block> for BlockchainDb<Block> {
...
+
+				if  number % section_size.into() != Zero::zero() && number > One::one() {
```
rphmeier:
```suggestion
				if number % section_size.into() != Zero::zero() && number > One::one() {
```

andresilva:
Are we sure that this is better for the average case (where the tree is pruned?). The current implementation is simpler and seems like it would work better assuming we don't exhaust the cache.

core/client/header-metadata/src/lib.rs
```diff
@@ -22,7 +22,7 @@ use parking_lot::RwLock;
 use lru_cache::LruCache;
 
 /// Set to the expected max difference between `best` and `finalized` blocks at sync.
-const LRU_CACHE_SIZE: usize = 5_000;
+const LRU_CACHE_SIZE: usize = 10_000;
```
bkchr: This test: https://github.com/paritytech/substrate/pull/3776/files#diff-c0982d5a95c967fde94ecf0a10d342dbR2252
Seems to rely that the value is 5000.

CC @andresilva

marcio-diaz:
> Are we sure that this is better for the average case (where the tree is pruned?). The current implementation is simpler and seems like it would work better assuming we don't exhaust the cache.

Yes, it's hard to choose a good and simple algorithm. I was ok with the current one, until I ran it in the database you gave me :). Is there a max height in the case of tree pruned?

The current algorithm is constant time in a very specific case. For instance  `lca(1, 25_000), lca(2,  25_000)` will take ~25k iterations each query.

The proposed algorithm, if we choose `section_size = sqrt(25000) = 500` will take at most 1k iterations, for any query (with height <= 25k). Combined with the cache, it seems to be fast enough. 

Also, I think is not much more expensive (constant space) to combine both approaches.

andresilva:
The database I gave you is an edge case that shouldn't happen again going forward. We should make sure we don't pessimize the average case in lieu of an edge case. Either way, if we're going to include this change this needs tests specifically for the edge conditions of this code (i.e. all ifs should be covered).

bkchr:
Status?

marcio-diaz:
Closing, I'm not 100% convinced on the algorithm either. I'll make benchmarks and rewrite it.

PR #3774, opened 2019-10-07
Chain key value storage

cheme:
This is ongoing work on a new storage kind for substrate. I do open the PR to follow the progress, and get feedback.
The amount of code start to grow, and I am at a point where I want to stabilize, clean and document code, for this putting some roadmap in place will help.

Seems also like a good time to start design discussion.

# Description

Some info needs to be in synch with blockchain state (information accessible or any block), but does not require the proof property.

This PR intends to have such storage. Design goals are:
- lower overhead than trie storage: we store directly key values directly (with the need to have some history management: so double indexing and a bit of history reading).
- synch of info up to the extrinsics call:
  - support revert when evaluating an extrinsic, and state update for next extrinsic
  - support chain reorg

Basically the same thing as the existing state trie storage without the trie proof overhead.

# Use case

- child trie ids: 

Child trie key isolation can be based on parent storage path, but this does not play well with move or delete child trie operations. So associating a unique id with every child trie seems needed (see #2209). Yet those uniqueids are only an implementation details and should not be part of the state (and in case of full storage with pruning it could not (this case does not exists at this time but indicate a severe design limitation of #2209)).
Using this storage for those child trie unique id (or keyspace) seems fine. For instance, a child trie deletion and then re-creation (at the same address in the same transaction) is doable : deletion register the old unique id for pruning and delete the association from the new storage, creation get a new unique id for the new child trie so no delta for pruning is needed (global id for full history with pruning or index from the same storage if only using canonical).

An alternative to get back to #2209 state would be to use storage cache, but it is a bit less flexible (no direct access from statemachine overlay, so it would be temporary code until move or delete operation are really needed), and it is subject to exploit (as his the current implementation) : stores one rocksdb key value per block history as a linked list.

- offchain local storage:

cc\ @todr , iirc there is such local storage in offchain storage api but not yet implemented.
It seems to me this replies the the local offchain storage, it will requires exposing the extrinsic (see TODOs), and probably handling bigger content (see TODOs on indexing and maybe handling file as content). Also maybe adding the per key lock that is in place in offchain storage and not needed in others use cases.

- any technical info that can be different between client, eg current use cases of storage-cache.

- info that should remain identical between client but does not fit a trie storage (we still need to synch some proof on the trie). can even be extended to state storage with lazy proof access (or only a few recalculated trie levels).

# TODOs in this PR

- [x] overlay db handle: push transaction at the end of a block execution
- [x] state-db branches local storage, usage of branch index
- [x] state-db handling tree history of data. Indexing by branch index and block number.
- [x] canonical storage: simple indexing by block number, no need for branch index.
- [x] client testing (canonical): client test should be extended, notably assert the block number send from state-db.
- [x] clean code.

# TODOs for other PR

Those TODOs even if a bit to much to be reviewable in this pr are strictly needed for this storage.
- [ ] lazy pruning key set (no need to keep the set in memory). Additionally we may be pruning to often: since we do not have to maintain a deathrow delta having less frequent pruning can be fine.
- [ ] Genesis storage initialization (may be needed for client test) and 'eset_storage'
- [ ] Serialized implementation scaling (do not store full history in one rocksdb value (obviously don't scale), but implement two strategie:
				- linear: size limite range index and value: linked list of range with possibly included values in range.
				- mmr: range with indexing of range as in mmr allows quick access for latest blocks and reduce significantly access for very big history (eg a block index). Size limits similar to previous impl (may need to allow splitting mmr nodes index too).
				- bench with different size limits.
- [ ] Full history mode (archive all):
				- serialized with history: likely a first indexing by branch index and canonical only indexing behind it.
				- storing association block -> branch index : that was done in one of my previous branch (client-db-ix) where I mistakenly implements branch index management at client level.
				- storing last branch ix at client level.
				- transmitting branchix in commitset of state-db and initiating state-db at right branch index
				- this mode is way less efficient than standard canonical only mode so there could be a plugged canonical only storage mode (same as current) with complement to the full trie indexed one.
- [ ] storage cache: quite needed.

# Optional TODOs

- [ ] Api similar to offchain, add a prefix, have reserved columns for some static prefix (or add both column and prefix).
				- means that result transaction will also be grouped by those.
				- means adding extrinsics to query the storage (such ext will probably be needed soon for testing (but gated behind a feature).

- [ ] full history with pruning: currently all non canonical state is only use in state-db in memory. Implementing full history for this storage means basically that those key do not have to be in memory until canonicalisation. This is not possible currently for state db trie because it will result in key collision. Still a costy approach would be to extend trie library to add the couple branch number block number to the encoded nodes in the underlying db (we already prefix with the node partial path). Then stored branch will becomes [Encoded branch ++ array of child (branch ix + block ix)]. This will not change hash calculation, but would certainly be a bit costy memory wise (two u64 per child, a bit less if we do some small optim like last childs if undefined are same as previous). This could be use along with a canonical compact encoding, but it would involve a pretty heavy cannonicalisation process (rewrite all branch pointer and node index for canonicalized content).

Also this pr could be split for review:
- overlay (all change in state-machine)
- statedb (state-db change with tree historied data)
- client code (client with linear historied data)

This could also be organized in a clean 3 commits history.


burdges:
What gets stored in such a data structure?

cheme:
I am storing a keyspace ids see initialized #3804, I also think it could match some offchain storage needs (at some point there was this unimplemented local storage interface, I do not know if the code was cleaned since cc @tomusdrw ).

I also may want to try storing state with lazy proof calculation or partially lazy, but that is really just an idea at this point.

So key values.

cheme:
Code should be readable, yet it is a bit log. So please tell me if I shall split the PR (as proposed in description).

core/utils/historical-data/README.md
```diff
@@ -0,0 +1,18 @@
...
+Crate with methods to manage data that stores its own history.
+
+This covers:
+- linear history driven data, eg. transactional layers for overlay.
+- long term storage with multiple branch, eg. offchain storage.
+
+General design is container where query and update requires global
+history context.
+
+History is serialize as a per item basis.
```
thiolliere: I'm not sure I understand what that means ?

core/utils/historical-data/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+	pub value: V,
+	/// The moment in history when the value got set.
+	pub index: I,
+}
+
+impl<V, I> From<(V, I)> for HistoricalValue<V, I> {
+	fn from(input: (V, I)) -> HistoricalValue<V, I> {
+		HistoricalValue { value: input.0, index: input.1 }
+	}
+}
```
thiolliere: Usually I prefer when From/Into are used to convert between types, here `(V, I)` is not really a type, implementing a function `new` seems to make more sense here to me.

core/utils/historical-data/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+
+// Utility function for panicking cast (enabling casts similar to `as` cast for number).
+fn as_u<U: num_traits::Bounded, I: TryInto<U>>(i: I) -> U {
```
thiolliere: rename it saturating into

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,447 @@
...
+/// It means that using transaction in a module got a direct allocation cost.
+#[cfg(feature = "std")]
+const ALLOCATED_HISTORY: usize = 2;
+
+#[derive(Debug, Clone)]
+#[cfg_attr(any(test, feature = "test"), derive(PartialEq))]
+/// Arraylike buffer with in place byte data.
+/// Can be written as is in underlying
+/// storage.
+/// Could be extended to direct access memory too.
```
thiolliere: docs should go before other attributes

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,447 @@
...
+/// Currently at two for committed and prospective only.
+/// It means that using transaction in a module got a direct allocation cost.
+#[cfg(feature = "std")]
+const ALLOCATED_HISTORY: usize = 2;
+
+#[derive(Debug, Clone)]
+#[cfg_attr(any(test, feature = "test"), derive(PartialEq))]
+/// Arraylike buffer with in place byte data.
+/// Can be written as is in underlying
+/// storage.
```
thiolliere: you can merge those lines I think

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,447 @@
...
+/// Could be extended to direct access memory too.
+pub struct Serialized<'a, F>(SerializedBuff<'a>, PhantomData<F>);
+
+#[derive(Debug)]
+#[cfg_attr(any(test, feature = "test"), derive(PartialEq))]
+enum SerializedBuff<'a> {
```
thiolliere: doc ?

also can't you just have the variant copy on write and just copy it on writes ?

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,447 @@
...
+use rstd::marker::PhantomData;
+use rstd::borrow::Cow;
+use crate::HistoricalValue;
+
+
+/// Array like buffer for in memory storage.
+/// By in memory we expect that this will
+/// not required persistence and is not serialized.
+#[cfg(not(feature = "std"))]
+pub(crate) type MemoryOnly<V, I> = Vec<HistoricalValue<V, I>>;
```
thiolliere: maybe rename InMemory ?

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,447 @@
...
+			}
+		}
+	}
+}
+
+/// Serialized specific behavior.
+pub trait SerializedConfig {
+	/// encoded empty slice
+	fn empty() -> &'static [u8];
+	/// size at start for encoding version.
```
thiolliere: does that means length used at start of value to encode the version ?

at start of what ?

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,447 @@
...
+impl SerializedConfig for DefaultVersion {
+	fn empty() -> &'static [u8] {
+		&DEFAULT_VERSION_EMPTY_SERIALIZED
+	}
+	fn version_len() -> usize {
+		1
+	}
+}
+
+// encoding size as u64
```
thiolliere: encoded size of what ?

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,447 @@
...
+const SIZE_BYTE_LEN: usize = 8;
+
+// Basis implementation to be on par with implementation using
```
thiolliere:
```suggestion
// Basis implementation must be on par with InMemory
```

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,447 @@
...
+// vec like container. Those method could be move to a trait
+// implementation.
+// Those function requires checked index.
+impl<'a, F: SerializedConfig> Serialized<'a, F> {
+
+	pub fn into_owned(self) -> Serialized<'static, F> {
+    Serialized(SerializedBuff::Cow(Cow::from(self.0.into_owned())), PhantomData)
+  }
```
thiolliere: weird formatting

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,447 @@
...
+	}
+}
+
+// encoding size as u64
+const SIZE_BYTE_LEN: usize = 8;
+
+// Basis implementation to be on par with implementation using
+// vec like container. Those method could be move to a trait
+// implementation.
+// Those function requires checked index.
```
thiolliere: yes and a trait would allow centralized documentation

core/utils/historical-data/src/linear.rs
thiolliere: like `Cow<Vec<u8>>` ?

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,447 @@
...
+#[cfg(feature = "std")]
+const ALLOCATED_HISTORY: usize = 2;
+
+#[derive(Debug, Clone)]
+#[cfg_attr(any(test, feature = "test"), derive(PartialEq))]
+/// Arraylike buffer with in place byte data.
+/// Can be written as is in underlying
+/// storage.
+/// Could be extended to direct access memory too.
+pub struct Serialized<'a, F>(SerializedBuff<'a>, PhantomData<F>);
```
thiolliere: you should say somewhere how it is supposed to be organized, looking at the code to understand how this structure is actually encoded is difficult

core/utils/historical-data/src/linear.rs
cheme: This is indeed looking a bit odd, I am not totally sure if it is possible. In my first implementation it was a simple 'cow<[u8]>', but I probably reach a case where my handle was an '&mut Vec<u8>' and reach the conclusion that I would not want to go through cloning and replacing the buffer afterward. In practice this is only use for InMemory case (direct change of the value in the hashmap), but not with rocksdb.

core/client/db/src/lib.rs
```diff
@@ -537,11 +585,10 @@ impl<Block> client::backend::BlockImportOperation<Block, Blake2Hasher>
...
 	fn update_storage(
 		&mut self,
-		update: StorageCollection,
-		child_update: ChildStorageCollection,
+		update: FullStorageCollection,
 	) -> ClientResult<()> {
-		self.storage_updates = update;
-		self.child_storage_updates = child_update;
+		self.storage_updates = update.top;
+		self.child_storage_updates = update.children;
```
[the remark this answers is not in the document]
cheme: Yes that is part of the tasks I kept out of the pr (in the TODO list for storage_cache management (I always think of this method as the main point for state storage update, but it only really go to the cache update and the rpc)).
Change for kv where put alongside change for state in 'BlockImportOperation' db_update field. Maybe I should use a new field (using a pair helps ensuring I do not have some omission but at this point it could get split).
Thinking twice it could go through 'update_storage', it may be a better place to handle such changes.

core/utils/historical-data/src/linear.rs
thiolliere: isn't the funciton [to_mut](file:///home/thiolliere/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc/rust/html/std/borrow/enum.Cow.html#method.to_mut) doing already the job ?

cheme: Yes, probably need to store some Cow in the InMemory Hashmap then (I do not know a good way to handle hashmap 'get_mut' otherwhise).

cheme:
> some comments, hmm maybe I shouldn't concentrate on the code but just on the general design.

My suggestion to read that is the data lifecycle.
- state machine and overlay change
- state db and fork aware data (historical tree)
- client code (historical linear encoding and pruning)
as stated in the description, I can try to split that in three.
More generally the code is mostly a non trie oriented variant of the trie key val,  so instead of building trie encoded node change set in the statemachine, we only manipulate key value delta.

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,428 @@
...
+
+// Basis implementation must be on par with InMemory.
+// Those method could be move to a 'VecLike' trait.
+//
+// Those function requires prior index checking.
+impl<'a, F: SerializedConfig> Serialized<'a, F> {
+
+	pub fn into_owned(self) -> Serialized<'static, F> {
+		Serialized(Cow::from(self.0.into_owned()), PhantomData)
+	}
```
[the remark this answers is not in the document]
thiolliere: also this function seem never used, nor it is documented, also Serialized implements Clone and the only difference between this and clone is that internally the value is owned instead of borrowed but usre of linear don't access this so from his point of view there is no difference

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,428 @@
...
+  fn eq(&self, other: &Serialized<'b, F>) -> bool {
+		self.0.eq(&other.0)
+	}
+}
+
+impl<'a, F> Eq for Serialized<'a, F> { }
+
+/// Internal buffer, it is either a readonly view other the
+/// serialized data or the pending changes.
+type SerializedBuff<'a> = Cow<'a, [u8]>;
```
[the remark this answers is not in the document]
thiolliere: Actually doesn't seems to have more than Cow semantic, the type renaming is for me more confusing than helpful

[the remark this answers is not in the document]
cheme: yeah, removing the alias could be better, will do.

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,428 @@
...
+	/// variant of push where part of the value is in a second slice.
+	pub(crate) fn push_extra(&mut self, val: HistoricalValue<&[u8], u64>, extra: &[u8]) {
+		let len = self.len();
+		let start_ix = self.index_start();
+		let end_ix = self.0.len();
+		// A sized buffer and multiple index to avoid to big copy
+		// should be use here.
+		let mut new_ix = self.0[start_ix..end_ix].to_vec();
+		// truncate here can be bad
+		self.0.to_mut().truncate(start_ix + SIZE_BYTE_LEN);
```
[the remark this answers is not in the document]
cheme: not to much :), I dropped the comment (just rewrite with copy_within)

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,428 @@
...
+		self.read_le_usize(len - SIZE_BYTE_LEN) as usize
+	}
+
+	pub(crate) fn clear(&mut self) {
+		self.write_le_usize(F::version_len(), 0);
+		self.0.to_mut().truncate(F::version_len() + SIZE_BYTE_LEN);
+	}
+
+	#[cfg(test)]
+	fn truncate(&mut self, index: usize) {
```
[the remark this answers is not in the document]
cheme: The truncate until is a bit more costy as it is handling non terminal removal and need to rewrite the indexes.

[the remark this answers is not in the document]
thiolliere: yes but truncate is used for test so that is not important you can still reuse it with comment in case someone want finally to use in production but I would prefer avoiding as much unecessary code as possible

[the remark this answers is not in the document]
cheme: true :+1:

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,417 @@
...
+		let state = self.read_le_u64(start_ix);
+		HistoricalValue {
+			value: &self.0[start_ix + SIZE_BYTE_LEN..end_ix],
+			index: state,
+		}
+	}
+
+}
+
+const EMPTY_SERIALIZED: [u8; SIZE_BYTE_LEN] = [0u8; SIZE_BYTE_LEN];
```
thiolliere: maybe rename NO_VERSION_EMPTY_SERIALIZED

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,417 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Linear historical data.
+//!
+//! Current encoding is a single encoded succession of values and
+//! their history index (version 1 is used for it).
```
thiolliere: on time you say state index another time you say history index. actually I prefer using state index in the whole file.

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,417 @@
...
+//! The frame for n elements is:
+//!
+//! `1 byte version ++ (u64 le encoded state index ++ byte value of element) * n
+//! ++ (u64 le encoded index of element) * n - 1 ++ n encoded as le u64`
```
thiolliere:
```suggestion
//! ++ (u64 le encoded index of element in the frame) * n - 1 ++ n encoded as le u64`
```

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,417 @@
...
+impl<F> Into<Serialized<'static, F>> for Vec<u8> {
+	fn into(self) -> Serialized<'static, F> {
+		Serialized(Cow::Owned(self), PhantomData)
+	}
+}
+
+// Utility function for basis implementation.
+impl<'a, F: SerializedConfig> Serialized<'a, F> {
+	
+	// Index at end, also contains the encoded size
```
thiolliere:
```suggestion
	// Index at the end of the element part, start of part with the index for each element in the buffer. (contains the encoded size, as it is the last part in the buffer)
```

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,417 @@
...
+		}
+		let i = self.index_start() + (position - 1) * SIZE_BYTE_LEN;
+		self.read_le_usize(i)
+	}
+
+	fn slice_copy(&mut self, start_from: usize, start_to: usize, size: usize) {
+		self.0.to_mut().copy_within(start_from..start_from + size, start_to);
+	}
+
+	// Usize encoded as le u64 (for historical value).
```
thiolliere:
```suggestion
	// Read u64 encoded as le (for history index).
```

what do you mean by Usize encoded as le u64, history index is u64 itself

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,417 @@
...
+	fn read_le_u64(&self, pos: usize) -> u64 {
+		let mut buffer = [0u8; SIZE_BYTE_LEN];
+		buffer.copy_from_slice(&self.0[pos..pos + SIZE_BYTE_LEN]);
+		u64::from_le_bytes(buffer)
+	}
+
+	// Usize encoded as le u64 (only for internal indexing).
```
thiolliere:
```suggestion
	// Read usize encoded as le u64 (only for internal indexing).
```

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,417 @@
...
+	fn read_le_usize(&self, pos: usize) -> usize {
+		let mut buffer = [0u8; SIZE_BYTE_LEN];
+		buffer.copy_from_slice(&self.0[pos..pos + SIZE_BYTE_LEN]);
+		u64::from_le_bytes(buffer) as usize
+	}
+
+	// Usize encoded as le u64.
```
thiolliere:
```suggestion
	// Write usize encoded as le u64.
```

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,417 @@
...
+	fn write_le_usize(&mut self, pos: usize, value: usize) {
+		let buffer = (value as u64).to_le_bytes();
+		self.0.to_mut()[pos..pos + SIZE_BYTE_LEN].copy_from_slice(&buffer[..]);
+	}
+
+	// Usize encoded as le u64.
```
thiolliere:
```suggestion
	// Append usize encoded as le u64.
```

core/utils/historical-data/src/linear.rs
```diff
@@ -0,0 +1,417 @@
...
+	fn append_le_usize(&mut self, value: usize) {
+		let buffer = (value as u64).to_le_bytes();
+		self.0.to_mut().extend_from_slice(&buffer[..]);
+	}
+
+	// Usize encoded as le u64.
```
thiolliere:
```suggestion
	// Write u64 encoded as le.
```

core/state-machine/src/backend.rs
```diff
@@ -250,7 +321,8 @@ impl error::Error for Void {
 /// tests.
 pub struct InMemory<H: Hasher> {
 	inner: HashMap<Option<Vec<u8>>, HashMap<Vec<u8>, Vec<u8>>>,
```
thiolliere: must be renamed trie_inner or trie

core/state-machine/src/backend.rs
```diff
@@ -250,7 +321,8 @@ impl error::Error for Void {
...
-	trie: Option<TrieBackend<MemoryDB<H>, H>>,
+	kv: Option<InMemoryKvBackend>,
```
thiolliere: must be renamed kv_inner or kv

core/state-machine/src/backend.rs
```diff
@@ -250,7 +321,8 @@ impl error::Error for Void {
...
+	state: Option<StateBackend<MemoryDB<H>, H, InMemoryKvBackend>>,
```
thiolliere: once you merge master you could renamed it with a more obvious name like as_state_backend_place_holder but I can't really find a good name, but at least the comment

core/state-machine/src/backend.rs
```diff
@@ -283,29 +357,66 @@ impl<H: Hasher> Clone for InMemory<H> {
...
 impl<H: Hasher> InMemory<H> {
+	/// Key value inmemory storage is either used from the
+	/// state db or its own field.
+	/// This function access the right field.
+	fn kv(&self) -> &InMemoryKvBackend {
+		if let Some(kv) = self.kv.as_ref() {
+			kv
+		} else {
+			self.state.as_ref().unwrap().kv_backend()
+		}
```
thiolliere: this else condition doesn't make sense, state is only used in the function for returning the immutable reference in as_state_backend.

core/state-machine/src/backend.rs
```diff
@@ -283,29 +357,66 @@ impl<H: Hasher> Clone for InMemory<H> {
...
+	}
+
+	/// Access the current key value db in a mutable way.
+	fn kv_mut(&mut self) -> &mut InMemoryKvBackend {
+		if let Some(kv) = self.kv.as_mut() {
+			kv
+		} else {
+			self.state.as_mut().unwrap().kv_backend_mut()
+		}
```
thiolliere: same useless else condition

core/state-machine/src/backend.rs
```diff
@@ -283,29 +357,66 @@ impl<H: Hasher> Clone for InMemory<H> {
...
+	/// Extract key values from `InMemory`, by emptying it.
+	fn extract_kv(&mut self) -> InMemoryKvBackend {
+		if let Some(kv) = self.kv.take() {
+			kv
+		} else {
+			std::mem::replace(
+				self.state.as_mut().unwrap().kv_backend_mut(),
+				Default::default(),
+			)	
+		}
```
thiolliere: same remove else condition

core/state-machine/src/backend.rs
```diff
@@ -283,29 +357,66 @@ impl<H: Hasher> Clone for InMemory<H> {
...
 	}
 }
 
 impl<H: Hasher> From<HashMap<Option<Vec<u8>>, HashMap<Vec<u8>, Vec<u8>>>> for InMemory<H> {
 	fn from(inner: HashMap<Option<Vec<u8>>, HashMap<Vec<u8>, Vec<u8>>>) -> Self {
 		InMemory {
 			inner: inner,
-			trie: None,
+			state: None,
+			kv: Some(Default::default()),
```
thiolliere: All this from sounds now a bit weird to me because it is not obvious if it should make an InMemory with the kv storage or without the kv storage. usually I prefer a new function that makes it obvious. However I'm ok with this. But if we really want to make kv optional then add a doc in the from implementation stating that kv is activated

core/client/db/src/lib.rs
```diff
@@ -448,7 +492,7 @@ impl<Block: BlockT> HeaderMetadata<Block> for BlockchainDb<Block> {
 /// Database transaction
 pub struct BlockImportOperation<Block: BlockT, H: Hasher> {
 	old_state: CachingState<Blake2Hasher, RefTrackingState<Block>, Block>,
-	db_updates: PrefixedMemoryDB<H>,
+	db_updates: (PrefixedMemoryDB<H>, InMemoryKvBackend),
```
thiolliere: maybe it is better to avoid tuple here ? you could do trie_db_update and kv_db_update.
or maybe create a new struct ?

core/state-machine/src/backend.rs
```diff
@@ -118,25 +140,30 @@ pub trait Backend<H: Hasher>: std::fmt::Debug {
...
+		&StateBackend<Self::TrieBackendStorage, H, Self::KvBackend>
+	> {
 		None
 	}
 
 	/// Calculate the storage root, with given delta over what is already stored
 	/// in the backend, and produce a "transaction" that can be used to commit.
 	/// Does include child storage updates.
-	fn full_storage_root<I1, I2i, I2>(
+	fn full_storage_root<I1, I2i, I2, I3>(
```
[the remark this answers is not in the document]
cheme: Yes kv does not influe in the root, it is mainly here to complete the transaction.
Ideally we could have a function that does only the root: this would be good from my point of view (there is quite a few place where we do not use the produced transaction).

core/state-machine/src/backend.rs
cheme: I remember why I use it this way, it was simply to avoid copy when producing the 'as_state_backend' : instead of cloning the values as we do for the trie part it is moving the container.
So after using `as_state_backend` this is getting use instead of the kv field. I could probably switch to simply cloning the `kv` field.

core/state-machine/src/backend.rs
cheme: Yes, it is only to avoid cloning it, would probably be better with a box. But yes the option here does not have meaning, in current state you need either kv being Some or state being Some.
But yes I could probably just clone.

core/client/db/src/lib.rs
cheme: Yes, that is a bit of refactoring, but those key value pr should try to get as decoupled as possible from the trie state (the opposite of what I did).
That would be a big but probably needed refactoring, I am switching those PR state to in_progress thus.

core/state-machine/src/backend.rs
thiolliere: ok this is correct indeed

thiolliere: would be good to have some comment about this complex behavior

core/client/db/src/lib.rs
```diff
@@ -1288,7 +1423,75 @@ impl<Block: BlockT<Hash=H256>> Backend<Block> {
...
+) -> Result<(), io::Error> {
+
+	for (key, change) in commit.kv.iter() {
+		let (mut ser, new) = if let Some(stored) = db.get(columns::KV, key)? {
+			(Ser::from_vec(stored.to_vec()), false)
+		} else {
+			if change.is_some() {
+				(Ser::default(), true)
+			} else {
+				break;
```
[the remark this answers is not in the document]
cheme: yes it is the case of deleting a value that is not here.

cheme:

PR #3775, opened 2019-10-07
Move `Externalities` into its own crate

bkchr:
Instead of having the `Externalities` trait defined in `primitives`, this pr moves it into the `substrate-externalities` crate. This makes it possible to use `Externalities` without depending on `primitives`. In the given pr that makes no big difference, but will be required by some later pr of me. Besides moving the trait, the following was done:
- `Externalities` scoping was moved as well into the new crate, from `sr-io`. It makes sense to have the stuff in one place.
- We don't expose the `KeyStore` or the offchain `Externalities` anymore in the `Externalities` directly. Instead the `Externalities` now support extensions and the keystore and offchain ext are registered as extensions when they are available. This extensions feature is currently internal, but could be extended in the future to support adding custom extensions, but more about this topic when we have custom runtime interfaces.
- The hasher is dropped from the `Externalities` trait. Having the hasher in the trait does not makes any sense at all and up to this point is was hardcoded as `Blake2Hasher` anyway. If we get to the point, where we want to make this fully generic, I'm willing to refactor this again ;) But currently, it just makes stuff more complicated.
- Crated a new crate `primitives-storage` for storage related primitives. This is just a sub-crate of `primitives` which is exported by it. It just serves the purpose of not depending on the complete `primitives` crate :D  

Besides these changes, most of the stuff is just code adaption to the said changes.


core/externalities/src/extensions.rs
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	/// Register the given extension.
+	pub fn register<E: Any>(&mut self, ext: E) {
+		self.extensions.insert(ext.type_id(), Box::new(ext));
+	}
+
+	/// Return a mutable reference to the requested extension.
+	pub fn get_mut(&mut self, ext_type_id: TypeId) -> Option<&mut dyn Any> {
+		self.extensions.get_mut(&ext_type_id).map(DerefMut::deref_mut)
```
tomusdrw: Why expose `Any` instead of downcasting here?

core/externalities/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+	///
+	/// This function should not be used directly and instead [`Externalities::extension`] should
+	/// be used.
+	fn extension_by_type_id(&mut self, type_id: TypeId) -> Option<&mut dyn Any>;
+}
+
+/// Extension for the [`Externalities`] trait.
+pub trait ExternalitiesExt {
+	/// Tries to find a registered extension and returns a mutable reference.
+	fn extension<T: Any>(&mut self) -> Option<&mut T>;
```
tomusdrw: While the API seems nice, the main issue I see (that I have considered for `ChainSpec` extensions as well) is that because the API is not properly typed it's really easy to register a different extension than one actually wants to use further down i.e. someone wraps `OffchainExt`in `Arc` and registers this, while other parts of code expect `OffchainExt`.
This kind of integration is also usually not very well tested, so it might be hard to spot a bug.

That said, I'm not opposing it as I don't really have a better solution to this problem, just noting that we have to be careful with this (maybe additional docs on the structs that are being passed as extensions or even a requirement of some marker trait).

core/externalities/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+	/// If the storage root equals the default hash as defined by the trie, the key in the top-level
+	/// storage map will be removed.
+	fn child_storage_root(&mut self, storage_key: ChildStorageKey) -> Vec<u8>;
+
+	/// Get the change trie root of the current storage overlay at a block with given parent.
+	fn storage_changes_root(&mut self, parent: H256) -> Result<Option<H256>, ()>;
+
+	/// Tries to find a registered extension by the given `type_id` and returns it as a `&mut dyn Any`.
+	///
+	/// This function should not be used directly and instead [`Externalities::extension`] should
```
tomusdrw: If it should not be used directly might be better to move it to some other trait then (like `ExtensionStore`) and implement `ExternalitiesExt for Externalities + ExtensionStore`.
That way people will not be confused by it.

core/state-machine/src/ext.rs
```diff
@@ -152,15 +143,31 @@ where
 		self.storage_transaction = None;
 	}
 
+	/// Register the given extension for this instance.
+	pub fn register_extension<E: Any + Sized>(&mut self, ext: E) {
+		println!("REGISTER: {:?} {}", ext.type_id(), std::any::type_name::<E>());
```
tomusdrw: stray `println`? Maybe use `info!`?

core/state-machine/src/ext.rs
```diff
@@ -482,31 +519,29 @@ where	H: Hasher,
...
 	}
 
-	fn offchain(&mut self) -> Option<&mut dyn offchain::Externalities> {
-		self.offchain_externalities.as_mut().map(|x| &mut **x as _)
-	}
-
-	fn keystore(&self) -> Option<BareCryptoStorePtr> {
-		self.keystore.clone()
+	fn extension_by_type_id(&mut self, type_id: TypeId) -> Option<&mut dyn Any> {
+		println!("REQUEST: {:?}", type_id);
```
tomusdrw: Same

core/state-machine/src/ext.rs
```diff
@@ -152,15 +143,31 @@ where
...
 	}
 
+	/// Register the given extension for this instance.
+	pub fn register_extension<E: Any + Sized>(&mut self, ext: E) {
+		println!("REGISTER: {:?} {}", ext.type_id(), std::any::type_name::<E>());
+		self.extensions.register(ext);
+	}
+
+	/// Reset the internal state.
+	pub fn reset(&mut self) {
```
tomusdrw: That might be really flaky, can we add some documentation near the field declaration that you must reset the internal state here? To avoid situation in future, where people add/alter fields and forget that they need to be re-set.

tomusdrw: BTW why is that even better than creating externalities from scratch every time?

And also what's the logic behind WHAT should be cleared? Seems things like `overlay` are not. Maybe we can choose a better name for this? Seems it's only for randomness and transaction data?

core/state-machine/src/lib.rs
```diff
@@ -167,48 +163,52 @@ fn always_untrusted_wasm<E, R: Decode>() -> ExecutionManager<DefaultHandler<R, E
...
+		let mut ext = ext::Ext::new(
+			overlay,
 			backend,
 			changes_trie_storage,
-			offchain_ext,
-			overlay,
+		);
+
+		if let Some(keystore) = keystore {
+			ext.register_extension(keystore);
```
tomusdrw: Hmm, I see that persistent `Ext` is now needed to prevent the extensions from being dropped? Maybe we can extract the `ExtensionStore` then and just pass it to externalities?

core/state-machine/src/ext.rs
```diff
@@ -152,15 +143,31 @@ where
...
+		self.storage_transaction = None;
+		self.changes_trie_transaction = None;
+		self.id = rand::random();
+	}
+
+	/// Returns a mutable reference to the overlayed changes.
+	pub fn mut_overlayed_changes(&mut self) -> &mut OverlayedChanges {
```
tomusdrw: Isn't the convention usually to add `mut` at the end? i.e. `overlayed_changes_mut`?

core/state-machine/src/ext.rs
bkchr: Ohh, debugging letfovers :sweat_smile:

core/externalities/src/extensions.rs
bkchr: `Externalities` can just expose a function that returns an any, as we can not have a function that takes a generic. Otherwise the trait could not be made into an object.

tomusdrw: Ah, right. Makes sense.

core/sr-primitives/src/lib.rs
```diff
@@ -67,6 +67,9 @@ pub use sr_arithmetic::{
 /// Re-export 128 bit helpers from sr_arithmetic
 pub use sr_arithmetic::helpers_128bit;
 
+#[cfg(feature = "std")]
+pub use externalities::set_and_run_with_externalities;
```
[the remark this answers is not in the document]
bkchr: Why? As you see, it is used in almost every srml crate for testing. It makes more sense to re-export it from here then from `sr-io` imho.

core/externalities/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+	/// be used.
+	fn extension_by_type_id(&mut self, type_id: TypeId) -> Option<&mut dyn Any>;
+}
+
+/// Extension for the [`Externalities`] trait.
+pub trait ExternalitiesExt {
```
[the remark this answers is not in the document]
bkchr: https://github.com/paritytech/substrate/pull/3775#discussion_r331968360

core/client/src/client.rs
```diff
@@ -1058,18 +1058,24 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
-					},
-					None,
-					NeverOffchainExt::new(),
-				)?;
+				let (_, storage_update, changes_update) = self.executor
+					.call_at_state::<_, _, NeverNativeValue, fn() -> _>(
+						transaction_state,
+						&mut overlay,
+						"Core_execute_block",
+						&<Block as BlockT>::new(
```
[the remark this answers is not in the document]
tomusdrw: Yeah, would definitely help with readability.

core/client/src/client.rs
```diff
@@ -1460,12 +1466,14 @@ impl<B, E, Block, RA> CallRuntimeAt<Block> for Client<B, E, Block, RA> where
...
 		let capabilities = context.capabilities();
-		let mut offchain_extensions = match context {
-			ExecutionContext::OffchainCall(ext) => ext.map(|x| x.0),
+		let offchain_extensions = match context {
+			ExecutionContext::OffchainCall(Some(ext)) => Some(
+				OffchainExt::new(offchain::LimitedExternalities::new(capabilities, ext.0))
+			),
 			_ => None,
-		}.map(|ext| offchain::LimitedExternalities::new(capabilities, ext));
+		};
```
tomusdrw: nit:
```suggestion
let offchain_extensions = if let ExecutionContext::OffchainCall(Some(ext)) = context {
	Some(OffchainExt::new(offchain::LimitedExternalities::new(capabilities, ext.0)))
} else {
	None
};
```

core/externalities/src/extensions.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+impl Extensions {
+	/// Create new instance of `Self`.
+	pub fn new() -> Self {
+		Self::default()
+	}
+
+	/// Register the given extension.
+	pub fn register<E: Any + Extension>(&mut self, ext: E) {
```
tomusdrw: I think `Any` can be dropped here, no?
```suggestion
	pub fn register<E: Extension>(&mut self, ext: E) {
```

bkchr: No, I call `type_id()` and also would like to keep it. To make sure we have the correct bounds. I don't see any problem in using it.

bkchr:
> This new separation between externalities and primitives still seems pretty weird and arbitrary to me. With Externalities being re-exported from primitives and `CodeExecutor` still living there, I don't see why it needs to be in its own crate, but I guess I'll wait to see why in the subsequent PRs.

I need `Externalities` to not live in `primtives`, as I will need it to use them from a different crate, which then will be imported by `primtivies` as well. And a crate can not depend on its own, so I need to pull stuff out, to make this possible. But you are right, you will see this in the subsequent pr.

`CodeExecutor` accepts something that implements `Externalities`, but that is all. Why should `CodeExecutor` live in the `externalities` crate? This is just some trait that can live in primitives.

PR #3667, opened 2019-09-23
Add hardware sr25519 signer to subkey

chesterliliang:
This is a prototype of how subkey calling a hardware to sign data or transfer with sr25519. Gavin will take the device back for test when he returns from China. 
I am not sure if it is appropriate to ask a new branch for this work. But if subkey wants work with kinds of signer hardwares, it may be a good start.

[clabot:check]

Issue #3396, opened 2019-08-14
SURI HD derivation confuses `ChainCode` with index `i`


maciejhirsz:
Disclaimer: I'm not a cryptographer, talked to @burdges about it to make sure I'm reading stuff right. Jeff can fill in details if I missed anything.

Current [sr25519 derivation code](https://github.com/paritytech/substrate/blob/a622361559b31f1eb9eb8ce222ad6f94bf4b79c4/core/primitives/src/sr25519.rs#L469-L476) feeds `DeriveJunction`s as `ChainCode`s to soft/hard derivation methods, using empty byte arrays for `i` in both cases. This seems to be fine for what we are doing (both `ChainCode` and `i` end up in the same hash anyway), but is not the intended use of the API.

The consequences are that we are missing the extra entropy from the `ChainCode` product of previous key expansion in subsequent derivations. This is probably fine as we still have complete entropy of the original secret key, and my understanding after talking to Jeff as of why the extra entropy exists in BIP32 in the first place is that nobody knows :P.

TL;DR: We are feeding derivation junctions as chain codes instead of `i`, and throwing away chain codes from previous expansions, which is probably fine.

burdges:
Right `i` should come from the derivation path, not the `ChainCode`.

I'll first note that [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#extended-keys) is ambiguous about the `ChainCode`'s real function, only that it "prevent [derived keys] from depending solely on the [parent key]".  

It's quite unambiguous that `ChainCode`s should be machine generated bytes though, meaning they should not be user controlled parts of the derivation path.

I'm not so worried about lacking the extra entropy from the `ChainCode`.  In fact, our starting secret key has 252 bits of entropy, more than enough.  In theory, it only has 128 bits of security since the public key is known, but again enough.  I should probably decided if the chain code should be optional or not too.

Also, we take system randomness when creating derived nonces in https://github.com/w3f/schnorrkel/blob/master/src/derive.rs#L185 although this does not work when doing hard derivations since we do not pass them a `RngCore`.

In fact, I think `ChainCode` should really be viewed as de facto hard derivations that can be turned into soft derivations later by sharing the `ChainCode`.   Is this functionality useful?  I donno, but maybe by someone somewhere.

[...]

PR #3777, opened 2019-10-07
Wookong hardware wallet support

gavofyork:
Rework (mostly a rewrite) of #3667.

### DIFF
--- a/subkey/subhd/src/lib.rs
+++ b/subkey/subhd/src/lib.rs
@@ -1,108 +1,106 @@
-// -*- mode: rust; -*-
-//
-// This file is part of subhd, substrate.
 // Copyright (c) 2017-2019 Chester Li and extropies.com
+// Copyright 2018-2019 Parity Technologies (UK) Ltd.
+// This file is part of Substrate/subhd.
+
+// Substrate is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 // See LICENSE for licensing information.
 //
-// Authors:
+// Original author:
 // - Chester Li<chester@lichester.com>
 
-//! hd wallet layer for subkey
-extern crate wookong_solo;
-extern crate hex_view;
-use hex_view::HexView;
+//! Rust bindings for Wookong hardware wallet.
 
-use primitives::{sr25519, sr25519::{ Signature,Public}};
+use primitives::{
+	sr25519::{self, Signature, Public, Pair}, crypto::{DeriveJunction, Pair as TraitPair}
+};
 use wookong_solo::{wk_getpub, wk_sign, wk_generate, wk_format, wk_import};
+
 /// Error define
-/// DeviceNotfound: device not connected
-/// DeviceNotInit: empty device, generate or import first
-/// DeviceError: general error of command
-/// TODO: give detail error
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
-pub enum HDWalletError {
-    DeviceNotfound,
-    DeviceNotInit,
-    DeviceError
-}
-/// seed type
-type Seed = [u8; 32];
-/// subhd Result
-pub type HDWalleResult<T> = Result<T, HDWalletError>;
-/// struct device
-pub struct HDDevice;
-/// method of subhd
-pub trait HDwallet{
-    /// get public key
-    /// Address in defualt return
-    /// .into make it into public key
-    fn hd_getpub(&self) ->  HDWalleResult<Public>;
-    /// sign the message
-    fn hd_sign(&self, message:&[u8]) -> HDWalleResult<Signature>;
-    /// generate keypair and return seed
-    fn hd_generate(&self) -> HDWalleResult<Seed>;
-    /// import seed
-    fn hd_import(&self, seed:&[u8]) -> HDWalleResult<()>;
-    /// clear the device to empty
-    fn hd_format(&self) -> HDWalleResult<()>;
-    /// format seed hex string
-    fn to_hex(&self, data:&[u8]) -> String;
+pub enum Error {
+	/// Device not connected.
+	DeviceNotfound,
+	/// Empty device; key must be generated or imported first.
+	DeviceNotInit,
+	/// General error of command.
+	DeviceError,
 }
 
-/// implement of methods
-impl HDwallet for HDDevice {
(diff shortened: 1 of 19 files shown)


gavofyork:
@chesterliliang i'm having trouble getting this to work on linux. are there specific drivers needed? here's the output:

```
gav@pop-os:~/Core/substrate$ sudo ./target/release/subkey --wookong inspect
[sudo] password for gav: 
U��1---------init--g_dev=0------
                                ----------!multi part!------------
                                                                  ----------!multi part!------------
                                                                                                    ----------!multi part!------------
              ----------!multi part!------------
                                                ----------!multi part!------------
                                                                                  ----------!multi part!------------
                                                                                                                    Public Key URI `` is account:
                           Network ID/version: substrate
                                                          Public key (hex): 0x0000000000000000000000000000000000000000000000000000000000000000
                        Address (SS58): 5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM
```

note that it leave the terminal in a broken state after running, too.

chesterliliang:
@gavofyork Linux should have CP2102 driver.  The port could be listed by 'mesg | grep ttyS*' 
Here is the C code to init uart in solo crate
```
int uart_init(int *dev)
{
    int fd; 
#ifdef __linux__git
    fd = open("/dev/ttyUSB0", O_RDWR | O_NOCTTY | O_NONBLOCK); 
#elif TARGET_OS_MAC
    fd = open("/dev/tty.SLAB_USBtoUART", O_RDWR | O_NOCTTY | O_NONBLOCK); 
#else
#   //error "Unsupport platform for uart"
#endif
```
I will try to figure out why.

gavofyork:
it looks a bit like it's opening & writing to the standard i/o instead of the UART.

gavofyork:
```
gav@pop-os:~/Core/substrate/scripts$ mesg | grep ttyS*
gav@pop-os:~/Core/substrate/scripts$ 
```

chesterliliang:
Device not in the list.. well, do you use ubuntu 18.x?  And just for a remind, plz plug the usb connector to the blue board.  I put it into another one sometimes..

gavofyork:
ubuntu 19.04

chesterliliang:
Here is a reference to see if the device works on linux:
https://www.silabs.com/community/interface/knowledge-base.entry.html/2018/07/31/how_to_check_if_the-5tUc
and try to install the driver may help:
https://www.silabs.com/products/development-tools/software/usb-to-uart-bridge-vcp-drivers

The product will switch to USB-HID finally, so it will not need driver for every platform. I thought UART is simple enough for early use but it seems it's not ture. T_T

chesterliliang:
@gavofyork is your 19.04 the desktop or server version?

gavofyork:
desktop; i'm back to working on macos and it's ok again

subkey/subhd/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+	fn reset(&self) -> Result<()> { Err(Error::DeviceNotFound) }
+}
+
+/// Unit type representing the Wookong hardware wallet.
+pub struct Wookong;
+
+impl Wallet for Wookong {
+	type Pair = sr25519::Pair;
+	fn derive_public(&self, _path: &[DeriveJunction]) -> Result<Public> {
+		// TODO: pass through _path and _password to mutate the key on the device accordingly.
```
gavofyork: @chesterliliang this needs implementing

subkey/subhd/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+		} else if rv != 0 {
+			return Err(Error::DeviceNotFound);
+		} else if rv == 0 {
+			Ok(sr25519::Public::from_raw(pk))
+		} else {
+			return Err(Error::DeviceError);
+		}
+	}
+	fn sign(&self, _path: &[DeriveJunction], message: &[u8]) -> Result<Signature> {
+		// TODO: pass through _path and _password to mutate the key on the device accordingly.
```
gavofyork: @chesterliliang this needs implementing

subkey/subhd/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+	type Pair = T;
+	fn derive_public(&self, _: &[DeriveJunction]) -> Result<<Self::Pair as TraitPair>::Public> { Err(Error::DeviceNotFound) }
+	fn sign(&self, _: &[DeriveJunction], _: &[u8]) -> Result<<Self::Pair as TraitPair>::Signature> { Err(Error::DeviceNotFound) }
+	fn generate(&self) -> Result<Self::Pair> { Err(Error::DeviceNotFound) }
+	fn import(&self, _: &<Self::Pair as TraitPair>::Seed) -> Result<()> { Err(Error::DeviceNotFound) }
+	fn reset(&self) -> Result<()> { Err(Error::DeviceNotFound) }
+}
+
+/// Unit type representing the Wookong hardware wallet.
+pub struct Wookong;
```
gavofyork: @chesterliliang if the device can be made to support multiple root keys (which would be nice, i think), then this could be changed thus:

```

subkey/subhd/src/lib.rs
chesterliliang: Should we verify password as well when do the sign?

subkey/subhd/src/lib.rs
chesterliliang: If they are root keys(seed? before derive), we could find them by index, it seems hardware can not know which one is the right by account.

subkey/subhd/src/lib.rs
gavofyork: yeah - sorry the `_password` is old and doesn't need to be implemented at all.

subkey/subhd/src/lib.rs
gavofyork: no - can ignore `_password` - it's part of an old design and not relevant any more.

subkey/subhd/src/lib.rs
gavofyork: hmm, then they can be listed which isn't so great for privacy. couldn't it just store the public keys along with the roots and cycle through those that it know until it either finds the correct one or returns some `KeyNotFound` error?

chesterliliang: It's possible but must under the same derive path or same rule to manage the path. Seems not a big trouble. Will try to give an implementation for further discussion.

burdges:
Just fyi @chesterliliang we'll eventually "repair" the missuse of i and the chain code by SURI.  See https://github.com/w3f/schnorrkel/issues/45 and https://github.com/paritytech/substrate/issues/3396  We'll play some tricks to keep all current derivations the same, which requires a coordinated upgrade of both subkey and schnorrkel, but I wanted to warn you. :)

chesterliliang:
@burdges Thanks Jeff.  Don't worry about us since we just start to work on derive part.  Could you plz provide one, just one test vector if the issue is fixed?

burdges:
We'll avoid currently allowable test vectors changing.  Yet, subkey will continue to not use chaincodes for the questionable purpose for which bip32 added them, so maybe some future wallets will use it in incompatible ways.

gnunicorn:
is this still being worked on?

chesterliliang:
No... Team has financial problem.. plz close it...

PR #3778, opened 2019-10-07
support crypto primitives for no_std introducing `full_crypto` feature

brenzi:
This adds a new compiler feature `full_crypto` to
* substrate-primitives
* substrate-application-crypto

# Why
The upstream logic knows only two use cases: 
* within a `std` environment -> use crypto primitives like `sr25519::Pair`
* within wasm, a `no_std` environment with the goal of yielding lightweight binaries -> use host calls for crypto primitives

As we use substrate-primitives within an SGX enclave to sign extrinsics, we're actually adding a third use case:
* within a `no_std` environment but needing crypto primitives -> no solution so far.

So far, we just patched cargo dependencies to our own fork of these substrate crates, but this is hard to maintain and we think others might benefit from our generalization.

# CI
In order to make sure we can test this new feature automatically, we wrote a minimal crate:
https://github.com/scs/test-no-std/tree/substrate-primitives

unfortunately, the linker still fails, but as long as there is no `duplicate lang item`error, the build is to be considered a success.


burdges:
Is rand appearing here?  We've reached rand 0.7 by now? 

As I've noted before https://github.com/w3f/schnorrkel/issues/31#issuecomment-520213230 I think rand_os activates std before 0.2.1, so before rand 0.7.  We avoid `OsRng`, `ThreadRng`, etc. on chain, but require roughly `OsRng` for substraTEE, etc. 

We'll want `RngCore5As4` from https://github.com/w3f/schnorrkel/commit/8150ef6333df6238d8d861aec74bf7ccd87119aa if we want both ed25519-dalek and rand_core 0.5 (rand 0.7).  Are cargo features an issue?  I think not..

brenzi:
@burdges: What we need for substraTEE are: signing & verifying and hashing. For random we use Intel SGX features. 
This also means we're not actually using Schnorrkel ourselves within the enclave, because - in contrast to ed25519 - sr25519 seems to require random for deriving a keypair from a seed. So we want to verify sr25519 signatures, but for signing we use an ed25519 pair.

I'm not sure I understand your remarks. `no_std` works for us in the enclave with sr25519 and ed25519. However, because of a cargo bug, we still need to patch to avoid `duplicate lang item std`:
```
[patch.crates-io]
ed25519-dalek = { git = "https://github.com/scs/ed25519-dalek.git", branch = "no_std_sgx"}
```

brenzi:
added CI comments ^^^^

burdges:
> sr25519 seems to require random for deriving a keypair from a seed. 

We do not require randomness when expanding a keypair from a seed:
https://docs.rs/schnorrkel/0.1.1/src/schnorrkel/keys.rs.html#114

We do use randomness when doing HDKD in https://docs.rs/schnorrkel/0.1.1/src/schnorrkel/derive.rs.html#177 which I should probably replace with a `t.witness_bytes` call so that you can replace `thread_rng`  using `attach_rng`.  I tried to write the code more semantically once upon a time, but this'll make more sense now I think.  Thanks!

brenzi:
@burdges: thanks for your helpful hints. This means we actually might use sr25519 keys for our enclave.

However: I consider this off-topic for this PR.

brenzi:
@bkchr, @kianenigma: This PR just got flagged "in progress". While it's true that it is not ready for merging/review, I'd like to ask you for a statement if you are willing to accept our added use case and the approach of this PR

gavofyork:
yeah; looks reasonable to me.

core/primitives/src/ed25519.rs
```diff
@@ -152,13 +153,20 @@ impl<'de> Deserialize<'de> for Public {
...
 
-#[cfg(feature = "std")]
+#[cfg(all(feature = "with_crypto", feature = "std"))]
 impl std::hash::Hash for Public {
 	fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
 		self.0.hash(state);
 	}
 }
 
+#[cfg(all(feature = "with_crypto", not(feature = "std")))]
```
bkchr: Duplicate.

core/primitives/src/ed25519.rs
```diff
@@ -236,13 +244,21 @@ impl std::fmt::Debug for Signature {
...
 
-#[cfg(feature = "std")]
+#[cfg(all(feature = "with_crypto", feature = "std"))]
 impl std::hash::Hash for Signature {
 	fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
 		std::hash::Hash::hash(&self.0[..], state);
 	}
 }
 
+#[cfg(all(feature = "with_crypto", not(feature = "std")))]
```
bkchr: Duplicate impl.

bkchr: (Imprort it from `rstd`)

core/primitives/src/crypto.rs
```diff
@@ -18,6 +18,8 @@
 //! Cryptographic utilities.
 // end::description[]
 
+use rstd::vec::Vec;
+use core::hash::Hash;
```
bkchr:
```suggestion
```

core/primitives/src/crypto.rs
```diff
@@ -18,6 +18,8 @@
 //! Cryptographic utilities.
 // end::description[]
 
+use rstd::vec::Vec;
```
bkchr:
```suggestion
use rstd::{vec::Vec, hash::Hash};
```

core/primitives/src/crypto.rs
```diff
@@ -254,9 +255,10 @@ pub enum PublicError {
 }
 
 /// Key that can be encoded to/from SS58.
-#[cfg(feature = "std")]
+#[cfg(feature = "with_crypto")]
 pub trait Ss58Codec: Sized {
 	/// Some if the string is a properly encoded SS58Check address.
+	#[cfg(feature = "std")]
```
bkchr: Why is this only available with `std` feature?

core/primitives/Cargo.toml
```diff
@@ -67,20 +68,27 @@ std = [
...
 	"sha2",
-	"schnorrkel",
+	"schnorrkel/std",
 	"regex",
 	"num-traits/std",
 	"externalities",
 	"primitives-storage/std",
 ]
+
+with_crypto = [
```
bkchr: I think I would call it `full_crypto`.

And I would also like to see an comment what the feature does.

core/primitives/src/ed25519.rs
brenzi: ok

core/primitives/src/ed25519.rs
brenzi: ok

core/primitives/src/crypto.rs
brenzi: done

core/primitives/src/crypto.rs
brenzi: true, should be safe to enable

core/primitives/Cargo.toml
brenzi: done

core/primitives/src/crypto.rs
brenzi: done

bkchr:
@brenzi you need to update the `Cargo.lock` and commit that. (Just run `cargo build` for a moment)

brenzi:
It appears to me that cargo doesn't forward features correctly:
```
error[E0412]: cannot find type `Pair` in module `app`
  --> core/finality-grandpa/primitives/src/lib.rs:38:31
   |
38 | pub type AuthorityPair = app::Pair;
   |                               ^^^^ not found in `app`
help: possible candidates are found in other modules, you can import them into scope
   |
25 | use app_crypto::Pair;
   |
25 | use app_crypto::ed25519::Pair;
   |
25 | use app_crypto::sr25519::Pair;
   |
25 | use sr_primitives::testing::sr25519::Pair;
   |

error[E0277]: the trait bound `app::Signature: std::hash::Hash` is not satisfied
  --> core/finality-grandpa/primitives/src/lib.rs:33:2
   |
33 |     app_crypto!(ed25519, GRANDPA);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::hash::Hash` is not implemented for `app::Signature`
   |

```
@bkchr: do you have a clue?

brenzi:
oops, sorry, that was too quick. The following still fails:
```
cargo build -p substrate-primitives --no-default-features --features full_crypto
```
I'd suggest to include this line in CI. How should I proceed to do that?

core/primitives/src/crypto.rs
brenzi: In the end I get trouble becase I'd need to add `Mutex`:
https://github.com/paritytech/substrate/blob/48e6e349cd4d64fb94ea476089f9e4f900f31cf3/core/primitives/src/crypto.rs#L329

brenzi: I don't need this, so I can live with `#[cfg(feature = "std")]`

bkchr: Okay

bkchr:
> oops, sorry, that was too quick. The following still fails:
> 
> ```
> cargo build -p substrate-primitives --no-default-features --features full_crypto
> ```
> 
> I'd suggest to include this line in CI. How should I proceed to do that?

Are you familiar with gitlab ci? You just need to edit `.gitlabci.yml`. I would propose you copy the `srml-staking` test and modify for your needs.

.gitlab-ci.yml
```diff
@@ -156,6 +156,23 @@ test-srml-staking:                 &test-srml-staking
...
+    # Enable debug assertions since we are running optimized builds for testing
+    # but still want to have debug assertions.
+    RUSTFLAGS: -Cdebug-assertions=y
+    RUST_BACKTRACE: 1
+  except:
+    variables:
+      - $DEPLOY_TAG
+  script:
+    - cd core/primitives/
+    - time cargo build --release --verbose --no-default-features --features full_crypto
```
bkchr: Please don't build it in release, that just waste resources^^

brenzi: ok.

brenzi:
@bkchr It seems I broke the CI script:
```
Running with gitlab-runner 12.4.1 (05161b14)
  on ci2-docker 8a60f15c
Using Docker executor with image parity/rust-builder:latest ...
ERROR: Preparation failed: Error response from daemon: Conflict. The container name "/runner-8a60f15c-project-145-concurrent-0-cache-c33bcaa1fd2c77edfc3893b41966cea8" is already in use by container "81f69a0237714f11351d20325b818d3400aa19476e4057b1d43bcaefcfbe0075". You have to remove (or rename) that container to be able to reuse that name. (cache_container.go:97:0s)
Will be retried in 3s ...
```
but why and how?

core/application-crypto/src/ed25519.rs
```diff
@@ -21,6 +21,7 @@ use crate::{RuntimePublic, KeyTypeId};
 pub use primitives::ed25519::*;
 
 mod app {
+	use crate::Vec;
```
bkchr: No this is not correct, please fix the macro to include `Vec` with an absolute path.

brenzi: ok

brenzi:
@bkchr: It seems that my new CI test is not run:
https://github.com/paritytech/substrate/blob/536d6c3df57c6f4f53f7ecb0b968a541a1735eee/.gitlab-ci.yml#L274

brenzi:
I'll have to put this on wait because it seem's the latest version causes `no_std` collisions when testing against https://github.com/scs/test-no-std/tree/substrate-primitives. 
I'll revert once that's fixed

brenzi:
the last commit fixes it. My only remaining question is if my new CI test is actually running.

brenzi:
could we include building https://github.com/scs/test-no-std/tree/substrate-primitives in CI as well? It's a minimal crate that checks for `no_std` collisions. Maybe you already have means to test such things for wasm? Better ways?

.gitlab-ci.yml
```diff
@@ -275,6 +271,25 @@ build-linux-subkey:
     - cp -r scripts/docker/subkey.Dockerfile ./artifacts/subkey/
     - sccache -s
 
+build-full-crypto-feature:
```
bkchr: You should copy from the `srml-staking` test. This is not correct. You are not using the correct stage etc. Please compare with the `srml-staking` test and also move this test above line 223.

brenzi: thought it should be build stage because it's build, not test.
I had that before but then I got errors. Let's see what my last commit does

brenzi: done

core/application-crypto/src/ed25519.rs
```diff
@@ -21,6 +21,7 @@ use crate::{RuntimePublic, KeyTypeId};
 pub use primitives::ed25519::*;
 
 mod app {
+	use rstd::vec::Vec;
```
bkchr:
```suggestion
```

core/consensus/aura/Cargo.toml
```diff
@@ -37,3 +37,11 @@ test-client = { package = "substrate-test-runtime-client", path = "../../test-ru
...
 env_logger = "0.7.0"
 tempfile = "3.1.0"
+
+[features]
+default = ["full_crypto"]
+
+# This feature enables all crypto primitives for `no_std` builds like microcontrollers 
+# or Intel SGX. 
+# For the regular wasm runtime builds this should not be used.
+full_crypto = []
```
bkchr: That is actually not good.

You should change the macro to generate different output, if the `full_crypto` feature is added to the `primtives` crate.

(You probably need to create 2 different macros to make this possible)

brenzi: hmm, but wouldn't that change be more invasive and less understandable than pseudo-adding this feature to each crate using the macro?
Is your goal to leave the crates entirely untouched that have no use for the full_crypto feature?

bkchr: I don't want to add a feature, just because I want to use a macro. If the user does not provides the feature, the compiler will report weird errors. Macros should be self contained when possible and it is possible in this case.

brenzi: alright. this feels like manually doing the job cargo should do. But I agree.

bkchr: Yeah, I understand your pain, but I don't want to distribute this pain to everyone using this :D

brenzi: I think I've found a better way:
```
#[cfg(any(feature = "full_crypto", feature = "std"))]
```
because `std` always implies `full_crypto` we can safely put this logic. In the case of `babe` there is no `full_crypto`feature, but `std` will be there, given that you could accept putting this in all crates that use the app_crypto macro:

```
[features]
default = ["std"]
```

brenzi: just pushed my suggestion

core/application-crypto/src/lib.rs
```diff
@@ -53,9 +53,9 @@ pub use traits::*;
 #[macro_export]
 macro_rules! app_crypto {
 	($module:ident, $key_type:expr) => {
-		#[cfg(feature="std")]
+		#[cfg(any(feature = "full_crypto", feature = "std"))]
```
bkchr: This will not work. When you for example want to use the app crypto from grandpa in your SGX env, the code will not be accessible.

core/application-crypto/src/lib.rs
```diff
@@ -53,9 +53,9 @@ pub use traits::*;
 #[macro_export]
 macro_rules! app_crypto {
```
bkchr: What I wanted from you:
```
#[macro_export]
#[cfg(feature = "full_crypto")]
macro_rules! app_crypto {
    // Call 2 macros, one does the public and signature implementation and the other one does the Pair impl
}

#[macro_export]
#[cfg(not(feature = "full_crypto"))]
macro_rules! app_crypto {
    // Just call the macro that does the public and signature implementation
}
```

core/application-crypto/src/lib.rs
brenzi: that's true. It won't be a problem for us now, but if we one day want to run the entire client in SGX (or some other no_std env) this will be an issue

core/application-crypto/src/lib.rs
brenzi: Ok, I'll take care of this

brenzi: done, but I had to further split macros in order to get rid of all `full_crypto` compiler switches within macros

core/application-crypto/src/lib.rs
brenzi: ...discussion continued below

core/primitives/src/crypto.rs
```diff
@@ -758,7 +770,10 @@ pub trait Pair: CryptoType + Sized + Clone + Send + Sync + 'static {
 	/// be equivalent to no password at all.
 	///
 	/// `None` is returned if no matches are found.
-	fn from_string_with_seed(s: &str, password_override: Option<&str>) -> Result<(Self, Option<Self::Seed>), SecretStringError> {
+	#[cfg(feature = "std")]
+    fn from_string_with_seed(s: &str, password_override: Option<&str>)
```
bkchr: Spaces in front of the function name

brenzi: done. hopefully there's only this one occurrence.

bkchr:
Ty!

brenzi:
@bkchr thanks for your patience and support too.

Issue #3782, opened 2019-10-08
More efficient storage proofs when verifier already knows values


jimpo:
The `substrate-state-machine` crate exposes APIs for generating and verifying storage proofs using the Merkle-Patricia trie. The main function signature for verification is:

```rust
/// Check storage read proof, generated by `prove_read` call.
pub fn read_proof_check<H, I>(
	root: H::Out,
	proof: Vec<Vec<u8>>,
	keys: I,
) -> Result<HashMap<Vec<u8>, Option<Vec<u8>>>, Box<dyn Error>>
where
	H: Hasher,
	H::Out: Ord,
	I: IntoIterator,
	I::Item: AsRef<[u8]>;
```

Notably, the function does not take key-value pairs as input and verifies that they are right. Rather, the proof encodes the values themselves and "checking the proof" extracts and returns them.

Not only is this an usual definition of "proof", but it can unnecessary increase the proof size in contexts where the verifier already has the values that it wants to check (through computation or checking that some value has not changed) and the values are large.

An interface more like the following would enable proof size savings.

```rust
/// Check storage read proof, generated by `prove_read` call.
pub fn read_proof_check<H, I>(
	root: H::Out,
	proof: Vec<Vec<u8>>,
	items: I,
) -> Result<(), Box<dyn Error>>
where
	H: Hasher,
	H::Out: Ord,
	I: IntoIterator,
	I::Item: (Vec<u8>, Vec<u8>);  // Maybe this could be &[u8], but not sure about lifetimes yet
```

### Proof structure

Merkle trie proofs currently are the subset of nodes in the trie traversed when reading some value. As such it contains all of the leaf nodes of the values read. The nodes are encoded and the proof is a vector of encoded trie nodes.

### Removing values from proofs

The easiest way to reduce the proof size would be to remove the leaf nodes from the proof. Instead, the proof would consist of only the branch and extension nodes traversed when reading the values. In order to verify, the verifier attempts a lookup in the trie and if it encounters a read of a trie node by hash that it does not have in the database, then it attempts to construct a leaf node for the value matching the partial key. This can be implemented with a custom `HashDB` since the signature for getting a value on a `HashDB` is

```rust
pub trait HashDB<H: Hasher, T>: Send + Sync + AsHashDB<H, T> {
	/// Look up a given hash into the bytes that hash to it, returning None if the
	/// hash is not known.
	fn get(&self, key: &H::Out, prefix: Prefix) -> Option<T>;
}
```

[...]

PR #3783, opened 2019-10-08
Logic for checking Substrate proofs from within runtime module.

jimpo:
See PR title.

### DIFF
--- a/srml/bridge/src/error.rs
+++ b/srml/bridge/src/error.rs
@@ -1,0 +1,25 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
+// This file is part of Substrate.
+
+// Substrate is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Common error types for the srml-bridge crate.
+
+#[derive(PartialEq)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub enum Error {
+	StorageRootMismatch,
+	StorageValueUnavailable,
+}
+
--- a/srml/bridge/src/lib.rs
+++ b/srml/bridge/src/lib.rs
@@ -32,6 +32,9 @@
 // Ensure we're `no_std` when compiling for Wasm.
 #![cfg_attr(not(feature = "std"), no_std)]
 
+mod error;
+mod storage_proof;
+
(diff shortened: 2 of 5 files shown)


srml/bridge/src/error.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```
kianenigma:
```suggestion
// Copyright 2019 Parity Technologies (UK) Ltd.
```

srml/bridge/src/storage_proof.rs
```diff
@@ -0,0 +1,103 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```
kianenigma:
```suggestion
// Copyright 2019 Parity Technologies (UK) Ltd.
```

srml/bridge/src/lib.rs
```diff
@@ -242,7 +246,7 @@ mod tests {
 				MockBridge::tracked_bridges(1),
 				Some(BridgeInfo {
 					last_finalized_block_number: 42,
-					last_finalized_block_hash: test_header.hash(), // FIXME: This is broken
```
kianenigma: I don't know the details but is this now _unbroken_?

srml/bridge/src/storage_proof.rs
```diff
@@ -0,0 +1,103 @@
...
+	db: MemoryDB<H>,
+}
+
+impl<H> StorageProofChecker<H>
+	where H: Hasher
+{
+	/// Constructs a new storage proof checker.
+	///
+	/// This returns an error if the given proof is invalid with respect to the given root.
+	pub fn new(root: H::Out, proof: Vec<Vec<u8>>) -> Result<Self, Error> {
```
kianenigma: This API does look a bit strange to me that you build a new instance of `checked` and via the constructor you also give it the _task_ that it has to do (i.e. proof to check). If it woks for you then it's okay but I was expecting `new` to actually return a new instance of something that _can check_ and a `.check` to actually check.

kianenigma: Though looking at the test I see it more clear of how you intend to use it. Just a thought.

srml/bridge/src/lib.rs
jimpo: Yes, this is now fixed in this PR. Was a compilation issue.

srml/bridge/src/storage_proof.rs
jimpo: Yeah... I'm also not thrilled about this API, but it's kind of a direct consequence of the proof format. See https://github.com/paritytech/substrate/issues/3782.

`StateProofCheckerAndReader` is somewhat more accurate of a description, but in `substrate-state-machine`, it's just called `check_read_proof` usually so ¯\_(ツ)_/¯.

srml/bridge/src/lib.rs
HCastano: `testing::Header` didn't have a `hash()` method available, which is why it was broken. But yeah, it's fixed now

kianenigma: My comment was due to the fact that only looking at the _local scope of the code_, I only identified that `test_header.hash();` now has a variable and no logic has changed and I was like _hmm.. how does that solve the problem?_ 😬

PR #3791, opened 2019-10-10
Move srml RPC extensions to separate crates

tomusdrw:
The main point of the PR is to generalize the RPC extension so that they can be used in down stream projects (see https://github.com/paritytech/polkadot/pull/460).
Before they were part of `node-rpc` crate, and were too closely tied with `node-primitives`. Now they are enclosed under `srml` directory inside the module they are related to.


node/primitives/Cargo.toml
```diff
@@ -5,11 +5,7 @@ authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
 [dependencies]
-client = { package = "substrate-client", path = "../../core/client", default-features = false }
```
kianenigma: why was this here in the first place? seems _very out of place_

tomusdrw:
@kianenigma I actually have on my list to automate this process, cause there is a lot of boilerplate code needed now, and I imagine that most of the RPCs will be super simple - basically directly expose the `RuntimeApi`. We can discuss how this could be done, but then the example would be basically:
1. Specify `RuntimeApi`
2. Add some macro magic
3. ???
4. RPC exposed.

node/rpc/src/lib.rs
```diff
@@ -25,43 +25,32 @@
...
-		accounts::{Accounts, AccountsApi},
-		contracts::{Contracts, ContractsApi},
-	};
+	use srml_system_rpc::{System, SystemApi};
+	use srml_contracts_rpc::{Contracts, ContractsApi};
 
 	let mut io = jsonrpc_core::IoHandler::default();
 	io.extend_with(
-		AccountsApi::to_delegate(Accounts::new(client.clone(), pool))
+		SystemApi::to_delegate(System::new(client.clone(), pool))
```
kianenigma: Very detailed but to avoid confusion with similar names defined in `node/runtime` I would consider renaming these structs that implement rpc api to some name that suggests more clearly what they are doing, e.g.
```suggestion
		SystemApi::to_delegate(SystemRpcHandler::new(client.clone(), pool))
```

tomusdrw:
> This RPC code, despite being placed in srml is non-runtime code. Hence, almost all meaningful interactions with would have to use the runtime api (I assume for this reason we keep a ref to client). So it seems the whole concept of rpc api for srml module is quite entangled with runtime apis.

Yes, to all 3 questions. This is a conscious choice to be able to migrate the runtime in the future, but keep the RPCs backward compatible. Otherwise the RPC implementation would have to be tightly coupled with things like storage layout of srml, etc. So definitely every RPC we exposed will have some RuntimeApi behind it.

srml/contracts/rpc/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+
+//! Runtime API definition required by Contracts RPC extensions.
+//!
+//! This API should be imported and implemented by the runtime,
+//! of a node that wants to use the custom RPC extension
+//! adding Contracts access methods.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use rstd::prelude::Vec;
```
bkchr:
```suggestion
use rstd::vec::Vec;
```

PR #3792, opened 2019-10-10
Decouple randomness-collective-flip

xlc:
Decouple randomness-collective-flip and other modules

### DIFF
--- a/node-template/runtime/src/lib.rs
+++ b/node-template/runtime/src/lib.rs
@@ -33,7 +33,7 @@ pub use sr_primitives::BuildStorage;
 pub use timestamp::Call as TimestampCall;
 pub use balances::Call as BalancesCall;
 pub use sr_primitives::{Permill, Perbill};
-pub use support::{StorageValue, construct_runtime, parameter_types};
+pub use support::{StorageValue, construct_runtime, parameter_types, traits::Randomness};
 
 /// An index to a block.
 pub type BlockNumber = u32;
@@ -341,7 +341,7 @@ impl_runtime_apis! {
 		}
 
 		fn random_seed() -> <Block as BlockT>::Hash {
-			RandomnessCollectiveFlip::random_seed()
+			<RandomnessCollectiveFlip as Randomness<Hash>>::random_seed()
 		}
 	}
 
--- a/node-template/runtime/src/lib.rs
+++ b/node-template/runtime/src/lib.rs
@@ -341,7 +341,7 @@ impl_runtime_apis! {
 		}
 
 		fn random_seed() -> <Block as BlockT>::Hash {
-			<RandomnessCollectiveFlip as Randomness<Hash>>::random_seed()
+			RandomnessCollectiveFlip::random_seed()
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


srml/support/src/traits.rs
```diff
@@ -717,3 +717,23 @@ pub trait InitializeMembers<AccountId> {
 impl<T> InitializeMembers<T> for () {
 	fn initialize_members(_: &[T]) {}
 }
+
+// A trait that is able to provide randomness.
+pub trait Randomness<Output> {
+	/// Get a "random" value
+	///
+	/// Being a deterministic block chain, real randomness is difficult to come by. This gives you
```
```suggestion
	/// Being a deterministic blockchain, real randomness is difficult to come by. This gives you
```

node-template/runtime/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub use sr_primitives::BuildStorage;
 pub use timestamp::Call as TimestampCall;
 pub use balances::Call as BalancesCall;
 pub use sr_primitives::{Permill, Perbill};
-pub use support::{StorageValue, construct_runtime, parameter_types};
+pub use support::{StorageValue, construct_runtime, parameter_types, traits::Randomness};
```
bkchr: Why do we need the trait here?

node/runtime/src/lib.rs
```diff
@@ -22,7 +22,7 @@
 
 use rstd::prelude::*;
 use support::{
-	construct_runtime, parameter_types, traits::{SplitTwoWays, Currency}
+	construct_runtime, parameter_types, traits::{SplitTwoWays, Currency, Randomness}
```
bkchr: Same.

srml/contracts/src/exec.rs
```diff
@@ -753,7 +753,7 @@ where
 	}
 
 	fn random(&self, subject: &[u8]) -> SeedOf<T> {
-		randomness_collective_flip::Module::<T>::random(subject)
+		<T::Randomness as Randomness<T::Hash>>::random(subject)
```
bkchr:
```suggestion
		T::Randomness::random(subject)
```

I don't think that we need this cast?

node-template/runtime/src/lib.rs
xlc: Otherwise have this error
```
error[E0599]: no function or associated item named `random_seed` found for type `srml_randomness_collective_flip::Module<Runtime>` in the current scope
   --> node/runtime/src/lib.rs:594:30
    |
594 |             RandomnessCollectiveFlip::random_seed()
    |                                       ^^^^^^^^^^^ function or associated item not found in `srml_randomness_collective_flip::Module<Runtime>`
    |
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope, perhaps add a `use` for it:
```

xlc:
@kianenigma No there isn't previous discussion about this. Just me want to write some code that doesn't involve too much thinking.

node-template/runtime/src/lib.rs
bkchr: Ahh it is used there. Okay :)

PR #3794, opened 2019-10-10
Allow zero fee transactions without a "live" account

kianenigma:
At the moment, a free transactions will fail if origin has less than ED, or falls below it. This can easily be the case if you have just created an account and want to do something free. Quoting from @gnunicorn: you don't have to be forced to go to a bank if you want to buy something free. 

I am not sure if we want this. This allows all transactions that incur _zero fee_ to not be checked agains the internal logics of balances module (via `withdraw()`), including being checked for _minimum balance_. 

But, sudo is actually already `FreeOperational` and this is the reason it is not really _functional_ at the moment, if the sudo keys runs out of money.

cc @gnunicorn this solves your problem as well.

I will make a follow-up PR and organise some of the errors of [`Reservable`]`Currency` trait to make it look nicer.

### DIFF
--- a/srml/balances/src/lib.rs
+++ b/srml/balances/src/lib.rs
@@ -1253,25 +1253,20 @@ impl<T: Trait<I>, I: Instance + Clone + Eq> SignedExtension for TakeFees<T, I> {
 	) -> TransactionValidity {
 		// pay any fees.
 		let fee = Self::compute_fee(len, info, self.0);
-		let imbalance = match <Module<T, I>>::withdraw(
-			who,
-			fee,
-			WithdrawReason::TransactionPayment,
-			ExistenceRequirement::KeepAlive,
-		) {
-			Ok(imbalance) => imbalance,
-			Err(msg) => {
-				// TODO: relying on string error is quite bad here.
-				// if we the transaction is operational, and the error is due to existential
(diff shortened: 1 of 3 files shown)


srml/balances/src/lib.rs
```diff
@@ -1260,7 +1260,16 @@ impl<T: Trait<I>, I: Instance + Clone + Eq> SignedExtension for TakeFees<T, I> {
 			ExistenceRequirement::KeepAlive,
 		) {
 			Ok(imbalance) => imbalance,
-			Err(_) => return InvalidTransaction::Payment.into(),
+			Err(msg) => {
+				// TODO: relying on string error is quite bad here.
+				// if we the transaction is operational, and the error is due to existential
```
bkchr:
```suggestion
				// if the transaction is operational, and the error is due to existential
```

kianenigma:
> Don't we need to check that the weight is 0 and if it is also operational, we don't try to withdraw anything?

The current condition is a sueprset of this, Allowing operationals to do a bit more than just the situation fo `FreeOperational`. But yeah we can add that to the condition too.

kianenigma:
closed via https://github.com/paritytech/substrate/pull/3796

PR #3796, opened 2019-10-11
Fix semantics of ExistenceRequirement::KeepAlive.

gavofyork:


gavofyork:
Well, if you look at the documentation, this is strictly a fix: no semantics have changed. I'm ok with the idea of adding an additional `ExistenceRequirement::MustContinueExistence` which requires that the account continues to exist after the operation, though I don't know if there's any need for it.

This "fix" will now allow the possibility of storage-leaks: an account may now be created (i.e. nonce is written) without an ED balance and therefore that may never be cleaned up. I would be tempted to make the `CheckNonce` logic aware of this and avoid writing a new nonce value if the account is dead.

PR #1966, opened 2019-03-11
Fix Index reuse for zero ExistentialDeposit configured chains

mnaamani:
In chains that choose to allow zero ExistentialDeposit, the definition of a dead account is no longer the same as when existential deposit is greater than zero. In other words an account with zero balance is not considered dead. This has an impact on when an account index is (incorrectly) reused. 

In addition it really only makes sense for an account index to remain associated with the same account as long as the total balance is still positive. After an account is reaped it is reasonable to create a new account index.

This PR addresses these two points:

1. prevents incorrect account index reuse by changing implementation of the `IsDeadAccount` trait for the balances module.

2. Only calling new_account() when both the freebalance and reservedbalance components of an account don't exist

PR #3797, opened 2019-10-11
Fix treasury so pot can be spend entirely

thiolliere:
### Context
When you change existencialDeposit to 1 then the test 

https://github.com/paritytech/substrate/blob/7adfd837565f870882cb9726e546dc9df6804885/srml/treasury/src/lib.rs#L591-L604

fails at 602 because treasury hasn't been withdrawn because it wasn't allowed to.

### Done

* fix the test suite to be more accurate
* fix the failing test resulting of it by changing treasury existence requirement.

### DIFF
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -84,7 +84,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// and set impl_version to equal spec_version. If only runtime
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
-	spec_version: 174,
+	spec_version: 175,
 	impl_version: 175,
 	apis: RUNTIME_API_VERSIONS,
 };
--- a/srml/treasury/src/lib.rs
+++ b/srml/treasury/src/lib.rs
@@ -315,7 +315,7 @@ impl<T: Trait> Module<T> {
 			&Self::account_id(),
 			imbalance,
 			WithdrawReason::Transfer,
-			ExistenceRequirement::KeepAlive
+			ExistenceRequirement::AllowDeath,
 		) {
 			print("Inconsistent state - couldn't settle imbalance for funds spent by treasury");
 			// Nothing else to do here.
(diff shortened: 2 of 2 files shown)


thiolliere:
I though that `ED = 0` was invalid. https://github.com/paritytech/substrate/pull/1966#discussion_r264639764

But something I just realized is that treasury won't get updated if the dilution is less than existential deposit :-/

kianenigma:
My general sentiment is that it is okay and we should not try and make _everything be compatible with everything_. Treasury wants a `Currency`. The current balances module that srml provides is completely based on the idea of _anything less than ED is considered insubstantial and discarded_. Hence, if we are passing the current `srml-balances` with an ED > 0 to treasury as currency, everything that is going on here seems reasonable to me, including: 

> But something I just realized is that treasury won't get updated if the dilution is less than existential deposit :-/

if we could ban `ED = 0` in balances and provide `no-ed` feature that handles this + a different implementation of `is_dead_account()` (as mentioned in your linked comment) might be a better approach rather than trying to _force_ treasury (which doesn't really care about ed) to  work with balances (which does care about ed and doesn't really interpret ed = 0 nicely).

thiolliere:
OK, so this PR just make that pot can be send entirely.
If OnDilution put into pot less than ED and pot is empty then it is considered insubstantial so ok to discard. 
Same if treasury module is used for OnUnbalanced and the unbalanced stuff is less than ED and pot is empty then it is considered insubstantial so ok to discard.

Note: maybe we could extend Currency trait to be able to gather some amount that could be less than ED, but this is is not done in this PR

gavofyork:
`KeepAlive` was there because I felt it better than it not be possible to kill the treasury account and without thinking too deeply about it, that hasn't changed.

gavofyork:
I'm ok with the idea of "fixing" the ED=0 case (i.e. accounts never get deleted); it can be easily done by checking if `FreeBalance < ED && ReservedBalance < ED`, which has the nice advantage that for `ED==0` it will short-circuit  to `true` without any storage lookups and otherwise will typically get away with only a single storage lookup since it'll typically be `FreeBalance` that is non-zero rather than `ReservedBalance`.

thiolliere:
I don't have interest in fixing ED=0 but there is still an issue that if the proposal or burning make the treasury empty then (which is legit to happen) the treasury will keep its fund instead of being emptied.

I'll fix in an overseeding PR

PR #3799, opened 2019-10-11
Move sr-arithmetic to a new crate and add in a fuzzer

### LINKED ISSUE #3745: Fuzzing and Organisation for Runtime Arithmetic Types
- Relocate all the arithmetic stuff into a separate crate. Probably one file per each type defined in `sr_arithmetic.rs`.
- Create better and more fuzzing code for all tests that bombard the code with random values. A lot can be done here. Any taker's imagination is the limit. Nonetheless, the minimum is to remove all the tests that have a random number generator in them and replace them with fuzzing code. 

#### Optionally 
- Use the benchmarks to apply potentially any optimizations to the code
  - [ ] Some ideas of using LE ordering which were tried with @jimpo https://github.com/paritytech/substrate/tree/kiz-limb-arithmetic-2 (last two commits)
  - [ ] Use criterion and more comprehensive set of benchmarks.
  - [ ] Try and re-use memory: `add_assign` instead of `add`. Also read and apply the suggestions [here](https://skanthak.homepage.t-online.de/division.html).

expenses:
This closes #3745. This commit splits up `sr_arithmetic.rs` into a new crate files, translates the previous 'fuzzing' tests into [`honggfuzz`](https://github.com/rust-fuzz/honggfuzz-rs) binaries (I was going off of [`parity-scale-codec`](https://github.com/paritytech/parity-scale-codec) which uses the same, happy to change fuzzing tool though) and adds a (currently failing) `multiply_by_rational` test that was uncovered via fuzzing.

@kianenigma


expenses:
The fuzzing code found that this line: https://github.com/paritytech/substrate/blob/0cd9bbfb721b5f0f31982f681bdf1e77612ecc79/core/sr-primitives/src/sr_arithmetic.rs#L1667

sometimes adds 1 to the value of the result when it shouldn't.

expenses:
We should consider re-implementing the algorithms used by `num-bigint` for `BigUint`: https://github.com/rust-num/num-bigint/blob/master/src/algorithms.rs

expenses:
It might be better to use `num-bigint` instead after https://github.com/rust-num/num-bigint/pull/101 lands.

kianenigma:
> We should consider re-implementing the algorithms used by num-bigint for BigUint: https://github.com/rust-num/num-bigint/blob/master/src/algorithms.rs

They are taken from the same source. There might be just some optimisations to apply. 

> It might be better to use num-bigint instead after rust-num/num-bigint#101 lands.

This is great for a user to add to their runtime if they want it but to be added to the main substrate runtime, it can only happen if we are sure the library does not have a big footprint. This was the main incentive to write this mini library ourselves. 

is this PR review ready?

core/sr-arithmetic/fuzzer/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "sr-arithmetic-fuzzer"
+version = "0.1.0"
+authors = ["Ashley <ashley.ruglys@gmail.com>"]
```
kianenigma: preferably use the same format as the rest of the crates.

core/sr-arithmetic/src/biguint.rs
```diff
@@ -0,0 +1,825 @@
...
+			"BigUint {{ {:?} ({:?})}}",
+			self.digits,
+			u128::try_from(self.clone()).unwrap_or_else(|_| 0),
+		)
+	}
+}
+
+impl PartialEq for BigUint {
+	fn eq(&self, other: &Self) -> bool {
+		// sadly, we have to reallocate here as strip mutably uses self.
```
kianenigma: Hmm I seem to have forgotten to do this: 

I think this can just use `self.cmp(other) == Ordering::Equal`. Or a similar implementation, we can avoid re-allocation here.

core/sr-arithmetic/src/biguint.rs
```diff
@@ -0,0 +1,825 @@
...
+								let d: $type = d.into();
+								acc += d << (SHIFT * i);
+							}
+							Ok(acc)
+						}
+					}
+				}
+			)*
+		};
+	}
```
kianenigma: I don't see a closing bracket for `macro_rules `

kianenigma: one too much indented here.

core/sr-arithmetic/src/biguint.rs
```diff
@@ -0,0 +1,825 @@
...
+
+macro_rules! impl_from_for_smaller_than_word {
+		($($type:ty),+) => {
+			$(impl From<$type> for BigUint {
+				fn from(a: $type) -> Self {
+					Self { digits: vec! [a.into()] }
+				}
+			})*
+		}
+	}
```
kianenigma: and here

core/sr-arithmetic/src/biguint.rs
```diff
@@ -0,0 +1,825 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
```
kianenigma: Has anything significant changed here? 

- If so, would be nice to highlight them since this PR is meant to just re-locate files + fuzz them. Both of which should not need any significant change to the lib itself. 
- If not but you intend to so, I'd be happy to do that in a separate PR. I really find it bad review-wise when a PR that re-locates stuff also changes them.

core/sr-arithmetic/src/parts_per_x.rs
```diff
@@ -0,0 +1,551 @@
+
```
kianenigma: header missing + how do you like the name `per_things.rs`? we used to call them like this, even in the code.

core/sr-arithmetic/src/parts_per_x.rs
```diff
@@ -0,0 +1,551 @@
...
+/// ```
+///
+/// ```rust,should_panic
+/// # fn main() {
+/// sr_arithmetic::assert_eq_error_rate!(12, 10, 1);
+/// # }
+/// ```
+#[macro_export]
+#[cfg(feature = "std")]
+macro_rules! assert_eq_error_rate {
```
kianenigma: How did this end up here?

core/sr-arithmetic/src/traits.rs
```diff
@@ -0,0 +1,262 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```
kianenigma: This file should not exists at all. We have two options here: 

- extract sr-arithmatic in a way that it no longer depends on sr-primitives, so that SR-primitives can depend on it. This is afaik for sure doable and needs more effort and scaffolding I suppose 
- just put the arithmetic stuff in the sr-primitves but in their own sub-module for better organisation. 

I was actually glad that you did the former approach, and it is much better if it works out, but for sure not like this.

core/sr-arithmetic/src/traits.rs
```diff
@@ -0,0 +1,262 @@
...
+	Zero, One, Bounded, CheckedAdd, CheckedSub, CheckedMul, CheckedDiv,
+	CheckedShl, CheckedShr
+};
+use rstd::ops::{
+	Add, Sub, Mul, Div, Rem, AddAssign, SubAssign, MulAssign, DivAssign,
+	RemAssign, Shl, Shr
+};
+
+/// A lazy value.
+pub trait Lazy<T: ?Sized> {
```
kianenigma: Not sure why these are here.

core/sr-arithmetic/src/traits.rs
```diff
@@ -0,0 +1,262 @@
...
+	/// This will compute the value if the function is invoked for the first time.
+	fn get(&mut self) -> &T;
+}
+
+impl<'a> Lazy<[u8]> for &'a [u8] {
+	fn get(&mut self) -> &[u8] { &**self }
+}
+
+/// Extensible conversion trait. Generic over both source and destination types.
+pub trait Convert<A, B> {
```
kianenigma: likewise.

core/sr-arithmetic/src/helpers_128bit.rs
```diff
@@ -72,6 +72,15 @@ pub fn multiply_by_rational(mut a: u128, mut b: u128, c: u128) -> Result<u128, &
 		mem::swap(&mut a, &mut b);
 	}
 
+	// Attempt to perform the division first
```
kianenigma: nice 👍

core/sr-arithmetic/fuzzer/src/biguint.rs
```diff
@@ -0,0 +1,175 @@
+use honggfuzz::fuzz;
```
kianenigma: license header + short instruction on how to use on the module doc please.

core/sr-arithmetic/src/parts_per_x.rs
kianenigma: yeah this is used in substrate as well, if you need it here and you want to stay independent, sadly you have to either duplicate it or come up with another way

core/sr-arithmetic/src/biguint.rs
thiolliere: but we can do in following PR

core/sr-arithmetic/src/traits.rs
thiolliere: I think that Lazy and Convert should stay in sr-primitive as they are not arithmetic related

thiolliere:
> the trait file seems to be mistakingly moved. I would recommend you to:
> 
>     * Either make the sr-arithmeric a new module of the primitives for simplicity and focus on the fuzzing.

Maybe we should keep SimpleArithmetic trait and Saturating conversion here in the new crate it seems fine, only Lazy and Convert doesn't make sense here and should be moved back to sr-primitives.

core/sr-arithmetic/src/biguint.rs
expenses: No, I haven't significantly changed any internal logic

core/sr-arithmetic/src/parts_per_x.rs
expenses: Would it make sense to move `assert_eq_error_rate` into `rstd`?

expenses:
One problem with having all the arithmetic stuff in a `sr-primitives` is that it makes the fuzzer binaries larger and we end up hitting against this issue: https://github.com/rust-fuzz/honggfuzz-rs/issues/19.

expenses:
After removing `#[macro_export]`, I'm getting the following warning:
```
warning: unused macro definition
  --> core/sr-arithmetic/src/per_things.rs:11:1
   |
11 | / macro_rules! assert_eq_error_rate {
12 | |     ($x:expr, $y:expr, $error:expr $(,)?) => {
13 | |         assert!(
14 | |             ($x) >= (($y) - ($error)) && ($x) <= (($y) + ($error)),
...  |
20 | |     };
21 | | }
   | |_^
   |
   = note: `#[warn(unused_macros)]` on by default
```

Not sure how to fix this, as the macro is definitely used.

xlc:
> After removing `#[macro_export]`, I'm getting the following warning:
> 
> ```
> warning: unused macro definition
>   --> core/sr-arithmetic/src/per_things.rs:11:1
>    |
> 11 | / macro_rules! assert_eq_error_rate {
> 12 | |     ($x:expr, $y:expr, $error:expr $(,)?) => {
> 13 | |         assert!(
> 14 | |             ($x) >= (($y) - ($error)) && ($x) <= (($y) + ($error)),
> ...  |
> 20 | |     };
> 21 | | }
>    | |_^
>    |
>    = note: `#[warn(unused_macros)]` on by default
> ```
> 
> Not sure how to fix this, as the macro is definitely used.

Try `#[cfg(test)]` as it is only used by test code

expenses:
Thanks!

core/sr-arithmetic/src/biguint.rs
kianenigma: Great, let's keep it that way for the reason mentioned and I'll be happy to work on any sort of optimisation of you find suitable, in a separate PR.

core/sr-arithmetic/src/parts_per_x.rs
kianenigma: I don't think so, rstd should be the representation of rust's `std`. 

It is used only in testing, Maybe you can put it in some testing util folder. 

But as I said, not a big concern as far as this PR is concerned.

Cargo.toml
```diff
@@ -53,6 +53,8 @@ members = [
 	"core/service/test",
 	"core/session",
 	"core/sr-api-macros",
+	"core/sr-arithmetic",
+	"core/sr-arithmetic/fuzzer",
```
kianenigma: I am not sure with how this fuzzing stuff works, but can you double check if this is needed?

core/sr-arithmetic/fuzzer/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "sr-arithmetic-fuzzer"
+version = "2.0.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+
+[dependencies]
+honggfuzz = "0.5"
+sr-arithmetic = { path = ".." }
+primitive-types = "0.6"
+num-bigint = { git = "https://github.com/expenses/num-bigint", branch = "as-slice" }
```
kianenigma: Since this is fuzz code, it is okay to depend on git dependencies, but is there any other way to not depend on this branch?

core/sr-arithmetic/src/traits.rs
```diff
@@ -0,0 +1,240 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Primitives for the runtime modules.
+
+use rstd::prelude::*;
+use rstd::{self, convert::{TryFrom, TryInto}};
+//use crate::codec::{Codec, Encode, Decode, HasCompact};
```
kianenigma:
```suggestion

```

core/sr-arithmetic/src/traits.rs
```diff
@@ -0,0 +1,240 @@
...
+//use crate::generic::{Digest, DigestItem};
```
kianenigma:
```suggestion

```

core/sr-primitives/src/lib.rs
```diff
@@ -64,14 +63,14 @@ pub use primitives::{TypeId, crypto::{key_types, KeyTypeId, CryptoType}};
 pub use app_crypto::RuntimeAppPublic;
 
 /// Re-export top-level arithmetic stuff.
-pub use sr_arithmetic::{
+pub use arithmetic::{
```
kianenigma: So here we have the dilemma: 

always import arithmetic stuff into sr-primitives and re-export 
or refactor once and for all and make all other crates depend on sr-arithmetic. 

It should be more optimised to do the latter, right? you might need some primitive traits but not the arithmetic. There is no reason for your runtime to pull in arithmetic in this case. 
baseically we are making `sr-primitives` ,maybe too fat here. 

What do you think @thiolliere ?

core/sr-primitives/src/traits.rs
```diff
@@ -17,28 +17,20 @@
...
 use std::fmt::{Debug, Display};
 #[cfg(feature = "std")]
 use serde::{Serialize, Deserialize, de::DeserializeOwned};
 use primitives::{self, Hasher, Blake2Hasher, TypeId};
-use crate::codec::{Codec, Encode, Decode, HasCompact};
+use crate::codec::{Codec, Encode, Decode};
 use crate::transaction_validity::{
 	ValidTransaction, TransactionValidity, TransactionValidityError, UnknownTransaction,
 };
+pub use arithmetic::traits::*;
```
kianenigma: Same dilemma as above

Cargo.toml
expenses: Having this here just means that the fuzzer doesn't have it's own `Cargo.lock`. I'm happy to remove this.

core/sr-arithmetic/fuzzer/Cargo.toml
expenses: Yeah, that's not quite ideal, but to compare the `BigUint` implementations I need access to the underlying slice, so I had to use my own branch. I'll try and make a PR for that.

core/sr-primitives/src/lib.rs
expenses: I'm happy either way.

core/sr-arithmetic/fuzzer/Cargo.toml
expenses: https://github.com/rust-num/num-bigint/pull/12 does what I want, but the repo seems to be moving pretty slowly.

expenses:
> The fuzzing code found that this line:
> 
> https://github.com/paritytech/substrate/blob/0cd9bbfb721b5f0f31982f681bdf1e77612ecc79/core/sr-primitives/src/sr_arithmetic.rs#L1667
> 
> sometimes adds 1 to the value of the result when it shouldn't.

Actually, if you remove this line the fuzzer doesn't crash at all, so the issue here is that we want `multiply_by_rational` to round up sometimes, but (probably?) not all the time.

expenses:
> This is great for a user to add to their runtime if they want it but to be added to the main substrate runtime, it can only happen if we are sure the library does not have a big footprint. This was the main incentive to write this mini library ourselves.

What's the best way to check the size of the runtime footprint? I think it would be worthwhile to verify this.

kianenigma:
> What's the best way to check the size of the runtime footprint? I think it would be worthwhile to verify this.

Looking into the `.wasm` files created in `node/runtime` should give you some insight. But we have discussed this internally, and the outcome was always to not depend (for now at least) on external libs.

Cargo.toml
thiolliere: in parity-scale-codec the fuzzer is part of the workspace as well, I think this is fine

core/sr-arithmetic/fuzzer/Cargo.toml
thiolliere: as fuzzer is part of workspace then the test-suite will fail if this branch is changed, I think this is not fine, not sure what the best way to solve this.

core/sr-primitives/src/lib.rs
thiolliere: I'm ok either way but I agree with kian.

core/sr-arithmetic/fuzzer/src/biguint.rs
```diff
@@ -0,0 +1,208 @@
...
+use sr_arithmetic::biguint::{BigUint, Single};
+use std::convert::TryFrom;
+use num_traits::ops::checked::CheckedDiv;
+
+type S = u128;
+
+fn main() {
+	loop {
+		fuzz!(|data: (Vec<Single>, Vec<Single>, bool)| {
+			let (mut digits_u, mut digits_v, rem) = data;
```
kianenigma: I would immediately convert these two to 

```rust 
let u = BigUint::from_limbs(&digits_u);
let v = BigUint::from_limbs(&digits_v);
let ue = S::try_from(u.clone()).unwrap();
let ve = S::try_from(v.clone()).unwrap();
``` 

and clone for the rest of the function. I had to create them separately in `run_with_data_set` because they were separate functions at the time. You can for sure reduce some redundant code here

core/sr-arithmetic/fuzzer/src/biguint.rs
```diff
@@ -0,0 +1,208 @@
...
+
+			digits_u.reverse();
+			digits_v.reverse();
+
+			let num_u = num_bigint::BigUint::new(digits_u.clone());
+			let num_v = num_bigint::BigUint::new(digits_v.clone());
+
+			// Equality
+
+			assert_biguints_eq(&u, &num_u);
```
kianenigma: This function is useful later on when you apply arithmetic, but here it is quite trivial. Both `num_bigint` and `BigUint` are built directly from the same slice. 

What you could do, is to compare the `u.cmp(v)` and `num_u.cmp(num_v)`. Now you are just checking `if digits_u == digits_v`

core/sr-arithmetic/fuzzer/Cargo.toml
thiolliere: This is used for the function `fn assert_biguints_eq(a: &BigUint, b: &num_bigint::BigUint)`
I think you can come up with a solution by converting `a` into num_bigint and use the partialeq implementation of num_bigint.

Or you can use the function https://docs.rs/num-bigint/0.2.3/num_bigint/struct.BigUint.html#method.to_radix_be or https://docs.rs/num-bigint/0.2.3/num_bigint/struct.BigUint.html#method.to_radix_le

core/sr-arithmetic/fuzzer/src/biguint.rs
```diff
@@ -0,0 +1,208 @@
...
+			assert_biguints_eq(&w, &num_w);
+
+			// Division
+
+			if v.len() == 1 && v.get(0) != 0 {
+				let w = u.clone().div_unit(v.get(0));
+				let num_w = num_u.clone() / &num_v;
+				assert_biguints_eq(&w, &num_w);
+			} else {
+				let w = u.clone().div(&v, rem).map(|(w, _)| w);
```
kianenigma: Can be a bit better: 
Check remainder as well (unless if `v.len() == 1`)

core/sr-primitives/src/lib.rs
kianenigma: Let's leave it for another PR if needed then.

core/sr-arithmetic/fuzzer/Cargo.toml
kianenigma: > > The fuzzing code found that this line:
> > https://github.com/paritytech/substrate/blob/0cd9bbfb721b5f0f31982f681bdf1e77612ecc79/core/sr-primitives/src/sr_arithmetic.rs#L1667
> > 
> > sometimes adds 1 to the value of the result when it shouldn't.
> 
> Actually, if you remove this line the fuzzer doesn't crash at all, so the issue here is that we want `multiply_by_rational` to round the value, but (probably?) not all the time.

Can you paste an example crash?

core/sr-arithmetic/fuzzer/src/biguint.rs
```diff
@@ -0,0 +1,208 @@
...
+				let num_w = num_u.clone().checked_div(&num_v);
+
+				// assert that the division was checked equally:
+				// assert_eq!(w.is_some(), num_w.is_some());
```
kianenigma: and for this: strip both u and v at the very beginning. aJust better to do so. 
Then run and compare division only when u.len() > v.len()

In this case you can remote all the `if let` and unwrap both of them immediately, as they both _must_ return. then you can also either have `// assert_eq!(w.is_some(), num_w.is_some());` uncommented (though' I think unwrapping implies this)

core/sr-arithmetic/fuzzer/src/biguint.rs
```diff
@@ -0,0 +1,208 @@
...
+
+			assert_biguints_eq(&w, &num_w);
+
+			// Division
+
+			if v.len() == 1 && v.get(0) != 0 {
+				let w = u.clone().div_unit(v.get(0));
+				let num_w = num_u.clone() / &num_v;
+				assert_biguints_eq(&w, &num_w);
+			} else {
```
kianenigma:
```suggestion
			} else if u.len() > v.len() {
```

core/sr-arithmetic/fuzzer/src/rational128.rs
```diff
@@ -0,0 +1,77 @@
...
+				u128::from_be_bytes(c_bytes),
+			);
+
+			println!("++ Equation: {} * {} / {}", a, b, c);
+
+			let truth = mul_div(a, b, c);
+			// if Err just use the truth value. We don't care about such cases. The point of this
+			// fuzzing is to make sure: if `multiply_by_rational` returns `Ok`, it must be 100% accurate
+			// returning `Err` is fine.
+			let result = multiply_by_rational(a, b, c).unwrap_or(truth);
```
kianenigma: Oh, my bad. I think I should have made the doc more accurate. The new implementation of `multiply_by_rational ` should _always return a 100& accurate value as long as it fits in u128_. So this must change a bit and unwrap_or() is not correct

kianenigma: And I suppose about the +- 1problem you can ignore it 🤔

core/sr-arithmetic/fuzzer/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+num-traits = "*"
```
kianenigma: better to specify a format here?

core/sr-arithmetic/fuzzer/Cargo.toml
kianenigma: Did you consider this? I suppose you can go to  `to_str_radix` and then easily convert every 4 chunks into a `Single`.

core/sr-arithmetic/src/per_things.rs
```diff
@@ -0,0 +1,535 @@
+
```
kianenigma: license header.

core/sr-arithmetic/src/traits.rs
```diff
@@ -0,0 +1,238 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Primitives for the runtime modules.
+
+use rstd::prelude::*;
+use rstd::{self, convert::{TryFrom, TryInto}};
+//use crate::weights::DispatchInfo;
```
kianenigma:
```suggestion

```

node/runtime/src/lib.rs
```diff
@@ -84,8 +84,8 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// and set impl_version to equal spec_version. If only runtime
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
-	spec_version: 175,
-	impl_version: 175,
+	spec_version: 176,
```
kianenigma: Don't think we have changed any behaviour here?

core/sr-arithmetic/fuzzer/src/biguint.rs
kianenigma: will this be solved or do you disagree?

core/sr-arithmetic/fuzzer/Cargo.toml
expenses: Converting to `num_bigint::BigUint` seems like the easiest approach for now.

core/sr-arithmetic/fuzzer/src/biguint.rs
kianenigma: 🏴

core/sr-arithmetic/fuzzer/src/biguint.rs
kianenigma: 🏴

core/sr-arithmetic/fuzzer/src/biguint.rs
kianenigma: 🏴

core/sr-arithmetic/fuzzer/src/biguint.rs
kianenigma: 🏴

core/sr-arithmetic/fuzzer/src/biguint.rs
kianenigma: 🏴

core/sr-arithmetic/src/parts_per_x.rs
kianenigma: Please add the licence header. 🏴

core/sr-arithmetic/src/traits.rs
kianenigma: 🏴

core/sr-arithmetic/src/traits.rs
kianenigma: 🏴

node/runtime/src/lib.rs
kianenigma: 🏴

kianenigma:
@expenses a noticeable number number of my reviews have been marked as _resolved_ while I see no code change, nor an explanation. It is okay if you disagree and don't want to apply something, but please don't just resolve a review without any further information. 

I have went through all of the notes again and marked (hopefully all of the) ones that were _mistakingly_ marked as resolved with  a 🏴. Please give them a look and see if you will fix them or argue why not. 

Github hides some review comments automatically, keep an eye on that as well as it can be a holy pain :P

core/sr-arithmetic/fuzzer/Cargo.toml
kianenigma: 🏴 ????

core/sr-arithmetic/src/per_things.rs
```diff
@@ -0,0 +1,535 @@
...
+use serde::{Serialize, Deserialize};
+
+use rstd::{ops, prelude::*, convert::TryInto};
+use codec::{Encode, Decode};
+use crate::traits::{SaturatedConversion, UniqueSaturatedInto, Saturating};
+
+/// Copied from `sr-primitives` and documented there.
+#[cfg(feature = "std")]
+#[cfg(test)]
+macro_rules! assert_eq_error_rate {
```
kianenigma: this can probably be place somewhere nicer, no? 

We generally might need this for testing in this crate when there is a rounding error. So I would put it in `lib.rs`. 

And I don't know why we just thought of this 😁: why not import this into sr-primitives from here and just not define it there?

kianenigma: also, I think cfg(test) is enough. `std` is out feature that explicitly says not wasm, i.e. native code. while `test` is also not-wasm, it is more expressive to say just `test`. This macro is and will never be used in any real code, just for test, regardless of wasm/std.

expenses: > And I don't know why we just thought of this 😁: why not import this into sr-primitives from here and just not define it there?

Sorry, I'm not entirely sure what you're suggesting. Previously, I was defining `assert_eq_error_rate` in `sr-arithmetic` then re-exporting it in `sr-primitives`. Should we go back to doing that?

> also, I think cfg(test) is enough. std is out feature that explicitly says not wasm, i.e. native code. while test is also not-wasm, it is more expressive to say just test. This macro is and will never be used in any real code, just for test, regardless of wasm/std.
 
Done :^)

kianenigma: > Sorry, I'm not entirely sure what you're suggesting. Previously, I was defining assert_eq_error_rate in sr-arithmetic then re-exporting it in sr-primitives. Should we go back to doing that?

Correct, I thought it is defined twice now. It is okay.

core/sr-arithmetic/fuzzer/src/rational128.rs
kianenigma: ?

core/sr-arithmetic/src/biguint.rs
```diff
@@ -0,0 +1,816 @@
...
+}
+
+#[cfg(feature = "std")]
+impl rstd::fmt::Debug for BigUint {
+	fn fmt(&self, f: &mut rstd::fmt::Formatter<'_>) -> rstd::fmt::Result {
+		write!(
+			f,
+			"BigUint {{ {:?} ({:?})}}",
+			self.digits,
+			u128::try_from(self.clone()).unwrap_or_else(|_| 0),
```
bkchr:
```suggestion
			u128::try_from(self.clone()).unwrap_or(0),
```

core/sr-arithmetic/src/fixed64.rs
```diff
@@ -0,0 +1,206 @@
...
+	type Output = Self;
+
+	fn sub(self, rhs: Self) -> Self::Output {
+		Self(self.0 - rhs.0)
+	}
+}
+
+impl CheckedSub for Fixed64 {
+	fn checked_sub(&self, rhs: &Self) -> Option<Self> {
+		if let Some(v) = self.0.checked_sub(rhs.0) {
```
bkchr: Shorter:
```
self.0.checked_sub(rhs.0).map(Self)
```

core/sr-arithmetic/src/fixed64.rs
```diff
@@ -0,0 +1,206 @@
...
+			Some(Self(v))
+		} else {
+			None
+		}
+	}
+}
+
+impl CheckedAdd for Fixed64 {
+	fn checked_add(&self, rhs: &Self) -> Option<Self> {
+		if let Some(v) = self.0.checked_add(rhs.0) {
```
bkchr:
```
self.0.checked_add(rhs.0).map(Self)
```

core/sr-arithmetic/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Minimal fixed point arithmetic primitives and types for runtime.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+// to allow benchmarking
+#![cfg_attr(feature = "bench", feature(test))]
+#[cfg(feature = "bench")] extern crate test;
```
bkchr: We use https://crates.io/crates/criterion for benchmarking.

core/sr-arithmetic/src/per_things.rs
```diff
@@ -0,0 +1,535 @@
...
+				Self(x)
+			}
+		}
+
+		impl From<codec::Compact<$name>> for $name {
+			fn from(x: codec::Compact<$name>) -> Self {
+				x.0
+			}
+		}
+
```
bkchr: Can be replaced by `derive(CompactAs)`.

core/sr-arithmetic/fuzzer/src/biguint.rs
```diff
@@ -0,0 +1,183 @@
...
+//!
+//! # More infomation
+//! More information about `honggfuzz` can be found
+//! [here](https://docs.rs/honggfuzz/0.5.45/honggfuzz/).
+
+use honggfuzz::fuzz;
+use sr_arithmetic::biguint::{BigUint, Single};
+use std::convert::TryFrom;
+
+type S = u128;
```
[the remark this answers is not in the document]
expenses: That's the type that we're testing conversion from. It might be better to just remove it actually.

expenses:
@bkchr I'm a little hesitant to make these changes, because it might be better to save them for another PR. @kianenigma what do you think?

core/sr-arithmetic/fuzzer/src/biguint.rs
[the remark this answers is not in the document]
kianenigma: That's my historical short hand for `type **S**aturated = u128`. Should stay as a type in case we chnage it but can be named the full version.

kianenigma:
@expenses The recommendations don't seem to be any logic change but rather syntactic improvements. Those can be applied safely.

The migration to criterion can be saved for another PR as it is a bit bulky and I don't want to stall this PR for it too much. @bkchr I have added benchmarks in the past both with and without criterion. If we really want to have as a _"we benchmark only with criterion"_, we should at some point a) make it a strict rule b) migrate the already existing ones. 

Personally, I don't think it is worth migrating them and making it strict rule. Given that the required benchmark is fairly simple (like the ones in this crate) the test crate gets the job done okay.

PR #3800, opened 2019-10-11
core/finality-grandpa: Request block sync from network after import timeout

mxinden:
## Goal

In order to participate in the *Grandpa* gossip process a *Grandpa* voter needs
to have access to the blocks referenced in the *Grandpa* messages. Blocks are
being synchronized atonomously via the network *sync* service. In case this does
not work as expected e.g. due to delayed messages where the corresponding blocks
are already purged, we can use the new `set_sync_fork_request` API added in
#3633 to tell the network:

1. Which blocks are relevant for the current round.

2. Where to retrieve these relevant blocks from.

This is further described in #3629.

An initial solution piggy-backing on the *Grandpa neighbor messages* was
suggested in #3755. After further discussions an easier (intermediate) solution
via *Grandpa vote messages* and the *UntilImported* component is suggested with
this pull request. In addition the *UntilImported* component already has the
notion of withholding messages and the notion of a import timeout.

## 2nd solution (*Grandpa vote messages* and *UntilImported*)

The *UntilImported* component within `core/finality-grandpa` withholds messages
from the upper *Grandpa* layers until all blocks referenced in these messages
are imported on the local node.

With the goal above in mind we can notify the network *sync* service after a
certain timeout of waiting that we need a given block and which peers probably
have it.

## Pull request status

I am opening this up for architectual feedbacks. Changes are grouped by their
functionality. This pull request replaces #3755. I will follow up on comments in
#3755 in later pull requests.


core/finality-grandpa/src/communication/mod.rs
```diff
@@ -436,7 +443,9 @@ impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
 			format!("Failed to receive on unbounded receiver for round {}", round.0)
 		));
 
-		let incoming = incoming.select(out_rx);
+		// TODO: Do I understand correctly that we combine the incoming Grandpa messages with the messages we are
+		// sending out to other nodes and in addition thereby also send to ourself?
```
rphmeier: yes! this means that the logic for importing our own votes is unified with the logic for importing others' votes in the GRANDPA crate - reducing code paths in the vote processing pipeline and making bugs less likely.

core/finality-grandpa/src/until_imported.rs
```diff
@@ -165,28 +190,35 @@ impl<Block: BlockT, Status, I, M> Stream for UntilImported<Block, Status, I, M>
...
+			for (&block_hash, &mut (block_number, ref mut last_log, ref senders, ref v)) in &mut self.pending {
 				if let Some(number) = self.status_check.block_number(block_hash)? {
+					// TODO: Now that we pass the number all the way down here, should we check if it is the same as we
+					// are expecting it to be?
 					known_keys.push((block_hash, number));
 				} else {
 					let next_log = *last_log + LOG_PENDING_INTERVAL;
+					// TODO: Should this not be >=? We want to log whenever we want to log every LOG_PENDING_INTERVAL,
+					// right?
 					if Instant::now() <= next_log {
```
mxinden: @rphmeier this should be `>=`, right?

core/finality-grandpa/src/until_imported.rs
```diff
@@ -64,23 +70,32 @@ pub(crate) trait BlockUntilImported<Block: BlockT>: Sized {
...
+	status_check: BlockStatus,
+	// TODO: Why is this called inner? Why not being more descriptive and say finality_msg_stream?
 	inner: Fuse<I>,
 	ready: VecDeque<M::Blocked>,
 	check_pending: Interval,
-	pending: HashMap<Block::Hash, (Instant, Vec<M>)>,
+	/// Mapping block hashes to their block number, the point in time it was first encountered (Instant), nodes that
+	/// have the corresponding block (inferred by the fact that they send a message referencing it) and a list of
+	/// Grandpa messages referencing the block hash.
+	pending: HashMap<Block::Hash, (NumberFor<Block>, Instant, Vec<network::PeerId>, Vec<(M)>)>,
```
andresilva: This should be a `HashSet<PeerId>`, we might see multiple messages from the same peer referencing the same hash.

andresilva:
Regarding the issue I mentioned above in my review. We could start by asking the peer that sent us the vote, and then fallback to asking any peer we're connected to that is on the same round or further from the round whose block data we're looking for.

Another issue with the current implementation is that only validators will be actively trying to fetch block data. Full nodes just running the GRANDPA observer don't create any round streams and therefore don't run the code in `UntilImported`. An easy fix is to have the observer create the round streams and just consume and drop all messages, this will make sure that the logic in `UntilImported` is run.

core/network/src/protocol/sync.rs
```diff
@@ -456,14 +456,20 @@ impl<B: BlockT> ChainSync<B> {
...
-			}
-			return;
+			debug!(
+				target: "sync",
+				"Explicit sync request for block {:?} with no peers specified. \
+				 Syncing from all connected peers {:?} instead.",
+				hash, peers,
+			);
+
+			peers = self.peers.iter().map(|(id, _)| id.clone()).collect();
```
arkpar: Should filter out peers that have `best_block < number`. We don't want to request anything from peers that are definitely not yet synced.

core/finality-grandpa/src/environment.rs
```diff
@@ -370,6 +370,8 @@ impl<Block: BlockT> SharedVoterSetState<Block> {
 
 /// The environment we run GRANDPA in.
 pub(crate) struct Environment<B, E, Block: BlockT, N: Network<Block>, RA, SC, VR> {
+	// TODO: As far as I can tell this does not follow the pattern of having an outer struct protecting an inner struct
+	// via a lock in order to be Sync. How about renaming this to `client` like we do in many other places?
```
andresilva: Indeed, I did the rename.

core/finality-grandpa/src/until_imported.rs
```diff
@@ -64,23 +70,31 @@ pub(crate) trait BlockUntilImported<Block: BlockT>: Sized {
 }
 
 /// Buffering imported messages until blocks with given hashes are imported.
-pub(crate) struct UntilImported<Block: BlockT, Status, I, M: BlockUntilImported<Block>> {
+pub(crate) struct UntilImported<Block: BlockT, BlockStatus, BlockSyncRequester, I, M: BlockUntilImported<Block>> {
 	import_notifications: Fuse<Box<dyn Stream<Item = BlockImportNotification<Block>, Error = ()> + Send>>,
-	status_check: Status,
+	block_sync_requester: BlockSyncRequester,
+	status_check: BlockStatus,
+	// TODO: Why is this called inner? Why not being more descriptive and say finality_msg_stream?
```
andresilva: This is used with both the global per-set streams (commits, catch-ups) and the round streams (votes). Maybe rename it to `msg_stream`?

core/finality-grandpa/src/until_imported.rs
```diff
@@ -165,28 +182,35 @@ impl<Block: BlockT, Status, I, M> Stream for UntilImported<Block, Status, I, M>
 
 		if update_interval {
 			let mut known_keys = Vec::new();
-			for (&block_hash, &mut (ref mut last_log, ref v)) in &mut self.pending {
+			for (&block_hash, &mut (block_number, ref mut last_log, ref v)) in &mut self.pending {
 				if let Some(number) = self.status_check.block_number(block_hash)? {
+					// TODO: Now that we pass the number all the way down here, should we check if it is the same as we
+					// are expecting it to be?
```
andresilva: This will be checked later on in `wait_completed`, we pass the canon number to it (the one we fetch here and store in `known_keys`. The `wait_completed` can look at the vote itself and the target number to figure out what the original number was.

core/finality-grandpa/src/until_imported.rs
andresilva: I removed the TODO but we can address this in a follow up PR if this one gets merged soon.

PR #3801, opened 2019-10-11
Provide macro for exporting functions from wasm

bkchr:
The macro generates the functions with the signature we expect for wasm
functions. This macro is useful for tests where we need to call into
wasm. Parameter passing is done by SCALE encoding the input and output
parameters.

CC @lamafab

### DIFF
--- a/core/executor/runtime-test/src/lib.rs
+++ b/core/executor/runtime-test/src/lib.rs
@@ -5,48 +5,23 @@
 #[cfg(feature = "std")]
 include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
-use rstd::{vec::Vec, slice, vec};
+#[cfg(not(feature = "std"))]
+use rstd::{vec::Vec, vec};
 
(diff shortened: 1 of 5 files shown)


core/executor/runtime-test/src/lib.rs
```diff
@@ -5,48 +5,23 @@
 #[cfg(feature = "std")]
 include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
-use rstd::{vec::Vec, slice, vec};
+#[cfg(not(feature = "std"))]
+use rstd::{vec::Vec, vec};
```
andresilva: I guess this is only needed under `no-std` because the macro below only expands under `no-std`. Would it make more sense to have the macro just use the full path to `rstd::vec::Vec` instead of relying on this import?

bkchr: The macro does not rely on this, the functions defined in this file depend on this. So, this is fine to include here.

PR #3802, opened 2019-10-11
node: re-use testnet genesis spec for staging testnet

andresilva:
We have two different chain specs staging and testnet which are rather similar. I've made slight changes to the testnet genesis so that we can reuse that code when defining staging chain spec (used in flaming fir and under `--chain staging`).

### DIFF
--- a/node/cli/src/chain_spec.rs
+++ b/node/cli/src/chain_spec.rs
@@ -117,83 +117,19 @@ fn staging_testnet_config_genesis() -> GenesisConfig {
 	)];
 
 	// generated with secret: subkey inspect "$secret"/fir
-	let endowed_accounts: Vec<AccountId> = vec![
+	let root_key: AccountId = hex![
 		// 5Ff3iXP75ruzroPWRP2FYBHWnmGGBSb63857BgnzCoXNxfPo
-		hex!["9ee5e5bdc0ec239eb164f865ecc345ce4c88e76ee002e0f7e318097347471809"].unchecked_into(),
-	];
+		"9ee5e5bdc0ec239eb164f865ecc345ce4c88e76ee002e0f7e318097347471809"
+	].unchecked_into();
 
-	const ENDOWMENT: Balance = 10_000_000 * DOLLARS;
-	const STASH: Balance = 100 * DOLLARS;
+	let endowed_accounts: Vec<AccountId> = vec![root_key.clone()];
 
-	GenesisConfig {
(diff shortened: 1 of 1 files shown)


test-utils/chain-spec-builder/Cargo.toml
```diff
@@ -5,7 +5,10 @@ authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
 [dependencies]
-clap = { version = "~2.32.0", features = ["yaml"] }
+ansi_term = "0.12.1"
+clap = { version = "~2.32.0" }
```
bkchr:
```suggestion
```

It is not used anymore? Or does structopt requires it?

test-utils/chain-spec-builder/Cargo.toml
```diff
@@ -5,7 +5,10 @@ authors = ["Parity Technologies <admin@parity.io>"]
...
+keystore = { package = "substrate-keystore", path = "../../core/keystore" }
 node-cli = { path = "../../node/cli" }
 primitives = { package = "substrate-primitives",  path = "../../core/primitives" }
-substrate-service = { path = "../../core/service" }
+rand = "0.7.2"
+structopt = "0.2.0"
```
bkchr:
```suggestion
structopt = "0.3.3"
```

Use latest.

test-utils/chain-spec-builder/src/main.rs
```diff
@@ -1,49 +1,275 @@
...
+	file.write_all(json.as_bytes())
+		.map_err(|err| err.to_string())?;
+
+	Ok(())
 }
 
 fn main() {
-	let json = generate_chain_spec();
-	println!("{}", json);
+	let code = match run() {
```
bkchr: Please use this https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/question-mark-in-main-and-tests.html

test-utils/chain-spec-builder/Cargo.toml
andresilva: The reason I did this was because we're using `structopt = "0.2.0"` elsewhere, and I didn't want to add a new version. Might make sense to update structopt everywhere if it's easy.

test-utils/chain-spec-builder/Cargo.toml
andresilva: Not sure, need to check but probably not needed.

PR #3804, opened 2019-10-11
Avoid key collision for child trie

cheme:
Update 11/11/19, this PR will revert to initial approach where unique id is provided by the srml storage.


cheme:
Closing in favor of #4162

PR #3805, opened 2019-10-11
Fix Typo

c410-f3r:
The `chain::error::FutureResult` doc is currently referring to the wrong structure.


andresilva:
Thanks for the PR!

PR #3806, opened 2019-10-11
It's Clippy time!

c410-f3r:
Compliance with all default hard errors and crates that are explicitly
using `#![deny(warnings)]`.
There are some legitimate warnings like `println!("")` -> `println!()` but
they aren't considered in this PR.

### DIFF
--- a/core/primitives/src/ed25519.rs
+++ b/core/primitives/src/ed25519.rs
@@ -41,6 +41,7 @@ use crate::{crypto::{Public as TraitPublic, UncheckedFrom, CryptoType, Derive}};
 type Seed = [u8; 32];
 
 /// A public key.
+#[cfg_attr(feature = "std", derive(Hash))]
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Encode, Decode, Default)]
 pub struct Public(pub [u8; 32]);
 
@@ -149,13 +150,6 @@ impl<'de> Deserialize<'de> for Public {
 	fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> where D: Deserializer<'de> {
 		Public::from_ss58check(&String::deserialize(deserializer)?)
 			.map_err(|e| de::Error::custom(format!("{:?}", e)))
-	}
-}
-
-#[cfg(feature = "std")]
-impl std::hash::Hash for Public {
-	fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-		self.0.hash(state);
 	}
 }
 
--- a/core/primitives/src/sr25519.rs
+++ b/core/primitives/src/sr25519.rs
@@ -47,6 +47,7 @@ use schnorrkel::keys::{MINI_SECRET_KEY_LENGTH, SECRET_KEY_LENGTH};
 const SIGNING_CTX: &[u8] = b"substrate";
 
 /// An Schnorrkel/Ristretto x25519 ("sr25519") public key.
+#[cfg_attr(feature = "std", derive(Hash))]
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Encode, Decode, Default)]
 pub struct Public(pub [u8; 32]);
 
@@ -148,13 +149,6 @@ impl<'de> Deserialize<'de> for Public {
 	fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> where D: Deserializer<'de> {
 		Public::from_ss58check(&String::deserialize(deserializer)?)
 			.map_err(|e| de::Error::custom(format!("{:?}", e)))
-	}
-}
-
-#[cfg(feature = "std")]
-impl std::hash::Hash for Public {
-	fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-		self.0.hash(state);
 	}
 }
 
--- a/srml/support/src/hash.rs
+++ b/srml/support/src/hash.rs
@@ -59,7 +59,7 @@ impl StorageHasher for Twox64Concat {
 	type Output = Vec<u8>;
 	fn hash(x: &[u8]) -> Vec<u8> {
 		twox_64(x)
-			.into_iter()
+			.iter()
 			.chain(x.into_iter())
 			.cloned()
 			.collect::<Vec<_>>()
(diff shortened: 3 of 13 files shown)


core/primitives/src/ed25519.rs
```diff
@@ -41,7 +41,7 @@ use crate::{crypto::{Public as TraitPublic, UncheckedFrom, CryptoType, Derive}};
 type Seed = [u8; 32];
 
 /// A public key.
-#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Encode, Decode, Default)]
+#[derive(Eq, PartialOrd, Ord, Clone, Encode, Decode, Default)]
```
DemiMarie-parity: A better fix is to use `cfg_attr` to guard `derive(Hash)` here, instead of manually implementing `Hash` or `PartialEq`.

core/primitives/src/sr25519.rs
```diff
@@ -47,7 +47,7 @@ use schnorrkel::keys::{MINI_SECRET_KEY_LENGTH, SECRET_KEY_LENGTH};
 const SIGNING_CTX: &[u8] = b"substrate";
 
 /// An Schnorrkel/Ristretto x25519 ("sr25519") public key.
-#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Encode, Decode, Default)]
+#[derive(Eq, PartialOrd, Ord, Clone, Encode, Decode, Default)]
```
DemiMarie-parity: Ditto

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -160,6 +160,8 @@ macro_rules! implement_per_thing {
 		{
 			type Output = N;
 			fn mul(self, b: N) -> Self::Output {
+				use core::ops::Div;
```
```suggestion
				use rstd::ops::Div;
```

preferably all at the top.

srml/support/src/hash.rs
```diff
@@ -59,7 +59,7 @@ impl StorageHasher for Twox64Concat {
 	type Output = Vec<u8>;
 	fn hash(x: &[u8]) -> Vec<u8> {
 		twox_64(x)
-			.into_iter()
```
kianenigma: why? the value (return value of `twox_64`) is being consumed and I find it nice that the code is implying ti by `into_`. Is there a performance implication here?

srml/system/src/lib.rs
```diff
@@ -578,7 +578,7 @@ impl<T: Trait> Module<T> {
 		// We perform early return if we've reached the maximum capacity of the event list,
 		// so `Events<T>` seems to be corrupted. Also, this has happened after the start of execution
 		// (since the event list is cleared at the block initialization).
-		if <Events<T>>::append([event].into_iter()).is_err() {
+		if <Events<T>>::append([event].iter()).is_err() {
```
kianenigma: same as the other comment about iter

srml/support/src/hash.rs
niklasad1: It is equivalent to `iter()` because of arrays don't implement `IntoIterator [T; X]`, see https://github.com/rust-lang/rust/issues/25725. Thus, a readability thingy

core/consensus/babe/primitives/src/lib.rs
```diff
@@ -18,6 +18,7 @@
 #![deny(warnings)]
 #![forbid(unsafe_code, missing_docs, unused_variables, unused_imports)]
 #![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::too_many_arguments)]
```
niklasad1: can you to scope it down to individual methods/functions instead?

c410-f3r: This lint is triggered by some function declared in `#![deny(warnings)]` and `decl_runtime_apis!` but it is now being removed to avoid `clippy` attributes

c410-f3r:
To remove all hard errors, more research is needed, specially inside the declarative `decl_runtime_apis!` macro. Nevertheless, the ones covered in this PR are a great start IMO

core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -178,7 +178,7 @@ macro_rules! implement_per_thing {
 					// `rem_multiplied_upper` is less than $max^2 therefore divided by $max it fits
 					// in $type. remember that $type always fits $max.
 					let mut rem_multiplied_divided_sized =
-						(rem_multiplied_upper / upper_max) as $type;
+						rem_multiplied_upper.div(upper_max) as $type;
```
tomusdrw: Hmm, how is that better?

c410-f3r: It is just to avoid `https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_arithmetic_impl`. Probably a false-positive

core/sr-primitives/src/traits.rs
```diff
@@ -475,9 +475,10 @@ pub trait Hash: 'static + MaybeSerializeDebug + Clone + Eq + PartialEq {
 #[cfg_attr(feature = "std", derive(Debug, Serialize, Deserialize))]
 pub struct BlakeTwo256;
 
-impl Hash for BlakeTwo256 {
+impl self::Hash for BlakeTwo256 {
```
bkchr: Why we need this self here?

c410-f3r: Clippy thinks that `Hash` alone is the std `Hash`, which triggers https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq

core/sr-primitives/src/sr_arithmetic.rs
bkchr: Could you please revert this? We should not include fixes for false positives in clippy.

core/sr-primitives/src/traits.rs
bkchr: Same.

andresilva: Please revert this one.

bkchr:
@c410-f3r you need to merge master.

c410-f3r:
@bkchr Merged

bkchr:
Ty

c410-f3r:
Thanks

PR #3809, opened 2019-10-12
deps: update clap and structopt

andresilva:
Update clap and structopt to the latest version.


node/cli/Cargo.toml
```diff
@@ -63,4 +63,4 @@ tempfile = "3.1.0"
 
 [build-dependencies]
 cli = { package = "substrate-cli", path = "../../core/cli" }
-structopt = "0.2.0"
+structopt = "0.3.3"
```
bkchr: I'm not sure, but shouldn't it be enough to have it in the normal dependencies?

andresilva: No, `build.rs` imports stuff from `structopt`. Just tried and it doesn't build.

bkchr: Yeah I know that, I just thought that build deps are merged with normal deps as well.

Issue #1897, opened 2019-02-28
Rewrite `construct_runtime!` as procedural macro


bkchr:
The declarative macro `construct_runtime!` should be rewritten as procedural macro. The current version of the macro can be found here: https://github.com/paritytech/substrate/blob/master/srml/support/src/runtime.rs#L78-L267
All the `__decl_*` helper macros in the same file should be converted as well.

Expected outcome:
- Much better error reporting
- Better readable code
- More tests 

Expected time frame, around week, depending on the rust/procedural macro knowledge.

alleycat-at-git:
@bkchr I'm currently implementing this refactoring and one thing I struggle with is how to call `impl_outer_event` that is located in `srml-support` crate from the `srml-support-procedural` crate. Basically I'm rewriting the `__decl_outer_event` macro and it uses `crate::impl_outer_event` at the expansion. Obviously, when I do the same in `srml-support-procedural` crate I get an error. Is there a workaround of that except an obvious rewriting of `impl_outer_event`?

bkchr:
Hey, first ty that you are doing this :)

Could you please take a look here:
https://github.com/paritytech/substrate/blob/66042f1cc737f4eb5ca8d05db9d981c3e4bd3693/srml/support/procedural/src/storage/transformation.rs#L85-L91

Basically, we will re-export the macro from the `srml-support` module. So, you will need to include this crate in a hidden module and then you can access everything `srml-support` exports.
I hope that is understandable, if not, just ask again :)

alleycat-at-git:
@bkchr #3810 Ready for review

Issue #3788, opened 2019-10-09
Improve decl_storage parsing error message


thiolliere:
Currently the storage line attributes are parsed with:

https://github.com/paritytech/substrate/blob/bebe88a64c7199fcbcfdfae0ba6eb72e9e797b8d/srml/support/procedural/src/storage/mod.rs#L106-L120

this options are not good for error messages: like if the expression in `build(expression)` is wrong the error message is
```
error: expected `:`
   --> srml/example/src/lib.rs:307:26
    |
307 |         Accounts get(accounts) build(|config| {
    |                                ^^^^^

```

This must be changed.

Additionally the derivation of parse for enum is wrong as well as it doesn't show the possible variants.

thiolliere:
I think Parse trait should be implementde by hand.

thiolliere:
Fixed by https://github.com/paritytech/substrate/pull/4731

PR #3810, opened 2019-10-12
Refactor `construct_runtime` to procedural

alleycat-at-git:
**Description**
Refactor `construct_runtime` macro from declarative to procedural. Reference issue #1897.


srml/support/procedural/src/construct_runtime/transformations.rs
```diff
@@ -0,0 +1,377 @@
...
+	let outer_event = try_tok!(decl_outer_event_or_origin(
+		&name,
+		modules.iter().filter(|module| module.name.to_string() != "System"),
+		&system_module,
+		&scrate,
+		DeclOuterKind::Event,
+	));
+	let outer_origin = try_tok!(decl_outer_event_or_origin(
+		&name,
+		modules.iter().filter(|module| module.name.to_string() != "System"),
```
bkchr: You don't need to call `to_string` always. Ident implements the correct `PartialEq` for that.

srml/support/procedural/src/lib.rs
```diff
@@ -215,3 +216,63 @@ use proc_macro::TokenStream;
...
+/// Construct a runtime, with the given name and the given modules.
+///
+/// The parameters here are specific types for `Block`, `NodeBlock`, and `InherentData`
+/// and the modules that are used by the runtime.
+/// `Block` is the block type that is used in the runtime and `NodeBlock` is the block type
+/// that is used in the node. For instance they can differ in the extrinsics type.
+///
+/// # Example:
+///
+/// ```nocompile
```
bkchr: Please make this compile.

srml/support/procedural/src/construct_runtime/transformations.rs
```diff
@@ -0,0 +1,377 @@
+use super::parse::{ModuleDeclaration, ModulePart, RuntimeDefinition, WhereSection};
```
bkchr: Please move all of this into the `mod.rs` file.
Having these `transformation` file is somewhat useless.

alleycat-at-git:
@bkchr Thanks for your feedback! 

> We decided to move away from the automatic Parse implementation, as it generates very bad error messages. Could you please do the same and implement on your own?

Can you give a couple of examples where this can go bad? As I moved through the implementation I checked various errors that could arise and they seemed reasonable to me. Like in the most general case this approach leads to error "Token X is expected" or "Identifier is expected" which is pretty helpful IMO. It's true that automatic derive works badly for enums thought, but I made a wrapper around the only enum I had so that it produces a meaningful error.

bkchr:
Could you provide the said `trybuild` tests, so that I can see the errors?

thiolliere:
> Can you give a couple of examples where this can go bad?

not only the enums but the option as well and maybe other types as well, see https://github.com/paritytech/substrate/issues/3788

srml/support/procedural/src/lib.rs
alleycat-at-git: @bkchr I tried to make it compile, but it has some obstacles:
Since the macro expansion relies on having `srml-support` crate in dependencies, it looks like it's impossible to have this doc test in `srml-support-procedural` crate (adding `srml-support` to `srml-support-procedural` will have a circular dependency). We can move the docs to `srml-support` crate but then `srml-support-procedural::construct_runtime` will be undocumented or having a duplicate docs which is not very nice also. To me it looks like it's not worth of having the doc test, but rather leave the nocompile option as it is.

alleycat-at-git:
@bkchr @thiolliere Made necessary updates

srml/support/procedural/src/construct_runtime/parse.rs
```diff
@@ -0,0 +1,211 @@
...
+#[derive(ToTokens, Debug)]
+pub struct ModuleEntryWrapper {
+	pub inner: ModuleEntry,
+}
+
+impl Parse for ModuleEntryWrapper {
+	fn parse(input: ParseStream) -> Result<Self> {
+		ModuleEntry::parse(input)
+			.map(|inner| ModuleEntryWrapper { inner })
+			.map_err(|_| input.error("Expected `default` or module export name (e.g. Call, Event, etc.)"))
```
thiolliere: is there span in the output here ?

srml/support/procedural/src/construct_runtime/parse.rs
```diff
@@ -0,0 +1,211 @@
...
+pub enum ModuleEntry {
+	Default(Token![default]),
+	Part(ModulePart),
+}
+
+#[derive(Parse, ToTokens, Debug, Clone)]
+pub struct ModulePart {
+	pub name: Ident,
+	pub generics: syn::Generics,
+	pub args: ext::Opt<ext::Parens<ext::Punctuated<Ident, Token![,]>>>,
```
thiolliere: I'm not sure this option is legit as well, if the inner thing is not a valid ident for example then the whole option will be None and the error will be on the invalidity of ModuleEntryWrapper I think

thiolliere:
I think the parsing still needs some improvment ideally everything is implemented by hand, like you can take example of srml/staking/reward-curve.

for example this input:
```rust
construct_runtime! {
	pub enum Runtime where
		Block = Block,
		NodeBlock = Block,
	{
		System: system,
		Balance: balances::<Instance1>::{Call(toto), Origin<I>},
	}
}
```

result in:

```
error: no rules expected the token `;`
  --> $DIR/missing_origin_generic_on_module_with_instance.rs:3:1
   |
3  | / construct_runtime! {
4  | |     pub enum Runtime where
5  | |         Block = Block,
6  | |         NodeBlock = Block,
...  |
11 | |     }
12 | | }
   | |_^ no rules expected this token in macro call
   |

error[E0433]: failed to resolve: use of undeclared type or module `system`
 --> $DIR/missing_origin_generic_on_module_with_instance.rs:9:11
  |
9 |         System: system,
  |                 ^^^^^^ use of undeclared type or module `system`

```

srml/support/procedural/src/construct_runtime/parse.rs
thiolliere: also only inherent have this args, all other must be checked to have none

srml/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -0,0 +1,395 @@
...
+			token: modules_token,
+		},
+		..
+	} = definition;
+
+	// Assert we have system module declared
+	let system_module = match find_system_module(modules.iter()) {
+		Some(sm) => sm,
+		None => {
+			return syn::Error::new(
```
bkchr: I think this is really bad style. This `construct_runtime` should return a `Result` and this `Result` should be handled in the calling method.

srml/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -0,0 +1,395 @@
...
+use parse::{ModuleDeclaration, ModulePart, RuntimeDefinition, WhereSection};
+use proc_macro::TokenStream;
+use proc_macro2::{Span, TokenStream as TokenStream2};
+use quote::quote;
+use srml_support_procedural_tools::syn_ext as ext;
+use srml_support_procedural_tools::{generate_crate_access, generate_hidden_includes};
+use syn::Ident;
+
+// try macro but returning tokenized error
+macro_rules! try_tok(( $expre : expr ) => {
```
bkchr: Please remove this.

srml/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -0,0 +1,395 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+mod parse;
+
+use parse::{ModuleDeclaration, ModulePart, RuntimeDefinition, WhereSection};
+use proc_macro::TokenStream;
```
bkchr: Please use only `proc_macro2` down here.

srml/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -0,0 +1,395 @@
...
+	block: &'a syn::TypePath,
+	unchecked_extrinsic: &'a syn::TypePath,
+	module_declarations: impl Iterator<Item = &'a ModuleDeclaration>,
+	scrate: &'a TokenStream2,
+) -> TokenStream2 {
+	let modules_tokens = module_declarations.filter_map(|module_declaration| {
+		let maybe_config_part = module_declaration
+			.module_parts()
+			.into_iter()
+			.filter(|part| part.name == "Inherent")
```
bkchr: This filtering is done almost in every method and leads to a lot of code duplication, please do this in a function.

bkchr:
@alleycat-at-git hey, any update? :)

expenses:
@alleycat-at-git is it ok if I try and finish this one off?

alleycat-at-git:
> @alleycat-at-git is it ok if I try and finish this one off?

@bkchr @expenses I'll try to finish it myself this weekend

srml/support/procedural/src/construct_runtime/mod.rs
alleycat-at-git: It will be cumbersome since `syn::parse_macro_input` works only with `proc_macro`. But I reduced the usage to the minimum

alleycat-at-git:
@bkchr @thiolliere Ready for review. `continuous-integration/gitlab-check_polkadot` but it seems like the error is not related to the code.

gavofyork:
Some conflicts...

alleycat-at-git:
> Some conflicts...

@gavofyork Fixed

paint/support/procedural/src/construct_runtime/parse.rs
```diff
@@ -0,0 +1,354 @@
...
+    pub node_block: syn::TypePath,
+    pub unchecked_extrinsic: syn::TypePath,
+}
+
+impl Parse for WhereSection {
+    fn parse(input: ParseStream) -> Result<Self> {
+        let _: Token![where] = input.parse()?;
+        let mut seen_keys = HashSet::new();
+        let mut definitions = HashMap::new();
+        while !input.peek(token::Brace) {
```
bkchr: This `while` can be replaced with a:
```let definitions = Punctuated::parse_terminated(input)?;```

File: paint/support/procedural/src/construct_runtime/parse.rs [hunk 11]
```diff
@@ -0,0 +1,354 @@
...
+            definitions.insert(kind, value);
+            if !input.peek(Token![,]) {
+                if !input.peek(token::Brace) {
+                    return Err(input.error("Expected `,` or `{`"));
+                }
+                break;
+            }
+            let _: Token![,] = input.parse()?;
+        }
+        let expected_seen_keys: HashSet<_> = vec![
```
bkchr: Here, we should collected the `definitions` into a `Vec<WhereDefinition>`.

Then we need a function:
```
fn remove_kind(kind: WhereKind, definitions: &mut Vec<_>) -> Result<WhereDefinition> {
    if let Some(pos) = definitions.iter().position(|d| d.kind == kind) {
        Ok(definitions.remove(pos))
    else {
         Error::("Could not find bla")
    }
}
```

After calling this function, we should return an error for every value that is still in the vector (as they are duplicates)

paint/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+    block: &'a syn::TypePath,
+    unchecked_extrinsic: &'a syn::TypePath,
+    module_declarations: impl Iterator<Item = &'a ModuleDeclaration>,
+    scrate: &'a TokenStream2,
+) -> TokenStream2 {
+    let modules_tokens = module_declarations.filter_map(|module_declaration| {
+        let maybe_config_part = module_declaration
+            .module_parts()
+            .into_iter()
+            .filter(|part| part.name == "Inherent")
```
bkchr: Replace `filter()` and `next()` with `find()`.

paint/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+fn decl_validate_unsigned<'a>(
+    runtime: &'a Ident,
+    module_declarations: impl Iterator<Item = &'a ModuleDeclaration>,
+    scrate: &'a TokenStream2,
+) -> TokenStream2 {
+    let modules_tokens = module_declarations
+        .filter(|module_declaration| {
+            module_declaration
+                .module_parts()
+                .into_iter()
```
bkchr: `ModuleDeclaration` should have a function `exists`.

paint/support/procedural/src/construct_runtime/mod.rs
bkchr: `ModuleDeclaration` should have a function `find`.

paint/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+    runtime: &'a Ident,
+    module_declarations: impl Iterator<Item = &'a ModuleDeclaration>,
+    scrate: &'a TokenStream2,
+) -> TokenStream2 {
+    let modules_tokens = module_declarations
+        .filter_map(|module_declaration| {
+            let generics: Vec<_> = module_declaration
+                .module_parts()
+                .into_iter()
+                .filter(|part| part.name == "Config")
```
bkchr: There only can exists one `Config` per line.

Use the proposed: `ModuleDeclaration::find().map()`

paint/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+                .map(|part| part.generics)
+                .collect();
+            if generics.len() == 0 {
+                None
+            } else {
+                let transformed_generics: Vec<_> = generics[0]
+                    .params
+                    .iter()
+                    .map(|param| quote!(<#param>))
```
bkchr: This can be just a `map()` on the value from above.

paint/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+        .filter_map(|module_declaration| {
+            let parts = module_declaration.module_parts();
+            let parts_len = parts.len();
+            let filtered_names: Vec<_> = parts
+                .into_iter()
+                .filter(|part| part.name != "Module")
+                .map(|part| part.name.clone())
+                .collect();
+            // Theres no `Module` entry
+            if filtered_names.len() == parts_len {
```
bkchr: Use
`ModuleDeclaration::find().map()`

paint/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+                .map(|name| quote!(#module::#name)),
+        );
+        let type_decl = quote!(
+            pub type #type_name = #module::Module <#(#generics),*>;
+        );
+        types.extend(type_decl);
+        names.push(&module_declaration.name);
+    }
+    // Make nested tuple structure like (((Babe, Consensus), Grandpa), ...)
+    let all_modules = if names.len() < 2 {
```
bkchr: This can be converted to:
```
names.iter().fold(Tokenstream::default(), |combined, name| {
    quote!((#name, #combined))
})
```

paint/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+use paint_support_procedural_tools::{generate_crate_access, generate_hidden_includes};
+use parse::{ModuleDeclaration, ModulePart, RuntimeDefinition, WhereSection};
+use proc_macro::TokenStream;
+use proc_macro2::{Span, TokenStream as TokenStream2};
+use quote::quote;
+use syn::{Ident, Result};
+
+pub fn construct_runtime(input: TokenStream) -> TokenStream {
+    let definition = syn::parse_macro_input!(input as RuntimeDefinition);
+    match construct_runtime_parsed(definition) {
```
bkchr:
```suggestion
    construct_runtime_parsed(definition).unwrap_or_else(|e| e.to_compile_error()).into()
```

paint/support/procedural/src/construct_runtime/parse.rs
```diff
@@ -0,0 +1,354 @@
...
+pub enum WhereKind {
+    Block,
+    NodeBlock,
+    UncheckedExtrinsic,
+}
+
+impl WhereKind {
+    pub fn parse(input: ParseStream, seen_keys: &mut HashSet<WhereKind>) -> Result<Self> {
+        let lookahead = input.lookahead1();
+        if lookahead.peek(keyword::Block) {
```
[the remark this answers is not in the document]
alleycat-at-git: I made it to avoid storing keyword tokens inside the `WhereKind` enum (storing keyword tokens is required for meaningful error messages for duplicates). This would introduce too much pain with manually implementing Hash, PartialEq, etc. for `WhereKind`.

However, I found a better approach that looks good from every perspective.

paint/support/procedural/src/construct_runtime/parse.rs
alleycat-at-git: From the `parse_terminated`: `Parsing continues until the end of this parse stream`. So this method is usually used with parsing content inside braces like `braced!` macro that effectively cuts out part of the token stream. Since we have no clear delimiters here like braces, currently I see no better way of doing it.

paint/support/procedural/src/construct_runtime/parse.rs
alleycat-at-git: Nice suggestion - this makes the code much simpler :)

paint/support/procedural/src/construct_runtime/mod.rs
alleycat-at-git: I'm not sure what you mean here.  The thing is that in case of 2 params decl macros accept something like `<T> <I>` instead of `<T, I>` - that's what this code does.

paint/support/procedural/src/construct_runtime/mod.rs
alleycat-at-git: To me, it looked like the old decl macro used `first one wins` rule for multiple declarations. So if we're gonna disallow the double declarations I'm gonna need to add it to code. In this case, I also need to understand what to do with `Balances::{default, Config}` - is that an error?

paint/support/procedural/src/construct_runtime/mod.rs
bkchr: Where do we accept `<T><I>`?

And what I wanted is, you should not call `collect` in line 188. Use `find` and then you just get an `Option` that you can map directly.

paint/support/procedural/src/construct_runtime/mod.rs
bkchr: Yeah, we should disallow multiple. Adding this to the declarative macro would be a huge pain.

`default, Config` should also be an error.

paint/support/procedural/src/construct_runtime/parse.rs
bkchr: Good point.

paint/support/procedural/src/construct_runtime/mod.rs
alleycat-at-git: https://github.com/paritytech/substrate
> Where do we accept <T><I>?

/blob/ba72d757254efcda1375bd447212e5149cb432a8/primitives/sr-primitives/src/lib.rs#L523-L529

As far as I can see it accepts `<T>, <I>` vs `<T, I>`

alleycat-at-git: > And what I wanted is, you should not call collect in line 188
I see, I made the fix to make the code shorter

bin/node-template/runtime/src/lib.rs
```diff
@@ -243,7 +243,7 @@ construct_runtime!(
 		Timestamp: timestamp::{Module, Call, Storage, Inherent},
 		Aura: aura::{Module, Config<T>, Inherent(Timestamp)},
 		Grandpa: grandpa::{Module, Call, Storage, Config, Event},
-		Indices: indices::{default, Config<T>},
```
alleycat-at-git: @bkchr Note that I had to fix it, since now `construct_runtime` errors on this

bkchr:
@alleycat-at-git ready for review again?

alleycat-at-git:
> @alleycat-at-git ready for review again?

@bkchr Yes

gavofyork:
conflicts need resolving @alleycat-at-git  and need to do a final review @bkchr @thiolliere .

alleycat-at-git:
> conflicts need resolving @alleycat-at-git and need to do a final review @bkchr @thiolliere .

@gavofyork @bkchr @thiolliere  - fixed conflicts

alleycat-at-git:
> One last thing that needs to be done before merging, please convert your files to use tabs and not spaces.

@bkchr fixed

PR #3811, opened 2019-10-12
chain spec builder: add generate mode for authority keys

andresilva:
This PR overhauls the existing chain-spec-builder utility. It adds a new subcommand `generate` which takes the number of authorities and endowed accounts, creates a chainspec, and generates keystores for each authority preloaded with their respective keys. I changed the app to use `structopt` instead of `clap` and the existing functionality (taking authority seeds, etc.) is available under the subcommand `new`.
This is useful for quickly setting up test networks with arbitrary number of nodes.


test-utils/chain-spec-builder/src/main.rs
```diff
@@ -1,49 +1,263 @@
...
+		/// The number of authorities.
+		#[structopt(long, short)]
+		authorities: usize,
+		/// The number of endowed accounts.
+		#[structopt(long, short, default_value = "0")]
+		endowed: usize,
+		/// The path where the chain spec should be saved.
+		#[structopt(long, short, default_value = "./chain_spec.json")]
+		chain_spec_path: PathBuf,
+		/// Path to use when saving generated keystores for each authority.
```
bkchr: Please also add that each authority creates its own folder `auth-0` etc.

test-utils/chain-spec-builder/src/main.rs
```diff
@@ -1,49 +1,263 @@
...
+		},
+	};
+
+	let json = generate_chain_spec(
+		authority_seeds,
+		endowed_accounts,
+		sudo_account,
+	)?;
+
+	let mut file = File::create(chain_spec_path)
```
bkchr: These 6 lines can be abbreviated with https://doc.rust-lang.org/std/fs/fn.write.html

So
```
fs::write(chain_spec_path, json).map_err(ToString::to_string)
```

PR #3812, opened 2019-10-12
Bump dependencies, respecting semver

DemiMarie-parity:


bkchr:
You have done this just last week? Do we really need this every week?

DemiMarie-parity:
@bkchr Ideally, this would be automated.

kianenigma:
Such PRs can potentially break polkadot as well. So they should be checked pre-merge. I just some dependency divergence between some cli dep that breaks polkadot build

kianenigma:
namely updating `structopt` which I assume happened recently.

bkchr:
structopt was not updated in this pr.

kianenigma:
Yes I had a general comment about "Such PRs", not this one in particular. I didn't really look into where when it was updated and I suppose it doesn't matter now. Next polkadot merge can fix it.

DemiMarie-parity:
@kianenigma is there a reason this check is not part of CI?

kianenigma:
Not sure. I practice, we can just add a check to every PR that tries to build polkadot with this branch. cc @gabreal

bkchr:
This check would be useless. People are already confused about the check runtime test. When polkadot could not be built anymore, how should this be solved by an external contributor? Yes, that does not work.

kianenigma:
I never said it would be a holy grail but it does not hurt anyone if there is a check that tells you "Hey polkadot will not build with this". If it has a sane reason, then it just helps you be more mindful and aware of it, like a warning. Similar to check runtime it can be optional. I really think it is a step forward at the least :D 🤷🏻‍♂️

bkchr:
People already don't understand check runtime and think they have done something wrong.
I also think that we require much more important CI stuff (like a bot).

PR #3813, opened 2019-10-13
Fix #1536: do not require to construct a block for encoding it

### LINKED ISSUE #1536: Do not require to construct a block for encoding it
In situations this [this](https://github.com/paritytech/substrate/blob/9650ffb710bb3b6742cb5b18851b489eca25d4bc/core/client/src/client.rs#L725) we need to create a block instance to encode it. It would be nice to make a block encodable without constructing one to prevent copying the extrinsics/header.

circlespainter:
Fixes #1536.

_WARNING_: day 3 Rust developer and total Substrate newbie here, plus the proposed solution is surprisingly simple so it is entirely possible that, at this stage, this is just mindless babbling; I apologize in advance if that turns out to be the case.

This proposal follows from the fact that SCALE [encodes tuples and structs in the same way](https://substrate.dev/docs/en/overview/low-level-data-format#tuples-and-structures) and [`parity-scale-codec`](https://github.com/paritytech/parity-scale-codec) seems to [support tuple encoding](https://github.com/paritytech/parity-scale-codec/blob/4a559de5f6fece90205a159e008b86304d4c8750/src/codec.rs#L812).

Concretely I added an `encode_from(&header, &extrinsics) -> Vec<u8>` function to the `Block` trait and straightforwardly implemented it by just encoding a pair; finally I replaced `Block::new` with that function in `client.rs` (that seems so far the only code that copies the header to encode the block).

Other approaches I considered were:

- Just building a tuple in-place in `client.rs`: duplicates logic that should really be kept and maintained with `Block`.
- Exporting a `block_tuple` macro together with `Block` implementations: slightly more efficient but less clear and less structured, e.g. doesn't require impls. to provide one; less clear that it is available, as it requires a separate import and is not bundled with the trait, and also when it should be used.
- Defining a new `struct` (e.g. `BlockEnvelope`) that borrows the headers and adding codec support for it: encoding refs by expanding them into the actual content might be confusing and possibly incoherent with other cases (and so it might increase technical debt).
- Custom encoding function: less maintainable.

### DIFF
--- a/core/sr-primitives/src/traits.rs
+++ b/core/sr-primitives/src/traits.rs
@@ -687,8 +687,8 @@ pub trait Block: Clone + Send + Sync + Codec + Eq + MaybeSerializeDebugButNotDes
 	fn hash(&self) -> Self::Hash {
 		<<Self::Header as Header>::Hashing as Hash>::hash_of(self.header())
 	}
-	/// Encodes the block from header and extrinsics, that is without creating an instance of it.
-	fn encode_from(header: &Self::Header, extrinsics: &Vec<Self::Extrinsic>) -> Vec<u8>;
+	/// Create an encoded block from the given `header` and `extrinsics` without requiring to create an instance.
+	fn encode_from(header: &Self::Header, extrinsics: &[Self::Extrinsic]) -> Vec<u8>;
 }
 
 /// Something that acts like an `Extrinsic`.
--- a/core/sr-primitives/src/traits.rs
+++ b/core/sr-primitives/src/traits.rs
@@ -687,6 +687,8 @@ pub trait Block: Clone + Send + Sync + Codec + Eq + MaybeSerializeDebugButNotDes
 	fn hash(&self) -> Self::Hash {
 		<<Self::Header as Header>::Hashing as Hash>::hash_of(self.header())
 	}
+	/// Encodes the block from header and extrinsics, that is without creating an instance of it.
+	fn encode_from(header: &Self::Header, extrinsics: &Vec<Self::Extrinsic>) -> Vec<u8>;
 }
 
 /// Something that acts like an `Extrinsic`.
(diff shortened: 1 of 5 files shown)


circlespainter:
[clabot:check]

core/sr-primitives/src/traits.rs
```diff
@@ -687,6 +687,8 @@ pub trait Block: Clone + Send + Sync + Codec + Eq + MaybeSerializeDebugButNotDes
 	fn hash(&self) -> Self::Hash {
 		<<Self::Header as Header>::Hashing as Hash>::hash_of(self.header())
 	}
+	/// Encodes the block from header and extrinsics, that is without creating an instance of it.
+	fn encode_from(header: &Self::Header, extrinsics: &Vec<Self::Extrinsic>) -> Vec<u8>;
```
```suggestion
	fn encode_from(header: &Self::Header, extrinsics: &[Self::Extrinsic]) -> Vec<u8>;
```

core/sr-primitives/src/traits.rs
```diff
@@ -687,6 +687,8 @@ pub trait Block: Clone + Send + Sync + Codec + Eq + MaybeSerializeDebugButNotDes
 	fn hash(&self) -> Self::Hash {
 		<<Self::Header as Header>::Hashing as Hash>::hash_of(self.header())
 	}
+	/// Encodes the block from header and extrinsics, that is without creating an instance of it.
```
```suggestion
	/// Create an encoded block from the given `header` and `extrinsics` without requiring to create an instance.
```

circlespainter:
@bkchr Thanks for the review! I'll apply the changes today or tomorrow. I have a question about `lib.rs` and the version bump: is it correct to have it only for `impl` and not `spec`? I think so because neither the behavior nor the format have changed but it looks like this is the first time that those numbers are misaligned.

bkchr:
> @bkchr Thanks for the review! I'll apply the changes today or tomorrow. I have a question about `lib.rs` and the version bump: is it correct to have it only for `impl` and not `spec`? I think so because neither the behavior nor the format have changed but it looks like this is the first time that those numbers are misaligned.

That is correct, but I would even accept this pr without changes to these values. Nevertheless, just bumping `impl` is correct. We always reset `impl` to `spec`, when we bump `spec`.

PR #3816, opened 2019-10-14
refactor: Transaction-Payment module

### LINKED ISSUE #3291: Adjust byte fee as well with the block fullness targeting multiplier
> Length fee is on the other hand static. In principle, length fee should adjust the same way as the weight fee does.

cc @keorn

kianenigma:
This PR refactors some leftover from the weight / fee system. The incentive is that I intend to write a document soon about this concept. Going through the details, I found it to be pointless to do so, given that I also intend to refactor some of it. Once this is merged, I will make a conceptual doc about weight. 

### Refactors 

- A new minimal module (with no dispatchables) that maintains the `SignedExtension` for that takes fees. Based on historical dilemmas regarding transaction fees (see here), I have named it _transaction-payment_, referring to a minimum inclusion payment that each transaction pays, nothing more. 
- System module no linger maintains a multiplier for fees. `transaction-payment` does.  System module only takes care of maintaining a per-block weight and length accumulator. `transaction-payment` interprets this as desired to deduct fees. 
- Balances module no longer takes care of any payment related concept. `transaction-payment` does. 

Fairly speaking, while the diff is big, the PR is almost insubstantial and does not alter any significant logic. The only logic change is the fix of #3291 

### Still to do 

- [x] closes #3291 
- [x] polkadot PR https://github.com/paritytech/polkadot/pull/482

### Maybe here, maybe another PR: 

- ~~deprecate `TARGET_BLOCK_FULLNESS` and attach it to the type that is used as `FeeMultiplierUpdate`. (probably learn from `DealWithFees` and how they attach a const to a type).~~ DONE here https://github.com/paritytech/substrate/pull/3823


node/runtime/src/lib.rs
```diff
@@ -504,6 +511,7 @@ construct_runtime!(
 		Authorship: authorship::{Module, Call, Storage, Inherent},
 		Indices: indices,
 		Balances: balances::{default, Error},
+		TransactionPayment: transaction_payment::{Module, Storage}, // TODO double check this
```
kianenigma: Is this the correct?

srml/transaction-payment/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! # Transaction Payment Module
+//!
+//! This module provides the basic logic needed to pay the absolute minimum amount needed for a
+//! transaction to be included. This includes:
+//!   - _weight fee_: A fee proportional to amount of Weight a transaction consumes. (TODO: explain
```
kianenigma: I should generally improve this doc.

srml/balances/src/mock.rs
```diff
@@ -92,34 +71,38 @@ impl system::Trait for Runtime {
...
 	type MaximumBlockLength = MaximumBlockLength;
 	type AvailableBlockRatio = AvailableBlockRatio;
 	type Version = ();
 }
+parameter_types! {
+	pub const TransactionBaseFee: u64 = 0;
+	pub const TransactionByteFee: u64 = 1;
+}
+impl transaction_payment::Trait for Runtime {
+		type Currency = Module<Runtime>;
```
xlc: indent

kianenigma: already outdate :P thanks

node/runtime/src/lib.rs
JoshOrndorff: Yes

JoshOrndorff:
@kianenigma Let's work together on the docs. Here's an issue and PR to track progress on it with some initial sketching.
https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/284
https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/288

I'm also going to add this to Substrate recipes.

node/executor/src/lib.rs
```diff
@@ -88,17 +88,15 @@ mod tests {
 	}
 
 	/// Default transfer fee
-	fn transfer_fee<E: Encode>(extrinsic: &E) -> Balance {
+	fn transfer_fee<E: Encode>(extrinsic: &E, fm: Fixed64) -> Balance {
```
thiolliere: fm stands for fee multiplier ? I would maybe write this in the doc of the function and rename to a more sensible name

node/executor/src/lib.rs
```diff
@@ -217,6 +215,10 @@ mod tests {
 			None,
 		).0;
 		assert!(r.is_ok());
+
+		let mut fm: Fixed64 = Default::default();
+		t.execute_with(|| { fm = TransactionPayment::next_fee_multiplier(); });
```
thiolliere:
```suggestion
		let fm = t.execute_with(TransactionPayment::next_fee_multiplier);
```

node/executor/src/lib.rs
thiolliere: I mean just for this one because it is in function argument, later use of fm are fine

srml/transaction-payment/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+};
+
+type Multiplier = Fixed64;
+type BalanceOf<T> =
+	<<T as Trait>::Currency as Currency<<T as system::Trait>::AccountId>>::Balance;
+type NegativeImbalanceOf<T> =
+	<<T as Trait>::Currency as Currency<<T as system::Trait>::AccountId>>::NegativeImbalance;
+
+pub trait Trait: system::Trait {
+	/// The currency type of the chain.
```
thiolliere: actually one chain could use different currency type. 

I can suggest this,
```suggestion
	/// The currency type used.
```
but it is not better :-/

srml/transaction-payment/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+	type TransactionBaseFee: Get<BalanceOf<Self>>;
+
+	/// The fee to be paid for making a transaction; the per-byte portion.
+	type TransactionByteFee: Get<BalanceOf<Self>>;
+
+	/// Convert a weight value into a deductible fee based on the currency type.
+	type WeightToFee: Convert<Weight, BalanceOf<Self>>;
+
+	/// Update the multiplier of the next block, based on the previous block's weight.
+	// TODO: maybe this does not need previous weight and can just read it
```
thiolliere: But for polkadot we do need previous block weight isn't it ?
then we can remove this todo, someone can implement a constant function indeed and this is fine.

srml/transaction-payment/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+		adjusted_fee.saturating_add(tip)
+	}
+}
+
+#[cfg(feature = "std")]
+impl<T: Trait + Send + Sync> rstd::fmt::Debug for ChargeTransactionPayment<T> {
+	fn fmt(&self, f: &mut rstd::fmt::Formatter) -> rstd::fmt::Result {
+		write!(f, "ChargeTransactionPayment<{:?}>", self.0)
+	}
+}
```
thiolliere: isn't the derive Default good enough ?

srml/transaction-payment/src/lib.rs
kianenigma: I am still not sure about it. 

If possible, I would like to remove the hardcoded function api and make it just a simple `Convert<Fixed64, Fixed64>`. But first, a technical issue that I haven't really looked into: storage is not readable outside of externalities and I haven't looked into how it can be done here.

thiolliere: oh ok that make sense then,

also this convert is called inside externalities.

srml/transaction-payment/src/lib.rs
JoshOrndorff: The currency in which fees will be paid.

node/executor/src/lib.rs
kianenigma: much better, thanks

srml/transaction-payment/src/lib.rs
kianenigma: This convert is called inside externalities, but if I want to read directly from the storage of `System` in `node/runtime/src/impl.rs`, then I am not in externalities (right?) and for this, unfortunately, we have to hardcode the weight as the argument. Hopefully this can be further refactored.

thiolliere: you can do, if AllExtrinsicsWeight is public:
```rust
system::AllExtrinsicsWeight::get()
```
At runtime if it called inside some externalities then this is will run ok otherwise it will crash with an error.

srml/transaction-payment/src/lib.rs
thiolliere: oh ya indeed :-/

PR #3817, opened 2019-10-14
Fix asynchronous transaction rejections.

tomusdrw:
Before #3650 the transactions were imported to the pool in a synchronous manner,. Because of that it was simple for `submit_and_watch_extrinsic` subscription to return an error if the transaction was rejected, instead of starting the subscription.

The async refactoring missed one case that handles the asynchronous verification in the pool - errors in that verification only lead to a warning and the subscription was started anyway, but never produced any message.

This PR fixes this behavior. CC @jacogr

### DIFF
--- a/core/rpc/src/author/mod.rs
+++ b/core/rpc/src/author/mod.rs
@@ -164,11 +164,10 @@ impl<B, E, P, RA> AuthorApi<ExHash<P>, BlockHash<P>> for Author<B, E, P, RA> whe
 			Ok(
 				self.pool
 					.submit_and_watch(&generic::BlockId::hash(best_block_hash), dxt)
-					.map(|res| res.map_err(|e|
-						e.into_pool_error()
-							.map(error::Error::from)
-							.unwrap_or_else(|e| error::Error::Verification(Box::new(e)).into())
-					))
+					.map_err(|e| e.into_pool_error()
+						.map(error::Error::from)
+						.unwrap_or_else(|e| error::Error::Verification(Box::new(e)).into())
+					)
 			)
 		};
 
--- a/core/rpc/src/author/mod.rs
+++ b/core/rpc/src/author/mod.rs
@@ -26,10 +26,9 @@ use log::warn;
 use client::{self, Client};
 use rpc::futures::{
 	Sink, Future,
-	stream::Stream as _,
 	future::result,
 };
-use futures03::{StreamExt as _, compat::Compat};
+use futures03::{StreamExt as _, compat::Compat, future::ready};
 use api::Subscriptions;
 use jsonrpc_pubsub::{typed::Subscriber, SubscriptionId};
 use codec::{Encode, Decode};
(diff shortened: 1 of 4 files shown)


core/rpc/src/author/tests.rs
```diff
@@ -31,6 +31,9 @@ use test_client::{
 	self, AccountKeyring, runtime::{Extrinsic, Transfer, SessionKeys}, DefaultTestClientBuilderExt,
 	TestClientBuilderExt,
 };
+use rpc::futures::{
```
bkchr: The curly braces can be removed.

core/rpc/src/author/tests.rs
```diff
@@ -132,6 +135,29 @@ fn should_watch_extrinsic() {
...
+	let client = Arc::new(test_client::new());
+	let pool = Arc::new(Pool::new(Default::default(), FullChainApi::new(client.clone())));
+	let keystore = KeyStore::new();
+	let p = Author {
+		client,
+		pool: pool.clone(),
+		subscriptions: Subscriptions::new(Arc::new(runtime.executor())),
+		keystore: keystore.clone(),
+	};
+	let (subscriber, id_rx, _data) = ::jsonrpc_pubsub::typed::Subscriber::new_test("test");
```
bkchr:
```suggestion
	let (subscriber, id_rx, _data) = jsonrpc_pubsub::typed::Subscriber::new_test("test");
```

core/rpc/src/author/mod.rs
```diff
@@ -162,42 +161,45 @@ impl<B, E, P, RA> AuthorApi<ExHash<P>, BlockHash<P>> for Author<B, E, P, RA> whe
...
-					.map(error::Error::from)
-					.unwrap_or_else(|e| error::Error::Verification(Box::new(e)).into())
-				))
+			Ok(
+				self.pool
+					.submit_and_watch(&generic::BlockId::hash(best_block_hash), dxt)
+					.map(|res| res.map_err(|e|
+						e.into_pool_error()
+							.map(error::Error::from)
+							.unwrap_or_else(|e| error::Error::Verification(Box::new(e)).into())
```
bkchr: Why don't we use a second `map_err` here?
We could also just make one `map_err` that outputs the correct error directly.

core/rpc/src/author/tests.rs
```diff
@@ -132,6 +135,29 @@ fn should_watch_extrinsic() {
 	);
 }
 
+#[test]
+fn should_return_watch_validation_error() {
```
bkchr: Is that testing the problem that was described?

tomusdrw: Yes, the test fails on `master` and passes here.

jacogr:
Tested this branch, and it fixes the issue, actual RPC outputs with a previous failure -

```
2019-10-15 10:41:49          API-WS: calling author_submitAndWatchExtrinsic {"id":10,"jsonrpc":"2.0","method":"author_submitAndWatchExtrinsic","params":["0x350283ff0c9f6509b840bea296f9e75a9c730be8b07c4ba6b1d4449e991c97c2b158ad4e00a167c64820acc31551c49f2425aa914b137a351bb11e558f995777e40f774b1b8a3636c15a0b06004dff3d570792a2d31804cef33820af76471f1341421f0c850100000600ff8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a483a85f102"]}
2019-10-15 10:41:49          API-WS: received {"jsonrpc":"2.0","error":{"code":1010,"message":"Invalid Transaction","data":"Payment"},"id":10}
```

(Previously, as well as on Kusama, used to get a subscription id back, not the actual error)

PR #3820, opened 2019-10-15
Auto import SimpleDispatchInfo

kianenigma:
Similar to how we recently brought common things like `StorageValue, StorageMap,...` into scope, I assume it is good to use also do the same with `SimpleDispatchInfo` as it is used everywhere. 

Not sure if it is needed, some might say it is ambiguous. Hence, I only touched example, balances and system. If approved, will fix the rest of srml.


bkchr:
I'm against this. We should never include stuff into the scope automatically, as this maybe breaks some stuff. The `Storage..` stuff are traits and we have done this move after we had `SomeTrait as _` to not clash with any other import.

kianenigma:
That does sound reasonable. It would also be an unused import if the user wants to use a custom weight type.

PR #3821, opened 2019-10-15
Runtime logging.

tomusdrw:
This PR introduces a way to log formatted messages from the runtime. 

# Disclaimer
It's extremely important to note that doing any logging in the runtime may significantly increase the size of the WASM blob and will affect the performance, so it should be used very carefuly!

I've made a bunch of tests to ensure that the runtime size is **NOT** affected by this changes at all - the compiler is smart enough to remove dead code (i.e. unused `Debug` implementations), but still extra care has been taken to avoid increasing the size by accident (namely the `RuntimeDebug` custom derive).

The most important things in this PR:

### `RuntimeDebug` (`primitives/debug-derive`)
A custom derive for `core::fmt::Debug` which generates an **empty implementation** for WASM, but regular `Debug` when running `std`.
This allows us to:
1. Remove `MaybeDebug` traits and just use `rstd::fmt::Debug`
2. Sprinkle logs that are executed even in WASM, but are more lightweight because they don't print the struct content (just a message).
3. If full `Debug` is required it can be easily changed to `derive(Debug)` or custom implementation.

### `srml_support::debug`

A support module exporting macros to print log entries (stderr; re-exports from `log` crate). Logging from WASM requires calling `RuntimeLogger::init()` to pass the logs to the host side (note it affects performance and size of the blob).
Also a wrapper for `sr_io::print` is introduced, allowing to print more structured messages (note it should be used sparingly, as it increases the size of the wasm blob).

A lot of use cases would be totally ok if logging worked only if the code is running natively (i.e. avoiding the impact to wasm blob size / perf). We introduce `native` module to support that case.
The same macros are exported from there, but they are no-op on WASM.
This should be a preferred method of logging / printing.

## Rest
The rest is just replacing trait bounds and conditional `Debug` implementation with `RuntimeDebug`.


core/primitives/debug-derive/src/impls.rs
```diff
@@ -0,0 +1,222 @@
...
+use quote::quote;
+use proc_macro2::TokenStream;
+use syn::{Ident, Data, DeriveInput, parse_quote};
+
+fn ident2str(name: &Ident) -> String {
+	format!("{}", name)
+}
+
+pub fn debug_derive(ast: &DeriveInput) -> proc_macro::TokenStream {
+	let name_str = ident2str(&ast.ident);
```
bkchr:
```suggestion
	let name_str = ast.ident.to_string();
```

core/primitives/debug-derive/src/impls.rs
```diff
@@ -0,0 +1,222 @@
...
+			fn fmt(&self, fmt: &mut core::fmt::Formatter) -> core::fmt::Result {
+				#implementation
+			}
+		}
+	};
+
+	gen.into()
+}
+
+#[cfg(not(any(test, feature = "std")))]
```
bkchr: Did you thought of using `CARGO_CFG_STD`?

core/primitives/debug-derive/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+
+extern crate proc_macro;
+
+mod impls;
+
+use proc_macro::TokenStream;
+
+#[proc_macro_derive(RuntimeDebug)]
+pub fn debug_derive(input: TokenStream) -> TokenStream {
+	match syn::parse(input) {
```
bkchr: Why not use `parse_macro_input!`?

core/primitives/debug-derive/src/impls.rs
tomusdrw: What does it do? I've actually changed the condition a bit, because `test` is unnecessary (we run tests in `std` anyway).

core/primitives/debug-derive/src/lib.rs
tomusdrw: We pretty much handle all structs and enums anyway. I think the custom type would pretty much look exactly like the generic ones from `syn`.
Unless I'm missing some obvious advantage.

core/primitives/debug-derive/src/impls.rs
bkchr: I mean, instead of having the feature exposed by the derive macro, just check if the `CARGO_CFG_STD` is set while compiling a crate. Cargo sets environment variables for each cfg that is enabled for a crate and as the proc-macro is running as part of the build process, you can inspect this environment variables.

core/primitives/debug-derive/src/lib.rs
bkchr:
```
impls::debug_derive(&parse_macro_input!(input))
```

This would be all you for the function. It automatically returns an error internally, if the parsing failes.

core/primitives/debug-derive/src/impls.rs
tomusdrw: Seems that the env variable is not set when compiling tests and I also find this approach more fragile than the usual feature-based conditional compilation.

core/sr-arithmetic/src/per_things.rs
```diff
@@ -26,8 +26,8 @@ macro_rules! implement_per_thing {
 		/// A fixed point representation of a number between in the range [0, 1].
 		///
 		#[doc = $title]
-		#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug, Ord, PartialOrd))]
-		#[derive(Encode, Decode, Default, Copy, Clone, PartialEq, Eq, CompactAs)]
+		#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Ord, PartialOrd))]
+		#[derive(Encode, Decode, Default, Copy, Clone, PartialEq, Eq, Debug, CompactAs)]
```
bkchr: Why not `RuntimeDebug` here?

core/sr-arithmetic/src/rational128.rs
```diff
@@ -19,8 +19,7 @@ use crate::helpers_128bit;
 use num_traits::Zero;
 
 /// A wrapper for any rational number with a 128 bit numerator and denominator.
-#[derive(Clone, Copy, Default, Eq)]
-#[cfg_attr(feature = "std", derive(Debug))]
+#[derive(Clone, Copy, Default, Eq, Debug)]
```
bkchr: Same.

core/sr-io/Cargo.toml
```diff
@@ -9,15 +9,17 @@ edition = "2018"
 rustc_version = "0.2.3"
 
 [dependencies]
-rstd = { package = "sr-std", path = "../sr-std", default-features = false }
-primitives = { package = "substrate-primitives", path = "../primitives", default-features = false }
 codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false }
+environmental = { version = "1.0.1", optional = true }
```
bkchr: This is not used anymore in `sr-io`.

core/sr-io/with_std.rs
```diff
@@ -192,6 +192,22 @@ impl OtherApi for () {
...
+		message: &[u8],
+	) {
+		let target = std::str::from_utf8(target).unwrap_or("invalid utf8");
+		let msg = std::str::from_utf8(message).unwrap_or("invalid utf8");
+
+		log::log!(
+			target: target,
+			log::Level::from(level),
+			"{}",
+			msg
```
bkchr:
```suggestion
			msg,
```

core/sr-primitives/src/traits.rs
```diff
@@ -1147,9 +1167,12 @@ impl Printable for &str {
 	}
 }
 
-impl Printable for u64 {
+#[impl_for_tuples(1, 12)]
+impl Printable for Tuple {
+	for_tuples!( where #( Tuple: Printable )* );
```
```suggestion
```

This is done automatically :) 

The syntax is only required when you need another bound.

srml/contracts/src/lib.rs
```diff
@@ -988,9 +985,14 @@ impl<T: Trait + Send + Sync> Default for CheckBlockGasLimit<T> {
 	}
 }
 
-#[cfg(feature = "std")]
-impl<T: Trait + Send + Sync> std::fmt::Debug for CheckBlockGasLimit<T> {
-	fn fmt(&self, _: &mut std::fmt::Formatter) -> std::fmt::Result {
+impl<T: Trait + Send + Sync> rstd::fmt::Debug for CheckBlockGasLimit<T> {
+	#[cfg(feature = "std")]
+	fn fmt(&self, f: &mut rstd::fmt::Formatter) -> rstd::fmt::Result {
+		write!(f, "CheckBlockGasLimit<T>")
```
bkchr:
```suggestion
		write!(f, "CheckBlockGasLimit")
```

srml/support/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
 [dependencies]
+log = "0.4"
```
bkchr: Don't we need to disable the `std` feature?

srml/support/src/debug.rs
```diff
@@ -0,0 +1,211 @@
...
+macro_rules! runtime_print {
+	($($arg:tt)+) => {
+		use core::fmt::Write;
+		let mut w = $crate::debug::Writer::default();
+		let _ = core::write!(&mut w, $($arg)+);
+		w.print();
+	}
+}
+
+pub use crate::runtime_print;
```
bkchr: Above you import it and re-export it as `print`. I think one should be enough :D

srml/system/src/lib.rs
```diff
@@ -540,6 +541,7 @@ impl<T: Trait> Module<T> {
 	/// This will update storage entries that correspond to the specified topics.
 	/// It is expected that light-clients could subscribe to this topics.
 	pub fn deposit_event_indexed(topics: &[T::Hash], event: T::Event) {
+		support::runtime_print!("{:?}", event);
```
bkchr: Debug leftovers?

srml/system/src/lib.rs
```diff
@@ -826,6 +828,7 @@ impl<T: Trait + Send + Sync> SignedExtension for CheckWeight<T> {
 		info: DispatchInfo,
 		len: usize,
 	) -> Result<(), ApplyError> {
+		support::runtime_print!("{:?}", _call);
```
bkchr: Leftover?

srml/system/src/lib.rs
```diff
@@ -994,9 +1012,14 @@ impl<T: Trait + Send + Sync> SignedExtension for CheckEra<T> {
 #[derive(Encode, Decode, Clone, Eq, PartialEq)]
 pub struct CheckGenesis<T: Trait + Send + Sync>(rstd::marker::PhantomData<T>);
 
-#[cfg(feature = "std")]
-impl<T: Trait + Send + Sync> rstd::fmt::Debug for CheckGenesis<T> {
-	fn fmt(&self, _f: &mut rstd::fmt::Formatter) -> rstd::fmt::Result {
+impl<T: Trait + Send + Sync> Debug for CheckGenesis<T> {
+	#[cfg(feature = "std")]
+	fn fmt(&self, f: &mut rstd::fmt::Formatter) -> rstd::fmt::Result {
+		write!(f, "CheckGenesis<T>")
```
```suggestion
		write!(f, "CheckGenesis")
```

srml/system/src/lib.rs
```diff
@@ -1023,9 +1046,14 @@ impl<T: Trait + Send + Sync> SignedExtension for CheckGenesis<T> {
 #[derive(Encode, Decode, Clone, Eq, PartialEq)]
 pub struct CheckVersion<T: Trait + Send + Sync>(rstd::marker::PhantomData<T>);
 
-#[cfg(feature = "std")]
-impl<T: Trait + Send + Sync> rstd::fmt::Debug for CheckVersion<T> {
-	fn fmt(&self, _f: &mut rstd::fmt::Formatter) -> rstd::fmt::Result {
+impl<T: Trait + Send + Sync> Debug for CheckVersion<T> {
+	#[cfg(feature = "std")]
+	fn fmt(&self, f: &mut rstd::fmt::Formatter) -> rstd::fmt::Result {
+		write!(f, "CheckVersion<T>")
```
```suggestion
		write!(f, "CheckVersion")
```

core/primitives/debug-derive/src/impls.rs
```diff
@@ -0,0 +1,218 @@
...
+			vars: Vec<Ident>,
+		},
+		Named {
+			names: Vec<Ident>,
+			this: Option<Ident>,
+		},
+	}
+
+	impl Fields {
+		fn new<'a>(fields: impl Iterator<Item=&'a syn::Field>, this: Option<Ident>) -> Self {
```
bkchr: You want to use [this](https://docs.rs/syn/1.0.5/syn/token/struct.SelfValue.html) for `this`.

srml/support/src/debug.rs
tomusdrw: The other one is conditional.

tomusdrw:
> Now we generate just an empty implementation for Debug on no_std, didn't you said somewhere that you at least want to output the name on no_std? Or did I dream this?

Yes, I was running a bunch of tests and chatted with you in DMs. Outputing the type name is fine - according to the tests I did, the most code is generated by long enum matches (so if you want to print a variant as well).

So currently the main goals are:
1. Get rid of `MaybeDebug`, and conditional `Debug` implementation to simplify code.
2. Avoid accidentally bloating the WASM blob by putting a log statement that prints out a huge struct / enum.

I figured that printing out the type is not really that useful, cause there are really no places where you actually don't know the type statically already. If you put a log line:
`log::log!("Something happened: {:?}", call)`

IT really doesn't matter that much if you get the type of the `call` in the log line or not as the log line most likely idenitifies your place uniquely anyway and then you can just read the type from the code.

However I'm totally fine with printing the type in `Debug` - it will increase the wasm blob size only slightly (obviously only if you actually log that particular struct, as the rest is removed by the compiler anyway).

srml/support/src/debug.rs
bkchr: https://github.com/paritytech/substrate/pull/3821/files/4fba98da71c79869165c0acec0de4d1899a0fb2e#diff-77a9e51bda04f3b5bf1ee2a5de1b7864R93 hmm?

bkchr:
> E.g. if you put a log line: `log::log!("Something happened: {:?}", call)` it really doesn't matter that much if you get the type of the `call` in the log line or not as the log line most likely idenitifies your place uniquely anyway and then you can just read the type from the code.
> 
> However I'm totally fine with printing the type in `Debug` - it will increase the wasm blob size only slightly (obviously only if you actually log that particular struct, as the rest is removed by the compiler anyway).

I think the problem will be that people don't understand and will open 1mio issues for this. As I understand the bloat size, could we try to just print "Stripped out in wasm" or something similar, to make the people aware that it actually called the `Debug` implementation. I would hope that the compiler/linker is that smart to put the same string into the same section and so, we only have this once in the final binary. If the compiler/linker are not smart enough, we need to put this message into some global const, to make the compiler/linker aware that it is always the same.

tomusdrw:
@bkchr Yup, that actually sounds good. I'm 99% sure compiler is smart enough to deduplicate `&'static str`, but will run a test to confirm it :)

srml/support/src/debug.rs
tomusdrw: Oh, you caught me :D

srml/support/Cargo.toml
tomusdrw: I think it's actually off by default: https://github.com/rust-lang-nursery/log/blob/master/Cargo.toml#L40

PR #3822, opened 2019-10-15
Add a note about --document-private-items to the readme

expenses:
I had totally forgotten about this flag, but it's quite useful, so it would be nice to have a note about it.


thiolliere:
oh interesting I wasn't aware of this flag, the generated doc doesn't show if item is private or public though.

bkchr:
Our readme is not for documenting cargo doc features.

Nevertheless Ty for the pr, but I will close this.

PR #3823, opened 2019-10-15
Better Parameterisation for Fee system

kianenigma:
Makes is slightly more _substrate-style_ by using `Get` instead of floating constant or hardcoded values. 

--- 

- [ ] Needs a Polkadot PR.


core/sr-primitives/src/sr_arithmetic.rs
```diff
@@ -557,6 +557,12 @@ impl Fixed64 {
 		Self(int.saturating_mul(DIV))
 	}
 
+	/// Return the inner value. Only for testing.
+	#[cfg(any(feature = "std", test))]
```
thiolliere: if this is only for testing, then why not ?
```suggestion
	#[cfg(test)]
```

kianenigma: Fair point and I (naively) used the gating feature that I used here https://github.com/paritytech/substrate/pull/3823/files#diff-24bd3cac4cae1c620f82f7aa79ef28ceR707

Declaring the function with just `#[cfg(test)]` was not enough and it cannot be found in tests though 🤔  and I had to make it `test or std`. Do you know the reason of the difference

kianenigma: I agree that logically it should be just `cfg[test]`. Something is wrong maybe in cargo files

thiolliere: oh actually this is because the test configure is only set for the tested crate.
if you test the runtime then sr-arithmetic isn't compiled with `test`.

I don't know what is the prefered way to solve this

thiolliere:
code is OK but wonder about API again maybe we would prefer not having this code in runtime/impl.rs as probably quite some other runtime will implement the same.

In contracts it is organized a bit differently and maybe we should inspiration from it:
on constracts::trait there is this 
```rust
type ComputeDispatchFee: ComputeDispatchFee<<Self as Trait>::Call, BalanceOf<Self>>;
```

and then in contracts crate there is :
```rust
/// The default dispatch fee computor computes the fee in the same way that
/// the implementation of `ChargeTransactionPayment` for the Balances module does. Note that this only takes a fixed
/// fee based on size. Unlike the balances module, weight-fee is applied.
pub struct DefaultDispatchFeeComputor<T: Trait>(PhantomData<T>);
impl<T: Trait> ComputeDispatchFee<<T as Trait>::Call, BalanceOf<T>> for DefaultDispatchFeeComputor<T> {
	fn compute_dispatch_fee(call: &<T as Trait>::Call) -> BalanceOf<T> {
              ...
	}
}
```

core/sr-primitives/src/sr_arithmetic.rs
kianenigma: (won't work as discussed on off-band. will leave the comment open)

kianenigma:
So the dilemma is to leave it in `impl.rs` or move it to `srml/transaction-payment`. 

In practice, I am fine with both. We are providing a default behaviour for something which at the end of the day the user should define. 

But what I don't like about putting this `targetedAdjustment` in a __substrate module__ is that it is too `polkadot`-specific. In otherwords, naming a polkadot requirement as the default of substrate is not good. Although, I have made the struct configurable here by being able to change the 25% value. But still the whole logic of "Changing fees based on congestion" seems polkadot-specific to me.

 I am even tempted to wipe this ting from substrate node and only leave it in polkadot repo. Substrate node can just use `()` or `ConvertInto`, meaning that fees will not change over time.

4meta5:
> But what I don't like about putting this targetedAdjustment in a substrate module is that it is too polkadot-specific. In otherwords, naming a polkadot requirement as the default of substrate is not good.

Agreed, I am pro putting this logic in Polkadot and using `()` for the Substrate node. I appreciate the separation and think the default for a Substrate node should be no change in fees over time. In this case, documentation could guide users to configure fees in a similar way to Polkadot as an example (@JoshOrndorff  should be kept aware bc of https://github.com/substrate-developer-hub/recipes/pull/62), but I really don't like having the Polkadot slow-adjusting fee model as the default in Substrate...

I can expand on this, but I basically don't expect most developers building with Substrate to adopt the same fee structure as a relay chain. The economics of parachains/sovereign chains should and will be very different. Personally, I think something like [BRAQ](https://github.com/Joystream/whitepaper/blob/master/paper.pdf) (section 4.2, page 9) or some similar model for subsidizing actions within a defined period of blocks might be more appropriate than charging per method call.

node/runtime/src/impls.rs
```diff
@@ -155,50 +144,67 @@ mod tests {
...
-		let mut fm = Fixed64::default();
-		let mut iterations: u64 = 0;
-		loop {
-			let next = FeeMultiplierUpdateHandler::convert((block_weight, fm));
-			fm = next;
-			if fm == Fixed64::from_rational(-1, 1) { break; }
-			iterations += 1;
-		}
-		println!("iteration {}, new fm = {:?}. Weight fee is now zero", iterations, fm);
+		let block_weight = 0;
```
thiolliere: is it on purpose that block weight changed from 1000 to 0 ?

kianenigma: yeah just makes a bit more sense: the test is supposed to simulate an _empty_ chain. Nonetheless the test is not that important as you see and just runs a simulation.

kianenigma:
@thiolliere we are writing a doc about weights and I thought I would keep this Iced to apply very small changes to docs and prevent opening new PRs every day.

PR #3826, opened 2019-10-15
Persist block announcements

arkpar:
Currently block announcements may be ignored occasionally. This PR changes handling announcements in the same way as explicit sync requests.

### DIFF
--- a/core/network/src/protocol.rs
+++ b/core/network/src/protocol.rs
@@ -1122,11 +1122,12 @@ impl<B: BlockT, S: NetworkSpecialization<B>, H: ExHashT> Protocol<B, S, H> {
 
 		match self.sync.on_block_announce(who.clone(), hash, &announce, is_their_best) {
 			sync::OnBlockAnnounce::Nothing => {
-				// try_import is only true when we have all data required to import block
+				// `on_block_announce` returns `OnBlockAnnounce::ImportHeader`
+				// when we have all data required to import the block
 				// in the BlockAnnounce message. This is only when:
 				// 1) we're on light client;
 				// AND
-				// 2) parent block is already imported
+				// 2) parent block is already imported and not pruned.
 				return CustomMessageOutcome::None
 			}
 			sync::OnBlockAnnounce::ImportHeader => () // We proceed with the import.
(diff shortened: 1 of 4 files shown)


core/network/src/protocol.rs
```diff
@@ -1121,18 +1121,12 @@ impl<B: BlockT, S: NetworkSpecialization<B>, H: ExHashT> Protocol<B, S, H> {
...
-			}
 			sync::OnBlockAnnounce::Nothing => {
 				// try_import is only true when we have all data required to import block
 				// in the BlockAnnounce message. This is only when:
 				// 1) we're on light client;
 				// AND
-				// - EITHER 2.1) announced block is stale;
-				// - OR 2.2) announced block is NEW and we normally only want to download this single block (i.e.
-				//           there are no ascendants of this block scheduled for retrieval)
+				// 2) parent block is already imported
```
arkpar: @svyatonik please check if this is correct

core/network/src/protocol/sync.rs
```diff
@@ -1009,79 +1042,42 @@ impl<B: BlockT> ChainSync<B> {
...
-				}
-			}
+		let requires_additional_data = !self.role.is_light() || !known_parent;
+		if !requires_additional_data {
+			trace!(target: "sync", "Importing new header announced from {}: {} {:?}", who, hash, header);
+			return OnBlockAnnounce::ImportHeader
 		}
 
-		if ancient_parent {
+		if ancient_parent && !requires_additional_data {
```
svyatonik: There's a `if !requires_additional_data {}` right above - so either this `if` should be moved before the other one, or we can merge both in single one with `if ancient_parent { OnBlockAnnounce::Nothing } else { OnBlockAnnounce::ImportHeader }`. At the same time, I've noticed that currently on light client `fn block_status()` never returns `InChainPruned`. I'll prepare a follow-up PR that returns `InChainPruned` for headers that have been replaced by CHT.

core/network/src/protocol.rs
svyatonik: I'd change that to smth like "parent block is already imported and not pruned" (given my other comment). Also "try_import is only true" should be changed to "on_block_announce returns OnBlockAnnounce::ImportHeader".

core/network/src/protocol/sync.rs
arkpar: I suppose there should be a distinction between pruned state and pruned block here. For now the code treats any kind of pruning as a no go for import.

Issue #2572, opened 2019-05-14
Refactor TestExternalities


thiolliere:
As I wanted to extend functionnalities of TestExternalities it seems I ended up just copying the Ext implementation with InMemory backend and thins.

I can't make use of Ext directly because TestExternalities would be a self referencing struct.

I think we could update the test to create TextExternalities in two call one for creating backend stuff and another one for creating the referencing Ext.

Also test could make use of with_storage. All in all it seems code is redundant between TestExternalities, BasicExternalities and Ext

Kianenigma:
I am not much aware of everything that happens under the hood when you simply create an externalities (storage, backend etc.) so can't judge much, but already had in mind to propose: 

- We can make an approach like the `ExtBuilder` sort-of standard and refactor all modules to use it + isolate the struct in a generic way (if possible) as _the standard substrate test framewrok_.

- Not sure if it is possible but having some kind of a `before_each()` and `afte_each()` in my fantasies also sounded nice-to-have.

thiolliere:
maybe we can do that:
```
impl<H: Hasher> TestExternalities<H> where H::Out: Ord {
       /// Return externalities
       pub fn ext(&mut self) -> Ext<H, InMemory<H>, ChangesTrieInMemoryStorage<H>, NeverOffchainExt> {
               Ext::new(&mut self.overlay, &self.backend, Some(&self.changes_trie_storage), None)
       }
}
```
and then in most test we have to add `.ext()` I have a wip branch there, note test of sr-api-macros fails for a reason still unknown to me. https://github.com/paritytech/substrate/tree/gui-refactor-test

Kianenigma:
Okay, I see your changes but I see that we are sorta not on the same page and maybe they should be different issues. What I envisioned and wanted to add was a way to minimize the boilerplate needed to set up the test. I should talk to you in person about this.

thiolliere:
My main concern was I wanted to use Ext and InMemoryBackend stuff as much as possible instead of a custom implementation of Externalities.
Yes sure

thiolliere:
might be easier to solve with https://github.com/paritytech/substrate/pull/3793

thiolliere:
Test externalities is refactored, maybe we can improve test tools in general but this is another issue.

PR #3830, opened 2019-10-16
Code redundancy between ext implementation and testing.

cheme:
This PR extract common code between state db `Ext` implementation and state db testing implementation.

Code logic does not change except for:

- `Ext` deletes parent key on empty child trie root (as was testing).
- fix a wrong implementation of 'child_storage_hash'.
- when ext is use from testing it does produce some trace and memoize transaction, which is not an issue at all.


bkchr:
I don't understand the separation into `ExtBasic` and `ExtBasicMut`, what is the idea behind it?

My second question would be, why doesn't `TestExternalities` just don't stores `Ext` internally and uses it? Probably with a `DerefMut` implementation we could drop the implementation of `Externalities` for `TestExternalities`.

cheme:
> I don't understand the separation into ExtBasic and ExtBasicMut, what is the idea behind it?

Nothing really interesting there, just that I could not use `&mut` on overlay for `TestExternalities` (in 'basis' function).
Both are just a technical detail, they are just here to avoid changing function prototype to much when extraction common code.

> why doesn't TestExternalities just don't stores Ext internally and uses it?

I could do it that way, only issue is that I cannot get my &mut from the test externalities overlay for ExtBasic case, but for ExtBasicMut it should be fine.

bkchr:
By using `Ext` directly in `TestExternalities`, I would not expect your newly introduced structs to exists. If we create a trait for overlay, we can make `Ext` either store references or owned values.

cheme:
I did start putting a trait in place for overlay_change (seems interesting), but it started to be a rather big change so I moved back to the easy/quick removal of ExtBasis by considering the case where 'TestExternality' is not mut to be irrelevant.

thiolliere:
(note that TestExternalities are most of the time created just to call execute_with on it, maybe we can refactor by breaking this TestExternalities structure as well)

core/state-machine/src/testing.rs
```diff
@@ -48,6 +46,20 @@ pub struct TestExternalities<H: Hasher<Out=H256>=Blake2Hasher, N: ChangesTrieBlo
...
+		Ext {
+			overlay: &mut self.overlay,
+			backend: &self.backend,
+			_phantom: std::marker::PhantomData,
+			storage_transaction: None,
+			changes_trie_storage: Some(&self.changes_trie_storage),
+			changes_trie_transaction: None,
+			id: 0,
+			extensions: None,
+		}
```
thiolliere: the only difference with Ext::new is that `id` is not random isn't it ? is `id` important to be zero here ? otherwise just use Ext::new

thiolliere:
I don't know about `Ext deletes parent key on empty child trie root (as was testing).`

but otherwise I approve

core/state-machine/src/testing.rs
```diff
@@ -118,7 +130,8 @@ impl<H: Hasher<Out=H256>, N: ChangesTrieBlockNumber> TestExternalities<H, N> {
 	///
 	/// Returns the result of the given closure.
 	pub fn execute_with<R>(&mut self, execute: impl FnOnce() -> R) -> R {
-		externalities::set_and_run_with_externalities(self, execute)
+		let mut ext = self.ext();
+		externalities::set_and_run_with_externalities(&mut ext, execute)
 	}
```
thiolliere: ext is dropped, does it contains some value that are thus discarded ?

the fields: `storage_transaction` and `changes_trie_transaction` are not kept after this execute_with, this is a change, does it has some consequences ?

bkchr: As we are not really storing `TestExternalities` in the database, I think this is fine.

bkchr:
@thiolliere didn't you had an issue for this?

core/state-machine/src/testing.rs
cheme: It is without consequence at this point (it was one of the main difference between both implementation), but it could be good to add those to the test externalities struct (currently it means there was probably no low level test on this memoization)).

cheme:
I still need to fix broken tests (a few many).

thiolliere:
Fix https://github.com/paritytech/substrate/issues/2572

bkchr:
@cheme Please implement `Deref` and `DerefMut` for `TestExternalities`. This should make fixing the tests much easier.

cheme:
I cannot implement 'Deref' (except by using inner mutability). Not sure I want to go in this direction.

bkchr:
Hmm okay :(

bkchr:
@cheme tests still failing.

core/state-machine/src/testing.rs
thiolliere: > (currently it means there was probably no low level test on this memoization)).

Or not between multiple call to execute_with.

thiolliere:
looks good to me

bkchr:
@cheme still not compiling.

cheme:
Thanks for noticing it again, I merged a bit too fast :(

PR #3832, opened 2019-10-16
gossip: futures 03 Receiver

kigawas:
#3099


gavofyork:
Need resolving

core/finality-grandpa/src/communication/mod.rs
```diff
@@ -145,12 +146,19 @@ impl<B, S, H> Network<B> for Arc<NetworkService<B, S, H>> where
...
+	#[allow(deprecated)]
 	fn messages_for(&self, topic: B::Hash) -> Self::In {
 		let (tx, rx) = oneshot::channel();
 		self.with_gossip(move |gossip, _| {
-			let inner_rx = gossip.messages_for(GRANDPA_ENGINE_ID, topic);
+			let inner_rx = gossip
+				.messages_for(GRANDPA_ENGINE_ID, topic)
+				.map(|x| Ok(x))
+				.compat()
+				.boxed(); // Box::new() will break
```
bkchr:
```suggestion
				.boxed();
```

core/network/src/protocol/consensus_gossip.rs
```diff
@@ -627,6 +629,13 @@ mod tests {
 		}
 	}
 
+	macro_rules! stream_assert_eq {
+		($stream: expr, $expected_value: expr) => {
+			let mut rt = Runtime::new().unwrap(); // use tokio 0.1's runtime on stable rust
```
bkchr: Before we just used `poll` why can we not use `poll_next` now?

kigawas: Do you mean `.await`? You can only use it in async functions but test function cannot be async

bkchr: Just use [this](https://docs.rs/futures-preview/0.3.0-alpha.17/futures/executor/fn.block_on_stream.html), then you don't need to import tokio.

core/finality-grandpa/src/communication/mod.rs
```diff
@@ -145,8 +146,11 @@ impl<B, S, H> Network<B> for Arc<NetworkService<B, S, H>> where
 	S: network::specialization::NetworkSpecialization<B>,
 	H: network::ExHashT,
 {
-	type In = NetworkStream;
+	type In = NetworkStream<
+		Box<dyn Stream<Item = network_gossip::TopicNotification, Error = ()> + Send + 'static>,
+	>;
 
+	#[allow(deprecated)]
```
kianenigma: what's the deprecated piece of this fn?

core/finality-grandpa/src/communication/mod.rs
```diff
@@ -159,7 +163,11 @@ impl<B, S, H> Network<B> for Arc<NetworkService<B, S, H>> where
 		// waiting for the oneshot to resolve and from there on acting like a normal message channel.
 		let (tx, rx) = oneshot::channel();
 		self.with_gossip(move |gossip, _| {
-			let inner_rx = gossip.messages_for(GRANDPA_ENGINE_ID, topic);
+			let inner_rx = gossip
+				.messages_for(GRANDPA_ENGINE_ID, topic)
+				.map(|x| Ok(x))
+				.compat()
```
kianenigma: If I have understood correctly, at some point when the call site of `gossip.message_for()` also returns a compatible future, this can+should be removed. If this is the case, you can mark `#[allow(deprecated)]` as `TODO: #3099 ...` or something similar.

core/finality-grandpa/src/communication/mod.rs
bkchr: What is deprecated in here?

kigawas: boxed

bkchr: Then just call `Box::new()`.

kigawas: Tried, but it didn't work here on rust nightly :(

bkchr: Why?

kigawas: Np, fixed

PR #3833, opened 2019-10-16
Recover transaction pool on light client

svyatonik:
It has been broken since #3538 

TLDR: this PR introduces 'generic' light tx pool, which should work on all chains. There's a way to override different parts of this 'generic' implementation with chain-specific implementations. The chain specific implementation could, sometimes, be better than generic.

**PART 1. What needs to be implemented.**

Basically, it is:
1.1) validation of transactions on light client;
1.2) removal of 'mined' transactions on light client;
1.3) removal of 'invalid' transactions on light client;
1.4) moving transactions between 'ready' and 'future' queues on light client.

**PART 2. Validation of transactions on light client.**

Options that I have considered:
2.1) do not validate transactions at all - just broadcast to full nodes
2.1.PROS) fastest option;
2.1.CONS) I assume the submitter wants to know if his transaction is valid asap + it is impossible to prune invalid transactions from the pool just by looking at the blocks (it will never be included);
2.2) require specific implementation for every chain:
2.1.PROS) chain-specific implementation could be simpler/faster than generic. Like we could use proof-of-read of account nonce instead of proof-of-call of `fn TaggedTransactionQueue::validate_transaction()`. Or we might have a chain where validation could be performed on full node, if all required information is available from block headers;
2.2.CONS) it isn't always possible to write better than generic validation + writing specific implementation is an optimization task + we might want to get slow, but functional light client in the beginning;
2.3) provide generic implementation, but allow overriding it by chain-specific implementations:
2.3.PROS) easy-to-start + might be optimized later;
2.3.CONS) requires proof-of-call from remote nodes

So I've chose the option 2.3. In the code it is in the [`LightChainApi`](https://github.com/paritytech/substrate/blob/34ea521e009115e6bd3184d8f16e96c6ed907941/core/transaction-pool/src/api.rs#L130). The specific implementation could be provided when transaction pool [is created](https://github.com/paritytech/substrate/blob/34ea521e009115e6bd3184d8f16e96c6ed907941/node/cli/src/service.rs#L277).

**Part 3. Removal of 'mined' transactions on light client.**

Options that I have considered:
3.1) chain-specific removal. E.g. we can fetch proof-of-read of account nonce from remote node and prune all transactions with nonce < than provided
3.1.PROS) may be faster if we could figure out whether tx has been included from the header
3.1.CONS) the same as in 2.2.CONS
3.2) generic removal by fetching proof-of-inclusion
3.2.PROS) -
3.2.CONS) should be fast enough for most of implementations

I've chose the option 3.2. But in this PR there's no proof-of-inclusion - instead we fetch bodies of all transactions for every block when txpool is not empty and prune these transactions from the pool. If PR will be accepted, I'll file an issue about this optimization.

**Part 4 and 5. Moving transactions between 'future', 'ready' and 'invalid' states.**

First of all, I've made a controversial decision that on light client we do not accept transactions to the 'future' queue. That's because it is impossible to know whether transaction should be moved to the 'ready' queue without revalidation. And I wanted to avoid much revalidaton calls, because they're quite heavy in 'generic' implementation. Also - 'ready'/'future' state depends on transactions that are in transaction pool. Like the same transaction could be moved to 'ready' queue if tags that it requires are satisfied by other transactions from the 'ready' queue. At the same time, on other node it could be moved to the 'future' queue if the 'satisfying' transaction is missing. That said, I'm not sure - whether this restriction makes sense, or not. So any opinion/advice is appreciated.

So after this restriction, we only have 'ready' and 'invalid' states. And the only possible transition is from 'ready' to 'invalid' state. Iiuc, on full client, this transition is detected when some of mined transactions are providing the same tags => then the 'ready' transaction is revalidated. On light client it isn't a good idea to revalidate every transaction from every block (just to receive tags that it provides). So the (not best, but functional) solution is to perform revalidation at some intervals. By default, all ready transactions are revalidated every 60 seconds or every 20 blocks (whatever happens first). These constants can be overrided. And the whole mechanism could be overrided by passing chain-specific 'TranasctionPoolMaintainer' implementation to `fn ServiceBuilder::with_transaction_pool_maintainer()`.

**Part 6, unrelated. Accounts RPC.**

When I was testing this PR, I found that "Accounts RPC methods" from node are (now?) required to submit transaction to the pool. So I've also fixed it in this PR. The fix is localized in the `node/rpc/src/*` and additional argument is added to the `fn ServiceBuilder::with_rpc_extensions()`. But if required, I could extract this fix to the separate PR.


core/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -182,18 +187,100 @@ impl<B: ChainApi> ValidatedPool<B> {
...
+		enum Status { Future, Ready, Failed, Dropped };
+
+		let (initial_statuses, final_statuses) = {
+			let mut pool = self.pool.write();
+
+			// remove all passed transactions from the ready/future queues
+			// (this may remove additional transactions as well)
+			let mut initial_statuses = HashMap::new();
+			let mut txs_to_resubmit = Vec::with_capacity(transactions.len());
+			while !transactions.is_empty() {
```
tomusdrw: Alternatively we could do `transactions.into_iter()` and maintain a `HashSet<H256>` of already removed / resubmitted transactions? Have you considered that? I wonder if the code is simpler in such case.

core/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -182,18 +187,100 @@ impl<B: ChainApi> ValidatedPool<B> {
...
+							},
+							base::Imported::Future { .. } => {
+								final_statuses.insert(hash, Status::Future);
+							},
+						},
+						Err(err) => {
+							// we do not want to fail if single transaction import has failed
+							// nor we do want to propagate this error, because it could tx unknown to caller
+							// => let's just notify listeners (and issue debug message)
+							debug!(target: "txpool", "[{:?}] Removing invalid transaction from update: {}", hash, err);
```
tomusdrw: Why not a `warn`? Do you expect it might be common?

core/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -182,18 +187,100 @@ impl<B: ChainApi> ValidatedPool<B> {
...
+			// (this may remove additional transactions as well)
+			let mut initial_statuses = HashMap::new();
+			let mut txs_to_resubmit = Vec::with_capacity(transactions.len());
+			while !transactions.is_empty() {
+				let hash = transactions.keys().next().cloned().expect("transactions is not empty; qed");
+
+				let removed = pool.remove_from_ready(&hash);
+				for removed_tx in removed {
+					let removed_hash = removed_tx.hash.clone();
+					let tx_to_resubmit = if let Some(updated_tx) = transactions.remove(&removed_hash) {
```
tomusdrw: I'm not sure I fully understand the logic here. So we can have a different transaction under `removed_hash` it seems, but yet we never really store it's hash anywhere and don't notify listeners about `hash(updated_tx)` only about `hash(removed_tx)`.

This may lead to a situation where `updated_tx` get's accepted, and we actually send an info that `hash(removed_tx)` is now ready.

srml/system/rpc/src/lib.rs
```diff
@@ -73,46 +85,135 @@ where
...
+				match maybe_best_header {
+					Some(best_header) => Ok(best_header),
+					None => Err(ClientError::UnknownBlock(format!("{}", best_hash))),
+				}
+			));
+		let future_nonce = future_best_header.and_then(move |best_header|
+			fetcher.remote_call(RemoteCallRequest {
+				block: best_hash,
+				header: best_header,
+				method: "AccountNonceApi_account_nonce".into(),
```
tomusdrw: Hmm, that seems very fragile, can't we store that on the `trait` level as consts to avoid harcoding strings like this in the code?

tomusdrw: We should at least add a test that this is the same as the runtime API.

core/transaction-pool/src/api.rs
```diff
@@ -86,3 +94,82 @@ impl<T, Block> txpool::ChainApi for FullChainApi<T, Block> where
...
+			.and_then(|header_hash| self.client.expect_header(BlockId::Hash(header_hash))
+				.map(|header| (header_hash, header)));
+		let (block, header) = match header_and_hash {
+			Ok((header_hash, header)) => (header_hash, header),
+			Err(err) => return Box::new(ready(Err(err.into()))),
+		};
+		let remote_validation_request = self.fetcher.remote_call(RemoteCallRequest {
+			block,
+			header,
+			method: "TaggedTransactionQueue_validate_transaction".into(),
```
tomusdrw: Same as with other grumble. We really need to derive these strings from the declaration/implementation.

core/transaction-pool/src/api.rs
```diff
@@ -86,3 +94,82 @@ impl<T, Block> txpool::ChainApi for FullChainApi<T, Block> where
...
+	pub fn new(client: Arc<T>, fetcher: Arc<F>) -> Self {
+		LightChainApi {
+			client,
+			fetcher,
+			_phantom: Default::default(),
+		}
+	}
+}
+
+impl<T, F, Block> txpool::ChainApi for LightChainApi<T, F, Block> where
```
tomusdrw: Any chance we could have a test for this? Or at least run transaction pool tests using `LightChainApi` or sth.

bkchr:
(this pr was accidently linked in another one as being fixed)

core/service/src/builder.rs
```diff
@@ -68,19 +72,21 @@ use transaction_pool::txpool::{self, ChainApi, Pool as TransactionPool};
...
 	keystore: Arc<RwLock<Keystore>>,
 	fetcher: Option<TFchr>,
+	remote_blockchain: Option<Arc<dyn RemoteBlockchain<TBl>>>,
 	select_chain: Option<TSc>,
 	import_queue: TImpQu,
 	finality_proof_request_builder: Option<TFprb>,
 	finality_proof_provider: Option<TFpp>,
 	network_protocol: TNetP,
 	transaction_pool: Arc<TExPool>,
+	transaction_pool_maintainer: TExPoolM,
```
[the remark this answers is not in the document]
svyatonik: OK - so I'll make `ChainApi` private (to `transaction-pool` crate) then and instead expose `TransactionPool` trait and two implementations - `FullTransactionPool` and `LightTransactionPool`.

core/transaction-pool/graph/src/base_pool.rs
```diff
@@ -83,8 +83,7 @@ pub struct PruneStatus<Hash, Ex> {
 }
 
 /// Immutable transaction
-#[cfg_attr(test, derive(Clone))]
-#[derive(PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
svyatonik: Revalidate on light node means that we: (1) read all transactions from the pool (2) call `validate_transaction()` on remote node (3) resubmit updated transactions back to the pool. This at least requires `clone()` of `Transaction.data: Extrinsic` at (1), because it is an argument of `validate_transaction()`. And there have been clones of this data before - [here](https://github.com/paritytech/substrate/blob/c12e1d1f6ed7f59f58e30310078c6c6553ceb780/core/transaction-pool/graph/src/pool.rs#L311) for example.

What happens at (3) - basically I'm removing transactions that have been in the moment (1) from the pool. That's because their requires/provides could have been updated in revalidate call. Like they could became dependent on each other, or dependency could go away, or some transactions could became invalid. But what if between (1) and (3) new transaction has been inserted to the pool and this transaction depends on some tx from (1). I had couple of options - the most straightforward was to remove this transaction from the pool too. But we don't have revalidation results for this tx => I'm using requires/provides from previous `validate()` call (which is exactly what `base_pool::Transaction` stores). But on removal we only get the `Arc<Transaction>` (tbh - I do not know why - if we could get rid of `Arc`s that'd solve the problem) => that's why I need to clone contents of this `Arc`.

core/transaction-pool/graph/src/base_pool.rs
```diff
@@ -356,6 +366,14 @@ impl<Hash: hash::Hash + Member + Serialize, Ex: ::std::fmt::Debug> BasePool<Hash
 		removed
 	}
 
+	/// Removes ready transaction represented by the hash and all other ready transactions
+	/// that depend on it.
+	///
+	/// Returns a list of actually removed transactions.
+	pub fn remove_from_ready(&mut self, hash: &Hash) -> Vec<Arc<Transaction<Hash, Ex>>> {
```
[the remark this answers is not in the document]
svyatonik: Well, I just wanted to avoid calling `remove_invalid()` from `resubmit()` just because transactions that `resubmit()` deals with are not necessary invalid. So thought that it is better to hide the fact that it actually uses the same algorithm as `remove_invalid()` behind the name. But will remove as you've suggested.

core/transaction-pool/graph/src/base_pool.rs
```diff
@@ -181,6 +182,11 @@ impl<Hash: hash::Hash + Eq, Ex> Default for BasePool<Hash, Ex> {
 }
 
 impl<Hash: hash::Hash + Member + Serialize, Ex: ::std::fmt::Debug> BasePool<Hash, Ex> {
+	/// Start rejecting future transactions.
+	pub fn reject_future_transactions(&mut self) {
```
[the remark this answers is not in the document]
svyatonik: I believe with separate `LightTransactionPool` and `FullTransactionPool` it'll be easier - will do.

core/transaction-pool/graph/src/pool.rs
```diff
@@ -165,6 +172,36 @@ impl<B: ChainApi> Pool<B> {
 		)
 	}
 
+	/// Revalidate all ready transactions.
+	///
+	/// Returns future that performs validation of all ready transactions and
+	/// then resubmits all transactions back to the pool.
+	pub fn revalidate_ready(&self, at: &BlockId<B::Block>) -> impl Future<Output=Result<(), B::Error>> {
+		let validated_pool = self.validated_pool.clone();
+		let ready = self.validated_pool.ready().map(|tx| tx.data.clone());
```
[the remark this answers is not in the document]
svyatonik: Nope - this only clones encoded runtime extrinsic (without validation results - requires/provides/...). I do not think that `drain()` is a good idea, because revalidation could take some time and (if it, hopefully, ends successfully) the light node won't propagate transactions when revalidation is in progress. Also, drain will return `Arc<Transaction>` (as I have told above) => we'll need this clone anyway :/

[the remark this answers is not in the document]
svyatonik: Ahh, and revalidation could fail because e.g. timeout, missing peers, ... And we would need to reinsert tx with previous validation results if there would be `drain()`. Now we just emit warning in this case.

core/transaction-pool/graph/src/pool.rs
```diff
@@ -266,7 +304,7 @@ impl<B: ChainApi> Pool<B> {
 					&at,
 					known_imported_hashes,
 					pruned_hashes,
-					reverified_transactions,
+					reverified_transactions.into_iter().map(|(_, xt)| xt).collect(),
```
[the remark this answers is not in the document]
svyatonik: This'll result in `Vec<&Transaction>` and we need `Vec<Transaction>`

core/transaction-pool/graph/src/pool.rs
```diff
@@ -131,7 +136,9 @@ impl<B: ChainApi> Pool<B> {
 		let validated_pool = self.validated_pool.clone();
 		self.verify(at, xts, force)
 			.map(move |validated_transactions| validated_transactions
-				.map(|validated_transactions| validated_pool.submit(validated_transactions)))
+				.map(|validated_transactions| validated_pool.submit(validated_transactions
+					.into_iter()
+					.map(|(_, tx)| tx))))
```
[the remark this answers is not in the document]
svyatonik: This will be `Iterator<Item=&Transaction>`, but we need `Iterator<Item=Transaction>` here

core/transaction-pool/graph/src/validated_pool.rs
svyatonik: IMO this would require more `clone()` calls because we'll be unable to check if updated validation result exists for the transaction.

svyatonik: (and as a result more calls to `HashMap::insert()` because we need to overwrite validation result with updated)

core/transaction-pool/graph/src/validated_pool.rs
svyatonik: It maybe my misuderstanding as well. I don't think that we can have different transaction under `remove_hash`. And I don't clearly understand what do you mean by "we never really store it's hash anywhere and don't notify ...".. Let me describe this while with example and check that we both on the same page. Let's say at some moment we had a pool with:

HashOf(Tx1) => { data=Tx1, requires=[], provides=[1] }
HashOf(Tx2) => { data=Tx2, requires=[1], provides=[2] }
HashOf(Tx3) => { data=Tx3, requires=[], provides=[3] }

At this moment light client decides to revalidate its transactions. It sends revalidation requests for [Tx1, Tx2, Tx3] to remote nodes. Let's say while we wait for the response, another transactions are inserted to the pool:

HashOf(Tx4) => { data=Tx4, requires=[3], provides=[] }
HashOf(Tx5) => { data=Tx5, requires=[], provides=[] }

Let's say we have received responses (aka updated validation results):

HashOf(Tx1) => { data=Tx1, requires=[3], provides=[] }
HashOf(Tx2) => is now considered invalid
HashOf(Tx3) => { data=Tx3, requires=[], provides=[3] }

Then the `resubmit()` is called with these results. The `while !transactions.is_empty() {}` we're talking about must remove transactions Tx1, Tx2, Tx3 along with all transactions that depend on these tx. In our case they're: Tx4. While removing, this `while` builds a map of `TxHash => Either<UpdatedValidationResult, PreviousValidationResult>`. In our case Tx1, Tx2, Tx3 will have `UpdatedValidationResult` and Tx4 will have `PreviousValidationResult`. So the resulting map is:

HashOf(Tx1) => { data=Tx1, requires=[3], provides=[] }
HashOf(Tx2) => is now considered invalid
HashOf(Tx3) => { data=Tx3, requires=[], provides=[3] }
HashOf(Tx4) => { data=Tx4, requires=[3], provides=[] }

(mind that Tx5 is not affected by resubmit at all and stays in pool). So we're always referring to the same transactions with their hashes. Could you, please, check my logic here - maybe I'm missing something you're talking about?

svyatonik: (While responding to this, I've found that the order of insertion in `resubmit()` makes sense if we reject 'future' transactions. I.e. if we'll try to insert Tx1 before Tx3, then Tx1 will be rejected. So we either need to: (1) insert in the correct order (2) clear 'future' queue at the end of `resubmit()` call (with proper notifications, of course) or (3) support future queue on light client. I'd still prefer to avoid future transactions here => will try to go with (2) here)

srml/system/rpc/src/lib.rs
svyatonik: I've made a separate issue for that - #3875

core/transaction-pool/src/api.rs
svyatonik: There are tests in [`pool_maintainer.rs`](https://github.com/paritytech/substrate/pull/3833/files#diff-9db06f15e12d6c0feeaabcd41489533bR356) - `should_remove_transactions_from_the_light_pool`, `should_schedule_transactions_revalidation_at_light_pool` and `should_revalidate_transactions_at_light_pool`. Are they enough or do you want to test something else?

svyatonik:
So I've made a new `TransactionPool` trait. However, I haven't introduced two implementations - `FullTransactionPool` and `LightTransactionPool`. This is because I want to provide ability to customize `PoolApi` and `Maintainer` for (at least) light nodes (see notes about this in description). And when I made `LightTransactionPool` and `FullTransactionPool` generic over `PoolApi` and `Maintaner`, the implementation was exactly the same.

Main changes are - tx pool clients are now generic over `TransactionPool` instead of `ChainApi`. Also maintain procedure is now a part of the pool implementation itself.

I had to wrap all the futures that `TransactionPool` returns in the `Box` because otherwise too many impl details are to be exposed (also to avoid multi-multi-lines `Future`s declarations) and I'm unable to return `impl Future` from the trait, obviously. Also there's a bit more methods than simply `submit()`, `ready()` and `maintain()` - it appears that we need to subscribe to pool transactions (hence `import_notification_stream`), we need to report poll status to telemetry (hence `status`), ...

After doing this, I'm not sure that traitifying `TransactionPool` was a perfect idea - we still have a single implementation (`BasicTransactionPool` instead of `txpool::Pool`) that is generic over `PoolApi` and `Maintainer`. What do you think on that?

core/service/src/builder.rs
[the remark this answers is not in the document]
svyatonik: *(see top level comment)*

core/transaction-pool/graph/src/base_pool.rs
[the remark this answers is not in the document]
tomusdrw: We return `Arc<Transaction>` exactly to avoid cloning of the transaction content, and it has to exists in multiple places (in the pool and in the pending iterator for instance).

The code that clones `tx.data` could actually be optimized to first try to `Arc::try_unwrap` and only `clone` if there is another reference active (I suppose that 99% of cases would actually succeed in `try_unwrap`).

I'd really prefer to avoid making the transactions cloneable, they can be significant in size and it's going to be quite wasteful, not to mention a possible DoS vector. Do you think we could make a similar trick with `Arc::try_unwrap` in your new code and maybe avoid implementing `Clone` but rather adding some method that has more siginifcant name like `Transaction::duplicate()` to be able to document the rationale of not having `Clone`?

core/transaction-pool/graph/src/base_pool.rs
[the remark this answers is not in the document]
tomusdrw: Thanks! I'm just afraid that someone might call `remove_from_ready` thinking that it won't mark the transactions as invalid, and it's not really true.

core/transaction-pool/graph/src/validated_pool.rs
tomusdrw: Thanks for explanations! Yes, that totally makes sense, for some reason I was confused with what exactly we store in the `HashMap<Hash, ValidatedTransaction>`. Maybe renaming `transactions` to `updated_transactions` and adding a short comment that we store re-validated transactions there would make it clearer.

Re futures: indeed `(2)` seems like a better way to go, so you would temporarily need to allow `futures` queue for this specific calls and then just clear it after.

core/transaction-pool/src/api.rs
tomusdrw: That should do!

core/transaction-pool/Cargo.toml
```diff
@@ -18,3 +18,4 @@ txpool = { package = "substrate-transaction-graph", path = "./graph" }
 [dev-dependencies]
 keyring = { package = "substrate-keyring", path = "../../core/keyring" }
 test-client = { package = "substrate-test-runtime-client", path = "../../core/test-runtime/client" }
+consensus_common = { package = "substrate-consensus-common", path = "../../core/consensus/common" }
```
tomusdrw: Why is this needed? Can't find references.

core/transaction-pool/graph/src/base_pool.rs
```diff
@@ -169,22 +169,29 @@ pub struct BasePool<Hash: hash::Hash + Eq, Ex> {
 	recently_pruned_index: usize,
 }
 
-impl<Hash: hash::Hash + Eq, Ex> Default for BasePool<Hash, Ex> {
+impl<Hash: hash::Hash + Member + Serialize, Ex: ::std::fmt::Debug> Default for BasePool<Hash, Ex> {
```
tomusdrw: You don't need `::` before `std`

core/transaction-pool/graph/src/base_pool.rs
```diff
@@ -169,22 +169,29 @@ pub struct BasePool<Hash: hash::Hash + Eq, Ex> {
...
 		}
 	}
-}
 
-impl<Hash: hash::Hash + Member + Serialize, Ex: ::std::fmt::Debug> BasePool<Hash, Ex> {
-	/// Start rejecting future transactions.
-	pub fn reject_future_transactions(&mut self) {
-		self.reject_future_transactions = true;
+	/// Changes reject future transactions flag. Returns previous flag value.
+	pub fn reject_future_transactions(&mut self, reject: bool) -> bool {
```
tomusdrw: Does it really have to be `pub`? Can we make it `pub(crate)`?

core/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -219,6 +219,12 @@ impl<B: ChainApi> ValidatedPool<B> {
 				}
 			}
 
+			// if we're rejecting future transactions, then insertion order matters here:
+			// if tx1 depends on tx2, then if tx1 is inserted before tx2, then it goes
+			// to the future queue and gets rejected immediately
+			// => let's temporary stop rejection and clear future queue before return
+			let reject_future_transactions = pool.reject_future_transactions(false);
```
tomusdrw: To be fair I'd prefer either `pool.with_futures_enabled(FnOnce)` or a guard that resets to previous value on `Drop`.
Just setting a flag and flipping it back in the end is error prone, especially if you accidentally `return` or use `?` at some point .

core/transaction-pool/src/lib.rs
```diff
@@ -20,10 +20,246 @@
...
+
+/// Extrinsic hash type for a pool.
+pub type ExHash<P> = <P as TransactionPool>::Hash;
+/// Block hash type for a pool.
+pub type BlockHash<P> = <<P as TransactionPool>::Block as BlockT>::Hash;
+/// Extrinsic type for a pool.
+pub type ExtrinsicFor<P> = <<P as TransactionPool>::Block as BlockT>::Extrinsic;
+
+/// Transaction pool interface.
+pub trait TransactionPool: Send + Sync {
```
tomusdrw: The trait should live outside of this crate, we don't want other implementations to depend on the entire pool.

Maybe put it to `primitives` or some `transaction-pool/api` or wherever other service-generic fields reside.

core/transaction-pool/src/lib.rs
```diff
@@ -20,10 +20,246 @@
...
+	/// Error type.
+	type Error: From<txpool::error::Error> + txpool::error::IntoPoolError;
+
+	/// Returns a future that imports a bunch of unverified extrinsics to the pool.
+	fn submit_at(
+		&self,
+		at: &BlockId<Self::Block>,
+		xts: impl IntoIterator<Item=ExtrinsicFor<Self>> + 'static,
+		force: bool,
+	) -> Box<dyn Future<Output=Result<Vec<Result<Self::Hash, Self::Error>>, Self::Error>> + Send + Unpin>;
```
tomusdrw: We could use associated types for this, but I suppose it might be an overkill. I think a box here is fine.

core/transaction-pool/src/lib.rs
```diff
@@ -20,10 +20,246 @@
...
+	fn remove_invalid(
+		&self,
+		hashes: &[Self::Hash],
+	) -> Vec<Arc<txpool::base_pool::Transaction<Self::Hash, ExtrinsicFor<Self>>>>;
+
+	/// Returns pool status.
+	fn status(&self) -> txpool::base_pool::Status;
+
+	/// Get an iterator for ready transactions ordered by priority
+	fn ready(&self) -> BestIterator<Self::Hash, ExtrinsicFor<Self>>;
```
tomusdrw: Hmm, would be best to avoid leaking `BestIterator` from the crate. Maybe let's just `Box` it here as well.

core/transaction-pool/src/maintainer.rs
```diff
@@ -479,37 +525,41 @@ mod tests {
...
+			*revalidation_status.lock() = TxPoolRevalidationStatus::Scheduled(Some(Instant::now()), None)
+		);
 		assert_eq!(status.ready, 0);
 
 		// when revalidation is scheduled by block number - it is performed
 		let fetcher = build_fetcher();
-		let maintainer = DefaultLightTransactionPoolMaintainer::new(fetcher.clone(), None, None);
-		*maintainer.revalidation_status.lock() = TxPoolRevalidationStatus::Scheduled(None, Some(0));
-		let status = with_fetcher_maintain(fetcher, &maintainer);
+		//let maintainer = DefaultLightTransactionPoolMaintainer::new(fetcher.clone(), None, None);
```
tomusdrw: Uncomment or remove

core/transaction-pool/src/lib.rs
```diff
@@ -20,10 +20,246 @@
...
+> where
+	Block: BlockT<Hash=H256>,
+	Backend: 'static + client::backend::Backend<Block, Blake2Hasher>,
+	Client<Backend, Executor, Block, Api>: ProvideRuntimeApi,
+	<Client<Backend, Executor, Block, Api> as ProvideRuntimeApi>::Api: TaggedTransactionQueue<Block>,
+	Executor: 'static + Send + Sync + client::CallExecutor<Block, Blake2Hasher>,
+	Api: 'static + Send + Sync,
+{
+	/// Create new basic full transaction pool with default API and maintainer.
+	pub fn default_full(options: Options, client: Arc<Client<Backend, Executor, Block, Api>>) -> Self {
```
tomusdrw: why not `new_full`? or just `full`?

core/transaction-pool/src/lib.rs
```diff
@@ -20,10 +20,246 @@
...
+	fn maintain(
+		&self,
+		id: &BlockId<Self::Block>,
+		retracted: &[BlockHash<Self>],
+	) -> Box<dyn Future<Output=()> + Send + Unpin>;
+}
+
+/// Basic implementation of transaction pool that can be customized by providing
+/// different PoolApi and Maintainer.
+pub struct BasicTransactionPool<PoolApi, Maintainer, Block>
```
tomusdrw: Cant' we merge `PollApi` and `Maintainer` somehow? You can't really have `FullChainApi + LightPoolMaintainer`, can you? Alternatively maybe just merge `BasicTransactionPool` with the maintainer implementation? The only drawback would be that you have to duplicate method implementations I guess?

tomusdrw: Or perhaps let's move `fn maintain` trait to a separate thread, and then have:
```rust
// generic implementation that does not care about Light/Full stuff
impl TransactionPool for BasicTransactionPool<...>  {

}
// Light/Full-specific maintainer implementation
impl TransactionPoolMaintain for BasicTransactionPool<
  LightChainApi<..>
> {
 fn maintain(...) {}
}
```
This will allow us to reduce the amount of code a bit and rid of `Default*PoolMaintainer` implementations.

I think you could also simplify pool creation with by switchin to no-maintainer and using bare functions:
```rust
fn new_full<Client, Block>(client: Arc<Client>) -> BasicTransactionPool<
  FullChainApi<Client, Block>
> where 
  Block: traits::Block,
  Client: ProvideRuntimeApi + HeaderBackend,
{

}
```
You don't really need all of those bounds to create stuff, only when you add behavior (i.e. imlement `TransactionPool` trait).

core/transaction-pool/src/lib.rs
svyatonik: Have made a separate crate - `transaction-pool/api` (will add a top-level comment about that)

core/transaction-pool/src/lib.rs
svyatonik: (will leave a top-level comment about that)

svyatonik:
So here's now `core/src/transaction-pool/api` - the create that defines the `TransactionPool` trait, `TransactionPoolMaintainer` and `InPoolTransaction` trait. Besides that, it declares `MaintainableTransactionPool` struct that is able to merge two structs that are implementing `TransactionPool` and `TransactionPoolMaintainer` into a single struct that implements both.

Normally, everyone would use 'basic' implementations from `core/src/transaction-pool` - `BasicPool<FullChainApi>` and `BasicPool<LightChainApi>`, though one could customize Api implementation. There are also two default maintainer implementations in this crate - `FullBasicPoolMaintainer` and `LightBasicPoolMaintainer`.

No crates except those who create transaction pools are now depend on `substrate-transaction-pool` crate. Though, it is still referenced by significant amount of crates, because basic implementation is used in tests

svyatonik:
Corresponding fix of polkadot is [here](https://github.com/paritytech/polkadot/compare/fix_polkadot_for_light_tx_pool3?expand=1). I'm not opening PR because don't know if (and when) this PR will be approved. Will open polkadot PR once if it is approved.

svyatonik:
@tomusdrw Thanks for merging with master! I've resolved another bunch of conflicts - mostly caused by #4145 . The main change is that now `offchain::TransactionPool` is implemented for `txpool_api::MaintainableTransactionPool` (before it has been implemented for `transaction_graph::Pool`). It would be better to have blanket impl, but `TransactionPool` traits are in different crates && I can't move common `trait TransactionPool` to primitives, because it depends on `sr_primitives`.

(note that corresponding PR in polkadot isn't opened until this PR is approved)

PR #3834, opened 2019-10-16
Create opaque struct for StorageProof.

jimpo:
Passing around `Vec<Vec<u8>>` everywhere is gross and confusing and breaks encapsulation.


core/client/src/client.rs
```diff
@@ -705,18 +705,18 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
 
 		cht::for_each_cht_group::<Block::Header, _, _, _>(cht_size, blocks, |_, cht_num, cht_blocks| {
 			let cht_proof = self.changes_trie_roots_proof_at_cht(cht_size, cht_num, cht_blocks)?;
-			proof.extend(cht_proof);
+			proofs.push(cht_proof);
 			Ok(())
 		}, ())?;
 
-		Ok(proof.into_iter().collect())
+		Ok(merge_storage_proofs(proofs))
```
cheme: I prefer the old version where we did not cat vecs: here we are eating memory. Maybe an iterator version of merge_storage_proofs will get the best of both world.
Actually merge_storage_proofs takes an iterator as input, so there is maybe something todo, does not seems super easy though since it neends to transform for_each_cht_group to an iterator.

core/client/src/light/call_executor.rs
```diff
@@ -218,11 +216,7 @@ pub fn prove_execution<Block, S, E>(
 
 	// execute method + record execution proof
 	let (result, exec_proof) = executor.prove_at_trie_state(&trie_state, &mut changes, method, call_data)?;
-	let total_proof = init_proof.into_iter()
-		.chain(exec_proof.into_iter())
-		.collect::<HashSet<_>>()
-		.into_iter()
-		.collect();
+	let total_proof = merge_storage_proofs(vec![init_proof, exec_proof]);
```
cheme: vec allocation can be avoided here, especially since merge_storage_proofs takes an iterator as input.

core/finality-grandpa/Cargo.toml
```diff
@@ -27,6 +27,7 @@ network = { package = "substrate-network", path = "../network" }
 srml-finality-tracker = { path = "../../srml/finality-tracker" }
 fg_primitives = { package = "substrate-finality-grandpa-primitives", path = "primitives" }
 grandpa = { package = "finality-grandpa", version = "0.9.0", features = ["derive-codec"] }
+state-machine = { package = "substrate-state-machine", path = "../state-machine" }
```
cheme: We should put 'StorageProof' in a different crate (trie or primitive) to avoid those new dependencies.

core/network/Cargo.toml
```diff
@@ -46,6 +46,7 @@ erased-serde = "0.3.9"
 void = "1.0.2"
 zeroize = "0.10.1"
 babe-primitives = { package = "substrate-consensus-babe-primitives", path = "../consensus/babe/primitives" }
+state-machine = { package = "substrate-state-machine", path = "../../core/state-machine" }
```
cheme: new deps could probably be avoided.

jimpo: We can re-export from client, but I don't see the reason. `substrate-state-machine` is an implicit dependency anyway, why not make it explicit?

core/finality-grandpa/Cargo.toml
jimpo: It's nice to have it as part of the `proving_backend` module so that the internals can stay private/opaque.

core/state-machine/src/proving_backend.rs
```diff
@@ -29,6 +30,47 @@ use crate::trie_backend::TrieBackend;
 use crate::trie_backend_essence::{Ephemeral, TrieBackendEssence, TrieBackendStorage};
 use crate::{Error, ExecutionError, Backend};
 
+/// A proof that some set of key-value pairs are included in the storage trie.
```
kianenigma: Looks pretty reasonable. Where is this verification process (walking the nodes and building the sub-tree) that you mentioned in the docs happening?

core/network/Cargo.toml
cheme: semantically 'network' depending on 'state-machine' feels a bit odd. I would rather reexport from client.

core/state-machine/src/proving_backend.rs
jimpo: It's a bit spread around, but mostly `create_proof_check_backend`.

core/client/src/light/call_executor.rs
jimpo: How exactly do you recommend creating this iterator? I think something like `iter::once(init_proof).chain(iter::once(exec_proof))` is a bit clunky.

core/finality-grandpa/Cargo.toml
jimpo: It is not re-exported from `client::light::fetcher`.

core/client/src/light/call_executor.rs
jimpo: This really doesn't seem like a good reason to sacrifice readability to me. `merge_storage_proofs` internally has to create a new `HashSet` and much larger vector of trie nodes anyway.

cheme: Of course it is just an alloc among lot of others, I think crazy iter expression are rather common in the code base, but not abusing them is probably a good thing.

Regarding the other comment where we grow memory, you also can use a different approach, I know some poeple like this kind of api, I never really use it but it may be good here.

```rust
struct ProofMerger(Hashset<StorageProof>);

impl ProofMerger {
  fn add(self, proof: StorageProof) {
    self.0.extends(proof);
    self
  }
  fn merge(self) -> StorageProof {
    StorageProof { trie_nodes: self.0.into_iter().collect() }
  }

}
```
in this case would look like 

```rust
let total_proof = ProofMerger(init_proof)
  .add(exec_proof)
  .merge();
```

cheme: Not that it is really only nitpick, should not prevent merging this PR for me.

core/state-machine/src/proving_backend.rs
```diff
@@ -29,6 +30,47 @@ use crate::trie_backend::TrieBackend;
...
+///
+/// The proof consists of the set of serialized nodes in the storage trie accessed when looking up
+/// the keys covered by the proof. Verifying the proof requires constructing the partial trie from
+/// the serialized nodes and performing the key lookups.
+#[derive(Debug, PartialEq, Eq, Clone, Encode, Decode)]
+pub struct StorageProof {
```
bkchr: I think this should be moved to `substrate-primitives`. State-machine just compiles on native, but the proof stuff is required on `no-std` as well e.g. in Cumulus.

jimpo: OK, I added a `TrieProofWithValues` struct in `substrate-trie` which has a `no_std` option. `StorageProof` in `substrate-state-machine` now wraps that.

The reason these need to be separate is because a `TrieBackend` is an abstration over a single trie which is defined in `substrate-state-machine`. In particular, it allows storage accesses to child tries which traverse multiple tries, a concept foreign to `substrate-trie`.

Currently, these structs are both very simple and are basically just a vec of encoded nodes. In the future, though, it would be nice to create trie proofs that omit redundant internal hashes and storage proofs that internally contain multiple separate trie proofs, one for the root trie and one for each child trie.

core/network/src/protocol.rs
```diff
@@ -1226,7 +1226,7 @@ impl<B: BlockT, S: NetworkSpecialization<B>, H: ExHashT> Protocol<B, S, H> {
 					error
 				);
 				self.peerset_handle.report_peer(who.clone(), RPC_FAILED_REPUTATION_CHANGE);
-				Default::default()
+				StorageProof::empty()
```
bkchr: Why not implement `Default` for `StorageProof`?

core/state-machine/src/proving_backend.rs
bkchr: Could we please get a method `into_inner()` or similar?

core/network/src/protocol.rs
jimpo: I don't think the concept of a "default proof" makes any sense.

core/state-machine/src/proving_backend.rs
jimpo: OK, yes, I removed the whole TrieProof thing and there's a `StorageProof::iter_nodes(self)` method which iterates over all trie nodes.

Issue #3639, opened 2019-09-18
Sync cannot download explicitly requested forks as light client if necessary


rphmeier:
#3633 introduced the beginning of an API for downloading important forks (according to the higher-level consensus algorithm). The problem is that some of these forks may be orphaned or too old for a full node to import.

#3629 is for GRANDPA to use this API 

This API should be extended with a parameter where the consensus code can inform sync that it only needs the header-chain. This is the case for GRANDPA. The behavior of the sync code if only the headers are needed should be to:
  - attempt to download and import the chain as a full client, if the state of the common ancestor hasn't been pruned.
  - otherwise, download only the header chain of the fork. it could go through the client, but doesn't have to, as long as any callers of the `sync_fork` API receive the header chain.

Importing the header chain via the client (as a "sort-of-light" mode) would be a bit nicer w.r.t. continuity; otherwise if the caller of `sync_fork` wants to return those headers later on they'll have to maintain their own storage rather than using the shared client utilities.

That needs some investigation as to whether the client can be configured to sometimes act as a light client in import and sometimes not -- @svyatonik any thoughts on that point?

svyatonik:
If that would be possible, I think that having 'own storage' is a better option. IMO there are many hidden assumptions in the full client code. Like if block just have been imported, its state is available (e.g. to run offchain, ...), its body is available (e.g. to maintain memory pool, ...), ... And if it is possible to localize the code (headers-only import) in a single place, it would be better, imo.

rphmeier:
Still, it seems cleaner to have it all go through the client. I suppose we could add another `BlockOrigin` where import notifications aren't fired for such blocks. That should make sure that none of those hidden assumptions (which shouldn't exist in the first place...) get triggered. Either way, if those assumptions are panicking, that is a bug.

arkpar:
This is implemented in #3942

PR #3835, opened 2019-10-16
cli: disable pruning on validators

andresilva:
Currently GRANDPA needs to import all blocks that are voted on, and therefore requires to have state available for those blocks. During periods of long finality stalls we run the risk of state being pruned due to depth (we prune after 4096 blocks regardless of finality), or e.g. a short-lived fork being pruned out. This PR changes the default pruning mode to be set to archive if the node is a validator, otherwise it defaults to keeping state for the last 256 blocks. The node will refuse to start if the node is a validator and pruning is enabled, unless the flag `--unsafe-pruning` is given as well.

Currently the node doesn't seem to detect whether the existing database is an archive or pruned. This means that if you start with a pruned database, and restart the node as archive, the node will either crash due to missing state (if you are a validator trying to author), or it just won't import any blocks. The reverse (archive -> pruned) seems to work. This might be a bit confusing to end-users, especially if they just enable/disable `--validator` which implicitly switches the pruning mode.

cc @arkpar

### DIFF
--- a/core/cli/src/lib.rs
+++ b/core/cli/src/lib.rs
@@ -674,13 +674,6 @@ where
 	config.database_path = db_path(&base_path, config.chain_spec.id());
 	config.database_cache_size = cli.database_cache_size;
 	config.state_cache_size = cli.state_cache_size;
-	config.pruning = match cli.pruning {
-		Some(ref s) if s == "archive" => PruningMode::ArchiveAll,
-		None => PruningMode::default(),
-		Some(s) => PruningMode::keep_blocks(s.parse()
-			.map_err(|_| error::Error::Input("Invalid pruning mode specified".to_string()))?
-		),
-	};
 
 	let is_dev = cli.shared_params.dev;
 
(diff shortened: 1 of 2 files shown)


kianenigma:
If it is really a short term patch that will be replaced anyhow (post #3639), and can be done manually via flag, a warning in the case of `is_validator || !archive_mode` might also be enough.

andresilva:
@kianenigma I would like to make it particularly hard for validators to run with what we consider unsafe settings. A warning is easily ignored IMO, hence why I added the `--unsafe-pruning` flag, so that this becomes obvious to anyone that has to type it out.

gavofyork:
> Currently the node doesn't seem to detect whether the existing database is an archive or pruned. This means that if you start with a pruned database, and restart the node as archive, the node will either crash due to missing state (if you are a validator trying to author), or it just won't import any blocks.

we should probably address that in this patch then otherwise we'll have a torrent of validators wondering why their node is crashing. i'd suggest just discarding the (pruned) DB if they are wanting to use archive and then it'll just resync from scratch.

andresilva:
@arkpar Any suggestion on how to verify whether a given state db is an archive or pruned on startup? (in particular something that is migration friendly). I guess we could check for the existence of a pruning journal on the database?

arkpar:
@andresilva `Client::block_status` can be used to check if the state is available

PR #3841, opened 2019-10-17
Use Bytes for contract rpc input_data

kwingram25:
Switch to using Bytes for CallRequest.input_data in place of Vec<u8> for easier UI interaction


srml/contracts/rpc/src/lib.rs
```diff
@@ -41,7 +42,7 @@ pub struct CallRequest<AccountId, Balance> {
 	dest: AccountId,
 	value: Balance,
 	gas_limit: number::NumberOrHex<u64>,
-	input_data: Vec<u8>,
+	input_data: Bytes
```
bkchr:
```suggestion
	input_data: Bytes,
```

PR #3843, opened 2019-10-17
Remove "Call" from construct_runtime!

enfipy:
Remove `Call` from `construct_runtime!` for `RandomnessCollectiveFlip` due to the following error:
```
error[E0277]: `srml_randomness_collective_flip::Call<Runtime>` doesn't implement `std::fmt::Debug`
```


shawntabrizi:
Can you also update `node`?

https://github.com/paritytech/substrate/blob/master/node/runtime/src/lib.rs#L530

bkchr:
As discussed on riot, all tests are happy on master, so the issue must be something different.

PR #3844, opened 2019-10-17
Make build work with rustc 1.37 stable and RUSTC_BOOTSTRAP=1

infinity0:
By relaxing some checks we can easily make the build work using Debian's system rustc (which I maintain), avoiding having to install rustc nightly:

~~~~
$ apt install rustc libstd-rust-dev-wasm32-cross libssl-dev
$ RUSTC_BOOTSTRAP=1 cargo build
[..]
   Compiling transaction-factory v0.0.1 ($PWD/test-utils/transaction-factory)
    Finished dev [unoptimized + debuginfo] target(s) in 11m 07s
~~~~

### DIFF
--- a/core/sr-std/without_std.rs
+++ b/core/sr-std/without_std.rs
@@ -14,7 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
-#[cfg(feature = "nightly")]
 #[doc(hidden)]
 pub extern crate alloc;
 
--- a/core/utils/wasm-builder/src/prerequisites.rs
+++ b/core/utils/wasm-builder/src/prerequisites.rs
@@ -16,6 +16,7 @@
 
 use std::{process::{Command, Stdio}, fs};
 
+use std::env;
 use tempfile::tempdir;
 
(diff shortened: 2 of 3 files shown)


infinity0:
[clabot:check]

bkchr:
I don't understand your changes, what means stable forced to nightly? And why is the env variable called `RUSTC_BOOTSTRAP`.

infinity0:
`RUSTC_BOOTSTRAP` is a not-really-secret environment variable for the rustc compiler to switch on nightly features. It is used to bootstrap the next version of rustc, hence the name, but can also be used to "force" a stable rustc compiler to switch on nightly features, hence my naming of that function.

bkchr:
:(

gavofyork:
?

bkchr:
I'm neither happy with the style nor the way it was implemented, but I will come up with a follow up pr at some point.

infinity0:
@bkchr If you could state your issues I'd be very happy to address them in a follow-up PR myself! :)

bkchr:
Ty, but it is already fixed and merged into master. In general, you used whitespaces and I was not happy about they way you implemented it. Nothing serious.

PR #3846, opened 2019-10-17
safe multi-era slashing for NPoS

rphmeier:
Based on https://research.web3.foundation/en/latest/polkadot/slashing/npos/

This PR implements safe multi-era slashing for NPoS. The fundamental problem that this tries to solve is that any particular validator or nominator's stake is largely shared across many eras. However, those participants are liable for up to 100% of their stake in each era. When performing multi-era slashing, we have a careful balance of ensuring that 100% slashes from multiple eras don't _overslash_ (i.e. punish more than the validator or nominator was ever on the hook for in that period of time) without placing too low of a cap on the amount of stake that can be slashed.

 For the purposes of the economic model, it is easiest to think of each validator
 of a nominator which nominates only its own identity.

 The act of nomination signals intent to unify economic identity with the validator - to take part in the
 rewards of a job well done, and to take part in the punishment of a job done badly.

 There are 3 main difficulties to account for with slashing in NPoS:
   - A nominator can nominate multiple validators and be slashed via any of them.
   - Until slashed, stake is reused from era to era. Nominating with N coins for E eras in a row
     does not mean you have N*E coins to be slashed - you've only ever had N.
   - Slashable offences can be found after the fact and out of order.

 The algorithm implemented in this module tries to balance these 3 difficulties.

 First, we only slash participants for the _maximum_ slash they receive in some time period,
 rather than the sum. This ensures a protection from overslashing.

Second, we do not want the time period (or "span") that the maximum is computed
 over to last indefinitely. That would allow participants to begin acting with
 impunity after some point, fearing no further repercussions. For that reason, we
 automatically "chill" validators and withdraw a nominator's nomination after a slashing event,
 requiring them to re-enlist voluntarily (acknowledging the slash) and begin a new
 slashing span.

Typically, you will have a single slashing event per slashing span. Only in the case
 where a validator releases many misbehaviors at once, or goes "back in time" to misbehave in
 eras that have already passed, would you encounter situations where a slashing span
 has multiple misbehaviors. However, accounting for such cases is necessary 
 to deter a class of "rage-quit" attacks.

major TODOs:
  - [x] Better description :)
  - [x] Rewards to reporters
  - [x] Pruning of slashing-related metadata
  - [x] integration of the new slashing logic
  - [x] Testing


srml/staking/src/lib.rs
```diff
@@ -615,9 +607,22 @@ decl_storage! {
...
+		/// All slashing events on nominators, mapped by era to the highest slash value of the era.
+		NominatorSlashInEra:
+			double_map EraIndex, twox_128(T::AccountId) => Option<BalanceOf<T>>;
+
+		/// Slashing spans for stash accounts.
+		SlashingSpans: map T::AccountId => Option<slashing::SlashingSpans>;
+
+		/// Records information about the maximum slash of a stash within a slashing span,
+		/// as well as how much reward has been paid out.
+		SpanSlash: map (T::AccountId, slashing::SpanIndex)
```
rphmeier: this is something that's fairly likely to change in the near future. what's the typical approach in SRML for making storage entries versionable?

srml/offences/src/lib.rs
```diff
@@ -113,44 +108,18 @@ where
...
-				offenders_count.saturating_sub(previous_offenders_count),
-				validator_set_count,
-			);
-			let numerator = new_fraction.saturating_sub(previous_fraction);
-			let denominator = Perbill::one().saturating_sub(previous_fraction);
-			denominator.saturating_mul(numerator)
-		} else {
-			new_fraction.clone()
-		};
+		let slash_perbill: Vec<_> = (0..concurrent_offenders.len())
```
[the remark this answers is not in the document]
tomusdrw: Just saw that in the `slashing` module we implement a rolling window approach and only slash for maximum (not sum) value anyway.

srml/staking/src/slashing.rs
```diff
@@ -0,0 +1,772 @@
...
+
+		let last_length = next_start - self.last_start;
+		self.prior.insert(0, last_length);
+		self.last_start = next_start;
+		self.span_index += 1;
+		true
+	}
+
+	// an iterator over all slashing spans in _reverse_ order - most recent first.
+	pub(crate) fn iter(&'_ self) -> impl Iterator<Item = SlashingSpan> + '_ {
```
tomusdrw: Maybe call it `rev_iter` then?

srml/staking/src/slashing.rs
```diff
@@ -0,0 +1,772 @@
...
+		let pruned = match old_idx {
+			Some(o) => {
+				self.prior.truncate(o);
+				let new_earliest = self.span_index - self.prior.len() as SpanIndex;
+				Some((earliest_span_index, new_earliest))
+			}
+			None => None,
+		};
+
+		// readjust the ongoing span, if it started before the beginning of the winow.
```
tomusdrw:
```suggestion
		// readjust the ongoing span, if it started before the beginning of the window.
```

srml/staking/src/slashing.rs
```diff
@@ -0,0 +1,772 @@
...
+};
+use rstd::vec::Vec;
+use codec::{Encode, Decode};
+
+/// The proportion of the slashing reward to be paid out on the first slashing detection.
+/// This is f_1 in the paper.
+const REWARD_F1: Perbill = Perbill::from_percent(50);
+
+/// The index of a slashing span - unique to each stash.
+pub(crate) type SpanIndex = u32;
```
tomusdrw: Do we really need `pub(crate)` everywhere? The module is not really public anyway.

srml/staking/src/slashing.rs
```diff
@@ -0,0 +1,772 @@
...
+	let mut reward_payout = Zero::zero();
+
+	for nominator in &exposure.others {
+		let stash = &nominator.who;
+
+		// the era slash of a nominator always grows, if the validator
+		// had a new max slash for the era.
+		let era_slash = {
+			let own_slash_prior = prior_slash_p * nominator.value;
+			let own_slash_by_validator = slash * nominator.value;
```
tomusdrw: Why do we multiple `slash` value and the exposed balance? Shouldn't this be just a percentage?

Also shouldn't we subtract validators `own_slash` from the remaining `slash` (before calling this function) and only slash nominators the rest?

srml/staking/src/slashing.rs
```diff
@@ -0,0 +1,772 @@
...
+	if !value.is_zero() {
+		ledger.active -= value;
+
+		// Avoid there being a dust balance left in the staking system.
+		if ledger.active < T::Currency::minimum_balance() {
+			value += ledger.active;
+			ledger.active = Zero::zero();
+		}
+
+		ledger.total -= value;
```
tomusdrw: We are sure it won't overflow, right? `ledger.total >= ledger.active`?

srml/staking/src/slashing.rs
```diff
@@ -0,0 +1,772 @@
...
+					return
+				}
+				(payout, slashed_imbalance)
+			},
+		};
+
+		// take rewards out of the slashed imbalance.
+		let (mut reward_payout, mut value_slashed) = slashed_imbalance.split(reward_payout);
+
+		let per_reporter = reward_payout.peek() / (reporters.len() as u32).into();
```
tomusdrw: Should we add a comment that `reporters` are not empty, because we guard against this earlier? just to future-proof the code.

srml/staking/src/lib.rs
kianenigma: oh, if it is deployed then it needs manual, one-time  migration code `on_initialised`.

srml/staking/src/lib.rs
```diff
@@ -1034,67 +1037,10 @@ impl<T: Trait> Module<T> {
...
-			amount: slash,
-		});
-
-		// trigger the event
-		Self::deposit_event(
-			RawEvent::Slash(stash.clone(), slash)
-		);
-
-		imbalance
+	/// Chill a stash account.
```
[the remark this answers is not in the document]
kianenigma: I think chill is actually better: we actually have a helper named `kill_stash()`. This is used to force-unbond and clean the data. Chill clearly says: you are still bonded, but we've force-chill-ed you to make sure you are not being slashed over and over. 

I think it is okay

srml/offences/src/lib.rs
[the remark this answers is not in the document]
rphmeier: right - accounting for the prior slash is done now in the slashing module

srml/staking/src/lib.rs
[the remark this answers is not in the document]
rphmeier: `chill` uses `chill_stash` internally after my changes

srml/staking/src/slashing.rs
rphmeier: This data structure is canonically iterated in reverse order, so I think it's better to call it `iter`.

srml/staking/src/slashing.rs
rphmeier: what do you suggest instead? `pub(crate)` is better than `pub` for this case, and we need items to be used in the module above.

srml/staking/src/slashing.rs
rphmeier: > Also shouldn't we subtract validators own_slash from the remaining slash (before calling this function) and only slash nominators the rest?

There is a way of doing that, but it's really involved and the Web3 Research team mentioned that we do in fact want slashes to hit nominators at least a little so that they have a disincentive to nominate untrusted candidates.

rphmeier: `slash` is a proportion here, also

srml/staking/src/slashing.rs
rphmeier: yeah, that's an invariant of the data structure.

srml/staking/src/slashing.rs
tomusdrw: If the module is not public, there isn't really a difference, is there? I prefer `pub` as it's more concise.

srml/offences/src/lib.rs
```diff
@@ -113,44 +108,18 @@ where
...
-				} else {
-					old_fraction.clone()
-				}
-			})
-			.collect::<Vec<_>>();
-
-		T::OnOffenceHandler::on_offence(&concurrent_offenders, &slash_perbill);
+		T::OnOffenceHandler::on_offence(
+			&concurrent_offenders,
+			&slash_perbill,
```
kianenigma: nit: from the way that `slash_perbill ` is being created, I don't see why it has to be a collection rather than a single value. Seems to be always the same?

srml/staking/src/lib.rs
kianenigma: This https://github.com/paritytech/substrate/pull/3947/files should make it at least less of an annoyance.

srml/staking/src/lib.rs
```diff
@@ -1212,31 +1158,32 @@ impl<T: Trait> Module<T> {
...
 		CurrentEraStartSessionIndex::mutate(|v| {
 			*v = start_session_index;
 		});
 		let bonding_duration = T::BondingDuration::get();
 
-		if current_era > bonding_duration {
-			let first_kept = current_era - bonding_duration;
-			BondedEras::mutate(|bonded| {
-				bonded.push((current_era, start_session_index));
+		BondedEras::mutate(|bonded| {
```
kianenigma: nit: the entire funciton seems to be like 

```
/// some high level explanation
function_that_does_it();

/// more logic 
one_liner_that_does_it();
```

i would consider putting this into a `update_bonded_eras() `

burdges:
>These lots and lots of small offences are not things that substantially affect security, otherwise they would be large offences.

Yes exactly.  I'd expect slashing code would often flag repeated slashing offenses, so slashing only max puts the onus on the slashing code to detect a more dangerous correlation.

Importantly, max normally simplifies writing slashing code for individual offenses.  If we treated them additively, then really any modules that reports slashing should worry about repeated slashing and maybe their interactions with other modules that report related offenses.  In using max, we declare that slashing reports have no relationship by default. 

If slashes are very small then they could just impact rewards which we still handle additively.

> Once you misbehave as a validator, you have to voluntarily start a new slashing span (i.e. end the max range) in order to validate again. So if this validator re-registers and misbehaves again, it's a sum, not a max.

I'll clarify:  We always kick validators when they gets slashed (enough).  We do so for numerous reasons, but one reason is protecting the validator and their nominators.  We asked if nominators themselves need similar protections when slashing across historical eras, and concluded yes.  You can think of slashing spans as a way to track how much and when we kick the nominators themselves, as oppose to validators. 

I think the current code chills them, but extreme cases expose a trade off between security for the validator and security for the network.  Alfonso and I found a simple mechanism to make place this trade off under dynamic control by governance https://research.web3.foundation/en/latest/polkadot/slashing/npos/#suppressed-nominators-in-phragmen but implementing it can wait and we're actually still refining it slightly.

> These little misbehaviors that do nothing don't lead to validator being slashed a lot, but more importantly, their nominators don't get slashed a lot either - no "rage-quit" attack is possible.

We're still vulnerable to rage quit and validator extortion attacks technically, but much much less so.  In particular, if a validator wants to get their nominators slashed by proportion p, like say 100%, then they must retroactively commit offenses or correlated offenses that warrant a slash of p.  We've many fewer such offenses so it's much less likely they can do so in a way that governance cannot explore and appropriately revert.  If they say validate an invalid block (100%) then governance can look at uncles, etc.  We cannot code all investigation tricks into slashing logic.

srml/staking/src/slashing.rs
rphmeier: There is a difference in intent. It also stops `pub use` from higher up to fail, meaning the author of that line would have to stop and think about changing the intent of the code.

rphmeier:
I need to add some code for automatically de-registering a validator from nomination lists on kick that isn't O(n) in the number of nominators. That seems to involve a storage migration.

burdges:
Not O(n)?  Is your trick to identify validators by their slashing span not going to work?

rphmeier:
It will work, but it needs a storage upgrade that I have to finesse a little.

kianenigma:
did this PR end up touching phragmen code significantly? I want to start to 
1- Remove the whole self-vote thing. Staking will be responsible to add all validators as nominators as well. 
2- Make any other steps needed to move it to a separate crate. 
3- Based on the discussion we had, will probably switch to accepting iterators. Although this can be done later. I prefer first having a clean interface and then optimising based on that.

burdges:
We confused ourselves slightly here @rphmeier because I'd envisioned slashes still being additive for nominators when they get caused by different validators within the same era.  I think the data structures do not currently track enough for this, which sucks and makes everything less efficient.

rphmeier:
> did this PR end up touching phragmen code significantly?

No, I just added preprocessing before calling phragmén.

> We confused ourselves slightly here @rphmeier because I'd envisioned slashes still being additive for nominators when they get caused by different validators within the same era. I think the data structures do not currently track enough for this, which sucks and makes everything less efficient.

@burdges we're tracking an era_slash for the nominator, which is the total slashed in the era (the sum to the right of the max). Updating the slash via a particular validator from P to P' is just adding the difference `(P' - P)(x_{n,v,e})`. `fn slash_nominators` does that.
`P' > P` is an invariant of calling the function. But it follows the formula from the research post about slashes via multiple validators in the same era being additive. If that's not the desired behavior, then the document is wrong.

burdges:
Appears fine now @rphmeier  Yay :)

rphmeier:
One of the later additions to this PR is that slashes are immediately computed and not-so-immediately applied. Slashes are deferred in a FIFO queue (by era of detection) where governance or Sudo can intervene and prevent the slash from actually being applied for `SlashDeferDuration` eras.

gavofyork:
> where any root origin

there's only one :)

but it would be good to get this done with `EnsureOrigin` so that the original can be configured as part of the config trait. see treasury pallet for a simple example of how this can be done

rphmeier:
>  it would be good to get this done with EnsureOrigin so that the original can be configured as part of the config trait

Done - I set this in node-runtime to be supermajority of council.

bin/node/runtime/src/lib.rs
```diff
@@ -265,6 +265,8 @@ impl staking::Trait for Runtime {
 	type SessionsPerEra = SessionsPerEra;
 	type BondingDuration = BondingDuration;
 	type SlashDeferDuration = SlashDeferDuration;
+	/// A super-majority of the council can cancel the slash.
+	type SlashCancelOrigin = collective::EnsureProportionAtLeast<_3, _4, AccountId, CouncilCollective>;
```
gavofyork: 👍

palette/staking/src/lib.rs
```diff
@@ -1018,14 +1148,44 @@ decl_module! {
...
+
+			let mut slash_indices = slash_indices;
+			slash_indices.sort_unstable();
+			let mut unapplied = <Self as Store>::UnappliedSlashes::get(&era);
+
+			for (removed, index) in slash_indices.into_iter().enumerate() {
+				let index = index as usize;
+
+				// if `index` is not duplicate, `removed` must be <= index.
+				if removed > index { return Err("duplicate index") }
```
gavofyork:
```suggestion
				ensure!(removed <= index, "duplicate index");
```

palette/staking/src/lib.rs
```diff
@@ -1018,14 +1148,44 @@ decl_module! {
...
+				if index >= unapplied.len() { return Err("slash record index out of bounds") }
```
gavofyork:
```suggestion
				ensure!(index < unapplied.len(), "slash record index out of bounds");
```

gavofyork:
at least the slash-deferring logic seems reasonable.

PR #3852, opened 2019-10-18
Add derive path

chesterliliang:
1. Add derive path
2. Upgrade wookong-solo to 0.1.4
3. Path not send to device due to new firmware needed. 

Plan to add new command 'select' for switch multi seed.
```
fn get_seed_amount() ->u32
fn select(public_key, path) -> Result<<Self::Pair as TraitPair>::Public>;
fn choose(index) -> Result<<Self::Pair as TraitPair>::Public>;
```


core/primitives/src/crypto.rs
```diff
@@ -704,7 +704,7 @@ pub trait Pair: CryptoType + Sized + Clone + Send + Sync + 'static {
 				return Ok(r)
 			}
 		}
-
+
```
kianenigma:
```suggestion
```

subkey/subhd/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2018"
 [dependencies]
 hex-view = "0.1.3"
 primitives = { package = "substrate-primitives",  version = "*", path = "../../core/primitives" }
-wookong-solo = "0.1.2"
+wookong-solo = {version = "0.1.4"}
```
kianenigma:
```suggestion
wookong-solo = { version = "0.1.4" }
```

subkey/subhd/src/lib.rs
```diff
@@ -70,10 +74,16 @@ pub struct Wookong;
 
 impl Wallet for Wookong {
 	type Pair = sr25519::Pair;
-	fn derive_public(&self, _path: &[DeriveJunction]) -> Result<Public> {
-		// TODO: pass through _path and _password to mutate the key on the device accordingly.
+	fn derive_public(&self, path: &[DeriveJunction]) -> Result<Public> {
 		let mut pk: [u8; 32] = [0u8; 32];
-		let rv = wk_getpub(&mut pk);
+		let mut vec = Vec::new();
+		for j in path {
```
kianenigma: I am no expert in derivation paths nor the hardware wallet integration, but regardless I don't get this loop: 

you are matching to check the type of junction, but you are doing the same thing regardless of the result?

subkey/subhd/src/lib.rs
```diff
@@ -35,6 +35,10 @@ pub enum Error {
 	DeviceNotInit,
 	/// General error of command.
 	DeviceError,
+	/// No seed matching the give account
+	NoAccount,
```
kianenigma: Are these used anywhere? or is this a WIP PR?

subkey/subhd/src/lib.rs
gavofyork: indeed; this is definitely wrong behaviour. the device needs to derive correctly according to the path.

kianenigma:
any updates?

chesterliliang:
@kianenigma We are working on the firmware part https://github.com/wookong-dot/sr25519-embedded-c
As soon as it is finished, we will test it with subkey and may submit fixes.

chesterliliang:
@burdges the implement of soft derive in schnorrkel calls 'witness_bytes' in 'derive_secret_key' which need rand_hack() to get random number for nonce. Could you please make it the 'attach_rng' to let extern random in to do that?

burdges:
It's now fixed in master, thanks!  I created that problem when I removed the Clone bound on transcripts.  I fixed it by adding a `derive_key_simple_rng` method in https://github.com/w3f/schnorrkel/commit/ad2190b1c0bbb204690a9959f7cb9427cb073273

chesterliliang:
@burdges Thanks a lot. It works well.

bkchr:
Any update?

chesterliliang:
We finished the derive work in the firmware but realized we need deliver another EVB to you for the test. That will not be a good way since the update will come up constantly. So we decide to find some way to simulate the behavior of firmware on desktop OS. QEMU is the first choice and I found some samples written in RUST, so I  can not hold myself to try to use RUST to write entire firmware.... Not a wise plan and wasted a lot of time. So may be you could close this PR and we will do it again when ready.

bkchr:
Okay :)

Issue #1262, opened 2018-12-12
Get rid of wasm-gc


pepyakin:
[wasm-gc](https://github.com/alexcrichton/wasm-gc) got deprecated but we still use it in our build pipeline.

The author mentions that:

> The Rust compiler now natively supports --gc-sections when linking wasm executables, which means wasm executables already have 90% of their garbage removed when coming out of the compiler.

But for us this is not entirely true and the size of the runtime is:

- 882K baseline
- 636K with wasm-gc

And

> The wasm-pack (and wasm-bindgen) project will already run this by default for you, so there's no need to run it again.

But we don't use neither of these.

We need to figure out what do we want to use.

pepyakin:
According to `twiggy diff node_runtime.wasm node_runtime.compact.wasm`:

```
 Delta Bytes │ Item
─────────────┼─────────────────────────────────
     -252601 ┊ <total>
      -68707 ┊ custom section '.debug_str'
      -56447 ┊ custom section '.debug_line'
      -52165 ┊ custom section '.debug_info'
      -16040 ┊ custom section '.debug_ranges'
      -10261 ┊ custom section '.debug_pubnames'
       -6072 ┊ custom section '.debug_pubtypes'
       -1432 ┊ custom section '.debug_abbrev'
         -20 ┊ "function names" subsection
         -18 ┊ custom section '.debug_macinfo'
          -3 ┊ __wasm_call_ctors
          -1 ┊ elem[0]
          -1 ┊ func[357]
```

so major source of the cruft is `debug_` sections.

There is a tool called `wasm-strip` in the wabt suite, which strips all custom sections.

```
882K baseline
643K after wasm-strip
636K after wasm-gc
```

however, if we decide to depend on wabt in our build pipeline we could then use wabt's wasm-opt, which could optimize binary (including pruning of dead symbols) and produce a binary with size of  `561K`.

bkchr:
Why do we have debug sections in a release build? Shouldn't they be removed?

pepyakin:
I'd say they should be removed! But for now this is not the case (and it has been this way for a while).
And I'm not sure why is this that way, but I asked why.

pepyakin:
So, a quick recap: 

- substrate uses a wasm file to encode logic of blockchain. We call it wasm runtime
- an example wasm runtime is written in rust, built with cargo. Rust uses LLVM LLD linker to link rust libraries and produce a final wasm blob
[...]

PR #3854, opened 2019-10-18
Obviate manual dependency on wasm-gc executable

bddap:
This PR Improves user experience by replacing the dependency on the [wasm-gc](https://crates.io/crates/wasm-gc) executable with its programatic interface, [wasm-gc-api](https://crates.io/crates/wasm-gc).


bddap:
[clabot:check]

kianenigma:
except, fyi, the fact that https://github.com/paritytech/substrate/issues/1262 is trying to remove wasm-gc altogether.

PR #3856, opened 2019-10-19
Price to Weight ratio as type parameter

JoshOrndorff:
Builds on top of https://github.com/paritytech/substrate/pull/3823 but has _both_ desirable properties:
* All constants declared in `parameter_types!` are used as associated types with `get`
* All transaction fee calculation constants are in one place

@kianenigma and I had both previously wanted to use an _instance_ of a struct to calculate weight fees because we could give it a field to represent the slope. But `transaction_payment::Trait` requires an associated _type_. Generating a type when you really want a constant is exactly what `parameter_types!` is for. The solution is to make the fee:weight ratio itself a type parameter for the struct as I've done here.

### DIFF
--- a/node/runtime/src/impls.rs
+++ b/node/runtime/src/impls.rs
@@ -22,7 +22,7 @@ use sr_primitives::traits::{Convert, Saturating};
 use sr_primitives::{Fixed64, Perbill};
 use support::traits::{OnUnbalanced, Currency, Get};
 use crate::{
-	Balances, System, Authorship, MaximumBlockWeight, NegativeImbalance, WeightToFee,
+	Balances, System, Authorship, MaximumBlockWeight, NegativeImbalance,
 	TargetedFeeAdjustment,
 };
 
@@ -49,10 +49,17 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
 	fn convert(x: u128) -> Balance { x * Self::factor() }
 }
 
-/// Simply multiply by a coefficient denoted by the `Get` implementation.
-impl Convert<Weight, Balance> for WeightToFee {
-	fn convert(x: Weight) -> Balance {
-		Balance::from(x).saturating_mul(Self::get())
+// M is the coefficient in units of currency / weight
+pub struct LinearWeight<Multiplier>(Multiplier);
+
+impl<Multiplier> Convert<Weight, Balance> for LinearWeight<Multiplier>
+ 	where Multiplier: Get<u128> {
+
+	fn convert(w: Weight) -> Balance {
+		// substrate-node a weight of 10_000 (smallest non-zero weight) to be mapped to 10^7 units of
+		// fees, hence:
+		let multiplier = Multiplier::get();
+		Balance::from(w).saturating_mul(multiplier)
 	}
 }
 
--- a/node/runtime/src/impls.rs
+++ b/node/runtime/src/impls.rs
@@ -50,17 +50,17 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
 }
 
 /// Convert from weight to balance via a simple coefficient multiplication
-/// The associated type M encapsulates a constant in units of balance per weight
-pub struct LinearWeight<M>(M);
+/// The associated type C encapsulates a constant in units of balance per weight
+pub struct LinearWeightToFee<C>(C);
 
-impl<M> Convert<Weight, Balance> for LinearWeight<M>
- 	where M: Get<Balance> {
+impl<C> Convert<Weight, Balance> for LinearWeightToFee<C>
+ 	where C: Get<Balance> {
 
 	fn convert(w: Weight) -> Balance {
 		// substrate-node a weight of 10_000 (smallest non-zero weight) to be mapped to 10^7 units of
 		// fees, hence:
-		let multiplier = M::get();
-		Balance::from(w).saturating_mul(multiplier)
+		let coefficient = C::get();
+		Balance::from(w).saturating_mul(coefficient)
 	}
 }
 
--- a/node/runtime/src/impls.rs
+++ b/node/runtime/src/impls.rs
@@ -51,7 +51,7 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
 
 /// Convert from weight to balance via a simple coefficient multiplication
 /// The associated type C encapsulates a constant in units of balance per weight
-pub struct LinearWeightToFee<C>(C);
+pub struct LinearWeightToFee<C>(rstd::marker::PhantomData<C>);
 
 impl<C> Convert<Weight, Balance> for LinearWeightToFee<C>
  	where C: Get<Balance> {
--- a/node/runtime/src/impls.rs
+++ b/node/runtime/src/impls.rs
@@ -49,16 +49,17 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
 	fn convert(x: u128) -> Balance { x * Self::factor() }
 }
 
-// M is the coefficient in units of currency / weight
-pub struct LinearWeight<Multiplier>(Multiplier);
+/// Convert from weight to balance via a simple coefficient multiplication
+/// The associated type M encapsulates a constant in units of balance per weight
+pub struct LinearWeight<M>(M);
 
-impl<Multiplier> Convert<Weight, Balance> for LinearWeight<Multiplier>
- 	where Multiplier: Get<u128> {
+impl<M> Convert<Weight, Balance> for LinearWeight<M>
+ 	where M: Get<Balance> {
 
 	fn convert(w: Weight) -> Balance {
 		// substrate-node a weight of 10_000 (smallest non-zero weight) to be mapped to 10^7 units of
 		// fees, hence:
-		let multiplier = Multiplier::get();
+		let multiplier = M::get();
 		Balance::from(w).saturating_mul(multiplier)
 	}
 }
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -65,7 +65,7 @@ pub use staking::StakerStatus;
 
 /// Implementations of some helper traits passed into runtime modules as associated types.
 pub mod impls;
-use impls::{CurrencyToVoteHandler, Author, LinearWeight};
+use impls::{CurrencyToVoteHandler, Author, LinearWeightToFee};
 
 /// Constant values used within the runtime.
 pub mod constants;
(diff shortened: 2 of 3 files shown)


node/runtime/src/impls.rs
```diff
@@ -49,10 +49,17 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
...
 
-/// Simply multiply by a coefficient denoted by the `Get` implementation.
-impl Convert<Weight, Balance> for WeightToFee {
-	fn convert(x: Weight) -> Balance {
-		Balance::from(x).saturating_mul(Self::get())
+// M is the coefficient in units of currency / weight
+pub struct LinearWeight<Multiplier>(Multiplier);
+
+impl<Multiplier> Convert<Weight, Balance> for LinearWeight<Multiplier>
+ 	where Multiplier: Get<u128> {
```
kianenigma:
```suggestion
 	where Multiplier: Get<Balance> {
```

node/runtime/src/impls.rs
```diff
@@ -49,10 +49,17 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
 	fn convert(x: u128) -> Balance { x * Self::factor() }
 }
 
-/// Simply multiply by a coefficient denoted by the `Get` implementation.
-impl Convert<Weight, Balance> for WeightToFee {
-	fn convert(x: Weight) -> Balance {
-		Balance::from(x).saturating_mul(Self::get())
+// M is the coefficient in units of currency / weight
+pub struct LinearWeight<Multiplier>(Multiplier);
```
kianenigma:
```suggestion
pub struct LinearWeight<M>(M);
```

kianenigma:
I do recall our discussion, but can't quite see what the PR is trying to achieve here. Can you elaborate a bit more? `WeightToFee`'s should internal implementation seems to be the same and it is only a matter of naming? 

I think the logic of a simple coefficient multiplier is simple enough to be handled by just one struct, rather than one encapsulating the number and one encapsulating the multiplication. What do you think?

node/runtime/src/impls.rs
```diff
@@ -49,10 +49,17 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
 	fn convert(x: u128) -> Balance { x * Self::factor() }
 }
 
-/// Simply multiply by a coefficient denoted by the `Get` implementation.
-impl Convert<Weight, Balance> for WeightToFee {
-	fn convert(x: Weight) -> Balance {
-		Balance::from(x).saturating_mul(Self::get())
+// M is the coefficient in units of currency / weight
```
kianenigma:
```suggestion
/// M is the coefficient in units of currency / weight
```

and it should be a proper doc of what it is

kianenigma:
Although, maybe @thiolliere would be happier like this? you commented that you think having both `Get<>` and `Convert<>` implemented for one struct is confusing and you would rather have them separated. If you think this is better, I am happy to proceed.

JoshOrndorff:
> I think the logic of a simple coefficient multiplier is simple enough to be handled by just one struct, rather than one encapsulating the number and one encapsulating the multiplication. What do you think?

Here's what the two types are representing
- `TransactionWeightFree` -- A wrapper type to encapsulate the constant coefficient. We do this for constants all over.
- `LinearWeight` -- A to so linear conversion with _some_ scale factor. Because we can't use an instance of `LinearWeight`, we must specify that coefficient as a type parameter.

Perhaps this could exemplified more simply with a more general example. Consider this alternative weight to fee converter.
```rust
/// Convert from weight to balance via a simple coefficient multiplication
/// The type parameters encapsulate the coefficients
pub struct QuadraticWeight<C0, C1, C2>(C0, C1, C2);

impl<M> Convert<Weight, Balance> for QuadraticWeight<C0, C1, C2>
 	where C0: Get<Balance>, C1: Get<Balance>, C2: Get<Balance> {

	fn convert(w: Weight) -> Balance {
		let c0 = C0::get();
		let c1 = C1::get();
		let c2 = C2::get();
		let w0 = Balance::from(w);

		//TODO use saturating arithmetic
		c0 + c1 * w0 + c2 * w1 * w1
	}
}
```

Then to actually use the quadratic converter in your runtime you would need to do something like this
```rust
parameter_types! {
	pub const TransactionBaseFee: Balance = 1 * CENTS;
	pub const TransactionByteFee: Balance = 10 * MILLICENTS;

	// Now there are even more structs because there are even more constants
	pub const WeightFeeConstant: Balance = 1_000;
	pub const WeightFeeLinear: Balance = 100;
	pub const WeightFeeQuadratic : Balance = 10;

	// for a sane configuration, this should always be less than `AvailableBlockRatio`.
	pub const TargetedFeeAdjustment: Perbill = Perbill::from_percent(25);
}
```
The above code separate the ultimate weight to fee calculation into _four_ structs. One actual struct that represents quadratic calculation, and three that just wrap the constants. It wouldn't make sense to implement the entire conversion on any one of the three constant-wrapping types here.

JoshOrndorff:
I still like this idea, but I also had another idea that you may prefer.

New idea: Turn an _instance_ of 
```rust
parameter_types!{
  // --snip--
  pub const WeightToFee: LinearWeightFee = LinearWeightFee{ coefficient: 10_000 };
}
```
I've prototyped it https://github.com/paritytech/substrate/commit/420e162b184965934e785d15391755fdaefbf0a6 but I can't get it to compile. I'll wait for feedback before doing anything else.

kianenigma:
Separating the coefficient into a separate struct only makes sense in an example like `QuadraticWeight` where you have multiple of them, because then you cannot distinguish between `Get<C0>, Get<C1>` etc. unless if you create a custom trait; 

in `node_runtime` there is only one and I still think simplicity of not having multiple config structs laying around is more important. One struct can be `Convert<_, _> + Get<_>` and that works perfectly fine. The case of a user wanting to implement quadratic is more of an _example_, and not enough of a reason to be added to both substrate node and polakdot.

Even if not in node-runtime, we should put this example in the docs or maybe even in node-template. 

Nonetheless, I will add some naming suggestions here to make it better, if we decide to merge.

kianenigma:
> I still like this idea, but I also had another idea that you may prefer.
> 
> New idea: Turn an _instance_ of
> 
> ```rust
> parameter_types!{
>   // --snip--
>   pub const WeightToFee: LinearWeightFee = LinearWeightFee{ coefficient: 10_000 };
> }
> ```
> 
> I've prototyped it [420e162](https://github.com/paritytech/substrate/commit/420e162b184965934e785d15391755fdaefbf0a6) but I can't get it to compile. I'll wait for feedback before doing anything else.

I don't think you should pursue this. Using `self` in `Convert` will not work as it is called as a type function like `T::AssociatedType::function()`, hence will never have self.

node/runtime/src/impls.rs
```diff
@@ -49,10 +49,18 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
 	fn convert(x: u128) -> Balance { x * Self::factor() }
 }
 
-/// Simply multiply by a coefficient denoted by the `Get` implementation.
-impl Convert<Weight, Balance> for WeightToFee {
-	fn convert(x: Weight) -> Balance {
-		Balance::from(x).saturating_mul(Self::get())
+/// Convert from weight to balance via a simple coefficient multiplication
+/// The associated type M encapsulates a constant in units of balance per weight
+pub struct LinearWeight<M>(M);
```
kianenigma:
```suggestion
pub struct LinearWeightToFee<C>(C);
```

(Sorry if I am taking back my own word :D, if it is a coefficient then C makes more sense. sry)

node/runtime/src/impls.rs
```diff
@@ -49,10 +49,18 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
...
+
+impl<M> Convert<Weight, Balance> for LinearWeight<M>
+ 	where M: Get<Balance> {
+
+	fn convert(w: Weight) -> Balance {
+		// substrate-node a weight of 10_000 (smallest non-zero weight) to be mapped to 10^7 units of
+		// fees, hence:
+		let multiplier = M::get();
```
kianenigma:
```suggestion
		let coefficient = C::get();
```

node/runtime/src/lib.rs
```diff
@@ -178,7 +178,7 @@ parameter_types! {
 	pub const TransactionBaseFee: Balance = 1 * CENTS;
 	pub const TransactionByteFee: Balance = 10 * MILLICENTS;
 	// setting this to zero will disable the weight fee.
-	pub const WeightToFee: Balance = 1_000;
+	pub const TransactionWeightFee: Balance = 1_000;
```
kianenigma:
```suggestion
	pub const WeightFeeCoefficient: Balance = 1_000;
```

node/runtime/src/lib.rs
```diff
@@ -188,7 +188,7 @@ impl transaction_payment::Trait for Runtime {
 	type OnTransactionPayment = DealWithFees;
 	type TransactionBaseFee = TransactionBaseFee;
 	type TransactionByteFee = TransactionByteFee;
-	type WeightToFee = WeightToFee;
+	type WeightToFee = LinearWeight<TransactionWeightFee>;
```
kianenigma:
```suggestion
	type WeightToFee = LinearWeightToFee< WeightFeeCoefficient >;
```

thiolliere:
Yes I think I like this, the parameter type is only used in the same file, and there is no like hidden implementation on top of it in another file.

Also maybe LinearWeight structure could be moved to srml/transaction-payment as an optionnal useful structure to use.

node/runtime/src/lib.rs
thiolliere: I see you like parameter types, and I'm ok now. But we can keep in mind that this could be written
`LinearWeightToFee::new(1_000)`  or `LinearWeightToFee { coefficient: 1_000 }`

JoshOrndorff: I also tried this approach but couldn't get it to compile. @kianenigma [replied above](https://github.com/paritytech/substrate/pull/3856#issuecomment-544254122)

> I don't think you should pursue this. Using self in Convert will not work as it is called as a type function like T::AssociatedType::function(), hence will never have self.

So I'll stick with the type parameter for now. Would be happy to learn if this approach can work though, because I do like it better.

thiolliere: no indeed I forget it was just a type not a variable

JoshOrndorff:
I've included the renames @kianenigma requested. I'm ready to merge this back into https://github.com/paritytech/substrate/pull/3823 and if necessary finish any last pieces there.

PR #3857, opened 2019-10-19
 Add an OnUnbalanced hook for contract rent payments

### LINKED ISSUE #2455: Add OnUnbalanced hook for contract rent payment
This allow other modules to collect the rent payment and and redistribute or burn it. For example I would like to add the rent payments to validator rewards.

Similar to how balances module offers `TransferPayment` hook

https://github.com/paritytech/substrate/blob/3669cf7b7e2ab6bb130715a082f4db9b6014743b/srml/balances/src/lib.rs#L761

expenses:
This closes #2455.


srml/contracts/src/rent.rs
```diff
@@ -165,6 +168,8 @@ fn try_evict_or_and_pay_rent<T: Trait>(
 			..contract
 		});
 
+		T::RentPayment::on_unbalanced(NegativeImbalanceOf::<T>::wrap(dues));
```
kianenigma:
```suggestion
		T::RentPayment::on_unbalanced(imbalance);
```

kianenigma: and further up do `let imbalance = T::Currency::withdraw(..)`

srml/support/src/traits.rs
```diff
@@ -203,6 +203,9 @@ pub trait Imbalance<Balance>: Sized {
 	/// Drop an instance cleanly. Only works if its `value()` is zero.
 	fn drop_zero(self) -> Result<(), Self>;
 
+	/// Wrap a balance as an imbalance.
```
```suggestion

```

Should be gone, likewise for the implementation

srml/contracts/src/rent.rs
expenses: Oops! Thanks!

expenses:
Ok, I've fixed the code to use `imbalance`. A polkadot PR will have to wait a bit due to `polkadot-master` being significantly behind `master`.

kianenigma:
Okay at some point before the update of `polkadot-master` it should be taken care of. 

I will leave it to @pepyakin or @jimpo who know contracts better to maybe have a final look?

srml/contracts/src/rent.rs
```diff
@@ -165,6 +169,9 @@ fn try_evict_or_and_pay_rent<T: Trait>(
 			..contract
 		});
 
+		let imbalance = imbalance.expect("Imbalance has been checked above");
```
pepyakin: I am not sure if I am following the proof. Where exactly it's been checked? What prevents it from being `None` here?

expenses: I should clarify that. `imbalance` is `Some` if `insufficient_rent || pay_rent` is `true`. [Below, on line 142](https://github.com/paritytech/substrate/pull/3857/files#diff-18b5bc312e600d7b4d3c7ea047d935ccL142) `insufficient_rent || !can_withdraw_rent` is checked, and if `true` then it returns, so when `pay_rent` is checked below, `imbalance` has to be `Some`.

expenses: I've clarified the proof:
```
imbalance is Some if can_withdraw_rent && (insufficient_rent || pay_rent) is true;
if can_withdraw_rent is false, the function has returned;
pay_rent is true;
qed
```

expenses: nope, nevermind, I was right the first time.

srml/contracts/src/rent.rs
```diff
@@ -137,7 +137,11 @@ fn try_evict_or_and_pay_rent<T: Trait>(
 			dues_limited < rent_budget < balance - subsistence < balance - existential_deposit;
 			qed",
 		);
-	}
+
```
thiolliere: why don't you just handle the imbalance here ? with `T::RentPayment::on_unbalanced(imbalance);` directly.

expenses: That works!

expenses: The tests seem to fail due to the withdrawn amount being zero: 
```
---- tests::restorations_dirty_storage_and_different_storage stdout ----
thread 'tests::restorations_dirty_storage_and_different_storage' panicked at 'assertion failed: `(left != right)`
  left: `0`,
 right: `0`', srml/contracts/src/tests.rs:305:3
```

thiolliere: does in case of insufficient_rent the imbalance be dropped instead of put in RentPayment ?

expenses:
@thiolliere yes, because
```rust
if insufficient_rent || !can_withdraw_rent {
    ...
    return (RentOutcome::Evicted, Some(tombstone_info));
}
```

happens before `T::RentPayment::on_unbalanced(imbalance);` is called.

srml/contracts/src/rent.rs
thiolliere: but indeed the withdraw is correctly 0 as the rent_budget of the contract is zero.

I'm not sure if the test is relevant, if yes then we can just disable non null withdraw in case of this restoration tests.
you can do it with:
```diff
@@ -297,12 +297,16 @@ fn compile_module<T>(wabt_module: &str)
        Ok((wasm, code_hash))
 }
 
+pub const DISABLE_NON_ZERO_RENT_ASSERT: AtomicBool = AtomicBool::new(false);
+
 // A hook for rent payments to check that the rent is not zero.
 pub struct NonZeroRentHook;
 
 impl OnUnbalanced<NegativeImbalanceOf<Test>> for NonZeroRentHook {
        fn on_unbalanced(amount: NegativeImbalanceOf<Test>) {
-               assert_ne!(amount.peek(), BalanceOf::<Test>::zero());
+               if !DISABLE_NON_ZERO_RENT_ASSERT.load(..) {
+                       assert_ne!(amount.peek(), BalanceOf::<Test>::zero());
+               }
        }
 }
```

expenses: Ok, good idea.

expenses: I think that for the time being, I'm just going to get rid of the test though.

expenses:
Oh, haha, no changes to polkadot are needed because it doesn't use `srml-contracts`.

PR #3861, opened 2019-10-19
Add SECP256k1/ECDSA support for transaction signing

gavofyork:
This introduces SECP256k1 crypto into Substrate. It also switches Node from `AnySigner` to `MultiSigner` for its transaction signatures. To make this work, it refactors the crypto/account framework to introduce a difference between `Public` (the public key type) and `AccountId` (the type used to identify accounts). This is necessary to retain backwards compatibility to Kusama/`AnySigner` and to avoid potential alignment and size pessimisations, keeping everything to a clean 32 bytes.

As far as account crypto goes, there are three levels:

- `Pair` (for creating transactions)
- `Public` (always derivable from `Pair` through the associated type `Verify::Signer`, and in the case of SECP256k1/ECDSA, also from a valid `Signature` with the ECDSA Recover opreation)
- `AccountId` (always derivable from `Public` using `IdentifyAccount::AccountId`)

Signatures are verified using an `AccountId` only, not a `Public` key. For simple cryptographies, `AccountId` continues to be equivalent to `Public` and everything works as before. However for `MultiSigner`, things are a little different, as we want to retain 32 byte `AccountId`s. The `MultiSignature` type itself encodes the underlying cryptography (ECDSA/*25519), and in the case of the *25519 crypto, the public key is bytewise equivalent to the public key and this a normal Verify operation can happen. However, in the case of ECDSA/SECP256k1 crypto, the `AccountId` is the hash of the 33 byte public key (to securely reduce it to 32 bytes), and as such we cannot use the normal Verify operation. Instead, we use the Recover operation, and ensure that the `AccountId` derived from the recovered public key matches that we expect.

As such, the `Public` datatype now gets used directly much less. In principle this opens to the door to an Ethereum-like account ID compression where we always hash the public key and throw away some number of bytes from it, giving us a modest space saving.

It adds an ECDSA Recover host function for the compressed public key which is needed for our signature format.

The last commit 8257a6a17e7f369e46fb381a48ea6133c189a2f8 may be viewed independently and just adds support to `subkey` for viewing the seed of a derived keypair where possible.


core/executor/src/host_interface.rs
```diff
@@ -761,6 +761,7 @@ impl_wasm_host_interface! {
 		}
 
 		ext_secp256k1_ecdsa_recover(
+			compressed: u32,
```
bkchr: This will break Kusama.

core/sr-io/src/lib.rs
```diff
@@ -205,15 +205,20 @@ export_api! {
...
 		/// Verify and recover a SECP256k1 ECDSA signature.
 		/// - `sig` is passed in RSV format. V should be either 0/1 or 27/28.
-		/// - returns `Err` if the signature is bad, otherwise the 64-byte pubkey (doesn't include the 0x04 prefix).
+		/// - returns `Err` if the signature is bad, otherwise the 64-byte raw pubkey (doesn't include the 0x04 prefix).
 		fn secp256k1_ecdsa_recover(sig: &[u8; 65], msg: &[u8; 32]) -> Result<[u8; 64], EcdsaVerifyError>;
+
+		/// Verify and recover a SECP256k1 ECDSA signature.
+		/// - `sig` is passed in RSV format. V should be either 0/1 or 27/28.
+		/// - returns `Err` if the signature is bad, otherwise the 33-byte compressed pubkey.
+		fn secp256k1_ecdsa_recover_compressed(sig: &[u8; 65], msg: &[u8; 32]) -> Result<[u8; 33], EcdsaVerifyError>;
```
bkchr: Why not merge both functions?

core/sr-io/without_std.rs
```diff
@@ -938,7 +943,20 @@ impl CryptoApi for () {
...
 		let mut pubkey = [0u8; 64];
 		match unsafe {
-			ext_secp256k1_ecdsa_recover.get()(msg.as_ptr(), sig.as_ptr(), pubkey.as_mut_ptr())
+			ext_secp256k1_ecdsa_recover.get()(0, msg.as_ptr(), sig.as_ptr(), pubkey.as_mut_ptr())
+		} {
+			0 => Ok(pubkey),
+			1 => Err(EcdsaVerifyError::BadRS),
+			2 => Err(EcdsaVerifyError::BadV),
+			3 => Err(EcdsaVerifyError::BadSignature),
+			_ => unreachable!("`ext_secp256k1_ecdsa_recover` only returns 0, 1, 2 or 3; qed"),
```
bkchr: You return `4` in the implementation as well.

core/executor/src/host_interface.rs
```diff
@@ -788,9 +789,11 @@ impl_wasm_host_interface! {
 				_ => return Ok(3),
 			};
 
-			context.write_memory(pubkey_data, &pubkey.serialize()[1..65])
-				.map_err(|_| "Invalid attempt to set pubkey in ext_secp256k1_ecdsa_recover")?;
-
+			match compressed {
+				0 => context.write_memory(pubkey_data, &pubkey.serialize()[1..65]),
```
bkchr: why not `if compressed == 0 {} else {}`?

Then we don't have this new return value `4`

core/sr-io/without_std.rs
gavofyork: yeah but it can never be reached, hence the `unreachable`

core/sr-io/src/lib.rs
gavofyork: how?

core/executor/src/host_interface.rs
gavofyork: ugh. right.

core/sr-io/src/lib.rs
bkchr: Can we not just expose the `compressed` parameter?

gavofyork: not without having a dynamic `enum` return type, which would hamper usability.

gavofyork: not without having a dynamic `enum` return type, which would hamper usability.

core/primitives/src/ecdsa.rs
```diff
@@ -0,0 +1,621 @@
...
+#[cfg(feature = "std")]
+use secp256k1::{PublicKey, SecretKey};
+
+/// A secret seed (which is bytewise essentially equivalent to a SecretKey).
+///
+/// We need it as a different type because `Seed` is expected to be AsRef<[u8]>.
+#[cfg(feature = "std")]
+type Seed = [u8; 32];
+
+/// The Blake2-256 hash of the ECDSA 33-byte compressed public key.
```
jimpo: This comment seems incorrect as the defined struct is 32 bytes.

core/sr-primitives/src/traits.rs
```diff
@@ -50,47 +50,83 @@ impl<'a> Lazy<[u8]> for &'a [u8] {
...
+	type AccountId = Self;
+	fn into_account(self) -> Self { self }
+}
+
+impl IdentifyAccount for primitives::sr25519::Public {
+	type AccountId = Self;
+	fn into_account(self) -> Self { self }
+}
+
+impl IdentifyAccount for primitives::ecdsa::Public {
```
jimpo: It's weird that with the `MultiSigner` `into_account` hashes the key to 32 bytes whereas with this it does not. Why?

core/sr-io/src/lib.rs
jimpo: It's possible to leave these two API functions separate and still have a single Wasm host function that takes a `compressed` parameter (since that just returns a u32 status and writes bytes to the output).

gavofyork: i just added the extra function - backwards compatible now.

core/sr-io/without_std.rs
gavofyork: outdated.

core/sr-primitives/src/traits.rs
gavofyork: Basically because the public key of the *25519 crypto is a straight 32-bytes, whereas the public key of SECP256k1 is an annoying 33 bytes. Since 32 bytes is a nice alignment that might have some good optimisations (e.g. for cache lines), and since `AccountId` is used copiously, I don't especially want to risk pessimising everything for the sake of supporting SECP256k1.

This basically just follows the pattern of Ethereum where the public key is hashed and that hash is used for the public account (though unlike Ethereum, we don't throw away 12 bytes of the hash).

For the *25519 crypto, it remains backwards compatible with Kusama.

core/executor/src/host_interface.rs
```diff
@@ -790,7 +790,39 @@ impl_wasm_host_interface! {
...
 				.map_err(|_| "Invalid attempt to set pubkey in ext_secp256k1_ecdsa_recover")?;
+			Ok(0)
+		}
+
+		ext_secp256k1_ecdsa_recover_compressed(
+			msg_data: Pointer<u8>,
+			sig_data: Pointer<u8>,
+			pubkey_data: Pointer<u8>,
+		) -> u32 {
+			let mut sig = [0u8; 65];
```
jimpo: Could you at least factor out the common code between this and `ext_secp256k1_ecdsa_recover` into a separate function?

core/sr-primitives/src/traits.rs
jimpo: I was asking more why `primitives::ecdsa::public as IdentifyAccount` does not do the hashing, not  why `MultiSigner` does. I feel like it's be better if they were consistent with each other.

gavofyork: `MultiSigner` is what forces the need for 32 byte `AccountId`s, given it's used to be backwards compatible with Kusama and the other two entries are naturally 32 bytes.

gavofyork:
Line width cannot be fixed due to limitation in `hex` macro macro.

core/primitives/src/crypto.rs
```diff
@@ -604,7 +700,10 @@ pub trait Pair: CryptoType + Sized + Clone + Send + Sync + 'static {
 	fn from_phrase(phrase: &str, password: Option<&str>) -> Result<(Self, Self::Seed), SecretStringError>;
 
 	/// Derive a child key from a series of given junctions.
-	fn derive<Iter: Iterator<Item=DeriveJunction>>(&self, path: Iter) -> Result<Self, Self::DeriveError>;
+	fn derive<Iter: Iterator<Item=DeriveJunction>>(&self,
+		path: Iter,
+		seed: Option<Self::Seed>
```
kianenigma:
```suggestion
		seed: Option<Self::Seed>,
```

core/primitives/src/ecdsa.rs
```diff
@@ -0,0 +1,608 @@
...
+
+#[cfg(feature = "std")]
+impl std::hash::Hash for Signature {
+	fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+		std::hash::Hash::hash(&self.0[..], state);
+	}
+}
+
+impl Signature {
+	/// A new instance from the given 64-byte `data`.
```
kianenigma:
```suggestion
	/// A new instance from the given 65-byte `data`.
```

Issue #3815, opened 2019-10-14
Where's the `start` function in sr-sandbox ?


yjhmelody:
confused about https://github.com/paritytech/substrate/blob/master/core/sr-sandbox/src/lib.rs#L177
```
If execution of `start` function generated a trap, then `Err(Error::Execution)` be returned.
```

I see the sandbox's Instance code and tests. I dont know where the `start` function is,
Like https://github.com/paritytech/substrate/blob/master/core/sr-sandbox/with_std.rs#L313

Also, what's the state param mean in Instance::new()?

And any different in sr-sandbox and executor/sandbox?

bkchr:
CC @pepyakin

pepyakin:
start function is a term that is coming from wasm spec. It is a function that may be defined by a wasm module to be executed at instantiation time. Note that rustc by default doesn’t emit it.

state parameter is needed for passing around additional parameter. Use case is to pass around some user defined context in the host functions. For instance, in sr-sandbox uses it to pass around the context of type associated with the instance and it has a type of T

yjhmelody:
@pepyakin 
So that  a wasm code which not include `start`, won't be run if don't use `invoke` function ?

NikVolf:
@yjhmelody if you have no start function, this start step is just no-op, so no need to disable anything

pepyakin:
what @NikVolf said. Ignoring the start function if it is present in the module would make such execution environment non-spec compilant since executing the `start` function as the last step of instantiation is mandated by the spec.

PR #3864, opened 2019-10-21
Fix ambiguity about start function in sr-sandbox doc

NikVolf:


NikVolf:
asked in #3815

bkchr:
This needs resolving and also does not require a bump of the impl_version. Just ignore the check runtime test.

PR #3866, opened 2019-10-21
Fix deserialization of `Bytes`

tomusdrw:
Closes: #3833 
Related: https://github.com/paritytech/parity-common/pull/243


tomusdrw:
@NikVolf Actually this bug was just introduced around a month ago as part of optimization of that path.

Issue #3722, opened 2019-09-30
Offchain Workers: deterministic bookkeeping


rphmeier:
cc @tomusdrw , @jimpo 

Offchain workers are a feature of Substrate allowing us to provide code in the Runtime which may be non-deterministic, and is intended to be executed for each _new_ block to perform tasks such as:
  - querying a price feed
  - making an HTTP request
  - creating a DHT entry
  - replicating a file

For nondeterministic tasks such as the above, it would be bad to execute the offchain logic on ancient blocks while performing a major synchronization. You might end up re-executing logic initially triggered years ago, to absolutely no effect. Because of this, offchain workers are designed not to execute for every block in the chain, as a full node.

I've encountered two use-cases which fall into another category of execution: deterministic bookkeeping. These are situations where the computation is deterministic but data-heavy, and we want to off-load data (typically trie nodes) off of the chain state (where trie roots are kept). For these use-cases, the current operation of offchain workers does not seem to be sufficient.

**Example 1: Merkle Mountain Ranges (MMR)**
#2053 , https://github.com/mimblewimble/grin/blob/master/doc/mmr.md

For many kinds of auxiliary blockchain protocols, it's important to be able to prove that some ancient block header is an ancestor of the finalized chain head. MMRs provide a good way of doing that.

We want to write a runtime module to keep track of the _peaks_ (roots) of a bunch of different merkle tries - there will be `log2(N)` of these for `N` blocks (and `N` trie nodes in total). You can add to the MMR with only the peaks, and prove ancestry if you have all the nodes.

We'd want full nodes to keep track of all of the MMR nodes by keeping them in offchain storage. However, if even one block in the chain is not executed, it is possible to end up in a situation where ancestry can no longer be proven.

**Example 2: Historical Slashing**

`srml-staking` and `srml-session` are designed so that validators and nominators can be slashed for a long bonding duration while they wait for their money to be withdrawable. Keeping months' worth of historical validator sets, session keys, and nominator assignments on-chain is too heavy, so we instead keep a trie root encoding the historical validator sets for every session. Full nodes are intended to keep this trie root.

For slashing, the situation isn't as severe. However, for security it would be best to have as many full nodes as possible be able to report misbehavior. If a full node doesn't execute the off-chain worker, it may not have the trie nodes necessary to issue a report of a misbehavior that it witnesses - reducing the effectiveness of fishermen.

**Final notes**

[...]

PR #3868, opened 2019-10-21
grandpa: report equivocations

andresilva:
This PR adds equivocation reporting for GRANDPA. Equivocation detection is implemented in [finality-grandpa](https://github.com/paritytech/finality-grandpa) crate and exposed through `Environment` hooks (`prevote_equivocation` and `precommit_equivocation`).

The GRANDPA runtime module is extended with a `report_equivocation` method that takes an equivocation proof and a session membership proof. This method validates the membership proof using an abstract `KeyOwnerProofSystem` (historical session module currently) and the equivocation proof (different vote targets, valid signatures, etc.), if both are valid an offence is reported to an abstract `ReportOffence` type (offences module currently). Additionally, the method `submit_report_equivocation_extrinsic` creates and submits a signed transaction for reporting an equivocation.

From the node-side, when we want to report an equivocation we need to create the session membership proof. In particular, the membership proof can only be created at a block when the given validator is part of the current session. We figure out the last block at which the current set id is active in a given chain by looking at previously signalled authority changes and create the membership proof at that block. We then call `submit_report_equivocation_extrinsic` at the best block.

### Important

1. In order to report an equivocation we need to send a signed transaction, therefore we need to have some key available for this, taking into account that this is also the account that will be potentially rewarded. I created a new app-specific crypto ("fish") that should be used for all equivocation reporting tasks (this will also be used for BABE later on). Any comments on this?

1. The current implementation for generating membership proofs requires having state available for the block on which the session where the equivocation happened was valid. This is not problematic right now since we require validators to run as archive. A future implementation could leverage offchain workers for generating full membership proof tries when a session ends and storing them locally.

**TODO**:
- [x] Publish finality-grandpa 0.9.1 (includes fixes for building under no-std)
- [X] Fix execution context of `submit_report_equivocation_extrinsic` call (needs to be offchain context but it's not easy to use that now outside of actual offchain workers)
- [X] Abstract srml/grandpa reporting logic to trait (so that we can implement it for `()` to disable reporting, needed to integrate in node-template)
- [x] Merge #4200 and drop cherry-picked changes
- [x] Cleanup
- [x] Tests
- [x] Docs

polkadot companion: paritytech/polkadot#1000


burdges:
We only need the key signing a report transaction to be an account key so that it can pay some fee, right?  We've no existing mechanism for actions signed by session keys to change their controller account fees yet?  We could maybe permit free reports by the grandpa ed25519 key if the deduplication works well enough, not so universally applicable though.  We can chat about this when you like.  :)

bin/node/executor/tests/submit_transaction.rs
```diff
@@ -15,7 +15,7 @@
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
 use node_runtime::{
-	Call, Executive, Indices, Runtime, SubmitTransaction, UncheckedExtrinsic,
+	Call, Executive, Indices, Runtime, SubmitImOnlineTransaction, UncheckedExtrinsic,
```
kianenigma: Whoever loses the race of merging first needs to update this as they will conflict :D 
https://github.com/paritytech/substrate/pull/4517/files#diff-81c5ce5884e59ad2b5de584efa307f0aR34

vkandy:
Would you consider allowing custom implementation of `GrandpaEquivocationOffence` to be passed to `report_offence`? We want to be able to change the formula for calculating the slash fraction. If we can avoid cloning the pallet just to be able to change the formula that would be perfect.

andresilva:
@vkandy Yeah I think that's doable, I'll add support for this.

vkandy:
@andresilva thank you!

andresilva:
I will add some more tests but I think this is ready for a review. Please read the PR description wrt the reporting requiring an app-specific account with funds and also @burdges comments on it. The easiest way to implement this is using appcrypto and my idea is that this key will be shared for any kind of reporting/fishermen activity. I'd appreciate opinions on this.

primitives/core/src/lib.rs
```diff
@@ -108,8 +108,11 @@ impl ExecutionContext {
 		match self {
 			Importing | Syncing | BlockConstruction =>
 				offchain::Capabilities::none(),
-			// Enable keystore by default for offchain calls. CC @bkchr
-			OffchainCall(None) => [offchain::Capability::Keystore][..].into(),
+			// Enable keystore and transaction pool by default for offchain calls.
+			OffchainCall(None) => [
+				offchain::Capability::Keystore,
+				offchain::Capability::TransactionPool,
```
andresilva: @tomusdrw Does this make sense? It would be hard for me to specify my own context in GRANDPA code since it requires providing `Externalities` (I think it would defeat the purpose of using offchain for this).

frame/grandpa/src/lib.rs
```diff
@@ -189,10 +211,81 @@ decl_module! {
...
+		/// against the extracted offender. If both are valid, the offence
+		/// will be reported.
+		///
+		/// FIXME: I have no clue about the weight (but we're checking two
+		/// ed25519 signatures).
+		#[weight = SimpleDispatchInfo::FixedOperational(10_000_000)]
+		fn report_equivocation(
+			origin,
+			equivocation_proof: EquivocationProof<T::Hash, T::BlockNumber>,
+			// key_owner_proof: <T::HandleEquivocation as equivocation::HandleEquivocation<T>>::KeyOwnerProof,
```
andresilva: I don't know how to fix this. @bkchr mentioned that this is probably some rustc bug.

dud1337:
Any updates on the bug fix?

andresilva:
@dud1337 I believe the overall approach is correct and good to go. There's some minor grumbles with the PR (I commented in the PR about them) and a question regarding the approach for defining the reporting account (using an app specific key). This is missing reviews and some extra tests.

bin/node/executor/tests/submit_transaction.rs
kianenigma: You lost :D

bin/node/primitives/src/lib.rs
```diff
@@ -62,3 +62,32 @@ pub type Header = generic::Header<BlockNumber, BlakeTwo256>;
...
+/// App-specific crypto used for reporting equivocation/misbehavior in BABE and
+/// GRANDPA. The crypto used is sr25519 and the account must be minimally funded
+/// in order to pay for transaction fees. Any rewards for misbehavior reporting
+/// will be paid out to this account.
+pub mod report {
+	use sp_core::crypto::KeyTypeId;
+
+	/// Key type for the reporting module. Used for reporting BABE and GRANDPA
+	/// equivocations.
+	pub const KEY_TYPE: KeyTypeId = KeyTypeId(*b"fish");
```
kianenigma: comes from fisherman?

client/finality-grandpa/src/environment.rs
```diff
@@ -924,22 +929,133 @@ where
...
+			}
+		};
 	}
 }
 
+/// Report the given equivocation to the GRANDPA runtime module. This method
+/// generates a session membership proof of the offender and then submits an
+/// extrinsic to report the equivocation. In particular, the session membership
+/// proof must be generated at the block at which the given set was active which
+/// isn't necessarily the best block if there are pending authority set changes.
```
kianenigma: when and how do we have pending authorities? is this a reflection of the queued validators in session module?

frame/grandpa/src/equivocation.rs
```diff
@@ -0,0 +1,270 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//!
```
kianenigma:
```suggestion
```

frame/grandpa/src/lib.rs
```diff
@@ -183,10 +205,81 @@ decl_module! {
...
-			ensure_signed(origin)?;
-			// FIXME: https://github.com/paritytech/substrate/issues/1112
+		/// Report voter equivocation/misbehavior. This method will verify the
+		/// equivocation proof and validate the given key ownership proof
+		/// against the extracted offender. If both are valid, the offence
+		/// will be reported.
+		///
+		/// FIXME: I have no clue about the weight (but we're checking two
+		/// ed25519 signatures).
+		#[weight = SimpleDispatchInfo::FixedOperational(10_000_000)]
```
kianenigma: Can be ignored for now but this value is 1/10 of the block limit. I don't think this is your intention?

client/finality-grandpa/src/environment.rs
```diff
@@ -924,22 +929,133 @@ where
 	fn prevote_equivocation(
 		&self,
 		_round: RoundNumber,
-		equivocation: ::finality_grandpa::Equivocation<Self::Id, Prevote<Block>, Self::Signature>
+		equivocation: finality_grandpa::Equivocation<Self::Id, Prevote<Block>, Self::Signature>,
 	) {
 		warn!(target: "afg", "Detected prevote equivocation in the finality worker: {:?}", equivocation);
-		// nothing yet; this could craft misbehavior reports of some kind.
+		match report_equivocation(
```
mxinden: Why not have `report_equivocation` a method within an `Environment` `impl`? Would save us on passing the first 3 arguments.

client/finality-grandpa/src/environment.rs
```diff
@@ -924,22 +929,133 @@ where
...
+			&self.client.clone(),
```
mxinden:
```suggestion
			&self.client,
```

client/finality-grandpa/src/environment.rs
```diff
@@ -924,22 +929,133 @@ where
...
 	fn precommit_equivocation(
 		&self,
 		_round: RoundNumber,
-		equivocation: Equivocation<Self::Id, Precommit<Block>, Self::Signature>
+		equivocation: finality_grandpa::Equivocation<Self::Id, Precommit<Block>, Self::Signature>,
 	) {
 		warn!(target: "afg", "Detected precommit equivocation in the finality worker: {:?}", equivocation);
-		// nothing yet
+		match report_equivocation(
+			&self.client.clone(),
```
mxinden:
```suggestion
			&self.client,
```

client/finality-grandpa/src/environment.rs
```diff
@@ -580,6 +584,7 @@ where
 	Block: 'static,
 	B: Backend<Block>,
 	C: crate::ClientForGrandpa<Block, B> + 'static,
+	C::Api: GrandpaApi<Block>+ SessionMembership<Block, Error = sp_blockchain::Error>,
```
mxinden:
```suggestion
	C::Api: GrandpaApi<Block> + SessionMembership<Block, Error = sp_blockchain::Error>,
```

client/finality-grandpa/src/lib.rs
```diff
@@ -615,6 +617,7 @@ pub fn run_grandpa_voter<Block: BlockT, BE: 'static, C, N, SC, VR>(
 	NumberFor<Block>: BlockNumberOps,
 	DigestFor<Block>: Encode,
 	C: ClientForGrandpa<Block, BE> + 'static,
+	C::Api: GrandpaApi<Block>+ SessionMembership<Block, Error = sp_blockchain::Error>,
```
mxinden:
```suggestion
	C::Api: GrandpaApi<Block> + SessionMembership<Block, Error = sp_blockchain::Error>,
```

client/finality-grandpa/src/lib.rs
```diff
@@ -710,6 +713,7 @@ where
 	Block: BlockT,
 	B: Backend<Block> + 'static,
 	C: ClientForGrandpa<Block, B> + 'static,
+	C::Api: GrandpaApi<Block>+ SessionMembership<Block, Error = sp_blockchain::Error>,
```
mxinden:
```suggestion
	C::Api: GrandpaApi<Block> + SessionMembership<Block, Error = sp_blockchain::Error>,
```

client/finality-grandpa/src/lib.rs
```diff
@@ -896,6 +899,7 @@ where
 	NumberFor<Block>: BlockNumberOps,
 	SC: SelectChain<Block> + 'static,
 	C: ClientForGrandpa<Block, B> + 'static,
+	C::Api: GrandpaApi<Block>+ SessionMembership<Block, Error = sp_blockchain::Error>,
```
mxinden:
```suggestion
	C::Api: GrandpaApi<Block> + SessionMembership<Block, Error = sp_blockchain::Error>,
```

bin/node/runtime/src/lib.rs
```diff
@@ -845,6 +869,16 @@ impl_runtime_apis! {
 		}
 	}
 
+	impl sp_session::SessionMembership<Block> for Runtime {
+		fn generate_session_membership_proof(
+			session_key: (sp_core::crypto::KeyTypeId, Vec<u8>),
```
rphmeier: Shouldn't the `Vec<u8>` here be a concrete key type?

client/finality-grandpa/src/environment.rs
rphmeier: @kianenigma From the runtime perspective, we don't have pending authorities. However, GRANDPA set changes are pending until finalized. When finality lags we can have a DAG of unfinalized changes. In normal circumstances a GRANDPA set-change is not pending for more than a few seconds.

client/finality-grandpa/src/environment.rs
```diff
@@ -924,22 +929,133 @@ where
...
+		// there is no pending change, the latest block for the current set is
+		// the best block.
+		None => best_header.hash(),
+	};
+
+	// generate membership proof at that block
+	let membership_proof = match client
+		.runtime_api()
+		.generate_session_membership_proof(
+			&BlockId::Hash(current_set_latest_hash),
```
rphmeier: Hopefully with the work of @drahnr on #3722  we would be able to call all of this stuff at the most recent block instead, as the offchain DB would hold the set membership tries.

bin/node/primitives/src/lib.rs
andresilva: Yeah, initially I set it to "rprt" but thought this was cooler :P

frame/grandpa/src/lib.rs
andresilva: I take that's way too much :relaxed: We should definitely come up with some sensitive value before merging (not necessarily too accurate for now).

bin/node/runtime/src/lib.rs
andresilva: The session module only deals with opaque key types.

rphmeier: That is true, but the implementation of `KeyOwnerProofSystem` is written as

```rust
impl<T: Trait, D: AsRef<[u8]>> frame_support::traits::KeyOwnerProofSystem<(KeyTypeId, D)>
	for Module<T>
```

I did it this way so that modules can avoid using `Vec`ly-typed keys with the `KeyOwnerProofSystem`, because the underlying bytes are the same. It's more understandable and more efficient to avoid `Vec`.

primitives/finality-grandpa/src/lib.rs
```diff
@@ -157,6 +164,241 @@ impl<N: Codec> ConsensusLog<N> {
...
+		match self {
+			Equivocation::Prevote(ref equivocation) => &equivocation.identity,
+			Equivocation::Precommit(ref equivocation) => &equivocation.identity,
+		}
+	}
+}
+
+/// Verifies the equivocation proof by making sure that both votes target
+/// different blocks and that its signatures are valid.
+pub fn check_equivocation_proof<H, N>(report: &EquivocationProof<H, N>) -> Result<(), ()>
```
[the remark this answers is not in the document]
rphmeier: I prefer it the way it is, because `self.check()` just feels a bit weird to me organizationally.

primitives/session/src/lib.rs
```diff
@@ -44,6 +47,33 @@ sp_api::decl_runtime_apis! {
 		/// Returns the list of public raw public keys + key type.
 		fn decode_session_keys(encoded: Vec<u8>) -> Option<Vec<(Vec<u8>, KeyTypeId)>>;
 	}
+
+	/// Historical session membership runtime api.
+	pub trait SessionMembership {
```
rphmeier: Maybe we can just hide this in a GRANDPA runtime API to avoid leaking runtime-internal details about the proof. After all, I don't think node-side code will inspect this proof deeply.

frame/session/src/historical.rs
```diff
@@ -253,26 +254,16 @@ impl<T: Trait> ProvingTrie<T> {
...
 impl<T: Trait, D: AsRef<[u8]>> frame_support::traits::KeyOwnerProofSystem<(KeyTypeId, D)>
 	for Module<T>
 {
-	type Proof = Proof;
+	type Proof = MembershipProof;
 	type IdentificationTuple = IdentificationTuple<T>;
 
+	// The session index of a given membership proof and the validator count for
+	// that session.
+	type ExtraData = (SessionIndex, ValidatorCount);
```
rphmeier: This approach seems difficult to reconcile with the trait system. The property that users of `KeyOwnerProofSystem` need would be something like `Proof::ExtraData: Has<SessionIndex> + Has<ValidatorCount>` which can only be expressed via `frunk`'s `HList` ATM. Either that or you require `ExtraData = (SessionIndex, ValidatorCount)` which feels fragile.

The way it was done for parachain-equivocations in Polkadot was to add a trait and then you can just ensure `Proof: GetSessionIndex + GetValidatorCount`

bin/node/runtime/src/lib.rs
rphmeier: https://github.com/paritytech/polkadot/pull/951

frame/grandpa/src/lib.rs
andresilva: This is what I did in the initial implementation, the problem is that it requires always having a `KeyOwnerProofSystem` in order to instantiate GRANDPA (e.g. node-template doesn't use any of this), so I refactored all that into the `HandleEquivocation` trait that we can implement for `()` in order to equivocation reporting. We can't implement `KeyOwnerProofSystem` for `()` we'd need something like struct `NoopKeyOwnerProofSystem<Key, Proof, Id>(PhantomData<(Key, Proof, Id)>)`.

bin/node/primitives/src/lib.rs
kianenigma: "fish" is cooler 🕶

bin/node/executor/tests/submit_transaction.rs
andresilva: We both did actually :D

frame/grandpa/src/lib.rs
```diff
@@ -359,6 +452,22 @@ impl<T: Trait> Module<T> {
...
 		}
 	}
+
+	/// Submits an extrinsic to report an equivocation. This method will sign an
+	/// extrinsic with a call to `report_equivocation` with any reporting keys
+	/// available in the keystore and will push the transaction to the pool.
+	/// Only useful in an offchain context.
+	pub fn submit_report_equivocation_extrinsic(
+		equivocation_proof: EquivocationProof<T::Hash, T::BlockNumber>,
+		key_owner_proof: <T::HandleEquivocation as equivocation::HandleEquivocation<T>>::KeyOwnerProof,
```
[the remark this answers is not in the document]
andresilva: We would have to parameterize `EquivocationProof` on `K` (`K = <T::HandleEquivocation as equivocation::HandleEquivocation<T>>::KeyOwnerProof`), which would make it impossible to define the runtime API since we don't have access to that type there (https://github.com/paritytech/substrate/pull/3868/files#diff-233fff7c89a6b4ca96ca4ee31c62088cR483).

frame/session/src/historical.rs
andresilva: Implemented this now and reverted the changes to the historical module.

bin/node/runtime/src/lib.rs
andresilva: Updated to do the same thing.

frame/grandpa/src/lib.rs
```diff
@@ -183,10 +205,81 @@ decl_module! {
...
+			let reporter_id = ensure_signed(origin)?;
+
+			// FIXME: this is a hack needed because the typed argument version above fails to
+			// compile (due to missing codec implementation)
+			let key_owner_proof = Decode::decode(&mut &key_owner_proof[..])
+				.map_err(|_| "Key owner proof decoding failed.")?;
+
+			// validate the membership proof and extract session index and
+			// validator set count of the session that we're proving membership of
+			let (offender, session_index, validator_set_count) =
```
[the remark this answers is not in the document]
andresilva: This is done by checking against the `set_id -> session_id` mapping in the module's storage.

primitives/session/src/lib.rs
andresilva: Done.

andresilva:
> the weight of the call to report_equivocation may be set to 0 with taking further steps to avoid DoS.

Moving this to a `SignedExtension` just feels like moving the problem somewhere else, i.e. instead of DoSing the runtime you DoS the transaction pool. IMO there is no getting around finding a proper weight for the method and users paying the fees in order to report.

montekki:
> Moving this to a `SignedExtension` just feels like moving the problem somewhere else, i.e. instead of DoSing the runtime you DoS the transaction pool. IMO there is no getting around finding a proper weight for the method and users paying the fees in order to report.

Could you describe a possible attack vector? The way I understand the inner workings of the tx pool is following:

1) all TXs enter the pool with a public api `submit_at` that among other things calls `verify` on the transactions.
2) In any case for each single TX we will end up in `verify_one` where `validate_transaction` from the `frame/executive` will be called.
3) `validate_transaction` among other things will call `SignedExtension` validation logic on this TX.
4) If that fails the transaction is not let into the pool.

Now, I am wondering how the DoS vector would look like here. For instance Eve could send a bunch of transactions that do not pass the `SignedExtension` validation phase. By "a bunch" I mean enough to make their discarding a) computaionally heavy b) bandwidth-heavy. In that case we have a problem with all TXs validated in the same fashion (not only this one case) unless we take some further steps, like banning Eve for a bad reputation or something.

A side note: the discussion on this validation pipeline regularly pops up so we probably should document it somewhere in detail, discussing all invariants.

rphmeier:
I guess the DoS is that you can have an unlimited number of accounts submit the same (valid) report. But it should be easier to implement restrictions on the number of transactions in the pool so that most, if not all, duplicates are rejected. Or, in practice a block author will include this TX as a local transaction.

montekki:
> I don't think we want duplicate or invalid reports successfully going into, but being rejected by the runtime.

If this is ok: say two honest fishermen detected the same double vote, you would then need to somehow pick one and only that one gets a reward. That might not be very motivating for the fishermen?

primitives/finality-grandpa/src/lib.rs
```diff
@@ -157,6 +164,241 @@ impl<N: Codec> ConsensusLog<N> {
...
+				&$message($equivocation.second.0.clone()),
+				&$equivocation.identity,
+				&$equivocation.second.1,
+				$equivocation.round_number,
+				report.set_id,
+			)?;
+
+			return Ok(());
+		};
+	}
```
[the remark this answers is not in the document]
andresilva: I am unsure of what to do here, the reason I'm using a macro is because the two types we're checking even though have the same fields are completely separate types. I could add some methods to `EquivocationProof` to extract most of the data and then check it without the macro but I think this would add too much boilerplate (e.g. `fn first_hash(), first_number(), first_message(), ...`) and on each of these implementations we'd have the same code just matching for a `Prevote` or a `Precommit`. Since checking the proof is the only thing we're doing with it I think it is not worth to add that boilerplate?

I have removed the clones though :)

primitives/finality-grandpa/src/lib.rs
```diff
@@ -157,6 +164,241 @@ impl<N: Codec> ConsensusLog<N> {
...
+	) -> Self {
+		Equivocation::Precommit(equivocation)
+	}
+}
+
+impl<H, N> Equivocation<H, N> {
+	/// Returns the authority id of the equivocator.
+	pub fn offender(&self) -> &AuthorityId {
+		match self {
+			Equivocation::Prevote(ref equivocation) => &equivocation.identity,
```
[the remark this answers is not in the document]
andresilva: That doesn't work because the inner equivocations have different types (same problem as above with the macro thing).

client/finality-grandpa/src/environment.rs
andresilva: Done.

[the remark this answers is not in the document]
andresilva: Done.

frame/grandpa/src/lib.rs
andresilva: Moved validation to a `SignedExtension` so the weight is now 0.

andresilva:
I have addressed all review comments and also addressed some things discussed on chat:
- Moved `KeyOwnerProofSystem` out of `HandleEquivocation` and into the main module trait;
- Moved equivocation report validation to a `SignedExtension`. The weight of the call is now 0 so the reporting account doesn't need to be funded;
- Moved key ownership proof runtime API method to the grandpa runtime API and the proof itself is exposed as an `OpaqueKeyMembershipProof` (at the runtime API boundary we don't know the exact type of the proof so we need to deal with an encoded version).

frame/grandpa/src/lib.rs
andresilva: No longer an issue since I moved `KeyOwnerProof` to the module trait and there's no extra indirection now.

bin/node/runtime/src/lib.rs
```diff
@@ -569,6 +574,23 @@ impl pallet_authority_discovery::Trait for Runtime {}
...
+		<Self::KeyOwnerProofSystem as KeyOwnerProofSystem<(KeyTypeId, Vec<u8>)>>::Proof;
+
+	type KeyOwnerIdentification = <Self::KeyOwnerProofSystem as KeyOwnerProofSystem<(
+		KeyTypeId,
+		Vec<u8>,
+	)>>::IdentificationTuple;
+
+	type HandleEquivocation = pallet_grandpa::EquivocationHandler<
+		Self::KeyOwnerIdentification,
+		TransactionSubmitterOf<node_primitives::report::ReporterId>,
```
rphmeier, line 591: modularity here is really good 👍

bin/node/runtime/src/lib.rs
```diff
@@ -786,6 +810,31 @@ impl_runtime_apis! {
...
+			let key_owner_proof = key_owner_proof.decode()?;
+
+			Grandpa::submit_report_equivocation_extrinsic(
+				equivocation_proof,
+				key_owner_proof,
+			)
+		}
+
+		fn generate_key_ownership_proof(
+			authority_id: fg_primitives::AuthorityId,
```
rphmeier, line 830: We probably want to provide `set_id` as a parameter here, since @drahnr will be changing `Historical::prove` to accept `SessionIndex`. I believe we have a `SetId -> SessionIndex` lookup somewhere.

andresilva: Yeah we have a map `SetId -> Session` in the grandpa pallet which maps to the latest session in the set. I'll update this API (and keep the parameter unused for now).

bin/node/primitives/src/lib.rs
```diff
@@ -62,3 +62,34 @@ pub type Header = generic::Header<BlockNumber, BlakeTwo256>;
...
+/// GRANDPA. Any rewards for misbehavior reporting will be paid out to this
+/// account.
+pub mod report {
+	use super::{Signature, Verify};
+	use frame_system::offchain::AppCrypto;
+	use sp_core::crypto::KeyTypeId;
+
+	/// Key type for the reporting module. Used for reporting BABE and GRANDPA
+	/// equivocations.
+	pub const KEY_TYPE: KeyTypeId = KeyTypeId(*b"fish");
```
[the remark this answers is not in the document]
andresilva: Agreed will add to follow-up PR.

primitives/session/src/lib.rs
```diff
@@ -46,6 +49,32 @@ sp_api::decl_runtime_apis! {
...
+	pub session: SessionIndex,
+	/// Trie nodes of a merkle proof of session membership.
+	pub trie_nodes: Vec<Vec<u8>>,
+	/// The validator count of the session on which the specific key is a member.
+	pub validator_count: ValidatorCount,
+}
+
+impl MembershipProof {
+	/// Returns a session this proof was generated for.
+	pub fn session(&self) -> SessionIndex {
```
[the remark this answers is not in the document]
andresilva: Yeah they don't.

primitives/finality-grandpa/src/lib.rs
```diff
@@ -157,6 +164,242 @@ impl<N: Codec> ConsensusLog<N> {
...
+	set_id: SetId,
+	buf: &mut Vec<u8>,
+) -> Result<(), ()>
+where
+	H: Encode,
+	N: Encode,
+{
+	localized_payload_with_buffer(round, set_id, message, buf);
+
+	#[cfg(not(feature = "std"))]
```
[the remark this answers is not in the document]
andresilva: native/wasm version, tbh I already wrote this specific function some time ago I no longer remember why this was needed. Will check again and change in follow-up if unnecessary.

frame/grandpa/src/lib.rs
```diff
@@ -351,14 +429,35 @@ impl<T: Trait> Module<T> {
...
+		// because we only update this `on_new_session` which isn't called
+		// for the genesis session.
+		SetIdSession::insert(0, 0);
+	}
+
+	/// Submits an extrinsic to report an equivocation. This method will sign an
+	/// extrinsic with a call to `report_equivocation` with any reporting keys
+	/// available in the keystore and will push the transaction to the pool.
+	/// Only useful in an offchain context.
+	pub fn submit_report_equivocation_extrinsic(
```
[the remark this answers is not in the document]
andresilva: It means that the transaction wasn't sent, either there are no signing accounts available locally for this or the given chain doesn't have equivocation reporting enabled at all. I can add this to the docs.

frame/grandpa/src/lib.rs
```diff
@@ -351,14 +429,35 @@ impl<T: Trait> Module<T> {
...
+		equivocation_proof: EquivocationProof<T::Hash, T::BlockNumber>,
+		key_owner_proof: T::KeyOwnerProof,
+	) -> Option<()> {
+		T::HandleEquivocation::submit_equivocation_report(equivocation_proof, key_owner_proof)
+			.ok()?;
+
+		Some(())
```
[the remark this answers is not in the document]
andresilva: 👍

frame/grandpa/src/lib.rs
```diff
@@ -351,14 +429,35 @@ impl<T: Trait> Module<T> {
...
 		if !authorities.is_empty() {
 			assert!(
 				Self::grandpa_authorities().is_empty(),
 				"Authorities are already initialized!"
 			);
 			Self::set_grandpa_authorities(authorities);
 		}
+
+		// NOTE: initialize first session of first set. this is necessary
+		// because we only update this `on_new_session` which isn't called
```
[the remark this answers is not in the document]
andresilva: Yeah initially the code here was only on genesis builder, I moved it here and didn't update the comment.

frame/grandpa/src/lib.rs
```diff
@@ -183,11 +223,49 @@ decl_module! {
 
 		fn deposit_event() = default;
 
-		/// Report some misbehavior.
+		/// Report voter equivocation/misbehavior. This method will verify the
+		/// equivocation proof and validate the given key ownership proof
+		/// against the extracted offender. If both are valid, the offence
+		/// will be reported.
+		///
+		/// Since the weight is 0 in order to avoid DoS pre-validation is implemented in a
```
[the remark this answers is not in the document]
andresilva: This is missing some commas, should read as:
"Since the weight is 0, in order to avoid DoS, pre-validation is implemented in a..."

frame/grandpa/src/equivocation.rs
```diff
@@ -0,0 +1,406 @@
...
+		write!(f, "ValidateEquivocationReport<T>")
+	}
+}
+
+/// Custom validity error used when validating equivocation reports.
+#[derive(Debug)]
+#[repr(u8)]
+pub enum ReportEquivocationValidityError {
+	/// The proof provided in the report is not valid.
+	InvalidEquivocationProof = 1,
```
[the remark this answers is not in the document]
andresilva: Where should they go then? Or do I just look for the next free `u8`?

frame/grandpa/src/equivocation.rs
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	fn validate(
+		&self,
+		_who: &Self::AccountId,
+		call: &Self::Call,
+		_info: &DispatchInfoOf<Self::Call>,
+		_len: usize,
+	) -> TransactionValidity {
+		let (equivocation_proof, key_owner_proof) = match call.is_sub_type() {
```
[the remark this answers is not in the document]
andresilva: Initially this was checked in the dispatchable and there was no signed extension. We want to avoid users being able to spam the chain with fake equivocation reports, one way to do it is to have a proper weight here and make sure that users pay for the transaction (this was what I did initially). But this requires that the reporting account has enough balance to pay these fees. So instead we decided to make the transaction free and validation happens in the signed extension, if the signed extension is not added then this doesn't work properly.

bin/node/cli/Cargo.toml
```diff
@@ -85,8 +85,9 @@ pallet-balances = { version = "2.0.0-dev", path = "../../../frame/balances" }
 pallet-transaction-payment = { version = "2.0.0-dev", path = "../../../frame/transaction-payment" }
 frame-support = { version = "2.0.0-dev", default-features = false, path = "../../../frame/support" }
 pallet-im-online = { version = "2.0.0-dev", default-features = false, path = "../../../frame/im-online" }
-pallet-authority-discovery = { version = "2.0.0-dev",  path = "../../../frame/authority-discovery" }
-pallet-staking = { version = "2.0.0-dev",  path = "../../../frame/staking" }
+pallet-authority-discovery = { version = "2.0.0-dev", path = "../../../frame/authority-discovery" }
+pallet-staking = { version = "2.0.0-dev", path = "../../../frame/staking" }
+pallet-grandpa = { version = "2.0.0-dev", path = "../../../frame/grandpa" }
```
AurevoirXavier, line 90: This should be in `dev-dependencies`

PR #3870, opened 2019-10-21
Explicitly declare decl_storage! getters as functions

rphmeier:
This is a backwards-compatible (test included) convenience improvement for `decl_storage!`.

Instead of declaring a field as 

```rust
MyField get(getter_name): ...;
```

we can declare as 

```rust
MyField get(fn getter_name): ...;
```

This makes it clear that a function is being declared, and also means that searches for `fn getter_name` don't come up empty when reading code and looking for definitions. I also ported over all our SRML modules to follow this pattern.

### DIFF
--- a/srml/support/procedural/src/storage/parse.rs
+++ b/srml/support/procedural/src/storage/parse.rs
@@ -113,11 +113,16 @@ struct DeclStorageLine {
 	pub default_value: ext::Opt<DeclStorageDefault>,
 }
 
+#[derive(Parse, ToTokens, Debug)]
+struct DeclStorageGetterBody {
+	fn_keyword: Option<Token![fn]>,
+	ident: Ident,
+}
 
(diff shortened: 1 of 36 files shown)


srml/support/procedural/src/storage/parse.rs
```diff
@@ -113,11 +113,16 @@ struct DeclStorageLine {
 	pub default_value: ext::Opt<DeclStorageDefault>,
 }
 
+#[derive(Parse, ToTokens, Debug)]
+struct DeclStorageGetterBody {
```
bkchr: This is not required. You can just add the `fn_token: Option<syn::token::Fn>` to the getter struct.

bkchr:
Could you please update the docs of the macro for this change?

rphmeier:
could you tell me which docs you mean? i did another pass, but I had already gotten through `srml-support/procedural`.

PR #3872, opened 2019-10-22
Remove sr-arithmetic/fuzzer from workspace to fix windows builds

expenses:
Currently, builds do not work on windows due to a `honggfuzz-rs` compile error:
```
error: honggfuzz-rs does not currently support Windows but works well under WSL (Windows Subsystem for Linux)
 --> C:\Users\Me\.cargo\registry\src\github.com-1ecc6299db9ec823\honggfuzz-0.5.45\build.rs:7:1
  |
7 | compile_error!("honggfuzz-rs does not currently support Windows but works well under WSL (Windows Subsystem for Linux)");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

We could solve this by either removing the fuzzer from the workspace (which seems much simpler to me) or by conditionally compiling honggfuzz-rs via `cfg(not(target_family="windows"))`.


bkchr:
I would go the conditional compilation way. Otherwise this will not be run.

expenses:
@bkchr I don't see the benefit in running the code and getting an error saying that windows is not supported by the fuzzer vs that same error occurring in compilation.

bkchr:
I would expect that conditional compilation does not makes the fuzzer error on windows, as it is not run/compiled.

expenses:
Sorry, to clarify, we could use an alternative main function like:
```rust
#[cfg(target_family="windows")]
fn main() {
    println!("Error: honggfuzz-rs does not currently support Windows but works well under WSL (Windows Subsystem for Linux)");
}
```
and mark the fuzzer main function with `#[cfg(not(target_family="windows"))]`, but I think it makes sense to let the compilation fail instead. Additionally, the fuzzers aren't run via `cargo run` but via `cargo hfuzz`, which won't compile on windows either.

PR #3873, opened 2019-10-22
Watch existing extrinsics RPC

NikVolf:
not sure about naming though @tomusdrw

### DIFF
--- a/core/rpc/api/src/author/mod.rs
+++ b/core/rpc/api/src/author/mod.rs
@@ -86,7 +86,7 @@ pub trait AuthorApi<Hash, BlockHash> {
 
 	/// Watch multiple extrinsics (own or from network)
 	#[pubsub(
-		subscription = "author_extrinsicTrack",
+		subscription = "author_extrinsicUpdate",
 		subscribe,
 		name = "author_watchExtrinsic"
 	)]
@@ -95,12 +95,4 @@ pub trait AuthorApi<Hash, BlockHash> {
 		subscriber: Subscriber<Status<Hash, BlockHash>>,
 	    hash: Hash,
 	);
-
-	/// Unsubscribe from multiple extrinsics watching.
-	#[pubsub(
-		subscription = "author_extrinsicTrack",
-		unsubscribe,
-		name = "author_unwatchExtrinsic"
-	)]
-	fn untrack_extrinsic(&self, metadata: Option<Self::Metadata>, id: SubscriptionId) -> Result<bool>;
 }
--- a/core/rpc/api/src/author/mod.rs
+++ b/core/rpc/api/src/author/mod.rs
@@ -67,7 +67,7 @@ pub trait AuthorApi<Hash, BlockHash> {
 		subscribe,
 		name = "author_submitAndWatchExtrinsic"
 	)]
-	fn watch_extrinsic(&self,
+	fn submit_and_watch_extrinsic(&self,
 		metadata: Self::Metadata,
 		subscriber: Subscriber<Status<Hash, BlockHash>>,
 		bytes: Bytes
@@ -90,7 +90,7 @@ pub trait AuthorApi<Hash, BlockHash> {
 		subscribe,
 		name = "author_watchExtrinsic"
 	)]
-	fn track_extrinsic(&self,
+	fn watch_extrinsic(&self,
     	metadata: Self::Metadata,
 		subscriber: Subscriber<Status<Hash, BlockHash>>,
 	    hash: Hash,
--- a/core/rpc/api/src/author/mod.rs
+++ b/core/rpc/api/src/author/mod.rs
@@ -83,4 +83,25 @@ pub trait AuthorApi<Hash, BlockHash> {
 		metadata: Option<Self::Metadata>,
 		id: SubscriptionId
 	) -> Result<bool>;
+
+	/// Watch multiple extrinsics (own or from network)
+	#[pubsub(
+		subscription = "author_multiExtrinsicUpdate",
+		subscribe,
+		name = "author_watchExtrinsics"
+	)]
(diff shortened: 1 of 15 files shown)


core/rpc/api/src/author/mod.rs
```diff
@@ -83,4 +83,24 @@ pub trait AuthorApi<Hash, BlockHash> {
 		metadata: Option<Self::Metadata>,
 		id: SubscriptionId
 	) -> Result<bool>;
+
+	/// Watch multiple extrinsics (own or from network)
+	#[pubsub(
+		subscription = "author_extrinsicTrack",
+		subscribe,
+		name = "author_trackExtrinsic"
```
tomusdrw:
```suggestion
		name = "author_watchExtrinsic"
```

core/rpc/api/src/author/mod.rs
```diff
@@ -83,4 +83,24 @@ pub trait AuthorApi<Hash, BlockHash> {
...
+    	metadata: Self::Metadata,
+		subscriber: Subscriber<Status<Hash, BlockHash>>,
+	    hash: Hash,
+	);
+
+	/// Unsubscribe from multiple extrinsics watching.
+	#[pubsub(
+		subscription = "author_extrinsicTrack",
+		unsubscribe,
+		name = "author_untrackExtrinsic"
```
tomusdrw:
```suggestion
		name = "author_unwatchExtrinsic"
```

Have you tried naming the unsubscribe method the same? I guess that will hopefuly work, but in both methods we will need to unsubscribe from both subscriptions in such case.

core/rpc/api/src/author/mod.rs
```diff
@@ -83,4 +83,24 @@ pub trait AuthorApi<Hash, BlockHash> {
...
+	)]
+	fn track_extrinsic(&self,
+    	metadata: Self::Metadata,
+		subscriber: Subscriber<Status<Hash, BlockHash>>,
+	    hash: Hash,
+	);
+
+	/// Unsubscribe from multiple extrinsics watching.
+	#[pubsub(
+		subscription = "author_extrinsicTrack",
```
tomusdrw: Yeah, it really is a pitty that we don't support two subscriptions with the same name. Opened an issue for this: https://github.com/paritytech/jsonrpc/issues/511

core/rpc/api/src/author/mod.rs
tomusdrw: I'd prefer this name, even if unsubscribe is going to be unintuitively named differently.

core/transaction-pool/graph/src/pool.rs
```diff
@@ -165,6 +165,13 @@ impl<B: ChainApi> Pool<B> {
 		)
 	}
 
+	pub fn watch(
```
tomusdrw: Missing docs?

core/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -182,6 +182,10 @@ impl<B: ChainApi> ValidatedPool<B> {
 		}
 	}
 
+	pub fn watch(&self, hash: ExHash<B>) -> Watcher<ExHash<B>, BlockHash<B>> {
```
tomusdrw: Missing docs

core/rpc/api/src/author/mod.rs
NikVolf: naming seems working, but how do you unsubscribe from both?

core/transaction-pool/graph/src/pool.rs
```diff
@@ -165,6 +165,17 @@ impl<B: ChainApi> Pool<B> {
...
 
+	/// Watch existing transaction
+	///
+	/// Get notified when some existing transaction is finished verifying or gets finalized
+	/// in a new block.
+	pub fn watch(
+		&self,
+		hash: ExHash<B>,
+	) -> Watcher<ExHash<B>, BlockHash<B>> {
+		self.validated_pool.watch(hash)
```
tomusdrw: Hmm, actually just realised that exposing this might open up a DoS vector, since we don't really validate if the hash is in the pool (which for one reason is good - you can subscribe to something before you saw it), but on the other hand you may wait for it forever.

What's your thoughts on this?

core/rpc/api/src/author/mod.rs
tomusdrw: I guess if we call the same `self.subscription.unsubscribe` it actually does unsubscribe from any of the two.

core/transaction-pool/graph/src/pool.rs
NikVolf: We can hard limit txpool number of watchers?

NikVolf:
Probably should wait a little until https://github.com/paritytech/jsonrpc/pull/512

core/rpc/api/src/author/mod.rs
```diff
@@ -83,4 +83,16 @@ pub trait AuthorApi<Hash, BlockHash> {
...
 		id: SubscriptionId
 	) -> Result<bool>;
+
+	/// Watch multiple extrinsics (own or from network)
+	#[pubsub(
+		subscription = "author_extrinsicUpdate",
+		subscribe,
+		name = "author_watchExtrinsic"
+	)]
+	fn track_extrinsic(&self,
```
tomusdrw: Maybe let's rename the function `watch_extrinsic` now and the other to `submit_and_watch` to avoid confusion.

NikVolf: Yeah I was just reluctant about this long name

NikVolf:
I am unsure if I get this runtime impl version updated @bkchr or ignore this ci error? Dependency was changed

PR #3874, opened 2019-10-22
Make tests work after the changes introduced in #3793

HCastano:


srml/bridge/Cargo.toml
```diff
@@ -17,6 +17,7 @@ system = { package = "srml-system", path = "../system", default-features = false
 trie = { package = "substrate-trie", path = "../../core/trie", default-features = false }
 
 [dev-dependencies]
+externalities = { package = "substrate-externalities", path = "../../core/externalities" }
```
bkchr: You should not require this import.

HCastano:
> lgtm other than the unnecessary dependency. but why didn't you push this directly to your branch?

I wanted to ping Jim about this without DM-ing him about the changes.

PR #3876, opened 2019-10-22
RPC to query transaction fee + weight + info 

kianenigma:
I am opening the PR to allow @jacogr to test this (regardless of the below dilemmas)

closes https://github.com/paritytech/substrate/issues/3933

My dilemmas are: 

#### Signature 

Preference is not to sign a tx to know how much it will cost. In reality, signature does not matter actually in the cost.  We can either accept an input to the rpc which is __exactly the same as `unchecked_extrinsic`__ and just put a dummy 0 signature in it. We don't want to check in the receiver side. This is simpler to implement but there is a waste of bytes in it we could also hardcode the length of the signature and just add it to the final weight. Currently we do this (assume signature length is always 64 or any other constant) and the RPC accepts `call, extra, origin` (which is practically everything in an extrinsic except for the signature)

#### Encoding

Regardless of the approach, by looking at the examples, I see that the RPC interface decodes the parameters into the given type. This is again easier but it does not make sense to me that we implicitly decode some bytes into for instance `Origin` in the rpc layer and then encode it again to get the length. Better would be to 
- Accept every parameter as raw bytes. 
- Keep their length
- instead of in the rpc later, decode them in the runtime layer (this might have some performance implication I suppose) as a way to make sure they are valid.
- just use the encoded length that we already knew.

--- 

- [x] Will inevitably need a polkadot PR


kianenigma:
@tomusdrw @bkchr your opinions will be highly appreciated.

srml/transaction-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+	TransactionPaymentApi as TransactionPaymentRuntimeApi, RuntimeDispatchInfo,
+};
+use self::gen_client::Client as TransactionPaymentClient;
+
+
+/// A wrapper around the request information to query a dispatchable info.
+#[derive(Serialize, Deserialize)]
+pub struct DispatchInfoRequest<AccountId, Extra, Call> {
+	/// Sender aka. origin of the dispatch.
+	pub origin: AccountId,
```
jacogr: For an extrinsic, this is actually `Address`, not `AccountId` (i.e. it can be sent from an `AccountIndex` as well - not 100% sure where the reverse lookup into `AccountId` is done when it hits the call)

jacogr:
As you suggested, would also prefer an interface where a fully-encoded extrinsic can be passed, like we do here https://github.com/paritytech/substrate/blob/master/core/rpc/api/src/author/mod.rs#L39

srml/transaction-payment/rpc/src/lib.rs
kianenigma: right -- for the same reason they can have various length actually.

kianenigma:
If I don't come to a strong inclination by the end of today, and no feedback, I will change the API to receive just the raw bytes of extrinsic.

kianenigma:
the approach of decoding things like `Call` and `Extra` at the RPC layer is wrong on a whole new level: I realized they all have to implement `serde` stuff, and I really don't want to.

node/runtime/src/lib.rs
```diff
@@ -653,7 +655,12 @@ impl_runtime_apis! {
         })
 		}
 
-		fn verify(payload: &Vec<u8>, signature: &EncodedSignature, authority_id: &EncodedAuthorityId) -> bool {
+		fn verify(
+			payload: &Vec<u8>,
+			signature:
```
tomusdrw: I think the type and var name would fit on a single line :)

node/runtime/src/lib.rs
```diff
@@ -701,6 +706,22 @@ impl_runtime_apis! {
 		}
 	}
 
+	impl transaction_payment_rpc_runtime_api::TransactionPaymentApi<
```
tomusdrw: Shall we have a macro to implement stuff like this? It feels pretty repetitive to implement all these forwarding traits. Like:
```rust
runtime_api!{
  impl TransactionPaymentApi<Block, AccountId, Balance, Call, SignedExtra> for Runtime {
    fn query_info_unsigned(AccountId, Call, SignedExtra) -> RuntimeDispatchInfo<Balance> delegate to TransactionPayment;
  }
}
```

tomusdrw: Only when I started writing it I realised that there is probably no way to make it any shorter, but leaving here for posterity, maybe someone will figure out how to do it better.

srml/transaction-payment/Cargo.toml
```diff
@@ -10,6 +10,7 @@ rstd = { package = "sr-std", path = "../../core/sr-std", default-features = fals
 sr-primitives = { path = "../../core/sr-primitives", default-features = false }
 support = { package = "srml-support", path = "../support", default-features = false }
 system = { package = "srml-system", path = "../system", default-features = false }
+transaction-payment-rpc-runtime-api = { package = "srml-transaction-payment-rpc-runtime-api", path = "./rpc/runtime-api", default-features = false }
```
tomusdrw: Shall you enable `std`?

srml/transaction-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! RPC interface for the t
```
tomusdrw: Line bre
aks
are
unnecessary here

srml/transaction-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+use jsonrpc_core::{Error, ErrorCode, Result};
+use jsonrpc_derive::rpc;
+use sr_primitives::{
+	generic::BlockId,
+	traits::{Block as BlockT, ProvideRuntimeApi},
+};
+use srml_transaction_payment_rpc_runtime_api::{
+	TransactionPaymentApi as TransactionPaymentRuntimeApi, RuntimeDispatchInfo,
+};
+use self::gen_client::Client as TransactionPaymentClient;
```
tomusdrw: You probably want to make it public

srml/transaction-payment/rpc/src/lib.rs
tomusdrw: Shouldn't this be coming from `Extrinsic` trait or at least be somehow bound to that trait?

srml/transaction-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+	pub origin: AccountId,
+	/// The call type.
+	pub call: Call,
+	/// Signed extensions.
+	pub extra: Extra,
+}
+
+#[rpc]
+pub trait TransactionPaymentApi<BlockHash, AccountId, Balance, Call, Extra> {
+	#[rpc(name = "transaction_payment_query_unsigned")]
```
tomusdrw: The convention is `<namespace>_<camelCasedMethodName>`.

srml/transaction-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/// A struct that implements the [`TransactionPaymentApi`].
+pub struct TransactionPayment<C, B> {
+	client: Arc<C>,
+	_marker: std::marker::PhantomData<B>,
+}
+
+impl<C, B> TransactionPayment<C, B> {
+	pub fn new(client: Arc<C>) -> Self {
```
tomusdrw: Missing docs

srml/transaction-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+	Call,
+	Extra,
+> TransactionPaymentApi<
+	<Block as BlockT>::Hash,
+	AccountId,
+	Balance,
+	Call,
+	Extra,
+> for TransactionPayment<C, Block> where
+	Block: BlockT,
```
tomusdrw: Perhaps move the short bounds to `impl<` part, to make it more concise?

srml/transaction-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+	C: Send + Sync + 'static + ProvideRuntimeApi + HeaderBackend<Block>,
+	C::Api: TransactionPaymentRuntimeApi<Block, AccountId, Balance, Call, Extra>,
+	AccountId: Codec,
+	Balance: Codec,
+	Call: Codec,
+	Extra: Codec,
+{
+	fn query_info_unsigned(
+		&self,
+		request: DispatchInfoRequest<AccountId, Extra, Call>
```
tomusdrw:
```suggestion
		request: DispatchInfoRequest<AccountId, Extra, Call>,
```

srml/transaction-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+	) -> Result<RuntimeDispatchInfo<Balance>> {
+		let api = self.client.runtime_api();
+		let best = self.client.info().best_hash;
```
tomusdrw: The convention should be to allow specifying the hash as (optional) last argument and only fall back to `best_hash` if it's not provided.

srml/transaction-payment/rpc/src/lib.rs
tomusdrw: I think it would be best to accept raw bytes here but it seems you've noticed that already. Then we can just decode to expected type or even pass it raw to the runtime (I prefer the former though, as you can return more meaningful error messages).

We could probably have a wrapper type that does this encoding automatically, i.e. implements `serde::{Ser, De}` for every type that is `scale::{Encode, Decode}`.

srml/transaction-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+		let at = BlockId::hash(best);
+
+		let DispatchInfoRequest { origin, extra, call } = request;
+		api.query_info_unsigned(&at, origin, call, extra).map_err(|e| Error {
+			code: ErrorCode::ServerError(1),
```
tomusdrw: Might be good to actually share this error code with other modules as well.

srml/transaction-payment/src/lib.rs
```diff
@@ -97,7 +98,39 @@ decl_module! {
...
+	{
+		// TODO: We can not decode this stuff and receive some raw bytes from the RPC, save the
+		// lengths, then decode and check that they are decodable (aka. correct) and just use the
+		// values.
+		let encoded_extra = extra.encode();
+		let encoded_call = call.encode();
+		let encoded_sender = sender.encode();
+
+		let dispatch_info = <T::Call as GetDispatchInfo>::get_dispatch_info(&call);
+		let len = encoded_call.len() + encoded_extra.len() + encoded_sender.len() + 64;
```
tomusdrw: What's the magic `64` for?

jacogr: This should go away with raw - since the signature is not provided, it assumed that is is the 64 bytes from `Signature` (but also needs the actual signing flags, i.e. `0x83` which is present in the raw extrinsic as well)

So this magic will disappear.

kianenigma:
@tomusdrw thanks for the extensive review! The PR was a bit novice and I was hoping you'd only see the description and not the (messy) code but too late for that 🗡 

So I conclusion would be that: 1- will change it to accepting raw bytes 2- I will expect and at some point decode an `unchecked_extrinsic` but never check the signature. I assume it can be zero filled.

tomusdrw:
@kianenigma Yes, I think using `UncheckedExtrinsic` makes most sense. There callers of this API need to be able to construct extrinsic anyway and not coupling the RPC with some specific seems like a good idea.

srml/transaction-payment/rpc/src/lib.rs
kianenigma: I am not sure if this really makes any sense here, I only followed examples after realizing that apparently all of them require the best block as the argument. 

How is this used by the runtime api? if the given block is not the best, the runtime call is made to the runtime state at the given particular block? this sounds quite expensive to do actually and I don't know if it makes sense for anyone to ask for previous fees?

tomusdrw: Yes, we call into the state of a specific block if the hash is passed and that state is not pruned yet.
It should be as expensive as calling into latest state block unless I'm missing something obvious here.

The reason for that is that you might have multiple calls that you want to do in parallel (or sequentially) and would be good to enforce that they are run against the same state.
So pretty much you fetch the latest hash yourself and then spawn a bunch of calls and pass that hash. Now you know that the estimates are coming from the same latest block and not from potentially different states. To be fair I can't yet imagine why this would be useful for estimating the tip, but would prefer to maintain consistency between RPCs here.

kianenigma: Yes my main concern was that what if someone (just because they can it is fun) decides to query the fee price of some old past. 

This brings an interesting thought: the runtime code itself is also part of the state right (`:code`)? given this, if there is a function `foo()` which is even removed from the runtime, I can still query it by given a block hash at which this function existed in the runtime.  

Thanks for the explanation. Will add the optional block hash.

tomusdrw: > if there is a function foo() which is even removed from the runtime, I can still query it by given a block hash at which this function existed in the runtime.

Yes, but that's also why we have `RuntimeApi`s, so that we can query what functions are available or change the implementation without affecting the API.

srml/transaction-payment/rpc/src/lib.rs
kianenigma: Totally changed. now we accept the unchecked_extrinsic, totally same as a normal one.

node/runtime/src/lib.rs
kianenigma: Yes I agree and I still have it on my list to do this very soon; but since this is rather a small api and having it in soon for kusama would make many validators + Jaco happy, I decided to do that first. Noted and hopefully very soon we'll further discuss how this boilerplate can be put into a macro.

srml/transaction-payment/Cargo.toml
kianenigma: enabled.

srml/transaction-payment/rpc/src/lib.rs
kianenigma: cat on the keyborad.

srml/transaction-payment/rpc/src/lib.rs
kianenigma: done

srml/transaction-payment/rpc/src/lib.rs
kianenigma: Done just to be sure what if namespace is `name_space` or let's stick to `namespace`? for now I named it `payment_queryInfo`

srml/transaction-payment/rpc/src/lib.rs
kianenigma: Can you give a new look and see if it fine? afair in most of substrate, once the line is too ling we just put all the bounds after the `where`. I also think this is more readable to just stick to one type of bound rather than some after `where` and some not.

srml/transaction-payment/rpc/src/lib.rs
kianenigma: changed.

srml/transaction-payment/src/lib.rs
kianenigma: no longer there.

srml/transaction-payment/src/lib.rs
```diff
@@ -97,7 +98,33 @@ decl_module! {
...
-impl<T: Trait> Module<T> {}
+impl<T: Trait> Module<T> {
+	/// Query the data that we know about the fee of a given `call`.
+	///
+	/// As this module is not and cannot be aware of the internals of a signed extension, it only
+	/// interprets them as some encoded value and takes their length into account.
+	///
+	/// All dispatchables must be annotated with weight and will have some fee info. This function
+	/// always returns.
+	// TODO: we can actually make it understand `ChargeTransactionPayment`, but would be some hassle
```
tomusdrw: Do you plan to address this TODO? If not maybe let's create an issue.

kianenigma: Yes I will create a potential issue but I am not sure if is even needed.. this enables the RPC to be tip-aware, while it might be waaaay too much hassle for no reason. The client itself is building the tx at the end of the day. It already knows the tip.

srml/transaction-payment/src/lib.rs
```diff
@@ -95,7 +96,33 @@ decl_module! {
...
```
kianenigma:
```suggestion
	// NOTE: we can actually make it understand `ChargeTransactionPayment`, but would be some hassle
```

Issue #3879, opened 2019-10-22
Treasury pot is both kept and spend when emptied.


thiolliere:
If after funds has been spend the pot goes below ED then the settle [here](https://github.com/paritytech/substrate/blob/master/srml/treasury/src/lib.rs#L314-L323) will fail and the pot thus will be kept.

So funds are kept and spend at the same time.

PR #3880, opened 2019-10-22
Fix treasury kept and spend when emptied

thiolliere:
Fix https://github.com/paritytech/substrate/issues/3879
Superseed https://github.com/paritytech/substrate/pull/3797

## Done:

* treasury account is created at genesis config.
* pot doesn't include existential deposit thus pot can be entirely spend with no issue.

## TODO:

* [ ] update polkadot: just includes Config in construct_runtime
* [x] fix some tests using runtime genesis config

## Alternatives:

* we could also keep old behavior with treasury being created when it happens that dilution is more than ED. but seem better that we create it once and for all.

## Notes:

* I still kept resolve_creating to ensure it still works on Kusama if there is no treasury account then it will just get created once on_dilution mint more than ED instead of never be created.
  Also in case someone forget to include Config in construct runtime (thus account is not created) then this account can be created later as well.

### DIFF
--- a/srml/treasury/src/lib.rs
+++ b/srml/treasury/src/lib.rs
@@ -640,4 +640,27 @@ mod tests {
 			assert_eq!(Treasury::pot(), 75); // Pot has finally changed
 		});
 	}
+
+	// Treasury account doesn't get deleted if amount approved to spend is all its free balance.
+	// i.e. pot should not include existential deposit needed for account survival.
+	#[test]
+	fn treasury_account_doesnt_get_deleted() {
+		new_test_ext().execute_with(|| {
+			Treasury::on_dilution(100, 100);
+			assert_eq!(Treasury::pot(), 100);
+			let treasury_balance = Balances::free_balance(&Treasury::account_id());
+
+			assert_ok!(Treasury::propose_spend(Origin::signed(0), treasury_balance, 3));
+			assert_ok!(Treasury::approve_proposal(Origin::ROOT, 0));
+
+			<Treasury as OnFinalize<u64>>::on_finalize(2);
+			assert_eq!(Treasury::pot(), 100); // Pot hasn't changed
+
+			assert_ok!(Treasury::propose_spend(Origin::signed(0), Treasury::pot(), 3));
+			assert_ok!(Treasury::approve_proposal(Origin::ROOT, 1));
+
+			<Treasury as OnFinalize<u64>>::on_finalize(4);
+			assert_eq!(Treasury::pot(), 0); // Pot is emptied
+		});
+	}
 }
--- a/srml/treasury/src/lib.rs
+++ b/srml/treasury/src/lib.rs
@@ -345,7 +345,7 @@ impl<T: Trait> Module<T> {
 
 impl<T: Trait> OnUnbalanced<NegativeImbalanceOf<T>> for Module<T> {
 	fn on_unbalanced(amount: NegativeImbalanceOf<T>) {
-		// Must resolve into existing but better be safe.
+		// Must resolve into existing but better to be safe.
 		let _ = T::Currency::resolve_creating(&Self::account_id(), amount);
 	}
 }
--- a/srml/treasury/src/lib.rs
+++ b/srml/treasury/src/lib.rs
@@ -237,7 +237,7 @@ decl_storage! {
 		build(|_config| {
 			// Create Treasury account
 			let _ = T::Currency::make_free_balance_be(
-				&MODULE_ID.into_account(),
+				&<Module<T>>::account_id(),
 				T::Currency::minimum_balance(),
 			);
 		});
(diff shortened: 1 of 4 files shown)


srml/treasury/src/lib.rs
```diff
@@ -233,6 +233,15 @@ decl_storage! {
 		/// Proposal indices that have been approved but not yet awarded.
 		Approvals get(fn approvals): Vec<ProposalIndex>;
 	}
+	add_extra_genesis {
+		build(|_config| {
+			// Create Treasury account
+			let _ = T::Currency::make_free_balance_be(
+				&MODULE_ID.into_account(),
```
kianenigma: `&Self::accountid()` doesn't work here right?

srml/treasury/src/lib.rs
```diff
@@ -325,21 +335,31 @@ impl<T: Trait> Module<T> {
...
 		T::Currency::free_balance(&Self::account_id())
+			// Must never be less than 0 but better be safe.
+			.saturating_sub(T::Currency::minimum_balance())
 	}
 }
 
 impl<T: Trait> OnUnbalanced<NegativeImbalanceOf<T>> for Module<T> {
 	fn on_unbalanced(amount: NegativeImbalanceOf<T>) {
-		T::Currency::resolve_creating(&Self::account_id(), amount);
+		// Must resolve into existing but better be safe.
```
kianenigma:
```suggestion
		// Must resolve into existing but better to be safe.
```

srml/treasury/src/lib.rs
```diff
@@ -325,21 +335,31 @@ impl<T: Trait> Module<T> {
...
+		let _ = T::Currency::resolve_creating(&Self::account_id(), amount);
```
kianenigma: Now that we have better runtime logging I do get tempted to do something like `resolve_exisiting().unwrap_or_else(|_| { log("CODE RED, SOS"); resolve_creating() )`. 

But I assume it depends on the proof. If it is logically impossible to cause account creation then this is also good.

srml/treasury/src/lib.rs
```diff
@@ -311,6 +320,7 @@ impl<T: Trait> Module<T> {
 			Self::deposit_event(RawEvent::Burnt(burn))
 		}
 
+		// Budget doesn't include existential deposit thus the account must stay alive.
```
kianenigma: So we create the treasury account once with ED and with `KeepAlive` we argue that it will never go beyond ED; qed

right?

srml/treasury/src/lib.rs
thiolliere: no actually inside build function self is GenesisConfig

srml/treasury/src/lib.rs
thiolliere: yes the buget to spend is balance - ED so even if the whole budget is spend then this doesn't go below ED. But indeed I'll rephrase a bit for clarification

srml/treasury/src/lib.rs
xlc: `Module<T>::account_id()` should work

thiolliere: oh yes indeed bette. done.

srml/treasury/src/lib.rs
thiolliere: inline doc improved

srml/treasury/src/lib.rs
thiolliere: interesting indeed a lot of such safe operation could be logged 🤔

thiolliere:
I would very suspect that treasury does exist already in kusama, but anyway it is fine. and I added both tests.

another solution would be to lazily create the account if it happens that it doesn't exist but whatever is OK

thiolliere:
test are failing at least because of some CI issue, let's retry them

srml/treasury/src/lib.rs
```diff
@@ -600,17 +626,76 @@ mod tests {
...
+			assert_ok!(Treasury::approve_proposal(Origin::ROOT, 0));
+
+			<Treasury as OnFinalize<u64>>::on_finalize(2);
+			assert_eq!(Treasury::pot(), 100); // Pot hasn't changed
+
+			assert_ok!(Treasury::propose_spend(Origin::signed(0), Treasury::pot(), 3));
+			assert_ok!(Treasury::approve_proposal(Origin::ROOT, 1));
+
+			<Treasury as OnFinalize<u64>>::on_finalize(4);
+			assert_eq!(Treasury::pot(), 0); // Pot is emptied
```
kianenigma:
```suggestion
			assert_eq!(Treasury::pot(), 0); // Pot is emptied
			assert_eq!(Balances::free_balance(&Treasury::account_id()), 1); // but the account is still there 
```

PR #3882, opened 2019-10-22
Error message when changing pruning mode

arkpar:
See #3835

### DIFF
--- a/core/state-db/src/lib.rs
+++ b/core/state-db/src/lib.rs
@@ -41,6 +41,11 @@ use noncanonical::NonCanonicalOverlay;
 use pruning::RefWindow;
 use log::trace;
 
+const PRUNING_MODE: &[u8] = b"mode";
+const PRUNING_MODE_ARCHIVE: &[u8] = b"archive";
+const PRUNING_MODE_ARCHIVE_CANON: &[u8] = b"archive_canonical";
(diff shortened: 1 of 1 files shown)


core/state-db/src/lib.rs
```diff
@@ -200,18 +221,41 @@ impl<BlockHash: Hash, Key: Hash> StateDbSync<BlockHash, Key> {
...
+			Some(v) => Err(Error::InvalidPruningMode(String::from_utf8_lossy(v).into())),
+			None => Ok(()),
+		}
+	}
+
 	pub fn insert_block<E: fmt::Debug>(&mut self, hash: &BlockHash, number: u64, parent_hash: &BlockHash, mut changeset: ChangeSet<Key>) -> Result<CommitSet<Key>, Error<E>> {
+		let mut meta = ChangeSet::default();
+		if number == 0 {
+			// Save pruning mode when writing first block.
+			meta.inserted.push((to_meta_key(PRUNING_MODE, &()), self.mode.id().into()));
```
andresilva: Can't we do this above on `new`? When checking meta if it comes back as `None` then we update it.

arkpar: `new` and `StateDB` in general does not write to the database directly.  It returns a `CommitSet` that is applied in the upper level code. Returning a `CommitSet` from `new` seems less elegant and would also require some restructuring of the calling code.

PR #3887, opened 2019-10-22
Split the telemetry net status report in two

### LINKED ISSUE #3541: Send the telemetry network state separately
Currently, every 5 seconds, a node sends its state to the telemetry. This includes a huge structure of network state. This unfortunately floods the telemetry server.

We should instead send this network state more infrequently, for example every 30 seconds, and separately from the rest of the status.

cc @maciejhirsz

tomaka:
Fix #3541 

Two changes:

- The `network_state()` method on the `AbstractService` trait now takes a `Duration` for the interval at which the state should be reported. This is IMO a more sensible design, as it allows the informant to decide how often it should be displayed.

- Splits the `"system.interval"` telemetry report in two: the `"system.interval"` report no longer contains the `"network_state"` field. Instead, there's a new `"system.network_state"` message that contains a single field named `"state"`. This new message is sent only every 30 seconds.

The telemetry backend should obviously be updated accordingly. cc @maciejhirsz

### DIFF
--- a/core/service/src/lib.rs
+++ b/core/service/src/lib.rs
@@ -74,9 +74,10 @@ pub struct NewService<TBl, TCl, TSc, TNetStatus, TNet, TTxPool, TOc> {
 	select_chain: Option<TSc>,
 	network: Arc<TNet>,
 	/// Sinks to propagate network status updates.
-	network_status_sinks: Arc<Mutex<Vec<mpsc::UnboundedSender<(
+	/// For each element, every time the `Interval` fires we push an element on the sender.
+	network_status_sinks: Arc<Mutex<Vec<(tokio_timer::Interval, mpsc::UnboundedSender<(
 		TNetStatus, NetworkState
-	)>>>>,
+	)>)>>>,
 	transaction_pool: Arc<TTxPool>,
 	/// A future that resolves when the service has exited, this is useful to
 	/// make sure any internally spawned futures stop when the service does.
@@ -296,9 +297,12 @@ macro_rules! new_impl {
 		let client_ = client.clone();
 		let mut sys = System::new();
 		let self_pid = get_current_pid().ok();
-		let (netstat_tx, netstat_rx) = mpsc::unbounded::<(NetworkStatus<_>, NetworkState)>();
-		network_status_sinks.lock().push(netstat_tx);
-		let tel_task = netstat_rx.for_each(move |(net_status, network_state)| {
+		let (state_tx, state_rx) = mpsc::unbounded::<(NetworkStatus<_>, NetworkState)>();
+		network_status_sinks.lock().push((
+			tokio_timer::Interval::new_interval(std::time::Duration::from_millis(5000)),
+			state_tx
+		));
+		let tel_task = state_rx.for_each(move |(net_status, _)| {
 			let info = client_.info();
 			let best_number = info.chain.best_number.saturated_into::<u64>();
 			let best_hash = info.chain.best_hash;
@@ -325,7 +329,6 @@ macro_rules! new_impl {
 			telemetry!(
 				SUBSTRATE_INFO;
 				"system.interval";
-				"network_state" => network_state,
 				"peers" => num_peers,
 				"height" => best_number,
 				"best" => ?best_hash,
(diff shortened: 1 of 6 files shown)


core/service/src/lib.rs
```diff
@@ -746,19 +761,27 @@ fn build_network_future<
...
+						best_seen_block: network.best_seen_block(),
+						num_sync_peers: network.num_sync_peers(),
+						num_connected_peers: network.num_connected_peers(),
+						num_active_peers: network.num_active_peers(),
+						average_download_per_sec: network.average_download_per_sec(),
+						average_upload_per_sec: network.average_upload_per_sec(),
+					};
+					let state = network.network_state();
+
+					if !status_sinks[n].1.unbounded_send((status.clone(), state.clone())).is_ok() {
```
bkchr:
```suggestion
					if status_sinks[n].1.unbounded_send((status.clone(), state.clone())).is_err() {
```

core/service/src/lib.rs
```diff
@@ -746,19 +761,27 @@ fn build_network_future<
...
-				average_upload_per_sec: network.average_upload_per_sec(),
-			};
-			let state = network.network_state();
-
-			status_sinks.lock().retain(|sink| sink.unbounded_send((status.clone(), state.clone())).is_ok());
+		{
+			let mut status_sinks = status_sinks.lock();
+			for n in (0..status_sinks.len()).rev() {
+				while let Ok(Async::Ready(_)) = status_sinks[n].0.poll() {
+					let status = NetworkStatus {
```
bkchr: `state` and `status` should be build outside of the for loop, only once.

core/service/src/lib.rs
```diff
@@ -746,19 +761,27 @@ fn build_network_future<
...
-				average_download_per_sec: network.average_download_per_sec(),
-				average_upload_per_sec: network.average_upload_per_sec(),
-			};
-			let state = network.network_state();
-
-			status_sinks.lock().retain(|sink| sink.unbounded_send((status.clone(), state.clone())).is_ok());
+		{
+			let mut status_sinks = status_sinks.lock();
+			for n in (0..status_sinks.len()).rev() {
+				while let Ok(Async::Ready(_)) = status_sinks[n].0.poll() {
```
bkchr: This should use `FuturesUnordered`.

core/service/src/lib.rs
tomaka: I went for creating the state inside the loop, because that loop is called every single time something happens on the network (i.e. very very often). In practice, the inside of the loop will only be reached three times every 5 seconds.

core/service/src/lib.rs
tomaka: I'm not sure how you see that. The intervals are streams and not futures, and I would have to maintain a parallel list of senders.

core/service/src/lib.rs
bkchr: You still call clone below.

core/service/src/lib.rs
bkchr: Yeah, just implement your own future. 
Above you have written that this outer future is called "very very often" and every time you will iterate the list of status sinks and ask "hey is your timer finished?".
`FuturesUnordered` is exactly for such use cases as it will not iterate the full list every time we poll this future.

```
struct StatusSink {
    sender: Option<Sender>,
    timer: Option<Timer>,
}

impl StatusSink {
    fn send_status(&mut self, status) -> bool {
        self.sender.as_mut().unwrap().unbounded_send(status).is_err()
    }
}

impl Future for StatusSink {
    type Item = Self;
    type Error = ();

    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
        match self.timer.as_mut().unwrap().poll()? {
            Async::Ready(_) => StatusSink { sender: self.sender.take(), timer: self.timer.take() },
            Async::NotReady => NotReady,
        }
    }
}
```

tomaka:
I did it in a cleaner way.

core/service/src/status_sinks.rs
```diff
@@ -0,0 +1,149 @@
...
+	///
+	/// Panics if not called within the context of a task.
+	pub fn poll(&mut self, mut status_grab: impl FnMut() -> T) {
+		loop {
+			match self.entries.poll() {
+				Ok(Async::Ready(Some((sender, interval)))) => {
+					let status = status_grab();
+					if sender.unbounded_send(status).is_ok() {
+						self.entries.push(YieldAfter {
+							// Note that since there's a small delay between the moment a task is
```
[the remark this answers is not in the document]
tomaka: Ideally the `Delay` struct would return the moment when it has been triggered, and we could do `triggered_moment + interval` instead of `Instant::now() + interval`.
`tokio-timer` doesn't do that, but `futures_timer` does, I think.

PR #3888, opened 2019-10-22
Support disabling the addition of the default bootnode

bkchr:


bkchr:
I don't wanted to break the current default behavior.

PR #3891, opened 2019-10-22
Throw an error if a bootnode is registered with two different peer ids

bkchr:


core/network/src/error.rs
```diff
@@ -28,13 +28,16 @@ pub enum Error {
 	Io(std::io::Error),
 	/// Client error
 	Client(client::error::Error),
+	/// Some other error occurred.
+	Other(String),
```
tomaka: Why not `DuplicateBootnodeAddress` instead of `Other`?

bkchr: Renamed

core/network/src/error.rs
```diff
@@ -28,13 +28,16 @@ pub enum Error {
 	Io(std::io::Error),
 	/// Client error
 	Client(client::error::Error),
+	/// The same bootnode (based on address) is given with two different peer ids.
+	DuplicateBootnode(String),
```
tomaka:
```suggestion
	DuplicateBootnode {
		address: Multiaddr,
		first_id: PeerId,
		second_id: PeerId,
	}
```

bkchr: Done

PR #3894, opened 2019-10-23
Bump polkadot-master

expenses:
Hello, I need to be able to make a polkadot PR for #3857, but polkadot won't compile with the current `master` branch of substrate so first I need `polkadot-master` to get updated then https://github.com/paritytech/polkadot/pull/490 to get merged. Having to do things this way is a bit annoying. Perhaps if polkadot used a submodule of substrate instead of relying on this `polkadot-master` branch things would be easier.


expenses:
It turns out that I do not need to make said PR, however it would be good to keep `polkadot-master` updated.

bkchr:
We update `polkadot-master`, when we prepare a pr for Polkadot. This is not the best workflow, but the best we can do currently. Also we just update the branch by pushing to it, so I will close this pr for now.

thiolliere:
see the polkadot part in the guidelines https://github.com/paritytech/substrate/blob/master/CONTRIBUTING.adoc

PR #3898, opened 2019-10-23
Fix phragmen election to wasm 

kianenigma:
I seem to have forgotten to include `rstd::prelude` 
- also include it in `substrate-node`
- also remove a very ambiguous type def (we basically had `type UncheckedExtrinsic` defined twice). For the future I prefer and recommend moving all of the `type Extra`, `Type Extrinsic` etc. form node-runtime to node-primitives).

### DIFF
--- a/node/cli/src/chain_spec.rs
+++ b/node/cli/src/chain_spec.rs
@@ -235,11 +235,11 @@ pub fn testnet_genesis(
 			members: vec![],
 			phantom: Default::default(),
 		}),
-		elections: Some(ElectionsConfig {
-			members: vec![],
-			presentation_duration: 1 * DAYS,
+		elections_phragmen: Some(ElectionsConfig {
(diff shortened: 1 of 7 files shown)


kianenigma:
The module also panics if temr-duration == 0 which I think will be the case since Kusama's genesis does not include this module at all and the default will be used. Fixed here

node/cli/src/chain_spec.rs
```diff
@@ -235,11 +235,11 @@ pub fn testnet_genesis(
 			members: vec![],
 			phantom: Default::default(),
 		}),
-		elections: Some(ElectionsConfig {
-			members: vec![],
-			presentation_duration: 1 * DAYS,
+		elections_phragmen: Some(ElectionsConfig {
+			members: endowed_accounts.iter().take(2).cloned().collect(),
```
kianenigma: I assume for testing it is okay to have some members and _real_ values.

PR #3899, opened 2019-10-23
node: spawn grandpa voter as essential task

andresilva:
This used to be correctly spawned as an essential task but was changed back in #3382 (probably when handling merge conflicts).

### DIFF
--- a/node/cli/src/service.rs
+++ b/node/cli/src/service.rs
@@ -109,7 +109,6 @@ macro_rules! new_full_start {
 /// concrete types instead.
 macro_rules! new_full {
 	($config:expr, $with_startup_data: expr) => {{
-		use futures::future::Future;
 		use futures::sync::mpsc;
 		use network::DhtEvent;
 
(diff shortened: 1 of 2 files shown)


node/cli/src/service.rs
```diff
@@ -109,6 +109,7 @@ macro_rules! new_full_start {
 /// concrete types instead.
 macro_rules! new_full {
 	($config:expr, $with_startup_data: expr) => {{
+		use futures::future::Future;
```
bkchr:
```suggestion
```

node/cli/src/service.rs
```diff
@@ -168,13 +169,15 @@ macro_rules! new_full {
 			};
 
 			let babe = babe::start_babe(babe_config)?;
+			let babe = babe.select(service.on_exit()).then(|_| Ok(()));
```
bkchr: We do this automatically in `spawn_task` and `spawn_essential_task`.
```suggestion
```

PR #3902, opened 2019-10-24
Switch sr-arithmetic benchmarking to criterion

expenses:
As mentioned by @bkchr and @kianenigma in #3799, we should try and use criterion for benchmarking where possible. I wanted to replace all old `#[bench]` benchmarks with criterion in this pr, but unfortunately the subkey benchmarks aren't exposed, the node executor benchmarks require an unwanted amount of code duplication and the phragmen benchmarks are complied via a macro. I'm not sure it's worth it to try and convert these benchmarks at the moment.

I did rename the core/primitives benchmark file from `benches.rs` to `bench.rs` to be more consistent though.

https://github.com/paritytech/substrate/blob/eda5fe5cf0f8e56d90f8cd913ffcb118362df559/subkey/src/vanity.rs#L163-L173
https://github.com/paritytech/substrate/blob/eda5fe5cf0f8e56d90f8cd913ffcb118362df559/node/executor/src/lib.rs#L1215-L1230
https://github.com/paritytech/substrate/blob/eda5fe5cf0f8e56d90f8cd913ffcb118362df559/core/phragmen/benches/phragmen.rs#L158-L174


core/sr-arithmetic/benches/bench.rs
```diff
@@ -0,0 +1,76 @@
+use criterion::{Criterion, criterion_group, criterion_main};
```
kianenigma: License header + how to run

core/sr-arithmetic/benches/bench.rs
```diff
@@ -0,0 +1,76 @@
...
+criterion_group!{
+    name = benches;
+    config = Criterion::default();
+    targets =
+        bench_addition_2_digit, bench_addition_4_digit,
+        bench_subtraction_2_digit, bench_subtraction_4_digit,
+        bench_multiplication_2_digit, bench_multiplication_4_digit,
+        bench_division_4_digit
+}
+criterion_main!(benches);
```
kianenigma: newline EOF

core/sr-arithmetic/benches/bench.rs
```diff
@@ -0,0 +1,76 @@
...
+use sr_arithmetic::biguint::{BigUint, Single};
+
+fn random_big_uint(size: usize) -> BigUint {
+	use rand::Rng;
+	let mut rng = rand::thread_rng();
+	let digits: Vec<_> = (0..size).map(|_| rng.gen_range(0, Single::max_value())).collect();
+	BigUint::from_limbs(&digits)
+}
+
+fn bench_addition_2_digit(c: &mut Criterion) {
```
kianenigma: I think now with criterion you can create a vector of limb sizes that you want `vec![2, 4, 6, 8, 10, ...]` and pass that and then configure the benchmark to generate numbers based on that... I have to look it up my memory of criterion is a bit foggy now but I am pretty sure criterion is much much more flexible and can probably circumvent this ugly pattern of `bench_size_x()`, `bench_size_y()` that I added initially.

bkchr:
The reason I want to see this change `#[bench]` is unstable and requires a nightly compiler and I'm not sure they will ever stabilize it. :)

core/sr-arithmetic/Cargo.toml
```diff
@@ -26,3 +26,7 @@ std = [
 	"serde",
 	"substrate-debug-derive/std",
 ]
+
+[[bench]]
+name = "bench"
+harness = false
```
bkchr:
```suggestion
harness = false

```

core/sr-arithmetic/benches/bench.rs
bkchr:
```suggestion
criterion_main!(benches);

```

core/sr-arithmetic/benches/bench.rs
expenses: You can now just run the benchmarks with `cargo bench` so I'm not sure we need run instructions.

core/sr-arithmetic/benches/bench.rs
expenses: Done :^)

core/sr-arithmetic/benches/bench.rs
```diff
@@ -0,0 +1,68 @@
+// Copyright 2019 Parity Technologies
```
gavofyork: wrong license - please ensure it's the same text as the file from which the code has been copied.

core/sr-arithmetic/benches/bench.rs
kianenigma: ahh nice I didn't even have to look it up :D thanks.

core/sr-arithmetic/benches/bench.rs
kianenigma: okay 👍

core/sr-arithmetic/benches/bench.rs
```diff
@@ -0,0 +1,70 @@
...
+fn bench_op<F: Fn(&BigUint, &BigUint)>(c: &mut Criterion, name: &str, op: F) {
+	let mut group = c.benchmark_group(name);
+
+	for size in [2, 4, 6, 8, 10].iter() {
+		group.throughput(Throughput::Elements(*size));
+		group.bench_with_input(BenchmarkId::from_parameter(size), size, |bencher, &size| {
+			let a = random_big_uint(size as usize);
+			let b = random_big_uint(size as usize);
+
+			bencher.iter(|| op(&a, &b));
```
kianenigma: hmm I just ran it and being amazed of why division is as fast as addition: we don't do division `p / q` if the size of `q` is equal to `p`. Need to be reworked. I'd leave it up to you regarding how to bake it out of criterion, but ideally, given `p` with `size` limbs, we should chose a secondary random number between `[2, size-1]` and build a number of that size and divide them.

kianenigma:
@expenses done and review needed? please update the labels as needed :)

expenses:
I don't think I can set labels, sorry.

PR #3905, opened 2019-10-24
Adds function `to_substrate_wasm_fn_return_value`

bkchr:
Instead of replicating this piece of code over and over again, just move
it to a function that does it.

### DIFF
--- a/core/executor/src/wasmi_execution.rs
+++ b/core/executor/src/wasmi_execution.rs
@@ -357,8 +357,9 @@ fn call_in_wasm_module(
 		},
 		|res, memory| {
 			if let Some(I64(r)) = res {
+				let r = u64::from_le(r as u64);
 				let offset = r as u32;
-				let length = (r as u64 >> 32) as usize;
+				let length = (r >> 32) as usize;
 				memory.get(offset, length).map_err(|_| Error::Runtime).map(Some)
 			} else {
 				Ok(None)
--- a/core/executor/src/wasmi_execution.rs
+++ b/core/executor/src/wasmi_execution.rs
@@ -357,9 +357,8 @@ fn call_in_wasm_module(
 		},
 		|res, memory| {
 			if let Some(I64(r)) = res {
-				let r = u64::from_le(r as u64);
 				let offset = r as u32;
-				let length = (r >> 32) as usize;
+				let length = (r as u64 >> 32) as usize;
 				memory.get(offset, length).map_err(|_| Error::Runtime).map(Some)
 			} else {
 				Ok(None)
--- a/core/primitives/src/lib.rs
+++ b/core/primitives/src/lib.rs
@@ -286,3 +286,25 @@ impl From<LogLevel> for log::Level {
 		}
 	}
 }
+
+/// Converts the given value into an `u64` by following the Substrate calling convention.
+///
+/// When Substrate calls into Wasm it expects a fixed signature for functions exported
+/// from the Wasm blob. The return value of this signature is always a `u64` in little endian.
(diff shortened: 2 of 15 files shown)


core/primitives/src/lib.rs
```diff
@@ -286,3 +286,25 @@ impl From<LogLevel> for log::Level {
 		}
 	}
 }
+
+/// Converts the given value into an `u64` by following the Substrate calling convention.
+///
+/// When Substrate calls into Wasm it expects a fixed signature for functions exported
+/// from the Wasm blob. The return value of this signature is always a `u64` in little endian.
```
pepyakin: I am not sure what is meant here since `endianess` can only apply to values stored in memory. When we are talking about just regular values the values are always stored in direct order.

core/primitives/src/lib.rs
```diff
@@ -286,3 +286,25 @@ impl From<LogLevel> for log::Level {
...
+/// This `u64` stores the pointer to the encoded return value and the length of this encoded value.
+/// The first `32bit` are reserved for the pointer and the last `32bit` for the length.
```
pepyakin: This on the contrary depends on the bit numbering scheme: the first would mean "the rightmost" bit in LSB and "the leftmost" in MSB. I'd go for "high/low" parts.

core/primitives/src/lib.rs
```diff
@@ -286,3 +286,25 @@ impl From<LogLevel> for log::Level {
...
+	let ptr = encoded.as_ptr() as u64;
+	let length = encoded.len() as u64;
+	let res = ptr | (length << 32);
+
+	// Leak the output vector to avoid it being freed.
+	// This is fine in a WASM context since the heap
+	// will be discarded after the call.
+	rstd::mem::forget(encoded);
+
+	res.to_le()
```
pepyakin: As mentioned above this doesn't make a lot of sense:

1. This is a function that is supposed to be used only in wasm thus conversion to little endian would do nothing
2. If this code were used on a big-endian machine, this would be incorrect since the bytes would be swapped.

core/primitives/src/lib.rs
```diff
@@ -286,3 +286,25 @@ impl From<LogLevel> for log::Level {
 		}
 	}
 }
+
+/// Converts the given value into an `u64` by following the Substrate calling convention.
```
pepyakin: I think a better wording would be "encodes the given value into a buffer and returns the pointer and length encoded as a single u64" or something like this.

core/executor/src/wasmi_execution.rs
```diff
@@ -357,8 +357,9 @@ fn call_in_wasm_module(
 		},
 		|res, memory| {
 			if let Some(I64(r)) = res {
+				let r = u64::from_le(r as u64);
```
pepyakin: As discussed below, I don't think that this makes sense.

core/primitives/src/lib.rs
bkchr: Wasm is always little endian. So yes, the conversion is somewhat useless, but will be optimized away anyway. 

Now on the host machine the current code works, because we are running it on little endian machines as well, but would you run it on a big endian machine, it would fail.

This representation is just for communicating between the wasm runtime and the host. Think of networking, in networking you store data in big endian. You need a fixed endianess, to know how to decode the data on the other side.

core/executor/src/wasmi_execution.rs
bkchr: If you run this code on a big endian machine, it makes sense, because otherwise it would be incorrect.

PR #3908, opened 2019-10-24
Bring back `SubmitSignedTransaction` trait.

tomusdrw:
We now support a case where `AccountId != Public`. This introduces a small change to `CreateTransaction` trait API (namely the pubkey is passed along the account id).


node/runtime/src/lib.rs
```diff
@@ -424,7 +424,7 @@ impl sudo::Trait for Runtime {
 	type Proposal = Call;
 }
 
-type SubmitTransaction = TransactionSubmitter<ImOnlineId, Runtime, UncheckedExtrinsic>;
+type SubmitTransaction = TransactionSubmitter<ImOnlineId,  Runtime, UncheckedExtrinsic>;
```
bkchr:
```suggestion
type SubmitTransaction = TransactionSubmitter<ImOnlineId, Runtime, UncheckedExtrinsic>;
```

Issue #814, opened 2018-09-27
Grafana integration


gavofyork:
An adjunct to telemetry.

- Figure out Grafana JSON data source format.
- Create a grafana-data-source crate a bit like `substrate-telemetry` crate that can take key-value pairs (a bit like `telemetry!` macro) and serve/output in a format that grafana likes.
- Create awesome grafana visualisations.

expenses:
This can be closed now! :D

PR #3913, opened 2019-10-25
Grafana integration

expenses:
See #814. This PR adds a HTTP server that (in a currently very bare-bones way) conforms to the [grafana-json-datasource](https://github.com/simPod/grafana-json-datasource) API.

Screen Shot 2019-10-25 at 5 58 25 PM

This PR probably has a fair bit to go before it can be merged. Some notes:

* Currently, all metrics are stored as `f32`s, which is probably fine (although a bit awkward) as we're serializing into json anyway.
* The server is currently turned on by default, but it might be a good idea to put it behind a flag instead.
* I haven't yet implemented the [`table` query response type](https://github.com/simPod/grafana-json-datasource#query), but I'm not sure if this is absolutely needed.
* I also haven't bothered with the [annotations API](https://github.com/simPod/grafana-json-datasource#annotations), not sure about that one either.
* At the moment, the metrics grow indefinitely. This is definitely something that needs changing, but I'm not sure when a cleanup should occur. 
* Ideally, we want to switch to a small, embedded timeseries database.


core/grafana-data-source/src/lib.rs
```diff
@@ -0,0 +1,56 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```
kianenigma:
```suggestion
// Copyright 2019 Parity Technologies (UK) Ltd.
```

expenses:
Ok, I think I've gotten things to a good-enough place now. _Ideally_, we'd use a database like [Akumuli](https://github.com/akumuli/Akumuli) in the same way that we use RocksDB, but there are no rust bindings for that (I'm looking into it, no promises) and it's probably overkill.

Grafana doesn't seem to complain that the tables mode doesn't return a table, so that's probably a non-issue.

I've added a future that removed metrics that are over a week old that is ran every day, which seems reasonable for the time being.

core/grafana-data-source/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+mod util;
+
+pub use server::run_server;
+pub use util::now_millis;
+
+type Metrics = HashMap<&'static str, Vec<(f32, i64)>>;
+
+lazy_static! {
+	/// The `RwLock` wrapping the metrics. Not intended to be used directly.
+    pub static ref METRICS: RwLock<Metrics> = RwLock::new(Metrics::new());
```
tomaka: Bad indentation.

Also, I'd suggest hiding this.
```suggestion
    #[doc(hidden)]
    pub static ref METRICS: RwLock<Metrics> = RwLock::new(Metrics::new());
```

core/grafana-data-source/test/src/main.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+fn main() {
+	let handle = spawn(|| {
+		let mut rng = rand::thread_rng();
+
+		loop {
+			let random = rng.gen_range(0.0, 1000.0);
+			record_metrics!(
+                "random data" => random,
```
tomaka: Bad indentation.

core/service/Cargo.toml
```diff
@@ -38,6 +38,8 @@ rpc = { package = "substrate-rpc", path = "../../core/rpc" }
 tel = { package = "substrate-telemetry", path = "../../core/telemetry" }
 offchain = { package = "substrate-offchain", path = "../../core/offchain" }
 parity-multiaddr = { package = "parity-multiaddr", version = "0.5.0" }
+authority-discovery-primitives = { package = "substrate-authority-discovery-primitives", path = "../authority-discovery/primitives", default-features = false }
```
tomaka: I suppose this was not intended?

```suggestion
```

core/grafana-data-source/src/util.rs
```diff
@@ -0,0 +1,43 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+/// Get the current unix timestamp in milliseconds.
+pub fn now_millis() -> i64 {
+	chrono::Utc::now().timestamp_millis()
```
tomaka: Could use `SystemTime` so that we don't depend on `chrono`:

Something like that (pseudo-code, needs to handle negative values):

```suggestion
	let dur = SystemTime::UNIX_EPOCH.elapsed().unwrap(); dur.as_secs().saturating_mul(1_000_000).saturating_add(dur.subsec_nanos() / 1000)
```

core/grafana-data-source/src/util.rs
```diff
@@ -0,0 +1,43 @@
...
+// find the index of a timestamp
+pub fn find_index(slice: &[(f32, i64)], timestamp: i64) -> usize {
+	slice.binary_search_by_key(&timestamp, |&(_, timestamp)| timestamp)
+		.unwrap_or_else(|index| index)
+}
+
+// Evenly select `num_points` points from a slice
+pub fn select_points<T: Copy>(slice: &[T], num_points: usize) -> Vec<T> {
+	(0 .. num_points - 1)
+		.map(|i| slice[i * slice.len() / (num_points - 1)])
```
tomaka: The server will panic if the user does a query with `num_points` equal to 0 or 1. We should try to avoid any "raw" arithmetic.

core/grafana-data-source/src/server.rs
```diff
@@ -0,0 +1,134 @@
...
+				.body(Body::from(string))
+				.map_err(|e| e.into())
+		})
+	)
+}
+
+/// Start the data source server.
+///
+/// The server shuts down cleanly when `shutdown` resolves.
+pub fn run_server<F>(address: &std::net::SocketAddr, shutdown: F) -> impl Future<Item=(), Error=()>
```
tomaka: I'm not sure how hyper works exactly (as it's quite un-idiomatic in this regard), I don't think we should pass a `shutdown` parameter here. Instead, we should do `run_server(...).select(exit)` in the service. When `exit` triggers, the `Future` here would then be destroyed and would shut down the server.

rphmeier: The service has some APIs for avoiding explicit `select`ion because it is error-prone to omission. I think the API is called `spawn_task`, where we pass a future that we want to be combined with the overall `exit`.

expenses: I'm happy to use something other than hyper if it would help, I just chose it because it's already used in substrate.

core/grafana-data-source/src/util.rs
expenses: I'm not sure we can get rid of `chrono` without replacing it with something else because we need a way to parse the timestamp strings coming from grafana.

kianenigma:
Do we have any indication of the **block-time** stored somewhere? If so, I think it would be super helpful to report it somewhere. Some matters that can be addressed with it: 
- increased block times due to validators being offline
- Give us an indication of how much stuff we're packing into `on_initialize`, `on_finalize` on _special_ occasions, like end of session, end of era etc.

expenses:
Yeah, I was wondering about that. It doesn't seem to be reported anywhere in telemetry though, so I haven't bothered to add it yet.

kianenigma:
Telementry and any UI app, I imagine, don't want to store this and should only care for the last block time (or last N at most). Ideally, I see grafanara storing all block times and we can easily identify the outliers. Anyhow, yeah I suppose this should be sqeuezed out of somewhere, if available, and no need to pass it to telementry imo.

core/grafana-data-source/src/server.rs
expenses: Where should `spawn_task` be called in this case? I see it only being used in the node currently.

tomaka: What I was suggesting is that instead of doing [this](https://github.com/paritytech/substrate/pull/3913/files#diff-1c0bedf4d6831513aa1122c56b1003bdR1118), you do something like this:

```rust
grafana_data_source::run_server(&port).select(exit.clone())
```

And the `exit` parameter would no longer be passed to `run_server`.

But as I mentioned I would have to look into the docs of hyper to see if it properly shuts down when you kill the future, so you tell me if that makes sense.

expenses: Yep, `ctrl`-`c` seems to stall forever if I use `select` instead. Plus, the code ends up looking more like `.select(exit.clone()).map(|_| ()).map_err(|_| ())`, which is less pleasant.

expenses:
Ok, I've changed the test code to be closer to how the node spawns and kills futures. It seems that `with_graceful_shutdown` only shuts the server down when the server has something to respond to; I can run the test binary ctrl-c immediately and it won't finish until either a day has passed and the `clean_up` code gets run, or I switch to the browser tab.

Honestly, at this point I'm tempted to just use something like the [prometheus client library](https://crates.io/crates/prometheus) and just tell users to run that server instead.

expenses:
This still happens even if I use something like
```rust
Server::bind(address)
		.serve(|| service_fn(api_response))
		.with_graceful_shutdown(shutdown.clone())
		.select2(shutdown)
```

expenses:
The issues with `Future`s seem to be resolved with `async fn`s, so I'm waiting on that.

expenses:
The code complies, but currently panics with a mis-matched `tokio` version: `Thread 'main-tokio-0' panicked at 'timer error: timer is shutdown', /Users/ashley/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-timer-0.3.0-alpha.6/src/delay.rs:119`, so we're waiting on `tokio` `0.2`.

tomaka:
I didn't see that you switched to the new hyper.
I'm not a fan of this, as this adds a hard requirement on `tokio`.
If we use hyper, in my opinion we should do the networking manually and use the non-I/O components of hyper only.

expenses:
Hmm ok, I didn't realise that this was a problem. Hyper has a `tokio` feature that we can disable, but it removes [`AddrIncoming`](https://docs.rs/hyper/0.13.0-alpha.4/hyper/server/conn/struct.AddrIncoming.html), the connection stream, so we've gotta find an alternative to that.

OR we could use something other than hyper.

OR I could revert the previous commits, convert the hyper future to `0.3` and still use `async` for `run_server` and `clean_up`, which might fix the problems I've been having.

DemiMarie-parity:
@tomaka may I ask why?

tomaka:
@DemiMarie-parity If you use tokio sockets, you can only ever use a tokio executor to process them.

gavofyork:
> Hmm ok, I didn't realise that this was a problem. Hyper has a `tokio` feature that we can disable, but it removes [`AddrIncoming`](https://docs.rs/hyper/0.13.0-alpha.4/hyper/server/conn/struct.AddrIncoming.html), the connection stream, so we've gotta find an alternative to that.
> 
> OR we could use something other than hyper.
> 
> OR I could revert the previous commits, convert the hyper future to `0.3` and still use `async` for `run_server` and `clean_up`, which might fix the problems I've been having.

Was there a decision on this? I'd actually quite like to see this merged...

tomaka:
> Was there a decision on this?

I think it's fine to merge now and open an issue about that.

gavofyork:
> > Was there a decision on this?
> 
> I think it's fine to merge as-is and open an issue about that.

Ok. @expenses just need to get the CI passing then..

expenses:
Ok, I've switched to using `async-std` for networking. This both fixes the problems with mismatched `tokio` versions, and gets rid of `hyper`'s `tokio` dependency.

expenses:
`async-std`'s lack of support for wasm remains a problem though :^<

Cargo.toml
```diff
@@ -41,6 +41,8 @@ members = [
 	"client/transaction-pool",
 	"client/transaction-pool/graph",
 	"client/utils/wasm-builder-runner",
+	"client/grafana-data-source",
+    "client/grafana-data-source/test",
```
tomaka:
```suggestion
	"client/grafana-data-source/test",
```

client/grafana-data-source/test/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+		record_metrics!(
+			"random data" => random,
+			"random^2" => random * random
+		);
+	}
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+	ctrlc::set_handler(|| {
```
tomaka: I don't think we need the `ctrlc` thing at all.

core/service/Cargo.toml
```diff
@@ -0,0 +1,62 @@
+[package]
```
tomaka: I think this was added by accident.

client/grafana-data-source/test/src/main.rs
expenses: That was useful for testing to make sure things would work in the same way as the node, but it's not really needed now.

gavofyork:
@expenses please update the lock file. in the future i'd recommend you work in branches off this repo so that stuff like this can be done by anyone else without having to open a new PR.

expenses:
@gavofyork done :^)

tomaka:
> i'd recommend you work in branches off this repo so that stuff like this can be done by anyone else without having to open a new PR.

For what it's worth, GitHub allows contributors to edit branches corresponding to PRs, even if they are outside of the repo.

expenses:
Also, I'm not a part of the parity github organization yet 😉😉

gavofyork:
image

gavofyork:
Full test suite passed on build server.

gavofyork:
> > i'd recommend you work in branches off this repo so that stuff like this can be done by anyone else without having to open a new PR.
> 
> For what it's worth, GitHub allows contributors to edit branches corresponding to PRs, even if they are outside of the repo.

that has never worked for me. can you explain how?

image

tomaka:
> that has never worked for me. can you explain how?

Hmmm
The option is enabled by default but people can disable it, that's maybe why

bkchr:
For me it also never worked.

"You can set commit permissions when you first create a pull request from a fork. For more information, see "Creating a pull request from a fork." Additionally, you can modify an existing pull request to let repository maintainers make commits to your branch."

Most people probably have disabled it.

expenses:
Oh, I was assuming it was enabled on this PR (as I thought that was the default) but it's not. Sorry 😅

expenses:
checkbox

dlguddus:
Hi. We are implementing a Prometheus exporter runtime in substrate and connecting into grafana server. I am afraid it is a duplicated job with almost same objective.

I am curious why you don't use already well known open source program "prometheus" and implemented subset of its feature by hand?

Our repository is here
https://github.com/nodebreaker-carlsagan/substrate/tree/prometheus_v0.1/client/prometheus

Thank you.

expenses:
@dlguddus I decided not to use Prometheus because it's overkill if you just want to record metrics. For most use-cases, the advantages of Prometheus such as persistent metrics and being able to store metrics from multiple nodes are not needed. I believe that we should be aiming for good defaults in Substrate, and using a simple in-memory database is a better default than needing Prometheus.

That said, it would be great if we could support both. With #4187 we could do this, or if you setup your Prometheus observer to use the existing `record_metrics` macro, that would be great!

dlguddus:
As a validator, we are running about 30 nodes. We expect it would be more than 100 next year. Definitely we need generalized data import from our monitoring center. This is our common usecase, as a validator because we are not only validating polkadot.

As you can see in our implementation, prometheus does not store any data in local PC, but just realtime passing the data to monitoring(grafana) server. So I expect it is lighter than your implementation which stores data in local disk.

Do I understand the specs correctly?

dlguddus:
And, I think #4187 is definitely a must go work because it can minimize complexity of codebase for extracting data for various data exporter such as telemetry, prometheus, grafana-api, rpc, lcd, and more. 

We will communicate there to contribute and discuss. Thanks for inviting!

PR #3922, opened 2019-10-26
Basic extrinsic pool benchmarks

NikVolf:
```
cargo bench -p substrate-transaction-graph
```

### DIFF
--- a/core/transaction-pool/graph/benches/basics.rs
+++ b/core/transaction-pool/graph/benches/basics.rs
@@ -42,7 +42,6 @@ impl ChainApi for TestApi {
 	type Error = error::Error;
 	type ValidationFuture = futures::future::Ready<error::Result<TransactionValidity>>;
 
-	/// Verify extrinsic at given block.
 	fn validate_transaction(
 		&self,
 		at: &BlockId<Self::Block>,
@@ -79,7 +78,6 @@ impl ChainApi for TestApi {
 		})
 	}
 
-	/// Returns a block number given the block id.
 	fn block_id_to_number(&self, at: &BlockId<Self::Block>) -> Result<Option<NumberFor<Self>>, Self::Error> {
 		Ok(match at {
 			BlockId::Number(num) => Some(*num),
@@ -87,7 +85,6 @@ impl ChainApi for TestApi {
 		})
 	}
 
-	/// Returns a block hash given the block id.
 	fn block_id_to_hash(&self, at: &BlockId<Self::Block>) -> Result<Option<BlockHash<Self>>, Self::Error> {
 		Ok(match at {
 			BlockId::Number(num) => Some(H256::from_low_u64_be(*num)).into(),
@@ -95,7 +92,6 @@ impl ChainApi for TestApi {
 		})
 	}
 
-	/// Hash the extrinsic.
 	fn hash_and_length(&self, uxt: &ExtrinsicFor<Self>) -> (Self::Hash, usize) {
 		let encoded = uxt.encode();
 		(blake2_256(&encoded).into(), encoded.len())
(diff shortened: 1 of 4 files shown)


core/transaction-pool/graph/benches/basics.rs
```diff
@@ -0,0 +1,150 @@
+use criterion::{criterion_group, criterion_main, Criterion};
```
tomusdrw: preamble missing

core/transaction-pool/graph/benches/basics.rs
```diff
@@ -0,0 +1,150 @@
...
+
+fn uxt(transfer: Transfer) -> Extrinsic {
+	Extrinsic::Transfer(transfer, Default::default())
+}
+
+fn bench_configured(pool: Pool<TestApi>, number: u64) {
+	let mut nonce = 1;
+	let mut futures = Vec::new();
+
+	for _ in 0..number {
```
tomusdrw: You could just do `for nonce in 1..=number`.

core/transaction-pool/graph/benches/basics.rs
```diff
@@ -0,0 +1,150 @@
...
+
+		let mut requires = vec![];
+		if nonce > 1 && self.nonce_limit.is_some() {
+			requires.push(to_tag(nonce-1, from.clone()));
+		};
+
+		let mut provides = vec![];
+		provides.push(to_tag(nonce, from.clone()));
+		if self.nonce_limit.is_some() {
+			provides.push(to_tag(nonce+1, from.clone()));
```
tomusdrw: That will definitely work, but the outcome will be really tricky. Basically if `nonce_limit` is set every transaction is actually providing two tags:
`nonce` and `nonce+1`
So this means that every two consecutive transactions will overlap.
Since the pool only supports exactly one transaction that provides given tag, when you first import `TxA(nonce, nonce+1)` and then `TxB (nonce+1, nonce+2)`, `TxA` will either be removed from the pool (if the priority of `TxB` is higher) or `TxB` will not be accepted. In the first case I think it may break the entire chain of transactions and move them to future since there will be no transaction that can provide `nonce`, so tags `required` by `TxB` will never be satisfied.

core/transaction-pool/graph/benches/basics.rs
```diff
@@ -0,0 +1,150 @@
...
+			amount: 5,
+			nonce,
+		});
+
+		nonce += 1;
+
+		futures.push(pool.submit_one(&BlockId::Number(1), xt));
+	}
+
+	let _res = block_on(futures::future::join_all(futures.into_iter()));
```
tomusdrw: I think you should check that the result here is successful, I suspect you will learn that some transactions were not imported.

core/transaction-pool/graph/benches/basics.rs
```diff
@@ -0,0 +1,150 @@
...
+
+	// instantly producing "blocks" and pruning all ready until no ready
+	let mut block_num = 2;
+	loop {
+		if let Some(ready) = pool.ready().next() {
```
tomusdrw: Maybe just `while let`?

core/transaction-pool/graph/benches/basics.rs
```diff
@@ -0,0 +1,150 @@
...
+			//dbg!(ready.clone());
+			block_on(pool.prune_tags(
+				&BlockId::Number(block_num),
+				vec![to_tag(ready.data.transfer().nonce, ready.data.transfer().from.clone())],
```
tomusdrw: why not just use `ready.provided_tags`? The tag here is only one of the 2 this transaction provides in case you run with `nonce_limit`. Everything should work fine, but it doesn't necessarily reflect the actual workflow.

core/transaction-pool/graph/benches/basics.rs
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	let _res = block_on(futures::future::join_all(futures.into_iter()));
+
+	// instantly producing "blocks" and pruning all ready until no ready
+	let mut block_num = 2;
+	loop {
+		if let Some(ready) = pool.ready().next() {
+			//dbg!(ready.clone());
+			block_on(pool.prune_tags(
```
tomusdrw: The function is actually designed to prune multiple transactions at once, pruning them one-by-one is going to take way longer than actually necessary.

NikVolf: it's funny that you mention it, so when I switched to batch posting all ready to prune_tags, I got 10x slowdown :)

core/transaction-pool/graph/benches/basics.rs
```diff
@@ -0,0 +1,158 @@
...
+use test_runtime::{Block, Extrinsic, Transfer, H256, AccountId};
+use sr_primitives::{
+	generic::BlockId,
+	transaction_validity::{TransactionValidity, TransactionTag as Tag},
+};
+use primitives::blake2_256;
+
+#[derive(Clone, Debug, Default)]
+struct TestApi {
+	nonce_limit: Option<u64>,
```
marcio-diaz: I think you are only using it as a `bool`.

core/transaction-pool/graph/benches/basics.rs
```diff
@@ -0,0 +1,158 @@
...
+		let nonce = uxt.transfer().nonce;
+		let from = uxt.transfer().from.clone();
+
+		futures::future::ready(
+			Ok(Ok(ValidTransaction {
+				priority: 4,
+				requires: if nonce > 1 && self.nonce_limit.is_some() { vec![to_tag(nonce-1, from.clone())] } else {vec![] },
+				provides:
+					if self.nonce_limit.is_some() {
+						vec![to_tag(nonce+1, from)]
```
marcio-diaz: Isn't `provides: vec![to_tag(nonce, from)]` enough?

tomusdrw:
Turns out the batch pruning was slow, because all pruned transactions were just re-imported to the pool right after (since we didn't tell the pool, that the transactions are not valid any more). The loop was just finishing after 10 iterations (reaching `longevity` period) and it was either 10 iterations of one-remove-one-resubmit, or all-remove-all-resubmit.

marcio-diaz:
> Turns out the batch pruning was slow, because all pruned transactions were just re-imported to the pool right after (since we didn't tell the pool, that the transactions are not valid any more). The loop was just finishing after 10 iterations (reaching `longevity` period) and it was either 10 iterations of one-remove-one-resubmit, or all-remove-all-resubmit.

Nice catch! @NikVolf @tomusdrw

PR #3927, opened 2019-10-26
Substrate EVM

sorpaas:
This implements an EVM execution module in SRML, to allow unmodified EVM code to be executed in a Substrate-based blockchains.

#### Execution Lifecycle

Accounts that the EVM can touch are stored in `srml_evm::Accounts`. Users can use call `Module::deposit_balance` and `Module::withdraw_balance` to transfer balance from normal `balances` module into the EVM module. There's one-to-one correspondence of `balances` account and EVM external accounts, defined via `Trait::ConvertAccountId`.

Once balance is in, users can then use `Module::call` and `Module::create` to execute normal EVM transactions, which corresponds to Ethereum message call transactions and contract creation transactions.

#### Compatibility

The module should be able to produce nearly identical result as running on Ethereum mainnet, including gas cost and balance changes. Observable behaviors by the contract that might be different include:

* Block hashes. We use the `system` module to fetch block hashes. The available length may not be 256, but depends on the configuration of the `system` module.
* Difficulty and coinbase. This value does not quite make sense in this module, and is currently hard coded to zero.

I think we currently don't aim to make unobservable behaviors, such as state root, to be the same. We also don't aim to follow the exact same transaction / receipt format. However, given one Ethereum transaction and one Substrate balances user private key, one should be able to convert any Ethereum transaction into a transaction compatible to use in this module.

The gas configurations are currently hard-coded to Istanbul hard fork. It can later be expanded to support earlier hard fork configurations.

#### EVM Engine

The current underlying EVM engine is [SputnikVM](https://github.com/sorpaas/rust-evm). The engine is overhauled so that it's [modular](https://github.com/corepaper/evm). One of the important reasons for this is that in the future we'll want to allow users to swap out components like gasometer, and inject their own customized ones.

#### Status

TODOs before this PR can be merged:

- [x] Remove all TODO comments in code.
- [x] Make sure gasometer compliance with Ethereum.
- [x] Fix issues regarding EVM version support.
- [x] Precompile interface and custom system call support.
- [x] Refactoring duplicate code of call/create.

Improvements that can be done, but maybe better in a future PR:

- More customization. Allow selecting gasometer and gas cost configurations.
- Performance. The current substate merging/reverting logic can be optimized a lot.


Cargo.toml
```diff
@@ -130,3 +131,7 @@ is-it-maintained-open-issues = { repository = "paritytech/substrate" }
 [profile.release]
 # Substrate runtime requires unwinding.
 panic = "unwind"
+
+[patch.crates-io]
```
sorpaas: Those lines can be removed after https://github.com/paritytech/parity-common/pull/250 is merged and `impl-rlp 0.2.1` is released.

srml/evm/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! EVM execution module for Substrate
+
+// Ensure we're `no_std` when compiling for Wasm.
+#![cfg_attr(not(feature = "std"), no_std)]
+
+extern crate alloc;
```
xlc: I don't think this is needed?

srml/evm/src/backend.rs
```diff
@@ -0,0 +1,202 @@
...
+	fn block_difficulty(&self) -> U256 {
+		U256::zero()
+	}
+
+	fn block_gas_limit(&self) -> U256 {
+		U256::zero()
+	}
+
+	fn chain_id(&self) -> U256 {
+		U256::one()
```
pepyakin: Can we actually pass `runtime_io::chain_id()` here?

srml/evm/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+				&sender,
+				value,
+				WithdrawReason::Reserve,
+				ExistenceRequirement::KeepAlive,
+			)?;
+
+			let bvalue = U256::from(UniqueSaturatedInto::<u64>::unique_saturated_into(value));
+			let address = T::ConvertAccountId::convert_account_id(sender);
+			Accounts::mutate(&address, |account| {
+				account.balance += bvalue;
```
xlc: will it be more performant if the account balance is moved to its own storage? otherwise this will unnecessarily read and write `Account::code` which can be large.

srml/evm/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+/// Trait for converting account ids of `balances` module into
+/// `H160` for EVM module.
+///
+/// Accounts and contracts of this module are stored in its own
+/// storage, in an Ethereum-compatible format. In order to communicate
+/// with the rest of Substrate module, we require an one-to-one
+/// mapping of Substrate account to Ethereum address.
+pub trait ConvertAccountId<A> {
+	/// Given a Substrate address, return the corresponding Ethereum address.
+	fn convert_account_id(account_id: A) -> H160;
```
xlc:
```suggestion
	fn convert_account_id(account_id: &A) -> H160;
```

srml/evm/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+			T::Currency::deposit_creating(&sender, value);
+
+			Ok(())
+		}
+
+		#[weight = SimpleDispatchInfo::FixedNormal(10_000)]
+		fn call(origin, target: H160, input: Vec<u8>, value: U256, gas_limit: u32) -> Result {
+			let sender = ensure_signed(origin)?;
+			let source = T::ConvertAccountId::convert_account_id(sender.clone());
+			let gas_price = T::FeeCalculator::gas_price();
```
xlc: it will be good if `gas_price` can be calculated from `tip` from balances module somehow

srml/evm/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+			let mut backend = Backend::<T>::new(&vicinity);
+			let mut executor = StackExecutor::new_with_precompile(
+				&backend,
+				gas_limit as usize,
+				&backend::GASOMETER_CONFIG,
+				T::Precompiles::execute,
+			);
+
+			let total_fee = gas_price * U256::from(gas_limit);
+			if Accounts::get(&source).balance < value + total_fee {
```
xlc: is this needed if `executor.withdraw` will fail when balance is not enough anyway?

srml/evm/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+				source,
+				target,
+				value,
+				input,
+				gas_limit as usize,
+			);
+
+			let ret = match reason {
+				ExitReason::Succeed(_) => Ok(()),
+				_ => Err("Execute message call failed")
```
xlc: is it possible to get more detailed reasons?

srml/evm/src/lib.rs
xlc: `value + total_fee` will overflow

srml/evm/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+			let mut backend = Backend::<T>::new(&vicinity);
+			let mut executor = StackExecutor::new_with_precompile(
+				&backend,
+				gas_limit as usize,
+				&backend::GASOMETER_CONFIG,
+				T::Precompiles::execute,
+			);
+
+			let total_fee = gas_price * U256::from(gas_limit);
+			if Accounts::get(&source).balance < value + total_fee {
```
xlc: `value + total_fee` will overflow

srml/evm/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+
+			let mut backend = Backend::<T>::new(&vicinity);
+			let mut executor = StackExecutor::new_with_precompile(
+				&backend,
+				gas_limit as usize,
+				&backend::GASOMETER_CONFIG,
+				T::Precompiles::execute,
+			);
+
+			let total_fee = gas_price * U256::from(gas_limit);
```
xlc: It is unlikely, but this may overflow

srml/evm/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
```
xlc: It is unlikely, but this may overflow

srml/evm/src/backend.rs
```diff
@@ -0,0 +1,202 @@
...
+					}
+
+					if delete_empty && account.is_empty() {
+						Accounts::remove(address);
+					} else {
+						Accounts::insert(address, account);
+					}
+				},
+				Apply::Delete { address } => {
+					Accounts::remove(address);
```
xlc: Also `AccountStorages::remove_prefix(address);`?

srml/evm/src/backend.rs
xlc: @pepyakin I did a quick search, looks like `runtime_io::chain_id` is hardcoded to return 42?shouldn't it be configurable in chain spec?

pepyakin: That's a good question which unfortunately I cannot answer.

srml/evm/src/lib.rs
pepyakin: Actually, I'd assume that this module should eventually be integrated with the weights system. Otherwise, gas would be just a number that doesn't influence the block weight.

srml/evm/src/lib.rs
sorpaas: I moved code to its own storage. I'm currently still keeping `balance` and `nonce` in one place because they're usually changed together.

srml/evm/src/lib.rs
sorpaas: Line below this withdraws the total fee, which fails if `balance` is lower than `total_fee`. However, the check for value, if without this line, only happens inside call, after all the substate structures are created. That is due to the behavior of `CALL` in Ethereum -- it adds stipend first into gas counter (which is considered even when the call failed), and only after then transfer condition is checked.

I'm fine removing this or keeping this, but currently I think keeping this would be slightly clearer. I do think there may be ways to refactor the stipend logic so that things can be nicer.

srml/evm/src/lib.rs
sorpaas: I added different error messages for all four return reason. Normal exit `Succeed`, Common out-of-gas errors `Error`, Explicit revert `Revert` and VM fatal error (such as operation not supported) `Fatal`.

srml/evm/src/backend.rs
```diff
@@ -127,15 +113,15 @@ impl<'vicinity, T: Trait> BackendT for Backend<'vicinity, T> {
 	}
 
 	fn code_size(&self, address: H160) -> usize {
-		self.account(address).code.len()
+		AccountCodes::get(&address).len()
```
xlc: I think you can use `decode_len` to avoid decode the whole value

srml/evm/src/lib.rs
xlc: It will be the gas limit influence the block weight not gas price?

xlc: It will be the gas limit influence the block weight not gas price? Gas price should affect the transaction priority.

srml/evm/src/backend.rs
xlc: #3990

srml/evm/src/lib.rs
pepyakin: Yes, gas limit should play a similar role to that of the weight. I think gas price should be a derivative from the fee computed for the given gas limit and the price paid for that amount of weight.

sorpaas: https://github.com/paritytech/substrate/issues/4007

PR #3930, opened 2019-10-27
runtime: Account nicknames

gavofyork:
Simple "nicknames" module for attaching names to accounts.

### DIFF
--- a/srml/nicks/src/lib.rs
+++ b/srml/nicks/src/lib.rs
@@ -1,4 +1,4 @@
-// Copyright 2017-2019 Parity Technologies (UK) Ltd.
+// Copyright 2019 Parity Technologies (UK) Ltd.
 // This file is part of Substrate.
 
 // Substrate is free software: you can redistribute it and/or modify
--- a/srml/nicks/src/lib.rs
+++ b/srml/nicks/src/lib.rs
@@ -1,0 +1,331 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
+// This file is part of Substrate.
+
+// Substrate is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! # Nicks Module
+//!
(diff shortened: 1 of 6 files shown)


srml/nicks/src/lib.rs
```diff
@@ -0,0 +1,331 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2019 Parity Technologies (UK) Ltd.
```

srml/nicks/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+//! ## Overview
+//!
+//! Nicks is a trivial module for keeping track of account names on-chain. It makes no effort to
+//! create a name hierarchy, be a DNS replacement or provide reverse lookups.
+//!
+//! ## Interface
+//!
+//! ### Dispatchable Functions
+//!
+//! Only the sudo key can call the dispatchable functions from the Sudo module.
```
bkchr: Still mentions `sudo`.

srml/nicks/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+		///
+		/// The dispatch origin for this call must be _Root_ or match `T::KillOrigin`.
+		///
+		/// # <weight>
+		/// - O(1).
+		/// - Limited storage reads.
+		/// - One DB write (event).
+		/// - Unknown weight of derivative `proposal` execution.
+		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
```
kianenigma:
```suggestion
		#[weight = SimpleDispatchInfo::FreeOperational]
```

srml/nicks/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+		///
+		/// If the account doesn't already have a name, then a fee of `ReservationFee` is reserved
+		/// in the account.
+		///
+		/// The dispatch origin for this call must be _Signed_.
+		///
+		/// # <weight>
+		/// - O(1).
+		/// - Limited storage reads.
+		/// - One DB write (event).
```
bkchr: We have a second write for inserting the name.

srml/nicks/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+		}
+
+		/// Clear an account's name and return the deposit. Fails if the account was not named.
+		///
+		/// The dispatch origin for this call must be _Signed_.
+		///
+		/// # <weight>
+		/// - O(1).
+		/// - Limited storage reads.
+		/// - One DB change.
```
bkchr: We are writing an event and taking the name.

gavofyork:
> Looks good, but don't we need to implement OnFreeBalanceZero to clear the nickname?

As long as ReservationFee is greater than ExistentialDesposit (will add the note to the docs), then no, since there will always need to be at least that reserved.

bkchr:
Ahh okay, makes sense :)

PR #3936, opened 2019-10-27
`decl_storage!` check for duplicate `config()`/`get()`

bkchr:


bkchr:
Rustc will complain about the `GenesisConfig` having duplicate fields and this error message is not very intuitive to the user IMHO.

bkchr:
I'm with you that I would like to have a better error message, directly indicating where to find the duplicate, but sadly that is not possible AFAIK. (at the moment)

Nevertheless, I think we should catch as much errors as possible in the macro, rustc will generate you something, but the context will always be something different.

PR #3937, opened 2019-10-28
Tip payment is a different withdraw reason.

gavofyork:
Several subsystems (staking, elections-phragmen) routinely introduce account locks. Though account locks are able to cover only a subset of funds, allowing there to be unlocked funds to do stuff like sending new transactions, often times it's easy to accidentally lock all funds in the account, preventing all further activity until the account is topped up with fresh, unlocked funds.

This introduces a new kind of `WithdrawReason`: `Tip`, which allows us to distinguish between transaction payment that goes primarily to the system `::TransactionPayment` and transaction payment that goes primarily to a validator `::Tip`. By doing this we can allow transactions to draw down on locked funds if all funds are exclusively used for `::TransactionPayment` and not `::Tip`, thereby removing the ability for a validator to recycle funds en masse and theoretically hold locked funds with nothing at stake.

### DIFF
--- a/core/primitives/src/sr25519.rs
+++ b/core/primitives/src/sr25519.rs
@@ -129,9 +129,6 @@ impl std::fmt::Display for Public {
 	}
 }
 
-#[cfg(not(feature = "std"))]
-use core as std;
-
 impl rstd::fmt::Debug for Public {
 	#[cfg(feature = "std")]
 	fn fmt(&self, f: &mut rstd::fmt::Formatter) -> rstd::fmt::Result {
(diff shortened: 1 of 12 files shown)


xlc:
Does it make sense to reuse the Transfer reason for tip payment? in the sense it is transfer fund to validators for additional service.

gavofyork:
Thing is, it's not a straight transfer to validators - it depends on the chain's logic. Some chains might take e.g. 20% of the tip for the treasury.

kianenigma:
at the first glance I can add: 

shouldn't this now update the `lock()` in staking/elections to lock for `reasons::all().except(Tip)`?

gavofyork:
No - Should be `reasons::all().except(TransactionPayment)`

kianenigma:
right about the lock type -- just needs to set those properly (now afaiu all locks are set for `::all()`) and then good to go.

PR #3939, opened 2019-10-28
Fix a import+prune+replace case for multi-provides transactions.

### LINKED ISSUE #3923: Transaction pool might panic if tags are pruned but no hashes provided on prune
Maybe it is unavoidable, but it is still might be an attack vector for some usage scenarios

```
Every time transaction is imported its hash is placed in `ready` map and tags in `provided_tags`;
Every time transaction is removed from the queue we remove the hash from `ready` map and from `provided_tags`;
Hence every hash retrieved from `provided_tags` is always present in `ready`;
qed
```

tomusdrw:
Closes #3923 

Fix a potential panic in the transaction pool implementation in the following scenario:
1. We import transaction that provides more than one tag.
2. We prune only one of those tags. But it results in the transaction being removed anyway.
3. We import another transaction that provides the previously un-pruned tag.
4. Now since the pruning in (2) was not done correctly, we still believe that the transaction is in the pool, hence treat the new one as a replacement transaction. This leads to a situation where we try to remove a non-existing hash.

This situation has been found by @NikVolf by writing benchmarks, however it's very unlikely to happen in reality with current chain implementations that do not alter the way `TransactionValidity` is generated (i.e., rely on the account-nonce based model).

I've also added a simplified test case, based on the test that Nikolay prepared.


core/transaction-pool/src/tests.rs
```diff
@@ -181,3 +189,29 @@ fn should_ban_invalid_transactions() {
...
+	block_on(pool.submit_one(&BlockId::number(0), xt.clone())).unwrap();
+	assert_eq!(pool.status().ready, 1);
+
+	// remove the transaction that just got imported.
+	block_on(pool.prune_tags(&BlockId::number(1), vec![vec![209]], vec![])).unwrap();
+	assert_eq!(pool.status().ready, 0);
+
+	// so now let's insert another transaction that also provides the 155
+	let xt = uxt(Alice, 210);
+	block_on(pool.submit_one(&BlockId::number(1), xt.clone())).unwrap();
```
NikVolf: I'd always prefer expect vs unwrap everywhere since it saves some time later when modifying logic to see what fails

gavofyork:
@tomusdrw test failing...

tomusdrw:
Fixed now.

PR #3940, opened 2019-10-28
Updates to elections-phragmen and some runtime docs. 

kianenigma:
- some leftover doc updates for weights (related to the ongoing doc PR)
- storage -> associated type for elections
- Store the equivalent of `exposure.total` in elections as well. if executed according to the plan, no logic has changed.

needless to say, needs to be taken into account in the next polkadot update.

### DIFF
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -327,6 +327,9 @@ impl collective::Trait<CouncilCollective> for Runtime {
 parameter_types! {
 	pub const CandidacyBond: Balance = 10 * DOLLARS;
 	pub const VotingBond: Balance = 1 * DOLLARS;
+	pub const TermDuration: BlockNumber = 7 * DAYS;
+	pub const DesiredMembers: u32 = 13;
+	pub const DesiredRunnersUp: u32 = 7;
 }
 
 impl elections_phragmen::Trait for Runtime {
@@ -335,6 +338,9 @@ impl elections_phragmen::Trait for Runtime {
 	type CurrencyToVote = CurrencyToVoteHandler;
 	type CandidacyBond = CandidacyBond;
 	type VotingBond = VotingBond;
+	type TermDuration = TermDuration;
+	type DesiredMembers = DesiredMembers;
+	type DesiredRunnersUp = DesiredRunnersUp;
 	type LoserCandidate = ();
 	type BadReport = ();
 	type KickedMember = ();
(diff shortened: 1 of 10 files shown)


node/runtime/src/lib.rs
```diff
@@ -328,6 +328,9 @@ impl collective::Trait<CouncilCollective> for Runtime {
 parameter_types! {
 	pub const CandidacyBond: Balance = 10 * DOLLARS;
 	pub const VotingBond: Balance = 1 * DOLLARS;
+	pub const TermDuration: BlockNumber = 7 * DAYS;
+	pub const DesiredMembers: u32 = 13;
```
kianenigma: should be replaced maybe with more meaningful numbers.

node/runtime/src/lib.rs
```diff
@@ -82,7 +82,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 189,
```
gavofyork: definitely a different spec.
```suggestion
	spec_version: 190,
```

node/runtime/src/lib.rs
```diff
@@ -328,6 +328,9 @@ impl collective::Trait<CouncilCollective> for Runtime {
 parameter_types! {
 	pub const CandidacyBond: Balance = 10 * DOLLARS;
 	pub const VotingBond: Balance = 1 * DOLLARS;
+	pub const TermDuration: BlockNumber = 7 * DAYS;
+	pub const DesiredMembers: u32 = 13;
+	pub const DesiredRunnersUp: u32 = 7;
```
gavofyork:
```suggestion
	pub const DesiredRunnersUp: u32 = 13;
```

node/runtime/src/lib.rs
```diff
@@ -328,6 +328,9 @@ impl collective::Trait<CouncilCollective> for Runtime {
 parameter_types! {
 	pub const CandidacyBond: Balance = 10 * DOLLARS;
 	pub const VotingBond: Balance = 1 * DOLLARS;
+	pub const TermDuration: BlockNumber = 7 * DAYS;
```
gavofyork:
```suggestion
	pub const TermDuration: BlockNumber = 10 * MINUTES;
```

until we're ready

gavofyork:
needs fixing...

kianenigma:
beast is sick -- have to retry in a short while.

gavofyork:
some legit failures.

kianenigma:
Just ran all the tests locally after the mini-patch. Now it should be all about the CI.

PR #3941, opened 2019-10-28
Implement Debug for some structures

c410-f3r:
`NetworkConfiguration`, `TransportConfig`, `NodeKeyConfig` and `Secret`.

Needs a new release of the `rust-libp2p` crate. This PR is just a
reminder.

### DIFF
--- a/core/network/src/config.rs
+++ b/core/network/src/config.rs
@@ -234,7 +234,7 @@ impl From<multiaddr::Error> for ParseErr {
 }
 
 /// Network service configuration.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct NetworkConfiguration {
 	/// Directory path to store general network configuration. None means nothing will be saved.
 	pub config_path: Option<String>,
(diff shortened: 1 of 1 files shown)


core/network/src/config.rs
```diff
@@ -372,7 +372,7 @@ pub enum NodeKeyConfig {
 pub type Ed25519Secret = Secret<ed25519::SecretKey>;
 
 /// The configuration options for obtaining a secret key `K`.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
```
tomaka: I'd prefer a manual implementation of `Debug` here that doesn't print the value of `K`.

Something like:

```rust
impl<K> fmt::Debug for Secret<K> {
	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
		match self {
			Secret::Input(_) => f.debug_tuple("Secret::Input").finish(),
			Secret::File(path) => f.debug_tuple("Secret::File").field(path).finish(),
			Secret::New => f.debug_tuple("Secret::New").finish(),
		}
	}
}
```

In practice I think that we never use anything like `Secret<Vec<u8>>`, but I think we're never careful enough when it comes to this.

c410-f3r: Thanks! Turns out that this suggestions is pretty good already so I simply copied and pasted

c410-f3r:
Thanks!

PR #3942, opened 2019-10-28
Header-only sync for old forks

arkpar:
* Sync requests coming from GRANDPA are now done in header-only mode.
* Tweaked import pipeline to enable header-only import.

### DIFF
--- a/core/consensus/common/src/block_import.rs
+++ b/core/consensus/common/src/block_import.rs
@@ -42,7 +42,7 @@ pub enum ImportResult {
 /// Auxiliary data associated with an imported block result.
 #[derive(Debug, Default, PartialEq, Eq)]
 pub struct ImportedAux {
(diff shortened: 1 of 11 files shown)


core/consensus/common/src/block_import.rs
```diff
@@ -35,11 +35,15 @@ pub enum ImportResult {
...
 	/// Block parent is not in the chain.
 	UnknownParent,
+	/// Parent state is missing.
+	MissingState,
 }
 
 /// Auxiliary data associated with an imported block result.
 #[derive(Debug, Default, PartialEq, Eq)]
 pub struct ImportedAux {
+	/// ONly the header has been imported. Block body verification was skipped.
```
marcio-diaz:
```suggestion
	/// Only the header has been imported. Block body verification was skipped.
```

PR #3944, opened 2019-10-28
Implement contract_getStorage RPC API

pepyakin:
This PR adds a new srml-contracts API for querying a contract's storage. This API is accessible by RPC and directly via Rust. This PR should ease https://github.com/paritytech/srml-contracts-waterfall/issues/6

The API is named `contracts_getStorage` and returns `Ok(Some(Bytes))` if the value is found, `Ok(None)` if the value is not set, or `Err` if the specified contract doesn't exist or is a tombstone.

cc @Stefie, @kwingram25


srml/contracts/rpc/src/lib.rs
```diff
@@ -121,4 +138,57 @@ where
 
 		Ok(exec_result)
 	}
+
+	fn get_storage(
+		&self,
+		address: AccountId,
+		key: Bytes,
```
tomusdrw: Why not use `H256` if you need to check the length anyway?

srml/contracts/rpc/src/lib.rs
```diff
@@ -121,4 +138,57 @@ where
...
+				// Handle general API calling errors.
+				Error {
+					code: ErrorCode::ServerError(RUNTIME_ERROR),
+					message: "Runtime trapped while querying storage.".into(),
+					data: Some(format!("{:?}", e).into()),
+				})?
+			.map_err(|e|
+				// Convert contract errors to RPC errors.
+				match e {
+					GetStorageError::ContractDoesntExist => Error {
```
tomusdrw: Maybe a `From<GetStorageError> for Error`?

srml/contracts/rpc/runtime-api/src/lib.rs
```diff
@@ -45,6 +45,18 @@ pub enum ContractExecResult {
 	Error,
 }
 
+pub type GetStorageResult = Result<Option<Vec<u8>>, GetStorageError>;
+
+/// The possible errors that can happen querying the storage of a contract.
+#[derive(Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+pub enum GetStorageError {
```
tomusdrw: Why not use the type from `srml` directly?

srml/contracts/rpc/runtime-api/src/lib.rs
```diff
@@ -45,6 +45,18 @@ pub enum ContractExecResult {
 	Error,
 }
 
+pub type GetStorageResult = Result<Option<Vec<u8>>, GetStorageError>;
+
+/// The possible errors that can happen querying the storage of a contract.
+#[derive(Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
```
tomusdrw: Is this actually neeed?

srml/contracts/rpc/runtime-api/src/lib.rs
```diff
@@ -45,6 +45,18 @@ pub enum ContractExecResult {
 	Error,
 }
 
+pub type GetStorageResult = Result<Option<Vec<u8>>, GetStorageError>;
```
tomusdrw: Missing docs?

srml/contracts/rpc/src/lib.rs
pepyakin: Oh, didn't realize that it is an option!

srml/contracts/rpc/src/lib.rs
pepyakin: This code is in crate A. `GetStorageError` is declared in crate `B`. `Error` is declared in crate `C`. I think orphan rules won't allow to do this? Unless you have something in mind that I don't : )

tomusdrw: Right, I guess we could wrap in a `struct GetStorageError(other::GetStorageError)` and then just `map_err(GetStorageError)?`.

Having a separate function would be ok too. I just feel bad seeing the "meat" of this function being diluted with tedious error handling.

pepyakin: Ok, you were right, this looks better!

srml/contracts/rpc/runtime-api/src/lib.rs
pepyakin: As per offline discussion, this would require another primitives crate. I am not sure if adding another crate is justified at this point, since the only use case for the moment would be avoiding one type definition, convertation logic.

kianenigma: Can you briefly say what was the offline reason of not using the type in the SRML?

kianenigma: introducing the primitives, only for this is an overkill, I agree. If you had other stuff in the contracts module which need to be shared and sort of postponed creating a primitive, then maybe now would be a good time. 

> This is not the end of the world, but it would be weird that it would depend on the entire crate

Yeah, the reason was that I also assumed this dependency is _bad/weird_, but then I was like: why though? Like, _really_ why? The rpc crate of one crate is really meant for that crate and only that crate imo. Hence, maybe it is fine to actually to depend on it?

I also don't have a strong opinion here and don't want to block the PR, so merge whenever needed, this is not a change request. But I do like the discussion.

PR #3945, opened 2019-10-28
State machine call proof backend

### LINKED ISSUE #3881: State machine incomplete proof
Current state machine proving backend only get proof for state machine backend operation 'storage' and 'child_storage'.
This is problematic in case such as cumulus where we also need the proof of operation such as 'storage_root' or any possible operation done by a runtime.
I believe it is also an issue for light client 'RemoteCall' proof.

So, the idea here would be to extract the proof directly from the 'HashDB' used by the trie backend. That way all trie accessed data will be in proof.

I got a prototype working branch https://github.com/cheme/substrate/tree/full_proof (https://github.com/paritytech/substrate/compare/master...cheme:full_proof) used to assert this approach.
Still needed before making a PR :
- [ ] assert we can change the proof in all case and have a single implementation.
- [ ] get rid of this mutex (only use for some trait constraint but shouldn't be here

Another word on it:
the approach (getting value at hashdb level) is expecting trie crate to do its query in a optimal way: that is not querying non needed node info.
This is rather relevant for multiple implementation, but we can say that if trie crate queries non needed node, the proof check will fail.
For 'storage_root' operation there is no problem currently : it just does insert and remove operation for all the delta.
For iterator like operation a reexecution need to go into all queried value: it means:
- implementation cannot buff the whole trie.
- implementation cannot do read ahead of a few elements.

So current implementation of iterator_with_prefix is a bit off: it uses triedb iterator seek on the prefix (that is good), but the end iteration condition involve querying a first value that does not start with the prefix.
This is not optimal as a trie with prefix primitive (does not exist) would stop by itself without this last element query.
The impact of this is that 'clear_prefix' currently have an unexpected semantic: if a proof was produced by a optimal implementation, the optimal implementation will have to add some nodes to the proof to be compatible with current runtime extrinsic.

cheme:
Closes #3881 , this PR replaces previous state machine proof backend to use a proof recorder that simply record over 'hash_db' trie accessed nodes.

This allows handle of proof/witness in case of cumulus, and is needed for completeness when doing call proof.

I did not totally replace the former proof recorder of state machine that is still use by change trie (change trie proof only requires trie or child trie proof I think cc/ @svyatonik ).

One thing I am not really happy in this PR is that the recorder is behind a 'RWLock', that is related to the need to use it as a hashdb for trie backend which requires Sync (and we need inner mutability). I spent a bit of time trying to get rid of it but did not manage to get something fine (previous recorder did use dedicated 'trie' crate primitives and could run behind a simple 'RcRefcell'.


bkchr:
Needs updating.

core/state-machine/src/proving_backend.rs
```diff
@@ -178,45 +183,67 @@ impl<'a, S, H> ProvingBackendEssence<'a, S, H>
...
+		let root = essence.root().clone();
+		let recorder = ProofRecorderBackend {
+			backend: essence.backend_storage(),
+			proof_recorder: proof_recorder,
+		};
+		ProvingBackend(TrieBackend::new(recorder, root))
 	}
 
-	/// Consume the backend, extracting the gathered proof in lexicographical order by value.
+	/// Extracting the gathered proof in undefined order.
```
cheme: CC\ @jimpo , I broke your ordering description here (I cannot order the recording of node on a call/exec). Is that something problematic?

jimpo: Oops, no I think I forgot to update the comment when I updated. Undefined is good. :-)

gavofyork:
Conflicts...

cheme:
after a few restart, the tests are passing again, status_sinks::tests::basic_usage seems a bit unreliable under heavy ci load (couldn't reproduce locally).

PR #3947, opened 2019-10-28
Add translate API for storage values

gavofyork:
This helps migrating from one version of a storage item to another. A follow-up PR should work with enumerable maps, too.

### DIFF
--- a/srml/support/src/storage/generator/value.rs
+++ b/srml/support/src/storage/generator/value.rs
@@ -60,6 +60,23 @@ impl<T: FullCodec, G: StorageValue<T>> storage::StorageValue<T> for G {
 		G::from_optional_value_to_query(value)
 	}
 
+	fn translate<O: Decode, F: FnOnce(Option<O>) -> Option<T>>(f: F) -> Result<T, ()> {
+		let key = Self::storage_value_final_key();
+
+		// attempt to get the length directly.
+		let maybe_old = match unhashed::get_raw(&key) {
+			Some(old_data) => Some(O::decode(&mut &old_data).ok_or(())?),
+			None => None,
+		};
+		let maybe_new = f(maybe_old);
+		if let Some(new) = maybe_new {
+			new.with_encoded(|d| unhashed::put_raw(&key, d));
+		} else {
+			unhashed::kill(&key);
+		}
+		Ok(new)
+	}
+
 	fn put<Arg: EncodeLike<T>>(val: Arg) {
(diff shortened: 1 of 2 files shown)


thiolliere:
Also I don't know if this is the best API, maybe it is more safe to do for type migration:

* create a new storage`MyStorageV2` with new types
* call a function to populate it and remove the old one.

kianenigma:
@thiolliere you can find both a test here, not sure how it can be migrated to this PR
https://github.com/paritytech/substrate/pull/3948

srml/support/src/storage/generator/mod.rs
```diff
@@ -30,3 +30,58 @@ pub use linked_map::{StorageLinkedMap, Enumerator, Linkage};
...
+			let old_key = ValueOld::storage_value_final_key();
+			let old_raw_value = unhashed::get_raw(&old_key).unwrap();
+
+			// put the old value in the new key.
+			let key = Value::storage_value_final_key();
+			unhashed::put_raw(&key, &old_raw_value);
+
+			// translate
+			let translate_fn = |old: Option<u32>| -> Option<(u64, u64)> {
+				Some(old.map_or_else(|| unimplemented!(), |o| (o.into(), (o*2).into())))
```
bkchr:
```suggestion
				old.map(|o| (o.into(), (o*2).into()))
```

srml/support/src/storage/generator/mod.rs
```diff
@@ -30,3 +30,58 @@ pub use linked_map::{StorageLinkedMap, Enumerator, Linkage};
...
+		let t = GenesisConfig::default().build_storage().unwrap();
+		TestExternalities::new(t).execute_with(|| {
+			// write and get the encoded value via proxy
+			ValueOld::put(1111);
+			let old_key = ValueOld::storage_value_final_key();
+			let old_raw_value = unhashed::get_raw(&old_key).unwrap();
+
+			// put the old value in the new key.
+			let key = Value::storage_value_final_key();
+			unhashed::put_raw(&key, &old_raw_value);
```
bkchr:
```suggestion
			unhashed::put_raw(&key, &1111u32.encode());
```

Then you can drop all these `ValueOld` stuff.

kianenigma: That's exactly how I did it at the first place and in the other test of elections-phragmen. But I find this to be much more expressive for a test that can serve as a guideline as well.

thiolliere: I'm ok with both way

PR #3948, opened 2019-10-28
Storage migration of elections-phragmen

kianenigma:


srml/elections-phragmen/src/lib.rs
```diff
@@ -159,6 +162,11 @@ decl_storage! {
 		/// The present candidate list. Sorted based on account id. A current member can never enter
 		/// this vector and is always implicitly assumed to be a candidate.
 		pub Candidates get(fn candidates): Vec<T::AccountId>;
+
+		/// Has the storage format been updated?
+		/// NOTE: Only use and set to false if you have used an early version of this module. Should
+		/// be set to true otherwise.
+		pub DidUpdate: bool = false;
```
pepyakin: DidMigrate? Since DidUpdate has already been used for in a different context (e.g. see `timestamp`)

srml/elections-phragmen/src/lib.rs
```diff
@@ -626,6 +638,36 @@ impl<T: Trait> Module<T> {
...
 		ElectionRounds::mutate(|v| *v += 1);
 	}
+
+	/// Perform the storage update needed to migrate the module from the initial version of the
+	/// storage.
+	///
+	/// If decoding the old storage fails in any way, the consequence is that we start with an empty
+	/// set.
+	fn do_update() {
+		let members_key = "PhragmenElection Members";
```
kianenigma: oh I can already get `hashed_key()`. will be updated.

kianenigma:
okay -- the test case should be enough for ensure this will do what it is supposed to do; but I would triple check every line of it.

srml/elections-phragmen/src/lib.rs
```diff
@@ -159,6 +161,11 @@ decl_storage! {
 		/// The present candidate list. Sorted based on account id. A current member can never enter
 		/// this vector and is always implicitly assumed to be a candidate.
 		pub Candidates get(fn candidates): Vec<T::AccountId>;
+
+		/// Has the storage format been updated?
+		/// NOTE: Only use and set to false if you have used an early version of this module. Should
+		/// be set to true otherwise.
+		DidMigrate: bool = false;
```
xlc: should just initialise this to `true` by here? also will it better to use a `u8` version byte instead?
upgrade existing chain won't reinitialise this and will have `DidMigrate` to be `None` which defaults to `false`

srml/elections-phragmen/src/lib.rs
```diff
@@ -626,6 +637,33 @@ impl<T: Trait> Module<T> {
...
+	/// storage.
+	///
+	/// If decoding the old storage fails in any way, the consequence is that we start with an empty
+	/// set.
+	fn do_migrate() {
+		// old storage format.
+		let old_members: Vec<T::AccountId> = unhashed::get_raw(&<Members<T>>::hashed_key())
+			.map_or_else(Default::default, |bytes| Decode::decode(&mut &*bytes).unwrap_or_default());
+		let old_runners: Vec<T::AccountId> = unhashed::get_raw(&<RunnersUp<T>>::hashed_key())
+			.map_or_else(Default::default, |bytes| Decode::decode(&mut &*bytes).unwrap_or_default());
```
bkchr:
```suggestion
			.and_then(|bytes| Decode::decode(&mut &*bytes).ok()).unwrap_or_default();
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -626,6 +637,33 @@ impl<T: Trait> Module<T> {
...
+
+	/// Perform the storage update needed to migrate the module from the initial version of the
+	/// storage.
+	///
+	/// If decoding the old storage fails in any way, the consequence is that we start with an empty
+	/// set.
+	fn do_migrate() {
+		// old storage format.
+		let old_members: Vec<T::AccountId> = unhashed::get_raw(&<Members<T>>::hashed_key())
+			.map_or_else(Default::default, |bytes| Decode::decode(&mut &*bytes).unwrap_or_default());
```
bkchr:
```suggestion
			.and_then(|bytes| Decode::decode(&mut &*bytes).ok()).unwrap_or_default();
```

srml/elections-phragmen/src/lib.rs
```diff
@@ -626,6 +637,33 @@ impl<T: Trait> Module<T> {
...
+			.map(|r| (r, Zero::zero()))
+			.collect();
+		let new_members: Vec<(T::AccountId, BalanceOf<T>)> = old_members
+			.into_iter()
+			.map(|r| (r, Zero::zero()))
+			.collect();
+
+		<Members<T>>::put(new_members);
+		<RunnersUp<T>>::put(new_runners);
+
```
bkchr:
```suggestion
```

srml/elections-phragmen/src/lib.rs
bkchr: Good ideas.

gavofyork: no need for version byte. once the upgrade is through, a second update should remove the item from storage, and then the storage item declaration may be removed from code completely.

this process can be repeated whenever necessary and doesn't leave an additional set of storage items around indefinitely.

gavofyork: can be left without default, or as is with = false.

in general we'd want an additional genesis item that set it to true, but setting a default isn't the way to do that.

srml/elections-phragmen/src/lib.rs
```diff
@@ -373,6 +380,10 @@ decl_module! {
 
 		/// What to do at the end of each block. Checks if an election needs to happen or not.
 		fn on_initialize(n: T::BlockNumber) {
+			if !DidMigrate::get() {
```
gavofyork: `DidMigrate::exists` is a little nicer.

kianenigma:
Changed to `didMigrate: bool` with no default and checking with `::exists()`. 

I saw your PR and I think that's also not that far from being ready. We _could_ merge them together and then do the migration, or take it step by step and do this manually.

srml/elections-phragmen/src/lib.rs
```diff
@@ -373,6 +380,10 @@ decl_module! {
 
 		/// What to do at the end of each block. Checks if an election needs to happen or not.
 		fn on_initialize(n: T::BlockNumber) {
+			if !DidMigrate::exists() {
```
bkchr:
```suggestion
			if !DidMigrate::get() {
```

`false` is the default value.

thiolliere:
if the runtime upgrade happens on a new era, then session::on_initialize is triggered before election::on_initialize, and the session on initialize will call staking::on_session_ending which will call phragmen::elect(..) which will panics because phragmen::on_initialize won't have been called already.

isn't it ? can we just say this won't happen this way. otherwise we can also make phragmen::on_initialize prior to session (in construct_runtime) (but this might break other assumption I'm not aware of)

thiolliere:
just a note: we are sure that no other module inside their on_initialize indirectly call on method of election-phragment isn't it ?

(to be more precise no other module that have a higher position in construct_runtime!)

kianenigma:
> just a note: we are sure that no other module inside their on_initialize indirectly call on method of election-phragment isn't it ?
> 
> (to be more precise no other module that have a higher position in construct_runtime!)

Good point, but not as far as I can reason about.

PR #3949, opened 2019-10-28
Remove footgun around session keys/handlers

bkchr:
- `OpaqueKeys` now has an associated type `KeyTypeIdProviders`. This can
be used in the runtime as input for `SessionHandler` from the session
trait.
- `impl_opaque_keys` now works with modules and extracts the `KeyTypeId`
from the module directly.
- Added some checks to the `session` storage initialization that checks
that the `SessionHandler` and `Keys` use the same number of keys and
that the order is equal.

### DIFF
--- a/core/sr-primitives/src/traits.rs
+++ b/core/sr-primitives/src/traits.rs
@@ -979,7 +979,7 @@ pub trait ValidateUnsigned {
 	fn validate_unsigned(call: &Self::Call) -> TransactionValidity;
 }
 
-/// Opaque datatype that may be destructed into a series of raw byte slices (which represent
+/// Opaque datatype that may be destructured into a series of raw byte slices (which represent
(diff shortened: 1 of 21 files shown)


core/sr-primitives/src/traits.rs
```diff
@@ -979,14 +979,14 @@ pub trait ValidateUnsigned {
 	fn validate_unsigned(call: &Self::Call) -> TransactionValidity;
 }
 
-/// Opaque datatype that may be destructured into a series of raw byte slices (which represent
+/// Opaque datatype that may be destructed into a series of raw byte slices (which represent
```
rphmeier: Destructured is correct here.

```suggestion
/// Opaque datatype that may be destructured into a series of raw byte slices (which represent
```

PR #3950, opened 2019-10-29
Retire `storage_items!`

### LINKED ISSUE #1466: Retire `storage_items!`
The `storage_items!` macro should be removed in favor of `decl_storage_items!`.

[Here](https://github.com/paritytech/substrate/blob/8ef14908ad9d46840db58808a72516ba9fb6602f/srml/support/src/storage/generator.rs#L197) the macro can be found.

While that is done, `StorageList` trait should be removed as well.

expenses:
Closes #1466.


bkchr:
@expenses these changes seem to break some tests, so you will need to take a closer look what is happening. Maybe `decl_storage` returns some other default value than `storage_Items!`.

expenses:
The test appears to have timed out. I tried to run the test suite on `master`, but that took forever as well, so I'm not sure what's going on here.

bkchr:
Yeah one test runs into an infinite loop, but as you see:
```
tests::authorities_call_works ... FAILED
```

Another one already failed. You should be able to replicate this with:
```
cargo test --release -p substrate-consensus-aura
```

bkchr:
> the test module of storage_items file have some tests, I pretty sure they are not redundant, they should be moved in srml-support-test crate

good point.

expenses:
I'm having some trouble with this failing test:
```rust
---- tests::authorities_call_works stdout ----
thread 'tests::authorities_call_works' panicked at 'assertion failed: `(left == right)`
  left: `[]`,
 right: `[Public(d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d (5GrwvaEF...)), Public(8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48 (5FHneW46...)), Public(90b5ab205c6974c9ea841be688864633dc9ca8a357843eeacf2314649965fe22 (5FLSigC9...))]`', core/consensus/aura/src/lib.rs:874:3
```

The authorities are set in the [`test-runtime/genesismap`](https://github.com/paritytech/substrate/blob/master/core/test-runtime/src/genesismap.rs) `GenesisConfig` here: https://github.com/paritytech/substrate/blob/efe7add76acc57f9d47f526ab09034c89e0f8c73/core/test-runtime/client/src/lib.rs#L98-L104

I think we want to remove this `GenesisConfig` and use the one that `decl_storage!` generates with `config()`'d fields, but I'm not entirely sure. All the fields are different apart from `authorities` and I'm not sure about what to do about: https://github.com/paritytech/substrate/blob/71b2f209cf759ef18bd73c5eb8311e614dde71c5/core/test-runtime/src/genesismap.rs#L60-L85

expenses:
I restored the `storage_items.rs` tests by the way.

bkchr:
`decl_storage!` generates a `GenesisConfig` as well. You need to add the `authorities` to this `GenesisConfig` and call `assimilate_storage` in `genesis_map`.

expenses:
@bkchr great, thanks!

bkchr:
Ty

PR #3951, opened 2019-10-29
Remove deprecated `Client::backend`

### LINKED ISSUE #2532: Refactor to not use client.backend & client.import_lock
With #2240 client's `fn backend(&self)` and `fn import_lock(&self)` have officially been deprecated (they were already meant for testing-only but used in other case). With that PR we are now seeing many more warnings because they were used more extensively than they were supposed to. 

We should refactor the code in the following two ways and remove these functions:

 - Code that uses `client.backend()` should rather directly be bound to `Backend` and receive this from `service`
 - Move the `import_lock` from client over to the `ImportQueue`, which means refactoring the way that import and execution happens, because the lock needs to happen around both.

expenses:
Closes #2532.

### DIFF
--- a/core/client/src/client.rs
+++ b/core/client/src/client.rs
@@ -342,15 +342,6 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 		self.backend.state_at(*block)
 	}
 
-	/// Expose backend reference. To be used in tests only
-	#[doc(hidden)]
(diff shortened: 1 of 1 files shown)


gnunicorn:
@expenses did you check whether this breaks polkadot? And send a PR if it does?

tomaka:
> did you check whether this breaks polkadot?

I can't find the string `backend()` or `backend(` in Polkadot's code, so it should be fine.

PR #3952, opened 2019-10-29
More robust punishment

gavofyork:
- Introduce option to alwaus for a new era
- Force offending validators to take manual action to avoid deselection

### DIFF
--- a/srml/staking/src/lib.rs
+++ b/srml/staking/src/lib.rs
@@ -1420,6 +1420,14 @@ impl<T: Trait> Module<T> {
 			}
 		});
 	}
+
+	/// Ensures that at the end of the current session there will be a new era.
+	fn ensure_new_era() {
+		match ForceEra::get() {
+			Forcing::ForceAlways | Forcing::ForceNew => (),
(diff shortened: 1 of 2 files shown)


srml/staking/src/lib.rs
```diff
@@ -1502,6 +1524,13 @@ impl <T: Trait> OnOffenceHandler<T::AccountId, session::historical::Identificati
 				continue
 			}
 
+			// Auto deselect validator on any offence and force a new era if they haven't previously
+			// been deselected.
+			if <Validators<T>>::exists(stash) {
```
tomusdrw: I think it would be better to move it after the `amount` check. We don't really slash anything if the amount is zero. Now we will deselect the validator even if they are not really slashed.

gavofyork:
> Doesn't it make them invulnerable to future punishments while still allowing them to affect the network?

No. It just removes them from being elected by default next session.

srml/staking/src/lib.rs
gavofyork: That's exactly the point.

> Force offending validators (*even those offenders whose slash was zero*) to take manual action to avoid deselection

PR #3953, opened 2019-10-29
Catch native panics when executing the wasm runtime

bkchr:
As with the native runtime, we now catch all native panics when we
execute the wasm runtime. The panics inside the wasm runtime were
already catched before by the wasm executor automatically, but any panic
in the host functions could bring down the node. The recent switch to
execute the native counterpart of the host function in `sr-io`, makes
this change required. The native `sr-io` functions just `panic` when
something is not provided or any other error occured.

### DIFF
--- a/core/executor/src/native_executor.rs
+++ b/core/executor/src/native_executor.rs
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::{result, cell::RefCell, panic::UnwindSafe};
+use std::{result, cell::RefCell, panic::{UnwindSafe, AssertUnwindSafe}};
 use crate::error::{Error, Result};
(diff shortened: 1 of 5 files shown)


core/executor/src/native_executor.rs
```diff
@@ -92,15 +92,45 @@ impl<D: NativeExecutionDispatch> NativeExecutor<D> {
...
-			let runtime = cache.fetch_runtime(ext, self.fallback_method, self.default_heap_pages)?;
-			f(runtime, ext)
+			let (runtime, code_hash) = cache.fetch_runtime(
+				ext,
+				self.fallback_method,
+				self.default_heap_pages,
+			)?;
+
+			let runtime = AssertUnwindSafe(runtime);
+			let ext = AssertUnwindSafe(ext);
```
tomaka: cc #3036

This is not a blocker for this PR, but doing it properly would basically require refactoring half of Substrate.

bkchr: Yeah.

PR #3954, opened 2019-10-29
Remove the `exit` parameter from importing and exporting blocks

tomaka:
Small refactoring.

Importing and exporting blocks are now entirely futures, and we can therefore remove the `exit` parameter. If they want to interrupt an ongoing import or export, the user can simply stop polling the corresponding future instead.

This only breaks the API of the service builder, but **not** the API of substrate-cli. Hence no Polkadot PR is necessary.
I kept the API of substrate-cli the way it is, because I'm not sure how I would accomodate the changes for the `Worker` trait of Polkadot.


seunlanlege:
this is pretty cool :sunglasses:, needs a rebase

tomaka:
Fixed the conflicts

client/cli/src/lib.rs
```diff
@@ -388,7 +388,25 @@ impl<'a> ParseAndPrepareExport<'a> {
 			None => Box::new(stdout()),
 		};
 
-		builder(config)?.export_blocks(exit.into_exit(), file, from.into(), to.map(Into::into), json)?;
+		// Note: while we would like the user to handle the exit themselves, we handle it here
+		// for backwards compatibility reasons.
```
bkchr: What kind of backwards compatibility? To what?

tomaka: To keep the API of `substrate-cli` backwards-compatible.
"We would like the user [of substrate-cli] to handle the exit"

bkchr: And when do you plan to let the user handle this?

tomaka: If we let the user handle this, then we'd need to propagate the changes to Polkadot, which in turns would require a refactoring of `Worker` trait first, which itself is more complicated than it looks because it requires passing a `&Service`.

PR #3955, opened 2019-10-29
*: Refactor authority discovery (key mngmt, runtime API)

mxinden:
#3914 disabled the *authority discovery* module due to issues with the session
key management (see pull request description for details). This pull requests
addresses theses issues and thus re-enables the *authority discovery* module. It
does the following:

- Before disabling the *authority discovery* module on *master* (#3914), it was
  wrongly configured as a session change handler without introducing its own
  session keys.

  This pull requests **reintroduces the *srml authority discovery* module** as a
  session handler. #3949 made the mistake described above impossible.

- Previously the *authority discovery* module was ment to piggyback on another
  session key type to keep the overall amount of keys low. This is a premature
  optimization.

  This pull request **introduces specific app crypto** (public, pair and
  signature) types for the *authority discovery* module instead.

- Previously the *authority discovery* module would do all crypto related
  operations (sign, verify) through runtime APIs, thereby one would be able to
  change the crypto type with a runtime upgrade, given that anything within
  `/core` would treat it as an opaque blob. This is a stretch goal, but not
  worth the complexity as of today. In addition with the first change described
  above, the *authority discovery* module depends on specific crypto key types,
  preventing a crypto type change with a runtime upgrade in the first place.

  This pulll request **removes the crypto related runtime APIs** and uses a
  pointer to the Substrate key store instead.

- (This pull request delays the initial operations of the *authority discovery*
  module to give the libp2p Kademlia module time to bootstrap own and other
  nodes' k-buckets.)

---

This pull request still includes a couple of `TODO` sections wherever I am
looking for input from reviewers. In addition, given that this pull request does
include key management and given that I am not very familiar with this critical
subsystem, it would be great to have someone more familiar give this a very
thorough review.


mxinden:
This pull request is ready for review, but still has a couple of `TODO` comments which I would like to discuss with the reviewers. So please do not merge quite yet.

srml/authority-discovery/src/lib.rs
```diff
@@ -237,126 +186,67 @@ mod tests {
...
 
-		GenesisConfig::<Test> { keys: keys }
-			.assimilate_storage(&mut t)
-			.unwrap();
-
-		// Create externalities.
-		let mut externalities = TestExternalities::new(t);
-		externalities.register_extension(KeystoreExt(key_store));
+		// Needed for `session::OneSessionHandler::on_new_session`.
+		let first_authorities_and_account_ids: Vec<(&AuthorityId, AuthorityId)>  = first_authorities.clone()
```
bkchr:
```suggestion
		let first_authorities_and_account_ids: Vec<(&AuthorityId, AuthorityId)> = first_authorities.clone()
```

node/cli/src/chain_spec.rs
```diff
@@ -220,7 +246,8 @@ pub fn testnet_genesis(
 		}),
 		session: Some(SessionConfig {
 			keys: initial_authorities.iter().map(|x| {
-				(x.0.clone(), session_keys(x.2.clone(), x.3.clone(), x.4.clone()))
+				// TODO: Again, is this safe to do?
```
bkchr: What is safe?

node/cli/src/chain_spec.rs
```diff
@@ -61,8 +62,13 @@ pub fn flaming_fir_config() -> Result<ChainSpec, String> {
...
 
-fn session_keys(grandpa: GrandpaId, babe: BabeId, im_online: ImOnlineId) -> SessionKeys {
-	SessionKeys { grandpa, babe, im_online, }
+fn session_keys(
+	grandpa: GrandpaId,
+	babe: BabeId,
+	im_online: ImOnlineId,
+	authority_discovery: AuthorityDiscoveryId,
+) -> SessionKeys {
+	SessionKeys { grandpa, babe, im_online, authority_discovery}
```
bkchr:
```suggestion
	SessionKeys { grandpa, babe, im_online, authority_discovery }
```

core/primitives/src/crypto.rs
```diff
@@ -897,6 +897,8 @@ pub mod key_types {
 	pub const AURA: KeyTypeId = KeyTypeId(*b"aura");
 	/// Key type for ImOnline module, built-in.
 	pub const IM_ONLINE: KeyTypeId = KeyTypeId(*b"imon");
+	/// Key type for AuthorityDiscovery module, built-in.
+	pub const AUTHORITY_DISCOVERY: KeyTypeId = KeyTypeId(*b"audi");
```
bkchr: image

core/authority-discovery/src/lib.rs
```diff
@@ -244,22 +260,18 @@ where
...
-
-			let is_verified = self
-				.client
-				.runtime_api()
-				.verify(&id, &addresses, &signature, &authority_id.clone())
-				.map_err(Error::CallingRuntime)?;
+			} = schema::SignedAuthorityAddresses::decode(value).map_err(Error::DecodingProto)?;
+			// TODO: Should we rather have a `TryFrom<&[u8]> for Signature` instead of encoding a scale encoded
+			// signature as Protobuf?
+			let signature = AuthoritySignature::decode(&mut &signature[..]).map_err(Error::EncodingDecodingScale)?;
```
bkchr: Why not use SCALE here?

core/authority-discovery/primitives/src/lib.rs
```diff
@@ -20,31 +20,28 @@
...
-use sr_primitives::RuntimeDebug;
 
-#[derive(codec::Encode, codec::Decode, Eq, PartialEq, Clone, RuntimeDebug)]
-#[cfg_attr(feature = "std", derive(Hash))]
-pub struct Signature(pub Vec<u8>);
-#[derive(codec::Encode, codec::Decode, Eq, PartialEq, Clone, RuntimeDebug)]
-#[cfg_attr(feature = "std", derive(Hash))]
-pub struct AuthorityId(pub Vec<u8>);
+mod app {
+	use app_crypto::{app_crypto, key_types::AUTHORITY_DISCOVERY, sr25519};
```
bkchr: Hmm, I think we should follow `im-online` and provide support for `sr25519` and `ed25519`. The module probably does not have a hard requirement on any crypto.

core/primitives/src/crypto.rs
mxinden: ... No, *listen* in Latin @bkchr :P

bkchr: Okay :P

node/cli/src/chain_spec.rs
mxinden: This comment is related to the comment above asking whether it is safe to use the same seed for `BabeId`, `ImOnlineId` and `AuthorityDiscoveryId`. I guess given that this is only ever used as a  testnetwork it does not matter @bkchr?

core/authority-discovery/primitives/src/lib.rs
mxinden: Right now I am using the `BareCryptoStore` trait to access the keystore within the authority discovery module. The trait only offers specific functions for sr25519 and ed25519, which I could not use if we make the crypto type generic. Are you suggesting to access the `Store` directly and use `key_pair` to retrieve the key pair?

gavofyork: i don't think it's necessary at this stage, and it wouldn't be the only module that supports a specific type of crypto.

node/cli/src/chain_spec.rs
```diff
@@ -116,6 +131,9 @@ fn staging_testnet_config_genesis() -> GenesisConfig {
 		hex!["00299981a2b92f878baaf5dbeba5c18d4e70f2a1fcd9c61b32ea18daf38f4378"].unchecked_into(),
 		// 5C4vDQxA8LTck2xJEy4Yg1hM9qjDt4LvTQaMo4Y8ne43aU6x
 		hex!["00299981a2b92f878baaf5dbeba5c18d4e70f2a1fcd9c61b32ea18daf38f4378"].unchecked_into(),
+		// TODO: Is it safe to just copy like ImOnlineId did from BabeId?
```
gavofyork: should be fine.

node/cli/src/service.rs
```diff
@@ -174,6 +179,18 @@ macro_rules! new_full {
 
 			let babe = babe::start_babe(babe_config)?;
 			service.spawn_essential_task(babe);
+
+			// TODO: Safe to unwrap here?
```
gavofyork: never.

core/authority-discovery/src/lib.rs
mxinden: Friendly ping @bkchr.

srml/authority-discovery/src/lib.rs
```diff
@@ -17,34 +17,25 @@
...
-pub trait Trait: system::Trait + session::Trait {
-	type AuthorityId: RuntimeAppPublic + Default + FullCodec + PartialEq;
-}
+pub trait Trait: system::Trait + session::Trait {}
 
 decl_storage! {
 	trait Store for Module<T: Trait> as AuthorityDiscovery {
-		/// The current set of keys that may issue a heartbeat.
-		Keys get(fn keys): Vec<T::AuthorityId>;
+		// Keys of the current authority set.
```
4meta5:
```suggestion
		/// Keys of the current authority set.
```

core/authority-discovery/src/lib.rs
bkchr: Sounds reasonable with the `[u8]`.

mxinden: After a private conversation with @bkchr, we settled on first SCALE and then Protobuf encoding as it is implemented right now. Feel free to revive the discussion.

srml/authority-discovery/src/lib.rs
```diff
@@ -55,79 +46,40 @@ decl_module! {
...
-	/// Sign the given payload with the private key corresponding to the given authority id.
-	pub fn sign(
-		payload: &Vec<u8>,
-	) -> Option<(
-		<<T as Trait>::AuthorityId as RuntimeAppPublic>::Signature,
-		T::AuthorityId,
-	)> {
-		let authority_id = Module::<T>::authority_id()?;
-		authority_id.sign(payload).map(|s| (s, authority_id))
+	pub fn authorities() -> Vec<AuthorityId> {
```
shawntabrizi: This function and `Self::keys()` do the same thing. Do you need both?

Could you just do `get(fn authorities)` in the storage above?

srml/authority-discovery/src/lib.rs
```diff
@@ -55,79 +46,40 @@ decl_module! {
...
-	) -> bool {
-		authority_id.verify(payload, &signature)
-	}
-
-	fn initialize_keys(keys: &[T::AuthorityId]) {
+	fn initialize_keys(keys: &[AuthorityId]) {
 		if !keys.is_empty() {
-			assert!(Keys::<T>::get().is_empty(), "Keys are already initialized!");
-			Keys::<T>::put(keys);
+			assert!(Keys::get().is_empty(), "Keys are already initialized!");
```
shawntabrizi:
```suggestion
			assert!(Keys::exists(), "Keys are already initialized!");
```

Does `Keys::exists()` work for your needs here?

srml/authority-discovery/src/lib.rs
```diff
@@ -55,79 +46,40 @@ decl_module! {
...
 	where
 		I: Iterator<Item = (&'a T::AccountId, Self::Key)>,
 	{
-		let keys = validators.map(|x| x.1).collect::<Vec<_>>();
+		let keys = authorities.map(|x| x.1).collect::<Vec<_>>();
 		Self::initialize_keys(&keys);
 	}
 
-	fn on_new_session<'a, I: 'a>(_changed: bool, _validators: I, next_validators: I)
+	fn on_new_session<'a, I: 'a>(_changed: bool, _authorities: I, next_authorities: I)
```
shawntabrizi: Is there a runtime event fired elsewhere when this is triggered? If not, adding an event in this module would make sense to me.

srml/authority-discovery/src/lib.rs
mxinden: I am fine with either way. My reasoning of introducing of using a function here is to not expose any internal implementation details. Let's say that we would like to keep track of the current *authority discovery* authority set in the future, it would only need a change within this module and not a change outside of it. Does that make sense @shawntabrizi?

shawntabrizi: I see. So there is a world where maybe `authorities` returns something different than the strict storage of `Keys`.

It makes sense.

srml/authority-discovery/src/lib.rs
mxinden: Can you expand on this @shawntabrizi? I don't understand what that event would do.

srml/authority-discovery/src/lib.rs
mxinden: So far I have tried to be consistent with https://github.com/paritytech/substrate/blob/master/srml/im-online/src/lib.rs#L506. In case we want to use `exists` I am happy to do that, but rather as a follow up in both modules at once.

srml/authority-discovery/src/lib.rs
shawntabrizi: There would be an event which is emitted from the runtime module when the authorities change, probably with the list of new authorities.

It would be a way to notify the outside world that this change has occured, for example to block explorers or front-ends like Polkadot-JS Apps.

It may not make sense in this context depending on how the authorities are changed. For example, if another module is expected to trigger this change, and that module emits an event, then you probably do not need to emit one yourself, but I wanted to bring it up so you make that judgement.

mxinden: The logic in this pull request follows a *pull approach*, having the *core* module pull the current authority set on an interval from the *srml* module. Thus this event would not be needed within the actual logic.

In regards to notifying the outside applications of the authority set change via this event: I doubt this is useful, as the *authority discovery* authority set is the same as the *Babe* (or *Grandpa* with some caveats) set at any given point in time.

Does this explanation make sense?

mxinden:
@bkchr @gavofyork @shawntabrizi thanks for the reviews! Do you have further comments?

srml/authority-discovery/src/lib.rs
shawntabrizi: 👍

srml/authority-discovery/src/lib.rs
bkchr: `exits()` is not correct here. We could use `decode_len()`, but that makes no real difference here, as this is only called at genesis. `Keys` will always be initialized by the `GenesisConfig`, but we need to make sure that we did not set them to something before. So, we check if the set keys are empty.

srml/authority-discovery/src/lib.rs
```diff
@@ -55,79 +46,40 @@ decl_module! {
...
 	where
 		I: Iterator<Item = (&'a T::AccountId, Self::Key)>,
 	{
 		// Remember who the authorities are for the new session.
-		Keys::<T>::put(next_validators.map(|x| x.1).collect::<Vec<_>>());
+		Keys::put(next_authorities.map(|x| x.1).collect::<Vec<_>>());
```
bkchr:
```suggestion
		if changed {
			Keys::put(authorities.map(|x| x.1).collect::<Vec<_>>());
		}
```

I think we need to use here the `authorities`. You probably don't want to connect to the authorities of the next session?

srml/authority-discovery/src/lib.rs
bkchr:
```suggestion
	fn on_new_session<'a, I: 'a>(_changed: bool, authorities: I, _next_authorities: I)
```

core/authority-discovery/src/lib.rs
```diff
@@ -112,23 +123,31 @@ where
 	pub fn new(
 		client: Arc<Client>,
 		network: Arc<Network>,
-		dht_event_rx: Receiver<DhtEvent>,
+		key_store: BareCryptoStorePtr,
+		dht_event_rx: Pin<Box<dyn Stream<Item = DhtEvent> + Send>>,
 	) -> Self {
 		// Kademlia's default time-to-live for Dht records is 36h, republishing records every 24h. Given that a node
```
bkchr: Not perfect to depend on any "default value" that probably could change at any moment.

core/authority-discovery/src/lib.rs
```diff
@@ -296,6 +304,44 @@ where
 		self.address_cache
 			.retain(|peer_id, _addresses| current_authorities.contains(peer_id))
 	}
+
+	fn get_priv_keys_within_authority_set(&mut self) -> Result<Vec<AuthorityPair>> {
```
gavofyork: this is a non-trivial function, so it needs some docs.

core/authority-discovery/src/lib.rs
```diff
@@ -296,6 +304,44 @@ where
...
+			.filter_map(|pub_key| {
+				self.key_store.read().sr25519_key_pair(key_types::AUTHORITY_DISCOVERY, &pub_key)
+			})
+			.map(std::convert::Into::into)
+			.collect();
+
+		Ok(keys)
+	}
+
+	fn get_own_public_keys_within_authority_set(&mut self) -> Result<HashSet<AuthorityId>> {
```
gavofyork: same

core/authority-discovery/src/lib.rs
mxinden: Docs added in https://github.com/paritytech/substrate/pull/3955/commits/89f501888c87e5fabeead46ed6e3bdf09d3410c2.

core/authority-discovery/src/lib.rs
mxinden: Docs added in https://github.com/paritytech/substrate/pull/3955/commits/89f501888c87e5fabeead46ed6e3bdf09d3410c2.

core/authority-discovery/src/lib.rs
mxinden: Right. Kademlia does not expose these values to the outside as of today (https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/lib.rs). If we want to do that in the future I would prefer doing that in a follow up pull request.

PR #3956, opened 2019-10-29
grandpa: fix handling of catch-up requests

andresilva:
GRANDPA has a protocol for requesting data from its peers in order to catch up to the latest round. Currently, we restrict these requests to other authorities, assuming that non-authorities are running the GRANDPA observer protocol which does not have the data available to answer catch up requests.

With the current setup for sentry nodes, the validator is only connected to non-authority nodes (the sentries), and therefore if it ever gets behind on the round it cannot issue catch-up requests to anyone. Also, the assumption that only authorities can answer catch-up requests may be incorrect depending on the network service setup.

This PR makes this configurable, so that on Kusama where the GRANDPA observer protocol isn't currently enabled, we can issue catch-up requests to all full nodes rather than just authorities.

### DIFF
--- a/core/finality-grandpa/src/communication/gossip.rs
+++ b/core/finality-grandpa/src/communication/gossip.rs
@@ -1868,7 +1868,7 @@ mod tests {
 	}
 
 	#[test]
-	fn doesnt_send_catch_up_requests_to_non_authorities() {
+	fn doesnt_send_catch_up_requests_to_non_authorities_when_observer_enabled() {
 		let (val, _) = GossipValidator::<Block>::new(
 			config(),
 			voter_set_state(),
--- a/core/finality-grandpa/src/communication/gossip.rs
+++ b/core/finality-grandpa/src/communication/gossip.rs
@@ -512,6 +512,40 @@ enum PendingCatchUp {
 	},
 }
 
+/// Configuration for the round catch-up mechanism.
+enum CatchUpConfig {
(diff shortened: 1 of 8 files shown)


core/finality-grandpa/src/communication/gossip.rs
```diff
@@ -512,6 +512,40 @@ enum PendingCatchUp {
...
+/// Configuration for the round catch-up mechanism.
+enum CatchUpConfig {
+	/// Catch requests are enabled, our node will issue them whenever it sees a
+	/// neighbor packet for a round further than `CATCH_UP_THRESHOLD`. If
+	/// `only_from_authorities` is set, the node will only send catch-up
+	/// requests to other authorities it is connected to. This is useful if the
+	/// GRANDPA observer protocol is live on the network, in which case full
+	/// nodes (non-authorities) don't have the necessary round data to answer
+	/// catch-up requests.
+	Enabled { only_from_authorities: bool },
```
mxinden: In the *sentry* node case, one would set `only_from_authorities` to `false` for validators only publicly routable through *sentry* nodes?

andresilva: This is more related to when the observer protocol is enabled (in which case we need to make the distinction). After we introduce `--sentry`, the nodes will announce themselves as authorities threfore the private/validator nodes will be able to request from the sentries (even when the observer protocol is enabled and this is therefore set to `true`).

PR #3957, opened 2019-10-29
Allow passing a custom database when creating the Service

tomaka:
Relevant for #2416

Right now the `substrate-client-db` crate has a `kvdb-rocksdb` compile-time feature. When enabled (which is the case by default), we use RocksDB. When disabled (which is the case for the in-browser experiments), we print a warning and use the memorydb instead.

This pull request modifies this mechanic.

In both `substrate_service::config::Configuration` and `substrate_client_db::DatabaseSettings` it is now possible to pass either a path to a database, in which case it will be opened using RocksDB, or a custom implementation of the `kvdb::KeyvalueDB` trait. The in-browser node would manually open a database (using the browser's local-storage) and pass it at initialization.

The `kvdb-rocksdb` feature still exists in `substrate_client_db`. If it is disabled, we return an error at initialization in the case of a path to a database. This is safer than just printing a warning. Another option could be to simply remove at compile-time the option to pass a path, rather than producing a runtime error, but from my experience putting items behind Cargo features usually becomes a source of annoying compilation errors.

Thanks to this change, I also slightly cleaned up `substrate_client_db::Backend::new_test`.

### DIFF
--- a/core/cli/src/lib.rs
+++ b/core/cli/src/lib.rs
@@ -28,7 +28,7 @@ pub mod informant;
 
 use client::ExecutionStrategies;
 use service::{
-	config::{Configuration, ConfigurationDb},
+	config::{Configuration, DatabaseConfig},
 	ServiceBuilderExport, ServiceBuilderImport, ServiceBuilderRevert,
 	RuntimeGenesis, ChainSpecExtension, PruningMode, ChainSpec,
 };
@@ -340,7 +340,7 @@ impl<'a> ParseAndPrepareExport<'a> {
 	{
 		let config = create_config_with_db_path(spec_factory, &self.params.shared_params, self.version)?;
 
-		if let ConfigurationDb::Path { ref path, .. } = &config.database {
+		if let DatabaseConfig::Path { ref path, .. } = &config.database {
 			info!("DB path: {}", path.display());
 		}
 		let from = self.params.from.unwrap_or(1);
@@ -422,9 +422,9 @@ impl<'a> ParseAndPreparePurge<'a> {
 			spec_factory, &self.params.shared_params, self.version
 		)?;
 		let db_path = match config.database {
-			ConfigurationDb::Path { path, .. } => path,
+			DatabaseConfig::Path { path, .. } => path,
 			_ => {
-				println!("Cannot purge custom database implementation");
+				eprintln!("Cannot purge custom database implementation");
 				return Ok(());
 			}
 		};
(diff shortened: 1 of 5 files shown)


core/cli/src/lib.rs
```diff
@@ -419,7 +421,13 @@ impl<'a> ParseAndPreparePurge<'a> {
 		let config = create_config_with_db_path::<(), _, _, _>(
 			spec_factory, &self.params.shared_params, self.version
 		)?;
-		let db_path = config.database_path;
+		let db_path = match config.database {
+			ConfigurationDb::Path { path, .. } => path,
+			_ => {
+				println!("Cannot purge custom database implementation");
```
arkpar: I'd use `error!` for this

core/service/src/config.rs
```diff
@@ -96,6 +94,23 @@ pub struct Configuration<C, G, E = NoExtension> {
 	pub dev_key_seed: Option<String>,
 }
 
+/// Configuration of the database of the client.
+#[derive(Clone)]
+pub enum ConfigurationDb {
```
arkpar: `DbConfiguration` or even `DatabaseConfig` would match our coding guide better

arkpar:
LG in general, minor grumbles

core/client/Cargo.toml
```diff
@@ -32,6 +32,7 @@ header-metadata = { package = "substrate-header-metadata", path = "header-metada
 [dev-dependencies]
 env_logger = "0.7.0"
 tempfile = "3.1.0"
+client-db = { package = "substrate-client-db", path = "./db", features = ["kvdb-rocksdb"] }
```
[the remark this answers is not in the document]
tomaka: I added this in order to force the `kvdb-rocksdb` feature, otherwise the test fails.
Alternatively, I could maybe change the test to use an in-memory database.

core/cli/src/lib.rs
tomaka: I went for `println!` because this is a CLI function, and the rest of the function also uses `println!` several times.

arkpar: Normally errors should be printed to stderr.

core/client/db/src/lib.rs
```diff
@@ -191,16 +194,29 @@ impl<B: BlockT> StateBackend<Blake2Hasher> for RefTrackingState<B> {
...
+pub enum DatabaseSettingsSrc {
+	/// Load a database from a given path. Recommended for most uses.
+	Path {
+		/// Path to the database.
+		path: PathBuf,
+		/// Cache size in bytes. If `None` default is used.
+		cache_size: Option<usize>,
+	},
+
+	/// Use a custom already-open database. Recommended only for testing purposes, or in
```
bkchr: Minor minor nitpick. Could you just remove the everything after the first sentence?

This is exactly what I need for Cumulus and this does not fit into one of these categories ;)

PR #3959, opened 2019-10-29
node: add sentry mode flag

andresilva:
This PR adds a `--sentry` flag to the node. The node is started as a validator (i.e. authority role, pruning restrictions, runs full GRANDPA protocol, etc.), but no keystore is passed to any consensus module so that the node **cannot** actively participate in it.

### DIFF
--- a/core/cli/src/lib.rs
+++ b/core/cli/src/lib.rs
@@ -236,6 +236,17 @@ where
 	}
 }
 
+/// Returns a string displaying the node role, special casing the sentry mode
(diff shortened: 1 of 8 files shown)


iammelea:
> This PR adds a `--sentry` flag to the node. The node is started as a validator (i.e. authority role, pruning restrictions, runs full GRANDPA protocol, etc.), but no keystore is passed to any consensus module so that the node **cannot** actively participate in it.

I believe Some people is missunderstanding this text.
So clarify.
Sentry flag need prunning archive option too for work in private with validators and Finnish blocks fine.

PR #3915, opened 2019-10-25
Check given Grandpa validator set against set found in storage

HCastano:

Issue #2053, opened 2019-03-20
Log2 Ancestors Module


rphmeier:
The idea behind log2 ancestors is to keep a record of some block hashes in the history of the chain.

```rust
fn initialise() {
    let number = current_number() - 1;
    let parent_hash = parent_hash();

    let mut index = 0;
    let mut working_number = number;
    while index < 32 {
        Ancestors::put(index, (number, parent_hash)); // storage
        if working_number % 2 != 0 { break }
        working_number /= 2;
        index += 1;
    }

    // post-condition: Ancestors(index) stores the hash and
    // number of the last block whose number was divisible by 2^index.
}
```

We can use this module to safely skip backwards in the finalized chain.

jimpo:
If I understand correctly, the ancestry proofs would consist of a mix of ~log (m - n) headers and storage proofs where `m` is the height of the chain and `n` is the height of the ancestor.

Like say the chain height is 17 and you want to prove inclusion of block 5. Then the proof is:

- Block 16 header + storage proof with respect to block 17 state root
- Block 8 header + storage proof with respect to block 16 state root
- Block 6 header + storage proof with respect to block 8 state root
- Block 5 header + storage proof with respect to block 6 state root

This is pretty much the same structure as a [Merkle Mountain Range](https://github.com/mimblewimble/grin/blob/master/doc/mmr.md), but more indirect. With a Merkle mountain range, instead of just storing the block hash at the indices, you store the root of a full binary Merkle tree to the previous power of two index. Thus the inclusion proof for the above scenario would be:

- Merkle root over all blocks 1-16 + storage proof with respect to block 17 state root
- Regular merkle branch proving inclusion of header in the above merkle root

This has the same nice property that ancestry proofs for recent blocks are short. The code is barely any more complicated, but does involve one hash evaluation for each execution of the inner loop, which is expected to be constant and log(n) in the worst case.

BUT with the MMR approach, nodes would need separate storage for the non-pruned part of the header tree. This doesn't feel that different from the original proposal, however, where only nodes running in archive mode can generate proofs.

rphmeier:
The MMR approach seems to require a lot of substrate storage changes and other implementation effort (to avoid duplication - I can't see it being efficient to embed an MMR in the regular state trie).

Whereas the log2-ancestors thing gets us most of the way there with barely any implementation overhead.

jimpo:
[...]

PR #3963, opened 2019-10-30
Verify Ancestry between Headers

HCastano:
Adds an `AncestryProofChecker` that is used to verify whether a given `child` header is an ancestor of a given `ancestor` header. It does this by working its way through a hash map of `Hashes => Headers` which _should_ contain all the headers between `child` and `ancestor`. In the future I think it's possible to change this method to use something like the Log2 Ancestry proposed by Rob in #2053.

One thing I've been thinking about that I don't know the answer to is how to ensure that somebody doesn't send a proof that contains an arbitrarily large number of headers. This would be bad since we're checking every header in the proof until we find something what we're looking for, or run out of headers.

The diff here isn't totally accurate, since it's built on top of #3915 which is still waiting to go in. The only thing that was actually added here was `srml/bridge/src/ancestry_proof.rs`.


srml/bridge/src/ancestry_proof.rs
```diff
@@ -0,0 +1,164 @@
...
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+
+//! A way to check whether or not two headers are related to eachother.
```
jimpo: typo: eachother

srml/bridge/src/ancestry_proof.rs
```diff
@@ -0,0 +1,164 @@
...
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
```
jimpo: nit: remove newline

srml/bridge/src/ancestry_proof.rs
```diff
@@ -0,0 +1,164 @@
...
+
+//! A way to check whether or not two headers are related to eachother.
+
+use crate::error::Error;
+
+// Question: Since this is from std, am I allowed to use it?
```
jimpo: No. If you needed a map you can use `rstd::collections::btree_map::BTreeMap`. But you shouldn't need a map here.

srml/bridge/src/ancestry_proof.rs
```diff
@@ -0,0 +1,164 @@
...
+use std::collections::HashMap;
+use sr_primitives::traits::Header;
+
+/// A struct used to check whether or not two headers
+/// are related to one another.
+pub struct AncestryProofChecker<H: Header> {
```
jimpo: Does this need to be a struct? I feel like a simple `verify_ancestry` function is fine.

srml/bridge/src/ancestry_proof.rs
```diff
@@ -0,0 +1,164 @@
...
+	proof: HashMap<H::Hash, H>,
```
jimpo: Why can't the proof just be a `Vec` of headers? It's up to the prover to ensure they are in order.

srml/bridge/src/lib.rs
```diff
@@ -160,6 +161,39 @@ impl<T: Trait> Module<T> {
 	}
 }
 
+// A naive way to check whether a `child` header is an ancestor
+// of an `ancestor` header. For this it requires a proof which
+// is a header chain, This could be updated to use something like
+// Log2 Ancestors (#2053) in the future.
+fn verify_ancestry<H>(proof: Vec<H>, ancestor: H, child: H) -> std::result::Result<(), Error>
```
jimpo: nit: Why does this need to be a `std::result::Result` and not just `Result`?

srml/bridge/src/lib.rs
```diff
@@ -160,6 +161,39 @@ impl<T: Trait> Module<T> {
...
+where
+	H: Header
+{
+	let mut curr_header = &proof[0];
+	if curr_header.hash() != child.hash() {
+		return Err(Error::AncestorNotFound);
```
jimpo: I'd make the error `InvalidAncestryProof`. I think that reflects better that it is unknown whether one is an ancestor of the other, simply that we can't check because the proof is invalid.

srml/bridge/src/lib.rs
```diff
@@ -160,6 +161,39 @@ impl<T: Trait> Module<T> {
...
+	}
+
+	let mut parent_hash = curr_header.parent_hash();
+
+	// If we find that the header's parent hash matches our ancestor's hash we're done
+	for i in 1..proof.len() {
```
jimpo: Iterate over the headers directly, not the indices.

`for header in &proof[1..].iter()`

srml/bridge/src/lib.rs
```diff
@@ -160,6 +161,39 @@ impl<T: Trait> Module<T> {
...
+		curr_header = &proof[i];
+
+		// Need to check that blocks are actually related
+		if curr_header.hash() != *parent_hash {
+			break;
+		}
+
+		parent_hash = curr_header.parent_hash();
+		if *parent_hash == ancestor.hash() {
```
jimpo: If this is the only thing `ancestor` is used for, make the function parameter `ancestor_hash` instead of `ancestor` (header).

srml/bridge/src/lib.rs
```diff
@@ -160,6 +161,39 @@ impl<T: Trait> Module<T> {
 	}
 }
 
+// A naive way to check whether a `child` header is an ancestor
```
jimpo: `child` header is a *descedent* of ...

srml/bridge/src/lib.rs
```diff
@@ -160,6 +161,39 @@ impl<T: Trait> Module<T> {
...
+// of an `ancestor` header. For this it requires a proof which
+// is a header chain, This could be updated to use something like
+// Log2 Ancestors (#2053) in the future.
+fn verify_ancestry<H>(proof: Vec<H>, ancestor: H, child: H) -> std::result::Result<(), Error>
+where
+	H: Header
+{
+	let mut curr_header = &proof[0];
+	if curr_header.hash() != child.hash() {
```
jimpo: The proof shouldn't contain the child header -- that is redundant. It should be only the headers *between* ancestor and descendent.

srml/bridge/src/lib.rs
```diff
@@ -303,4 +337,90 @@ mod tests {
 			assert_eq!(MockBridge::num_bridges(), 1);
 		});
 	}
+
+	fn get_related_block_headers() -> (Header, Header, Header) {
```
jimpo: A helper function like `fn build_header_chain(first_parent_hash: HashOutput, len: usize) -> Vec<Header>` would probably make tests more flexible and concise.

srml/bridge/src/lib.rs
HCastano: In another case I was having an issue where it was using a Substrate `Result` instead of an `std::Result` which is what I wanted. Looks like the import is gone now though.

HCastano:
Sorry for those force pushes, my rebase didn't go as smoothly as I expected :P

Issue #3419, opened 2019-08-16
Extend `SignedExtension` to include `Origin` specification


tomusdrw:
Currently the `Origin` has a special meaning in the code and signed and unsigned transactions follow different paths in the validation and execution phase.

We could attempt to unify them and encapsulate all the logic in `Extra`. So that `SignedExtension`s would be allowed to return additional `Origin` information that would then be passed to `dispatch`.

That would make `ValidateUnsigned` and all `_unsigned` functions of `SignedExtension` obsolete, since both signed and unsigned extrinsics would follow the same path.

Possible simplifications in:
1. `im-online` - we introduce another `SignedExtra` which would check the signature inside the payload (not the signature of the transaction, since it's an unsigned transaction) and provide a special `Origin` containing the `AuthorityId` that signed that payload.
2. `claims` - details below
3. Any module using `ValidateUnsigned` trait or `SignedExtension::validate_unsigned`.
4. Clean and elegant `srml::executive` validation and dispatch.

Original convo:
> we could, i guess, extend SignedExtension API to include Origin-specification.
though it's a fairly fundamental change and i'm not sure how aggregation could work
but if we did it, then claims module would introduce a new Origin::EthereumClaimant(EthereumAddress)
and return that from its SignedExtension's validate (or pre_dispatch)
but there's a lot of interplay that would need to be figured out.
signed and unsigned are currently handled very differently.

bkchr:
I don't think that we should merge `validate_unsigned` and `validate`. The separation actually gives us the ability to deny any unsigned transaction for that we don't have a signed extension registered. This is important, as the programmer can easily forget to include a `SignedExtension`.

tomusdrw:
I imagine that we would disallow all transactions by default and `SignedExtensions` `validate` function would return the origin only if it "whitelists" a particular transaction. So that means that the most basic `SignedExtension` would just return a `Origin::Signed(x)` (after checking the signature) and that would get aggregated with other extensions. Unsigned would only be pass through if one of the extensions returns `Origin::None` (which means it does authorize the call).

I still think we would need one `Extension` that is able to just delegate validation to something similar to `ValidateUnsigned`. Having every pallet expose a separate extension that needs to be added to the list of `Extra` seems like an overkill.

bkchr:
[...]

PR #3972, opened 2019-10-30
Increase priority of ImOnline heartbeats

tomusdrw:
Currently the `priority` is set to `0` which means it has the lowest possible value in the transaction pool.

We have a notion of `Operational` transactions (`DispatchClass`), however it only applies to transactions with `CheckWeight` `SignedExtension`. Since heartbeat is actually an `UnsignedTransaction` it's currently not using this mechanism (see #3419)

### DIFF
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -82,7 +82,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 190,
-	impl_version: 190,
+	impl_version: 191,
 	apis: RUNTIME_API_VERSIONS,
 };
 
--- a/srml/im-online/src/lib.rs
+++ b/srml/im-online/src/lib.rs
@@ -80,6 +80,7 @@ use sr_primitives::{
 	traits::{Convert, Member, Printable, Saturating}, Perbill,
 	transaction_validity::{
 		TransactionValidity, TransactionLongevity, ValidTransaction, InvalidTransaction,
+		TransactionPriority,
 	},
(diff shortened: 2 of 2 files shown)


node/runtime/src/lib.rs
```diff
@@ -82,7 +82,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 190,
-	impl_version: 190,
+	impl_version: 191,
```
pepyakin: Since this change actually alters behavior of the runtime it looks like the `spec_version` should be bumped as well.

srml/im-online/src/lib.rs
```diff
@@ -532,7 +533,7 @@ impl<T: Trait> support::unsigned::ValidateUnsigned for Module<T> {
 			}
 
 			Ok(ValidTransaction {
-				priority: 0,
+				priority: TransactionPriority::max_value(),
```
rphmeier: is there anything to stop validators from issuing _too many_ heartbeats?

pepyakin: Had the same question. My second thought was that this might be a common problem among operational transactions.

tomusdrw: Since `provides` tag is `(current_session, authority_id)` there can exist only one such transaction in the pool. The authority could spam transactions from different sessions (or from different `authority_id`) but they should be rejected through validation. So there can only be as much heartbeats in the pool as there is authorities. If the heartbeat is included in block it should invalidate duplicates from being included.

rphmeier: OK, thanks for the clarification.

PR #3974, opened 2019-10-30
Add the code for compiling node-cli for WASM-browser

tomaka:
cc #2416 

Adds the necessary code to make node-cli compile for the browser.
See the `README.md` in `browser-demo` for instructions.

Note that this doesn't work yet. The node compiles and can start, but panics at runtime most notably because of time-related functions. See #2416 for more details.

# How this works

`node-cli` now has two features: `cli` and `browser`. The `cli` feature is enabled by default and provides all CLI-related functions. The `browser` feature is disabled by default and provides the functions that the JavaScript code has to call in order to run the node.

Contrary to my previous prototype, I think this is quite a clean design. It makes sense to me that the same package ("node-cli") can be used either as a binary or from the browser, depending on how it is compiled.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -200,7 +200,6 @@ check-web-wasm:
     - time cargo web build -p substrate-consensus-aura
     - time cargo web build -p substrate-consensus-babe
     - time cargo web build -p substrate-consensus-common
-    - time cargo web build -p substrate-service --no-default-features
     - time cargo web build -p substrate-telemetry
     # Note: the command below is a bit weird because several Cargo issues prevent us from compiling the node in a more straight-forward way.
     - time cargo build --manifest-path=node/cli/Cargo.toml --no-default-features --features "browser" --target=wasm32-unknown-unknown


core/service/Cargo.toml
```diff
@@ -29,7 +29,7 @@ consensus_common = { package = "substrate-consensus-common", path = "../../core/
 network = { package = "substrate-network", path = "../../core/network" }
 chain-spec = { package = "substrate-chain-spec", path = "../chain-spec" }
 client = { package = "substrate-client", path = "../../core/client" }
-client_db = { package = "substrate-client-db", path = "../../core/client/db", features = ["kvdb-rocksdb"] }
+client_db = { package = "substrate-client-db", path = "../../core/client/db" }
```
tomaka: Hmm, this might break `node-template` and Polkadot.

tomaka: I added `rocksdb` as an enabled-by-default feature on `substrate-service`. When you import the service, you have to explicitly say `default-features = false` if you don't want rocksdb. This should be fairly fail-safe.

For context, what happens in the worst case scenario is an error at initialization, so it's not that big of a deal. What I'm mostly trying to avoid is the surprise factor.

node/cli/browser-demo/ws.js
```diff
@@ -0,0 +1,148 @@
+// Copyright 2018-2019 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2019 Parity Technologies (UK) Ltd.
```

node/cli/src/browser.rs
```diff
@@ -0,0 +1,159 @@
+// Copyright 2018-2019 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2019 Parity Technologies (UK) Ltd.
```

PR #3975, opened 2019-10-31
Deprecate ValidateUnsigned and prevent duplicate heartbeats

tomusdrw:
This PR deprecates `ValidateUnsigned` as it's better to unify this with `SignedExtensions` (they are more powerful).

It also fixes an issue observed on Kusama, where outdated hearbeats are included in blocks, but fail. The reason for this is that `ValidateUnsigned` was currently only used for transaction pool, and since the culling in pool is delayed, we can still include a bunch of "invalid" transactions that are already in the pool.

To prevent this we need to return an `ApplyError` early in the process. For `SignedExtensions` this happens during `pre_dispatch` checks, but there was no equivalent for `ValidateUnsigned`. This PR introduces `pre_dispatch` to the trait and makes sure that `ImOnline` is using it correctly.

The next step would be to migrate from `ValidateUnsigned` in `ImOnline`, I'll do this in a separate PR, but I think it's still worth to fix `ValidateUnsigned` and explicitly deprecate it for other blockchains that might be using it.

Related: #3091 #3419


core/sr-primitives/src/traits.rs
```diff
@@ -966,10 +965,27 @@ pub trait RuntimeApiInfo {
 /// the transaction for the transaction pool.
 /// During block execution phase one need to perform the same checks anyway,
 /// since this function is not being called.
+#[deprecated(note = "Use SignedExtensions instead.")]
```
bkchr: How should we use `SignedExtensions` for this? You mean we should create an extra signed extension for each unsigned transaction and add this to the signed extensions used by the `UncheckedExtrinsic`? 

This would mean that we accept any unsigned transaction, when we forget to add the appropriate signed extension?

tomusdrw: Yes. It's a fair point that we should only allow unsigned transactions that are coming from one of the whitelisted modules (currently the `validate_unsigned` function of `SignedExtension` will just allow everything). But it should be easy to do a wrapper type that simply disallows a transaction if it's unsigned and none of the modules returns non-default `TransactionValidity`.

Actually it might be best to just return `InvalidTransaction` by default from `validate_unsigned` when we get rid of `UnsignedValidator`.

bkchr: Yeah, good ideas :)

kianenigma: maybe one of our meta-modules (system etc.) can register a unified `CheckAndHandleUnsigned` signed extension. What this does is it rejects everything except for a handful ones that are explicitly whitelisted. This can be injected into system via `type UnsignedValidatorModules` which would be a tuple of modules that implement some trait (maybe even the current `ValidateUnsigned`)

This basically reverses the problem Basti said 

> This would mean that we accept any unsigned transaction, when we forget to add the appropriate signed extension?

everything is rejected unless a module explicitly ping system module about it.

kianenigma: Okay what I said ^^ is in the same line of: 

> But it should be easy to do a wrapper type that simply disallows a transaction if it's unsigned and none of the modules returns non-default TransactionValidity.

i just understood it :D

srml/im-online/src/lib.rs
```diff
@@ -233,24 +233,20 @@ decl_module! {
...
 			let current_session = <session::Module<T>>::current_index();
-			ensure!(current_session == heartbeat.session_index, "Outdated heartbeat received.");
 			let exists = <ReceivedHeartbeats>::exists(
 				&current_session,
 				&heartbeat.authority_index
 			);
 			let keys = Keys::<T>::get();
 			let public = keys.get(heartbeat.authority_index as usize);
 			if let (true, Some(public)) = (!exists, public) {
-				let signature_valid = heartbeat.using_encoded(|encoded_heartbeat| {
```
kianenigma: why is this gone?

andresilva: It's done in `validate_unsigned` as well before dispatching the transaction. @tomusdrw are `pre_dispatch` checks run as part of block execution? Otherwise someone could include an invalid transaction into a block if the signature is only checked there.

tomusdrw: Yes, that's the main difference between `validate_transaction`(run only by the pool, to determine if transaction should enter it or not) and `pre_dispatch` run during block processing to make sure the transaction is valid. A transaction with failing `pre_dispatch` should cause a block import to panic. We are only able to recover from it during block authorship (we just skip transaction, mark it as invalid in the pool and try others).

The reason why we have pre-dispatch failing transactions in the pool is asynchrony between `validate_transaction` and the actual block creation. Some transactions might become invalid because some other transactions were imported first, and such relationship is not possible to express via `provides`.

andresilva: Thanks for the explanation! :)

PR #3976, opened 2019-10-31
Remove NetworkSpecialization::on_event

tomaka:
Some refactoring/clean-up.

Calling `NetworkSpecialization::on_event` was initially the way that we had planned to receive DHT events.
However the `NetworkWorker` will now generate these DHT events on its stream output, and these events can be made accessible by passing a channel when building the `Service` (using `with_dht_event_tx` method).

This PR consequently deprecates `on_event`, with the intent of removing it in the not-so-far future.

I don't think Polkadot needs DHT events, but if there's ever a need for them that isn't satisfied with `with_dht_event_tx`, we can always revisit this.

### DIFF
--- a/core/network/src/protocol/specialization.rs
+++ b/core/network/src/protocol/specialization.rs
@@ -42,7 +42,8 @@ pub trait NetworkSpecialization<B: BlockT>: Send + Sync + 'static {
 	);
 
 	/// Called when a network-specific event arrives.
-	fn on_event(&mut self, event: Event);
+	#[deprecated(note = "This method is never called; please use `with_dht_event_tx` when building the service")]
+	fn on_event(&mut self, event: Event) {}
 
 	/// Called on abort.
 	#[deprecated(note = "This method is never called; aborting corresponds to dropping the object")]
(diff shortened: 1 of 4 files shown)


core/network/src/protocol/specialization.rs
```diff
@@ -42,7 +42,8 @@ pub trait NetworkSpecialization<B: BlockT>: Send + Sync + 'static {
 	);
 
 	/// Called when a network-specific event arrives.
-	fn on_event(&mut self, event: Event);
+	#[deprecated(note = "This method is never called; please use `with_dht_event_tx` when building the service")]
+	fn on_event(&mut self, event: Event) {}
```
bkchr: Why not remove it completely? We don't have anything that uses it and we don't have anything released. Deprecating does not makes any sense here.

tomaka: Well, there's Polkadot. The idea was to then make a Polkadot PR, then later actually remove them.

bkchr: Hmm not sure. It is not that something updates Polkadot automatically. You could already open the pr to Polkadot and everybody would be happy.

Issue #3912, opened 2019-10-25
Pass arguments to off-chain worker


xlc:
How to pass arguments to off-chain worker? AFAIK it is possible to configure session key for offchain worker to consume. But can this session key be arbitrary string?

For example, I would like to pass API key to off-chain worker so it can fetch data from exchanges. The offchain code can check if it has API key configured to do actual work.

In the end, I would like to build an oracle module that allow people to stake to become an operator. Then they can configure a session key and one or more external services API keys to start submit data using offchain worker. This makes a permisionless decentralised oracle network possible.

thiolliere:
it seems arguments to off-chain worker can benefit from the architecture of the chain spec. with availability to changes keys at specific block numbers and extend it.

thiolliere:
hmm did I actually understood correctly ?

pass argument to off-chain worker from where ? from the runtime ? or from the cli ?

if it is from the runtime this can be done indeed

thiolliere:
> In the end, I would like to build an oracle module that allow people to stake to become an operator. Then they can configure a session key and one or more external services API keys to start submit data using offchain worker. This makes a permisionless decentralised oracle network possible.

I don't understand how this make use of arguments ? this key is just in the trie ?

tomusdrw:
> For example, I would like to pass API key to off-chain worker so it can fetch data from exchanges. The offchain code can check if it has API key configured to do actual work.

This should be just inserted to the local offchain worker database under some known key. You can set it up during service or CLI initialisation.

xlc:
Yes like @tomusdrw said. Pass arguments from runtime is easy as offchain worker can just read storage. I would like to pass arguments from CLI or API. Preferably from API so it can be rotated without reboot the node.

xlc:
Fixed by #4694

PR #3977, opened 2019-10-31
expose offchain worker storage prefix

gnunicorn:
Expose the offchain worker storage prefix so other components can reuse it when they want to communicate with the offchain worker via storage.

E.g. store a configuration at a specific key at startup that the offchain worker than reads and uses to do its job (in this case: an SMTP setup).


core/offchain/src/api.rs
```diff
@@ -71,6 +71,7 @@ fn unavailable_yet<R: Default>(name: &str) -> R {
 }
 
 const LOCAL_DB: &str = "LOCAL (fork-aware) DB";
+/// Local Storage Prefix used by the Offchain Worker API to
 pub const STORAGE_PREFIX: &[u8] = b"storage";
```
tomusdrw: Perhaps in such case it's better suited in  `primitives/offchain`?

xlc:
Relates to #3912?

gnunicorn:
@xlc yes. this is effectively what we are doing to get the smtp config from the node into the offchain worker in gather.

https://github.com/gnunicorn/gather/blob/efe97032870d3f7780728c5341624ac9d530d8be/src/service.rs#L82-L90

however. I intend to follow up the next few days with a proper way to expose other functionality to the offichain worker, that could be stateful.

PR #3978, opened 2019-10-31
Move config path generation into the service config for reusability

gnunicorn:
As of now, if a node wants to store its own information in the configuration directory chose through the cli, it has no way of knowing which directory that is supposed to be. This PR refactors the `Configuration` and its helper functions into a single entity that a) decides upon all directories by using its base directory, b) means every other component now only needs access to config to figure out the correct place to store stuff and c) allow nodes to store their own specific configuration and cache files.


core/cli/src/lib.rs
```diff
@@ -666,14 +680,10 @@ where
 		)?
 	}
 
-	let base_path = base_path(&cli.shared_params, version);
-
-	config.keystore_path = cli.keystore_path.unwrap_or_else(
-		|| keystore_path(&base_path, config.chain_spec.id())
-	);
+	config.keystore_path = cli.keystore_path.unwrap_or_else( || config.in_chain_config_dir("keystore"));
```
```suggestion
	config.keystore_path = cli.keystore_path.unwrap_or_else(|| config.in_chain_config_dir("keystore"));
```

core/cli/src/lib.rs
```diff
@@ -556,18 +571,16 @@ fn fill_transaction_pool_configuration<C, G, E>(
 }
 
 /// Fill the given `NetworkConfiguration` by looking at the cli parameters.
-fn fill_network_configuration(
+fn fill_network_configuration
```
```suggestion
fn fill_network_configuration(
```

core/cli/src/lib.rs
```diff
@@ -556,18 +571,16 @@ fn fill_transaction_pool_configuration<C, G, E>(
...
+(
```
```suggestion
```

core/cli/src/lib.rs
```diff
@@ -307,18 +307,33 @@ pub struct ParseAndPrepareBuildSpec<'a> {
...
+		C: Default,
 		G: RuntimeGenesis,
 		E: ChainSpecExtension,
 	{
 		info!("Building chain spec");
 		let raw_output = self.params.raw;
 		let mut spec = load_spec(&self.params.shared_params, spec_factory)?;
-		with_default_boot_node(&mut spec, self.params, self.version)?;
+
+		if spec.boot_nodes().is_empty() {
```
```suggestion
		if spec.boot_nodes().is_empty() && !self.params.disable_default_bootnode {
```

Why did you remove this?

core/service/src/config.rs
```diff
@@ -161,6 +169,10 @@ impl<C, G, E> Configuration<C, G, E> where
 		configuration
 	}
 
+}
+
+impl<C, G, E> Configuration<C, G, E> {
+
```
bkchr:
```suggestion
```

core/service/src/config.rs
```diff
@@ -120,11 +122,17 @@ impl<C, G, E> Configuration<C, G, E> where
 {
 	/// Create default config for given chain spec.
 	pub fn default_with_spec(chain_spec: ChainSpec<G, E>) -> Self {
+		Self::default_with_spec_and_base_path(chain_spec, Default::default())
```
bkchr: I don't think that the default `config_dir` should be an empty path.

core/cli/src/lib.rs
gnunicorn: not intentionally, this was moved before your patch came in and I forgot it in the merge. Good catch!

core/service/src/config.rs
gnunicorn: If I see this correctly, this is only used in tests now, for which this doesn't matter. 

What would you suggest this should be used instead? Should I put in a tmp-path (though it is never used) or the current working directory? Or rename it to  ´default_with_spec_with_empty_path´ to make it explicit?

gavofyork:
waiting for resolving and @bkchr's signoff.

core/service/src/config.rs
bkchr: Remove it?
If some tests need to call this, then it should call `Default::default()` itself and be aware that maybe something is written to this path.

gnunicorn: actually, no one seems to keep using it. removing.

core/cli/src/lib.rs
```diff
@@ -666,14 +679,10 @@ where
...
-	let base_path = base_path(&cli.shared_params, version);
-
-	config.keystore_path = cli.keystore_path.unwrap_or_else(
-		|| keystore_path(&base_path, config.chain_spec.id())
-	);
+	config.keystore_path = cli.keystore_path.unwrap_or_else(|| config.in_chain_config_dir("keystore"));
 
 	config.database = DatabaseConfig::Path {
-		path: db_path(&base_path, config.chain_spec.id()),
+		path: config.database_path(),
```
tomaka: This line is kind of a red flag. We store `config.database_path()` in `config.database.path`.

gnunicorn: the general change of this PR is to move the generation of relative paths from the internal `*_path(&base_path`-style onto the configuration object. `database_path` there is only a handy alias for `self.base_path.clone().join("database")`. 

What exactly is the red flag?

core/service/src/config.rs
```diff
@@ -170,6 +176,27 @@ impl<C, G, E> Configuration<C, G, E> where
...
+		self.config_dir.clone().map(|mut path| {
+			path.push("chains");
+			path.push(self.chain_spec.id());
+			path.push(sub);
+			path
+		})
+	}
+	
+	/// The basepath for the db directory
+	pub fn database_path(&self) -> Option<PathBuf> {
```
tomaka: This function really doesn't make sense. There's already a `database.path` field in `self`, and what I would expect by  calling this function is that the value of this field gets returned.

In my opinion we should remove `path` from `DatabaseConfig` and split `NetworkConfiguration` in two, with one version without the paths in `Configuration` and one with the paths passed when building the network. This way, there would be on incoherent state in `Configuration`.

As a quick hack I'd be fine with `NetworkConfiguration::config_path`/`NetworkConfiguration::net_config_path` set to null, then appropriately set later.

gnunicorn: I am fine with removing that function and have the setter function directly call `in_chain_config_dir("db")` analog to how this is done for `keystore_path`. It was just meant as a handy function to have the `db` only be written once rather than multiple places throughout the code base.

gnunicorn:
addressing some minor refactor. please wait before merging.

gnunicorn:
finally pushed.

PR #3980, opened 2019-10-31
Check polkadot compilation.

tomusdrw:
Run a `cargo check` on the polkadot `master` with substrate pointing to current PR.


.gitlab-ci.yml
```diff
@@ -319,6 +319,20 @@ check_warnings:
...
+  stage:                           build
+  <<:                              *docker-env
+  allow_failure:                   true
+  dependencies:
+    - test-linux-stable
+  script:
+    - git clone --depth 1 https://github.com/paritytech/polkadot.git
+    - cd polkadot
+    - git grep "polkadot-master" -l | xargs sed -i "s/polkadot-master/$CI_COMMIT_REF_NAME/"
+    - time cargo check
```
bkchr:
```suggestion
    - BUILD_DUMMY_WASM_BINARY=1 time cargo check
```

.gitlab-ci.yml
```diff
@@ -319,6 +319,20 @@ check_warnings:
       fi
   allow_failure:                   true
 
+check_polkadot:
+  stage:                           build
```
bkchr:
```suggestion
  stage:                           test
```

Otherwise it is not run while testing a pr.

tomusdrw: Why not? Does the stage really matter? I just want it to run after the checks are green as this might be a bit heavy.

.gitlab-ci.yml
bkchr: Hmm ahh yeah. Did not thought about this.

tomusdrw:
It fails, hence it works. :thinking:

bkchr:
Hmm, there is a compile error:
```
error[E0609]: no field `database_path` on type `substrate_service::config::Configuration<CustomConfiguration, polkadot_runtime::GenesisConfig>`
   --> service/src/lib.rs:149:23
    |
149 |     let db_path = config.database_path.clone();
    |                          ^^^^^^^^^^^^^ unknown field
    |
    = note: available fields are: `impl_name`, `impl_version`, `impl_commit`, `roles`, `transaction_pool` ... and 24 others
```

PR #3981, opened 2019-10-31
Avoid sending heartbeat if we are already considered online.

tomusdrw:
Authority can now be online by producing a block or an uncle, so it doesn't make sense to produce a heartbeat any more.

It would be best to actually defer producing a heartbeat a bit (like to the middle of the session period) instead of sending it right away. But that seems to be more involved than I anticipated CC @andresilva.


srml/im-online/src/lib.rs
```diff
@@ -405,14 +416,21 @@ impl<T: Trait> Module<T> {
...
-			// because at this point the heartbeat was definitely submitted.
-			Self::set_worker_status(block_number, true);
+			results.push(
+				T::SubmitTransaction::submit_unsigned(call)
+					.map_err(|_| OffchainErr::SubmitTransaction)
+			);
 		}
+
+		// fail only after trying all keys.
+		results.into_iter().collect::<Result<Vec<_>, OffchainErr>>()?;
```
bkchr: That is not correct: https://doc.rust-lang.org/src/core/result.rs.html#1299

bkchr: (It will still fail after the first error)

tomusdrw: That's actually ok, I just don't want to break the loop on the first error. Since we now guard against `is_online_in_current_session` we should not produce a duplicate report , so we don't set the worker as finished - we will retry next time and just produce the ones that are missing.

But to be fair the error shouldn't really ever happen, it would only happen if the pool is full and cannot accept new transactions (although our have the highest priority so they should kick everything else) or the transaction is somehow incorrect.

PR #3985, opened 2019-10-31
grandpa: Re-add Grandpa runtime API for genesis authority set.

jimpo:
This is a follow-up to #3734 to ensure new nodes on Kusama (or any Substrate chain deployed before #3734) can still get the initial authority set in order to sync. Without this, existing nodes would be fine since the authority set is persisted locally by the grandpa service, but new nodes would not be able to read the initial authority set.

The main difference from #3734 is that the `GenesisAuthoritySetProvider` uses the Grandpa runtime API, which is backwards compatible, instead of a direct storage lookup.

I have tested that this works locally with the following steps:

1. Checkout master. Change `MILLISECONDS_PER_BLOCK` in `node/runtime/src/constants.rs` to 6000 and build. Save the built `substrate` binary as `substrate-master`.
2. Checkout this branch. Change `MILLISECONDS_PER_BLOCK` in `node/runtime/src/constants.rs` to 6000 and build.
3. Run `./substrate-master build-spec --dev --raw > my-spec.json`.
4. Run `./substrate-master --chain my-spec.json --alice --base-path ./tmp-alice`.
5. Run `./substrate-master --chain my-spec.json --bob --base-path ./tmp-bob --bootnodes /ip4/172.17.0.1/tcp/30333/p2p/Qmc73KjWHef5HY7S3cjxLJrqvtJLZbcTHMmBwnHMZ5QXoM --port 30332`.
6. Observe that the bob node follows finalized blocks from Alice.
7. Kill Bob and Alice.
8. Re-run Alice using the newly build `target/release/substrate` instead of `substrate-master`.
9. Re-run Bob using the newly build `target/release/substrate` instead of `substrate-master`.
10. Observe that Bob gets in sync.
11. Go to https://polkadot.js.org/apps/#/extrinsics and point at the local Alice Substrate node. Use `sudo(proposal), system, setCode` and set the code to the new `target/release/wbuild/node-runtime/node_runtime.compact.wasm`.
12. Observe that Alice upgrades by querying the `state_getRuntimeVersion` RPC for the spec version.
13. Observe the Bob stays in sync.
14. Remove the ./tmp-bob database.
15. Re-run Bob with a clean database and observe that it gets in sync.

### DIFF
--- a/core/finality-grandpa/primitives/src/lib.rs
+++ b/core/finality-grandpa/primitives/src/lib.rs
@@ -23,9 +23,10 @@ extern crate alloc;
 
 #[cfg(feature = "std")]
 use serde::Serialize;
-use codec::{Encode, Decode, Codec};
+use codec::{Encode, Decode, Input, Codec};
 use sr_primitives::{ConsensusEngineId, RuntimeDebug};
 use client::decl_runtime_apis;
+use rstd::borrow::Cow;
 use rstd::vec::Vec;
 
 mod app {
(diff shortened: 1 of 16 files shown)


rphmeier:
LGTM after master merge. Thanks!

Issue #3695, opened 2019-09-26
Storage root mismatch on testnet


iammelea:
Running one testnet this logs in  one sentry, no stop just bucle logs, after stop and start , works well
```
Hash: given=a4efcabf0dd2c407588236b059e657acf6e55eb07e3175e81a832eee1e0130be, expected=dcfa4fd44da5b9319cf6d7fac53e5d2e341d30f64dc5bc0b35b79d89e4232f6a

====================

Version: 2.0.0-4117bb9-x86_64-linux-gnu

stack backtrace:
   0: substrate_panic_handler::set::{{closure}}
   1: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:481
   2: std::panicking::begin_panic
   3: srml_executive::Executive<System,Block,Context,UnsignedValidator,AllModules>::execute_block
   4: node_runtime::api::dispatch
   5: std::panicking::try::do_call
   6: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:82
   7: substrate_executor::native_executor::safe_call
   8: std::thread::local::LocalKey<T>::with
   9: <node_executor::Executor as substrate_executor::native_executor::NativeExecutionDispatch>::dispatch
  10: <substrate_executor::native_executor::NativeExecutor<D> as substrate_primitives::traits::CodeExecutor<substrate_primitives::hasher::blake2::Blake2Hasher>>::call
  11: substrate_state_machine::StateMachine<B,H,N,T,O,Exec>::execute_aux
  12: substrate_state_machine::StateMachine<B,H,N,T,O,Exec>::execute_call_with_native_else_wasm_strategy
  13: substrate_state_machine::StateMachine<B,H,N,T,O,Exec>::execute_using_consensus_failure_handler
  14: <substrate_client::call_executor::LocalCallExecutor<B,E> as substrate_client::call_executor::CallExecutor<Block,substrate_primitives::hasher::blake2::Blake2Hasher>>::call_at_state
  15: <&substrate_client::client::Client<B,E,Block,RA> as substrate_consensus_common::block_import::BlockImport<Block>>::import_block
  16: <substrate_finality_grandpa::import::GrandpaBlockImport<B,E,Block,RA,PRA,SC> as substrate_consensus_common::block_import::BlockImport<Block>>::import_block
[...]

Issue #3628, opened 2019-09-17
Runtime upgrades causing node to drop most peers


drewstone:
I am trying to run a runtime upgrade that changes a single parameter and version number. I picked the compact wasm blobl from the `wbuild` directory after building from release. Upon sending the TX, the node that receives the TX stalls and drops all its peers. Why is this?

bkchr:
How big is the update? Was the node already running the new native runtime or still the old native runtime?

drewstone:
the old native runtime. The upgrade is roughly 1.3MB @bkchr . Things sometime keep chugging along but eventually hit
```
2019-09-17 14:03:11 Prepared block for proposing at 182 [hash: 0x77a61b53677f35e27f70345bedc34b968372517872892bc1a2fd4dab887b7bc7; parent_hash: 0x063a…83e3; extrinsics: [0xa99f…fb9c, 0xc063…e607]]
2019-09-17 14:03:11 Discarding proposal for slot 261453015; block production took too long
2019-09-17 14:03:18 Starting consensus session on top of parent 0x063aa8039fa9c43ecda3f7634c90fec891fbf69781fc28cf6ba82961b64c83e3
```

drewstone:
@bkchr and adding in @kianenigma 
This happened to one local node after attempted a sudo set_code
```
====================

Version: 1.0.0-a06948d-x86_64-macos

stack backtrace:
   0: backtrace::backtrace::trace
   1: backtrace::capture::Backtrace::new
   2: substrate_panic_handler::set::{{closure}}
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::continue_panic_fmt
   5: rust_begin_unwind
   6: core::panicking::panic_fmt
   7: core::result::unwrap_failed
   8: <substrate_state_machine::ext::Ext<H,N,B,T,O> as substrate_primitives::traits::Externalities<H>>::storage
   9: substrate_executor::wasm_runtimes_cache::RuntimesCache::fetch_runtime
  10: std::thread::local::LocalKey<T>::with
  11: substrate_state_machine::StateMachine<B,H,N,T,O,Exec>::execute_aux
  12: substrate_state_machine::StateMachine<B,H,N,T,O,Exec>::execute_using_consensus_failure_handler
  13: <substrate_client::call_executor::LocalCallExecutor<B,E> as substrate_client::call_executor::CallExecutor<Block,substrate_primitives::hasher::blake2::Blake2Hasher>>::contextual_call
[...]

Issue #3580, opened 2019-09-10
node crash and  when start again, it always panic.


yaanhyy:
It look like that the data which stored in db is  not identical with the data stored in remote node db?
here is the log:
2019-09-10 10:05:26 Syncing  0.0 bps, target=#174602 (4 peers), best: #170051 (0xb14e…0cf7), finalized #158442 (0x25eb…9840), ⬇ 31.8kiB/s ⬆ 22.1kiB/s
Hash: given=edac50c8e697c899ce2bf388686bdd33b9db3983d51a98190ab903a4bec8ce7c, expected=67e81b16f1d35c35c64f08f35a4e9939bd8820d20a57c04f3f102a4a56dfd7e1

====================

stack backtrace:
   0: substrate_panic_handler::set::{{closure}}::h6c64ce3a7889097c (0x55f2b5b2ed9e)
   1: std::panicking::rust_panic_with_hook::h2e01ca08bfa50f6c (0x55f2b5cb863c)
             at src/libstd/panicking.rs:481
   2: std::panicking::begin_panic::h7a7473a1c7cf749c (0x55f2b514b8d5)
   3: <node_runtime::Runtime as substrate_client::runtime_api::runtime_decl_for_Core::Core<sr_primitives::generic::block::Block<sr_primitives::generic::header::Header<u32,sr_primitives::traits::BlakeTwo256>,sr_primitives::generic::unchecked_extrinsic::UncheckedExtrinsic<<srml_indices::Module<node_runtime::Runtime> as sr_primitives::traits::StaticLookup>::Source,node_runtime::Call,sr_primitives::AnySignature,(srml_system::CheckVersion<node_runtime::Runtime>, srml_system::CheckGenesis<node_runtime::Runtime>, srml_system::CheckEra<node_runtime::Runtime>, srml_system::CheckNonce<node_runtime::Runtime>, srml_system::CheckWeight<node_runtime::Runtime>, srml_balances::TakeFees<node_runtime::Runtime>)>>>>::execute_block::h6bb511705724960c (0x55f2b5158733)
   4: node_runtime::api::dispatch::hc8f40040c6da3def (0x55f2b5142f2b)
   5: std::panicking::try::do_call::h61ac906458cbbf61 (0x55f2b511efa2)
   6: __rust_maybe_catch_panic (0x55f2b5cbd4da)
             at src/libpanic_unwind/lib.rs:80
[...]

PR #3989, opened 2019-11-01
Possible fix to storage cache

marcio-diaz:
This is a possible fix to an issue that causes sync to stall in some cases.

In order to reproduce the issue you can create a fork (using `--reserved-only` or Blockade partitions) and then join the network. Depending on the length of the fork it may delay or stall sync, producing an error similar to `Thread 'import-queue-worker-0' panicked at 'Storage root must match that calculated.', srml/executive/src/lib.rs:271`.

#### Detailed (possible) explanation:
- After creating the forks, restart the clients removing `--reserved-only`, joining the network.
- When restarting the nodes `cache.modifications` will be empty.
- We can get cache misses while executing best block, filling `local_cache`.
- When other node is discovered, it imports and execute common ancestor block.
- Since `cache.modifications` is still empty, `is_allowed()` returns `true`, and we can retrieve a wrong value for the key, when querying at the ancestor block.

**Disclaimer**: I'm not very familiar with this code so the explanation may be wrong.

#### Possible solutions:
- As in the current version of the PR, eliminate the propagation of `local_cache` into `cache.lru_*`. This fixes the bug but looks like it doesn't  make full use of the caches. 
- Inserting the modification when propagating. Here I'm not sure from where should I take the `commit_hash` and `commit_number`. Well, I could take it from parent, but then probably I need to add some logic.
- Just remove the `return true` form `is_allowed` when `modifications` is empty.

\cc @arkpar , @cheme 

Closes: #3695,  #3580, https://github.com/paritytech/polkadot/issues/413
Related: #3628

### DIFF
--- a/core/client/db/src/storage_cache.rs
+++ b/core/client/db/src/storage_cache.rs
@@ -321,27 +321,29 @@ impl<H: Hasher, B: BlockT> CacheChanges<H, B> {
 		// Propagate cache only if committing on top of the latest canonical state
 		// blocks are ordered by number and only one block with a given number is marked as canonical
 		// (contributed to canonical state cache)
-		// if let Some(_) = self.parent_hash {
-		// 	let mut local_cache = self.local_cache.write();
-		// 	if is_best {
-		// 		trace!(
-		// 			"Committing {} local, {} hashes, {} modified root entries, {} modified child entries",
-		// 			local_cache.storage.len(),
-		// 			local_cache.hashes.len(),
-		// 			changes.len(),
-		// 			child_changes.iter().map(|v|v.1.len()).sum::<usize>(),
-		// 		);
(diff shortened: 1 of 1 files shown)


cheme:
:+1: I am not very familiar with the internal, but what the test describes looks like a race condition. I would put a state counter (atomic uint) that increment on cache clear, and value updates would only be allowed iff the state counter did not change. (Probably stating the obvious).

arkpar:
Indeed, when restarting the cache fails to identify the canonical chain properly. For now I'd just disable cache when there are no modifications.

marcio-diaz:
> Indeed, when restarting the cache fails to identify the canonical chain properly. For now I'd just disable cache when there are no modifications.

Sounds good to me, thanks.

PR #3991, opened 2019-11-01
Add events for im_online

gavofyork:
Also cleanup the code a bit.

### DIFF
--- a/srml/im-online/src/lib.rs
+++ b/srml/im-online/src/lib.rs
@@ -200,11 +200,14 @@ pub trait Trait: system::Trait + session::historical::Trait {
 decl_event!(
 	pub enum Event<T> where
 		<T as Trait>::AuthorityId,
+		IdentificationTuple = IdentificationTuple<T>,
 	{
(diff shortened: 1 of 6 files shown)


srml/im-online/src/lib.rs
```diff
@@ -507,46 +512,32 @@ impl<T: Trait> session::OneSessionHandler<T::AccountId> for Module<T> {
...
-		let current_session = <session::Module<T>>::current_index();
-
+		let session_index = <session::Module<T>>::current_index();
 		let keys = Keys::<T>::get();
 		let current_validators = <session::Module<T>>::validators();
 
-		for (auth_idx, validator_id) in current_validators.into_iter().enumerate() {
-			if !Self::is_online_in_current_session_aux(auth_idx as u32, &validator_id) {
-				let full_identification = T::FullIdentificationOf::convert(validator_id.clone())
-					.expect(
```
gavofyork: this is a weak and dangerous proof; it should never have been here.

nothing should be able to panic in initialize_block/finalize_block.

Issue #3883, opened 2019-10-22
Extend Externalities with `for_keys_with_prefix` instead of just clear_prefix


thiolliere:
Currently Externalities only provides way to clear_prefix and clear_child_prefix which allows to clear all keys with prefix in main trie or a child trie.

In order to do so the current backend does iterate over keys with such prefix and removes them. It would be great to expose a function `for_keys_with_prefix(prefix, callback)` which would allow to iterate on all keys with prefix in main trie and in child tries.

The main interest is for childtrie, if you could isolate some storage into its child trie you could then iterate on all values in this child trie which can be usefull

## Cons

* the function `fn for_keys_with_prefix(prefix, callback)` would need to pass a callback from wasm to native. Which might be unsuitable.
  the wasm callback will have to be executed on each value found.
  as the only intereset for me is to iterate on all value of a childtrie maybe the solution is actually to have a new externality which is `fn all_value` which returns all values in vec.

Issue #3648, opened 2019-09-19
Possible child trie key collision when pruning


cheme:
Currently two different child tries will store the encoded value of their trie node in rocksdb without any distinction of origin.

Therefore if two child tries got similar contents, part of their underlying trie node can be identical and the same rocksdb entry will be shared.

Since we do not use reference counting (for good reason) on our stored trie encoded node, it results in problematic key collisions, and makes child trie unusable if pruning is activated.

Tackling this can be done by using reference counting for our child trie rocksdb entries (this is not the prefered approach) or by isolating every child trie datas (prefixing rocksdb key values with a unique child trie id was the chosen approach in #2209).

Note that in the current code there is no distinction between rocksdb entries coming from the top level trie storage and the child trie storages, so prefix approach will need to either apply prefix to top level key or change the api to allow some trie (top) to store in a dedicated rocksdb column.

A previous PR #2209 did implement this, but a caveat was that the unique child trie technical identifier where stored on chain see https://github.com/paritytech/substrate/pull/2209#issuecomment-528811294 .

Note that usage of keyspace for keys under a child trie allows some efficient deletion. When a child trie get fully removed, the fact that next written entries for the same child trie key are using a new keyspace makes a prunning only on keyspace possible (no need to register every keys and possibility to make the actual pruning at rocksdb level).

abebeos:
Could someone please add sections like this to this issue, and clarify if this is still an unsolved issue?

## Current Status

## Expected

## Possible Solution(s)

cheme:
Sorry, this was fixed in https://github.com/paritytech/substrate/pull/4162 (long ago).

PR #3992, opened 2019-11-01
Introduce new storage type: prefixed map

thiolliere:
## Context

Prefixed map is a new storage available in decl_storage.

it differ from map as it stores it keys at `twox_256(prefix)++map_hasher(key)` instead of `map_hasher(prefix ++ key)`.

Thus it allows to remove all key of the map using `sr-io::kill_prefix`.

But the final goal is to make it also iterable using https://github.com/paritytech/substrate/issues/3883

In the end this should replace linked_map as iteration is more efficient (just there is no longer ordering). and also Maps as it allows to have more functionality 

## To discuss:

* should we directly put this in a child trie instead of behind a prefix in top trie ?
  * I would prefer waiting for https://github.com/paritytech/substrate/issues/3648 before having child trie in decl_storage actually.

* the value is stored at `twox_256(prefix)++map_hasher(key)`:
  * the first part is using twox_256 for hashing the prefix because we have to be sure that no other value share the same prefix. in case we only use twox_128 then an attacker could try to craft a key which have blake2_256(key)[0..16] == twox_128(prefix). Thus inserting a wrong value into this map.
    
    Introducing this map has the consequence to limit the size of our hashers. introducing a hasher with an output longer than 256 bit is no longer relevant as it can already collision with this map.

    If this is an issue then it can be solve in two way:
    * waiting for child trie
    * allow to remove key with a given prefix and a given size only. thus we would call `sr-io::kill_prefix_with_size(MAP_PREFIX, 32+size_of_map_hasher)`

  * the second part is map_hasher(key). currently a hasher is mandatory though it seems it could be interesting to introduce a no_hasher variant (or hasher=identity). thus value would be stored at `twox_256(prefix) ++ key.encode()`. The only interest I see from this is that it would avoid some hasher call.

### DIFF
--- a/srml/metadata/src/lib.rs
+++ b/srml/metadata/src/lib.rs
@@ -278,6 +278,15 @@ pub enum StorageHasher {
 	Twox64Concat,
 }
 
+/// The kind of map for a map storage entry type.
+#[derive(Clone, PartialEq, Eq, Encode, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(Decode, Serialize))]
+pub enum StorageMapKind {
+	Map,
+	LinkedMap,
+	PrefixedMap,
+}
+
 /// A storage entry type.
 #[derive(Clone, PartialEq, Eq, Encode, RuntimeDebug)]
 #[cfg_attr(feature = "std", derive(Decode, Serialize))]
@@ -287,7 +296,7 @@ pub enum StorageEntryType {
 		hasher: StorageHasher,
 		key: DecodeDifferentStr,
 		value: DecodeDifferentStr,
-		is_linked: bool,
+		kind: StorageMapKind,
 	},
 	DoubleMap {
 		hasher: StorageHasher,
@@ -342,8 +351,10 @@ pub enum RuntimeMetadata {
 	V6(RuntimeMetadataDeprecated),
 	/// Version 7 for runtime metadata. No longer used.
 	V7(RuntimeMetadataDeprecated),
-	/// Version 8 for runtime metadata.
-	V8(RuntimeMetadataV8),
+	/// Version 8 for runtime metadata. No longer used.
+	V8(RuntimeMetadataDeprecated),
+	/// Version 9 for runtime metadata.
+	V9(RuntimeMetadataV9),
 }
 
 /// Enum that should fail.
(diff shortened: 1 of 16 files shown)


xlc:
This looks like `double_map` with with first key of `()`?

https://github.com/paritytech/substrate/blob/ec7c6cf1779b88e75137ef6f6f7bf67ecd0f75a5/srml/system/src/lib.rs#L416

gavofyork:
It's not so great to have a long, insecure hash as the first key as it's not inconceivable that someone could craft a module which intentionally has storage keys that collided with another module. A better way of forming the first key could be to simply append a short twox hash (8 byte, perhaps) of the item name to either the module name or a twox hash of it. This more or less guarantees it being unrealistic to craft a storage item collision, at least making it super-obvious in the module name.

thiolliere:
So finally we will do instead of a new storage type just an optional attribute on linked_map

```
MyLinkedMap prefixed(): linked_map u32 => u32
```

stored at `twox(ModuleName)++twox(MyLinkedMap)++hasher(key)`.
This would implemente the same trait StorageLinkedMap (though with change of the generator to have this two prefixes), also in the metadata we need a new field for the additional prefix as well (or a complete new storage variant)

(Or otherwise we could also make use of blake2hasher if we can ensure this is done at compile time but I prefer the first way)

I'll work on this starting next week after my days off

gavofyork:
sounds good.

bkchr:
As we want to remove `LinkedMap` in the future anyway. I would propose that you add the new `prefixed()` attribute to `StorageMap`. 
If it is a prefixed `StorageMap`, we also implement `Iterator` or a custom trait to make it iterable like a `LinkedMap`.

gavofyork:
how far off is this? i could do with it for a PR i'm working on...

gavofyork:
in particular, i could do with iteration on a `double_map` - is that already prototyped?

thiolliere:
for double_map you want to iterate over all values?

I can implement `Iterator<Item=(Key1, Key2, Value)>` similarly to the incoming prefixed map it seems.

But if you want to iterate on first keys only it seems it requires new function in sr-io.

thiolliere:
superseeded by https://github.com/paritytech/substrate/pull/4185

PR #3993, opened 2019-11-01
executor: Move runtime version caching out of WasmRuntime interface.

jimpo:
The runtime version is now fetched and cached at a higher level, not within the WasmRuntime trait implementations. The purpose is primarily to reduce duplication.

### DIFF
--- a/core/executor/src/wasm_runtime.rs
+++ b/core/executor/src/wasm_runtime.rs
@@ -56,10 +56,7 @@ pub enum WasmExecutionMethod {
 struct VersionedRuntime {
 	runtime: Box<dyn WasmRuntime>,
 	/// Runtime version according to `Core_version`.
-	///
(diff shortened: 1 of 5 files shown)


core/executor/src/wasm_runtime.rs
```diff
@@ -58,6 +52,16 @@ pub enum WasmExecutionMethod {
...
 }
 
+/// A Wasm runtime object along with its cached runtime version.
+struct VersionedRuntime {
+	runtime: Box<dyn WasmRuntime>,
+	/// Runtime version according to `Core_version`.
+	///
+	/// Can be `None` if the runtime doesn't provide the information or there was an error while
+	/// fetching it.
+	version: Option<RuntimeVersion>,
```
bkchr: The core interface is required by any valid substrate runtime, so there always need to be a version.

If the runtime does not provides a version, it should be handled as invalid.

So, please remove the Option :)

PR #3994, opened 2019-11-01
Optional serde for phragmen support

kianenigma:
useful for third party tools that might use the algorithm outside of substrate.

### DIFF
--- a/core/phragmen/src/lib.rs
+++ b/core/phragmen/src/lib.rs
@@ -33,6 +33,8 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+#[cfg(feature = "std")]
+use serde::{Serialize, Deserialize};
 use rstd::{prelude::*, collections::btree_map::BTreeMap};
 use sr_primitives::RuntimeDebug;
 use sr_primitives::{helpers_128bit::multiply_by_rational, Perbill, Rational128};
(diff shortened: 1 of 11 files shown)


core/phragmen/src/lib.rs
```diff
@@ -116,6 +118,7 @@ pub struct PhragmenResult<AccountId> {
 /// This, at the current version, resembles the `Exposure` defined in the staking SRML module, yet
 /// they do not necessarily have to be the same.
 #[derive(Default, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```
bkchr:
```suggestion
#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
```

core/phragmen/src/lib.rs
```diff
@@ -33,6 +33,8 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+#[cfg(feature = "std")]
```
bkchr:
```suggestion
```

core/phragmen/src/lib.rs
```diff
@@ -33,6 +33,8 @@
...
+use serde::{Serialize, Deserialize};
```
bkchr:
```suggestion
```

Issue #3700, opened 2019-09-26
Remove panics in srml-support on corrupted storage state.


thiolliere:
In few places we panic in case of corrupted storage state.

https://github.com/paritytech/substrate/blob/429380ea7d109bce29d6aba2c66f33e79c61ea11/srml/support/src/storage/generator/linked_map.rs#L113-L114

https://github.com/paritytech/substrate/blob/429380ea7d109bce29d6aba2c66f33e79c61ea11/srml/support/src/storage/unhashed.rs#L23-L27

To remove them I see two strategy:
* silently handle those error in those functions, like get would return None in case it can't decode and enumeration will stop in case it can't continue (or even maybe update the linkage)
* provides some method with Result output. this would allow module programmer to make some logic to handle corrupted state.
* make the types implement a trait which is:
  ```rust
  trait OnCorruptedState {
    fn on_corrupted_state(state: &[u8]) -> Option<Self>;
  }
  ```
  and this function would be called on when this happens

The question is more do we want to write specific code when state is corrupted ? or should silently fix the storage state by removing values.

note that our silent fix will be somewhat inconsistent like for linked_map if enumeration is broken by a corrupted linkage then some (key, value) are still part of the map (like accessible with get) but not part of the enumeration.

thiolliere:
also with the first strategy:

> silently handle those error in those functions, like get would return None in case it can't decode and enumeration will stop in case it can't continue (or even maybe update the linkage)

I wonder if there is some log feature in the runtime, like we could at least log than an error has happended.

bkchr:
I like your `OnCorruptedState` idea :) 

If we could integrate this somewhat easily into the `decl_storage!` macro, we could by default just return the default value, but we could give the user the possibility to provide a closure that would be running as part of `on_corrupted_state` to give us the new value.

kianenigma:
I reckong storage migration _could_ also be a subset of `OnCorruptedState `?

thiolliere:
> I reckong storage migration _could_ also be a subset of `OnCorruptedState `?

I can imagine having version in the encoding of the values, like one byte for the version (`$Version++encode(value)`) and if version failed to decode then this on corrupted state could be called. But I don't feel that this is the prefered way to do migration.

arkpar:
What's wrong with panicking? Don't we already use panics to bail out of invalid extrinsics and such?

thiolliere:
[...]

PR #3997, opened 2019-11-01
Remove panic on corruped storage

thiolliere:
Fix corrupted storage panic

In the long term we should implement https://github.com/paritytech/substrate/issues/3700

### DIFF
--- a/palette/support/src/storage/child.rs
+++ b/palette/support/src/storage/child.rs
@@ -26,8 +26,12 @@ use codec::{Codec, Encode, Decode};
 
 /// Return the value of the item in storage under `key`, or `None` if there is no explicit entry.
 pub fn get<T: Decode + Sized>(storage_key: &[u8], key: &[u8]) -> Option<T> {
-	runtime_io::storage::child_get(storage_key, key).map(|v| {
-		Decode::decode(&mut &v[..]).expect("storage is not null, therefore must be a valid type")
+	runtime_io::storage::child_get(storage_key, key).and_then(|v| {
+		Decode::decode(&mut &v[..]).map(Some).unwrap_or_else(|_| {
+			// TODO #3700: error should be handleable.
+			runtime_print!("ERROR: Corrupted state in child trie at {:?}/{:?}", storage_key, key);
+			None
+		})
 	})
 }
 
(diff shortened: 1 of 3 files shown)


srml/support/src/storage/generator/linked_map.rs
```diff
@@ -151,22 +159,32 @@ where
...
-					it always points to existing keys; qed");
-		res.1.next = linkage.next;
-		unhashed::put(prev_key.as_ref(), &res);
+		if let Some(mut res) = read_with_linkage::<_, _, G>(prev_key.as_ref()) {
+			res.1.next = linkage.next;
+			unhashed::put(prev_key.as_ref(), &res);
+		} else {
+			// TODO #3700: error should be handleable.
+			runtime_print::print(
+				"ERROR: Corrupted state: linked map {}: previous value doesn't exist at {:x}",
```
bkchr: We need to call into the else branch below to write the correct head.

thiolliere: Yes, I was mixed about what was the best and conclude the least we touch things the best it is.

Here either:
* we let the head where it is, knowing that the list has been cut in two, with second part having a wrong head.
* move the head to next values potentially cutting a large part of the list.

bkchr: But this means we will print out this message over and over again. (Given there is no modifcation in between)

thiolliere: I don't think this code should attempt to fix stuff, but ok maybe we can do it.

Then we should also maybe:
* remove entry when in storage::get decode has failing, 
* set the next to None when enumeration is broken,

thiolliere: oh ya ok at least we can make previous being None for the next element. indeed

bkchr:
Any update?

thiolliere:
actually I'm confused about what to do:
for remove_linkage it seems we could do:
* if both previous key and next key are corrupted (like the linkage what the actual value corrupted that somehow decoded correctly) then do no touch anything
* else if previous key is corrupted then change head and set the previous of the next one to None.
* else if next key is corrupted then set None to the next of the previous one.

Is that correct ? I still kind of prefer not touching anything if possible but if you agree on this then I make it.

gavofyork:
looks strictly better than before, which is enough for a merge.

thiolliere:
Resolved

PR #4000, opened 2019-11-02
Add `ExistenceRequirement` to `Currency` trait

### LINKED ISSUE #3986: Add Existence Requirement to Transfer
We should add a parameter to `transfer` in the Balances module so that the sender can prefer to have the transaction fail rather than have the account reaped.

Currently, extrinsics pay an `inclusion_fee = base_fee + byte_fee + weight_fee`. This is charged prior to execution of the extrinsic and doesn't know anything about the internal logic of the extrinsic.

The inclusion fee is updated for block `N+1` based on the weight of block `N` compared to the target weight according to [TargetedFeeAdjustment](https://github.com/paritytech/polkadot/blob/v0.6.5/runtime/src/impls.rs#L69L76) so you can calculate the inclusion fee for the next block. However, it is possible that a transaction is in the pool for several blocks before inclusion. The sender may intend to leave enough balance to keep the account but have it reaped.

**Users**

This is a problem for exchanges who manage thousands of accounts. Not having this feature makes them leave a large margin of safety on ED to prevent nonce reset. This is undesirable because it represents a lot of wasted capital.

**Example**

As a concrete example, imagine the following scenario:
```
balance = 10
ED = 1
current_inclusion_fee = 1
next_block_inclusion_fee = 1.5
```
The user attempts to transfer 7 tokens, but the extrinsic is in the pool for several blocks and the `inclusion_fee` increases to 2.1. Substrate then:
- Takes the inclusion fee
- Executes the extrinsic, leaving `balance = 0.9` and reaping the account

The user should be able to add an optional parameter to `transfer` to avoid executing the transaction if it would lead to the account getting reaped.

expenses:
Closes #3986.

### DIFF
--- a/node/executor/src/lib.rs
+++ b/node/executor/src/lib.rs
@@ -468,11 +468,16 @@ mod tests {
 				},
 				EventRecord {
 					phase: Phase::ApplyExtrinsic(1),
+					event: Event::treasury(treasury::RawEvent::Deposit(1984800000000)),
+					topics: vec![],
+				},
+				EventRecord {
+					phase: Phase::ApplyExtrinsic(1),
 					event: Event::balances(balances::RawEvent::Transfer(
 						alice().into(),
 						bob().into(),
 						69 * DOLLARS,
-						1 * CENTS
+						1 * CENTS,
 					)),
 					topics: vec![],
 				},
@@ -512,6 +517,11 @@ mod tests {
 				},
 				EventRecord {
 					phase: Phase::ApplyExtrinsic(1),
+					event: Event::treasury(treasury::RawEvent::Deposit(1984780231392)),
+					topics: vec![],
+				},
+				EventRecord {
+					phase: Phase::ApplyExtrinsic(1),
 					event: Event::balances(
 						balances::RawEvent::Transfer(
 							bob().into(),
@@ -525,6 +535,11 @@ mod tests {
 				EventRecord {
 					phase: Phase::ApplyExtrinsic(1),
 					event: Event::system(system::Event::ExtrinsicSuccess),
+					topics: vec![],
+				},
+				EventRecord {
+					phase: Phase::ApplyExtrinsic(2),
+					event: Event::treasury(treasury::RawEvent::Deposit(1984780231392)),
 					topics: vec![],
 				},
 				EventRecord {
--- a/srml/balances/src/lib.rs
+++ b/srml/balances/src/lib.rs
@@ -407,7 +407,7 @@ decl_module! {
 		) {
 			let transactor = ensure_signed(origin)?;
 			let dest = T::Lookup::lookup(dest)?;
-			<Self as Currency<_>>::transfer(&source, &dest, value)?;
+			<Self as Currency<_>>::transfer(&transactor, &dest, value)?;
 		}
 
 		/// Set the balances of a given account.
--- a/srml/balances/src/lib.rs
+++ b/srml/balances/src/lib.rs
@@ -407,7 +407,7 @@ decl_module! {
 		) {
 			let transactor = ensure_signed(origin)?;
 			let dest = T::Lookup::lookup(dest)?;
-			Self::transfer_inner(&transactor, &dest, value, ExistenceRequirement::AllowDeath)?;
+			<Self as Currency<_>>::transfer(&source, &dest, value)?;
 		}
 
 		/// Set the balances of a given account.
--- a/srml/balances/src/lib.rs
+++ b/srml/balances/src/lib.rs
@@ -1143,7 +1143,7 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
 		transactor: &T::AccountId,
 		dest: &T::AccountId,
 		value: T::Balance,
-		existential_requirement: ExistenceRequirement,
+		existence_requirement: ExistenceRequirement,
 	) -> Result {
 		let from_balance = Self::free_balance(transactor);
 		let to_balance = Self::free_balance(dest);
--- a/srml/balances/src/lib.rs
+++ b/srml/balances/src/lib.rs
@@ -397,6 +397,8 @@ decl_module! {
 		///      `T::OnNewAccount::on_new_account` to be called.
 		///   - Removing enough funds from an account will trigger
 		///     `T::DustRemoval::on_unbalanced` and `T::OnFreeBalanceZero::on_free_balance_zero`.
+		///   - `transfer_keep_alive` works the same way as `transfer`, but has an additional
+		///     check that the transfer will not kill the origin account.
 		///
 		/// # </weight>
 		#[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
@@ -465,12 +467,14 @@ decl_module! {
 			<Self as Currency<_>>::transfer(&source, &dest, value)?;
 		}
 
-		/// Transfer some liquid free balance to another account, while checking that the transfer
-		/// will not kill the account.
+		/// Same as the [`transfer`] call, but with a check that the transfer will not kill the
+		/// origin account.
 		///
-		/// 99% of the time you want `transfer` instead.
+		/// 99% of the time you want [`transfer`] instead.
+		///
+		/// [`transfer`]: struct.Module.html#method.transfer
 		#[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
-		pub fn transfer_some(
+		pub fn transfer_keep_alive(
 			origin,
 			dest: <T::Lookup as StaticLookup>::Source,
 			#[compact] value: T::Balance
@@ -1169,7 +1173,7 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
 		if transactor != dest {
 			if existential_requirement == ExistenceRequirement::KeepAlive {
 				if new_from_balance < Self::minimum_balance() {
-					return Err("payment would kill account");
+					return Err("transfer would kill account");
 				}
 			}
 
--- a/srml/balances/src/lib.rs
+++ b/srml/balances/src/lib.rs
@@ -411,22 +411,6 @@ decl_module! {
 			Self::transfer_inner(&transactor, &dest, value, ExistenceRequirement::AllowDeath)?;
 		}
 
-		/// Transfer some liquid free balance to another account, while checking that the transfer
-		/// will not kill the account.
-		///
-		/// 99% of the time you want `transfer` instead.
-		#[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
-		pub fn transfer_some(
-			origin,
-			dest: <T::Lookup as StaticLookup>::Source,
-			#[compact] value: T::Balance
-		) {
-			let transactor = ensure_signed(origin)?;
-			let dest = T::Lookup::lookup(dest)?;
-
-			Self::transfer_inner(&transactor, &dest, value, ExistenceRequirement::KeepAlive)?;
-		}
-
 		/// Set the balances of a given account.
 		///
 		/// This will alter `FreeBalance` and `ReservedBalance` in storage. it will
@@ -481,6 +465,23 @@ decl_module! {
 			let dest = T::Lookup::lookup(dest)?;
 			<Self as Currency<_>>::transfer(&source, &dest, value)?;
 		}
+		
+		/// Transfer some liquid free balance to another account, while checking that the transfer
+		/// will not kill the account.
+		///
+		/// 99% of the time you want `transfer` instead.
+		#[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
+		pub fn transfer_some(
+			origin,
+			dest: <T::Lookup as StaticLookup>::Source,
+			#[compact] value: T::Balance
+		) {
+			let transactor = ensure_signed(origin)?;
+			let dest = T::Lookup::lookup(dest)?;
+
+			Self::transfer_inner(&transactor, &dest, value, ExistenceRequirement::KeepAlive)?;
+		}
+
 	}
 }
 
--- a/srml/balances/src/lib.rs
+++ b/srml/balances/src/lib.rs
@@ -407,7 +407,24 @@ decl_module! {
 		) {
 			let transactor = ensure_signed(origin)?;
 			let dest = T::Lookup::lookup(dest)?;
-			<Self as Currency<_>>::transfer(&transactor, &dest, value)?;
+
+			Self::transfer_inner(&transactor, &dest, value, ExistenceRequirement::AllowDeath)?;
+		}
+
+		/// Transfer some liquid free balance to another account, while checking that the transfer
+		/// will not kill the account.
+		///
+		/// 99% of the time you want `transfer` instead.
+		#[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
+		pub fn transfer_some(
+			origin,
+			dest: <T::Lookup as StaticLookup>::Source,
+			#[compact] value: T::Balance
+		) {
+			let transactor = ensure_signed(origin)?;
+			let dest = T::Lookup::lookup(dest)?;
+
+			Self::transfer_inner(&transactor, &dest, value, ExistenceRequirement::KeepAlive)?;
 		}
 
 		/// Set the balances of a given account.
@@ -852,42 +869,7 @@ where
 	}
 
 	fn transfer(transactor: &T::AccountId, dest: &T::AccountId, value: Self::Balance) -> Result {
-		let from_balance = Self::free_balance(transactor);
-		let to_balance = Self::free_balance(dest);
-		let would_create = to_balance.is_zero();
-		let fee = if would_create { T::CreationFee::get() } else { T::TransferFee::get() };
-		let liability = match value.checked_add(&fee) {
-			Some(l) => l,
-			None => return Err("got overflow after adding a fee to value"),
-		};
-
-		let new_from_balance = match from_balance.checked_sub(&liability) {
-			None => return Err("balance too low to send value"),
-			Some(b) => b,
-		};
-		if would_create && value < T::ExistentialDeposit::get() {
-			return Err("value too low to create account");
-		}
-		Self::ensure_can_withdraw(transactor, value, WithdrawReason::Transfer.into(), new_from_balance)?;
-
-		// NOTE: total stake being stored in the same type means that this could never overflow
-		// but better to be safe than sorry.
-		let new_to_balance = match to_balance.checked_add(&value) {
-			Some(b) => b,
-			None => return Err("destination balance too high to receive value"),
-		};
-
-		if transactor != dest {
-			Self::set_free_balance(transactor, new_from_balance);
-			if !<FreeBalance<T, I>>::exists(dest) {
-				Self::new_account(dest, new_to_balance);
-			}
-			Self::set_free_balance(dest, new_to_balance);
-			T::TransferPayment::on_unbalanced(NegativeImbalance::new(fee));
-			Self::deposit_event(RawEvent::Transfer(transactor.clone(), dest.clone(), value, fee));
-		}
-
-		Ok(())
+		Self::transfer_inner(transactor, dest, value, ExistenceRequirement::AllowDeath)
 	}
 
 	fn withdraw(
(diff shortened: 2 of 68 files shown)


expenses:
CC @joepetrowski

srml/balances/src/lib.rs
```diff
@@ -851,7 +854,12 @@ where
 		}
 	}
 
-	fn transfer(transactor: &T::AccountId, dest: &T::AccountId, value: Self::Balance) -> Result {
+	fn transfer(
+		transactor: &T::AccountId,
+		dest: &T::AccountId,
+		value: Self::Balance,
+		existence_requirement: ExistenceRequirement,
```
joepetrowski: Not a dev so take this with a grain of salt. The only logic is based around the `KeepAlive` variant, so I wonder if you can make this an optional parameter and then do `if let Some(existence_requirement)`. It would remove the opinion of choosing a default in other places in the codebase.

expenses: Yeah, that's a fair point. Perhaps we could use `AllowDeath` as the default though?

xlc:
This is a major breaking changes breaking JS sdk (the interface of dispatchable call `transfer`) and other substrate-based chains (`Currency` trait).

One way to avoid the breaking change is add a new method and have the existing one calling the new one with a existence requirement `AllowDeath` to preserve existing semantics. The `Currency` trait should also have default implementation to avoid breaking other non substrate runtime modules.

expenses:
yeah, we could perhaps do something like have `transfer(origin, dest, value)` that calls another balance api named something like `transfer_more_complex(origin, dest, value, existence_requirement)` with `ExistenceRequirement`.

node/cli/src/factory_impl.rs
```diff
@@ -150,7 +150,8 @@ impl RuntimeAdapter for FactoryState<Number> {
 			function: Call::Balances(
 				BalancesCall::transfer(
 					indices::address::Address::Id(destination.clone().into()),
-					(*amount).into()
+					(*amount).into(),
+					balances::ExistenceRequirement::AllowDeath
```
kianenigma:
```suggestion
					balances::ExistenceRequirement::AllowDeath,
```

node/cli/src/service.rs
```diff
@@ -533,7 +533,9 @@ mod tests {
 				let version = service.client().runtime_version_at(&best_block_id).unwrap().spec_version;
 				let signer = charlie.clone();
 
-				let function = Call::Balances(BalancesCall::transfer(to.into(), amount));
+				let function = Call::Balances(BalancesCall::transfer(
+					to.into(), amount, balances::ExistenceRequirement::AllowDeath
```
kianenigma:
```suggestion
					to.into(),
					amount,
					balances::ExistenceRequirement::AllowDeath,
```

node/executor/benches/bench.rs
```diff
@@ -137,7 +138,9 @@ fn test_blocks(genesis_config: &GenesisConfig, executor: &NativeExecutor<Executo
 	block1_extrinsics.extend((0..20).map(|i| {
 		CheckedExtrinsic {
 			signed: Some((alice(), signed_extra(i, 0))),
-			function: Call::Balances(balances::Call::transfer(bob().into(), 1 * DOLLARS)),
+			function: Call::Balances(balances::Call::transfer(
+				bob().into(), 1 * DOLLARS, ExistenceRequirement::AllowDeath
```
kianenigma: ditto

node/executor/src/lib.rs
```diff
@@ -368,7 +369,7 @@ mod tests {
 				},
 				CheckedExtrinsic {
 					signed: Some((alice(), signed_extra(0, 0))),
-					function: Call::Balances(balances::Call::transfer(bob().into(), 69 * DOLLARS)),
+					function: Call::Balances(balances::Call::transfer(bob().into(), 69 * DOLLARS, ExistenceRequirement::KeepAlive)),
```
kianenigma:
```suggestion
					function: Call::Balances(default_transfer_call()),
```

node/executor/src/lib.rs
```diff
@@ -390,7 +391,7 @@ mod tests {
...
```
kianenigma:
```suggestion
					function: Call::Balances(default_transfer_call()),
```

srml/balances/src/lib.rs
kianenigma: This would make it also compatible with all the old transactions I believe since `encode(None) == []`, hence the new format can be decoded from the old bytes.

srml/contracts/src/tests.rs
```diff
@@ -35,7 +35,7 @@ use sr_primitives::{
 };
 use support::{
 	assert_ok, assert_err, impl_outer_dispatch, impl_outer_event, impl_outer_origin, parameter_types,
-	storage::child, StorageMap, StorageValue, traits::{Currency, Get},
+	storage::child, StorageMap, StorageValue, traits::{Currency, Get, ExistenceRequirement},
```
kianenigma: I would _maybe_ derive `Default` for `ExistenceRequirement ` and use it here + it makes sense as imo it _actually is the default_ based on the comment above.

srml/executive/src/lib.rs
```diff
@@ -301,7 +301,10 @@ mod tests {
 	};
 	use support::{
 		impl_outer_event, impl_outer_origin, parameter_types, impl_outer_dispatch,
-		traits::{Currency, LockIdentifier, LockableCurrency, WithdrawReasons, WithdrawReason},
+		traits::{
+			Currency, LockIdentifier, LockableCurrency, WithdrawReasons, WithdrawReason,
+			ExistenceRequirement
```
kianenigma:
```suggestion
			ExistenceRequirement, 
```

and place in one line up to 100 chars if possible

srml/executive/src/lib.rs
```diff
@@ -422,7 +425,10 @@ mod tests {
 			balances: vec![(1, 211)],
 			vesting: vec![],
 		}.assimilate_storage(&mut t).unwrap();
-		let xt = sr_primitives::testing::TestXt(sign_extra(1, 0, 0), Call::Balances(BalancesCall::transfer(2, 69)));
+		let xt = sr_primitives::testing::TestXt(
+			sign_extra(1, 0, 0),
+			Call::Balances(BalancesCall::transfer(2, 69, ExistenceRequirement::KeepAlive))
```
kianenigma:
```suggestion
			Call::Balances(BalancesCall::transfer(2, 69, ExistenceRequirement::KeepAlive)),
```

srml/executive/src/lib.rs
```diff
@@ -503,7 +509,10 @@ mod tests {
 	fn bad_extrinsic_not_inserted() {
 		let mut t = new_test_ext(1);
 		// bad nonce check!
-		let xt = sr_primitives::testing::TestXt(sign_extra(1, 30, 0), Call::Balances(BalancesCall::transfer(33, 69)));
+		let xt = sr_primitives::testing::TestXt(
+			sign_extra(1, 30, 0),
+			Call::Balances(BalancesCall::transfer(33, 69, ExistenceRequirement::KeepAlive))
```
kianenigma:
```suggestion
			Call::Balances(BalancesCall::transfer(33, 69, ExistenceRequirement::KeepAlive)),
```

srml/executive/src/lib.rs
```diff
@@ -521,7 +530,10 @@ mod tests {
 	fn block_weight_limit_enforced() {
 		let mut t = new_test_ext(10000);
 		// given: TestXt uses the encoded len as fixed Len:
-		let xt = sr_primitives::testing::TestXt(sign_extra(1, 0, 0), Call::Balances(BalancesCall::transfer(33, 0)));
+		let xt = sr_primitives::testing::TestXt(
+			sign_extra(1, 0, 0),
+			Call::Balances(BalancesCall::transfer(33, 0, ExistenceRequirement::KeepAlive))
```
kianenigma:
```suggestion
			Call::Balances(BalancesCall::transfer(33, 0, ExistenceRequirement::KeepAlive)),
```

srml/executive/src/lib.rs
```diff
@@ -557,9 +570,18 @@ mod tests {
 
 	#[test]
 	fn block_weight_and_size_is_stored_per_tx() {
-		let xt = sr_primitives::testing::TestXt(sign_extra(1, 0, 0), Call::Balances(BalancesCall::transfer(33, 0)));
-		let x1 = sr_primitives::testing::TestXt(sign_extra(1, 1, 0), Call::Balances(BalancesCall::transfer(33, 0)));
-		let x2 = sr_primitives::testing::TestXt(sign_extra(1, 2, 0), Call::Balances(BalancesCall::transfer(33, 0)));
+		let xt = sr_primitives::testing::TestXt(
+			sign_extra(1, 0, 0),
+			Call::Balances(BalancesCall::transfer(33, 0, ExistenceRequirement::KeepAlive))
```
kianenigma:
```suggestion
			Call::Balances(BalancesCall::transfer(33, 0, ExistenceRequirement::KeepAlive)),
```

srml/executive/src/lib.rs
```diff
@@ -557,9 +570,18 @@ mod tests {
...
+		);
+		let x1 = sr_primitives::testing::TestXt(
+			sign_extra(1, 1, 0),
+			Call::Balances(BalancesCall::transfer(33, 0, ExistenceRequirement::KeepAlive))
```
kianenigma:
```suggestion
			Call::Balances(BalancesCall::transfer(33, 0, ExistenceRequirement::KeepAlive)),
```

srml/executive/src/lib.rs
```diff
@@ -557,9 +570,18 @@ mod tests {
...
+		);
+		let x2 = sr_primitives::testing::TestXt(
+			sign_extra(1, 2, 0),
+			Call::Balances(BalancesCall::transfer(33, 0, ExistenceRequirement::KeepAlive))
```
kianenigma:
```suggestion
			Call::Balances(BalancesCall::transfer(33, 0, ExistenceRequirement::KeepAlive)),
```

srml/support/src/traits.rs
```diff
@@ -153,7 +153,7 @@ impl<Imbalance> OnUnbalanced<Imbalance> for () {
 }
 
 /// Simple boolean for whether an account needs to be kept in existence.
-#[derive(Copy, Clone, Eq, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq, Encode, Decode, Debug)]
```
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, Encode, Decode)]
#[cfg_attr(feature = "std", derive(Debug))]
```

Debug only on std, or use the new `RuntimeDebug` (which I am not 100% sure how it works but you can find examples probably cc @tomusdrw )

srml/transaction-payment/src/lib.rs
```diff
@@ -227,7 +227,9 @@ impl<T: Trait + Send + Sync> SignedExtension for ChargeTransactionPayment<T>
 mod tests {
 	use super::*;
 	use codec::Encode;
-	use support::{parameter_types, impl_outer_origin, impl_outer_dispatch};
+	use support::{
+		parameter_types, impl_outer_origin, impl_outer_dispatch, traits::ExistenceRequirement
```
kianenigma:
```suggestion
		parameter_types, impl_outer_origin, impl_outer_dispatch, 
		traits::ExistenceRequirement,
```

(not par the guideline but IMO a very good convention: newline per sub::module)

srml/transaction-payment/src/lib.rs
```diff
@@ -239,7 +241,9 @@ mod tests {
 	use rstd::cell::RefCell;
 	use transaction_payment_rpc_runtime_api::RuntimeDispatchInfo;
 
-	const CALL: &<Runtime as system::Trait>::Call = &Call::Balances(BalancesCall::transfer(2, 69));
+	const CALL: &<Runtime as system::Trait>::Call = &Call::Balances(
+		BalancesCall::transfer(2, 69, ExistenceRequirement::KeepAlive
+	));
```
kianenigma:
```suggestion
	);
```

srml/transaction-payment/src/lib.rs
```diff
@@ -239,7 +241,9 @@ mod tests {
 	use rstd::cell::RefCell;
 	use transaction_payment_rpc_runtime_api::RuntimeDispatchInfo;
 
-	const CALL: &<Runtime as system::Trait>::Call = &Call::Balances(BalancesCall::transfer(2, 69));
+	const CALL: &<Runtime as system::Trait>::Call = &Call::Balances(
+		BalancesCall::transfer(2, 69, ExistenceRequirement::KeepAlive
```
kianenigma:
```suggestion
		BalancesCall::transfer(2, 69, ExistenceRequirement::KeepAlive),
```

srml/utility/src/lib.rs
```diff
@@ -63,7 +63,10 @@ decl_module! {
 mod tests {
 	use super::*;
 
-	use support::{assert_ok, assert_noop, impl_outer_origin, parameter_types, impl_outer_dispatch};
+	use support::{
+		assert_ok, assert_noop, impl_outer_origin, parameter_types, impl_outer_dispatch,
+		traits::ExistenceRequirement
```
kianenigma:
```suggestion
		traits::ExistenceRequirement,
```

srml/balances/src/lib.rs
bkchr: None is does not encode to an empty array.

kianenigma: right it is `0x00`. nevermind this then :D

xlc: @jacogr How does polkadot.js handles missing trailing parameter on a dispatchable call? Does it throw or adds a default value? For later case, make `existence_requirement: Option< ExistenceRequirement>` will be a tolerable change.

xlc:
Transactions are only valid for a given runtime version anyway so old unconfirmed transactions will be invalided anyway on runtime upgrade. So I mostly care about existing polkadot js code still works and existing runtime module outside of Substrate don't need unnecessary breaking changes (i.e. add new method with default implementation to a trait is fine, modify existing will be a breaking change). I understand breaking changes are sometimes unavoidable but in this case it not hard to avoid break one of the most commonly used method (`transfer`).

srml/balances/src/lib.rs
jacogr: It expects all params, so all calls will fail when this is introduced. Even if did unexpected things and added a default value, that would not make sense unless it is an `Option` - in this case it would make sense "not to break". (Not the case, only length check atm, so all expected)

(There are no completely empty types at all, `Option` also needs a single byte to encode none)

gavofyork:
I'm ok with adding `fn transfer_some` which only transfers *some* of the balance, never *all*. If the `value` is too high, then the transfer will fail.

That said, I'm not sure if this really is a pain point given that we now have a means of getting the total fee. @joepetrowski under what conditions could an account accidentally be reaped when the exchange didn't want it to be given that we now have the RPC for determining fees?

joepetrowski:
> That said, I'm not sure if this really is a pain point given that we now have a means of getting the total fee. @joepetrowski under what conditions could an account accidentally be reaped when the exchange didn't want it to be given that we now have the RPC for determining fees?

I thought about that after posting the issue and added a [follow-up comment](https://github.com/paritytech/substrate/issues/3986#issuecomment-548795217) to demonstrate that you can calculate a _max_ fee but not the _exact_ fee as you don't know for how many blocks your tx will be in the queue.

Based on my comment above, for Polkadot, you can calculate the fee within 0.2% for a 50-block mortal, which seems like a reasonable bound.

That said,
- [W3F Research](https://research.web3.foundation/en/latest/polkadot/Token%20Economics/#1-fast-adjusting-mechanism) says that a fast-adjusting fee may have `v = 0.0033`, which would have a margin of 18%. Is this implemented somewhere and I just haven't seen it? (@kianenigma )
- Other Substrate-based chains will have unique means of adjusting fees.

IMO, `fn transfer_some` is not critical but is a nice guarantee to have in that it will (a) reduce maintenance with governance changes and (b) make integration of other Substrate chains easier.

kianenigma:
> @joepetrowski under what conditions could an account accidentally be reaped when the exchange didn't want it to be given that we now have the RPC for determining fees?

A very very very rare but nonetheless possible case that you query the fee, and take that into account, the fee is slightly increased by the time your transaction is included, because, lets say that the network is congested. 

@joepetrowski can't we say that the way to prevent this is to NOT submit an immortal transaction? this way you can know a bound on the fee.

joepetrowski:
> @joepetrowski can't we say that the way to prevent this is to NOT submit an immortal transaction? this way you can know a bound on the fee.

Yeah, immortals should be a no-go in most cases anyway, especially for an exchange as they could easily be replay attacked.

srml/support/src/traits.rs
tomusdrw: Indeed, please use `RuntimeDebug`. It's just regular `derive`.

expenses:
So we're clear on removing the current changes in this PR, and adding a `transfer_some` method that does the same thing?

kianenigma:
Based on the discussion, yes in my opinion. That seems to be the safest option. yet, I won't be shocked if we change/remove it in the future.. but it doesn't seem like a lot of effort to add it so I'd be okay with that. The only complication I see would be for generic-asset. It should just call the normal transfer?  

PS. This will be substrate's 4000'th PR once merged 🚀

expenses:
Alright, I'll get on that. Sorry to disappoint, I think it's only the 4000th issue/PR opened unfortunately, not _that_ special :P

xlc:
generic-asset doesn't have existential deposit policy implemented at all so you shouldn't need to worry about it.

srml/balances/src/lib.rs
```diff
@@ -1151,3 +1133,55 @@ where
 		Self::total_balance(who).is_zero()
 	}
 }
+
+impl<T: Trait<I>, I: Instance> Module<T, I> {
```
joepetrowski: Why not in `impl Currency`?https://github.com/paritytech/substrate/blob/91e73bd9609241a5c7a14a6af9bfe6677a494dc3/srml/balances/src/lib.rs#L789

srml/balances/src/lib.rs
```diff
@@ -1151,3 +1133,55 @@ where
...
+		// NOTE: total stake being stored in the same type means that this could never overflow
+		// but better to be safe than sorry.
+		let new_to_balance = match to_balance.checked_add(&value) {
+			Some(b) => b,
+			None => return Err("destination balance too high to receive value"),
+		};
+
+		if transactor != dest {
+			if existential_requirement == ExistenceRequirement::KeepAlive {
+				if new_from_balance < T::ExistentialDeposit::get() {
```
joepetrowski:
```suggestion
				if new_from_balance < Self::minimum_balance() {
```

srml/balances/src/lib.rs
expenses: I don't want to hassle `Currency` implementors with writing an implementation of `transfer_inner`/`transfer_some`.

srml/balances/src/lib.rs
```diff
@@ -464,6 +464,23 @@ decl_module! {
 			let dest = T::Lookup::lookup(dest)?;
 			<Self as Currency<_>>::transfer(&source, &dest, value)?;
 		}
+
+		/// Transfer some liquid free balance to another account, while checking that the transfer
+		/// will not kill the account.
+		///
+		/// 99% of the time you want `transfer` instead.
+		#[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
+		pub fn transfer_some(
```
thiolliere: I feel that `_some` is not very meaningful suffix ?
maybe transfer_keep_alive.

srml/balances/src/lib.rs
```diff
@@ -464,6 +464,23 @@ decl_module! {
 			let dest = T::Lookup::lookup(dest)?;
 			<Self as Currency<_>>::transfer(&source, &dest, value)?;
 		}
+
+		/// Transfer some liquid free balance to another account, while checking that the transfer
+		/// will not kill the account.
+		///
+		/// 99% of the time you want `transfer` instead.
```
thiolliere: I feel this should more directly point to transfer function or repeat the weight paragraph.
```
		/// Same as `transfer` call, transfer some liquid free balance to another account, but only if from account is kept alive.
        ///
		/// 99% of the time you want `transfer` instead.

File: srml/balances/src/lib.rs (hunk 20 above)
Comment (thiolliere) [reply srml/balances/src/lib.rs]: we could change transfer signature of Currency trait to:
```rust
	fn transfer(transactor: &T::AccountId, dest: &T::AccountId, value: Self::Balance, liveness: ExistenceRequirement) -> Result {
```

srml/balances/src/lib.rs
thiolliere: we could maybe merge both by adding an argument: ExistentialRequirement, but that would require a few more work on UI part I think

expenses: I think we agreed above that not changing `transfer` and only adding an extra `transfer_x` function would work better. Renaming to `transfer_keep_alive` sounds good though.

srml/balances/src/lib.rs
expenses: I think that we should try to avoid breaking anything if we can. See @xlc's first comment in this PR.

thiolliere: Xlc comment is about not changing the Call. We can still have transfer and transfer_keep_alive in Call and have `transfer(...., liveness: ExistenceRequirement)` in Currency trait.

I feel generalizing the trait here doesn't harm

gavofyork: Depends if anyone else has implemented anything on it - third parties might have done. That said, if it is to be changed, now is the time to do it.

srml/balances/src/lib.rs
```diff
@@ -1151,3 +1137,55 @@ where
 		Self::total_balance(who).is_zero()
 	}
 }
+
+impl<T: Trait<I>, I: Instance> Module<T, I> {
+	fn transfer_inner(
+		transactor: &T::AccountId,
+		dest: &T::AccountId,
+		value: T::Balance,
+		existential_requirement: ExistenceRequirement,
```
kianenigma:
```suggestion
		existence_requirement: ExistenceRequirement,
```

or does it just make more sense like this to me? :D

thiolliere:
It seems you need to merge master to make CI happy

PR #4001, opened 2019-11-02
state machine key value storage

cheme:
This PR extract the state machine logic from #3774 to try to make it more reviewable.
This PR cover state machine and its new backend use to handle non state key values.
This is a very limited interface but a next step could be to extend api with the ability to chose a 'collection' (#3774 first target is purely technical : keyspace for child trie, but it could be extended, for instance we could push events targeting offchain).


core/state-machine/src/trie_backend.rs
```diff
@@ -22,17 +22,23 @@ use trie::{Trie, delta_trie_root, default_child_trie_root, child_delta_trie_root
...
 use crate::trie_backend_essence::{TrieBackendEssence, TrieBackendStorage, Ephemeral};
 use crate::Backend;
+use crate::kv_backend::KvBackend;
+use primitives::storage::well_known_keys::CHILD_STORAGE_KEY_PREFIX;
+use std::collections::HashMap;
 
 /// Patricia trie-based backend. Transaction type is an overlay of changes to commit.
-pub struct TrieBackend<S: TrieBackendStorage<H>, H: Hasher> {
+/// A simple key value backend is also accessible for direct key value storage.
+pub struct TrieBackend<S: TrieBackendStorage<H>, H: Hasher, K: KvBackend> {
```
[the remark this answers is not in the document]
cheme: 'KvBackend' ? at start I call this kv storage 'offstate' which was worst, I was also considering deriving the name from 'chain-aux' as it is basically auxilliary storage but synch with chain state. At the very end I did not find good idea, and end up using kv.
One thing I noticed is that the trie backend name could be change: it is use as the main backend of state-machine, notably in client code.
Either changing its name, but considering its relationship with trie_backend_essence, a new struct with could be use with both backend.

core/client/db/src/lib.rs
```diff
@@ -81,7 +83,12 @@ const MIN_BLOCKS_TO_KEEP_CHANGES_TRIES_FOR: u32 = 32768;
 const DEFAULT_CHILD_RATIO: (usize, usize) = (1, 10);
 
 /// DB-backed patricia trie state, transaction type is an overlay of changes to commit.
-pub type DbState = state_machine::TrieBackend<Arc<dyn state_machine::Storage<Blake2Hasher>>, Blake2Hasher>;
+/// A simple key value backend is also accessible for direct key value storage.
+pub type DbState = state_machine::TrieBackend<
+	Arc<dyn state_machine::Storage<Blake2Hasher>>,
+	Blake2Hasher,
+	Arc<dyn state_machine::KvBackend>,
+>;
```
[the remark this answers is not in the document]
cheme: Switching from TrieBackend containing kvbackend and trie backend essence to a 'StateMachineBackend' (not sure about the name) containing trie backend and kv backend can be an idea indeed.

core/state-machine/src/trie_backend.rs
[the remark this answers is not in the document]
thiolliere: oh no KbBackend is ok, what confuse me is that TrieBackend struct contains a KvBackend which is not part of the trie.
The actual name TrieBackend is now confusing, but actually I feel like this struct could be kept as it is and KvBackend could be in another struct, but I might be wrong.

core/state-machine/src/state_backend.rs
```diff
@@ -0,0 +1,248 @@
...
+		}
+	}
+
+	/// Get backend essence reference.
+	pub fn trie(&self) -> &TrieBackend<S, H> {
+		&self.trie
+	}
+
+	/// Get backend storage reference.
+	pub fn backend_storage(&self) -> &S {
```
thiolliere:
```suggestion
	pub fn trie_backend_storage(&self) -> &S {
```

core/state-machine/src/state_backend.rs
```diff
@@ -0,0 +1,248 @@
...
+		&self.kv
+	}
+
+	/// Get key value storage backend mutable reference.
+	pub fn kv_backend_mut(&mut self) -> &mut K {
+		&mut self.kv
+	}
+
+	/// Consumes self and returns underlying storage.
+	pub fn into_storage(self) -> S {
```
thiolliere:
```suggestion
	pub fn into_trie_storage(self) -> S {
```

core/state-machine/src/backend.rs
```diff
@@ -198,10 +230,30 @@ impl<'a, T: Backend<H>, H: Hasher> Backend<H> for &'a T {
...
+
+	fn child_pairs(&self, child_storage_key: &[u8]) -> Vec<(Vec<u8>, Vec<u8>)> {
+		(*self).child_pairs(child_storage_key)
+	}
+
+	fn kv_in_memory(&self) -> crate::InMemoryKvBackend {
+		(*self).kv_in_memory()
+	}
+
+
```
thiolliere:
```suggestion
```

core/state-machine/src/backend.rs
```diff
@@ -283,21 +358,54 @@ impl<H: Hasher> Clone for InMemory<H> {
...
+	}
+
+	fn kv_mut(&mut self) -> &mut InMemoryKvBackend {
+		if let Some(kv) = self.kv.as_mut() {
+			kv
+		} else {
+			self.trie.as_mut().unwrap().kv_backend_mut()
+		}
+	}
+
```
thiolliere:
```suggestion
	/// Extract kv from in memory by emptying it.
```

cheme:

PR #4002, opened 2019-11-02
state-db key value storage

cheme:
This PR extract the state-db logic from #3774 to try to make it more reviewable.
This PR cover state-db additions.

As an extract from #3774, some TODO are present on state machine and client interfaces and tests are lacking.

- dedicated journal  (`KvJournalRecord`) are added (this could be merge in existing journal but focus here on non breaking change) to store block change, and pruning.
- fork are handled with a sequential 'branch id' used as a per value index for memory historic of data (see historical-data tree). 
The branch id are not persisted since this is state db.
- pruning which is more a client problematic is handled in this pr in the state machine test over a test backend.
- apply commit operation got new variants, because some key value content do not make sense.

Note that some pruning enhancement described in #3774 could be added.


cheme:

PR #4003, opened 2019-11-02
client canonical key value storage

cheme:
This PR extract the client and client-db logic from #3774 to try to make it more reviewable.

As an extract from #3774, some TODO are present on state machine and state-db interfaces and tests are not functional.

- client db is using a new column to store `Serialized` sequential history of canonical key values.
History serialization is defined in historical-data crate.
This means that archive all mode is limited to canonical blocks (implementation of archive all seems doable but change the backend storage logic to use a branch indexing).
- client is using key value storage (`StorageAt`) at a targeted block (last canonical block). A typical query involve querying the statedb first, followed by the canonical client storage.


cheme:

PR #4011, opened 2019-11-04
Limit longevity of im-online heartbeats.

tomusdrw:
By talking with @jacogr I realised that the heartbeat transactions effectively live forever in the pool (as we issue `immortal` transactions in the runtime). This PR limits the longevity of those transactions to 128 blocks (~13mins with 6s block time).

### DIFF
--- a/srml/im-online/src/lib.rs
+++ b/srml/im-online/src/lib.rs
@@ -79,7 +79,7 @@ use sr_primitives::{
 	RuntimeDebug,
 	traits::{Convert, Member, Printable, Saturating}, Perbill,
 	transaction_validity::{
-		TransactionValidity, TransactionLongevity, ValidTransaction, InvalidTransaction,
+		TransactionValidity, ValidTransaction, InvalidTransaction,
 		TransactionPriority,
 	},
 };
--- a/srml/im-online/src/lib.rs
+++ b/srml/im-online/src/lib.rs
@@ -74,6 +74,7 @@ use app_crypto::RuntimeAppPublic;
 use codec::{Encode, Decode};
 use primitives::offchain::{OpaqueNetworkState, StorageKind};
 use rstd::prelude::*;
+use rstd::convert::TryInto;
 use session::historical::IdentificationTuple;
 use sr_primitives::{
 	RuntimeDebug,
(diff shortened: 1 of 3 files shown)


srml/im-online/src/lib.rs
```diff
@@ -596,7 +596,7 @@ impl<T: Trait> support::unsigned::ValidateUnsigned for Module<T> {
 				priority: TransactionPriority::max_value(),
 				requires: vec![],
 				provides: vec![(current_session, authority_id).encode()],
-				longevity: TransactionLongevity::max_value(),
+				longevity: 128,
```
bkchr: Hmm, why not make this parameter configurable? Or make it proportional to the length of a session?

Whatever, but I would not hardcode this parameter.

tomusdrw: Fair enough. We can make a parameter for this I suppose, the actual session length is not easy to retrieve, as it's controlled by `ShouldEndSession` handler, so for instance for Kusama it's controlled by babe, but it's fine to just pass it as parameter.

bkchr: AFAIK the session length is fixed. You are right that we call into `Babe` to know if a session should end, but `Babe` internally does not change the length or whatever. 
Even if the session length is not 100% correct, it will not make anything worse in this use case here. The transaction would just be valid for a little bit more time than it should be.

gavofyork:
> heartbeat transactions effectively live forever

shouldn't they be discarded once they become invalid (i.e. once they make it into a block)?

tomusdrw:
@gavofyork yes, if the pool receives the notification about block being imported. I'll implement transaction pool period revalidation to fix issues of stalled transactions in the pool.

srml/im-online/src/lib.rs
```diff
@@ -519,7 +527,11 @@ impl<T: Trait> session::OneSessionHandler<T::AccountId> for Module<T> {
 		where I: Iterator<Item=(&'a T::AccountId, T::AuthorityId)>
 	{
 		// Tell the offchain worker to start making the next session's heartbeats.
-		<GossipAt<T>>::put(<system::Module<T>>::block_number());
+		// Since we consider producing blocks as being online,
+		// the hearbeat is defered a bit to prevent spaming.
+		let block_number = <system::Module<T>>::block_number();
```
bkchr: Why does this prevent spamming?

tomusdrw: Instead of sending heartbeats at the beginning of the session we will wait a bit and only the validators that didn't have a chance to produce a block so far will send the hearbeats. So roughly we will decrease the amount of "spammy" heartbeat transactions by half - I consider all of those transactions spammy, as they are just part of the inner workings and just "take place" for users transactions / introduce a bit of noise.

bkchr: Ahh makes sense :)

PR #4012, opened 2019-11-04
Update RPC deps.

kianenigma:
Another small one while developing tools around the rpc client. 

This might be breaking for
- https://github.com/paritytech/substrate-subxt
- https://github.com/scs/substrate-api-client

the core of the issue is that if your crate depends on a different version of `jsonrpc-core-client`, compared to `substrate-rpc-api` who  also depends on `jsonrpc-core-client`, you get all sorts of `trait not implemented for type` errors, as expected.

### DIFF
--- a/core/test-runtime/src/genesismap.rs
+++ b/core/test-runtime/src/genesismap.rs
@@ -84,7 +84,7 @@ impl GenesisConfig {
 		let mut storage = (map, self.child_extra_storage.clone());
 		let mut config = system::GenesisConfig::default();
 		config.authorities = self.authorities.clone();
-		config.assimilate_storage(&mut storage);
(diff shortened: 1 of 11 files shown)


core/test-runtime/src/genesismap.rs
```diff
@@ -84,7 +84,7 @@ impl GenesisConfig {
 		let mut storage = (map, self.child_extra_storage.clone());
 		let mut config = system::GenesisConfig::default();
 		config.authorities = self.authorities.clone();
-		config.assimilate_storage(&mut storage);
+		config.assimilate_storage(&mut storage).unwrap();
```
bkchr:
```suggestion
		config.assimilate_storage(&mut storage).expect("Adding `system::GensisConfig` to the genesis");
```

PR #4020, opened 2019-11-05
Check aura slot numbers are strictly increasing

NikVolf:


rphmeier:
@NikVolf why is it in-progress?

NikVolf:
well happy path test seem to enough there if @andresilva ok with that

client/consensus/aura/src/lib.rs
```diff
@@ -708,7 +804,7 @@ pub fn import_queue<B, C, P, T>(
 	};
 	Ok(BasicQueue::new(
 		verifier,
-		block_import,
+		Box::new(AuraBlockImport::<_, _, _, P>::new(block_import, client)),
```
rphmeier: The `AuraBlockImport` is only used in the `ImportQueue`, but isn't being used in `start_aura`. Usually, we have the caller instantiate the `BlockImport` outside of the consensus crate to avoid this issue.

NikVolf: Not sure what are you proposing to change

rphmeier: The `AuraBlockImport `should be created outside this function, in the service

NikVolf: @rphmeier  Now when I dug into some of the service code, it starts to make sense (for me), thanks!

client/consensus/aura/src/lib.rs
```diff
@@ -721,6 +807,8 @@ pub fn import_queue<B, C, P, T>(
 	B: BlockT,
 	C: 'static + ProvideRuntimeApi + BlockOf + ProvideCache<B> + Send + Sync + AuxStore,
 	C::Api: BlockBuilderApi<B> + AuraApi<B, AuthorityId<P>> + ApiExt<B, Error = sp_blockchain::Error>,
+	C: 'static + ProvideRuntimeApi + BlockOf + ProvideCache<B> + Send + Sync + AuxStore + HeaderBackend<B>,
```
bkchr: You are replicating the line from 2 lines above here.

bkchr: (mostly)

PR #4025, opened 2019-11-05
Correctly serialize code in chain spec as hex

jimpo:
Due to a bug, the runtime code was previously serialized as a JSON array of numbers, pretty printed one byte per line. This could be seen by running `$ substrate build-spec`.

This fixes the `decl_storage!` proc macro so that the `serde` directive [already in the SRML system module code](https://github.com/paritytech/substrate/blob/master/srml/system/src/lib.rs#L421) is not ignored.

### DIFF
--- a/srml/support/procedural/src/storage/genesis_config/genesis_config_def.rs
+++ b/srml/support/procedural/src/storage/genesis_config/genesis_config_def.rs
@@ -23,9 +23,9 @@ use quote::quote;
 use super::super::{DeclStorageDefExt, StorageLineTypeDef};
 
 pub struct GenesisConfigFieldDef {
-	pub doc: Vec<syn::Meta>,
 	pub name: syn::Ident,
 	pub typ: syn::Type,
+	pub attrs: Vec<syn::Meta>,
 	pub default: TokenStream,
 }
 
@@ -114,17 +114,16 @@ impl GenesisConfigDef {
 				.unwrap_or_else(|| quote!( Default::default() ));
 
 			config_field_defs.push(GenesisConfigFieldDef {
-				doc: line.doc_attrs.clone(),
 				name: config_field,
 				typ,
+				attrs: line.doc_attrs.clone(),
 				default,
 			});
 		}
 
 		for line in &def.extra_genesis_config_lines {
-			let doc = line.attrs.iter()
-				.filter_map(|a| a.parse_meta().ok())
-				.filter(|m| m.name() == "doc")
+			let attrs = line.attrs.iter()
+				.map(|a| a.parse_meta().expect("attribute cannot be parsed"))
 				.collect();
 
 			let default = line.default.as_ref().map(|e| quote!( #e ))
@@ -132,9 +131,9 @@ impl GenesisConfigDef {
 
 
 			config_field_defs.push(GenesisConfigFieldDef {
-				doc,
 				name: line.name.clone(),
 				typ: line.typ.clone(),
+				attrs,
 				default,
 			});
 		}
--- a/srml/support/procedural/src/storage/genesis_config/genesis_config_def.rs
+++ b/srml/support/procedural/src/storage/genesis_config/genesis_config_def.rs
@@ -18,7 +18,7 @@
 
 use srml_support_procedural_tools::syn_ext as ext;
 use proc_macro2::TokenStream;
-use syn::parse_quote;
+use syn::{spanned::Spanned, parse_quote};
 use quote::quote;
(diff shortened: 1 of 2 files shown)


srml/support/procedural/src/storage/genesis_config/genesis_config_def.rs
```diff
@@ -114,27 +114,26 @@ impl GenesisConfigDef {
...
 				default,
 			});
 		}
 
 		for line in &def.extra_genesis_config_lines {
-			let doc = line.attrs.iter()
-				.filter_map(|a| a.parse_meta().ok())
-				.filter(|m| m.name() == "doc")
+			let attrs = line.attrs.iter()
+				.map(|a| a.parse_meta().expect("attribute cannot be parsed"))
```
bkchr: First rule of the macro club, never panic ;)

If we want to accept all attributes here, we don't need to parse them.

Just store `syn::Attribute` in `attrs`.

So:
`let attrs = line.attrs.clone()`

srml/support/procedural/src/storage/genesis_config/genesis_config_def.rs
```diff
@@ -114,27 +114,27 @@ impl GenesisConfigDef {
...
 				default,
 			});
 		}
 
 		for line in &def.extra_genesis_config_lines {
-			let doc = line.attrs.iter()
+			let attrs = line.attrs.iter()
 				.filter_map(|a| a.parse_meta().ok())
-				.filter(|m| m.name() == "doc")
+				.filter(|m| m.name() == "doc" || m.name() == "serde")
```
bkchr: You should just store the `line.attrs` directly and don't do any filtering here.

srml/support/procedural/src/storage/genesis_config/genesis_config_def.rs
```diff
@@ -23,9 +23,9 @@ use quote::quote;
 use super::super::{DeclStorageDefExt, StorageLineTypeDef};
 
 pub struct GenesisConfigFieldDef {
-	pub doc: Vec<syn::Meta>,
 	pub name: syn::Ident,
 	pub typ: syn::Type,
+	pub attrs: Vec<syn::Meta>,
```
bkchr:
```suggestion
	pub attrs: Vec<syn::Attribute>,
```

srml/support/procedural/src/storage/genesis_config/genesis_config_def.rs
jimpo: So there's two issues. One is that depending on the attribute, you might need to apply it both to the field declaration and the default assignment (for like a `cfg` attribute). Or that some attributes might not make sense at all. The other is that attrs is filtered for doc attrs in the regular storage lines (like not the add_extra_genesis lines) and so it's a `syn::Meta` there, not a `syn::Attribute`. So that would also have to be unrestricted if we made it `Attribute`. What do you think?

bkchr: Okay, then we should do the following:
```
line.attrs.iter().filter_map(|a| a.parse_meta().ok()).try_for_each(|m| if m.name() == "cfg" { Error::new("cfg not supported, bla, good error message") } else { Ok(()) })?;
```

bkchr: If that check is successful, we store the `Vec<Attribute>`, as said before.

srml/support/procedural/src/storage/genesis_config/genesis_config_def.rs
```diff
@@ -114,31 +116,40 @@ impl GenesisConfigDef {
...
 
 		for line in &def.extra_genesis_config_lines {
-			let doc = line.attrs.iter()
-				.filter_map(|a| a.parse_meta().ok())
-				.filter(|m| m.path().is_ident("doc"))
-				.collect();
+			let attrs = line.attrs.iter()
+				.map(|attr| {
+					let meta = attr.parse_meta()?;
+					if meta.path().is_ident("doc") || meta.path().is_ident("serde") {
```
bkchr: Can you please do it like I proposed? 

Next week the next one is coming and want to use feature xy for whatever reason. We just need to disallow `#[cfg()]` for now.

jimpo:
@bkchr All of the chain spec files I see in the substrate and Polkadot repos ([including Kusama](https://raw.githubusercontent.com/paritytech/polkadot/master/service/res/kusama.json)) have raw genesis data, not runtime genesis data that uses this serialization. So I think it's not an issue?

bkchr:
Good point.

PR #4026, opened 2019-11-05
Treasury rewards should pay the remainder of the 10%

gavofyork:
This alters the functionality of staking so that 10% inflation per year is fixed and any of the 10% not paid to stakers instead goes to the treasury.

It no longer attempts to keep stakers "only at" the level of inflation, as before.

This also adds a new event to the `treasury` called `Deposit`, given out when funds are received through the `Treasury` imbalance handler.

### DIFF
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -241,7 +241,7 @@ impl session::historical::Trait for Runtime {
 srml_staking_reward_curve::build! {
 	const REWARD_CURVE: PiecewiseLinear<'static> = curve!(
 		min_inflation: 0_025_000,
-		max_inflation: 0_100_000,
+		max_inflation: 0_100_000,   // 10% - must be equal to MaxReward below.
 		ideal_stake: 0_500_000,
 		falloff: 0_050_000,
 		max_piece_count: 40,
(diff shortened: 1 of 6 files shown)


node/runtime/src/lib.rs
```diff
@@ -253,20 +253,23 @@ parameter_types! {
 	pub const SessionsPerEra: sr_staking_primitives::SessionIndex = 6;
 	pub const BondingDuration: staking::EraIndex = 24 * 28;
 	pub const RewardCurve: &'static PiecewiseLinear<'static> = &REWARD_CURVE;
+	pub const MaxReward: Perbill = Perbill::from_percent(10);
+	// ^^^ 10% - must be equal to max_inflation, above.
```
gavofyork: This is annoyingly the same value in two different places; the `curve!` macro requires a literal and we can't extract the parameters once it's created.

thiolliere: The only way I see to remove this duplication is:
* make max_reward perbill an output of curve macro
* (or wait for more const in rust: all arithmetic operation needed by curve could potentially be const)
* (or make the curve build at runtime)

srml/treasury/src/lib.rs
```diff
@@ -346,29 +348,12 @@ impl<T: Trait> Module<T> {
...
 
-/// Mint extra funds for the treasury to keep the ratio of portion to total_issuance equal
-/// pre dilution and post-dilution.
-///
-/// i.e.
-/// ```nocompile
-/// portion / total_issuance_before_dilution ==
-///   (portion + minted) / (total_issuance_before_dilution + minted_to_treasury + minted)
-/// ```
-impl<T: Trait> OnDilution<BalanceOf<T>> for Module<T> {
```
kianenigma: If we are to not have this trait and just use imbalances for it (as it seems here), I think it can be removed as a trait (have not done a clone build but should already give some warnings afaiu)

Issue #4013, opened 2019-11-04
Implement Send and Sync for TestExternalitities


geigerzaehler:
It would be helpful if `substrate_state_machine::TestExternalities` implemented `Send` and `Sync` so it can be used in more elaborate tests.

In my particular use case I’m creating a client test double that simulates the node runtime with `TestExternalities`. I would like to use the client across different threads which requires the `Send` for `TestExternalities`.

bkchr:
Hmmm. We could probably make it `Send`, but probably not `Sync`.

Why don't you just add a function that creates the externalities and you call this function from your threads? Then you don't need it to have `Send` implemented.

geigerzaehler:
> Why don't you just add a function that creates the externalities and you call this function from your threads? Then you don't need it to have `Send` implemented.

Unfortunately this does not allow me to share the externalities between threads. In my example I’m running a webserver where I want to provide the test double as a “request context”. The webserver requires the context to implement `Send` because it spawns multiple threads. So I’m limited by the API the server gives me.

One way to work around this would obviously be to call the text externalities only in a single fixed thread and implement an adapter that uses message passing to call the externalities from a different thread. Implementing this adapter is not easy though. It would be a lot easier if TestExternalities implemented `Send`.

bkchr:
Do you modify the externalities somewhere in your server or what are you doing with them?

geigerzaehler:
> Do you modify the externalities somewhere in your server or what are you doing with them?

Yes. This is roughly the setup we have
```rust
pub struct Client {
    test_ext: TestExternalities,
}

impl Client {
    fn transfer(&mut self) {
        self.test_ext.execute_with(|| {
            // runtime code
        })
    }
}

fn main() {
    let test_ext = TestExternalities::new();
    let client = Arc::new(Mutex::new(Client { test_ext }));
    let service = warp::any().map(move || {
        client.clone().lock().unwrap().transfer();
        warp::reply()
    });
    warp::serve(service).run(([127, 0, 0, 1], 8080))
}
```
This fails to compile. The closure passed to `map` may be run in another trait to `client` needs to implement `Send`. But this is only the case if `TestExternalities` implements `Send` too.

PR #4030, opened 2019-11-06
Make `TestExternalities` implement `Send`

bkchr:
Fixes: https://github.com/paritytech/substrate/issues/4013
CC @geigerzaehler

### DIFF
--- a/core/externalities/src/extensions.rs
+++ b/core/externalities/src/extensions.rs
@@ -27,7 +27,12 @@ use std::{collections::HashMap, any::{Any, TypeId}, ops::DerefMut};
 ///
 /// As extensions are stored as `Box<Any>`, this trait should give more confidence that the correct
 /// type is registered and requested.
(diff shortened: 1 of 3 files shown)


core/externalities/src/extensions.rs
```diff
@@ -51,7 +56,11 @@ macro_rules! decl_extension {
 		$( #[ $attr ] )*
 		$vis struct $ext_name (pub $inner);
 
-		impl $crate::Extension for $ext_name {}
+		impl $crate::Extension for $ext_name {
+			fn as_mut_any(&mut self) -> &mut dyn std::any::Any {
+				self
```
tomusdrw: Can't this be a default implementation?

bkchr: No, that doesn't work :(

geigerzaehler:
Awesome. Thanks!

Issue #3623, opened 2019-09-16
Improvements for verifier fork-choice applications


rphmeier:
Verifiers, being inherently parallelizable, should not be making decisions about fork-choice, since they don't have up-to-date information about the best block and any fork-choice that they make could be pre-empted by another import.

Two items:
  - The verifier shouldn't have the option to return a fork-choice in `BlockImportParams`. fork-choice determination be done completely within the `BlockImport` structure, because it does sequential import of blocks.
  - It would be nice for the verifier to return data that is passed through the pipeline to the block import down the line. Some investigation is needed to see how this can be done with minimal impact on `BlockImport` composability.

PR #4031, opened 2019-11-06
Allow import withouth state verification

arkpar:
This is another take on reverted #3942
Instead of doing header-only sync we do regular sync here and skip state enactment if parent state is not available.

### DIFF
--- a/core/client/src/client.rs
+++ b/core/client/src/client.rs
@@ -967,7 +967,7 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 				}
 				storage_changes
 			},
-			_ => Default::default()
+			_ => None,
 		};
 
 		let is_new_best = finalized || match fork_choice {
--- a/core/client/src/client.rs
+++ b/core/client/src/client.rs
@@ -838,15 +838,22 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 			finalized,
 			auxiliary,
 			fork_choice,
+			allow_missing_state,
 		} = import_block;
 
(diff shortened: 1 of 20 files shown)


gavofyork:
Master merge is need to fix the test

rphmeier:
This is higher priority than switching to async/await.

core/client/src/client.rs
```diff
@@ -927,22 +936,39 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
+				}
+				if let Some(storage_changes) = storage_changes.clone() {
+					operation.op.update_storage(storage_changes.0, storage_changes.1)?;
+				}
+				if let Some(Some(changes_update)) = changes_update {
+					operation.op.update_changes_trie(changes_update)?;
+				}
+				storage_changes
+			},
+			_ => Default::default()
```
rphmeier:
```suggestion
			_ => None,
```

I'd prefer explicitness for readability

core/finality-grandpa/src/tests.rs
```diff
@@ -984,6 +985,7 @@ fn allows_reimporting_change_blocks() {
 			bad_justification: false,
 			needs_finality_proof: false,
 			is_new_best: true,
+			header_only: false,
```
rphmeier: why is any of this exposed to the consensus engine (`header_only`, `allow_missing_state`), which only cares about headers?

arkpar: Same reasons as #3623. Fixing the pipeline is out of scope of this PR.

PR #4035, opened 2019-11-06
Decrease peer reputation on bad transactions

arkpar:


core/service/src/lib.rs
```diff
@@ -629,10 +636,13 @@ where
 				let import_future = import_future
 					.then(move |import_result| {
 						match import_result {
-							Ok(_) => report_handle.report_peer(who, reputation_change),
+							Ok(_) => report_handle.report_peer(who, reputation_change_good),
 							Err(e) => match e.into_pool_error() {
 								Ok(txpool::error::Error::AlreadyImported(_)) => (),
-								Ok(e) => debug!("Error adding transaction to the pool: {:?}", e),
+								Ok(e) => {
```
marcio-diaz: Is there a problem when decreasing reputation of already imported? (why did you change it?)

arkpar: Transactions are propagated by push. We can't know in advance if the peer we're sending to already has it. So we don't punish or reward for duplicates.

PR #4036, opened 2019-11-06
Remove dependencies on OpenSSL

DemiMarie-parity:
We really really don’t want to depend on OpenSSL, as that requires us
make a new release for non-Linux platforms whenever an upstream OpenSSL
security vulnerability affects us.

We currently depend on OpenSSL only via the `native-tls` crate, which
we depend on via `hyper-tls` and `tokio-tls`.  This PR is a
work-in-progress because it only removes the first dependency.  Removing
the second one will require a change to the `websocket` crate.

### DIFF
--- a/core/offchain/src/api/http.rs
+++ b/core/offchain/src/api/http.rs
@@ -47,11 +47,10 @@ pub fn http() -> (HttpApi, HttpWorker) {
 		requests: FnvHashMap::default(),
 	};
 
-	let tls = hyper_rustls::HttpsConnector::new(1);
 	let engine = HttpWorker {
 		to_api,
 		from_api,
-		http_client: hyper::Client::builder().build(tls),
+		http_client: hyper::Client::builder().build(hyper_rustls::HttpsConnector::new(1)),
 		requests: Vec::new(),
 	};
 
--- a/core/offchain/src/api/http.rs
+++ b/core/offchain/src/api/http.rs
@@ -47,12 +47,11 @@ pub fn http() -> (HttpApi, HttpWorker) {
 		requests: FnvHashMap::default(),
 	};
 
+	let tls = hyper_rustls::HttpsConnector::new(1);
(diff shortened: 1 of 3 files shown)


core/offchain/src/api/http.rs
```diff
@@ -554,9 +554,7 @@ enum WorkerToApi {
 /// Wraps around a `hyper::Client` with either TLS enabled or disabled.
 enum HyperClient {
```
tomaka: We should remove the `HyperClient` enum entirely now.

DemiMarie-parity:
@tomaka can you re-review?

core/offchain/src/api/http.rs
```diff
@@ -47,12 +47,11 @@ pub fn http() -> (HttpApi, HttpWorker) {
...
 	};
 
+	let tls = hyper_rustls::HttpsConnector::new(1);
 	let engine = HttpWorker {
 		to_api,
 		from_api,
-		// TODO: don't unwrap; we should fall back to the HttpConnector if we fail to create the
-		// Https one; there doesn't seem to be any built-in way to do this
-		http_client: HyperClient::new(),
+		http_client: hyper::Client::builder().build(tls),
```
tomaka:
```suggestion
		http_client: hyper::Client::builder().build(hyper_rustls::HttpsConnector::new(1)),
```

core/offchain/src/api/http.rs
```diff
@@ -47,12 +47,11 @@ pub fn http() -> (HttpApi, HttpWorker) {
 		requests: FnvHashMap::default(),
 	};
 
+	let tls = hyper_rustls::HttpsConnector::new(1);
```
tomaka:
```suggestion
```

DemiMarie-parity:
@tomusdrw [`jsonrpsee` is not on `crates.io`](https://crates.io/crates/jsonrpsee)

PR #4037, opened 2019-11-06
Ban incoming invalid transactions

arkpar:


gavofyork:
@arkpar I presume this won't indefinitely ban transactions which are just slightly before their time (e.g. Nonce too high)?

arkpar:
No, the ban is temporary

PR #4042, opened 2019-11-07
Add a --no-private-ipv4 CLI option

tomaka:
Fix https://github.com/paritytech/polkadot/issues/543

If passed, the `--no-private-ipv4` CLI option will prevent the networking from trying to connect to the ranges `10.0.0.0/8`, `172.16.0.0/12` and `192.168.0.0/16`.

### DIFF
--- a/core/cli/src/lib.rs
+++ b/core/cli/src/lib.rs
@@ -343,7 +343,7 @@ impl<'a> ParseAndPrepareBuildSpec<'a> {
 			];
 			spec.add_boot_node(addr)
 		}
-		
+
 		let json = service::chain_ops::build_spec(spec, raw_output)?;
 
 		print!("{}", json);
@@ -625,6 +625,7 @@ fn fill_network_configuration(
 
 	config.transport = TransportConfig::Normal {
 		enable_mdns: !is_dev && !cli.no_mdns,
+		allow_private_ipv4: !cli.no_private_ipv4,
 		wasm_external_transport: None,
 	};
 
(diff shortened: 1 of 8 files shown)


tomaka:
Thinking about it, should we maybe do `--only-global-ipv4` instead?
This would also cover `127.0.0.0/8`, `169.254.0.0/16` (loopback), and so on.

arkpar:
> Thinking about it, should we maybe do `--only-global-ipv4` instead?
> This would also cover `127.0.0.0/8`, `169.254.0.0/16` (loopback), and so on.

Don't we filter these out in discovery already?

arkpar:
Would that prevent the peerset from connecting to already known private addresses when using `--no-private-ipv4`? It should be possible to switch this CLI option on and off without clearing the peerset file.

tomaka:
> Don't we filter these out in discovery already?

No, as they might be legitimate.
Maybe we should filter them out of Kademlia, but I'm not sure if it's ok to do so.

tomaka:
> Would that prevent the peerset from connecting to already known private addresses when using --no-private-ipv4? It should be possible to switch this CLI option on and off without clearing the peerset file.

We don't have a peerset file yet, but yes that would prevent all already-known addresses from being attempted.

tomaka:
cc @joepetrowski 
Devops might want to enable this option.

gregorst3:
@tomaka  Hello, there is an internal ip range missing: 100.64.0.0/10  https://datatracker.ietf.org/doc/html/rfc6598.

PR #4044, opened 2019-11-07
Move required CreateTransaction::Public bounds to the trait definition.

tomusdrw:
When @gnunicorn attempted to use the trait it turned out that the `where` clause makes it really cumbersome (close to impossible, when combined with `decl_module` macros).

This PR moves the bound to `CreateTransaction` trait to alleviate that.

Still a draft, cause we are figuring out if there can be any more changes to increase the ergonomy of this.


bkchr:
If someone is interested, the corresponding rust issue: https://github.com/rust-lang/rust/issues/44591 :D

PR #4045, opened 2019-11-07
Increase parallel downloads to 5

arkpar:
Logs show that sometimes all 3 peers timeout.


core/network/src/protocol/sync.rs
```diff
@@ -1228,7 +1234,7 @@ fn peer_block_request<B: BlockT>(
 	attrs: &message::BlockAttributes,
 	major_sync: bool,
 ) -> Option<(Range<NumberFor<B>>, BlockRequest<B>)> {
-	let max_parallel = if major_sync { 1 } else { 3 };
+	let max_parallel = if major_sync { 1 } else { 5 };
```
bkchr: Why not make this configurable and when we have found a reasonable value, we can set this?

PR #4049, opened 2019-11-08
Fix in_memory backend for child tries.

thiolliere:
when converting the in memory to a triebackend, if the main trie is empty then child trie root hash were not inserted into the main trie.

This can happen for instance with an empty trie with only one transaction:
```rust
vec![(Some(b"1".to_vec()), b"2".to_vec(), Some(b"3".to_vec()))]
```
But actually I'm not completely sure such a transaction is valid, maybe transaction must include the value to write in the main trie.

### DIFF
--- a/core/state-machine/src/backend.rs
+++ b/core/state-machine/src/backend.rs
@@ -246,10 +246,11 @@ impl error::Error for Void {
 	fn description(&self) -> &str { "unreachable error" }
 }
 
-/// In-memory backend. Fully recomputes tries on each commit but useful for
-/// tests.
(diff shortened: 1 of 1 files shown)


gavofyork:
how is that a transaction?

thiolliere:
> how is that a transaction?

are you saying such a transaction is invalid or what transaction means in this context ?

I meant transaction as a storage transaction:
https://github.com/paritytech/substrate/blob/6fb3758336e74e4979a92e1644301adf9be8870c/core/state-machine/src/backend.rs#L364-L366

PR #4050, opened 2019-11-08
grandpa: progressively increase target gossip peers

andresilva:
Currently in GRANDPA gossip we have the notion of politeness to restrict who we should gossip messages to (i.e. only send messages to nodes that are close to the messages' round). Still, assuming that all nodes are up-to-date with the latest round we will potentially gossip messages to all of the peers we are connected to. This results in a lot of churn in the network as we keep receiving the same message multiple times from different peers.

This PR adds logic for restricting the number of peers we gossip to and progressively increase the set as necessary (i.e. if GRANDPA doesn't make any progress). The logic for filtering peers follows the given state transitions:
- State 0: don't send to anyone (only if our local node is not an authority)
- State 1: send to random `sqrt(authorities)`
- State 2: send to all authorities
- State 3: send to random `sqrt(non-authorities)`
- State 4: send to all non-authorities

Transitions will be triggered on repropagation attempts by the underlying gossip layer, which should happen every 30 seconds. Assuming GRANDPA is live and making progress these messages should be expired before any repropagation attempts happen.

The current implementation is a bit hackish, but as we have already discussed refactoring consensus gossip to extract it from the network layer, we should have more control over the implementation and be able to implement this logic more cleanly afterwards.

### DIFF
--- a/core/finality-grandpa/src/communication/gossip.rs
+++ b/core/finality-grandpa/src/communication/gossip.rs
@@ -92,6 +92,7 @@ use substrate_telemetry::{telemetry, CONSENSUS_DEBUG};
 use log::{trace, debug, warn};
 use futures::prelude::*;
 use futures::sync::mpsc;
+use rand::Rng;
 
 use crate::{environment, CatchUp, CompactCommit, SignedMessage};
 use super::{cost, benefit, Round, SetId};
@@ -483,6 +484,14 @@ impl<N: Ord> Peers<N> {
 	fn peer<'a>(&'a self, who: &PeerId) -> Option<&'a PeerInfo<N>> {
 		self.inner.get(who)
 	}
+
+    fn authorities(&self) -> usize {
+        self.inner.iter().filter(|(_, info)| info.roles.is_authority()).count()
+    }
+
+    fn non_authorities(&self) -> usize {
+        self.inner.iter().filter(|(_, info)| !info.roles.is_authority()).count()
+    }
 }
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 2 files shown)


core/finality-grandpa/src/communication/gossip.rs
```diff
@@ -1183,6 +1255,13 @@ impl<Block: BlockT> network_gossip::Validator<Block> for GossipValidator<Block>
 				Some(x) => x,
 			};
 
+            if let MessageIntent::Broadcast { previous_attempts } = intent {
+                // early return if the message isn't allowed at this stage.
+                if !inner.message_allowed(peer, previous_attempts) {
```
andresilva: This filter is currently applied before the GRANDPA politeness filter, should we do it afterwards instead?

core/finality-grandpa/src/communication/gossip.rs
```diff
@@ -980,6 +989,69 @@ impl<Block: BlockT> Inner<Block> {
...
+        if peer.roles.is_authority() {
+            let authorities = self.peers.authorities();
+
+            // the target node is an authority, on the first attempt we start by
+            // sending the message to only `sqrt(authorities)` (if we're
+            // connected to at least `MIN_AUTHORITIES`).
+            if previous_attempts == 0 && authorities > MIN_AUTHORITIES {
+                let authorities = authorities as f64;
+                let p = (authorities.sqrt()).max(MIN_AUTHORITIES as f64) / authorities;
+                rand::thread_rng().gen_bool(p)
```
arkpar: This is not exactly same as selecting exactly sqrt random peers and sending to them. I think it is still fine here though.

core/finality-grandpa/src/communication/gossip.rs
arkpar: This check is quicker, isn't it?

arkpar:
Spaces are used instead of tabs in a few places. Integration test fails. Perhaps it needs to be changed to include a few sentry peers.

core/finality-grandpa/src/communication/gossip.rs
andresilva: Yeah, unfortunately there's no easy way to do this with the current interface, should be doable after refactoring consensus gossip.

andresilva:
It seems that editorconfig got broken in my editor hence the spaces. Looking into the integration test now.

andresilva:
I updated the filtering logic to make it more lenient for global messages since there should be a lot less global messages than round messages (just commits and in principle just 1 per round), and we want these to propagate to non-authorities fast enough so that they can observe finality. The logic for filtering global messages follows the given state transitions:

- State 0: send to random `sqrt(authorities)` ++ `sqrt(non-authorities)`.
- State 1: send to all authorities
- State 2: send to all non-authorities

PR #4052, opened 2019-11-08
Translation for linked map

rphmeier:
Similar to #3947 , but for `LinkedMap`.

I did a lot of cleanup also of code that didn't do a great job of using Rust traits... I mostly made a bunch of stuff more permissive so that we could use the definition of a `StorageMap<K, V>` to also look up keys for a `StorageMap<K2, V2>`, which is necessary for translation. Most of that hinges around an internal `KeyFormat` trait that every `StorageMap` has as an assoc. type, which is un-opinionated about the actual types passed to it.

Still needs a test, but is ready for review.


rphmeier:
I didn't touch any code w.rt. declaration of storage values, but i've got a new compilation error where it can't infer the type of a default `Vec` from a `JustVec: Vec<u32>`. Any ideas?

srml/support/src/storage/generator/linked_map.rs
```diff
@@ -47,31 +47,59 @@ pub trait StorageLinkedMap<K: FullCodec, V: FullCodec> {
...
 
 	/// Generate the hashed key for head
 	fn storage_linked_map_final_head_key() -> <Self::Hasher as StorageHasher>::Output {
-		Self::Hasher::hash(Self::head_key())
+		<Self::KeyFormat as KeyFormat>::storage_linked_map_final_head_key()
+	}
+}
+
+/// A type-abstracted key format used for a family of linked-map types.
+pub trait KeyFormat {
```
thiolliere: > I mostly made a bunch of stuff more permissive so that we could use the definition of a StorageMap<K, V> to also look up keys for a StorageMap<K2, V2>, which is necessary for translation. Most of that hinges around an internal KeyFormat trait that every StorageMap has as an assoc. type, which is un-opinionated about the actual types passed to it.

I'm not sure I understand why you introduce this trait ? to access the old value before the translation ? because in this PR it seems the whole code (both translation implementation and the test) doesn't make use of this KeyFormat extraction.

thiolliere:
> I didn't touch any code w.rt. declaration of storage values, but i've got a new compilation error where it can't infer the type of a default Vec from a JustVec: Vec<u32>. Any ideas?

Is that related to new rust release ? which version are you using ?

srml/support/src/storage/generator/linked_map.rs
rphmeier: it is used, heavily. 

> to access the old value before the translation ?

yes, exactly.

rphmeier:
> rustc 1.37.0 (eae3437df 2019-08-13)

srml/support/src/storage/generator/linked_map.rs
thiolliere: why would you want to access it before the translation ? isn't translation the first and only call you need to do ?

rphmeier: Yes, translation is the first and only call you need to do. `KeyFormat` is used within the body of the `translate` function to access the old values before the translation has actually occurred - that's how I interpreted "accessing" the old value before translations.

thiolliere: oh ok then I would have just introduced two private function
```rust
fn read_head<Key, Value>(key: key, prefix: &'static [u8]) -> Value {...}
fn final_key<Key>(key: key, prefix: &'static [u8]) -> Vec<u8> {...}
```
instead of the new trait but ok.

srml/support/procedural/src/storage/instance_trait.rs
```diff
@@ -28,6 +28,7 @@ pub(crate) const DEFAULT_INSTANTIABLE_TRAIT_NAME: &str = "__GeneratedInstantiabl
 // prefix for consts in trait Instance
 pub(crate) const PREFIX_FOR: &str = "PREFIX_FOR_";
 pub(crate) const HEAD_KEY_FOR: &str = "HEAD_KEY_FOR_";
+pub(crate) const LINKED_MAP_KEY_FORMAT_FOR: &str = "LINKED_MAP_KEY_FORMAT_FOR_";
```
thiolliere: I don't think this should be in instance_trait, it is not related to instance trait it is just the name of a struct to generate along side storage struct.

The two other keywords are prefix for consts in the trait Instance.

srml/support/procedural/src/storage/storage_struct.rs
```diff
@@ -156,21 +156,51 @@ pub fn decl_and_impl(scrate: &TokenStream, def: &DeclStorageDefExt) -> TokenStre
...
+					);
+
+					let format_type = quote!( #name<#optional_instance>);
+
+					let optional_phantom = optional_instance
+						.as_ref()
+						.map(|instance| quote!( #scrate::rstd::marker::PhantomData<#instance> ));
+
+					let struct_decl = quote!(
+						#visibility struct #format_type(#optional_phantom);
```
thiolliere: Lacks documentation, user see a new struct named LINKED_MAP_KEY_FORMAT_FOR_MyStorage with no other comment

rphmeier:
> I agree this KeyFormat could be used for further new features but I would prefer to introduce it when we see the usecase.

I didn't just add it for fun; it's necessary to implement `translate`.

srml/support/src/storage/generator/linked_map.rs
rphmeier: Including those as free functions (i.e. fixing logic for key generation) defeats the purpose of having a trait function for computing the final keys in the first place. The point of having `fn storage_linked_map_final_key` is so that the implementation can decide how to compute the key. 

All of the changes I included are a refactoring - the code before this diff seriously misunderstood Rust traits.

srml/support/src/storage/generator/linked_map.rs
```diff
@@ -360,4 +401,50 @@ where
...
+		loop {
+			let old_raw_key = G::KeyFormat::storage_linked_map_final_key(prefix, &current_key);
+			let (val, linkage) = read_with_linkage::<K2, V2>(old_raw_key.as_ref()).ok_or(())?;
+			let next = linkage.next.clone();
+
+			let val = translate_val(val);
+			let linkage = translate_linkage(linkage);
+
+			// now that we've successfully migrated the storage, kill the old key.
+			unhashed::kill(old_raw_key.as_ref());
```
bkchr: You are removing the old one, without knowing if there will be an error.

rphmeier:
>  I think translation should be done in memory and when everything is correct, we should write everything at once.

I disagree, because the map could be really big.

srml/support/src/storage/generator/linked_map.rs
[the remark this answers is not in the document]
rphmeier: 1. this is for linked maps only
2. i don't understand. this is a `linked_map::KeyFormat`. `for a family of linked-map types`.

It's a type-abstracted interface in the sense that one implementation can be used for all possible linked-map keys.

srml/support/src/storage/generator/linked_map.rs
```diff
@@ -360,4 +401,50 @@ where
 			Ok(len)
 		}
 	}
+
+	fn translate<K2, V2, TK, TV>(translate_key: TK, translate_val: TV) -> Result<(), ()>
+		where K2: FullCodec + Clone, V2: Decode, TK: Fn(K2) -> K, TV: Fn(V2) -> V
```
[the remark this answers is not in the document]
rphmeier: no, it's part of the trait definition.

srml/support/src/storage/generator/linked_map.rs
rphmeier: The only possible error afterwards is if the next iteration of the loop doesn't decode correctly with `K2` and `V2`. We've already decoded correctly as `K2` and `V2`, so there are only 2 ways that can happen:
  - `K2` and `V2` aren't actually the correct old format and only decoded before by chance.
  - The storage is in a malformed state already. We can't fix that.

srml/support/src/storage/generator/linked_map.rs
```diff
@@ -360,4 +401,50 @@ where
...
+			let new_raw_key = G::storage_linked_map_final_key(&translate_key(current_key.clone()));
+			unhashed::put(new_raw_key.as_ref(), &(&val, &linkage));
+
+			match next {
+				None => break,
+				Some(next) => { current_key = next },
+			}
+		}
+
+		// update head at end, since it means that we made it through the entire
```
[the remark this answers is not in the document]
rphmeier: we would expect to bail on the first implementation of the loop and not afterwards

srml/support/src/storage/generator/linked_map.rs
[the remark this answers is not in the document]
bkchr: Yeah, sorry, I misinterpreted something.

gavofyork:
Indeed. Also, the typical use case for this would be in migration. If there were an unexpected error, then it wouldn't be able to panic (as that would brick the chain). It also couldn't not migrate, since the rest of the runtime would be expecting the new format. Realistically, it would probably just try to ignore it and reset the item to default, therefore it's probably fine to write as we go along in practice and not allow for the possibility of an atom write/rollback.

bkchr:
We should at least remove all the old entries, so nothing stays in the storage. The caller than would be responsible for inserting a default value or whatever.

rphmeier:
> We should at least remove all the old entries, so nothing stays in the storage. The caller than would be responsible for inserting a default value or whatever.

This is actually impossible, isn't it? We've failed to read the linkage for some entry, which means we can't iterate the map further. The only thing that seems possible to do is to ensure that the last value's linkage is set to `None`. The rest of the `linked_map` will be orphaned in storage, but it's clear we don't know how to read it anymore.

PR #4055, opened 2019-11-08
Batch gossip messages

arkpar:
Should help with tokio overhead and allocation fragmentation


andresilva:
There's a test that needs fixing, was changed in master.

PR #4057, opened 2019-11-08
Substrate runtime interface 2.0

bkchr:
The current runtime interface lives in `sr-io` and `wasm-executor`. Any change that is done in `sr-io` requires manual changes to the interface declaration in `wasm-executor`. This pr introduces an attribute proc macro `#[runtime_interface]` for declaring and implementing a runtime interface. The runtime interface is declared like a nomal rust trait declaration with default method implementations. This declaration is split into the native implementation, the wasm implementation which calls the corresponding external function and the implementation of the external function on the host that calls into the native implementation. 

Values between the runtime and the host are not SCALE encoded by default. There are different strategies for passing the arguments between the runtime and the host. The idea behind this is to not waste any performance for encoding the data when it is not required. Data like `[u8]` or `Vec<u8>` is not encoded and it passed directly to the host. This imitates the current behavior as it is done for the storage apis etc. 

For more information, please look into the documentation :)


core/runtime-interface/proc-macro/src/runtime_interface/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use crate::utils::generate_runtime_interface_include;
+
+use proc_macro2::{Span, TokenStream};
+
+use syn::{Ident, ItemTrait, Result};
+
+use inflector::Inflector;
+
```
tomusdrw: What's with the empty lines?

core/runtime-interface/proc-macro/src/runtime_interface/host_function_interface.rs
```diff
@@ -0,0 +1,415 @@
...
+	ItemTrait, TraitItemMethod, Result, ReturnType, Ident, TraitItem, Pat, Error, Signature,
+	spanned::Spanned,
+};
+
+use proc_macro2::{TokenStream, Span};
+
+use quote::{quote, ToTokens};
+
+use inflector::Inflector;
+
```
tomusdrw: ditto

core/sr-io/src/lib.rs
```diff
@@ -46,352 +65,774 @@ pub enum EcdsaVerifyError {
...
+#[runtime_interface]
+pub trait Misc {
+	/// The current relay chain identifier.
+	fn chain_id(&self) -> u64 {
+		externalities::Externalities::chain_id(*self)
+	}
+
+	/// Print a number.
+	fn print_num(val: u64) {
+		log::debug!(target: "runtime", "{}", val);
```
tomusdrw: Shouldn't this be an `info`? I think now by default it's not going to display anything, and previously `print` was a major way to understand what's happening in the runtime.

core/sr-std/with_std.rs
```diff
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
+pub use std::alloc;
```
tomusdrw: I thought we want to avoid the entire `alloc`? We do some whitelist in without-std, no?

gavofyork:
Looks like it needs a polkadot PR?

core/sr-io/src/lib.rs
bkchr: This was done actually here: https://github.com/paritytech/substrate/commit/0eeee4ddeeebf6e1a2edbb987113eb9847db7040

I think that if you need to debug the runtime `RUST_LOG=runtime=debug` is probably better than people always asking what kind of output this is.

core/runtime-interface/proc-macro/src/runtime_interface/mod.rs
bkchr: I do this to separate different dependencies.

core/sr-std/with_std.rs
bkchr: Yeah. On `no_std` we just export some of it.

PR #4058, opened 2019-11-08
Weight annotation for block hooks. 

kianenigma:
Adds the ability to annotate the block hooks (`on_initialize`, `on_finalize`) with weights. The struct passed to the weight_attribute should only implement `WeighData`. The implementation can use the `block_number` as the argument and calculate the weight based on that. 

The executive module, upon initialising a block, will aggregate and sum the weight of all the `on_initialise` and `on_finalize` functions of all the modules (`AllModules`) and writes them directly, without any checks, the the system module. This function is also made public with a very extensive documentation about it being dangerous, for any other runtime code. See `pub fn register_extra_weight_unchecked`. The value passed is actually checked to not exceed the limit, or for overflow. The term unchecked means that no logical check exists to ensure that all the on_initialise and on_finalise functions **will not eat up all the weight in a block**. This is the responsibility of the runtime dev. There is one check added to `node/runtime/src/lib.rs` to always ensure at least substrate-node
 will not consume more than half of the block weight (for the first few thousand blocks at the least). 

- [ ] This needs a PR is polkadot runtime as well. 

Unlike `Call` variants, the default value of a `fn on_initialise()` with no annotation is `zero()`. This might have to change? I don't have a strong opinion but imo it makes more sense to have these hooks to be free by default, unless stated otherwise. 

- A follow up PR should probably sweep substrate and properly annotate all of them.

Needless to say, no one is charged any fee and such for these weights. This is only about the limits. 

#### Some ideas for future: 

- Instead of the test in `node/runtime`, all modules can optionally implement a `PostRuntimeConstruction<AllModules>` that does some arbitrary check to make sure that the module is properly implemented and can integrate with other modules. `AllModules` is probably not enough for all use cases and we cannot make assumptions about which other modules live in `AllModules` but this is slightly more general than what we currently have (currently I think the only way is to do some storage checks on genesis). We can also do some one-time check on storage state/corruption as well in this code. 

- The process of creating a struct and implementing it for the appropriate signature is quite daunting. We should soon allow 

```
#[weight = decl_weight!(|a: u32, b: u64|) { Weight::from(a + b) }]
fn some_dispatch(a: u32, b: u64)
```

which I think would work fine if the closure's signature matches and it still returns the correct type.


gavofyork:
Would be good to allow some kind of dynamic hinting from inside the function itself. Maybe something like:

```
fn on_initialize() {
  if let Some(operations) = BigThingToDo::take() {
    for operation in operations {
      let _weight_of_operation = Self::do_operation(operation);
      #[note_weight(_weight_of_operation)]
    }
}
```

kianenigma:
> Would be good to allow some kind of dynamic hinting from inside the function itself. Maybe something like:
> 
> ```
> fn on_initialize() {
>   if let Some(operations) = BigThingToDo::take() {
>     for operation in operations {
>       let _weight_of_operation = Self::do_operation(operation);
>       #[note_weight(_weight_of_operation)]
>     }
> }
> ```

For such cases, any code within the runtime has the freedom to call `register_extra_weight` (which I want to rename to `register_extra_weight_unchecked` to further note that it should be used with responsibility) https://github.com/paritytech/substrate/pull/4058/files#diff-24bd3cac4cae1c620f82f7aa79ef28ceR609

srml/executive/src/lib.rs
```diff
@@ -652,4 +696,134 @@ mod tests {
...
+
+			Executive::initialize_block(&Header::new_from_number(1));
+			// NOTE: might need updates over time if system and balance introduce new weights. For
+			// now only accounts for the custom module.
+			assert_eq!(<system::Module<Runtime>>::all_extrinsics_weight(), 150 + 25);
+		})
+	}
+}
+
+#[cfg(test)]
```
kianenigma: This is some separate test that I actually ended up integrating into the main `mod tests`.  But at some point it had a compile error that I still want to investigate and lear why it was failing. I will remove this later. 

See the code at the bottom:  when calling `Executive::initialize_block(&Header::new_from_number(1));` it would complain that `Executive` does not have the function because the `UncheckedExtrnisic` type is not `Checkable`, which in my opinion perfectly is.

```
error[E0277]: the trait bound `sr_primitives::generic::unchecked_extrinsic::UncheckedExtrinsic<srml_indices::address::Address<substrate_primitives::sr25519::Public, u32>, tests_weights::Call, substrate_primitives::sr25519::Signature, ()>: sr_primitives::traits::Checkable<srml_system::ChainContext<tests_weights::Runtime>>` is not satisfied
   --> srml/executive/src/lib.rs:769:4
    |
769 |             <UncheckedExtrinsic as Checkable<ChainContext<Runtime>>>::check(a, &Default::default());
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `sr_primitives::traits::Checkable<srml_system::ChainContext<tests_weights::Runtime>>` is not implemented for `sr_primitives::generic::unchecked_extrinsic::UncheckedExtrinsic<srml_indices::address::Address<substrate_primitives::sr25519::Public, u32>, tests_weights::Call, substrate_primitives::sr25519::Signature, ()>`
    |
    = help: the following implementations were found:
              <sr_primitives::generic::unchecked_extrinsic::UncheckedExtrinsic<Address, Call, Signature, Extra> as sr_primitives::traits::Checkable<Lookup>>
```

gavofyork:
ok, but that's a bit problematic in the case that weights are not used for the chain for some reason, no?

kianenigma:
> ok, but that's a bit problematic in the case that weights are not used for the chain for some reason, no?

Not as far as I know, correct me if I am wrong: Weights are a storage in `system`, which is mandatory for any srml runtime. Hence I'd argue that any other module that wants to work with the srml model must deal with weights and provide them and such. Hence I don't see a problem with having this utility function as well a 1st class resident of `srml-system`. 

If you don't want weights, you can disable them by the default to zero or such other _tricks_.

gavofyork:
yeah i guess it's not a big deal.

gavofyork:
looks like you have 3 long lines.

PR #4059, opened 2019-11-08
Typed chain state queries over rpc.

bddap:
This pr adds the ability to query values from chain state in a type-safe way, over rpc.

The decl_storage macro already generates implementations of the srml_support::storage::generator traits (StorageValue, StorageMap, StorageLinkedMap, and StorageDoubleMap). Those traits are currently used to provide type-safe storage access to runtime modules. This pr uses those same traits to provide type-safe storage access over json rpc. The approach keeps users from manually calculating storage keys, a potentially error prone process.

This pr adds minimal functionality, just the ability to get a typed value using a typed key. If the changes are well received, I'd like to look into additional functionality, like *subscribing* to typed keys, subscribing to events, creating and submitting typed extrinsics, etc.

The new workspace member is called substrate-rpc-custom as it allows queries on storage defined custom runtime modules, as opposed to substrate-rpc-api, which is mostly runtime agnostic. The new workspace member depends on both srml-support and on substrate-rpc-api. It was placed in "core/rpc/custom", alongside other rpc utilities, for discoverability.


core/rpc/custom/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "substrate-rpc-custom"
+version = "0.1.0"
+authors = ["Andrew Dirksen <andrew@dirksen.com>"]
```
tomusdrw: Should include ParityTech I suppose?

gnunicorn:
@tomusdrw let's merge and resolve the dependency in the upcoming re-org?!?

tomusdrw:
Superseded by #4079

PR #4061, opened 2019-11-09
grandpa: clear stale periodic block announcements

andresilva:
The GRANDPA voter periodically re-announces blocks it has voted on if it stopped making progress. When the node stops being a validator we would keep re-announcing the last blocks we voted on, since the periodic background announcer would assume that no progress is being made in the GRANDPA protocol. This PR makes it so that when we complete a round we make sure to clear all blocks that were added to the periodic background announcer.


andresilva:
I'm removing the periodic block announcer instead, we are able to explicitly sync forks now so this shouldn't be necessary.

PR #4069, opened 2019-11-09
Pass startup_time to telemetry

tomaka:
cc @maciejhirsz 

Passes the node startup time to the telemetry when connected.
This way, if we disconnect then reconnect, the telemetry knows that it's just the connection that closed and not the node as a whole.

### DIFF
--- a/core/service/src/builder.rs
+++ b/core/service/src/builder.rs
@@ -44,7 +44,10 @@ use sr_primitives::traits::{
 	Block as BlockT, Extrinsic, ProvideRuntimeApi, NumberFor, One, Zero, Header, SaturatedConversion
 };
 use substrate_executor::{NativeExecutor, NativeExecutionDispatch};
-use std::{io::{Read, Write, Seek}, marker::PhantomData, sync::Arc, sync::atomic::AtomicBool};
+use std::{
+	io::{Read, Write, Seek},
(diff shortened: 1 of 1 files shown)


core/service/src/builder.rs
```diff
@@ -1095,6 +1098,9 @@ ServiceBuilder<
 				endpoints,
 				wasm_external_transport: config.telemetry_external_transport.take(),
 			});
+			let startup_time = SystemTime::UNIX_EPOCH.elapsed()
+				.map(|dur| dur.as_millis())
```
bkchr: Isn't seconds enough?

arkpar: This would pass machine start time, right? We want node (process) run time.

arkpar: Telemetry should show node running time regardless of when we connected to telemetry. E.g. if the node was started an hour ago, and it only connected to telemetry 15 minutes ago, it should still show "1 hour". For this to work we need to pass our real startup time I believe.

tomaka: This code is called only once, when we build the service.
`startup_time` is then cloned (well, copy'ed) when we connect to the telemetry.

arkpar: I misread it

tomaka: > Isn't seconds enough?

@maciejhirsz told me milliseconds were better.

PR #4070, opened 2019-11-09
Randomized tests on storage cache

NikVolf:
along with some extra test cases

### DIFF
--- a/core/client/db/src/storage_cache.rs
+++ b/core/client/db/src/storage_cache.rs
@@ -668,7 +668,6 @@ mod tests {
 
 	#[test]
 	fn simple_fork() {
-		//init_log();
 		let root_parent = H256::random();
 		let key = H256::random()[..].to_vec();
 		let h1 = H256::random();
@@ -692,7 +691,6 @@ mod tests {
 
 	#[test]
 	fn double_fork() {
-		//init_log();
 		let root_parent = H256::random();
 		let key = H256::random()[..].to_vec();
 		let h1 = H256::random();
--- a/core/client/db/src/storage_cache.rs
+++ b/core/client/db/src/storage_cache.rs
@@ -27,6 +27,7 @@ use state_machine::{backend::Backend as StateBackend, TrieBackend};
 use log::trace;
 use super::{StorageCollection, ChildStorageCollection};
 use std::hash::Hash as StdHash;
+
 const STATE_CACHE_BLOCKS: usize = 12;
 
 type StorageKey = Vec<u8>;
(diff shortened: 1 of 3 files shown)


NikVolf:
Not sure such tests should be applied on the cache level or on the state storage + cache level though, @arkpar

arkpar:
Unit tests are fine with in-memory backend I suppose. We could add some tests with pre-populated backend and some that simulate backend changes.

gavofyork:
test failures...

core/client/db/src/storage_cache.rs
```diff
@@ -765,3 +833,490 @@ mod tests {
 		assert_eq!(s.storage(&key).unwrap(), None);
 	}
 }
+
+#[cfg(test)]
+mod qc {
+	use std::collections::{HashMap, hash_map::Entry};
+
+    use quickcheck::{quickcheck, TestResult, Arbitrary};
```
kianenigma:
```suggestion
	use quickcheck::{quickcheck, TestResult, Arbitrary};
```

NikVolf:
> Unit tests are fine with in-memory backend I suppose. We could add some tests with pre-populated backend and some that simulate backend changes.

Yep, the same api can be generalised and reused there as well!

NikVolf:
> test failures...

fixed

core/client/db/src/storage_cache.rs
```diff
@@ -654,6 +666,62 @@ mod tests {
 		assert!(s.storage(&key).unwrap().is_none());
 	}
 
+	#[test]
+	fn simple_fork() {
+		//init_log();
```
kianenigma:
```suggestion

```

core/client/db/src/storage_cache.rs
```diff
@@ -654,6 +666,62 @@ mod tests {
...
+		let s = CachingState::new(InMemory::<Blake2Hasher>::default(), shared.clone(), Some(h2a.clone()));
+		assert_eq!(s.storage(&key).unwrap().unwrap(), vec![2]);
+	}
+
+	#[test]
+	fn double_fork() {
+		//init_log();
+		let root_parent = H256::random();
+		let key = H256::random()[..].to_vec();
+		let h1 = H256::random();
```
kianenigma: This is a nit that you can ignore but I would def. consider if I can hide this process of generating `N` random `h256::random()`s somewhere. In the easiest case, if you are fine with using them behind an index (as opposed to h1, h1a etc.) you can do something along the lines of `(0..N).map(H256::random).collect()` or sth similar.

NikVolf: it will be hard to semantically distinguish forks and so on but thanks!

PR #4077, opened 2019-11-10
srml-timestamp: define max timestamp drift in millis

andresilva:
Timestamps in srml-timestamp used to be defined in seconds but were changed to milliseconds later on. When importing a block we allow for a drift in the timestamp but the constant was defined in seconds and wasn't updated. This PR fixes the max drift definition to 30 seconds, defined in millis.

### DIFF
--- a/srml/timestamp/src/lib.rs
+++ b/srml/timestamp/src/lib.rs
@@ -288,7 +288,7 @@ impl<T: Trait> ProvideInherent for Module<T> {
 	}
 
 	fn check_inherent(call: &Self::Call, data: &InherentData) -> result::Result<(), Self::Error> {
-		const MAX_TIMESTAMP_DRIFT: u64 = 30 * 1000;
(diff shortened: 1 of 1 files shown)


srml/timestamp/src/lib.rs
```diff
@@ -288,7 +288,7 @@ impl<T: Trait> ProvideInherent for Module<T> {
 	}
 
 	fn check_inherent(call: &Self::Call, data: &InherentData) -> result::Result<(), Self::Error> {
-		const MAX_TIMESTAMP_DRIFT: u64 = 60;
+		const MAX_TIMESTAMP_DRIFT: u64 = 30 * 1000;
```
kianenigma: I think this refactoring from seconds to millisecons has had one or two of such incidents so far; maybe enough to consider introducing a very basic `enum Time` to make it a _bit harder_ to become inconsistent in the future.

thiolliere: also we could suffix the unit to all constant as MAX_TIMESTAP_DRIFT_SECS

PR #4079, opened 2019-11-10
Typed chain state queries over rpc.

tomusdrw:
Re-opening of #4059

Since we plan to heavily reorganize substrate directory structure soon, to prevent churn and issues during rebase on the updated master I've updated the #4059 to address the comments.
@bddap Hope you don't mind :)


srml/support/rpc/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "srml-support-rpc"
+version = "0.1.0"
```
bkchr:
```suggestion
version = "2.0.0"
```

PR #4083, opened 2019-11-11
Upgrade from futures-preview to futures 0.3.1, and remove futures 0.1 where currently possible

expenses:
As #3099 mentions, we want to switch to stable futures. While we are still blocked in several places (tokio 0.2, jsonrpc, libp2p etc), we can switch from the preview branch of futures to the proper `0.3.1`, and get away from futures `0.1` in some places.

Note that in order to stabilise [`futures-timer`](https://crates.io/crates/futures-timer), the `Interval` stream was moved into [`async-std`](https://crates.io/crates/async-std) under the `unstable` feature (as the stream api hasn't been stabilised yet). This isn't great, as it adds more dependencies than we really need. Perhaps it would be best to move `Interval` into our own lib.


tomaka:
> This isn't great, as it adds more dependencies than we really need.

I'm not so worried about adding dependencies, but more about the fact that it's behind the `unstable` flag, meaning that it can break at any time.

core/rpc/Cargo.toml
```diff
@@ -8,7 +8,8 @@ edition = "2018"
 api = { package = "substrate-rpc-api", path = "./api" }
 client = { package = "substrate-client", path = "../client" }
 codec = { package = "parity-scale-codec", version = "1.0.0" }
-futures03 = { package = "futures-preview", version = "0.3.0-alpha.19", features = ["compat"] }
+futures = { version = "0.3.1", features = ["compat"] }
+futures01 = { package = "futures", version = "0.1.29" }
```
tomaka: That was a `dev-dependency` before, is there a reason for moving it to an actual dependency?

core/cli/src/lib.rs
```diff
@@ -102,7 +101,7 @@ pub struct VersionInfo {
 /// Something that can be converted into an exit signal.
 pub trait IntoExit {
 	/// Exit signal type.
-	type Exit: Future<Item=(),Error=()> + Send + 'static;
+	type Exit: Future<Output=()> + Unpin + Send + 'static;
```
tomaka: As a heads up, this breaks Polkadot.

core/rpc/Cargo.toml
expenses: Nope, fixed!

core/cli/src/lib.rs
expenses: Great, I'll make a PR for that then.

tomaka:
You might have to replace all the `cargo web build -p ...` in `.gitlab-ci.yml` with `cargo build -p ... --target=wasm32-unknown-unknown` in order to make it work :grimacing:

tomaka:
`async-std` unfortunately doesn't compile for `wasm32-unknown-unknown` :grimacing: 
cc https://github.com/async-rs/async-std/issues/220

expenses:
@tomaka ah ok. What should I do here?

core/network/src/debug_info.rs
```diff
@@ -76,7 +76,7 @@ impl<TSubstream> DebugInfoBehaviour<TSubstream> {
 			ping: Ping::new(PingConfig::new()),
 			identify,
 			nodes_info: FnvHashMap::default(),
-			garbage_collect: Box::new(Interval::new(GARBAGE_COLLECT_INTERVAL).map(|()| Ok(())).compat()),
+			garbage_collect: Box::new(interval(GARBAGE_COLLECT_INTERVAL).map(|()| Ok(())).compat()),
```
tomaka: That should work, I think (with better formatting):

```suggestion
			garbage_collect: Box::new(futures::stream::unfold((), |()| async { Delay::new(GARBAGE_COLLECT_INTERVAL).await; ((), ()) }).map(|()| Ok(())).compat()),
```

core/network/src/protocol.rs
```diff
@@ -430,8 +431,8 @@ impl<B: BlockT, S: NetworkSpecialization<B>, H: ExHashT> Protocol<B, S, H> {
 		let behaviour = LegacyProto::new(protocol_id, versions, peerset);
 
 		let protocol = Protocol {
-			tick_timeout: Box::new(futures_timer::Interval::new(TICK_TIMEOUT).map(|v| Ok::<_, ()>(v)).compat()),
-			propagate_timeout: Box::new(futures_timer::Interval::new(PROPAGATE_TIMEOUT).map(|v| Ok::<_, ()>(v)).compat()),
+			tick_timeout: Box::new(interval(TICK_TIMEOUT).map(|v| Ok::<_, ()>(v)).compat()),
```
tomaka: Same idea.

gavofyork:
@tomaka is the webwasm stuff something that can/should be fixed in this PR?

tomaka:
The WASM CI is still failing because there's still a dependency on `async-std` which can be removed by using `stream::unfold` rather than `async_std::Interval` (as explained above).

gavofyork:
ok - @expenses you ok doing that?

expenses:
Yep!

client/authority-discovery/src/lib.rs
```diff
@@ -68,6 +67,8 @@ use prost::Message;
 use sr_primitives::generic::BlockId;
 use sr_primitives::traits::{Block as BlockT, ProvideRuntimeApi};
 
+type Interval = Box<dyn Stream<Item = ()> + Unpin>;
```
expenses: I think this is preferable to `future::Unfold<(), ...>`, but not ideal

expenses:
@tomaka this should be ready now

PR #4089, opened 2019-11-11
Do nothing for zero imbalances

### LINKED ISSUE #4088: OnUnbalanced triggers even when zero
In a recent runtime upgrade (Kusama) the `treasury.Deposit`-event was added.
This event is triggered quite frequently: https://polkascan.io/pre/kusama-cc2/event?module=treasury&event=Deposit&page=1

Most instances of the event have a zero value; Examples:
- https://polkascan.io/pre/kusama-cc2/event/582836-3
- https://polkascan.io/pre/kusama-cc2/event/582808-3

It looks like only the events that are triggered at the end of a `session` have a non-zero value; Examples: 
1. End of session: 257
- https://polkascan.io/pre/kusama-cc2/event/582742-4 (~0.18 KSM)
- https://polkascan.io/pre/kusama-cc2/event/582742-6 (~260.882 KSM)
2. End of session: 256
- https://polkascan.io/pre/kusama-cc2/event/580733-3 (~32.103 KSM)
- https://polkascan.io/pre/kusama-cc2/event/580733-5 (~260.838 KSM)

I have the following questions/remarks:
1. Why are some of those events triggered mid-session?
2. Why do those mid-session events have a zero value?
3. Shouldn't these mid-session events simply be omitted (bloat/no added value)?
4. Why are there 2 such events at the end of the session?
5. If these 2 events have distinct meaning, shouldn't it be two distinct event types?
6. If they do not have a distinct meaning, why not combine these two events into one event with a combined value?

gavofyork:
Closes #4088 
Introduce a new `TryDrop` trait to get around the fact that `Imbalance` trait is itself generic. Then always `try_drop` before actually acting on the imbalance. This means that zero imbalances will be silently dropped without being explicitly handled, as they should be.

### DIFF
--- a/node/runtime/src/impls.rs
+++ b/node/runtime/src/impls.rs
@@ -25,7 +25,7 @@ use crate::{Balances, System, Authorship, MaximumBlockWeight, NegativeImbalance}
 
 pub struct Author;
 impl OnUnbalanced<NegativeImbalance> for Author {
-	fn on_unbalanced(amount: NegativeImbalance) {
+	fn on_nonzero_unbalanced(amount: NegativeImbalance) {
 		Balances::resolve_creating(&Authorship::author(), amount);
 	}
 }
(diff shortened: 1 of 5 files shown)


gavofyork:
@kianenigma *bump*

PR #4090, opened 2019-11-11
EVM should store contracts' balances in an account.

gavofyork:
EVM currently remembers its contracts' balances outside of the issuance system. This leads to `TotalIssuance` being less than it should be as it doesn't count any account balances of EVM contracts. This fixes that in the same way it's handled for treasury module.

### DIFF
--- a/srml/evm/src/lib.rs
+++ b/srml/evm/src/lib.rs
@@ -27,12 +27,15 @@ use rstd::vec::Vec;
 use support::{dispatch::Result, decl_module, decl_storage, decl_event};
 use support::traits::{Currency, WithdrawReason, ExistenceRequirement};
 use system::ensure_signed;
+use sr_primitives::ModuleId;
 use sr_primitives::weights::SimpleDispatchInfo;
-use sr_primitives::traits::UniqueSaturatedInto;
+use sr_primitives::traits::{UniqueSaturatedInto, AccountIdConversion};
 use primitives::{U256, H256, H160};
 use evm::{ExitReason, ExitSucceed, ExitError};
 use evm::executor::StackExecutor;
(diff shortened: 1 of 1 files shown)


srml/evm/src/lib.rs
```diff
@@ -117,12 +120,13 @@ decl_module! {
 		fn deposit_balance(origin, value: BalanceOf<T>) -> Result {
 			let sender = ensure_signed(origin)?;
 
-			T::Currency::withdraw(
+			let imbalance = T::Currency::withdraw(
 				&sender,
 				value,
 				WithdrawReason::Reserve.into(),
-				ExistenceRequirement::KeepAlive,
+				ExistenceRequirement::AllowDeath,
```
gavofyork: No reason to keep the sender alive.

srml/evm/src/lib.rs
```diff
@@ -27,13 +27,16 @@ use rstd::vec::Vec;
...
+use sr_primitives::ModuleId;
 use sr_primitives::weights::SimpleDispatchInfo;
-use sr_primitives::traits::UniqueSaturatedInto;
+use sr_primitives::traits::{UniqueSaturatedInto, AccountIdConversion};
 use primitives::{U256, H256, H160};
 use evm::{ExitReason, ExitSucceed, ExitError};
 use evm::executor::StackExecutor;
 use evm::backend::ApplyBackend;
 
+const MODULE_ID: ModuleId = ModuleId(*b"py/ethvm");
```
sorpaas: What does the “py/“ prefix mean in the module id?

gavofyork: short for "parity"

PR #4091, opened 2019-11-11
Fix sync stalling on moving head (best block) prior to the tip of the chain sometimes

arkpar:
* Work around the reorg-to-not-the-best-block by not restarting sync in such cases.
* Small optimization to avoid copying huge `peers` structure needlessly.

### DIFF
--- a/core/client/src/client.rs
+++ b/core/client/src/client.rs
@@ -1923,7 +1923,7 @@ pub(crate) mod tests {
 	use super::*;
 	use primitives::blake2_256;
 	use sr_primitives::DigestItem;
-	use consensus::{BlockOrigin, SelectChain};
+	use consensus::{BlockOrigin, SelectChain, BlockImport};
 	use test_client::{
 		prelude::*,
 		client_db::{Backend, DatabaseSettings, DatabaseSettingsSrc, PruningMode},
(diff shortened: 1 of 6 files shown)


core/client/src/client.rs
```diff
@@ -1587,16 +1587,8 @@ impl<'a, B, E, Block, RA> consensus::BlockImport<Block> for &'a Client<B, E, Blo
...
-		{
-			BlockStatus::InChainWithState | BlockStatus::Queued => {},
-			BlockStatus::Unknown => return Ok(ImportResult::UnknownParent),
-			BlockStatus::InChainPruned if allow_missing_state => {},
-			BlockStatus::InChainPruned => return Ok(ImportResult::MissingState),
-			BlockStatus::KnownBad => return Ok(ImportResult::KnownBad),
-		}
-
+		// Own status must be checked first. If the block and ancestry is pruned
+		// this function must return `AlreadyInChain` rather than `MissingState`
```
gavofyork: this comment seems incongruous to the code that follows it:

```
BlockStatus::InChainPruned => {}
```

so it doesn't return `AlreadyInChain` ...

arkpar: good catch

gavofyork:
maybe a test would be an idea?

Issue #3780, opened 2019-10-07
Applying WASM snapshot fails occasionally


raykyri:
Our nodes occasionally panic when the executor's WASM snapshot gets out of sync. This happens more frequently on different nodes than others (e.g. nodes that have been running for longer / were started earlier). 

We're using this version of Substrate: https://github.com/paritytech/substrate/commit/e0d486d617cbe24f0c91cd6aa0dfcfb6c4e4dcc2

```
Oct 07 00:35:25 edg09-1 run_node.sh[20824]: ====================
Oct 07 00:35:25 edg09-1 run_node.sh[20824]: Version: 1.0.0-5a52390-x86_64-linux-gnu
Oct 07 00:35:25 edg09-1 run_node.sh[20824]: stack backtrace:
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:    0: substrate_panic_handler::set::{{closure}}
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:    1: std::panicking::rust_panic_with_hook
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:              at src/libstd/panicking.rs:481
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:    2: std::panicking::continue_panic_fmt
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:              at src/libstd/panicking.rs:384
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:    3: rust_begin_unwind
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:              at src/libstd/panicking.rs:311
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:    4: core::panicking::panic_fmt
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:              at src/libcore/panicking.rs:85
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:    5: core::result::unwrap_failed
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:              at src/libcore/result.rs:1084
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:    6: std::thread::local::LocalKey<T>::with
Oct 07 00:35:25 edg09-1 run_node.sh[20824]:    7: substrate_state_machine::StateMachine<B,H,N,T,O,Exec>::execute_aux
[...]

PR #4097, opened 2019-11-13
Additional wasm diagnostics

pepyakin:
Related #3780 

This PR makes the erasing procedure in wasmi runtimes failable. If the erasing fails it will be returned as an error instead of panicking on expect.

### DIFF
--- a/core/executor/src/wasmi_execution.rs
+++ b/core/executor/src/wasmi_execution.rs
@@ -28,7 +28,7 @@ use crate::sandbox;
 use crate::allocator;
 use crate::wasm_utils::interpret_runtime_api_result;
 use crate::wasm_runtime::WasmRuntime;
-use log::trace;
+use log::{error, trace};
 use parity_wasm::elements::{deserialize_buffer, DataSegment, Instruction, Module as RawModule};
 use wasm_interface::{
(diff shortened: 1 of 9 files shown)


core/executor/src/wasmi_execution.rs
```diff
@@ -568,9 +551,8 @@ impl WasmRuntime for WasmiRuntime {
 		method: &str,
 		data: &[u8],
 	) -> Result<Vec<u8>, Error> {
-		self.with(|module| {
-			call_in_wasm_module(ext, module, method, data, &self.host_functions)
-		})
+		self.state_snapshot.apply(&self.instance)?;
```
bkchr: This will fail the execution. Why not reset the instance?

pepyakin: What do you mean by reset here?

If you mean call `invalidate_runtime`, then I'd expect it to happen as part of handling `Err`.

bkchr: That does not invalidates the runtime. I mean reset `instance` and `module` of your current runtime and retry the execution.

The problem when a call fails, it could invalidate a transaction when this happens while building a block or when importing one.

PR #4105, opened 2019-11-14
Bump dependencies again

DemiMarie-parity:
We no longer depend on old versions of `syn` et al, which should reduce
our compile-times substantially.


node/cli/Cargo.toml
```diff
@@ -23,14 +23,14 @@ crate-type = ["cdylib", "rlib"]
...
-codec = { package = "parity-scale-codec", version = "1.0.6" }
+codec = { package = "parity-scale-codec", version = "1.1.0" }
 serde = { version = "1.0.102", features = [ "derive" ] }
 futures = "0.1.29"
 hex-literal = "0.2.1"
 jsonrpc-core = "14.0.3"
 log = "0.4.8"
-rand = "0.7.2"
-structopt = "0.3.3"
+rand = "0.6.5"
```
tomaka: Why the downgrade?

DemiMarie-parity: `cargo upgrade` bug

DemiMarie-parity:
@tomaka Polkadot won’t build unless I replace `BitVec` with something else.

Issue #3803, opened 2019-10-11
contracts: implement `ext_runtime_get_storage`


ascjones:
In order to allow smart contracts to query runtime storage, a new `ext_` function is required:
```
fn ext_runtime_get_storage(key_ptr: u32, key_len: u32) -> u32
```
I've implemented a quick prototype: https://github.com/paritytech/substrate/compare/aj-contracts-get-storage.

However this is [hard coded](https://github.com/paritytech/substrate/compare/aj-contracts-get-storage#diff-96eada255f897b2fc434cb93b380661fR706) to hash the key `twox_128(key)`

Assuming we don't expect the contract itself to perform the key hashing, the remaining problem to solve is the API for correct hashing of keys for different storage types:

- Value `twox_128(key)`
- Map `blake2_256(prefix + key)`
- Map2 `twox_128(prefix + key)`
- LinkedMap `blake2_256(prefix + key)`
- LinkedMap2 `twox_128(prefix + key)`
- DoubleMap `blake2_256(prefix + key) + blake2_256(key2)`
- DoubleMap2 `twox_128(prefix + key) + twox_128(key2)`

Rel: https://github.com/paritytech/ink/pull/204

xlc:
Related previous discussion: #2214

xlc:
Access the storage directly will make it very hard / impossible to upgrade runtime module storage format because it means all the contracts have to upgrade as well.

PR #4108, opened 2019-11-14
srml-contracts: Add ability to read substrate's storage

pepyakin:
Fixes https://github.com/paritytech/substrate/issues/3803

Essentially, this PR adds a function:

```rust
// Retrieve the value at the under the given key from the **runtime** storage and return 0.
// If there is no entry under the given key then this function will return 1 and
// clear the scratch buffer.
//
// - key_ptr: the pointer into the linear memory where the requested value is placed.
// - key_len: the length of the key in bytes.
fn ext_get_runtime_storage(ctx, key_ptr: u32, key_len: u32) -> u32
```


srml/contracts/src/wasm/runtime.rs
```diff
@@ -835,6 +835,29 @@ define_env!(Env, <E: Ext>,
 		ctx.ext.block_number().encode_to(&mut ctx.scratch_buf);
 		Ok(())
 	},
+
+	// Retrieve the value at the under the given key from the **runtime** storage and return 0.
+	// If there is no entry under the given key then this function will return 1 and
+	// clear the scratch buffer.
+	//
+	// - key_ptr: the pointer into the linear memory where the requested value is placed.
+	// - key_len: the length of the key in bytes.
```
Robbepop: The first sentence doesn't make sense grammatically.

Robbepop: Also I would suggest the following docs:

```
/// Retrieve the value at the given key from the **runtime** storage and return 0.
///
/// # Error
///
/// - If there is no entry at the given key return 1.
/// - A runtime might restrict access to a certain subset of keys. If the supplied
///   key is not part of that restricted whitelist of keys; return 1.
///
/// # Parameters
///
/// - `key_ptr`: The pointer into the contiguous memory where the requested value is placed.
/// - `key_len`: The length of the key in number of bytes.
```

Robbepop: These docs include your idea with the whitelisted keys. I think we should really implement this as it is a really useful safety feature. We might be inclined to return a different value from `1` though in this case.

Robbepop: I generally would also like an `# Errors` and `# Parameters` section for all the `ext_` function docs because it looks clearer and also is more similar to what normal Rust docs look like.

srml/contracts/src/wasm/mod.rs
```diff
@@ -1634,4 +1662,103 @@ mod tests {
...
+		// "\01\02\03\04" - Some(0x14144020)
+		// "\02\03\04\05" - None
+		*mock_ext.runtime_storage_keys.borrow_mut() = [
+			([1, 2, 3, 4].to_vec(), Some(0x14144020u32.to_le_bytes().to_vec())),
+			([2, 3, 4, 5].to_vec().to_vec(), None),
+		]
+		.iter()
+		.cloned()
+		.collect();
+		let _ = execute(
```
Robbepop: Why `let` here required if there is also an `unwrap`?

srml/contracts/src/wasm/mod.rs
```diff
@@ -1634,4 +1662,103 @@ mod tests {
 		assert_eq!(output, ExecReturnValue { status: 17, data: hex!("5566778899").to_vec() });
 		assert!(!output.is_success());
 	}
+
+	const CODE_GET_RUNTIME_STORAGE: &str = r#"
```
Robbepop: Have you thought about using `include_bytes!` instead of embedding all those raw strings into the actual code?

pepyakin: Yeah, it's one of the options. One another is to move those declarations in a separate `.rs` file. It is useful because some wasm test fixtures are shared between integration tests and unit tests. We can even combine these two approaches. 

However, I think this is better done as a follow up

srml/contracts/src/wasm/mod.rs
pepyakin: `let` here is not required. Note all other usages of this pattern - it is deliberate stylistic choise to note that there is a result. Don't feel strongly about this one though, and can be removed. Although also, not within this PR.

Robbepop: It confused me so maybe it confuses others as well. Agree that we shouldn't do that in this PR if it is a pattern but you might want to think about removing it because it might confuse others.

srml/contracts/src/wasm/runtime.rs
pepyakin: ah, I meant to remove `at the` part, thanks.

as per whitelist let's discuss it on the meeting tmrw.

wrt doc sections: yeah sounds good. As with other feedback items that do not relate to this particular PR, I'd move them as a follow up. (Otherwise we end up with mixed styles, which I consider to not be not worth it even though it would improve the situation of one particular item)

gavofyork: > really useful safety feature

How, exactly? All storage is necessarily public and open. A proof can always be given about the last block's state, or attested to in the current block by the block author. So *at best* you're delaying the ability for a contract to "read" storage by one block assuming that the block author doesn't want to facilitate things.

Robbepop: Hm ... so the thoughts were as followed: A contract could read storage from a runtime entry that is arbitrarily large and thus requires an arbitrarily amount of time (resources) to query as far as I understood. So the idea by @pepyakin and me was that whitelisting keys (entries) by the runtime would fix this problem.

pepyakin: @gavofyork 

To give more context: for substrate contracts we limit the maximal size of stored values by some limit (~16Kb). Up to this number, the size of the value for storage loads don't influence the run time significantly and the idea that we would apply flat rate for the loads.

However, if you allow loading around the whole storage, this assumption might not hold and then the value size aspect kicks in. This is really annoying since we don't have a way to find out the size of the storage value to compute and charge gas for that. This can be abused if you have contract calls that devote a bit of gas and then make a big load (e.g. `:code`).

srml/contracts/src/wasm/runtime.rs
```diff
@@ -835,6 +835,29 @@ define_env!(Env, <E: Ext>,
...
+		let key_buf = mem::replace(&mut ctx.scratch_buf, Vec::new());
+
+		match ctx.ext.get_runtime_storage(&key_buf) {
+			Some(value_buf) => {
+				// The given value exists.
+				ctx.scratch_buf = value_buf;
+				Ok(0)
+			}
+			None => {
+				ctx.scratch_buf = Vec::new();
```
jimpo: Maybe assign back to `key_buf` and `clear` it to avoid reallocating memory?

srml/contracts/src/wasm/runtime.rs
jimpo: Hmm, yeah, this does seem like a problem. Another issue is that even if the value you are loading isn't too big, one of it's ancestor nodes in the trie could have a huge attached value (eg. doing a lookup for `:code:nothing-here`). As long as storage lookups require a trie traversal, this would also incur a large performance penalty. I don't see any options to `Get` in RocksDB that allows the caller to limit the size of the value.

pepyakin:
Yeah, it is somewhat expected. We are pushing assigning the gas schedule to be one of the latest things to do, because it is very sensitive for the implementation details.

pepyakin:
The failures don't look related to this PR.

PR #4109, opened 2019-11-14
Add serde for Signature types

fubuloubu:
### Description
This PR adds support for `serde` serialization and deserialization to the Signature types of all cryptosystems that Substrate uses. It also adds support to the `MultiSIgnature` enum class through automatic derivation.

Please note that hex-string encoding was chosen for Signature classes for simplicity.

### DIFF
--- a/primitives/core/src/ecdsa.rs
+++ b/primitives/core/src/ecdsa.rs
@@ -528,6 +528,7 @@ mod test {
 	use super::*;
 	use hex_literal::hex;
 	use crate::crypto::DEV_PHRASE;
+	use serde_json;
 
 	#[test]
 	fn default_phrase_should_be_used() {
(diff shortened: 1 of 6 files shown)


fubuloubu:
Note sure I understand the CI failure with the dependency rules

burdges:
Just fyi, there are many crypto crates that initially got their serialization size wrong, due to serializing as a slice, not as a fixed size array.

fubuloubu:
@burdges I copied the implemention from `Public`

primitives/core/src/ecdsa.rs
```diff
@@ -190,6 +190,21 @@ impl rstd::convert::TryFrom<&[u8]> for Signature {
...
+impl Serialize for Signature {
+	fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error> where S: Serializer {
+		serializer.serialize_str(&hex::encode(self))
+	}
+}
+
+#[cfg(feature = "std")]
+impl<'de> Deserialize<'de> for Signature {
+	fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> where D: Deserializer<'de> {
+		Ok(Signature::from_slice(&hex::decode(&String::deserialize(deserializer)?)
```
tomusdrw: That will crash the node if we ever attempt to deserialize a wrongly sized signature. Please use `try_from` instead and return an error instead of panicking.

fubuloubu:
@tomusdrw is there a special location for serialization tests? I don't see a similar one for `Pair`

primitives/core/src/ecdsa.rs
fubuloubu: Resolved in 4e2c7f0

fubuloubu: Resolved in 4e2c7f0

fubuloubu:
Implemented all suggested changes

burdges:
I have not looked at this code much yet, sorry. 

I only meant that several signature crates had a bug where they serialized with a 4 or 8 byte length.  It's a bug because they know exactly the length, so this makes it hard to correspond to existing formats.  I think you're not delegating serde to any signature crates here, so no problem if any dependencies fix their own serde though.  

Implicitly, I'm also warning against making the same mistake, and instead represent the signature type, except..  It's not so problematic if it stays consistent and substrate deals with this sort of migration lots lately.  :)

fubuloubu:
Rebased `master`

fubuloubu:
Like to note that test failures are unrelated:
```
Running `/builds/parity/substrate/target/release/deps/substrate_primitives-8ba8cc773cf96d8d`

running 62 tests
test ...
test ecdsa::test::signature_serialization_doesnt_panic ... ok
test ecdsa::test::signature_serialization_works ... ok
test ed25519::test::signature_serialization_doesnt_panic ... ok
test ed25519::test::signature_serialization_works ... ok
test sr25519::test::signature_serialization_doesnt_panic ... ok
test sr25519::test::signature_serialization_works ... ok
test ...

test result: ok. 62 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

PR #4111, opened 2019-11-15
Fixes `paint-authority-discovery` build and remove warnings

bkchr:


paint/authority-discovery/src/lib.rs
```diff
@@ -198,12 +198,6 @@ mod tests {
 			.map(AuthorityId::from)
 			.collect();
 
-		// Needed for `session::OneSessionHandler::on_new_session`.
```
mxinden: :+1:

PR #4112, opened 2019-11-15
Finality proof: fetch new authorities set instead of previous one

svyatonik:
Finality proof includes the proof(previous-authorities-set) instead of proof(new-authorities-set). This PR fixes that. Thanks to @jimpo for the report


svyatonik:
(I think one test should be updated to test this properly => linprogress)

client/finality-grandpa/src/finality_proof.rs
```diff
@@ -314,11 +314,10 @@ pub(crate) fn prove_finality<Block: BlockT<Hash=H256>, B: BlockchainBackend<Bloc
 
 		if let Some(justification) = blockchain.justification(current_id)? {
 			// check if the current block enacts new GRANDPA authorities set
-			let parent_id = BlockId::Number(current_number - One::one());
-			let new_authorities = authorities_provider.authorities(&parent_id)?;
+			let new_authorities = authorities_provider.authorities(&current_id)?;
 			let new_authorities_proof = if current_authorities != new_authorities {
 				current_authorities = new_authorities;
```
jimpo: Shouldn't `authorities_set_id` also get incremented?

client/finality-grandpa/src/finality_proof.rs
```diff
@@ -338,10 +337,12 @@ pub(crate) fn prove_finality<Block: BlockT<Hash=H256>, B: BlockchainBackend<Bloc
 			if justifies_end_block || justifies_authority_set_change {
 				// check if the proof is generated by the requested authority set
 				if finality_proof.is_empty() {
```
jimpo: Why does it only check that the justification is correct if there are no other proof fragments?

client/finality-grandpa/src/finality_proof.rs
```diff
@@ -338,10 +337,12 @@ pub(crate) fn prove_finality<Block: BlockT<Hash=H256>, B: BlockchainBackend<Bloc
...
+					let parent_id = BlockId::Number(current_number - One::one());	
+					let prev_authorities = authorities_provider.authorities(&parent_id)?;
```
jimpo: Couldn't you just keep a reference to `current_authorities` above before reassigning instead of doing another query to the `authorities_provider`? Or just reassign `current_authorities` (and perhaps `authorities_set_id`) below this?

client/finality-grandpa/src/finality_proof.rs
```diff
@@ -505,12 +506,13 @@ fn check_finality_proof_fragment<Block: BlockT<Hash=H256>, B, J>(
 	// and now verify new authorities proof (if provided)
 	if let Some(new_authorities_proof) = proof_fragment.authorities_proof {
 		// it is safe to query header here, because its non-finality proves that it can't be pruned
-		let header = blockchain.expect_header(BlockId::Hash(proof_fragment.block))?;
-		let parent_hash = *header.parent_hash();
-		let parent_header = blockchain.expect_header(BlockId::Hash(parent_hash))?;
+		let header = match proof_fragment.unknown_headers.iter().rev().next().cloned() {
```
jimpo: Please leave a comment explaining this.

client/finality-grandpa/src/finality_proof.rs
```diff
@@ -609,18 +611,22 @@ pub(crate) mod tests {
...
-			proof: StorageProof
+			proof: StorageProof,
 		) -> ClientResult<AuthorityList> {
 			self.0(hash, header, proof)
 		}
 	}
 
 	#[derive(Debug, PartialEq, Encode, Decode)]
-	pub struct TestJustification(pub bool, pub Vec<u8>);
+	pub struct TestJustification(pub bool, pub Option<(u64, Vec<(AuthorityId, u64)>)>, pub Vec<u8>);
```
jimpo: Use `AuthorityList` instead of `Vec<(AuthorityId, u64)>`.

jimpo: Is this just a `Some` because you don't want to go through all the tests and give a value? Any reason not to make it required?

client/finality-grandpa/src/finality_proof.rs
svyatonik: Nope - `authorities_set_id` are passed by the requester (who wants finality proof) and are only used to check that we have the same set (to fail if there have been forced changes).

client/finality-grandpa/src/finality_proof.rs
svyatonik: Because, as I said above, we only check that the **initial set** is the same on both requester and prover

client/finality-grandpa/src/finality_proof.rs
svyatonik: Yes - that's the reason.

jimpo: I feel like there's not so many uses that it would be difficult to add the authority set and that's a good additional check to have in place anyway. And you could drop the first `bool` arg since `false` is the same as providing an incorrect set `id`.

PR #4117, opened 2019-11-17
Bump dependencies, respecting semver

DemiMarie-parity:
We really need a bot to make these PRs, but in the meantime, I can do so with almost no effort.
✄ -----------------------------------------------------------------------------


client/service/test/Cargo.toml
```diff
@@ -9,9 +9,9 @@ tempdir = "0.3.7"
 tokio = "0.1.22"
 futures = "0.1.29"
 log = "0.4.8"
-env_logger = "0.7.0"
+env_logger = "0.7.1"
 fdlimit = "0.1.1"
-futures03 = { package = "futures-preview", version = "=0.3.0-alpha.19", features = ["compat"] }
+futures03 = { package = "futures-preview", version = "0.3.0-alpha.19", features = ["compat"] }
```
thiolliere: I don't know how alpha works, if they release a new alpha is it correctly considered a breaking change ?

DemiMarie-parity: Not sure.  In any case, we should switch to `futures` now that futures 0.3 is out.

gavofyork:
closing since it's now stale. this really needs to be done and merged immediately upon CI pass.

PR #4118, opened 2019-11-17
Document that no balance can be strictly between 0 and EXISTENTIAL_DEPOSIT

DemiMarie-parity:
✄ -----------------------------------------------------------------------------

### DIFF
--- a/paint/balances/src/lib.rs
+++ b/paint/balances/src/lib.rs
@@ -42,13 +42,25 @@
 //! - **Total Issuance:** The total number of units in existence in a system.
 //! - **Reaping an account:** The act of removing an account by resetting its nonce. Happens after its balance is set
 //! to zero.
-//! - **Free Balance:** The portion of a balance that is not reserved. The free balance is the only balance that matters
-//! for most operations. When this balance falls below the existential deposit, most functionality of the account is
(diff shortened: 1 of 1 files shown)


kianenigma:
@joepetrowski I think I recommended to you that we could have a genesis level check [here](https://github.com/paritytech/substrate/blob/a04b8bdd0c043297ebe31b3183d663b55baa6e6b/paint/balances/src/lib.rs#L371) to make sure a faulty chain spec cannot defy this rule? Maybe you could add it in this PR @DemiMarie-parity ?

emielvanderhoek:
Kusama has 'only' 380 accounts with a non-zero balance, but there are more than 1210 accounts with an AccountIndex in Kusama genesis. Does this issue address those 830 accounts?

joepetrowski:
> Kusama has 'only' 380 accounts with a non-zero balance, but there are more than 1210 accounts with an AccountIndex in Kusama genesis. Does this issue address those 830 accounts?

No, having an index won't affect that.  Those accounts will have their respective index until they are reaped.  

The logic in https://github.com/paritytech/substrate/pull/4151/files will just panic if it finds addresses with less than ED but it won't reap them. I am wondering if that check will panic on these accounts (registered with an index but no balance). @kianenigma

PR #4120, opened 2019-11-18
add hashed_key_for() getter method for StorageDoubleMap

elim0322:
I noticed that there isn't `hashed_key_for()` getter method for `StorageDoubleMap` whereas one exists in `StorageMap`. This PR adds the method.


elim0322:
[clabot:check]

PR #4123, opened 2019-11-18
elections-phragmen: renounce candidacy [+ remove dead code]

kianenigma:
gist of it, and the only notable TODO: 

```
/// renounce one's intention to be a candidate for the next election round. 3 potential
/// outcomes exist:
/// - `origin` is a candidate and not elected in any set. In this case, the bond is
///   unreserved, returned and origin is removed as a candidate.
/// - `origin` is a current runner up. In this case, the bond is taken and origin is removed
///   as a runner, consequently not being a candidate for the next round anymore. Indeed, a
///   runner up should generally avoid calling this function unless if they want a quick
///   exit out of the system. Otherwise, staying a runner and hoping to be elected down the
///   road makes more sense.
/// - `origin` is a current member. In this case, the bond is unreserved and origin is
///   removed as a member, consequently not being a candidate for the next round anymore.
///   Similar to [`remove_voter`], if replacement runners exists, they are immediately used.
///   Otherwise, a new election is started immediately.
// TODO: maybe don't re-run phragmen since it is expensive.
```

alternatively, we can: 

- not run the election if a member leaves, and more importantly maybe don't return the bond if they are already elected.
- Or, even if we want to run the election, we should put a temp storage there to make sure we only run phragmen at most once per block

---

also cleans the migration code.


kianenigma:
(needs a `/denounce/renounce/` in the entire PR. My english mistake)

AlistairStewart:
Yes, should be renounce. We might end up rerunning Phragmén multiple times in a block in the corner case that several council members withdraw simultaneously and we don't have enough candidates to fill the council, which could lead to a very long block time.

paint/elections-phragmen/src/lib.rs
```diff
@@ -315,14 +312,81 @@ decl_module! {
...
+		/// - `origin` is a current member. In this case, the bond is unreserved and origin is
+		///   removed as a member, consequently not being a candidate for the next round anymore.
+		///   Similar to [`remove_voter`], if replacement runners exists, they are immediately used.
+		///   Otherwise, a new election is started immediately.
+		// TODO: maybe don't re-run phragmen since it is expensive.
+		#[weight = SimpleDispatchInfo::FixedOperational(2_000_000)]
+		fn denounce_candidacy(origin) {
+			let who = ensure_signed(origin)?;
+
+			// NOTE: this function attempts the 3 arms and fails if none are matched. Unlike other
```
kianenigma: I am not really set on this and can change in case of any objections. The code would admittedly be more readable if we read all 3 sets first and do the checks.

paint/elections-phragmen/src/lib.rs
```diff
@@ -315,14 +312,81 @@ decl_module! {
...
+			}
+
+			let mut runners_with_stake = Self::runners_up();
+			if let Ok(index) = runners_with_stake
+				.binary_search_by(|(ref r, ref _s)| r.cmp(&who))
+			{
+				runners_with_stake.remove(index);
+				// slash bond
+				let (imbalance, _) = T::Currency::slash_reserved(&who, T::CandidacyBond::get());
+				T::KickedMember::on_unbalanced(imbalance);
```
kianenigma:
```suggestion
				T::LoserCandidate::on_unbalanced(imbalance);
```

gavofyork:
Ok - remove candidates, re-filling from runners up. Never re-run Phragmén.

> `origin` is a current runner up. In this case, the bond is taken and origin is removed as a runner, 

Bond should not be taken - can be returned as normal.

kianenigma:
> Bond should not be taken - can be returned as normal.

The reason why it is the way it is: Currently this rule applies: you will lose your bond unless if you are a winner. Both losers and runners who are not elevated, will lose their bond once outgoing. Maybe I should change this as well? if someone is a runner, and then no longer a runner (out-voted), should they be given their bond back?

kianenigma:
Okay -- all runners are now immune to losing their bond.

gavofyork:
You should lose your bond only if you outright lose in an election.

You should not lose your bond if you're always a winner or a runner-up.

If you're a winner or a runner-up and then *at the next election* lose completely, then you'll lose your bond. i.e. It's up to you to quit (or just not enter) before you lose.

gavofyork:
conflicts...

PR #4124, opened 2019-11-18
Relocate weight to paint + decouple from extensions

kianenigma:
- Makes `SignedExtension` independent of the type of the weight data, allowing a potential user to actually not have any sort of weight or class by setting `type Info = ()`. 

- Moves the weight related stuff to support, since they are part of our _opinionated_ runtime. 

Looks like we need a follow up PR to move all the extrinsic stuff to paint, but I'll do that in a follow up, since this has a bit of logic change in signed extension.


paint/elections-phragmen/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2018"
 codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false, features = ["derive"] }
 sr-primitives = { path = "../../primitives/sr-primitives", default-features = false }
 phragmen = { package = "substrate-phragmen", path = "../../primitives/phragmen", default-features = false }
-paint-support = { path = "../support", default-features = false }
+support = { package = "paint-support", path = "../support", default-features = false }
```
kianenigma: as soon as names are settled, one should make a PR and unify all of such imports.

paint/support/src/weights.rs
```diff
@@ -218,3 +219,39 @@ impl SimpleDispatchInfo {
 		Self::FixedNormal(0)
 	}
 }
+
+/// Implementation for unchecked extrinsic.
+impl<Address, Call, Signature, Extra> GetDispatchInfo
+	for sr_primitives::generic::UncheckedExtrinsic<Address, Call, Signature, Extra>
```
kianenigma: should fix these

paint/elections-phragmen/Cargo.toml
tomusdrw: I thought we no longer want to do renames?

paint/contracts/src/lib.rs
```diff
@@ -1034,6 +1034,7 @@ impl<T: Trait + Send + Sync> SignedExtension for CheckBlockGasLimit<T> {
 	type AccountId = T::AccountId;
 	type Call = <T as Trait>::Call;
 	type AdditionalSigned = ();
+	type Info = DispatchInfo;
```
tomusdrw:
```suggestion
	type DispatchInfo = DispatchInfo;
```
I know `Info` is shorter to type, but it's really generic.

paint/support/src/weights.rs
```diff
@@ -39,14 +39,18 @@
...
+	generic::{CheckedExtrinsic, UncheckedExtrinsic},
+};
 
 /// Re-export priority as type
-pub use crate::transaction_validity::TransactionPriority;
+pub use sr_primitives::transaction_validity::TransactionPriority;
 
 /// Numeric range of a transaction weight.
-pub type Weight = u32;
+pub type Weight = u64;
```
tomusdrw: That might be a breaking change. But I guess a bump of `spec_version` covers that, right?

paint/support/src/dispatch.rs
```diff
@@ -23,11 +23,11 @@ pub use paint_metadata::{
 	FunctionMetadata, DecodeDifferent, DecodeDifferentArray, FunctionArgumentMetadata,
 	ModuleConstantMetadata, DefaultByte, DefaultByteGetter, ModuleErrorMetadata, ErrorMetadata
 };
+pub use crate::weights::{
```
tomusdrw: Is it really worth to re-export? This stuff is in the same crate, I don't see a point - it will just introduce ambiguity.

paint/transaction-payment/src/lib.rs
```diff
@@ -56,7 +56,7 @@ type NegativeImbalanceOf<T> =
 
 pub trait Trait: system::Trait {
 	/// The currency type in which fees will be paid.
-	type Currency: Currency<Self::AccountId>;
+	type Currency: Currency<Self::AccountId> + Send + Sync;
```
tomusdrw: Can you elaborate why is `Send + Sync` required? I wouldn't expect to see that in any part of runtime.

primitives/sr-primitives/src/traits.rs
```diff
@@ -713,6 +712,9 @@ pub trait SignedExtension: Codec + Debug + Sync + Send + Clone + Eq + PartialEq
 	/// The type that encodes information that can be passed from pre_dispatch to post-dispatch.
 	type Pre: Default;
 
+	/// An opaque set of information attached to the transaction.
```
tomusdrw: Should we add some examples to the description? Like what it could be used for or when it's not needed, etc.

primitives/sr-primitives/src/traits.rs
```diff
@@ -713,6 +712,9 @@ pub trait SignedExtension: Codec + Debug + Sync + Send + Clone + Eq + PartialEq
...
+	type Info: Copy;
```
tomusdrw: `Copy` is a pretty strict, why not just `Clone`?

paint/support/src/weights.rs
pepyakin: Actually yes. Per our discussion with @kianenigma it shouldn't matter too much since we are not persisting it anywhere.

paint/elections-phragmen/Cargo.toml
kianenigma: indeed, but see my note: 

> as soon as names are settled, one should make a PR and unify all of such imports.

at least now that almost everywhere we use `paint-support` as `support`, having consistency is better IMO + makes `sed`-ing much easier. As mentioned, I'm more than happy to kill all of these as soon as we are done with paint -> palette.

paint/contracts/src/lib.rs
kianenigma: fair enough 👍

paint/support/src/weights.rs
kianenigma: Oh now I see what happened int he issue: I will undo this. This needs not too much work as sergei said, but probably another +50 file changes since a lot of tests might change. issue here: https://github.com/paritytech/substrate/issues/4141

paint/support/src/dispatch.rs
kianenigma: AFAIK: This stuff, similar to other re-exports of the module are needed to be accessed from within the macro. I learned this by example. @bkchr can probably either explain more or correct me.

paint/transaction-payment/src/lib.rs
kianenigma: Good point, I didn't really care much since I have seen these two in all SignedExtension type. For instance: https://github.com/paritytech/substrate/blob/0f87c92d2fb4bd7d9c235b1fb0caffc2ce58f803/paint/system/src/lib.rs#L853

So I assumed it is okay. But I will look it up. 

although, we use `Send + Sync` in `Member`  which is used around the runtime, so just by observation, I would not assume that we ban them in runtime. Although it does makes sense that a single threaded runtime should not really need them.

primitives/sr-primitives/src/traits.rs
kianenigma: I did face the dilemma and at the time the rational was: while we are making `Type Info` be generic but we currently know that we'll use it with a type that has `Copy` and this makes the syntax much much easier for now. If we see Copy as merely a marker that clone is cheap, I think it is fair to say that this bundled information will always be cheap to copy.

kianenigma: naah made it clone, was easier than what I thought to refactor.

primitives/sr-primitives/src/traits.rs
kianenigma: Did, though I must admit that this is not the end of the story, Once more of the refactors are done, we can technically use this to allow different runtimes, even different dispatches to carry different information in their `#[weight]`, but it would be super vague to formulate that + we will move stuff around soon again, so will probably change. I'll make a note for myself to elaborate on this after the repo is stable.

kianenigma:
will need some minor refactoring in polkadot

paint/transaction-payment/src/lib.rs
kianenigma: Needed due to `AccountId` being `Member`.

palette/executive/src/lib.rs
```diff
@@ -118,8 +118,10 @@ impl<
 		WeighBlock<System::BlockNumber>,
 > ExecuteBlock<Block> for Executive<System, Block, Context, UnsignedValidator, AllModules>
 where
-	Block::Extrinsic: Checkable<Context> + Codec,
-	CheckedOf<Block::Extrinsic, Context>: Applyable<AccountId=System::AccountId> + GetDispatchInfo,
+Block::Extrinsic: Checkable<Context> + Codec,
```
bkchr:
```suggestion
	Block::Extrinsic: Checkable<Context> + Codec,
```

palette/membership/src/lib.rs
```diff
@@ -24,10 +24,12 @@
...
 use rstd::prelude::*;
 use support::{
-	decl_module, decl_storage, decl_event, traits::{ChangeMembers, InitializeMembers},
+	decl_module, decl_storage, decl_event,
+	traits::{ChangeMembers, InitializeMembers},
+	weights::SimpleDispatchInfo,
 };
 use system::ensure_root;
-use sr_primitives::{traits::EnsureOrigin, weights::SimpleDispatchInfo};
+use sr_primitives::{traits::EnsureOrigin};
```
bkchr:
```suggestion
use sr_primitives::traits::EnsureOrigin;
```

PR #4125, opened 2019-11-18
[WIP] Refactor gossiping system

tomaka:
**This PR is a work in progress. It isn't finished yet (there are panics and TODOs), but I'm far ahead enough and I'm opening it in order to raise awareness around the changes and possibly a discussion.**

Refactors the gossiping system of Substrate.
These changes are (more or less) the outcome of the meeting we had on the second day of the retreat.

## Changes to the API surface of the network

- Removes `NetworkService::with_gossip` and `NetworkService::gossip_consensus_message`. The `consensus_gossip` module has also disappeared.
- Replaces these with three new methods: `NetworkService::events_stream`, `NetworkService::register_notif_protocol` and `NetworkService::write_notif`.
- Adds new elements to the `Event` enum`: when we open a notifications substream, when we close a notifications substream, and when we receive a message on a notifications substream.

(the terminology "notification" used below can be seen a synonym of "gossiping message", except that is a more generic concept)

The way you use the new API is as follow: first, register a "notifications protocol" using `register_notif_protocol`. At the same time, call `events_stream` in order to obtain a `Stream<Item = Event>`. The events received on this stream inform you about everything that happens on the network, most notably which notifications channels we open/close and the notifications that we receive. Call `write_notif` in order to send a notification on an open channel.

Just like it is the case right now, it is the network which decides who we open channels with.

## New crate: substrate-network-gossip

The new network API is way more low-level than previously. Most notably, there is no validation of messages and no method that does multicasting.

The reason for that is that these capabilities have been moved to a new crate called `substrate-network-gossip`. The `substrate-finality-grandpa` crate now depends on `substrate-network-gossip` rather than `substrate-network` alone.

The idea behind this design is that rather than having one pool of messages inside of the network crate that is used by all gossiping protocols combined, each protocol would have its own messages pool.

There are two ideas behind this design:

- Other gossiping protocols (such as the ones in Polkadot) may want to use a different gossiping strategy than the one in `substrate-network-gossip`.
- In the future, we can easily add strong typing in the `substrate-network-gossip` crate.
- In the future, we can also move the peerset in the `substrate-network-gossip` crate.

## Changes in substrate-finality-grandpa

The finality-grandpa crate has been adjusted for these changes:

- When creating a GrandPA engine, you had to pass an implementation of the `Network` trait located in `finality-grandpa`. This was providing a networking abstraction. We instead require a different trait, located in `substrate-network-gossip`.
- The `Network` trait has been removed entirely.
- Registering a validator is now done by creating a `GossipEngine` (calling `GossipEngine::new`) rather than a `register_validator` method on the `Network`.
- It is now a `GossipEngine` that is passed around everywhere rather than a `N` that implements `Network`. Since `GossipEngine` is only generic over the type of block, I removed a lot of template parameters.

## Changes inside of substrate-network

First of all, everything has been made backwards-compatible with the current network.

The protocols handler (i.e. the task dedicated to each connection) is aware of all the registered notifications protocols. In addition to the "legacy substream" (the one we use for everything at the moment), it tries to open a substream for each notification protocol. If the remote doesn't support the protocol, then it's not a big deal.

When we send a notifications message, the handler tries to see if the corresponding protocol-specific substream is open. If so, it sends the message on it. If not, it encodes a `ConsensusMessage` and sends it on the single main substream.

When we receive a `ConsensusMessage` on the single main substream, we report it as an `Event` as if it came on the dedicated protocol-specific substream.

The protocol-specific substreams are unidirectional, which means that each side has to open their own. One substream is used for messages from A to B, and a different one is used for messages from B to A. The substream handshake is as follow: after the protocol has been negotiated using multistream-select, the receiver has to send back a single protocol-specific handshake message. This is an occasion for the receiver to gracefully refuse a substream if it wishes so, by closing the substream before sending the handshake message. After B sends a handshake message, only A is allowed to send messages on this substream.


mxinden:
Thanks for tackling this and the detailed pull request description! I have
reviewed most of the changes and thus have a couple of comments.

> I'm opening it in order to raise awareness around the changes and possibly a
> discussion.

Given that this pull request is still marked as *work-in-progress*, can you
expand on the level of details you would like to see discussions around? E.g.
(1) 'General Architecture', (2) 'APIs' or (3) 'Implementations'.

tomaka:
> Given that this pull request is still marked as work-in-progress, can you
expand on the level of details you would like to see discussions around?

The PR is mostly finished, except for fixing the tests, maybe add documentation, and tackling the remaining TODOs.
It would be nice to refactor the code in `network-gossip` (which is quite messy), but I'm afraid to touch things too much here.

client/network/src/generic_proto/behaviour.rs
```diff
@@ -386,20 +387,49 @@ impl<TSubstream> GenericProto<TSubstream> {
...
+		});
+	}
+
+	/// Sends a notification to a peer.
+	///
+	/// Has no effect if the custom protocol is not open with the given peer.
+	///
+	/// Also note that even we have a valid open substream, it may in fact be already closed
+	/// without us knowing, in which case the packet will not be received.
+	pub fn write_notif(
```
rphmeier: `write_notification` please

client/network/src/generic_proto/handler/group.rs
```diff
@@ -119,13 +122,22 @@ pub enum NotifsHandlerIn {
...
-	/// Sends a message through a custom protocol substream.
+	/// Sends a message through the custom protocol substream.
 	Send {
-		/// Name of the protocol for the message, or `None` to force the legacy protocol.
+		/// The message to send.
+		message: Vec<u8>,
+	},
+
+	/// Sends a notifications message.
+	SendNotif {
```
rphmeier: `SendNotification`

tomaka:
Please note that my confidence in these changes is not super high. We should deploy it to a test network first.

client/network/src/service.rs
```diff
@@ -716,34 +767,38 @@ impl<B: BlockT + 'static, S: NetworkSpecialization<B>, H: ExHashT> Stream for Ne
...
-					return Ok(Async::Ready(Some(Event::Dht(ev)))),
-				Ok(Async::Ready(None)) => CustomMessageOutcome::None,
+				Ok(Async::Ready(Some(BehaviourOut::BlockImport(origin, blocks)))) =>
+					self.import_queue.import_blocks(origin, blocks),
+				Ok(Async::Ready(Some(BehaviourOut::JustificationImport(origin, hash, nb, justification)))) =>
+					self.import_queue.import_justification(origin, hash, nb, justification),
+				Ok(Async::Ready(Some(BehaviourOut::FinalityProofImport(origin, hash, nb, proof)))) =>
+					self.import_queue.import_finality_proof(origin, hash, nb, proof),
+				Ok(Async::Ready(Some(BehaviourOut::Event(ev)))) => {
+					self.events_streams.retain(|sender| sender.unbounded_send(ev.clone()).is_ok());
```
mxinden: We expect `unbounded_send` not to fail in many other places. Instead of silently dropping the events stream here, how about logging the failure?

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,254 @@
...
+					} else {
+						break;
+					}
+				}
+			}
+		});
+
+		async_std::task::spawn(async move {
+			let mut stream = Compat01As03::new(network.events_stream());
+			while let Some(Ok(event)) = stream.next().await {
```
mxinden: In case sending fails within network worker we drop the sender, thus the channel will be closed. I think we should either log the send failure in network worker, or log that the channel is closed here.

Network worker:

```rust
				Ok(Async::Ready(Some(BehaviourOut::Event(ev)))) => {
					self.events_streams.retain(|sender| sender.unbounded_send(ev.clone()).is_ok());
					return Ok(Async::Ready(Some(ev)));
				},
```

client/network/src/service.rs
```diff
@@ -659,6 +715,8 @@ pub struct NetworkWorker<B: BlockT + 'static, S: NetworkSpecialization<B>, H: Ex
 	from_worker: mpsc::UnboundedReceiver<ServerToWorkerMsg<B, S>>,
 	/// Receiver for queries from the light client that must be processed.
 	light_client_rqs: Option<mpsc::UnboundedReceiver<RequestData<B>>>,
+	/// Senders for events that happen on the network.
+	events_streams: Vec<mpsc::UnboundedSender<Event>>,
```
mxinden:
```suggestion
	event_streams: Vec<mpsc::UnboundedSender<Event>>,
```

I would call this `event_streams` given that a single `stream` already implies the plural for `event`.

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,254 @@
...
+				data: data.clone(),
+			});
+		}
+	}
+
+	pub fn announce(&self, block: B::Hash, associated_data: Vec<u8>) {
+		// TODO: self.announce_block(block, associated_data)
+	}
+
+	pub fn set_sync_fork_request(&self, peers: Vec<network::PeerId>, hash: B::Hash, number: NumberFor<B>) {
```
mxinden: While I know it is a convenient place to put this, this is not related to gossiping but much rather a way to communicate with another component (sync). I am guessing that only the Grandpa module needs this function. Thus this could e.g. be a channel between the Grandpa and the Sync instance. What do you think?

client/network-gossip/src/state_machine.rs
```diff
@@ -0,0 +1,920 @@
...
+
+			trace!(target: "gossip", "Propagating to {}: {:?}", id, message);
+			batch.push(message.clone())
+		}
+		protocol.send_consensus(id.clone(), batch);
+	}
+}
+
+/// Validates consensus messages.
+pub trait Validator<B: BlockT>: Send + Sync {
```
mxinden: With the trait definition here, can we remove the one in `consensus_gossip.rs`?

tomaka: Oh I meant to delete `consensus_gossip.rs` and forgot.

client/network/src/service.rs
tomaka: It can only fail if the receiving side gets dropped. IMO dropping the receiver is a clean and legitimate way to indicate that we're no longer interested in events.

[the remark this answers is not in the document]
tomaka: I guess we could put the message as a `Bytes` (which is kind of the equivalent of an `Arc<Vec<u8>>`).

client/network-gossip/src/bridge.rs
tomaka: There's actually already an abstraction for that, but it's used only in one module: https://github.com/paritytech/substrate/blob/2fa6a2647ee430604139c2b9efb2e156983d166f/client/finality-grandpa/src/lib.rs#L271

client/network-gossip/src/bridge.rs
tomaka: To me that's a legitimate way to end the task.
If the `NetworkWorker` gets destroyed, then we want this background task to get closed as well.

client/network/src/service.rs
```diff
@@ -416,6 +421,57 @@ impl<B: BlockT + 'static, S: NetworkSpecialization<B>, H: ExHashT> NetworkServic
...
+		});
+	}
+
+	/// Returns a stream containing the events that happen on the network.
+	///
+	/// If this method is called multiple times, the events are duplicated.
+	///
+	/// The stream never ends (unless the `NetworkWorker` gets shut down).
+	// Note: when transitioning to stable futures, remove the `Error` entirely
+	pub fn events_stream(&self) -> impl Stream<Item = Event, Error = ()> {
```
mxinden:
```suggestion
	pub fn event_stream(&self) -> impl Stream<Item = Event, Error = ()> {
```

client/network/src/service.rs
```diff
@@ -716,34 +775,40 @@ impl<B: BlockT + 'static, S: NetworkSpecialization<B>, H: ExHashT> Stream for Ne
...
-					return Ok(Async::Ready(Some(Event::Dht(ev)))),
-				Ok(Async::Ready(None)) => CustomMessageOutcome::None,
+				Ok(Async::Ready(Some(BehaviourOut::BlockImport(origin, blocks)))) =>
+					self.import_queue.import_blocks(origin, blocks),
+				Ok(Async::Ready(Some(BehaviourOut::JustificationImport(origin, hash, nb, justification)))) =>
+					self.import_queue.import_justification(origin, hash, nb, justification),
+				Ok(Async::Ready(Some(BehaviourOut::FinalityProofImport(origin, hash, nb, proof)))) =>
+					self.import_queue.import_finality_proof(origin, hash, nb, proof),
+				Ok(Async::Ready(Some(BehaviourOut::Event(ev)))) => {
+					self.event_streams.retain(|sender| sender.unbounded_send(ev.clone()).is_ok());
```
mxinden: Follow up to https://github.com/paritytech/substrate/pull/4125#discussion_r348355881 (can't comment there no more).

Either a `Bytes` or we make the protocol name available here so we can filter early on?

client/network-gossip/src/bridge.rs
mxinden: Ah, and with the above down and upgrade of the `Arc` we ensure to end the other spawned task as well?

client/network/src/service.rs
mxinden: That makes sense. Thanks!

client/network-gossip/src/bridge.rs
mxinden: Right. This is used as a way to do dependency injection and was meant for the finality grandpa module only.

Are you suggesting to have `NetworkService` implement this trait and thus still passing a clone of `NetworkService` into grandpa?

arkpar:
How does this handle backwards compatibility?

mxinden:
> First of all, everything has been made backwards-compatible with the current network.

@arkpar I guess the goal is to be fully backwards compatible, even though I remember @tomaka mentioning somewhere else that there was something missing. Can you add more details here @tomaka?

---

What else is missing? Would you like another round of review on this pull request or help with anything in particular?

tomaka:
> What else is missing? Would you like another round of review on this pull request or help with anything in particular?

There is kind of a fundamental issue: what if the remote opens a gossip substream before we do so? This can't happen with the code right now but could theoretically happen and will happen in the future once we deprecate the legacy protocol.

We would then either need to report to the API user that we received a message from a closed connection (not great), or report to the API user that we opened a connection, but that it is not unavailable for syncing and light requests.

tomaka:
Because of the size of the changes, I'm really not confident of the robustness of this code.
Instead I'm going to split this PR in two, and first merge the gossiping extraction code.
Then later we can merge the second part, which is making the gossiping open new substreams.

tomaka:
Closing as it was superseded by #4284

mxinden:
@tomaka https://github.com/paritytech/substrate/pull/4284 *only* extracted the gossip code into its own module. We would still need (/want) to have the gossiping system open new substreams, correct?

mxinden:
Friendly ping @tomaka. (Not implying that you have to do the work, but trying to define our goals.)

tomaka:
Yes, we still have to do part 2, except that at this point I'd prefer to start from scratch.
I purposefully didn't delete the branch of this PR in order to be able to read the code for inspiration.
Let me know if you're interested.

mxinden:
Effort continued in https://github.com/paritytech/substrate/pull/4909.

Issue #4099, opened 2019-11-13
repo reorg tracking issues


gnunicorn:
First batch in #4074 .

Next up:
- [x] remove `palette`-dependencies from `client` (move inherents to primitives):
- [x] move into primitives: #4184
- [x] rename all the crates to their area:
  - [x] `/primitives`: prefixed `sp-`
  - [x] `/client`: prefixed `sc-`
  - [x] clean up aliases, enforce long name usage
- [x] update versions, to bump below 1.0:
  - [x] networking
  - [x] consensus
- [x] remove client internal deps:
  - [x] basic-authorship
  - [x] aura
  - [x] babe
  - [x] finality-grandpa
  - [x] client-db
  - [x] network
  - [x] rpc
- [x] test reorg:
  - [x] network

probably post 2.0:
- [ ] merge remaining client into service
- [ ] clean up frame-support
- [ ] clean up frame-system
- [ ] Merge sp-serializer with sp-core or sp-runtime

gnunicorn:
PR pending. closing.

PR #4126, opened 2019-11-18
Move inherents to primitives

gnunicorn:
Next step towards #4099, this PR removes any dependency from `client` to `paint`. This is being achieved by moving all inherents into primitives, creating three more super thin crates, that all others then reference. The newly added files, therefore, are just code moves with minor modifications regarding import names.

The new crates are prefixed already with `sp-` as per the convention that will be following.


client/consensus/aura/src/lib.rs
```diff
@@ -654,9 +660,9 @@ fn register_aura_inherent_data_provider(
 	inherent_data_providers: &InherentDataProviders,
 	slot_duration: u64,
 ) -> Result<(), consensus_common::Error> {
-	if !inherent_data_providers.has_provider(&paint_aura::INHERENT_IDENTIFIER) {
+	if !inherent_data_providers.has_provider(&aura_primitives::inherents::INHERENT_IDENTIFIER) {
```
tomusdrw: Can't we just import it directly? I find it a bit confusing to have both `use aura_primitives::...` at the top and then use full path to import some other stuff (I know the other one is `pub use`, but still the full path here feels a bit too long).

primitives/timestamp/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "sp-timestamp"
+version = "2.0.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+
+[dependencies]
+sr-api = { path = "../sr-api", default-features = false }
+rstd = { package = "sr-std", path = "../sr-std", default-features = false }
```
tomusdrw: no renames?

gnunicorn: I mostly kept the existing names and copied `Cargo.toml` files to keep the changes low. I would fix this in the upcoming PR if you don't mind.

tomusdrw: Sure!

gavofyork:
looks good - merge when ready @gnunicorn

PR #4129, opened 2019-11-18
Support custom runtime interfaces in native executor

bkchr:
This makes it possible to use custom runtime interfaces within your
runtime by registering them at the native executor.

### DIFF
--- a/client/executor/Cargo.toml
+++ b/client/executor/Cargo.toml
@@ -37,13 +37,17 @@ assert_matches = "1.3.0"
 wabt = "0.9.2"
 hex-literal = "0.2.1"
 runtime-test = { package = "substrate-runtime-test", path = "runtime-test" }
+test-runtime = { package = "substrate-test-runtime", path = "../../test/utils/runtime" }
+runtime-interface = { package = "substrate-runtime-interface", path = "../../primitives/runtime-interface" }
 client-api = { package = "substrate-client-api", path = "../api" }
 substrate-offchain = { path = "../offchain/" }
 state_machine = { package = "substrate-state-machine", path = "../../primitives/state-machine"  }
 test-case = "0.3.3"
 
 [features]
-default = []
+default = [ "std" ]
+# This crate does not have `no_std` support, we just require this for tests
+std = []
 wasm-extern-trace = []
 wasmtime = [
 	"cranelift-codegen",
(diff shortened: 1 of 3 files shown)


client/executor/src/native_executor.rs
```diff
@@ -283,3 +339,34 @@ macro_rules! native_executor_instance {
...
+		fn say_hello_world(data: &str) {
+			println!("Hello world from: {}", data);
+		}
+	}
+
+	native_executor_instance!(
+		pub MyExecutor,
+		test_runtime::api::dispatch,
+		test_runtime::native_version,
+		(my_interface::HostFunctions, my_interface::HostFunctions),
```
tomusdrw: Shouldn't a name clash be an error?

client/executor/Cargo.toml
```diff
@@ -37,13 +37,17 @@ assert_matches = "1.3.0"
...
+runtime-interface = { package = "substrate-runtime-interface", path = "../../primitives/runtime-interface" }
 client-api = { package = "substrate-client-api", path = "../api" }
 substrate-offchain = { path = "../offchain/" }
 state_machine = { package = "substrate-state-machine", path = "../../primitives/state-machine"  }
 test-case = "0.3.3"
 
 [features]
-default = []
+default = [ "std" ]
+# This crate does not have `no_std` support, we just require this for tests
```
tomusdrw: Can you elaborate more why this is needed? I thought that tests run in `std` anyway? We don't seem to be using that feature anywhere in the code afaict.

gnunicorn:
:dancing_banana_gif:

brenzi:
cool, looking forward to this one

client/executor/src/native_executor.rs
bkchr: While implementing it, I was not sure. Maybe we want to support at one point that users even can overwrite the default runtime interface functions. 

(This is just on the host side, so the stuff that is registered in the executor)

client/executor/Cargo.toml
bkchr: `#[runtime_interface]` requires a `std` feature. Otherwise the generated code expects wasm. I could probably at some point change the feature selection to `any(std, test)`.

client/executor/src/native_executor.rs
clangenb: A feature we would welcome very much for more easy Intel SGX support. :)

bkchr: Which functions do you need to override and why?

clangenb: We had running an runtime in a trusted execution environment, (which implies `no_std` , but not wasm). With the previous implementation we just patched the sr-io module, which I hoped to circumvent now. 

But now I believe that this is not the core issue. Is it correct that in order to do that, one would need to implement a `no_std` compatible `Externalities` trait?

bkchr: Hmm?

PR #4132, opened 2019-11-18
Add instrumentation, with performance profiling capability

mattrutherford:
Make available a flexible way to gather performance metrics from anywhere in the codebase.

This initial implementation is basic but extendable and sufficient to begin collecting some key values for analytic purposes.

The Subscriber implementation currently uses a Vec (with natural ordering of elements) to store spans. Removal of items should be efficient based on my assumption that we're going to be mostly removing items near the end of the stack, however it might make sense to either use Hashmap, or at some later stage convert to a more elaborate system to improve efficiency.

This currently supports sending the metrics via telemetry, but in a follow up PR we can add CSV, Grafana or any other implementation.

I decided against feature-gating this because the tracing library is designed to have minimal overhead in the case where a span is not enabled. 

Activating spans requires a restart but it's possible we can in future provide an RPC to toggle things on/off.

We may want to switch to using multiple `Layer`s in future if needs grow more complex.


client/instrumentation/src/lib.rs
```diff
@@ -0,0 +1,151 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```
kianenigma:
```suggestion
// Copyright 2019 Parity Technologies (UK) Ltd.
```

client/instrumentation/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+use std::sync::atomic::{AtomicU64, Ordering};
```
kianenigma: I won't leave comments everywhere but in general for substrate: at least a one liner as module doc. All pub items should also have proper rustdoc (///, not //), including all public fields.

paint/support/src/dispatch.rs
```diff
@@ -1009,9 +1050,19 @@ macro_rules! decl_module {
...
+			#[cfg(feature = "std")]
+			{
+				use $crate::tracing;
+				let span = tracing::span!(tracing::Level::INFO, stringify!($name));
+				let _enter = span.enter();
+				let result: $crate::dispatch::DispatchResult<$error_type> = (move || { { $( $impl )* } Ok(()) })();
+				result
+			}
+			#[cfg(not(feature = "std"))]
+			return {
```
kianenigma:
```suggestion
		 {
```

I think you don't need the returns + no need to store 'result' everywhere

paint/support/src/dispatch.rs
```diff
@@ -1009,9 +1050,19 @@ macro_rules! decl_module {
...
+				{ $( $impl )* }
+				Ok(())
+			};
```
kianenigma:
```suggestion
			}
```

paint/support/Cargo.toml
```diff
@@ -20,6 +20,7 @@ once_cell = { version = "0.2.4", default-features = false, optional = true }
 state-machine = { package = "substrate-state-machine", path = "../../primitives/state-machine", optional = true }
 bitmask = { version = "0.5.0", default-features = false }
 impl-trait-for-tuples = "0.1.3"
+tracing = { version = "0.1.2", optional = true }
```
bkchr: Please use the latest version.

client/instrumentation/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+			}
+		}
+		None => (s.to_string(), Level::TRACE)
+	}
+}
+
+impl Subscriber for ProfilingSubscriber {
+	fn enabled(&self, metadata: &Metadata<'_>) -> bool {
+		for x in &self.targets {
+			if metadata.target().starts_with(x.0.as_str()) && metadata.level() <= &x.1 {
```
marcio-diaz: Try to use more meaningful names and maybe transform `x` from tuple to struct.

paint/support/src/dispatch.rs
mattrutherford: good catch - I had just copied old code without thinking enough

client/instrumentation/src/lib.rs
mattrutherford: I've swapped out `x` for `t` here, but seeing as it's a small private fn and concise 1 letter variable is easier to read - I think is best compromise. I did however follow your suggestion and updated some other variables in this module to have better names.

client/instrumentation/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "substrate-instrumentation"
```
pepyakin: I wonder if we can come up with a more specific name for that? Instrumentation sounds a bit too geeneral and for me implies a totally different thing.

How about "substrate-timings"?

mattrutherford: I wasn't totally sure about the name and I'm happy to make this change if people want it. The reason I went with instrumentation is because it's possible we might want to use it for other forms of instrumentation in addition to, or other than, timing. There's potential to selectively capture a lot of information using this tool - I'm not sure which exact scenarios would work better than our current logging system, but thought to allow that flexibility. However - if we want to keep the scope of this crate nice and focused then let's make the change - I'm happy either way. If so, should the CLI param be `--timing` ?

marcio-diaz: Or `--time-profiling`. Probably you can add `telemetry` as default so we don't needed to type it (although a better default may be just `stdout`).

mattrutherford: that's a good point, stdout would be most useful for people I guess - I'll add that and include another CLI param to configure it

mattrutherford: also, just another thought regarding name - maybe `timing` or `timings` is too generic sounding - what about `profiling`?

client/instrumentation/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+fn send_telemetry(span_datum: SpanDatum) {
+	telemetry!(SUBSTRATE_INFO; "instrumentation.profiling";
+		"name" => span_datum.name,
+		"target" => span_datum.target,
+		"line" => span_datum.line,
+		"time" => span_datum.overall_time.as_nanos(),
+	);
+}
```
pepyakin:
```suggestion
}

```

client/instrumentation/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+
+use parking_lot::Mutex;
+use tracing_core::{event::Event, Level, metadata::Metadata, span::{Attributes, Id, Record}, subscriber::Subscriber};
+
+use substrate_telemetry::{telemetry, SUBSTRATE_INFO};
+
+/// Used to configure how to receive the metrics
+pub enum Receiver {
+	Telemetry,
+}
```
pepyakin: I think for now it might be OK, but I'd actually revisit this approach in favor of traits.

client/instrumentation/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+//!
+//! This crate is unstable and the API and usage may change.
+//!
+//! # Usage
+//!
+//! Monitor performance throughout the codebase via the creation of `Span`s.
+//! A span is set in the following way:
+//! ```ignore
+//! let span = tracing::span!(tracing::Level::INFO, "my_span_name");
+//!	let _enter = span.enter();
```
pepyakin:
```suggestion
//! let _enter = span.enter();
```

client/instrumentation/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+license = "GPL-3.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+
+[dependencies]
+log = { version = "0.4.8" }
+parking_lot = "0.9.0"
+tracing-core = "0.1.7"
+
+substrate-telemetry = { path = "../telemetry" }
```
pepyakin:
```suggestion
substrate-telemetry = { path = "../telemetry" }

```

client/instrumentation/src/lib.rs
mattrutherford: Ok, sounds like a good idea - will revisit in follow-up PR

mattrutherford:
Added feature to output tracing info to logger, substrate can now be run with `-l off` to isolate the output from tracing, eg:

```
substrate --dev -l off -d ~/Downloads/data14 --tracing-targets=paint=info
2019-11-21 09:17:54 Tracing: INFO paint_babe: on_initialize, line: 235, time: 49592ns
2019-11-21 09:17:54 Tracing: INFO paint_authorship: on_initialize, line: 213, time: 113336ns
2019-11-21 09:17:54 Tracing: INFO paint_session: on_initialize, line: 461, time: 51683ns
2019-11-21 09:17:54 Tracing: INFO paint_democracy: on_initialize, line: 333, time: 58781ns
2019-11-21 09:17:54 Tracing: INFO paint_elections_phragmen: on_initialize, line: 172, time: 23242ns
2019-11-21 09:17:54 Tracing: INFO paint_grandpa: on_initialize, line: 171, time: 106ns
2019-11-21 09:17:54 Tracing: INFO paint_randomness_collective_flip: on_initialize, line: 70, time: 36391ns
2019-11-21 09:17:54 Tracing: INFO paint_babe: on_initialize, line: 235, time: 41206ns
2019-11-21 09:17:54 Tracing: INFO paint_authorship: on_initialize, line: 213, time: 76272ns
2019-11-21 09:17:54 Tracing: INFO paint_session: on_initialize, line: 461, time: 34059ns
2019-11-21 09:17:54 Tracing: INFO paint_democracy: on_initialize, line: 333, time: 142362ns
2019-11-21 09:17:54 Tracing: INFO paint_elections_phragmen: on_initialize, line: 172, time: 25504ns
2019-11-21 09:17:54 Tracing: INFO paint_grandpa: on_initialize, line: 171, time: 118ns
2019-11-21 09:17:54 Tracing: INFO paint_randomness_collective_flip: on_initialize, line: 70, time: 37463ns
2019-11-21 09:17:54 Tracing: INFO paint_timestamp: set, line: 213, time: 40815ns
2019-11-21 09:17:54 Tracing: INFO paint_finality_tracker: final_hint, line: 113, time: 23561ns
2019-11-21 09:17:54 Tracing: INFO paint_babe: on_finalize, line: 235, time: 1858ns
2019-11-21 09:17:54 Tracing: INFO paint_timestamp: on_finalize, line: 213, time: 1824ns
2019-11-21 09:17:54 Tracing: INFO paint_authorship: on_finalize, line: 213, time: 1557ns
2019-11-21 09:17:54 Tracing: INFO paint_transaction_payment: on_finalize, line: 83, time: 26863ns
2019-11-21 09:17:54 Tracing: INFO paint_staking: on_finalize, line: 666, time: 24598ns
2019-11-21 09:17:54 Tracing: INFO paint_finality_tracker: on_finalize, line: 113, time: 67976ns
2019-11-21 09:17:54 Tracing: INFO paint_grandpa: on_finalize, line: 171, time: 33710ns
2019-11-21 09:17:54 Tracing: INFO paint_treasury: on_finalize, line: 114, time: 264ns
2019-11-21 09:17:54 Tracing: INFO paint_contracts: on_finalize, line: 457, time: 1101ns
```

Default option is now to use logger, telemetry output configurable by cli with `--tracing-receiver <RECEIVER>` - `[possible values: Log, Telemetry]`

Also renamed `intrumentation` to `substrate-tracing` after considering it for a while - I'm still open to discussion, but this seems best to me.

mattrutherford:
after latest merge, ^^^ becomes:
```
substrate --dev -l off --tracing-targets pallet=info
2019-11-21 09:57:21 Tracing: INFO pallet_babe: on_initialize, line: 161, time: 11996ns
2019-11-21 09:57:21 Tracing: INFO pallet_authorship: on_initialize, line: 165, time: 70713ns
2019-11-21 09:57:21 Tracing: INFO pallet_session: on_initialize, line: 461, time: 29739ns
2019-11-21 09:57:21 Tracing: INFO pallet_democracy: on_initialize, line: 333, time: 32754ns
2019-11-21 09:57:21 Tracing: INFO pallet_elections_phragmen: on_initialize, line: 172, time: 9505ns
2019-11-21 09:57:21 Tracing: INFO pallet_grandpa: on_initialize, line: 171, time: 80ns
2019-11-21 09:57:21 Tracing: INFO pallet_randomness_collective_flip: on_initialize, line: 70, time: 53743ns
2019-11-21 09:57:21 Tracing: INFO pallet_babe: on_initialize, line: 161, time: 22861ns
2019-11-21 09:57:21 Tracing: INFO pallet_authorship: on_initialize, line: 165, time: 91879ns
2019-11-21 09:57:21 Tracing: INFO pallet_session: on_initialize, line: 461, time: 13084ns
2019-11-21 09:57:21 Tracing: INFO pallet_democracy: on_initialize, line: 333, time: 28046ns
2019-11-21 09:57:21 Tracing: INFO pallet_elections_phragmen: on_initialize, line: 172, time: 7844ns
2019-11-21 09:57:21 Tracing: INFO pallet_grandpa: on_initialize, line: 171, time: 70ns
2019-11-21 09:57:21 Tracing: INFO pallet_randomness_collective_flip: on_initialize, line: 70, time: 27002ns
2019-11-21 09:57:21 Tracing: INFO pallet_timestamp: set, line: 126, time: 19900ns
2019-11-21 09:57:21 Tracing: INFO pallet_babe: on_finalize, line: 161, time: 910ns
2019-11-21 09:57:21 Tracing: INFO pallet_timestamp: on_finalize, line: 126, time: 837ns
2019-11-21 09:57:21 Tracing: INFO pallet_authorship: on_finalize, line: 165, time: 938ns
2019-11-21 09:57:21 Tracing: INFO pallet_transaction_payment: on_finalize, line: 83, time: 12840ns
2019-11-21 09:57:21 Tracing: INFO pallet_staking: on_finalize, line: 666, time: 13243ns
2019-11-21 09:57:21 Tracing: INFO pallet_finality_tracker: on_finalize, line: 59, time: 41528ns
2019-11-21 09:57:21 Tracing: INFO pallet_grandpa: on_finalize, line: 171, time: 17260ns
2019-11-21 09:57:21 Tracing: INFO pallet_treasury: on_finalize, line: 114, time: 145ns
2019-11-21 09:57:21 Tracing: INFO pallet_contracts: on_finalize, line: 457, time: 2879ns
```

client/instrumentation/Cargo.toml
mattrutherford: after some more consideration I have renamed to substrate-tracing, to align with the use of the tracing library for substrate, but also happy to switch to the above suggestion of @pepyakin if we want to be more focused/descriptive, my preference is slightly toward the current name, but not by much

palette/support/src/dispatch.rs
```diff
@@ -861,7 +862,17 @@ macro_rules! decl_module {
 			$crate::sr_primitives::traits::OnInitialize<$trait_instance::BlockNumber>
 			for $module<$trait_instance$(, $instance)?> where $( $other_where_bounds )*
 		{
-			fn on_initialize(_block_number_not_used: $trait_instance::BlockNumber) { $( $impl )* }
+			fn on_initialize(_block_number_not_used: $trait_instance::BlockNumber) {
+				#[cfg(feature = "std")]
```
tomusdrw: There is a nice `if_std!` macro in `sr-std` for that.

mattrutherford: Cool, thanks for letting me know about that stuff. It's done now

client/tracing/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+//! Monitor performance throughout the codebase via the creation of `Span`s.
+//! A span is set in the following way:
+//! ```
+//! let span = tracing::span!(tracing::Level::INFO, "my_span_name");
+//! let _enter = span.enter();
+//! ```
+//! To begin timing, a span must be entered. When the span is dropped, the execution time
+//! is recorded and details sent to the `Receiver` which defines how to process it.
+//!
+//! Currently we provide a single `Telemetry` variant for `Receiver`
```
kianenigma:
```suggestion
//! Currently we provide a single `Telemetry` variant for `Receiver`.
```

kianenigma: Don't we also have Log somewhere?

palette/support/src/dispatch.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017-2019 Parity Technologies (UK) Ltd.
+// Copyright 2019 Parity Technologies (UK) Ltd.
```
kianenigma: no each file need to be have a license for the year it was created :D this can stay 2017-2019

palette/support/src/dispatch.rs
```diff
@@ -1009,9 +1042,16 @@ macro_rules! decl_module {
...
 			$origin: $origin_ty $(, $param: $param_ty )*
 		) -> $crate::dispatch::DispatchResult<$error_type> {
-			{ $( $impl )* }
-			// May be unreachable.
-			Ok(())
+			use $crate::rstd::if_std;
+			if_std! {
+				use $crate::tracing;
+				let span = tracing::span!(tracing::Level::INFO, stringify!($name));
+				let _enter = span.enter();
```
kianenigma: hmm I think I read somewhere that these stop timing when they go out of scope? I assume the scoping here is okay and they end when the function impl goes out of scope? I am worried that the `_enter` is created an immediately dropped since it is in a separate scope

client/tracing/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+impl Default for TracingReceiver {
+	fn default() -> Self {
+		Self::Log
+	}
+}
+
+#[derive(Debug)]
+struct SpanDatum {
```
kianenigma: are we naming German now here? what does `Datum` stand for? :D

client/tracing/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+			TracingReceiver::Log => print_log(span_datum),
+			TracingReceiver::Telemetry => send_telemetry(span_datum),
+		}
+	}
+}
+
+fn print_log(span_datum: SpanDatum) {
+	let message = format!(
+		"Tracing: {} {}: {}, line: {}, time: {}ns",
+		span_datum.level, span_datum.target, span_datum.name, span_datum.line, span_datum.overall_time.as_nanos()
```
kianenigma:
```suggestion
		span_datum.level, 
		span_datum.target, 
		span_datum.name, 
		span_datum.line, 
		span_datum.overall_time.as_nanos(),
```

palette/support/src/dispatch.rs
kianenigma: I suppose it is fine for now, but in the future, maybe we'd want to actually make this part in dispatch.rs be gated with a special feature? so that not all nodes (as many of them, even when running the native compile) spend time on this and report and only those that have been compiled with a special flag do. I assume, for instance, as a validator behind a sentry node you don't want this stuff, nor do you want to report anything to telementry.

client/cli/src/params.rs
```diff
@@ -491,6 +509,20 @@ pub struct RunCmd {
 	#[structopt(long = "force-authoring")]
 	pub force_authoring: bool,
 
+	/// Comma separated list of targets for tracing
+	#[structopt(long = "tracing-targets")]
+	pub tracing_targets: Option<String>,
```
marcio-diaz: I think you should add `value_name = "TARGETS"` so it look consistent with other flags.

client/tracing/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+		match self.receiver {
+			TracingReceiver::Log => print_log(span_datum),
+			TracingReceiver::Telemetry => send_telemetry(span_datum),
+		}
+	}
+}
+
+fn print_log(span_datum: SpanDatum) {
+	let message = format!(
+		"Tracing: {} {}: {}, line: {}, time: {}ns",
```
marcio-diaz: `{} ns` looks better and can facilitate parsing. Maybe try using tabs so times/columns can be easily compared.

palette/support/src/dispatch.rs
mattrutherford: I looked into this at the start and I'd say this isn't necessary - in the runtime we're calling these spans maximum a few hundred times per second, and with no targets enabled there will be (for practical purposes) no real cost to it in this situation. Here's a link to an expanded version of `balances`  https://gist.github.com/mattrutherford/8b0395a1e061c4b5f0fdbd39aea70382 

Let me know what you think?

palette/support/src/dispatch.rs
mattrutherford: yeah, it does stay in scope - see here: https://gist.github.com/mattrutherford/8b0395a1e061c4b5f0fdbd39aea70382

give it a try if in doubt: 
```./target/release/substrate --dev --tracing-targets=pallet=info``` 

Values seem to be in expected range.

client/tracing/src/lib.rs
mattrutherford: "a single piece of information, as a fact, statistic, or code; an item of data."

client/tracing/src/lib.rs
mattrutherford: I don't think tabs will work well because there can be a huge variation in module path

mattrutherford: but have added your suggestion for spacing the units

client/cli/src/params.rs
mattrutherford: done

mattrutherford:
> Looks good in general; I would still consider compiling this conditionally.

I think it's OK as is - see answer above: https://github.com/paritytech/substrate/pull/4132#discussion_r349534991

Also I think we should try to avoid bloating features, unless really required.

Let me know if you think differently?

gavofyork:
can put it in for now and feature gate if profiling turns this out as a bottleneck.

mattrutherford:
OK, thanks all. Just re-trying the CI job and will merge on success

PR #4133, opened 2019-11-18
fix full_crypto

brenzi:
This fixes issues with the `full_crypto` feature. We had trouble using the feature with 
https://github.com/scs/substrate-api-client

More specifically, the following test crate only builds with this patch
https://github.com/scs/substrate-api-client/tree/cl-update-to-6ae3b6c4ddc03d4cdb10bd1d417b95d20f4c1b6e/test_no_std

Without this patch rust complains about missing `MaybeHash` and `Vec` not found in this scope

### DIFF
--- a/primitives/application-crypto/src/ed25519.rs
+++ b/primitives/application-crypto/src/ed25519.rs
@@ -24,6 +24,7 @@ pub use primitives::ed25519::*;
 
 mod app {
 	use primitives::testing::ED25519;
+	use rstd::vec::Vec;
 	crate::app_crypto!(super, ED25519);
 
 	impl crate::traits::BoundToRuntimeAppPublic for Public {
--- a/primitives/application-crypto/src/ed25519.rs
+++ b/primitives/application-crypto/src/ed25519.rs
@@ -24,7 +24,6 @@ pub use primitives::ed25519::*;
 
 mod app {
 	use primitives::testing::ED25519;
-	use rstd::vec::Vec;
 	crate::app_crypto!(super, ED25519);
 
 	impl crate::traits::BoundToRuntimeAppPublic for Public {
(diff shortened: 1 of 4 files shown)


primitives/application-crypto/src/ed25519.rs
```diff
@@ -24,6 +24,7 @@ pub use primitives::ed25519::*;
 
 mod app {
 	use primitives::testing::ED25519;
+	use rstd::vec::Vec;
```
bkchr: This is not correct. The macro should internally use a full path to vec. (rstd is probably already exported by application crypto)

And then use `$crate::rstd::vec::Vec`

brenzi: Not sure I understood you right. The above actually built correctly for me (but CI complained...probably rust version?). However, I've changed it to `$crate.....$`

brenzi:
It seems I'm in a deadlock. I can't appease the build of substrate with the build of substrate-api-client at the same time. With this commit I get:
```
error[E0412]: cannot find type `Vec` in this scope
   --> /home/abrenzikofer/.cargo/git/checkouts/substrate-6b44b349a63f0531/5d921fd/primitives/application-crypto/src/lib.rs:145:28
    |
145 |             fn to_raw_vec(&self) -> Vec<u8> { self.0.to_raw_vec() }
    |                                     ^^^ not found in this scope
    | 
   ::: /home/abrenzikofer/.cargo/git/checkouts/substrate-6b44b349a63f0531/5d921fd/primitives/application-crypto/src/ed25519.rs:27:5
    |
27  |     crate::app_crypto!(super, ED25519);
    |     ----------------------------------- in this macro invocation
    |
help: possible candidates are found in other modules, you can import them into scope
    |
26  |     use rstd::prelude::Vec;
    |
26  |     use rstd::vec::Vec;
    |

```
for https://github.com/scs/substrate-api-client/tree/cl-update-to-6ae3b6c4ddc03d4cdb10bd1d417b95d20f4c1b6e

brenzi:
ok, this might work

primitives/application-crypto/src/ed25519.rs
brenzi: see below...

brenzi:
won't bother with CI, it seems that all recent commits on master fail these checks....

PR #4134, opened 2019-11-18
Renaming and documentation for ApplyResult, ApplyOutcome and et al

pepyakin:
This PR features:

- Removing unused variants from `ApplyError`, preserving the enumeration (not sure about that, should we just use `TransactionValidity` directly?),
- Renaming the `ApplyResult` and `ApplyOutcome` to (hopefully) less confusing ones. 
- Adding documentation that (again hopefuly) further reduces confusion.

This change doesn't require version bump, although it changes the external API. Therefore [here](https://github.com/paritytech/polkadot/pull/596) is the accompanying Polkadot PR.


bkchr:
(This breaks the block authoring of Kusama)

pepyakin:
Can you shed some light on why is that?

bkchr:
We can not decode `ApplyResult`/`InclusionOutput` anymore. It is solvable, but will require a little bit more effort.
We need to add a compatibility layer as shown here: https://github.com/paritytech/substrate/pull/3585. This would require that all validators update before we send out the runtime update to the network. (we should increment the authoring version as well)

primitives/block-builder/runtime-api/src/lib.rs
```diff
@@ -18,16 +18,19 @@
...
 
 sr_api::decl_runtime_apis! {
 	/// The `BlockBuilder` api trait that provides the required functionality for building a block.
 	#[api_version(3)]
 	pub trait BlockBuilder {
-		/// Apply the given extrinsics.
-		fn apply_extrinsic(extrinsic: <Block as BlockT>::Extrinsic) -> ApplyResult;
+		/// Apply the given extrinsic.
+		///
+		/// Returns an inclusion outcome which specifies if this extrinsic should be included in
```
bkchr: Hmm, this "should be". Isn't it more like "could be"? 
In some sense you are right, we still could discard the prospective, but it is more like that the runtime should decide and not the node if a transaction should be included.

primitives/sr-primitives/src/lib.rs
```diff
@@ -348,56 +348,41 @@ impl From<ed25519::Signature> for AnySignature {
...
-	/// General error to do with the permissions of the sender.
-	NoPermission,
-
-	/// General error to do with the state of the system in general.
-	BadState,
-
+/// Reason why an extrinsic couldn't be included into a block.
+pub enum InclusionError {
 	/// Any error to do with the transaction validity.
 	Validity(transaction_validity::TransactionValidityError),
```
bkchr: If the enum just has one variant, we should drop it and just use `TransactionValidityError`.

primitives/sr-primitives/src/lib.rs
```diff
@@ -451,6 +436,27 @@ impl From<&'static str> for DispatchError {
...
+
+/// The outcome of inclusion of an extrinsic into a block.
+///
+/// This type is typically used in the context of `BlockBuilder` to signal that the extrinsic
+/// in question cannot be included. It is fair to say that a valid block doesn't contain any
+/// extrinsic that would have had a negative inclusion outcome. On successful inclusion this type
+/// supplies the result of the extrinsic dispatch.
+///
+/// Examples of reasons preventing inclusion in a block:
+/// - More block weight is required to process the extrinsic than is left in the block being built.
```
kianenigma: I wouldn't name weight here and just say block resources. Weight is Palette specific while I suppose these error types are.

primitives/sr-primitives/src/lib.rs
```diff
@@ -451,6 +436,27 @@ impl From<&'static str> for DispatchError {
...
+///   This doesn't neccessarily mean that the extrinsic is invalid, since it can still be
+///   included in the next block if it has enough spare weight available.
+/// - The sender doesn't have enough funds to pay the transaction inclusion fee. Including such
+///   a transaction in the block doesn't make sense.
+/// - The extrinsic supplied a bad signature. This transaction won't become valid ever.
```
kianenigma: I would make it very clear that a dispatch has two phases 1- *pre-dispatch stuff* and 2- *the dispatch itself*. 

A signature check, and any other code that we might put in `Executive::apply_extrinsic_xxx` is in the first group (`.check`). For instance a signature check gives you `InclusionError`

A weight check, nonce and other things that happen either in signedExtensions or dispatch code itself are in the second group. I think a wrong nonce gives you `Ok(DispatchError)` (while an all okay dispatch gives `Ok(())`). 

So I would make this distinction very clear in the docs. 

(take the above with a grain of slat and double check, I didn't look super deep into the code.)

kianenigma: Based on this, without any explanation, I find it confusing that signature check and weight check are enumerated next to each other, as it seems that they are natively different types of checks.

primitives/sr-primitives/src/lib.rs
```diff
@@ -451,6 +436,27 @@ impl From<&'static str> for DispatchError {
 	}
 }
 
+/// This type specifies the outcome of dispatching a call to a module.
+///
+/// In case of failure an error specific to the module is returned.
```
tomusdrw: I would add a note that even in case of failure the extrinsic is included in the block and all the state changes done in `pre_dispatch` and `post_dispatch` are applied. Namely for instance the sender pays fees.

primitives/sr-primitives/src/lib.rs
```diff
@@ -451,6 +436,27 @@ impl From<&'static str> for DispatchError {
...
+pub type DispatchOutcome = Result<(), DispatchError>;
+
+/// The outcome of inclusion of an extrinsic into a block.
+///
+/// This type is typically used in the context of `BlockBuilder` to signal that the extrinsic
+/// in question cannot be included. It is fair to say that a valid block doesn't contain any
```
tomusdrw: The second sentence is a bit convoluted to me, can't we simply write:

> A block containing extrinsics that have a negative inclusion outcome is invalid. A negative result can only occur in the context of `BlockBuilder`, where such extrinsics are detected and removed from the block that is being created and the transaction pool. To rehash: every extrinsic in a valid block must return a positive `InclusionOutcome`.

primitives/sr-primitives/src/lib.rs
tomusdrw: Yeah, I would redirect the reader from this documentation to how `apply/dispatch + SignedExtensions` work in general to get more details into the lifecycle of extrinsic.

primitives/sr-primitives/src/lib.rs
```diff
@@ -348,56 +348,41 @@ impl From<ed25519::Signature> for AnySignature {
...
-/// Reason why an extrinsic couldn't be applied (i.e. invalid extrinsic).
-pub enum ApplyError {
-	/// General error to do with the permissions of the sender.
-	NoPermission,
-
-	/// General error to do with the state of the system in general.
-	BadState,
-
+/// Reason why an extrinsic couldn't be included into a block.
+pub enum InclusionError {
```
tomusdrw: Since there is only one variant left I wonder if the existence of this type is justified. I think it would be way clearer to return `TransactionValidityError` instead of a wrapper type.

client/block-builder/src/lib.rs
```diff
@@ -43,7 +43,7 @@ use sr_api::{Core, ApiExt, ApiErrorFor};
 pub use runtime_api::BlockBuilder as BlockBuilderApi;
 
 /// Error when the runtime failed to apply an extrinsic.
-pub struct ApplyExtrinsicFailed(pub sr_primitives::ApplyError);
+pub struct ApplyExtrinsicFailed(pub sr_primitives::InclusionError);
```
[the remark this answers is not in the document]
pepyakin: My logic was that I'd leave this as is to denote that `apply_extrinsic` failed because of the following reason which is represented by `InclusionError`.

I thought maybe it would be better to name `apply_extrinsic` as `include_extrinsic`, but that is getting out of hand...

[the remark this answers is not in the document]
pepyakin: That said, I am not strong on it. Give me a sign (thumbs up will do) and I will make the change

primitives/block-builder/runtime-api/src/lib.rs
pepyakin: ehm, doesn't feel difference between these, so will stick with `is`.

primitives/sr-primitives/src/lib.rs
pepyakin: So yeah, I actually left a comment on it somewhere (in the commit msg, I guess? : P) but now you say this I think I see no reason to keep it this way.

Let's do it. However, would you be OK with the fact that we would return `TransactionValidityError` from `apply_extrinsic` (that you [stated here](https://github.com/paritytech/substrate/pull/4134/files#r347827672))?

primitives/sr-primitives/src/lib.rs
pepyakin: Ahhh, that's a good point. I am mistakingly assumed that we are in the pallete context.

primitives/sr-primitives/src/lib.rs
pepyakin: Yeah, they are different checks in practice, but we are dealing with the non-pallete context, from the block builder API perspective, from that PoV they are essentially the same, isn't it This also relates to referencing the user to the documentation of `SignedExtensions` and co.

pepyakin:
So I guess if we are more or less happy with this refactoring, as the next step I'd like to get some directions on how to cleanly integrate this PR into polkadot/kusama.

Apart from updating [the accompanying PR](https://github.com/paritytech/polkadot/pull/596) with updating the authoring version there, what else should we do here? Do you have an idea @bkchr ?

gavofyork:
@pepyakin re: compatibility with kusama, it's more about just ensuring that UI people are given appropriate warning.

conflicts need sorting, but otherwise should be good.

client/block-builder/src/lib.rs
```diff
@@ -107,21 +130,43 @@ where
...
-					extrinsics.push(xt);
-					Ok(())
-				}
-				Err(e) => {
-					Err(ApplyExtrinsicFailed(e))?
-				}
-			}
-		})
+		if self
+            .api
```
bkchr: Everything here are spaces!

pepyakin: yuk!

PR #4136, opened 2019-11-18
add div and checked_div to Fixed64

xlc:


primitives/sr-arithmetic/src/fixed64.rs
```diff
@@ -241,4 +266,25 @@ mod tests {
 		saturating_mul_acc_test!(u64);
 		saturating_mul_acc_test!(u128);
 	}
+
+	#[test]
+	fn div_works() {
+		let a = Fixed64::from_rational(12, 10);
+		let b = Fixed64::from_rational(1, 100);
+		assert_eq!(a / b, Fixed64::from_rational(120, 1));
```
kianenigma: This could cover more cases such as: 

- a / b where a < b && b > 1 
- a / b where a < b && b < 1 (you already have this one)
- a / b where a < b && b > 1 
- a / b where a < b && b < 1 

and maybe detail of panic, if it does, as mentioned in the doc?

xlc: Will add the tests. For panic, I just checked and it won't actually panic on div 0. Should I make it panic just to match the default behaviour?

PR #4139, opened 2019-11-19
Transaction pool notification stream

seunlanlege:
Previously the function `import_notification_stream` returned `Stream<Item=()>` which isn't very useful.

I'm working on an instant-seal consensus implementation for single node enviroments, and for implementation purposes
I need to get the imported transactions over a stream. This PR implements `Clone` for `transaction_pool::base::Transaction<Hash, Extrinsic>`
which allows the transaction to be cloned for each subscriber of the notification stream.

### DIFF
--- a/client/transaction-pool/graph/src/base_pool.rs
+++ b/client/transaction-pool/graph/src/base_pool.rs
@@ -83,8 +83,7 @@ pub struct PruneStatus<Hash, Ex> {
 }
 
 /// Immutable transaction
-#[cfg_attr(test, derive(Clone))]
-#[derive(PartialEq, Eq)]
+#[derive(PartialEq, Eq, Clone)]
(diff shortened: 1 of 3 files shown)


client/transaction-pool/graph/src/base_pool.rs
```diff
@@ -83,8 +83,7 @@ pub struct PruneStatus<Hash, Ex> {
 }
 
 /// Immutable transaction
-#[cfg_attr(test, derive(Clone))]
-#[derive(PartialEq, Eq)]
+#[derive(PartialEq, Eq, Clone)]
```
tomusdrw: The transaction should not be cloneable. It's too heavy to do so frequently. If there is no way to retrieve an `Arc` then please just return `Hash` in the notification stream.

PR #4140, opened 2019-11-19
Fix the status-sinks tests, hopefully

tomaka:
This makes the test less covering, but at least it will pass CI.


gavofyork:
@gnunicorn do we expect polkadot and/or dependency rules to be green?

gnunicorn:
dependency rules, no. Polkadot I am not sure. I thought we had a PR and it was merged. @bkchr would know.

gavofyork:
only for v0.6, i think

PR #4143, opened 2019-11-19
Manual Seal

seunlanlege:
Manual Seal is a block production engine for single-node environments for testing.

It provides two RPC methods `engine_createBlock` and `engine_finalizeBlock` for creating and finalizing blocks respectively.

Alongside manual seal is instant seal, which creates blocks as soon as new transactions are available


kianenigma:
Interesting; is this mostly used for testing then? can you elaborate a bit more?

joshua-mir:
If this is based on what we have on parity-ethereum, yes @kianenigma  - (a `--dev` chain on parity-ethereum uses the nstantSeal engine, which is intended for local-only debugging)

client/consensus/instant-seal/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+		A: txpool::ChainApi + 'static,
+		C: SelectChain<B> + 'static,
+{
+	let block_import = Arc::new(Mutex::new(block_import));
+	let env = Arc::new(Mutex::new(env));
+	let select_chain = Arc::new(select_chain);
+	let inherent_data_providers = Arc::new(inherent_data_providers);
+
+	// propose a new block everytime a transaction is imported
+	pool.import_notification_stream()
```
tomusdrw: Wouldn't that lead to having two blocks proposed in parallel at the same height? Since the code is asynchronous here, we might be spawn two futures by accident.

I think it would be best to make sure blocks are created synchronously only in one `future` and the notification stream just triggers the other task to do the work.

tomusdrw:
Rather than having `InstantSeal` engine I think it would be better to have an engine that is manually controlled. `InstantSeal` can then build on top of this, but also we can just expose the engine functions over RPC. More control allows more sophisticated testing (especially if we build a simple UI on top) and might be super useful in the future to do some simulations.

Imagine if you could take an existing chain, locally swap consensus to "manual" and then check what result your transaction would have, how it changes over time (empty blocks being produced), etc. Using just the same tools that you use already (polkadot apps).

client/consensus/instant-seal/src/lib.rs
seunlanlege: Great catch, yeah I was worried about that initially, but I checked the futures crate to see how it handled `for_each` and it appears that it will poll the future produced by the closure to completion before polling the next item of the stream. 

https://github.com/rust-lang-nursery/futures-rs/blob/c4291883bcc9d3ae8cc3186b2eec27b096dafb66/futures-util/src/stream/stream/for_each.rs#L70-L87

there's also `for_each_concurrent` which will do exactly what you've described.

tomusdrw: Cool. Good point, didn't know about `for_each_concurrent` actually. Anyways, if that's the case we should put a guard inside the closure and never produce a block if there is no pending transactions.
Currently what may happen is:
1. You have 1000 transactions that get imported to the pool concurrently.
2. This means you will get 1000 notifications on the `import_notifications_stream`.
3. Which in turn means that you will produce 1000 blocks
4. But the first block will most likely contain most of the transactions (all transactions that were able to be imported already)
5. Then second block will have the rest
6. And the rest 998 blocks will be empty.


Also it would be good to write some tests for this engine, to make sure we don't run into similar issues as in Parity Ethereum.

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+		block: BlockImportParams<B>,
+		cache: HashMap<CacheKeyId, Vec<u8>>,
+	) -> Result<ImportResult, Self::Error> {
+		// TODO: strip out post-digest.
+
+		self.inner.import_block(block, cache)
+	}
+}
+
+/// The verifier for the manual seal engine; instantly finalizes.
```
tomusdrw: Would probably be good to control finality manually as well.

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! An instant sealing engine: we always author on every slot, and accept all blocks.
```
tomusdrw: needs an update

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+		S: Stream<Item=EngineCommand> + 'static,
+		C: SelectChain<B> + 'static,
+{
+	let block_import = Arc::new(Mutex::new(block_import));
+	let env = Arc::new(Mutex::new(env));
+	let select_chain = Arc::new(select_chain);
+	let inherent_data_providers = Arc::new(inherent_data_providers);
+	let moved_pool = pool.clone();
+
+	seal_block_channel
```
tomusdrw: Can't this just be a loop?

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+		.for_each(move |command| {
+			let select_chain = select_chain.clone();
+			let env = env.clone();
+			let inherent_data_providers = inherent_data_providers.clone();
+			let block_import = block_import.clone();
+			let moved_pool = moved_pool.clone();
+
+			async move {
```
tomusdrw: Why do we need another closure here? Can't we make the upper one `async`?

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,65 @@
...
+use jsonrpc_derive::rpc;
+use futures::channel::mpsc;
+
+/// The "engine" receives these messages over a channel
+pub enum EngineCommand {
+	/// Tells the engine to propose a new block
+	///
+	/// if force == true, it will create empty blocks.
+	SealNewBlock {
+		force: bool
```
tomusdrw: Would rename `force` to `createEmpty`

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,65 @@
...
+	},
+	/// Tells the engine to create a fork
+	///
+	/// TODO: implement CreateFork message handling
+	CreateFork
```
tomusdrw: Actually `CreateFork` is the same as `SealNewBlock`, but with additional `parent_hash` parameter. I believe we should just merge the two and have `Optional<Hash>`, where `None` means to build a block on top of the latest one.

client/consensus/manual-seal/src/lib.rs
seunlanlege: how would this work?

client/consensus/manual-seal/src/lib.rs
seunlanlege: The RPC sends the block authorship task commands over a `futures::mpsc`, also with the way the api is designed, I've implemented an instant_seal authorship task.


https://github.com/paritytech/substrate/pull/4143/files#diff-4cecb2a4202c6cedf89a9417017c3ac0R215-R250

client/consensus/manual-seal/src/lib.rs
seunlanlege: rust's move semantics, the async block would move the variables and they'd be unavailable for the next call to `for_each`. would work if `for_each` took an `FnOnce`

client/consensus/manual-seal/src/lib.rs
tomusdrw: via RPC I suppose, so basically you mark the block as final whenever you like. With instant finality there is no way to revert the latest state or build a fork.

client/consensus/manual-seal/src/lib.rs
tomusdrw: I mean instead of `for_each` there should be a way to write a loop that consumes elements of the stream using `async/await`, no?

client/consensus/manual-seal/src/lib.rs
tomusdrw:
```
.for_each(async move |command| {
            // <cut, since not needed>
			//async move {
           ...
})
```
Wouldn't that work? Why have a regular closure that returns an `async` block if you can have an `async` closure instead?

seunlanlege: async closures aren't finalized yet

client/consensus/manual-seal/src/lib.rs
seunlanlege: I can look into this

client/transaction-pool/src/lib.rs
```diff
@@ -25,6 +25,8 @@ mod maintainer;
 pub mod error;
 #[cfg(test)]
 mod tests;
+#[cfg(feature = "test")]
+pub mod test_helpers;
```
seunlanlege: @tomusdrw needed to re-use some code from the transaction pool tests, not sure if this is the right approach,

tomusdrw: I'd call the module `mod testing`, but the feature `test-helpers`. I think this convention is more widespread in substrate. But looks good apart from that.

seunlanlege:
@tomusdrw :wave:  this is ready

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+	{
+		self.inner.check_block(block)
+	}
+
+	fn import_block(
+		&mut self,
+		block: BlockImportParams<B>,
+		cache: HashMap<CacheKeyId, Vec<u8>>,
+	) -> Result<ImportResult, Self::Error> {
+		// TODO: strip out post-digest.
```
rphmeier: no TODOs

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+{
+	let pool = basic_pool.pool();
+
+	while let Some(command) = seal_block_channel.next().await {
+		match command {
+			EngineCommand::SealNewBlock {
+				create_empty,
+				parent_hash,
+				sender
+			} => {
```
rphmeier: refactor: wouldn't it be better to have a separate function for each of these blocks?

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+					}
+					Ok(other) => rpc::send_result(sender, Err(other.into())),
+					Err(e) => {
+						log::warn!("Failed to import block: {:?}", e);
+						rpc::send_result(sender, Err(e.into()))
+					}
+				};
+			}
+			EngineCommand::FinalizeBlock { hash, sender } => {
+				// TODO(seun): Justification support?
```
rphmeier: what justification? also TODO needs to be resolved

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+						log::info!("Successfully finalized block: {}", hash);
+						rpc::send_result(sender, Ok(()))
+					}
+				}
+			}
+		}
+	}
+}
+
+pub async fn run_instant_seal<B, CB, H, E, A, C>(
```
rphmeier: needs docs!

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,191 @@
...
+	},
+	/// Tells the engine to finalize the block with the supplied hash
+	FinalizeBlock {
+		hash: Hash,
+		sender: Sender<()>,
+	}
+}
+
+#[rpc]
+pub trait ManualSealApi<Hash> {
```
rphmeier: needs docs

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,191 @@
...
+
+/// A struct that implements the [`ManualSealApi`].
+pub struct ManualSeal<Hash> {
+	import_block_channel: mpsc::Sender<EngineCommand<Hash>>,
+}
+
+/// return type of `engine_createBlock`
+#[derive(Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct CreatedBlock<Hash> {
+	pub hash: Hash,
```
rphmeier: docs on fields

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,191 @@
...
+			let result = sink.send(
+				EngineCommand::SealNewBlock {
+					create_empty,
+					parent_hash,
+					sender: Some(sender),
+				}
+			).await;
+			if let Err(e) = result {
+				return Err(map_error(e))
+			};
```
rphmeier: could this be rewritten with `map_err` and `?` ?

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,191 @@
...
+
+			match receiver.await {
+				// all good
+				Ok(Ok(block)) => Ok(block),
+				// error from the authorship task
+				Ok(Err(e)) => {
+					Err(Error {
+						code: ErrorCode::ServerError(500),
+						message: format!("{}", e),
+						data: None
```
rphmeier:
```suggestion
						data: None,
```

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,191 @@
...
+					})
+				}
+				// channel has been dropped
+				Err(_) => {
+					Err(Error {
+						code: ErrorCode::ServerError(500),
+						message: "Server is shutting down".into(),
+						data: None
```
rphmeier:
```suggestion
						data: None,
```

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,191 @@
...
+						code: ErrorCode::ServerError(500),
+						message: format!("{}", e),
+						data: None
+					})
+				}
+				// channel has been dropped
+				Err(_) => {
+					Err(Error {
+						code: ErrorCode::ServerError(500),
+						message: "Server is shutting down".into(),
```
rphmeier: Where did you draw this error message and code from?

Maybe we can rephrase to "Consensus process terminating", if the only reason that this can happen is that the `manual_seal` future is terminating.

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+		let header = backend.blockchain().header(BlockId::Number(1)).unwrap().unwrap();
+		let (tx, rx) = futures::channel::oneshot::channel();
+		sink.send(EngineCommand::FinalizeBlock {
+			sender: Some(tx),
+			hash: header.hash()
+		}).await.unwrap();
+		// assert that the background task returns ok
+		assert_eq!(rx.await.unwrap().unwrap(), ());
+	}
+	// TODO: test with fork blocks
```
rphmeier: remove TODO!

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+			created_block,
+			CreatedBlock {
+				hash: created_block.hash.clone(),
+				aux: ImportedAux {
+					header_only: false,
+					clear_justification_requests: false,
+					needs_justification: false,
+					bad_justification: false,
+					needs_finality_proof: false,
+					is_new_best: true
```
rphmeier:
```suggestion
					is_new_best: true,
```

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
```
rphmeier:
```suggestion
					is_new_best: true,
```

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+		C: SelectChain<B> + 'static,
+{
+	let pool = basic_pool.pool();
+
+	while let Some(command) = seal_block_channel.next().await {
+		match command {
+			EngineCommand::SealNewBlock {
+				create_empty,
+				parent_hash,
+				sender
```
rphmeier:
```suggestion
				sender,
```

client/consensus/manual-seal/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+				let (params, header) = match result {
+					Ok(block) => {
+						let (header, body) = block.deconstruct();
+						(BlockImportParams {
+							origin: BlockOrigin::Own,
+							header: header.clone(),
+							justification: None,
+							post_digests: Vec::new(),
+							body: Some(body),
+							finalized: false,
```
rphmeier: I feel like this should be an option from the `SealNewBlock` message. Otherwise, you would always have to send 2 messages to get an instantly-finalized block.

rphmeier:
Thanks! Generally looks good. Cool to see a first example of RPCs being generated directly from a consensus crate.

client/consensus/manual-seal/src/lib.rs
seunlanlege: makes sense.

client/consensus/manual-seal/src/lib.rs
seunlanlege: actually this was originally your TODO, :sweat_smile:

rphmeier: i thought that it was :) but I wouldn't mark a PR as "pleasereview" with an active TODO.

seunlanlege: Yeah, I understand, will keep that in mind for next time. 

> actually this was originally your TODO

and so what I meant by this is, I have no idea how to resolve this, I would assume to empty the post-digests `Vec` or are there any other additional operations required?

rphmeier: If you haven't added a post-digest anywhere during the sealing process then the TODO should just be removed

client/consensus/manual-seal/src/finalize_block.rs
```diff
@@ -0,0 +1,64 @@
...
+	generic::BlockId,
+};
+use std::{
+	sync::Arc,
+	marker::PhantomData
+};
+use hash_db::Hasher;
+use client_api::backend::Backend as ClientBackend;
+
+pub struct FinalizeBlockParams<B: BlockT, H, CB> {
```
rphmeier: missing docs on public items. This is a blanket rule in the Substrate and Polkadot codebases, and I hope I don't have to leave any more comments like this in the future.

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,200 @@
...
+	TryFutureExt,
+	SinkExt
+};
+use serde::{Deserialize, Serialize};
+use sp_runtime::Justification;
+
+type FutureResult<T> = Box<dyn jsonrpc_core::futures::Future<Item = T, Error = Error> + Send>;
+pub type Sender<T> = Option<oneshot::Sender<std::result::Result<T, crate::Error>>>;
+
+/// message sent by rpc to the background authorship task
```
rphmeier:
```suggestion
/// Message sent to the background authorship task, usually by RPC.
```

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,200 @@
...
+		oneshot,
+	},
+	TryFutureExt,
+	SinkExt
+};
+use serde::{Deserialize, Serialize};
+use sp_runtime::Justification;
+
+type FutureResult<T> = Box<dyn jsonrpc_core::futures::Future<Item = T, Error = Error> + Send>;
+pub type Sender<T> = Option<oneshot::Sender<std::result::Result<T, crate::Error>>>;
```
rphmeier: missing docs on this and a couple other items in the module

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,212 @@
...
+
+/// report any errors or successes encountered by the authorship task back
+/// to the rpc
+pub fn send_result<T: std::fmt::Debug>(
+	sender: Sender<T>,
+	result: std::result::Result<T, crate::Error>
+) {
+	sender.map(|sender| {
+		sender.send(result)
+			.expect("receiving end isn't dropped until it receives a message; qed")
```
tomusdrw: I guess it might be when the node is terminating, best to just print an error here instead of panicking imho.

client/consensus/manual-seal/src/seal_new_block.rs
```diff
@@ -0,0 +1,173 @@
...
+	/// Environment trait object for creating a proposer
+	pub env: &'a mut E,
+	/// SelectChain object
+	pub select_chain: &'a C,
+	/// block import object
+	pub block_import: &'a mut BoxBlockImport<B>,
+	/// inherent data provide
+	pub inherent_data_provider: &'a inherents::InherentDataProviders,
+	/// phantom data to pin the hasher type.
+	pub _phantom: PhantomData<H>
```
tomusdrw: Same here, I don't think the `H` type is useful anywhere and it doesn;'t need to be bound to the params.

client/consensus/manual-seal/src/seal_new_block.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+/// seals a new block with the given params
+pub async fn seal_new_block<B, C, CB, E, P, H>(params: SealBlockParams<'_, B, C, CB, E, P, H>)
```
tomusdrw:
```suggestion
pub async fn seal_new_block<B, C, CB, E, P, H>(params: SealBlockParams<'_, B, C, CB, E, P>)
```

client/consensus/manual-seal/src/seal_new_block.rs
```diff
@@ -0,0 +1,173 @@
...
+		sender,
+		back_end,
+		select_chain,
+		block_import,
+		env,
+		inherent_data_provider,
+		..
+	} = params;
+
+	if pool.status().ready == 0 && !create_empty {
```
tomusdrw: This is subject to race conditions. The pool might not be empty now, but it will be empty when we start creating the block.
I think I'd go through the whole block production and then reject a block if it's actually empty, then risk unpredictable behavior here.

client/consensus/manual-seal/src/seal_new_block.rs
```diff
@@ -0,0 +1,173 @@
...
+					match header {
+						Some(header) => Ok(header),
+						None => Err(Error::BlockNotFound)
+					}
+				})
+		}
+		None => select_chain.best_chain().map_err(Error::from)
+	};
+
+	let header = match header {
```
tomusdrw: The error handling here is tedious, can't we use `map_err` and `?` operator and call` rpc::send_result` in single place?

client/transaction-pool/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2018"
 [dependencies]
 codec = { package = "parity-scale-codec", version = "1.0.0" }
 derive_more = "0.99.2"
-futures = { version = "0.3.1", features = ["compat"] }
+futures = { version = "0.3.1", features = ["compat", "compat"] }
```
tomusdrw: Such `compat`, wow.

```suggestion
futures = { version = "0.3.1", features = ["compat"] }
```

client/transaction-pool/Cargo.toml
```diff
@@ -17,7 +17,10 @@ sc-transaction-graph = { path = "./graph" }
 sp-transaction-pool = { path = "../../primitives/transaction-pool" }
 sc-client-api = { path = "../api" }
 sp-blockchain = { path = "../../primitives/blockchain" }
+test-client = { package = "substrate-test-runtime-client", path = "../../test-utils/runtime/client" }
```
tomusdrw: No renames here either.

client/transaction-pool/Cargo.toml
```diff
@@ -17,7 +17,10 @@ sc-transaction-graph = { path = "./graph" }
...
 
 [dev-dependencies]
-sp-keyring = { path = "../../primitives/keyring" }
-substrate-test-runtime-client = { path = "../../test-utils/runtime/client" }
+keyring = { package = "sp-keyring", path = "../../primitives/keyring" }
```
tomusdrw: Nor here.

client/consensus/manual-seal/src/seal_new_block.rs
seunlanlege: > The pool might not be empty now, but it will be empty when we start creating the block.

Hmm, I don't see how this might be possible, can you elaborate?

client/consensus/manual-seal/src/seal_new_block.rs
seunlanlege: that would be possible if `seal_new_block` returned a `Result`

client/consensus/manual-seal/src/seal_new_block.rs
tomusdrw: You can have multiple concurrent tasks running at `proposer.propose` since they are asynchronous. Also transaction pool culling happens asynchronously, so you can assume that the transaction you see in the pool right now are going to be there when you actually get pending iterator. The pool is designed to be accessed from multiple threads (or concurrent tasks), so there are race conditions.

I'd leave this check here as an optimisation, but then would double check if produced block has no transactions before importing.

tomusdrw: Also that got me thinking - what will happen if you trigger creation of two blocks at the same time (without providing `parent` parameter). I suspect that instead of producing 2 blocks on the canon chain it would produce two blocks at the same height, which most likely is not what we want.

client/consensus/manual-seal/src/seal_new_block.rs
tomusdrw: It's possible now too, you just need to introduce an extra function:
```
fn seal_new_block(...) -> () {
  fn helper_function(..) -> Result<(), _> {
     ... // current logic using `.map_err` and `?`
  } 

  if let Err(e) = helper_function(...) {
     rpc::send_result(e);
  }  
}

File: client/consensus/manual-seal/src/seal_new_block.rs (hunk 30 above)
Comment (seunlanlege) [reply client/consensus/manual-seal/src/seal_new_block.rs]: > what will happen if you trigger creation of two blocks at the same time (without providing parent parameter).

The authorship task processes `EngineCommands` synchronously not concurrently, so one will wait for the other.

>  but then would double check if produced block has no transactions before importing.

yeah this would ideal.

**Comment by seunlanlege** [comment]
SYN @tomusdrw

File: client/consensus/manual-seal/src/error.rs [hunk 35]
```diff
@@ -0,0 +1,53 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
```
tomusdrw: I guess the dates in all files now need an update

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,212 @@
...
+			).await;
+			result.map_err(map_error)?;
+
+			match receiver.await {
+				// all good
+				Ok(Ok(block)) => Ok(block),
+				// error from the authorship task
+				Ok(Err(e)) => {
+					Err(Error {
+						code: ErrorCode::ServerError(500),
```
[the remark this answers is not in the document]
tomusdrw: Still applies?

client/consensus/manual-seal/src/rpc.rs
```diff
@@ -0,0 +1,220 @@
...
+	}
+}
+
+/// convert a futures::mpsc::SendError to jsonrpc::Error
+fn map_error(error: SendError) -> Error {
+	if error.is_disconnected() {
+		log::warn!("Received sealing request but Manual Sealing task has been dropped");
+	}
+
+	Error {
```
tomusdrw: This is being repeated multiple times. Can you instead introduce an `enum` with different errors and just implement `From<Error> for jsonrpc_core::Error` to do the conversion in a single place instead? Like:
```rust
enum Error {
  ConsensusTerminating,
  ...
}
impl From<Error> for jsonrpc_core::Error {
 fn from(e: Error) -> Self {
   match e {
     Error::ConsensusTerminating => jsonrpc_core::Error {
        code: ...,
        ...
     } 
   }
}
```

client/consensus/manual-seal/src/seal_new_block.rs
```diff
@@ -0,0 +1,165 @@
...
+				rpc::send_result(&mut sender, Err(err.into()))
+			})?;
+
+		let block = proposer.propose(id, Default::default(), Duration::from_secs(5))
+			.await
+			.map_err(|err| {
+				rpc::send_result(&mut sender, Err(Error::ProposerError(format!("{}", err))))
+			})?;
+
+		if block.extrinsics().len() == 0 && !create_empty {
```
tomusdrw: `.is_empty()` instead of `.len() == 0`?

client/consensus/manual-seal/src/seal_new_block.rs
```diff
@@ -0,0 +1,165 @@
...
+		};
+
+		match block_import.import_block(params, HashMap::new()) {
+			Ok(ImportResult::Imported(aux)) => {
+				rpc::send_result(&mut sender, Ok(CreatedBlock { hash: <B as BlockT>::Header::hash(&header), aux }))
+			}
+			Ok(other) => rpc::send_result(&mut sender, Err(other.into())),
+			Err(e) => {
+				log::warn!("Failed to import block: {:?}", e);
+				rpc::send_result(&mut sender, Err(e.into()))
```
tomusdrw: It's cool that now we have one line to send the result, but isn't it really possible to extract the logic and just have single place were we actually call that method? Doing `map_err` all over this function seems quite redundant.

client/consensus/manual-seal/src/error.rs
rphmeier: all the new files at least :)

client/consensus/manual-seal/src/seal_new_block.rs
```diff
@@ -0,0 +1,149 @@
...
+				Error::StringError(format!("{}", err))
+			}).await?;
+		let id = inherent_data_provider.create_inherent_data()?;
+		let proposal = proposer.propose(id, Default::default(), Duration::from_secs(5), false.into())
+			.map_err(|err| {
+				// <<<E as Environment<B>::Proposer> as Proposer<B>>::Error: fmt::Display
+				Error::StringError(format!("{}", err))
+			}).await?;
+
+		if proposal.block.extrinsics().is_empty() && !create_empty {
```
tomusdrw: I don't think it will ever be the case. `proposal.block.extrinsics()` will always contain inherent data. Can we compare that with the len of `inherent_data` that we pass to `proposer`?

client/consensus/manual-seal/src/seal_new_block.rs
```diff
@@ -0,0 +1,149 @@
...
+
+		let mut proposer = env.init(&header)
+			.map_err(|err| {
+				// <E as Environment<B>::Error>: fmt::Display
+				Error::StringError(format!("{}", err))
+			}).await?;
+		let id = inherent_data_provider.create_inherent_data()?;
+		let proposal = proposer.propose(id, Default::default(), Duration::from_secs(5), false.into())
+			.map_err(|err| {
+				// <<<E as Environment<B>::Proposer> as Proposer<B>>::Error: fmt::Display
```
tomusdrw:
```suggestion
```
Same here.

client/consensus/manual-seal/src/seal_new_block.rs
```diff
@@ -0,0 +1,149 @@
...
+				match back_end.blockchain().header(BlockId::Hash(hash))? {
+					Some(header) => header,
+					None => return Err(Error::BlockNotFound(format!("{}", hash))),
+				}
+			}
+			None => select_chain.best_chain()?
+		};
+
+		let mut proposer = env.init(&header)
+			.map_err(|err| {
```
tomusdrw: I'd remove the comments and keep it in a single line, cause imho the comment will get out of date easily.
```suggestion
			.map_err(|err| Error::StringError(format("{}", err))
```

client/consensus/manual-seal/src/seal_new_block.rs
```diff
@@ -0,0 +1,149 @@
...
+				// <E as Environment<B>::Error>: fmt::Display
+				Error::StringError(format!("{}", err))
+			}).await?;
+		let id = inherent_data_provider.create_inherent_data()?;
+		let proposal = proposer.propose(id, Default::default(), Duration::from_secs(5), false.into())
```
tomusdrw: Where does the `from_secs(5)` come from? Maybe move it to a constant if that's an upper limit that we want to set for the manual seal.

seunlanlege:
@tomusdrw done, :sweat_smile:

xlc:
Is this been integrated anywhere?

PR #4144, opened 2019-11-19
Upgrade trie-db crate to 0.16.0.

jimpo:
The new version of `trie-db` has some breaking API changes, but the node encoding format remains the same.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6420,7 +6420,7 @@ dependencies = [
  "parity-scale-codec 1.0.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "sr-std 2.0.0",
  "substrate-primitives 2.0.0",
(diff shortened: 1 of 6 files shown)


Cargo.lock
```diff
@@ -6960,7 +6972,7 @@ name = "twox-hash"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "rand 0.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rand 0.3.23 (registry+https://github.com/rust-lang/crates.io-index)",
```
cheme: A bit unexpected version change in the dep.

cheme:
I guess trie-bench crate also need an update to new trie-db crate.

Cargo.lock
niklasad1: I would guess that this is because of https://github.com/rust-lang/cargo/issues/5702

PR #4145, opened 2019-11-19
Offchain execution extensions

tomusdrw:
This PR is an attempt to clean up how we deal with externalities extensions for offchain calls (i.e. the additional APIs available for some contexts), namely: `OffchainWorker` extensions, `TransactionPool` and `Keystore`.

Instead of passing things separately this PR introduces a `ExecutionExtensions` idea, where all `ExternalitiesExtensions` can be registered (avoiding cycles though). Later on in the call stack we pass `Extensions` instead of arbitrary `Option<KeystorePtr>` or `Option<OffchainExt>` stuff.
This cleans up some APIs and also avoids uneccesary direct dependencies.

CC @andresilva - finally :)


primitives/state-machine/src/lib.rs
```diff
@@ -497,8 +487,12 @@ where
 	Exec: CodeExecutor,
 {
 	let proving_backend = proving_backend::ProvingBackend::new(trie_backend);
+	let mut extensions = Extensions::new();
+	if let Some(keystore) = keystore {
```
tomusdrw: Do we really need `keystore` here? If that can be invoked from the network context (i.e. light client requesting to prove some execution) we can end up signing something with a local key and giving it back as a response.

primitives/state-machine/src/lib.rs
```diff
@@ -542,8 +536,12 @@ where
 	H: Hasher<Out=H256>,
 	Exec: CodeExecutor,
 {
+	let mut extensions = Extensions::new();
+	if let Some(keystore) = keystore {
+		extensions.register(keystore);
```
tomusdrw: Same here.

primitives/sr-io/with_std.rs
```diff
@@ -347,18 +349,22 @@ fn with_offchain<R>(f: impl FnOnce(&mut dyn offchain::Externalities) -> R, msg:
 }
 
 impl OffchainApi for () {
+	fn submit_transaction(data: Vec<u8>) -> Result<(), ()> {
```
bkchr: This file is not used anymore. It was probably readded in your big refactoring pr.

https://github.com/paritytech/substrate/pull/4130

client/src/light/call_executor.rs
```diff
@@ -111,9 +112,8 @@ impl<Block, B, Local> CallExecutor<Block, Blake2Hasher> for
 		initialize_block: InitializeBlock<'a, Block>,
 		_manager: ExecutionManager<EM>,
 		native_call: Option<NC>,
-		side_effects_handler: Option<OffchainExt>,
 		recorder: &Option<ProofRecorder<Block>>,
-		enable_keystore: bool,
+        extensions: Option<Extensions>,
```
andresilva: indented with spaces.

client/api/src/lib.rs
```diff
@@ -15,25 +15,25 @@
...
 pub mod call_executor;
 pub mod client;
-pub mod offchain;
+pub mod error;
+pub mod execution_extensions;
+pub mod light;
+pub mod notifications;
 
-pub use error::*;
+// TODO: avoid re-exports
```
andresilva: Should this have been handled?

PR #4146, opened 2019-11-19
Fix serialization of `transaction_paymentInfo`

tomusdrw:
We serialize to `serde_json::Value` internally in `jsonrpc` and without `arbitrary_precision` feature the `u128` (used as `Balance`) is not support and causes a failure.

Also fixes the format of returned response to be more javascripty.

CC @jacogr


kigawas:
Looks like polkadot.js stuck at connecting websocket :(

Issue #3921, opened 2019-10-26
Transaction Weight mechanism not considering proposals


xlc:
Disaptchable calls can be dispatched indirectly (and many of them are designed to be dispatched indirectly, such as `system::set_code`). The transaction weight for those indirects calls are not considered at all right now.

e.g. `collective::vote` is `FixedOperational(200_000)` and may dispatch an arbitrary proposal, such as `membership::add_member` which is `FixedNormal(50_000)`. The final weight for such transaction should be 250_000 instead of ignoring the weight of `membership::add_member`.

Related: #3730 can make dynamically increase weight possible when the proposal is dispatched.

kianenigma:
related to https://github.com/paritytech/substrate/issues/3730

also @sorpaas I think you will need the same thing for EVM, right?

sorpaas:
Yeah I think it's related to EVM module as well.

xlc:
I think this is fixed already.

Issue #3730, opened 2019-10-01
Proposal: Allow Dynamic Weights


kianenigma:
Regards to [this](https://github.com/paritytech/substrate/pull/3729/files#r329874645) conversation about weights and this part of it:

> and calculating it would take about as long as actually doing it

I was thinking maybe, maybe, we can allow this with a special type of weight (e.g. `#[weight = Dynamic(Foo)]`) which is allowed to be updated after the call has been dispatched. The requirement comes from what I have been discussing with @pepyakin recently about how he can use the weight system for his contract module and at least for his situation: a dispatch's operations can be separated into two parts:
- Fixed: preparing contract, gas metering primitives etc. equivalent of calling an `noop` contract.
- Dynamic: dependent directly on some input or internal state which cannot be predicted without actually dispatching the call first. 

I was thinking we could make it a pattern of passing some data from the `Ok(())` of the `.dispatch()` and then pass it into `post_dispatch` to make updates current block weight, charge more/less fee etc. `.post_dispatch` would automatically call `Foo::update_weight()` with some parameter `T` returned from `.dispatch()`.

I am not sure yet if this is general enough to be considered in substrate, or if we want to stick to the current simple approach of _weight must be calculated before the dispatch_, so this is to discuss this before any prototyping takes place.

xlc:
Any chain use tx fee as a security mechanism and have dispatch methods that are not `O(1)` will need this ability.
One tricky thing is how to (re)charge for the additional amount of the tx fee. It can either over-charge upfront (like max gas fee) and refund the unused part, or simply charge additional fee per increase weight (this means weight can never decrease, which is fine).

kianenigma:
> Any chain use tx fee as a security mechanism and have dispatch methods that are not O(1) will need this ability.

Yeah -- now the approach is that if a particular transaction can cost more, it should handle it manually somehow.

> One tricky thing is how to (re)charge for the additional amount of the tx fee. It can either over-charge upfront (like max gas fee) and refund the unused part, or simply charge additional fee per increase weight (this means weight can never decrease, which is fine).

Yeah indeed, this is exactly how the smart contract was discussed to work with the proposal of this issue.

kianenigma:
We have been talking quite a lot with @pepyakin off band about this and the gist of it, at least from my PoV is:

#### Primitives

[...]

PR #4147, opened 2019-11-19
pallet-contracts: Migrate to weights mechanism

pepyakin:
This PR changes srml-contracts to tap on the weight system instead of relying on a separate gas mechanism. 

In this change we treat 1 gas unit as 1 weight unit. I wasn't sure about getting rid of the notion of gas althogether since we might want to scale it relative to weight later on and because gas is familiar lingo for the existing smart-contract developers. Apart from that, it would also introduce some backward compatibility fall out.

This change also makes an assumption that the weight to price is a linear function.

The transition is not complete though, because there are basically two problems.

# Dynamic Weights

It is not yet clear how to implement `ext_dispatch_call`, a contract-to-runtime call, within the weights system. The problem here is that `ext_dispatch_call(T::Call)` has no information regarding how much weight the given `T::Call` would consume and thus there is no way to ensure that executing this call won't be out of given gas/weight limit budget. Related issues were filled and partially implemented (https://github.com/paritytech/substrate/issues/3730, https://github.com/paritytech/substrate/issues/3921), but it is still not enough. We will work this out with @kianenigma.

# Block weight limit

The current values for the weight budget are a bit off. To understand the issue let's take a simple example where a contract transfers some funds to another account. The situation right now is that the fees for transfers are charged with gas, trying to emulate the behavior of the balances module. The transfer fee in balances module is defined in terms of balance units. We use approximation of the gas cost (we can do that since we know the price of one unit of gas in terms of balance units) for the fees and charge that. At the moment of writing, the fee is set to [1 "cent"](https://github.com/paritytech/substrate/blob/ae9c6eac90606c7e612df52d37e89ff1e6d9203b/bin/node/runtime/src/lib.rs#L406-L407). 

The problem is that with current values set in node-runtime the whole weight limit costs 1 cent, so it is not possible to perform even one transfer.

There are different cures to that to name a few:

1. We don't need to have fees for transfers and it is a remnant of pre-weights system. We just need to remove it.
1. We can just tune the fees themselves. This might allow a weird loophole where it makes sense to transfer balances via contracts rather than specialized balances.

I opted-in to the last one, since merging of https://github.com/paritytech/substrate/pull/4590 will allow us to basically use `T::Currency::transfer` directly without involving any emulation!


frame/contracts/src/lib.rs
```diff
@@ -1035,40 +1101,55 @@ impl<T: Trait + Send + Sync> SignedExtension for CheckBlockGasLimit<T> {
...
-			Call::__PhantomItem(_, _)  => unreachable!("Variant is never constructed"),
+	fn post_dispatch(pre: Self::Pre, _info: DispatchInfo, _len: usize) {
+		if let Some(DynamicWeightData {
+			transactor,
+			imbalance,
+		}) = pre {
+			let (gas_left, gas_spent) = GasUsageReport::take();
+
+			// These should be OK since we don't buy more
+			let unused_weight = gas_left as Weight;
```
jimpo: The gas/weight conversion is kind of clunky. :-/

frame/contracts/src/lib.rs
```diff
@@ -1035,40 +1101,55 @@ impl<T: Trait + Send + Sync> SignedExtension for CheckBlockGasLimit<T> {
...
+
+	fn pre_dispatch(
+		self,
+		who: &Self::AccountId,
+		call: &Self::Call,
+		_: DispatchInfo,
+		_: usize,
+	) -> Result<Self::Pre, TransactionValidityError> {
+		Self::perform_pre_dispatch_checks(who, call)
+			.map_err(Into::into)
```
jimpo: Is the `map_err` necessary? It looks like the error type is the same.

frame/contracts/src/lib.rs
```diff
@@ -1035,40 +1101,55 @@ impl<T: Trait + Send + Sync> SignedExtension for CheckBlockGasLimit<T> {
...
+		who: &Self::AccountId,
 		call: &Self::Call,
 		_: Self::DispatchInfo,
 		_: usize,
 	) -> TransactionValidity {
-		let call = match call.is_sub_type() {
-			Some(call) => call,
-			None => return Ok(ValidTransaction::default()),
-		};
+		Self::perform_pre_dispatch_checks(who, call)
```
jimpo: Hmm, wouldn't this do the actual withdrawal instead of just checking `ensure_can_withdraw`?

frame/contracts/src/lib.rs
```diff
@@ -1035,40 +1101,55 @@ impl<T: Trait + Send + Sync> SignedExtension for CheckBlockGasLimit<T> {
...
+			transactor,
+			imbalance,
+		}) = pre {
+			let (gas_left, gas_spent) = GasUsageReport::take();
+
+			// These should be OK since we don't buy more
+			let unused_weight = gas_left as Weight;
+			let spent_weight = gas_spent as Weight;
+
+			let refund = T::WeightToFee::convert(unused_weight);
```
jimpo: I don't think this is right with a non-linear weight model. Say `L` is the weight limit, `S` is the amount spent, and `U` is unused. Then clearly `L = S + U`. But now say you have a quadratic weight function `fee(weight)`. Then you first withdraw `fee(L)` then deposit back `fee(U)`. But really you wanted to just withdraw `fee(S)`. But `fee(S) = fee(L) - fee(U)` only if `fee` is linear.

Instead, the refund should just be `imbalance - T::WeightToFee::convert(spent_weight)`.

frame/contracts/src/lib.rs
```diff
@@ -867,8 +858,10 @@ decl_event! {
 
 decl_storage! {
 	trait Store for Module<T: Trait> as Contract {
-		/// Gas spent so far in this block.
-		GasSpent get(fn gas_spent): Gas;
+		/// The amount of gas left from the execution of the latest contract.
+		///
+		/// This value is transient and removed before block finalization.
```
jimpo: It's removed sooner, right? We can say it is removed immediately after runtime call dispatch (by the signed extension).

frame/contracts/src/lib.rs
```diff
@@ -1008,19 +999,94 @@ impl Default for Schedule {
...
+					// might attempt to include this transaction again.
+					return Err(InvalidTransaction::ExhaustsResources.into());
+				}
+
+				// Compute the fee corresponding for the given gas_weight_limit and attempt
+				// withdrawing from the origin of this transaction.
+				let fee = T::WeightToFee::convert(gas_weight_limit);
+
+				// Compute and store the effective price per unit of gas.
+				let gas_price = {
```
jimpo: This seems problematic, as it is assuming a linear weight-to-fee conversion, which may be not be the case.

frame/contracts/src/lib.rs
```diff
@@ -725,11 +714,13 @@ impl<T: Trait> Module<T> {
 			DirectAccountDb.commit(ctx.overlay.into_change_set());
 		}
 
-		// Refund cost of the unused gas.
+		// Save the gas usage report.
 		//
 		// NOTE: This should go after the commit to the storage, since the storage changes
```
jimpo: This `NOTE` is not relevant anymore, right?

frame/contracts/src/lib.rs
```diff
@@ -542,21 +541,16 @@ decl_module! {
...
-		) -> Result {
-			let origin = ensure_signed(origin)?;
-
-			let (mut gas_meter, imbalance) = gas::buy_gas::<T>(&origin, gas_limit)?;
+		) -> dispatch::Result {
+			let _origin = ensure_signed(origin)?;
 
 			let schedule = <Module<T>>::current_schedule();
-			let result = wasm::save_code::<T>(code, &mut gas_meter, &schedule);
+			let result = wasm::save_code::<T>(code, &schedule);
```
jimpo: Why does `put_code` no longer require gas? Shoudn't it at least be priced higher than regular runtime call data since it's being put into storage?

pepyakin: This is a good question.

TBH, I am not sure why it got gas in the first place. If you think about it, `put_code` is similar to a regular dispatchable function, although yeah with a rather pricy transaction byte fee. I will change the price down the line.

frame/contracts/src/lib.rs
pepyakin: Indeed, not anymore! Will address that

frame/contracts/src/lib.rs
pepyakin: It will, but I assumed that's alright since we are in validation context therefore all changes are discarded. Or you see some other problem?

frame/contracts/src/lib.rs
pepyakin: Yeah, this is a remnant of testing different approaches, namely trying weight as u32 and u64. This will be definitely addressed down the line

frame/contracts/src/lib.rs
pepyakin: Good point, for now there is assumption that weight is linear. As you've noticed this is not the only place where the code assumes that.

I think there is more to it than just this and we might need to change the weight mechanism a bit. That said, I think we might just proceed with this assumption, documenting that the contracts module is compatible with linear weight model at the moment)

Thanks for the idea of fix though, might be useful

bin/node/runtime/src/lib.rs
```diff
@@ -391,11 +391,21 @@ impl pallet_treasury::Trait for Runtime {
...
+	// transfer.
+	//
+	// This basically means that in order to transfer funds it would make more sense to use the
+	// contracts module rather than balances module.
+	//
+	// This should be fixed in nearest future by using the synchronous transfer directly. It will
+	// cut fees from the senders account balance directly.
+	pub const ContractTransferFee: Balance = 1000;
+	pub const ContractCreationFee: Balance = 1000;
+	pub const ContractFee: Balance = 1000;
```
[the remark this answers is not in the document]
pepyakin: Regarding static assertions I am not sure. They are established by the downstream module and actually depend on the environment the pallet is embedded into.

Regarding the costs my rough plan is to provide an integration guide that would dive into things like this. Not these values though, since I expect them to be gone as soon as we migrate to transactional storage.

frame/contracts/src/lib.rs
```diff
@@ -575,21 +574,16 @@ decl_module! {
 		/// You can instantiate contracts only with stored code.
 		pub fn put_code(
```
kianenigma: >  although yeah with a rather pricy transaction byte fee. I will change the price down the line.

You could start experimenting with `FunctionOf()` here and make this guys weight a function of the code length, if that is what you are looking for here.

frame/contracts/src/lib.rs
kianenigma: Question about such transient values: 

Since they are never stored on chain, hence never used in consensus and state root, why do we even bother storing them on the same trie which imposes all the hashing overhead? Or are they not imposing any overhead in such uses?

frame/contracts/src/lib.rs
```diff
@@ -1081,40 +1145,54 @@ impl<T: Trait + Send + Sync> SignedExtension for CheckBlockGasLimit<T> {
...
-					Ok(ValidTransaction::default())
-				}
-			},
-			Call::__PhantomItem(_, _)  => unreachable!("Variant is never constructed"),
+	fn post_dispatch(pre: Self::Pre, _info: DispatchInfo, _len: usize) {
+		if let Some(DynamicWeightData {
+			transactor,
+			imbalance,
+		}) = pre {
+			let (gas_left, gas_spent) = GasUsageReport::take();
```
kianenigma: Where is this value coming from and for what does it account for? just the contract ops? or the inner calls to runtime as well?

frame/contracts/src/tests.rs
```diff
@@ -300,17 +281,13 @@ fn compile_module<T>(wabt_module: &str)
 	Ok((wasm, code_hash))
 }
 
-// Perform a simple transfer to a non-existent account supplying way more gas than needed.
-// Then we check that the all unused gas is refunded.
+/// The payment for gas is handled outside of contract execution (e.g `call`) functions.
 #[test]
-fn refunds_unused_gas() {
-	ExtBuilder::default().gas_price(2).build().execute_with(|| {
+fn call_doesnt_pay_for_gas() {
```
kianenigma: ideally `contracts::Call()` should have some fixed weight that does actually incur some basic fee, neh?

frame/contracts/src/tests.rs
```diff
@@ -2330,19 +2307,24 @@ fn cannot_self_destruct_in_constructor() {
 	});
 }
 
+// TODO: This test relies on put_code, which no longer requires any gas.
 #[test]
 fn check_block_gas_limit_works() {
```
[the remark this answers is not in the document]
kianenigma: I think crafting the above would help us big time in understanding the consequences of each.

kianenigma:
> The problem here is that `ext_dispatch_call(T::Call)` has no information regarding how much weight the given `T::Call` would consume and thus there is no way to ensure that executing this call won't be out of given gas/weight limit budget. 

Any call's weight can be pre-computed via `get_dispatch_info`. ain't that enough? 

We also added `register_extra_weight_unchecked` in system which _could allow a call to sneak in some additional weight, aside from what is named in `#[weight]` and visible in `get_dispatch_info`_. For such cases you can record the total block weight before and after executing the call and that should tell you if this opaque call did or did not use `register_extra_weight_unchecked`. 

All in all, given `register_extra_weight_unchecked`, we remove the assumption that all weight must be known pre-dispatch. So your _problem statement_ is actually something that we have to deal with and now is part of our assumption: 

> **there is not way to ensure that executing this call won't be out of given gas/weight limit 

which is true. Given `register_extra_weight_unchecked` there really isn'y anymore.

frame/contracts/src/lib.rs
```diff
@@ -1054,19 +1043,94 @@ impl Default for Schedule {
...
+		};
+
+		match *call {
+			Call::claim_surcharge(_, _) | Call::update_schedule(_) | Call::put_code(_) => Ok(None),
+			Call::call(_, _, gas_limit, _) | Call::instantiate(_, gas_limit, _, _) => {
+				// Compute how much block weight this transaction can take at its limit, i.e.
+				// if this transaction depleted all provided gas to zero.
+				let gas_weight_limit = Weight::try_from(gas_limit)
+					.map_err(|_| InvalidTransaction::ExhaustsResources)?;
+				let weight_available = <T as system::Trait>::MaximumBlockWeight::get()
```
kianenigma: Could be made into a function in system: `get_remaining_block_weight`

frame/contracts/src/lib.rs
thiolliere: the overhead depend on the implementation of the client but currently our client has a cache-overlay-changes, so this value will be write in this cache, and the removed from this cache. at the end of the block build, this cache-overlay-changes will be committed without those values

Maybe we can find a solution not to call the sr-io api and just put that in the wasm memory but I don't know.

frame/contracts/src/lib.rs
```diff
@@ -427,8 +430,8 @@ pub trait Trait: frame_system::Trait {
 	/// The maximum size of a storage value in bytes.
 	type MaxValueSize: Get<u32>;
 
-	/// The maximum amount of gas that could be expended per block.
-	type BlockGasLimit: Get<Gas>;
+	/// Convert a weight value into a deductible fee based on the currency type.
```
thiolliere: I think as the current implementation only support linear WeightToFee we should require this in the doc

frame/contracts/src/lib.rs
```diff
@@ -1054,19 +1043,94 @@ impl Default for Schedule {
...
+				// Compute the fee corresponding for the given gas_weight_limit and attempt
+				// withdrawing from the origin of this transaction.
+				let fee = T::WeightToFee::convert(gas_weight_limit);
+
+				// Compute and store the effective price per unit of gas.
+				let gas_price = {
+					let divisor = gas_weight_limit.unique_saturated_into();
+					fee
+						.checked_div(&divisor)
+						.unwrap_or(1.into())
```
thiolliere: this ratio can go very inaccurate if WeightToFee ratio is less than 1 or if it is different from a natural number. (i.e. 1.5)
Maybe we could make use of some Perbill here, maybe we should introduce in sp_arithmetic a perbill with a natural part as well `(1.5 = 1 + 500_000_000/1_000_000_000)`

thiolliere:
also should we bump bin/node/runtime/src/lib.rs version:
```rust
	// Per convention: if the runtime behavior changes, increment spec_version
	spec_version: 204,
	impl_version: 204,
```
?

frame/contracts/src/lib.rs
pepyakin: Yeah, avoiding hitting the storage layer would be good anyway.

Some of the transient values have to survive through a runtime call boundary so they cannot be stored in memory.

frame/contracts/src/lib.rs
pepyakin: Yeah, I think it would be actually great!

frame/contracts/src/lib.rs
pepyakin: This value is saved in the `execute_wasm` function. It basically the total consumed gas: contract ops, inner calls to runtime, and everything else.

frame/contracts/src/lib.rs
pepyakin: Added! Named it `remaining_weight` though to be inline with other weight related functions in system

pepyakin:
@kianenigma yeah, as you say it yourself, it turns out `get_dispatch_info` is not enough since `register_extra_weight_unchecked` will add additional weight after the execution. Or, put in a different perspective, there is no way for the caller of `register_extra_weight_unchecked` to satisfy it requirements, namely that it doesn't overrun the max block limit.

It seems to me that we have to know the upper bound of the call at hand before execution. Otherwise, there is no way to fulfil `register_extra_weight_unchecked` requirements.

kianenigma:
> It seems to me that we have to know the upper bound of the call at hand before execution. Otherwise, there is no way to fulfil register_extra_weight_unchecked requirements.

Weight are surely from day 0 defined to be the upper bound. At the moment we don't have good estimates but this is an empirical issue from the substrate core, not your side. You can still assume

 - All weights are upper bound.
- Except for calls that use `register_extra_weight_unchecked`, which is not used by anyone atm and will stay limited. 

For calls that do use `register_extra_weight_unchecked`, IFAIK there's **really** no way to know how long will a nested call take and how much weight it will consume. You can know it ahead of time, but that will take as long as just executing it in the first place ([related](https://en.wikipedia.org/wiki/Halting_problem)). You can't do it just by examining it in a cheap way. 

How about banning certain calls, or providing a different, more complicated ways to call them? for example if you want to call sudo from your contract, you need to put some deposit down etc. 

What we can do at substrate level to help a bit with this is to add a `is_dynamic` flag or similar to `DispatchInfo` so you'd know for each call, if it is safe to assume the weight is actually the upper bound, or not.

pepyakin:
For posterity, we moved this discussion to the other channels. We should do a write up after we come to a conclusion though. My current thinking is that we won't be able to fix it in this PR.

frame/contracts/src/tests.rs
pepyakin: This ensures for now that this payment is not performed within the `::call` method itself as it used to be. Now, we totally rely that the weight mechanism will take care of it.

However, you are definitely right! I will add weight annotation to the `::call` method.

frame/contracts/src/lib.rs
pepyakin: I think that shouldn't matter if the 1 weight = 1 gas, right? Apart from that, In upcoming https://github.com/paritytech/substrate/pull/4590 there are two instances of using gas_price:

1. `approx_gas_for_balance` to compute gas fee just before calling into the runtime. This should be changed with direct weight payments in dynamic weight world.
2. `approx_gas_for_balance` for getting gas amount for `TransferFeeToken`. #4590 [will use](https://github.com/paritytech/substrate/pull/4649/files#diff-7d69b0e04bf2c554f4fd753326b32a64R576) gas cost for the transfer. I believe we can change others as well.
3. Apart from `approx_gas_for_balance` we also have `ext_gas_price`, which I think we might want to drop.

So I am wondering if a TODO here would do?

frame/contracts/src/gas_weight_conv.rs
```diff
@@ -0,0 +1,195 @@
...
+					let divisor = gas_weight_limit.unique_saturated_into();
+					fee.checked_div(&divisor).unwrap_or(1.into())
+				};
+
+				//
+				// The place where we set GasPrice for the execution of this transaction.
+				//
+				<GasPrice<T>>::put(gas_price);
+
+				let imbalance = match T::Currency::withdraw(
```
athei: You call this from `SignedExtension::validate` and `SignedExtension::pre_dispatch`? Aren't you charging the `who` double by doing this?

kianenigma: Whatever we do in `validate()` is not persistent. The state is dropped thereafter.

athei: But changes in `pre_dispatch` are persisted, right?

gnunicorn:
may be re-opened at a later time.

pepyakin:
To clarify: this, of course, doesn't imply that we drop this feature. We close it for clearing the PR queue. In future, we can re-open this PR or perhaps open another one.

athei:
I am currently working on this by building on the currently merged weight refund mechanism and the unmerged weight rework (#5446). I will start a new PR for this when it is ready for review.

PR #4148, opened 2019-11-19
Publish the dispatch info in extrinsic events

gavofyork:
Place the DispatchInfo (which contains weight information about the
extrinsic that is otherwise difficult to determine) in the dispatch
result value.

CC @jacogr


gavofyork:
afaik the RPC is useful only for extrinsics that have yet to happen, not that have happened.

jacogr:
100%, the RPC is pre-execution, the events are post-execution. 

So the events are especially useful in explorers that display historical transactions with all associated info, where the RPC is especially useful in wallets where transactions are made.

emielvanderhoek:
The request came from me (for Polkascan). There seems no other way to accurately determine fees on transaction level for historical transactions. We should be able to determine the `weight` component of the fees with this change.

PR #4156, opened 2019-11-20
fix runtime check script

gnunicorn:
Since moving some crates around, the runtime check passed when it shouldn't have. This adapts the script to the new paths.

### DIFF
--- a/.maintain/gitlab/check_runtime.sh
+++ b/.maintain/gitlab/check_runtime.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 #
 #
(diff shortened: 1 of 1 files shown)


gnunicorn:
However, looking into the script, it only checks a subset of potential file changes. Since we have been moving more things into the primitives, as the distinct API that both sides share and the dependency tree has become a lot clearer, I noticed that the current checklist is incomplete, for example `core/primitives` was being included in `core/sr-primitives` and exposed. Those a change in it would not make the script alert us.

I am thinking whether there is a more sophisticated way to really use the entire cargo dependency tree to make sure we aren't missing anything important.

gnunicorn:
superseeded.

PR #4157, opened 2019-11-20
Better docs for TxPool::import_notification_stream

seunlanlege:
A more detailed explanation on what consumers should do with the stream.

### DIFF
--- a/client/transaction-pool/graph/src/pool.rs
+++ b/client/transaction-pool/graph/src/pool.rs
@@ -279,7 +279,9 @@ impl<B: ChainApi> Pool<B> {
 			)))
 	}
 
-	/// Return an event stream of transactions imported to the pool.
+	/// Return an event stream of notifications for when transactions are imported to the pool.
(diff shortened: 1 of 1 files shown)


seunlanlege:
cc @tomusdrw

client/transaction-pool/graph/src/pool.rs
```diff
@@ -279,7 +279,9 @@ impl<B: ChainApi> Pool<B> {
 			)))
 	}
 
-	/// Return an event stream of transactions imported to the pool.
+	/// Return an event stream of notifications for when transactions are imported to the pool.
+	/// consumers of this stream should use the `ready` method to actually get the
```
```suggestion
	/// 
	/// Consumers of this stream should use the `ready` method to actually get the
```

client/transaction-pool/graph/src/pool.rs
```diff
@@ -279,7 +279,9 @@ impl<B: ChainApi> Pool<B> {
...
+	/// newly imported transactions.
```
```suggestion
	/// pending transactions in the right order.
```

PR #4158, opened 2019-11-20
custom weight function wrapper

kianenigma:
Allows easily declaring weights dynamically based on the function argument types. Can be reviewed but need to wait for https://github.com/paritytech/substrate/pull/4124 to have proper testing first. 

cc @JoshOrndorff


JoshOrndorff:
I like the idea of writing the weighting function right next to the actual dispatchable call. I'm not a fan of the mapping from integers to `DispatchClass`es because:

1. It is not obvious when I see the weighting function, what that second integer actually means.
2. It provides an opportunity for code to panic (by providing a u8 that doesn't map to a `DispatchClass`).

I think if you want syntactic sugar to make this easier for the programmer, then allow them to omit the dispatch class entirely and have it default to `Normal`. IMO the integers are just one more thing that an author would need to consult the docs for.

paint/example/src/lib.rs
```diff
@@ -482,6 +482,15 @@ decl_module! {
 			Ok(())
 		}
 
+		// TODO: this stuff will move to a new test.
+		#[weight = sr_primitives::weights::FunctionOf(
+			|args: (&u32, &T::AccountId, &BalanceOf<T>)| (args.0 + 100) as Weight,
+			1u8,
```
xlc: Can we do this?
```suggestion
			DispatchClass::Operational,
```

paint/example/src/lib.rs
```diff
@@ -482,6 +482,15 @@ decl_module! {
 			Ok(())
 		}
 
+		// TODO: this stuff will move to a new test.
+		#[weight = sr_primitives::weights::FunctionOf(
+			|args: (&u32, &T::AccountId, &BalanceOf<T>)| (args.0 + 100) as Weight,
```
xlc: Is it possible to have type inferring to avoid unnecessary code?

```suggestion
			|(val, _, _)| val + 100,
```

paint/example/src/lib.rs
kianenigma: You can use any type here that implements `Into<DispatchClass>`. We can make a balnket impl for this. My general impression was that we prefer number encoding for brevity.

paint/example/src/lib.rs
kianenigma: Sadly, as far s I tried, no

kianenigma:
I might not be able to follow a lot here. This can be either merged or left stale for the time being. It will be needed for the next gen. of weights and the new struct is not yet used. Merging it is also harmless and it is done on my opinion. Just have to make sure that there are no merge conflicts, specifically with https://github.com/paritytech/substrate/pull/4165

frame/support/src/weights.rs
```diff
@@ -258,3 +291,57 @@ impl<Call: Encode, Extra: Encode> GetDispatchInfo for sr_primitives::testing::Te
...
+	}
+
+	decl_module! {
+		pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+			// no arguments, fixed weight
+			#[weight = SimpleDispatchInfo::FixedNormal(1000)]
+			fn f0(_origin) { unimplemented!(); }
+
+			// weight = a x 10 + b
+			#[weight = FunctionOf(|args: (&u32, &u32)| args.0 * 10 + args.1, DispatchClass::Normal)]
```
kianenigma: we can probably make this a macro to not need to annotate the args on every function.

kianenigma:
@gavofyork you can merge this already if you need it, but I am working on closures around `FunctionOf` to further ease the use of it.

frame/support/src/weights.rs
```diff
@@ -236,6 +247,35 @@ impl SimpleDispatchInfo {
 	}
 }
 
+/// A struct to represent a weight which is a function of the input arguments. The given closure's
+/// arguments must match the dispatch's argument types, given as a tuple.
```
thiolliere: Maybe we could improve doc so the meaning of the different fields are directly understandable
```suggestion
/// arguments must match the dispatch's argument types, given as a tuple.
/// type is defined as: `FunctionOf(WeightDataFunction, ClassifyDispatchClass, PaysFeeBoolean)`
/// * WeightDataFunction is used to compute weight
/// * ClassifyDispatchClass is used to get the dispatch class
/// * PaysFeeBoolean is used to get whether it pays fees
```

frame/support/src/weights.rs
```diff
@@ -236,6 +247,35 @@ impl SimpleDispatchInfo {
...
+	F : Fn(Args) -> Weight
+{
+	fn weigh_data(&self, args: Args) -> Weight {
+		(self.0)(args)
+	}
+}
+
+impl<Args, F, Class> ClassifyDispatch<Args> for FunctionOf<F, Class>
+where
+	Class: Into<DispatchClass> + Clone,
```
thiolliere: I'm not entirely sure about the interest of having the Class abstract here ? can't we just have DispatchClass as second field of FunctionOf directly ?

kianenigma: Yeah I will change it to that. The purpose was to provide a syntactic simplicity. I will do that later with a macro.

gavofyork:
how are we doing here @kianenigma ?

kianenigma:
@gavofyork  I will address @thiolliere 's comments soon and label the PR for you merge it in. But don't use it _everywhere_ just yet; maybe just in occasions that we really need a function-like weight. I have some pending changes to 
1- further ease the use of `FuncitonOf()` via macro 
2- deprecate `SimpleDispatchInfo`.

Those will come in another PR

frame/support/src/weights.rs
kianenigma: removed.

frame/support/src/weights.rs
```diff
@@ -236,6 +236,36 @@ impl SimpleDispatchInfo {
 	}
 }
 
+/// A struct to represent a weight which is a function of the input arguments. The given items have
+/// the following types:
+///
+/// - `F`: a closure with the same argument list as the dispatched, wrapped in a tuple.
+/// - `DispatchClass`: class of the dispatch.
+/// - `bool`: weather this dispatch pays fee or not.
```
thiolliere:
```suggestion
/// - `bool`: whether this dispatch pays fee or not.
```

gavofyork:
@kianenigma tests failing...

PR #4162, opened 2019-11-20
Fix key collision for child trie

cheme:
This PR superseed #3804, in favor of a simple implementation. It reverts to the initial approach of #2209.
CC\ @arkpar @thiolliere 

# Polkadot required change: https://github.com/paritytech/polkadot/commit/f724806e9b2361f3688d6684516f48446c8e29c4

To avoid key collision for child trie, a unique id will be prepend to its rocksdb storage key values.
This prefix need to be crypto strong to avoid collision with other keys (we use the same column as the top trie).
Definition of this prefix is the direct responsibility of palet module using child tries.
If a module does not use a collision resistant unique id for the child trie it can break the whole db, so this is a bit less robust than previous implementation.

The palet module will probably need to store this unique id on chain but could also use static unique id in case of a bounded number of child trie.
An example of unbounded child trie usage is in contract module (we use the hash part of trie_id).

## Breaking change

- runtime extern function

This PR keep 'deprecated_host_interface' child trie function.
**I think I should delete them** because using them will only result in problem (only usable for archive mode). At this point it just act as previously by using childinfo with an empty uniqueid and no root.

Storage host child trie function were all change: this means that a chain linking to this function will be broken.
Checking latest kusama runtime, those are not use, and I do not know of any case where they are use, therefore I choose to change them **this can be a breaking change** but I think it is not.
If it ends up being a breaking change, new ext function will be needed.
The new prototype is meant to be extensible, therefore we add a 'child info' encoded byte and a u32 type.
Enum `ChildInfo` is created from them.

- child trie common api
A new parameter is use: the `ChildInfo` enum. At this point it only allow a strong crypto id with possibly a root to avoid a useless query.

For 'srml support', the api do not use `ChildInfo` but ~~simply an optional unique id with option None reserved for id given by system~~ directly a unique id. This is very questionable design as it forbid usage of child root directly. On the other hand it is a 'simple' api. So **should we align to use childinfo**?

- genesis json
the genesis encoding does change to include the child information, this means that existing json chainspec containing some child trie will need to be updated.
I do not know any chainspec using child trie, so this should not be an issue.

- rpc
We need to break rpc (unique id being set/defined/generated by palet module we have no way to only use storage key as before).
I firstly implemented it by added 'unique id' after 'storage key' parameter, but finally find it more extensible to use the same api as for externalities.
That way we can already use rpc without fetching child root (child info starting with 32 byte of root, then unique id and child type 2).
This can be use to get some partial proof (only the child query part of the proof). 

## Next steps

- change contract module to memoïze child root (either by direct runtime-io usage or change support child api) and use it in child info, this can be included in this pr.

- putting child trie with a sequential uniqueid stored in state that is not crypto resistant should be easy, but it requires writing all child trie into their own rocksdb column. This would be equivalent to final design of #2209. This can be implemented in a follow up pr by using another child type key in rpc and ext interface and using None in general interface.
In fact api already take unique id as optional for this next PR.

- Using child trie with a sequential defined id that is not stored in state (stored on kv storage) will be for later (putting the kv storage to store this information in place being a substantial task that may require additional design).


client/src/light/backend.rs
```diff
@@ -284,18 +285,17 @@ where
 		check_genesis_storage(&top, &children)?;
 
 		// this is only called when genesis block is imported => shouldn't be performance bottleneck
-		let mut storage: HashMap<Option<Vec<u8>>, StorageOverlay> = HashMap::new();
+		let mut storage: HashMap<Option<(Vec<u8>, OwnedChildInfo)>, StorageOverlay> = HashMap::new();
```
cheme: Usage of 'OwnedChildInfo' in the key is rather questionable here, yet it seem to work well with current implementation.

primitives/state-machine/src/backend.rs
```diff
@@ -219,7 +283,10 @@ impl Consolidate for () {
 	}
 }
 
-impl Consolidate for Vec<(Option<Vec<u8>>, Vec<u8>, Option<Vec<u8>>)> {
+impl Consolidate for Vec<(
+		Option<(Vec<u8>, OwnedChildInfo)>,
+		Vec<(Vec<u8>, Option<Vec<u8>>)>,
+	)> {
 	fn consolidate(&mut self, mut other: Self) {
 		self.append(&mut other);
```
cheme: This will produce multiple definition with the same key (child trie storage key and ownedinfo), a btreemap could be use instead, but this is closer to existing code (it already factor some data redundancy).

primitives/state-machine/src/backend.rs
```diff
@@ -357,22 +426,33 @@ impl<H: Hasher> From<Vec<(Option<Vec<u8>>, Vec<u8>, Option<Vec<u8>>)>> for InMem
...
 
-	fn child_storage(&self, storage_key: &[u8], key: &[u8]) -> Result<Option<Vec<u8>>, Self::Error> {
-		Ok(self.inner.get(&Some(storage_key.to_vec())).and_then(|map| map.get(key).map(Clone::clone)))
+	fn child_storage(
+		&self,
+		storage_key: &[u8],
+		child_info: ChildInfo,
+		key: &[u8],
+	) -> Result<Option<Vec<u8>>, Self::Error> {
+		Ok(self.inner.get(&Some((storage_key.to_vec(), child_info.to_owned())))
```
cheme: In Memory costy instantiation for key. Generally In memory could change to some { top :  , children: } with child info into value of children.

cheme:
I did start thinking on how things can evolve if the api switch to using hierarchical child trie (it start to seem usefull for case such as palet module in its own child trie which itself use some child trie).

The point is that changing child api at this point is done under the asumption that child trie are not being use in any runtime, but future changes would need to keep the existing sr-io interface in place.
Therefore it seems interesting to foresee such changes and maybe put in place an api that is conformant right now.
This is mainly for sr-io and rpc.


First idea would be to use multiple storage_key as parameter, but it would also require multiple (child_info, child_type) for each step.

Another idea would be that hierarchical trie would use their own child trie_type and child_info encoding.
Every hierachical level of child trie will need its own child type, child info and storage key, this can be set/encoded in child info. It will then resolve to a simple vec of child info.
So we do not change the current pr interface.
Storage_key will get processed by 'resolve_child_info' to produce multiple storage key (storage key with a given child_type is seen a concatenation of multiple scale encoded storage key), it will be the same for ChildInfo.
`resolve_child_info` then would also resolve storage key when some new internal api is put in place.


At this point open questions should be:
- shoud we encode child_type in child_info from the start (if we plan to encode them for hierarchical trie it makes less sense to have a separate parameter)?
- should we make storage key part of child info (I think they do not have similar meaning and that it is probably a bad idea)? or simply rely on the child type for an additional storage key transformation in the future.
- should we make an interface that is more like a Vec<(storage_key, child_info, child_type)> from the start? For sr_io I guess it would result in encoding, but for rpc it could looks better (but changing rpc later seems doable even if it can be a burden)?

gavofyork:
would this break kusama? (i don't think so as we don't use child tries, but want to check)

cheme:
As long as there is no wasm runtime containing 'ext_child...' function it should not.
I did check that on latest polkadot build, checking every previous runtime version would be a bit tedious.
I think I should just test a synch with this PR (I did it on my overlay_change pr and it is not much work).
For sure this will break rpc for child because additional content is needed for the query.

gavofyork:
ok - there will be child tries in use once we introduce crowdfund module, so it would be nice to get this in soonish. will need resolving. who is best to review? @arkpar maybe?

primitives/sr-io/src/lib.rs
```diff
@@ -92,16 +92,23 @@ pub trait Storage {
...
 	/// Returns the data for `key` in the child storage or `None` if the key can not be found.
-	fn child_get(&self, child_storage_key: &[u8], key: &[u8]) -> Option<Vec<u8>> {
+	/// A key collision free unique id is required as parameter.
+	fn child_get(
+		&self,
+		child_storage_key: &[u8],
+		child_definition: &[u8],
+		child_type: u32,
+		key: &[u8],
+	) -> Option<Vec<u8>> {
```
thiolliere: I think, parameter here lacks documentation, also note that we can use any scale encodable structure here. But maybe its fine

frame/support/src/storage/child.rs
```diff
@@ -19,14 +19,31 @@
 //! This module is a currently only a variant of unhashed with additional `storage_key`.
 //! Note that `storage_key` must be unique and strong (strong in the sense of being long enough to
 //! avoid collision from a resistant hash function (which unique implies)).
+//!
+//! A **key collision free** unique id is required as parameter to avoid key collision
+//! between child tries.
+//! This unique id management and generation responsability is delegated to pallet module.
+//! In the future this id could be generated and stored out of chain state, and using None as a
+//! parameter will result into this mode of operation.
```
thiolliere: I think we can remove these lines as it is quite hypotetical. I know I said maybe this could handled by system or something. But at the end I also feel very fine to have modules managing it themself. Like they currently manage their prefix for storages.

Maybe just remove all these optional unique id we will open an issue for ideas.

Later we can make it optional very easily anyway

primitives/externalities/src/lib.rs
```diff
@@ -45,33 +45,59 @@ pub trait Externalities: ExtensionStore {
...
 	/// Read child runtime storage.
-	fn child_storage(&self, storage_key: ChildStorageKey, key: &[u8]) -> Option<Vec<u8>>;
+	fn child_storage(
+		&self,
+		storage_key: ChildStorageKey,
+		child_info: ChildInfo,
+		key: &[u8],
+	) -> Option<Vec<u8>>;
 
 	/// Set storage entry `key` of current contract being called (effective immediately).
```
thiolliere: doc will have to be changed here I think, but future PR

primitives/sr-io/src/lib.rs
cheme: Yes that is one of the thing I did ask myself, is it worth separating both parameters.
My justification was that child info and child storage key should be handle separately because storage key can potentially change more often than child_definition (and is mostly use as key in our overlays and others temporary structure). But it sure could make sense to have both in the same field (especially if we consider a possibility of hierarchical trie where we would need to decode/encode the multiple level of storage key from/to child_storage_key depending on the child_type).
I will document that after rebasing to master.

cheme:
For the review, we discuss this with both @arkpar and @thiolliere , so both opinion and/or reviews are very welcome.
The current design is in a middle-ground between having a very specific interface that will break often and having a full generic one that makes no sense; yet it is still open for discussion (sr-io and rpc may look a bit to generic (there is a encoded payload for child_info), but it seems at the same time a bit more future proof).

primitives/sr-io/src/lib.rs
thiolliere: > My justification was that child info and child storage key should be handle separately because storage key can potentially change more often than child_definition (and is mostly use as key in our overlays and others temporary structure).

Yes I understand from the runtime perspective both are very linked and will be changed together, from the overlay (maybe trie level) having the key to uniquely identify a childtrie is enough.

I'm not sure of what is best either

frame/support/src/storage/child.rs
```diff
@@ -19,14 +19,29 @@
...
+//! A **key collision free** unique id is required as parameter to avoid key collision
+//! between child tries.
+//! This unique id management and generation responsability is delegated to pallet module.
 // NOTE: could replace unhashed by having only one kind of storage (root being null storage key (storage_key can become Option<&[u8]>).
 
 use crate::rstd::prelude::*;
 use codec::{Codec, Encode, Decode};
 
+/// Unique id managed by the system is not implemented yet.
+const SOME: &'static str = "Managed child unique ids are not yet implemented";
```
thiolliere: I was thinking removing this unimplemented part as well. I think all the `Option<UniqueID>` should be changed for UniqueID directly.
Later if we reorganize differently by having this unique ID generated by system or any other way, we can make it optional.

client/executor/src/deprecated_host_interface.rs
```diff
@@ -406,7 +448,10 @@ impl_wasm_host_interface! {
 		) -> Pointer<u8> {
 			let storage_key = context.read_memory(storage_key_data, storage_key_len)
 				.map_err(|_| "Invalid attempt to determine storage_key in ext_child_storage_root")?;
-			let value = runtime_io::storage::child_root(&storage_key);
+			// This can only work with archive mode.
+			let value = runtime_io::storage::child_root(
+				&storage_key,
+			);
```
thiolliere: can this only works with archive mode ? I thought that now runtime_io::storage::child_root was fixed.

cheme: That is the 'deprecated_host_interface' where I kept old behavior.
In fact this should be deleted since not use in kusama, but I was not bold enough to do it (that is the first bold point in description).
I think I shall remove it, anyway I will test kusama synch again (last test was a bit to far away and with parasitic trie behavior).
So idea would be to remove all child related call from deprecated host, agree?

primitives/state-machine/src/overlayed_changes.rs
```diff
@@ -149,10 +149,17 @@ impl OverlayedChanges {
...
+		child_info: ChildInfo,
+		key: Vec<u8>,
+		val: Option<Vec<u8>>,
+	) {
 		let extrinsic_index = self.extrinsic_index();
-		let map_entry = self.prospective.children.entry(storage_key).or_default();
-		let entry = map_entry.entry(key).or_default();
+		let map_entry = self.prospective.children.entry(storage_key)
+			.or_insert_with(|| (Default::default(), child_info.to_owned()));
+		let entry = map_entry.0.entry(key).or_default();
```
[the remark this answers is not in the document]
cheme: I just realize that this can go wrong if a user change child info and spawn a child info with root or a child_info without root (then Eq will mismatch). But I think it would indicate bad use of the api (if you memoize root you should do it globabally and not use some new child_info everywhere.
I can also change Eq to only make root being None always equal to root being Some.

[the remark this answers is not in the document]
thiolliere: > if user change child info

what does that mean ? like in the future when there is another extension one could change the extension here ? would it work ?
I thought changing the extension wasn't doable.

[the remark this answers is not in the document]
cheme: By user I meant runtime developer, but it would be a bug. (triggered by using a child trie api both with root memoization and without).

primitives/core/storage/src/lib.rs
```diff
@@ -156,3 +159,155 @@ impl<'a> ChildStorageKey<'a> {
...
+/// Information related to a child state.
+pub enum ChildInfo<'a> {
+	Default(ChildTrie<'a>),
+}
+
+/// Owned version of `ChildInfo`.
+/// To be use in persistence layers.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "std", derive(PartialEq, Eq, Hash, PartialOrd, Ord))]
+pub enum OwnedChildInfo {
```
thiolliere: I feel like usually it goes the other way around with ChildInfo and ChildInfoRef.
But I'm completely fine with that

primitives/state-machine/src/basic.rs
```diff
@@ -214,20 +238,24 @@ impl Externalities for BasicExternalities {
 					.expect("Map only feed by valid keys; qed"),
 			);
 			if &empty_hash[..] == &child_root[..] {
-				top.remove(&storage_key);
+				top.remove(storage_key.as_slice());
 			} else {
-				top.insert(storage_key, child_root);
+				top.insert(storage_key.clone(), child_root);
```
thiolliere:
```suggestion
				top.insert(storage_key, child_root);
```

primitives/state-machine/src/trie_backend.rs
```diff
@@ -87,11 +93,22 @@ impl<S: TrieBackendStorage<H>, H: Hasher> Backend<H> for TrieBackend<S, H> where
...
-	fn for_child_keys_with_prefix<F: FnMut(&[u8])>(&self, storage_key: &[u8], prefix: &[u8], f: F) {
+	fn for_child_keys_with_prefix<F: FnMut(&[u8])>(
+		&self,
+		storage_key: &[u8],
+		_child_info: ChildInfo,
+		prefix: &[u8],
+		f: F,
+	) {
 		self.essence.for_child_keys_with_prefix(storage_key, prefix, f)
 	}
```
thiolliere: this is wrong _child_info must be used

cheme: Clearly :+1:

thiolliere:
The unique ID used by to prepend kvdb keys for the child trie nodes is not stored inside the main trie.

This can be ok but we have to keep in mind because of this we can't iterate on all the value contained in a block:
* we can only iterate on values in the main trie, in order to get the values in a child trie we have its hash but we also need the unique_id. As this unique_id is now set by the runtime, this is not straightforward.

Maybe it is fine this way, but otherwise we could also put this unique id in the main trie alongside the child trie root hash.

cheme:
That is indeed a big limitation, previous implementation did thing this way even for the type, making the child_info parameter optional.
Currently I am thinking that it can be desirable, this design really tie the child trie usage to its runtime. I also think we may want iterable child trie outside of runtime context (eg if we want multiple level of child trie with chained deletion), but it would be another child type (as child trie with system sequential id could be). In this case Child info would have an optional couple uniqueid - root.
But it may be preferable to create it when it gets needed.

gavofyork:
this is needed ASAP since we'll want to use child tries in crowdfund module, so should be prioritised.

thiolliere:
> I also think we may want iterable child trie outside of runtime context (eg if we want multiple level of child trie with chained deletion), but it would be another child type

Yes indeed it looks fine this way then.

At some point I was thinking that now that we have all module after a prefix such feature such as storage usage of a module would require this feature.

I'm ok with the current design.

cheme:
Added polkadot required change https://github.com/paritytech/polkadot/commit/3aa711a36bcb39b28e3384764ef00e414f7718e1 in description.

cheme:
@thiolliere latest commit https://github.com/paritytech/substrate/pull/4162/commits/d4c73345bf35b387b4aefa68920f988b12c34c35 changes a bit ChainInfo to be use by frame support. The chaininfo code looks overall a bit worst, but it is to avoid some allocation on every calls.
This should allow storing root (otherwhise on had to call directly sr-io from its runtime).

client/chain-spec/src/chain_spec.rs
```diff
@@ -99,7 +106,7 @@ enum Genesis<G> {
 	Runtime(G),
 	Raw(
 		HashMap<StorageKey, StorageData>,
-		HashMap<StorageKey, HashMap<StorageKey, StorageData>>,
+		HashMap<StorageKey, (HashMap<StorageKey, StorageData>, Vec<u8>, u32)>,
```
thiolliere: maybe add some doc about what is supposed to be Vec<u8> and u32 here

primitives/core/storage/src/lib.rs
```diff
@@ -156,3 +159,198 @@ impl<'a> ChildStorageKey<'a> {
...
+/// Type of child.
+/// It does not strictly define different child type, it can also
+/// be related to technical consideration or api variant.
+#[repr(u32)]
+pub enum ChildType {
+	/// Default, it uses a cryptographic strong unique id as input.
+	CryptoUniqueId = 1,
+	/// Default, but with a root registerd to skip root fetch when querying.
+	/// Root is a 32 byte hash.
+	CryptoUniqueIdRoot32Api = 2,
```
thiolliere: Does having the thing only implemented for root hash being 32 byte to too limiting ? don't we end to something a bit similar to what does fix https://github.com/paritytech/substrate/pull/4231

cheme: We could add the hash length before the root in data? A bit extra memory but it would avoid consuming new child_type id.
I think it may be fine (since the root variant is no longer buildable from slice but only from a owned child info), I will implement it.

primitives/sr-io/src/lib.rs
```diff
@@ -117,15 +135,21 @@ pub trait Storage {
...
+		child_definition: &[u8],
+		child_type: u32,
 		key: &[u8],
 		value_out: &mut [u8],
 		value_offset: u32,
 	) -> Option<u32> {
 		let storage_key = child_storage_key_or_panic(child_storage_key);
-		self.child_storage(storage_key, key)
+		let child_info = ChildInfo::resolve_child_info(child_type, child_definition)
+			.expect("Invalid child definition");
```
thiolliere: It seems all the validation of child specification are done here, should we also check that storage_key path match the child_info variants: which means contains ':default' as there is only Default child tries for now ?

primitives/state-machine/src/ext.rs
```diff
@@ -474,29 +516,48 @@ where
...
+
+					self.backend.child_storage_root(storage_key, child_info.as_ref(), delta)
+				};
+
+				let root = root.encode();
+				if is_empty {
+					self.overlay.set_storage(storage_key.into(), None);
+				} else {
+					self.overlay.set_storage(storage_key.into(), Some(root.clone()));
+				}
```
thiolliere: Why is this interesting ?
this root hash gets invalid if another value is set.

primitives/state-machine/src/backend.rs
thiolliere: Something I wonder, should we actually keeps this childInfo ? in the real backend those information are not stored.

primitives/state-machine/src/ext.rs
cheme: Yes I think it is not working properly, IIRC correctly it was there to be able when calling 'storage_root' to get a storage root that takes account of child storage root. But in fact 'storage_root' should not ('full_storage_root' should be use).
I think a simple test at state-machine level (to avoid storage_cache) like :
```
let a = self.child_storage(storage_key, key);
self.child_root(storage_key);
let b = self.child_storage(storage_key, key);
assert_eq(a, b)
```
will fail (unless a is not set).
This will need fix in another pr (child_root transaction and root also need to be stored for backend).

primitives/sr-io/src/lib.rs
cheme: I do not know, I did not want to merge both definition.

primitives/state-machine/src/backend.rs
cheme: I need it to calculate the trie delta (when calling full_storage_root I need to rebuild the roots so I need to know where and how children are stored (to query existing trie state)).

cheme: Sorry, I misread the question, in the real backend I think child info is stored with the delta (in value of the map).

primitives/sr-io/src/lib.rs
thiolliere: currently to parameters are related, if a user do call:
(Lets say we create a new child called second_kind)
```
child::set(
  storage_key=$prefix ++ :default ++ $some_path,
  child_info=SecondKind($some_uid,
  key=b"1"
)
```
Then this call will not fail, depending on how the backend is written it can even not panic, this wrong argument will be hard to debug for the user.

primitives/state-machine/src/ext.rs
thiolliere: So currently it doesn't do anything right ?

cheme: I was looking at it, I remember better, it is specific to externalities: 'storage_root' for externalities got the same semantic as 'full_storage_root' for the overlay, and this was used to cache the result of the child trie root, so on call to child root only we will get the right value.
I will try my test and possibly remove the caching from child_trie_root and those line.

primitives/core/storage/src/lib.rs
```diff
@@ -156,3 +160,244 @@ impl<'a> ChildStorageKey<'a> {
...
+				}),
+		}
+	}
+
+	/// Create child info from a linear byte packed value and a given type. 
+	pub fn resolve_child_info(child_type: u32, data: &'a[u8]) -> Option<Self> {
+		match child_type {
+			x if x == ChildType::CryptoUniqueId as u32 => Some(ChildInfo::new_default(data)),
+			x if x == ChildType::CryptoUniqueIdRootApi as u32 => {
+				let data_cursor = &mut data.clone();
```
thiolliere:
```suggestion
				let data_cursor = &mut &data[..];
```

primitives/core/storage/src/lib.rs
```diff
@@ -156,3 +160,244 @@ impl<'a> ChildStorageKey<'a> {
...
+				// u32 is considered enough for a root size.
+				let number: u32 = match Decode::decode(data_cursor) {
```
thiolliere: Maybe more readable:
```suggestion
				let number: u32 = Decode::decode(data_cursor)
					.unwrap_or_else(|_| None)?;
```

primitives/core/storage/src/lib.rs
```diff
@@ -156,3 +160,244 @@ impl<'a> ChildStorageKey<'a> {
...
+					Ok(number) => number,
+					Err(_) => return None,
+				};
+				let offset = data.len() - data_cursor.len();
+				if data.len() >= number as usize {
+					Some(ChildInfo::Default(ChildTrie {
+						offset,
```
thiolliere: currenty the offset is u32 not scale encoded so it is always 4 byte. did you mean to use scale encoding ?

thiolliere: also you can use this trait parity_scale_codec::CompactLen but I'm fine with having offset

primitives/state-machine/src/ext.rs
cheme: Ok I wrote a test and it is actually not broken.
The fact is that child storage root is always fetch from the trie backend (never from the changed value in the overlay), so it is ok to get some cache, but it should definitely be rewrite to get unitary cache from each call to child_storage_root (caching root value from trie backend should be more prioritary).

thiolliere:
I still havn't reviewed "primitives/trie/src/lib.rs" but two things:
* As discussed a bit in private, I think it is quite difficult from the runtime to make an efficient use of the feature: being able to give the root of the child trie for better performance (in function in sr-io).

  All in all I think this is more a cache layer to do this, currently in this architecture implementing such a cache seems not straightforward. But ideally I think when you fetch a childtrie we should cache its root so that a future fetch won't go through the main trie anymore
* I feel that not storing the unique id in the main trie might leads to future difficulties:
  * we can't get the whole trie: main trie + child tries with just the root hash, we need some runtime information on top of this.
    getting all storage used by a module won't be possible just with the module prefix.
  * note: we will probably give put this unique id in the metadata.

Otherwise looks good, still have to review substrate-trie.

primitives/core/storage/src/lib.rs
cheme: :+1: I switch to compact (that was my intent, otherwhise the offset field is useless).

cheme:
- I can remove the root type, I think it can make sense for rpc and light but it is true for runtime, a cache in Ext would be probably easier to use (and with a larger scope: I will remove that tomorrow).
- About storing the unique id I fully share your views on the limitation of this PR, but I am not sure I want to go back to state storage like in my first PR.

primitives/state-machine/src/ext.rs
thiolliere: maybe we could write a short comment on top of this if

thiolliere:
OK I approve the PR as it is, still need to understand "primitives/trie/src/lib.rs" but looks ok.

if another one want to discuss things, this are few points we could think more to me:

* storing the unique id on chain so that from the block hash we can traverse the whole trie+childtries. otherwise you need information from the runtime about what unique id he was using.
  currently contract use as unique_id the hash part of the child_storage_key, so it is actually stored in the trie.
   All child tries following this convention are explorable.

   If we really need this feature for some childtries we can always create a new childtrie type later.

* (minor) in sr-io we have this three argument to specify child_storage_key (=path in main trie to access the child storage root), child_type (=default), child_info (for child_type=default then unique_id), this could be organized differently, but current spec is extensible and works well.

primitives/state-machine/src/ext.rs
cheme: Yes, I already forgot once how it works, so it seems more than usefull.

gavofyork:
@arkpar maybe worth you taking a look before merge?

client/chain-spec/src/chain_spec.rs
arkpar: Or better yet, use a struct. Tuples of this size are hard to follow.

arkpar:
`child_type` and `child_data` seem to be always passed together. Why not replace them with a single enum?  Do we even need `child_type`? Are there any types other than unique-id?

primitives/state-machine/src/backend.rs
```diff
@@ -26,6 +26,12 @@ use trie::{
 	trie_types::{TrieDBMut, Layout},
 };
 use codec::{Encode, Codec};
+use primitives::storage::{ChildInfo, OwnedChildInfo};
+
+pub(crate) type StorageTuple = (
```
arkpar: This whole type would look much better as a struct with another struct for the second `HashMap` value

cheme:
At the start of the PR we got two child types, it was cut down to one.
We currently do not strictly need it, it is just for future child trie kind. 
It is true we can always create new host function when we got more than one kind. For the host function it can be fine as it will only default to this kind and it seems that we can have versionable host function now, just the rpc and light query to break/update.

cheme:
Putting status 'inprogress' while I switch code to using struct instead of tuple.

cheme:
I did switch to struct where there is tuple, this is very likely to conflict with other PR but looks better (change to genesis conf seems to work directly with polkadot (json parser is quiet flexible)).
@arkpar, @thiolliere , should we remove child type (in this case I will also rename all child_info field to unique_id, and I am not sure if I also should remove ChildInfo enum to use Vec and &[u8] instead)?

arkpar:
I'm generally fine with the way it is now.

gavofyork:
@cheme needs resolving

PR #4165, opened 2019-11-21
PaysFee for DispatchInfo

xlc:
This allow more flexible fee logic such as adjust fee based on DispatchClass.

I want to have operational transactions have weight but no fees.


kianenigma:
> I want to have operational transactions have weight but no fees.

Fair and indeed a problem right now that we should fix. 

I will think a bit about this and maybe and see if I can come up with a more general way of expressing fee logic in `#[weight]` without making it too complicated. Ideally, I want to get rid of a horrible function `pay_len_fee()` which is total crap as well. But this PR also seems like a good start. Thanks :)

kianenigma:
Needs further test fixing as well.

kianenigma:
@xlc how about this: 

- get rid of `pay_length_fee()`. 
- A third trait that all weight representatives must implement named `Trait PaysFee { fn pays_fee() -> bool { true; } }` + a blanket impl for all structs that have the other two traits (`WeighData`, `ClassifyDispatch`).
- new filed in `DispatchInfo` named `pays_fee: bool` which is being set by `dispatch.rs`. 

Then, whoever calls `get_dispatch_info()` know this and it can be expressed in the `#[weight]` tag. `SimpleDispatchInfo` does not show it atm but we can readily make it possible.

palette/support/src/weights.rs
```diff
@@ -76,6 +76,13 @@ pub trait WeighBlock<BlockNumber> {
 	fn on_finalize(_: BlockNumber) -> Weight { Zero::zero() }
 }
 
+pub trait PaysFee {
```
kianenigma:
```suggestion
/// Indicates if dispatch function should pay fees or not. If set to false, the block resource limits are applied, yet no fee is deducted. 
pub trait PaysFee {
```

palette/support/src/weights.rs
```diff
@@ -76,6 +76,13 @@ pub trait WeighBlock<BlockNumber> {
...
+	/// Indicates if dispatch function should pays fee.
```
kianenigma:
```suggestion

```

palette/support/src/weights.rs
```diff
@@ -210,6 +208,20 @@ impl<T> ClassifyDispatch<T> for SimpleDispatchInfo {
 	}
 }
 
+impl PaysFee for SimpleDispatchInfo {
```
kianenigma: Okay but I don't think this is enough for your use case right? I would allow a way to express this through the annotation (similar to https://github.com/paritytech/substrate/pull/4158, or we can just wait for that). Technically you want to have a, lets say operational, transaction with weight 100, once with fees and once without. I think this simple Enum is no longer enough and we need something better, like https://github.com/paritytech/substrate/pull/4158

palette/transaction-payment/src/lib.rs
```diff
@@ -154,28 +154,28 @@ impl<T: Trait + Send + Sync> ChargeTransactionPayment<T> {
...
-		};
+			let len_fee = base.saturating_add(per_byte.saturating_mul(len));
 
-		let weight_fee = {
-			// cap the weight to the maximum defined in runtime, otherwise it will be the `Bounded`
-			// maximum of its data type, which is not desired.
-			let capped_weight = info.weight.min(<T as system::Trait>::MaximumBlockWeight::get());
-			T::WeightToFee::convert(capped_weight)
-		};
+			let weight_fee = {
```
kianenigma: hmm I think now we want to have any fee only when `pays_fee == true`? then weight fee should also go inside the `if`, no?

xlc: this is already inside the if. It can still pay tips even pays_fee is false, I guess that make sense.

palette/support/src/weights.rs
xlc: Maybe I should have `FreeNormal` and `FreeOperational` takes a weight? Does it make sense to actually have a weightless transaction?

kianenigma: At the simplest, you can make all of them get a `bool` that gets mapped to pays_fee. So you can say: `FreeOperation(true)` and `FreeOperational(false)`. But I think this is no longer _Simple_ stuff as the name says.  I think we can ditch SimpleDispatchInfo entirely and come up with better stuff. I will think about it.

xlc: Yeah. I think this can be done in later PR.

palette/support/src/weights.rs
```diff
@@ -136,17 +144,8 @@ pub struct DispatchInfo {
...
-impl DispatchInfo {
-	/// Determine if this dispatch should pay the base length-related fee or not.
-	pub fn pay_length_fee(&self) -> bool {
-		match self.class {
-			DispatchClass::Normal => true,
-			// For now we assume all operational transactions don't pay the length fee.
-			DispatchClass::Operational => false,
-		}
-	}
+	/// Does this transaction pays fee.
```
kianenigma:
```suggestion
	/// Does this transaction pay fees.
```

palette/support/src/weights.rs
kianenigma: sure, will make an issue for it.

palette/transaction-payment/src/lib.rs
kianenigma: ahh yeas indeed, i did not see the indentation. Kinda hard to check this stuff in gh ui :D

palette/transaction-payment/src/lib.rs
```diff
@@ -476,7 +478,8 @@ mod tests {
 			// like a FreeNormal
 			let free_transaction = DispatchInfo {
 				weight: 0,
-				class: DispatchClass::Normal
+				class: DispatchClass::Normal,
+				pays_fee: false
```
kianenigma:
```suggestion
				pays_fee: false,
```

PR #4166, opened 2019-11-21
Revert "Fix serialization of `transaction_paymentInfo`"

tomusdrw:
Reverts paritytech/substrate#4146

### DIFF
--- a/palette/transaction-payment/rpc/runtime-api/src/lib.rs
+++ b/palette/transaction-payment/rpc/runtime-api/src/lib.rs
@@ -27,7 +27,6 @@ use serde::{Serialize, Deserialize};
 /// Some information related to a dispatchable that can be queried from the runtime.
 #[derive(Eq, PartialEq, Encode, Decode, Default)]
 #[cfg_attr(feature = "std", derive(Debug, Serialize, Deserialize))]
-#[cfg_attr(feature = "std", serde(rename_all = "camelCase"))]
 pub struct RuntimeDispatchInfo<Balance> {
 	/// Weight of this dispatch.
 	pub weight: Weight,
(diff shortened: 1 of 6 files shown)


kianenigma:
I have also faced some issues with `arbitrary_precision ` of serde and json_rpc crate in other repos... shan't we for now make the fee RPC return a (potentially lossy) `SaturatedInto<u64>` of the fee? that's the only problem here. I can make a quick patch for that

palette/transaction-payment/rpc/runtime-api/src/lib.rs
```diff
@@ -27,7 +27,6 @@ use serde::{Serialize, Deserialize};
 /// Some information related to a dispatchable that can be queried from the runtime.
 #[derive(Eq, PartialEq, Encode, Decode, Default)]
 #[cfg_attr(feature = "std", derive(Debug, Serialize, Deserialize))]
-#[cfg_attr(feature = "std", serde(rename_all = "camelCase"))]
```
jacogr: Not sure this specific line should be reverted - this doesn't affect anything and as it stands the naming here is inconsistent with what is used everywhere else.

primitives/sr-primitives/src/weights.rs
```diff
@@ -95,7 +95,6 @@ impl<BlockNumber: Copy> WeighBlock<BlockNumber> for SingleModule {
 /// A generalized group of dispatch types. This is only distinguishing normal, user-triggered transactions
 /// (`Normal`) and anything beyond which serves a higher purpose to the system (`Operational`).
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-#[cfg_attr(feature = "std", serde(rename_all = "camelCase"))]
```
jacogr: Same here with `camelCase`, don't believe it should be reverted.

PR #4167, opened 2019-11-21
Verify Grandpa proofs from within runtime

HCastano:
I'm looking to get some feedback on this. For the bridge module, we need to be able to verify the validity of Grandpa proofs. The problem is however, that there is no way to do this from within the runtime as the current Grandpa proof code requires `std`. What I've done here is taken some of the code from the Substrate `finality-grandpa` crate and modified it to make it work in a `no_std` environment. The two main changes were:

1. Switch `HashMap` and `HashSet` to `BTreeMap` and `BTreeSet`
2. Move from using `AuthorityPair` for signature verification, to using `AuthoritySignature` + `RuntimeAppPublic`

Ideally this code wouldn't be duplicated in two places, but right now I'm just trying to make this work in the context of the bridge module.

Before merging this PR I'd like to have the `submit_finalized_headers()` function actually using the new code to verify Grandpa proofs.


HCastano:
As @svyatonik pointed out to me, not all the functions are needed by the bridge module (e.g `from_commit()`. While this is true, I think they'd be useful for creating unit tests.

srml/bridge/src/justification.rs
```diff
@@ -0,0 +1,288 @@
...
+/// to the commit target block. Due to the current voting strategy the precommit
+/// targets should be the same as the commit target, since honest voters don't
+/// vote past authority set change blocks.
+///
+/// This is meant to be stored in the db and passed around the network to other
+/// nodes, and are used by syncing nodes to prove authority set handoffs.
+#[derive(Encode, Decode)]
+pub struct GrandpaJustification<Block: BlockT> {
+	round: u64,
+	pub(crate) commit: Commit<Block>,
```
rphmeier: pub(crate) field should have docs. I'd prefer an accessor method though.

srml/bridge/src/justification.rs
```diff
@@ -0,0 +1,288 @@
...
+/// A GRANDPA justification for block finality, it includes a commit message and
+/// an ancestry proof including all headers routing all precommit target blocks
+/// to the commit target block. Due to the current voting strategy the precommit
+/// targets should be the same as the commit target, since honest voters don't
+/// vote past authority set change blocks.
+///
+/// This is meant to be stored in the db and passed around the network to other
+/// nodes, and are used by syncing nodes to prove authority set handoffs.
+#[derive(Encode, Decode)]
+pub struct GrandpaJustification<Block: BlockT> {
```
rphmeier: I feel like the contents of this module should be in the `pallette/grandpa` crate, since it's generally useful.

One thing - we have to deal with validator-set handoffs as well.

srml/bridge/src/justification.rs
```diff
@@ -0,0 +1,288 @@
...
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
```
rphmeier: missing crate docs. For something this complex there should be more than one line when you add some

srml/bridge/src/justification.rs
```diff
@@ -0,0 +1,288 @@
...
+use client::{CallExecutor, Client};
+use client::backend::Backend;
+use client::error::Error as ClientError;
+use codec::{Encode, Decode};
+use core::cmp::{Ord, Ordering};
+use fg::{Commit, Error, Message}; // Q: Should I make any of these a part of fg_primitives?
```
rphmeier: yeah, ideally `fg_primitives` re-exports this stuff

srml/bridge/src/justification.rs
andresilva: IMO it should go into `primitives/finality-grandpa` and then it gets used by both client and pallet code (and also by the bridge pallet).

srml/bridge/src/justification.rs
```diff
@@ -0,0 +1,288 @@
...
+	set_id: SetIdNumber,
+) -> Result<(), ()> {
+	let as_public = id.clone();
+	let encoded_raw = localized_payload(round, set_id, message);
+
+	// Since `app::Public` implements `RuntimeAppPublic` we can call `verify()`
+	if as_public.verify(&encoded_raw, signature) {
+		Ok(())
+	} else {
+		// debug!(target: "afg", "Bad signature on message from {:?}", id);
```
andresilva: We probably want to keep this for the client code that uses it once this is refactored to re-use the same code.

srml/bridge/src/justification.rs
```diff
@@ -0,0 +1,288 @@
...
+	message: &Message<Block>,
+	id: &AuthorityId,
+	signature: &AuthoritySignature,
+	round: RoundNumber,
+	set_id: SetIdNumber,
+) -> Result<(), ()> {
+	let as_public = id.clone();
+	let encoded_raw = localized_payload(round, set_id, message);
+
+	// Since `app::Public` implements `RuntimeAppPublic` we can call `verify()`
```
andresilva: Is this comment really needed?

HCastano: It was more a note to self, so I guess not

srml/bridge/src/justification.rs
HCastano: I'm going to wait until I rebase past `the-big-reorg` before moving the code into `primitives/finality-grandpa`.

srml/bridge/src/justification.rs
HCastano: Looks like we'd have to pull in the `finality-grandpa` crate to do this, so it's not really worth it.

srml/bridge/src/lib.rs
```diff
@@ -241,14 +240,19 @@ where
...
 		&justification,
 		(hash, number),
 		set_id,
 		voters,
-	).unwrap();
+	);
 
-	Ok(())
+	// Uh, there's got to be a nicer way of doing this...
+	match j {
```
niklasad1: you could do (provided that your error type implement client::Error -> Error):

```
	GrandpaJustification::<B>::decode_and_verify_finalizes(
		&justification,
		(hash, number),
		set_id,
		voters,
	)?;

	Ok(())
```

HCastano: Ideally yes, but the error types don't implement that :P. I don't really want to implement it since at the end of the day I don't actually care about the error that comes from `decode_and_verify_finalizes`, I just want to know that an error happened.

niklasad1: ok, then I guess you could chain `map(|_| ()).map_err(|_| Error::InvalidFinalityProof)` but it is almost uglier :P

HCastano:
There are a couple of things that I want to do in later PRs:

1. Move the `justification` code into `finality-grandpa/primitives`
2. Properly handle authority set changes

I think it's fine to get this PR in without these two things. Once this is merged I can then go ahead and rebase the whole bridge branch past `the-big-reorg` which will make my life easier when dealing with (1). As for (2), I need to do a bit more digging before handling it.

srml/bridge/Cargo.toml
```diff
@@ -8,9 +8,15 @@ edition = "2018"
 
 [dependencies]
 codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false }
+client = { package = "substrate-client", path = "../../core/client" }
```
jimpo: You can't import `client`, `std` cannot be disabled.

Also, you can't import `state-machine` below.

srml/bridge/Cargo.toml
```diff
@@ -8,9 +8,15 @@ edition = "2018"
...
+fg = { package = "substrate-finality-grandpa", path = "../../core/finality-grandpa/" }
```
jimpo: You can't import this.

srml/bridge/Cargo.toml
```diff
@@ -8,9 +8,15 @@ edition = "2018"
...
 fg_primitives = { package = "substrate-finality-grandpa-primitives", path = "../../core/finality-grandpa/primitives" }
+grandpa = { package = "finality-grandpa", version = "0.9.1", features = ["derive-codec"] }
```
jimpo: need `default-features = false`.

srml/bridge/Cargo.toml
```diff
@@ -8,9 +8,15 @@ edition = "2018"
 
 [dependencies]
 codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false }
+client = { package = "substrate-client", path = "../../core/client" }
+fg = { package = "substrate-finality-grandpa", path = "../../core/finality-grandpa/" }
 fg_primitives = { package = "substrate-finality-grandpa-primitives", path = "../../core/finality-grandpa/primitives" }
```
jimpo: needs `default-features = false`

srml/bridge/Cargo.toml
```diff
@@ -8,9 +8,15 @@ edition = "2018"
...
+grandpa = { package = "finality-grandpa", version = "0.9.1", features = ["derive-codec"] }
 hash-db = { version = "0.15.2", default-features = false }
 primitives = { package = "substrate-primitives",  path = "../../core/primitives" }
```
jimpo: needs `default-features = false`

srml/bridge/src/justification.rs
```diff
@@ -0,0 +1,296 @@
...
+	/// ancestry proofs finalize the given block.
+	pub(crate) fn decode_and_verify_finalizes(
+		encoded: &[u8],
+		finalized_target: (Block::Hash, NumberFor<Block>),
+		set_id: u64,
+		voters: &VoterSet<AuthorityId>,
+	) -> Result<GrandpaJustification<Block>, ClientError> where
+		NumberFor<Block>: grandpa::BlockNumberOps,
+	{
+
```
jimpo: nit: remove newline

srml/bridge/src/justification.rs
```diff
@@ -0,0 +1,296 @@
...
+	}
+}
+
+impl<Block: BlockT> PartialOrd for BlockHashKey<Block> {
+	fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+		Some(self.0.as_ref().cmp(other.0.as_ref()))
+	}
+}
+
+impl<Block: BlockT> PartialEq for BlockHashKey<Block> {
```
jimpo: Just `derive(PartialEq)`.

srml/bridge/src/justification.rs
```diff
@@ -0,0 +1,296 @@
...
+
+		Ok(route)
+	}
+
+	fn best_chain_containing(&self, _block: Block::Hash) -> Option<(Block::Hash, NumberFor<Block>)> {
+		None
+	}
+}
+
+pub(crate) fn localized_payload<E: Encode>(round: RoundNumber, set_id: SetIdNumber, message: &E) -> Vec<u8> {
```
jimpo: drop the `pub(crate)`?

srml/bridge/src/lib.rs
```diff
@@ -49,36 +58,34 @@ use system::{ensure_signed};
...
 			current_validator_set: validator_set,
 		}
 	}
 }
 
 type BridgeId = u64;
 
-pub trait Trait: system::Trait {}
+pub trait Trait: system::Trait<Hash=H256> {
+	type Block: BlockT<Hash=H256, Header=Self::Header>;
```
jimpo: nit: I think you can drop the `Hash=H256`, as it is implied by the `Header=Self::Header`.

HCastano: Looks like the compiler isn't happy without the `Hash=H256`.

HCastano:
I think it's going to be a little annoying to get rid of the `client` and `core/finality-grandpa` dependencies. It looks like I'm going to have to move a few things into `finality-grandpa/primitives`, but it's not totally clear to me how just yet.

HCastano:
I think some of the `#[cfg(test)]` stuff looks a little gross, but we should be able to get rid of it once we move the code into `primitives/finality-grandpa`

srml/bridge/Cargo.toml
HCastano: I'm gonna wait on #3868 before getting rid of it

srml/bridge/src/lib.rs
```diff
@@ -430,8 +524,177 @@ mod tests {
 		invalid_proof.insert(5, fake_ancestor);
 
 		assert_err!(
-			verify_ancestry(invalid_proof, ancestor.hash(), child),
+			verify_ancestry(invalid_proof, ancestor.hash(), &child),
 			Error::InvalidAncestryProof
 		);
 	}
+
+	// Currently stealing this from `core/finality-grandpa/src/test.rs`
```
jimpo: Do you expect it to be replaced? Otherwise, don't say "Currently".

srml/bridge/src/storage_proof.rs
```diff
@@ -17,11 +17,12 @@
 //! Logic for checking Substrate storage proofs.
 
 use hash_db::{Hasher, HashDB, EMPTY_PREFIX};
-use state_machine::StorageProof;
 use trie::{MemoryDB, Trie, trie_types::TrieDB};
 
 use crate::Error;
 
+pub(crate) type StorageProof = Vec<Vec<u8>>;
```
jimpo: `pub`

HCastano:
I talked with Jim about some of the options for moving this forward. Since there's a lot of code that came from `core/finality-grandpa` we could either

1) Merge this code as is and refactor it later
2) Put this on ice, and make a new PR once #3868 is merged

We think that the best way forward is (2) as it'll allow us to have a cleaner implementation that doesn't duplicate Grandpa code.

HCastano:
Closing in favour of paritytech/parity-bridges-common#19.

PR #4169, opened 2019-11-21
Introduce efficient Hash-based RNG streamer

gavofyork:
This is a simple generic module that lets you take a `Hash`'s `Output` type (conveniently give by the `Randomness` trait) and turn it into an infinite sequence of numbers, as cryptographically secure as your `Hash` algorithm and the seed, efficiently using the entropy to minimise the number of additional hashes done in the case that the original bytes were insufficient.


primitives/sr-primitives/src/random_number_generator.rs
```diff
@@ -0,0 +1,92 @@
...
+/// let random_seed = BlakeTwo256::hash(b"Sixty-nine");
+/// let mut rng = <RandomNumberGenerator<BlakeTwo256>>::new(random_seed);
+/// assert_eq!(rng.pick_u32(100), 47);
+/// assert_eq!(rng.pick_item(&[1, 2, 3]), Some(&2));
+/// ```
+///
+/// This can use any cryptographic `Hash` function as the means of entropy-extension, and avoids
+/// needless extensions of entropy.
+#[derive(Encode, Decode)]
+pub struct RandomNumberGenerator<Hashing: Hash> {
```
rphmeier: A `Pseudo` qualifier in the name or at the very least a note in the docs to clarify that this is deterministic based on seed and hash function would help this avoid becoming a footgun

primitives/sr-primitives/src/random_number_generator.rs
```diff
@@ -0,0 +1,92 @@
...
+	current: Hashing::Output,
+	offset: u32,
+}
```
rphmeier: This PRNG has an expected periodicity of ~`2^(bits(n)/2)` due to the birthday problem. For a 256-bit hash, we'd expect to reach a cycle after about 2^128 hashes - that's fine, but for a 64-bit or 80-bit hash it may be less so. I think this should be documented with a suggestion not to use short hashes.

primitives/sr-primitives/src/random_number_generator.rs
```diff
@@ -0,0 +1,92 @@
...
+		if self.offset() + needed > self.current.as_ref().len() {
+			// rehash
+			self.current = Hashing::hash(self.current.as_ref());
+			self.offset = 0;
+		}
+		let data = &self.current.as_ref()[self.offset()..self.offset() + needed];
+		self.offset += needed as u32;
+		let raw = u32::decode(&mut TrailingZeroInput::new(data)).unwrap_or(0);
+		if max < u32::max_value() {
+			raw % (max + 1)
```
rphmeier: for small values of `max`, this introduces bias. e.g. `max = 100`. We need to draw 1 byte.
```rust
0..100 % 100 => 0..100
100..200 % 100 => 0..100
200..256 => 0..56
```

So `0..56` have 3 possible drawings where `56..256` only have 2.

rphmeier: the typical thing to do here is loop when the value is out-of-bounds

rphmeier:
LGTM other than review comments

burdges:
It should run faster if you use ChaCha: 
```
use rand_core::{RngCore,SeedableRng};
use rand_chacha::ChaChaRng;

impl H256 {
    fn make_rng(&self) -> ChaChaRng {
        ChaChaRng::from_seed(self.0)
    }
}
```
I donno if you want to minimize rand crate dependencies here or something.

rphmeier:
@burdges I think because `ChaChaRng20` doesn't provide a way to marshall, which we would want.

primitives/sr-primitives/src/random_number_generator.rs
gavofyork: Even a 64-bit hash would result in a birthday only after ~2^32 bits (= 512 MB) have been drained. Given that this happens within runtime logic and generally within the same block, if you're pulling upwards of 1GB of data out of a single hash in one block, you probably have other problems to worry about than the possibility of a repeated sequence.

rphmeier: I wouldn't use it like that, but I wouldn't be surprised if someone uses it in an offchain worker.

gavofyork: (I put a note in there anyway)

primitives/sr-primitives/src/random_number_generator.rs
gavofyork: addressed.

burdges:
Yes, folks rarely expose the ChaCha state for marshalling, not even sure all it's security properties hold if an adversary can control the whole ChaCha state.

primitives/sr-primitives/src/random_number_generator.rs
```diff
@@ -0,0 +1,103 @@
...
+			offset: 0,
+		}
+	}
+
+	fn offset(&self) -> usize { self.offset as usize }
+
+	/// Returns a number at least zero, at most `max`.
+	pub fn pick_u32(&mut self, max: u32) -> u32 {
+		let needed = (4 - max.leading_zeros() / 8) as usize;
+		let top = ((1 << (needed as u64 * 8)) / ((max + 1) as u64) * ((max + 1) as u64) - 1) as u32;
```
rphmeier: what's the significance of `2^needed_bits / ((max + 1)^2 - 1)`?

gavofyork:
we don't get uniform sampling, actually, since `rand_distr` doesn't support `no_std`.

primitives/sr-primitives/src/random_number_generator.rs
gavofyork: well, it's part of the wider expression which, due to integer division, gives you the top of the non-biased range.

Essentially: `max_unbiased := floor(raw_range / target_range) * target_range`, then just loop until the value is in the unbiased area.

gavofyork:
I moved what I was using this for over to `ChaCha`, but I still think this code is useful since a) I don't know how much bloat `rand_chacha` and `rand_core` is going to pull in to the wasm build; and b) we don't get `rand_distr` because of lack of `no_std` support therefore this code is strictly more useful for drawing from distributions.

burdges:
I missed that `Rng` provides only a thin wrapper around `rand::distributions` now.  I think the `Distribution` trait and [Uniform](https://docs.rs/rand/0.7.2/rand/distributions/struct.Uniform.html) exist in rand itself, but..

We're well into rand's churn territory here, so maybe this depends upon the version, and maybe even rand 0.7 should not be considered stable enough yet.

gavofyork:
> I missed that Rng provides only a thin wrapper around rand::distributions now

I missed that it provided anything at all outside of `rand_distr`! :)

PR #4170, opened 2019-11-21
Society pallet

gavofyork:
Society is basically a membership provider where member-candidates are voted in by existing members. Candidacy/entrance requirements can be put in place, as well as ongoing random verification requirements. A suspension & appeals process exists for when candidates/members fail to make or retain the approval of the membership.

TODO:
- [x] Challenges
- [x] Suspension
- [x] Remove linear vesting and just have a multi-cliff system to make slash-rewards less problematic. 
- [x] Tests


frame/society/src/lib.rs
```diff
@@ -0,0 +1,652 @@
...
+			let who = ensure_signed(origin)?;
+
+			let pos = pos as usize;
+			<Bids<T>>::mutate(|b|
+				if pos < b.len() && b[pos].1 == who {
+					b.remove(pos);
+					Self::deposit_event(RawEvent::Unbid(who));
+					Ok(())
+				} else {
+					Err("bad position")
```
niklasad1: This error message isn't great because it will be `bad position` if `who` is not bidder (i.e, not in the vector)

So I would suggest two different errors instead.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,671 @@
...
+		for skeptic in (0..members.len().integer_sqrt()).map(pick_member) {
+			for (_, c, _) in candidates.iter() {
+				<Votes<T>>::insert(c, skeptic, Vote::Skeptic);
+			}
+		}
+	}
+
+	/// Attempt to slash the payout of some member. Return the payout block number most in the
+	/// future along with the total amount deducted.
+	fn slash_payout(who: &T::AccountId, value: BalanceOf<T>) -> (T::BlockNumber, BalanceOf<T>) {
```
shawntabrizi: It is not 100% clear what the returned block number should be here, and what it will be used for. (not currently used in the module)

Right now, it will return the block number of the _first_ payout which still has funds in it. This is not to me "the most in the future".

frame/society/src/lib.rs
gavofyork: meh - users shouldn't be calling into this directly so really shouldn't be an issue. if it can be done for zero extra cost, sure, but i don't think it can...

frame/society/src/lib.rs
gavofyork: it used to be used and perhaps it'll be used in the future... can be removed if the PR is complete and it's still unused though.

frame/society/src/lib.rs
niklasad1: Right, feel free to ignore me then but a comment about it would good so it won't be interpreted as a bug by future readers.

FWIW, I think rust handles `&static str` just as a "string constant" (stored in text/data segment) and it will not have any additional runtime cost (except copy it from flash to RAM at startup) but it will produce a slightly bigger binary which might be an issue for WASM.

gavofyork: i mean for finding the alternative condition, not for the extra string constant.

frame/society/src/tests.rs
```diff
@@ -24,25 +24,43 @@ use support::{assert_ok, assert_noop};
 #[test]
 fn founding_works() {
 	EnvBuilder::new().with_members(vec![]).execute(|| {
+		// Account 1 is set as the founder origin and can start a society
+		// This allows the module to bootstrap on an already running chain
```
gavofyork: might be a point worth making, but here isn't the right place.

frame/society/src/lib.rs
```diff
@@ -484,31 +492,33 @@ impl<T: Trait> Module<T> {
...
-							// really shouldn't fail given the conditional we're in.
+							// Really shouldn't fail given the conditional we're in.
 							let _ = Self::unset_vouching(&who);
 							value.saturating_sub(tip)
 						}
 					};
 
 					Self::bump_payout(&candidate, maturity, value);
 
+					// We track here the total_approvals so that every user has a unique range
```
gavofyork: "every user"?

gavofyork: i think you mean "every voter for each candidate".

frame/society/src/lib.rs
```diff
@@ -537,15 +549,17 @@ impl<T: Trait> Module<T> {
 
 			// if at least one candidate was accepted...
 			if !accepted.is_empty() {
-				// select one as primary, randomly chosen from the accepted, weighted by approvals.
```
gavofyork: don't remove this!! it's describing what we're trying to do.

frame/society/src/lib.rs
```diff
@@ -297,6 +302,18 @@ decl_module! {
 			Self::deposit_event(RawEvent::Founded(founder));
 		}
 
+		/// Allow founder origin to make judgement on a suspended member.
+		fn judge_suspended_member(origin, who: T::AccountId, forgive: bool) {
+			T::FounderOrigin::ensure_origin(origin)?;
```
gavofyork: Should be some other origin

frame/society/src/lib.rs
```diff
@@ -415,14 +418,38 @@ impl<T: Trait> Module<T> {
 		)
 	}
 
-	/// Remove a member from the members list.
+	/// Check a user is a members.
```
gavofyork: ... is a member.

frame/society/src/lib.rs
```diff
@@ -623,8 +650,35 @@ impl<T: Trait> Module<T> {
...
+			// Increase payout time of existing payouts by the suspension time.
+			<Payouts<T>>::mutate(&who, |payouts| {
+				payouts.iter_mut().map(|v| v.0 += suspended_period).collect::<Vec<_>>()
+			});
+			<SuspendedMembers<T>>::remove(who);
+		}
+	}
+
+	/// Remove all information about this user from this module.
+	fn remove_user(who: &T::AccountId) {
```
gavofyork: accounts are either candidates, members or voters. there are no "users".

frame/society/src/lib.rs
```diff
@@ -161,7 +161,7 @@ decl_storage! {
 		}): Vec<T::AccountId>;
 
 		/// The set of suspended members.
-		pub SuspendedMembers get(fn suspended_member): map T::AccountId => Option<T::BlockNumber>;
+		pub SuspendedMembers get(fn suspended_member): linked_map T::AccountId => Option<T::BlockNumber>;
```
gavofyork: we don't use linked maps any more. a `map` should be fine; if you need to enumerate keys from the UI then either store them in the value, or use the `twox_128_concat` hasher.

frame/society/src/lib.rs
```diff
@@ -146,6 +146,10 @@ decl_storage! {
 		pub Candidates get(candidates):
 			Vec<(BalanceOf<T>, T::AccountId, BidKind<T::AccountId, BalanceOf<T>>)>;
 
+		/// The set of suspended candidates.
+		pub SuspendedCandidates get(suspended_candidates):
+			linked_map T::AccountId => Option<(BalanceOf<T>, BidKind<T::AccountId, BalanceOf<T>>)>;
```
gavofyork: No linked maps, please.

frame/society/src/lib.rs
```diff
@@ -314,6 +321,35 @@ decl_module! {
 			}
 		}
 
+		/// Allow founder origin to make judgement on a suspended candidate.
+		fn judge_suspended_candidate(origin, who: T::AccountId, approve: Option<bool>) {
+			T::FounderOrigin::ensure_origin(origin)?;
```
gavofyork: don't coalesce origins without good reason.

frame/society/src/lib.rs
```diff
@@ -470,8 +505,7 @@ impl<T: Trait> Module<T> {
 			let candidates = <Candidates<T>>::take();
 			members.reserve(candidates.len());
 
-			let maturity = <system::Module<T>>::block_number()
-				+ Self::lock_duration(members.len() as u32);
+			let maturity = Self::maturity();
```
gavofyork: well done; you've just managed to double the time we spend decoding members.

gavofyork: don't split out storage accesses into low-level getters without thinking.

frame/society/src/lib.rs
```diff
@@ -681,9 +699,40 @@ impl<T: Trait> Module<T> {
...
-	/// `kind` lets you know whether there's either a deposit at stake or a member's vouching rights.
-	fn suspend_candidate(_who: &T::AccountId, _kind: BidKind<T::AccountId, BalanceOf<T>>) {
-//		unimplemented!();
+	/// Pay an accepted candidate their bid value.
+	fn pay_accepted_candidate(
+		candidate: &T::AccountId,
+		value: BalanceOf<T>,
+		kind: BidKind<T::AccountId, BalanceOf<T>>
+	) {
+		let maturity = Self::maturity();
```
gavofyork: wow.
amazing. you've turned decoding the members from an `1x` in my original code to `(N+2)x` in your code. this is obviously brain-dead.

frame/society/src/lib.rs
shawntabrizi: The takeaway for others here is:

> helpers shouldn't be pulling in storage; in general they shouldn't be using globals at all. it's forgivable if all conceivable instances of usage would all use those globals.

This pattern should be easy to identify upon review

frame/society/src/lib.rs
```diff
@@ -0,0 +1,792 @@
...
+				value.saturating_sub(tip)
+			}
+		};
+
+		Self::bump_payout(candidate, maturity, value);
+	}
+
+	/// Remove a suspended candidate. Slash their deposit or give strike to their voucher.
+	fn slash_suspended_candidate(who: &T::AccountId) {
+		if let Some((_, kind)) = <SuspendedCandidates<T>>::take(who) {
```
gavofyork: this is an unnecessary pessimisation. just pass in the value at call and use `::remove` instead.

shawntabrizi: I got rid of this function and just moved the logic into the code. Thanks for the sharp eyes.

frame/society/src/lib.rs
niklasad1: ahh, alright I think it would require an extra branch

shawntabrizi:
I still need to add quality pallet documentation to the top of this, but it _should_ be ready for code review.

Different from the spec:
* A failed voucher results in a strike rather than inability to vouch ever again.
* Challenges happen in a single period rather than split between two periods (with a response period). Every challenge period, a new random user from the member pool will be selected for a challenge.
* Suspensions opt for a `SuspensionJudgementOrigin` to decide the outcome of suspended candidates and members rather than an appeals process.

If any of these changes are not acceptable, please just let me know and I can follow up.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,894 @@
...
+				let primary = accepted.iter().find(|e| e.1 > primary_point)
+					.expect("e.1 of final item == total_approvals; \
+						worst case find will always return that item; qed")
+					.0.clone();
+				
+				let accounts = accepted.into_iter().map(|x| x.0).collect::<Vec<_>>();
+
+				// Then write everything back out, signal the changed membership and leave an event.
+				members.sort();
+				<Members<T, I>>::put(members.clone());
```
shawntabrizi: This might be really stupid. I don't understand the relative costs between cloning a large vec like `members` versus reading it twice.

It seems I would need to do one of these two choices. Please let me know which is right.

thiolliere: can't the following works 
```suggestion
				<Members<T, I>>::put(&members);
```
With EncodeLike mecanism you should be able to avoid most clone to write to storage now

shawntabrizi: I get the following error

```
error[E0277]: the trait bound `&&mut std::vec::Vec<<T as frame_system::Trait>::AccountId>: _::_parity_scale_codec::EncodeLike<std::vec::Vec<<T as frame_system::Trait>::AccountId>>` is not satisfied
   --> frame/society/src/lib.rs:702:26
    |
702 |                 <Members<T, I>>::put(&members);
    |                                      -^^^^^^^
    |                                      |
    |                                      the trait `_::_parity_scale_codec::EncodeLike<std::vec::Vec<<T as frame_system::Trait>::AccountId>>` is not implemented for `&&mut std::vec::Vec<<T as frame_system::Trait>::AccountId>`
    |                                      help: consider removing 1 leading `&`-references
    |
    = help: the following implementations were found:
              <std::vec::Vec<T> as _::_parity_scale_codec::EncodeLike<&[U]>>
              <std::vec::Vec<T> as _::_parity_scale_codec::EncodeLike<std::vec::Vec<U>>>
    = note: required by `sp_api_hidden_includes_decl_storage::hidden_include::StorageValue::put`

Comment (thiolliere) [reply frame/society/src/lib.rs]: hmm yes ok, indeed you can do
```suggestion
				<Members<T, I>>::put(codec::Ref::from(&members));
```

or 
```suggestion
				<Members<T, I>>::put(&members[..]);
```

frame/society/src/lib.rs
```diff
@@ -0,0 +1,891 @@
...
+
+	/// Something that provides randomness in the runtime.
+	type Randomness: Randomness<Self::Hash>;
+
+	/// The minimum amount of a deposit required for a bid to be made.
+	type CandidateDeposit: Get<BalanceOf<Self, I>>;
+
+	/// The proportion of the unpaid reward that gets deducted in the case that either a skeptic
+	/// doesn't vote or someone votes in the wrong way.
+	type WrongSideDeduction: Get<BalanceOf<Self, I>>;
```
shawntabrizi: Should this be a percentage?

> The proportion of the unpaid reward...

kianenigma:
also needs weight annotation; start with the basis of min (cheapest tx weight in the entire system) being 10_000 and build on top of that, for the time being.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1035 @@
...
+		/// The number of blocks between membership challenges.
+		const ChallengePeriod: T::BlockNumber = T::ChallengePeriod::get();
+
+		// Used for handling module events.
+		fn deposit_event() = default;
+
+		/// Make a bid for entry.
+		///
+		/// Alters Bids (O(N) decode+encode, O(logN) search, <=2*O(N) write).
+		pub fn bid(origin, value: BalanceOf<T, I>) -> Result {
```
shawntabrizi: I am worried that we do not check that a user may already be an existing bid. This would cause them to re-reserve some funds.

These reserved funds would not have any process to then be unreserved.

However, checking this might be expensive since we need to search the bids for an existing bid.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1148 @@
...
+		}): Vec<T::AccountId>;
+
+		/// The set of suspended members.
+		pub SuspendedMembers get(fn suspended_member): map T::AccountId => Option<()>;
+
+		/// The current bids.
+		Bids: Vec<(BalanceOf<T, I>, T::AccountId, BidKind<T::AccountId, BalanceOf<T, I>>)>;
+
+		/// Members currently vouching, stored ordered
+		Vouching: Vec<T::AccountId>;
```
shawntabrizi: @gavofyork correct me if I am wrong here, but I see no reason for `Vouching` to be a `Vec`.

We only ever add and remove from it a single user at a time.

Reading all the vouchers from storage and doing a binary search to add/remove each time seems way less efficient than simply storing a `map T::AccountId => Option<()>` to keep track of this.

gavofyork:
> Different from the spec:
> * A failed voucher results in a strike rather than inability to vouch ever again.

Please revert back to spec.

frame/society/src/lib.rs
gavofyork: no

frame/society/src/lib.rs
gavofyork: that very much remains to be seen, and i suspect empirical testing will be the only way to find out. for now i'm assuming a single DB read of (likely) less than 1KB followed by a binary chop search and insert will be more efficient overall than using the main trie for 30-40 items. i might be wrong though.

shawntabrizi: Since tries are prefixed now by module and storage item, the extra nodes under this branch for a map wont affect performance of any other module or storage item query... right?

frame/society/src/lib.rs
shawntabrizi: removed

frame/society/src/lib.rs
shawntabrizi: Since tries are prefixed now by module and storage item, the extra nodes under a branch for a storage map should not affect performance of any other module or storage item queries.

So in this case, it makes sense to use a map.

Will follow up with benchmarks after the holidays.

frame/society/src/lib.rs
```diff
@@ -623,8 +650,35 @@ impl<T: Trait> Module<T> {
...
+		}
+	}
+
+	/// Remove all information about this user from this module.
+	fn remove_user(who: &T::AccountId) {
+		let _ = Self::remove_member(who);
+		<Payouts<T>>::remove(who);
+		<Strikes<T>>::remove(who);
+		<SuspendedMembers<T>>::remove(who);
+		// TODO: Clean up vouching
```
gavofyork: stray TODO

frame/society/src/lib.rs
gavofyork: avoid linked_map - use blake2_128_concat or twox64_concat (if the `AccountId` only comes from a signed origin).

frame/society/src/lib.rs
gavofyork: avoid linked map.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// - `pos`: Position in the `Bids` vector of the bid who should be unvouched.
+		///
+		/// # <weight>
+		/// - One storage read to check the signer is a vouching member.
+		/// - One storage read and write to retrieve and update the bids.
+		/// - A vector lookup to check if the user is at the specified position.
+		/// - One vouching storage removal.
+		/// - One event.
+		/// # </weight>
+		///
```
gavofyork: blank lines.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		///              approved (`true`) or rejected (`false`).
+		///
+		/// # <weight>
+		/// - One storage read and O(log M) search to check user is a member.
+		/// - One storage read to lookup candidate index.
+		/// - One storage read and O(C) search to check that user is a candidate.
+		/// - One storage write to add vote to votes.
+		/// - One event.
+		/// # </weight>
+		///
```
gavofyork: blank lines

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// - `pos`: Position in the `Bids` vector of the bid who wants to unbid.
+		///
+		/// # <weight>
+		/// - One storage read and write to retrieve and update the bids.
+		/// - A vector lookup to check if the user is at the specified position.
+		/// - Up to one unreserve balance action.
+		/// - Up to vouching storage removal.
+		/// - One event.
+		/// # </weight>
+		///
```
gavofyork: blank lines.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// 	- One storage read to retrieve all current bids.
+		/// 	- One storage read to retrieve all current candidates.
+		/// 	- One storage read to retrieve all members.
+		/// - O(B + C + log M) to check user is not already a part of society.
+		/// - External Module Operations:
+		/// 	- One balance reserve operation.
+		/// - Write Operations:
+		/// 	- 
+		/// - One event.
+		/// # </weight>
```
gavofyork: blank lines, and a stray comment line.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		///
+		/// - `approve`: A boolean which says if the candidate should be
+		///              approved (`true`) or rejected (`false`).
+		///
+		/// # <weight>
+		/// - One storage read and O(log M) search to check user is a member.
+		/// - One storage write to add vote to votes.
+		/// - One event.
+		/// # </weight>
+		///
```
gavofyork: blank lines

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// The dispatch origin for this call must be _Signed_.
+		///
+		/// # <weight>
+		/// - One storage read and O(log M) search to check signer is a member.
+		/// - One storage read to get all payouts for a member.
+		/// - One storage read to get the current block number.
+		/// - Up to one currency transfer call.
+		/// - One storage write or removal to update the member's payouts.
+		/// # </weight>
+		///
```
gavofyork: blank lines

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// # <weight>
+		/// - One storage read to check `Head`.
+		/// - One storage write to add the first member to society.
+		/// - One storage write to add new Head.
+		/// - One event.
+		///
+		/// This is marked as `Operational` since it is a low level system action.
+		/// # </weight>
+		///
```
gavofyork: blank lines.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// - Up to one storage write with O(log M) binary search to add a member back to society.
+		/// - Up to 3 storage removals to clean up a removed member.
+		/// - Up to one storage write with O(B) search to remove vouched bid from bids.
+		/// - Up to one additional event if unvouch takes place.
+		/// - One storage removal.
+		/// - One event for the judgement.
+		///
+		/// This is marked as `Operational` since it comes from a privileged origin.
+		/// # </weight>
+		///
```
gavofyork: blank lines.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// 	- Up to one storage write to ban the vouching member from vouching again.
+		/// - Rebid Logic
+		/// 	- Storage mutate with O(log B) binary search to place the user back into bids.
+		/// - Up to one additional event if unvouch takes place.
+		/// - One storage removal.
+		/// - One event for the judgement.
+		///
+		/// This is marked as `Operational` since it comes from a privileged origin.
+		/// # </weight>
+		///
```
gavofyork: blank lines.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+
+				// Remove suspended candidate
+				<SuspendedCandidates<T, I>>::remove(who);
+			} else {
+				Err(Error::<T, I>::NotSuspended)?
+			}
+		}
+
+		fn on_initialize(n: T::BlockNumber) {
+
```
gavofyork: blank line.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// 	- One storage read to check for suspended candidate.
+		/// 	- One storage read to check for suspended member.
+		/// 	- One storage read to retrieve all current bids.
+		/// 	- One storage read to retrieve all current candidates.
+		/// 	- One storage read to retrieve all members.
+		/// - O(B + C + log M) to check user is not already a part of society.
+		/// - External Module Operations:
+		/// 	- One balance reserve operation.
+		/// - Write Operations:
+		/// 	-
```
gavofyork: empty line.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+
+			let deposit = T::CandidateDeposit::get();
+			T::Currency::reserve(&who, deposit)?;
+
+			Self::put_bid(who.clone(), value.clone(), BidKind::Deposit(deposit));
+			Self::deposit_event(RawEvent::Bid(who, value));
+			Ok(())
+		}
+
+		/// A bid can remove their bid for entry into society.
```
gavofyork:
```suggestion
		/// A bidder can remove their bid for entry into society.
```

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		///
+		/// By doing so, they will have their candidate deposit returned or
```
gavofyork: shouldn't be on its own paragraph.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// they will unvouch their voucher.
+		///
+		/// The dispatch origin for this call must be _Signed_.
```
gavofyork: Should mention that the signing account must be a bidder.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+			ensure!(!Self::is_candidate(&who), Error::<T, I>::AlreadyCandidate);
+			ensure!(!Self::is_member(&who), Error::<T, I>::AlreadyMember);
+
+			<Vouching<T, I>>::insert(&voucher, VouchingStatus::Vouching);
+			Self::put_bid(who.clone(), value.clone(), BidKind::Vouch(voucher.clone(), tip));
+			Self::deposit_event(RawEvent::Vouch(who, value, voucher));
+			Ok(())
+		}
+
+		/// As a vouching member, unvouch a bid. This only works while vouched user is still a bid.
```
gavofyork:
```suggestion
		/// As a vouching member, unvouch a bid. This only works while vouched user is still only a bidder (and not a candidate).
```

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		///
+		/// The dispatch origin for this call must be _Signed_.
```
gavofyork: ...and the signed origin should be a current bidder.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		///
+		/// Parameters:
+		///
+		/// - `pos`: Position in the `Bids` vector of the bid who should be unvouched.
+		///
+		/// # <weight>
```
gavofyork: the weights don't mention complexity of storage accesses.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+					Ok(())
+				} else {
+					Err(Error::<T, I>::BadPosition)?
+				}
+			)
+		}
+
+		/// As a member, vote on a candidate.
+		///
+		/// The dispatch origin for this call must be _Signed_.
```
gavofyork: needs to mention that the signing account must be a member.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		///
+		/// Parameters:
+		///
+		/// - `candidate`: The candidate that the member would like to bid on.
+		/// - `approve`: A boolean which says if the candidate should be
+		///              approved (`true`) or rejected (`false`).
+		///
+		/// # <weight>
```
gavofyork: doesn't mention storage access complexity.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+
+			Self::deposit_event(RawEvent::Vote(candidate, voter, approve));
+		}
+
+		/// As a member, vote on the defender.
+		///
+		/// The dispatch origin for this call must be _Signed_.
+		///
+		/// Parameters:
+		///
```
gavofyork: blank line.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+
+			let vote = if approve { Vote::Approve } else { Vote::Reject };
+			<Votes<T, I>>::insert(&candidate, &voter, vote);
+
+			Self::deposit_event(RawEvent::Vote(candidate, voter, approve));
+		}
+
+		/// As a member, vote on the defender.
+		///
+		/// The dispatch origin for this call must be _Signed_.
```
gavofyork: ...and a member.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+			<DefenderVotes<T, I>>::insert(&voter, vote);
+
+			Self::deposit_event(RawEvent::DefenderVote(voter, approve));
+		}
+
+		/// Transfer the first matured payout for the sender and remove it from the records.
+		///
+		/// This extrinsic can be called multiple times to claim all matured payouts.
+		///
+		/// The dispatch origin for this call must be _Signed_.
```
gavofyork: ...and a member.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		///
+		/// # <weight>
+		/// - One storage read and O(log M) search to check signer is a member.
+		/// - One storage read to get all payouts for a member.
+		/// - One storage read to get the current block number.
+		/// - Up to one currency transfer call.
```
gavofyork:
```suggestion
		/// - One currency transfer call.
```

describe what happens on success. if this call results in an `Ok` return, then it'll have exactly one balance operation.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+
+		/// A bid can remove their bid for entry into society.
+		///
+		/// By doing so, they will have their candidate deposit returned or
+		/// they will unvouch their voucher.
+		///
+		/// The dispatch origin for this call must be _Signed_.
+		///
+		/// Parameters:
+		///
```
gavofyork: blank line.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+			Self::put_bid(who.clone(), value.clone(), BidKind::Deposit(deposit));
+			Self::deposit_event(RawEvent::Bid(who, value));
+			Ok(())
+		}
+
+		/// A bid can remove their bid for entry into society.
+		///
+		/// By doing so, they will have their candidate deposit returned or
+		/// they will unvouch their voucher.
+		///
```
gavofyork: Should mention the payment it will take here with a `Payment: ` prefixed line. see `identity` module for examples.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+					return Ok(())
+				}
+			}
+			Err(Error::<T, I>::NoPayout)?
+		}
+
+		/// Found the society.
+		///
+		/// This is done as a discrete action in order to allow for the
+		/// module to be included into a running chain.
```
gavofyork: Mention that the society cannot be founded twice.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		///
+		/// The dispatch origin for this call must be from the _FounderOrigin_.
+		///
+		/// Parameters:
+		///
+		/// - `founder` - The first member and head of the newly founded society.
+		///
+		/// # <weight>
```
gavofyork: needs to mention complexity of each storage access.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// - One storage read to check `Head`.
+		/// - One storage write to add the first member to society.
+		/// - One storage write to add new Head.
+		/// - One event.
+		///
+		/// This is marked as `Operational` since it is a low level system action.
```
gavofyork: Incorrect. It's nowhere near "low-level" enough to be operational. Operational stuff is for extrinsics that are critical to the security of the network.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		#[weight = SimpleDispatchInfo::FixedOperational(10_000)]
+		fn found(origin, founder: T::AccountId) {
+			T::FounderOrigin::ensure_origin(origin)?;
+			ensure!(!<Head<T, I>>::exists(), Error::<T, I>::AlreadyFounded);
+			Self::add_member(&founder)?;
+			<Head<T, I>>::put(&founder);
+			Self::deposit_event(RawEvent::Founded(founder));
+		}
+
+		/// Allow founder origin to make judgement on a suspended member.
```
gavofyork: Incorrect now.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		///
+		/// The dispatch origin for this call must be from the _SuspensionJudgementOrigin_.
+		///
+		/// Parameters:
+		///
+		/// - `who` - The suspended member to be judged.
+		/// - `forgive` - A boolean representing whether the suspension judgement origin
+		///               forgives (`true`) or rejects (`false`) a suspended member.
+		///
+		/// # <weight>
```
gavofyork: Misses storage access complexity.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// - One storage read to check `who` is a suspended member.
+		/// - Up to one storage write with O(log M) binary search to add a member back to society.
+		/// - Up to 3 storage removals to clean up a removed member.
+		/// - Up to one storage write with O(B) search to remove vouched bid from bids.
+		/// - Up to one additional event if unvouch takes place.
+		/// - One storage removal.
+		/// - One event for the judgement.
+		///
+		/// This is marked as `Operational` since it comes from a privileged origin.
```
gavofyork: Incorrect. None of this module is even remotely security-critical to the network.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+					);
+				}
+			}
+
+			<SuspendedMembers<T, I>>::remove(&who);
+			Self::deposit_event(RawEvent::SuspendedMemberJudgement(who, forgive));
+		}
+
+
+		/// Allow founder origin to make judgement on a suspended candidate.
```
gavofyork: Incorrect.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		/// - Reject Logic
+		/// 	- Up to one repatriate reserved currency action.
+		/// 	- Up to one storage write to ban the vouching member from vouching again.
+		/// - Rebid Logic
+		/// 	- Storage mutate with O(log B) binary search to place the user back into bids.
+		/// - Up to one additional event if unvouch takes place.
+		/// - One storage removal.
+		/// - One event for the judgement.
+		///
+		/// This is marked as `Operational` since it comes from a privileged origin.
```
gavofyork: Incorrect.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+		#[weight = SimpleDispatchInfo::FixedOperational(50_000)]
+		fn judge_suspended_candidate(origin, who: T::AccountId, judgement: Judgement) {
+			T::SuspensionJudgementOrigin::ensure_origin(origin)?;
+			if let Some((value, kind)) = <SuspendedCandidates<T, I>>::get(&who) {
+				match judgement {
+					Judgement::Approve => {
+						// Founder origin has approved this candidate
+						// Make sure we can pay them
+						let pot = Self::pot();
+						ensure!(pot > value, Error::<T, I>::InsufficientPot);
```
gavofyork:
```suggestion
						ensure!(pot >= value, Error::<T, I>::InsufficientPot);
```

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+
+	/// Check a user is a member.
+	fn is_member(m: &T::AccountId) -> bool {
+		<Members<T, I>>::get()
+			.binary_search(m)
+			.is_ok()
+	}
+
+	/// Add a member to the sorted members list. Will not add a duplicate member.
+	fn add_member(m: &T::AccountId) -> DispatchResult {
```
gavofyork: This is a dangerous function as it can both return an `Err` and has global side-effects.

Docs should state as much and make it clear that it *must* be at the border point between checks and actions in a dispatch function.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+	fn rotate_challenge(members: &mut Vec<T::AccountId>) {
+		// Assume there are members, else don't run this logic.
+		if !members.is_empty() {
+			// End current defender rotation
+			if let Some(defender) = Self::defender() {
+				let mut approval_count = 0;
+				let mut rejection_count = 0;
+				// Tallies total number of approve and reject votes for the defender.
+				members.iter()
+				.filter_map(|m| <DefenderVotes<T, I>>::get(m))
```
gavofyork: missing indent level

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+				.for_each(|v|{
```
gavofyork:
```suggestion
				.for_each(|v| {
```

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+					match v {
+						Vote::Approve => { approval_count += 1 }
+						_ => {rejection_count += 1}
```
gavofyork:
```suggestion
						_ => rejection_count += 1,
```

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1372 @@
...
+			// End current defender rotation
+			if let Some(defender) = Self::defender() {
+				let mut approval_count = 0;
+				let mut rejection_count = 0;
+				// Tallies total number of approve and reject votes for the defender.
+				members.iter()
+				.filter_map(|m| <DefenderVotes<T, I>>::get(m))
+				.for_each(|v|{
+					match v {
+						Vote::Approve => { approval_count += 1 }
```
gavofyork:
```suggestion
						Vote::Approve => approval_count += 1,
```

frame/society/src/lib.rs
shawntabrizi: Ack. Misunderstood what operational extrinsics should be used for.

frame/society/src/lib.rs
shawntabrizi: Made it so that this function cannot fail. If the user is already a member, we simply do nothing :)

frame/society/src/lib.rs
```diff
@@ -68,6 +68,14 @@
...
 //! module. Some subset of this treasury is placed in a Society Pot, which is used
 //! to determine the number of accepted bids.
+//!
+//! #### Rate of Growth
+//! 
+//! The membership society can grow at a rate of 10 accepted candidates per rotation period up
+//! to the max membership threshold. Once this threshold is met, candidate selections
+//! are stalled until there is space for new members to join. This can be resolved by
+//! voting out existing members through the random challenges or by using governance
+//! to increase the maximum membership count.
```
shawntabrizi: Note the behavior when max membership is reached.

It is possible that we have `max_membership + 10` total users in the member list. This is because the approved candidates have clean way to "fail" being added to the membership pool.

The scenario can happen as so:
* There are 90 active members in society
* There are at least 10 suspended members in society
* There are 10 approved candidates that will make it in during the next rotation period
* The suspension judgement origin adds 10 suspended members back to the society, reaching the limit of 100
* Then the rotation period occurs and 10 candidates are added on top

As mentioned with inline comments, this has no consensus critical issue or side effects. Being over the max or being at the max is exactly the same behavior.

frame/society/src/lib.rs
```diff
@@ -1431,7 +1442,7 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
 						}
 					});
 
-				if approval_count < rejection_count {
+				if approval_count <= rejection_count {
```
shawntabrizi: It makes sense for this to be skewed toward rejection. If the society member cannot even vote for themselves in their own membership challenge, they should be kicked from society.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1520 @@
...
+		/// - External Module Operations:
+		/// 	- One balance reserve operation. O(X)
+		/// 	- Up to one balance unreserve operation if bids.len() > MAX_BID_COUNT.
+		/// - Events:
+		/// 	- One event for vouch.
+		/// 	- Up to one event for AutoUnbid if bid.len() > MAX_BID_COUNT.
+		///
+		/// Total Complexity: O(M + B + C + logM + logB + X)
+		/// # </weight>
+
```
gavofyork: blank line.

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1520 @@
...
+		///
+		/// # <weight>
+		/// Key: B (len of bids), X (balance unreserve)
+		/// - One storage read and write to retrieve and update the bids. O(B)
+		/// - Either one unreserve balance action O(X) or one vouching storage removal. O(1)
+		/// - One event.
+		///
+		/// Total Complexity: O(B + X)
+		/// # </weight>
+
```
gavofyork: blank line

frame/society/src/lib.rs
```diff
@@ -0,0 +1,1520 @@
...
+		/// - External Module Operations:
+		/// 	- One balance reserve operation. O(X)
+		/// 	- Up to one balance unreserve operation if bids.len() > MAX_BID_COUNT.
+		/// - Events:
+		/// 	- One event for new bid.
+		/// 	- Up to one event for AutoUnbid if bid.len() > MAX_BID_COUNT.
+		///
+		/// Total Complexity: O(M + B + C + logM + logB + X)
+		/// # </weight>
+
```
gavofyork: blank line.

gavofyork:
A few blank lines in there, I'll take a more in-depth look at the changed logic tomorrow.

gavofyork:
CI failure unrelated.

PR #4174, opened 2019-11-21
Updating GRANDPA to v0.10.0

rphmeier:
Includes more diagnostics and persists more votes. Also allows for better resuming of completed, but not concluded rounds on startup.


client/finality-grandpa/src/environment.rs
```diff
@@ -815,6 +815,59 @@ where
...
+			target: "afg", "Voter {} concluded round {} in set {}. Estimate = {:?}, Finalized in round = {:?}",
+			self.config.name(),
+			round,
+			self.set_id,
+			state.estimate.as_ref().map(|e| e.1),
+			state.finalized.as_ref().map(|e| e.1),
+		);
+
+		self.update_voter_set_state(|voter_set_state| {
+			// NOTE: we don't use `with_current_round` here, it is possible that
```
andresilva: The reason we can't use `with_current_round` here is because the round being concluded is no longer current (must have been previously completed).

client/finality-grandpa/src/environment.rs
```diff
@@ -815,6 +815,59 @@ where
...
+				};
+
+			let mut completed_rounds = completed_rounds.clone();
+
+			if let Some(already_completed) = completed_rounds.rounds
+				.iter_mut().find(|r| r.number == round)
+			{
+				let n_existing_votes = already_completed.votes.len();
+				already_completed.votes.extend(
+					historical_votes.seen().iter().skip(n_existing_votes).cloned()
```
andresilva: Maybe add a comment here about the contract that `historical_votes` seen in completed must be a prefix of `historical_votes` in concluded.

PR #4175, opened 2019-11-21
Expose some of session module's storage as public, as well as the set_keys transaction

Sushisource:
Seemingly there's no reason not to do this, as anyone can always do it the "hard way" by constructing storage keys or extrinsics.

This was inspired by the fact that a module I'm writing which managers validator sets (without involving staking in any way) needs to be able to check that a new validator has already registered their keys with `set_keys` before allowing them to be added to the set. I had no way to check that without manually constructing storage hashes.


bkchr:
While this proposed solution probably solves your problem, it feels more like a "hack". Why not introduce a trait that does this. It provides a function to check if a validator is already registered and to get the key.

bkchr:
(By doing this, you also do not depend on `Session` and everybody could reduce your module with a different `Session` implementation)

Sushisource:
@bkchr You mean a trait that my module would implement, that the `session` module calls any time someone registers their keys? Strictly speaking you'd still need to depend on it, since it'd be the one defining that trait (and I already depend on it for the other traits like `SessionHandler`), but it would reduce the coupling.

I can do that, but it seems roughly equivalent to just adding an `is_registered` public function that my module could call. How about that?

bkchr:
Yeah `is_registered` sounds good :) But make it a trait directly please :)

Sushisource:
@bkchr Sounds good to me. Just want to clarify what you mean by "a trait directly".

Something like `trait ValidatorRegistration<ValidatorId>` defining `is_registered` which `Session` would then implement?

bkchr:
Yep 👍

frame/session/src/lib.rs
```diff
@@ -333,6 +333,16 @@ impl<V> SelectInitialValidators<V> for () {
 	}
 }
 
+pub trait ValidatorRegistration<ValidatorId> {
```
bkchr: Some docs + moving the trait to `support::traits`, otherwise looks good :)

PR #4181, opened 2019-11-22
Optimize randomness collective flip

marcio-diaz:
This is my humble attempt to optimize randomness collective flip.
It seems to take more time than it should (at `on_initialize`).

This PR basically replaces a vector for a linked map, removing a `mutate` (to avoid reading the whole vector?). Maybe people with more knowledge about the runtime have a better idea.

TODO:

- [ ] - Write benchmark
- [ ] - Add more tests.
- [ ] - Add length variable to storage (?).


paint/randomness-collective-flip/src/lib.rs
```diff
@@ -239,11 +233,12 @@ mod tests {
...
 			setup_blocks(81);
 
-			let random_material = CollectiveFlip::random_material();
+			let random_material_0 = CollectiveFlip::random_material(0);
+			let random_material_1 = CollectiveFlip::random_material(1);
 
-			assert_eq!(random_material.len(), 81);
-			assert_ne!(random_material[0], random_material[1]);
-			assert_eq!(random_material[0], genesis_hash);
+			// assert_eq!(random_material.len(), 81);
```
expenses: Unless you're going to be re-adding these checks later, I'd just remove the line.

marcio-diaz: Yes, the plan is to re-add it.

palette/randomness-collective-flip/src/lib.rs
```diff
@@ -130,15 +124,16 @@ impl<T: Trait> Randomness<T::Hash> for Module<T> {
...
 		if !hash_series.is_empty() {
+			let block_number = <system::Module<T>>::block_number();
+			let index = block_number_to_index::<T>(block_number);
+
 			// Always the case after block 1 is initialised.
 			hash_series.iter()
 				.cycle()
-				.skip(index)
+				.skip(index.try_into().ok().unwrap())
 				.take(RANDOM_MATERIAL_LEN as usize)
```
marcio-diaz: Note to myself: write expect.

palette/randomness-collective-flip/src/lib.rs
```diff
@@ -130,15 +124,16 @@ impl<T: Trait> Randomness<T::Hash> for Module<T> {
...
-		let hash_series = <RandomMaterial<T>>::get();
 		if !hash_series.is_empty() {
+			let block_number = <system::Module<T>>::block_number();
+			let index = block_number_to_index::<T>(block_number);
+
 			// Always the case after block 1 is initialised.
 			hash_series.iter()
 				.cycle()
-				.skip(index)
+				.skip(index.try_into().ok().unwrap())
```
marcio-diaz: Note to myself: write expect.

marcio-diaz:
>     * A vector of N elements needs one hashing to get the key and one large decode. A linked map of N elements needs 1 hashing to get the head key and `N` small decodes and potentially `N-1` other hashings based on what is the `NextKey`, if it is hashed on unhashed. Based on this I would (totally unscientifically) predict that vector should work better actually. All in all: I am a bit skeptical of it, also since the vector is known to be 81 Hashes which I think would not exceed more than a few kilobytes. But interested to see some bechmarks that prove me wrong :)

I did a very un-scientific-don't-trust-on-it benchmark that seems to show this is better. The thing is: who is better at accessing only one element?

kianenigma:
> who is better at modifying only one random element? 

in that case I think linked map can be better if it only decode keys and skip data? but the code at least has a `Data<T>::enumerate()::collect();`. This decodes and goes through all of the 81 elements.

marcio-diaz:
> > who is better at modifying only one random element?
> 
> in that case I think linked map can be better if it only decode keys and skip data? but the code at least has a `Data<T>::enumerate()::collect();`. This decodes and goes through all of the 81 elements.

Yes, I hope that's not awful (need to check). On the other side that code is called when you need the randomness, not at every single block.

gavofyork:
The `Vec` should be a *lot* faster, since storage i/o, not decoding, should be the bottleneck, especially for sizes as small as 81. Since it's a `linked_map` it will use Blake2 to hash keys, and iterating over the full 81 keys will thus require 81 round-trips across the native/client bridge, 81 trie queries and/or `HashMap` lookups, 81 Blake2 hashes, 81 allocations and 81 decodes in all cases.

PR #4182, opened 2019-11-22
Rename Palette to FRAME

shawntabrizi:
To further avoid confusion between "Palette" and "Pallet", the libraries formerly known as the SRML will be called FRAME.

**FRAME: Framework for Runtime Aggregation of Modularised Entities**

> FRAME is our framework for developing new Substrate pallets. Pallets are a special kind of Rust module from which Substrate runtimes can be composed. Each pallet has its own discrete logic which can modify the features and functionality of your blockchain's state transition function.


xlc:
Please tell me this is the final rename...

gavofyork:
Me too.

gavofyork:
I think the idea is to keep renaming it until everyone if of the opinion that a name they don't like is better than having to change it _again_.

docs/README.adoc
```diff
@@ -446,8 +446,8 @@ substrate-trie
 sr-api, sr-io, sr-primitives, sr-sandbox, sr-std, sr-version
 * Substrate Runtime Module Library (SRML)
 [source, shell]
-pallet-assets, pallet-balances, palette-consensus, pallet-contracts, palette-council, pallet-democracy, pallet-example,
-palette-executive, palette-metadata, pallet-session, pallet-staking, palette-support, palette-system, pallet-timestamp,
+pallet-assets, pallet-balances, frame-consensus, pallet-contracts, frame-council, pallet-democracy, pallet-example,
```
gnunicorn: seems like this was wrong before already, please apply
```suggestion
pallet-assets, pallet-balances, palett-consensus, pallet-contracts, palett-council, pallet-democracy, pallet-example,
```

docs/README.adoc
```diff
@@ -446,8 +446,8 @@ substrate-trie
 sr-api, sr-io, sr-primitives, sr-sandbox, sr-std, sr-version
 * Substrate Runtime Module Library (SRML)
```
gnunicorn: This also needs renaming, I guess?

docs/Structure.adoc
```diff
@@ -99,7 +99,7 @@ Subkey is a client library to generate keys and sign transactions to send to a s
 ....
 +---------------+       +----------------+
 |               |       |                |
-|    runtime    +<------+   palette      |
+|    runtime    +<------+   frame      |
```
gnunicorn:
```suggestion
|    runtime    +<------+   frame        |
```

missing spaces.

frame/im-online/src/lib.rs
```diff
@@ -131,7 +131,7 @@ pub mod ed25519 {
 
 /// The local storage database key under which the worker progress status
 /// is tracked.
-const DB_KEY: &[u8] = b"palette/im-online-worker-status";
+const DB_KEY: &[u8] = b"frame/im-online-worker-status";
```
gnunicorn: this might be a problem on a soft-upgrade ...

shawntabrizi: We should probably remove this from the key all together right?

docs/README.adoc
shawntabrizi: There are actually a number of docs which still need to be fixed

shawntabrizi: I patched this, but we should do a docs pass at another time (so we dont block polkadot release)

kigawas:
> I think the idea is to keep renaming it until everyone is of the opinion that sticking with a name they don't like is better than having to change it _again_.

Can we stick to "srml"? Short and understandable

nodebreaker-carlsagan:
me three.

hskang9:
@shawntabrizi @gavofyork me four.

lamafab:
> > I think the idea is to keep renaming it until everyone is of the opinion that sticking with a name they don't like is better than having to change it _again_.
> 
> Can we stick to "srml"? Short and understandable

@kigawas @nodebreaker-carlsagan @hskang9 @gavofyork  We would generally appreciate if you didn't name things after Substrate (or Rust). Therefore all Polkadot implementations (with different languages, frameworks, etc.) are somewhat on equal ground. Please avoid "srml".

kigawas:
> > > I think the idea is to keep renaming it until everyone is of the opinion that sticking with a name they don't like is better than having to change it _again_.
> > 
> > 
> > Can we stick to "srml"? Short and understandable
> 
> @kigawas @nodebreaker-carlsagan @hskang9 @gavofyork We would generally appreciate if you didn't name things after Substrate (or Rust). Therefore all Polkadot implementations (with different languages, frameworks, etc.) are somewhat on equal ground. Please avoid "srml".

if you regard "srml" as something like "gnu" or "unix" then it's just a name with 4 characters.

kigawas:
And "frame" or "pallet" is too generic - sounds like a chemistry factory

image
image
image

lamafab:
@kigawas 
> if you regard "srml" as something like "gnu" or "unix" then it's just a name with 4 characters.

Acronyms are fine if they describe some generic components (Unix > Multics: "Multiplexed Information and Computing System"). But "srml" explicitly names **Substrate**, which is a Rust Framework.

xlc:
@lamafab I am not offering opinions but wanting to point out the code in this repository is Substrate, a Rust framework. If someone reimplement it with another language, it will just be another different framework with a different name. I don't see why the name of this framework matters to other PRE (Polkadot Runtime Environment) framework.

gnunicorn:
This issue has been resolved and the PR merged. If you want to have a conversation about preferred names, this isn't the place for it. Locked the conversation to prevent further spamming the participants.

PR #4184, opened 2019-11-22
Remove all (non-dev) `client` references from `frame`

gnunicorn:
Moving a few things into `primitives` (already prefixing correctly), allows us to get rid of `../client` references in `palette` - the only remaining one is a dev-dependency for tests.

This also moves back `peerset` and moves around a few traits and imports, namely `client_api::{blockhain, error}` are now to be top-level imported from the newly creates `blockchain` primitive.

refs #4099 .


Cargo.toml
```diff
@@ -125,6 +123,8 @@ members = [
 	"primitives/sr-version",
 	"primitives/state-machine",
 	"primitives/timestamp",
+	"primitives/transaction-pool",
```
tomusdrw: I strongly feel that both `transaction-pool` and `transaction-graph` should stay in `client`. We just need a proper tratification (already implemented in #3833).

Cargo.toml
```diff
@@ -105,9 +103,9 @@ members = [
 	"primitives/keyring",
 	"primitives/offchain",
 	"primitives/panic-handler",
-	"primitives/peerset",
 	"primitives/phragmen",
 	"primitives/rpc",
+	"primitives/rpc/api",
```
tomusdrw: I don't feel like `rpc/api` definitions fit the `primitives` directory. It's just a set of API that are exposed by the default substrate node. As a side effect it also generates a bunch of `rpc-client` structs, though that seem to be required for some `paint` module. I wonder if the rpc client doesn't deserve a separate place (maybe `util`?).

Perhaps we try too much to obey the rules in the script without actually looking deeper what we try to achieve. Seems to me that `primitives` is now just becoming a `common` stuff between runtime and client, instead of being a merely a set of primitive types/structs.

gnunicorn: agreed, this is wrong. moved it back to client, moved the only import out of palette into utils

Cargo.toml
gnunicorn: moved the depending crate out into utils for now, moved these two back into client.

client/api/src/lib.rs
```diff
@@ -18,20 +18,17 @@
...
-pub mod blockchain;
 pub mod call_executor;
 pub mod client;
-pub mod error;
 pub mod execution_extensions;
 pub mod light;
 pub mod notifications;
 
-// TODO: avoid re-exports
+pub use sp_blockchain as blockchain;
```
tomusdrw: What are we going to do with this crate? It seems that now we either import `sp_blockchain` directly or use `client_api::blockchain`. Would be good to have some consistency.

Also I'm still not in favor of re-exports in that crate, I think it makes more sense to mention the module name on the import path.

Cargo.toml
```diff
@@ -139,6 +137,8 @@ members = [
 	"test/utils/transaction-factory",
 	"utils/build-script-utils",
 	"utils/fork-tree",
+	"utils/frame/rpc/support",
```
tomusdrw: So this one is being moved to `uitls` only because `transaction-pool` dependency, right? That's going to be solved after #3833 (includes traitifcation of the pool)

utils/frame/rpc/support/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "frame-support-rpc"
+name = "substrate-frame-rpc-support"
```
tomusdrw: I'm not a huge fan of names here, why is it both `substrate` and `frame` and why `rpc` is before `support`?

client/api/src/lib.rs
gnunicorn: this is mostly for convenience for me for now. I consider it part of the unification to remove this and other lines in here in favor of explicit crate names.

Cargo.toml
gnunicorn: No, `support` is here because it depends on both frame and client:
```
frame-support = { path = "../../../../frame/support" }
substrate-primitives-storage = { path = "../../../../primitives/core/storage" }
sc-rpc-api = { path = "../../../../client/rpc/api" }
```

Which isn't allowed in neither `frame` nor `client` and thus should live elsewhere that makes on such assumption.

The one depending on transaction pool is `utils/frame/rpc/system`:

```

frame-system-rpc-runtime-api = { path = "../../../../frame/system/rpc/runtime-api" }
substrate-primitives = { path = "../../../../primitives/core" }
sp-blockchain = { path = "../../../../primitives/blockchain" }
sc-transaction-graph = { path = "../../../../client/transaction-pool/graph" }
```

Also here because it also depends on frame and client. The later could probably be moved back into frame once the traitification of pool is through. The ealier I am not sure.

But I think we should think hard and good about where the RPC-related-stuff of frame should be sitting and how strongly we want to have bindings through expressed primitives or how else to structure this.

utils/frame/rpc/support/Cargo.toml
gnunicorn: because they depend on both substrate and frame (not only either, but both), taking either out implies they wouldn't need the other IMHO...

`rpc` before the key modules because I thinking of a generic pattern to describe the integration: `substrate-frame` to define the integration barriers) + `-rpc` signifying the feature upon which the integration is happening + `-pallet` to name the specific pallet this relates to.

But as said before, I consider this a temporary place for either of them and just a reminder, that we need to fix that API here to make the distinction clear and move it back.

.maintain/ensure-deps.sh
```diff
@@ -3,43 +3,62 @@
...
-# Such rule is just: `client<frame`.
+# Such rule is just: `client crates must not depend on anything in /frame`.
 
 # The script should be run from the main repo directory!
 
 set -u
 
+# HARD FAILING
+MUST_NOT=(
+	"client crates must not depend on anything in /frame"
```
tomusdrw: I :heart:  the format, pretty clever :)

.maintain/ensure-deps.sh
```diff
@@ -3,43 +3,62 @@
...
+
+# ONLY DISPLAYED, script still succeeds
 PLEASE_DONT=(
-	"client<frame"
-	"client<node"
-	"frame<node"
-	"frame<client"
-	"primitives<frame"
-	"primitives<client"
+	"primitives crates must not depend on anything in /client"
```
tomusdrw: should not?

gnunicorn:
```suggestion
	"primitives crates should not depend on anything in /client"
```

PR #4186, opened 2019-11-22
Switch to new kvdb-rocksdb

bkchr:
This pr switches to the new kvdb-rocksdb, that will be released in the next days. However, we already need it now to fix the linking errors in Polkadot.


gavofyork:
@bkchr a few failing tests here it seems...

client/db/src/utils.rs
```diff
@@ -213,8 +213,24 @@ pub fn open_database(
 		#[cfg(feature = "kvdb-rocksdb")]
 		DatabaseSettingsSrc::Path { path, cache_size } => {
 			let mut db_config = DatabaseConfig::with_columns(Some(NUM_COLUMNS));
-			db_config.memory_budget = *cache_size;
-			let path = path.to_str().ok_or_else(|| client::error::Error::Backend("Invalid database path".into()))?;
+
+			if let Some(cache_size) = cache_size {
+				let state_col_budget = (*cache_size as f64 * 0.7) as usize;
+				let other_col_budget = state_col_budget - cache_size;
```
shawntabrizi:
```suggestion
				let other_col_budget =  cache_size - state_col_budget;
```

PR #4191, opened 2019-11-23
Remove `frame/` prefix from `im_online` DB_KEY.

shawntabrizi:
Caught in the renaming process. It is not obvious to me why `frame/` or `palette/` or `srml/` was included in the DB key.

If there is rational, would love to know what that is before this is closed.

### DIFF
--- a/frame/im-online/src/lib.rs
+++ b/frame/im-online/src/lib.rs
@@ -131,7 +131,7 @@ pub mod ed25519 {
 
 /// The local storage database key under which the worker progress status
 /// is tracked.
-const DB_KEY: &[u8] = b"frame/im-online-worker-status";
+const DB_KEY: &[u8] = b"im-online-worker-status";
 
 /// It's important to persist the worker state, since e.g. the
 /// server could be restarted while starting the gossip process, but before
--- a/frame/im-online/src/lib.rs
+++ b/frame/im-online/src/lib.rs
@@ -131,7 +131,7 @@ pub mod ed25519 {
 
 /// The local storage database key under which the worker progress status
 /// is tracked.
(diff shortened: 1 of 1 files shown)


bkchr:
Why do you want to remove it? It probably exists to make the prefix unique.

shawntabrizi:
@bkchr It got hit in the rename, and if there will be any further renames in the future (crossing my fingers against that), having this string be using in the DB_KEY may break something?

As for uniqueness, `im-online-worker-status` seems about as unique as `frame/im-online-worker-status`.

gavofyork:
it won't break anything, but we might want to prefix it with `parity/`.

shawntabrizi:
> it won't break anything, but we might want to prefix it with `parity/`.

Again, I am not exactly sure if we are really protecting anything here. It seems like if there would be a key conflict, it is because someone is trying to copy our key of `im-online-worker-status` (a pretty specific string with specific meaning). In which case, a prefix will not stop them from copying the key :P 

But I have added `parity/` per your suggestion :)

gnunicorn:
Just wondering if that (or the previous rename) has any effect, when deploying. I suppose that previously recorded imlonline tags won't be found between when you stopped the old and now used the new version? Could that cause issues because we are using an incomplete cache? Like us not reporting online nodes properly and thus slashing more than we should? Should we brace for impact on this or ensure to deploy at an era/session change to prevent this becoming a problem?

//cc @tomusdrw

tomusdrw:
@gnunicorn That's only a key for local database entry, which we use to synchronize between workers that potentially run in parallel. The worst thing that may happen is that the node will produce two reports during the transition period, and the old local storage entry will never be removed from the database, but both of those things seem minor.

PR #4192, opened 2019-11-23
Make weights compatible with u64

### LINKED ISSUE #4141: Make weights compatible with u64
Currently `type Weight = u32`. 

It is likely that we will not want to change this, but it should be easily possible, and the code should always respect and use this type, rather than hardcoding u32. This issue is a simple refactor of tests and codes that use the u32. At most, some overflow errors might happen that need to be dealt with, with either saturating or checked arithmetic. 

cc @pepyakin

DarkEld3r:
Closes #4141.


kianenigma:
need to check if this change is _breaking_ in anyway for Kusama

bin/node/runtime/src/impls.rs
```diff
@@ -339,8 +339,8 @@ mod tests {
 			100 * kb,
 			mb,
 			10 * mb,
-			Weight::max_value() / 2,
-			Weight::max_value()
+			(u32::max_value() / 2).into(),
```
bkchr: This is against the whole purpose of this pr.

DarkEld3r:
Current state: `Weight` is used everywhere instead of `u32`, so changing the underlying `Weight` type no longer requires additional refactoring. There is an issue with `u64`, though: the `weight_to_fee_should_not_overflow_on_large_weights` test fails in that case. I have digged into that a little, but I need more time.

I think that we can merge this pull request as it is and create a separate issue for dealing with the overflow.

gavofyork:
CI fails.

DarkEld3r:
I have fixed the benchmarks, though `gitlab-check_polkadot ` is failing now. As far as I can see, it is unrelated to my changes.

PR #4194, opened 2019-11-25
Fixed shared cache race on import

arkpar:
* Make sure cache is updated under impror lock so that the best block info is always consistent.
* `is_best` functor was checked under cache lock, which was supposed to guarantee consistency. But in reality it does not, because cache is also updated when validating transactions outside of import lock. So this is removed.
* Changed import lock to be `Rwlock` instead of `Mutex` for extra performance.

### DIFF
--- a/client/src/call_executor.rs
+++ b/client/src/call_executor.rs
@@ -97,7 +97,7 @@ where
 		)
 		.map(|(result, _, _)| result)?;
 		{
-			let _ = self.backend.get_import_lock().read();
+			let _lock = self.backend.get_import_lock().read();
 			self.backend.destroy_state(state)?;
 		}
 		Ok(return_data.into_encoded())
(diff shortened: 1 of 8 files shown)


arkpar:
> LGTM. Probably [this comment](https://github.com/paritytech/substrate/blob/master/client/api/src/backend.rs#L309) must be updated as well, because `Backend` actually acquires the lock itself starting from this PR. Also maybe worth mentioning that no states should be destroyed when import lock is held (this seems a bit fragile)?

Good point. It would be better to lock it outside.

client/src/call_executor.rs
```diff
@@ -96,7 +96,10 @@ where
 			None,
 		)
 		.map(|(result, _, _)| result)?;
-		self.backend.destroy_state(state)?;
+		{
+			let _ = self.backend.get_import_lock().read();
```
bkchr: That is not correct. It drops immediately the lock!

It is also not required to create a sub context here, as the functions end after it anyway.

arkpar: > That is not correct. It drops immediately the lock!

How so? The `RwLockReadGuard` returned by `read` lives till the end of the block.

> It is also not required to create a sub context here, as the functions end after it anyway.

Here it actually makes sense because we don't want to hold the lock for the duration of `into_encoded`

bkchr: https://github.com/rust-lang/rust/issues/40096

PR #4196, opened 2019-11-25
Remove `keystore` when doing calls and producing a proof

tomusdrw:
Follow up on #4145 

Removes `keystore` parameter from calls that might be invoked in untrusted environment. I believe we were passing `None` to all of them anyway.


bkchr:
Needs some fixes :)

PR #4199, opened 2019-11-25
Use timestamp deltas in grafana-data-store

expenses:
Currently, we're storing metrics for grafana in vectors of `(f32, i64)`, which have a `size_of` of 16. This PR replaces this system by storing a single `i64` base timestamp, and `u32` timestampdeltas from that, so we end up halving this to 8 bytes (`(f32, u32)`). `u32::max_value()` is `4294967295` milliseconds, or 49 days, while we clean up the metrics every 24 hours. This PR also just cleans up the code as well.


client/grafana-data-source/src/lib.rs
```diff
@@ -24,35 +24,51 @@
...
-			metrics.entry($key).or_insert_with(Vec::new).push(($value as f32, now));
-		)*
+		$crate::record_metrics_slice(&[
+			$( ($key, $value as f32), )*
+		]);
 	}
 );
+
+/// Write metrics to `METRICS` as a slice.
+pub fn record_metrics_slice(metrics: &[(&'static str, f32)]) -> Result<(), Error> {
```
niklasad1: this could be more generic by taking: `IntoIterator<Item = (&'static str, f32)>` instead

client/grafana-data-source/src/database.rs
```diff
@@ -0,0 +1,130 @@
...
+		})
+	}
+
+	fn make_absolute(&self, base_timestamp: i64) -> (f32, i64) {
+		(self.value, base_timestamp + self.delta_timestamp as i64)
+	}
+}
+
+fn find_index(slice: &[Datapoint], base_timestamp: i64, timestamp: i64) -> usize {
+	slice.binary_search_by_key(&timestamp, |datapoint| {
```
niklasad1: this requires the slice to be sorted, I can't see that it is sorted?

client/grafana-data-source/src/database.rs
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	pub fn datapoints_between(&self, key: &str, from: i64, to: i64, max_datapoints: usize) -> Option<Vec<(f32, i64)>> {
+		self.storage.get(key)
+			.map(|vec| {
+				let from = find_index(vec, self.base_timestamp, from);
+				let to = find_index(vec, self.base_timestamp, to);
+				let slice = &vec[from .. to];
+
+				if max_datapoints == 0 {
```
niklasad1: this could be performed before `b_searching` the datapoints but perhaps very unlikely.

client/grafana-data-source/src/database.rs
```diff
@@ -0,0 +1,130 @@
...
+					Vec::new()
+				} else if max_datapoints >= slice.len() {
+					// Just convert the slice as-is
+					slice.iter()
+						.map(|dp| dp.make_absolute(self.base_timestamp))
+						.collect()
+				} else {
+					// We have more datapoints than we need, so we need to skip some
+					(0 .. max_datapoints - 1)
```
niklasad1: hmm, this looks rather complex. Why can't you just do:

```rust
slice.iter().take(max_datapoints).map(|dp| dp.make_absolute(self.base_timestamp))
```

Then, remove the if conditions?

client/grafana-data-source/src/database.rs
```diff
@@ -0,0 +1,130 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use crate::Error;
+
+pub struct Database {
```
niklasad1: add docs for public items,  perhaps add `#[warn(missing_docs)]` in the crate root (lib.rs)

client/grafana-data-source/src/database.rs
```diff
@@ -0,0 +1,130 @@
...
+				.collect();
+		}
+
+		self.base_timestamp = new_base_timestamp;
+
+		Ok(())
+	}
+}
+
+#[derive(Clone)]
```
niklasad1: why not Copy?

client/grafana-data-source/src/database.rs
niklasad1: this assumes the `&[Datapoint]` is sorted by monotonic increasing timestamps?

client/grafana-data-source/src/database.rs
expenses: But we need to sample `max_datapoints` evenly from the `slice`, not just from the start.

client/grafana-data-source/src/database.rs
expenses: Yep, but they are. This might break if someone messes with their system clock, but we're assuming that they wont.

client/grafana-data-source/src/database.rs
expenses: The compiler should optimize the `max_datapoints == 0` to happen first regardless.

client/grafana-data-source/src/lib.rs
```diff
@@ -24,35 +24,51 @@
...
+/// Error type that can be returned by either `record_metrics` or `run_server`.
+#[derive(Debug, derive_more::Display)]
+pub enum Error {
+	Hyper(hyper::Error),
+	Serde(serde_json::Error),
+	Http(hyper::http::Error),
+	Timestamp(TryFromIntError),
+}
+
+impl std::error::Error for Error {}
```
tomusdrw: You should implement `fn source`

client/grafana-data-source/src/database.rs
tomusdrw: Can we get some tests for this struct? Would also be good to add some docs.

client/grafana-data-source/src/database.rs
```diff
@@ -0,0 +1,130 @@
...
+	pub fn keys_starting_with<'a>(&'a self, base: &'a str) -> impl Iterator<Item = &'static str> + 'a {
+		self.storage.keys().filter(move |key| key.starts_with(base)).cloned()
+	}
+
+	pub fn datapoints_between(&self, key: &str, from: i64, to: i64, max_datapoints: usize) -> Option<Vec<(f32, i64)>> {
+		self.storage.get(key)
+			.map(|vec| {
+				let from = find_index(vec, self.base_timestamp, from);
+				let to = find_index(vec, self.base_timestamp, to);
+				let slice = &vec[from .. to];
```
tomusdrw: Is this guaranteed to never panic? what if `from == to` in case both are not found?

client/grafana-data-source/src/database.rs
```diff
@@ -0,0 +1,130 @@
...
+					Vec::new()
+				} else if max_datapoints >= slice.len() {
+					// Just convert the slice as-is
+					slice.iter()
+						.map(|dp| dp.make_absolute(self.base_timestamp))
+						.collect()
+				} else {
+					// We have more datapoints than we need, so we need to skip some
+					(0 .. max_datapoints - 1)
+						.map(|i| &slice[i * slice.len() / (max_datapoints - 1)])
```
tomusdrw: Clever, but you could just use `Iterator::step_by` I suppose?

client/grafana-data-source/src/database.rs
niklasad1: ok, I see but it is not evenly `shuffled` with the current implementation because it will always start with zero and then it will be increased by a multiple of `slice length` divided by `max_datapoints` together with last all the time.

Ideally, you would use [SliceRandom:shuffle](https://docs.rs/rand/0.6.5/rand/seq/trait.SliceRandom.html#tymethod.shuffle) instead and just take take the first `max_datapoints` or simply just generate random `index` where to start sample from.

expenses: Sorry, I didn't really mean 'sample' (as-in random), I just meant that we need to get out `max_datapoints` taken from regular intervals.

client/grafana-data-source/src/database.rs
expenses: Unfortunately, the step wouldn't be even every time. For example, if `slice.len()` is 1.5x `max_datapoints`.

client/grafana-data-source/src/database.rs
expenses: `&slice[x..x]` is just `&[]` so this should be fine.

client/grafana-data-source/src/lib.rs
expenses: I might just take the `pub` off actually.

client/grafana-data-source/src/database.rs
expenses: `Database` isn't public.

client/grafana-data-source/src/database.rs
tomusdrw: Right. What about a case where index is not found? I think we will get `x > vec.len()` and panic in such case?

client/grafana-data-source/src/lib.rs
expenses: The type signature would be `IntoIterator<Item = &'a (&'static str, f32)>` to make it work with slices, which is less nice. I chose to clarify when it should be used instead.

gavofyork:
CI fails & conflicts.

client/grafana-data-source/src/lib.rs
niklasad1: fair enough

client/grafana-data-source/src/lib.rs
```diff
@@ -23,36 +23,63 @@
 //! [Grafana]: https://grafana.com/
 //! [`grafana-json-data-source`]: https://github.com/simPod/grafana-json-datasource
 
+#[warn(missing_docs)]
```
niklasad1:
```suggestion
#![warn(missing_docs)]
```

PR #4200, opened 2019-11-25
Add documentation to SubmitSignedTransaction and actually make it work

tomusdrw:
With recent changes to crypto stuff (notably `IdentifyAccount` that makes the `AccountId` diverge from `PublicKey`) the `SubmitSignedTransaction` trait family was not really working correctly.

The main complications arise because of application-crypto type wrappers (`app-crypto`) and signed aggregators (`MultiSigner`), this requires a bit of `From/TryInto/Into` magic and some additional trait bounds, that on the first sight look unnecessary in this code :)

This PR is an attempt to fix that and also make it more convenient to use local keys for transaction signing (see `SigningAccountFinder`). This additional trait bound can be used to create and submit signed transactions from offchain calls using all local keys (intersected with some list of accounts).

Still WiP, as I want to work on tests a bit more, but wanted to make this visible for @kianenigma @gnunicorn and @andresilva


frame/system/src/offchain.rs
```diff
@@ -130,9 +162,70 @@ pub trait SubmitUnsignedTransaction<T: crate::Trait, Call> {
...
+	///
+	/// This method should intersect given list of accounts with the ones
+	/// supported locally and submit signed transaction containing given `Call`
+	/// with every of them.
+	///
+	/// Returns a vector of results and account ids that were supported.
+	fn submit_signed(
+		accounts: impl IntoIterator<Item= T::AccountId>,
+		call: impl Into<Call> + Clone,
+	) -> Vec<(T::AccountId, Result<(), ()>)> {
```
andresilva: Do you think it makes sense to add an option to stop on the first successful submission?

tomusdrw: Why would that be useful? I think we should always submit with all accounts that are there (i.e. we do (or planned to) support mutliple validators on one node, right).

andresilva: In the equivocation report I was bailing out after the first success, but yeah probably makes sense to report from all available accounts anyway.

tomusdrw:
Finally ready for review. Added tests and a bunch of utility methods (like `can_sign` - CC @gnunicorn )

gavofyork:
tests failing?

tomusdrw:
@gavofyork I believe it was caused by a new compiler version. Merged with master and it seems fine. Polkadot failure seems unrelated (some `clap` issues).

kianenigma:
https://github.com/paritytech/substrate/pull/4517 is _eagerly_ waiting for this :D

gavofyork:
@tomusdrw is this merge correct?

tomusdrw:
Yup, thanks! Could I get a second review on this? @bkchr maybe if you have some time? :smiling_imp:

frame/system/src/offchain.rs
```diff
@@ -130,9 +167,114 @@ pub trait SubmitUnsignedTransaction<T: crate::Trait, Call> {
...
+	/// Create and submit signed transactions from all local accounts.
+	///
+	/// This method submits a signed transaction from all local accounts
+	/// for given application crypto.
+	///
+	/// Returns a vector of results and account ids that were supported.
+	#[must_use]
+	fn submit_signed(
+		call: impl Into<Call> + Clone,
+	) -> Vec<(T::AccountId, Result<(), ()>)> {
```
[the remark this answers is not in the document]
tomusdrw: I like `impl` syntax on input parameters, but returning an `Iterator` seems a bit unnecessary for me. Also I don't think if it's even possible on a trait method.

bin/node/executor/tests/basic.rs
```diff
@@ -0,0 +1,837 @@
...
+	System, TransactionPayment, Event,
+	TransferFee, TransactionBaseFee, TransactionByteFee,
+	constants::currency::*,
+};
+use node_primitives::{Balance, Hash};
+use wabt;
+use node_testing::keyring::*;
+
+mod common;
+use self::common::{*, sign};
```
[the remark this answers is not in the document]
tomusdrw: This doesn't work as there is a conflict with glob imports (`common` and `keystore`), so we need to define `sign` explicitly.

frame/system/src/lib.rs
```diff
@@ -628,11 +649,15 @@ impl<T: Trait> Module<T> {
 	}
 
 	/// Start the execution of a particular block.
+	///
+	/// Setting `is_for_inspection` flag will not cause the events
```
bkchr: I think the comment can be removed? Or at least be fixed.

tomusdrw: Good catch, thanks!

tomusdrw:
@bkchr convinced me to change `offchain_worker` `RuntimeApi` and pass the entire header to make the initialization properly, I'm working on this right now, hence the `inprogress` tag.

PR #4201, opened 2019-11-25
Check in block authoring that we can author with current authoring version

bkchr:
We need to make sure that we can still author with the current block authoring implementation, because there could have been a runtime update in between that made the current implementation incompatible.


gavofyork:
@arkpar : didn't already implement this 18 months ago? :-)

arkpar:
I did here: https://github.com/paritytech/substrate/pull/256/files#diff-76964ec8de481638ce29d022262cfc53R454
But it seems it was lost in one of many refactorings.

gavofyork:
CI fails.

gavofyork:
@bkchr will need a polkadot PR...

bkchr:
Yeah, just waiting for https://github.com/paritytech/substrate/pull/4236 and than will open a pr for both :)

PR #4211, opened 2019-11-26
Integrate grafana and tracing

marcio-diaz:
This starts the integration between grafana and tracing output:

Screenshot_2019-11-26 Substrate Dashboard - Grafana

I opened the PR to get ideas on how to best integrate this two modules and on the usefulness of it.

TODO:
- Make the aggregation more precise (module, function, scopes?).
- Support multiple receivers? (e.g. stdout, grafana, telemetry)

### DIFF
--- a/client/tracing/src/lib.rs
+++ b/client/tracing/src/lib.rs
@@ -39,6 +39,7 @@ use parking_lot::Mutex;
 use tracing_core::{event::Event, Level, metadata::Metadata, span::{Attributes, Id, Record}, subscriber::Subscriber};
 
 use substrate_telemetry::{telemetry, SUBSTRATE_INFO};
+use grafana_data_source::{self, record_metrics};
 
 /// Used to configure how to receive the metrics
 #[derive(Debug, Clone)]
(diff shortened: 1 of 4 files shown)


client/tracing/src/lib.rs
```diff
@@ -196,3 +200,7 @@ fn send_telemetry(span_datum: SpanDatum) {
 		"time" => span_datum.overall_time.as_nanos(),
 	);
 }
+
+fn send_grafana(span_datum: SpanDatum) {
+	record_metrics!(span_datum.target => span_datum.overall_time.as_nanos());
```
mattrutherford: yeah, not sure the correct way to represent this for grafana, but seeing as we're limited to just a single `key => value`, we probably need to just concatenate `span_datum.target` and `span_datum.name`.

maybe `record_metrics!(format!("{}::{}", span_datum.target, span_datum.name) => span_datum.overall_time.as_nanos());` or something similar, if that works for grafana?

expenses:
Oh, nice! :D

PR #4212, opened 2019-11-26
skip passed slots and wake up service correctly

rphmeier:
This provides a fast-path for avoiding proposing for slots that have already passed. The service future was also not being woken up correctly when essential tasks exited.


client/consensus/slots/src/lib.rs
```diff
@@ -122,6 +122,19 @@ pub trait SimpleSlotWorker<B: BlockT> {
 		let (timestamp, slot_number, slot_duration) =
 			(slot_info.timestamp, slot_info.number, slot_info.duration);
 
+		{
+			let slot_now = SignedDuration::default().slot_now(slot_duration);
+			if slot_now > slot_number {
+				// if this is behind, return.
+					debug!(target: self.logging_target(),
```
gavofyork:
```suggestion
				debug!(target: self.logging_target(),
```

andresilva:
```suggestion
				debug!(target: self.logging_target(),
```

PR #4215, opened 2019-11-26
Make block proposing remaining duration configurable

sorpaas:
This tries to address the issue that proposers giving up proposing blocks due to long proposing time, causing the network not producing blocks for a long time.

`remaining_duration`, which is the parameter used to determine how long the proposing logic tolerant execution, is now splited into two -- `slot_remaining_duration`, for inner proposing logic to determine pushing extrinsic, and `proposing_remaining_duration`, for outer `on_slot` to determine the actual timeout. The timeout is now configurable by each consensus engine. For example, an engine can change this to be determined by last proposing slot `actual_slot - expected_slot > expected_slot - parent_slot`.

When review, please note that the behavior is changed that `on_slot` will not time out on proposing by default. I talked with @rphmeier and we think that's fine, but please let me know if you think this will create issues!

### DIFF
--- a/client/consensus/slots/src/lib.rs
+++ b/client/consensus/slots/src/lib.rs
@@ -124,8 +124,12 @@ pub trait SimpleSlotWorker<B: BlockT> {
 	}
 
 	/// Remaining duration for proposing. None means unlimited.
-	fn proposing_remaining_duration(&self, _slot_info: &SlotInfo) -> Option<Duration> {
-		None
+	fn proposing_remaining_duration(
+		&self,
+		_head: &B::Header,
+		slot_info: &SlotInfo
+	) -> Option<Duration> {
+		Some(self.slot_remaining_duration(slot_info))
 	}
 
 	/// Implements the `on_slot` functionality from `SlotWorker`.
@@ -208,7 +212,7 @@ pub trait SimpleSlotWorker<B: BlockT> {
 		};
 
 		let slot_remaining_duration = self.slot_remaining_duration(&slot_info);
-		let proposing_remaining_duration = self.proposing_remaining_duration(&slot_info);
+		let proposing_remaining_duration = self.proposing_remaining_duration(&chain_head, &slot_info);
 		let logs = self.pre_digest_data(slot_number, &claim);
 
 		// deadline our production to approx. the end of the slot
--- a/client/consensus/slots/src/lib.rs
+++ b/client/consensus/slots/src/lib.rs
@@ -38,7 +38,7 @@ use inherents::{InherentData, InherentDataProviders};
 use log::{debug, error, info, warn};
 use sr_primitives::generic::BlockId;
 use sr_primitives::traits::{ApiRef, Block as BlockT, Header, ProvideRuntimeApi};
-use std::{fmt::Debug, ops::Deref, pin::Pin, sync::Arc};
+use std::{fmt::Debug, ops::Deref, pin::Pin, sync::Arc, time::{Instant, Duration}};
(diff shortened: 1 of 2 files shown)


client/consensus/slots/src/lib.rs
```diff
@@ -192,45 +207,43 @@ pub trait SimpleSlotWorker<B: BlockT> {
...
-				Err(consensus_common::Error::ClientImport("Timeout in the Slots proposer".into())),
-		});
+			}));
 
 		let block_import_params_maker = self.block_import_params();
 		let block_import = self.block_import();
 		let logging_target = self.logging_target();
 
 		Box::pin(proposal_work.map_ok(move |(block, claim)| {
-			// minor hack since we don't have access to the timestamp
```
sorpaas: Because the proposing is now offloaded to a dedicated threadpool, the timeout above by `future::select(Delay)` is now reliable, and this "minor hack" can be removed.

client/consensus/slots/src/slots.rs
```diff
@@ -71,6 +71,8 @@ pub fn time_until_next(now: Duration, slot_duration: u64) -> Duration {
 pub struct SlotInfo {
 	/// The slot number.
 	pub number: u64,
+	/// The last slot number produced.
+	pub last_number: u64,
```
sorpaas: This parameter not currently used, but is added so that engines can implement logic on `proposing_remaining_duration` based on parent/last proposing slot.

client/consensus/slots/src/lib.rs
```diff
@@ -113,6 +113,21 @@ pub trait SimpleSlotWorker<B: BlockT> {
...
+		if now < slot_info.ends_at {
+			slot_info.ends_at.duration_since(now)
+		} else {
+			Duration::from_millis(0)
+		}
+	}
+
+	/// Remaining duration for proposing. None means unlimited.
+	fn proposing_remaining_duration(&self, _slot_info: &SlotInfo) -> Option<Duration> {
+		None
```
andresilva: Why not default to `Some(slot_remaining_duration)`? I think it's safer to make sure inner proposal logic is always timed out externally.

client/consensus/slots/src/slots.rs
rphmeier: What does `produced` mean here?

What we would probably want to compare against is the slot of the parent header

sorpaas: @rphmeier It should be the slot when we last tried to propose blocks, so it's not necessarily `number - 1`. I think this parameter would be useful if we just want to calculate the `remaining_duration` based on past rough average proposing time?

I added reference of chain head header to `proposing_remaining_duration`. In this way, if the engine wants to calculate the duration based on parent slot, they can do it by parsing the header digest.

client/consensus/slots/src/lib.rs
sorpaas: Okay I changed it to that.

sorpaas: Still, I want to note that if we really get a faulty proposer that stuck in an infinite loop, this won't really save us. We currently don't kill offloaded proposing logic. The threadpool has a few spots, if that fills up, no new proposing thread will be created, meaning the offloading future will just always return pending. 

I don't think we need to deal with that extreme situation. If we do, the only (?) solution would probably be to kill those offloaded thread. I currently think that will be a bad idea, as it may poison locks and damage internal state.

client/consensus/slots/src/slots.rs
rphmeier: I think that's the best we could hope for. Thanks

PR #4218, opened 2019-11-27
Add sha2-256 hash function

kigawas:
Widely used hash function, supported by [bitcoin](https://en.bitcoin.it/wiki/Script#Crypto) and [ethereum](https://solidity.readthedocs.io/en/v0.5.13/units-and-global-variables.html#mathematical-and-cryptographic-functions)

### DIFF
--- a/primitives/core/src/hashing.rs
+++ b/primitives/core/src/hashing.rs
@@ -17,6 +17,7 @@
 //! Hashing functions.
 
 use blake2_rfc;
+use sha2::{Digest, Sha256};
 use tiny_keccak::{Hasher, Keccak};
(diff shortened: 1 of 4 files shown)


primitives/core/src/hashing.rs
```diff
@@ -131,3 +132,11 @@ pub fn keccak_256(data: &[u8]) -> [u8; 32] {
 	keccak.finalize(&mut output);
 	output
 }
+
+pub fn sha2_256(data: &[u8]) -> [u8; 32] {
```
tomusdrw: Missing docs

PR #4219, opened 2019-11-27
Prune some duplicate dependencies

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

* Remove ed25519-dalek 0.9.1
* Remove hex 0.3.2
* Remove parity-wasm 0.40.3
* Remove pwasm-utils 0.11.0
* Remove wasmi-validation 0.2.0
* Remove quickcheck 0.8.5
* Remove tempdir (Replace tempdir with tempfile)
* Remove derive_more 0.15


client/keystore/src/lib.rs
```diff
@@ -317,12 +317,16 @@ impl BareCryptoStore for Store {
 #[cfg(test)]
 mod tests {
 	use super::*;
-	use tempdir::TempDir;
+	use tempfile::TempDir;
 	use primitives::{testing::{SR25519}, crypto::{Ss58Codec}};
 
+	fn tempdir_with_prefix(prefix: &str) -> TempDir {
```
bkchr: As you see is the prefix always the same, so you don't need to take it as parameter. And the actual name of the directory is not important, aka we don't need the prefix at all.

koushiro: Okay, I will fix it :)

koushiro: PTAL again

PR #4221, opened 2019-11-27
fix some discrepancies in generic-asset

kigawas:


gavofyork:
needs resolving. @kigawas

Issue #3885, opened 2019-10-22
put_code takes too much time


pepyakin:
https://github.com/paritytech/substrate/issues/3819 reports that `put_code` can take a lot of time for a big-but-still-reasonable-sized (IMO, 240k, see the issue) contract. While, [integration of wasmtime](https://github.com/paritytech/substrate/issues/3869) can fix that, the situation can imply that put_code requires some optimization.

One clue how to improve this might be in the insight that we are doing multiple iterations over the wasm binary. We could try to leverage the streaming approach to read, check and instrument binary in one go. This however would require quite some refactorings of the wasm-utils side.

expenses:
I've done some benchmarking on the wasm file in #3819 and (as expected) the hot functions are `parity_wasm::elements::deserialize_buffer` (4.46s) and `wasmi_validation::validate_module` (3.00s). The file actually fails to validate as it uses `f64`s, so a streaming approach would definitely speed things up by returning early.

pepyakin:
Oh, thanks for looking into this!

Good to know. I think we can't rely on an early return since we always have to take into account the worst case (e.g. an attacker can always put `f64` at the very end of code section).

Stefie:
A little update on this:

I just ran into the same issue with:
- The latest versions of Substrate FRAME-contracts (`spec_version: 195`, `impl_version: 196`, at commit [`# 6e9c675`](https://github.com/paritytech/substrate/commit/6e9c6759130d04c72389cf3fba96d8bcf95caef8))
- Latest version of cargo-contract ([`#c54fef9`](https://github.com/paritytech/cargo-contract/commit/c54fef90ded5a092f72c5bb8ec6e52f2fccf144c))
- and the latest version of ink! (https://github.com/paritytech/ink/commit/414eca1b526ab6fa020237e52af620a1b847bcba)) 

with a more than reasonably sized contract (52k).

The problem persists and might be worse than expected. It has already been reported in #4153, that deploying contracts much smaller than 240k will bring block production to halt.

The compiled ERC20 example  in the [`ink/examples/lang2/erc20`](https://github.com/paritytech/ink/tree/master/examples/lang2/erc20) folder has a size of **52k** and will cause this bug. 
The .wasm file of the [Flipper example](https://github.com/paritytech/ink/tree/master/examples/lang2/flipper) has a size of **17k** and can be deployed without problems.

expenses:
@Stefie I might try and write some benchmarks for this so we can examine why it's taking so long.

[...]

Issue #3819, opened 2019-10-15
Avoid missing a slot on slow proposer


jeremymj:
I want to upload a contract file to the node, when the contract file is about 100kB, it can be successfully uploaded; but when I upload the contract at 240kB,the node can't generate a new block,the prompt message is as follows:
`2019-10-15 17:50:56 Discarding proposal for slot 261855508; block production took too long`
I compiled the substrate code  version is commit `1ae7a901d9b2a7f4fefc7b5f5846c2d13ef6d8c7` 
by command `cargo build  --release`
and started the node by `./target/release/substrate --dev`,
the test contract wasm file I am using is as follows: [reproduce_wasm_file.zip](https://github.com/paritytech/substrate/files/3728594/reproduce_wasm_file.zip)

thiolliere:
Is this a transaction weight issue ?

If an extrinsic take too much time to be executed then it shouldn't be put into the block, the weight mecanism should prevent block from being too big, I think the weight for this extrinsic is wrong.

I'm not sure how to solve it though.

cc @kianenigma

thiolliere:
sidenote:
* if I put panic before the line https://github.com/paritytech/substrate/blob/9b8bfd72126be453cda302caff28a6c99adee14c/srml/contracts/src/wasm/prepare.rs#L343 then extrinsic is invalid and block are produced
* if I put a panic after this line https://github.com/paritytech/substrate/blob/9b8bfd72126be453cda302caff28a6c99adee14c/srml/contracts/src/wasm/prepare.rs#L344 then this issue happens

maybe some code could be optimised as well so we can handle contract of this size, but also maybe this contract size is quite big and the issue should be solve by reducing it with features like dynamic linking.

kianenigma:
From the looks of it, the block did take too long and the weight system did not catch it basically. But the transaction is actually one of the dynamic ones that we currently don't really deal with in the weight system. This can also be fixed by making the authorship code multi-threaded.

 I must try and first reproduce this. 

cc @pepyakin @andresilva

jeremymj:
Thank you for your follow-up on this issue! I am now trying to optimize the contract business logic, reduce the contract size, and look forward to your updates.

tomusdrw:
I believe we should mark the extrinsic as invalid if we timeout during it's execution and it's the first one in the block.

@kianenigma I don't see how we could make authorship multi-threaded, you mean to somehow verify correctness separately?

[...]

PR #4222, opened 2019-11-27
`Contracts::put_code` benchmarks

expenses:
#3885 and #3819 note that `put_code` can take too long on large wasm files, and cause block production to fail. This PR adds benchmarks for this using compiled [ink examples](https://github.com/paritytech/ink/tree/master/examples).


expenses:
Here's the resulting time for `erc20` by the way:

```
wasms/erc20.wasm        time:   [6.4462 ms 6.5215 ms 6.5997 ms]
                        thrpt:  [7.6973 MiB/s 7.7897 MiB/s 7.8806 MiB/s]
```

This seems pretty fast, so maybe the problems only occur under certain conditions?

expenses:
@Stefie maybe you could run these benchmarks on your machine?

thiolliere:
this does benchmark put_code in native. I don't remember precisely but thoughts wasm was used always for block production. maybe in wasm the put_code performance are very different 🤔 ?

expenses:
Oh, good point! What would be the best way to benchmark in wasm then?

thiolliere:
> Oh, good point! What would be the best way to benchmark in wasm then?

I don't know @pepyakin ?

bkchr:
@expenses you can setup a wasm blob like this one: https://github.com/paritytech/substrate/tree/master/primitives/runtime-interface/test-wasm

Expose a function `benchmark_put_code` or whatever. This function internally calls 1000x `put_code`. And you call this function from your benchmark. See [here](https://github.com/paritytech/substrate/blob/b3a7c8e463ff7a63ce20e1eb14d8ccc63de4a412/primitives/runtime-interface/src/lib.rs#L154) for on how to call it.

Stefie:
> @Stefie maybe you could run these benchmarks on your machine?

Sorry, I was out of office yesterday. If you tell me how, I'm happy to do that! I never ran any benchmarking tests for Substrate.

The times I mentioned in https://github.com/paritytech/substrate/issues/3885 were with the `latest` Docker image from https://hub.docker.com/r/parity/substrate/tags

expenses:
@bkchr is this even possible, given that `GenesisConfig`s and `TestExternalities` only works under `std`?

bkchr:
@expenses please look at the linked code, you create the storage etc from std and just call into wasm to execute the benchmark method.

expenses:
Screen Shot 2019-12-04 at 4 31 23 PM

Here's the output from instrumentation.

expenses:
Closing due to solutions having arised.

PR #4223, opened 2019-11-27
The crate rename

gnunicorn:
refs #4099 .

Rename crates in `/primitives` and `/client`: prefix them with with `sp-`and `sc-` respectively, remove any unnecessary suffixing `-primitives`, `-runtime-api`. Fixup all the used names inside the repo.

## Upgrading

You'll find the bash script `rename-crates-for-2.0.sh` in the `.maintain`-folder. It is the script that was run and should be working the same way on your substrate-github-referencing project by renaming all the references in `Cargo.toml` and `*.rs`-files. You might have to `cargo clean` and build on a clean project again, if you encounter errors.


gnunicorn:
force pushed a new run of the script against master (rather than attempting to fix the enormous conflict)

.maintain/rename-crates-for-2.0.sh
```diff
@@ -0,0 +1,112 @@
...
+    "substrate-network sc-network"
+    "substrate-offchain sc-offchain"
+    "substrate-peerset sc-peerset"
+    "substrate-rpc-servers sc-rpc-server"
+    "substrate-rpc sc-rpc"
+    "substrate-service sc-service"
+    "substrate-service-test sc-service-test"
+    "substrate-state-db sc-state-db"
+    "substrate-telemetry sc-telemetry"
+    "substrate-tracing sc-transaction"
```
tomusdrw: ?
```suggestion
    "substrate-tracing sc-tracing"
```

.maintain/rename-crates-for-2.0.sh
```diff
@@ -0,0 +1,112 @@
...
+    "sr-arithmetic sp-arithmetic"
+    "sr-arithmetic-fuzzer sp-arithmetic-fuzzer"
+    "sr-io sp-io"
+    "sr-primitives sp-runtime"
+    "sr-sandbox sp-sandbox"
+    "sr-staking-primitives sp-staking"
+    "sr-std sp-std"
+    "sr-version sp-version"
+    "substrate-state-machine sp-state-machine"
+    "substrate-transaction-pool-runtime-api sp-transaction-pool"
```
tomusdrw: This has been changed in #3833 it's now `sp-transaction-pool-runtime-api` already, we also have `sp-transaction-pool-api` and I believe that both should be merged into `sp-transaction-pool`, but let's do that in a separate PR. I guess this line can be removed from the script then as it has no effect, unless we want to help migrating other projects too.

gnunicorn: I'd leave it in, it does no harm and makes conversions for others idea. I just applied it against a clean master this morning and no bad side effects.

PR #4226, opened 2019-11-27
Hard-cap the execution time of custom contract calls.

tomusdrw:
To prevent RPC spam, make sure we disallow contract queries that may take forever to complete. The limit is now based on the assumption that:
1. Gas cost is roughly equivalent to weight (CC @pepyakin to confirm)
2. One weight is roughly one nano second (seems like something that is assumed in linked W3F doc).

Current default block weight limit is `1e9` as well, so this limit can be interpreted as either:
1. Up to 5s of native execution.
2. Or 5x max block size.

I'd rather be conservative here - if this ever creates issues we can revisit later.


frame/contracts/rpc/src/lib.rs
```diff
@@ -39,7 +39,16 @@ const RUNTIME_ERROR: i64 = 1;
...
 const CONTRACT_IS_A_TOMBSTONE: i64 = 3;
 
-// A private newtype for converting `GetStorageError` into an RPC error.
+/// A rough estimate of how much gas a decent hardware consumes per second,
+/// using native execution.
+/// This value is used to set the upper bound for maximal contract calls to
+/// prevent blocking the RPC for too long.
+///
+/// Based on W3F research spreadsheet:
+/// https://docs.google.com/spreadsheets/d/1h0RqncdqiWI4KgxO0z9JIpZEJESXjX_ZCK6LFX6veDo/view
```
HCastano: I don't have permissions to view this, so I'm guessing others won't either. Who do we talk to about making this public?

frame/contracts/rpc/src/lib.rs
```diff
@@ -148,6 +157,19 @@ where
 			data: None,
 		})?;
 
+		let max_gas_limit = 5 * GAS_PER_SECOND;
```
HCastano: I'd add the "interpretation" you put in the PR description as a comment here

HCastano:
Some nits, but looks good

PR #4227, opened 2019-11-27
Implement all storage after prefix

thiolliere:
This updates the way storages are implemented now all storage are prefixed resulting in:

* storage value at `twox128(module) ++ twox128(storagename)`
* linked_map and map at `twox128(module)++twox128(storagename)++hasher(key)`
* linked_map head at `twox128(module) ++ twox128("HeadOf" ++ storagename)`
* double_map at `twox128(module)++twox128(storagename)++hasher(key1)++hasher(key2)`

TODO:
I have no idea how we want to launch this PR in regards to Kusama, nor what the schedules are, but here it is.

NOTE:
* some improvement could be done by returning arrays instead of Vector for generating the final keys, but this is not very important.
* this doesn't implement any new sr-io method to iterate on anything

CC @bkchr @jacogr

### DIFF
--- a/frame/support/procedural/src/storage/storage_struct.rs
+++ b/frame/support/procedural/src/storage/storage_struct.rs
@@ -16,11 +16,11 @@
 
 //! Implementation of storage structures and implementation of storage traits on them.
 
-use proc_macro2::TokenStream;
+use proc_macro2::{TokenStream, Ident, Span};
 use quote::quote;
 use super::{
 	DeclStorageDefExt, StorageLineTypeDef,
-	instance_trait::{PREFIX_FOR, HEAD_KEY_FOR},
+	instance_trait::INHERENT_INSTANCE_NAME,
 };
 
 fn from_optional_value_to_query(is_option: bool, default: &Option<syn::Expr>) -> TokenStream {
@@ -78,17 +78,14 @@ pub fn decl_and_impl(scrate: &TokenStream, def: &DeclStorageDefExt) -> TokenStre
 		let from_optional_value_to_query =
 			from_optional_value_to_query(line.is_option, &line.default_value);
 
-		let final_prefix = if let Some(instance) = def.module_instance.as_ref() {
-			let instance = &instance.instance_generic;
-			let const_name = syn::Ident::new(
-				&format!("{}{}", PREFIX_FOR, line.name.to_string()), proc_macro2::Span::call_site()
-			);
-			quote!( #instance::#const_name.as_bytes() )
+		// Contains accessor to instance, used to get prefixes
+		let instance_or_inherent = if let Some(instance) = def.module_instance.as_ref() {
+			instance.instance_generic.clone()
 		} else {
-			let prefix = format!("{} {}", def.crate_name, line.name);
-			quote!( #prefix.as_bytes() )
+			Ident::new(INHERENT_INSTANCE_NAME, Span::call_site())
 		};
 
+		let storage_name_str= syn::LitStr::new(&line.name.to_string(), line.name.span());
 
 		let storage_generator_trait = &line.storage_generator_trait;
 		let storage_struct = &line.storage_struct;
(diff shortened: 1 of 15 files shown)


jacogr:
We will need a bump - the API needs some way of detecting how to format this. At this point basically all storage keys will be wrong, so effectively nothing the API touches will work. 

(Remember that the APIs need to support all versions of chains out in the wild, so some pointers that indicate "format stuff in this way" is always good)

frame/support/procedural/src/storage/storage_struct.rs
```diff
@@ -78,17 +78,14 @@ pub fn decl_and_impl(scrate: &TokenStream, def: &DeclStorageDefExt) -> TokenStre
...
+		// Contains accessor to instance, used to get prefixes
+		let instance_or_inherent = if let Some(instance) = def.module_instance.as_ref() {
+			instance.instance_generic.clone()
 		} else {
-			let prefix = format!("{} {}", def.crate_name, line.name);
-			quote!( #prefix.as_bytes() )
+			Ident::new(INHERENT_INSTANCE_NAME, Span::call_site())
 		};
 
+		let storage_name_str= syn::LitStr::new(&line.name.to_string(), line.name.span());
```
bkchr:
```suggestion
		let storage_name_str = syn::LitStr::new(&line.name.to_string(), line.name.span());
```

frame/support/procedural/src/storage/storage_struct.rs
```diff
@@ -144,30 +149,18 @@ pub fn decl_and_impl(scrate: &TokenStream, def: &DeclStorageDefExt) -> TokenStre
...
-					let const_name = syn::Ident::new(
-						&format!("{}{}", HEAD_KEY_FOR, line.name.to_string()), proc_macro2::Span::call_site()
-					);
-					quote!( #instance::#const_name.as_bytes() )
-				} else {
-					let prefix = format!("head of {} {}", def.crate_name, line.name);
-					quote!( #prefix.as_bytes() )
-				};
+				let head_prefix_str = syn::LitStr::new(
+					&format!("HeadOf{}", line.name.to_string()),
```
bkchr: CC @jacogr just to make sure you have seen this :)

jacogr: Yes, sadly this little line took me, erm, 2.5 tries to get right - first time forgot about it, second time I forgot that it is twox (instead of hasher as per previous). But confirmed working with validators & nominators on a dev chain using this branch, can iterate from head.

freddieee:
> This updates the way storages are implemented now all storage are prefixed resulting in:
> 
> * storage value at `twox128(module) ++ twox128(storagename)`
> * linked_map and map at `twox128(module)++twox128(storagename)++hasher(key)`
> * linked_map head at `twox128(module) ++ twox128("HeadOf" ++ storagename)`
> * double_map at `twox128(module)++twox128(storagename)++hasher(key1)++hasher(key2)`
> 
> TODO:
> I have no idea how we want to launch this PR in regards to Kusama, nor what the schedules are, but here it is.
> 
> NOTE:
> 
> * some improvement could be done by returning arrays instead of Vector for generating the final keys, but this is not very important.
> * this doesn't implement any new sr-io method to iterate on anything
> 
> CC @bkchr @jacogr

Hi， Could u please explain it in more detail? What does the "++" means here? I try to use `xxhashAsHex` function to get the `twox128(module)` and `twox128(storagename)` value and concat them together as a storage key, but still it seems to be a invalid key.

jacogr:
https://github.com/polkadot-js/api/blob/master/packages/metadata/src/Decorated/storage/fromMetadata/createFunction.ts#L62

freddieee:
> https://github.com/polkadot-js/api/blob/master/packages/metadata/src/Decorated/storage/fromMetadata/createFunction.ts#L62

Thanks for your reply! I did exactly the same thing. For example, I get `0x26aa394eea5630e07c48ae0c9558cef702a5c1b19ab7a04f536c519aca4983ac` from the function you referred with  Prefix "System", Method "Number". But with this storage key I cannot retrieve any value from rpc call `state_getStorage`

jacogr:
Ok, so it really is not the place to (excuse the term) revive PRs from the dead, a PR is not an issue.

I would suggest asking around in the Substrate technical chat, the keys as explained here and implemented in the JS API (a well as the Python version), works, so it is not an API-use issue, nor an implementation issue.

PR #4231, opened 2019-11-27
The storage runtime interface should not enforce a hash type

bkchr:
Currently the runtime interface enforces `H256` as hash type, but in the
future people could use whatever they want as hash type. The hash type
always needs to match between the runtime and the node, but that is
already required.

### DIFF
--- a/client/db/src/lib.rs
+++ b/client/db/src/lib.rs
@@ -174,7 +174,7 @@ impl<B: BlockT> StateBackend<Blake2Hasher> for RefTrackingState<B> {
 		self.state.storage_root(delta)
 	}
 
-	fn child_storage_root<I>(&self, storage_key: &[u8], delta: I) -> (Vec<u8>, bool, Self::Transaction)
+	fn child_storage_root<I>(&self, storage_key: &[u8], delta: I) -> (H256, bool, Self::Transaction)
 		where
 			I: IntoIterator<Item=(Vec<u8>, Option<Vec<u8>>)>,
 	{
@@ -193,7 +193,9 @@ impl<B: BlockT> StateBackend<Blake2Hasher> for RefTrackingState<B> {
 		self.state.child_keys(child_key, prefix)
 	}
 
-	fn as_trie_backend(&mut self) -> Option<&state_machine::TrieBackend<Self::TrieBackendStorage, Blake2Hasher>> {
+	fn as_trie_backend(
+		&mut self,
+	) -> Option<&state_machine::TrieBackend<Self::TrieBackendStorage, Blake2Hasher>> {
 		self.state.as_trie_backend()
 	}
 }
--- a/primitives/externalities/src/lib.rs
+++ b/primitives/externalities/src/lib.rs
@@ -44,7 +44,7 @@ pub trait Externalities: ExtensionStore {
 
 	/// Get child storage value hash. This may be optimized for large values.
 	///
-	/// Returns an `Option` that holds the raw hash.
+	/// Returns an `Option` that holds the SCALE encoded hash.
 	fn child_storage_hash(&self, storage_key: ChildStorageKey, key: &[u8]) -> Option<Vec<u8>>;
 
 	/// Read original runtime storage, ignoring any overlayed changes.
--- a/primitives/externalities/src/lib.rs
+++ b/primitives/externalities/src/lib.rs
@@ -22,8 +22,6 @@
 //!
 //! This crate exposes the main [`Externalities`] trait.
 
-use primitive_types::H256;
-
 use std::any::{Any, TypeId};
 
 use primitives_storage::ChildStorageKey;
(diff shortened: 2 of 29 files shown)


primitives/externalities/src/lib.rs
```diff
@@ -42,26 +40,40 @@ pub trait Externalities: ExtensionStore {
 	fn storage(&self, key: &[u8]) -> Option<Vec<u8>>;
 
 	/// Get storage value hash. This may be optimized for large values.
-	fn storage_hash(&self, key: &[u8]) -> Option<H256>;
+	fn storage_hash(&self, key: &[u8]) -> Option<Vec<u8>>;
 
 	/// Get child storage value hash. This may be optimized for large values.
-	fn child_storage_hash(&self, storage_key: ChildStorageKey, key: &[u8]) -> Option<H256>;
+	///
+	/// Returns an `Option` that holds the raw hash.
```
thiolliere: what does raw hash means in this context ?

looking at the implementation:
```rust
	fn child_storage_hash(&self, storage_key: ChildStorageKey, key: &[u8]) -> Option<Vec<u8>> {
		self.child_storage(storage_key, key).map(|v| Blake2Hasher::hash(&v).encode())
```
it seems SCALE encoded hash as well no ?

```suggestion
	/// Returns an `Option` that holds the SCALE encoded hash.
```

primitives/externalities/src/lib.rs
```diff
@@ -42,26 +40,40 @@ pub trait Externalities: ExtensionStore {
 	fn storage(&self, key: &[u8]) -> Option<Vec<u8>>;
 
 	/// Get storage value hash. This may be optimized for large values.
-	fn storage_hash(&self, key: &[u8]) -> Option<H256>;
+	fn storage_hash(&self, key: &[u8]) -> Option<Vec<u8>>;
```
thiolliere: I suppose we don't make hasher as a associated type because we want to have this trait being able to be a trait object isn't it ?

primitives/state-machine/src/basic.rs
```diff
@@ -118,27 +119,31 @@ impl Externalities for BasicExternalities {
 		self.top.get(key).cloned()
 	}
 
-	fn storage_hash(&self, key: &[u8]) -> Option<H256> {
-		self.storage(key).map(|v| Blake2Hasher::hash(&v))
+	fn storage_hash(&self, key: &[u8]) -> Option<Vec<u8>> {
+		self.storage(key).map(|v| Blake2Hasher::hash(&v).encode())
```
thiolliere: should we put in the doc that it uses `[u8; 32]` as a hash ?

primitives/state-machine/src/proving_backend.rs
```diff
@@ -145,6 +146,7 @@ impl<'a, S, H> ProvingBackendRecorder<'a, S, H>
 		key: &[u8]
 	) -> Result<Option<Vec<u8>>, String> {
 		let root = self.storage(storage_key)?
+			.and_then(|r| Decode::decode(&mut &r[..]).ok())
```
thiolliere: why doesn't it returns an error here ?

primitives/state-machine/src/trie_backend.rs
```diff
@@ -159,16 +160,17 @@ impl<S: TrieBackendStorage<H>, H: Hasher> Backend<H> for TrieBackend<S, H> where
...
-		H::Out: Ord
+		H::Out: Ord,
 	{
 		let default_root = default_child_trie_root::<Layout<H>>(storage_key);
 
 		let mut write_overlay = S::Overlay::default();
 		let mut root = match self.storage(storage_key) {
-			Ok(value) => value.unwrap_or(default_root.clone()),
+			Ok(value) =>
+				value.and_then(|r| Decode::decode(&mut &r[..]).ok()).unwrap_or(default_root.clone()),
```
thiolliere: this decode fails only on corrupted storage if I understand correctly, maybe it worth a warn or at least a comment

client/db/src/lib.rs
```diff
@@ -174,7 +174,7 @@ impl<B: BlockT> StateBackend<Blake2Hasher> for RefTrackingState<B> {
 		self.state.storage_root(delta)
 	}
 
-	fn child_storage_root<I>(&self, storage_key: &[u8], delta: I) -> (Vec<u8>, bool, Self::Transaction)
+	fn child_storage_root<I>(&self, storage_key: &[u8], delta: I) -> (H256, bool, Self::Transaction)
```
cheme: Just curious, I do not really understand this change (all the switch of child trie api from 'Vec<u8>' to H256), it seems like it goes in the opposite direction as the pr (iirc child api was not using H::Out or H256 in prevision of using different hash type for different child tries (but there is multiple others things to change to achieve that so it could seem a bit pointless)).

bkchr: That comes in a different pr, where I completely drop the `BlockT<Hash = H256>` stuff in the codebase. However it is not that important here, as this is the `state_machine::Backend` that is a node only internface.

primitives/externalities/src/lib.rs
bkchr: Yep.

primitives/externalities/src/lib.rs
bkchr: Yes

primitives/state-machine/src/basic.rs
bkchr: `BasicExternalities` need to be adapted when we have someone that uses a different hash.

client/db/src/lib.rs
cheme: > drop the BlockT<Hash = H256>
That would be amazing (I remember starting something ,but dropping the refactoring in the middle (the environmental part being tricky)) :+1: 
I aggree it is not very important.

Issue #4216, opened 2019-11-26
Hash raw secp256k1 Public Keys instead of compressed ones


fubuloubu:
In #3861, `secp256k1` keypairs were introduced. However, do to their construction, it wasn't possible to uniquely store a public key as 32 bytes without breaking the larger API, therefore the key types were hashed [here](https://github.com/paritytech/substrate/pull/3861/files#diff-2d4fe59416e64524ebe5d49fb1cd82c2R249) and [here](https://github.com/paritytech/substrate/pull/3861/files#diff-2d4fe59416e64524ebe5d49fb1cd82c2R293) in order to fit the keys securely into `AccountId`.

The choice of Public datatype for `secp256k1` keys in Substrate was the 33 byte "compressed" version of the key, which is then hashed via `blake2b` to produce the 32 byte size. If instead, the 64-byte "raw" version of the key (without type identifiers) was used for this hash (and the hashing algorithm were configurable to `keccak256`, i.e. #3624) this would directly match the address format that Ethereum uses and allow more directly compatible usage of Substrate with Ethereum-based applications that already have significant state on them, and allow a better degree of inter-operation between the two (say, for Dothereum).

I think long term what would be most preferable would be to allow users to directly configure how addresses are displayed, and be able to specify their own conversion and encoding formats between public keys and addresses (making Substrate overall a more flexible tool), but this small change will help bridge that gap a lot better than how it is currently set up.

burdges:
We support secp256k1 primarily to be Ethereum friendly, so yes being compatible sounds good.  I presume the 64 bit form is Affine.  

I'll note however that secp256k1 ecdsa signatures cannot be batch verified like ed25519 or sr25519, so we should probably make them cost slightly more whenever we start doing batch verification.  I suppose keccak256 should cost more too until hardware venders add support.  It'd be simpler if we can pick _one_ somewhat Ethereum friendly secp256k1 ecdsa signature form and stick with it, regardless of where it uses blake2b or keccak256.

fubuloubu:
Using `keccak256` for `AccountId` would definitely be preferred, but I think assumes the use case is Ethereum-only (it could be Bitcoin?)

PR #4234, opened 2019-11-27
Use raw public key instead of compressed

fubuloubu:
Fixes: #4216

### DIFF
--- a/primitives/core/src/ecdsa.rs
+++ b/primitives/core/src/ecdsa.rs
@@ -46,9 +46,9 @@ use secp256k1::{PublicKey, SecretKey};
 #[cfg(feature = "full_crypto")]
 type Seed = [u8; 32];
 
-/// The ECDSA 33-byte compressed public key.
+/// The ECDSA 64-byte raw public key.
 #[derive(Clone, Encode, Decode)]
-pub struct Public(pub [u8; 33]);
+pub struct Public(pub [u8; 64]);
 
 impl PartialOrd for Public {
 	fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
@@ -72,7 +72,7 @@ impl Eq for Public {}
 
 impl Default for Public {
 	fn default() -> Self {
-		Public([0u8; 33])
+		Public([0u8; 64])
 	}
 }
 
@@ -84,8 +84,8 @@ pub struct Pair {
 	secret: SecretKey,
 }
 
-impl AsRef<[u8; 33]> for Public {
-	fn as_ref(&self) -> &[u8; 33] {
+impl AsRef<[u8; 64]> for Public {
+	fn as_ref(&self) -> &[u8; 64] {
 		&self.0
 	}
 }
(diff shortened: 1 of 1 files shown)


primitives/core/src/ecdsa.rs
```diff
@@ -332,16 +332,27 @@ pub enum PublicError {
...
+	pub fn from_raw(data: [u8; 64]) -> Self {
 		Public(data)
 	}
 
+	/// A new instance from the given 65-byte `data`.
+	///
+	/// NOTE: No checking goes on to ensure this is a real public key. Only use it if
+	/// you are certain that the array actually is a pubkey. GIGO!
+    pub fn from_full(data: [u8; 65]) -> Self {
+        let raw_key = &data[1..];
```
bkchr: Whitespaces

fubuloubu:
Also, as noted in https://github.com/paritytech/substrate/issues/4216#issuecomment-559147275 using `keccak256` instead of `blake2_256` in `AccountId` hashing logic would make this _more_ Ethereum-compatible, but as I noted in my response it would be better if `AccountId` wasn't assumed 32 bytes (instead just `Sized`) and was user configurable with how addresses are encoded.

I could add `keccak_256` to this PR though.

gavofyork:
`AccountId` is:

```
type AccountId: Parameter + Member + MaybeSerializeDeserialize + Debug + MaybeDisplay + Ord + Default;
```

So basically, you can have it be any type you want. There's no need for this PR, though since it doesn't really break anything and makes things strictly more compatible I'm ok merging it. But you can in any case make a 20-byte `AccountId` if you really want and you can make a new `Signature` type which is compatible with it (and a new `MultiSignature` if you want to support other crypto at the same time, too).

gavofyork:
For now, though, some of the CI is failing and will need to be fixed before it can be merged.

fubuloubu:
> [Y]ou can in any case make a 20-byte `AccountId` and use sha256, keccak256, ripemd-160 or whatever else if you really want and you can make a new `Signature` type which is compatible with it (and a new `MultiSignature` if you want to support other crypto at the same time, too).

Ah, well I will have to try a different method of doing what I want. I was using `MultiSigner` and `MultiSignature`, which builds in that presumption of the `blake2` hash, and I assumed there was no other way to do it, but I see where I am wrong though. This PR is definitely still foundational however.

Will fix CI issues.

primitives/core/src/ecdsa.rs
```diff
@@ -558,7 +569,7 @@ mod test {
 		);
 		let public = pair.public();
 		assert_eq!(public, Public::from_raw(
-			hex!("028db55b05db86c0b1786ca49f095d76344c9e6056b2f02701a7e7f3c20aabfd91")
+			hex!("8db55b05db86c0b1786ca49f095d76344c9e6056b2f02701a7e7f3c20aabfd913ebbe148dd17c56551a52952371071a6c604b3f3abe8f2c8fa742158ea6dd7d4")
 		));
 		let message = b"";
 		let signature = hex!("3dde91174bd9359027be59a428b8146513df80a2a3c7eda2194f64de04a69ab97b753169e94db6ffd50921a2668a48b94ca11e3d32c1ff19cfe88890aa7e8f3c00");
```
fubuloubu: One of the CI failures is due to line width of the line 3 above this one. How did this one get configured not to trip the line size check?

fubuloubu:
As noted above, the 120 char line limit is failing because I made a hex string larger. It doesn't seem like I should be splitting it up, so it is probably best to ignore this error?

The other two errors do not seem to be my fault, as far as I can tell.

Please advise.

primitives/core/src/ecdsa.rs
gavofyork: yeah unfortunately i don't think it's possible, so can ignore it.

PR #4236, opened 2019-11-27
Update kvdb-rocksdb again

bkchr:


arkpar:
Build fails

bkchr:
I'm now waiting for the release of the crate, which should happen today :)

dvdplm:
> which should happen today

It did happen. :)

bkchr:
Yeah, I'm already at updating :D

PR #4237, opened 2019-11-28
Parameter trait for Time::Moment

kigawas:
Without this we cannot have a Moment parameter in dispatchable functions

### DIFF
--- a/frame/support/src/traits.rs
+++ b/frame/support/src/traits.rs
@@ -26,6 +26,8 @@ use sr_primitives::{
 	traits::{MaybeSerializeDeserialize, SimpleArithmetic, Saturating},
 };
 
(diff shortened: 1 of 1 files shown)


frame/support/src/traits.rs
```diff
@@ -622,7 +622,7 @@ bitmask! {
 }
 
 pub trait Time {
-	type Moment: SimpleArithmetic + FullCodec + Clone + Default + Copy;
+	type Moment: SimpleArithmetic + FullCodec + Clone + Default + Copy + Debug;
```
```suggestion
	type Moment: SimpleArithmetic + Parameter + Default + Copy;
```

bkchr: Probably needs an import for `Parameter`.

PR #4240, opened 2019-11-28
Check block CLI command

arkpar:
A CLI command to re-validate existing block. This is useful for debugging/tracing block execution. 
Also fixed `--pruning` option for import/export commands and got rid of macros in service, such as `import_blocks!`

### DIFF
--- a/client/cli/src/lib.rs
+++ b/client/cli/src/lib.rs
@@ -762,7 +762,7 @@ where
 		Some(s) => {
 			if role == service::Roles::AUTHORITY && !cli.unsafe_pruning {
 				return Err(error::Error::Input(
-						"Validators should run with state pruning disabled (i.e. archive). \
+					"Validators should run with state pruning disabled (i.e. archive). \
 					You can ignore this check with `--unsafe-pruning`.".to_string()
 				));
 			}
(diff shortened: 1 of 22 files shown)


client/cli/src/lib.rs
```diff
@@ -694,76 +738,31 @@ fn fill_config_keystore_password<C, G, E>(
...
-	config.pruning = match cli.pruning {
+	config.pruning = match &cli.pruning {
 		Some(ref s) if s == "archive" => PruningMode::ArchiveAll,
 		None if role == service::Roles::AUTHORITY => PruningMode::ArchiveAll,
 		None => PruningMode::default(),
 		Some(s) => {
 			if role == service::Roles::AUTHORITY && !cli.unsafe_pruning {
 				return Err(error::Error::Input(
-					"Validators should run with state pruning disabled (i.e. archive). \
+						"Validators should run with state pruning disabled (i.e. archive). \
```
gavofyork:
```suggestion
					"Validators should run with state pruning disabled (i.e. archive). \
```

client/cli/src/params.rs
```diff
@@ -868,6 +871,10 @@ impl<CC, RP> StructOpt for CoreParams<CC, RP> where
 			ImportBlocksCmd::augment_clap(SubCommand::with_name("import-blocks"))
 				.about("Import blocks from file.")
 		)
+		.subcommand(
+			CheckBlockCmd::augment_clap(SubCommand::with_name("check-block"))
+			.about("Re-validate a known block.")
```
gavofyork:
```suggestion
				.about("Re-validate a known block.")
```

Issue #4233, opened 2019-11-27
"Address already in use" should be handled gracefully


iammelea:
after update to 0.6.18 node crash, stop and start no solved
```
kusasentry  | 2019-11-27 13:20:13 ----------------------------
0|kusasentry  | 2019-11-27 13:20:13 This chain is not in any way
0|kusasentry  | 2019-11-27 13:20:13       endorsed by the       
0|kusasentry  | 2019-11-27 13:20:13      KUSAMA FOUNDATION      
0|kusasentry  | 2019-11-27 13:20:13 ----------------------------
0|kusasentry  | 2019-11-27 13:20:13 Node name: melea-◮👁◭ Vali→ [CdEm1E_5Yb9dr]
0|kusasentry  | 2019-11-27 13:20:13 Roles: SENTRY
0|kusasentry  | 2019-11-27 13:20:13 Highest known block at #850155
0|kusasentry  | 2019-11-27 13:20:13 Local node identity is: QmW7AsPownc6ZS7smc3SbMKioUuUk4ch7xxr7az1ajzxjH
0|kusasentry  | 2019-11-27 13:20:13 Unable to bind server to 127.0.0.1:9944. Trying random port.
0|kusasentry  | ====================
0|kusasentry  | Version: 
0|kusasentry  | 0.6.18-63d2c0e3-x86_64-linux-gnu
0|kusasentry  | stack backtrace:
0|kusasentry  |    0: substrate_panic_handler::set::{{closure}}
0|kusasentry  |    1: std::panicking::rust_panic_with_hook
0|kusasentry  |              at src/libstd/panicking.rs:477
0|kusasentry  |    2: std::panicking::continue_panic_fmt
0|kusasentry  |              at src/libstd/panicking.rs:380
0|kusasentry  |    3: rust_begin_unwind
0|kusasentry  |              at src/libstd/panicking.rs:307
0|kusasentry  |    4: core::panicking::panic_fmt
0|kusasentry  |              at src/libcore/panicking.rs:85
0|kusasentry  |    5: core::result::unwrap_failed
0|kusasentry  |              at src/libcore/result.rs:1165
0|kusasentry  |    6: <std::future::GenFuture<T> as core::future::future::Future>::poll
0|kusasentry  |    7: <futures_util::future::select::Select<A,B> as core::future::future::Future>::poll
0|kusasentry  |    8: <futures_util::compat::compat03as01::Compat<Fut> as futures::future::Future>::poll
[...]

PR #4241, opened 2019-11-28
Handle Grafana server bind errors gracefuly.

tomusdrw:
Closes https://github.com/paritytech/substrate/issues/4233

### DIFF
--- a/client/grafana-data-source/src/server.rs
+++ b/client/grafana-data-source/src/server.rs
@@ -110,12 +110,50 @@ impl<T> tokio_executor::TypedExecutor<T> for Executor
 	}
 }
 
+/// An error that may occur during server runtime.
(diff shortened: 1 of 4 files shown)


client/grafana-data-source/src/server.rs
```diff
@@ -110,12 +110,50 @@ impl<T> tokio_executor::TypedExecutor<T> for Executor
 	}
 }
 
+/// An error that may occur during server runtime.
+#[derive(Debug, derive_more::Display, derive_more::From)]
+pub enum RunError {
+	/// Propagated hyper server error.
+	Hyper(hyper::Error),
+	/// Initial bind IO error.
+	Io(async_std::io::Error),
```
tomaka: This is the same as `std::io::Error`. Use that to make CI pass.

```suggestion
	Io(std::io::Error),
```

PR #4243, opened 2019-11-28
Commission is stated as percent of rewards

gavofyork:


frame/staking/src/lib.rs
```diff
@@ -1252,8 +1252,8 @@ impl<T: Trait> Module<T> {
 	/// nominators' balance, pro-rata based on their exposure, after having removed the validator's
 	/// pre-payout cut.
 	fn reward_validator(stash: &T::AccountId, reward: BalanceOf<T>) -> PositiveImbalanceOf<T> {
-		let off_the_table = reward.min(Self::validators(stash).validator_payment);
-		let reward = reward - off_the_table;
+		let off_the_table = Self::validators(stash).commission * reward;
+		let reward = reward.saturating_sub(off_the_table);
```
tomusdrw: We can't really end up with anything greater than `reward`, can we? Just making sure, I'm in favour of the change anyway, just to be safe.

gavofyork: i don't think so since `Perbill` can only go up to 1, but if i didn't do it then an auditor would surely pick up on it and request the change anyway :-)

thiolliere: Yes Perbill is between 0 and 1

tomusdrw: A nitpicky auditor would suggest to replace the multiplication with `saturating_mul` as well, just in case `Perbill` goes over `1` :rofl:

thiolliere: unfortunately Perbill doesn't implement saturating_mul with balance only saturing_mul with another Perbill hehe

gavofyork:
Passes all tests locally. CI appears to be gummed up again.

PR #4251, opened 2019-11-29
Remove unused host_interface as it has been moved to deprecated_host_interface

thiolliere:
host_interface.rs has been moved to deprecated_host_interface.rs but the file still is in the repo


gavofyork:
Will need a fix for polkadot.

bkchr:
The `check_polkadot` error is not related to this change.

PR #4256, opened 2019-11-30
Economic fixes for democracy module

gavofyork:
Democracy currently has a state bloat attack since public proposals are allowed to include `Call`s which get stored indefinitely in state. This revamps that system so that only a fixed-length hash of the `Call` is stored, and only when the motion is accepted does the underlying preimage `Call` get uploaded to the chain. To protect the chain from griefing, there is a substantial refundable deposit for uploading preimage data on the chain of around 1 cent per byte (around $10,000 for 1MB), but this is waived in the case that the preimage is for a motion that has already passed and is in the enactment period.

Once a proposal is enacted, then the preimage is dropped and the deposit refunded. A preimage deposit can also be dropped/refunded once it times out, the time out period is the `voting_period + enactment_period`.

Since there already exist proposals in Kusama, an additional sudo function is introduced; `clear_proposals` to get rid of them around upgrade time.

Additionally, the tests have been updated to reflect the fact that the per-block logic happens during `initialize_block` rather than at `finalize_block`.

# Breaking change

Democracy module requires additional configuration items `PreimageByteDeposit`, `Slash` in the config trait. Here's an example of how it might look:

```
parameter_types! {
	...
	pub const PreimageByteDeposit: Balance = 1 * CENTS;
}
...
impl democracy::Trait for Runtime {
	...
	type PreimageByteDeposit = PreimageByteDeposit;
	type Slash = Treasury;
}
```

### DIFF
--- a/frame/democracy/src/lib.rs
+++ b/frame/democracy/src/lib.rs
@@ -22,7 +22,7 @@ use rstd::prelude::*;
 use rstd::{result, convert::TryFrom};
 use sr_primitives::{
 	RuntimeDebug,
-	traits::{Zero, Bounded, CheckedMul, CheckedDiv, EnsureOrigin, Hash, Dispatchable},
+	traits::{Zero, Bounded, CheckedMul, CheckedDiv, EnsureOrigin, Hash, Dispatchable, Saturating},
 };
 use codec::{Ref, Encode, Decode, Input, Output, Error};
 use support::{
@@ -227,30 +227,33 @@ pub trait Trait: system::Trait + Sized {
 
 	/// Period in blocks where an external proposal may not be re-submitted after being vetoed.
 	type CooloffPeriod: Get<Self::BlockNumber>;
+
+	/// The amount of balance that must be deposited per byte of preimage stored.
+	type PreimageByteDeposit: Get<BalanceOf<Self>>;
 }
 
 /// Info regarding an ongoing referendum.
 #[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug)]
-pub struct ReferendumInfo<BlockNumber: Parameter, Proposal: Parameter> {
+pub struct ReferendumInfo<BlockNumber: Parameter, Hash: Parameter> {
 	/// When voting on this referendum will end.
 	end: BlockNumber,
 	/// The proposal being voted on.
-	proposal: Proposal,
+	proposal_hash: Hash,
 	/// The thresholding mechanism to determine whether it passed.
 	threshold: VoteThreshold,
 	/// The delay (in blocks) to wait after a successful referendum before deploying.
 	delay: BlockNumber,
 }
 
-impl<BlockNumber: Parameter, Proposal: Parameter> ReferendumInfo<BlockNumber, Proposal> {
+impl<BlockNumber: Parameter, Hash: Parameter> ReferendumInfo<BlockNumber, Hash> {
 	/// Create a new instance.
 	pub fn new(
 		end: BlockNumber,
-		proposal: Proposal,
+		proposal_hash: Hash,
 		threshold: VoteThreshold,
 		delay: BlockNumber
 	) -> Self {
-		ReferendumInfo { end, proposal, threshold, delay }
+		ReferendumInfo { end, proposal_hash, threshold, delay }
 	}
 }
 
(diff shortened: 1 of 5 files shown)


frame/democracy/src/lib.rs
```diff
@@ -642,6 +673,70 @@ decl_module! {
...
+		/// Remove an expired proposal preimage and collect the deposit.
+		#[weight = SimpleDispatchInfo::FixedNormal(10_000)]
+		fn reap_preimage(origin, proposal_hash: T::Hash) {
+			let who = ensure_signed(origin)?;
+
+			if let Some((_, old, deposit, then)) = <Preimages<T>>::get(&proposal_hash) {
+				let now = <system::Module<T>>::block_number();
+				if now >= then + T::EnactmentPeriod::get() + T::VotingPeriod::get() {
+					// allowed to claim the deposit.
+					let _ = T::Currency::repatriate_reserved(&old, &who, deposit);
```
shawntabrizi: Brutal...

frame/democracy/src/lib.rs
```diff
@@ -827,9 +922,24 @@ impl<T: Trait> Module<T> {
 	}
 
 	/// Enact a proposal from a referendum.
-	fn enact_proposal(proposal: T::Proposal, index: ReferendumIndex) {
-		let ok = proposal.dispatch(system::RawOrigin::Root.into()).is_ok();
-		Self::deposit_event(RawEvent::Executed(index, ok));
+	fn enact_proposal(proposal_hash: T::Hash, index: ReferendumIndex) -> Result {
+		if let Some((encoded_proposal, who, amount, _)) = <Preimages<T>>::take(&proposal_hash) {
+			if let Ok(proposal) = T::Proposal::decode(&mut &encoded_proposal[..]) {
```
shawntabrizi: It seems there is a way that a reserved balance is permanently stuck.

If the `encoded_proposal` fails to decode, you will never get to the `unreserve` action below.

Since you `take` on the `Preimages`, you will also not be able to `reap_preimage`.

Seems like the unreserve step should happen before the decode or in both of the else statements.

frame/democracy/src/lib.rs
shawntabrizi: So this is literal a race to get the deposit by everyone?

There seems to be no mechanism that a user who knows their proposal failed can try to claim their reserved balance back first...

frame/democracy/src/lib.rs
```diff
@@ -227,39 +227,45 @@ pub trait Trait: system::Trait + Sized {
...
 	trait Store for Module<T: Trait> as Democracy {
 		/// The number of (public) proposals that have been made so far.
 		pub PublicPropCount get(fn public_prop_count) build(|_| 0 as PropIndex) : PropIndex;
-		/// The public proposals. Unsorted.
-		pub PublicProps get(fn public_props): Vec<(PropIndex, T::Proposal, T::AccountId)>;
+		/// The public proposals. Unsorted. The second item is the proposal's hash.
+		pub PublicProps get(fn public_props): Vec<(PropIndex, T::Hash, T::AccountId)>;
+		/// Map of hashes to the proposal preimage, along with who registered it and their deposit.
+		/// The block number is the block at which it was deposited.
+		pub Preimages: map T::Hash => Option<(Vec<u8>, T::AccountId, BalanceOf<T>, T::BlockNumber)>;
```
shawntabrizi: A lot of items in this tuple. Why not use a struct?

Also why use a `Vec<u8>` rather than a `T::Proposal`?

frame/democracy/src/lib.rs
shawntabrizi: Maybe an authorization check, which skips the "date" check.

The owner of the data can call this function at any time, but if someone else wants to reap the rewards, they need to wait until after expiry.

gavofyork: it's a race, but it's not brutal - the only reason you'd ever use have funds reserved is if you know for sure that the proposal will pass before it hits the enactment period; otherwise, you'll just use the `_imminent` variant.

frame/democracy/src/lib.rs
gavofyork: if you propose a hash that resolves to a blob which is not a valid `Proposal` encoding, then you deserve to lose access to the reserved funds.

frame/democracy/src/lib.rs
gavofyork: using a `Proposal` complicates the economics of `note_preimage`.

basically, proposals can be really big and laborious to `encode` and `decode`; we can work around this by keeping them an opaque blob, at least until we have collected funds and/or know we want to use it. if we don't, then anyone can craft an annoying complex (to encode/decode) proposal and push it in here from an account that doesn't have enough funds to reserve. this would open up a griefing attack, forcing us to attach a large and/or complex weight/cost to calling this function.

keeping it opaque avoids this.

frame/democracy/src/lib.rs
gavofyork: addressed.

frame/democracy/src/lib.rs
```diff
@@ -227,39 +229,48 @@ pub trait Trait: system::Trait + Sized {
...
 
 /// Info regarding an ongoing referendum.
 #[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug)]
-pub struct ReferendumInfo<BlockNumber: Parameter, Proposal: Parameter> {
+pub struct ReferendumInfo<BlockNumber: Parameter, Hash: Parameter> {
 	/// When voting on this referendum will end.
 	end: BlockNumber,
-	/// The proposal being voted on.
-	proposal: Proposal,
+		/// The hash of the proposal being voted on.
```
```suggestion
	/// The hash of the proposal being voted on.
```

GitHub white spacing rendering is an issue.... sorry :/

frame/democracy/src/lib.rs
```diff
@@ -364,11 +399,10 @@ decl_module! {
 		/// # </weight>
 		#[weight = SimpleDispatchInfo::FixedNormal(5_000_000)]
```
marcio-diaz: Maybe update `weight`, a transfer is `1_000_000`.

gavofyork:
> Looks good. Although, I don't get why we can't reap the pre-image automatically (on `bake_referendum`?).

because it may not have an associated `bake_referendum`.

frame/democracy/src/lib.rs
gavofyork: i'd prefer to address all weights in a separate PR.

PR #4257, opened 2019-12-01
Introduce `VestingCurrency` trait

shawntabrizi:
This takes the "vesting" part of the Balances module into its own `VestingCurrency` trait.

This will help solve the problem outlined here:
https://github.com/paritytech/polkadot/issues/612

In summary, vesting can only be set directly in the balances module through genesis config. The `claims` module wants to apply some vesting schedule if the user claims their tokens after genesis.

We extract the following functionality into this new trait:

* `vesting_balance`: Returns the vesting balance of a user.
* `add_vesting_schedule`: Allows you to add a new vesting schedule to an account. It will fail if there is any existing schedule.
* `remove_vesting_schedule`: Allows you to remove a vesting schedule from an account.

I believe then this will allow us to use this trait in the Claims module where it will be able to introduce some logic around vesting.

### DIFF
--- a/frame/balances/src/lib.rs
+++ b/frame/balances/src/lib.rs
@@ -167,7 +167,7 @@ use support::{
 	traits::{
 		UpdateBalanceOutcome, Currency, OnFreeBalanceZero, OnUnbalanced, TryDrop,
 		WithdrawReason, WithdrawReasons, LockIdentifier, LockableCurrency, ExistenceRequirement,
-		Imbalance, SignedImbalance, ReservableCurrency, Get,
+		Imbalance, SignedImbalance, ReservableCurrency, Get, VestingCurrency,
 	},
 	weights::SimpleDispatchInfo,
 	dispatch::Result,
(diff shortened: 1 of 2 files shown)


frame/balances/src/lib.rs
```diff
@@ -1209,6 +1196,44 @@ where
...
+		locked: T::Balance,
+		per_block: T::Balance,
+		starting_block: T::BlockNumber
+	) {
+		let vesting_schedule = VestingSchedule {
+			locked,
+			per_block,
+			starting_block
+		};
+		<Vesting<T, I>>::insert(who, vesting_schedule);
```
gavofyork: this would clobber any pre-existing vesting schedule. it should probably return a result and fail if there already exists one.

shawntabrizi: Updated to `add_vesting_schedule`, which returns `Err` if `Vesting::exists(who)`.

frame/support/src/traits.rs
```diff
@@ -602,6 +602,29 @@ pub trait LockableCurrency<AccountId>: Currency<AccountId> {
...
+
+	/// Adds a vesting schedule to a given account.
+	///
+	/// If there already exists a vesting schedule for the given account, an `Err` is returned
+	/// and nothing is updated.
+	fn add_vesting_schedule(
+		who: &AccountId,
+		locked: Self::Balance,
+		per_block: Self::Balance,
+		starting_block: Self::Moment,
```
MOZGIII, line 621: Since the `Self::Moment` can be anything, why don't we change the `per_block` and `starting_block` to `per_moment` and `starting_moment`?
Also, would it work with timestamp pallet's UNIX milliseconds instead of block numbers? Seem like it should.

PR #4261, opened 2019-12-01
pallet-evm: weight calculation for call/create based on gas provided

sorpaas:
This makes `EVM::call` and `EVM::create` to calculate their weights based on gas provided. It does not yet refund unused gas.


frame/evm/src/lib.rs
```diff
@@ -23,13 +23,14 @@ mod backend;
...
-use rstd::vec::Vec;
+use rstd::{vec::Vec, marker::PhantomData};
 use support::{dispatch::Result, decl_module, decl_storage, decl_event};
+use support::weights::{Weight, WeighData, ClassifyDispatch, DispatchClass, PaysFee};
 use support::traits::{Currency, WithdrawReason, ExistenceRequirement};
 use system::ensure_signed;
 use sp_runtime::ModuleId;
 use support::weights::SimpleDispatchInfo;
-use sp_runtime::traits::{UniqueSaturatedInto, AccountIdConversion};
+use sr_runtime::traits::{UniqueSaturatedInto, AccountIdConversion, SaturatedConversion};
```
niklasad1:
```suggestion
use sp_runtime::traits::{UniqueSaturatedInto, AccountIdConversion, SaturatedConversion};
```

PR #4262, opened 2019-12-01
Improve error message on `can_author_with` failure

bkchr:


primitives/sr-version/src/lib.rs
```diff
@@ -161,10 +161,30 @@ pub struct NativeVersion {
...
-	pub fn can_author_with(&self, other: &RuntimeVersion) -> bool {
-		self.runtime_version.spec_name == other.spec_name &&
-			(self.runtime_version.authoring_version == other.authoring_version ||
-			self.can_author_with.contains(&other.authoring_version))
+	///
+	/// # Return
+	///
+	/// - Returns `Ok(())` when authoring is supported.
+	/// - Returns `Err(_)` with a detailed error when authoring is not supported.
+	pub fn can_author_with(&self, other: &RuntimeVersion) -> Result<(), String> {
```
niklasad1: not really related to this PR, but why doesn't `NativeVersion` impl the `trait CanAuthorWith` instead?

PR #4266, opened 2019-12-02
Fix phragmen elections

gavofyork:
CC @kianenigma

### DIFF
--- a/frame/elections-phragmen/src/lib.rs
+++ b/frame/elections-phragmen/src/lib.rs
@@ -84,9 +84,8 @@
 
 use rstd::prelude::*;
 use sp_runtime::{print, traits::{Zero, StaticLookup, Bounded, Convert}};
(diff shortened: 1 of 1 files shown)


frame/elections-phragmen/src/lib.rs
```diff
@@ -826,9 +811,25 @@ mod tests {
 		fn get() -> u64 { TERM_DURATION.with(|v| *v.borrow()) }
 	}
 
+	thread_local! {
```
bkchr:
```suggestion
thread_local! {
```

gavofyork:
there is "a test"; i added `TestChangeMembers`, which makes *all* tests check this logic :-)

bkchr:
> there is "a test"; i added `TestChangeMembers`, which makes _all_ tests check this logic :-)

Good point :) :+1:

PR #4274, opened 2019-12-02
client/authority-discovery: Fix interval_at to fire on start

mxinden:
The authority discovery module publishes and queries (onto) the DHT on an interval. Given that the `futures-timer` crate removed `Interval` the concept of an `Interval` was introduced into the authority discovery module directly.

In order for the custom implementation to equal the old `futures-timer` implementation, it should support an instant `Interval` as well as *queuing* events when the `Interval` is not polled for a while.

---

Maybe worth discussing after this pull request:

I don't think the authority discovery module should have its own implementation of `Interval`, but rather I would like it to move to its own library, or us depending on an alternative to `futures-timer` with an `Interval` equivalent implementation.

As of now the unit smoke tests depend on timing assumptions that might not hold in all environments (milliseconds scale). With the paragraph above in mind I would like to remove them longterm.

### DIFF
--- a/client/authority-discovery/src/lib.rs
+++ b/client/authority-discovery/src/lib.rs
@@ -457,14 +457,11 @@ fn hash_authority_id(id: &[u8]) -> Result<libp2p::kad::record::Key> {
 }
 
 fn interval_at(start: Instant, duration: Duration) -> Interval {
-	let stream = futures::stream::unfold((), move |_| {
-		let wait_time = start.saturating_duration_since(Instant::now());
+	let stream = futures::stream::unfold(start, move |next| {
+		let time_until_next =  next.saturating_duration_since(Instant::now());
 
-		futures::future::join(
-			Delay::new(wait_time),
-			Delay::new(duration)
-		).map(|_| Some(((), ())))
-	}).map(drop);
+		Delay::new(time_until_next).map(move |_| Some(((), next + duration)))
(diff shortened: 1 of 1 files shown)


mxinden:
If we don't care about the idea of *queuing* events when not polled for a while, we could also go with:

```rust
use futures::{FutureExt, StreamExt, Stream};
use futures_timer::Delay;
use std::time::{Duration, Instant};
use std::pin::Pin;

fn interval_at(start: Instant, duration: Duration) -> Pin<Box<dyn Stream<Item = ()> + Send>> {
    let initial_delay = Delay::new(start.saturating_duration_since(Instant::now()));
    let interval = futures::stream::repeat(()).then(move |_| Delay::new(duration));

    initial_delay.into_stream()
        .chain(interval)
        .boxed()
}

fn main() {
    futures::executor::block_on(async {
        let mut interval = interval_at(Instant::now() + Duration::from_secs(5), Duration::from_secs(1));
        while let Some(()) = interval.next().await {
            println!("Hello, world!");
        }
    })
}
```

(@montekki thanks for the input)

montekki:
```rust
fn interval_at(start: Instant, duration: Duration) -> Pin<Box<dyn Stream<Item = ()> + Send>> {
```

Not really sure about this return type, esp. the `Pin` part.

PR #4280, opened 2019-12-03
Rename folder: primitives/sr-primitives -> primitives/runtime

kigawas:
This follows #4223, it should not break anything

`sr-*` can also be renamed in other prs


tomusdrw:
Just spoke with @ben and we will prepare a meta issue with the list of all expected renames. This should be a series of smaller PRs that shouldn't really be conflicting. I'll probably do it in few hours.

kigawas:
@tomusdrw 
How about the meta issue? I suggest we only rename one folder per pr to mitigate conflicts

tomusdrw:
@kigawas Sorry for the delay, here it is: https://github.com/paritytech/substrate/issues/4318

bkchr:
Needs to be resolved.

PR #4281, opened 2019-12-03
Overflow audits

DemiMarie-parity:
This adds checks for overflow in some places.  In others, it flags (using words like ERROR or WARN) where more attention is needed.


frame/assets/src/lib.rs
```diff
@@ -179,7 +179,9 @@ decl_module! {
 			ensure!(origin_balance >= amount, "origin account balance must be greater than or equal to the transfer amount");
 
 			Self::deposit_event(RawEvent::Transferred(id, origin, target.clone(), amount));
+			// we check above that this will not overflow
 			<Balances<T>>::insert(origin_account, origin_balance - amount);
+			// sum of all issuances cannot exceed u64, so this will not overflow WARN
```
gavofyork: this is not true, as i said in the chat. however, *in the polkadot codebase* `TotalIssuance` will make it above 2**96 and `Balance` is `u128`.

frame/authorship/src/lib.rs
```diff
@@ -238,6 +238,7 @@ impl<T: Trait> Module<T> {
 
 			T::EventHandler::note_uncle(
 				author.clone().unwrap_or_default(),
+				// WARN `uncle` cannot be after `now`
```
gavofyork: why is this true?

gavofyork: ok - i guess these are *our* uncles.

frame/balances/src/lib.rs
```diff
@@ -851,6 +864,7 @@ where
 	fn issue(mut amount: Self::Balance) -> Self::NegativeImbalance {
 		<TotalIssuance<T, I>>::mutate(|issued|
 			*issued = issued.checked_add(&amount).unwrap_or_else(|| {
+				// cannot overflow: max_value >= any Self::Balance
```
gavofyork: tautologies like this comment don't need to be in the codebase.

frame/balances/src/lib.rs
```diff
@@ -819,6 +831,7 @@ where
 	type NegativeImbalance = NegativeImbalance<T, I>;
 
 	fn total_balance(who: &T::AccountId) -> Self::Balance {
+		// WARN cannot overflow: total balance does not exceed u64
```
gavofyork: incorrect.

frame/authorship/src/lib.rs
DemiMarie-parity: @gavofyork I was not sure.  That is why I included this message!

DemiMarie-parity: I added an explicit check for this.

kianenigma:
What is the status here? 

if done, it would be nice to get it in for 2.0 @gnunicorn

gnunicorn:
needs a merge first.

gnunicorn:
someone needs to pick this up,  merge and resolve the conflicts first.

bkchr:
@DemiMarie-parity ?

DemiMarie-parity:
@bkchr will do

bkchr:
Needs a merge.

gnunicorn:

PR #4284, opened 2019-12-03
Extract consensus_gossip.rs and put it in its own crate

tomaka:
Extracted from #4125 

This extracts `consensus_gossip.rs` from the networking crate and puts it in its own crate.

In details:

- Removes `consensus_gossip.rs` from the network, as well as the gossiping-related methods: `with_gossip` and `gossip_consensus_message`.
- Instead, you can register a gossiping protocol using `register_notifications_protocol`. Call `event_stream` in order to obtain a stream of the events that happen on the network. Call `write_notifications` in order to send a message to a peer we're connected to.
- The `NetworkService` no longer returns events and is now a `Future` again. Everything is done through `event_stream`.
- The abstraction level of `finality_grandpa::Network` has changed to be lower-level. It's GrandPa that sets up the gossiping system.
- Adds a new `executor` parameter to `GrandpaParams` and to `run_grandpa_observer`, which must implement the `futures::task::Spawn` trait and that is used to spawn GrandPa's background tasks. This trait has been implemented on `service::SpawnTaskHandler`, so that it's the service that spawns GrandPa's tasks (which is also something we've been willing to do for a while).

**Note**: the Polkadot PR is **not** ready. I'll work on it while this gets reviewed.


client/finality-grandpa/src/communication/mod.rs
```diff
@@ -97,50 +95,6 @@ mod benefit {
 	pub(super) const PER_EQUIVOCATION: i32 = 10;
 }
 
-/// A handle to the network. This is generally implemented by providing some
-/// handle to a gossip service or similar.
-///
-/// Intended to be a lightweight handle such as an `Arc`.
-pub trait Network<Block: BlockT>: Clone + Send + 'static {
```
rphmeier: I wouldn't mind something like this going into `consensus-common` either - where we can focus on the properties a network needs to provide to a consensus engine.

client/finality-grandpa/src/communication/tests.rs
```diff
@@ -46,56 +43,36 @@ struct TestNetwork {
...
 		let _ = self.sender.unbounded_send(Event::Announce(block));
 	}
 
-	/// Notify the sync service to try syncing the given chain.
-	fn set_sync_fork_request(&self, _peers: Vec<network::PeerId>, _hash: Hash, _number: NumberFor<Block>) {}
+	fn set_sync_fork_request(
+		&self,
+		_peers: Vec<network::PeerId>,
+		_hash: Hash,
+		_number: NumberFor<Block>
```
rphmeier:
```suggestion
		_number: NumberFor<Block>,
```

client/finality-grandpa/src/communication/tests.rs
```diff
@@ -104,14 +81,19 @@ impl network_gossip::ValidatorContext<Block> for TestNetwork {
 	fn broadcast_message(&mut self, _: Hash, _: Vec<u8>, _: bool) {	}
 
 	fn send_message(&mut self, who: &network::PeerId, data: Vec<u8>) {
-		<Self as super::Network<Block>>::send_message(self, vec![who.clone()], data);
+		<Self as network_gossip::Network<Block>>::write_notification(
+			self,
+			who.clone(),
+			GRANDPA_ENGINE_ID,
+			data
```
rphmeier:
```suggestion
			data,
```

client/finality-grandpa/src/communication/tests.rs
```diff
@@ -287,11 +271,14 @@ fn good_commit_leads_to_relay() {
 			// send a message.
 			let sender_id = id.clone();
 			let send_message = tester.filter_network_events(move |event| match event {
-				Event::MessagesFor(topic, sender) => {
-					if topic != global_topic { return false }
-					let _ = sender.unbounded_send(network_gossip::TopicNotification {
-						message: commit_to_send.clone(),
-						sender: Some(sender_id.clone()),
+				Event::EventStream(sender) => {
+					let _ = sender.unbounded_send(NetworkEvent::NotifOpened {
```
rphmeier: I'm not a huge fan of the abbreviated naming

client/finality-grandpa/src/environment.rs
```diff
@@ -375,20 +375,20 @@ impl<Block: BlockT> SharedVoterSetState<Block> {
...
 	pub(crate) authority_set: SharedAuthoritySet<Block::Hash, NumberFor<Block>>,
 	pub(crate) consensus_changes: SharedConsensusChanges<Block::Hash, NumberFor<Block>>,
-	pub(crate) network: crate::communication::NetworkBridge<Block, N>,
+	pub(crate) network: crate::communication::NetworkBridge<Block>,
 	pub(crate) set_id: SetId,
 	pub(crate) voter_set_state: SharedVoterSetState<Block>,
 	pub(crate) voting_rule: VR,
 }
 
-impl<B, E, Block: BlockT, N: Network<Block>, RA, SC, VR> Environment<B, E, Block, N, RA, SC, VR> {
```
rphmeier: 👍 good that we're removing some bounds here.

client/finality-grandpa/src/tests.rs
```diff
@@ -456,6 +459,7 @@ fn add_forced_change(
 fn finalize_3_voters_no_observers() {
 	let _ = env_logger::try_init();
 	let mut runtime = current_thread::Runtime::new().unwrap();
+	let threads_pool = futures03::executor::ThreadPool::new().unwrap();
```
rphmeier: @romanb migrated tests in the external `finality-grandpa` crate over to `LocalPool`. Would that be better here also? Does it make a difference?

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,299 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use crate::Network;
```
rphmeier: missing docs in crate

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,299 @@
...
+		let gossip_engine = GossipEngine {
+			inner: inner.clone(),
+			engine_id,
+		};
+
+		let res = executor.spawn({
+			let inner = Arc::downgrade(&inner);
+			async move {
+				loop {
+					let _ = futures_timer::Delay::new(Duration::from_millis(1100)).await;
```
rphmeier: why the delay?

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,299 @@
...
+			let mut stream = Compat01As03::new(event_stream);
+			while let Some(Ok(event)) = stream.next().await {
+				match event {
+					Event::NotifOpened { remote, engine_id: msg_engine_id } => {
+						if msg_engine_id != engine_id {
+							continue;
+						}
+						let mut inner = inner.lock();
+						let inner = &mut *inner;
+						// TODO: for now we hard-code the roles to FULL; fix that
```
rphmeier: that should be done in this PR, or an issue should be filed. Is that a regression?

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,299 @@
...
+		if res.is_err() {
+			log::error!(target: "gossip", "Failed to spawn background task");
+		}
+
+		let res = executor.spawn(async move {
+			let mut stream = Compat01As03::new(event_stream);
+			while let Some(Ok(event)) = stream.next().await {
+				match event {
+					Event::NotifOpened { remote, engine_id: msg_engine_id } => {
+						if msg_engine_id != engine_id {
```
rphmeier: i don't get why it's necessary to echo the engine ID in the notifications. shouldn't it be implicit from context?

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,299 @@
...
+use network::message::generic::ConsensusMessage;
+use network::{Event, config::Roles, ReputationChange};
+
+use futures::{prelude::*, channel::mpsc, compat::Compat01As03, task::SpawnExt as _};
+use libp2p::PeerId;
+use parking_lot::Mutex;
+use sp_runtime::{traits::{Block as BlockT, NumberFor}, ConsensusEngineId};
+use std::{sync::Arc, time::Duration};
+
+pub struct GossipEngine<B: BlockT> {
```
rphmeier: missing docs

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,299 @@
...
+	pub fn report(&self, who: PeerId, reputation: ReputationChange) {
+		self.inner.lock().context.report_peer(who, reputation);
+	}
+
+	/// Registers a message without propagating it to any peers. The message
+	/// becomes available to new peers or when the service is asked to GossipEngine
+	/// the message's topic. No validation is performed on the message, if the
+	/// message is already expired it should be dropped on the next garbage
+	/// collection.
+	pub fn register_gossip_message(
```
rphmeier: Why would we register a message without propagating to existing peers, but send it to new peers? Is this used somewhere?

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,299 @@
...
+	fn report_peer(&mut self, who: PeerId, reputation: ReputationChange) {
+		self.network.report_peer(who, reputation);
+	}
+
+	fn disconnect_peer(&mut self, who: PeerId) {
+		self.network.disconnect_peer(who)
+	}
+
+	fn send_consensus(&mut self, who: PeerId, messages: Vec<ConsensusMessage>) {
+		// TODO: send batch
```
rphmeier: can't merge TODO

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,299 @@
...
+	/// Notifies the sync service to try and sync the given block from the given
+	/// peers.
+	///
+	/// If the given vector of peers is empty then the underlying implementation
+	/// should make a best effort to fetch the block from any peers it is
+	/// connected to (NOTE: this assumption will change in the future #3629).
+	///
+	/// Note: this method isn't strictly related to gossiping and should eventually be moved
+	/// somewhere else.
+	pub fn set_sync_fork_request(&self, peers: Vec<network::PeerId>, hash: B::Hash, number: NumberFor<B>) {
```
rphmeier: Is this really something that should be provided by the gossip crate? Not a blocker, just seems less than ideal.

client/network-gossip/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+pub use self::bridge::GossipEngine;
```
rphmeier: module docs please!

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,299 @@
...
+			let inner = Arc::downgrade(&inner);
+			async move {
+				loop {
+					let _ = futures_timer::Delay::new(Duration::from_millis(1100)).await;
+					if let Some(inner) = inner.upgrade() {
+						let mut inner = inner.lock();
+						let inner = &mut *inner;
+						inner.state_machine.tick(&mut *inner.context);
+					} else {
+						break;
```
mxinden: Can you add a note here that this branch is reached whenever the `GossipEngine` and thus the `GossipEngineInner` is dropped and thus the tick is not needed no more?

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,299 @@
...
+	pub fn abort(&self) {
+		self.inner.lock().state_machine.abort();
+	}
+
+	pub fn report(&self, who: PeerId, reputation: ReputationChange) {
+		self.inner.lock().context.report_peer(who, reputation);
+	}
+
+	/// Registers a message without propagating it to any peers. The message
+	/// becomes available to new peers or when the service is asked to GossipEngine
```
mxinden: What does it mean `to GossipEngine` a message?

client/network-gossip/src/bridge.rs
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	/// Broadcast all messages with given topic.
+	pub fn broadcast_topic(&self, topic: B::Hash, force: bool) {
+		let mut inner = self.inner.lock();
+		let inner = &mut *inner;
+		inner.state_machine.broadcast_topic(&mut *inner.context, topic, force);
+	}
+
+	/// Get data of valid, incoming messages for a topic (but might have expired meanwhile)
```
mxinden:
```suggestion
	/// Get data of valid, incoming messages for a topic (but might have expired meanwhile).
```

client/network-gossip/src/bridge.rs
tomaka: This is a much deeper refactoring than the scope of this PR.

client/network-gossip/src/bridge.rs
tomaka: We get events from all the registered protocols, not just ours.

client/finality-grandpa/src/tests.rs
tomaka: The `ThreadPool` is a global threads pool where tasks are dispatched.
The difference is that with a `LocalPool` things are more deterministic, but if the code doesn't have any bug, that shouldn't change anything.

client/network-gossip/src/bridge.rs
tomaka: That was already there before my refactoring.

client/network/src/service.rs
```diff
@@ -716,34 +751,42 @@ impl<B: BlockT + 'static, S: NetworkSpecialization<B>, H: ExHashT> Stream for Ne
...
-					return Ok(Async::Ready(Some(Event::Dht(ev)))),
-				Ok(Async::Ready(None)) => CustomMessageOutcome::None,
+				Ok(Async::Ready(Some(BehaviourOut::BlockImport(origin, blocks)))) =>
+					self.import_queue.import_blocks(origin, blocks),
+				Ok(Async::Ready(Some(BehaviourOut::JustificationImport(origin, hash, nb, justification)))) =>
+					self.import_queue.import_justification(origin, hash, nb, justification),
+				Ok(Async::Ready(Some(BehaviourOut::FinalityProofImport(origin, hash, nb, proof)))) =>
+					self.import_queue.import_finality_proof(origin, hash, nb, proof),
+				Ok(Async::Ready(Some(BehaviourOut::Event(ev)))) => {
+					self.event_streams.retain(|sender| sender.unbounded_send(ev.clone()).is_ok());
```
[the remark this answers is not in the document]
tomaka: The notification messages within `Event` are `Bytes`.

client/network-gossip/src/bridge.rs
tomaka: `tick` is to be called periodically, hence the delay.
This `loop { Delay::new(...).await; /* foo */ }` is same as `Interval::new(...).for_each(/* foo */)`.

client/network-gossip/src/bridge.rs
tomaka: There's now a `ConsensusBatch` message that groups multiple consensus messages into one, but I don't really see what this brings and looks like an XY solution to me.
This TODO is here because ideally we should send a `ConsensusBatch` message, but I also don't think there's a point in doing so.

tomaka: If we split the gossiping into multiple substreams (as planned), then the `ConsensusBatch` will disappear anyway.

client/network-gossip/src/bridge.rs
mxinden: How about adding a note here, creating a GitHub issue and following up on it later? 

(I am happy to help tackle this in a follow up pull request.)

client/network/src/service.rs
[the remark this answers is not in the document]
mxinden: But e.g. a `DhtEvent::ValueFound` is not. I guess we can follow up on this later on in case it ever crystallizes as a performance bottleneck.

client/network-gossip/src/bridge.rs
rphmeier: oh, that's kind of strange. why do we get events from all subprotocols?

client/network-gossip/src/bridge.rs
rphmeier: > This TODO is here because ideally we should send a ConsensusBatch message, but I also don't think there's a point in doing so.

That's fine, but it needs an issue filed if you want this PR to be merged

tomaka:
Since I'm passing a `ThreadsPool` in order to execute tasks, tokio-timer's `Delay` unfortunately errors. I therefore transitioned `finality-grandpa` to `futures-timer`.

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -346,18 +183,13 @@ impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
...
-			Ok(())
-		});
+		let executor = Compat::new(executor);
+		executor.execute(Box::new(rebroadcast_job.select(on_exit.clone()).then(|_| Ok(()))))
+			.expect("failed to spawn grandpa rebroadcast job task");
+		executor.execute(Box::new(reporting_job.select(on_exit.clone()).then(|_| Ok(()))))
+			.expect("failed to spawn grandpa reporting job task");
 
-		(bridge, startup_work)
+		(bridge, futures::future::ok(()))
```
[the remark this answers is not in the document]
tomaka: That's done

client/network-gossip/src/bridge.rs
tomaka: I don't think there's any real drawback in doing so, except that we duplicate all messages once event stream (which is why the message itself is a `Bytes`, which is kind of an `Arc<Vec<u8>>`, rather than just a `Vec<u8>`).

I did that in order to not inflict ourself any artificial constraint later, for example if you want the same code to register two different protocols, or if you want to know whether the node you're talking to also has a substream to a different specific protocol open.

And it's also easier to implement this way.

mxinden:
Let us know once this is ready for another review @tomaka.

tomaka:
This is ready. All tests are passing and documentation has been written.

The corresponding Polkadot PR still needs to be written, though.

client/network/src/protocol.rs
```diff
@@ -1758,6 +1794,12 @@ pub enum CustomMessageOutcome<B: BlockT> {
 	BlockImport(BlockOrigin, Vec<IncomingBlock<B>>),
 	JustificationImport(Origin, B::Hash, NumberFor<B>, Justification),
 	FinalityProofImport(Origin, B::Hash, NumberFor<B>, Vec<u8>),
+	/// Notifications protocols have been opened with a remote.
```
mxinden:
```suggestion
	/// Notification protocols have been opened with a remote.
```

client/network/src/protocol.rs
```diff
@@ -1758,6 +1794,12 @@ pub enum CustomMessageOutcome<B: BlockT> {
...
+	NotificationsStreamOpened { remote: PeerId, protocols: Vec<ConsensusEngineId>, roles: Roles },
```
mxinden:
```suggestion
	NotificationStreamOpened { remote: PeerId, protocols: Vec<ConsensusEngineId>, roles: Roles },
```

client/service/src/lib.rs
```diff
@@ -451,22 +462,32 @@ fn build_network_future<
 			(status, state)
 		});
 
+		// Processing DHT events.
+		while let Ok(Async::Ready(Some(event))) = event_stream.poll() {
```
mxinden: Given that the authority discovery module has a reference to the network service, it could register its own event stream, thus the `build_network_future` function does not need to be aware of the authority discovery module no more, correct? Happy to do that in a follow up.

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -151,157 +105,37 @@ pub(crate) fn global_topic<B: BlockT>(set_id: SetIdNumber) -> B::Hash {
...
-					Err(_) => Err(())
-				}
-			},
-			NetworkStream::PollingTopicNotifications(stream) => {
-				stream.poll()
-			},
-		}
-	}
+/// Registers the notifications protocol towards the network.
+pub(crate) fn register_dummy_protocol<B: BlockT, N: Network<B>>(network: N) {
```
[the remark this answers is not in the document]
tomaka: It's a dummy because don't handle the incoming messages. We don't even create any channel, so there's no resource leak.

client/service/src/lib.rs
tomaka: Exactly. I was leaving that for a later refactoring, but that's totally feasible.

client/finality-grandpa/src/communication/mod.rs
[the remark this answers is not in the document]
mxinden: I find this indirection a bit confusing. Why not do:

```rust
	network.register_notifications_protocol(fg_primitives::GRANDPA_ENGINE_ID);
```

in `lib.rs`?

tomaka:
Small update: Polkadot PR almost ready, I just have to fix a tests module.

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -151,157 +105,37 @@ pub(crate) fn global_topic<B: BlockT>(set_id: SetIdNumber) -> B::Hash {
...
-
+	) -> Self {
 		let (validator, report_stream) = GossipValidator::new(
 			config,
 			set_state.clone(),
 		);
 
 		let validator = Arc::new(validator);
-		service.register_validator(validator.clone());
+		let gossip_engine = GossipEngine::new(gossip_engine, executor, GRANDPA_ENGINE_ID, validator.clone());
```
mxinden: My previous comment might not have been very clear. I meant to rename the left side of the equation. Now one creates a `GossipEngine` with a `gossip_engine` as an argument. The latter should be a `network` or a `service`, no?

tomaka:
Polkadot PR now ready: https://github.com/paritytech/polkadot/pull/680

client/network-gossip/src/bridge.rs
mxinden: Would be fixed via https://github.com/paritytech/substrate/pull/4359.

client/service/src/lib.rs
mxinden: Would be fixed with https://github.com/paritytech/substrate/pull/4344.

PR #4285, opened 2019-12-03
client: Introduce --sentry-nodes flag

mxinden:
Enable operators to specify the public addresses of sentry nodes infront of a validator node so that the validator node can announce the sentry node addresses instead of its own public addresses on the DHT via the authority discovery module.

### Commit structure

This pull request is based on https://github.com/paritytech/substrate/pull/4274, thus it should only be merged after https://github.com/paritytech/substrate/pull/4274. One can ignore the first two commits when reviewing this pull request.

Commit 3 (updating `.maintain/sentry-node/docker-compose.yml`) and commit 5 (shortening line length to 100 instead of 120 char) should not need much review.

Commit 4 (https://github.com/paritytech/substrate/commit/a5b5db7f4c1bda0ac906bb04fa67826eabbe9b5b) is the core of this pull request.

### Polkadot

Given that this changes the signature of `AuthorityDiscovery::new`, it will break Polkadot. I will follow up with another pull request on github.com/pairtytech/polkadot.

### DIFF
--- a/.maintain/sentry-node/docker-compose.yml
+++ b/.maintain/sentry-node/docker-compose.yml
@@ -38,13 +38,12 @@ services:
       - "--base-path"
       - "/tmp/alice"
       - "--chain=local"
-      - "--key"
-      - "//Alice"
       - "--port"
       - "30333"
       - "--validator"
-      - "--name"
-      - "AlicesNode"
+      - "--alice"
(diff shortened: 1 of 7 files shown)


mxinden:
Corresponding Polkadot pull request: https://github.com/paritytech/polkadot/pull/648.

PR #4290, opened 2019-12-04
Use balance capped to u64 for transaction payment RPC.

tomusdrw:
`payment_fee` is not likely to ever exceed `u64` anyway, and by not using `Balance` directly we can avoid running into issues with `u128` balances.

Fixes: https://github.com/paritytech/polkadot/issues/654


gavofyork:
Ci failing.

PR #4292, opened 2019-12-04
Revalidate some transactions on every block import.

tomusdrw:
This should fix transactions stalling in the pool for a long time.

I'm still running tests for Kusama to make sure it works correctly.

### DIFF
--- a/client/transaction-pool/src/maintainer.rs
+++ b/client/transaction-pool/src/maintainer.rs
@@ -23,7 +23,7 @@ use futures::{
 	Future, FutureExt,
 	future::{Either, join, ready},
 };
-use log::{warn, debug};
+use log::{warn, debug, trace};
 use parking_lot::Mutex;
 
 use client_api::{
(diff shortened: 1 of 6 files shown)


client/transaction-pool/src/maintainer.rs
```diff
@@ -74,6 +74,14 @@ where
 		id: &BlockId<Block>,
 		retracted: &[Block::Hash],
 	) -> Box<dyn Future<Output=()> + Send + Unpin> {
+		let now = std::time::Instant::now();
+		let took = move || {
+			let took = std::time::Instant::now().duration_since(now);
+			format!("Took {} ms", took.as_millis())
+		};
+
```
marcio-diaz: Next time you can do something like
```
use tracing;
let span = tracing::span!(tracing::Level::DEBUG, "retracted");
let _enter = span.enter();
```
inside a context, to send the time to trace/grafana.

tomusdrw: Can I also display that in logs?

marcio-diaz: Yeah, you need to use `--tracing-targets`  with an appropriated pattern (maybe `substrate_client`) and `--tracing-receiver` (but is defaulting to Log).

tomusdrw:
Run some tests on Kusama and it seems to be working correctly (as in: not breaking anything). We now revalidate 16 transactions from the front of the transaction pool (after pruning), to make sure that what we have is still valid for the next block.

gavofyork:
Merged since CI failure (no connectivity) clearly spurious.

PR #4293, opened 2019-12-04
Add linear back-off for aura slot workers

drewstone:
This implements the linear-back off that exists for babe worker, for the aura worker. It does so by simple piggybacking off the previous slot number.

@rphmeier this still hasn't solved my issue as I just tried running an upgrade where all I did is increment the spec/impl versions. The upgrade still stalled. I'm using the polkadotjs UI, maybe there's something in my process I'm doing incorrectly?

### DIFF
--- a/client/consensus/aura/src/lib.rs
+++ b/client/consensus/aura/src/lib.rs
@@ -321,8 +321,11 @@ impl<H, B, C, E, I, P, Error, SO> slots::SimpleSlotWorker<B> for AuraWorker<C, E
 		// in normal cases we only attempt to issue blocks up to the end of the slot.
 		// when the chain has been stalled for a few slots, we give more lenience.
 		let slot_lenience = slot_info.number.saturating_sub(parent_slot + 1);
+		println!("{:?}, {:?}, {:?}", slot_remaining, parent_slot, slot_lenience);
 		let slot_lenience = std::cmp::min(slot_lenience, BACKOFF_CAP);
+		println!("{:?}, {:?}, {:?}", slot_remaining, parent_slot, slot_lenience);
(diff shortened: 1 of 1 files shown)


client/consensus/aura/src/lib.rs
```diff
@@ -302,6 +302,29 @@ impl<H, B, C, E, I, P, Error, SO> slots::SimpleSlotWorker<B> for AuraWorker<C, E
...
+		let slot_remaining = self.slot_remaining_duration(slot_info);
+		let parent_slot = slot_info.last_number;
+
+		// we allow a lenience of the number of slots since the head of the
+		// chain was produced, minus 1 (since there is always a difference of at least 1)
+		//
+		// linear back-off.
+		// in normal cases we only attempt to issue blocks up to the end of the slot.
+		// when the chain has been stalled for a few slots, we give more lenience.
+		let slot_lenience = slot_info.number.saturating_sub(parent_slot + 1);
```
drewstone: From the looks of it, slot_lenience will always be 0 until it wraps around?

rphmeier: I think `parent_slot` here isn't right; it's the most recent slot number that the slot worker tried to author on (which will always be ~`n_authorities` slots ago in Aura). 

You should extract the slot from the parent header. #4020 has some code for that.

drewstone: Ok @rphmeier I presume this logic existed before I changed but its updated now. Testing soon once all this jazz builds.

drewstone:
Awesome, an upgrade went through with a bit of delay but successful nonethless! Thanks @rphmeier for the guidance

PR #4297, opened 2019-12-05
client/authority-discovery: Increase margins for interval test

mxinden:
`interval_at_is_queuing_ticks` test shows to be [flaky](https://gitlab.parity.io/parity/substrate/-/jobs/322417). This commit
increases the time margins while also reducing the amount of necessary
time ticks.

In case the test is still flaky on our CI system after this patch, I think we
should remove it.

### DIFF
--- a/client/authority-discovery/src/lib.rs
+++ b/client/authority-discovery/src/lib.rs
@@ -547,24 +547,26 @@ mod tests {
 	}
 
 	#[test]
-	fn interval_at_is_queuing_events() {
+	fn interval_at_is_queuing_ticks() {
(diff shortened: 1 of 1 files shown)


bkchr:
Just some general nodes, sleeps in tests are never good. We will probably see at some point that it will fail again when the CI system is under high load. However, I also don't have a good idea for testing a timer.

mxinden:
@bkchr I agree, but I rather have some tests that flake in **rare** occasions than no tests. I suggested to move `interval_at` into a separate crate, but people were not that excited about that idea.

PR #4298, opened 2019-12-05
Limit number of uncles that can be provided.

marcio-diaz:
Looks like the vector of headers in `set_uncles` can be unbounded.


gavofyork:
What stops the blockbuilder from blindly producing blocks which can't be imported because they have too many uncles? Just practical considerations?

marcio-diaz:
> What stops the blockbuilder from blindly producing blocks which can't be imported because they have too many uncles? Just practical considerations?

Right, I think implementing  `check_inherent` and fix `create_inherent` should be enough.  Or I'm missing something else?

I could add the same check to the number of uncles, but I need to check if suffices.

frame/authorship/src/lib.rs
```diff
@@ -314,14 +315,14 @@ impl<T: Trait> Module<T> {
...
-	type Error = MakeFatalError<()>;
+	type Error = InherentError;
 	const INHERENT_IDENTIFIER: InherentIdentifier = INHERENT_IDENTIFIER;
 
 	fn create_inherent(data: &InherentData) -> Option<Self::Call> {
 		let uncles = data.uncles().unwrap_or_default();
 		let mut set_uncles = Vec::new();
 
-		if !uncles.is_empty() {
+		if !uncles.is_empty() && uncles.len() <= MAX_UNCLES {
```
arkpar: Block builder might actually push more uncles than can be included since it can't pre-verify them. This function should just take MAX_UNCLES valid uncles.

marcio-diaz: Right, I guess it's safe here.

PR #4300, opened 2019-12-05
Allow members of groups to change their key

gavofyork:


frame/membership/src/lib.rs
```diff
@@ -186,6 +188,31 @@ decl_module! {
...
+		/// May only be called from `Signed` origin of a current member.
+		#[weight = SimpleDispatchInfo::FixedNormal(50_000)]
+		fn change_key(origin, new: T::AccountId) {
+			let remove = ensure_signed(origin)?;
+
+			if remove != new {
+				let mut members = <Members<T, I>>::get();
+				let location = members.binary_search(&remove).ok().ok_or("not a member")?;
+				members[location] = new.clone();
+				let _location = members.binary_search(&new).err().ok_or("already a member")?;
```
niklasad1: hmm, shouldn't this be performed after sorting?

After `removed` gets replaced with `new` the `Vec` is not necessarily sorted?

bkchr: Hmm? The vector is sorted after inserting the new key? 

When you insert the new key, the sorting is very likely not the same as before.

niklasad1: I missed that the `let _location = members.binary_search(&new).err().ok_or("already a member")?;` expected it be an error sorry about that.

I still find `Err(already a member)`confusing because if got `[10, 20, 30]` and want to replace `30 -> 40` it will end up in error: `already a member`

Am I misunderstanding something?

bkchr: Ahh yeah! Fuck that is right, we should search before inserting.

bkchr: Ahh, now I understand why the test works. Yep, that is an error.

PR #4302, opened 2019-12-05
client/authority-discovery: Limit the amount of sentries per authority

mxinden:
When receiving more addresses for a given authority than a defined
threshold (5), the authority discovery drops the remaining in order to
prevent a single authority to fill all priority group slots.


bkchr:
Do we also limit the number of addresses we send?

mxinden:
@bkchr not as of today. I added this as a defense against malicious authorities. What do you think of printing a warning when specifying more sentry nodes than the threshold allows @bkchr?

bkchr:
Sounds like an idea.

PR #4312, opened 2019-12-06
Disable record_metrics_slice in grafana macro on wasm

expenses:
This is a lot easier than just disabling the macro entirely.

### DIFF
--- a/client/grafana-data-source/src/lib.rs
+++ b/client/grafana-data-source/src/lib.rs
@@ -47,9 +47,13 @@ lazy_static! {
 #[macro_export]
 macro_rules! record_metrics(
 	($($key:expr => $value:expr,)*) => {
-		$crate::record_metrics_slice(&[
-			$( ($key, $value as f32), )*
-		]);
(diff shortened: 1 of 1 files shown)


client/grafana-data-source/src/lib.rs
```diff
@@ -47,9 +47,13 @@ lazy_static! {
 #[macro_export]
 macro_rules! record_metrics(
 	($($key:expr => $value:expr,)*) => {
-		$crate::record_metrics_slice(&[
-			$( ($key, $value as f32), )*
-		]);
+		if cfg!(not(target_os = "unknown")) {
```
bkchr:
```suggestion
		if cfg!(not(feature = "std")) {
```

expenses:
@bkchr Sorry, I should have been clearer! This is to prevent crashes when running the browser demo in `bin/node/cli/browser-demo`. This happens due to time system calls not being supported in wasm:

Screen Shot 2019-12-06 at 2 54 16 PM

Could you please revert that change?

PR #4317, opened 2019-12-06
Fix `check_polkadot` CI check

bkchr:


mxinden:
As far as I remember without the commit reference it would be looking for `master` instead of the commit itself.

Issue #4318, opened 2019-12-06
Remaining renames (metaissue)


tomusdrw:
A tentative list of issues that are still remaining to wrap up the "big refactor" thing.
Most likely it should be one PR per rename (i.e. keep them small and focused) - it's should be just a mechanical work.
Feel free to leave a comment if you work on one.

Crate names:
- [x] sc-application-crypto -> sp-application-crypto
- [x] sp-transaction-pool-api -> sp-transaction-pool
- [x] sp-core-storage -> sp-storage
- [x] transaction-factory -> substrate-transaction-factory
- [x] move chain-spec-builder to bin 
- [ ] sc-client-api -> sc-api ?
- [ ] sc-client-db -> sc-db ?

Folder names:
- [x] primitives/sr-arithmetic -> primitives/arithmetic
- [x] primitives/sr-api -> primitives/api
- [x] primitives/sr-io -> primitives/io
- [x] primitives/sr-primitives -> primitives/runtime
- [x] primitives/sr-sandbox -> primitives/sandbox
- [x] primitives/sr-staking-primitives -> primitives/staking
- [x] primitives/sr-std -> primitives/std
- [x] primitives/sr-version -> primitives/version
- [x] primitives/block-builder/runtime-api -> primitives/block-builder
- [x] test/utils -> test-utils
- [x] client/rpc/api -> client/rpc-api
- [x] client/utils/wasm-builder-runner -> utils/wasm-builder-runner
- [x] client/grafana-data-source -> utils/grafana-data-source (?)

Other:
- [ ] Merge `sp-serializer` with sp-core or sp-runtime
- [x] Merge `sp-transaction-pool-runtime-api` with `sp-transaction-pool`

tomusdrw:
Tackling:
> Merge sp-transaction-pool-runtime-api with sp-transaction-pool

gnunicorn:
refs #4099 

We also have to clean up the remaining `../client`-dependencies in `/primitives` ... maybe means moving `consensus-slots` from client to primitives...

tomusdrw:
Yeah, I think the `SlotData` trait just belongs to primitives. Either `common` or a new crate for `slots` primitives.

kigawas:
primitives/sr-primitives -> primitives/runtime
#4280 
primitives/sr-std -> primitives/std
#4327

gnunicorn:
[...]

PR #4320, opened 2019-12-06
Merge txpool-runtime-api with txpool-api

tomusdrw:
Related: #4318

To avoid crate bloat we merge runtime-api of the transaction pool (~10 lines crate) with the primitives of the transaction pool. The downside is `no_std` support, but I think it's manageable.


bkchr:
Hmm, more crates provide better parallelization opportunities for rustc/cargo. So, why do remove the number of crates? :D

tomusdrw:
> Hmm, more crates provide better parallelization opportunities for rustc/cargo. So, why do remove the number of crates? :D

That's a good point, but I also feel that too many crates causes cognitive overload. And while I'd like us to for instance to split `sp-core/sp-runtime` to smaller (more domain-specific) crates, I feel like separating `runtime-api` from it's primitives is just unnecessary. I'm not very opinionated on this particular change though, so if there is another person that has doubts let's just close this PR.

rockbmb:
> The downside is `no_std` support, but I think it's manageable.

Hello @tomusdrw,

In the sentence I quoted could you perhaps have meant

> The downside is no `no_std` support, but I think it's manageable.

?

tomusdrw:
@rockbmb I meant that there is additional maintenance cost to support `no_std`, because we need to add `default-features = false` to all crates and have `std` feature explicitly enable `std` feature of crates we depend on.
`no_std` is definitely supported if you, because we use that `RuntimeApi` in all default runtimes in this repo.

PR #4321, opened 2019-12-06
Remove depenendcy on sc-consensus-slots

tomusdrw:
This cleans up `[dependencies]` section of all `primitives` from `client` deps :tada: :)
We are only left with `[dev-dependencies]` now (I guess that requires moving the tests to separate `test` crate, cc @ben).


primitives/consensus/common/src/lib.rs
```diff
@@ -181,3 +181,20 @@ impl<Block: BlockT> CanAuthorWith<Block> for AlwaysCanAuthor {
...
 	}
 }
+
+/// A type from which a slot duration can be obtained.
+pub trait SlotData {
+	/// Gets the slot duration.
+	fn slot_duration(&self) -> u64;
+
+	/// The static slot key
+	const SLOT_KEY: &'static [u8];
```
niklasad1:
```suggestion
	const SLOT_KEY: &[u8];
```

primitives/consensus/common/src/lib.rs
```diff
@@ -181,3 +181,20 @@ impl<Block: BlockT> CanAuthorWith<Block> for AlwaysCanAuthor {
...
+}
+
+impl SlotData for u64 {
+	fn slot_duration(&self) -> u64 {
+		*self
+	}
+
+	const SLOT_KEY: &'static [u8] = b"aura_slot_duration";
```
niklasad1:
```suggestion
	const SLOT_KEY: &[u8] = b"aura_slot_duration";
```

tomusdrw:
I've tried to apply @niklasad1 suggestions, but I think omitting lifetime is not yet supported for traits:
```
   Compiling sc-executor v2.0.0 (/home/builds/remote-builds/substrate2/client/executor)
error[E0106]: missing lifetime specifier
   --> primitives/consensus/common/src/lib.rs:191:18
    |
191 |     const SLOT_KEY: &[u8];
    |                     ^ expected lifetime parameter

error[E0106]: missing lifetime specifier
   --> primitives/consensus/common/src/lib.rs:199:18
    |
199 |     const SLOT_KEY: &[u8] = b"aura_slot_duration";
    |                     ^ expected lifetime parameter

error: aborting due to 2 previous errors
```

niklasad1:
Sorry about that my suggestions were wrong, see https://github.com/rust-lang/rust/issues/51370

PR #4326, opened 2019-12-08
More comprehensive identity system

### LINKED ISSUE #4311: Extend naming system
Naming system could become more sophisticated:

Introduce ability for council to verify names. Deposit `FEE+BOND`; council verifies your information (by checking your website on which the account ID should be published, and querying you on riot); if good, `BOND` is returned, `FEE` goes to treasury. If bad, `BOND+FEE` goes to treasury.

In the future, more identification fields can be added.

```
#[derive(Encode, Decode)]
pub enum Field {
  Web,
  Riot,
}
/// Information that has been verified by the `VerifierOrigin` (which will initially be configured
/// to be the council, but could be a community-led oracle, too).
Verified: map T::AccountId => Vec<(Field, Vec<u8>)>;
```

Secondly, introduce basic vouching system, so that outfits who control a lot of public accounts (e.g. Cryptium) are able to vouch for each individually without bothering the council for each one.

```
/// Ordered account IDs of accounts that have been vouched for by the key. The balance
/// is how much has been deposited so far for registration. Except for fee changes, It should
/// be proportional to the number of entries. The size of the `Vec` should be capped at
/// something sensible (e.g. 100). If additional entries are needed, then multiple levels of
/// indirection can be used.
Vouched: map T::AccountId => (T::Balance, Vec<T::AccountId>);
```

gavofyork:
Closes #4311 

A federated naming system, allowing for multiple registrars to be added from a specified origin. Registrars can set a fee to provide identity-verification service. Anyone can put forth a proposed identity for a fixed deposit and ask for review by any number of registrars (paying each of their fees). Registrar judgements are multi-tier allowing for more sophisticated opinions than a boolean true or false.

A super-user can remove accounts and slash the deposit.

The number of registrars should be limited, and the deposit made sufficiently large, to ensure no state-bloat attack is viable.

@jacogr incoming! :)

TODO:
- [x] Tests
- [x] Docs


frame/identity/src/lib.rs
```diff
@@ -0,0 +1,1040 @@
...
+impl Default for Data {
+	fn default() -> Self {
+		Self::None
+	}
+}
+
+/// An identifier for a single name registrar/identity verification service.
+pub type RegistrarIndex = u32;
+
+/// An attesttion of a registrar over how accurate some `IdentityInfo` is in describing an account.
```
shawntabrizi:
```suggestion
/// An attestation of a registrar over how accurate some `IdentityInfo` is in describing an account.
```

frame/identity/src/lib.rs
```diff
@@ -0,0 +1,1040 @@
...
+}
+
+/// Either underlying data or a hash of it, if the data is less than 32 bytes.
+///
+/// Can also be `None`.
+#[derive(Clone, Eq, PartialEq, RuntimeDebug)]
+pub enum Data {
+	/// No data here.
+	None,
+	/// The data is stored directly.
```
tomusdrw: Might be good to mention that it's up to 32 bytes, and if you encode a bigger `Raw` value, the rest will be discarded.

frame/identity/src/lib.rs
```diff
@@ -0,0 +1,1040 @@
...
+	fn encode(&self) -> Vec<u8> {
+		match self {
+			Data::None => vec![0u8; 1],
+			Data::Raw(ref x) => {
+				let l = x.len().min(32);
+				let mut r = vec![l as u8 + 1; l + 1];
+				&mut r[1..].copy_from_slice(&x[..l as usize]);
+				r
+			}
+			Data::BlakeTwo256(ref h) => (34u8..35u8).chain(h.iter().cloned()).collect(),
```
tomusdrw: could be 
```suggestion
			Data::BlakeTwo256(ref h) => once(34u8).chain(h.iter().cloned()).collect(),
```

frame/identity/src/lib.rs
```diff
@@ -0,0 +1,1040 @@
...
+	Email          = 0b0000000000000000000000000000000000000000000000000000000000010000,
+	PgpFingerprint = 0b0000000000000000000000000000000000000000000000000000000000100000,
+	Image          = 0b0000000000000000000000000000000000000000000000000000000001000000,
+}
+
+/// Wrapper type for `BitFlags<IdentityField>` that implements `Codec`.
+#[derive(Clone, Copy, PartialEq, Default, RuntimeDebug)]
+pub struct IdentityFields(BitFlags<IdentityField>);
+
+impl Eq for IdentityFields {}
```
tomusdrw: This can be just derived, no?

frame/identity/src/lib.rs
```diff
@@ -0,0 +1,1040 @@
...
+impl Encode for IdentityFields {
+	fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
+		self.0.bits().using_encoded(f)
+	}
+}
+impl Decode for IdentityFields {
+	fn decode<I: codec::Input>(input: &mut I) -> rstd::result::Result<Self, codec::Error> {
+		let field = u64::decode(input)?;
+		Ok(Self(<BitFlags<IdentityField>>::from_bits(field as u64).map_err(|_| "invalid value")?))
```
tomusdrw:
```suggestion
		Ok(Self(<BitFlags<IdentityField>>::from_bits(field).map_err(|_| "invalid value")?))
```

frame/identity/src/lib.rs
```diff
@@ -0,0 +1,1040 @@
...
+			if old_deposit < new_deposit {
+				T::Currency::reserve(&sender, new_deposit - old_deposit)?;
+			}
+			// do nothing if they're equal.
+			if old_deposit > new_deposit {
+				let _ = T::Currency::unreserve(&sender, old_deposit - new_deposit);
+			}
+
+			if subs.is_empty() {
+				<SubsOf<T>>::remove(&sender);
```
tomusdrw: What happens with the `deposit`? Should we error in case it's not unreserved (i.e. `new_deposit.is_zero()`)?

frame/identity/src/lib.rs
gavofyork: No - `BitFlags` doesn't impl it.

frame/identity/src/lib.rs
gavofyork: didn't compile.

frame/identity/src/lib.rs
gavofyork: zero subs, zero remaining deposit.

gavofyork:
> Looks good.
> 
> A small concern that it seems that there is a chance to get a judgment without paying a fee, by racing cancelling the request before a judgment is processed.

Well done! I was wondering if anyone would notice that. Yeah, if it turns out to be an issue (which I don't think it will since you'd pretty much need to be a validator to successfully beat the other tx in a race), then we can introduce an extra function `provide_paid_judgement` which only works if the fee is there.

frame/identity/src/lib.rs
```diff
@@ -0,0 +1,1040 @@
...
+
+/// Information concerning the identity of the controller of an account.
+///
+/// NOTE: This should be stored at the end of the storage item to facilitate the addition of extra
+/// fields in a backwards compatible way through a specialized `Decode` impl.
+#[derive(Clone, Encode, Decode, Eq, PartialEq, RuntimeDebug)]
+#[cfg_attr(test, derive(Default))]
+pub struct IdentityInfo {
+	/// Additional fields of the identity that are not catered for with the structs explicit
+	/// fields. Stored ordered.
```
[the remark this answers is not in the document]
gavofyork: Yeah, turns out this isn't important so will remove it.

frame/identity/src/lib.rs
```diff
@@ -0,0 +1,1040 @@
...
+	pub info: IdentityInfo,
+}
+
+impl <
+	Balance: Encode + Decode + Copy + Clone + Debug + Eq + PartialEq + Zero + Add,
+> Registration<Balance> {
+	fn total_deposit(&self) -> Balance {
+		self.deposit + self.judgements.iter()
+			.map(|(_, ref j)| if let Judgement::FeePaid(fee) = j { *fee } else { Zero::zero() })
+			.fold(Zero::zero(), |a, i| a + i)
```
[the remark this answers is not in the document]
gavofyork: Sadly not - `Balance` doesn't impl `iter::Sum` or the things that it depends on.

xlc:
> > Looks good.
> > A small concern that it seems that there is a chance to get a judgment without paying a fee, by racing cancelling the request before a judgment is processed.
> 
> Well done! I was wondering if anyone would notice that. Yeah, if it turns out to be an issue (which I don't think it will since you'd pretty much need to be a validator to successfully beat the other tx in a race), then we can introduce an extra function `provide_paid_judgement` which only works if the fee is there.

Can't you add some tips to increase the priority of a transaction and therefore make it very likely the cancel tx get executed before the judgment tx? With a long-ish block time and short tx propagation time, the chance of success front-running can be pretty high.

PR #4327, opened 2019-12-09
Rename: primitives/sr-std -> primitives/sp-std

kigawas:


bkchr:
I think we should stop renaming all imports directly and just use `sp-std`.

kigawas:
> I think we should stop renaming all imports directly and just use `sp-std`.

You mean rename the `use rstd` stuff?

tomusdrw:
@bkchr @kigawas 
I would remove the crate renames (`X = { package = "sp-X", ...}`) separately to simplify reviewing. It can also be applied as a script to the entire codebase easily, I think @gnunicorn had some ideas how to do it.

@kigawas `sp-` prefix should be dropped from the folder names, so we should rename to `primitives/std` and the crate name should be `sp-std` (it's `substrate-primitives-std`).

frame/balances/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2018"
 serde = { version = "1.0.101", optional = true }
 safe-mix = { version = "1.0.0", default-features = false }
 codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false, features = ["derive"] }
-rstd = { package = "sp-std", path = "../../primitives/sr-std", default-features = false }
+sp-std = { package = "sp-std", path = "../../primitives/std", default-features = false }
```
gnunicorn:
```suggestion
sp-std = { path = "../../primitives/std", default-features = false }
```

would recommend we remove the `package = ...` everywhere as well.

primitives/sr-arithmetic/fuzzer/Cargo.lock
```diff
@@ -308,7 +308,7 @@ dependencies = [
 ]
 
 [[package]]
-name = "sr-std"
+name = "std"
```
gnunicorn: have you tested whether the fuzzer still works?

gnunicorn:
@bkchr , @tomusdrw moving the discussion to the meta issue.

primitives/sr-arithmetic/fuzzer/Cargo.lock
kigawas: Looks like this is not in CI

kigawas: Already fixed the cargo.lock file, but the ci code still needs to add

kigawas:
full crypto failure will be fixed by #4339

PR #4329, opened 2019-12-09
Remove slog-async Github

gnunicorn:
By putting that record we are using directly into the crates–internal use only. Leaving a FIXME comment on the top to remove this once the PR has cleared.

Because this is completely internally an update removing this file and replacing the dependency later would only require a patch-level update.

### DIFF
--- a/client/telemetry/src/lib.rs
+++ b/client/telemetry/src/lib.rs
@@ -69,6 +69,7 @@ pub use slog_scope::with_logger;
 pub use slog;
 
 mod worker;
+mod async_record;
 
 /// Configuration for telemetry.
 pub struct TelemetryConfig {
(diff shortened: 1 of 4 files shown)


client/telemetry/src/lib.rs
```diff
@@ -69,6 +69,7 @@ pub use slog_scope::with_logger;
 pub use slog;
 
 mod worker;
+mod async_record;
```
tomaka: That's very OCD and nit-picky, but what about putting this *above* `mod worker;` in order to have alphabetical order?

tomaka: Thank you :heart:

gnunicorn: if we had rustfmt my editor would have done that automatically ;)

PR #4330, opened 2019-12-09
Don't assume a fixed length for `endowed_accounts`

bkchr:


thiolliere:
maybe it should be mentionned in the doc that the first half of the endowments account are choosen as members

bkchr:
It's not mentioned that we replace `endowed_accounts` with a default value either. I'm not sure and it's also more file local and nothing that should be used from outside of this file IMHO.

PR #4332, opened 2019-12-09
Fix warnings when compiling runtime.

tomusdrw:
Fixes warnings that are displayed when compiling runtime (usually hidden, because of wasm-builder does not output anything if it's successful).

Also removes `dispatch::Result` imports, as they are super confusing - especially the ones in `node-template` and in examples.


primitives/sr-io/src/lib.rs
```diff
@@ -770,8 +771,10 @@ pub fn panic(info: &core::panic::PanicInfo) -> ! {
 	}
 }
 
+/// A default OOM handler for WASM environment.
 #[cfg(all(not(feature = "disable_oom"), not(feature = "std")))]
 #[alloc_error_handler]
+#[allow(improper_ctypes)] // core::alloc::Layout is not FFI-safe.
```
pepyakin: Hm? 

1. I thought that FFI safeness has to do with `extern "C"`, since `extern` should assume Rust ABI.
2. We don't need `extern` here in the first place, I think

tomusdrw: I have no idea, just repeating what the warning says :)

JoshOrndorff:
Not a full review, but I can confirm this solves my `error: cannot find macro vec in this scope` error #4333 .

primitives/sr-io/src/lib.rs
bkchr: Given this: https://github.com/rust-lang/rust/issues/51540

The signature should be changed to `fn oom(_: core::alloc::Layout) -> !`

primitives/core/src/crypto.rs
```diff
@@ -18,7 +18,9 @@
 //! Cryptographic utilities.
 // end::description[]
 
-use rstd::{vec::Vec, hash::Hash};
+#[cfg(feature = "std")]
+use rstd::vec::Vec;
```
bkchr: `Vec` is in `std::prelude`. You probably want to use the `full-crypto` feature here.

bkchr: You really need to use the full-crypto feature here.

PR #4345, opened 2019-12-10
Allow referendums to begin out of order

gavofyork:
Right now, referendums can only begin if they end after the latest-ended existing referendum. This basically means that fast-tracking cannot work unless there is not an existing referendum happening.

This PR relaxes that constraint.

### DIFF
--- a/frame/democracy/src/lib.rs
+++ b/frame/democracy/src/lib.rs
@@ -537,11 +537,11 @@ decl_module! {
 			);
 			ensure!(proposal_hash == e_proposal_hash, "invalid hash");
 
+			<NextExternal<T>>::kill();
 			let now = <system::Module<T>>::block_number();
 			// We don't consider it an error if `vote_period` is too low, like `emergency_propose`.
 			let period = voting_period.max(T::EmergencyVotingPeriod::get());
-			Self::inject_referendum(now + period, proposal_hash, threshold, delay)?;
-			<NextExternal<T>>::kill();
+			Self::inject_referendum(now + period, proposal_hash, threshold, delay);
 		}
 
 		/// Veto and blacklist the external proposal hash.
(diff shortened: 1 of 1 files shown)


frame/democracy/src/lib.rs
```diff
@@ -537,11 +537,11 @@ decl_module! {
 			);
 			ensure!(proposal_hash == e_proposal_hash, "invalid hash");
 
-			<NextExternal<T>>::kill();
```
gavofyork: We shouldn't really be modifying storage as long as we might return an `Err`.

gavofyork:
Test added.

frame/democracy/src/lib.rs
```diff
@@ -900,26 +900,26 @@ impl<T: Trait> Module<T> {
...
 	}
 
 	/// Remove all info on a referendum.
 	fn clear_referendum(ref_index: ReferendumIndex) {
 		<ReferendumInfoOf<T>>::remove(ref_index);
+
+		NextTally::mutate(|i| if *i == ref_index {
+			*i += 1;
+			let end = ReferendumCount::get();
+			while !Self::is_active_referendum(*i) && *i < end {
```
bkchr: What if there is a referendum with a lower index that is still active?

gavofyork:
> After looking again into the code, I'm not sure if we should change `NextTally` to `LowestTally` or whatever, but that it points to the lowest still active index.
> 
> You also need to adapt the functions `active_referenda` and `maturing_referenda_at` as they still use `take_while`. (These functions also would not find any referendum that is still active and has a lower index as the one returned by `next_tally()`.)

Please check my changes (made before your comment, but possible after your review). There is no `take_while` in that file.

gavofyork:
You're right about the renaming - just fixed.

bkchr:
> > After looking again into the code, I'm not sure if we should change `NextTally` to `LowestTally` or whatever, but that it points to the lowest still active index.
> > You also need to adapt the functions `active_referenda` and `maturing_referenda_at` as they still use `take_while`. (These functions also would not find any referendum that is still active and has a lower index as the one returned by `next_tally()`.)
> 
> Please check my changes (made before your comment, but possibly after your review). There is no `take_while` in that file.

For some reason I have overlooked this. With the rename it is not much better/clearer.

Issue #4331, opened 2019-12-09
Validator with apparently zero backing chosen


gavofyork:
At some point, a validator with zero exposure was found to be offline:

https://polkascan.io/pre/kusama-cc3/event/146198-0

If shouldn't be possible to be selected as a validator with zero exposure. Either there's something wrong with Phragmen or there's something wrong with the exposure fetching as part of im_online.

kianenigma:
could very well be phragmen. I had it on my todo list to add a 'minimum entry barrier' parameter to it, since at the moment, it allows all candidates to be considered and potentially elected.

kianenigma:
Meaning: Phragmen indeed allows one with zero stake to be elected. Yet, I am sure that staking should never allow someone to bond with a value [less than ED](https://github.com/paritytech/substrate/blob/f588aa53af54029c6b45f26342204482f2e946c5/frame/staking/src/lib.rs#L841), so it is even more unclear to me where the issue is. Simple patch is to add the aforementioned `minimum_entry_barrier` parameter to phragmen but I am not sure if this the concern of phragmen. Staking should never allow a validator to have zero stake. They must bond more than ED and unbonding in a way that leads to dust [is also handled](https://github.com/paritytech/substrate/blob/f588aa53af54029c6b45f26342204482f2e946c5/frame/staking/src/lib.rs#L924).

burdges:
We want some minimum self stake I thought, which presumably gives the minimum_entry_barrier, no?  If we supported a separate minimum_exit/ellect_barrier then this requires tweaks to phragmen.

thiolliere:
Is it nearby the last session of an era ? if so I feel it could be related to https://github.com/paritytech/substrate/issues/4121

let's say there is new era at  session 5
at end of session 4 session pallet calls OnSessionEnding::on_session_ending(4, 6),
staking module actually executes this code
https://github.com/paritytech/substrate/blob/c6fc2e6240edeb31233b07748c6bf968c7c15553/frame/staking/src/lib.rs#L1610-L1617

So new_session(5) is called and returns the new validator set.
session pallet receive the new validator set but this set in its queue because he was asking for the validator set of session 6.

Then at the end of the session 5 the value in storage `staking::Stakers` are the new validator set but the value in `session::Validators` is still the old validator set.
So if the offence ask for a validator which is not in next one its exposure is 0.

proposed fix is https://github.com/paritytech/substrate/pull/4346

thiolliere:
I confirm:
[...]

Issue #4220, opened 2019-11-27
`ValidatorPrefs` Updates Should Apply at Next Era


joepetrowski:
If a current validator issues `staking::validate()` with new `ValidatorPrefs`, the effects are felt immediately. That is, the validator can change their commission mid-era and take the rewards from their nominators.

Related: https://github.com/paritytech/substrate/issues/4121

thiolliere:
if we fix this and allow the validator to change the preference only for future era is that enough ?
I can imagine a validator changing the preference just before a new era is planned, thus no nominators will have the time to change their nomination.

joepetrowski:
Validators will probably be more advanced and diligent than nominators, so they definitely could pull this off. That said, hopefully their nominators would only suffer one era of lost rewards and would then choose a different validator. Perhaps something for the research team to chime in on whether changes should apply in era `n + 1` or `n + 2`.

CC @burdges @AlistairStewart

burdges:
As you like, another option is era n+1 but only if the transaction appears before the final epoch of era n, after that it's n+2.

thiolliere:
Due to https://github.com/paritytech/substrate/pull/4474 ValidatorPrefs indeed apply at next era. though there is no delay for nominators to know this preference ahead of time (a validator can change its preference just before the new era)

Issue #4121, opened 2019-11-18
In staking rewards go to next validator set


thiolliere:
## Issue

`session.validators` contains the current set of validator validating and producing blocks,
`session.queued_keys` contains the next set.

in on_session_ending in case of a new era, the staking module elect a new set, put this set in `staking::CurrentElected` and returns it, session module put this new elected set in the queued keys.

But in `staking::reward_by_ids` we add reward points for a validator in `stakig::CurrentElected`.
And at the end of the era we reward according to this points the validators contained in `staking::CurrentElected`. This seems wrong as `staking::CurrentElected` at this moment contains the validator set for the next era not the current one.

## How to fix

same as session, staking should have information for next_session but also for current_session. Maybe storing the staking information for previous elected set so it can reward them correctly.

CC @jacogr

kianenigma:
I was also asked about this and it a bug, forgot to issue it actually and thanks for doing that. I think staking should also store 2 eras to always know the current live set. This will also be needed for offchian phragmen, as discussed internally.

thiolliere:
I think all of this boils down to just implementing  OnSessionEnding correctly:

currently it is:

https://github.com/paritytech/substrate/blob/524413a55c026764bc2a70378491ce5752c5b951/paint/staking/src/lib.rs#L1456-L1462

with OnSessionEnding being defined as:

https://github.com/paritytech/substrate/blob/524413a55c026764bc2a70378491ce5752c5b951/paint/session/src/lib.rs#L165-L184

I think staking should store the stake information for the session will_apply_at and use the information stored at session_ending to reward the current era. (and prune this information on session ending)

thiolliere:
I got a wip branch https://github.com/paritytech/substrate/compare/gui-staking-fix-wrong-reward-attribution?expand=1

But it needs to update test, also maybe there is some issue inside, and moreover we need to transition from current logic to this logic and this might need some additional conditions.

kianenigma:
https://github.com/paritytech/substrate/compare/gui-staking-fix-wrong-reward-attribution?expand=1#diff-cdf560c4db60ec139dae0470ea9988beR601

I would store just two storage items with known keys and hashes, instead of a map, since we are 100% sure we just want 1 additional entry to store the next + current validators.

thiolliere:
I think the current trait OnSessionEnding doesn't guarantee that there is only one era ahead.

[...]

PR #4346, opened 2019-12-10
Fix staking reward issue

thiolliere:
Fix https://github.com/paritytech/substrate/issues/4121
Fix https://github.com/paritytech/substrate/issues/4220
Fix https://github.com/paritytech/substrate/issues/4331

### TODO

* [ ] maybe better naming
* [ ] maybe improvement on storage structure
* [x] Kusama transition
* [ ] Test storage upgrade transition with some real chain put_code upgrade

### DONE

when implementing `OnSessionEnding::on_session_ending`, we must end era if the session ending is the last one of the era, this era end will reward validator and clear some storage. but we also potentially plan a new era in the future using the argument `will_apply_at`.
This new era planned must freeze the validators information for this future era: validator set, validators exposures, commissions and slot_stake.

This new era in planned with the storage `EraStartSessionIndex` which store the session index on which the era start from the current era to the futurest planned.
Era information is stored in `ValidatorForEra` and `SlotStakeForEra`.


frame/staking/src/mock.rs
```diff
@@ -443,8 +447,6 @@ pub fn advance_session() {
 }
 
 pub fn start_session(session_index: SessionIndex) {
-	// Compensate for session delay
-	let session_index = session_index + 1;
```
thiolliere: this is one reason why tests with start_session has been modified, it makes them clearer IMHO, also most tests don't use this but start_era instead.

frame/staking/src/mock.rs
```diff
@@ -409,11 +413,11 @@ pub fn check_nominator_exposure(stash: u64) {
 	);
 }
 
-pub fn assert_is_stash(acc: u64) {
+pub fn assert_is_stash(acc: AccountId) {
```
gavofyork: pass as `&`

gavofyork:
The staking/reward/exposure dataset could be huge and complicated. Have you analysed the various complexity here?

gavofyork:
I would prefer this be done in the way I outline in #4407

PR #4347, opened 2019-12-10
Break appcrypto dep cycle

gnunicorn:
Moving app crypto integration tests into dedicated sub-crate to fix dependency cycle, refs #4338 .

Also:
 - [x] fixes the crate name `sp-application-crypto` (rather than `sc-application-crypto`)
 - [x] update `.maintain/rename`-script
 - [x] start `docs/Upgrade.md` and record the steps to fix up.


primitives/application-crypto/test/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "sp-application-crypto-test"
+version = "2.0.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+description = "Integration tests for application-crypto"
+publish = false
+
+[dependencies]
+primitives = { package = "sp-core", path = "../../core", default-features = false }
+# codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false, features = ["derive"] }
```
bkchr: These 4 lines should be removed.

primitives/application-crypto/test/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+
```
bkchr: Missing license header

primitives/application-crypto/test/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+use sp_runtime::{generic::BlockId, traits::ProvideRuntimeApi};
+use primitives::{testing::{KeyStore, ED25519}, crypto::Pair};
+use test_client::{
+    TestClientBuilder, DefaultTestClientBuilderExt, TestClientBuilderExt,
+    runtime::{TestAPI, app_crypto::ed25519::{AppPair, AppPublic}},
+};
+
+#[test]
+fn ed25519_works_in_runtime() {
```
bkchr: Where is the other test?

primitives/application-crypto/test/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Integration tests for application crypto
+
+
+#[cfg(test)]
+mod ed25519 {
```
bkchr: Common ben, both of these modules can be removed and the tests can live in the crate.

bkchr: :P

gnunicorn: I went for minimal changes via copy-paste. Don't have much of an opinion other than that feels like appropriate compartmentalizing inline with what the original crate did. Consider that nothing here will ever be imported, either is fine with me. If you insist I will flatten it.

primitives/application-crypto/test/src/lib.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
```
tomusdrw: Instead of creating `src/lib.rs` you could just create `tests` folder and put every module in it's own file (`ed25519.rs`, `sr25519.rs`) and make this crate purely a test crate.

primitives/application-crypto/test/src/lib.rs
bkchr: Either flatten or what @tomusdrw says.

bkchr:
Needs resolving.

PR #4349, opened 2019-12-10
Including runners-up in check for defunct phragmen voter.

jnaviask:
The `is_defunct_voter` function is intended to check whether all of a given voter's selections are invalid. A selection is valid if it is a candidate in the next round, i.e. if it is a current member or runner-up (and thus automatically added as candidates), or on the candidates list.

However, the function as implemented did not check for runners-up. This PR adds this check to fix the bug, which would have permitted incorrect reports of defunct voters if they only voted for current runners-up. This PR also expands the `can_detect_defunct_voter` test case to catch the bug.

### DIFF
--- a/frame/elections-phragmen/src/lib.rs
+++ b/frame/elections-phragmen/src/lib.rs
@@ -536,7 +536,7 @@ impl<T: Trait> Module<T> {
 		if Self::is_voter(who) {
 			Self::votes_of(who)
 				.iter()
-				.all(|v| !Self::is_member(v) && !Self::is_candidate(v).is_ok())
+				.all(|v| !Self::is_member(v) && !Self::is_runner(v) && !Self::is_candidate(v).is_ok())
 		} else {
 			false
 		}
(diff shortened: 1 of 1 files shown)


frame/elections-phragmen/src/lib.rs
```diff
@@ -536,7 +536,7 @@ impl<T: Trait> Module<T> {
 		if Self::is_voter(who) {
 			Self::votes_of(who)
 				.iter()
-				.all(|v| !Self::is_member(v) && !Self::is_candidate(v).is_ok())
+				.all(|v| !Self::is_member(v) && !Self::is_runner(v) && !Self::is_candidate(v).is_ok())
```
bkchr: Could you add a test that failed without your changes?

jnaviask: Expanded a test case `can_detect_defunct_voter` to detect this bug.

gavofyork:
CC @kianenigma

kianenigma:
Thanks @jnaviask! +1

PR #4352, opened 2019-12-10
break sp-api dependency cycle

gnunicorn:
refs #4338 .

- [x] fix folder name
- [x] move `sp-test-primitives` into `/primitives` for easier reuse in doctest
- [x] move benches into `sp-api-test` to break the cycle
- [x] fix broken doctests (//cc @bkchr any better idea than just ignoring them?)


primitives/api/src/lib.rs
```diff
@@ -154,41 +154,11 @@ pub use sp_api_proc_macro::decl_runtime_apis;
 ///
 /// # Example
 ///
-/// ```rust
+/// ```ignore
```
bkchr: That is something I'm totally against and will not approve. Breaking tests for releasing is no good idea. The cyclic dependency you are mentioning is only a `dev-dependency`. We should just comment them out when we do a release.

You could also try to implement `Block` and `Header` for some fake types.

PR #4353, opened 2019-12-10
Fix cycle dependency in `sp-runtime-interface`

bkchr:
Tracking issue: https://github.com/paritytech/substrate/issues/4338


gnunicorn:
@bkchr breaks tests.

PR #4354, opened 2019-12-10
Fix a bunch of low work dependency cycles

gnunicorn:
tracking at #4338 . this

- [x] moves tests out of `sp-runtime-interface` to break cycle via `test-client`
- [x] replaces `test_client::runtime::Block` with `sp_test_primitive::Block` where possible. 
- [x] break the cycle in `sc-network` by moving tests outside


bkchr:
>     * [x]  moves tests out of `sp-runtime-interface` to break cycle with `test-client`

It does not use `test-client`?

gnunicorn:
@bkchr it does. and test-client uses sp-runtime-interface ... hence the cycle...

bkchr:
> @bkchr it does. and test-client uses sp-runtime-interface ... hence the cycle...

Where? I just use `substrate-executor` to call directly into the wasm blob. Maybe some other dependency builds this cycle?

gnunicorn:
on master cyclist reports it as:
> - [ ] `sp-runtime-interface` (`2.0.0`): Cycle through `sp-runtime-interface`(`2.0.0`) -> `sc-executor`(`2.0.0`) -> `sc-client-api`(`2.0.0`) -> `sc-client-db`(`2.0.0`) -> `sc-client`(`2.0.0`) -> `sc-block-builder`(`2.0.0`) -> `sp-api`(`2.0.0`) -> `sp-core`(`2.0.0`) -> `sp-runtime-interface`(`2.0.0`)

On this branch, this is resolved.  I think there are actually multiple in-between, like I also broke up unecessary dependencies in executor and api after...

PR #4355, opened 2019-12-11
Fix token economics url.

marcio-diaz:


frame/staking/src/lib.rs
```diff
@@ -173,7 +173,7 @@
 //! calculated using the era duration and the staking rate (the total amount of tokens staked by
 //! nominators and validators, divided by the total token supply). It aims to incentivise toward a
 //! defined staking rate. The full specification can be found
-//! [here](https://research.web3.foundation/en/latest/polkadot/Token%20Economics/#inflation-model).
+//! [here](https://research.web3.foundation/en/latest/polkadot/Token%20Economics.html).
```
bkchr:
```suggestion
//! [here](https://research.web3.foundation/en/latest/polkadot/Token%20Economics.html#inflation-model).
```

If you fix stuff, please make it properly.

marcio-diaz: Yeah, sorry, I didn't find it, and tried the url but didn't work for some reason.

PR #4356, opened 2019-12-11
Improve is_member and is_runner.

marcio-diaz:
Even if they are bounded, I prefer to fix it. From `O(NlogN)` to `O(logN)`.
Removes the use of `members_id` and `runners_up_id` that are linear.


marcio-diaz:
> This is a minor optimisation; it does not change the complexity.
> 
> `Self::members` is `O(N)` since it must construct and decode `N` items (all `Vec` storage items are `O(N)` to read or write).
> 
> The original complexity was `O(N + logN)`, not `O(NlogN)` as you asserted. The new complexity after your changes is also `O(N + logN)`. Nothing changes except that we avoid building a second `Vec`.

Yes, I'm still sleepy ^^

Issue #4363, opened 2019-12-11
Tip division should be configurable


joepetrowski:
Block producers should get 100% of the tip, but [currently](https://github.com/paritytech/substrate/blob/09917a50941922c3eb226acf93461b1ac5bb9fc9/frame/transaction-payment/src/lib.rs#L175) the tip gets included with the other pre-dispatch fees (base fee, byte fee, weight fee).

> The transaction fee is considered a base price. There will be a different field in the transaction called tip, and a user is free to put any amount of tokens in it or leave it at zero. Block producers receive 100% of the tip on top of the standard 20% of the fee, so they have an incentive to include transactions with large tips.

Source: https://research.web3.foundation/en/latest/polkadot/Token%20Economics.html#-2.-slow-adjusting-mechanism

**Note:** The W3F Research page may be out of date. Verify with research team that they still want this behavior.

Ideally, the tip division is configurable in the runtime. For example, one could decide that for normal fees, 20% go to the author and 80% to the Treasury, while for tips the split is 50/50.

CC @thiolliere

thiolliere:
if somebody wants to do it before the freeze I don't have enough time to get it, one can take a look at my PR, or do another way. I'll unassigned for now.

(note that it might look that I unassigned too late but the PR was there for a long time and I learned about the freeze just a few days ago back from days off).

PR #4364, opened 2019-12-11
Impl tip and fee payment distinctly

thiolliere:
Fix https://github.com/paritytech/substrate/issues/4363

This would allow polkadot to give all tip to author and fee partially to author and partially to treasury.

Two differences:
* if sender has enough funds to pay tip but not to pay fees and somehow the transaction get included then it will withdraw the tip payment and then fail the extrinsic. instead of just failing the extrinsic.
  I feel fine with this logic but otherwise we can just do ensure_can_withdraw before.
* the tip withdraw has only one reason `WithdrawReason::Tip` instead of `WithdrawReason::Tip|WithdrawReason::Payment` previously, should we still withdraw as both tip and payment reason ?

related polkadot update: https://github.com/paritytech/polkadot/pull/684

### DONE:

* I introduce Author trait in support, which returns the current author.
* priority is now `author_balance_diff/total_issuance/(transaction_weight/maximum_block_weight) * billion`
* pallet-aura has been modified to implement author traits, this needs a migration code. But are we doing migration code for such modules ?

### CRITIC:

The author balance is used to compute the priority but in-between block the traits return the default account. Maybe other validate function transfer to this default account (because the actual account they will transfer to will be set in on_initialise for example). Thus the priority will be wrong.


frame/transaction-payment/src/lib.rs
```diff
@@ -207,28 +207,36 @@ impl<T: Trait + Send + Sync> SignedExtension for ChargeTransactionPayment<T>
...
-		T::OnTransactionPayment::on_unbalanced(imbalance);
+
+		T::OnTransactionTipPayment::on_unbalanced(tip_imbalance);
+		T::OnTransactionFeePayment::on_unbalanced(fee_imbalance);
 
 		let mut r = ValidTransaction::default();
 		// NOTE: we probably want to maximize the _fee (of any type) per weight unit_ here, which
 		// will be a bit more than setting the priority to tip. For now, this is enough.
-		r.priority = fee.saturated_into::<TransactionPriority>();
+		r.priority = fee.saturating_add(tip).saturated_into::<TransactionPriority>();
```
thiolliere: This priority is now biased as the tip+fee is not actually proportional to the amount received by validator

thiolliere:
priority is fixed but I'm not very happy by how error-prone is the code now. I update main description

gnunicorn:
tests failing.

kianenigma:
I am slightly concerned about the priority computation; what is the reasoning behind it? it it absolutely necessary, given that we want to do as little as possible in `validate` phase?

frame/transaction-payment/src/lib.rs
```diff
@@ -214,32 +223,61 @@ impl<T: Trait + Send + Sync> SignedExtension for ChargeTransactionPayment<T>
...
-			T::OnTransactionPayment::on_unbalanced(imbalance);
+			T::OnTransactionFeePayment::on_unbalanced(imbalance);
 		}
 
-		let mut r = ValidTransaction::default();
-		// NOTE: we probably want to maximize the _fee (of any type) per weight unit_ here, which
-		// will be a bit more than setting the priority to tip. For now, this is enough.
-		r.priority = fee.saturated_into::<TransactionPriority>();
-		Ok(r)
+		let author_balance_diff = T::Currency::total_balance(&author)
```
kianenigma: This section could use a bit of documenting IMO; what is the author doing here? I imagine you are checking if any of the tip/fee was added to the author's balance in the two `if !fee.is_zero() {}` and `if !tip.is_zero() {}` blocks. Can't this be predicted, so that we can prevent another storage access? same goes with accessing `T::Currency::total_issuance()`. Note that this is called per transaction. Anything that we can avoid is a big gain.

kianenigma:
> tests failing.

seems unrelated to the PR, though I'd still like to hear about the concern above as well.

frame/transaction-payment/src/lib.rs
thiolliere: Indeed it seems that checking author balance is too costy.
I wanted to keep OnTransactionFeePayment generic, but maybe we should require this associated type to return how much is given to author.
And I'll try to remove total_issuance call.

frame/transaction-payment/src/lib.rs
```diff
@@ -72,6 +78,12 @@ pub trait Trait: frame_system::Trait {
 
 	/// Update the multiplier of the next block, based on the previous block's weight.
 	type FeeMultiplierUpdate: Convert<Multiplier, Multiplier>;
+
+	/// Author rewarded for transaction inclusion.
+	///
+	/// Its reward for including transaction is used to compute transaction priority. Thus the
+	/// value returned in-between blocks must be sensible in regards of Author reward.
+	type Author: Author<Self::AccountId>;
```
thiolliere: also this associated type can be error prone as usually in-between block default account is returned, and maybe in-between blocks default account can receive some funds for some other reason making the computation of priority wrong.

frame/support/src/traits.rs
```diff
@@ -216,6 +216,19 @@ impl<A> FindAuthor<A> for () {
...
 }
 
+/// A trait to get the author of the block.
+pub trait Author<AccountId> {
+	/// Return the account of the author of the block.
+	fn author() -> AccountId;
+}
+
+/// Return the default account id as author.
+impl<A: Default> Author<A> for () {
```
bkchr: I think we should remove this implementation for `()`. 

I think people could use `()` in their runtime and maybe fuck something up.

We could provide a type `TestAuthor` or whatever to make testing easier, but to remove the footgun.

frame/transaction-payment/src/lib.rs
```diff
@@ -178,9 +187,9 @@ impl<T: Trait + Send + Sync> ChargeTransactionPayment<T> {
 			let adjusted_fee = targeted_fee_adjustment.saturated_multiply_accumulate(adjustable_fee);
 
 			let base_fee = T::TransactionBaseFee::get();
-			base_fee.saturating_add(adjusted_fee).saturating_add(tip)
+			base_fee.saturating_add(adjusted_fee)
 		} else {
-			tip
+			0u32.into()
```
bkchr:
```suggestion
			Zero::zero()
```

frame/transaction-payment/src/lib.rs
```diff
@@ -214,32 +223,61 @@ impl<T: Trait + Send + Sync> SignedExtension for ChargeTransactionPayment<T>
...
 		let tip = self.0;
-		let fee = Self::compute_fee(len as u32, info, tip);
+		// Only mess with balances if tip is not zero.
+		if !tip.is_zero() {
+			let imbalance = match T::Currency::withdraw(
+				who,
+				tip,
+				WithdrawReason::Tip.into(),
+				ExistenceRequirement::KeepAlive,
+			) {
```
bkchr: Can we not just use
```suggestion
			).map_err(|_|  InvalidTransaction::Payment)?;
```

frame/transaction-payment/src/lib.rs
```diff
@@ -214,32 +223,61 @@ impl<T: Trait + Send + Sync> SignedExtension for ChargeTransactionPayment<T>
...
 				who,
 				fee,
-				if tip.is_zero() {
-					WithdrawReason::TransactionPayment.into()
-				} else {
-					WithdrawReason::TransactionPayment | WithdrawReason::Tip
-				},
+				WithdrawReason::TransactionPayment.into(),
 				ExistenceRequirement::KeepAlive,
 			) {
```
bkchr: Same `map_err`

frame/support/src/traits.rs
thiolliere: Yes and I fooled myself, node-template is wrong in this PR currently

bin/node-template/runtime/src/lib.rs
```diff
@@ -226,7 +226,9 @@ parameter_types! {
 
 impl transaction_payment::Trait for Runtime {
 	type Currency = balances::Module<Runtime>;
-	type OnTransactionPayment = ();
+	type OnTransactionTipPayment = ();
+	type OnTransactionFeePayment = ();
+	type Author = ();
```
thiolliere: node template should make use of aura for getting the author

frame/transaction-payment/src/lib.rs
thiolliere: yes done

frame/support/src/traits.rs
thiolliere: Done

thiolliere: Done

frame/transaction-payment/src/lib.rs
thiolliere: So I improved readibility, additional comment about cost:
* author account balance shouldn't be too costy to get, as it is should be in cache as it has been modified anyway so it is would stay in cache.
* getting the author though requires iterating through digest which are empty in-between blocks, and during blocks pallet-authorships save it into a temporary storage.

frame/aura/src/lib.rs
```diff
@@ -161,11 +161,31 @@ impl<T: Trait> FindAuthor<u32> for Module<T> {
 	}
 }
 
+impl<T: Trait> frame_support::traits::Author<T::AccountId> for Module<T> {
+	/// Fetch the author of the block.
+	///
+	/// This is safe to invoke from `System::on_initialize` until end of the block.
+	/// Returns default account id in between blocks.
+	fn author() -> T::AccountId {
```
gavofyork: why not just pass it in as an inherent at the beginning of the block?

gavofyork:
I'm not at all keen on this direction; it seems over-complicating things. I'll open an alternative PR.

thiolliere:
close in favor of Gav direction

PR #4369, opened 2019-12-11
Make sudo use `decl_error!`

shawntabrizi:
This is a small PR to confirm that I am using `decl_error` correctly here.

It upgrades the Sudo pallet to use `decl_error!`.


frame/sudo/src/lib.rs
```diff
@@ -122,8 +124,8 @@ decl_module! {
 		#[weight = SimpleDispatchInfo::FreeOperational]
 		fn sudo(origin, proposal: Box<T::Proposal>) {
 			// This is a public call, so we ensure that the origin is some signed account.
-			let sender = ensure_signed(origin)?;
-			ensure!(sender == Self::key(), "only the current sudo key can sudo");
+			let sender = ensure_signed(origin).map_err(|_| "origin must be signed")?;
```
DarkEld3r: Maybe `ModuleDispatchError::as_str` should be used here? It gives a less readable error description, but at least the same string won't be used everywhere.

shawntabrizi:
Let's wait until a small update to `decl_error!` make this a little bit better to merge this PR @thiolliere

gavofyork:
@shawntabrizi please reference an issue or PR for what you're waiting.

shawntabrizi:
Updated my comment to include a link to: https://github.com/paritytech/substrate/pull/4392

Basically, we expect every module to use `ensure_signed`, `ensure_root`, or `ensure_none`.

For this one situation, I suggested we look to support `From` trait between `frame_system` and any other pallet.

PR #4370, opened 2019-12-11
Use decl_error in the treasury module

DarkEld3r:
A slightly different approach than in #4369: `ModuleDispatchError::as_str` is used to convert errors.


shawntabrizi:
Let's wait until a small update to `decl_error!` make this a little bit better to merge this PR @thiolliere

frame/treasury/src/lib.rs
```diff
@@ -165,8 +168,8 @@ decl_module! {
 		/// # </weight>
 		#[weight = SimpleDispatchInfo::FixedOperational(100_000)]
 		fn reject_proposal(origin, #[compact] proposal_id: ProposalIndex) {
-			T::RejectOrigin::ensure_origin(origin)?;
-			let proposal = <Proposals<T>>::take(proposal_id).ok_or("No proposal at that index")?;
+			T::RejectOrigin::ensure_origin(origin).map_err(|e| Into::<&str>::into(e))?;
```
gavofyork: should also be using `.as_str()`

frame/treasury/src/lib.rs
```diff
@@ -183,9 +186,9 @@ decl_module! {
 		/// # </weight>
 		#[weight = SimpleDispatchInfo::FixedOperational(100_000)]
 		fn approve_proposal(origin, #[compact] proposal_id: ProposalIndex) {
-			T::ApproveOrigin::ensure_origin(origin)?;
+			T::ApproveOrigin::ensure_origin(origin).map_err(|e| Into::<&str>::into(e))?;
```
gavofyork: same.

frame/treasury/src/lib.rs
DarkEld3r: `ensure_origin` returns the `InvalidOrigin` struct that doesn't have `as_str` method because it doesn't implement `ModuleDispatchError`. It only has `From<InvalidOrigin> for &'static str` implementation. Am I missing something?

gavofyork: Then we should probably alter it so there are proper errors for invalid origin errors. Using strings for these kinds of super-well-known global errors seems dumb.

gavofyork: (Can go in a separate PR though, I guess)

gavofyork: See https://github.com/paritytech/substrate/issues/4409

PR #4371, opened 2019-12-11
Revamp reaping rules

gavofyork:
- Change `DispatchQueue` into a simple `Vec` item, which is fine as the entries are small and few (protected by governance). This makes is fast, cheap and easily enumerable.
- Simplify some APIs due to the above change allowing us to search the queue.
- Require a proposal preimage to not be in the queue for it to be reaped.
- Allow the owner of a preimage to reap before the public at large.
- Add a few tests.


frame/democracy/src/lib.rs
```diff
@@ -579,21 +578,13 @@ decl_module! {
...
-			let mut items = <DispatchQueue<T>>::get(when);
-			if items.get(which).and_then(Option::as_ref).map_or(false, |x| x.1 == what) {
-				items[which] = None;
-				<DispatchQueue<T>>::insert(when, items);
-			} else {
-				Err("proposal not found")?
-			}
+			let mut items = <DispatchQueue<T>>::get();
+			let original_len = items.len();
+			items.retain(|i| i.2 != which);
```
shawntabrizi: Does it matter that this function will continue to check items even if a match is found?

Something ever so slightly improved would stop after the first match since `Index` is unique.

frame/democracy/src/lib.rs
```diff
@@ -709,39 +700,42 @@ decl_module! {
...
+			// we always for a wait of at least (voting + enactment) to minimise the chance of a
+			// last minute preimage pre-enactment snatch. if a vote begins with no preimage
+			// registered, then you can be sure that the owner cannot yank it prior to enactment.
+			// if it is already registered, then
+			let (voting, enactment) = (T::VotingPeriod::get(), T::EnactmentPeriod::get());
+			let additional = if who == old { Zero::zero() } else { enactment };
+			ensure!(now >= then + voting + additional, "too early");
+
+			let queue = <DispatchQueue<T>>::get();
+			ensure!(queue.iter().all(|item| &item.1 != &proposal_hash), "imminent");
```
```suggestion
			ensure!(!queue.iter().any(|item| &item.1 == &proposal_hash), "imminent");
```

frame/democracy/src/lib.rs
```diff
@@ -709,39 +700,42 @@ decl_module! {
...
-			}
+			let (_, old, deposit, then) = <Preimages<T>>::get(&proposal_hash).ok_or("not found")?;
+			let now = <system::Module<T>>::block_number();
+			// we always for a wait of at least (voting + enactment) to minimise the chance of a
+			// last minute preimage pre-enactment snatch. if a vote begins with no preimage
+			// registered, then you can be sure that the owner cannot yank it prior to enactment.
+			// if it is already registered, then
+			let (voting, enactment) = (T::VotingPeriod::get(), T::EnactmentPeriod::get());
+			let additional = if who == old { Zero::zero() } else { enactment };
+			ensure!(now >= then + voting + additional, "too early");
```
shawntabrizi: I don't quite follow here. If the user who submitted the preimage wants to reap it, they do not follow this rule:

> we always for a wait of at least (voting + enactment)

Seems like you want to check for: 

```
now >= then + voting + enactment + additional
```

And `then` should really come from the original proposal timestamp not the preimage timestamp, but it does not look like it is easy to get that.

frame/democracy/src/lib.rs
shawntabrizi: Please let me know if this is the level of nit you are interested in these reviews.

frame/democracy/src/lib.rs
gavofyork: there will only ever be one of them (the same RefIndex cannot be dispatched twice), and it's `O(N)` in both instances. i think the simpler expression is more readable.

frame/democracy/src/lib.rs
gavofyork: docs updated

PR #4374, opened 2019-12-12
Introduce rebond

marcio-diaz:
From some channel:

> Marcio: Does it makes sense to introduce a `rebond`? Something that removes an `UnlockChunk` and adds it to the ledger balance?
> Jam: I would personally appreciate that - currently the flow is you have to wait for the bond to unlock completely, withdraw it, and then bond again.


frame/staking/src/lib.rs
```diff
@@ -910,6 +910,34 @@ decl_module! {
...
+		/// Rebond a portion of the stash scheduled to be unlocked.
+		///
+		/// # <weight>
+		/// - Insignificant complexity, bounded by `MAX_UNLOCKING_CHUNKS`.
+		/// - Can't increase storage, only decrease it.
+		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedNormal(500_000)]
+		fn rebond(origin, #[compact] value: BalanceOf<T>) {
+			let controller = ensure_signed(origin)?;
+			let mut ledger = Self::ledger(&controller).ok_or(Error::<T>::NotController)?;
```
kianenigma: You could check `ledger.unlocking.len()` here and short circuit if there hasn't been any pending funds to unbond.

frame/staking/src/tests.rs
```diff
@@ -1196,6 +1196,154 @@ fn too_many_unbond_calls_should_not_work() {
...
+				Some(StakingLedger {
+					stash: 11,
+					total: 1000,
+					active: 1000,
+					unlocking: vec![],
+				})
+			);
+
+			// trigger next era.
+			Timestamp::set_timestamp(10);
```
kianenigma: isn't this called inside start_era?

frame/staking/src/lib.rs
```diff
@@ -910,6 +910,34 @@ decl_module! {
...
+
+			let mut unlocking_balance: BalanceOf<T> = Zero::zero();
+
+			while let Some(mut unlock_chunk) = ledger.unlocking.pop() {
+				unlocking_balance += unlock_chunk.value;
+				ledger.active += unlock_chunk.value;
+
+				if unlocking_balance > value {
+					unlock_chunk.value = unlocking_balance - value;
+					ledger.active -= unlocking_balance - value;
```
kianenigma: I am not sure if this is correct w.r.t. accountability. AFAIK your unlock chunks are kept to keep track of how much you were staked at a particular time. Hence, a chunk should be treated as an atomic unit and not changed internally. I would revert the last update and then break.

kianenigma: Though in that case 

```
bond(1000)
unbond(1), unbond(1), unbond(1), unbond(100)

// rebond(10) would do nothing
```

but if there is a security compromise then this is how it should be. Though I am not sure.

marcio-diaz: You can create unlock chunks with `unbond` of any smaller value, so they are not keeping track of stake at a particular time. Afaik, they are there to unlock funds at the right era. So we can re-bond safely because a new `unbond` will give you a higher or equal era.

Another invariant that should hold in the ledger is `total = active + sum chunk.value for all chunk in ledger.unlocking`. Which I think is still valid after these changes.

frame/staking/src/tests.rs
marcio-diaz: I copy pasted from other test, will see if I can remove it :).

frame/staking/src/lib.rs
kianenigma: cc @rphmeier

rphmeier: > AFAIK your unlock chunks are kept to keep track of how much you were staked at a particular time. Hence, a chunk should be treated as an atomic unit and not changed internally

We use historical exposures for that, actually. Altering the unlock chunks is fine in that regard, and indeed we do it when slashing (see `StakingLedger::slash`).

The code here looks OK. A couple nits: 
1. In the case that we rebond a value covered by an exact number of unlock chunks, we always do an extra iteration of the loop. A tighter/more idiomatic form of the algorithm is probably possible
2. From a code clarity perspective - the `rebond` functionality should be on `StakingLedger` as opposed to being scattered into the module code, which should just invoke it.

marcio-diaz: 1) I tried a version without the extra iteration, but it was adding another if branch. After comparing them side-by-side, I decide to keep this one. But I can change it, np.
2) Right, I guess I can do the same with `unbond`.

marcio-diaz:
I need to label it as `breaksapi`, and bump spec and impl, right?

PR #4376, opened 2019-12-12
[client cli] generic blocknumber

### LINKED ISSUE #2653: Make `BlockNumber` in CLI generic as well
For some use cases in the CLI we require the `BlockNumber`, currently we assume the block number fits into `u32`. This is not correct and thus, we should make the CLI generic over the `BlockNumber` as well.

https://github.com/paritytech/substrate/pull/2602

niklasad1:
Attempt to close #2653

It changes that `block number` is hardcoded to u32 instead it stores a `String` and parses it at runtime to appropriate type.


client/cli/src/params.rs
```diff
@@ -734,6 +734,41 @@ pub struct BuildSpecCmd {
...
+
+impl BlockNumber {
+	/// Wrapper on top of `std::str::parse<N>` but with `Error` as a `String`
+	///
+	/// See `https://doc.rust-lang.org/std/primitive.str.html#method.parse` for more elaborate
+	/// documentation.
+	pub fn parse<N>(&self) -> Result<N, String>
+	where
+		N: FromStr,
+		N::Err: std::fmt::Debug,
```
niklasad1: just for printing the error which can be useful

client/cli/src/params.rs
```diff
@@ -734,6 +734,41 @@ pub struct BuildSpecCmd {
...
+		}
+	}
+}
+
+impl BlockNumber {
+	/// Wrapper on top of `std::str::parse<N>` but with `Error` as a `String`
+	///
+	/// See `https://doc.rust-lang.org/std/primitive.str.html#method.parse` for more elaborate
+	/// documentation.
+	pub fn parse<N>(&self) -> Result<N, String>
```
niklasad1: nothing prevents  parsing `BlockNumber to a signed integer thus it relies on each type's `FromStr` implementation.

For the primitives types `core::num` it will be `Overflow error`

PR #4377, opened 2019-12-12
Recover light client integration tests

svyatonik:
I've been sure that BABE' light client is broken since #3583 , because it has been removed from tests there + I've seen no consensus cache usage in the code. This was until this morning, when I've noticed that the full client now also doesn't call runtime for anything else than configuration at genesis block. Since this (runtime calls) was the only reason that light client was requiring additional care (storing epoch data in the consensus cache), there's no other reasons why light client should not work.

So I'm restoring integration tests for LC in this PR.


client/service/test/src/lib.rs
```diff
@@ -312,11 +310,7 @@ pub fn connectivity<G, E, Fb, F, Lb, L>(
 	const NUM_LIGHT_NODES: usize = 5;
 
 	let expected_full_connections = NUM_FULL_NODES - 1 + NUM_LIGHT_NODES;
-	let expected_light_connections = if light_node_interconnectivity {
-		expected_full_connections
-	} else {
-		NUM_FULL_NODES
-	};
+	let expected_light_connections = NUM_FULL_NODES;
```
HCastano: Should this be `NUM_LIGHT_NODES`?

svyatonik: It is a number of connections at light nodes. Light nodes only connect to full nodes => `NUM_FULL_NODES`.

PR #4379, opened 2019-12-12
Use multiple threads in integration tests

svyatonik:
spin off of #4377 

It seems unneeded now - integration tests are pretty stable now. At my laptop tests were running for ~1:30 with single test thread and ~0:40 with multiple threads.


gavofyork:
Depends on #4377

bkchr:
This should be now mergable?

PR #4383, opened 2019-12-12
Update networking code to libp2p 0.14

tomaka:
Libp2p 0.14 hasn't been released yet because we're stumbling upon a weird bug in secio.
However, here's the Substrate pull request that updates Substrate to the new version that uses stable futures.

The PR is ready to be reviewed. The only thing that we should do before merging is updating the `Cargo.toml`s to point to the published version.


tomaka:
Since the review is probably tedious, I must mention that I intentionally haven't touched any of the logic.

Some minor exceptions:

- In the config, we now use `telemetry::ExtTransport` rather than `libp2p::ExtTransport`. This makes it possible for the telemetry to have a different version of libp2p than the one of the networking.
- Changed a few `loop { match foo { Poll::Pending => break, ... } }` into `while let Poll::Ready(_) = foo { ... }`.
- The telemetry now boxes its transport rather than explicitly writing the type down. I had to add some map/map_err to adjust for the update in libp2p-websocket making PINGs and PONGs explicit.

client/telemetry/src/worker.rs
```diff
@@ -92,31 +75,48 @@ impl TelemetryWorker {
...
 		// an external transport on desktop and the fallback is used all the time.
 		#[cfg(not(target_os = "unknown"))]
 		let transport = transport.or_transport({
-			let inner = libp2p::dns::DnsConfig::new(libp2p::tcp::TcpConfig::new());
+			let inner = libp2p::dns::DnsConfig::new(libp2p::tcp::TcpConfig::new())?;
 			libp2p::websocket::framed::WsConfig::new(inner)
+				.and_then(|connec, _| {
+					let connec = connec
+						.with(|item| {
+							let item = libp2p::websocket::framed::OutgoingData::Binary(item);
```
twittner: In a future PR we may want to change this in order to support sending PINGs to the remote to detect connection loss.

client/network/src/discovery.rs
```diff
@@ -469,18 +458,24 @@ mod tests {
 						out,
 						secio,
 						endpoint,
-						libp2p::core::upgrade::Version::V1
-					)
+						upgrade::Version::V1
+					).map_ok(|(id, stream)| ((stream, id)))
 				})
-				.and_then(move |(peer_id, stream), endpoint| {
+				.and_then(move |(stream, peer_id), endpoint| {
```
expenses: What's this about?

tomaka: Ah, I got a compilation error here and initially thought it was because the variables were in the wrong order.

expenses:
Looks good as far as I can tell!

gavofyork:
Depends on libp2p 0.14 being released.

tomaka:
This should now be ready!

mxinden:
@tomaka Moving forward here sounds good to me, or would you prefer to test `libp2p:0.14.0-alpha.1` on a test network first?

tomaka:
> would you prefer to test libp2p:0.14.0-alpha.1 on a test network first?

I wouldn't mind testing first, but we don't really have a test network at our disposal, and I don't think this PR is much more sensitive than some others that Substrate merges.

Issue #4384, opened 2019-12-12
Make `decl_error!` slightly more friendly for `frame_system`


shawntabrizi:
Error messages propagating from one pallet to another is not handled that well right now.

If you use `ensure_signed` for example, the best you can do is:

```
let sender = ensure_signed(origin).map_err(|e| e.as_str())?;
```

This will convert the `frame_system::Error` object into a string, and as a result you will lose all of the great UI experiences that these errors are meant to provide.

One solution is to re-implement and map to a custom error message.

For example:

```
let sender = ensure_signed(origin).map_err(|_| Error::RequireSignedOrigin)?;
```

But then the user would need to define extra error types for every single module, since they all rely on some combination of `ensure_signed` or `ensure_root` or `ensure_none`.

```
decl_error! {
    enum Error {
        MyCustomError,
        RequireSignedOrigin,
        RequireRootOrigin,
        RequireNoOrigin,
    }
}
```

A proposed short term fix is to have `decl_error!` automatically add the 3 error types from `frame_system` for every new pallet:

* RequireSignedOrigin
* RequireRootOrigin
* RequireNoOrigin

Then to simply implement `From` between `frame_system` and the custom pallet being created.

At that point, the errors and UI should work as expected.

Long term, the solution here may need to be more robust and allow for verbose error tracing across multiple modules.

@thiolliere @bkchr

thiolliere:
yes we can do that behind an attribute for system errors as they are very common. Though I agree we probably need an error mecanism which is more scalable, and can wrap around other modules error.

shawntabrizi:
Optional sounds good :)

xlc:
My original idea for decl_error was make it similar to Event, each module have its own Error type and there is a Runtime Error type created from `construct_runtime`. Currently only the first part is in Substrate and the (outdated) second part is in #2880.

With this, runtime module can define an Error type in Trait like this

https://github.com/paritytech/substrate/pull/2880/files#diff-c41ce90d658462c61a0814b7279026a6R239

`type Error: From<Error> + From<system::Error> + From<&'static str>;`

And then the error type can be just converted from system::Error into T::Error.

If this runtime module uses other runtime module, the definition can be changed to 

`type Error: From<Error> + From<system::Error> +  From<balances::Error> + From<&'static str>;`

Related #4040

xlc:
FYI decl_error already add `CannotLookup`

[...]

PR #4387, opened 2019-12-12
Use `decl_error!` in Staking Pallet

DarkEld3r:


DarkEld3r:
In progress because of #4384.

shawntabrizi:
@DarkEld3r we can put this in review, I think we will not create any changes in the near/near future.

frame/staking/src/lib.rs
```diff
@@ -783,6 +783,32 @@ decl_event!(
...
+		/// Not a controller account.
+		NotController,
+		/// Not a stash account.
+		NotStash,
+		/// Stash is already bonded.
+		AlreadyBonded,
+		/// Controller is already paired.
+		AlreadyPaired,
+		/// Should be the root origin or the `T::SlashCancelOrigin`.
+		BadOrigin,
```
shawntabrizi:
```suggestion
		InvalidOrigin,
```

Let's use the term "InvalidOrigin" rather than "BadOrigin" :)

frame/staking/src/lib.rs
```diff
@@ -1163,7 +1191,7 @@ decl_module! {
 			T::SlashCancelOrigin::try_origin(origin)
 				.map(|_| ())
 				.or_else(ensure_root)
-				.map_err(|_| "bad origin")?;
+				.map_err(|_| Error::BadOrigin)?;
```
shawntabrizi:
```suggestion
				.map_err(|_| Error::InvalidOrigin)?;
```

gavofyork: I'd prefer for these general bad origin errors we just had one global error that everything can reference and that `try_origin` can return.

gavofyork: https://github.com/paritytech/substrate/issues/4409 - these grumbles can be fixed with that.

PR #4389, opened 2019-12-13
Update Balances Pallet events to help block explorers

### LINKED ISSUE #4382: Additional event-attribute balances.ReapedAccount
This is a request to add an additional attribute to the `balances`.`ReapedAccount`-event.
Currently only the account identifier is included as (first) event attribute (example: https://polkascan.io/pre/kusama-cc3/event/9618-9).

I'd like to have an additional (second) attribute with a `Balance` type that explicitly shows the amount that is burnt on reaping the account.

The context of this change is Polkascan's accounting project which allows us (and others) to have a complete accounting of the native token economy without querying account balances. In this particular case there is no way to identify the exact amount that is lost. It is a value greater than 0 and less than the existential deposit. The second event-attribute solves this.

shawntabrizi:
Fixes #4382 

The goal of this PR is to help an outside service like a block explorer to track the total token economy of the Balances Pallet through events, and without needing to querying individual account balances.

To accomplish this, we:

* Updated `ReapedAccount` event to emit the dust burned when an account is reaped.
* Added `BalanceSet` event to emit when root directly changes the balance of an account.

This PR includes the following changes:

* When an account's free or reserved balance drops below the ED, any dust amount will move to the other balance if it is above ED. This means a none of a user's balance will be burned if only one part of their total balance drops below ED. (See test)
* When an account is reaped (because both free and reserved balance is below ED), the `ReapedAccount` event now emits how much dust was burned.
* When root calls `set_balance` with a value less than ED for either reserved or free, we automatically set the value to 0.
* When `set_balance` is called, `BalanceSet` event is emitted with information about the user, final free balance, and final reserved balance.

I am not sure this solves all the problems around tracking tokens within the system, but it should help. 

Some example issues:

* We do not emit any information (necessarily) when a balance is moved between free and reserved.
* We do not emit any information (necessarily) when low level functions are used to manipulate the balance of a user:
    * Slash
    * Deposit Creating
    * etc...

Follow up questions in the original issue.


frame/balances/src/lib.rs
```diff
@@ -456,22 +458,35 @@ decl_module! {
 		) {
 			ensure_root(origin)?;
 			let who = T::Lookup::lookup(who)?;
+			let existential_deposit = T::ExistentialDeposit::get();
+			let mut final_free = new_free;
```
gavofyork: more cleanly written as `let new_free = if new_free < existential_deposit { Zero::zero() } else { new_free };` since this doesn't leave an intermediate value around nor needs use of `mut`.

PR #4392, opened 2019-12-13
Allow decl_error to implement From<frame_system::Error>

thiolliere:
Fix https://github.com/paritytech/substrate/issues/4384

decl_error now accept a new optional attribute, `#[substrate(impl_from_frame_system($frame_system))]` and if this line is given then it crate 3 new variants. and also implement `From<frame_system::Error>`.

Note: This PR is controversial, see related issue.

CC @shawntabrizi


shawntabrizi:
So, from any module, I could write:

```
let sender = ensure_signed(origin)?;
```

And it would just work?

thiolliere:
Yes

thiolliere:
implemented using attributes

thiolliere:
yes

* `#[substrate(impl_from(frame_system))]` is bit missleading to me, otherwise there is
* `#[substrate(from_frame_system(frame_system))]`, or
* `#[impl_from_frame_system(frame_system))]`

shawntabrizi:
@xlc @thiolliere @bkchr 

We decided to not merge this in, and instead just use `as_str` for now. Should hope to get a better error story in the future.

PR #4394, opened 2019-12-13
Add a `browser-utils` crate

expenses:
There is a lot of shared logic between the test substrate/polkadot in-browser light clients I've been working on, so I'm putting it in this `browser-utils` crate.

For clarity: I'm still working out how to best approach certain changes needed to get the light clients to work properly, so I'm merging things piece-by-piece.


utils/browser/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+use std::task::Poll;
+use std::pin::Pin;
+use chain_spec::Extension;
+
+pub use libp2p::wasm_ext::{ExtTransport, ffi::Transport};
+
+/// Set up a panic hook and console logging.
+pub fn set_hooks(log_level: log::Level) {
+    console_error_panic_hook::set_once();
+    let _ = console_log::init_with_level(log_level);
```
expenses: It might be a good idea to re-export these functions instead.

bin/node/cli/src/browser.rs
```diff
@@ -15,45 +15,27 @@
...
 	start_inner(wasm_ext)
+		.await
 		.map_err(|err| JsValue::from_str(&err.to_string()))
 }
 
-fn start_inner(wasm_ext: wasm_ext::ffi::Transport) -> Result<Client, Box<dyn std::error::Error>> {
-	console_error_panic_hook::set_once();
-	console_log::init_with_level(log::Level::Info);
+async fn start_inner(wasm_ext: Transport) -> Result<Client, Box<dyn std::error::Error>> {
+	set_hooks(log::Level::Debug);
```
gnunicorn: Are you intentionally setting a different log level then we had before?

utils/browser/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+futures01 = { package = "futures", version = "0.1.29" }
+log = "0.4.8"
+libp2p = { version = "0.13.1", default-features = false }
+console_error_panic_hook = "0.1.1"
+console_log = "0.1.2"
+js-sys = "0.3.22"
+wasm-bindgen = "0.2.55"
+wasm-bindgen-futures = "0.4.5"
+kvdb-web = "0.1.1"
+service = { package = "sc-service", path = "../../client/service", default-features = false }
```
gnunicorn: will need a version now. The following two lines, too.
```suggestion
service = { version = "2.0.0", package = "sc-service", path = "../../client/service", default-features = false }
```

utils/browser/src/lib.rs
gnunicorn: Yeah, not sure this makes a lot of sense to me. All this does is encapsulate two functions into one and swallow an error. First wanted to suggest you return the error ... but then the difference is almost non existent so I wonder if this function is that useful...

bin/node/cli/src/browser.rs
expenses: No, that was an experimental change and I forgot to change it back

bin/node/cli/browser-demo/build.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env sh
 wasm-pack build --target web --out-dir ./browser-demo/pkg --no-typescript --release ./.. -- --no-default-features --features "browser"
-python -m SimpleHTTPServer 8000
+python -m http.server 8000
```
tomaka: :+1: Python2 is now deprecated :sweat_smile:

tomaka:
IMO we should be more explicit in the `keystore`. The `keystore` should have a "memory-only" mode rather than using `#[cfg]` stuff. Once the `keystore` can be memory-only, we also no longer need to pass a dummy keystore path in the configuration.

What this PR does is okay-ish but doesn't go in the direction of having clean code.

We can either merge now and fix later, or improve this PR itself.

expenses:
@tomaka What can be cleaned up?

tomaka:
What I mentioned: not having platform-specific code in the `keystore`.

client/service/src/builder.rs
```diff
@@ -250,11 +250,15 @@ where TGen: RuntimeGenesis, TCSExt: Extension {
 		(),
 		TLightBackend<TBl>,
 	>, Error> {
+		#[cfg(not(target_os = "unknown"))]
```
tomaka: Well, the next obvious step is to adjust `config.keystore_path` to be an enum of either a path or "in memory".

expenses: 👍

tomaka:
This needs a Polkadot PR because of the change in the configuration now though.

gavofyork:
tests not passing either

bkchr:
Needs resolving.

expenses:
Not sure what's causing `wasm-bindgen-webidl` to not compile - that version seems to work fine on my machine and the build server.

PR #4395, opened 2019-12-13
[big refactor] Remove crate aliasing.

tomusdrw:
This PR removes (almost) all crate renames in the repository. The only exceptions are external crates: like `parity-scale-codec` and `finality-grandpa`.

Related: #4099 , #4318 

Open issues:
1. Some crates are re-exported I didn't rename them and I feel it looks a bit weird.
2. Constructing runtimes from unrenamed pallet has more severe consequences:
2.1. Renaming `system` does not really work well, in some places we just assume that it's renamed. There is a `where system = X` but I feel it doesn't really work well and IMHO should better be removed to simplify macros. You can always rename with `use X as system;`.
2.2. It affects `GenesisConfig` and generated `Event/Call`. `pallet_` prefix is present in many places.

My gut feeling is that we should still rename whenever we construct the runtime, but avoid renames in every other part of code (CC @bkchr, @gavofyork). Happy to add a commit to this PR to revert the renames in `node-template/runtime`, `node/runtime` and `test-utils/runtime`.


tomusdrw:
Resolving conflicts now.

gavofyork:
@tomusdrw is there a polkadot fix for this yet?

tomusdrw:
@gavofyork it shouldn't be necessary. This PR does not rename any of the crates, just changes how we import them internally in substrate. I'll check if it works correctly though tomorrow morning.

tomusdrw:
Ah, just realised that because I've changed couple of defaults in our macros it might now expect `frame_system` crate instead of renamed `system`. I can revert those though, cause as I said I'm not really convinced that avoiding crate renames when amalgamating runtime is a good idea.

gnunicorn:
@tomusdrw if the revert is easy enough, let's do that. I agree that having renames _within_ construct runtime is probably nicer and as those aren't what is usually mixed up (`runtime-primitives` vs `primitives`) there is no obvious value added enforcing that rule within FRAME. Personally would prefer if we did `use frame_system as system` in the code raser than aliasing with cargo though, but if that change is too much , let's do it later.

tomusdrw:
I've reverted the changes in `node-template`, but not in `node/runtime` or `test-runtime`. Also polkadot should not require any changes now, the current failure is unrelated to this PR.

thiolliere:
> Personally would prefer if we did use frame_system as system in the code raser than aliasing with cargo though

In rust 2015 people could do `extern crate MyCrate as MyCrateRename` and this MyCrateRename namespace could be used in other modules of your crate.
In rust 2018 people "deprecated" the use of `extern crate` and I remember hearing people saying that renames of crate should be done in Cargo toml level now.

`use frame_system as system` is not very viable as it is required to write it in every module.
In rust 2018 either we rename in cargo toml or with `extern crate` which I feel people tends to deprecate.

PR #4397, opened 2019-12-15
babe: exponential backoff on missed blocks

rphmeier:
Swaps out the linear backoff for an exponential one.


frame/support/src/storage/mod.rs
```diff
@@ -364,8 +364,10 @@ impl<Value: Decode> Iterator for PrefixIterator<Value> {
 	type Item = Value;
 
 	fn next(&mut self) -> Option<Self::Item> {
-		match sp_io::storage::next_key(&self.previous_key) {
-			Some(next_key) if next_key.starts_with(&self.prefix[..]) => {
```
rphmeier: this wasn't compiling on 1.39.0.

PR #4401, opened 2019-12-16
Identity module enhancements

gavofyork:
This adds two features:
- `SuperOf` public map item, for looking up the super account of a sub-account. Sub-accounts should be able to be looked up directly since the UI has no way of efficiently enumerating them.
- `fn set_account_id` dispatchable call so that registrars are able to reset their account ID.

### DIFF
--- a/frame/identity/src/lib.rs
+++ b/frame/identity/src/lib.rs
@@ -527,8 +527,6 @@ decl_module! {
 			}
 		}
 
-		// TODO: revamp any other instances where we modify SubsOf
-
 		/// Clear an account's identity info and all sub-account and return all deposits.
 		///
 		/// Payment: All reserved balances on the account are returned.
@@ -541,7 +539,7 @@ decl_module! {
 		/// # <weight>
 		/// - `O(R + S + X)`.
 		/// - One balance-reserve operation.
-		/// - Two storage mutations.
+		/// - `S + 2` storage deletions.
 		/// - One event.
 		/// # </weight>
 		fn clear_identity(origin) {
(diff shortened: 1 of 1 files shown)


frame/identity/src/lib.rs
```diff
@@ -506,13 +512,23 @@ decl_module! {
...
+
+			if ids.is_empty() {
 				<SubsOf<T>>::remove(&sender);
 			} else {
-				<SubsOf<T>>::insert(&sender, (new_deposit, subs));
+				<SubsOf<T>>::insert(&sender, (new_deposit, ids));
 			}
 		}
 
+		// TODO: revamp any other instances where we modify SubsOf
```
shawntabrizi: Still TODO?

frame/identity/src/lib.rs
```diff
@@ -531,8 +547,12 @@ decl_module! {
 		fn clear_identity(origin) {
```
shawntabrizi: Weight comment should be updated, cannot make a recommendation on GitHub but:

```
/// - Two storage mutations.
```

Is now 2 + n storage mutations where n is number of sub accounts.

frame/identity/src/lib.rs
```diff
@@ -759,8 +806,12 @@ decl_module! {
 			// Figure out who we're meant to be clearing.
 			let target = T::Lookup::lookup(target)?;
 			// Grab their deposit (and check that they have one).
+			let (subs_deposit, sub_ids) = <SubsOf<T>>::take(&target);
 			let deposit = <IdentityOf<T>>::take(&target).ok_or("not named")?.total_deposit()
-				+ <SubsOf<T>>::take(&target).0;
+				+ subs_deposit;
+			for sub in sub_ids.iter() {
+				<SuperOf<T>>::remove(sub);
```
shawntabrizi: Weight note update here too.

PR #4403, opened 2019-12-16
Stubs for removed functions to make wasm environment compatible with old runtimes

svyatonik:
Functions were removed in https://github.com/paritytech/substrate/pull/4162 => current client is unable to sync Flaming Fir (but we can reset this chain) and at some point (when commit is picked into `polkadot-master`) will be unable to sync Kusama (we do not want to reset this chain, right?).

### DIFF
--- a/client/executor/src/deprecated_host_interface.rs
+++ b/client/executor/src/deprecated_host_interface.rs
@@ -217,6 +217,26 @@ impl_wasm_host_interface! {
 			Ok(sp_io::storage::set(&key, &value))
 		}
 
+		ext_set_child_storage(
+			_storage_key_data: Pointer<u8>,
+			_storage_key_len: WordSize,
+			_key_data: Pointer<u8>,
+			_key_len: WordSize,
+			_value_data: Pointer<u8>,
+			_value_len: WordSize,
+		) {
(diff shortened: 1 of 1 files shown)


svyatonik:
For reference, if you'll try to sync FF now (with empty db), you'll get:
```
2019-12-16 14:27:52 Substrate Node
2019-12-16 14:27:52   version 2.0.0-13b1308-x86_64-linux-gnu
2019-12-16 14:27:52   by Parity Technologies, 2017-2019
2019-12-16 14:27:52 Chain specification: Flaming Fir
2019-12-16 14:27:52 Node name: mountainous-fire-8140
2019-12-16 14:27:52 Roles: FULL
2019-12-16 14:27:53 Initializing Genesis block/state (state: 0xd733…e696, header-hash: 0xb653…1d25)
2019-12-16 14:27:53 Loading GRANDPA authority set from genesis on what appears to be first startup.
2019-12-16 14:27:54 cannot create a runtime: Instantiation("Instantiation: Export ext_get_allocated_child_storage not found")
Error: Service(Client(Execution(InvalidCode("Instantiation(\"Instantiation: Export ext_get_allocated_child_storage not found\")"))))
```

PR #4404, opened 2019-12-16
Setting versions locally  towards 2.0 release

gnunicorn:
In order to allow to publish, `path`-dependencies must have a matching version specified, too. This PR attempts to make us ready for a 2.0 releasing to crates by :

- [x] cleaning up the path-definitions, move the version to the front and path to last place in all `Cargo.toml`
- [x] set the version on all path-defined dependencies
- [x] put sc-network and \*consensus\* versions into `0.8.0`, incl `pallet-aura` and `pallet-babe`
- [x] bump other pallets to 2.0

refs #4099


bin/node/runtime/Cargo.toml
```diff
@@ -14,59 +14,59 @@ rustc-hex = { version = "2.0", optional = true }
...
+frame-executive = { version = "2.0.0", default-features = false, path = "../../../frame/executive" }
+pallet-finality-tracker = { version = "2.0.0", default-features = false, path = "../../../frame/finality-tracker" }
+pallet-grandpa = { version = "2.0.0", default-features = false, path = "../../../frame/grandpa" }
+pallet-im-online = { version = "2.0.0", default-features = false, path = "../../../frame/im-online" }
+pallet-indices = { version = "2.0.0", default-features = false, path = "../../../frame/indices" }
+pallet-membership = { version = "2.0.0", default-features = false, path = "../../../frame/membership" }
+pallet-nicks = { version = "2.0.0", default-features = false, path = "../../../frame/nicks" }
+pallet-offences = { version = "2.0.0", default-features = false, path = "../../../frame/offences" }
+pallet-randomness-collective-flip = { version = "2.0.0", default-features = false, path = "../../../frame/randomness-collective-flip" }
+pallet-session = { features = ["historical"], path = "../../../frame/session", default-features = false }
```
tomusdrw: Why no version here?

bin/node/runtime/Cargo.toml
```diff
@@ -14,59 +14,59 @@ rustc-hex = { version = "2.0", optional = true }
 serde = { version = "1.0.102", optional = true }
 
 # primitives
-sp-authority-discovery = { path = "../../../primitives/authority-discovery", default-features = false }
-sp-consensus-babe = { path = "../../../primitives/consensus/babe", default-features = false }
+sp-authority-discovery = { version = "2.0.0", default-features = false, path = "../../../primitives/authority-discovery" }
+sp-consensus-babe = { version = "0.8", default-features = false, path = "../../../primitives/consensus/babe" }
 sp-block-builder = { path = "../../../primitives/block-builder", default-features = false}
```
tomusdrw: Why is it missing here?

bin/node/runtime/Cargo.toml
gnunicorn: that's a good question. I was doing a regexp-replace, it appears it didn't catch all cases.. need to figure out why and which ones...

bin/node/runtime/Cargo.toml
gnunicorn: I suppose this didn't match because of a missing space at the end...

gnunicorn: not intentional.

bin/node-template/Cargo.toml
```diff
@@ -18,25 +18,25 @@ tokio = "0.1.22"
...
-sc-consensus-aura = { path = "../../client/consensus/aura" }
-sp-consensus-aura = { path = "../../primitives/consensus/aura" }
-sp-consensus = { path = "../../primitives/consensus/common" }
-grandpa = { package = "sc-finality-grandpa", path = "../../client/finality-grandpa" }
-grandpa-primitives = { package = "sp-finality-grandpa", path = "../../primitives/finality-grandpa" }
-sc-client = {  path = "../../client/" }
-node-template-runtime = { path = "runtime" }
-sp-runtime = { path = "../../primitives/runtime" }
-sc-basic-authority = { path  = "../../client/basic-authorship"}
+sp-io = { version = "2.0.0", path = "../../primitives/io" }
```
gavofyork: why set versions as well as path? it's just redundant additional tedious data to maintain isn't it?

gnunicorn: Unfortunately, defining the version is a requirement to publish on `crates.io`. Cargo doesn't have any `publish --all` and won't resolve path to figure out which version to us–in general its publishing helper for bigger workspaces are pretty rudimentary. On the other side `path` is obviously a lot easier and nicer for development.

Fortunately, because cargo uses semver this isn't much of a deal. We can just define `version = "2.0.0"` and even if we bump the minor or patch version of that crate to make a new release, this doesn't need any update on this crate, but cargo will figure out the best version on `cargo update` (or `cargo build` if first time). Our internal `Cargo.lock` will also update without problems. This also allows us to do version updates more independently per crate, like updating network and consensus and apply the changes to the crates that follow without having to publish a breaking change, if those dependencies don't alter their externally perceived API.

Having both is the common way to stay sane, and is what we do in the [libp2p workspace repo](https://github.com/libp2p/rust-libp2p/blob/1aeb264cbc0e262af219c92e18eb0025724a9005/transports/dns/Cargo.toml#L13) and common practice on other workspace repos, too.

gavofyork: 👍

Issue #4365, opened 2019-12-11
Update Frame Pallets to use `decl_error!`


shawntabrizi:
Since https://github.com/paritytech/substrate/pull/3433 has been merged in, we have had the ability to create really nice error experiences for users using Substrate Pallets.

We just need to migrate existing errors into this new format.

I expect, since this will change a bunch of modules, we will do this in a few steps. This issue will track the progress:

- [x] assets
- [x] aura
- [x] authority-discovery
- [x] authorship
- [x] babe
- [x] balances
- [x] collective
- [x] contracts
- [x] democracy
- [x] elections
- [x] elections-phragmen
- [x] evm
- [x] example
- [x] executive
- [x] finality-tracker
- [x] generic-asset
- [x] grandpa
- [x] identity
- [x] im-online
- [x] indices
- [x] membership
- [x] metadata
- [x] nicks
- [x] offences
- [x] randomness-collective-flip
- [x] scored-pool
- [x] session
- [x] society
- [x] staking
- [x] sudo
- [x] support
- [x] system
- [x] timestamp
- [x] transaction-payment
- [x] treasury
- [x] utility

shawntabrizi:
Sudo: https://github.com/paritytech/substrate/pull/4369

DarkEld3r:
Treasury: #4370.

As far as I can see, there are no custom errors in `utility`.

DarkEld3r:
Stacking: #4387.

shawntabrizi:
Balances: https://github.com/paritytech/substrate/pull/4405

PR #4405, opened 2019-12-17
Update Balances Pallet for `decl_error!`

shawntabrizi:
Helps: https://github.com/paritytech/substrate/issues/4365

This is a PR to update the Balances Pallet to use `decl_error!`.

Errors available in the Balances module are:

```
decl_error! {
	pub enum Error for Module<T: Trait<I>, I: Instance> {
		/// Vesting balance too high to send value
		VestingBalance,
		/// Account liquidity restrictions prevent withdrawal
		LiquidityRestrictions,
		/// Got an overflow after adding
		Overflow,
		/// Balance too low to send value
		InsufficientBalance,
		/// Value too low to create account due to existential deposit
		ExistentialDeposit,
		/// Transfer/payment would kill account
		KeepAlive,
		/// A vesting schedule already exists for this account
		ExistingVestingSchedule,
		/// Beneficiary account must pre-exist
		DeadAccount,
	}
}
```

Feel free to nit-pick at the strings chosen here here.


frame/balances/src/lib.rs
```diff
@@ -930,29 +953,29 @@ where
 		let fee = if would_create { T::CreationFee::get() } else { T::TransferFee::get() };
 		let liability = match value.checked_add(&fee) {
 			Some(l) => l,
-			None => Err("got overflow after adding a fee to value")?,
+			None => Err(Error::<T, I>::Overflow)?,
```
DarkEld3r: Perhaps, it can be rewritten without a match?
```rust
let liability = match value.checked_add(&fee).ok_or(Error::<T, I>::Overflow)?;
```

frame/balances/src/lib.rs
```diff
@@ -930,29 +953,29 @@ where
...
+			Err(Error::<T, I>::ExistentialDeposit)?
 		}
 		Self::ensure_can_withdraw(transactor, value, WithdrawReason::Transfer.into(), new_from_balance)?;
 
 		// NOTE: total stake being stored in the same type means that this could never overflow
 		// but better to be safe than sorry.
 		let new_to_balance = match to_balance.checked_add(&value) {
 			Some(b) => b,
-			None => Err("destination balance too high to receive value")?,
+			None => Err(Error::<T, I>::Overflow)?,
```
DarkEld3r: And here.

PR #4414, opened 2019-12-17
Prioritize new blocks over old forks when syncing

### LINKED ISSUE #4398: Sync stopped
I've been running polkadot with just a single peer connection, in part to uncover these bugs.

Full trace (`-l sync=trace`) attached.

[synctrace-dead-1.txt](https://github.com/paritytech/substrate/files/3966060/synctrace-dead-1.txt)

Related #4367

arkpar:
Fixes #4398

* Prioritize new blocks over old forks when syncing
* Additionally check for existing blocks when downloading old forks as they might have been imported while the request was active.


primitives/consensus/common/src/import_queue/buffered_link.rs
```diff
@@ -137,12 +138,17 @@ impl<B: BlockT> BufferedLinkReceiver<B> {
 	/// it is as if this method always returned `Poll::Pending`.
 	pub fn poll_actions(&mut self, cx: &mut Context, link: &mut dyn Link<B>) {
```
arkpar: This function would panic occasionally on exit because `Stream::poll_next` is called for a finished stream.

arkpar:
> Sync changes looks good.
> 
> I'm not able to judge the overall idea of prioritizing normal sync over explicit forks sync, though. Iiuc, GRANDPA will be unable to finalize (or smth similar - I'm not aware) anything until it will sync these forks, right? So if we're far beyond the best known block (reported by peers), finalization will halt? But probably this is what we should expect - it looks similar to authoring on unsynced nodes. So probably just ignore this - I was just writing my thoughts.

When doing full sync finalization uses a different mechanism to catch up. GRANDPA only gossips when full sync is idle.

PR #4421, opened 2019-12-17
Fix Fees in Substrate

shawntabrizi:
This PR is the result of investigating the fee system in Substrate and possible issues around calculating fees.

When the `TargetedFeeAdjustment` calculation reached `-1`, the fee calculation mechanism would reduce the total fee a user paid to zero.

Instead, we now ensure that the `base_fee` is unaffected by this adjustment multiplier, so even when the network is so un-congested, the user will pay at a minimum some `base_fee`. This adjustment multiplier now only affects the length fee and weight fee of a transaction.

```
final_fee = base_fee + targeted_fee_adjustment(len_fee + weight_fee) + tip;
```

This includes a new test which verifies that the `compute_fee` function behaves as it should in various scenarios with this new formula.

### DIFF
--- a/bin/node/executor/src/lib.rs
+++ b/bin/node/executor/src/lib.rs
@@ -87,14 +87,14 @@ mod tests {
 
 	/// Default transfer fee
 	fn transfer_fee<E: Encode>(extrinsic: &E, fee_multiplier: Fixed64) -> Balance {
-		let length_fee = TransactionBaseFee::get() +
-			TransactionByteFee::get() *
-			(extrinsic.encode().len() as Balance);
+		let length_fee = TransactionByteFee::get() * (extrinsic.encode().len() as Balance);
 
 		let weight = default_transfer_call().get_dispatch_info().weight;
 		let weight_fee = <Runtime as pallet_transaction_payment::Trait>::WeightToFee::convert(weight);
 
-		fee_multiplier.saturated_multiply_accumulate(length_fee + weight_fee) + TransferFee::get()
+		let base_fee = TransactionBaseFee::get();
+
+		 base_fee + fee_multiplier.saturated_multiply_accumulate(length_fee + weight_fee) + TransferFee::get()
 	}
 
 	fn default_transfer_call() -> pallet_balances::Call<Runtime> {
@@ -537,7 +537,7 @@ mod tests {
 				},
 				EventRecord {
 					phase: Phase::ApplyExtrinsic(1),
-					event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1984780231392)),
+					event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1984788199392)),
 					topics: vec![],
 				},
 				EventRecord {
@@ -561,7 +561,7 @@ mod tests {
 				},
 				EventRecord {
 					phase: Phase::ApplyExtrinsic(2),
-					event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1984780231392)),
+					event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1984788199392)),
 					topics: vec![],
 				},
 				EventRecord {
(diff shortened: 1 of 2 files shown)


shawntabrizi:
I think there is still a deeper issue, this change does not actually affect the calculation of fees afaict

frame/transaction-payment/src/lib.rs
```diff
@@ -156,9 +161,9 @@ impl<T: Trait + Send + Sync> ChargeTransactionPayment<T> {
 	{
 		if info.pays_fee {
 			let len = <BalanceOf<T>>::from(len);
-			let base = T::TransactionBaseFee::get();
+			let base_fee = T::TransactionBaseFee::get();
```
kianenigma: better to be declared down there when it is used.

frame/transaction-payment/src/lib.rs
```diff
@@ -533,14 +540,93 @@ mod tests {
...
+
+			// Tip only, no fees works
+			let dispatch_info = DispatchInfo {
+				weight: 0,
+				class: DispatchClass::Operational,
+				pays_fee: false,
+			};
+			assert_eq!(ChargeTransactionPayment::<Runtime>::compute_fee(0, dispatch_info, 10), 10);
+			// No tip, only base fee works
+			assert_eq!(TransactionBaseFee::get(), 100);
```
kianenigma: I think it is evident from `ExtBuilder::default().fees(100, 10, 1)`

frame/transaction-payment/src/lib.rs
```diff
@@ -533,14 +540,93 @@ mod tests {
...
+			let dispatch_info = DispatchInfo {
+				weight: 0,
+				class: DispatchClass::Operational,
+				pays_fee: true,
+			};
+			assert_eq!(ChargeTransactionPayment::<Runtime>::compute_fee(0, dispatch_info, 0), 100);
+			// Tip + base fee works
+			assert_eq!(ChargeTransactionPayment::<Runtime>::compute_fee(0, dispatch_info, 69), 169);
+			// Len (byte fee) + base fee works
+			assert_eq!(TransactionByteFee::get(), 10);
```
kianenigma: likewise ^^

frame/transaction-payment/src/lib.rs
```diff
@@ -533,14 +540,93 @@ mod tests {
...
+							+ info.weight.min(MaximumBlockWeight::get()) as u64 * 2 /* weight * weight_to_fee */
+						) * 3 / 2
 				},
 			);
 
 		});
 	}
+
+	#[test]
+	fn compute_fee_works() {
```
kianenigma: I recommend the test to be broken into smaller ones, but up to you regarding how to do it. it is a bit too lengthy. Maybe: `compute_fee_works_with_multipier`, `compute_fee_works_without_multiplier`?

bin/node/executor/src/lib.rs
```diff
@@ -561,7 +561,7 @@ mod tests {
 				},
 				EventRecord {
 					phase: Phase::ApplyExtrinsic(2),
-					event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1984780231392)),
+					event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1984788199392)),
```
shawntabrizi: I don't know exactly how this number is calculated, I just copied the output of the failing test...

I would assume that is reasonable to do in this scenario.

kianenigma: maybe, but I wouldn't make that assumption, will have a look

kianenigma: This comes from the `_4, Treasury,` part of the `pub type DealWithFees = SplitTwoWays` given to `transaction_payment::Trait`. Actually I think it is a great opportunity to dig into it and make it a useful and critical assertion: **it must be 20% of the total fee of a particular transaction**. 

I really think this is important to be asserted properly. If not fixed here, please make an issue (and assign me)

PR #4422, opened 2019-12-17
Introduce frame::system::Error::InvalidOrigin

### LINKED ISSUE #4409: Introduce general BadOrigin runtime error
Right now module use various ways for reporting bad origins, even though the origin recognition APIs are all reusable library code. Some report a `&str` (of various content), others report a local `BadOrigin`-like error type.

Instead, there should be a single `BadOrigin` error (e.g. defined in `frame_system`) that all `ensure_origin`/`try_origin`/`ensure_root`/`ensure_nobody`/`ensure_signed` functions and their ilk use. If modules want to customise the error to something local, then they can with `map_err`, but a simple `ensure_root(origin)?` should Just Work.

DarkEld3r:
Key points:
- `InvalidOrigin` has been added to `frame::system::Error`.
- `EnsureOrigin` trait has been moved from `primitives/runtime/src/traits.rs` to `frame::system`.

Closes #4409.

### DIFF
--- a/frame/treasury/src/lib.rs
+++ b/frame/treasury/src/lib.rs
@@ -67,11 +67,11 @@ use frame_support::traits::{
 };
 use sp_runtime::{Permill, ModuleId};
 use sp_runtime::traits::{
-	Zero, EnsureOrigin, StaticLookup, AccountIdConversion, Saturating, ModuleDispatchError,
+	Zero, StaticLookup, AccountIdConversion, Saturating, ModuleDispatchError,
 };
 use frame_support::weights::SimpleDispatchInfo;
 use codec::{Encode, Decode};
-use frame_system::{self as system, ensure_signed};
+use frame_system::{self as system, ensure_signed, EnsureOrigin};
 
 type BalanceOf<T> = <<T as Trait>::Currency as Currency<<T as frame_system::Trait>::AccountId>>::Balance;
 type PositiveImbalanceOf<T> = <<T as Trait>::Currency as Currency<<T as frame_system::Trait>::AccountId>>::PositiveImbalance;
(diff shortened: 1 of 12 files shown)


primitives/runtime/src/traits.rs
```diff
@@ -134,28 +134,6 @@ impl<
 	}
 }
 
-/// An error type that indicates that the origin is invalid.
```
bkchr: I think the trait should stay here and we introduce the `BadOrigin` type directly here.

We already have some inherent errors that are declared by `decl_error!`. I would add a new one `BadOrigin` that implements `From<BadOrigin>` that is defined in this file here.

bkchr: I have some upcoming changes in my mind for the error stuff, that would conflict in the way it is implemented currently.

frame/treasury/src/lib.rs
```diff
@@ -168,7 +168,7 @@ decl_module! {
 		/// # </weight>
 		#[weight = SimpleDispatchInfo::FixedOperational(100_000)]
 		fn reject_proposal(origin, #[compact] proposal_id: ProposalIndex) {
-			T::RejectOrigin::ensure_origin(origin).map_err(|e| Into::<&str>::into(e))?;
+			T::RejectOrigin::ensure_origin(origin).map_err(|e| e.as_str())?;
```
gavofyork: why does it need to be `.as_str()`? can't the error be returned back directly?

niklasad1: I think `ensure_origin` returns `system::Error` which needs to be converted to `treasury::Error` accordingly to work but I think `Into::<&str>::into(e)` should still work.

bkchr:
Superseded by: https://github.com/paritytech/substrate/pull/4449

Issue #2303, opened 2019-04-17
A way to expose limited set of RPC methods


xlc:
Is it possible to only expose a limited set of RPC methods for diagnosis purpose only? Such like sync progress, block height, health, etc.

Those RPC should be very safe and readonly and cannot be (easily) abused so they can be open to public behind a general rate limiting API gateway.

RPC like submit transaction and query historical state have a larger attack surface so we would like to close them by default.

bkchr:
CC @tomusdrw @ascjones

tomusdrw:
I'd rather invest some time to make https://github.com/tomusdrw/jsonrpc-proxy more production ready than to build that into substrate. @xlc is that something that would work for you?

The permissioning configuration example can be found here:
https://github.com/tomusdrw/jsonrpc-proxy/blob/master/examples/permissions.json

Basically it's just the matter of running the proxy + pointing to the configuration file.

xlc:
Yes an easy configurable json rpc proxy will work too.

PR #4423, opened 2019-12-17
RPCs for reserved peers

rphmeier:


client/rpc-api/src/system/error.rs
```diff
@@ -43,6 +45,11 @@ impl From<Error> for rpc::Error {
 				message: format!("{}", e),
 				data: serde_json::to_value(h).ok(),
 			},
+			Error::MalformattedPeerArg(ref e) => rpc::Error {
+				code :rpc::ErrorCode::ServerError(BASE_ERROR + 2),
+				message: format!("{}", e),
```
niklasad1: already a `String` no need for `format!`

```suggestion
				message: e.clone(),
```

client/rpc-api/src/system/mod.rs
```diff
@@ -65,6 +67,21 @@ pub trait SystemApi<Hash, Number> {
 	#[rpc(name = "system_networkState", returns = "jsonrpc_core::Value")]
 	fn system_network_state(&self) -> Receiver<jsonrpc_core::Value>;
 
+	/// Adds a reserved peer. Returns the empty string or an error. The string
```
niklasad1: returns `()`?

client/rpc-api/src/system/mod.rs
```diff
@@ -65,6 +67,21 @@ pub trait SystemApi<Hash, Number> {
...
+	/// parameter should encode a `p2p` multiaddr.
+	///
+	/// `/ip4/198.51.100.19/tcp/30333/p2p/QmSk5HQbn6LhUwDiNMseVUjuRYhEtYj4aUZ6WfWoGURpdV`
+	/// is an example of a valid, passing multiaddr with PeerId attached.
+	#[rpc(name = "system_addReservedPeer", returns = "()")]
+	fn system_add_reserved_peer(&self, peer: String)
+		-> Compat<BoxFuture<'static, std::result::Result<(), jsonrpc_core::Error>>>;
+
+	/// Remove a reserved peer. Returns the empty string or an error. The string
```
niklasad1: returns `()`?

client/rpc-api/src/system/mod.rs
rphmeier: As I understand it, this will serialize to the empty string over RPC.

client/rpc-api/src/system/mod.rs
rphmeier: Same as above

client/rpc-api/src/system/mod.rs
niklasad1: Ok, I see

client/rpc-api/src/system/error.rs
```diff
@@ -43,6 +45,11 @@ impl From<Error> for rpc::Error {
 				message: format!("{}", e),
 				data: serde_json::to_value(h).ok(),
 			},
+			Error::MalformattedPeerArg(ref e) => rpc::Error {
+				code :rpc::ErrorCode::ServerError(BASE_ERROR + 2),
+				message: e.clone(),
+				data: Some(e.to_string().into()),
```
tomusdrw: `e` is just the string, so `data` would be just the same as `message`, right? I think we can just skip it then. `data` is meant to contain more detailed information about the error.
```suggestion
				data: None,
```

client/rpc-api/src/system/mod.rs
tomusdrw: Actually `()` is serialized as `null` afair. Might be better to return a `bool` maybe? There was some discussion around `author_insertKey` returning `null`, that it's not an expected result from a succesfull operation.

rphmeier:
going to push one more commit with tests, then will merge. Thanks for the review Niklas!

client/rpc/src/system/mod.rs
```diff
@@ -97,6 +103,34 @@ impl<B: traits::Block> SystemApi<B::Hash, <B::Header as HeaderT>::Number> for Sy
 		Receiver(Compat::new(rx))
 	}
 
+	fn system_add_reserved_peer(&self, peer: String)
```
tomusdrw: The implementation of all methods of this trait is really tedious. It should be auto-generated, but I guess that's a separate task.

rphmeier: I 100% agree. The architecture of having the RPC worker forward requests to something that answers them asynchronously is good, though.

tomusdrw: Yes, I guess we will move to purer architecture with `jsonrpsee`, so I guess no action needed here, I'm just whining :)

rphmeier: When I was writing this PR, using the compat between stable futures and 0.1 was a real headache. i took a quick stab at porting over `jsonrpc` to stable futures but then realized it was a decent amount of work.

xlc:
Can I call this RPC on a node with RPC exposed publicly and causing some troubles?

tomusdrw:
@xlc Yes, but you shouldn't really expose your RPC publicly without a proxy mentioned in #2303

xlc:
I am just start getting nervous on the _default_ setup of a node. The default setup should always be sane and secure. With all those configuration RPCs, a proxy become a must have component to run a public RPC node and I don't see much warning and documents about it.

tomusdrw:
@xlc I share your fears. If you have ideas where we could add more warnings, please submit a PR (happy to review it). Currently:
1. The default setup only listens on `localhost`.
2. There is a bunch of `--*-external` flags (I was actually sure that they are called `unsafe`, but just checked and it's not the case) that expose the server on all interfaces.

I guess the flags would be a good place to add more docs on the subject.

xlc:
#4445

PR #4424, opened 2019-12-17
pallet-evm: default implementation for FeeCalculator and ConvertAccountId and separate gas price

sorpaas:
* Provide default implementation for `FeeCalculator` and `ConvertAccountId`, so that using the EVM module is easier.
* Move gas price parameter to dispatch parameter, and change `FeeCalculator` to output minimal gas price. This allows `FeeCalculator` to depend on storage values and thus more flexible.

### DIFF
--- a/frame/evm/src/lib.rs
+++ b/frame/evm/src/lib.rs
@@ -23,8 +23,8 @@ mod backend;
 
 pub use crate::backend::{Account, Log, Vicinity, Backend};
 
-use sp_std::{vec::Vec, marker::PhantomData};
-use frame_support::{dispatch, decl_module, decl_storage, decl_event};
+use sp_std::vec::Vec;
+use frame_support::{ensure, dispatch, decl_module, decl_storage, decl_event};
 use frame_support::weights::{Weight, WeighData, ClassifyDispatch, DispatchClass, PaysFee};
 use frame_support::traits::{Currency, WithdrawReason, ExistenceRequirement};
 use frame_system::{self as system, ensure_signed};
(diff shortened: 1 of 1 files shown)


frame/evm/src/lib.rs
```diff
@@ -194,13 +217,14 @@ decl_module! {
 			Ok(())
 		}
 
-		#[weight = WeightForCallCreate::<T::FeeCalculator>::default()]
-		fn call(origin, target: H160, input: Vec<u8>, value: U256, gas_limit: u32)
-			-> dispatch::Result
-		{
+		#[weight = WeightForCallCreate]
+		fn call(
+			origin, target: H160, input: Vec<u8>, value: U256, gas_limit: u32, gas_price: U256,
```
gavofyork: style nit: it's `(long, long, long)` or `(
    high,
    high,
    high
)`. not a combination.

frame/evm/src/lib.rs
```diff
@@ -247,11 +271,14 @@ decl_module! {
 			ret
 		}
 
-		#[weight = WeightForCallCreate::<T::FeeCalculator>::default()]
-		fn create(origin, init: Vec<u8>, value: U256, gas_limit: u32) -> dispatch::Result {
+		#[weight = WeightForCallCreate]
+		fn create(
+			origin, init: Vec<u8>, value: U256, gas_limit: u32, gas_price: U256,
```
gavofyork: same

frame/evm/src/lib.rs
```diff
@@ -59,6 +63,25 @@ pub trait ConvertAccountId<A> {
 	fn convert_account_id(account_id: &A) -> H160;
 }
 
+/// Truncate the account id, taking the last 160-bit as the Ethereum address.
+pub struct TruncateConvertAccountId;
+
+impl<A: AsRef<[u8]>> ConvertAccountId<A> for TruncateConvertAccountId {
```
gavofyork: this might not be particularly secure without a hash first.

sorpaas: I changed this to `HashTruncateConvertAccountId` -- it takes a `Hasher`, hash the account id first, and then do the truncate operation.

PR #4426, opened 2019-12-17
Add Edgeware network ID to core/crypto and subkey

drewstone:
This adds the network ID 7 to the core crypto lib and adds Edgeware as a network option in subkey.


drewstone:
Hmm not sure why those old commits persisted..

PR #4429, opened 2019-12-18
Remove incorrect assumption that runners-up were sorted by account

gavofyork:
CC @kianenigma


bkchr:
Test needs to be updated.

PR #4440, opened 2019-12-18
grandpa: fix slow gossip test

andresilva:
In the GRANDPA test `good_commit_leads_to_relay` we test the gossip layer by sending a valid commit message and then making sure it is propagated after validation. This test was taking a long time to run, the reason was that we were only connected to one peer, so after receiving the commit from it we would refuse to resend the message to it until a periodic forced broadcast was triggered (which took 5 minutes to trigger). An easy fix is to add another peer to the test environment to whom the commit will be immediately propagated.


rphmeier:
I see, so we need to introduce the extra peer because we won't send the commit to the same peer who sent it, even though it is valid.

PR #4443, opened 2019-12-18
grandpa: log everything under afg target

andresilva:
For some reason we were logging some grandpa messages under the `finality` target rather than `afg`. I don't remember any clear reason for doing this and in practice it's one more logging target that we need to think about adding. I changed it so that all messages are logged under the `afg` target.


rphmeier:
It should definitely be an all-or-nothing. I could see everything moving over to a `finality` target when we have swappable consensus, but for now this will work.

Issue #4434, opened 2019-12-18
Document ordering of sp_transaction_pool::TransactionStatus


geigerzaehler:
I’m using the `watch_extrinsic` RPC method and would like to know what transaction status changes to expect. For example, is the first item in the stream always either `Ready` or `Invalid`? Does the stream terminate after receiving `Invalid` or `Usurped`?

It would be great if this was documented in the `TransactionStatus` enum and referenced in the `watch_extrinsic` documentation.

bkchr:
CC @tomusdrw

Issue #4438, opened 2019-12-18
Transaction Status should differentiate between InBlock and Finalized.


tomusdrw:
Currently the RPC stream returns `Finalized` event, whenever the transaction is included in block, which obviously is not technically correct.

We should rename current `Finalized` event to `InBlock` and introduce a new event that is triggered by actual finality.

Requested by @jacogr some time ago, but I didn't have time to look at this yet, happy to mentor though and help with the issue.

rphmeier:
> We should rename current Finalized event to InBlock and introduce a new event that is triggered by actual finality.

The chain state doesn't know that well about what's actually finalized

tomusdrw:
Yes I know, but I imagine that it will require another stream of events coming from the consensus protocol about finalised hashes, that can in turn be used by the transaction pool watcher to trigger these events.
It's not going to be trivial, for sure.

rphmeier:
I see, or the API code could correlate this a bit higher-up in the stack, by watching for finality notifications.

Sushisource:
Huge +1 to this. This was deeply confusing to discover that `Finalized` here did not mean what you might reasonably intuitively expect it to mean.

tomusdrw:
CC @jacogr 
We now have events for:
`inBlock`, `retracted` and later for `finalized` or `finalityTimeout`
note that only the latter two will terminate the stream.

PR #4446, opened 2019-12-18
Document TransactionStatus and fix termination conditions.

tomusdrw:
Resolves: #4434 
Related: #4438 

This PR adds documentation on the `TransactionStatus` type, outlining the ordering of events that may occur on specific transaction.
I've also noticed that `Usurped` and `Dropped` events were not causing stream termination, yet the transaction was no longer in the pool. Technically it may still re-enter the pool, but actually it's the case for every terminating state, and imho the callers should re-subscribe if they are interested. (CC @jacogr hope it's backward compatible for polkadot.js)

Additionally I renamed `Finalized` event to `InBlock` in the code, but kept backward compatibility on the RPC layer. This is an introduction to #4438

### DIFF
--- a/client/rpc-api/src/author/mod.rs
+++ b/client/rpc-api/src/author/mod.rs
@@ -60,6 +60,9 @@ pub trait AuthorApi<Hash, BlockHash> {
 	) -> Result<Vec<Hash>>;
 
 	/// Submit an extrinsic to watch.
+	///
+	/// See `sp_transaction_pool::TransactionStatus` for details on transaction
+	/// lifecycle.
 	#[pubsub(
 		subscription = "author_extrinsicUpdate",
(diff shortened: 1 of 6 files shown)


client/rpc-api/src/author/mod.rs
```diff
@@ -60,6 +60,9 @@ pub trait AuthorApi<Hash, BlockHash> {
 	) -> Result<Vec<Hash>>;
 
 	/// Submit an extrinsic to watch.
+	///
+	/// See `sp_transaction_pool::TransactionStatus` for details on transaction
```
bkchr:
```suggestion
	/// See [`TransactionStatus`](sp_transaction_pool::TransactionStatus) for details on transaction
```

PR #4448, opened 2019-12-18
Warn about using --rpc-external and --ws-external options

### LINKED ISSUE #4445: Print warnings when RPC are exposed publicly
It is not safe to expose RPC publicly without a proxy server to block some configuration RPC endpoints such as `system_addReservedPeer`.

When `--http-external` or `--ws-external` flag is passed, the node should print some warning message to inform user this is could be dangerous. Maybe add `--unsafe-http-external` / `--unsafe-ws-external` that suppress the warning.

The help message for `--*-external` flag should also mention they are unsafe.

Could even panic and require use of `--unsafe-*-external` when `--validator` flag is also used.

More documentations are also required (Out of scope of this issue, just for reference):

- Which endpoints are safe to expose and which are not
- How to setup a proxy to block / require auth for unsafe RPC endpoints

cc @tomusdrw

DarkEld3r:
Closes #4445.

### DIFF
--- a/client/cli/src/lib.rs
+++ b/client/cli/src/lib.rs
@@ -890,8 +890,8 @@ where
 			}
 		});
 
-	let rpc_interface: &str = if cli.rpc_external { "0.0.0.0" } else { "127.0.0.1" };
-	let ws_interface: &str = if cli.ws_external { "0.0.0.0" } else { "127.0.0.1" };
+	let rpc_interface: &str = interface_str(cli.rpc_external, cli.unsafe_rpc_external, cli.validator)?;
+	let ws_interface: &str = interface_str(cli.ws_external, cli.unsafe_ws_external, cli.validator)?;
 	let grafana_interface: &str = if cli.grafana_external { "0.0.0.0" } else { "127.0.0.1" };
 
 	config.rpc_http = Some(parse_address(&format!("{}:{}", rpc_interface, 9933), cli.rpc_port)?);
@@ -929,6 +929,28 @@ where
 	config.force_authoring = cli.shared_params.dev || cli.force_authoring;
 
 	Ok(config)
+}
+
+fn interface_str(
+	is_external: bool,
+	is_unsafe_external: bool,
+	is_validator: bool
+) -> Result<&'static str, error::Error> {
+	if is_external && is_validator {
+		return Err(error::Error::Input("--rpc-external and --ws-external options shouldn't be \
+		used if the node is running as vaidator.".to_owned()));
+	}
+
+	if is_external {
+		log::warn!("It isn't safe to expose RPC publicly without a proxy server. \
+		Use `--unsafe-rpc-external` to suppress the warning if you understand the risks.");
+	}
+
+	if is_external || is_unsafe_external {
+		Ok("0.0.0.0")
+	} else {
+		Ok("127.0.0.1")
+	}
 }
 
 /// Creates a configuration including the database path.
--- a/client/cli/src/lib.rs
+++ b/client/cli/src/lib.rs
@@ -934,19 +934,18 @@ where
 fn interface_str(
 	is_external: bool,
 	is_unsafe_external: bool,
-	is_validator: bool
+	is_validator: bool,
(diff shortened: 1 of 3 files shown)


client/cli/src/lib.rs
```diff
@@ -931,6 +931,28 @@ where
...
 }
 
+fn interface_str(
+	is_external: bool,
+	is_unsafe_external: bool,
+	is_validator: bool
+) -> Result<&'static str, error::Error> {
+	if is_external && is_validator {
+		return Err(error::Error::Input("--rpc-external and --ws-external options shouldn't be \
+		used if the node is running as vaidator.".to_owned()));
```
tomusdrw:
```suggestion
		used if the node is running as validator.".to_owned()));
```

client/cli/src/lib.rs
```diff
@@ -931,6 +931,28 @@ where
 	Ok(config)
 }
 
+fn interface_str(
+	is_external: bool,
+	is_unsafe_external: bool,
+	is_validator: bool
+) -> Result<&'static str, error::Error> {
+	if is_external && is_validator {
```
tomusdrw:
```suggestion
	if is_validator && !is_unsafe_external {
```

When you run a dev chain you probably don't care about the risk at all, it should be allowed to use `unsafe` flag to override this.

client/cli/src/lib.rs
```diff
@@ -931,6 +931,28 @@ where
...
+		return Err(error::Error::Input("--rpc-external and --ws-external options shouldn't be \
+		used if the node is running as vaidator.".to_owned()));
+	}
+
+	if is_external {
+		log::warn!("It isn't safe to expose RPC publicly without a proxy server. \
```
tomusdrw:
```suggestion
		log::warn!("It isn't safe to expose RPC publicly without a proxy server that filters available set of RPC methods. \
```

client/cli/src/lib.rs
```diff
@@ -931,6 +931,28 @@ where
...
+		Use `--unsafe-rpc-external` to suppress the warning if you understand the risks.");
```
tomusdrw: To be fair I wouldn't give an option to suppress the warning, I think `unsafe` should just be used to be able to run the node even if you are a validator. I'd keep the warning whenever you run with `is_external || is_unsafe_external`.

If you agree we should move this message to the validator error instead and update the condition above.

client/cli/src/lib.rs
HCastano:
```suggestion
		used if the node is running as validator.".to_owned()));
```

client/cli/src/params.rs
```diff
@@ -438,16 +438,30 @@ pub struct RunCmd {
...
+	/// Listen to all RPC interfaces.
+	///
+	/// Same as `--rpc-external`.
+	#[structopt(long = "unsafe-rpc-external")]
+	pub unsafe_rpc_external: bool,
+
 	/// Listen to all Websocket interfaces.
 	///
-	/// Default is local.
+	/// Default is local. Note: it isn't safe to expose RPC publicly without a proxy server.
```
tomusdrw:
```suggestion
	/// Default is local. Note: not all RPC methods are safe to be exposed publicly. Use a RPC proxy server to filter out dangerous methods. More details: (link to wiki).
```

Does anyone know if we have node configuration documentation somewhere? CC @shawntabrizi ?

client/cli/src/lib.rs
```diff
@@ -931,6 +931,28 @@ where
 	Ok(config)
 }
 
+fn interface_str(
+	is_external: bool,
+	is_unsafe_external: bool,
+	is_validator: bool
```
tomusdrw:
```suggestion
	is_validator: bool,
```

DarkEld3r:
In progress till I figure which link to the documentation should be used.

client/cli/src/params.rs
DarkEld3r: Unfortunately, it looks like we don't have such documentation.

Issue #4040, opened 2019-11-07
construct_runtime macro needs to generate runtime Error type


xlc:
It needs to generate runtime Error type similar to how it generate runtime Event type so different module error can all `Into<T::Error>` where `type Error: From<Error> + From<system::Error> + any_other_used_module::Error`

Otherwise `decl_error` is much less useful

Follow up of #2880 #3433

bkchr:
Why? Where do we need generic errors? Please provide an example.

xlc:
This code https://github.com/laminar-protocol/open-runtime-module-library/blob/master/tokens/src/lib.rs

if I add `type Error = Error` in `decl_module`, the dispatchable call return type will be `DispatchResult<Error>`. but `ensure_signed` returns `system::Error`, which will stops me doing `let from = ensure_signed(origin)?;` because `Error` can't convert to `system::Error`.

So I would like to add `type Error: From<Error> + From<system::Error> + From<LookupError?` to my module `Trait` and make the module error type `type Error = T::Error`. Now it will be able to return any of the supported error in my dispatchable calls.

The remaining problem will be construct a type satisfies the type bonds for `T::Error`. It should be generated by `construct_runtime` (or `impl_outer_error`).

thiolliere:
My feeling is that an error is different from an event, if you call let's say staking module and the error received is balance module saying balance too low, you may not be able to know what triggered this error, staking module must put meanings on what has triggered this too low balance.

On the other hand an event is meaningfull by itself, however it has been triggered this event happened on chain.

thiolliere:
considering this discussion https://github.com/paritytech/substrate/issues/4385#issuecomment-565434276
I can finally understand that error could be treated like an event, the user would just have an "event" related to what error happened and to further dig into that they would use different tools

xlc:
If we don't have the limitation of onchain storage, we can totally just use a popular Rust error handling library and good to go. But with the constraints, we just can't retain all the details.

What we can do is maybe come out some macro that switch between the lite error code or with full error stack trace based on std feature. But this should happen after we have the lite error code version fully implemented.

xlc:
Fixed by #4449

PR #4449, opened 2019-12-18
Make `decl_error!` errors usable

bkchr:
This pr implements support for returning errors of different pallets in
a pallet. These errors need to be declared with `decl_error!`.

The pr changes the following:

- Each dispatchable function now returns a `DispatchResult` which is an
alias for `Result<(), DispatchError>`.
- `DispatchError` is an enum that has 4 variants:
  - `Other`: For storing string error messages
  - `CannotLookup`: Variant that is returned when something returns a
  `sp_runtime::LookupError`
  - `BadOrigin`: Variant that is returned for any kind of bad origin
  - `Module`: The error of a specific module. Contains the `index`,
  `error` and the `message`. The index is the index of the module in
  `construct_runtime!`. `error` is the index of the error in the error
  enum declared by `decl_error!`. `message` is the message to the error
  variant (this will not be encoded).
- `construct_runtime!` now creates a new struct `ModuleToIndex`. This
struct implements the trait `ModuleToIndex`.
- `frame_system::Trait` has a new associated type: `ModuleToIndex` that
expects the `ModuleToIndex` generated by `construct_runtime!`.
- All error strings returned in any module are being converted now to `DispatchError`.
- `BadOrigin` is the default error returned by any type that implements `EnsureOrigin`.

Fixes: https://github.com/paritytech/substrate/issues/4409 https://github.com/paritytech/substrate/issues/4040

@jacogr this changes the encoding of the `DispatchError`. Do you want to have a bump of the metadata to detect this?


frame/balances/src/lib.rs
```diff
@@ -891,11 +891,11 @@ where
 		_amount: T::Balance,
 		reasons: WithdrawReasons,
 		new_balance: T::Balance,
-	) -> Result {
+	) -> DispatchResult {
 		if reasons.intersects(WithdrawReason::Reserve | WithdrawReason::Transfer)
 			&& Self::vesting_balance(who) > new_balance
 		{
-			return Err("vesting balance too high to send value");
+			Err("vesting balance too high to send value")?
```
niklasad1: personally I prefer `return Err("fooo".into())` because it more explicit than `Err("foo)?` but it's fine either way.

PR #4461, opened 2019-12-19
babe: remove unused slot_duration variable

andresilva:
Duplicate definition some lines above.


andresilva:
Ignoring the `check_polkadot` CI result as it must be unrelated to this PR.

PR #4462, opened 2019-12-19
Extend Utility pallet with multisig and pseudonyms

### LINKED ISSUE #4442: Origin rerouting utilities
Utility crate should be extended with two new dispatchable calls:

```
fn forward_sub(origin /*signed*/, sub_index: u32, call: Call) {
  // calls `call` from signed origin given by address blake2("forward_sub" ++ signed ++ sub_index) 
}
fn new_threshold(origin /*signed*/, sub_index: u32, threshold: u32, addresses: Vec<AccountId>) {
  // creates/registers a new multisig threshold signature. this will come from address blake2("forward_sub" ++ signed ++ sub_index)
}
fn approve_threshold(origin /*signed*/, from: AccountId, call_hash: Hash) {
  // approve a call happening from the multisig `from`
}
fn execute_threshold(origin /*signed*/, from: AccountId, call: Call) {
  // (approve and) execute a call from the multisig `from`. it must already have at least threshold-1 approvals.
}
```

gavofyork:
Adds three new new pieces of functionality to the Utility pallet:
- `fn as_sub`, allowing any account to change sender to some derivative account ID. The set of derivative addresses are formed as the Blake2 hash of the owner address and index, prefixed with a context string.
- `fn as_multi`, `fn approve_as_multi`, `fn cancel_as_multi`, allowing any set of accounts to proxy a message from a well known account ID, once a threshold of approvals is met.
- Alters the existing batch functionality so that it works with any origin.

This also cleans up the weights.

TODO: 
- [x] Tests for `Timepoint` stuff
- [x] Weights for everything
- [x] Docs (including complexity)
- [x] Module-level docs
- [x] Maybe refactor into `structs`
- [x] Batch should work with signed origins
- [x] Consider splitting deposit into base and threshold multiplier

Closes #4442

### DIFF
--- a/bin/node/runtime/src/lib.rs
+++ b/bin/node/runtime/src/lib.rs
@@ -128,9 +128,18 @@ impl frame_system::Trait for Runtime {
 	type ModuleToIndex = ();
 }
 
+parameter_types! {
+	// One storage item, up to 584 bytes.
+	pub const MultisigDeposit: Balance = 1 * DOLLARS;
+	pub const MaxSignatories: u16 = 16;
+}
+
 impl pallet_utility::Trait for Runtime {
 	type Event = Event;
 	type Call = Call;
+	type Currency = Balances;
+	type MultisigDeposit = MultisigDeposit;
+	type MaxSignatories = MaxSignatories;
 }
 
 parameter_types! {
@@ -515,7 +524,7 @@ construct_runtime!(
 		UncheckedExtrinsic = UncheckedExtrinsic
 	{
 		System: frame_system::{Module, Call, Storage, Config, Event},
-		Utility: pallet_utility::{Module, Call, Event},
+		Utility: pallet_utility::{Module, Call, Storage, Event<T>, Error},
 		Babe: pallet_babe::{Module, Call, Storage, Config, Inherent(Timestamp)},
 		Timestamp: pallet_timestamp::{Module, Call, Storage, Inherent},
 		Authorship: pallet_authorship::{Module, Call, Storage, Inherent},
--- a/bin/node/runtime/src/lib.rs
+++ b/bin/node/runtime/src/lib.rs
@@ -129,7 +129,7 @@ impl frame_system::Trait for Runtime {
 }
 
 parameter_types! {
-	// One storage item; value is size 4+4+8+32 bytes = 48 bytes.
+	// One storage item; value is size 4+4+16+32 bytes = 56 bytes.
 	pub const MultisigDepositBase: Balance = 30 * CENTS;
 	// Additional storage item size of 32 bytes.
 	pub const MultisigDepositFactor: Balance = 5 * CENTS;
--- a/bin/node/runtime/src/lib.rs
+++ b/bin/node/runtime/src/lib.rs
@@ -140,7 +140,8 @@ impl pallet_utility::Trait for Runtime {
 	type Event = Event;
 	type Call = Call;
 	type Currency = Balances;
-	type MultisigDeposit = MultisigDeposit;
+	type MultisigDepositBase = MultisigDepositBase;
+	type MultisigDepositFactor = MultisigDepositFactor;
 	type MaxSignatories = MaxSignatories;
 }
 
--- a/bin/node/runtime/src/lib.rs
+++ b/bin/node/runtime/src/lib.rs
@@ -129,9 +129,11 @@ impl frame_system::Trait for Runtime {
 }
 
 parameter_types! {
-	// One storage item, up to 584 bytes.
-	pub const MultisigDeposit: Balance = 1 * DOLLARS;
-	pub const MaxSignatories: u16 = 16;
+	// One storage item; value is size 4+4+8+32 bytes = 48 bytes.
+	pub const MultisigDepositBase: Balance = 30 * CENTS;
+	// Additional storage item size of 32 bytes.
+	pub const MultisigDepositFactor: Balance = 5 * CENTS;
+	pub const MaxSignatories: u16 = 100;
 }
 
 impl pallet_utility::Trait for Runtime {
--- a/frame/utility/src/lib.rs
+++ b/frame/utility/src/lib.rs
@@ -22,14 +22,15 @@
 
 use sp_std::prelude::*;
 use codec::{Encode, Decode};
-use sp_core::TypeId;
-use sp_core::blake2_256;
-use frame_support::{decl_module, decl_event, decl_storage, Parameter};
-use frame_support::weights::{
+use sp_core::{TypeId, blake2_256};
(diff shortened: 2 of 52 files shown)


frame/utility/src/lib.rs
```diff
@@ -15,48 +15,620 @@
...
+						<Multisigs<T>>::insert(&id, call_hash, m);
+						Self::deposit_event(RawEvent::MultisigApproval(who, timepoint, id));
+						return Ok(())
+					}
+				} else {
+					if (m.approvals.len() as u16) < threshold {
+						Err(Error::<T>::AlreadyApproved)?
+					}
+				}
+				call.dispatch(frame_system::RawOrigin::Signed(id.clone()).into())?;
```
pepyakin: Due to `?` it is possible to execute the call multiple times if it returns `Err`. This might be not desirable since it is not guaranteed nor required for a call to be side-effect free if it returns `Err`.

frame/utility/src/lib.rs
```diff
@@ -15,48 +15,620 @@
...
+			Ok(())
+		}
+	}
+}
+
+impl<T: Trait> Module<T> {
+	/// Derive a sub-account ID from the owner account and the sub-account index.
+	pub fn sub_account_id(who: T::AccountId, index: u16) -> T::AccountId {
+		let entropy = (b"modlpy/utilisuba", who, index).using_encoded(blake2_256);
+		T::AccountId::decode(&mut &entropy[..]).unwrap_or_default()
```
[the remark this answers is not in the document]
gavofyork: it assumes that it's no larger than a hash, but doesn't need to be as big, as the tests show. i think that this is fine. other parts of the codebase also assume that `AccountId`s can be decoded from unformatted data.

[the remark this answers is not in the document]
pepyakin: True. I was more concerned about that `T::AccountId` can be anything (theoretically) and it might be that not all variations of `entropy` are suitable for decoding.

If the other parts do assume that `AccountId` can be decoded from unformatted data then we might want to embrace that and document it explicitly. (Not to suggest doing this in this PR though).

frame/utility/src/lib.rs
```diff
@@ -15,48 +15,620 @@
...
+	/// The account who opened it (i.e. the first to approve it).
+	depositor: AccountId,
+	/// The approvals achieved so far, including the depositor.
+	approvals: Vec<AccountId>,
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as Utility {
+		/// The set of open multisig operations.
+		pub Multisigs: double_map hasher(twox_64_concat) T::AccountId, twox_64_concat([u8; 32])
```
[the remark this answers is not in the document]
gavofyork: yeah - good catch.

frame/utility/src/lib.rs
gavofyork: it is not guaranteed, but it is good practice.

gavofyork:
@pepyakin other points should be addressed in the latest changes.

frame/utility/src/lib.rs
```diff
@@ -15,65 +15,645 @@
...
-decl_event!(
+decl_event! {
 	/// Events type.
-	pub enum Event {
-		BatchExecuted(Vec<Result<(), DispatchError>>),
+	pub enum Event<T> where
+		AccountId = <T as system::Trait>::AccountId,
+		BlockNumber = <T as system::Trait>::BlockNumber
+	{
+		/// Batch of dispatches did not complete fully. Index of first failin dispatch given, as
```
shawntabrizi:
```suggestion
		/// Batch of dispatches did not complete fully. Index of first failing dispatch given, as
```

frame/utility/src/lib.rs
```diff
@@ -15,65 +15,645 @@
...
+		/// A new multisig operation has begun. First param is the account that is approving,
+		/// second is the multisig account.
+		NewMultisig(AccountId, AccountId),
+		/// A multisig operation has been approved by someone. First param is the account that is
+		/// approving, third is the multisig account.
+		MultisigApproval(AccountId, Timepoint<BlockNumber>, AccountId),
+		/// A multisig operation has been executed. First param is the account that is
+		/// approving, third is the multisig account.
+		MultisigExecuted(AccountId, Timepoint<BlockNumber>, AccountId, DispatchResult),
+		/// A multisig operation has been executed. First param is the account that is
```
shawntabrizi:
```suggestion
		/// A multisig operation has been cancelled. First param is the account that is
```

frame/utility/src/lib.rs
```diff
@@ -15,65 +15,645 @@
...
-	type Call: Parameter + Dispatchable<Origin=Self::Origin>;
+	type Call: Parameter + Dispatchable<Origin=Self::Origin> + GetDispatchInfo;
+
+	/// The currency mechanism.
+	type Currency: ReservableCurrency<Self::AccountId>;
+
+	/// The base amount of currency needed to reserve for creating a multisig execution.
+	///
+	/// This is held for an additional storage item whose value size is
+	/// `4 + sizeof((BlockNumber, Balance, AccountId))` bytes.
```
pepyakin:
```suggestion
	/// `4 + sizeof((BlockNumber, u32, Balance, AccountId))` bytes.
```

bin/node/runtime/src/lib.rs
```diff
@@ -128,9 +128,21 @@ impl frame_system::Trait for Runtime {
 	type ModuleToIndex = ();
 }
 
+parameter_types! {
+	// One storage item; value is size 4+4+8+32 bytes = 48 bytes.
```
pepyakin: I assume this is in the following order: height, index, deposit, depositor?
But isn't deposit is of `Balance` type which is defined as `u128`?

```suggestion
	// One storage item; value is size 4+4+16+32 bytes = 56 bytes.
```

frame/utility/src/lib.rs
```diff
@@ -15,65 +15,645 @@
...
+						Self::deposit_event(RawEvent::MultisigApproval(who, timepoint, id));
+						return Ok(())
+					}
+				} else {
+					if (m.approvals.len() as u16) < threshold {
+						Err(Error::<T>::AlreadyApproved)?
+					}
+				}
+
+				let result = call.dispatch(frame_system::RawOrigin::Signed(id.clone()).into());
```
shawntabrizi:
```suggestion
				let result = call.dispatch(origin.clone());
```

frame/utility/src/lib.rs
gavofyork: that's why it's `4 + ....` :-)

pepyakin: 🤦‍♂

PR #4466, opened 2019-12-20
Shared params in CLI API

arkpar:
Exposed common shared params in the CLI API and made some helper type aliases public in `service`.

This is required for https://github.com/paritytech/polkadot/pull/704


gavofyork:
tests failing

PR #4470, opened 2019-12-20
Clean up definition for custom ss58 address formats

sorpaas:
This cleans up definition of `Ss58AddressFormat`, ensuring that all custom ones are only present in the macro invocation, and nowhere else.


bkchr:
DOes not compile.

sorpaas:
@bkchr Fixed!

PR #4471, opened 2019-12-20
Rename ChainSpec field

JoshOrndorff:
This PR changes the names of fields and structs related to chain specifications. No logic has changed.

When [drafting docs on chain specification](https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/377) I found the field names in the `ChainSpec` struct confusing. Specifically I didn't like that the `ChainSpec` has two parts, `spec` and `genesis`.

The primary change introduced here is:
```diff
/// A configuration of a chain. Can be used to build a genesis block.
pub struct ChainSpec<G, E = NoExtension> {
-	spec: ChainSpecFile<E>,
+	client_spec: ClientSpec<E>,
	genesis: GenesisSource<G>,
}
```

Questions:
* Is such a naming change advisable at all?
* Are the names I've chosen appropriate?


client/chain-spec/src/chain_spec.rs
```diff
@@ -132,7 +132,7 @@ enum Genesis<G> {
 #[derive(Serialize, Deserialize, Clone, Debug)]
 #[serde(rename_all = "camelCase")]
 #[serde(deny_unknown_fields)]
-struct ChainSpecFile<E> {
+struct ClientSpec<E> {
```
bkchr: Maybe a comment would be nice as well here :)

JoshOrndorff:
You men it breaks compatibility with specs that are already stored as json? I don't think so. Consider this spec from the node template:

```json
{
  "name": "Local Testnet",
  "id": "local_testnet",
  "bootNodes": [
    "/ip4/127.0.0.1/tcp/30333/p2p/Qmaeaa2WT2oSGGUxpNEk2SpBbY2zPaH8UYSUDkMjdbrRAc"
  ],
  "telemetryEndpoints": null,
  "protocolId": null,
  "properties": null,
  "consensusEngine": null,
  "genesis": {
    ...
  }
}
```
It doesn't contain the name `spec` at all.

bkchr:
Yeah serialization is done here: https://github.com/paritytech/substrate/blob/a8e03411620f652b620930ec273edd782c8df711/client/chain-spec/src/chain_spec.rs#L277

It is flattened, so no problem here :)

Issue #4407, opened 2019-12-17
Lazy payouts


gavofyork:
Currently, all payouts from an era are made actively at the end of the era. This is a big task and will only get worse as time goes on and the validator/nominator sets grow. Right now it takes so long that it's stalling the Kusama chain (#4313). It is also problematic as the immediacy conflicts with the fact that the session keys lag one session behind the actual staking selection, leading to the wrong validators being paid for one session in every era where the validation set changes (#4331, #4346 and #4220).

Instead, payouts should happen lazily/passively.

The Session module should, at the end of a session, report the accrued points of each validator together with the era number that the validator set is from back up to the Staking module, giving it full information about which validators should be rewarded. (It should explicitly report so that the number of sessions that the validator set lags behind continues to be immaterial for the Staking module.)

The Staking module should should retain a `last_payout_era` value for all staked accounts, and should keep a number of past instances of the exposure information on-chain (e.g. last 100 eras) as well as the historical validator points for each session.

There should then be an explicit payout function which pays out any balance owed, taking into account all of the above information. Stakers would need to call it occasionally (at least once every  100 eras =~ 2 weeks on Kusama) to avoid missing out on their rewards.

This will fix #4121, #4331, and #4220 as well as help with #4313 . This fix will supersede #4346.

gavofyork:
CC @thiolliere

joepetrowski:
> Stakers would need to call it occasionally

Do you expect nominators to have to take any action, or do you only refer to validators here? I read this as validators only (triggering a payment to all their nominators) because:

1. Nomination should be relatively frictionless once you've made the nomination and you're already putting a lot of trust into the validators you nominate. Trusting them to call this function isn't a big stretch.
2. Letting nominators call this would require calling this function on every validator that is backed by the nominator and keeping tabs on when validators paid each nominator.

That said, perhaps nominators should have a final appeal? E.g. if the limit is 100 eras, then a nominator can report a single validator after 90 eras and claim their reward share from that validator.

rphmeier:
Doing payouts lazily/passively is that it conflicts with rewards accruing onto the stash's bond. If we do it lazily, it would allow validators to dodge part of slashes.

gavofyork:
i don't think it will matter in practise. the maximum amount of funds that would not be slashed will be around 4% of the annual (2 weeks/52 weeks), or, assuming 10% reward, 0.4% of your stake. that won't make much of a difference in terms of incentivisation.

[...]

Issue #4411, opened 2019-12-17
Fix inflation calculation to account for total stake


joepetrowski:
Currently ([source](https://github.com/paritytech/substrate/blob/40b1e530870cd5c1803fdd39b6386dcaa6a770fb/frame/staking/src/lib.rs#L1341L1350)), the staking rewards calculation calculates the total amount at stake by multiplying the `SlotStake` - the stake of the minimally staked validator - by the number of validators.

The actual amount at stake - if we consider any funds liable to be slashed as "at stake" - then the number of tokens at stake is higher since most validators will have more than the slot stake.

The calculation should find the sum of the amount at stake (sum of `exposure.total` from [`Stakers`](https://github.com/paritytech/substrate/blob/40b1e530870cd5c1803fdd39b6386dcaa6a770fb/frame/staking/src/lib.rs#L672)) and pass that to the payout function.

Related: https://github.com/paritytech/substrate/issues/4407

```
let validator_len: BalanceOf<T> = (validators.len() as u32).into();
let total_rewarded_stake = Self::slot_stake() * validator_len;

let (total_payout, max_payout) = inflation::compute_total_payout(
	&T::RewardCurve::get(),
	total_rewarded_stake.clone(),
	T::Currency::total_issuance(),
	// Duration of era; more than u64::MAX is rewarded as u64::MAX.
	era_duration.saturated_into::<u64>(),
);
```

joepetrowski:
After talking with @rphmeier, we may just want to use the median amount at stake instead of the sum. Either way, we should not use the minimum as it may be highly variable era-to-era.

rphmeier:
The research.web3.foundation site should describe this, if it doesn't already. Least-staked is OK, but bottom-tercile is probably the ideal metric

thiolliere:
This was fixed in lazy payouts PR

PR #4474, opened 2019-12-21
Lazy payouts

gavofyork:
Closes: #4407 

TODO: status of PR updated just below


gavofyork:
CC @thiolliere

frame/staking/src/lib.rs
```diff
@@ -668,8 +675,17 @@ decl_storage! {
 		/// Nominators for a particular account that is in action right now. You can't iterate
 		/// through validators here, but you can find them in the Session module.
 		///
-		/// This is keyed by the stash account.
-		pub Stakers get(fn stakers): map T::AccountId => Exposure<T::AccountId, BalanceOf<T>>;
+		/// This is keyed fist by the era index to allow bulk deletion and then the stash account.
+		///
+		/// Is it removed after `HISTORY_DEPTH` eras.
+		pub Stakers get(fn stakers): double_map EraIndex hasher(twox_64_concat),
+			twox_64_concat(T::AccountId) => Exposure<T::AccountId, BalanceOf<T>>;
```
thiolliere: I was wondering if having a `Vec<(T::Account, Exposure<..>)>` for historical stakers exposure was not better here, but iterating on prefix could be already efficient enough.

thiolliere:
OK, I see, seems future proof for when the number of staker will be bigger, also this decouple the session delay and staking logic (only constrain is having HISTORICAL_DEPTH > delay).

I can work on that now if wanted.

thiolliere:
> (It should explicitly report so that the number of sessions that the validator set lags behind continues to be immaterial for the Staking module.)

note: this is difficult to do because we have to compute the duration of the era and tell when the era is fully rewarded and we can make payments for it.

thiolliere:
Here is an implementation, because we need to get the length of the active era to compute its reward staking module still have to follow session endings and the era of current validator, such era is named active_era in the current code.

Actually this implementation ends up quite similar to https://github.com/paritytech/substrate/pull/4346

### notes:
* active_era is the era of the current validator set,
* current_era is the latest planned era.
* slot stakes has been removed as it is not used.
* I also fixed https://github.com/paritytech/substrate/issues/4411
* (also we probably don't want to have staking 2.0 before all those issue are fixed)

### todo
* port the test similarly to #4346 
* write migration
* rethink storage organisation for efficiency

thiolliere:
this is a draft that should make the API for session more explicit https://github.com/paritytech/substrate/pull/4609

thiolliere:
status of the PR updated https://github.com/paritytech/substrate/pull/4474#issuecomment-571227349

frame/staking/src/lib.rs
```diff
@@ -1237,6 +1312,42 @@ decl_module! {
...
+		fn payout_nominator(origin, era: EraIndex, validators: Vec<T::AccountId>)
+			-> DispatchResult
+		{
+			let who = ensure_signed(origin)?;
+			Self::do_payout_nominator(who, era, validators)
+		}
+
+		/// Make one staker's payout for one era.
+		///
+		/// - `who` is the nominator to pay out.
```
joepetrowski: Could you specify (for this and `payout_nominator`) that this function should be called by the _Controller?_
```suggestion
		/// - `who` is the validator to pay out.
```

thiolliere: yes definitely, done

frame/staking/src/lib.rs
```diff
@@ -688,33 +695,50 @@ decl_storage! {
...
-		CurrentEraPointsEarned get(fn current_era_reward): EraPoints;
+		/// Exposure of validator at era.
+		///
+		/// This is keyed fist by the era index to allow bulk deletion and then the stash account.
+		///
+		/// Is it removed after `HISTORY_DEPTH` eras.
+		pub ErasStakers get(fn eras_stakers):
+			double_map hasher(twox_64_concat) EraIndex, hasher(twox_64_concat) T::AccountId
+			=> Exposure<T::AccountId, BalanceOf<T>>;
+		// TODO: consider switching this to a simple map EraIndex => Vec<Exposure>
```
thiolliere: We can do this in a future PR, also I feel like it is hard to benchmark things currently in substrate

frame/staking/src/lib.rs
```diff
@@ -1236,6 +1294,57 @@ decl_module! {
...
+		/// # <weight>
+		/// - Number of storage read of `O(validators)`; `validators` is the argument of the call.
+		/// - Each storage read is `O(N)` size and decode complexity; `N` is the  maximum
+		///   nominations that can be given to a single validator.
+		/// - Computation complexity: `O(MAX_NOMINATIONS * logN)`; `MAX_NOMINATIONS` is the
+		///   maximum number of validators that may be nominated by a single nominator, it is
+		///   bounded only economically (all nominators are required to place a minimum stake).
+		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedNormal(500_000)]
+		// TODO: Limit the amount of nominators that can be assigned to a validator by Phragmen.
```
thiolliere: should we also limit this in the current PR ?

thiolliere: Also if decoding the structure Exposure which contains all nominations for the validator is to costy during this reward we could try to store on chain the nominations of a nominator so it only query one structure for himself and not one structure for each validator it has nominated

frame/staking/src/lib.rs
```diff
@@ -1551,31 +1718,33 @@ impl<T: Trait> Module<T> {
...
-			// Set the new validator set in sessions.
-			<CurrentElected<T>>::put(&elected_stashes);
+			// Insert current era informations
+			<ErasTotalStake<T>>::insert(&current_era, total_staked);
+			let default_pref = ValidatorPrefs::default();
+			for stash in &elected_stashes {
+				let pref = all_validators_and_prefs.get(stash)
+					// This should always succeed but better to be safe
+					.unwrap_or(&default_pref);
+				<ErasValidatorPrefs<T>>::insert(&current_era, stash, pref);
```
thiolliere: Here we can write the ErasValidatorPrefs for the current_era + 1 so nominators can nominate while knowing the preference in advance.

frame/staking/src/lib.rs
thiolliere: Note that having this structure as a `Vec<Exposure>` means decoding all validator exposure (which can contains a lot of nominators) when rewarding a single nominator and validator.
So instead of doing at most 16 (MAX_NOMINATION) exposure read we always do 1 read of all exposure.

frame/staking/src/lib.rs
gavofyork: that would likely take a lot of state and i/o at the phragmen time, so far from ideal. better to just limit the number of nominators at the phragmen stage.

gavofyork: we can introduce per-validator indices to denote which exposure index our nominator is for each validator: switch the parameter here to `fn payout_nominator(origin, era: EraIndex, validators: Vec<(T::AccountId, u32)>)` where the `u32` is the index in to the validator's exposure list.

we should still limit the number of nominators allowed for each validator to something sensible (e.g. 64) just to avoid a quadratic attack on decoding (e.g. a validator with 10,000 tiny nominators, thenn calling this function for each one would result in `O(10,000**2)` with only 10,000 dispatches paid.

frame/staking/src/lib.rs
```diff
@@ -1268,6 +1378,93 @@ impl<T: Trait> Module<T> {
...
+		}
+
+		nominator_ledger.next_reward = era + 1;
+		<Ledger<T>>::insert(&who, &nominator_ledger);
+
+		let mut reward = Perbill::zero();
+		let era_reward_points = <ErasRewardPoints<T>>::get(&era);
+		for validator in validators {
+			let commission = Self::eras_validator_prefs(&era, &validator).commission;
+			let validator_exposure = <ErasStakers<T>>::get(&era, &validator);
```
gavofyork: currently unbounded (phragmen can give a validator any number of nominators). that will need to be fixed.

frame/staking/src/migration.rs
```diff
@@ -111,6 +114,92 @@ pub mod inner {
...
+	//   * ErasStartSessionIndex
+	// * translate StakingLedger
+	// * removal of:
+	//   * Stakers
+	//   * SlotStake
+	//   * CurrentElected
+	//   * CurrentEraStart
+	//   * CurrentEraStartSessionIndex
+	//   * CurrentEraPointsEarned
+	pub fn to_v3<T: Trait>(version: &mut VersionNumber) {
```
gavofyork: is this code tested anywhere?

thiolliere: yes here but I agree it is not sure this take into account all state kusama can be in. https://github.com/paritytech/substrate/pull/4474/files#diff-6031f6ad0d6b579272b1748548a364ddR2863

frame/staking/src/lib.rs
```diff
@@ -1715,7 +1875,7 @@ impl<T: Trait> Convert<T::AccountId, Option<Exposure<T::AccountId, BalanceOf<T>>
 	for ExposureOf<T>
 {
 	fn convert(validator: T::AccountId) -> Option<Exposure<T::AccountId, BalanceOf<T>>> {
-		Some(<Module<T>>::stakers(&validator))
+		Some(<Module<T>>::eras_stakers(<Module<T>>::current_era(), &validator))
```
thiolliere: shouldn't it be active_era instead of current era ?

it is used by historical session (at some point it shouldn't be necessary as historical session as all validator set exposure and know which one is the current one. But this can be done in a later PR).

historical session states:
> 	/// A conversion from validator ID to full identification.
	///
	/// This should contain any references to economic actors associated with the
	/// validator, since they may be outdated by the time this is queried from a
	/// historical trie.
	///
	/// It must return the identification for the current session index.
	type FullIdentificationOf: Convert<Self::ValidatorId, Option<Self::FullIdentification>>;

frame/staking/src/lib.rs
thiolliere: note that if the unbounded issue is only for nominator reward we can store this information also per-nominator so it doesn't query this unbounded exposure for each validator but only one structure which holds all its information.

frame/staking/src/lib.rs
```diff
@@ -1366,66 +1543,60 @@ impl<T: Trait> Module<T> {
...
-				&T::RewardCurve::get(),
-				total_rewarded_stake.clone(),
-				T::Currency::total_issuance(),
-				// Duration of era; more than u64::MAX is rewarded as u64::MAX.
-				era_duration.saturated_into::<u64>(),
-			);
-
-			let mut total_imbalance = <PositiveImbalanceOf<T>>::zero();
+	fn end_session(session_index: SessionIndex) {
+		if ErasStartSessionIndex::get(Self::active_era() + 1) == session_index + 1 {
```
thiolliere: one tricky aspect of this condition is that if somehow one session is missed by session pallet then we are stuck in the current era forever.

thiolliere: this could be fixed by some emergency situation like this
```diff
diff --git a/frame/staking/src/lib.rs b/frame/staking/src/lib.rs
index 0c8b2320d..4bd052da3 100644
--- a/frame/staking/src/lib.rs
+++ b/frame/staking/src/lib.rs
@@ -1553,7 +1553,16 @@ impl<T: Trait> Module<T> {
        }
 
        fn end_session(session_index: SessionIndex) {
-               if ErasStartSessionIndex::get(Self::active_era() + 1) == session_index + 1 {
+               let active_era = Self::active_era();
+               let current_era = Self::current_era();
+               let next_era_start = ErasStartSessionIndex::get(active_era + 1);
+
+               // In case the next_era_start has been missed and there is some era planned then we end the
+               // current era.
+               // This must never happen
+               let emergency = next_era_start < session_index + 1 && active_era < current_era;
+
+               if next_era_start == session_index + 1 || emergency {
                        Self::end_era(session_index);
                }
        }
```

frame/staking/src/lib.rs
```diff
@@ -1655,17 +1837,27 @@ impl<T: Trait> Module<T> {
 	}
 }
 
+/// In this implementation `new_session(session)` must be called before `end_session(session-1)`
+/// i.e. the new session must be planned before the ending of the previous session.
+///
+/// `new_session(0)` must be called at genesis.
```
thiolliere: I think we should support less constraint, allow staking module to be plugged in a runtime that already have started. And also that the start of the era is not necessarily the first block when staking module has been put in the runtime.

thiolliere: Currently the staking module is required to be run from genesis. I think this is a constraint we don't want. I'm working on this.

frame/staking/src/lib.rs
```diff
@@ -1747,13 +1995,23 @@ impl <T: Trait> OnOffenceHandler<T::AccountId, pallet_session::historical::Ident
 
 		let reward_proportion = SlashRewardFraction::get();
 
-		let era_now = Self::current_era();
-		let window_start = era_now.saturating_sub(T::BondingDuration::get());
-		let current_era_start_session = CurrentEraStartSessionIndex::get();
+		let current_era = {
```
thiolliere: what is supposed to be this era_now ? the current era or the active era ?

frame/staking/src/lib.rs
thiolliere: This is fixed by introducing ErasStakersClipped which hold only the N biggest nominators for each stakers, such clipped exposure is only used for reward.

frame/staking/src/lib.rs
thiolliere: number of nominators is fixed for nominator reward using exposureclipped principle.

payout_nominator using nominator in validator exposure is done.

frame/staking/src/migration.rs
gavofyork: there are two other tests:
- use a test environment in `master` in order to create a toy state, then dump the keys in externalities; then make a test in this branch whose environment contains only those keys and check to make sure the upgrade completes correctly.
- dumping the state of kusama (there's an rpc `state_getPairs(prefix)` to make this easy), then place it in a dev chain's genesis and test to make sure it completes successfully.

Noc2:
Bruno just shared this PR on our technical channel. A question/idea I have: why not claim rewards automatically every time a user makes a specific transaction (e.g. a transfer). This might be more userfriendly, but maybe I'm missing something. In this case, just ignore my comment ;-)

joepetrowski:
> Bruno just shared this PR on our technical channel. A question/idea I have: why not claim rewards automatically every time a user makes a specific transaction (e.g. a transfer). This might be more userfriendly, but maybe I'm missing something. In this case, just ignore my comment ;-)

This only needs to be done max once per era and these functions read a lot of state. It would increase the transaction fees of other transactions because you'd have to (a) make `staking` depend on `balances` and (b) add additional logic like looking up if the account making a transfer is a Controller.

Noc2:
> This only needs to be done max once per era and these functions read a lot of state. It would increase the transaction fees of other transactions because you'd have to (a) make `staking` depend on `balances` and (b) add additional logic like looking up if the account making a transfer is a Controller.

Got it. You could, of course, integrate a last_payout_era check to call the function max once per era (this also makes it impossible to spam the network with it), but you still need to add some additional complexity. Also how about integrating it, for example, in the unbond function.

joepetrowski:
> Got it. You could, of course, integrate a last_payout_era check to call the function max once per era (this also makes it impossible to spam the network with it), but you still need to add some additional complexity. 

Yeah but you might want to call a function like transfer multiple times in an era.

> Also how about integrating it, for example, in the unbond function.

You should look at the code here... 
1. It only stores rewards for the last 84 eras.
2. If your destination is "Stash adding to stake", then waiting until unbond completely removes any compounding.
3. If I read the code correctly (@kianenigma , @thiolliere ?) then you have to call this once for every era. It doesn't give you a batch of eras' rewards.

thiolliere:
> If I read the code correctly then you have to call this once for every era. It doesn't give you a batch of eras' rewards.

yes

joepetrowski:
The original [issue](https://github.com/paritytech/substrate/issues/4407) for this states that rewards should accumulate in a storage item and be claimed in aggregate, but this implementation requires the claimer to submit a transaction for each era. This is quite inefficient, and fees could take a large amount of the reward for smaller stakers. What was the reason to change to this model? Besides the loss to fees, it seems quite inefficient to add 1 tx/staker/era to the system.

> There should then be an explicit payout function which pays out any balance owed, taking into account all of the above information. Stakers would need to call it occasionally (at least once every 100 eras =~ 2 weeks on Kusama) to avoid missing out on their rewards.

thiolliere:
> The original [issue](https://github.com/paritytech/substrate/issues/4407) for this states that rewards should accumulate in a storage item and be claimed in aggregate, but this implementation requires the claimer to submit a transaction for each era. This is quite inefficient, and fees could take a large amount of the reward for smaller stakers. What was the reason to change to this model? Besides the loss to fees, it seems quite inefficient to add 1 tx/staker/era to the system.
> 
> > There should then be an explicit payout function which pays out any balance owed, taking into account all of the above information. Stakers would need to call it occasionally (at least once every 100 eras =~ 2 weeks on Kusama) to avoid missing out on their rewards.

At the end of an era I just store the total payout of this era, we don't compute individual reward for each stakers as doing so would cost too much: computing the individual reward for each stakers imply reading all validator exposure and storing one value per staker which is almost same cost as actually rewarding all stakers.

I can change the extrinsic so that a stakers can ask for the reward of multiple era in the same extrinsic, but the length of this extrinsic will be proportional to the number of era asked.
a nominator will have to give an ordered vec of `era: EraIndex, validators: Vec<(T::AccountId, u32)>` and a validator will have to give a vec of era index (or maybe better a vec of range).
This bigger extrinsic will be a bit more efficient many individual indeed.

joepetrowski:
> I can change the extrinsic so that a stakers can ask for the reward of multiple era in the same extrinsic, but the length of this extrinsic will be proportional to the number of era asked.
a nominator will have to give an ordered vec of era: EraIndex, validators: Vec<(T::AccountId, u32)> and a validator will have to give a vec of era index (or maybe better a vec of range).
This bigger extrinsic will be a bit more efficient many individual indeed.

Yeah, this seems reasonable. To reduce the size of the extrinsic for nominators, I wonder if you could only accept a single vec of validators? Then automatically call `do_payout_nominator` every time someone submits a `nominate` extrinsic.

Then people would have the option of calling `payout_nominator` with a vec of their current set to get their payout without changing whom they nominate.

Not sure if this adds unacceptable complexity/risk, just wondering if we can reduce the size of the extrinsic to `(min_era_idx, max_era_idx, Vec<(T::AccountId, u32)>`.

frame/session/src/historical.rs
```diff
@@ -146,6 +147,9 @@ impl<T: Trait, I> crate::SessionManager<T::ValidatorId> for NoteHistoricalRoot<T
 
 		new_validators
 	}
+	fn start_session(end_index: SessionIndex) {
```
shawntabrizi:
```suggestion
	fn start_session(start_index: SessionIndex) {
```

Right?

frame/session/src/historical.rs
```diff
@@ -146,6 +147,9 @@ impl<T: Trait, I> crate::SessionManager<T::ValidatorId> for NoteHistoricalRoot<T
...
+		<I as SessionManager<_, _>>::start_session(end_index)
```
shawntabrizi:
```suggestion
		<I as SessionManager<_, _>>::start_session(start_index)
```

frame/session/src/historical.rs
```diff
@@ -108,6 +108,7 @@ pub trait SessionManager<ValidatorId, FullIdentification>: crate::SessionManager
 	/// If there was a validator set change, its returns the set of new validators along with their
 	/// full identifications.
 	fn new_session(new_index: SessionIndex) -> Option<Vec<(ValidatorId, FullIdentification)>>;
+	fn start_session(end_index: SessionIndex);
```
shawntabrizi:
```suggestion
	fn start_session(start_index: SessionIndex);
```

frame/session/src/historical.rs
thiolliere: yes right thanks

frame/staking/src/lib.rs
```diff
@@ -655,11 +679,14 @@ impl Default for Forcing {
 	fn default() -> Self { Forcing::NotForcing }
 }
 
+const HISTORY_DEPTH: EraIndex = 84;
```
shawntabrizi: Why is this not configurable in the Trait or as a storage item?

frame/staking/src/lib.rs
```diff
@@ -690,34 +718,73 @@ decl_storage! {
...
-		pub CurrentElected get(fn current_elected): Vec<T::AccountId>;
+		/// The active era index, the era currently rewarded.
+		///
+		/// Validator set of this era must be equal to `SessionInterface::validators`.
+		pub ActiveEra get(fn active_era): Option<EraIndex>;
 
-		/// The current era index.
-		pub CurrentEra get(fn current_era) config(): EraIndex;
+		/// The start of the active era.
+		pub ActiveEraStart get(fn active_era_start): Option<MomentOf<T>>;
```
shawntabrizi: These storage items seem to always be accessed together. We should combine them into a single item.

frame/staking/src/lib.rs
thiolliere: we can make it configurable in the Trait for no cost,
we can also make it configurable and changeable dynamically:
* either by adding a storage for previous removed era
* or by requiring some migration code when this value is changed

frame/staking/src/lib.rs
thiolliere: not in `fn convert`, `fn on_offence`, `start_session`, `end_session` but I guess grouping them doesn't harm indeed.

thiolliere:
> To reduce the size of the extrinsic for nominators, I wonder if you could only accept a single vec of validators? 

we could accept a single vec of validators but then we have to search the position of the nomination for each validator for each era on chain.
but we can still sort them and do a binary search there. This was what was done in a previous iteration of this PR see:
https://github.com/paritytech/substrate/pull/4474#discussion_r371859064

> Then automatically call do_payout_nominator every time someone submits a nominate extrinsic.

this is a tough suggestion: indeed we can just take before replacing it the previous nominations and ask for reward for all the era unrewarded yet with those validator, And that could be efficient enough but I feel like then the weight is very difficult to estimate. as it will correspond to all this parameters (difference between last reward era and active era, number of validator there, maybe most of those validator won't be actually in the validator set thus the weight estimation will be bigger than its actual cost).

frame/staking/src/lib.rs
shawntabrizi: You understand the code the best now. Do you suspect that governance will want to update this value? i.e. make a vote to increase or decrease the history depth?

If so, it should be a storage value.

Also depends on how often the value is accessed.

shawntabrizi: It is only accessed once per new era so a storage item totally makes sense. It seems the current logic may not be perfectly sound if the value is updated dynamically, i.e.

```
		// Clean old era information.
		if let Some(old_era) = current_era.checked_sub(HISTORY_DEPTH) {
			<ErasStakers<T>>::remove_prefix(old_era);
			<ErasStakersClipped<T>>::remove_prefix(old_era);
			<ErasValidatorPrefs<T>>::remove_prefix(old_era);
			<ErasValidatorReward<T>>::remove(old_era);
			<ErasRewardPoints<T>>::remove(old_era);
			<ErasTotalStake<T>>::remove(old_era);
			ErasStartSessionIndex::remove(old_era);
		}
```

If governance reduced the era history, some state will be not cleaned up.

This could be addressed in the extrinsic function which updates the storage item and also cleans up any old state which may have been missed.

thiolliere:
I did my best but I can't finish suggestions:
* about dynamic history_depth I got this implemented here: https://github.com/paritytech/substrate/compare/gav-lazy-rewards...gui-gav-lazy-reward-dynamic-history-depth?expand=1
  but I wanted to double check.
* about Joe extrinsic suggestion we can do something for stakers to be able to specify multiple eras. We can also do that after 2.0 by just adding two new extrinsics and not remove the old ones. Those new extrinsic would just allow to query reward for multiple era.

joepetrowski:
> about Joe extrinsic suggestion we can do something for stakers to be able to specify multiple eras. We can also do that after 2.0 by just adding two new extrinsics and not remove the old ones. Those new extrinsic would just allow to query reward for multiple era.

Yeah, understand there are complexity trade offs here. But I think calling multiple eras should be possible because:
- Tx fees will significantly affect smaller stakers.
- 1 tx/staker/era seems like unnecessary load on the chain.
- Most people will not change their nominations each era (although weight changes, not sure how to deal with that).

shawntabrizi:
@joepetrowski 1 tx per staker per era should be equalish to 1 N batch tx per staker per N eras right?

Maybe a slight overhead cost reduction, but it wouldn't be 1:1.

jacogr:
I agree with @joepetrowski here - the multi tx approach just seems less than optimal. From both a UX perspective and fees eating up all rewards for small nominators.

frame/staking/src/lib.rs
```diff
@@ -684,40 +718,76 @@ decl_storage! {
...
-		/// Nominators for a particular account that is in action right now. You can't iterate
-		/// through validators here, but you can find them in the Session module.
+		/// The current era index.
 		///
-		/// This is keyed by the stash account.
-		pub Stakers get(fn stakers):
-			map hasher(blake2_256) T::AccountId => Exposure<T::AccountId, BalanceOf<T>>;
+		/// This is the latest planned era, depending on how session module queues the validator
+		/// set, it might be active or not.
+		pub CurrentEra get(fn current_era): Option<EraIndex>;
```
shawntabrizi: Why is this an option? When it is `None`?

frame/staking/src/lib.rs
```diff
@@ -980,7 +1092,8 @@ decl_module! {
 					ledger.active = Zero::zero();
 				}
 
-				let era = Self::current_era() + T::BondingDuration::get();
+				// Note: in case there is no current era it is fine to bond one era more.
+				let era = Self::current_era().unwrap_or(0) + T::BondingDuration::get();
```
shawntabrizi: I question why this value would be `None` and why we would `unwrap_or(0)` here.

frame/staking/src/lib.rs
```diff
@@ -1257,6 +1425,21 @@ decl_module! {
...
 		}
+
+		#[weight = SimpleDispatchInfo::FixedOperational(500_000)]
+		fn set_history_depth(origin, #[compact] new_history_depth: EraIndex) {
+			ensure_root(origin)?;
+			if let Some(current_era) = Self::current_era() {
+				HistoryDepth::mutate(|history_depth| {
+					let last_kept = current_era.checked_sub(*history_depth).unwrap_or(0);
+					let new_last_kept = current_era.checked_sub(new_history_depth).unwrap_or(0);
+					for era_index in last_kept..new_last_kept {
```
shawntabrizi: Should only do this loop if `last_kept` < `new_last_kept` right?

frame/staking/src/lib.rs
thiolliere: as discussed on riot:

it is None before the first era is planned, i.e. before the first call for new_era from SessionManager

I wanted to make the least assumption possible. Now staking can be added into a runtime and the first era is planned whenever the session want to start to use it.
If I put zero here then the initial validator set must be provided with a runtime upgrade in case staking is added into the runtime or it start with an empty validator set 🤔 (with genesis things are configured and fine) Anyway the logic looked weird, the first era is planned when it is first planned by session make more sense to me.

frame/staking/src/lib.rs
thiolliere: before the first era is planned by the first call of SessionManager::new_session one user can bond and unbond.

in case the user unbond himself then this unbond will be unlocked one era more as if he has unbound in the first era.

frame/staking/src/lib.rs
```diff
@@ -1320,113 +1602,76 @@ impl<T: Trait> Module<T> {
...
-		match ForceEra::get() {
-			Forcing::ForceNew => ForceEra::kill(),
-			Forcing::ForceAlways => (),
-			Forcing::NotForcing if era_length >= T::SessionsPerEra::get() => (),
-			_ => return None,
+	/// Start a session potentially starting an era.
+	fn start_session(start_session: SessionIndex) {
+		let next_active_era = Self::active_era().map(|e| e + 1).unwrap_or(0);
+		if let Some(next_active_era_start_session_index) = Self::eras_start_session_index(next_active_era) {
+			if next_active_era_start_session_index == start_session {
```
shawntabrizi:
```suggestion
			if next_active_era_start_session_index <= start_session {
```

We can safely make this a less than or equal to right?

thiolliere: oh yes definitely
Also would be great to print a comment in case this `<` happen saying that something wrong happens, probably some sessions has been skipped.
(or at least but into comment that `<` should never happen but it is better in case some session has been skipped somehow. (or if two era start at the same session index..))

frame/staking/src/lib.rs
```diff
@@ -1270,6 +1453,100 @@ impl<T: Trait> Module<T> {
...
+		let mut ledger = <Ledger<T>>::get(&who).ok_or_else(|| Error::<T>::NotController)?;
+		if ledger.next_reward > era {
+			return Err(Error::<T>::InvalidEraToReward.into());
+		}
+
+		ledger.next_reward = era + 1;
+		<Ledger<T>>::insert(&who, &ledger);
+
+		let era_reward_points = <ErasRewardPoints<T>>::get(&era);
+		let commission = Self::validators(&ledger.stash).commission;
```
shawntabrizi: This is broken. Here we get the commission of the validator at the time of this call, not at the time of the era.

This means a validator can collect 84 eras worth of rewards, change their commission to 100%, and then take all the rewards, right?

shawntabrizi: I see you created a new storage item, we just need to use it here.

thiolliere: ah indeed I missed it we should take from eras_prefs same as in do_nominator_payout

frame/staking/src/lib.rs
```diff
@@ -1270,6 +1453,100 @@ impl<T: Trait> Module<T> {
...
+
+		nominator_ledger.next_reward = era + 1;
+		<Ledger<T>>::insert(&who, &nominator_ledger);
+
+		let mut reward = Perbill::zero();
+		let era_reward_points = <ErasRewardPoints<T>>::get(&era);
+
+		// Note: iterator take only `MAX_NOMINATIONS` to avoid querying more validator exposure
+		// than necessary. Anyway a nominator can't validate more than `MAX_NOMINATIONS`.
+		for (validator, nominator_index) in validators.into_iter().take(MAX_NOMINATIONS) {
```
shawntabrizi: Why not just return an error if validators length is greater than `MAX_NOMINTATIONS`?

Might be more user friendly.

frame/staking/src/lib.rs
```diff
@@ -1270,6 +1453,100 @@ impl<T: Trait> Module<T> {
...
 
+	fn do_payout_nominator(who: T::AccountId, era: EraIndex, validators: Vec<(T::AccountId, u32)>)
+		-> DispatchResult
+	{
+		let mut nominator_ledger = <Ledger<T>>::get(&who).ok_or_else(|| Error::<T>::NotController)?;
+		if nominator_ledger.next_reward > era {
+			return Err(Error::<T>::InvalidEraToReward.into());
+		}
+
+		nominator_ledger.next_reward = era + 1;
```
shawntabrizi: era is a user input. I believe you allow for an overflow error here :)

shawntabrizi: One way to solve this issue (beyond additional verification of the inputs), is to track the "last_reward" which is the last reward a user has claimed, rather than the `next_reward`

frame/staking/src/lib.rs
```diff
@@ -1270,6 +1453,100 @@ impl<T: Trait> Module<T> {
...
+		Ok(())
+	}
+
+	fn do_payout_validator(who: T::AccountId, era: EraIndex) -> DispatchResult {
+		let mut ledger = <Ledger<T>>::get(&who).ok_or_else(|| Error::<T>::NotController)?;
+		if ledger.next_reward > era {
+			return Err(Error::<T>::InvalidEraToReward.into());
+		}
+
+		ledger.next_reward = era + 1;
```
shawntabrizi: overflow here too

frame/staking/src/lib.rs
```diff
@@ -1270,6 +1453,100 @@ impl<T: Trait> Module<T> {
...
+		// Note: this is zero if the era is not finished yet.
+		let era_payout = <ErasValidatorReward<T>>::get(&era);
+		if let Some(imbalance) = Self::make_payout(&nominator_ledger.stash, reward * era_payout) {
+			Self::deposit_event(RawEvent::Reward(who, imbalance.peek()));
+		}
+
+		Ok(())
+	}
+
+	fn do_payout_validator(who: T::AccountId, era: EraIndex) -> DispatchResult {
```
shawntabrizi: This function should immediately fail if era > current era. A user can really screw themselves if we let them set "next_era" too high.

Really if we just track "last paid out era" and we put it into storage only after the payout, I think the user experience will be vastly improved. It allows someone to try and payout during a future era, the tx fails, and their "last era" tracker is untouched.

frame/staking/src/lib.rs
thiolliere: yes the user can reset its next_reward to zero using this error :-/
last_reward is OK but then it must be an Option which is a bit boring in the code,
let's just saturating add. the era index shouldn't wrap around.

frame/staking/src/lib.rs
thiolliere: > Really if we just track "last paid out era" and we put it into storage only after the payout, I think the user experience will be vastly improved. It allows someone to try and payout during a future era, the tx fails, and their "last era" tracker is untouched.

this doesn't change the issue, you point, if the user set a era index too high then he is screwed, user should claim their payout for `era < active_era` we can check that indeed.

frame/staking/src/lib.rs
```diff
@@ -902,7 +1007,13 @@ decl_module! {
 
 			let stash_balance = T::Currency::free_balance(&stash);
 			let value = value.min(stash_balance);
-			let item = StakingLedger { stash, total: value, active: value, unlocking: vec![] };
+			let item = StakingLedger {
+				stash,
+				total: value,
+				active: value,
+				unlocking: vec![],
+				next_reward: Self::current_era().unwrap_or(0),
```
thiolliere: I think this should be current_era + 1 or 0

frame/staking/src/lib.rs
thiolliere: but if we do saturating then last era index is either non rewardable or you can claim the reward indefinitely, I'll change for option

frame/staking/src/lib.rs
thiolliere: yes indeed it doesn't cost anything, done

frame/staking/src/lib.rs
thiolliere: I now query EraRewardPayout as `Option<Balance>` if the era has no payout (era out of history or era not yet finished) then we just return InvalidEra error without fetching and upgrading ledger.last_reward

frame/staking/src/lib.rs
thiolliere: thanks for having this done, looks good to me

frame/staking/src/lib.rs
```diff
@@ -1647,19 +1950,120 @@ impl<T: Trait> Module<T> {
...
+		Self::start_session(start_index)
+	}
+	fn end_session(end_index: SessionIndex) {
+		Self::end_session(end_index)
 	}
-	fn end_session(_end_index: SessionIndex) {}
 }
 
+/// This implementation has the same constrains as the implementation of
+/// `pallet_session::SessionManager`.
```
shawntabrizi: This session manager implementation is still using a deprecated storage item.

```
let exposure = <Stakers<T>>::get(&v);
```

Can we remove all the deprecated code from this PR, and hard code some of the storage names in the migration function? Would help mitigate issues like this.

frame/staking/src/lib.rs
shawntabrizi: It makes sense to me if we you call unbond, and `current_era` is `none`, you should be able to withdraw unbonded immediately.

frame/staking/src/lib.rs
thiolliere: I'm on it, I'll make deprecated storage out of the decl_storage

frame/staking/src/lib.rs
```diff
@@ -1288,6 +1438,111 @@ impl<T: Trait> Module<T> {
...
+	fn do_payout_nominator(who: T::AccountId, era: EraIndex, validators: Vec<(T::AccountId, u32)>)
+		-> DispatchResult
+	{
+		// validators len must not exceed `MAX_NOMINATIONS` to avoid querying more validator
+		// exposure than necessary.
+		if validators.len() > MAX_NOMINATIONS {
+			return Err(Error::<T>::InvalidNumberOfNominations.into());
+		}
+
+		// Note: if era has no reward to be claimed, era may be futur. better not to update
```
shawntabrizi:
```suggestion
		// Note: if era has no reward to be claimed, era may be future. better not to update
```

frame/staking/src/lib.rs
```diff
@@ -1288,6 +1438,111 @@ impl<T: Trait> Module<T> {
...
+
+		if let Some(imbalance) = Self::make_payout(&nominator_ledger.stash, reward * era_payout) {
+			Self::deposit_event(RawEvent::Reward(who, imbalance.peek()));
+		}
+
+		Ok(())
+	}
+
+	fn do_payout_validator(who: T::AccountId, era: EraIndex) -> DispatchResult {
+		// Note: if era has no reward to be claimed, era may be futur. better not to update
```
shawntabrizi:
```suggestion
		// Note: if era has no reward to be claimed, era may be future. better not to update
```

gnunicorn:
> I think we should still combine the one storage item I pointed out.

is this intended to happen before the merge? adding grumble, please remove once done/if not to be done.

shawntabrizi:
@gnunicorn all is good now!

PR #4480, opened 2019-12-22
Add tipping into treasury

### LINKED ISSUE #4477: Tip request in treasury
Additional dispatch items in treasury:
- `fn request_tip(origin /*signed*/, url: Vec<u8>)`; does nothing.
- `fn give_tip(origin /*council member*/, hash_of_url_concat_who: T::Hash, tip_value: BalanceOf<T>)`
- `fn send_tip(origin /*council member*/, hash_of_url_concat_who: T::Hash, url: Vec<u8>, who: T::AccountId)`

Once either `TipPeriod` time has elapsed after `(members().len() + 1) / 2` unique members have done `give_tip` or all members have done `give_tip`, then a tip is paid, which is the median of the `tip_value`s that have been recorded for the `url_hash`/`who` pair.

gavofyork:
Closes #4477 

TODO: 
- [x] Module-level docs
- [x] Function-level docs (with complexity)
- [x] Tests


kianenigma:
>  Custom weights(?)

https://github.com/paritytech/substrate/pull/4158

frame/support/src/traits.rs
```diff
@@ -53,15 +53,15 @@ impl<T: Default> Get<T> for () {
...
 /// in nature to `Get`, except it is designed to be lazy rather than active (you can't ask it to
 /// enumerate all values that it contains) and work for multiple values rather than just one.
-pub trait Contains<T> {
+pub trait Contains<T: Ord> {
 	/// Return `true` if this "contains" the given value `t`.
-	fn contains(t: &T) -> bool;
-}
+	fn contains(t: &T) -> bool { Self::sorted_members().binary_search(t).is_ok() }
 
-impl<V: PartialEq, T: Get<V>> Contains<V> for T {
```
gavofyork: this won't work if we also have specialisation for a module as we do now with the treasury module, hence the need for these changes and `ord_parameter_types`.

frame/treasury/src/lib.rs
```diff
@@ -107,7 +154,131 @@ pub trait Trait: frame_system::Trait {
...
+		Rejected(ProposalIndex, Balance),
+		/// Some of our funds have been burnt.
+		Burnt(Balance),
+		/// Spending has finished; this is the amount that rolls over until next spend.
+		Rollover(Balance),
+		/// Some funds have been deposited.
+		Deposit(Balance),
+		/// A new tip suggestion has been opened.
+		NewTip(Hash),
+		/// A tip suggestion has reached threshold and .
```
shawntabrizi: incomplete sentence

frame/treasury/src/lib.rs
```diff
@@ -107,7 +154,131 @@ pub trait Trait: frame_system::Trait {
...
+		TipClosing(Hash),
+		/// A new tip suggestion has been opened.
```
shawntabrizi:
```suggestion
		/// A new tip suggestion has been closed.
```

frame/treasury/src/lib.rs
shawntabrizi:
```suggestion
		/// A tip suggestion has reached threshold and is closing.
```

frame/treasury/src/lib.rs
```diff
@@ -107,7 +154,131 @@ pub trait Trait: frame_system::Trait {
...
+		TipClosed(Hash, AccountId, Balance),
+		/// A new tip suggestion has been opened.
```
shawntabrizi:
```suggestion
		/// A tip suggestion has been retracted.
```

frame/treasury/src/lib.rs
```diff
@@ -107,7 +154,131 @@ pub trait Trait: frame_system::Trait {
...
+);
+
+decl_error! {
+	/// Error for the treasury module.
+	pub enum Error for Module<T: Trait> {
+		/// Proposer's balance is too low.
+		InsufficientProposersBalance,
+		/// No proposal at that index.
+		InvalidProposalIndex,
+		/// The reason give is just too big.
```
shawntabrizi:
```suggestion
		/// The reason given is just too big.
```

frame/treasury/src/lib.rs
```diff
@@ -124,10 +295,183 @@ decl_module! {
...
+		/// - `reason`: The reason for, or the thing that deserves, the tip; generally this will be
+		///   a UTF-8-encoded URL.
+		/// - `who`: The account which should be credited for the tip.
+		///
+		/// Emits `NewTip` if successful.
+		///
+		/// # <weight>
+		/// - `O(R)` where `R` length of `reason`.
+		/// - One balance operation.
+		/// - One storage mutation (codec `O(R)`).
```
shawntabrizi: When you say `codec `O(R)` you are describing that interacting with this storage item will require us to parity-scale-codec encode/decode R items?

frame/treasury/src/lib.rs
```diff
@@ -107,7 +154,131 @@ pub trait Trait: frame_system::Trait {
...
+/// An open tipping "motion". Retains all details of a tip including information on the finder
+/// and the members who have voted.
+#[derive(Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+pub struct OpenTip<
+	AccountId: Parameter,
+	Balance: Parameter,
+	BlockNumber: Parameter,
+	Hash: Parameter,
+> {
+	/// The reason for the tip. Should be a human-readable UTF-8 encoded string. A URL would be
```
shawntabrizi:
```suggestion
	/// The hash of the reason for the tip. The hashed value should be a human-readable UTF-8 encoded string. A URL would be
```

frame/treasury/src/lib.rs
```diff
@@ -124,10 +295,183 @@ decl_module! {
...
+			let reason_hash = T::Hashing::hash(&reason[..]);
+			ensure!(!Reasons::<T>::exists(&reason_hash), Error::<T>::AlreadyKnown);
+			let hash = T::Hashing::hash_of(&(&reason_hash, &who));
+			ensure!(!Tips::<T>::exists(&hash), Error::<T>::AlreadyKnown);
+
+			let deposit = T::TipReportDepositBase::get()
+				+ T::TipReportDepositPerByte::get() * (reason.len() as u32).into();
+			T::Currency::reserve(&finder, deposit)?;
+
+			Reasons::<T>::insert(&reason_hash, &reason);
```
shawntabrizi: It seems that in general you have relaxed your position with on-chain strings.

Just to clarify, so as long as a user places a deposit for the data they place on chain, it seems okay to put strings/arbitrary bytes on the blockchain? Following a rule like this in the general, do we run into an issues similar to the data bloat which is found with smart contract platforms (and is addressed with storage rent)?

I guess I wonder if this kind of code is more of a "solution for now", and would be changed in the future once we have better integration with services like IPFS.

The alternative here is off-chain coordination of the `reason` string, which seems much harder for a UI to manage.

frame/treasury/src/lib.rs
```diff
@@ -87,6 +119,21 @@ pub trait Trait: frame_system::Trait {
 	/// Origin from which rejections must come.
 	type RejectOrigin: EnsureOrigin<Self::Origin>;
 
+	/// Origin from which tippers must come.
+	type Tippers: Contains<Self::AccountId>;
+
+	/// The period for which a tip remains open after is has achieved threshold tippers.
+	type TipCountdown: Get<Self::BlockNumber>;
+
+	/// The amount of the final tip which goes to the original reporter of the tip.
```
shawntabrizi:
```suggestion
	/// The percent of the final tip which goes to the original reporter of the tip.
```

frame/treasury/src/lib.rs
```diff
@@ -457,6 +826,146 @@ mod tests {
...
 	}
 
+	fn tip_hash() -> H256 {
+		BlakeTwo256::hash_of(&(BlakeTwo256::hash(b"awesome.dot"), 3u64))
+	}
+
+	#[test]
+	fn tip_new_cannot_be_used_twice() {
+		new_test_ext().execute_with(|| {
+			// Check that accumulate works when we have Some value in Dummy already.
```
shawntabrizi:
```suggestion
```

frame/treasury/src/lib.rs
```diff
@@ -457,6 +826,146 @@ mod tests {
...
+				Treasury::tip_new(Origin::signed(11), b"awesome.dot".to_vec(), 3, 10),
+				Error::<Test>::AlreadyKnown
+			);
+		});
+	}
+
+	#[test]
+	fn report_awesome_and_tip_works() {
+		new_test_ext().execute_with(|| {
+			// Check that accumulate works when we have Some value in Dummy already.
```
shawntabrizi:
```suggestion
```

frame/treasury/src/lib.rs
```diff
@@ -457,6 +826,146 @@ mod tests {
...
+			assert_eq!(Balances::reserved_balance(&0), 0);
+			assert_eq!(Balances::free_balance(&0), 102);
+			assert_eq!(Balances::free_balance(&3), 8);
+		});
+	}
+
+	#[test]
+	fn report_awesome_from_beneficiary_and_tip_works() {
+		new_test_ext().execute_with(|| {
+			// Check that accumulate works when we have Some value in Dummy already.
```
shawntabrizi:
```suggestion
```

frame/treasury/src/lib.rs
```diff
@@ -457,6 +826,146 @@ mod tests {
...
+			assert_ok!(Treasury::close_tip(Origin::signed(100), h.into()));
+			assert_eq!(Balances::reserved_balance(&0), 0);
+			assert_eq!(Balances::free_balance(&0), 110);
+		});
+	}
+
+	#[test]
+	fn close_tip_works() {
+		new_test_ext().execute_with(|| {
+			// Check that accumulate works when we have Some value in Dummy already.
```
shawntabrizi:
```suggestion
```

frame/treasury/src/lib.rs
```diff
@@ -457,6 +826,146 @@ mod tests {
...
+			assert_eq!(Balances::free_balance(&3), 10);
+
+			assert_noop!(Treasury::close_tip(Origin::signed(100), h.into()), Error::<Test>::UnknownTip);
+		});
+	}
+
+	#[test]
+	fn retract_tip_works() {
+		new_test_ext().execute_with(|| {
+			// Check that accumulate works when we have Some value in Dummy already.
```
shawntabrizi:
```suggestion
```

frame/treasury/src/lib.rs
```diff
@@ -457,6 +826,146 @@ mod tests {
...
+			assert_ok!(Treasury::retract_tip(Origin::signed(0), h.clone()));
+			System::set_block_number(2);
+			assert_noop!(Treasury::close_tip(Origin::signed(0), h.into()), Error::<Test>::UnknownTip);
+		});
+	}
+
+	#[test]
+	fn tip_median_calculation_works() {
+		new_test_ext().execute_with(|| {
+			// Check that accumulate works when we have Some value in Dummy already.
```
shawntabrizi:
```suggestion
```

frame/treasury/src/lib.rs
```diff
@@ -457,6 +826,146 @@ mod tests {
...
+			System::set_block_number(2);
+			assert_ok!(Treasury::close_tip(Origin::signed(0), h.into()));
+			assert_eq!(Balances::free_balance(&3), 10);
+		});
+	}
+
+	#[test]
+	fn tip_changing_works() {
+		new_test_ext().execute_with(|| {
+			// Check that accumulate works when we have Some value in Dummy already.
```
shawntabrizi:
```suggestion
```

frame/treasury/src/lib.rs
```diff
@@ -457,6 +826,146 @@ mod tests {
...
+			assert_ok!(Treasury::tip(Origin::signed(11), h.clone(), 10000));
+			assert_ok!(Treasury::tip(Origin::signed(12), h.clone(), 10000));
+			assert_ok!(Treasury::tip(Origin::signed(13), h.clone(), 0));
+			assert_ok!(Treasury::tip(Origin::signed(14), h.clone(), 0));
+			assert_ok!(Treasury::tip(Origin::signed(12), h.clone(), 1000));
+			assert_ok!(Treasury::tip(Origin::signed(11), h.clone(), 100));
+			assert_ok!(Treasury::tip(Origin::signed(10), h.clone(), 10));
+			System::set_block_number(2);
+			assert_ok!(Treasury::close_tip(Origin::signed(0), h.into()));
+			assert_eq!(Balances::free_balance(&3), 10);
```
shawntabrizi: To me this does not really check that tip changing does work.

Either I would want to look at the tip values in storage to see the value updated, or I would want to see a final resulting reward that would have changed given the tips changing.

In this case, if tips do not update `(0, 0, 10, 10000, 10000)` or  tips do update `(0, 0, 10, 100, 1000)`, the result of this test would be the same.

frame/treasury/src/lib.rs
gavofyork: correct.

frame/treasury/src/lib.rs
```diff
@@ -124,10 +295,183 @@ decl_module! {
...
+			const MAX_SENSIBLE_REASON_LENGTH: usize = 16384;
+			ensure!(reason.len() <= MAX_SENSIBLE_REASON_LENGTH, Error::<T>::ReasonTooBig);
+
+			let reason_hash = T::Hashing::hash(&reason[..]);
+			ensure!(!Reasons::<T>::exists(&reason_hash), Error::<T>::AlreadyKnown);
+			let hash = T::Hashing::hash_of(&(&reason_hash, &who));
+			ensure!(!Tips::<T>::exists(&hash), Error::<T>::AlreadyKnown);
+
+			let deposit = T::TipReportDepositBase::get()
+				+ T::TipReportDepositPerByte::get() * (reason.len() as u32).into();
```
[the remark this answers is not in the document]
gavofyork: because they're trusted values. we don't stop runtime developers from self-harming elsewhere and i don't want to make code less readable here.

frame/treasury/src/lib.rs
gavofyork: true; but for all instances where users can place data on-chain, there needs to be economic cost for each byte-block. this can be done through a per-time charge (rent) or through a deposit (where there's still a payment but through opportunity cost/inflation). deposits will be larger - probably be about 10x.

i don't think it's to be encouraged *in general*, but right now we have no option since there is no compelling and readily available alternative. it's not something i would rush into with polkadot mainnet but for kusama it seems ok.

frame/treasury/src/lib.rs
gavofyork: if tips did not update, then the tips would be 10000,10000,10000,0,0 and the result tip would be 10.

shawntabrizi: Of course, I was being stupid and missed the original tip from `tip_new`:

```
assert_ok!(Treasury::tip_new(Origin::signed(10), b"awesome.dot".to_vec(), 3, 10000));

File: frame/treasury/src/lib.rs [hunk 19]
```diff
@@ -124,10 +295,188 @@ decl_module! {
...
+		#[weight = SimpleDispatchInfo::FixedNormal(50_000)]
+		fn tip(origin, hash: T::Hash, tip_value: BalanceOf<T>) {
+			let tipper = ensure_signed(origin)?;
+			ensure!(T::Tippers::contains(&tipper), BadOrigin);
+
+			let mut tip = Tips::<T>::get(hash).ok_or(Error::<T>::UnknownTip)?;
+			if Self::insert_tip_and_check_closing(&mut tip, tipper, tip_value) {
+				Self::deposit_event(RawEvent::TipClosing(hash.clone()));
+			}
+			Tips::<T>::insert(&hash, tip);
```
wliyongfeng, line 450: Sorry for comment with this historical PR, but I want to ask were there any special considerations without emitting a `Tip` event here? Just to avoid too many events?

Actually it's a little troublesome without this event to handle tip off-chain.

thiolliere: I think we can emit some events if needed, it shouldn't be too may events AFAICT. You can open an issue or PR

wliyongfeng: > I think we can emit some events if needed, it shouldn't be too may events AFAICT. You can open an issue or PR

I see. Thanks.

PR #4481, opened 2019-12-23
ci: increase git cloning depth to 100

gabreal:
CI jobs are failing due to recently introduced shallow clones for ci checks e.g. https://gitlab.parity.io/parity/substrate/-/jobs/333878

Increasing the cloning depth to 100 should be a safe value :crossed_fingers: 

failing refs in local copy:
```
fatal: ambiguous argument 'origin/master...2003b9d8a08c9b9370a94cb0ee3c0f3e05607ce6': unknown revision or path not in the working tree.
```

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,7 +15,7 @@ stages:
 
 variables:
   GIT_STRATEGY:                    fetch
-  GIT_DEPTH:                       "3"
+  GIT_DEPTH:                       100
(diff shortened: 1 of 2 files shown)


bkchr:
We still get: 
`fatal: ambiguous argument 'origin/master...0ee1bc18a94be7fa73258c8b3d674ad428ab9302': unknown revision or path not in the working tree.`

Could you please also update `check-runtime` to fail properly as well. It should not succeed when it does not work (as it is currently).

gabreal:
@bkchr which pull request was that?

bkchr:
This one^^

PR #4482, opened 2019-12-23
Fix error on new version of structopt

kigawas:
structopt 0.3.6 added unimplemented `augment_clap`


niklasad1:
@kigawas 

I don't understand the reasoning behind this PR but it is related to https://github.com/TeXitoi/structopt/issues/315 i.e, `structopt v0.3.6` has a breaking change?

I think this doesn't concern `substrate` itself because of `Cargo.lock` but other crates that depend on it?

kigawas:
We need to lock the version of structopt to avoid unintentional change or block a normal cargo update.

CreepySkeleton:
`structopt` devs here. We yanked `0.3.6`, from now on everything should work fine. We will decide what to do with the new release - and how should we do it - ASAP.

CreepySkeleton:
@niklasad1 @kigawas Actually, there's no need for this PR anymore. We'll make sure that all the future releases pull the right dependencies.

niklasad1:
Closing because it is not needed anymore,

> @niklasad1 @kigawas Actually, there's no need for this PR anymore. We'll make sure that all the future releases pull the right dependencies.

Issue #830, opened 2018-09-27
Detailed memory usage report


arkpar:
Major memory consumers should track memory usage and that should be printed in the informant periodically and reported to telemetry. This includes

- [x] RocksDB cache size
- [x] state-db pruning and finalisation overlays
- [ ] libp2p buffers?
- [x] Block queue
- [x] Consensus and gossip message store
- [x] Transaction queue

gavofyork:
This is also a candidate for #814

guanqun:
use heapsize's heap_size_of_children?

arkpar:
> use heapsize's heap_size_of_children?

Probably. Except for RocksDB that has its own API for reporting memory usage.

PR #4483, opened 2019-12-23
Update to new kvdb/trie/memdb versions

NikVolf:
- No more `ElasticArray`
- Columns are now required, and no "default column" notion exists

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1012,14 +1012,6 @@ version = "1.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
-name = "elastic-array"
-version = "0.10.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "heapsize 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
 name = "enumflags2"
 version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


bkchr:
Looks good, but does not compile. What did change in the new trie version?

NikVolf:
> Looks good, but does not compile. What did change in the new trie version?

Yes, doing it already, looks like the version currently published is not compilable to wasm/no-std

bkchr:
I also just want to make sure that we not again break the storage root, because of a trie update ^^

NikVolf:
> What did change in the new trie version?

Haven't seen any logic changes, there were just some api changes (no ElasticArray array now, columns are now `u32`, not `Option`). I'm also currently adding the memory footprint calculations of database structures (#830)

bkchr:
(Please use merge commits)

Cargo.lock
```diff
@@ -8297,7 +8339,8 @@ dependencies = [
 "checksum kernel32-sys 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)" = "7507624b29483431c0ba2d82aece8ca6cdba9382bff4ddd0f7490560c056098d"
 "checksum kv-log-macro 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)" = "8c54d9f465d530a752e6ebdc217e081a7a614b48cb200f6f0aee21ba6bc9aabb"
 "checksum kvdb 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)" = "c1b2f251f01a7224426abdb2563707d856f7de995d821744fd8fa8e2874f69e3"
```
niklasad1: I think you have missed updating `kvdb-rocksdb` to `v0.3` but shouldn't be related to the errors related `trie-db` I guess

NikVolf:
@bkchr Tested with Kusama, just in case.

PR #4487, opened 2019-12-23
client/authority-discovery: Limit number of connections to authorities

mxinden:
Instead of connecting to all sentry nodes of all authorities, with this
patch the authority discovery module does the following:

- Choose one sentry node per authority at random.

- Choose MAX_NUM_AUTHORITY_CONN out of the above at random.

The module uses randomness to prevent hot spots, e.g. all nodes trying
to connect to a single node. If the authority discovery module would
choose the nodes to connect to at random on each new address that it
learns of, the node would go through a lot of connection churn.  Instead
it creates a random seed at start up and uses this seed for its RNG on
each update cycle to produce similar outputs.

While this should reduce churn a lot, it is in no way perfect. Alternative
suggestions welcome.


gavofyork:
test failing...

client/authority-discovery/src/addr_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	pub fn insert(&mut self, id: Id, mut addresses: Vec<Addr>) {
+		if addresses.is_empty() {
+			return;
+		}
+
+		addresses.sort_by(|a, b| {
+			a.as_ref()
+				.partial_cmp(b.as_ref())
```
niklasad1: Can you explain why `partial_cmp` is used here instead `cmp`?

AFAIK, u8 implements Ordering and it is infallible. `partial_cmp` should be used for comparisons that can fail for example floats.

client/authority-discovery/src/addr_cache.rs
```diff
@@ -0,0 +1,184 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+#![warn(missing_docs)]
```
niklasad1: This will not work because this is not the crate root.

client/authority-discovery/src/addr_cache.rs
bkchr: And why not use `sort_unstable`?

mxinden: Right, using `cmp` instead of `unwrap`ing a `parital_cmp` is a lot better :+1:, thanks!

Switched to `sort_unstable_by` as well.

client/authority-discovery/src/addr_cache.rs
mxinden: Removed. Attribute is already present at the root of the crate.

PR #4488, opened 2019-12-23
Fix replace_previous of Ready Transaction Queue.

tomusdrw:
By looking at the code I found a potential issue with `replace_previous`. The clean up procedure that was supposed to remove all traces of replaced transactions was not correct. The bug was not critical, it could only cause a slight incoherence in the internal pool tracking.

The main bug was assuming that newly imported transaction will always have empty `unlock` collection. Which is not necessarily true, if we replace a transaction in the middle of a chain:
`Tx1 -> Tx2 - > Tx3`, then we replace with `Tx2'` and that transaction should still unlock `Tx3`.

### DIFF
--- a/client/transaction-pool/graph/src/ready.rs
+++ b/client/transaction-pool/graph/src/ready.rs
@@ -562,7 +562,6 @@ mod tests {
 		ready.import(x)
 	}
 
-
 	#[test]
 	fn should_replace_transaction_that_provides_the_same_tag() {
 		// given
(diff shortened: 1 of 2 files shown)


client/transaction-pool/graph/src/ready.rs
```diff
@@ -517,6 +554,15 @@ mod tests {
...
 	}
 
+	fn import<H: hash::Hash + Eq + Member + Serialize, Ex>(
+		ready: &mut ReadyTransactions<H, Ex>,
+		tx: Transaction<H, Ex>
+	) -> error::Result<Vec<Arc<Transaction<H, Ex>>>> {
+		let x = WaitingTransaction::new(tx, ready.provided_tags(), &[]);
+		ready.import(x)
+	}
+
```
bkchr:
```suggestion
```

client/transaction-pool/graph/src/ready.rs
```diff
@@ -441,10 +478,10 @@ pub struct BestIterator<Hash, Ex> {
 }
 
 impl<Hash: hash::Hash + Member, Ex> BestIterator<Hash, Ex> {
-	/// Depending on number of satisfied requirements insert given ref
+	/// Depending on number of satisfied equirements insert given ref
```
bkchr:
```suggestion
	/// Depending on number of satisfied requirements insert given ref
```

PR #4489, opened 2019-12-23
Extract execution engines definitions into their own crates

pepyakin:
This PR extracts wasmi and wasmtime into their own execution engines.

This has a benefit that it makes adding another execution engine easier. However, the main reason behind this PR is to start refactoring executor crate: for now its responsibilities are [rather vague](https://github.com/paritytech/substrate/blob/042c9bb89c6ca8a432d0f66c70c97ebc36e49702/client/executor/src/lib.rs#L19-L27).

This PR does not alter any functionality.


primitives/runtime-interface/src/impls.rs
```diff
@@ -54,6 +54,10 @@ pub fn pointer_and_len_to_u64(ptr: u32, len: u32) -> u64 {
 }
 
 /// Splits an `u64` into the pointer and length.
+///
+/// Runtime API functions return a 64-bit value which encodes a pointer in the least-significant
+/// 32-bits and a length in the most-significant 32 bits. This interprets the returned value as a pointer,
+/// length tuple.
 pub fn pointer_and_len_from_u64(val: u64) -> (u32, u32) {
```
bkchr: I think we should move this function into the lib.rs file. As this file should only contain implementations of the traits for types.

client/executor/common/src/error.rs
```diff
@@ -35,7 +33,7 @@ pub enum Error {
 	Wasmi(wasmi::Error),
 	/// Wasmtime action error
 	#[cfg(feature = "wasmtime")]
-	Wasmtime(ActionError),
+	Wasmtime(wasmtime_jit::ActionError),
```
bkchr: Why do we still have wasmi and wasmtime related error messages. I would like to have them removed.

primitives/runtime-interface/src/impls.rs
pepyakin: Good point! Should've looked around more closely...

client/executor/common/src/error.rs
pepyakin: The engines still need to return error messages. How do you feel about returning the errors coming out of them as stringly-typed, i.e. `Other(String)`?

bkchr: Either this or boxing the error?

PR #4491, opened 2019-12-24
less dupes, cleanup

TriplEight:
- less dupes:
  `&name` is called "the yaml anchor", and used to reduce dupes, like that:
```
jobname:          &name
  stage:          test
  variables:
    var1:         1
  script:
    - do that

another-jobname:
  # it is the same as "jobname`, but the script is different
  <<:   *name
  script:
    - do this
```

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -287,14 +287,10 @@ build-linux-substrate:             &build-binary
           tee ./artifacts/substrate/VERSION;
       fi
     - sha256sum ./artifacts/substrate/substrate | tee ./artifacts/substrate/substrate.sha256
+    - printf '\n# building node-template\n\n'
+    - ./.maintain/node-template-release.sh ./artifacts/substrate/substrate-node-template.tar.gz
+    - cp -r .maintain/docker/substrate.Dockerfile ./artifacts/substrate/
     - sccache -s
-
-build-linux-node-template:
-  <<:                              *build-binary
-  script:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -286,25 +287,23 @@ build-linux-substrate:
...
-    - printf '\n# building node-template\n\n'
-    - ./.maintain/node-template-release.sh ./artifacts/substrate/substrate-node-template.tar.gz
-    - cp -r .maintain/docker/substrate.Dockerfile ./artifacts/substrate/
     - sccache -s
 
+build-linux-node-template:
+  <<:                              *build-binary
+  script:
+      - ./.maintain/node-template-release.sh ./artifacts/substrate/substrate-node-template.tar.gz
+      - cp -r .maintain/docker/substrate.Dockerfile ./artifacts/substrate/
```
bkchr: Why do we do this here? And not in the build substrate job above?

TriplEight: you're right, right now it's not effective to split these jobs

TriplEight: actually, we are compiling it right there: https://gitlab.parity.io/parity/substrate/-/jobs/339822#L706 , during `cargo build`.
Can we use that artifact?

bkchr: I was speaking about the docker file. Putting the node template into its own job sounds reasonable.

PR #4492, opened 2019-12-24
check_polkadot now runs nightly and on master (merges) only

TriplEight:
I've been asked to run this job only nightly and on merges, anyways it's always failing.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -351,11 +351,14 @@ check_warnings:
       fi
   allow_failure:                   true
 
-# Check whether Polkadot 'master' branch builds using this Substrate commit.
+# Nightly check whether Polkadot 'master' branch builds.
 check_polkadot:
   stage:                           build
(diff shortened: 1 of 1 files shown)


bkchr:
What is the purpose of running this job on nightly? Who requested that? Next question is, where do we see nightly jobs? Is anyone informed by the result of these jobs?

TriplEight:
I remember you hinting this, and a couple of other devs. That's why I did this PR.
There will be a chat bot telling about failed schedules.

bkchr:
When this is run per PR, what should we get from running it on nightly? We already know when a pr breaks polkadot when the check is run for each pr.

gavofyork:
To make it effective when running on PRs, we would need to merge every PR to master and re-run every time master changes. That's going to hit the build server much harder than just rerunning on master and for not much point as the vast majority of the time, merging and re-running gives the same result as the original.

PR #4493, opened 2019-12-24
Memory usage stats (database cache for now)

NikVolf:
on top of #4483 

- Also split `Client` `usage_info` from `chain_info`. More data (i/o stats, more detailed memory usage breakdown) is coming to the first one later, and it is needed only in few places.

As a part of #830

- [x] telemetry
- [ ] wait for kvdb update (currently working only with `parity-common/master` `trie/master`)


NikVolf:
Closing in favour of several small prs coming

Issue #4498, opened 2019-12-25
Broken MultiSignature verification for ECDSA keys


akru:
Looks like  `Verify` instance will not works for `ECDSA` because this instance recover the compressed public key:

https://github.com/paritytech/substrate/blob/master/primitives/runtime/src/lib.rs#L311

```rust
(MultiSignature::Ecdsa(ref sig), who) => {
	let m = sp_io::hashing::blake2_256(msg.get());
	match sp_io::crypto::secp256k1_ecdsa_recover_compressed(sig.as_ref(), &m) {
		Ok(pubkey) =>
			&sp_io::hashing::blake2_256(pubkey.as_ref())
				== <dyn AsRef<[u8; 32]>>::as_ref(who),
```

But  `ecdsa` module uses the uncompressed public key for `AccountId` derivation:

https://github.com/paritytech/substrate/blob/master/primitives/core/src/ecdsa.rs#L51
https://github.com/paritytech/substrate/blob/master/primitives/runtime/src/lib.rs#L249

PR #4499, opened 2019-12-25
Fix MultiSignature verification for ECDSA keys

akru:
**Motivation**

ECDSA signing aren't working in runtimes with `MultiSignature` signatures. https://github.com/paritytech/substrate/issues/4498

**Solution**

Use uncompressed recovery in `verify` routine.


akru:
I think this one is critical issue because Polkadot and Kusama affected. @bkchr

bkchr:
Hey, ty :) I checked the source code and found more places that required fixes and I also added some tests: https://github.com/paritytech/substrate/pull/4502

Nevertheless, ty again!

PR #4502, opened 2019-12-26
Make `MultiSigner` use compressed ECDSA public key

bkchr:
`MultiSigner` did not compressed the ECDSA public key before hashing it but `MultiSignature` expected that the public key is compressed when verifying a signature. Thus, the signature verification could never succeed. 

This pr changes the following:
- `MultiSigner` compresses the public key before hashing it with `blake256`.
- `Verify for ecdsa::Signature` compresses given public key before comparing it.
- `ecdsa::Public` is now an enum that distinguishes between the full and compressed format.

Fixes: #4498

### DIFF
--- a/primitives/runtime/src/lib.rs
+++ b/primitives/runtime/src/lib.rs
@@ -212,7 +212,7 @@ pub enum MultiSigner {
 	Ed25519(ed25519::Public),
 	/// An Sr25519 identity.
 	Sr25519(sr25519::Public),
-	/// An SECP256k1/ECDSA identity (actually, the Blake2 hash of the pub key).
+	/// An SECP256k1/ECDSA identity (actually, the Blake2 hash of the compressed pub key).
 	Ecdsa(ecdsa::Public),
 }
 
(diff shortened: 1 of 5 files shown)


primitives/runtime/src/lib.rs
```diff
@@ -246,7 +246,9 @@ impl traits::IdentifyAccount for MultiSigner {
 		match self {
 			MultiSigner::Ed25519(who) => <[u8; 32]>::from(who).into(),
 			MultiSigner::Sr25519(who) => <[u8; 32]>::from(who).into(),
-			MultiSigner::Ecdsa(who) => sp_io::hashing::blake2_256(who.as_ref()).into(),
+			MultiSigner::Ecdsa(who) => sp_io::hashing::blake2_256(
+				&who.as_compressed().expect("what should we do?")[..],
```
bkchr: @gavofyork what should we do here? I think the best would be to propagate the error via `IdentifyAccount`.

gavofyork: i believe this cannot fail for any valid keypair, so `.expect` should be fine

gavofyork:
but... what's the point of supporting uncompressed public key format?

bkchr:
> but... what's the point of supporting uncompressed public key format?

Good question, I just followed the initial implementation. However, I think that for ergonomic reasons the support for both sounds reasonable to me. If you say that people only use the compressed version, I can remove the support for full public keys.

gavofyork:
it does leave this annoying `.expect` thing in.

PR #4503, opened 2019-12-26
Support loading the URI from a file in subkey

bkchr:


bin/utils/subkey/src/main.rs
```diff
@@ -244,6 +247,29 @@ fn main() {
...
+/// Get `URI` from CLI or prompt the user.
+///
+/// `URI` is extracted from `matches` by using `match_name`.
+///
+/// If the `URI` given as CLI argument is a file, the file content is taken as `URI`.
+/// If no `URI` is given to the CLI, the user is prompted for it.
+fn get_uri(match_name: &str, matches: &ArgMatches) -> String {
+	if let Some(uri) = matches.value_of(match_name) {
+		let file = PathBuf::from(uri);
+		if file.is_file() && file.exists() {
```
kianenigma: is it needed to also do `.exists()`?

bkchr: Ahh yeah, I have misread the documentation. This is not required.

bkchr:
What you mean by prefer file over stdin? We don't do this?

kianenigma:
Was just wondering where the PR is coming from, with what motivation. In other words, what is the advantage of reading from a file rather than from the command line's stdin?

(and I noted that I personally think file is better but wanted to know your side as well. )

bkchr:
> Was just wondering where the PR is coming from, with what motivation. In other words, what is the advantage of reading from a file rather than from the command line's stdin?
> 
> (and I noted that I personally think file is better but wanted to know your side as well. )

Ahh okay :) The motivation came from using subkey :) I just don't wanted to copy my super secret uri into the clipboard :D 
Also I think it is a good pattern to support loading the URI from a file.

PR #4504, opened 2019-12-26
Improve subkey error reporting

NikVolf:
```
     Running `target/debug/subkey generate --words 1234123`

Error:
        Invalid number of words given for phrase: must be 12/15/18/21/24

```

### DIFF
--- a/bin/utils/subkey/src/main.rs
+++ b/bin/utils/subkey/src/main.rs
@@ -31,7 +31,7 @@ use sp_core::{
 };
 use sp_runtime::{traits::{IdentifyAccount, Verify}, generic::Era};
 use std::{
-	convert::{TryInto, TryFrom}, io::{stdin, Read}, str::FromStr, path::PathBuf, fs,
+	convert::{TryInto, TryFrom}, io::{stdin, Read}, str::FromStr, path::PathBuf, fs, fmt,
 };
 
 mod vanity;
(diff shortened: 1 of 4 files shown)


bin/utils/subkey/src/main.rs
```diff
@@ -236,15 +236,37 @@ fn main() {
...
 
-fn execute<C: Crypto>(matches: ArgMatches)
+#[derive(Debug, derive_more::Display, derive_more::From)]
+enum Error {
+	Static(&'static str),
+	Io(std::io::Error),
+	Formatted(String),
+}
+
+fn execute_proc<C: Crypto>(matches: ArgMatches)
```
bkchr: Just use `fn main() -> Result<(), Error>`

NikVolf: It does not really pretty-print error this way, see `nv-sk-err-msin` branch

bkchr: If you do this: https://github.com/paritytech/substrate/blob/c3a4c55cfc3a5d6f236dc184ea76f5b9ebe1eae4/client/network/src/error.rs#L51

The error message should look good.

NikVolf: Better, but still worse than here if you ask me (updated `nv-sk-err-msin` branch)

bkchr: The only different I see is that there is no new line after `Error:` in the output? I think hat is okay and less code is always better, IMHO^^

NikVolf: @bkchr updated

bin/utils/subkey/src/main.rs
```diff
@@ -270,7 +270,24 @@ fn get_uri(match_name: &str, matches: &ArgMatches) -> String {
 	}
 }
 
-fn execute<C: Crypto>(matches: ArgMatches)
+#[derive(derive_more::Display, derive_more::From)]
```
bkchr: Could you please change the function above to return an error as well? (get_uri)

PR #4505, opened 2019-12-27
WIP: prometheus exporter in substrate

nodebreaker0-0:
We are B-Harvest working on w3f granted project(https://github.com/w3f/Web3-collaboration/blob/master/grants/speculative/substrate%20x%20(prometheus%20%2B%20grafana)%20by%20B-Harvest.md). This PR is a draft implementation to request a review to be merged from parity team. We will do necessary modification if raised.

Please refer to README.md for detailed function information.

Issue #4187, opened 2019-11-22
Metrics crate integration with Grafana data source


expenses:
The [metrics](https://github.com/metrics-rs/metrics) crate provides a metrics-gathering facade similar to `log`. It would be great if we could integrate this with the Grafana data server from #3913, as we'd be able to easily switch it out for the [prometheus observer](https://github.com/metrics-rs/metrics/tree/master/metrics-observer-prometheus) or otherwise.

I'll probably end up doing this, but it should be fairly straightforward if anyone else wants to have a go.

pepyakin:
I guess this might be related 
сс @mattrutherford

mattrutherford:
Yes - looks like there is probably a lot of overlap with the tracing library that we just started to integrate https://github.com/paritytech/substrate/pull/4132 

Was thinking about adding a Receiver for substrate-tracing that can send to your Grafana data server.

Might be worth having a brain-storm some time to align on strategy for this stuff?

expenses:
Yup!

expenses:
I think we probably want to integrate the telemetry system into this as well.

marcio-diaz:
From validators channel:
> anyone working on prometheus for polkadot? the current json datasource doesnt allow us to configure alert in grafana

tomaka:
Now that Prometheus is well in place, I suppose we can close this issue?

PR #4511, opened 2019-12-30
R4R: prometheus exporter in substrate

nodebreaker0-0:
We are B-Harvest working on w3f granted project
(https://github.com/w3f/Web3-collaboration/blob/master/grants/speculative/substrate%20x%20(prometheus%20%2B%20grafana)%20by%20B-Harvest.md). 

This PR is a draft implementation to request a review to be merged from parity team. We will do necessary modification if raised.

Please refer to README.md for detailed function information.

Create new due to error:
[#4505 ]


expenses:
Hi, thanks for your PR! Have you considered integrating these changes with the existing metrics recording set up? https://github.com/paritytech/substrate/issues/4187

expenses:
@expenses To expand, it would be nice to be able to do
```rust
record_metrics!(
    a => 1,
    b => 2,
    ...
)
```

and have it go to either the built-in metrics server or the Prometheus server depending on a `--metrics-server` flag or something similar.
```

File: utils/prometheus/src/metrics.rs [hunk 1]
```diff
@@ -0,0 +1,96 @@
...
+    );
+    pub static ref P2P_NODE_DOWNLOAD: Result<IntGauge> = try_create_int_gauge(
+        "p2p_peers_receive_byte_per_sec",
+        "p2p_node_download_per_sec_byte"
+    );
+    pub static ref P2P_NODE_UPLOAD: Result<IntGauge> = try_create_int_gauge(
+        "p2p_peers_send_byte_per_sec",
+        "p2p_node_upload_per_sec_byte"
+    );
+}
```
expenses: Is there a way to create these gauges inside the `prometheus_gauge!` macro?

nodebreaker0-0:
Integrating the metric server is a good idea. but,
Your code must include Prometheus, and you will need to test a bit more because of the wasm and no_std compatibility issues.

The reason for using prometheus is due to grafana extensibility and to use the function completely, you must use prometheus code as below.
ex) prometheus :: register
record_metrics! Let's consider whether we can integrate the basic metrics we received through

But the heart of the project is v0.4, which will be integrated later.
https://github.com/nodebreaker0-0/substrate/blob/prometheus_v0.4/utils/prometheus/src/expansion.rs

Further consideration is needed for integrations that take into account extended metrics.
https://github.com/nodebreaker0-0/substrate/blob/prometheus_v0.4/utils/prometheus/README.md#seting-grafana

First, If the code runs without any problem,
I think we should consider metrics and macro integration,

utils/prometheus/src/networking.rs
```diff
@@ -0,0 +1,66 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use async_std::pin::Pin;
```
NikVolf: Can this code be moved inside a helper library in prometheus?

client/service/src/builder.rs
```diff
@@ -954,6 +956,17 @@ ServiceBuilder<
 				"bandwidth_upload" => bandwidth_upload,
 				"used_state_cache_size" => used_state_cache_size,
 			);
+			prometheus_gauge!(
+				MEMPOOL_SIZE => used_state_cache_size as u64,
```
NikVolf: it is not mempool

nodebreaker0-0: OK, I'll fix it.

utils/prometheus/src/networking.rs
nodebreaker0-0: Are you talking about writing /test code?

NikVolf: About `networking.rs`, the code seems pretty generic

nodebreaker0-0: Oh, alright,
Should I write that code in /prometheus/src/lib.rs?

utils/prometheus/src/metrics.rs
hskang9: Sure thing. I should include try_create_int_gauge in prometheus_gauge! Macro.

utils/prometheus/src/networking.rs
nodebreaker0-0: or 
I can create PR to use networking.rs code to be put in rust-prometheus’s helper library if the helper library means the cargo package. There is other way to import networking.rs in grafana-datasource module. Which would you prefer?

NikVolf: Exposing it in reusable way from grafana module and reusing here seems fine to me

client/service/src/builder.rs
nodebreaker0-0: Completion

utils/prometheus/src/networking.rs
nodebreaker0-0: complete
Imported by changing networking.rs in the grafana module to pub mod.

expenses:
> Your code must include Prometheus, and you will need to test a bit more because of the wasm and no_std compatibility issues.

This shouldn't be a problem, as I think that the metrics recording will only happen outside the runtime.

dlguddus:
> > Your code must include Prometheus, and you will need to test a bit more because of the wasm and no_std compatibility issues.
> 
> This shouldn't be a problem, as I think that the metrics recording will only happen outside the runtime.

Hi. I think the metric serving part should be separated from grafana-data-source so that we can connect our metric into the separated module. If you can separate it soon(2weeks from now)(we spent too much time on this project as now), we can migrate the metric part to newly separated metric serving module.

client/service/src/config.rs
```diff
@@ -75,6 +75,8 @@ pub struct Configuration<C, G, E = NoExtension> {
 	pub rpc_cors: Option<Vec<String>>,
 	/// Grafana data source http port. `None` if disabled.
 	pub grafana_port: Option<SocketAddr>,
+	/// Promteheus Port. `None` if disabled. and defult port 33333
```
mxinden:
```suggestion
	/// Prometheus Port.
	/// `None` if disabled and port 33333 by default.
```

utils/prometheus/README.md
```diff
@@ -0,0 +1,555 @@
+# Substrate Prometheus Node Exporter
+![grants](./photo_2019-12-13_16-32-53.jpg)
+## Introduction
+
+Prometheus is one of the most widely used monitoring tool for managing high availability services supported by [Cloud Native Computing Foundation](https://www.cncf.io/). By providing Prometheus exporter in substrate, node operators can easily adopt widely used display/alert tool such as Grafana without seting-up/operating external Prometheus agent through RPC connections. Easy access to such monitoring tools will benefit parachain develepers/operators and validators to have much higher availability quality of their services.
```
mxinden: - Nit: Given that this is within the binary, one would not speak of a Prometheus exporter, but a Prometheus metrics endpoint.

- Replacing `Prometheus agent` with `Prometheus push gateway`.

- Various typos.

```suggestion
Prometheus is one of the most widely used monitoring tool for managing high availability services supported by [Cloud Native Computing Foundation](https://www.cncf.io/). By providing Prometheus metrics in Substrate, node operators can easily adopt widely used display/alert tools such as Grafana and Alertmanager without setting-up/operating external Prometheus push gateways (which is an antipattern in the first place) through RPC connections. Easy access to such monitoring tools will benefit parachain developers/operators and validators to have much higher availability of their services.
```

utils/prometheus/README.md
```diff
@@ -0,0 +1,555 @@
...
+
+## List of Contents
```
mxinden:
```suggestion
## Table of Contents
```

utils/prometheus/README.md
```diff
@@ -0,0 +1,555 @@
...
+
+Hack Prometheus in Substrate
+ - Prometheus starter
```
mxinden: What is a `Prometheus starter`? Maybe a `Prometheus 101` or a `Prometheus primer`?

utils/prometheus/README.md
```diff
@@ -0,0 +1,555 @@
...
+
+Start Grafana
+ - Install Grafana
+
+## Substrate Dev hack
+### Prometheus starter
+
+Here is the entry point of prometheus core module in Parity Substrate.
+
+In existing sources, refer to the grapana source due to the issue of the wasm.
```
mxinden:
```suggestion
In existing sources, refer to the Grafana source due to the issue of the wasm.
```

mxinden: Can you add some more details here, I am not sure I follow the intention of the comment.

utils/prometheus/README.md
```diff
@@ -0,0 +1,555 @@
...
+        metrics::set(#$a, &$metric, $value);
+    )*
+  }
+);
+*/
+```
+
+
+
+Here is the dependancies of the module.
```
mxinden:
```suggestion
Here are the dependencies of the module.	
```

utils/prometheus/README.md
```diff
@@ -0,0 +1,555 @@
...
+## Substrate Dev hack
+### Prometheus starter
+
+Here is the entry point of prometheus core module in Parity Substrate.
+
+In existing sources, refer to the grapana source due to the issue of the wasm.
+
+utils/prometheus/src/lib.rs
+```rust
+#[macro_use]
```
mxinden: I am not sure I understand the reasoning for putting all of this source code here. It is very large as an example and difficult to maintain as there is no static analysis of code snippets within Markdown as far as I am aware.

utils/prometheus/README.md
```diff
@@ -0,0 +1,555 @@
...
+| Resource_receive_bytes_per_sec(Future) | IntGauge  |          | Operating System of bytes received                              |
+| Resource_send_bytes_per_sec(Future)    | IntGauge  |          | Operating System of bytes sent                                  |
+| Resource_cpu_use(Future)               | IntGauge  |          | Operating System cpu load                                       |
+| Resource_disk_use(Future)              | IntGauge  |          | Operating System disk use                                       |
+| validator_sign_prevote(Future)         | IntGauge  | validator addr | validator sign vote list                               	  |
+| validator_sign_precommit(Future)       | IntGauge  | validator addr | validator sign commit list                                |
+
+
+## Start Prometheus
+### Install prometheus
```
mxinden:
```suggestion
### Install Prometheus
```

utils/prometheus/README.md
```diff
@@ -0,0 +1,555 @@
...
+| p2p_peer_send_bytes_per_sec            | IntGauge  |          | number of bytes sent to a given peer                            |
+| Resource_receive_bytes_per_sec(Future) | IntGauge  |          | Operating System of bytes received                              |
+| Resource_send_bytes_per_sec(Future)    | IntGauge  |          | Operating System of bytes sent                                  |
+| Resource_cpu_use(Future)               | IntGauge  |          | Operating System cpu load                                       |
+| Resource_disk_use(Future)              | IntGauge  |          | Operating System disk use                                       |
+| validator_sign_prevote(Future)         | IntGauge  | validator addr | validator sign vote list                               	  |
+| validator_sign_precommit(Future)       | IntGauge  | validator addr | validator sign commit list                                |
+
+
+## Start Prometheus
```
mxinden: I don't think this is worth duplicating here. I would suggest to link to the Prometheus docs.

utils/prometheus/README.md
```diff
@@ -0,0 +1,555 @@
...
+sudo apt-get install grafana
+sudo service grafana-server start
+./prometheus
+```
+
+You can visit `http://localhost:3000/` to open grafana and create your own dashboard.
+
+> Tips: The default username and password are both admin. We strongly recommend immediately changing your username & password after login
+
+### Seting Grafana
```
mxinden:
```suggestion
### Setting up Grafana
```

utils/prometheus/README.md
```diff
@@ -0,0 +1,555 @@
...
+
+> The above example, you can save `prometheus.yml` at `~/volumes/prometheus` on your host machine
+
+You can visit `http://localhost:9090` to see prometheus data.
+
+
+
+## Start Grafana
+### Install Grafana
+https://grafana.com/docs/installation/debian/
```
mxinden: Again, I don't think it is worth explaining how to install Grafana nor Prometheus.

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+    let metric_families = prometheus::gather();
+    let mut buffer = vec![];
+    let encoder = TextEncoder::new();
+    encoder.encode(&metric_families, &mut buffer).unwrap();
+    Response::builder()
+      .status(StatusCode::OK)
+      .header("Content-Type", encoder.format_type())
+      .body(Body::from(buffer))
+      .map_err(Error::Http)
+      //.expect("Sends OK(200) response with one or more data metrics")
```
mxinden: Can you remove this?

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+  } else {
+    Response::builder()
+      .status(StatusCode::NOT_FOUND)
+      .body(Body::from("Not found."))
+      .map_err(Error::Http)
+      //.expect("Sends NOT_FOUND(404) message with no data metric")
```
mxinden: Same.

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+		async_std::task::spawn(future);
+	}
+}
+/// Initializes the metrics context, and starts an HTTP server
+/// to serve metrics.
+#[cfg(not(target_os = "unknown"))]
+pub  async fn init_prometheus(mut prometheus_addr: SocketAddr) -> Result<(), Error>{
+  use async_std::{net, io};
+  use grafana_data_source::networking::Incoming;
+	let listener = loop {
```
mxinden: formatting

mxinden: So in the worst case this is an endless loop?

utils/prometheus/src/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+///histogram Metrics a value in injection.
+pub fn set_histogram(histogram: &Result<Histogram>, value: f64) {
+    if let Ok(histogram) = histogram {
+        histogram.observe(value)
+    }
+}
+//All of the metrics in the prometheus are managed by the lazy_static.
+
+lazy_static! {
+    pub static ref FINALITY_HEIGHT: Result<IntGauge> = try_create_int_gauge(
```
mxinden: I think metric definitions should live close to the logic that they monitor. What do you think?

client/service/src/builder.rs
```diff
@@ -1103,7 +1115,17 @@ ServiceBuilder<
 				.then(|_| Ok(()))));
 			telemetry
 		});
-
+		// prometheus init
+		if let Some(port) = config.prometheus_port {
+			let future = select(
+					sc_prometheus::init_prometheus(port).boxed()
```
expenses: Bad style, use `sc_prometheus::init_prometheus(port).boxed(),`

utils/prometheus/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+log = "0.4.8"
+prometheus = { version = "0.7", features = ["nightly", "process"]}
+tokio = "0.2"
+futures-util = { version = "0.3.1", default-features = false, features = ["io"] }
+sp-runtime = { package = "sp-runtime",path = "../../primitives/runtime" }
+derive_more = "0.99"
+grafana-data-source = { version = "2.0.0", path = "../grafana-data-source" }
+
+[target.'cfg(not(target_os = "unknown"))'.dependencies]
+async-std = { version = "1.0.1", features = ["unstable"] }
```
expenses: Missing newline at EOF

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+use futures_util::{FutureExt,future::{Future}};
+use hyper::http::StatusCode;
+use hyper::Server;
+use hyper::{Body, Request, Response, service::{service_fn, make_service_fn}};
+pub use prometheus::{Encoder, HistogramOpts, Opts, TextEncoder};
+pub use prometheus::{Histogram, IntCounter, IntGauge};
+pub use sp_runtime::traits::SaturatedConversion;
+use std::net::SocketAddr;
+//#[cfg(not(target_os = "unknown"))]
+//mod networking;
```
expenses: Remove this commented out code

utils/prometheus/src/lib.rs
expenses: Yeah, but it's also what grafana does at the moment: https://github.com/paritytech/substrate/blob/ae46db60594ab195a2b5c90679b55fb5633489c3/utils/grafana-data-source/src/server.rs#L103-L122 so I don't think it's too bad.

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+#[macro_export]
+macro_rules! prometheus(
+  ($($a: expr; $metric:expr => $value:expr),*) => {
+    use $crate::{metrics::*};
+    $(
+        metrics::set(#$a, &$metric, $value);
+    )*
+  }
+);
+*/
```
expenses: Missing newline at EOF

utils/prometheus/src/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+/// histogram type metrics generation function
+pub fn try_create_histogram(name: &str, help: &str) -> Result<Histogram> {
+    let opts = HistogramOpts::new(name, help);
+    let histogram = Histogram::with_opts(opts)?;
+    prometheus::register(Box::new(histogram.clone()))?;
+    Ok(histogram)
+}
+
+///Gauge Metrics a value in injection.
```
expenses: Put a space between the slashes and the comment: `/// Gauge Metrics a value in injection.`

utils/prometheus/src/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+pub fn set_gauge(gauge: &Result<IntGauge>, value: u64) {
+    if let Ok(gauge) = gauge {
+        gauge.set(value as i64);
+    }
+}
+///histogram Metrics a value in injection.
```
expenses: Same here

utils/prometheus/src/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+pub fn set_histogram(histogram: &Result<Histogram>, value: f64) {
+    if let Ok(histogram) = histogram {
+        histogram.observe(value)
+    }
+}
+//All of the metrics in the prometheus are managed by the lazy_static.
```
expenses: Same here

utils/prometheus/src/metrics.rs
expenses: Agreed. They should either be defined in the metrics macro, or somewhere else in the module that they're called in.

utils/prometheus/src/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub static ref FINALITY_HEIGHT: Result<IntGauge> = try_create_int_gauge(
+        "consensus_finality_block_height_number",
+        "block is finality HEIGHT"
+
+    );
+    pub static ref BEST_HEIGHT: Result<IntGauge> = try_create_int_gauge(
+        "consensus_best_block_height_number",
+        "block is best HEIGHT"
+    );
+
```
expenses: Inconsistent newlines between `pub static ref` statements.

utils/prometheus/src/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub static ref BLOCK_INTERVAL_SECONDS: Result<Histogram> = try_create_histogram(
+        "consensus_block_interval_seconds",
+        "Time between this and last block(Block.Header.Time) in seconds"
+    );
```
expenses: Either add a newline here, or remove the above one.

utils/prometheus/src/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub static ref P2P_NODE_DOWNLOAD: Result<IntGauge> = try_create_int_gauge(
+        "p2p_peers_receive_byte_per_sec",
+        "p2p_node_download_per_sec_byte"
+    );
+    pub static ref P2P_NODE_UPLOAD: Result<IntGauge> = try_create_int_gauge(
+        "p2p_peers_send_byte_per_sec",
+        "p2p_node_upload_per_sec_byte"
+    );
+}
```
expenses: Newline at EOF

utils/prometheus/src/lib.rs
nodebreaker0-0: Which formatting?

utils/prometheus/src/metrics.rs
nodebreaker0-0: Which folder should I go to? or which name should edit ? I defined metrics on metrics.rs, but I don't know if it's wrong.

See tendermint repo

client/service/src/builder.rs
nodebreaker0-0: Why is it bad style? How should I change it?

client/service/src/config.rs
nodebreaker0-0: complete

utils/prometheus/README.md
nodebreaker0-0: complete

utils/prometheus/README.md
nodebreaker0-0: complete

utils/prometheus/README.md
nodebreaker0-0: complete

utils/prometheus/README.md
nodebreaker0-0: complete

utils/prometheus/src/lib.rs
nodebreaker0-0: complete

utils/prometheus/src/lib.rs
nodebreaker0-0: complete

utils/prometheus/Cargo.toml
nodebreaker0-0: complete

utils/prometheus/src/lib.rs
nodebreaker0-0: complete

utils/prometheus/src/lib.rs
nodebreaker0-0: complete

utils/prometheus/src/metrics.rs
nodebreaker0-0: complete

utils/prometheus/src/metrics.rs
nodebreaker0-0: complete

utils/prometheus/src/metrics.rs
nodebreaker0-0: complete

utils/prometheus/src/metrics.rs
nodebreaker0-0: complete

utils/prometheus/src/metrics.rs
nodebreaker0-0: complete

utils/prometheus/src/metrics.rs
nodebreaker0-0: complete

utils/prometheus/src/metrics.rs
nodebreaker0-0: complete

utils/prometheus/src/metrics.rs
nodebreaker0-0: complete

expenses:
I've made a PR on this fork with a few changes :^) https://github.com/nodebreaker0-0/substrate/pull/2

client/tracing/src/lib.rs
```diff
@@ -291,9 +290,6 @@ fn send_telemetry(span_datum: SpanDatum) {
...
 }
 
-fn send_grafana(span_datum: SpanDatum) {
-	let name = format!("{}::{}", span_datum.target, span_datum.name);
-	if let Err(e) = record_metrics!(&name => span_datum.overall_time.as_nanos(),) {
-		log::warn!("Unable to send metrics to grafana: {:?}", e);
-	}
+fn send_prometheus(span_datum: SpanDatum) {
+	unimplemented!()
 }
```
expenses: @mxinden do you know of a way that we could implement this API? It seems like prometheus gauges etc are intended to created as static refs etc.

primitives/consensus/babe/src/inherents.rs
```diff
@@ -32,59 +32,59 @@ pub const INHERENT_IDENTIFIER: InherentIdentifier = *b"babeslot";
...
+		let slot_number = timestamp / self.slot_duration;
+		inherent_data.put_data(INHERENT_IDENTIFIER, &slot_number)
+	}
 
-    fn error_to_string(&self, error: &[u8]) -> Option<String> {
-        Error::decode(&mut &error[..]).map(|e| e.into_string()).ok()
-    }
+	fn error_to_string(&self, error: &[u8]) -> Option<String> {
+		Error::decode(&mut &error[..]).map(|e| e.into_string()).ok()
+	}
```
expenses: Please revert this file back to paritytech/master.

utils/prometheus/README.md
expenses: Agreed

utils/prometheus/src/lib.rs
expenses: @nodebreaker0-0 this code shouldn't be indented this far. Should be a single tab.

primitives/consensus/babe/src/inherents.rs
nodebreaker0-0: Ask for your PR back?

Or do you need a rebase to paritytech/master?

nodebreaker0-0: ah 
 only need to revert this file.

client/tracing/src/lib.rs
nodebreaker0-0: My guess is that you can't implement it without creating a macro.

Or you can simply make a static reference.
https://github.com/tikv/rust-prometheus/blob/master/examples/example_edition_2018.rs

primitives/consensus/babe/src/inherents.rs
nodebreaker0-0: complete

nodebreaker0-0:
I think I've accidentally requested a review of my Github operation, sorry.

expenses:
I'm generally pretty happy with this PR now. I wanna get the opinions of @pepyakin, @mxinden and @mattrutherford before approving though.

client/tracing/src/lib.rs
mattrutherford: I don't yet understand fully how prometheus works, but to fully support this I think we would need to be able to: 

- batch together all values (spans) collected since last scrape 
- partition by target and name (by the use of labels perhaps).  

With all runtime targets currently in the code enabled, we can be approaching 200 measurements per second just across the runtime.  Is this feasible to do (while preserving timestamp) - the alternative would be to consider to lose some detail and aggregate the values between scrapes, which would fix the first point above.

expenses: Hmm. I'll have a go at this later.

mxinden:
I am sorry for the delay. I will give this another in-depth review in the upcoming days. Please don't merge this pull request yet.

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+
+pub use prometheus::core::{
+	GenericGauge as Gauge, AtomicF64 as F64, AtomicI64 as I64, AtomicU64 as U64
+};
+pub use lazy_static::lazy_static;
+
+pub fn create_gauge<T: Atomic + 'static>(name: &str, description: &str) -> Gauge<T> {
+	let opts = Opts::new(name, description);
+	let gauge = Gauge::with_opts(opts).expect("Creating Gauge Failed");
+	prometheus::register(Box::new(gauge.clone())).expect("Registering gauge failed");
```
mxinden: This will use the default (global) registry. While convenient I think we should not use it.

client/service/src/builder.rs
```diff
@@ -53,8 +53,42 @@ use sysinfo::{get_current_pid, ProcessExt, System, SystemExt};
 use sc_telemetry::{telemetry, SUBSTRATE_INFO};
 use sp_transaction_pool::{TransactionPool, TransactionPoolMaintainer};
 use sp_blockchain;
-use grafana_data_source::{self, record_metrics};
+use prometheus_endpoint::{create_gauge, Gauge, U64, F64};
 
+prometheus_endpoint::lazy_static! {
+	pub static ref FINALITY_HEIGHT: Gauge<U64> = create_gauge(
+		"consensus_finality_block_height_number",
```
mxinden: These should be prefixed with something substrate specific. E.g. `substrate_consensus_finality[...]`. Or using `dots` like https://github.com/paritytech/dotexporter does.

client/tracing/src/lib.rs
mxinden: > be approaching 200 measurements

@mattrutherford 200 measurements are not an issue for Prometheus. Prometheus can easily handle couple of mega bytes of metric data per scrape. See https://github.com/kubernetes/kube-state-metrics/issues/498 for some numbers.

> batch together all values (spans) collected since last scrape

This sounds more like we want a Histogram, right?

---

Would it be possible to tackle the effort within `client/tracing` as a follow up pull request? I would like to keep this one small. What do you think @mattrutherford @expenses @nodebreaker0-0?

utils/grafana-data-source/src/lib.rs
```diff
@@ -31,7 +31,7 @@ use parking_lot::RwLock;
 mod types;
```
mxinden: I am a bit confused. In some places we replace the Grafana logic, in some places it still exists. We should be consistent, me advocating for replacing Grafana. What do you all think?

utils/prometheus/README.md
```diff
@@ -0,0 +1,117 @@
...
+ - Install Grafana
+
+## Metrics
+
+substrate can report and serve the Prometheus metrics, which in their turn can be consumed by Prometheus collector(s).
+
+This functionality is disabled by default.
+
+To enable the Prometheus metrics, set in your cli command (--prometheus-addr,--prometheus-port ). 
+Metrics will be served under /metrics on 33333 port by default.
```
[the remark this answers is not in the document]
expenses: Yup

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+	fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+		match self {
+			Error::Hyper(error) => Some(error),
+			Error::Http(error) => Some(error),
+			Error::Io(error) => Some(error)
+		}
+	}
+}
+
+async fn request_metrics(req: Request<Body>) -> Result<Response<Body>, Error> {
```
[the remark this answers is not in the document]
expenses: because hyper needs you to pass in a function that returns a future.

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+			Ok(listener) => {
+				log::info!("Prometheus server started at {}", prometheus_addr);
+				break listener
+			},
+			Err(err) => match err.kind() {
+				io::ErrorKind::AddrInUse | io::ErrorKind::PermissionDenied if prometheus_addr.port() != 0 => {
+					log::warn!(
+						"Prometheus server to already {} port.", prometheus_addr.port()
+					);
+					prometheus_addr.set_port(0);
```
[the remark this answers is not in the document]
expenses: I believe we changed this with the embedded db because it wasn't clear what the behaviour should be in the case that it can't connect to the port.

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+pub struct Executor;
+
+#[cfg(not(target_os = "unknown"))]
+impl<T> hyper::rt::Executor<T> for Executor
+	where
+		T: Future + Send + 'static,
+		T::Output: Send + 'static,
+{
+	fn execute(&self, future: T) {
+		async_std::task::spawn(future);
```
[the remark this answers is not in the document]
expenses: This is so that we can use either tokio or a different executor.

expenses:
@mxinden
> Would it be possible to tackle the effort within client/tracing as a follow up pull request? I would like to keep this one small. What do you think @mattrutherford @expenses @nodebreaker0-0?

Yeah, I think that'd be the best decision here.

utils/grafana-data-source/src/lib.rs
expenses: Agreed.

client/tracing/src/lib.rs
mattrutherford: a follow up PR sounds sensible -  am I right in thinking that Histogram does not preserve individual observations? If so, that's not really what we want, because eg. there can be a particular individual timing of a span that we care about, with a specific set of inputs that occurs at a particular time and we want to preserve all that. Eg we can create a span as such:

```rust
let span = tracing::span!(
	tracing::Level::INFO,
	"execute_and_import_block",
	?hash,
	finalized,
	enact_state
);
```

So to just provide aggregates we lose much of the useful information.

utils/prometheus/Cargo.toml
```diff
@@ -9,14 +9,10 @@ edition = "2018"
 [dependencies]
 log = "0.4.8"
 hyper = { version = "0.13.1", default-features = false, features = ["stream"] }
+prometheus = { version = "0.7", features = ["nightly", "process"]}
```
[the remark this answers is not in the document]
expenses: It compiles for me on stable

utils/prometheus/src/lib.rs
expenses: What do you suggest we should do instead?

client/tracing/src/lib.rs
mxinden: > am I right in thinking that Histogram does not preserve individual observations?

Yes, that is correct, none of the metric types (counters, gauges, histograms, summary) preserve individual observations and that is very much by design for *metrics* compared to *logging* and *tracing*. *Metrics* make a trade off between the amount of metrics one can collect (within the millions per scrape across ones fleet) and the level of details per observation. I think https://peter.bourgon.org/blog/2017/02/21/metrics-tracing-and-logging.html is a great write-up of the difference for further details.

While this might make the *metrics* effort unsuitable for  a crate called *tracing*, maybe we can introduce an *observability* (the combination of the three - metrics, logging, tracing) in the future, allowing developers to instrument their code easily based on all 3 pillars.

Does all of this make sense @mattrutherford?

client/service/src/builder.rs
mxinden: For more details on metric nameing, see https://prometheus.io/docs/practices/naming/.

client/tracing/src/lib.rs
mattrutherford: I agree it's good to be clear on terminology, but that wasn't my main concern - only that if we're ripping out the Grafana server as part of the introduction of this Prometheus feature, is whether we'd lose this functionality to plot individual observations* - however that seems unavoidable based on what you said.

*For my use-case it's OK because I use the Telemetry Receiver in substrate-tracing to send the data to substrate-analytics which uses a PostgreSQL datasource, which in turn can be queried by Grafana (because I want to archive the data); however this is not trivial to set up, so the question is - do we really want to kill substrate Grafana server yet? It's a relatively new feature so maybe not many people use it, but I think it's something we should consider as part of our decision to remove it. Particularly as the current primary use-case for `substrate-tracing` is for profiling and that might be handy for developers to have any easy way to look at this.

Pinging @marcio-diaz @shawntabrizi @DarkEld3r in case you have opinion on this

client/service/src/builder.rs
expenses: The best solution imo is to use a `Registry` with a custom prefix: https://docs.rs/prometheus/0.7.0/prometheus/struct.Registry.html#method.new_custom

utils/prometheus/README.md
```diff
@@ -0,0 +1,117 @@
+# Substrate Prometheus Node Exporter
+![grants](./photo_2019-12-13_16-32-53.jpg)
```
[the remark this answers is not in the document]
nodebreaker0-0: This is a request from the W3F, which can only be placed in a private repo under consultation

client/tracing/src/lib.rs
expenses: Imo the Grafana server in its current state isn't really _good_ at doing anything. If Prometheus isn't going to work for tracing, we should probably find something that is. I'm happy enough to leave the Grafana server in for the moment until that happens.

utils/prometheus/README.md
nodebreaker0-0: ok, Agreed !Milestone 2 will provide a fully configured dashboard, so we will consider removing it in this release and tagging grafana official links the next release.

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+mod networking;
+
+pub use prometheus::core::{
+	GenericGauge as Gauge, AtomicF64 as F64, AtomicI64 as I64, AtomicU64 as U64
+};
+pub use lazy_static::lazy_static;
+
+pub fn create_gauge<T: Atomic + 'static>(name: &str, description: &str) -> Gauge<T> {
+	let opts = Opts::new(name, description);
+	let gauge = Gauge::with_opts(opts).expect("Creating Gauge Failed");
```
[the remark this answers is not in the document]
nodebreaker0-0: I agree with you.

client/tracing/src/lib.rs
mattrutherford: ok, good to know - and just to be clear that I have no problem removing it, if it's not going to be a big detriment to people - just wanted to make sure everyone was aware of full implications

client/service/src/builder.rs
nodebreaker0-0: Since you can use the metric in polkadot later, **dot_** is better than **substrate_**.
try to prefix dot_ all Do you agree?

utils/grafana-data-source/src/lib.rs
nodebreaker0-0: me too Agreed.

utils/prometheus/README.md
nodebreaker0-0: I agree. I'm dropping the table

utils/prometheus/src/lib.rs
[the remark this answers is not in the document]
nodebreaker0-0: Personal thinking is considering restructuring the rules.

1. By default, prometheus is disabled

2. Option to configure --prometheus-on

3. If you do not specify the port, use port 9955. If you are already using it, you must specify the port.

What do you think? @mxinden @expenses

utils/prometheus/README.md
```diff
@@ -0,0 +1,117 @@
+# Substrate Prometheus Node Exporter
```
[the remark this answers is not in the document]
nodebreaker0-0: complete

utils/prometheus/README.md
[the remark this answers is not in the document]
nodebreaker0-0: complete

utils/prometheus/README.md
```diff
@@ -0,0 +1,117 @@
...
+## Metrics
+
+substrate can report and serve the Prometheus metrics, which in their turn can be consumed by Prometheus collector(s).
+
+This functionality is disabled by default.
+
+To enable the Prometheus metrics, set in your cli command (--prometheus-addr,--prometheus-port ). 
+Metrics will be served under /metrics on 33333 port by default.
+
+### List of available metrics
```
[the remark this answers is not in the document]
nodebreaker0-0: complete

utils/prometheus/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+
+Consensus metrics, namespace: `substrate`
+
+| **Name**                               | **Type**  | **Tags** | **Description**                                                 |
+| -------------------------------------- | --------- | -------- | --------------------------------------------------------------- |
+| consensus_finality_block_height_number | IntGauge  |          | finality Height of the chain                                    |
```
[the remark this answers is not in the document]
nodebreaker0-0: complete

utils/prometheus/README.md
nodebreaker0-0: complete

utils/prometheus/README.md
nodebreaker0-0: complete

utils/prometheus/README.md
```diff
@@ -0,0 +1,117 @@
...
+            labels:
+              instance: local-validator
+```
+
+> Note：value of targets is ip:port which used by substrate monitor 
+
+### Start Prometheus
+
+```bash
+cd <prometheus file>
```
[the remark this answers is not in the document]
nodebreaker0-0: complete

utils/prometheus/README.md
nodebreaker0-0: complete

utils/prometheus/src/lib.rs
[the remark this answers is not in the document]
nodebreaker0-0: What do you think about removal? @expenses

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+		let listener = net::TcpListener::bind(&prometheus_addr).await;
+		match listener {
+			Ok(listener) => {
+				log::info!("Prometheus server started at {}", prometheus_addr);
+				break listener
+			},
+			Err(err) => match err.kind() {
+				io::ErrorKind::AddrInUse | io::ErrorKind::PermissionDenied if prometheus_addr.port() != 0 => {
+					log::warn!(
+						"Prometheus server to already {} port.", prometheus_addr.port()
```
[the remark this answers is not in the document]
nodebreaker0-0: complete

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+			}
+		}
+	};
+	let service = make_service_fn(|_| {
+		async {
+			Ok::<_, Error>(service_fn(request_metrics))
+		}
+	});
+
+	let _server = Server::builder(Incoming(listener.incoming()))
```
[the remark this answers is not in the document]
nodebreaker0-0: complete

client/tracing/src/lib.rs
dlguddus: I think tracing other's nodes is not (should not) the purpose of the prometheus exporter. Extracting information from nodes operated by individual entities(validators or full-node operators) is not sounding right as a perspective of information privacy.

I guess tracing functionality should remain in telemetry, and it should be also "off as default" for privacy of node operators.

client/service/src/builder.rs
expenses: The way I've got it set up in [my branch](https://github.com/paritytech/substrate/tree/ashley-prometheus) is that you can either pass in a Prometheus registry or have substrate create one for you, and this does the prefixing. This means that polkadot could pass in a different registry that prefixes all of substrate's metrics with `dot_` instead.

ddorgan:
Just some comments on what would be useful stats to provide. So first off to replace dotexporter (which is currently used to scrape some basic metrics) we would need the following info, unless anybody has any comments or suggested changes:

```
dot_chain_block_number{name="parity-polkadot",version="0.7.17",chain="Kusama CC3",block="finalized"} 655714
dot_chain_block_number{name="parity-polkadot",version="0.7.17",chain="Kusama CC3",block="head"} 655780
dot_peer_count{name="parity-polkadot",version="0.7.17",chain="Kusama CC3"} 275
dot_shouldHavePeers{name="parity-polkadot",version="0.7.17",chain="Kusama CC3"} 1
dot_isSyncing{name="parity-polkadot",version="0.7.17",chain="Kusama CC3"} 0
dot_specVersion{name="parity-polkadot",version="0.7.17",chain="Kusama CC3"} 1040
dot_rpc_healthy{name="parity-polkadot",version="0.7.17",chain="Kusama CC3"} 1
```

Some other items that came up recently that would be useful:
- Some afg stats, like prevotes counts and what round the host is on.
- Maybe a label on the existing head and final or a new value for the local epoch time when that block was imported. So we can easily detect stalls locally on the host and have a watchdog process attempt a restart of polkadot/substrate (and a warn message to alertmanager) at least once before our monitoring system contacts the oncall person if it detects there's still a failure. But 90% of the time a restart works.
- Some stats for memory allocation for different modules
-  Some stats for cpu usage for different modules/threads

Just a few things to start with!

expenses:
@ddorgan IMO, the best move is to to get this base PR merged first, then start implementing metrics.

utils/prometheus/Cargo.toml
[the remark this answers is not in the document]
expenses: We should check if these features are needed though.

expenses:
@mxinden would like your review again.

mxinden:
I will give this another review tomorrow. @nodebreaker0-0 @expenses thanks for all the fixes since the last review.

client/service/src/builder.rs
```diff
@@ -51,7 +51,71 @@ use sysinfo::{get_current_pid, ProcessExt, System, SystemExt};
...
+				"Number of network gossip peers",
+				registry
+			)?,
+			tx_count: create_gauge(
+				"transaction_count",
+				"Number of transactions",
+				registry
+			)?,
+			node_memory: create_gauge(
+				"memory_usage",
```
mxinden: As a convention Prometheus metric names should:

1. Use base units (See docs for details https://prometheus.io/docs/practices/naming/#base-units)

2. Have the unit as a suffix in the name of the metric (See docs for details https://prometheus.io/docs/practices/naming/#metric-names)

client/service/src/builder.rs
```diff
@@ -51,7 +51,71 @@ use sysinfo::{get_current_pid, ProcessExt, System, SystemExt};
...
+				"Node memory usage",
+				registry
+			)?,
+			node_cpu: create_gauge(
+				"cpu_usage",
```
mxinden: See `memory_usage`.

client/service/src/builder.rs
```diff
@@ -51,7 +51,71 @@ use sysinfo::{get_current_pid, ProcessExt, System, SystemExt};
...
+				"Node CPU usage",
+				registry
+			)?,
+			node_download: create_gauge(
+				"receive_byte_per_sec",
```
mxinden: Not a strong opinion: I would rename this to

```suggestion
				"received_per_sec_bytes",
```

to have the unit as a suffix.

client/service/src/builder.rs
```diff
@@ -51,7 +51,71 @@ use sysinfo::{get_current_pid, ProcessExt, System, SystemExt};
...
+				"Received bytes per second",
+				registry
+			)?,
+			node_upload: create_gauge(
+				"sent_byte_per_sec",
```
mxinden: See `receive_byte_per_sec`.

client/service/src/builder.rs
```diff
@@ -51,7 +51,71 @@ use sysinfo::{get_current_pid, ProcessExt, System, SystemExt};
...
 use sp_transaction_pool::{TransactionPool, TransactionPoolMaintainer};
 use sp_blockchain;
-use grafana_data_source::{self, record_metrics};
+use prometheus_exporter::{create_gauge, Gauge, U64, F64, Registry, PrometheusError};
+
+struct ServiceMetrics {
+	finality_height: Gauge<U64>,
+	best_height: Gauge<U64>,
+	peers_num: Gauge<U64>,
+	tx_count: Gauge<U64>,
```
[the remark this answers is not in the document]
mxinden: In case this is indeed a monotonically increasing counter, the convention is to suffix it with `total`, thus `tx_total` or `tx_count_total`, preferring the former.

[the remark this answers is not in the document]
mxinden: Is the abbreviation necessary here? If not, I would suggest using `transaction` instead of `tx`. Prometheus will compress the data very well on disc, thus no need for brevity here.

client/service/src/builder.rs
```diff
@@ -51,7 +51,71 @@ use sysinfo::{get_current_pid, ProcessExt, System, SystemExt};
...
+	node_download: Gauge<U64>,
+	node_upload: Gauge<U64>,
+	sync_target: Gauge<U64>,
+}
+
+impl ServiceMetrics {
+	fn register(registry: &Registry) -> Result<Self, PrometheusError> {
+		Ok(Self {
+			finality_height: create_gauge(
+				"finality_block_height_number",
```
mxinden: Instead of having two distinct metrics `finality_block_height_number` and `best_block_height_number`, one could also have a single metric `block_height_number` with two label instances `finality` and `best`. With this format one can easier run queries against the two in combination and display them in a single graph within the Prometheus web interface. (Grafana can display them in a single graph anyways.)

What do you think?

nodebreaker0-0: Are there two associations? For example
In the 10th best block, the finality block must be the 5th ..?

If there is finality block information in the best block, it is better to integrate it using gaugevec.

utils/prometheus/README.md
```diff
@@ -0,0 +1,45 @@
+# Substrate Prometheus Exporter
```
mxinden: What do you think of the following changes?

```diff
diff --git a/utils/prometheus/README.md b/utils/prometheus/README.md
index c6746d573..d7c895c3f 100644
--- a/utils/prometheus/README.md
+++ b/utils/prometheus/README.md
@@ -2,44 +2,25 @@
 
 ## Introduction
 
-Prometheus is one of the most widely used monitoring tool for managing highly available services supported by [Cloud Native Computing Foundation](https://www.cncf.io/). By providing Prometheus metrics in Substrate, node operators can easily adopt widely used display/alert tools such as Grafana and Alertmanager without setting-up/operating external Prometheus push gateways (which is an antipattern in the first place) through RPC connections. Easy access to such monitoring tools will benefit parachain developers/operators and validators to have much higher availability of their services.
+[Prometheus](https://prometheus.io/) is one of the most widely used monitoring
+tools for managing highly available services supported by [Cloud Native
+Computing Foundation](https://www.cncf.io/). By providing Prometheus metrics in
+Substrate, node operators can easily adopt widely used display/alert tools such
+as [Grafana](https://grafana.com/) and
+[Alertmanager](https://prometheus.io/docs/alerting/alertmanager/). Easy access
+to such monitoring tools will benefit parachain developers/operators and
+validators to have much higher availability of their services.
 
-## Table of Contents
+Metrics will be served under `/metrics` on TCP port 9615 by default.
 
-Metrics
 
-Start Prometheus
- - Edit Prometheus config file
+## Quick Start
 
+1. From the root of the repository start Substrate `cargo run --release`.
 
-## Metrics
+2. In another terminal run `curl localhost:9615/metrics` to retrieve the
+   metrics.
 
-Substrate can report and serve the Prometheus metrics, which in their turn can be consumed by Prometheus collector(s).
 
-This functionality is disabled by default.
-
-Metrics will be served under /metrics on 9615 port by default.
-
-
-## Start Prometheus
-### Edit Prometheus config file
-
-You can visit [prometheus.yml](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus.yml) to download default `prometheus.yml`.
-
-Then edit `prometheus.yml` and add `jobs` :
-
-```yaml
-      - job_name: kusama
-          static_configs:
-          - targets: ['localhost:9615']
-            labels:
-              instance: local-validator
-```
-
-> Note：value of targets is ip:port which used by substrate monitor 
-
-### Start Prometheus
-
-> The above example, you can save `prometheus.yml` at `~/volumes/prometheus` on your host machine
-
-You can visit `http://localhost:9090` to see prometheus data.
+To learn how to configure Prometheus see the Prometheus [Getting
+Started](https://prometheus.io/docs/prometheus/latest/getting_started/) guide.
```

utils/prometheus/Cargo.toml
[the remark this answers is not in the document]
mxinden: Compiles and exposes metrics just fine without `nightly`, @nodebreaker0-0 can you please remove the flag, unless we are missing something.

utils/prometheus/src/lib.rs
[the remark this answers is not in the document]
expenses: If you switch the executor added in #4688 to `async_std::task::spawn` and didn't have both the networking code and the custom `Executor`, you get a panic:

`
Thread 'async-std/executor' panicked at 'no current reactor', /home/expenses/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.4/src/io/driver/mod.rs:243
`

I'm in favour of leaving the `Executor` and networking code where it is for now.

mxinden:
@nodebreaker0-0 would you mind merging https://github.com/nodebreaker0-0/substrate/pull/5 which updates this branch to current master?

client/tracing/src/lib.rs
mattrutherford: OK - I can see what the misunderstanding is here and it warrants some more explanation - tracing is only capturing local data; which we then have the option to send somewhere, either Log (output tracing data via logger), Grafana server (built-in) or Telemetry. When using telemetry to send the tracing data, it is not expected (or desired) to send to the default telemetry url, so we override that via the cli, with eg: `--telemetry-url 'ws://localhost:8080 9'` to send the data to an analytics server (normally, that we control).

mxinden:
@expenses @nodebreaker0-0 what do you think of https://github.com/nodebreaker0-0/substrate/pull/6, using the global executor to spawn futures per connection?

nodebreaker0-0:
> @nodebreaker0-0 would you mind merging [nodebreaker0-0#5](https://github.com/nodebreaker0-0/substrate/pull/5) which updates this branch to current master?

nodebreaker0-0:
> @nodebreaker0-0 would you mind merging [nodebreaker0-0#5](https://github.com/nodebreaker0-0/substrate/pull/5) which updates this branch to current master?

PR completed

utils/prometheus/README.md
nodebreaker0-0: OK, simple is the best

client/service/src/builder.rs
mxinden: @nodebreaker0-0 I might not have explained that very well. Let me add some examples:

Instead of doing

```
finality_block_height_number{target="foo"} 11
best_block_height_number{target="foo"} 20
```

we could as well do

```
block_height_number{status="finalized", target="foo"} 11
block_height_number{status="best", target="foo"} 20
```

nodebreaker0-0: Good feedback, I think you can use vec
https://github.com/nodebreaker0-0/substrate/pull/7

I think we can discuss at.

expenses:
@mxinden How close do you think we are to being able to merge this?

utils/prometheus/src/lib.rs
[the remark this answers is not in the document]
mxinden: @expenses agreed.

nodebreaker0-0:
no-prometheus flag add!

nodebreaker0-0:
Fixed input validation in / metrics and enabling --no-prometheus.

utils/prometheus/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+		encoder.encode(&metric_families, &mut buffer).unwrap();
+
+		Response::builder()
+			.status(StatusCode::OK)
+			.header("Content-Type", encoder.format_type())
+			.body(Body::from(buffer))
+			.map_err(Error::Http)
+	} else {
+		Response::builder()
+		.status(StatusCode::NOT_FOUND)
```
mxinden: This line and the ones below would need to be indented. See `Response::builder().status(...)` above.

client/tracing/src/lib.rs
```diff
@@ -63,8 +62,8 @@ pub enum TracingReceiver {
 	Log,
 	/// Output to telemetry
 	Telemetry,
-	/// Output to Grafana
-	Grafana,
+	/// Output to Prometheus endpoint
```
mxinden: As far as I can tell, there has not been any objections to removing any Prometheus related functionality from the `tracing` crate.

client/cli/src/params.rs
```diff
@@ -476,9 +476,15 @@ pub struct RunCmd {
...
 	pub rpc_cors: Option<Cors>,
 
-	/// Specify Grafana data source server TCP Port.
-	#[structopt(long = "grafana-port", value_name = "PORT")]
-	pub grafana_port: Option<u16>,
+	/// Specify Prometheus endpoint TCP Port.
+	#[structopt(long = "prometheus-port", value_name = "PORT")]
+	pub prometheus_port: Option<u16>,
+
+	/// Disable connecting to the Substrate prometheus.
```
mxinden:
```suggestion
	/// Do not expose a Prometheus metric endpoint.
```

client/cli/src/params.rs
```diff
@@ -476,9 +476,15 @@ pub struct RunCmd {
...
+	///
+	/// prometheus is on by default on global chains.
```
mxinden:
```suggestion
	/// Prometheus metric endpoint is enabled by default.
```

client/cli/src/params.rs
mxinden: One does not disable others to connect to it, but the exposition in itself.

client/tracing/src/lib.rs
mattrutherford: That makes sense to me, at least for now.

client/cli/src/params.rs
nodebreaker0-0: complete

client/cli/src/params.rs
nodebreaker0-0: complete

utils/prometheus/src/lib.rs
nodebreaker0-0: complete

mxinden:
As far as I can tell https://github.com/paritytech/substrate/pull/4511/commits/ccb3179bfca127433db6767ce0f09752139929a2 is missing the following change in order to compile:

```diff
diff --git a/client/cli/src/params.rs b/client/cli/src/params.rs
index 128c66b30..1a4b28443 100644
--- a/client/cli/src/params.rs
+++ b/client/cli/src/params.rs
@@ -299,7 +299,6 @@ arg_enum! {
        pub enum TracingReceiver {
                Log,
                Telemetry,
-               Prometheus,
        }
 }
 
@@ -308,7 +307,6 @@ impl Into<sc_tracing::TracingReceiver> for TracingReceiver {
                match self {
                        TracingReceiver::Log => sc_tracing::TracingReceiver::Log,
                        TracingReceiver::Telemetry => sc_tracing::TracingReceiver::Telemetry,
-                       TracingReceiver::Prometheus => sc_tracing::TracingReceiver::Prometheus,
                }
        }
 }
```

nodebreaker0-0:
> This looks good to me once:
> 
> 1. The branch compiles (see small diff above).
> 2. The merge conflicts have been addressed.
> 
> Thanks for the hard work and again thanks for bearing with us.

Done

client/service/src/builder.rs
```diff
@@ -949,6 +1017,30 @@ ServiceBuilder<
...
+			let metrics = ServiceMetrics::register(&registry)?;
+
+			let future = select(
+				prometheus_exporter::init_prometheus(port, registry).boxed(),
+				exit.clone()
+			).map(drop);
+
+			let _ = to_spawn_tx.unbounded_send((
+				Box::pin(future),
+				From::from("prometheus-on-block")
```
mxinden: I am not sure what "prometheus-on-block" relates to. I would expect something like `prometheus-server` or `prometheus-endpoint`. Can you explain your thoughts?

NikVolf:
@nodebreaker0-0 you need to update Cargo.lock so that CI happy

run `cargo test --all` and commit the diff

client/service/src/builder.rs
nodebreaker0-0: I didn't understand a little bit about what this was, so I copied the Telemetry, but I think it's the `prometheus-endpoint`

nodebreaker0-0:
Real Done!. 
Thank you for joining me, see you again at Milestone 2.

gnunicorn:
please merge master and we are ready to merge.

mxinden:
Closing here in favor of https://github.com/paritytech/substrate/pull/4981 which includes the required *master merge*.

I am hoping to get https://github.com/paritytech/substrate/pull/4981 into *master* before the code freeze on Friday.

PR #4514, opened 2019-12-30
Insert key via node RPC for subkey

### LINKED ISSUE #4391: subkey can insert node keys via RPC
At the moment it’s quite cumbersome to add a session key to a node. I was following the [“Start you private network”](https://substrate.dev/docs/en/next/tutorials/start-a-private-network/customchain) tutorial which suggest using `subkey` to generate the keys and the PolkadotJS app to insert the keys using the `insert_key` API. It would be a lot more convenient if `subkey` would do this for you.

NikVolf:
Closes #4391

@tomusdrw

Maybe worth checking that it can only be submitted to the local node? (ie 127.0.0.1/localhost)


bin/utils/subkey/src/rpc.rs
```diff
@@ -0,0 +1,49 @@
...
+
+	pub fn insert_key(
+		&self,
+		key_type: String,
+		suri: String,
+		public: Bytes,
+	) {
+		let url = self.url.to_string();
+
+		rt::run(rt::lazy(move || {
```
tomusdrw: Do we really need `hyper` for that? Couldn't we just spawn on any runtime?

bin/utils/subkey/src/main.rs
```diff
@@ -384,6 +394,24 @@ where
...
 		}
+		("insert", Some(matches)) => {
+			let suri = get_uri("suri", &matches)?;
+			let pair = read_pair::<C>(Some(&suri), password)?;
+			let node_url = matches.value_of("node-url").unwrap_or("http://localhost:9933");
+			let key_type = matches.value_of("key-type").ok_or(Error::Static("Key type id is required"))?;
+
+			// Just checking
+			let _key_type_id = sp_core::crypto::KeyTypeId::try_from(key_type)
+				.map_err(|_| Error::Static("Unknown keytype - provide gran/babe/aura/acco/imon/audi"))?;
```
tomusdrw: I think that list will easily get out of date, maybe better to somehow get all items from the enum and convert them to `Display/Debug` here?

NikVolf:
> Maybe it's worth to add a flag too, to avoid having to run two separate commands?

I thought about adding it too, but seems rather clumsy to me

You will have one command like `subkey insert gran test.key http://localhost:9933` and then completely different layout to do the same thing via generate like `subkey genereate --node-url=... --key-type=...`

bin/utils/subkey/src/main.rs
NikVolf: Key-type is not strictly defined, any 4-character string is ok
I reduced variants so that it will not get out of date so easily

bin/utils/subkey/src/rpc.rs
```diff
@@ -0,0 +1,49 @@
...
+impl RpcClient {
+	pub fn new(url: String) -> Self { Self { url } }
+
+	pub fn insert_key(
+		&self,
+		key_type: String,
+		suri: String,
+		public: Bytes,
+	) {
+		let url = self.url.to_string();
```
bkchr:
```suggestion
		let url = self.url.clone();
```

bin/utils/subkey/src/main.rs
```diff
@@ -231,6 +232,15 @@ fn get_app<'a, 'b>(usage: &'a str) -> App<'a, 'b> {
 						If the value is a file, the file content is used as URI. \
 						If not given, you will be prompted for the URI.'
 				"),
+			SubCommand::with_name("insert")
+				.about("Insert a key to the local node")
```
bkchr:
```suggestion
				.about("Insert a key to the keystore of a node")
```

bin/utils/subkey/src/rpc.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright 2018-2019 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2019 Parity Technologies (UK) Ltd.
```

bkchr:
Inserting a freshly generated key does not makes that much of sense, as you most of the time only want to insert keys manually on genesis. Later you should stick to `rotate_keys` anyway IMHO.

bin/utils/subkey/src/rpc.rs
NikVolf: I'd suggest there is little use of worrying about it, since those are just synchronous requests initiated from cli, and reusing some runtime external will just complicate everything?

NikVolf: Chatted off github

PR #4515, opened 2019-12-31
Remove wrong assertion from phragmen

kianenigma:
Removes a reported `debug_assert!` from phragmen. This is wrong because in election-phragmen one can vote for themselves, and other candidates at the same time (not possible in staking since one account cannot be both validator and nominator.)

Also simplifies a bit by removing `own` from support. This makes phragmen unaware of self-vote.


primitives/phragmen/src/lib.rs
```diff
@@ -452,6 +438,8 @@ fn do_equalize<Balance, AccountId, C>(
 	// Nothing to do. This voter had nothing useful.
 	// Defensive only. Assignment list should always be populated.
 	if elected_edges.is_empty() { return 0; }
+	// Well, what can we equalize more? nothing.
+	if elected_edges.len() == 1 { return 0; }
```
gavofyork: these two conditions can be trivially reduced to one.

PR #4518, opened 2020-01-02
Update flamingfir spec for new testnet

ddorgan:
Update flaming fir spec file for new testnet after reset.


bkchr:
Please merge master to fix the tests :)

PR #4521, opened 2020-01-02
Fix `period` of offchain transactions.

### LINKED ISSUE #4484: Long-lived off-chain workers can't submit transactions
When execution of off-chain worker took more than block time - transaction submission from off-chain results in this error: `(offchain call) Error submitting a transaction to the pool: Pool(InvalidTransaction(InvalidTransaction::BadProof))`. But if the time of execution less - everything works fine.

I created a [repository](https://github.com/enfipy/substrate-offchain-example) with a delayed request in off-chain worker - to reproduce error just run `cargo run --release -- --dev` it will automatically insert dev key and start an off-chain worker that fails after 3s. To prove that it works without delay - change [this line](https://github.com/enfipy/substrate-offchain-example/blob/4338fdbe5e442d633e5b38031e587f583ed6afbe/runtime/src/example.rs#L84) with time query to 1 or fewer seconds.

tomusdrw:
Adds a bunch of documentation to `Era::period`, since mortal transactions' `period` can never exceed `BlockHashCount` parameter passed to `system` module (currently set to `250`).
Also fixes the default `period` in `CreateTransaction` implementation.

Resolves #4484


bin/node/runtime/src/lib.rs
```diff
@@ -500,7 +500,8 @@ impl frame_system::offchain::CreateTransaction<Runtime, UncheckedExtrinsic> for
 		account: AccountId,
 		index: Index,
 	) -> Option<(Call, <UncheckedExtrinsic as traits::Extrinsic>::SignaturePayload)> {
-		let period = 1 << 8;
+		// make sure it does not exceed `BlockHashCount`
```
NikVolf: How the line below that makes it certain?

bkchr: We could do:
```
BlockHashCount::get().checked_next_power_of_two().map(|c| c / 2).unwrap_or(2)
```

Issue #4209, opened 2019-11-26
IO Metrics


arkpar:
Collect statistics on both hi-level and low-level IO. Make statistics available over RPC.

### Hi level state IO
Storage writes/reads. Average value size (perhaps per module). Storage cache hits/misses. Benchmark heavy operations, such as `storage_root`.

###  Mid level IO:
`state-db` Reads/writes, journal size. Memory reads vs RocksDb reads.

###  Low level IO:
RocksDb statistics. Check if there's an API for retrieving internal statics. If there's none collect own.

Related: #830

PR #4525, opened 2020-01-03
Memory and I/O stats for backend database(s)

NikVolf:
As a part of #4209, #830

~~Based on not yet published `kvdb`/`kvdb-rockdb`, but can be reviewed~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -368,7 +368,7 @@ dependencies = [
  "futures 0.1.29 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "js-sys 0.3.34 (registry+https://github.com/rust-lang/crates.io-index)",
- "kvdb-web 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "kvdb-web 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "libp2p 0.13.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.6.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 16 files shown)


client/db/src/lib.rs
```diff
@@ -823,6 +823,37 @@ where
 	}
 }
 
+struct Frozen<T: Clone> {
+	at: std::time::Instant,
+	value: T,
+}
+
+pub(crate) struct FrozenForDuration<T: Clone> {
```
NikVolf: This is used to always return last 1-second stats

primitives/io/src/lib.rs
```diff
@@ -712,7 +712,7 @@ pub trait Offchain {
 		deadline: Option<Timestamp>,
 	) -> Result<u32, HttpError> {
 		self.extension::<OffchainExt>()
-			.expect("http_response_read_body can be called only in the offchain worker context")
```
pepyakin: Are these needed?

NikVolf: Nah, it's already in master, just got squashed

client/api/src/client.rs
```diff
@@ -87,7 +87,55 @@ pub struct ClientInfo<Block: BlockT> {
...
+pub struct MemoryInfo {
+	/// Size of state cache.
+	pub state_cache: usize,
+	/// Size of backend database cache.
+	pub database_cache: usize,
+}
+
+/// I/O statistics for client instance.
+#[derive(Default, Clone, Debug)]
+pub struct IoInfo {
```
gnunicorn: Are these stats persisted between restarts of the client? If not, we might want the docs to say that this only for the duration since last start to avoid confusion down the line?

Cargo.lock
```diff
@@ -1085,6 +1085,31 @@ dependencies = [
  "libc 0.2.66 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
+[[package]]
+name = "ethbloom"
```
NikVolf: bringing those is a bug, will get rid of it In a moment

client/api/src/client.rs
NikVolf: Not persisted

Sure, I'll add a comment

NikVolf: Added comment to `UsageInfo`

client/db/src/lib.rs
bkchr: A comment would be nice to explain what this does :)

client/src/light/blockchain.rs
```diff
@@ -314,5 +314,9 @@ pub mod tests {
 		fn cache(&self) -> Option<Arc<dyn BlockchainCache<Block>>> {
 			None
 		}
+
+		fn usage_info(&self) -> sc_client_api::UsageInfo {
+			sc_client_api::UsageInfo::default()
```
bkchr: Maybe the function should return an `Option`?

NikVolf:
@bkchr All addressed, thanks!

bkchr:
@NikVolf needs a last master merge.

PR #1669, opened 2019-02-04
Light GRANDPA import handler

svyatonik:
closes #269

I'm opening this PR to share the ideas && to make sure I'm on the right track. So the general idea is:
1) we do not pause syncing when see a block with consensus-related changes (which is just `DigestItem::AuthoritiesChange` now). This differs from original description of #269 and what has been discussed in #1268. But since full clients do not pause sync, I'm not sure - why light clients should (that's what bothering me). So light block importer only tries to get finalization of such blocks ASAP;
2) getting finalization = checking justification. To check justification, we need to know GRANDPA authorities set. One possible option is to use existing finality proof every time, but I think I've found a better way:
	2.1) assuming that every change of this set if accompanied with justification;
	2.2) when seeing a justification we could assume that it is created with previous set;
	2.3) if check-with-previous-set fails, we're fetching GRANDPA authorities set from the remote full node;
	2.4) if recheck-with-new-set works, then we're finalizing the block + storing new set + new set_id (which is increased by one on every change) locally

This should work when justifications are already created (i.e. when we're syncing from the scratch). But there's a case when this strategy will fail:
1) B1 which enacts new GRANDPA authorities set is imported on FULL client without justification (let's say it changes `set_id` from 10 to 11);
2) FULL announces this block to LIGHT && light imports the block without finalization (`set_id` is still 10 on light);
3) FULL creates justification J1 for B1, but it doesn't reach LIGHT yet;
4) B2 which updates `set_id` from 11 to 12 is imported by FULL && justification J2 is created;
5) light client tries to import B2+J2 && import fails, because it has missed 10 -> 11 change of `set_id`.

That's where I suppose finality proof could help (if changed a bit). So that new finality proof (which will work in both optimistic + pessimistic cases) will be:
**REQUEST**: prove finalization of BEST_BLOCK given that our best finalized is BEST_FINALIZED
**RESPONSE**:
&nbsp;&nbsp;&nbsp;&nbsp;assume that we have best justification for BEST_BLOCK+2
&nbsp;&nbsp;&nbsp;&nbsp;if GRANDPA set has NOT been changed in [BEST_FINALIZED; BEST_BLOCK+2]:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;we only need to respond with justification for BEST_BLOCK+2
&nbsp;&nbsp;&nbsp;&nbsp;if GRANDPA set has been changed in that range:
[...]

PR #4528, opened 2020-01-04
Fix light node sync problem

### LINKED ISSUE #4523: v1.0 light node cannot sync to the newest block
Seems  that v1.0 light node doesn't work. It cannot sync to the newest block. Apologies if I misunderstood.

# Reproduction steps
* run Alice full node
`
cargo run --   --base-path /tmp/alice   --chain=local   --alice   --node-key 0000000000000000000000000000000000000000000000000000000000000001   --telemetry-url ws://telemetry.polkadot.io:1024   --validator --log="sync=trace" 
`

* run Bob full node
`
cargo run --  --base-path /tmp/bob   --bootnodes /ip4/127.0.0.1/tcp/30333/p2p/QmQZ8TjTqeDj3ciwr93EJ95hxfDsb9pEYDizUAbWpigtQN   --chain=local   --bob   --port 30334   --telemetry-url ws://telemetry.polkadot.io:1024   --validator --log="sync=trace"
`

* wait alice and bob generate some new blocks

* run Charlie light node
`
cargo run --  --base-path /tmp/charlie  --bootnodes /ip4/127.0.0.1/tcp/30333/p2p/QmQZ8TjTqeDj3ciwr93EJ95hxfDsb9pEYDizUAbWpigtQN   --chain=local   --charlie  --port 30335   --telemetry-url ws://telemetry.polkadot.io:1024   --light --log="sync=trace"
`

Charlie light node cannot sync to the newest block

Err log of Charlie light node:
`
2020-01-03 10:50:10.570 libp2p-5 DEBUG sync  Dropping PeerId("QmQZ8TjTqeDj3ciwr93EJ95hxfDsb9pEYDizUAbWpigtQN") because "Verification failed: Could not fetch authorities at 0x629de053259ddd869b916c34cf059288bca9992b1c6590a6e6901f7c95c4c685: Error(InvalidAuthoritiesSet, State { next_error: None, backtrace: InternalBacktrace { backtrace: None } })"
`

Err log of Alice full  node
`
2020-01-03 10:44:41.393 Protocol TRACE sync  Remote call request 0 from QmeyzkWbNfP2zRppFivaBvvfDni3J3Z4cFt5r9pkwc4Lir (version at 0x629d…c685) failed with: Execution: Function: Module doesn't have export version
`

guoxbin:
Fixes #4523

v1.0 light node cannot sync to the newest block.
There are 2 problems:

* light node call a wrong method name `version`
`version` is a method of Core mudule, so  the method name should be `Core_version`

* the  import blocks channel of light node has a limited capacity
light node makes remote calls when importing and verifying new blocks and costs more time than full node does.  
When a light node has a lot of new blocks to sync, a limited capacity channel will be a problem.

### DIFF
--- a/core/client/src/light/call_executor.rs
+++ b/core/client/src/light/call_executor.rs
@@ -133,7 +133,7 @@ where
 	}
 
 	fn runtime_version(&self, id: &BlockId<Block>) -> ClientResult<RuntimeVersion> {
-		let call_result = self.call(id, "version", &[], ExecutionStrategy::NativeElseWasm, NeverOffchainExt::new())?;
+		let call_result = self.call(id, "Core_version", &[], ExecutionStrategy::NativeElseWasm, NeverOffchainExt::new())?;
 		RuntimeVersion::decode(&mut call_result.as_slice())
 			.ok_or_else(|| ClientErrorKind::VersionInvalid.into())
 	}
--- a/core/consensus/common/src/import_queue.rs
+++ b/core/consensus/common/src/import_queue.rs
@@ -417,7 +417,7 @@ impl<B: BlockT, V: 'static + Verifier<B>> BlockImportWorker<B, V> {
 		verifier: Arc<V>,
 		block_import: SharedBlockImport<B>,
 	) -> Sender<BlockImportWorkerMsg<B>> {
-		let (sender, port) = channel::bounded(4);
+		let (sender, port) = channel::unbounded();
 		let _ = thread::Builder::new()
 			.name("ImportQueueWorker".into())
 			.spawn(move || {


guoxbin:
[clabot:check]

bkchr:
Why will the channel be a problem? We have a limited channel to not blow up the memory usage.

guoxbin:
> Why will the channel be a problem? We have a limited channel to not blow up the memory usage.

When a light node verify a new block header, It needs to  get state (like authories). 

Since the light node doesn't maintain local state, It need make remote call to get the state, which slows down the verification speed and exhausts the import block channel.

arkpar:
cc @svyatonik

bkchr:
> > Why will the channel be a problem? We have a limited channel to not blow up the memory usage.
> 
> When a light node verify a new block header, It needs to get state (like authories).
> 
> Since the light node doesn't maintain local state, It need make remote call to get the state, which slows down the verification speed and exhausts the import block channel.

Nevertheless, I would stick to a bounded channel. We could increase the size. Maybe to 40 elements?

guoxbin:
> Nevertheless, I would stick to a bounded channel. We could increase the size. Maybe to 40 elements?

I know your concern. 

According to my recent study, I think the requirement of remote call is quite related to how the consensus module is implemented. 
If the block verification makes lots of remote calls, the sync from very early block number to a large best number may be a mission impossible.

I find that a light node will lose it's genesis state after a restart, which causes the node make a remote call even when querying state of genesis block. This causes things worse.

So I suggest:
Firstly, fix the genesis state lost problem, 
and then if unbounded(4) is acceptable, we don't need modify that.

Some code about genesis state lost problem
the genesis_state is initialized with None and no code will reload it from disk storage:

```RUST
pub struct Backend<S, F, H> {
	blockchain: Arc<Blockchain<S, F>>,
	genesis_state: RwLock<Option<InMemoryState<H>>>,
}
```

```RUST
impl<S, F, H> Backend<S, F, H> {
	/// Create new light backend.
	pub fn new(blockchain: Arc<Blockchain<S, F>>) -> Self {
		Self {
			blockchain,
			genesis_state: RwLock::new(None),
		}
	}
```

bkchr:
The problem is also that you are working here with deprecated code. We don't work on this anymore. Did you checked master to see if your problems are solved there?

guoxbin:
Yes, our project is based on v1.0. We may migrate to the newest version later, but this is not a simple work
I would like to help fix the problem on branch v1.0 for maybe someone else need a solid legacy vesion.

svyatonik:
just a side note: the idea behind light client was that it shouldn't have made **any** runtime calls during sync at all. So if in 1.0 it makes such calls, then it isn't actually a light client (iirc it has been syncing much longer than a full client).

To make light client working for 1.0, you'll at least need #1622 and #1669 (and even more, but these are essential), even though they're already obsoleted in the master branch. I'm not sure - whether we accept PRs to `v1.0` branch at all?

guoxbin:
Agree that light client shouldn't have made any runtime calls.

Branch v1.0 contains the commit by pr #1622. 
This makes the node have the genesis state for the first time, but the genesis state will be lost after restarting.

bkchr:
Yeah, I'm also not that convinced of merging something into version 1.0.

bkchr:
I would really encourage you to migrate to 2.0. 

For the time, I would propose that you just use a fork of Substrate 1.0 to add fixes you need. Sorry, but we don't have the resources to maintain this old version.

PR #4531, opened 2020-01-04
A Social Account Recovery Pallet

shawntabrizi:
This is social account recovery pallet.

Basic premise is that:
* A user can create a recovery process for their account by choosing some "friends" who can help recover access to an account if the private key is lost or something else of that nature.

Here is the recovery setup struct:

```rust
/// Configuration for recovering an account.
pub struct RecoveryConfig<BlockNumber, Balance, AccountId> {
	/// The minimum number of blocks since the start of the recovery process before the account
	/// can be recovered.
	delay_period: BlockNumber,
	/// The amount held in reserve of the `depositor`,
	/// to be returned once this configuration is removed.
	deposit: Balance,
	/// The list of friends which can help recover an account. Always sorted.
	friends: Vec<AccountId>,
	/// The number of approving friends needed to recover an account.
	threshold: u16,
}
```

* To configure your account to be recoverable, a deposit is reserved. The deposit amount scales linearly with the number of friends you choose.
* Any account (the "rescuer") can start to recover an account with a recovery process set up on it. This requires a deposit which can be claimed by the recovered account to deter malicious recovery attempts.
* Friends can vouch for the recovery attempt.
* Once the threshold of vouches are met, the rescuer can claim access to the lost account after the delay period has passed.
* Then the rescuer can make calls through this pallet on behalf of the lost account, doing whatever is needed.
* The user can call through the recovered account to remove all other active recoveries (including their own), reclaiming the deposit into the recovered account, and ultimately remove the rescue setup.

- [x] Tests
- [x] Docs
- [x] Weights


shawntabrizi:
If people wanted to review the end to end life cycle tests, now would be a good time to give feedback on the design. Otherwise, will keep this mostly on the back burner since I don't want to get too heavily involved before getting some initial feedback.

frame/recovery/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+			Self::deposit_event(RawEvent::RecoveryVouched(lost, rescuer, who));
+		}
+		
+		/// Allow a rescuer to claim their recovered account.
+		fn claim_recovery(origin, account: T::AccountId) {
+			let who = ensure_signed(origin)?;
+			// Get the recovery configuration for the lost account
+			let recovery_config = Self::recovery_config(&account).ok_or(Error::<T>::NotRecoverable)?;
+			// Get the active recovery process for the rescuer
+			let active_recovery = Self::active_recovery(&account, &who).ok_or(Error::<T>::NotStarted)?;
```
shawntabrizi: There is a possible optimization that can be done with the `claim_recovery` function.

You can see here that we read from storage both the `recovery_config` storage item and the `active_recovery` storage item.

We only need the `recovery_config` storage item to get the `threshold` value and check that it is less than or equal to the number of friends stored in the `active_recovery` storage item.

The optimization would be to copy the `threshold: u16` value into the `active_recovery` struct so that we don't need to read from the `recovery_config`.

So which is more efficient?

* One additional storage read per call to `claim_recovery`.
* One additional `u16` value per `active_recovery`. (this means across all active recoveries across all recoverable accounts).

shawntabrizi: My guess is that duplicating the u16 value is better overall, and I will likely make an update to fix this.

shawntabrizi:
This should be totally done and ready for review + merge :)

frame/recovery/src/tests.rs
```diff
@@ -0,0 +1,383 @@
...
+		assert_ok!(Recovery::initiate_recovery(Origin::signed(1), 5));
+		// Off chain, the user contacts their friends and asks them to vouch for the recovery attempt
+		assert_ok!(Recovery::vouch_recovery(Origin::signed(2), 5, 1));
+		assert_ok!(Recovery::vouch_recovery(Origin::signed(3), 5, 1));
+		assert_ok!(Recovery::vouch_recovery(Origin::signed(4), 5, 1));
+		// We met the threshold, lets try to recover the account...?
+		assert_noop!(Recovery::claim_recovery(Origin::signed(1), 5), Error::<Test>::DelayPeriod);
+		// We need to wait at least the delay_period number of blocks before we can recover
+		run_to_block(20);
+		assert_ok!(Recovery::claim_recovery(Origin::signed(1), 5));
```
gavofyork: why are `claim_recovery` and `close_recovery` two separate calls? wouldn't you always want to `close_recovery` after claiming?

shawntabrizi: There does need to be a separate `close_recovery` extrinsic for both repatriating the deposit  of malicious recovery attempts and for cleaning up other existing recovery processes (without executing a loop).

It could be that when we call `claim_recovery`, that we can close our own recovery process automatically, but I thought we can just leave that to the front-end to do, and keep our API really clean.

Let me know if you think there are reasons to design this differently.

frame/recovery/src/tests.rs
```diff
@@ -0,0 +1,383 @@
...
+		// Account 1 can use account 5 to close the active recovery process, claiming the deposited
+		// funds used to initiate the recovery process into account 5.
+		let call = Box::new(Call::Recovery(RecoveryCall::close_recovery(1)));
+		assert_ok!(Recovery::as_recovered(Origin::signed(1), 5, call));
```
gavofyork: this is ok, then 4 lines below, the next one is also ok... what changed?

frame/recovery/src/tests.rs
```diff
@@ -0,0 +1,383 @@
...
+		// Account 1 can then use account 5 to remove the recovery configuration, claiming the
+		// deposited funds used to create the recovery configuration into account 5.
+		let call = Box::new(Call::Recovery(RecoveryCall::remove_recovery()));
+		assert_ok!(Recovery::as_recovered(Origin::signed(1), 5, call));
+		// Account 1 should now be able to make a call through account 5 to get all of their funds
```
gavofyork: probably wants to be swapped with the line above.

also, it was apparently able to do this 5 lines before anyway?

frame/recovery/src/lib.rs
```diff
@@ -0,0 +1,665 @@
...
+		///
+		/// The dispatch origin for this call must be _Signed_ and registered to
+		/// be able to make calls on behalf of the recovered account.
+		///
+		/// Parameters:
+		/// - `account`: The recovered account you want to make a call on-behalf-of.
+		/// - `call`: The call you want to make with the recovered account.
+		///
+		/// # <weight>
+		/// - The weight of the `call`.
```
gavofyork: There's a storage lookup here, too. (`Self::recovered_account`)

shawntabrizi: Ah sorry about that. I will do another pass at weights.

gavofyork:
Needs to have master merged and fixed.

marceljay:
Hey Shawn, thanks for the intro at tonights's meetup, I couldn't make, it but watched it on the train. 
Is there some explanation how calling on behalf of another account works? I remember that consensus on substrate is decoupled from payments (which are just an "app") and thought this might be the reason, or am I wrong? Also I did not really understand what pallets are (a concept in rust?), I could not find a good intro in the devhub.

burdges:
Just fyi..

It appears custody solutions denominated in euros are required to provide account recovery using the users identity.  There was one [grant application](https://github.com/w3f/Web3-collaboration/pull/226) to [solve this without completely centralizing](https://docs.taler.net/anastasis.html). 

I'd kinda like us to fund it so we can make them use [Rogaway's Adept Secret Sharing](https://www.youtube.com/watch?v=qJv7-fxFVC8#t=43m10s) whenever he releases it, instead of their home brewed thing.  We do not however have enough interest from custody solutions to justify funding it though, but maybe useful here too.  

[RustySecrets](https://github.com/SpinResearch/RustySecrets) should eventually implement Rogaway's Adept Secret Sharing, btw. 

This would not impact this pull request, merely pointing this out.

shawntabrizi:
@marceljay All the magic around calling on behalf of another account comes from the `as_recovered` function:

```
#[weight = <Passthrough<T::AccountId, <T as Trait>::Call>>::new()]
fn as_recovered(origin,
	account: T::AccountId,
	call: Box<<T as Trait>::Call>
) -> DispatchResult {
	let who = ensure_signed(origin)?;
	// Check `who` is allowed to make a call on behalf of `account`
	ensure!(Self::recovered_account(&account) == Some(who), Error::<T>::NotAllowed);
	call.dispatch(frame_system::RawOrigin::Signed(account).into())
}
```
 Specifically the fact that we are able to create a new origin from within the runtime logic:

```
frame_system::RawOrigin::Signed(account)
```

The ability to do this is an extremely privileged action, but since we are runtime developers, we are allowed to do it. The only sense of identity that our runtime logic cares about is this origin, so we are essentially able to make any other pallet believe the call is coming from our lost account rather than the one we have access to.

The pallet term is relatively new, but means the same thing that "runtime module" did before. Check out this https://github.com/paritytech/substrate/pull/4182

> FRAME is our framework for developing new Substrate pallets. Pallets are a special kind of Rust module from which Substrate runtimes can be composed. Each pallet has its own discrete logic which can modify the features and functionality of your blockchain's state transition function.

marceljay:
Okay there was definitely some confusion on my side due to the changed wording. 
FRAME = SRML
Pallet = Module
(tbh, i liked the original names much better)

Regarding the code: my rust knowledge is too basic to understand the code. But I think my questions was rather how consensus is made based on certain function calls. Would definitely like to understand this in the future.

graceadam131:
Hey everybody, I am suggesting a notable individual in my life whom I recruited his administration to erase my criminal record in New York. He is an expert programmer. contact: (wisetechhacker gmail com) OR WhatsApp +1 (424) 283 6238

him for versatile hack, informal community hack, interpersonal organization check (blue identification) ledger hack and assets move, dmv hack, overhaul school results, charge card hack, 

redesign and eradicate terrible FICO ratings

PR #4532, opened 2020-01-05
Update Copyright to 2020

shawntabrizi:
Happy new year!

### DIFF
--- a/bin/node-template/runtime/build.rs
+++ b/bin/node-template/runtime/build.rs
@@ -1,4 +1,4 @@
-// Copyright 2019 Parity Technologies (UK) Ltd.
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
 // This file is part of Substrate.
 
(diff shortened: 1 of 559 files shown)


frame/recover/src/lib.rs
```diff
@@ -0,0 +1,339 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
```
joepetrowski: I think this file is from your other open PR and should be removed from this branch.

frame/bet/src/lib.rs
```diff
@@ -0,0 +1,984 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
```
bkchr: New file.

shawntabrizi: dunno how it got dirty, I used master afaik :/

PR #4535, opened 2020-01-06
client: allow reverting finalized blocks

andresilva:
This PR extends the revert functionality to allow reverting past finalized blocks. It also fixes the handling of leaves when reverting blocks.

### DIFF
--- a/client/db/src/lib.rs
+++ b/client/db/src/lib.rs
@@ -1485,50 +1485,70 @@ impl<Block> sc_client_api::backend::Backend<Block, Blake2Hasher> for Backend<Blo
 	}
 
 	fn revert(&self, n: NumberFor<Block>, revert_finalized: bool) -> ClientResult<NumberFor<Block>> {
-		let mut best = self.blockchain.info().best_number;
+		let mut best_number = self.blockchain.info().best_number;
+		let mut best_hash = self.blockchain.info().best_hash;
 		let finalized = self.blockchain.info().finalized_number;
 
-		let revertible = best - finalized;
+		let revertible = best_number - finalized;
 		let n = if !revert_finalized && revertible < n {
 			revertible
 		} else {
 			n
 		};
 
-		for c in 0 .. n.saturated_into::<u64>() {
-			if best.is_zero() {
-				return Ok(c.saturated_into::<NumberFor<Block>>())
-			}
-			let mut transaction = DBTransaction::new();
-			match self.storage.state_db.revert_one() {
-				Some(commit) => {
-					apply_state_commit(&mut transaction, commit);
-					let removed = self.blockchain.header(BlockId::Number(best))?.ok_or_else(
-						|| sp_blockchain::Error::UnknownBlock(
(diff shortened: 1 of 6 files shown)


client/db/src/lib.rs
```diff
@@ -1484,40 +1484,71 @@ impl<Block> sc_client_api::backend::Backend<Block, Blake2Hasher> for Backend<Blo
...
-						|| sp_blockchain::Error::UnknownBlock(
-							format!("Error reverting to {}. Block hash not found.", best)))?;
-					let key = utils::number_and_hash_to_lookup_key(best.clone(), &hash)?;
-					transaction.put(columns::META, meta_keys::BEST_BLOCK, &key);
-					transaction.delete(columns::KEY_LOOKUP, removed.hash().as_ref());
-					children::remove_children(&mut transaction, columns::META, meta_keys::CHILDREN_PREFIX, hash);
-					self.storage.db.write(transaction).map_err(db_err)?;
-					self.blockchain.update_meta(hash, best, true, false);
-					self.blockchain.leaves.write().revert(removed.hash().clone(), removed.number().clone(), removed.parent_hash().clone());
+		let revertible = best_number - finalized;
```
```suggestion
```

client/db/src/lib.rs
```diff
@@ -1484,40 +1484,71 @@ impl<Block> sc_client_api::backend::Backend<Block, Blake2Hasher> for Backend<Blo
...
+		let n = if !revert_finalized && revertible < n {
```
```suggestion
		let n = if !revert_finalized {
```

client/db/src/lib.rs
```diff
@@ -1484,40 +1484,71 @@ impl<Block> sc_client_api::backend::Backend<Block, Blake2Hasher> for Backend<Blo
...
+			revertible
```
```suggestion
			best_number - finalized
```

client/db/src/lib.rs
gavofyork: this change would alter the logic. explanation?

client/db/src/lib.rs
```diff
@@ -1484,40 +1484,71 @@ impl<Block> sc_client_api::backend::Backend<Block, Blake2Hasher> for Backend<Blo
...
+			n
+		};
+
+		let mut revert_blocks = || -> ClientResult<NumberFor<Block>> {
+			for c in 0 .. n.saturated_into::<u64>() {
+				if best_number.is_zero() {
+					return Ok(c.saturated_into::<NumberFor<Block>>())
+				}
+				let mut transaction = DBTransaction::new();
+				match self.storage.state_db.revert_one() {
```
arkpar: I believe this still won't revert past the last canonicalized block when pruning is enabled. `revert_one` returns `None` for finalized blocks.

andresilva: Yes, this only works with `pruning=archive` (i.e. for validators). It will stop reverting if it has no state.

kigawas:
In which case it'll need to revert finalized blocks?

andresilva:
@kigawas https://medium.com/@gavofyork/kusamas-first-adventure-2cd4f439a7a4 A writeup of why we needed this for Kusama.

PR #4542, opened 2020-01-06
Clean-ups in the network-gossip crate

tomaka:
Performs a few clean-ups related to the `sc-network-gossip` crate:

- Removes usage of `sc_network::Context`, and directly pass a `&mut dyn Network` instead.
- Since gossiping doesn't use `sc_network::Context` anymore, removes the `Context::send_consensus` method. The `Context` is now used exclusively for the `NetworkSpecialization` stuff, which is hopefully going to disappear in the future as well.
- Moves the `Validator` trait and its related traits and structs to a separate module.

### DIFF
--- a/client/network-gossip/src/validator.rs
+++ b/client/network-gossip/src/validator.rs
@@ -1,0 +1,103 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
+// This file is part of Substrate.
+
+// Substrate is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


tomaka:
Successfully tested on Kusama.

tomaka:
Polkadot PR: https://github.com/paritytech/polkadot/pull/732

client/network-gossip/src/validator.rs
```diff
@@ -0,0 +1,103 @@
...
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
```
rphmeier: missing module doc

PR #4544, opened 2020-01-06
client: add a block blacklist extension

andresilva:
We already have a "whitelist" feature through `ForkBlocks` which allows defining the canonical block at a given height in the chainspec. This PR adds a `BadBlocks` extension that allows defining bad block hashes, the client will refuse to import these blocks.


andresilva:
Building polkadot to test this on Kusama now.

andresilva:
Tested it on polkadot and it's working as expected.

PR #4548, opened 2020-01-06
Implement runtime version checks in `set_code`

bkchr:
Check that the new runtime code given to `set_code` fullfills some
requirements:

- `spec_name` matches
- `spec_version` does not decreases
- `impl_version` does not decreases
- Either `spec_version` and `impl_version` increase

### DIFF
--- a/client/executor/src/lib.rs
+++ b/client/executor/src/lib.rs
@@ -57,14 +57,15 @@ pub use sc_executor_common::{error, allocator, sandbox};
 /// - `call_data`: Will be given as input parameters to `function`
 /// - `execution_method`: The execution method to use.
 /// - `ext`: The externalities that should be set while executing the wasm function.
+///          If `None` is given, no externalities will be set.
 /// - `heap_pages`: The number of heap pages to allocate.
 ///
 /// Returns the `Vec<u8>` that contains the return value of the function.
-pub fn call_in_wasm<E: Externalities, HF: sp_wasm_interface::HostFunctions>(
+pub fn call_in_wasm<HF: sp_wasm_interface::HostFunctions>(
 	function: &str,
 	call_data: &[u8],
 	execution_method: WasmExecutionMethod,
-	ext: &mut E,
+	ext: Option<&mut dyn Externalities>,
 	code: &[u8],
 	heap_pages: u64,
 ) -> error::Result<Vec<u8>> {
@@ -74,7 +75,14 @@ pub fn call_in_wasm<E: Externalities, HF: sp_wasm_interface::HostFunctions>(
 		code,
 		HF::host_functions(),
 	)?;
-	instance.call(ext, function, call_data)
+
+	match ext {
+		Some(ext) => sp_externalities::set_and_run_with_externalities(
+			ext,
+			|| instance.call(function, call_data),
+		),
+		None => instance.call(function, call_data)
+	}
 }
 
 /// Provides runtime information.
--- a/primitives/io/src/lib.rs
+++ b/primitives/io/src/lib.rs
@@ -36,7 +36,7 @@ use sp_std::ops::Deref;
 #[cfg(feature = "std")]
 use sp_core::{
 	crypto::Pair,
-	traits::KeystoreExt,
+	traits::{KeystoreExt, CallInWasmExt},
 	offchain::{OffchainExt, TransactionPoolExt},
 	hexdisplay::HexDisplay,
 	storage::{ChildStorageKey, ChildInfo},
@@ -50,7 +50,7 @@ use sp_core::{
 };
 
 #[cfg(feature = "std")]
-use ::sp_trie::{TrieConfiguration, trie_types::Layout};
+use sp_trie::{TrieConfiguration, trie_types::Layout};
 
 use sp_runtime_interface::{runtime_interface, Pointer};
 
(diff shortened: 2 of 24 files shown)


rphmeier:
So we can't brick the chain by supplying the empty code block anymore?

gavofyork:
looks like a promising direction.

primitives/state-machine/src/lib.rs
```diff
@@ -204,8 +204,10 @@ impl<'a, B, H, N, T, Exec> StateMachine<'a, B, H, N, T, Exec> where
 		exec: &'a Exec,
 		method: &'a str,
 		call_data: &'a [u8],
-		extensions: Extensions,
+		mut extensions: Extensions,
 	) -> Self {
+		extensions.register(CallInWasmExt::new(exec.clone()));
```
tomusdrw: So it's now available to every runtime call I guess, what are the consequences here? Besides using it being super slow, as I guess a new wasm env has to be initiated.

client/executor/src/lib.rs
```diff
@@ -57,26 +57,55 @@ pub use sc_executor_common::{error, allocator, sandbox};
...
 /// - `heap_pages`: The number of heap pages to allocate.
 ///
 /// Returns the `Vec<u8>` that contains the return value of the function.
-pub fn call_in_wasm<E: Externalities, HF: sp_wasm_interface::HostFunctions>(
+pub fn call_in_wasm<HF: sp_wasm_interface::HostFunctions>(
 	function: &str,
 	call_data: &[u8],
 	execution_method: WasmExecutionMethod,
-	ext: &mut E,
+	ext: &mut dyn Externalities,
```
gavofyork: all the storage accesses (and there are a lot) go through externalities. does this really have an insubstantial effect on performance?

bkchr: We store the Externalities as trait pointer in the environmental anyway. The compiler can not optimize this anywax.

primitives/state-machine/src/lib.rs
bkchr: Why super slow? The wasm blob needs to be initialized when we want to call some function in it, but there is nothing we can optimize (as we don't know the blob in front of).

We need the extension while building/executing a block. But there is also no direct method that you can call from the runtime, besides the one that only can call `Core_version`.

primitives/io/src/lib.rs
```diff
@@ -351,6 +351,19 @@ pub trait Misc {
 	fn print_hex(data: &[u8]) {
 		log::debug!(target: "runtime", "{}", HexDisplay::from(&data));
 	}
+
+	/// Extract the runtime version of the given wasm blob by calling `Core_version`.
+	///
+	/// Returns the SCALE encoded runtime version and `None` if the call failed.
+	fn runtime_version(&mut self, wasm: &[u8]) -> Option<Vec<u8>> {
+		// Create some dummy externalities, `Core_version` should not write data anyway.
```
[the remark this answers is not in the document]
bkchr: What kind of side effects? Even if the call would write something, it would not leak into the Externalities of the actual block. If that was your question.

primitives/sandbox/with_std.rs
```diff
@@ -187,9 +188,12 @@ impl<T> ImportResolver for EnvironmentDefinitionBuilder<T> {
 			module_name.as_bytes().to_owned(),
 			field_name.as_bytes().to_owned(),
 		);
-		let externval = self.map.get(&key).ok_or_else(|| {
-			wasmi::Error::Instantiation(format!("Export {}:{} not found", module_name, field_name))
-		})?;
+		println!("{} {}", module_name, field_name);
+		let lol = ExternVal::HostFunc(HostFuncIndex(100));
+		let externval = self.map.get(&key).unwrap_or_else(|| {
+			// wasmi::Error::Instantiation(format!("Export {}:{} not found", module_name, field_name))
```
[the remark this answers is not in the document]
bkchr: Yeah, I had forgot some debugging stuff :sweat_smile: Good that you found it!

primitives/io/src/lib.rs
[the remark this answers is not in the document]
tomusdrw: Yes, that's what I meant, I wanted to be sure that the two contexts do not share anything.

primitives/state-machine/src/lib.rs
tomusdrw: Sure, I know there is no other way to call `Core_version` from a different runtime blob, but afair when talking with Sergei he was saying that initializing wasm runtime (alloc memory, load module, etc) takes quite some time, hence I've mentioned that it's most likely going to be way slower than any other operation or host function we have, so we need to be careful to avoid using it too often.
I was wondering if there are any other (more severe) consequences of exposing this kind of function to the runtime - i.e. if it can be somehow abused.

> . But there is also no direct method that you can call from the runtime, besides the one that only can call Core_version.

That's good to know, so it means that it's not easy to shoot yourself in a foot if you are a runtime developer, right?

bkchr: Yeah it is probably no easy operation, but better to include less transactions than to stop the network again ;) If something does not work or for whatever reason we need to put a blob in without any checks, I added `set_code_without_checks`. This will just do it the old way, so we are sure that we can put a blob in :) 

Yeah, the runtime developer can not access the `CallInRuntimeExt` in any other way. It is not exposed to the runtime. The developer can just call `misc::runtime_version` that uses this stuff internally.

tomusdrw: Okay, that's fair, but I would really add extra docs to make sure it's not being called as it's a possible DoS vector. Imagine if we do that check whenever someone proposes a runtime, then people can craft arbitrary runtimes and do arbitrary operations within `Core_version` and it wouldn't be priced correctly. So we should only make sure to run in during `set_code`, not during any preliminary checks (that accept unreviewed user's input).

bkchr:
I added some temporary execution time measuring into `runtime_version` and the host call takes `60ms`. Given the time it takes to execute the transaction itself, this sounds okay and reasonable.

gavofyork:
will need resolving

Issue #4456, opened 2019-12-19
Wasm executor should provide stubs for unknown externs


bkchr:
The wasm executor currently returns an error when it finds an extern for that we don't provide an implementation on the host. Instead of returning an error, the executor should be configurable to provide stubs for functions it does not know. Each stub should return an error when being called, which instructs the executor to stop the execution of the runtime.

- Make the executor configurable on what to do for unknown externs.
- If stub usage is enabled, we should provide the stub and print some `trace!` info about this.

Wasmi integration should be rather trivial. We have the resolver here: https://github.com/paritytech/substrate/blob/master/client/executor/wasmi/src/lib.rs#L280
If stub generation is enabled we should return a fixed index that corresponds to our stub.

Wasmtime maybe requires some more work, but that needs to be tested. Currently we register all the host functions we get here: https://github.com/paritytech/substrate/blob/master/client/executor/wasmtime/src/runtime.rs#L218
As the required extern functions are present in the wasm blob, it should be rather trivial to extract them and provide a stub for each unknown.

I can mentor/help on this.

bkchr:
Already implemented :)

cecton:
@bkchr also for wasmtime??

bkchr:
Yes

PR #4550, opened 2020-01-07
Wasm executor should provide stubs for unknown externs (wasmi)

cecton:
This is part of the implementation for #4456, only for wasmi

(Let me squash merge so I can put a proper commit message)


client/executor/wasmi/src/lib.rs
```diff
@@ -306,16 +322,17 @@ impl<'a> wasmi::Externals for FunctionExecutor<'a> {
...
-			.map_err(wasmi::Trap::from)
-			.map(|v| v.map(Into::into))
+
+		if let Some(function) = self.host_functions.get(index) {
+			function.execute(self, &mut args)
+				.map_err(|msg| Error::FunctionExecution(function.name().to_string(), msg))
+				.map_err(wasmi::Trap::from)
+				.map(|v| v.map(Into::into))
+		} else if self.enable_stub {
+			panic!("function {} does not exist", self.method);
```
bkchr: There can be more than 1 function that does not exist.
And please return an `Err`

cecton: If I return an Err it doesn't panic anymore and therefore it breaks my test. Does the test should panic?

client/executor/src/integration_tests/mod.rs
```diff
@@ -68,6 +88,60 @@ fn returning_should_work(wasm_method: WasmExecutionMethod) {
...
 
+#[test_case(WasmExecutionMethod::Interpreted)]
+#[cfg_attr(feature = "wasmtime", test_case(WasmExecutionMethod::Compiled))]
+#[should_panic(expected = "functions do not exist: missing_external, yet_another_missing_external")]
+fn call_not_existing_function_with_stub_enabled(wasm_method: WasmExecutionMethod) {
+	let mut ext = TestExternalities::default();
+	let mut ext = ext.ext();
+	let test_code = WASM_BINARY;
+
+	let _output = call_in_wasm_with_stub(
```
bkchr:
```suggestion
	call_in_wasm_with_stub(
```

client/executor/src/integration_tests/mod.rs
```diff
@@ -68,6 +88,60 @@ fn returning_should_work(wasm_method: WasmExecutionMethod) {
...
+
+#[test_case(WasmExecutionMethod::Interpreted)]
+#[cfg_attr(feature = "wasmtime", test_case(WasmExecutionMethod::Compiled))]
+#[should_panic(expected = "functions do not exist: missing_external, yet_another_missing_external")]
+fn call_yet_another_not_existing_function_with_stub_enabled(wasm_method: WasmExecutionMethod) {
+	let mut ext = TestExternalities::default();
+	let mut ext = ext.ext();
+	let test_code = WASM_BINARY;
+
+	let _output = call_in_wasm_with_stub(
```
bkchr:
```suggestion
	call_in_wasm_with_stub(
```

client/executor/wasmi/src/lib.rs
```diff
@@ -295,9 +306,17 @@ impl<'a> wasmi::ModuleImportResolver for Resolver<'a> {
 			}
 		}
 
-		Err(wasmi::Error::Instantiation(
-			format!("Export {} not found", name),
-		))
+		if self.enable_stub {
+			trace!("Could not find function {}, a stub will be provided instead.", name);
+			self.missing_functions.borrow_mut().push(name.to_string());
```
bkchr: Please also create 1 unique index per function.

client/executor/wasmi/src/lib.rs
```diff
@@ -306,16 +325,19 @@ impl<'a> wasmi::Externals for FunctionExecutor<'a> {
...
-			.map(|v| v.map(Into::into))
+
+		if let Some(function) = self.host_functions.get(index) {
+			function.execute(self, &mut args)
+				.map_err(|msg| Error::FunctionExecution(function.name().to_string(), msg))
+				.map_err(wasmi::Trap::from)
+				.map(|v| v.map(Into::into))
+		} else if self.enable_stub {
+			Err(Error::from(
+				format!("functions do not exist: {}", self.missing_functions.join(", "))
```
bkchr: Please only print the actual function that was tried to be called (by using the unique index generated in `resolve`).

client/executor/wasmi/src/lib.rs
```diff
@@ -561,14 +595,17 @@ impl WasmRuntime for WasmiRuntime {
 				error!(target: "wasm-executor", "snapshot restoration failed: {}", e);
 				e
 			})?;
-		call_in_wasm_module(ext, &self.instance, method, data, &self.host_functions)
+		call_in_wasm_module(
+			ext, &self.instance, method, data, &self.host_functions, self.enable_stub,
```
bkchr: Please put each parameter it its own line.

client/executor/wasmi/src/lib.rs
```diff
@@ -561,14 +595,17 @@ impl WasmRuntime for WasmiRuntime {
...
+			&self.missing_functions)
```
bkchr: `)` needs to go to a new line.

client/executor/wasmi/src/lib.rs
cecton: If I don't print all the not existing functions, why am I gathering them in a Vec? This implementation here was printing the function that was tried: https://github.com/paritytech/substrate/pull/4550/commits/e19948e8efa91f2ef745db609967bc7726285311

bkchr: Because you only want to know which function exactly was called. It does not helps us to know that 50 functions are not provided ^^

cecton: Ok so after discussion it finally :bulb: in my mind. It's the name of the missing export we need, not the name of the function in the module. Will fix this

client/executor/wasmi/src/lib.rs
```diff
@@ -536,6 +577,10 @@ pub struct WasmiRuntime {
 	state_snapshot: StateSnapshot,
 	/// The host functions registered for this instance.
 	host_functions: Vec<&'static dyn Function>,
+	/// Enable STUB for function called that are missing
+	enable_stub: bool,
```
pepyakin: Instead of `enable_stub` I'd like to suggest to use something like `allow_missing_imports`. The rationale is that in the wasm spec a missing import is specified an instantiation error, so `allow_missing_imports` would convey that this is a custom mode. `enable_stub` is on contrary a little bit vague since there might be different kinds of stubs out there.

Aside, expanding the role of this field in the doc would be great!

client/executor/wasmi/src/lib.rs
```diff
@@ -295,9 +319,19 @@ impl<'a> wasmi::ModuleImportResolver for Resolver<'a> {
 			}
 		}
 
-		Err(wasmi::Error::Instantiation(
-			format!("Export {} not found", name),
-		))
+		if self.enable_stub {
+			trace!("Could not find function {}, a stub will be provided instead.", name);
```
pepyakin: Let's add `target` here. Confusingly, it is `target: "wasm-executor"` in this module.

client/executor/wasmi/src/lib.rs
cecton: Fine by me. @bkchr any remark?

client/executor/wasmi/src/lib.rs
cecton: Can you provide me the exact message you want? I'm new :pray:

pepyakin: The message is actually good! I mean something like

```suggestion
			trace!(target: "wasm-executor", "Could not find function {}, a stub will be provided instead.", name);
```

This way it would allow us to filter messages with `RUST_LOG` that come only from this module.

client/executor/wasmi/src/lib.rs
bkchr: I like @pepyakin idea :) Sounds much better!

client/executor/src/integration_tests/mod.rs
```diff
@@ -47,6 +47,26 @@ fn call_in_wasm<E: Externalities>(
 		ext,
 		code,
 		heap_pages,
+		false,
```
bkchr: Won't the other tests fails because of missing imports? :D

client/executor/src/integration_tests/mod.rs
```diff
@@ -68,6 +88,60 @@ fn returning_should_work(wasm_method: WasmExecutionMethod) {
...
+
+	let output = call_in_wasm(
+		"test_calling_missing_external",
+		&[],
+		wasm_method,
+		&mut ext,
+		&test_code[..],
+		8,
+	);
+	assert!(output.is_err());
```
bkchr: Please check for the exact error you expect. Otherwise it could happen that the error type returned changes and the test still finishes successfully.

client/executor/wasmi/src/lib.rs
```diff
@@ -295,9 +319,19 @@ impl<'a> wasmi::ModuleImportResolver for Resolver<'a> {
...
+			let id = self.missing_function_id.borrow().clone();
+			self.missing_functions.borrow_mut().insert(id, name.to_string());
```
bkchr: We actually just require `missing_functions` to be a `Vec`.

The index for the function will be computed as:
```
self.host_functions.len() + self.missing_functions.len()
```
(before adding the new missing function name!)

And in invoke we do the following
```
if index >= host_functions.len() {
     Err("{} not found whatever", self.missing_functions[index - host_functions.len())
}
```
```

Comment (cecton) [reply client/executor/wasmi/src/lib.rs]: Yes that's what I implemented before as you can see here: https://github.com/paritytech/substrate/pull/4550/commits/233399499bd5cf097370b01f20bb66812dbdeb91

But I told you this looks way too clever and I should probably use a hashmap and you said you would have use a hashmap.

I don't mind reverting it but I'm not in favor of this because I find that it doesn't look like it brings performance improvement, maybe memory, but it certainly make the code harder to understand.

Now that I shared my humble opinion. What do you prefer? :grin:

File: client/executor/wasmi/src/lib.rs (hunk 8 above)
Comment (cecton) [reply client/executor/wasmi/src/lib.rs]: done

File: client/executor/wasmi/src/lib.rs (hunk 5 above)
Comment (cecton) [reply client/executor/wasmi/src/lib.rs]: done

File: client/executor/wasmi/src/lib.rs (hunk 4 above)
Comment (cecton) [reply client/executor/wasmi/src/lib.rs]: done

File: client/executor/wasmi/src/lib.rs (hunk 12 above)
Comment (bkchr) [reply client/executor/wasmi/src/lib.rs]: Yeah, I wanted to use a hashmap because I did not thought about this solution :sweat_smile: 
It clearly will improve the performance, as we just need to access the memory by index and not need to look into a hashmap. Please revert, but also make sure that `invoke` does not tries to use an invalid index.

File: client/executor/wasmi/src/lib.rs (hunk 1 above)
Comment (cecton) [reply client/executor/wasmi/src/lib.rs]: Done (now it unwrap() and catch the panic)

File: client/executor/wasmi/src/lib.rs (hunk 6 above)
Comment (cecton) [reply client/executor/wasmi/src/lib.rs]: done

File: client/executor/wasmi/src/lib.rs (hunk 7 above)
Comment (cecton) [reply client/executor/wasmi/src/lib.rs]: done

File: client/executor/src/integration_tests/mod.rs (hunk 10 above)
Comment (cecton) [reply client/executor/src/integration_tests/mod.rs]: done (but I think wasmtime is failing at instantiation because of the missing function. I will fix that)

File: client/executor/src/integration_tests/mod.rs (hunk 11 above)
Comment (cecton) [reply client/executor/src/integration_tests/mod.rs]: Done but I removed the test entirely as we test only with stub enabled

File: client/executor/wasmi/src/lib.rs (hunk 12 above)
Comment (cecton) [reply client/executor/wasmi/src/lib.rs]: done

File: client/executor/wasmi/src/lib.rs (hunk 4 above)
Comment (cecton) [reply client/executor/wasmi/src/lib.rs]: Ok so I think this is good but let me know

**Comment by cecton** [comment]
This is ready to be merged. I don't know why that test failed but it doesn't seem related to my changes.

File: client/executor/src/integration_tests/mod.rs [hunk 13]
```diff
@@ -32,6 +32,67 @@ use crate::WasmExecutionMethod;
...
+
+	pub struct WasmtimeHostFunctions;
+
+	impl HostFunctions for WasmtimeHostFunctions {
+		fn host_functions() -> Vec<&'static dyn Function> {
+			vec![MISSING_EXTERNAL_FUNCTION, YET_ANOTHER_MISSING_EXTERNAL_FUNCTION]
+		}
+	}
+
+	struct MissingExternalFunction;
```
bkchr: You only need one of these structs. Just put the name into the struct as parameter.

client/executor/src/integration_tests/mod.rs
```diff
@@ -40,13 +101,14 @@ fn call_in_wasm<E: Externalities>(
 	code: &[u8],
 	heap_pages: u64,
 ) -> crate::error::Result<Vec<u8>> {
-	crate::call_in_wasm::<E, sp_io::SubstrateHostFunctions>(
+	crate::call_in_wasm::<E, HostFunctions>(
```
bkchr: Use always the host functions with the fake functions.

And revert the `true` below to `false`.

client/executor/wasmi/src/lib.rs
```diff
@@ -536,6 +579,10 @@ pub struct WasmiRuntime {
 	state_snapshot: StateSnapshot,
 	/// The host functions registered for this instance.
 	host_functions: Vec<&'static dyn Function>,
+	/// Enable STUB for function called that are missing
```
bkchr:
```suggestion
	/// Enable stub generation for functions that are not available in `host_functions`.
	/// These stubs will error when the wasm blob tries to call them.
```

client/executor/src/integration_tests/mod.rs
cecton: done

client/executor/wasmi/src/lib.rs
```diff
@@ -22,6 +22,7 @@ use sc_executor_common::{
 	allocator,
 };
 use std::{str, mem};
+use std::cell::RefCell;
```
bkchr:
```suggestion
```

client/executor/wasmi/src/lib.rs
```diff
@@ -22,6 +22,7 @@ use sc_executor_common::{
 	allocator,
 };
 use std::{str, mem};
```
bkchr:
```suggestion
use std::{str, mem, cell::RefCell};
```

client/executor/wasmi/src/lib.rs
```diff
@@ -42,6 +43,8 @@ struct FunctionExecutor<'a> {
 	memory: MemoryRef,
 	table: Option<TableRef>,
 	host_functions: &'a [&'static dyn Function],
+	allow_missing_imports: bool,
+	missing_functions: &'a Vec<String>,
```
bkchr:
```suggestion
	missing_functions: &'a [String],
```

client/executor/wasmi/src/lib.rs
```diff
@@ -50,13 +53,17 @@ impl<'a> FunctionExecutor<'a> {
 		heap_base: u32,
 		t: Option<TableRef>,
 		host_functions: &'a [&'static dyn Function],
+		allow_missing_imports: bool,
+		missing_functions: &'a Vec<String>,
```
bkchr:
```suggestion
		missing_functions: &'a [String],
```

client/executor/wasmi/src/lib.rs
```diff
@@ -295,9 +318,19 @@ impl<'a> wasmi::ModuleImportResolver for Resolver<'a> {
 			}
 		}
 
-		Err(wasmi::Error::Instantiation(
-			format!("Export {} not found", name),
-		))
+		if self.allow_missing_imports {
+			trace!("Could not find function {}, a stub will be provided instead.", name);
```
bkchr:
```suggestion
			trace!(target: "wasm-executor", "Could not find function `{}`, a stub will be provided instead.", name);
```

client/executor/wasmi/src/lib.rs
```diff
@@ -269,14 +276,30 @@ impl<'a> Sandbox for FunctionExecutor<'a> {
...
+	missing_function_id: RefCell<usize>,
+}
+
+impl<'a> Resolver<'a> {
+	fn new(host_functions: &'a[&'static dyn Function], allow_missing_imports: bool) -> Resolver<'a> {
+		Resolver {
+			host_functions,
+			allow_missing_imports,
+			missing_functions: RefCell::new(Vec::new()),
+			missing_function_id: RefCell::new(host_functions.len()),
```
bkchr:
```suggestion
```

client/executor/wasmi/src/lib.rs
```diff
@@ -269,14 +276,30 @@ impl<'a> Sandbox for FunctionExecutor<'a> {
 	}
 }
 
-struct Resolver<'a>(&'a[&'static dyn Function]);
+struct Resolver<'a> {
+	host_functions: &'a[&'static dyn Function],
+	allow_missing_imports: bool,
+	missing_functions: RefCell<Vec<String>>,
+	missing_function_id: RefCell<usize>,
```
bkchr:
```suggestion
```

client/executor/wasmi/src/lib.rs
```diff
@@ -295,9 +318,19 @@ impl<'a> wasmi::ModuleImportResolver for Resolver<'a> {
 			}
 		}
 
-		Err(wasmi::Error::Instantiation(
-			format!("Export {} not found", name),
-		))
+		if self.allow_missing_imports {
+			trace!("Could not find function {}, a stub will be provided instead.", name);
+			let id = self.missing_function_id.borrow().clone();
```
bkchr:
```suggestion
			let id = self.missing_functions.borrow().len() + self.host_functions.len();
```

client/executor/wasmi/src/lib.rs
```diff
@@ -295,9 +318,19 @@ impl<'a> wasmi::ModuleImportResolver for Resolver<'a> {
...
+			self.missing_functions.borrow_mut().push(name.to_string());
+			*self.missing_function_id.borrow_mut() += 1;
```
bkchr:
```suggestion
```

client/executor/wasmi/src/lib.rs
```diff
@@ -295,9 +318,19 @@ impl<'a> wasmi::ModuleImportResolver for Resolver<'a> {
...
+
+			// NOTE: provide purposely an invalid index of the function
```
bkchr:
```suggestion
```

That is no invalid index.

client/executor/wasmi/src/lib.rs
```diff
@@ -306,16 +339,22 @@ impl<'a> wasmi::Externals for FunctionExecutor<'a> {
...
-			.map_err(|msg| Error::FunctionExecution(function.name().to_string(), msg))
-			.map_err(wasmi::Trap::from)
-			.map(|v| v.map(Into::into))
+
+		if let Some(function) = self.host_functions.get(index) {
+			function.execute(self, &mut args)
+				.map_err(|msg| Error::FunctionExecution(function.name().to_string(), msg))
+				.map_err(wasmi::Trap::from)
+				.map(|v| v.map(Into::into))
+		} else if self.allow_missing_imports {
```
bkchr:
```suggestion
		} else if self.allow_missing_imports && index >= self.host_functions.len() && index < self.host_functions.len() + self.missing_functions.len() {
```

client/executor/wasmi/src/lib.rs
```diff
@@ -306,16 +339,22 @@ impl<'a> wasmi::Externals for FunctionExecutor<'a> {
...
+			Err(Error::from(format!(
+				"function {} does not exist",
+				self.missing_functions
```
bkchr:
```suggestion
				self.missing_functions[index - self.host_functions.len()]
```

client/executor/wasmi/src/lib.rs
```diff
@@ -306,16 +339,22 @@ impl<'a> wasmi::Externals for FunctionExecutor<'a> {
...
+					.get(index - self.host_functions.len())
```
bkchr:
```suggestion
```

client/executor/wasmi/src/lib.rs
```diff
@@ -306,16 +339,22 @@ impl<'a> wasmi::Externals for FunctionExecutor<'a> {
...
+					.expect("invalid function index"),
```
bkchr:
```suggestion
```

client/executor/wasmi/src/lib.rs
```diff
@@ -306,16 +339,22 @@ impl<'a> wasmi::Externals for FunctionExecutor<'a> {
...
-			.map(|v| v.map(Into::into))
+
+		if let Some(function) = self.host_functions.get(index) {
+			function.execute(self, &mut args)
+				.map_err(|msg| Error::FunctionExecution(function.name().to_string(), msg))
+				.map_err(wasmi::Trap::from)
+				.map(|v| v.map(Into::into))
+		} else if self.allow_missing_imports {
+			Err(Error::from(format!(
+				"function {} does not exist",
```
bkchr:
```suggestion
				"Function `{}` is only a stub. Calling a stub is not allowed.",
```

client/executor/wasmi/src/lib.rs
```diff
@@ -360,7 +401,8 @@ fn call_in_wasm_module(
 		.and_then(|e| e.as_table().cloned());
 	let heap_base = get_heap_base(module_instance)?;
 
-	let mut fec = FunctionExecutor::new(memory.clone(), heap_base, table, host_functions)?;
+	let mut fec = FunctionExecutor::new(
+		memory.clone(), heap_base, table, host_functions, allow_missing_imports, missing_functions)?;
```
bkchr: Please put every parameter on its own line.

client/executor/wasmi/src/lib.rs
```diff
@@ -579,7 +636,8 @@ pub fn create_instance(
 	let data_segments = extract_data_segments(&code)?;
 
 	// Instantiate this module.
-	let instance = instantiate_module(heap_pages as usize, &module, &host_functions)
+	let (instance, missing_functions) = instantiate_module(
+			heap_pages as usize, &module, &host_functions, allow_missing_imports)
```
bkchr: Each parameter on its own line.

client/executor/wasmi/src/lib.rs
cecton: done

client/executor/wasmi/src/lib.rs
cecton: done

client/executor/wasmi/src/lib.rs
```diff
@@ -306,16 +334,20 @@ impl<'a> wasmi::Externals for FunctionExecutor<'a> {
...
+
+		if let Some(function) = self.host_functions.get(index) {
+			function.execute(self, &mut args)
+				.map_err(|msg| Error::FunctionExecution(function.name().to_string(), msg))
+				.map_err(wasmi::Trap::from)
+				.map(|v| v.map(Into::into))
+		} else if self.allow_missing_imports && index >= self.host_functions.len() && index < self.host_functions.len() + self.missing_functions.len() {
+			Err(Error::from(format!(
+				"Function `{}` is only a stub. Calling a stub is not allowed.",
+				self.missing_functions[index - self.host_functions.len()]
```
bkchr:
```suggestion
				self.missing_functions[index - self.host_functions.len()],
```

client/executor/wasmi/src/lib.rs
```diff
@@ -579,7 +635,12 @@ pub fn create_instance(
...
 
 	// Instantiate this module.
-	let instance = instantiate_module(heap_pages as usize, &module, &host_functions)
+	let (instance, missing_functions) = instantiate_module(
+		heap_pages as usize,
+		&module,
+		&host_functions,
+		allow_missing_imports,
+	)
 		.map_err(|e| WasmError::Instantiation(e.to_string()))?;
```
bkchr:
```suggestion
```

client/executor/wasmi/src/lib.rs
```diff
@@ -579,7 +635,12 @@ pub fn create_instance(
 	let data_segments = extract_data_segments(&code)?;
 
 	// Instantiate this module.
-	let instance = instantiate_module(heap_pages as usize, &module, &host_functions)
+	let (instance, missing_functions) = instantiate_module(
+		heap_pages as usize,
+		&module,
+		&host_functions,
+		allow_missing_imports,
+	)
```
bkchr:
```suggestion
	).map_err(|e| WasmError::Instantiation(e.to_string()))?;
```

PR #4553, opened 2020-01-07
do not chill indirectly-slashed nominators

rphmeier:
cc @burdges, @jacogr 

When a validator is slashed, its nominators are currently kicked until explicit re-nomination. This PR causes that not to happen. The phragmén election already pre-filters for votes cast _after_ a validator's most recent slash, so the nominators' votes on the slashed validator will be ignored until next explicit nomination.

The UX here is not ideal, since those votes on slashed validators will be silently ignored. User apps will be able to perform the same check as the staking code - I've spoken with @jacogr about this before, but pinging again here.

### DIFF
--- a/frame/staking/src/slashing.rs
+++ b/frame/staking/src/slashing.rs
@@ -392,9 +392,9 @@ fn slash_nominators<T: Trait>(
 			);
 
 			if target_span == Some(spans.span_index()) {
-				// Chill the nominator outright, ending the slashing span.
+				// End the span, but don't chill the nominator. its nomination
+				// on this validator will be ignored in the future.
 				spans.end_span(now);
-				<Module<T>>::chill_stash(stash);
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


gavofyork:
No tests to correct?

burdges:
We've worked over the trade offs here fairly well, so if you think this makes sense then go for it.  We should ideally avoid them being able to reoffend for free in future of course, but I gather this is happening in a way that makes that not happen.

Is there some weaker state of which we should elevate the importance?  We'd some scheme for chilling them only when they got slashed enough for example.  Or just not worry about it for now?

rphmeier:
> We should ideally avoid them being able to reoffend for free in future

Is that a risk introduced by this PR? This approximate behavior could always be emulated in the old system by ensuring that a nomination transaction immediately followed the slash report.

> We'd some scheme for chilling them only when they got slashed enough for example

At some point I guess, but the complexity of tracking so much state for each nominator sounds like a lot.

burdges:
> At some point I guess, but the complexity of tracking so much state for each nominator sounds like a lot.

I've now forgotten if this required tracking more state.  I'd believed it was equivalent at some point, but maybe you found something else that needed tracking.  If so, I do not remember cleanly sorting out cleanly if that record sticks with other records in the tree (cheap) or floats off elsewhwere (expensive).

PR #4555, opened 2020-01-07
Add translate function for storage prefixed map.

thiolliere:
Add a function to migrate values from one type to one another.

In case we fail to migrate one element we return an error with the key that has failed to decoded.
We could also remove values that fail to decode and finish the migration and at the end returning an error saying there is some value that has been removed.

### DIFF
--- a/frame/support/src/storage/mod.rs
+++ b/frame/support/src/storage/mod.rs
@@ -431,7 +431,8 @@ pub trait StoragePrefixedMap<Value: FullCodec> {
 	/// `TV` translates values.
 	///
 	/// Returns `Err` if the map could not be interpreted as the old type, and Ok if it could.
-	/// The `Err` contains the first hashed key which could not be migrated.
+	/// The `Err` contains the number of value that couldn't be interpreted, those value are
+	/// removed from the map.
 	///
 	/// # Warning
 	///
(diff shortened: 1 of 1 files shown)


frame/support/src/storage/mod.rs
```diff
@@ -422,6 +425,45 @@ pub trait StoragePrefixedMap<Value: FullCodec> {
...
+	{
+		let prefix = Self::final_prefix();
+		let mut previous_key = prefix.to_vec();
+		while let Some(next_key) = sp_io::storage::next_key(&previous_key)
+			.filter(|n| n.starts_with(&prefix[..]))
+		{
+			let value: OldValue = unhashed::get(&next_key)
+				// We failed to read the value.
+				// TODO TODO: should we remove all value after this ?
+				.ok_or_else(|| next_key.clone())?;
```
thiolliere: in case the value can't be decoded then we return an error, but what should we do then.
In the case of linkedmap if an element fails to decode during the translation then we end the linkedmap at the last correctly decoded element and we return an error.

What about here should we remove this value and continue to decode following values or should we stop at the first failure ?

frame/support/src/storage/mod.rs
```diff
@@ -422,6 +425,43 @@ pub trait StoragePrefixedMap<Value: FullCodec> {
...
+	fn translate_values<OldValue, TV>(translate_val: TV) -> Result<(), Vec<u8>>
+		where OldValue: Decode, TV: Fn(OldValue) -> Value
+	{
+		let prefix = Self::final_prefix();
+		let mut previous_key = prefix.to_vec();
+		while let Some(next_key) = sp_io::storage::next_key(&previous_key)
+			.filter(|n| n.starts_with(&prefix[..]))
+		{
+			let value: OldValue = unhashed::get(&next_key)
+				// We failed to read the value. Stop the translation and return an error.
```
thiolliere: in case the value can't be decoded then we return an error, but what should we do then.
In the case of linkedmap if an element fails to decode during the translation then we end the linkedmap at the last correctly decoded element and we return an error.

What about here should we remove this value and continue to decode following values or should we stop at the first failure ?

thiolliere:
related comment in regard to which implementation in case of error: https://github.com/paritytech/substrate/pull/4052#issuecomment-552399173

frame/support/src/storage/mod.rs
shawntabrizi: > What about here should we remove this value and continue to decode following values

This makes sense. Stopping at the first failure means that half of the storage will be one type and the other half another type. This would also prevent us from ever being able to call `translate_value` again.

thiolliere:
now key that couldn't be decoded are removed and an error is returned with the number of key that failed to decode.

frame/support/src/storage/mod.rs
```diff
@@ -472,9 +521,27 @@ mod test {
 
 			assert_eq!(MyStorage::iter().collect::<Vec<_>>(), vec![1, 2, 3, 4]);
 
+			// test removal
 			MyStorage::remove_all();
-
 			assert_eq!(MyStorage::iter().collect::<Vec<_>>(), vec![]);
+
+			// test migration
+			unhashed::put(&[&k[..], &vec![1][..]].concat(), &1u128);
```
bkchr: You inserting a `u128` and use `u32` in the translate function. You should actually store a `u32`.

And then call `iter()` to make sure that an empty vec is returned.

Than you should to the translation.

thiolliere: oh yes I updated tests

shawntabrizi:
> EDIT: actually maybe better not to remove them and just let them untranslated in the storage, anyway this will only happen if the storage is corrupted or the translate function expect wrong type. in both case I don't think we can solve anything from our side.

If you leave them untranslated, runtime code could really get messed up. Better to remove data then to have bad data where the runtime assumes it is good. imo

PR #4557, opened 2020-01-07
Fixed state pinning in state-db

arkpar:
Fixed a couple of potential soundness issues with pinned states in states in state-db

* Pinned state would not pin parent state, so it could still be discarded 
* Pinned state nodes still could be discarded if they had more than one references and only one of the referencing states is pinned.

Fixes https://github.com/paritytech/polkadot/issues/705

### DIFF
--- a/client/state-db/src/noncanonical.rs
+++ b/client/state-db/src/noncanonical.rs
@@ -37,7 +37,9 @@ pub struct NonCanonicalOverlay<BlockHash: Hash, Key: Hash> {
 	pending_canonicalizations: Vec<BlockHash>,
 	pending_insertions: Vec<BlockHash>,
 	values: HashMap<Key, (u32, DBValue)>, //ref counted
-	pinned: HashMap<BlockHash, HashMap<Key, DBValue>>, //would be deleted but kept around because block is pinned
+	//would be deleted but kept around because block is pinned, ref counted.
+	pinned: HashMap<BlockHash, u32>,
+	pinned_insertions: HashMap<BlockHash, Vec<Key>>,
 }
 
 #[derive(Encode, Decode)]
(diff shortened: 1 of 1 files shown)


client/state-db/src/noncanonical.rs
```diff
@@ -97,25 +92,32 @@ fn discard_descendants<BlockHash: Hash, Key: Hash>(
...
 	let mut discarded = Vec::new();
 	if let Some(level) = levels.get_mut(index) {
 		*level = level.drain(..).filter_map(|overlay| {
 			let parent = parents.get(&overlay.hash).expect("there is a parent entry for each entry in levels; qed").clone();
 			if parent == *hash {
-				parents.remove(&overlay.hash);
-				discarded.push(overlay.hash);
-				discard_values(&mut values, overlay.inserted, pinned.get_mut(hash));
+				discarded.push(overlay.hash.clone());
+				if pinned.contains_key(hash) {
```
svyatonik: Shouldn't it be `if pinned.contains_key(&overlay.hash)`? And `pinned_insertions.insert(overlay.hash.clone(), overlay.inserted);` two lines below?

client/state-db/src/noncanonical.rs
```diff
@@ -821,4 +851,64 @@ mod tests {
...
+		db.commit(&overlay.insert::<io::Error>(&h_11, 1, &H256::default(), c_11).unwrap());
+		db.commit(&overlay.insert::<io::Error>(&h_12, 1, &H256::default(), c_12).unwrap());
+		db.commit(&overlay.insert::<io::Error>(&h_21, 2, &h_11, c_21).unwrap());
+
+		overlay.pin(&h_21);
+
+		let mut commit = CommitSet::default();
+		overlay.canonicalize::<io::Error>(&h_12, &mut commit).unwrap();
+		db.commit(&commit);
+		overlay.apply_pending(); // 1_1 and 2_1 should be both pinned
```
svyatonik: What if we change this to `revert_pending()`. This method removes from `parents` unconditionally => when we unpin `h_21` later, states `1_1` and `0` won't be unpinned because there's no entry in `parents` for `h_21`.

This may not be an actual issue in current impl, because we only call `revert_pending()` if commit of block import operation has failed (at least that's my understanding) => no one is actually able to get state before block is inserted. But at state-db level, it may worth at least `debug_assert!()`. Or maybe not - just thought about that during review :)

arkpar: That's right. Pending states are not supposed to be pinned. I'll add a check for that.

PR #4561, opened 2020-01-07
Fix timer panics in the wasm light client

expenses:
Last change to get light clients to work :tada:! This PR relies on @tomaka 's wasm-timer crate, but only for replacements for `std::time::Instant` and `std::time::SystemTime`.

~Waiting on https://github.com/async-rs/futures-timer/pull/51.~


utils/browser/Cargo.toml
```diff
@@ -16,10 +16,10 @@ js-sys = "0.3.34"
 wasm-bindgen = "0.2.57"
 wasm-bindgen-futures = "0.4.7"
 kvdb-web = "0.3"
-sc-cli = { path = "../../client/cli", default-features = false }
-sc-service = { path = "../../client/service", default-features = false }
-sc-network = { path = "../../client/network" }
-sc-chain-spec = { path = "../../client/chain-spec" }
+cli = { package = "sc-cli", path = "../../client/cli", default-features = false }
```
tomaka: Why renames suddenly?

expenses: The other dependencies in `browser-utils` were renamed from `sc-x` to `x` in a merge, so I thought I'd do the same.

tomaka: There are now only 7 occurrences within Substrate where we rename the packages, 3 of which are here. So I think I'd not rename them.

expenses:
This PR is almost ready. But I think the switch from libp2p `0.14.0-alpha.1` to `0.15.0` seems to have broken the connection between the light client and other nodes, but without displaying any errors.

client/informant/src/display.rs
```diff
@@ -66,15 +69,29 @@ impl<B: BlockT> InformantDisplay<B> {
 			(SyncState::Downloading, Some(n)) => (format!("Syncing{}", speed), format!(", target=#{}", n)),
 		};
 
+		// Colour informant output if we're running natively.
+
+		#[cfg(not(target_os = "unknown"))]
```
tomaka: As usual, platform-specific code really isn't a great thing.
I'd prefer to have a runtime flag that allows enabling/disabling colour output.

client/network/src/transport.rs
```diff
@@ -126,11 +126,18 @@ pub fn build_transport(
...
 				.map_ok(|(id, muxer)| (id, core::muxing::StreamMuxerBox::new(muxer)))
-		})
+		});
 
-		.timeout(Duration::from_secs(20))
-		.map_err(|err| io::Error::new(io::ErrorKind::Other, err))
-		.boxed();
+	let transport = if cfg!(not(target_os = "unknown")) {
+		transport
+			.timeout(Duration::from_secs(20))
```
tomaka: Why not have the timeout with `unknown`? There's no other timeout anywhere in the stack, and no timeout means that we just continue trying to connect **forever** and never try a different address or a different node.

client/service/src/lib.rs
```diff
@@ -340,7 +341,10 @@ impl<TBl: Unpin, TCl, TSc: Unpin, TNetStatus, TNet, TTxPool, TOc> Future for
 		}
 
 		while let Poll::Ready(Some((task_to_spawn, name))) = Pin::new(&mut this.to_spawn_rx).poll_next(cx) {
-			(this.task_executor)(Box::pin(futures_diagnose::diagnose(name, task_to_spawn)));
+			#[cfg(not(target_os = "unknown"))]
+			let task_to_spawn = futures_diagnose::diagnose(name, task_to_spawn);
```
tomaka: Is `futures_diagnose` not compiling for `wasm32-unknown`?

client/telemetry/src/worker.rs
```diff
@@ -98,10 +98,10 @@ impl TelemetryWorker {
 						.map_ok(|data| BytesMut::from(data.as_ref()));
 					future::ready(Ok::<_, io::Error>(connec))
 				})
-		});
+		})
+			.timeout(CONNECT_TIMEOUT);
 
 		let transport = transport
-			.timeout(CONNECT_TIMEOUT)
```
tomaka: Same remark as above.

client/service/src/lib.rs
expenses: It does compile, but it panics because of `std::time::Instant::now()`.

client/network/src/transport.rs
expenses: Also panics because of `std::time::Instant::now()`.

client/informant/src/display.rs
DemiMarie-parity: On the one hand, I agree.  On the other hand, what if the flag is enabled when compiling for wasm?  That cannot possibly work.  So we need to give an error here.

client/network/src/protocol/legacy_proto/handler.rs
```diff
@@ -348,13 +348,12 @@ where
 
 			ProtocolState::Init { substreams, mut init_deadline } => {
 				match Pin::new(&mut init_deadline).poll(cx) {
-					Poll::Ready(Ok(())) => {
+					Poll::Ready(()) => {
```
DemiMarie-parity: :+1:

client/network/src/transport.rs
DemiMarie-parity: Is that still the case?

client/service/src/lib.rs
DemiMarie-parity: Can we make it not panic?

client/telemetry/src/worker.rs
```diff
@@ -98,10 +98,10 @@ impl TelemetryWorker {
 						.map_ok(|data| BytesMut::from(data.as_ref()));
 					future::ready(Ok::<_, io::Error>(connec))
 				})
-		});
+		})
+			.timeout(CONNECT_TIMEOUT);
```
DemiMarie-parity: nit: should be on the previous line

client/network/src/transport.rs
expenses: It was the case when I tested it 4 hours ago, yeah.

client/service/src/lib.rs
expenses: Yep, if we use `wasm-timer` in it.

client/informant/src/display.rs
tomaka: > That cannot possibly work.

Adding colours is just about inserting control characters within the string. If the flag is enabled on WASM, everything will still work perfectly.

expenses: I've added a `Configuration` option for this, but not a cli flag.

client/network/src/transport.rs
expenses: See https://github.com/libp2p/rust-libp2p/pull/1426.

client/cli/src/runtime.rs
```diff
@@ -121,7 +122,7 @@ where
 
 	let service = service_builder(config)?;
 
-	let informant_future = sc_informant::build(&service);
+	let informant_future = sc_informant::build(&service, colour_output);
```
tomaka: You can just put `true` here all the time. No need to add a field to `Configuration`.

client/informant/src/lib.rs
```diff
@@ -26,10 +26,10 @@ use std::time::Duration;
 mod display;
 
 /// Creates an informant in the form of a `Future` that must be polled regularly.
-pub fn build(service: &impl AbstractService) -> impl futures::Future<Output = ()> {
+pub fn build(service: &impl AbstractService, colour_output: bool) -> impl futures::Future<Output = ()> {
```
tomaka: An enum instead of a `bool` would also be appreciated.

client/service/src/lib.rs
expenses: See https://github.com/tomaka/futures-diagnose/pull/4.

expenses:
On ice as we're waiting for https://github.com/libp2p/rust-libp2p/issues/1414.

tomaka:
Why not merge this? It's basically ready.

expenses:
@tomaka it say that you've still got requested changes apparently.

client/informant/Cargo.toml
```diff
@@ -16,3 +16,4 @@ sc-network = { version = "0.8", path = "../network" }
 sc-service = { version = "0.8", default-features = false, path = "../service" }
 sp-blockchain = { version = "2.0.0", path = "../../primitives/blockchain" }
 sp-runtime = { version = "2.0.0", path = "../../primitives/runtime" }
+parity-util-mem = { version = "0.5.1", default-features = false, features = ["primitive-types"] }
```
tomaka: Nit: could you order alphabetically?

client/service/src/builder.rs
```diff
@@ -982,6 +983,10 @@ ServiceBuilder<
 				"disk_read_per_sec" => info.usage.as_ref().map(|usage| usage.io.bytes_read).unwrap_or(0),
 				"disk_write_per_sec" => info.usage.as_ref().map(|usage| usage.io.bytes_written).unwrap_or(0),
 			);
+			#[cfg(not(target_os = "unknown"))]
```
tomaka: Why is that now needed?

expenses: Because if I did that, it would have panicked when you started the light client. We now need to do it this way because you can't call `parity_util_mem::malloc_size` on wasm because `wasm_timer::wasm::Instant` doesn't impl `MallocSizeOf`.

PR #4562, opened 2020-01-07
State-db I/o metrics

NikVolf:
This allow collecting state cache requests and report cache hit ratio
@arkpar 

During the initial sync hit ratio seems to be close to 100%, by the way :)


client/db/src/stats.rs
```diff
@@ -0,0 +1,82 @@
+use std::sync::atomic::{AtomicU64, Ordering as AtomicOrdering};
```
bkchr: Copyright header missing.

client/api/src/client.rs
```diff
@@ -139,13 +143,15 @@ pub struct UsageInfo {
 impl fmt::Display for UsageInfo {
 	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
 		write!(f,
-			"caches: ({} state, {} db overlay), i/o: ({} tx, {} write, {} read, {} tx size)",
+			"caches: ({} state, {} db overlay), i/o: ({} tx, {} write, {} read, {} avg tx, {}/{} state cache read / total)",
```
bkchr: I really don't understand this output :sweat_smile: 

The first caches prints the size? In bytes? Maybe show in mb or something? At least append `b` or whatever to know what it is.

And this newly added state cache reads and state reads is also not understandable if you not read the code.

NikVolf: Those are key cache reads to total reads, this is just technical trace
but I will try improve on output, thanks!

NikVolf: Updated

client/db/src/stats.rs
```diff
@@ -0,0 +1,100 @@
...
+
+		fn unit(ops: &AtomicU64, bytes: &AtomicU64) -> UsageUnit {
+			UsageUnit { ops: ops.swap(0, AtomicOrdering::Relaxed), bytes: bytes.swap(0, AtomicOrdering::Relaxed) }
+		}
+
+		sp_state_machine::UsageInfo {
+			reads: unit(&self.reads, &self.bytes_read),
+			writes: unit(&self.writes, &self.bytes_written),
+			cache_reads: unit(&self.reads_cache, &self.bytes_read_cache),
+			memory: 0, // TODO:
```
marcio-diaz: TODO:?

client/db/src/lib.rs
```diff
@@ -886,7 +889,8 @@ pub struct Backend<Block: BlockT> {
 	shared_cache: SharedCache<Block, Blake2Hasher>,
 	import_lock: RwLock<()>,
 	is_archive: bool,
-	io_stats: FrozenForDuration<kvdb::IoStats>,
+	io_stats: FrozenForDuration<(kvdb::IoStats, StateUsageInfo)>,
+	state_usage: UsageStats,
```
marcio-diaz: `UsageStats` -> `StateStats`, maybe?

client/db/src/lib.rs
```diff
@@ -1242,13 +1247,23 @@ impl<Block: BlockT<Hash=H256>> Backend<Block> {
 
 			let finalized = if operation.commit_state {
 				let mut changeset: sc_state_db::ChangeSet<Vec<u8>> = sc_state_db::ChangeSet::default();
+				let mut ops: u64 = 0;
+				let mut bytes: u64 = 0;
 				for (key, (val, rc)) in operation.db_updates.drain() {
 					if rc > 0 {
+						ops += 1;
+						bytes += key.len() as u64 + val.len() as u64;
+
```
marcio-diaz: I feel there are too many `u64`. But maybe is me :).

client/db/src/stats.rs
NikVolf: It's rather complicated todo, I will write out

client/db/src/lib.rs
NikVolf: theoretically they can overflow when added before conversion, so it is a safer construct :)

client/db/src/lib.rs
```diff
@@ -1257,13 +1261,23 @@ impl<Block: BlockT> Backend<Block> {
...
 						changeset.inserted.push((key, val.to_vec()));
 					} else if rc < 0 {
+						ops += 1;
+						bytes += key.len() as u64;
+
 						changeset.deleted.push(key);
 					}
 				}
+				self.state_usage.tally_writes(ops, bytes);
```
marcio-diaz: Is this valid only for `PrunningMode::ArchiveAll`? Maybe you should make it explicit in that case.

client/db/src/storage_cache.rs
```diff
@@ -495,19 +499,25 @@ impl<S: StateBackend<HasherFor<B>>, B: BlockT> StateBackend<HasherFor<B>> for Ca
...
+			)
 		}
 		let mut cache = self.cache.shared_cache.lock();
 		if Self::is_allowed(Some(key), None, &self.cache.parent_hash, &cache.modifications) {
 			if let Some(entry) = cache.lru_storage.get(key).map(|a| a.clone()) {
 				trace!("Found in shared cache: {:?}", HexDisplay::from(&key));
-				return Ok(entry)
+				return Ok(
+					self.usage.tally_key_read(key, entry, true)
+				)
```
marcio-diaz: This looks a it strange to me, I would prefer that you do the tally and then return `Ok(entry)`. Same with others.

client/db/src/stats.rs
```diff
@@ -0,0 +1,103 @@
...
+/// Accumulated usage statistics for state queries.
+pub struct StateUsageStats {
+	started: std::time::Instant,
+	reads: AtomicU64,
+	bytes_read: AtomicU64,
+	writes: AtomicU64,
+	bytes_written: AtomicU64,
+	reads_cache: AtomicU64,
+	bytes_read_cache: AtomicU64,
+}
```
marcio-diaz: Maybe add `reads/writes` to `telemetry` and `record_metrics` if they are missing.

arkpar:
I'd rename `usage.rs` to `usage_stats.rs` or simply `stats.rs`

client/db/src/storage_cache.rs
```diff
@@ -296,10 +297,12 @@ pub struct CacheChanges<B: BlockT> {
 /// in `sync_cache` along with the change overlay.
 /// For non-canonical clones local cache and changes are dropped.
 pub struct CachingState<S: StateBackend<HasherFor<B>>, B: BlockT> {
+	/// Usage statistics
+	usage: StateUsageStats,
 	/// Backing state.
 	state: S,
 	/// Cache data.
-	pub cache: CacheChanges<B>,
+	pub cache: CacheChanges<H, B>,
```
arkpar: `CacheChanges` still only takes one generic param

client/db/src/storage_cache.rs
NikVolf: It just passes value through counter, why strange?

client/db/src/lib.rs
NikVolf: I don't think so, it is valid for any prunning mode

client/db/src/stats.rs
NikVolf: Yep, I'll open another PR for that when I figure out telemetry, ok?

client/db/src/storage_cache.rs
marcio-diaz: It's just a nit :). Because the return value of the function doesn't depend on `tally_key_read` as it looks. And a reader (like me) may be forced to check the definition unnecessarily. Also, I think is not very stylistic to write `Ok(func(..))`.

client/db/src/stats.rs
marcio-diaz: Ah, isn't it just adding the lines here? https://github.com/paritytech/substrate/blob/4ed0ad6fbd8b25b63105dc21710ace3c329de0d1/client/service/src/builder.rs#L968

NikVolf: Yeah I added previous metrics there, just unsure if anything in telemetry needs or displays those

marcio-diaz: I see, I know that for `record_metrics` is ok, and can be displayed using grafana.

client/db/src/light.rs
```diff
@@ -576,6 +576,9 @@ impl<Block> LightBlockchainStorage<Block> for LightStorage<Block>
 				writes: io_stats.writes,
 				reads: io_stats.reads,
 				average_transaction_size: io_stats.avg_transaction_size() as u64,
+				/// Light client does not track those
```
gavofyork:
```suggestion
				// Light client does not track those
```

(`///` would associate it with just one of "those" two fields)

client/db/src/lib.rs
```diff
@@ -1257,13 +1261,23 @@ impl<Block: BlockT> Backend<Block> {
 
 			let finalized = if operation.commit_state {
 				let mut changeset: sc_state_db::ChangeSet<Vec<u8>> = sc_state_db::ChangeSet::default();
+				let mut ops: u64 = 0;
+				let mut bytes: u64 = 0;
 				for (key, (val, rc)) in operation.db_updates.drain() {
 					if rc > 0 {
+						ops += 1;
```
cheme: Field db_updates are calculated trie nodes delta that are going to be written in rocksdb, so I think we should change the name, because it gives the wrong idea that the collected read stats (key value) are comparable with those writes (trie nodes).
We can also get write stats that correspond to our read, that would be field 'storage_updates' and 'child_storage_updates', IIRC those are send to storage cache for update and are complete in term of delta.
But we could also hook some 'write_cache' stat in state-machine overlay-db to compare what are actual post block modif ('storage_updates' fields) against what update the runtime did call. That way, we get an idea of the proportion of updates that are done on a same key. (at https://github.com/paritytech/substrate/blob/65ad8e9632f1a627fc9fb386dc5b60a54741d7a0/primitives/state-machine/src/overlayed_changes.rs#L235 and the child one)

client/db/src/lib.rs
```diff
@@ -1257,13 +1261,23 @@ impl<Block: BlockT> Backend<Block> {
...
+						bytes += key.len() as u64 + val.len() as u64;
+
 						changeset.inserted.push((key, val.to_vec()));
 					} else if rc < 0 {
+						ops += 1;
+						bytes += key.len() as u64;
+
```
cheme: We could store that in a different write counter.
With non archive mode it is actually an additional memory consumption until pruning, but with archive mode it is a no-op.
Or we check mode (it is maybe accessible in client-db).

client/db/src/storage_cache.rs
```diff
@@ -539,17 +547,25 @@ impl<S: StateBackend<HasherFor<B>>, B: BlockT> StateBackend<HasherFor<B>> for Ca
...
+				return Ok(
+					self.usage.tally_child_key_read(&key, entry, true)
+				)
 			}
 		}
 		trace!("Cache miss: {:?}", key);
 		let value = self.state.child_storage(storage_key, child_info, &key.1[..])?;
+
+		// just pass it through the usage counter
+		let value =	self.usage.tally_child_key_read(&key, value, false);
```
cheme: So we have something like state_read_cache - state_cache = state-read-db-backend.
One thing that we do not register regarding read query emitted from the runtime will be all scenario such as :
- querying a key that has been change during block processing: in this case the overlay-db from state machine return the value and we do not have stats.
- that's all I can think off :)
We could register them in https://github.com/paritytech/substrate/blob/65ad8e9632f1a627fc9fb386dc5b60a54741d7a0/primitives/state-machine/src/ext.rs#L157 I think it should be register as a different cache and added to the total read as it is done for storage cache.

PR #4563, opened 2020-01-07
grandpa: guarantee that vote limit is never lower than vote base

andresilva:
GRANDPA needs to keep track of all pending authority set changes across all forks, in order to limit its votes to not be past any authority set change block. If we have multiple pending changes across forks, we restrict our votes to the earliest one.

With BABE sessions are defined in number of slots (not blocks), which means that authority set changes can happen at different block numbers on different forks. Therefore we need to choose the correct limit according to the given round vote base block, to guarantee that we don't pick a lower limit from some other fork.


rphmeier:
late to the party, but LGTM

PR #4565, opened 2020-01-08
txpool: unify client generic name

kigawas:


bkchr:
Not sure about this change, while it is normally the `Client` that implements these traits it could theoretically be everything that provides the implementation.

kigawas:
Just imitating the `maintainer.rs` file in the same folder

Issue #2553, opened 2019-05-13
Fix the event topic list inefficiency


pepyakin:
With #2491 merged, we maintain a list of topics. Each item is paired with the block number in which the item was deposited. We did that to fix [the issue](https://github.com/paritytech/substrate/pull/2491#issuecomment-490866204) when a block has identical contents for two blocks, the notifications won't be triggered for the second block. 

However, this scheme is clumsy, redundant and inefficient, it just was the easiest way to leverage `::append`. A better solution would be to put a block number only once instead for each item.

pepyakin:
Another possible solution is to introduce another map similar to `EventTopics` and store some value there which is dependent on the current block (i.e. the simplest - the block number) for each topic, and make light clients to watch that as the primary way of fetching updates. A light client would load the list of event indexes for the topics only if the value in that extra map were changed.

cheme:
regarding #2491, I wonder why a storage double map is used for `EventTopics`, wouldn't a simple map be fine?

pepyakin:
I think I should’ve mentioned this in the docs: but this is the simplest way to make all storage locations to have the same prefix in tree. Thus it will allow us e.g. to use ‘kill_prefix’ there.

cheme:
ok, I get it, I think we can optimize that by using `kill_prefix` over the internal prefix generated by the macro for a simple map (a new function may be needed)

shawntabrizi:
@athei Can you look at this and figure a priority?

athei:
The inefficiency I see here is one additional integer per event emitted per integer. It doesn't contribute to storage bloat because this storage item is cleared in `on_initialize`. Also the process of emitting an event is dominated by the storage access. I wager that the single int does not contribute much.

I rate the priority pretty low.

PR #2491, opened 2019-05-06
Light client friendly events

pepyakin:
This PR proposes a solution for light client friendly SRML events.

Each deposited event can specify zero or more topics. A topic is a fixed sized value (atm, is just a hash). When an event is deposited with some topic, the runtime adds the index of this event in the `<Events<T>>` list to a mapping that maps from topics to lists of indexes.

Light clients can "subscribe" to changes of specific storage values (see #628). The idea is that light clients would subscribe to changes of the storage values for a topic they are interested in. When the light client detects a change, it fetches the contents of the indexes list for the specific topic and with this the light client can query events' data by the indexes.

Marking this as draft, since I don't know if this is the way that we want to go, I didn't test this with light-client nor inspected the storage values and there are some questions left

PR #4566, opened 2020-01-08
Use single map and `remove_all` for `EventTopics`

shawntabrizi:
https://github.com/paritytech/substrate/pull/2491 introduced light client friendly events.

It used a trick with the `double_map` api where it was able to clear a single map by removing a single prefix of the `double_map`.

See: https://github.com/paritytech/substrate/issues/2553#issuecomment-493411639

Now that all storage items use `StoragePrefixedMap`, we can now access the `remove_all` API directly from a single map.

This updates the implementation of `EventTopics` to use a single map and the `remove_all` api rather than the `double_map` hack.

### DIFF
--- a/frame/system/src/lib.rs
+++ b/frame/system/src/lib.rs
@@ -405,9 +405,6 @@ decl_storage! {
 		/// Mapping between a topic (represented by T::Hash) and a vector of indexes
 		/// of events in the `<Events<T>>` list.
 		///
-		/// The first key serves no purpose. This field is declared as double_map just
-		/// for convenience of using `remove_prefix`.
-		///
 		/// All topic vectors have deterministic storage locations depending on the topic. This
 		/// allows light-clients to leverage the changes trie storage tracking mechanism and
 		/// in case of changes fetch the list of events of interest.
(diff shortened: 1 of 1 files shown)


frame/system/src/lib.rs
```diff
@@ -405,18 +405,14 @@ decl_storage! {
...
 		/// All topic vectors have deterministic storage locations depending on the topic. This
 		/// allows light-clients to leverage the changes trie storage tracking mechanism and
 		/// in case of changes fetch the list of events of interest.
 		///
 		/// The value has the type `(T::BlockNumber, EventIndex)` because if we used only just
 		/// the `EventIndex` then in case if the topic has the same contents on the next block
 		/// no notification will be triggered thus the event might be lost.
-		EventTopics get(fn event_topics): double_map hasher(blake2_256) (), blake2_256(T::Hash)
-			=> Vec<(T::BlockNumber, EventIndex)>;
+		EventTopics get(fn event_topics): map T::Hash => Vec<(T::BlockNumber, EventIndex)>;
```
shawntabrizi: We could reduce the complexity of the hasher used on the key here.

I don't quite understand if this is something worth changing.

bkchr:
CC @jacogr

PR #4569, opened 2020-01-08
Build wasmtime support by default

arkpar:
Polkadot builds wasmtime support by default. Substrate node could do as well.

### DIFF
--- a/bin/node/cli/Cargo.toml
+++ b/bin/node/cli/Cargo.toml
@@ -105,7 +105,7 @@ structopt = "=0.3.7"
 vergen = "3.0.4"
 
 [features]
-default = ["cli"]
+default = ["cli", "wasmtime"]
 browser = [
 	"browser-utils",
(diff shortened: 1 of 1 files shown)


pepyakin:
I am not sure if this is a good idea. Since I overlooked the act of enabling it in polkadot and it is already there I won't be blocking this PR though.

But note that wasmtime has some questionable unsafe code (and unsafe code can theoretically non-local effects (although unlikely)) and it does not compile on ARM platforms.

gavofyork:
For ARM platforms, we can allow it to be disabled, surely? It would be nice to get rid of the unsafe code, but removal of that hasn't been forthcoming, it has been running fine on Kusama for months and we really do need the speedups for parachains.

pepyakin:
It would be hard to disable, but I think we can just fix so we will be able to compile it. I will prioritise brushing unsafe code.

PR #4571, opened 2020-01-08
Add test for cli keystore path generation

bkchr:


client/service/src/config.rs
```diff
@@ -107,17 +107,29 @@ pub struct Configuration<C, G, E = NoExtension> {
 /// Configuration of the client keystore.
 #[derive(Clone)]
 pub enum KeystoreConfig {
+	/// No config supplied.
+	None,
```
gavofyork: 👍

PR #4575, opened 2020-01-09
Update `per_thing` docs

shawntabrizi:
The macro which generates "generic" per-things had hard written comment about `permill`.

Updated.

### DIFF
--- a/primitives/arithmetic/src/per_things.rs
+++ b/primitives/arithmetic/src/per_things.rs
@@ -69,7 +69,7 @@ macro_rules! implement_per_thing {
 				Self::from_rational_approximation(p, q)
 			}
 
-			/// Converts from a fraction.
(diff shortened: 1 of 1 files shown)


primitives/arithmetic/src/per_things.rs
```diff
@@ -69,7 +69,7 @@ macro_rules! implement_per_thing {
 				Self::from_rational_approximation(p, q)
 			}
 
-			/// Converts a fraction into `Permill`.
+			/// Converts from a fraction.
```
bkchr:
```suggestion
			/// Converts a fraction into `Self`.
```

bkchr: IMHO "Converts from" sounds weird, but if you say that is correct, you can ignore this as well.

PR #4576, opened 2020-01-09
Modify doublemap syntax

thiolliere:
in the past the syntax for specifying the hasher for key1 was different from for key2.

Now key2 is same as key1: an optional `hasher($hash)` attribute that default to blake2_256 instead of `$hash($MyKey2Type)`

Maybe we can also rename this attribute with hasher1 and hasher2 if wanted.

Something not good is that the error message for the old syntax is not understandable because of the way decl_storage is parsing. maybe we should fix it before. Also this needs to be made before 2.0 otherwise we need to support both syntax.


frame/support/test/tests/final_keys.rs
```diff
@@ -71,12 +71,12 @@ mod instance {
...
 
-			pub DoubleMap: double_map u32, blake2_256(u32) => u32;
-			pub DoubleMap2: double_map hasher(twox_128) u32, blake2_128(u32) => u32;
+			pub DoubleMap: double_map u32, u32 => u32;
+			pub DoubleMap2: double_map hasher(twox_128) u32, hasher(blake2_128) u32 => u32;
 
 			pub TestGenericValue get(fn test_generic_value) config(): Option<T::BlockNumber>;
 			pub TestGenericDoubleMap get(fn foo2) config(test_generic_double_map):
-				double_map u32, blake2_256(T::BlockNumber) => Option<u32>;
+				double_map u32, hasher(blake2_256) T::BlockNumber => Option<u32>;
```
bkchr:
```suggestion
				double_map u32, T::BlockNumber => Option<u32>;
```

frame/support/test/tests/final_keys.rs
```diff
@@ -41,12 +41,12 @@ mod no_instance {
...
```
bkchr:
```suggestion
				double_map u32, T::BlockNumber => Option<u32>;
```

thiolliere:
> This is _extremely_ sensitive code. Any mistakes will brick Kusama at the next update. What have you done to make absolutely certain that this changes no logic whatsoever?

Indeed:
* the old syntax cannot compile as `$hash($type)` can't be parsed as a type
* logic in decl_storage hasn't changed, only the parsing.
* to check how current doublemap has been updated, one must be careful that the `$hash($type)` is translated to `hasher($hash) $type` or just `$type` if hasher is blake2_256

(PR for polkadot is on its way)

gavofyork:
can we compare binaries to ensure the code generated is exactly the same?

bkchr:
So, the syntax changed from:
`pub DataDM config(test_config): double_map hasher(twox_64_concat) u32, blake2_256(u32) => u64;`

to:
`pub DataDM config(test_config): double_map hasher(twox_64_concat) u32, u32 => u64;`

or more explicit:
`pub DataDM config(test_config): double_map hasher(twox_64_concat) u32, hasher(blake2_256) u32 => u64;`

The first `hasher` was already optional and defaulted to `blake2_256` when not given. So, we only need to look at newly introduced second hasher. If the second one is not given, it also defaults to `blake2_256` now. We only need to convert the old syntax from `twox_64_concat(u32)` to `hasher(twox_64_concat) u32` or `blake2_256(u32)` to `u32` (as `blake2_256` is the default value).

As @thiolliere already said, there are no more changes in the macro, just the parsing. This includes code generation, it is the same as before.

PR #4577, opened 2020-01-09
Add missing weights for Identity module.

marcio-diaz:
Afaik all dispatchable functions need a weight annotation.

Even though weights are gonna be updated later, I prefer to fix it now so we don't forget.

The weight is copy pasted from others functions in the module.

### DIFF
--- a/frame/identity/src/lib.rs
+++ b/frame/identity/src/lib.rs
@@ -540,6 +540,7 @@ decl_module! {
 		/// - At most O(2 * S + 1) storage mutations; codec complexity `O(1 * S + S * 1)`);
 		///   one storage-exists.
 		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedNormal(50_000)]
 		fn set_subs(origin, subs: Vec<(T::AccountId, Data)>) {
 			let sender = ensure_signed(origin)?;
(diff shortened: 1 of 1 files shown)


frame/identity/src/lib.rs
```diff
@@ -540,6 +540,7 @@ decl_module! {
 		/// - At most O(2 * S + 1) storage mutations; codec complexity `O(1 * S + S * 1)`);
 		///   one storage-exists.
 		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedNormal(50_000)]
```
kianenigma: Looking at the functions below, it looks like a few of them are `O(R) + 1 storage`, and then the rest, like this one, being `O(R, something_else_bounded) + (+1) storage`. But I only read the comments. Maybe the code is reflecting somethign else. 

What I did for the first version was: define a few weights based on the variance of complexity in the module. Assign functions to buckets based on complexity. This is super oversimplified, but at least it is relatively fair. A function taking more compute is likely to be assigned to a higher bucket with more weight. 

at the moment this pallet is just clustered by 10,000 and 50,000, which I think is too little. maybe we can make it 10k, 50k, 75k and 100k?

PR #2802, opened 2019-06-05
New sessions, kill consensus module

gavofyork:
Prototype of new sessions module. This does all the stuff discussed, and adds more stuff needed:

- No duplicate keys are allowed (all active keys are maintained in a sorted vec).
- Trait handlers/signals to allow control over when sessions are rotated (for BABE @DemiMarie-parity), when new keys are switched in (era) and when slots are disabled.
- Tuple-keys to allow for different keys to be stored and switch synchronously.
- Interface for the tuple keys to allow comparison for duplicate checking.
- Interface for checking proof-of-ownership over a key set.
- Some reformatting of staking to stop line-length checks complaining.
- Removes the consensus module and any notion of a general "authority set" or authority ID type. All consensus modules are responsible to define or figure out their own authority types and set. 

TODO:
- [x] Update docs
- [x] Update tests
- [x] Integrate into Babe
- [x] Integrate into Grandpa
- [x] Integrate into Aura
- [x] Integrate into Staking
- [x] Integrate into FinalityTracker/Grandpa
- [x] Integrate into overall runtime
- [x] Kill Consensus module
  - [x] Move authorities storage/config from consensus to Aura
  - [x] Move code config from consensus to System
- [x] Tidy up consensus crates
  - [x] Babe is currently falsely generic over crypto: `BabeApi` has a generic param - remove it.
  - [x] Create an `AuthorityId` type alias for Babe's `Public` and put it in the Babe primitives like for Aura; use it throughout.
  - [x] Aura is currently half-generic (client-side is generic, runtime-side is not). Re-genericise the runtime.
- [x] Manage Aura offline reporting somehow.

Addresses https://github.com/paritytech/polkadot/issues/47
Supersedes #2713 #2768 #2628

PR #4582, opened 2020-01-09
Remove references to frame_consensus from docs

### LINKED ISSUE #4574: pallet_aura docs still refer to frame_consensus
The module documentation for `pallet_aura` ([rendered here](https://substrate.dev/rustdocs/master/pallet_aura/index.html)) links to `frame_consensus` which I believe has been removed for quite a while.

Perhaps the line should just be removed, or maybe the link should be pointed at `pallet_babe` instead.

JoshOrndorff:
Fixes #4574 by removing references to the no-longer-existant `frame_consensus`.

Should anything more be done here? The docs for Aura used to say:

> The Consensus module does not relate directly to Aura, but serves to manage offline reporting by implementing `ProvideInherent` in a similar way.

### DIFF
--- a/frame/session/src/lib.rs
+++ b/frame/session/src/lib.rs
@@ -113,7 +113,6 @@
 //!
 //! ## Related Modules
 //!
-//! - [Consensus](../frame_consensus/index.html)
 //! - [Staking](../pallet_staking/index.html)
 //! - [Timestamp](../pallet_timestamp/index.html)
 
--- a/frame/session/src/lib.rs
+++ b/frame/session/src/lib.rs
@@ -52,7 +52,7 @@
 //! - **Session rotation process:** At the beginning of each block, the `on_initialize` function
 //! queries the provided implementation of `ShouldEndSession`. If the session is to end the newly
 //! activated validator IDs and session keys are taken from storage and passed to the
-//! `SessionHandler`. The validator set supplied by `OnSessionEnding` and the corresponding session
+//! `SessionHandler`. The validator set supplied by `SessionManager::new_session` and the corresponding session
 //! keys, which may have been registered via `set_keys` during the previous session, are written
 //! to storage where they will wait one session before being passed to the `SessionHandler`
 //! themselves.
--- a/frame/session/src/lib.rs
+++ b/frame/session/src/lib.rs
@@ -86,12 +86,10 @@
 //!
 //! ```
 //! use pallet_session as session;
-//! # fn not_executed<T: session::Trait>() {
 //!
-//! fn validators() -> Vec<<T as frame_system::Trait>::AccountId> {
-//!	<pallet_session::Module>::validators()
-//!}
-//! # }
+//! fn validators<T: session::Trait>() -> Vec<<T as frame_system::Trait>::AccountId> {
+//!	<pallet_session::Module<T>>::validators()
+//! }
 //! # fn main(){}
 //! ```
 //!
--- a/frame/session/src/lib.rs
+++ b/frame/session/src/lib.rs
@@ -89,7 +89,7 @@
 //! # fn not_executed<T: session::Trait>() {
 //!
 //! fn validators() -> Vec<<T as frame_system::Trait>::AccountId> {
-//!	<pallet_session::Module<T>>::validators()
+//!	<pallet_session::Module>::validators()
 //!}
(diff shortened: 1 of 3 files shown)


frame/session/src/lib.rs
```diff
@@ -113,7 +113,6 @@
 //!
 //! ## Related Modules
 //!
-//! - [Consensus](../frame_consensus/index.html)
```
bkchr: There are more occurrences in this module. However, the whole description of the module in the docs is not correct anymore.

bkchr: (maybe not the whole, but crucial parts)

JoshOrndorff:
I've pushed another commit that attempts a more thorough re-write of the docs for the Session pallet. This writeup is based on my understanding from reading the code and the Seminar about [interfacing with Session](https://www.youtube.com/watch?v=ku0MutmiAlk)

While the big-picture ideas are correct, I'm not confident that every detail I've written is correct and would appreciate a thorough review and perhaps mentoring if I missed the target. It seems the docs went stale with https://github.com/paritytech/substrate/pull/2802

frame/session/src/lib.rs
```diff
@@ -30,92 +30,75 @@
...
 //! - **Session rotation configuration:** Configure as either a 'normal' (rewardable session where rewards are
 //! applied) or 'exceptional' (slashable) session rotation.
-//! - **Session rotation process:** The session is changed at the end of the final block of the current session
-//! using the `on_finalize` method. It may be called by either an origin or internally from another runtime
-//! module at the end of each block.
+//! - **Session rotation process:** At the beginning of each block, the `on_initialize` function
+//! queries the provided implementation of `ShouldEndSession`. If the session is to end the newly
+//! activated validator IDs and session keys are taken from storage and passed to the
+//! `SessionHandler`. The validator set supplied by `OnSessionEnding` and the corresponding session
+//! keys, which may have been registered via `set_keys` during the previous session, are written ////! to storage where they will wait one session before being passed to the `SessionHandler`
```
thiolliere:
```suggestion
//! keys, which may have been registered via `set_keys` during the previous session, are written
//! to storage where they will wait one session before being passed to the `SessionHandler`
```

frame/session/src/lib.rs
```diff
@@ -30,92 +30,75 @@
...
 //!
-//! - `set_key` - Set a validator's session key for the next session.
-//! - `set_length` - Set a new session length to be applied upon the next session change.
-//! - `force_new_session` - Force a new session that should be considered either a normal (rewardable)
-//! or exceptional rotation.
-//! - `on_finalize` - Called when a block is finalized. Will rotate session if it is the last
-//! block of the current session.
+//! - `set_keys` - Set a validator's session keys for upcoming sessions.
+//! - `on_initialize` - Called when a block is initialized. Will rotate session if it is the first
+//! block of the new session.
```
thiolliere: on_initialize is not really a dispatchable function is it ? but if the docs are consistent I'm ok.

JoshOrndorff: Yeah, I thought that too, and left it because `on_finalize` was in there before. I'll remove it though. I think it makes more sense being removed.

bkchr:
@JoshOrndorff tests are failing.

frame/session/src/lib.rs
```diff
@@ -30,92 +30,72 @@
...
+//! This pallet provides the `PeriodicSessions` struct for simple periodic sessions.
 //! - **Session rotation configuration:** Configure as either a 'normal' (rewardable session where rewards are
 //! applied) or 'exceptional' (slashable) session rotation.
-//! - **Session rotation process:** The session is changed at the end of the final block of the current session
-//! using the `on_finalize` method. It may be called by either an origin or internally from another runtime
-//! module at the end of each block.
+//! - **Session rotation process:** At the beginning of each block, the `on_initialize` function
+//! queries the provided implementation of `ShouldEndSession`. If the session is to end the newly
+//! activated validator IDs and session keys are taken from storage and passed to the
+//! `SessionHandler`. The validator set supplied by `OnSessionEnding` and the corresponding session
```
thiolliere:
```suggestion
//! `SessionHandler`. The validator set supplied by `SessionManager::new_session` and the corresponding session
```

PR #4583, opened 2020-01-09
Allow node to add externalities extensions

gnunicorn:
Unfortunately, to make this work with the least code impact necessary (and without refactoring `ServiceBuilder`), this takes a similar approach as we already use for the `register_transaction_pool` function to expose it to the service builder, which can then in turn offer it to the outside.

This change is necessary for external libraries (in this case Substrate-LFS) to provide further functionality through the externalities of a chain, so they can be used in e.g. Offchain-Workers.


client/service/src/builder.rs
```diff
@@ -740,6 +741,46 @@ ServiceBuilder<
 		+ TransactionPoolMaintainer<Block=TBl, Hash = <TBl as BlockT>::Hash>,
 	TRpc: sc_rpc::RpcExtension<sc_rpc::Metadata> + Clone,
 {
+
+	/// Set an ExecutionExtensionsMaker
+	pub fn with_execution_extensions_maker(self, execution_extensions_maker: Box<dyn ExtensionsMaker>) -> Result<Self, Error> {
```
gnunicorn: This is an ugly API.

A different way to provide this would be to go through `service.client` e.g.:

```rust
		builder.client().execution_extensions().set_extensions_maker(Box::new(()));
```
I am open to discussing that as the "official" way to do this in 2.0 for now and maybe mark `builder.client()` as unstable. Which would make clear that if we changed how this works internally, this might break your code!

client/api/src/execution_extensions.rs
```diff
@@ -62,6 +62,18 @@ impl Default for ExecutionStrategies {
 	}
 }
 
+/// Generate the starting set of ExternalitiesExtensions based upon the given capabilities
+pub trait ExtensionsMaker: Send + Sync {
```
tomusdrw: I don't like `Maker` suffix that much, I would either call it `ExtensionsFactory` or `ExtensionsProvider` or just traitify `Extensions` itself and just have `fn new` function in that trait.

client/api/src/execution_extensions.rs
```diff
@@ -90,14 +104,20 @@ impl<Block: traits::Block> ExecutionExtensions<Block> {
 		keystore: Option<BareCryptoStorePtr>,
 	) -> Self {
 		let transaction_pool = RwLock::new(None);
-		Self { strategies, keystore, transaction_pool }
+		let extensions_maker = Box::new(());
```
tomusdrw:
```suggestion
		let extensions_maker = RwLock::new(Box::new(()));
		Self { strategies, keystore, extensions_maker, transaciton_pool }
```
to avoid repetitions

client/service/src/builder.rs
tomusdrw: I'd prefer going through `client` to be honest, you can do that after the service is actually built, no? Like this is something you can change at any point if you have a reference to client anyway, so why try to restrict it to service builder anyway?

bkchr: I don't understand why this is an ugly api? This just follows the builder pattern? Or do I miss something?

client/service/src/builder.rs
```diff
@@ -740,6 +741,46 @@ ServiceBuilder<
...
+		let ServiceBuilder {
```
bkchr: Just remove this let

client/service/src/builder.rs
```diff
@@ -740,6 +741,46 @@ ServiceBuilder<
...
+			finality_proof_request_builder,
+			finality_proof_provider,
+			network_protocol,
+			transaction_pool,
+			rpc_extensions,
+			remote_backend,
+			marker
+		} = self;
+
+		client.execution_extensions().set_extensions_maker(execution_extensions_maker);
```
bkchr: And use

```suggestion
		self.client.execution_extensions().set_extensions_maker(execution_extensions_maker);
```

client/service/src/builder.rs
```diff
@@ -740,6 +741,46 @@ ServiceBuilder<
...
+
+		Ok(ServiceBuilder {
```
bkchr: And return here `Ok(self)`.

client/api/src/execution_extensions.rs
bkchr: Sounds german :D I would also go with `Factory` or `Provider`.

client/service/src/builder.rs
tomaka: > I'd prefer going through client to be honest, you can do that after the service is actually built, no? Like this is something you can change at any point if you have a reference to client anyway, so why try to restrict it to service builder anyway?

Being able to do this after the service is built seems like a huge hack to me that we should explicitly forbid for the sake of it.

gnunicorn: Agree with @tomaka here. The _only_ reason the set_extensions_maker is wrapped in a `RwLock` is because the order in which the setup happens (first create the client, later call build – not the other way round) prevents us from doing it in a clean way–same for  the transaction-pool extensions. Changing _either_ after construction / during the run (which this makes possible) makes no sense and will have unexepected and unwanted side-effects. Just think of having different Externalities or Transaction Pools between two Exection-Callls ... really not expected behavior.

It's an API that I dearly would  like to see removed, thus I'd prefer to not leak it too wide or recommend its usage. Hence the wrapper we have here...

@bkchr it's ugly because internally and externally it _doesn't_ match the pattern around: it isn't accepting a `FnOnce` (nor `Fn`) like the others do; it isn't executed immediately and the result stored and it isn't even kept on the builder but directly set on client...

client/api/src/execution_extensions.rs
gnunicorn: Had that, Rust complained and couldn't understand the type given was okay. Will try again with latest rust stable.

client/api/src/execution_extensions.rs
gnunicorn: `Factory` and `Provider` both sound like more than it does, but sure I don't care what it is named.

client/service/src/builder.rs
tomaka: > It's an API that I dearly would like to see removed, thus I'd prefer to not leak it too wide or recommend its usage. Hence the wrapper we have here...

What about adding a comment about that in the code?
From my experience, if we don't explicitly mention to not do something, sooner or later someone will do a small PR against that design without understanding the implications.

gnunicorn: @tomaka I created a ticket and added a FIXME in the code.

bkchr:
Please merge master.

PR #4585, opened 2020-01-09
Introduce `OnReapAccount`

shawntabrizi:
When writing a FRAME pallet, I found myself wanting access to an `OnReapAccount` trait which signals to my module when an account is _fully_ reaped from the system.

I could imagine that such a trait could be regularly used by other pallets (maybe even our own) to ensure that when an account is reaped, all possible storage items for that account are cleaned up.

The only main change here besides the introduction of this new trait is actually using it for cleaning up the `AccountNonce` in the system module. Before, we were removing `AccountNonce` within the balances pallet, but now we will actually clean it up properly using this trait. This may beniefit other pallet creators who want to create alternatives to our balances module.

Open to feedback about why this is a good or bad idea.

### DIFF
--- a/bin/node-template/runtime/src/lib.rs
+++ b/bin/node-template/runtime/src/lib.rs
@@ -205,6 +205,8 @@ impl balances::Trait for Runtime {
 	type Balance = Balance;
 	/// What to do if an account's free balance gets zeroed.
 	type OnFreeBalanceZero = ();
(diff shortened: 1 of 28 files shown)


shawntabrizi:
Companion PR for Polkadot: https://github.com/paritytech/polkadot/pull/748

PR #4590, opened 2020-01-10
pallet-contracts: Use the transactional storage for contracts

pepyakin:
This builds upon #3263.


gnunicorn:
Closing as part of a general PR clean up.

---

This PR hasn't seen activity in quite a while (>30 days), thus I am closing this for now. This is not a comment on the approach, code quality or whether we like it or not and it would get merged eventually, but it won't get merged as is and seems to be stale. You are welcome to reopen or raise another PR about this any time later.

PR #4592, opened 2020-01-10
Fix runners-up candidacy submission check in council

kianenigma:
same issue as: https://github.com/paritytech/substrate/pull/4429/files

- Tests already existed; yet they also worked with the wrong binary search since the container to search was just one element. I reviewed all leftover of the `binary_seach_xxx` in that file and they only apply to members and candidates who _are_ actually sorted by account id; This should close this whole issue. 
- With or without this update, in the next Kusama term change the candidate list should be cleaned. Luckily the phragmen implementation also ignores (kinda by luck; needs issue to be cleaned) dupe candidates so we should be fine.


shawntabrizi:
This has a broken test @kianenigma @gavofyork

kianenigma:
should be fixed with https://github.com/paritytech/substrate/pull/4594

PR #4603, opened 2020-01-11
Refactor some tests in tx-pool

NikVolf:


client/transaction-pool/Cargo.toml
```diff
@@ -21,3 +21,4 @@ sp-blockchain = { version = "2.0.0", path = "../../primitives/blockchain" }
 [dev-dependencies]
 sp-keyring = { version = "2.0.0", path = "../../primitives/keyring" }
 substrate-test-runtime-client = { version = "2.0.0", path = "../../test-utils/runtime/client" }
+sc-client = { version = "2.0.0", path = "../../client"}
```
tomusdrw: I think whatever we need from `sc-client` (I guess `LongestChain`) should be just re-exported from `test-runtime-client`.

PR #4604, opened 2020-01-11
Additional zero bid logic for Society pallet

shawntabrizi:
This introduces some additional logic around users trying to join society with zero value bids.

1. We will only select one zero value bid to be a candidate per rotation period. The rest of the selected candidates will comprise of users whose bids have some non-zero value.
2. If the accepted candidate pool contains a user with bid value of zero, that user automatically becomes the Head no matter the number of approvals they recieved. Otherwise, a random candidate is selected as head, weighted by approvals (as it was before).

### DIFF
--- a/frame/society/src/lib.rs
+++ b/frame/society/src/lib.rs
@@ -1306,7 +1306,7 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
 				// select one as primary, randomly chosen from the accepted, weighted by approvals. 
 				// Choose a random number between 0 and `total_approvals`
 				let primary_point = pick_usize(&mut rng, total_approvals - 1);
-				// Find the user who falls on that point
+				// Find the zero bid or the user who falls on that point
 				let primary = accepted.iter().find(|e| e.2.is_zero() || e.1 > primary_point)
 					.expect("e.1 of final item == total_approvals; \
 						worst case find will always return that item; qed")
--- a/frame/society/src/lib.rs
+++ b/frame/society/src/lib.rs
@@ -1108,7 +1108,14 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
 		const MAX_BID_COUNT: usize = 1000;
 
 		match bids.binary_search_by(|bid| bid.value.cmp(&value)) {
-			Ok(pos) | Err(pos) => bids.insert(pos, Bid {
+			// Insert new elements after the existing ones. This ensures new bids
+			// with the same bid value are further down the list than existing ones.
+			Ok(pos) => bids.insert(pos + 1, Bid {
+				value,
+				who: who.clone(),
+				kind: bid_kind,
+			}),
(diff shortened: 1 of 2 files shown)


frame/society/src/tests.rs
```diff
@@ -742,3 +742,45 @@ fn max_limits_work() {
 		assert_eq!(Society::candidates().len(), 10);
 	});
 }
+
+#[test]
+fn zero_bid_works() {
+	// This tests:
+	// * Only one zero bid is selected.
+	// * That zero bid is places as head when accepted.
```
shawntabrizi:
```suggestion
	// * That zero bid is placed as head when accepted.
```

frame/society/src/lib.rs
```diff
@@ -1108,7 +1108,14 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
 		const MAX_BID_COUNT: usize = 1000;
 
 		match bids.binary_search_by(|bid| bid.value.cmp(&value)) {
-			Ok(pos) | Err(pos) => bids.insert(pos, Bid {
+			// Insert new elements after the existing ones. This ensures new bids
+			// with the same bid value are further down the list than existing ones.
```
gavofyork: i don't think it does - it only ensures that it's below at least *one* of the existing ones. we probably need to skip forward from the position until it gets to the end or the value changes, and then insert.

frame/society/src/lib.rs
```diff
@@ -1299,8 +1306,8 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
 				// select one as primary, randomly chosen from the accepted, weighted by approvals. 
 				// Choose a random number between 0 and `total_approvals`
 				let primary_point = pick_usize(&mut rng, total_approvals - 1);
-				// Find the user who falls on that point
-				let primary = accepted.iter().find(|e| e.1 > primary_point)
+				// Find the zero bid or the user who falls on that point
+				let primary = accepted.iter().find(|e| e.2.is_zero() || e.1 > primary_point)
```
gavofyork: will the zero bid always be the first in the list? if not then the primary point will be chosen if it comes before the zero bidder.

frame/society/src/lib.rs
shawntabrizi: I updated this, but not sure if it is done ergonomically.

frame/society/src/lib.rs
shawntabrizi: Yes. We must assume `bids` is sorted by value. If so, then `candidates` will be sorted by value. If so, then `accepted` will be sorted by value.

If you would like me to write this assumption down near this code I can do that.

shawntabrizi:
I will work on another implementation of this logic introducing a new storage item `ZeroBids`, but if you like this, then feel free to merge it in and I can scrap the other work.

gavofyork:
i think it would still be worth cleaning it up if you've already started, but can merge this as is since it's strictly better.

PR #4608, opened 2020-01-13
Disable equalize, avoiding consensus issue

gavofyork:


frame/staking/Cargo.toml
```diff
@@ -29,7 +29,7 @@ substrate-test-utils = { version = "2.0.0", path = "../../test-utils" }
 [features]
 equalize = []
```
bkchr:
```suggestion
```

kianenigma:
I am not sure if this will solve the stalling though; Native benchmarks on the chain state shows it is only responsible for a very small portion of the time taken.

PR #4612, opened 2020-01-13
Switch GrandPa to std futures (replaces #3909)

expenses:
This is an updated branch of https://github.com/paritytech/substrate/pull/3909.


client/finality-grandpa/src/tests.rs
```diff
@@ -1287,105 +1286,101 @@ fn voter_persists_its_votes() {
...
+					// we restart alice's voter
+					voter_tx.unbounded_send(()).unwrap();
+
+					// and we push our own prevote for block 30
+					let prevote = finality_grandpa::Prevote {
+						target_number: 30,
+						target_hash: block_30_hash,
+					};
+
+					Pin::new(&mut *round_tx.lock()).start_send(finality_grandpa::Message::Prevote(prevote)).unwrap();
```
expenses: This leads to a compile error: 
```
    --> client/finality-grandpa/src/tests.rs:1340:84
     |
1340 |                     Pin::new(&mut *round_tx.lock()).start_send(finality_grandpa::Message::Prevote(prevote)).unwrap();
     |                                                                                                   ^^^^^^^ expected associated type, found struct `primitive_types::H256`
     |
     = note: expected type `finality_grandpa::Prevote<<sp_runtime::generic::block::Block<sp_runtime::generic::header::Header<u64, sp_runtime::traits::BlakeTwo256>, substrate_test_runtime::Extrinsic> as sp_runtime::traits::Block>::Hash, <<sp_runtime::generic::block::Block<sp_runtime::generic::header::Header<u64, sp_runtime::traits::BlakeTwo256>, substrate_test_runtime::Extrinsic> as sp_runtime::traits::Block>::Header as sp_runtime::traits::Header>::Number>`
                found type `finality_grandpa::Prevote<primitive_types::H256, {integer}>`
     = note: consider constraining the associated type `<sp_runtime::generic::block::Block<sp_runtime::generic::header::Header<u64, sp_runtime::traits::BlakeTwo256>, substrate_test_runtime::Extrinsic> as sp_runtime::traits::Block>::Hash` to `primitive_types::H256` or calling a method that returns `<sp_runtime::generic::block::Block<sp_runtime::generic::header::Header<u64, sp_runtime::traits::BlakeTwo256>, substrate_test_runtime::Extrinsic> as sp_runtime::traits::Block>::Hash`
     = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html

```

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -137,7 +136,11 @@ pub(crate) struct NetworkBridge<B: BlockT, N: Network<B>> {
 	neighbor_sender: periodic::NeighborPacketSender<B>,
 }
 
-impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
+impl<B: BlockT, N: Network<B>> NetworkBridge<B, N>
+	where
+		B::Hash: Unpin,
```
tomaka: I don't know why you need that, but it's most certainly wrong.
At no point do we pin the `Hash` or block number

tomaka: There's probably some struct that implements `Future` somewhere whose `Unpin` trait implementation is automatically derived from its parameter while it doesn't need to and could be derived unconditonally.

expenses: That's probably a result of me using `Pin::into_inner(self)` when I don't need to. I'll try and sort that out.

expenses:
@rphmeier The changes needed for https://github.com/paritytech/finality-grandpa can be found here: https://github.com/expenses/finality-grandpa/tree/future

client/finality-grandpa/src/communication/mod.rs
expenses: Ok, resolved

rphmeier:
@expenses Do you have a PR ready to go? Differences from https://github.com/paritytech/finality-grandpa/pull/100 ?

expenses:
@rphmeier See https://github.com/paritytech/substrate/pull/4612.

client/finality-grandpa/src/tests.rs
arkpar: Try switching `round_communication` to return concrete type instead of anonymous `impl Sink`

rphmeier: Yeah, it's probably unable to project through the anonymous return type (intended behavior - `impl Trait` should not be leaky)

rphmeier: I confirmed that this works locally, but the behavior is definitely strange. It seems like `impl Trait<T>` can't unify projected versions of T with T.

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -321,14 +317,14 @@ impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
 			has_voted,
 		};
 
-		let out_rx = out_rx.map_err(move |()| Error::Network(
+		let out_rx = out_rx.map(Result::Ok).map_err(move |()| Error::Network(
```
tomaka: I think we can remove the `map` and `map_err` entirely.

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -300,16 +297,15 @@ impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
...
 					_ => {
 						debug!(target: "afg", "Skipping unknown message type");
-						return Ok(None);
+						future::ready(None)
 					}
 				}
 			})
-			.filter_map(|x| x)
-			.map_err(|()| Error::Network(format!("Failed to receive message on unbounded stream")));
+			.map(Ok);
```
tomaka: And remove this as well.

client/finality-grandpa/src/until_imported.rs
```diff
@@ -71,24 +73,30 @@ pub(crate) trait BlockUntilImported<Block: BlockT>: Sized {
...
+impl<Block: BlockT, BlockStatus, BlockSyncRequester, I, M: BlockUntilImported<Block>> Unpin for
+UntilImported<Block, BlockStatus, BlockSyncRequester, I, M> {
+}
+
 impl<Block, BlockStatus, BlockSyncRequester, I, M> UntilImported<Block, BlockStatus, BlockSyncRequester, I, M> where
 	Block: BlockT,
-	BlockStatus: BlockStatusT<Block>,
-	M: BlockUntilImported<Block>,
-	I: Stream,
+	BlockStatus: BlockStatusT<Block> + Unpin,
```
tomaka: Ideally we would use `pin-project` instead to not require `Unpin` here

expenses:
We should _probably_ field test this PR pretty heavily.

expenses:
@andresilva @rphmeier suggested I ping you for ideas on how to test this.

andresilva:
@expenses https://gist.github.com/andresilva/264e00476f125d3bd6716fb58478d543 have a look at these instructions for setting up a local testnet, feel free to pm me if you have any issues. Ideally, we should also build and deploy polkadot with these changes on one of our validators to do some field testing, coordinating with @ddorgan (we might have to wait a bit for this due to the issues with latest libp2p version).

expenses:
@andresilva I don't have the hardware to run the dockerised local testnet right now, but perhaps we could fieldtest with @ddorgan?

andresilva:
I guess we could deploy some validators on flaming fir (or spin up some temporary testnet) with these changes? cc @ddorgan

expenses:
@ddorgan updated the flaming fir nodes, and things seem to be going alright. I'm not the best at reading logs though. @rphmeier maybe you could just check that things are going smoothly if you're not too busy?

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -48,8 +42,8 @@ use sp_runtime::traits::{Block as BlockT, Hash as HashT, Header as HeaderT, Numb
 use sc_telemetry::{telemetry, CONSENSUS_DEBUG, CONSENSUS_INFO};
 
 use crate::{
-	CatchUp, Commit, CommunicationIn, CommunicationOut, CompactCommit, Error,
-	Message, SignedMessage,
+	CatchUp, Commit, CommunicationIn, CommunicationOutH,
```
mxinden: What does the `H` in `CommunicationOutH` stand for?

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -690,7 +677,7 @@ pub(crate) fn check_message_sig_with_buffer<Block: BlockT>(
 /// use the same raw message and key to sign. This is currently true for
 /// `ed25519` and `BLS` signatures (which we might use in the future), care must
 /// be taken when switching to different key types.
-struct OutgoingMessages<Block: BlockT> {
+pub struct OutgoingMessages<Block: BlockT> {
```
mxinden:
```suggestion
pub(crate) struct OutgoingMessages<Block: BlockT> {
```
That would be enough, right?

client/finality-grandpa/src/communication/tests.rs
```diff
@@ -44,11 +43,19 @@ struct TestNetwork {
 	sender: mpsc::UnboundedSender<Event>,
 }
 
-impl sc_network_gossip::Network<Block> for TestNetwork {
-	fn event_stream(&self) -> Box<dyn futures::Stream<Item = NetworkEvent, Error = ()> + Send> {
+impl TestNetwork {
+	fn event_stream_03(&self) -> Pin<Box<dyn futures::Stream<Item = NetworkEvent> + Send>> {
```
mxinden: Why this indirection through `event_stream_03`? Why not do this within `sc_network_gossip::Network::event_stream`?

client/finality-grandpa/src/lib.rs
```diff
@@ -608,13 +609,12 @@ pub fn run_grandpa_voter<B, E, Block: BlockT, N, RA, SC, VR, X, Sp>(
 							.expect("authorities is always at least an empty vector; elements are always of type string")
 					}
 				);
-				ready(())
+				future::ready(())
 			})
-			.unit_error()
-			.compat();
-		futures::future::Either::A(events)
+			.then(|_| { future::ready(()) });
```
mxinden: Small nit: Why not a `.map(|_| () }`?

mxinden: Or even `.map(drop)` like you do further below.

client/finality-grandpa/src/tests.rs
```diff
@@ -1284,105 +1283,99 @@ fn voter_persists_its_votes() {
...
```
mxinden: One should first call `poll_ready` before `start_send` on ` `Sink`. As a convenience function, given that we are within an `async` block, one can call `SinkExt::send`.

(I don't think this was introduced through this pull request, but the pull request enables the easy fix via `SinkExt::send`. Let me know if you want me to do this as a follow up.)

client/finality-grandpa/src/until_imported.rs
```diff
@@ -157,37 +164,36 @@ impl<Block, BStatus, BSyncRequester, I, M> Stream for UntilImported<Block, BStat
...
 				 	}
 				}
-				Ok(Async::NotReady) => break,
+				Poll::Pending => break,
 			}
 		}
 
 		let mut update_interval = false;
-		while let Async::Ready(Some(_)) = self.check_pending.poll().map_err(Error::Timer)? {
+		while let Poll::Ready(Some(Ok(()))) = Stream::poll_next(Pin::new(&mut this.check_pending), cx) {
```
mxinden: Wouldn't the following also work, given that the `Stream` is boxed and thus `Unpin`?

```suggestion
		while let Poll::Ready(Some(Ok(()))) = this.check_pending.poll_next_unpin(cx) {
```

client/finality-grandpa/src/communication/tests.rs
expenses: Yeah, that's a little wierd, but It's gonna get changed via https://github.com/paritytech/substrate/pull/4633 which removed pretty much all the remaining futures01 stuff.

client/finality-grandpa/src/communication/mod.rs
expenses: It's like `CommunicationOut` but "useful for forcing the hash to some type (e.g. `H256`) when the compiler can't do the inference.".

client/finality-grandpa/src/tests.rs
expenses: We actually can't use `SinkExt::send`, because that's that'd mean awaiting a function that borrows a mutex and ends up with: `*mut () cannot be sent between threads safely`.

rphmeier:
>  @rphmeier maybe you could just check that things are going smoothly if you're not too busy?

Seems to be finalizing 1 day later!

andresilva:
Flaming fir was having some issues when running nodes without this PR anyway, so I think it's probably unrelated (it's still running the libp2p version that we had issues with). I'm OK with not blocking this PR further, when we start using this in polkadot we can do some prior testing on our nodes as well. I'll give this a final review in a bit.

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -372,8 +365,8 @@ impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
 		voters: Arc<VoterSet<AuthorityId>>,
 		is_voter: bool,
 	) -> (
-		impl Stream<Item = CommunicationIn<B>, Error = Error>,
-		impl Sink<SinkItem = CommunicationOut<B>, SinkError = Error>,
+		impl Stream<Item = CommunicationIn<B>,>,
```
andresilva:
```suggestion
		impl Stream<Item = CommunicationIn<B>>,
```

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -401,12 +394,14 @@ impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
...
 			let voter::CommunicationOut::Commit(round, commit) = out;
-			Ok((round, commit))
+			future::ok((round, commit))
 		});
 
 		(incoming, outgoing)
 	}
+}
 
+impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
```
andresilva: Doesn't this have the same bounds as the impl above? Why create a new one just for this function.

client/finality-grandpa/src/until_imported.rs
```diff
@@ -70,13 +72,17 @@ pub(crate) trait BlockUntilImported<Block: BlockT>: Sized {
 }
 
 /// Buffering imported messages until blocks with given hashes are imported.
+#[pin_project::pin_project]
 pub(crate) struct UntilImported<Block: BlockT, BlockStatus, BlockSyncRequester, I, M: BlockUntilImported<Block>> {
-	import_notifications: Fuse<Box<dyn Stream<Item = BlockImportNotification<Block>, Error = ()> + Send>>,
+	#[pin]
```
andresilva: 🤯

tomaka: It's unnecessary here though, because `UnboundedReceiver` is `Unpin` :sweat_smile:

client/finality-grandpa/src/until_imported.rs
```diff
@@ -70,13 +72,17 @@ pub(crate) trait BlockUntilImported<Block: BlockT>: Sized {
...
+	import_notifications: Fuse<UnboundedReceiver<BlockImportNotification<Block>>>,
 	block_sync_requester: BlockSyncRequester,
 	status_check: BlockStatus,
+	#[pin]
 	inner: Fuse<I>,
 	ready: VecDeque<M::Blocked>,
-	check_pending: Box<dyn Stream<Item = (), Error = std::io::Error> + Send>,
+	#[pin]
```
tomaka: Here as well, not needed.

expenses:
Ok. If we're all happy with merging this PR, I'll do that in a few hours.

Issue #4533, opened 2020-01-06
Improve substrate factory


marcio-diaz:
Extend `substrate factory` command to also:

- [ ] Execute using wasm mode.
- [ ] Create all kind of transactions.
- [ ] Have the ability to express dependencies between tx, etc.
- [ ] Report stats about execution time.
- [ ] Support grafana.

Part of #1408.

bkchr:
I would really to get rid off most of the implementation. The problem is that `factory` builds blocks by manually replicating the consensus etc. I think that with https://github.com/paritytech/substrate/pull/4143 we could have a much better implementation that hopefully does not require to have all these hacks in it.

PR #4614, opened 2020-01-13
Improving tx factory for benchmarking

marcio-diaz:
Starting to test the idea of #4533. 

Can be run, for example, with:
`substrate benchmark --dev --blocks 2 --tx-per-block 10 --tracing-targets pallet --execution Native`
or 
`substrate benchmark --dev --blocks 5 --tx-per-block 10 --bench-file ./factory_tests/democracy.txt --mode Automata`

Some TODOs:

- [ ] Clean output.
- [ ] Print stats.
- [ ] Add arguments to the automaton.
- [ ] Use manual seal?.
- [ ] Improve the access to context, i.e. block number, etc.


bin/node/cli/src/cli.rs
```diff
@@ -50,32 +52,11 @@ impl GetSharedParams for CustomSubcommands {
...
-	///               ... x `rounds`
-	///
-	/// These three modes control manufacturing.
-	#[structopt(long="mode", default_value = "MasterToN")]
-	pub mode: node_transaction_factory::Mode,
-
-	/// Number of transactions to generate. In mode `MasterNToNToM` this is
-	/// the number of transactions per round.
+	/// Transaction name.
+	#[structopt(long="tx-name", default_value = "transfer")]
```
kianenigma: is there a module selected as well? or they will all be chosen from balances? Can I call remark here?

marcio-diaz: That's old, going to be fixed today. 

In general, this is just a draft. I'm exploring what can be done and if it's useful.

My current idea is to have:
- Simple command like `substrate bench` that executes all transactions (using default values) of all modules, N number of times and make a report. 
- Same for modules and specific transactions, i.e. `bench staking` (all tx in staking), `bench staking bond` (just execute bond N times).
- I found that many times we want to loop and add dependencies between tx, for that I would use some kind of automata and pass the file. So it will be `bench staking.txt`, with edges within the file as: `0 1 staking_unbond, 1 0 staking_rebond`, (`src_state dest_state tx`).
- Something I didn't figure out is how to specify the parameters of each tx. I was thinking to have a name (string) for a specific instance (value) of a type. Some edge like `0 1 balances_transfer alice 10`, could fetch the values like `Address::test_value("alice"), Balances::test_value("10")`.

Also, I'm not sure about the whole automaton idea, may be a waste of time. But it's nice to not need to recompile everything when you benchmark do something wrong :).

kianenigma: @shawntabrizi  @mattrutherford  I think you guys might want to align together?

kianenigma:
> I found that many times we want to loop and add dependencies between tx, for that I would use some kind of automata and pass the file. So it will be bench staking.txt, with edges within the file as: 0 1 staking_unbond, 1 0 staking_rebond, (src_state dest_state tx).

This is a really cool idea 👍 

--- 

Generally speaking, 

- tx factory calls into the runtime directly or pushes stuff to queue? 
- is it native or wasm?

marcio-diaz:
> > I found that many times we want to loop and add dependencies between tx, for that I would use some kind of automata and pass the file. So it will be bench staking.txt, with edges within the file as: 0 1 staking_unbond, 1 0 staking_rebond, (src_state dest_state tx).
> 
> This is a really cool idea +1
> 
> Generally speaking,
> 
>     * tx factory calls into the runtime directly or pushes stuff to queue?
> 
>     * is it native or wasm?

It doesn't use the tx queue. Can be run native or wasm, but still need to add some prints for wasm close to the call place.

Here is an example of input file that now runs with all tx returning `Ok()`: https://github.com/paritytech/substrate/blob/931424a7655d54466063bcbba4ad2a00d2860d0a/factory_tests/democracy.txt

kianenigma:
What's the prospect of this PR?

marcio-diaz:
> What's the prospect of this PR?

Closing, given the talk in benchmarking room.

PR #4615, opened 2020-01-13
Add trivial EnsureFounder verifier to society

gavofyork:
Needed for sensible usage of the pallet.


frame/society/src/lib.rs
```diff
@@ -1081,6 +1081,18 @@ decl_event! {
 	}
 }
 
+/// Simple ensure origin struct to filter for the founder account.
+pub struct EnsureFounder<T>(sp_std::marker::PhantomData<T>);
+impl<T: Trait> EnsureOrigin<T::Origin> for EnsureFounder<T> {
+	type Success = T::AccountId;
+	fn try_origin(o: T::Origin) -> Result<Self::Success, T::Origin> {
+		o.into().and_then(|o| match (o, Members::<T>::get().get(0)) {
```
bkchr: Isn't the list sorted by `AccountId`? So the first doesn't need to be the founder?

PR #4618, opened 2020-01-14
Make offchain worker calls more future proof.

tomusdrw:
This will help debugging possible issues when we migrate to future versions.

1. Display an error if version is not supported.
2. Support only version 1 and 2. Will error out when we ever switch to `v3`.

### DIFF
--- a/client/offchain/src/lib.rs
+++ b/client/offchain/src/lib.rs
@@ -101,13 +101,14 @@ impl<Client, Storage, Block> OffchainWorkers<
 		let has_api_v1 = runtime.has_api_with::<dyn OffchainWorkerApi<Block, Error = ()>, _>(
 			&at, |v| v == 1
 		);
-		let has_api_v2 = runtime.has_api::<dyn OffchainWorkerApi<Block, Error = ()>>(&at);
+		let has_api_v2 = runtime.has_api_with::<dyn OffchainWorkerApi<Block, Error = ()>, _>(
+			&at, |v| v == 2
+		);
 		let version = match (has_api_v1, has_api_v2) {
 			(_, Ok(true)) => 2,
 			(Ok(true), _) => 1,
 			_ => 0,
 		};
-
 		debug!("Checking offchain workers at {:?}: version:{}", at, version);
 		if version > 0 {
 			let (api, runner) = api::AsyncApi::new(
--- a/client/offchain/src/lib.rs
+++ b/client/offchain/src/lib.rs
@@ -139,6 +139,8 @@ impl<Client, Storage, Block> OffchainWorkers<
 			});
 			futures::future::Either::Left(runner.process())
 		} else {
+			let help = "Consider turning off offchain workers if they are not part of your runtime.";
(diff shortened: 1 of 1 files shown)


bkchr:
Maybe we should add some variant of [`CanAuthorWith`](https://github.com/paritytech/substrate/blob/master/primitives/consensus/common/src/lib.rs#L208) that prevents authoring when the `offchain` api is not present? Maybe that would be detected earlier?^^

client/offchain/src/lib.rs
```diff
@@ -139,6 +140,8 @@ impl<Client, Storage, Block> OffchainWorkers<
 			});
 			futures::future::Either::Left(runner.process())
 		} else {
+			let help = "Consider turning off offchain workers if they are not part of your runtime.";
```
bkchr: Problem is that validators need the offchain api, otherwise they don't send the heartbeat transactions. So, turning it off is not the best idea?

gavofyork:
CI failing, too.

gavofyork:
i don't see how not authoring when off-chain workers are not present helps. off-chain workers are designed to not be required for authoring.

bkchr:
> i don't see how not authoring when off-chain workers are not present helps. off-chain workers are designed to not be required for authoring.

Yeah I know, but still we currently have some connection. At least to my understanding. Not all validators could produce a block, but could also not execute the offchain worker to send the heartbeat. If you are not selected and not send the tx you are slashed. Sounds not like such an optional feature. In the future when we got offchain phragmen, it could probably be even worse? Or would we switch to on-chain phragmen when nobody provided an offchain one?

tomusdrw:
> Not all validators could produce a block, but could also not execute the offchain worker to send the heartbeat. 

Wouldn't then disabling authoring if you can't run offchain workers be even worse? Currently if you can't run offchain workers you at least have some chance of producing blocks, with the change you are proposing it would stall completely.

I suppose a good idea for validators now would be to monitor the node logs for error messages like these and then take some immediate actions. Perhaps better to somehow standardize critical messages reporting? I'm thinking about even posting to some URL given in CLI, so that you can configure IFTTT for instance to send you an email (this wouldn't require any programming knowledge).

tomusdrw:
BTW. CI issue magically resolved, I guess the failure was just spurious.

bkchr:
> BTW. CI issue magically resolved, I guess the failure was just spurious.

Yeah, I restarted it a second time.

bkchr:
> > Not all validators could produce a block, but could also not execute the offchain worker to send the heartbeat.
> 
> Wouldn't then disabling authoring if you can't run offchain workers be even worse? Currently if you can't run offchain workers you at least have some chance of producing blocks, with the change you are proposing it would stall completely.
> 
> I suppose a good idea for validators now would be to monitor the node logs for error messages like these and then take some immediate actions. Perhaps better to somehow standardize critical messages reporting? I'm thinking about even posting to some URL given in CLI, so that you can configure IFTTT for instance to send you an email (this wouldn't require any programming knowledge).

Yeah, you are right with the authoring.

I also thought about reporting, maybe we could also send some standardized message to telemetry.

PR #4622, opened 2020-01-14
Add `pallet-society` and `pallet-recovery` to substrate node

bkchr:


shawntabrizi:
We should also remove nicks and add Identity?

Issue #4402, opened 2019-12-16
`ReapedAcccount` event is spammed when multiple transactions are placed in the same block


shawntabrizi:
Take a look at Kusama CC-3 Block here:
https://polkadot.js.org/apps/#/explorer/query/0x3cf167740da9aac3559a95247cb015d2324fd259e9a12da570a8b782bca04909

In this block, a user submits multiple transactions which transfers their full balance. All but the first transaction failed (good).

However, you can see that the `ReapedAccount` event was emitted multiple times, when it should only really be emitted once here.

We should improve this.

dud1337:
After discussion with @laboon, this issue appears to be related to a negative "Active Accounts" integer on https://polkascan.io/pre/kusama-cc3

![undead_active](https://user-images.githubusercontent.com/5631021/72347470-95115b80-36d8-11ea-8163-e85fa704d7ce.png)

shawntabrizi:
@dud1337 indeed, that along with the fact that Kusama had zero fees for a while, so any account with zero nonce and zero value could submit a transaction and that would also trigger the `ReapedAccount` event.

Not sure if there is a good solution to this for situations where there are no fees.

PR #4628, opened 2020-01-15
Return early when fees/balances/values are zero

shawntabrizi:
Fixes https://github.com/paritytech/substrate/issues/4402

This PR adds a simple check before relevant balances functions that the "action to be performed" would result in a change.

For example: the function `can_slash` would normally query the free balance of a user and do a comparison function. However, if `can_slash` has `value.is_zero()`, we can simply return true without doing a storage read.

Further, a function like `set_lock` used to write zero amount locks, which we now skip. Same with `add_vesting_schedule`.

Ultimately, this PR also catches situations where some external logic, like the fee system, would try to do some action into the balances module with value zero. This may trigger unexpected side effects like emitting a `ReapedAccount` event when nothing changed in the system.

Please do not treat this PR as insubstantial. Would really want a second pair of eyes to confirm I have not broken any assumptions.

### DIFF
--- a/frame/balances/src/lib.rs
+++ b/frame/balances/src/lib.rs
@@ -886,6 +886,7 @@ where
 	}
 
 	fn can_slash(who: &T::AccountId, value: Self::Balance) -> bool {
+		if value.is_zero() { return true }
 		Self::free_balance(who) >= value
 	}
 
@@ -902,6 +903,7 @@ where
 	}
 
 	fn burn(mut amount: Self::Balance) -> Self::PositiveImbalance {
+		if amount.is_zero() { return PositiveImbalance::zero() }
 		<TotalIssuance<T, I>>::mutate(|issued| {
 			*issued = issued.checked_sub(&amount).unwrap_or_else(|| {
 				amount = *issued;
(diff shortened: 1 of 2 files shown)


frame/balances/src/lib.rs
```diff
@@ -961,6 +965,7 @@ where
 		value: Self::Balance,
 		existence_requirement: ExistenceRequirement,
 	) -> DispatchResult {
+		if value.is_zero() { return Ok(()) }
```
shawntabrizi: One side effect of this PR is that a `transfer` of zero just returns early.

Before, a transfer of zero would additionally charge a fee of `T::CreationFee` or `T::TransferFee` on top of the existing fee system.

Now, if the transfer is zero, because no logic is run, these additional fees are skipped. A user would still pay the normal transaction fee calculated with the `transaction_payment` module.

This all makes sense to me, but introduces a new behavior.

shawntabrizi: It also means a transfer of 0 emits no `Balances` events.

frame/balances/src/lib.rs
```diff
@@ -1194,20 +1217,22 @@ where
...
-			if l.id == id {
-				new_lock.take()
-			} else if l.until > now {
-				Some(l)
-			} else {
-				None
-			}).collect::<Vec<_>>();
-		if let Some(lock) = new_lock {
-			locks.push(lock)
+		if !amount.is_zero(){
```
gavofyork:
```suggestion
		if !amount.is_zero() {
```

why not just 

```
if amount.is_zero() { return }
```

shawntabrizi: :)

PR #4629, opened 2020-01-15
Refactor tx-pool maintenance and other high-level api

NikVolf:
- all tx-pool <-> chain interactions are going strictly through `ChainApi` now
- async-await for revalidation
- `BasicPool` now can be both light and full, depending on api and revalidation strategy


NikVolf:
Probably @svyatonik may take a look to see how light stuff is updated there

svyatonik:
iirc the main idea behind separating `Maintainer` and the `Pool` was that for some chains it would be better to implement their own maintenance procedure, based e.g. on information coming from blocks digests => avoid runtime calls (which are too expensive, especially on light clients). That way they could use our default `BasicPool` with their own maintenance procedure. I haven't reviewed in details, but would it be possible with this PR?

NikVolf:
> iirc the main idea behind separating `Maintainer` and the `Pool` was that for some chains it would be better to implement their own maintenance procedure, based e.g. on information coming from blocks digests => avoid runtime calls (which are too expensive, especially on light clients). That way they could use our default `BasicPool` with their own maintenance procedure. I haven't reviewed in details, but would it be possible with this PR?

I think it should be possible, just making custom `ChainApi`, like `LightApi` already does

svyatonik:
But by making custom `ChainApi` you could change how transaction is validated, not when and why it is validated. Detailed example: a chain with account+nonce + each tx from account increases nonce + when nonce is increased, `DigestItem::TransactionSubmitted(from_account)` is emitted. In this case custom light maintainer could only fetch body from remote node, theno prune and maybe revalidate this account transactions only when this digest is emitted. Default light maintainer will: (1) always fetch body from remote node when pool is empty (2) call revalidation periodically. Both actions aren't required in this case.

NikVolf:
Ah yes, this can be covered easy, we just add CustomMaintaincePool, that will take custom maintainer as argument

client/transaction-pool/graph/src/pool.rs
```diff
@@ -84,6 +86,9 @@ pub trait ChainApi: Send + Sync {
 
 	/// Returns hash and encoding length of the extrinsic.
 	fn hash_and_length(&self, uxt: &ExtrinsicFor<Self>) -> (Self::Hash, usize);
+
+	/// Return body
```
```suggestion
	/// Returns a block body given the block id.
```

To match other docs in the trait.

client/transaction-pool/src/lib.rs
```diff
@@ -49,6 +49,23 @@ pub struct BasicPool<PoolApi, Block>
...
 	pool: Arc<sc_transaction_graph::Pool<PoolApi>>,
+	api: Arc<PoolApi>,
+	revalidation_strategy: Arc<Mutex<RevalidationStrategy<NumberFor<Block>>>>,
+}
+
+/// Type of revalidation.
+pub enum RevalidationType {
+	/// Light revalidation type.
+	///
+	/// During maintaince, transaction pool makes periodic revalidation
```
tomusdrw: I guess you mean that the entire pool is being revalidated, right?
Cause currently the docs for Light and Full are not super clear about differences, If I didn't know exactly how it behaved before the changes I would have hard time deciding which one to choose.

Maybe add some more details on this logic?

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+}
+
+impl<N: Clone + Copy + SimpleArithmetic> RevalidationStatus<N> {
+	/// Called when revalidation is completed.
+	pub fn clear(&mut self) {
+		*self = Self::NotScheduled;
+	}
+
+	/// Returns true if revalidation is required.
+	pub fn is_required(
```
tomusdrw: I think this function should be called `next_state()`, cause calling it `is_required` and having `bool` as a return values is really confusing with it taking `&mut self`, I would really be surprised learning that subsequent calls to `is_required` can change the state.

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+	Light(RevalidationStatus<N>)
+}
+
+impl<N: Clone + Copy + SimpleArithmetic> RevalidationStrategy<N> {
+	pub fn clear(&mut self) {
+		if let Self::Light(status) = self { status.clear() }
+	}
+
+	pub fn resubmit_required(&mut self) -> bool {
+		if let Self::Light(_) = self { return false } else { return true }
```
tomusdrw:
```suggestion
		if let Self::Light(_) = self { false } else { true }
```

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+}
+
+enum RevalidationStrategy<N> {
+	Always,
+	Light(RevalidationStatus<N>)
+}
+
+impl<N: Clone + Copy + SimpleArithmetic> RevalidationStrategy<N> {
+	pub fn clear(&mut self) {
+		if let Self::Light(status) = self { status.clear() }
```
tomusdrw:
```suggestion
		if let Self::Light(status) = self {
			status.clear()
		}
```

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+	}
+
+	pub fn resubmit_required(&mut self) -> bool {
+		if let Self::Light(_) = self { return false } else { return true }
+	}
+
+	pub fn is_required(
```
tomusdrw: `RevalidationStrategy::is_required` sounds super confusing especially with `resubmit_required` being another function on the same type. What is required in the first case? It's not the strategy, is it?

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+		block: N,
+		revalidate_time_period: Option<std::time::Duration>,
+		revalidate_block_period: Option<N>,
+	) -> bool {
+		if let Self::Light(status) = self {
+			status.is_required(block, revalidate_time_period, revalidate_block_period)
+		} else { true }
+	}
+
+	pub fn amount(&self) -> Option<usize> {
```
tomusdrw: `amount` of what? `Strategy` is the context, it's not really obvious what amount we mean.

client/transaction-pool/src/lib.rs
```diff
@@ -57,16 +74,44 @@ impl<PoolApi, Block> BasicPool<PoolApi, Block>
...
+		pool_api: PoolApi,
+		revalidation_type: RevalidationType,
+	) -> Self {
+		let api = Arc::new(pool_api);
+		let cloned_api = api.clone();
 		BasicPool {
-			pool: Arc::new(sc_transaction_graph::Pool::new(options, pool_api)),
+			api: cloned_api,
+			pool: Arc::new(sc_transaction_graph::Pool::new(options, api)),
+			revalidation_strategy: Arc::new(Mutex::new(
```
tomusdrw: Maybe better move the mutability internally to the type? Some checks can be easily done without acquiring the lock in such case.

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+				lock.is_required(
+					block_number,
+					Some(std::time::Duration::from_secs(60)),
+					Some(20.into()),
+				),
+				lock.resubmit_required()
+			)
+		};
+
+		let revalidation_status = self.revalidation_strategy.clone();
```
tomusdrw:
```suggestion
		let revalidation_strategy = self.revalidation_strategy.clone();
```

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+		let revalidation_amount = revalidation_status.lock().amount();
+		let retracted = retracted.to_vec();
+
+		async move {
+			let hashes = api.block_body(&id).await
+				.unwrap_or_else(|e| {
+					log::warn!("Prune known transactions: error request {:?}!", e);
```
tomusdrw:
```suggestion
					log::error!(target: "txpool", "Unable to get block body for pruning: {:?}!", e);
```
as I don't think it should ever happen.

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+					None
+				})
+				.unwrap_or(Vec::new())
```
tomusdrw:
```suggestion
				.unwrap_or_default()
```

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+				.into_iter()
+				.map(|tx| pool.hash_of(&tx))
+				.collect::<Vec<_>>();
+
+			if let Err(e) = pool.prune_known(&id, &hashes) {
+				log::warn!("Cannot prune known in the pool {:?}!", e);
```
tomusdrw:
```suggestion
				log::error!(target: "txpool", "Cannot prune known in the pool {:?}!", e);
```

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+			}
+
+			if is_resubmit_required {
+				let mut resubmit_transactions = Vec::new();
+
+				for retracted_hash in retracted.into_iter() {
+					let txes = api.block_body(&BlockId::hash(retracted_hash.clone())).await
+						.unwrap_or(None)
+						.unwrap_or(Vec::new());
```
tomusdrw:
```suggestion
						.unwrap_or_default();
```

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+			if let Err(e) = pool.prune_known(&id, &hashes) {
+				log::warn!("Cannot prune known in the pool {:?}!", e);
+			}
+
+			if is_resubmit_required {
+				let mut resubmit_transactions = Vec::new();
+
+				for retracted_hash in retracted.into_iter() {
+					let txes = api.block_body(&BlockId::hash(retracted_hash.clone())).await
+						.unwrap_or(None)
```
tomusdrw: I think we should print the error as well.

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+				.collect::<Vec<_>>();
+
+			if let Err(e) = pool.prune_known(&id, &hashes) {
+				log::warn!("Cannot prune known in the pool {:?}!", e);
+			}
+
+			if is_resubmit_required {
+				let mut resubmit_transactions = Vec::new();
+
+				for retracted_hash in retracted.into_iter() {
```
tomusdrw:
```suggestion
				for retracted_hash in retracted {
```

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+					Some(std::time::Duration::from_secs(60)),
+					Some(20.into()),
+				),
+				lock.resubmit_required()
+			)
+		};
+
+		let revalidation_status = self.revalidation_strategy.clone();
+		let revalidation_amount = revalidation_status.lock().amount();
+		let retracted = retracted.to_vec();
```
tomusdrw: I suppose that avoiding allocating the vector here would require us to fight with lifetimes in the return type, right? I guess it's not worth to try to avoid it then.

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+
+			if let Err(e) = pool.prune_known(&id, &hashes) {
+				log::warn!("Cannot prune known in the pool {:?}!", e);
+			}
+
+			if is_resubmit_required {
+				let mut resubmit_transactions = Vec::new();
+
+				for retracted_hash in retracted.into_iter() {
+					let txes = api.block_body(&BlockId::hash(retracted_hash.clone())).await
```
tomusdrw:
```suggestion
					let block_transactions = api.block_body(&BlockId::hash(retracted_hash)).await
```

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+						.unwrap_or(None)
+						.unwrap_or(Vec::new());
+					for tx in txes {
```
tomusdrw:
```suggestion
					resubmit_transactions.extend(block_transactions)
```

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+
+	pub fn is_required(
+		&mut self,
+		block: N,
+		revalidate_time_period: Option<std::time::Duration>,
+		revalidate_block_period: Option<N>,
+	) -> bool {
+		if let Self::Light(status) = self {
+			status.is_required(block, revalidate_time_period, revalidate_block_period)
+		} else { true }
```
tomusdrw: If we introduce the background thread, revalidating front of the ready queue will not be required for `Full` mode.

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +175,158 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+				}
+			}
+
+			if is_revalidation_required {
+				if let Err(e) = pool.revalidate_ready(&id, revalidation_amount).await {
+					log::warn!("revalidate ready failed {:?}", e);
+				}
+			}
+
+			revalidation_status.lock().clear();
```
tomusdrw: That can get pretty racy, can't it? If we ever get two `maintain` runs in parallel they will mess up each other statuses.

client/transaction-pool/src/tests.rs
```diff
@@ -14,42 +14,70 @@
...
+};
+use std::collections::HashSet;
+use substrate_test_runtime_client::{
+	runtime::{AccountId, Block, BlockNumber, Extrinsic, Hash, Header, Index, Transfer},
+	AccountKeyring::{self, *},
 };
 
 struct TestApi {
-	pub modifier: Box<dyn Fn(&mut ValidTransaction) + Send + Sync>,
+	pub modifier: RwLock<Box<dyn Fn(&mut ValidTransaction) + Send + Sync>>,
```
tomusdrw: Could we integrate this structure under single `RwLock`? It's test code and I think it would benefit more from simplicity than performance, and better if no one deadlocks their tests by messing up locking order.

client/transaction-pool/src/tests.rs
```diff
@@ -220,3 +272,38 @@ fn should_correctly_prune_transactions_providing_more_than_one_tag() {
 	assert_eq!(pool.status().ready, 0);
 	assert_eq!(pool.status().future, 2);
 }
+
+#[test]
+fn maintaince_prune() {
```
tomusdrw:
```suggestion
fn should_prune_old_during_maintenance() {
```

client/transaction-pool/src/tests.rs
```diff
@@ -220,3 +272,38 @@ fn should_correctly_prune_transactions_providing_more_than_one_tag() {
...
+
+	pool.api.push_block(1, vec![xt.clone()]);
+
+	block_on(pool.maintain(&BlockId::number(1), &[]));
+	assert_eq!(pool.status().ready, 0);
+}
+
+
+#[test]
+fn maintaince_revalidate() {
```
tomusdrw:
```suggestion
fn should_revalidate_during_maintenance() {
```

primitives/transaction-pool/src/pool.rs
```diff
@@ -225,6 +226,13 @@ pub trait TransactionPool: Send + Sync {
 	fn hash_of(&self, xt: &TransactionFor<Self>) -> TxHash<Self>;
 }
 
+/// Trait for transaction pool maintaince.
+pub trait MaintainedTransactionPool : TransactionPool {
```
tomusdrw: Do we still need this separate trait? Perhaps better to just merge with `TransactionPool`?

svyatonik:
> Ah yes, this can be covered easy, we just add CustomMaintaincePool, that will take custom maintainer as argument

Sorry - missed that reponse somehow. Could you please add support for this case? Let's not destroy pre-designed features.

client/transaction-pool/src/lib.rs
NikVolf: It was this way in original, but yeah

primitives/transaction-pool/src/pool.rs
NikVolf: @svyatonik thinks we need it below

svyatonik: I don't have any specific opinion on whether we need this or not :) I just want my feature back! :)

client/transaction-pool/src/lib.rs
NikVolf: Light strategy revalidation can't run in parallel, since it is single-threaded, but I don't think we can enforce it here

client/transaction-pool/src/lib.rs
NikVolf: It can, if light client fails the block body fetch, which can happen from time to time

primitives/transaction-pool/src/pool.rs
tomusdrw: Let's just keep `TransactionPoolMaintainer` trait then and just implement it for `BasePool` by default.
If other maintainance is needed, one can:
1. Create a wrapper struct.
2. Implement `TransactionPool` and delegate to the original one
3. Implement different `TransactionPoolMaintainer`.

I like the fact that implementing the maintainer directly on the pool significantly reduces the amount of code and allows the maintainer to use internal methods. Maybe I'm still to be convinced on possible use cases of customizing only the maintainer, but from my POV it's not going to be common and I'm willing to trade-off easy of replacing with less code to maintain on our side.

NikVolf: @tomusdrw yeah my thoughts exactly

I don't think we need to keep bunch of confusing code just in case somebody may need it

NikVolf: > I don't have any specific opinion on whether we need this or not :) I just want my feature back! :)

Feature is there I think.
Start using it is not much more difficult that it would have been before :)

client/transaction-pool/src/tests.rs
NikVolf: I was planning to refactor this TestApi completely in the next PR, can we do it then?

client/transaction-pool/src/lib.rs
NikVolf: I redid in such way that it is no longer possible :)
But it is almost instant lock, not much more than just reading atomic

client/transaction-pool/src/lib.rs
NikVolf: I'll add warning - again, fetching remote body can fail.

client/transaction-pool/src/lib.rs
NikVolf: Yeah, it has to be moved in the async block

But it is usually just few items (hashes from discarded fork during reorg)

primitives/transaction-pool/src/pool.rs
NikVolf: @svyatonik and btw, I will be doing `FullPull` that will contain custom maintaince in next PR and address your concerns there, ok?

client/transaction-pool/src/lib.rs
NikVolf: Added :)

primitives/transaction-pool/src/pool.rs
svyatonik: I don't think I'll have any concerns other than supporting the custom maintenance algorithm right now, neither I have idea of what do you mean by `FullPull` :) So as I understand, we're doing something like that:
(1) `struct MyPool { inner: txpool::Pool<PoolApi> }`
(2) `impl TransactionPool for MyPool { delegate-everything-to-inner }`
(3) `impl MaintainedTransactionPool for MyPool { custom-logic }`

Looks good to me. I'll take a look at light-related code now

client/transaction-pool/src/api.rs
```diff
@@ -197,4 +206,28 @@ impl<Client, F, Block> sc_transaction_graph::ChainApi for LightChainApi<Client,
...
+		let fetcher = self.fetcher.clone();
+		async move {
+			let transactions = fetcher.remote_body({
+				RemoteBodyRequest {
+					header,
+					retry_count: None,
+				}
+			}).await;
+
+			Ok(Some(transactions.unwrap_or(Vec::new())))
```
svyatonik: Let's add `log::warn!()` before calling `unwrap_or()`

client/transaction-pool/graph/src/pool.rs
```diff
@@ -68,6 +68,8 @@ pub trait ChainApi: Send + Sync {
 	type Error: From<error::Error> + error::IntoPoolError;
 	/// Validate transaction future.
 	type ValidationFuture: Future<Output=Result<TransactionValidity, Self::Error>> + Send + Unpin;
+	/// Body future (since block body might be remote)
+	type BodyFuture: Future<Output = Result<Option<Vec<<Self::Block as traits::Block>::Extrinsic>>, Self::Error>> + Unpin + Send + 'static;
```
svyatonik: What's the difference between `None` and `Some(Vec::new())` - probably just keep it as `Vec<...>`?

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +177,162 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+		-> Pin<Box<dyn Future<Output=()> + Send>>
+	{
+		let id = id.clone();
+		let pool = self.pool.clone();
+		let api = self.api.clone();
+
+		let block_number = match api.block_id_to_number(&id) {
+			Ok(Some(number)) => number,
+			_ => {
+				log::trace!(target: "txqueue", "Skipping chain event - no numbrer for that block {:?}", id);
```
svyatonik:
```suggestion
				log::trace!(target: "txqueue", "Skipping chain event - no number for that block {:?}", id);
```

primitives/transaction-pool/src/pool.rs
NikVolf: cool,
(1) can be just `sc_transaction_graph:Pool`
BasicPool is already just `sc_transaction_graph:Pool` + maintaince

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +177,162 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+		let next_action = self.revalidation_strategy.lock().next(
+			block_number,
+			Some(std::time::Duration::from_secs(60)),
+			Some(20.into()),
+		);
+		let revalidation_strategy = self.revalidation_strategy.clone();
+		let retracted = retracted.to_vec();
+
+		async move {
+			let hashes = api.block_body(&id).await
```
svyatonik: We should only call this when pool is not empty. Otherwise light node would download block body of every imported block.

client/transaction-pool/src/lib.rs
```diff
@@ -130,3 +177,162 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
+					let block_transactions = api.block_body(&BlockId::hash(retracted_hash.clone())).await
+						.unwrap_or_else(|e| {
+							log::warn!("Failed to fetch block body {:?}!", e);
+							None
+						})
+						.unwrap_or_default();
+
+					resubmit_transactions.extend(block_transactions);
+				}
+				if let Err(e) = pool.submit_at(&id, resubmit_transactions, true).await {
```
svyatonik: I believe [this condition](https://github.com/paritytech/substrate/blob/5cd952bbf6405a165d0399fd3e5004961ace6819/client/transaction-pool/src/maintainer.rs#L87) should be saved.

client/transaction-pool/graph/src/pool.rs
NikVolf: Better probably follow what client trait `BlockBody` returns, it is `Result<Option<Vec<Extrinsic>>>`, what do you think?

client/transaction-pool/src/lib.rs
NikVolf: I'll add an issue for that.
We might want to run light strategy out of light client context actually

NikVolf:
@svyatonik thanks for comments, addressed!

NikVolf:
@tomusdrw can you resolve those couple of questions left?

client/transaction-pool/src/tests.rs
tomusdrw: Yes, please maybe make an issue for that to not forget.

client/transaction-pool/src/lib.rs
```diff
@@ -49,6 +49,25 @@ pub struct BasicPool<PoolApi, Block>
...
+	///
+	/// During maintaince, transaction pool makes periodic revalidation
+	/// of all transactions depending on number of blocks or time passed.
+	/// Also this kind of revalidation does not resubmit transactions from
+	/// retracted blocks, since it is too expensive.
+	Light,
+
+	/// Full revalidation type.
+	///
+	/// During maintaince, transaction pool revalidates some fixed amount of
```
tomusdrw:
```suggestion
	/// During maintenance, transaction pool revalidates some fixed amount of
```

client/transaction-pool/src/lib.rs
```diff
@@ -49,6 +49,25 @@ pub struct BasicPool<PoolApi, Block>
...
 	pool: Arc<sc_transaction_graph::Pool<PoolApi>>,
+	api: Arc<PoolApi>,
+	revalidation_strategy: Arc<Mutex<RevalidationStrategy<NumberFor<Block>>>>,
+}
+
+/// Type of revalidation.
+pub enum RevalidationType {
+	/// Light revalidation type.
+	///
+	/// During maintaince, transaction pool makes periodic revalidation
```
tomusdrw:
```suggestion
	/// During maintenance, transaction pool makes periodic revalidation
```

client/transaction-pool/src/lib.rs
tomusdrw: Don't think it's addressed. I really find it ugly to keep it in one line.

NikVolf: ops sorry!

PR #4630, opened 2020-01-15
Make Proposer instantiation potentially async.

rphmeier:
Needed for an upcoming Polkadot refactor and is strictly more general. This API was always intended to be used in an async context, so it doesn't break any underlying assumptions.

### DIFF
--- a/bin/node/cli/src/service.rs
+++ b/bin/node/cli/src/service.rs
@@ -404,6 +404,7 @@ mod tests {
 	use sc_service::AbstractService;
 	use crate::service::{new_full, new_light};
 	use sp_runtime::traits::IdentifyAccount;
+	use futures::prelude::*;
 
 	type AccountPublic = <Signature as Verify>::Signer;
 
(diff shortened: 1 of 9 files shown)


rphmeier:
(Polkadot PR incoming)

rphmeier:
From running some tests locally, it seems that `test_sync` is freezing when setting up a full node.

bin/node/cli/src/service.rs
```diff
@@ -535,13 +536,16 @@ mod tests {
...
-				)).expect("Error making test block").block;
+				let proposing = proposer_factory.init(&parent_header)
+					.and_then(|mut proposer| proposer.propose(
+						inherent_data,
+						digest,
+						std::time::Duration::from_secs(1),
+						RecordProof::Yes,
+					));
+
+				let new_block = futures::executor::block_on(proposing)
```
tomaka: May I suggest something like this instead?

```suggestion
				let new_block = futures::executor::block_on(async move {
				    let proposer = proposer_factory.init(&parent_header).await;
				    proposer.propose(
						inherent_data,
						digest,
						std::time::Duration::from_secs(1),
						RecordProof::Yes,
					).await
				});
```

rphmeier: still need to get the `async/await` into my bloodstream :)

PR #4631, opened 2020-01-15
client/finality-grandpa: Reintegrate periodic neighbor packet worker

mxinden:
The `NeighborPacketWorker` within `client/finality-grandpa` does two
things:

1. It receives neighbor packets from components within
`client/finality-grandpa`, sends them down to the `GossipEngine` in
order for neighboring nodes to receive.

2. It periodically sends out the most recent neighbor packet to the
`GossipEngine`.

In order to send out packets it had a clone to a `GossipEgine` within
an atomic reference counter and a mutex. The `NeighborPacketWorker` was
then spawned onto its own asynchronous task.

Instead of running in its own task, this patch reintegrates the
`NeighborPacketWorker` into the main `client/finality-grandpa` task not
requiring the `NeighborPacketWorker` to own a clone of the
`GossipEngine`.

The greater picture

This is a tiny change within a greater refactoring. The overall goal is
to **simplify** how finality-grandpa interacts with the network and to
**reduce** the amount of **unbounded channels** within the logic.

Why no unbounded channels: Bounding channels is needed for backpressure
and proper scheduling. With unbounded channels there is no way of
telling the producer side to slow down for the consumer side to catch
up.  Rephrased, there is no way for the scheduler to know when to favour
the consumer task over the producer task on a crowded channel and the
other way round for an empty channel.

Reducing the amount of shared ownership simplifies the logic and enables
one to use async-await syntax-suggar, given that one does not need to
hold a lock across poll invocations. Using async-await enables one to
use bounded channels without complex logic.


client/finality-grandpa/src/communication/mod.rs
```diff
@@ -134,7 +139,18 @@ pub(crate) struct NetworkBridge<B: BlockT, N: Network<B>> {
...
+	///
+	/// Packets send into this channel are processed by the `NeighborPacketWorker` and passed on to
+	/// the underlying `GossipEngine`.
 	neighbor_sender: periodic::NeighborPacketSender<B>,
+
+	/// `NeighborPacketWorker` processing packets send through the `NeighborPacketSender`.
+	//
+	// NetworkBridge is required to be clonable, thus one needs to be able to clone its children,
+	// thus one has to wrap neighor_packet_worker with an Arc Mutex.
+	neighbor_packet_worker: Arc<Mutex<periodic::NeighborPacketWorker<B>>>,
```
mxinden: Yes, in order to reduce the amount of `Mutex`es, I introduce a new one here. This one is within the crate only and can be removed once `NetworkBridge` does not need to be clonable anymore.

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -134,7 +139,18 @@ pub(crate) struct NetworkBridge<B: BlockT, N: Network<B>> {
 	service: N,
 	gossip_engine: GossipEngine<B>,
 	validator: Arc<GossipValidator<B>>,
+
+	/// Sender side of the neighbor packet channel.
+	///
+	/// Packets send into this channel are processed by the `NeighborPacketWorker` and passed on to
+	/// the underlying `GossipEngine`.
 	neighbor_sender: periodic::NeighborPacketSender<B>,
```
mxinden:
```suggestion
	neighbor_packet_sender: periodic::NeighborPacketSender<B>,
```

Happy to do this change, but rather as a follow up to reduce the noise within this pull request.

client/finality-grandpa/src/lib.rs
```diff
@@ -889,9 +900,9 @@ pub fn run_grandpa<B, E, Block: BlockT, N, RA, SC, VR, X, Sp>(
 	Block::Hash: Ord,
 	B: Backend<Block> + 'static,
 	E: CallExecutor<Block> + Send + Sync + 'static,
-	N: NetworkT<Block> + Send + Sync + Clone + 'static,
+	N: NetworkT<Block> + Send + Sync + Clone + Unpin + 'static,
 	SC: SelectChain<Block> + 'static,
-	NumberFor<Block>: BlockNumberOps,
+	NumberFor<Block>: BlockNumberOps + Unpin,
```
tomaka: This is wrong

client/finality-grandpa/src/communication/periodic.rs
```diff
@@ -56,56 +51,62 @@ impl<B: BlockT> NeighborPacketSender<B> {
...
-/// Does the work of sending neighbor packets, asynchronously.
-///
-/// It may rebroadcast the last neighbor packet periodically when no
-/// progress is made.
-pub(super) fn neighbor_packet_worker<B>(net: GossipEngine<B>) -> (
-	impl Future<Item = (), Error = ()> + Send + 'static,
-	NeighborPacketSender<B>,
-) where
-	B: BlockT,
+pub(super) struct NeighborPacketWorker<B: BlockT> {
```
rphmeier: could use a doc-string

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -391,6 +411,24 @@ impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
 	}
 }
 
+impl<B: BlockT, N: Network<B>> Future03 for NetworkBridge<B, N>
+where
+	NumberFor<B>: Unpin,
```
tomaka:
```suggestion
```
You probably have to add `impl<B, N> Unpin for NetworkBridge<B, N> {}` somewhere instead.

client/finality-grandpa/src/communication/periodic.rs
```diff
@@ -56,56 +51,62 @@ impl<B: BlockT> NeighborPacketSender<B> {
...
+			last: None,
+			delay,
+			rx,
+		}, NeighborPacketSender(tx))
+	}
+}
+
+impl <B: BlockT> Stream for NeighborPacketWorker<B>
+where
+	NumberFor<B>: Unpin,
```
tomaka:
```suggestion
```

client/finality-grandpa/src/lib.rs
```diff
@@ -551,10 +551,10 @@ pub fn run_grandpa_voter<B, E, Block: BlockT, N, RA, SC, VR, X, Sp>(
 	Block::Hash: Ord,
 	B: Backend<Block> + 'static,
 	E: CallExecutor<Block> + Send + Sync + 'static,
-	N: NetworkT<Block> + Send + Sync + Clone + 'static,
+	N: NetworkT<Block> + Send + Sync + Clone + Unpin + 'static,
 	SC: SelectChain<Block> + 'static,
 	VR: VotingRule<Block, Client<B, E, Block, RA>> + Clone + 'static,
-	NumberFor<Block>: BlockNumberOps,
+	NumberFor<Block>: BlockNumberOps + Unpin,
```
tomaka:
```suggestion
	NumberFor<Block>: BlockNumberOps,
```

client/finality-grandpa/src/lib.rs
```diff
@@ -551,10 +551,10 @@ pub fn run_grandpa_voter<B, E, Block: BlockT, N, RA, SC, VR, X, Sp>(
 	Block::Hash: Ord,
 	B: Backend<Block> + 'static,
 	E: CallExecutor<Block> + Send + Sync + 'static,
-	N: NetworkT<Block> + Send + Sync + Clone + 'static,
+	N: NetworkT<Block> + Send + Sync + Clone + Unpin + 'static,
```
tomaka: Not sure about this `Unpin`

client/finality-grandpa/src/lib.rs
```diff
@@ -650,13 +650,14 @@ struct VoterWork<B, E, Block: BlockT, N: NetworkT<Block>, RA, SC, VR> {
...
+	network: futures03::compat::Compat<NetworkBridge<Block, N>>,
 }
 
 impl<B, E, Block, N, RA, SC, VR> VoterWork<B, E, Block, N, RA, SC, VR>
 where
 	Block: BlockT,
- 	N: NetworkT<Block> + Sync,
-	NumberFor<Block>: BlockNumberOps,
+ 	N: NetworkT<Block> + Sync + Unpin,
+	NumberFor<Block>: BlockNumberOps + Unpin,
```
tomaka: Well, you get the idea

client/finality-grandpa/src/communication/mod.rs
mxinden: Thank you, thank you and thank you @tomaka :heart:

mxinden:
@tomaka can you take another look?

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -134,9 +139,22 @@ pub(crate) struct NetworkBridge<B: BlockT, N: Network<B>> {
 	service: N,
 	gossip_engine: GossipEngine<B>,
 	validator: Arc<GossipValidator<B>>,
+
+	/// Sender side of the neighbor packet channel.
+	///
+	/// Packets send into this channel are processed by the `NeighborPacketWorker` and passed on to
```
andresilva:
```suggestion
	/// Packets sent into this channel are processed by the `NeighborPacketWorker` and passed on to
```

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -134,9 +139,22 @@ pub(crate) struct NetworkBridge<B: BlockT, N: Network<B>> {
...
+	/// the underlying `GossipEngine`.
 	neighbor_sender: periodic::NeighborPacketSender<B>,
+
+	/// `NeighborPacketWorker` processing packets send through the `NeighborPacketSender`.
```
andresilva:
```suggestion
	/// `NeighborPacketWorker` processing packets sent through the `NeighborPacketSender`.
```

PR #4634, opened 2020-01-15
Use default in-memory keystore for commands that don't use keystore

### LINKED ISSUE #4623: (some) node subcommands are broken
...and they fail with `keystore` not being configured. Namely `import-blocks` and `factory` at least.

Also it would be nice to actually write some tests for the sub-commands so that we don't run into that issue in the future.

NikVolf:
Probably as a workaround until better refactoring

Fixes #4623

### DIFF
--- a/client/cli/src/lib.rs
+++ b/client/cli/src/lib.rs
@@ -401,7 +401,8 @@ impl<'a> ParseAndPrepareExport<'a> {
 		E: ChainSpecExtension,
 		Exit: IntoExit
 	{
-		let config = create_config_with_db_path(spec_factory, &self.params.shared_params, self.version)?;
+		let mut config = create_config_with_db_path(spec_factory, &self.params.shared_params, self.version)?;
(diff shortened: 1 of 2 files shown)


client/cli/src/lib.rs
```diff
@@ -749,6 +754,16 @@ fn input_keystore_password() -> Result<String, String> {
 		.map_err(|e| format!("{:?}", e))
 }
 
+/// Use in memory keystore config when it is not required at all.
+fn fill_config_keystore_in_memory<C, G, E>(config: &mut sc_service::Configuration<C, G, E>)
+	-> Result<(), String>
+{
+	match &mut config.keystore {
+		cfg @ KeystoreConfig::None => { *cfg = KeystoreConfig::InMemory; Ok(()) },
+		_ => Err(format!("Keystore config specified when it should not be!")),
```
```suggestion
		_ => Err("Keystore config specified when it should not be!".into()),
```

PR #4640, opened 2020-01-16
Rename basic-authority to basic-authorship

tomusdrw:
I believe by mistake we confused `authorship` with `authority` :)

May require a small fix in Polkadot as well.


.maintain/rename-crates-for-2.0.sh
```diff
@@ -43,6 +43,7 @@ TO_RENAME=(
     "sp-finality-granpda sp-finality-grandpa"
     "sp-sesssion sp-session"
     "sp-tracing-pool sp-transaction-pool"
+    "sc-basic-authority sc-basic-authorship"
```
gnunicorn: probably want to update line 92, too (otherwise updating an older work will be left an in a non-working state). Can't comment there :(

tomusdrw: Fixed, thanks for catching that.

PR #4643, opened 2020-01-16
Getting configuration from commands

cecton:
This PR adds a public function `.into_configuration()` on `ParseAndPrepare` that generates a `Configuration`.

The use case is for cumulus and other projects that would want to parse multiple command line arguments separately in order to start multiple nodes.

Related https://github.com/paritytech/cumulus/issues/34

Note: there is also a new extra argument `default_base_path` that allows the caller to define the `config_dir` before the `Configuration` gets created. This `default_base_path` can be overwritten by the argument `--base-path` (as proved in the test I added).


client/cli/src/lib.rs
```diff
@@ -1003,7 +1118,7 @@ where
 	S: FnOnce(&str) -> Result<Option<ChainSpec<G, E>>, String>,
 {
 	let spec = load_spec(cli, spec_factory)?;
-	let base_path = base_path(cli, version);
+	let base_path = base_path(cli, version, default_base_path);
```
tomusdrw: This could delegate to `create_build_spec_config` to avoid duplication

client/cli/src/lib.rs
```diff
@@ -295,6 +300,78 @@ impl<'a, CC, RP> ParseAndPrepare<'a, CC, RP> where CC: GetSharedParams {
 	}
 }
 
+impl<'a, CC, RP> ParseAndPrepare<'a, CC, RP> {
+	/// Convert ParseAndPrepare to Configuration
+	pub fn into_configuration<C, G, E, S>(
```
tomusdrw: Should existing `node/cli` and `node-template/cli` be using that function?

cecton:
> Looks ok, but I wonder if having a builder pattern here wouldn't make things simpler. Right now it seems we make it simpler by introducing yet another method (`into_configuration`). Do all other functions need to be public then?
> 
> All these `create_config_with_*` and `fill_*` functions are quite confusing to me. Without looking at the code I really have no idea which one should I use for my sub-command.

I don't think `create_config_with_*` and `fill_*` should be public anymore (@bkchr correct me if I'm wrong). There should be only 2 ways that should make sense and serve different purposes:
1. using the `parse_and_prepare()`
2. using the `into_configuration()`

The 1. is normally used for the 1st node as it also sets up the logger of the app. While the 2. can be used for extra nodes.

>  Should existing node/cli and node-template/cli be using that function?

I think no. Otherwise it would skip the logger setup (and the panic handler and some other things).

client/cli/src/lib.rs
cecton: (answered on https://github.com/paritytech/substrate/pull/4643#issuecomment-575127599 )

cecton:
@tomusdrw I'm not sure if it's used in polkadot actually. I just grepped the code and I couldn't find anything. Are you sure?

tomusdrw:
@cecton nope, just assumed it is. If it's not then even better :)

PR #4645, opened 2020-01-16
Async/await in transaction-graph

NikVolf:
So open mainly to discuss the decision

1. Combining async-await with trait implementation might force to write some lifetime bounds in both trait and trait impl, and probably also will need to propagate some bounds to the consumers.

2. But async code greatly improves thanks to async-await

So I wonder if we can avoid (1), which I hope, or is it tolerable, which I don't think so.

(Made pr against my other branch so that diff is not crazy)

### DIFF
--- a/client/rpc-api/src/author/error.rs
+++ b/client/rpc-api/src/author/error.rs
@@ -25,9 +25,6 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// Author RPC future Result type.
 pub type FutureResult<T> = Box<dyn rpc::futures::Future<Item = T, Error = Error> + Send>;
 
-/// Author RPC future Result type with specific lifetime.
-pub type FutureResultLt<'a, T> = Box<dyn rpc::futures::Future<Item = T, Error = Error> + Send + 'a>;
-
 /// Author RPC errors.
 #[derive(Debug, derive_more::Display, derive_more::From)]
 pub enum Error {
--- a/client/rpc-api/src/author/error.rs
+++ b/client/rpc-api/src/author/error.rs
@@ -25,6 +25,9 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// Author RPC future Result type.
 pub type FutureResult<T> = Box<dyn rpc::futures::Future<Item = T, Error = Error> + Send>;
 
+/// Author RPC future Result type with specific lifetime.
+pub type FutureResultLt<'a, T> = Box<dyn rpc::futures::Future<Item = T, Error = Error> + Send + 'a>;
+
 /// Author RPC errors.
 #[derive(Debug, derive_more::Display, derive_more::From)]
 pub enum Error {
--- a/client/transaction-pool/graph/src/pool.rs
+++ b/client/transaction-pool/graph/src/pool.rs
@@ -27,7 +27,6 @@ use serde::Serialize;
 use futures::{
 	Future, FutureExt,
 	channel::mpsc,
-	future::{Either, ready, join_all},
 };
 use sp_runtime::{
 	generic::BlockId,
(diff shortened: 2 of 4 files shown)


client/transaction-pool/graph/src/pool.rs
```diff
@@ -388,69 +369,75 @@ impl<B: ChainApi> Pool<B> {
...
-	fn verify_one(
+	async fn verify_one(
 		&self,
 		block_id: &BlockId<B::Block>,
 		block_number: NumberFor<B>,
 		xt: ExtrinsicFor<B>,
 		force: bool,
-	) -> impl Future<Output=(ExHash<B>, ValidatedTransactionFor<B>)> {
+	) -> (ExHash<B>, ValidatedTransactionFor<B>)
+	{
```
tomusdrw: Not a fan of curlies in solitude, can we merge it back with the previous line?

primitives/transaction-pool/src/pool.rs
```diff
@@ -167,58 +167,50 @@ pub trait TransactionPool: Send + Sync {
...
-		at: &BlockId<Self::Block>,
+	fn submit_and_watch<'a>(
+		&'a self,
+		at: &'a BlockId<Self::Block>,
 		xt: TransactionFor<Self>,
-	) -> Box<dyn Future<Output=Result<Box<TransactionStatusStreamFor<Self>>, Self::Error>> + Send + Unpin>;
-
-
-	// Block production / Networking
+	) -> std::pin::Pin<Box<dyn Future<Output=Result<Box<TransactionStatusStreamFor<Self>>, Self::Error>> + Send + 'a>>;
```
tomusdrw: Maybe just import `Pin`, cause it's used 3 times here already.

NikVolf: > Looks good to me, I think the lifetimes are acceptable.

I haven't also figured out how to patch RPC, probably some more lifetime bounds will be required there 🤷‍♀️

NikVolf:
@tomusdrw 

> Looks good to me, I think the lifetimes are acceptable.

I wonder how to make this possible:
```
	#[rpc(name = "author_submitExtrinsic")]
	fn submit_extrinsic<'a>(&'a self, extrinsic: Bytes) -> Box<dyn rpc::futures::Future<Item = T, Error = Error> + Send + 'a>;
```

client/transaction-pool/src/lib.rs
```diff
@@ -127,28 +129,40 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
 	fn submit_at(
 		&self,
 		at: &BlockId<Self::Block>,
-		xts: impl IntoIterator<Item=TransactionFor<Self>> + 'static,
-	) -> Box<dyn Future<Output=Result<Vec<Result<TxHash<Self>, Self::Error>>, Self::Error>> + Send + Unpin> {
-		Box::new(self.pool.submit_at(at, xts, false))
+		xts: Vec<TransactionFor<Self>>,
+	) -> PoolResult<Vec<Result<TxHash<Self>, Self::Error>>> {
+		let pool = self.pool.clone();
```
NikVolf: I removed lifetimes, but this seems to be unavoidable

Basically, due to how BasicPool is used elsewhere (rpc/etc), it has to carry shared pool reference with it

client/rpc-api/src/author/error.rs
```diff
@@ -25,6 +25,9 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// Author RPC future Result type.
 pub type FutureResult<T> = Box<dyn rpc::futures::Future<Item = T, Error = Error> + Send>;
 
+/// Author RPC future Result type with specific lifetime.
+pub type FutureResultLt<'a, T> = Box<dyn rpc::futures::Future<Item = T, Error = Error> + Send + 'a>;
```
tomusdrw: I think it would be better to do:
```
pub type FutureResult<'a, T> = Box<dyn Future<...> + Send + 'a>;

pub type StaticFutureResult<T> = FutureResult<'static, T>;
```
The `Lt` prefix does not seem intuitive to me.

NikVolf: This result with lifetime was not used already, so just deleted it

PR #4646, opened 2020-01-16
Expose proof generation and verifying api.

montekki:
Exposes proof generation and verification from `trie-db`.
Bumps `trie-db` and `memory-db` dependencies.


primitives/trie/src/lib.rs
```diff
@@ -1,5 +1,3 @@
-// Copyright 2015-2020 Parity Technologies (UK) Ltd.
-// This file is part of Substrate.
```
thiolliere: why ?

montekki: right

cheme:
Not sure if it should be part of this PR, but there is also : https://github.com/paritytech/trie/blob/d60e70bdd8c1e62af00459f9301c5002e69aaa21/trie-db/src/lib.rs#L105 that propose a way to have smaller proofs.
The method from this link targets proofs where we do not have content information (extrinsic execution or cumulus witness, or key value proof when we do not know the values).
It is still a processing / memory cost, and I am wondering if we should target support of multiple proof format. The method in this PR are specific enough that it seems very unlikely that someone would like to send a full proof instead.
I just realize that verification could be way faster if we verify against a previously checked proof (no need to recalculate the hash if they were already checked).
Just by curiosity what is the use case here (what I did understand from Jim was that it was mainly to check for change of values)?

montekki:
@cheme For instance it is intended to be used in [XCMP](https://research.web3.foundation/en/latest/polkadot/XCMP.html#solution-overview)

primitives/trie/src/lib.rs
```diff
@@ -727,4 +774,89 @@ mod tests {
 
 		assert_eq!(pairs, iter_pairs);
 	}
+
+    #[test]
```
bkchr: Spaces.

primitives/trie/src/lib.rs
```diff
@@ -119,6 +121,51 @@ pub mod trie_types {
...
+///
+/// For a key `K` that is included into the `db` a proof of inclusion is generated.
+/// For a key `K` that is not included into the `db` a proof of non-inclusion is generated.
+/// These can be later checked in [`verify_trie_proof`].
+///
+/// [`verify_trie_proof`]: ./fn.verify_trie_proof.html
+pub fn generate_trie_proof<'a, L: TrieConfiguration, I, K, DB>(
+    db: &DB,
+    root: TrieHash<L>,
+    keys: I
```
bkchr:
```suggestion
    keys: I,
```

primitives/trie/src/lib.rs
```diff
@@ -119,6 +121,51 @@ pub mod trie_types {
...
+/// Create a proof for a subset of keys in a trie.
+///
+/// The `keys` may contain any set of keys regardless of each one of them is included
+/// into the `db`.
+///
+/// For a key `K` that is included into the `db` a proof of inclusion is generated.
+/// For a key `K` that is not included into the `db` a proof of non-inclusion is generated.
+/// These can be later checked in [`verify_trie_proof`].
+///
+/// [`verify_trie_proof`]: ./fn.verify_trie_proof.html
```
bkchr:
```suggestion
```

Not understanding the reason for this line.

You already link the function above.

primitives/trie/src/lib.rs
```diff
@@ -119,6 +121,51 @@ pub mod trie_types {
...
+/// Verify a set of key-value pairs against a trie root and a proof.
+///
+/// Checks a set of keys with optional values for inclusion into the proof that was generated by
+/// [`generate_trie_proof`].
+/// If the value in the pair is supplied (`(key, Some(value))`), this key-value pair will be
+/// checked for inclusion into the proof.
+/// If the value is omitted (`(key, None)`), this key will be checked for non-inclusion into the
+/// proof.
+///
+/// [`generate_trie_proof`]: ./fn.generate_trie_proof.html
```
bkchr: The function is already linked above.

bkchr:
Benchmarks needs fixing.

montekki:
@bkchr yes, that is in process

primitives/trie/src/lib.rs
```diff
@@ -119,6 +121,47 @@ pub mod trie_types {
 	pub type TrieError<H> = trie_db::TrieError<H, super::Error>;
 }
 
+/// Create a proof for a subset of keys in a trie.
+///
+/// The `keys` may contain any set of keys regardless of each one of them is included
```
rphmeier: nit: "included into" should be "included in" - here and in other docs.

PR #4648, opened 2020-01-16
Add buffering to the TCP sockets

tomaka:
Attempt to fix the on-going performance issues.


tomaka:
Asked Jaco to test if it improves things.

tomaka:
Not working

PR #4649, opened 2020-01-16
Refactor the balances module

### LINKED ISSUE #4613: Refactor and optimise balances pallet
Balances pallet should be reorganised to optimise typical use-cases:
- Vesting should be split into its own pallet and utilise the locking subsystem. Claiming vested funds should be an explicit call which just reduces the `vesting` lock. This reduces the computation and storage access needed for making withdraw checks.
- Time-limits on locks should be removed as a feature. This allows a `locked_balance` field to be maintained (which is just the max of all the locks). This alone can be consulted when attempting a withdraw avoiding the need to iterate through the locks in for withdraw operations.
- Reserved balance, free balance and locks should be coalesced into a single storage item. Furthermore, there should be an additional member of the item whose type is configured by the config trait. This can be constructed in a similar manner to session keys, allowing other modules to deposit data into the same item. Since use-cases generally require reading various attributes of a single account, amalgamating data concerning an account from many different modules into the same storage item should increase the cache hits and improve performance, provided the size of the storage item's value doesn't get too large. (Ethereum makes this optimisation by combining balance, nonce, storage root and code hash into a single storage item.)

gavofyork:
Closes #4613 

Note that this "cheats" slightly and coalesces `WithdrawReasons` within balances to just two categories: TransactionPayment (Fees) and everything else (Misc). That allows an aggregate to be stored and avoids lock-list traversals in all balance operations without bloating the account storage item too much. It does mean that locking an account for `Transfers` now also locks it for `Reserve`s as well, but it's questionable whether they should ever really have been separate in the first place. I left `WithdrawReasons` as it is since I couldn't see any reason to dumb down the public API just because one implementation happens to implement it in a conservative way.

TODO:
- [x] Make sure I shouldn't be doing anything special with balances being instanceable.
- [x] Test migration.


frame/vesting/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+		if let Some(balance) = maybe_balance {
+			self.locked.max(balance) - balance
+		} else {
+			Zero::zero()
+		}
+	}
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as Sudo {
```
xlc:
```suggestion
	trait Store for Module<T: Trait> as Vesting {
```

frame/vesting/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+use frame_system::{self as system, ensure_signed, ensure_root};
+
+type BalanceOf<T> = <<T as Trait>::Currency as Currency<<T as frame_system::Trait>::AccountId>>::Balance;
+
+pub trait Trait: frame_system::Trait {
+	/// The overarching event type.
+	type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
+
+	/// The currency trait.
+	type Currency: Currency<Self::AccountId> + LockableCurrency<Self::AccountId>;
```
xlc: LockableCurrency is already Currency

frame/vesting/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+		BlockNumberToBalance: Convert<BlockNumber, Balance>
+	>(&self, n: BlockNumber) -> Balance {
+		// Number of blocks that count toward vesting
+		// Saturating to 0 when n < starting_block
+		let vested_block_count = n.saturating_sub(self.starting_block);
+		let vested_block_count = BlockNumberToBalance::convert(vested_block_count);
+		// Return amount that is still locked in vesting
+		let maybe_balance = vested_block_count.checked_mul(&self.per_block);
+		if let Some(balance) = maybe_balance {
+			self.locked.max(balance) - balance
```
xlc:
```suggestion
			self.locked.saturating_sub(balance)
```

frame/vesting/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+		if <Vesting<T>>::exists(who) {
+			Err(Error::<T>::ExistingVestingSchedule)?
+		}
+		let vesting_schedule = VestingInfo {
+			locked,
+			per_block,
+			starting_block
+		};
+		<Vesting<T>>::insert(who, vesting_schedule);
+		Ok(())
```
xlc: should also update the lock?

frame/vesting/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+	}
+
+	/// Remove a vesting schedule for a given account.
+	fn remove_vesting_schedule(who: &T::AccountId) {
+		<Vesting<T>>::remove(who);
```
xlc: and remove lock

frame/utility/src/lib.rs
```diff
@@ -703,19 +703,16 @@ mod tests {
...
 	parameter_types! {
 		pub const ExistentialDeposit: u64 = 0;
-		pub const TransferFee: u64 = 0;
 		pub const CreationFee: u64 = 0;
 	}
 	impl pallet_balances::Trait for Test {
 		type Balance = u64;
-		type OnFreeBalanceZero = ();
-		type OnReapAccount = System;
+type OnReapAccount = System;
```
xlc: indent

frame/vesting/src/lib.rs
```diff
@@ -0,0 +1,610 @@
...
+		}
+	}
+
+	/// Adds a vesting schedule to a given account.
+	///
+	/// If there already exists a vesting schedule for the given account, an `Err` is returned
+	/// and nothing is updated.
+	///
+	/// Is a no-op if the amount to be vested is zero.
+	fn add_vesting_schedule(
```
xlc: what happen if I want to make a sale with vesting schedule to an account with existing lock?
i.e.
Alice balance: $100, all locked by staking module
Alice participant a sale / IPO and have another $100 with a vesting schedule attached.
Alice now have $200 on balance, should have transferable amount of $0
But `T::Currency::set_lock` just add another lock with $100 amount.
Now Alice have $200 on balance, $100 locked with two locks, and $100 transferable.

This is only safe if there was no previous locks. I don't think it is safe to expose this to be used by other modules. More docs helps, but the use case of this method is still going to be very limited.

At least with #4655, it is possible to detect it is unsafe to add vesting schedule and return an error.

frame/balances/src/migration.rs
```diff
@@ -0,0 +1,84 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Some utilities for helping access storage with arbitrary key types.
```
xlc: this file seems useful as a general purpose migration utility, maybe move it to frame_support?

frame/vesting/src/lib.rs
gavofyork: This is desired behaviour. The idea is that vesting only applies to transferability and that the funds are otherwise usable. If you already had 100 locked for staking or whatever and you bought a pre-locked second 100, then you can think of is at the lock automatically being transferred to the new vesting-locked 100 and freeing up the old pre-vested 100.

frame/balances/src/migration.rs
gavofyork: yeah, probably should but i'd prefer to do it in a second PR.

frame/vesting/src/lib.rs
```diff
@@ -0,0 +1,610 @@
...
+			// * begin - Block when the account will start to vest
+			// * length - Number of blocks from `begin` until fully vested
+			// * liquid - Number of units which can be spent before vesting begins
+			for &(ref who, begin, length, liquid) in config.vesting.iter() {
+				let balance = T::Currency::free_balance(who);
+				assert!(!balance.is_zero(), "Currencies must be init'd before vesting");
+				// Total genesis `balance` minus `liquid` equals funds locked for vesting
+				let locked = balance.saturating_sub(liquid);
+				let length_as_balance = T::BlockNumberToBalance::convert(length);
+				let per_block = locked / length_as_balance.max(sp_runtime::traits::One::one());
```
shawntabrizi:
```suggestion
				let per_block = (locked / length_as_balance).max(sp_runtime::traits::One::one());
```

Looks like we are missing parenthesis here. We want to ensure `per_block` is at least 1.

i.e. 100 balance locked over 200 blocks, will have `per_block = 1`, which actually fully vest in 100 blocks.

shawntabrizi: Should probably add a test here for this.

frame/balances/src/lib.rs
```diff
@@ -306,40 +277,91 @@ decl_error! {
...
+	///
+	/// This balance is a 'reserve' balance that other subsystems use in order to set aside tokens
+	/// that are still 'owned' by the account holder, but which are suspendable.
+	pub reserved: Balance,
+	/// The amount that `free` may not drop below when withdrawing for *anything except transaction
+	/// fee payment*.
+	pub misc_frozen: Balance,
+	/// The amount that `free` may not drop below when withdrawing specifically for transaction
+	/// fee payment.
+	pub fee_frozen: Balance,
```
shawntabrizi: Just making a comment here while I dive into the code. It is not immediately clear to me the story around `misc_frozen` and `fee_frozen`.

A user will have a list of locks, each with a different id, amount, and a set of reasons.

The largest lock with reason ONLY for `TransactionPayment` will determine the `fee_frozen`.

The largest lock without `TransactionPayment` as the reason will determine the `misc_frozen`.

If a lock has some combination of reasons that include `TransactionPayment` and something else, we use the greatest of the two frozen values.

So I guess the question in my head is what is the most obvious scenario for having `misc_frozen` and `fee_frozen` be different values?

I assume `fee_frozen` would on average be less than `misc_frozen`, which would allow a user to use some balance specifically for fees, when they would otherwise be locked for other reasons?

What kinds of situations would this occur?

frame/vesting/src/lib.rs
gavofyork: image

The code is unchanged from before. The `.max(1)` is to ensure there's never a divide by zero.

frame/balances/src/lib.rs
gavofyork: Yes. Locks will typically be for: everything except fees (because we still want to let users pay for transactions with the locked funds - to unlock them for example); and everything (in the case where we don't even want them to pay fees with the locked funds). So basically `Reasons::Misc` and `Reasons::All`. It's very unlikely that `Reasons::Fees` would ever be used, but the code was simpler this way and it's strictly more flexible so I left it.

frame/vesting/src/lib.rs
gavofyork: For any sensible chain, and certainly the chains this is designed for, the average balance released per block is greater than *the smallest possible unit*, so this won't be an issue. In any case it's not something that needs fixing in this PR.

gavofyork: If you want, feel free to create an issue.

gavofyork:
@shawntabrizi any further comments?

gavofyork:
not sure we really can; re: execution-time it's not a big deal - it'll just be a long block. memory usage should be fine in principle, however if there's some issue with memory leakage (like that you found), then it could be a problem. i'd like to know what's going on with that before upgrading.

gavofyork:
ok - obvious way to test is to dump balances state, spin up a local testnet with it as genesis and then execute with wasm.

gavofyork:
Tested it on a dev chain, and can confirm it works fine in wasm. (Actually it's surprisingly fast migrating ~7k accounts)

Polkadot-Forum:
This pull request has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/moving-frame-out-of-substrate/907/1

PR #4651, opened 2020-01-16
Patch practical usability issues with Society

shawntabrizi:
This PR makes the following changes:

1. Currently, if you try to call `found` on the Society pallet on a running chain, it will fail since the `MaxMembers` storage item is default 0. This introduces a second parameter to the `found` extrinsic, `max_members` which will set this value to some non-zero value.

2. The Polkadot JS API does not recognize a storage item with type `Option<()>` as ever `isSome`. I change the use of `Option<()>` to `Option<bool>` for tracking suspended members. https://github.com/polkadot-js/api/issues/1724

3. This also updates the Substrate node so that Society has a genesis configuration, and thus running a new Substrate node will make society enabled and usable.


frame/society/src/lib.rs
```diff
@@ -813,10 +814,12 @@ decl_module! {
 		/// Total Complexity: O(1)
 		/// # </weight>
 		#[weight = SimpleDispatchInfo::FixedNormal(10_000)]
-		fn found(origin, founder: T::AccountId) {
+		fn found(origin, founder: T::AccountId, max_members: u32) {
```
shawntabrizi: Could also consider including a `pot` parameter which will bootstrap the pot, else we will wait 1 rotation period for it to grow.

If we include `pot`, we could also automatically transfer that amount from the founder to the Society account.

bin/node/cli/src/chain_spec.rs
```diff
@@ -295,6 +295,11 @@ pub fn testnet_genesis(
 		}),
 		pallet_membership_Instance1: Some(Default::default()),
 		pallet_treasury: Some(Default::default()),
+		pallet_society: Some(SocietyConfig {
+			members: endowed_accounts[0..3].to_vec(),
+			pot: 100000,
```
shawntabrizi: Double check this number

frame/society/src/lib.rs
gavofyork: the founder isn't meant to pay anything, and giving the `SetFounderOrigin` the capability of deducting as much as they want from any account in the system is obviously dumb.

gavofyork:
> to `Option<bool>`

why not to a straight `bool`? it's more ergonomic and doesn't provide a dead state to the developer.

bin/node/cli/src/chain_spec.rs
gavofyork: pot should probably just be zero.

bin/node/cli/Cargo.toml
```diff
@@ -72,6 +72,7 @@ pallet-transaction-payment = { version = "2.0.0", path = "../../../frame/transac
 frame-support = { version = "2.0.0", default-features = false, path = "../../../frame/support" }
 pallet-im-online = { version = "2.0.0", default-features = false, path = "../../../frame/im-online" }
 pallet-authority-discovery = { version = "2.0.0",  path = "../../../frame/authority-discovery" }
+pallet-society = { version = "2.0.0",  path = "../../../frame/society" }
```
bkchr: Why do we need this dependency?

shawntabrizi: It gets used here: https://github.com/paritytech/substrate/pull/4651/files#diff-c9f62bd741e70b8157fa630030c33a57R298

PR #4652, opened 2020-01-16
Default fork choice value and intermediates for block import parameters

sorpaas:
Addresses some of #3623. 

This PR tries to improve several issues in the consensus engines:

* Currently, if users pass on an incompatible `BlockImport` with a `Verifier`, consensus will silently fail.
* Because `Verifier` does not assume parents, it is not always possible for it to return the correct fork choice rule, but currently it is forced to do so.
* Methods to pipe-line data downwards from `Verifier` to all subsequent `BlockImport`s is not obvious.

In this PR:

* Changed `ForkChoiceStrategy` to be `Option`. If `None`, it indicates that the `BlockImport` or `Verifier` cannot currently make a decision on the fork choice rule. Subsequent `BlockImport` should properly modify the value. If the value is passed all the way down to the bottom `BlockImport`, then the block will not be imported.
* Added `intermediates` in `BlockImportParams`. This indicates values that should be processed by subsequent `BlockImport`s. If a value in `intermediates` is not processed, the block import fails. As an example usage of this, in PoW engine, `Verifier` uses `intermediates` to pass the calculated difficulty (if available) and the full block hash down to `BlockImport`, who will then use them to calculate the block auxiliary.
* Updated PoW engine to have proper separation of `BlockImport` and `Verifier`.


primitives/consensus/common/src/block_import.rs
```diff
@@ -142,13 +142,17 @@ pub struct BlockImportParams<Block: BlockT, Transaction> {
 	/// Is this block finalized already?
 	/// `true` implies instant finality.
 	pub finalized: bool,
+	/// Intermediate values that are interpreted by block importers. Each block importer,
+	/// when handled a value, removes it from the intermediate list. The final block importer
```
rphmeier:
```suggestion
	/// upon handling a value, removes it from the intermediate list. The final block importer
```

primitives/consensus/common/src/block_import.rs
```diff
@@ -142,13 +142,17 @@ pub struct BlockImportParams<Block: BlockT, Transaction> {
...
+	/// rejects block import if there are still intermediate values remain unhandled.
```
rphmeier:
```suggestion
	/// rejects block import if there are still intermediate values that remain unhandled.
```

(rephrasing for clarity / grammar)

primitives/consensus/common/src/block_import.rs
```diff
@@ -142,13 +142,17 @@ pub struct BlockImportParams<Block: BlockT, Transaction> {
...
+	pub intermediates: HashMap<Vec<u8>, Vec<u8>>,
 	/// Auxiliary consensus data produced by the block.
 	/// Contains a list of key-value pairs. If values are `None`, the keys
 	/// will be deleted.
 	pub auxiliary: Vec<(Vec<u8>, Option<Vec<u8>>)>,
 	/// Fork choice strategy of this import. This should only be set by a
 	/// synchronous import, otherwise it may race against other imports.
-	pub fork_choice: ForkChoiceStrategy,
+	pub fork_choice: Option<ForkChoiceStrategy>,
```
rphmeier: Please add docs describing the behavior of the final importer if `None` is supplied.

client/consensus/pow/src/lib.rs
```diff
@@ -243,55 +249,165 @@ impl<B: BlockT, C, S, Algorithm> PowVerifier<B, C, S, Algorithm> {
...
+		header: B::Header,
+		justification: Option<Justification>,
+		body: Option<Vec<B::Extrinsic>>,
+	) -> Result<(BlockImportParams<B, ()>, Option<Vec<(CacheKeyId, Vec<u8>)>>), String> {
+		let hash = header.hash();
+		let (checked_header, seal) = self.check_header(header)?;
+
+		let intermediate = PowIntermediate::<B, Algorithm::Difficulty> {
+			hash: hash,
+			difficulty: None,
```
rphmeier: should we not verify the difficulty here in the `Verifier`, as that's the parallelizable operation?

rphmeier:
LGTM except for question above

client/consensus/pow/src/lib.rs
sorpaas: Difficulty may not be parallelizable operation. The mining algorithm is always fixed (or can only be chosen from a set of validated ones), but for difficulty adjustment algorithms, they can be defined in runtime. In those cases, block importer will have to do a runtime call on the parent block to fetch the difficulty. Because parent block may not be available in `Verifier`, we can only set it to `None` here, meaning we do not yet know the value.

However, in mining routine, the value of difficulty is indeed available because we mine on local best block. In that case, difficulty is set to `Some` so we can avoid double calculation.

sorpaas: Difficulty may not be parallelizable operation. The mining algorithm is always fixed (or can only be chosen from a set of validated ones), but for difficulty adjustment algorithms, they can be defined in runtime. In those cases, block importer will have to do a runtime call on the parent block to fetch the difficulty. Because parent block may not be available in `Verifier`, we can only set it to `None` here, meaning we do not yet know the value.

However, in mining routine, the value of difficulty is indeed available because we mine on local best block. In that case, difficulty is set to `Some` so we can avoid double calculation.

andresilva:
For babe we might also be able to use `intermediates` to pass epoch data from verifier to block import (to avoid fetching it twice from tree). I think this can be done on a follow-up PR though.

PR #4657, opened 2020-01-17
Add some extra traces to basic-authorship

NikVolf:


bkchr:
Why is that not printed in the transaction pool?

NikVolf:
@bkchr I moved one trace about invalid in txpool

But pool status shouldn't just print itself every time it is queried, so I left one there

PR #4658, opened 2020-01-17
Drive by fix of doc of `Value`.

pepyakin:


primitives/wasm-interface/src/lib.rs
```diff
@@ -39,13 +39,17 @@ pub enum ValueType {
...
 #[derive(PartialEq, Debug, Clone, Copy)]
 pub enum Value {
-	/// An `i32` value.
+	/// A 32-bit integer.
 	I32(i32),
-	/// An `i64` value.
+	/// A 64-bit integer.
 	I64(i64),
-	/// An nan-preserving `f32` value.
+	/// A 32-bit float point number stored as raw bit pattern.
```
andresilva:
```suggestion
	/// A 32-bit floating-point number stored as raw bit pattern.
```

primitives/wasm-interface/src/lib.rs
```diff
@@ -39,13 +39,17 @@ pub enum ValueType {
...
+	///
+	/// You can materialize this value using `f32::from_bits`.
 	F32(u32),
-	/// An nan-preserving `f64` value.
+	/// A 64-bit float point number stored as raw bit pattern.
```
andresilva:
```suggestion
	/// A 64-bit floating-point number stored as raw bit pattern.
```

pepyakin:
Retrying CI...

PR #4661, opened 2020-01-17
client/finality-grandpa: Reintegrate gossip validator report stream

mxinden:
The `finality-grandpa` `GossipValidator` is called by the `GossipEngine`
in a synchronous fashion on each gossip message. Its main task is to
decide whether to gossip the given message on, or whether to drop it.

In addition it also updates the reputation of a node's peers based on
the incoming gossip messages. To do so it needs to be able to report the
reputation change which it does through an unbounded channel (in order
to stay synchronous).

Previously the receiving side of this channel would be handled by a new
task, polling the channel and forwarding the changes to a clone of the
`GossipEngine` that it would own.

Instead the receiver of the above mentioned channel is now being polled
by the `NetworkBridge` within its `Future::poll` implementation.
Reputation changes are reported through the already existing
`GossipEngine` instance within `NetworkBridge`.

For details on the overall goal, see d4fbb897c.

### DIFF
--- a/client/finality-grandpa/src/communication/gossip.rs
+++ b/client/finality-grandpa/src/communication/gossip.rs
@@ -83,15 +83,15 @@
 //! We only send polite messages to peers,
 
 use sp_runtime::traits::{NumberFor, Block as BlockT, Zero};
-use sc_network_gossip::{GossipEngine, MessageIntent, ValidatorContext};
+use sc_network_gossip::{MessageIntent, ValidatorContext};
 use sc_network::{config::Roles, PeerId, ReputationChange};
 use parity_scale_codec::{Encode, Decode};
 use sp_finality_grandpa::AuthorityId;
 
 use sc_telemetry::{telemetry, CONSENSUS_DEBUG};
-use log::{trace, debug, warn};
+use log::{trace, debug};
 use futures::prelude::*;
-use futures::sync::mpsc;
+use futures03::channel::mpsc;
 use rand::seq::SliceRandom;
 
 use crate::{environment, CatchUp, CompactCommit, SignedMessage};
@@ -1178,7 +1178,7 @@ impl<Block: BlockT> GossipValidator<Block> {
 	pub(super) fn new(
 		config: crate::Config,
 		set_state: environment::SharedVoterSetState<Block>,
-	) -> (GossipValidator<Block>, ReportStream)	{
+	) -> (GossipValidator<Block>, mpsc::UnboundedReceiver<PeerReport>)	{
 		let (tx, rx) = mpsc::unbounded();
 		let val = GossipValidator {
 			inner: parking_lot::RwLock::new(Inner::new(config)),
@@ -1186,7 +1186,7 @@ impl<Block: BlockT> GossipValidator<Block> {
 			report_sender: tx,
 		};
 
-		(val, ReportStream { reports: rx })
+		(val, rx)
 	}
 
 	/// Note a round in the current set has started.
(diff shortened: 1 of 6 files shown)


client/finality-grandpa/src/communication/mod.rs
```diff
@@ -148,9 +154,18 @@ pub(crate) struct NetworkBridge<B: BlockT, N: Network<B>> {
...
 	neighbor_packet_worker: Arc<Mutex<periodic::NeighborPacketWorker<B>>>,
+
+	/// Receiver side of the peer report stream populated by the gossip validator, forwarded to the
+	/// gossip engine.
+	//
+	// `NetworkBridge` is required to be clonable, thus one needs to be able to clone its children,
+	// thus one has to wrap gossip_validator_report_stream with an `Arc` `Mutex`. Given that it is
+	// just an `UnboundedReceiver`, one could also switch to a multi-producer-*multi*-consumer
+	// channel implementation.
+	gossip_validator_report_stream: Arc<Mutex<mpsc03::UnboundedReceiver<PeerReport>>>,
```
rphmeier: out of curiosity, what does the plan to phase out the unbounded channel look like after this?

client/finality-grandpa/src/communication/gossip.rs
```diff
@@ -1445,57 +1445,9 @@ impl<Block: BlockT> sc_network_gossip::Validator<Block> for GossipValidator<Bloc
...
-				Ok(Async::Ready(None)) => return Ok(Async::Ready(())),
-				Ok(Async::Ready(Some(PeerReport { who, cost_benefit }))) =>
-					self.net.report(who, cost_benefit),
-				Ok(Async::NotReady) => return Ok(Async::NotReady),
-			}
-		}
-	}
+pub(super) struct PeerReport {
+	pub who: PeerId,
+	pub cost_benefit: ReputationChange,
```
rphmeier: docs on this struct & members?

mxinden:
> out of curiosity, what does the plan to phase out the unbounded channel look like after this?

@rphmeier my plan roughly looks like this:

1. I have a pull request for removing the [`unbounded` within `round_communication`](https://github.com/paritytech/substrate/blob/master/client/finality-grandpa/src/communication/mod.rs#L346) given that the function returns a `Stream` and a `Sink`, which have asynchronous interfaces.

2. With https://github.com/paritytech/substrate/pull/4631 and this pull request and a couple of other changes (e.g. [integrating offband worker tasks](https://github.com/paritytech/substrate/blob/master/client/network-gossip/src/bridge.rs#L68)) `GossipEngine` would not need to be clonable anymore.

3. Once that happened I can remove the [`struct { inner: Arc<Mutex<>> }`](https://github.com/paritytech/substrate/blob/master/client/network-gossip/src/bridge.rs#L32) construction on `GossipEngine.

4. With exclusive ownership I can use async within [`messages_for`](https://github.com/paritytech/substrate/blob/master/client/network-gossip/src/state_machine.rs#L329), being able to remove its unbounded channel.

5. ....

client/finality-grandpa/src/communication/gossip.rs
mxinden: See https://github.com/paritytech/substrate/pull/4684

client/finality-grandpa/src/communication/mod.rs
mxinden: See https://github.com/paritytech/substrate/pull/4661#issuecomment-576257780.

PR #4662, opened 2020-01-17
Make Staking pallet using a proper Time module.

thiolliere:
(EDITED)

the staking modules make use of Time trait to get the time in millisecond, however this a wrong usage of such traits which doesn't specifies what is moment.

To solve this this PR introduce a new trait named UnixTime implemented by timestamp and used by staking.

As said by Kian it would be great to have more docs on the Time trait itself and usage it is suppose to have. But I struggle to define any documentation for this trait.

The migration is implemented in such a way that the active era start at the moment of the migration.
This is needed because even if the wrong usage of Time correctly returned a moment in millisecond there is no guarantee that this moment is since unix_epoch.

On Kusama however the old Time was indeed the millisecond since unix epoch thus we could improve kusama migration.


frame/support/src/traits.rs
```diff
@@ -653,9 +653,14 @@ bitmask! {
 	}
 }
 
+/// A trait to query the time of the current block.
 pub trait Time {
+	/// Type used for expressing timestamp.
 	type Moment: SimpleArithmetic + Parameter + Default + Copy;
 
+	/// Return the last timestamp set by inherent. If called before the inherent providing
```
bkchr: Hmm, why do we have the trait when it exactly replicates the behavior of the `timestamp` pallet? I would assume that time it something that strictly increases between blocks and could be in a simple case just the block number.

The documentation should go above the implementation of the `Time` trait for the `timestamp` pallet.

Here we should have a generic description what this trait provides you. IMHO.

thiolliere: it is used by staking module to compute the duration of an era.

So I wanted to make the documentation of this trait explicit enough so that is it usable by staking module as is, without enforcing a specific implementation of this trait.

Currently staking module do call this trait during on_initialize because it is called by session module during on_initialize. (thus the duration of the era is slightly off actually)

Also I just realized staking module expect to have this value expressed as millisecond, which is not enforced anywhere in those traits. maybe we should bound more on this trait so that Moment::year() returns the value for a duration of one year. (or something else to get this information).

bkchr: This sounds completely implementation specific to the staking pallet. We should document that the `Time` type of staking expects milliseconds.

I'm not against documentation on this trait, but when we put all these constraints, we can just import `timestamp` and remove the trait. Than we are sure that all the constraints are fulfilled, otherwise anyone could provide *anything* that implements this trait.

Maybe we should just be more specific on what the `now` method returns. This means, we should maybe rename the `now` function to `now_milliseconds` or similar. Or provide multiple of these methods that could return an `Option`. 

I'm not even sure how the `staking` module can expect that something is milliseconds, while even the type for the time is generic. I think there is some huge miss-use with this trait.

We need to clearly improve, but just putting some documentation above the the time trait, will not solve the `staking` problem IMHO. Maybe we should have a more specific time trait that exposes functions like `now_milliseconds`, `now_seconds` etc, so that the implementer is aware what is expected.

thiolliere: about the seconds/milliseconds issue yes, we can precise the unit of `now` or require a const in the trait which is `YEAR` or `HOUR` or `SECOND`.

for the issue about when now returns a correct value I don't maybe it is fine to put in the documentation of the implementation of timestamp that it is not available at genesis config and that before the timestamp inherent it return the time of previous block.
Then in staking module we can specify that Time is not called during genesis config. and for the computation of the era duration (which is slightly off) we can say accept that like an "imprecision" of the available clock (as we queryied just before it is updated), or we can also fix that and say in the doc of staking that Time is only called during on_finalize.

bkchr:
Any update on this?

thiolliere:
OK I made an alternative implementation, otherwise I can do now_millis but I feel like this way the moment of `now` is a bit more configurable.

frame/support/src/traits.rs
```diff
@@ -653,9 +653,15 @@ bitmask! {
...
+/// Trait to deals with time.
 pub trait Time {
+	/// Type for the time.
 	type Moment: SimpleArithmetic + Parameter + Default + Copy;
 
+	/// Duration for a Julian year (365.25 days).
+	fn year() -> Self::Moment;
+
+	/// Return current moment.
 	fn now() -> Self::Moment;
```
bkchr: Now you are mixing some time related functionality (`year()`) with a `moment` that can be anything. 

I still think we should keep this `Time` trait simple and say that `now()` will return a `moment` that increases between blocks, but without specifying what it is. It could also be the moment from the genesis of the chain.

I would add another trait `UnixTime` that clearly says that `now()` returns the `Duration` from the unix epoch. I would directly return `core::time::Duration`. This already brings all the functions for converting between the different time formats.

The timestamp module would implement both of these traits and return the same value for both `now()` calls, but as different types. `Staking` would require that `Time` implements `UnixTime`.

thiolliere: yes, I completely agree

frame/staking/src/lib.rs
```diff
@@ -859,8 +858,9 @@ decl_module! {
 
 		fn on_finalize() {
 			// Set the start of the first era.
-			if !<CurrentEraStart<T>>::exists() {
-				<CurrentEraStart<T>>::put(T::Time::now());
+			if !CurrentEraStart::exists() {
+				let now_as_millis_u64 = T::UnixTime::now().as_millis().saturated_into::<u64>();
```
thiolliere: note: could also implement Codec for duration, but anyway u64 is big enough

frame/staking/src/migration.rs
```diff
@@ -111,6 +112,15 @@ pub mod inner {
 		}
 	}
 
+	// migrate storage from v2 to v3: Insert into CurrentEraStart the current time.
+	pub fn to_v3<T: Trait>(version: &mut VersionNumber) {
+		if *version != 2 { return }
+		*version += 1;
+
+		let now_as_millis_u64 = T::UnixTime::now().as_millis().saturated_into::<u64>();
+		<Module<T> as Store>::CurrentEraStart::put(now_as_millis_u64);
```
thiolliere: We don't try to update from current era start because the moment might not be the same as UnixTime. (like it could be millisecond from the start of the chain)

But if we consider migration only for Kusama we could indeed do a translation.

Note: that this will decrease the duration of the current era, thus decreasing its reward

gnunicorn:
please update to latest master.

frame/staking/src/mock.rs
```diff
@@ -203,7 +203,7 @@ parameter_types! {
 }
 impl Trait for Test {
 	type Currency = pallet_balances::Module<Self>;
```
kianenigma:
```suggestion
	type Currency = Balances;
```

frame/staking/src/mock.rs
```diff
@@ -203,7 +203,7 @@ parameter_types! {
...
-	type Time = pallet_timestamp::Module<Self>;
+	type UnixTime = pallet_timestamp::Module<Self>;
```
kianenigma:
```suggestion
	type UnixTime = Timestamp;
```

kianenigma:
I think it should be explained in the timestamp module when another module (i.e. staking in this case) should use `Time` impl and and when `UnixTime`.

bkchr:
> I think it should be explained in the timestamp module when another module (i.e. staking in this case) should use `Time` impl and and when `UnixTime`.

Why? It depends on what the module requires on constraints and not what the timestamp module provides.

You could just have the requirement of time increasing which could be anything, even just the block number.

Or like staking you really need UNIX time.

thiolliere:
> I think it should be explained in the timestamp module when another module (i.e. staking in this case) should use `Time` impl and and when `UnixTime`.

to be honest I struggle to know what are the guarantee of Time, maybe that it can only increase.

kianenigma:
> **Why**? It depends on what the module requires on constraints and not what the timestamp module provides.

I think it is helpful if timestamp says: if you need _this_ then use `x`, if you need _that_ then use `y`. And all in all not a big concern, just a recommendation. Ignore if you think it is not necessary 👍

bkchr:
That needs to be documented on the trait, not the timestamp module. The timestamp module can describe what it provides.

frame/staking/src/lib.rs
```diff
@@ -660,6 +659,11 @@ pub trait Trait: frame_system::Trait {
 	/// For each validator only the `$MaxNominatorRewardedPerValidator` biggest stakers can claim
 	/// their reward. This used to limit the i/o cost for the nominator payout.
 	type MaxNominatorRewardedPerValidator: Get<u32>;
+
+	/// Deprecated Time.
+	///
+	/// Deprecated associated type wrongly used to query Time in millisecond.
+	type DeprecatedTime: Time;
```
bkchr: Why do we need this?

thiolliere: the main use it to be able to decode the old structure `ActiveInfo<MomentOf<T>>`.
But indeed I can just decode the EraIndex and not start field as I don't use it.

I updated code as such, (also fixing that StorageVersion wasn't set)

bkchr:
What is your plan with this @thiolliere ?

frame/staking/src/lib.rs
```diff
@@ -686,6 +686,8 @@ impl Default for Forcing {
 enum Releases {
 	V1_0_0Ancient,
 	V2_0_0,
+	V3_0_0,
+	V4_0_0,
 }
 
 impl Default for Releases {
```
thiolliere: I don't know what to do with this default releases.
If the storage is empty then it means either that the storage version is before the first runtime_upgrade that introduced this storage, or that this module has been introduced in a running chain (without building its genesis config).

frame/staking/src/lib.rs
```diff
@@ -686,11 +686,12 @@ impl Default for Forcing {
 enum Releases {
 	V1_0_0Ancient,
 	V2_0_0,
+	V3_0_0,
 }
 
 impl Default for Releases {
 	fn default() -> Self {
-		Releases::V2_0_0
+		Releases::V3_0_0
```
thiolliere: I'm not sure what to think about this default, because if we add staking in a running chain (so without executing GenesisConfig) then we must insert the storage version in the runtime upgrade.
I think a staking module without any storage version must be considered invalid now.

otherwise if a chain introduce staking module without inserting the storage version then the storage version depends on when it has been inserted, and storage version could be V2 or V3 there is no way to know it.

thiolliere:
> What is your plan with this @thiolliere ?

OK as the strategy for now is only to support Kusama I updated the runtime_upgrade code.
And merged master.

The PR is now ready for review

frame/timestamp/src/lib.rs
```diff
@@ -239,6 +241,23 @@ impl<T: Trait> Time for Module<T> {
...
+impl<T: Trait> UnixTime for Module<T> {
+	fn now() -> core::time::Duration {
+		// now is duration since unix epoch in millisecond as documented in
+		// `sp_timestamp::InherentDataProvider`.
+		let now = Self::now();
+		if now == T::Moment::zero() {
+			debug::error!(
+				"`pallet_timestamp::UnixTime::now` is called at genesis, invalid value returned: 0"
+			);
+		}
```
bkchr:
```suggestion
		sp_std::if_std! { 
			if now == T::Moment::zero() {
				debug::error!(
					"`pallet_timestamp::UnixTime::now` is called at genesis, invalid value returned: 0"
				);
			}
		}
```

bkchr:
Please merge master.

kianenigma:
Needed a polkadot companion PR  :(

thiolliere:
sorry done here https://github.com/paritytech/polkadot/pull/941

PR #4665, opened 2020-01-17
Remove `NetworkSpecialization`

rphmeier:
This PR removes the `NetworkSpecialization` from the network service.
`NetworkSpecialization` was an early attempt at customizing the node's network behavior for Polkadot, but there are now better APIs for doing so (`register_notifications_protocol`).

## Historical notes on backwards compatibility

This PR does not introduce network backwards-incompatibility. Potential network backwards incompatibility has been addressed by this series of issues:
  1. Polkadot: begin to ignore the `chain_status` parameter: https://github.com/paritytech/polkadot/issues/779
  2. Substrate: begin to ignore the `chain_status` in the handshake when decoding: #4675 
  3. A few releases down the road (once Polkadot no longer depends on it), remove the `chain_status` member from the handshake: https://github.com/paritytech/substrate/issues/4674

*Breaking API Changes*:
The service API is no longer the same. The `construct_simple_protocol` macro `with_network_protocol` function have been removed. Invocations of the macro and function should be deleted without loss of functionality (except when `NetworkSpecialization` was non-empty - then you need to port to the `register_notifications_protocol`).


rphmeier:
Once https://github.com/paritytech/polkadot/pull/788 is in we can start to look at merging this. It still needs a follow-up to enable the new network protocol.

client/network/src/protocol/legacy_proto/tests.rs
```diff
@@ -275,9 +279,21 @@ fn basic_two_nodes_requests_in_parallel() {
 	// Generate random messages with or without a request id.
 	let mut to_send = {
 		let mut to_send = Vec::new();
+		let mut existing_ids = HashSet::new();
 		for _ in 0..200 { // Note: don't make that number too high or the CPU usage will explode.
-			let msg = (0..10).map(|_| rand::random::<u8>()).collect::<Vec<_>>();
-			to_send.push(Message::<Block>::ChainSpecific(msg));
+			let req_id = loop {
```
mxinden: Just to make sure I understand this correctly, you have witnessed a u64 collision repeatedly? Even with the birthday attack in mind, that is rather unlikely, no? Am I missing something?

https://en.wikipedia.org/wiki/Birthday_attack#Mathematics

rphmeier: According to the diagram, this test would fail with a probability of about 1/10^15. So yeah, probably overkill. TBH, I thought that `RequestId` was `u32` (and I suppose the type may yet change), so it seems safer to protect against that. If you push for it I'll change it.

gnunicorn:
https://github.com/paritytech/polkadot/pull/788 has been merged into polkadot. Shall we try again to get this merged in substrate, too?

rphmeier:
@gnunicorn We need a non-trivial follow-up or it will break Polkadot master. I'm hoping to do that this week. Regardless, we should merge before feature-freeze.

gnunicorn:
Thanks for the update, @rphmeier .

One idea behind the freeze is to give us some time to upgrade any dependents accordingly without having many unrelated (API) changes... fixing Polkadot after the freeze/during alpha+beta is totally reasonable.

Could someone merge latest master in?

tomaka:
I don't think merging master is hard, but let me know if I should do it.

rphmeier:
> fixing Polkadot after the freeze/during alpha+beta is totally reasonable.

Our rule is usually to have a PR for Polkadot waiting. My TODO list is fairly long before I can get to that. We will probably have to make an exception here, because I'd like this to not go in 2.0.

tomaka:
Merged master.

gnunicorn:
@rphmeier anything else to do here before we can merge into substrate?

Re Polkadot: I think we can make an exception for the freeze.

gnunicorn:
needs another merge with master...

rphmeier:
OK, shall we merge directly before the freeze? I'd like to minimize breakage in Polkadot as much as possible.

andresilva:
@rphmeier We can also exclude this PR when backporting to `polkadot-master` until we have support for this in polkadot.

gnunicorn:
icing until the freeze (pun intended)

gnunicorn:
could someone with enough expertise update to master, resolved the conflicts and merge this into master?

//cc @rphmeier, @tomaka

tomaka:
Master-merging round 3 done.

tomaka:
The PR's ready to be merged. Since I haven't really followed the story regarding the Substrate freeze and Polkadot, I will let someone else do it.

rphmeier:
I have a PR in the works which adjusts Polkadot to use this.

PR #4671, opened 2020-01-18
Add rules and unfounding to society.

gavofyork:


frame/society/src/lib.rs
```diff
@@ -823,8 +829,37 @@ decl_module! {
...
+		///
+		/// The dispatch origin for this call must be Signed, and the signing account must be the
+		/// `Founder`.
+		///
+		/// # <weight>
+		/// - One storage read, O(M).
+		/// - Four storage removals O(1).
+		/// - One event.
+		///
+		/// Total Complexity: O(1)
```
NikVolf: Sorry, I probably lack context, but how `O(M)` in `One storage read` gives total `O(1)` here?

NikVolf: `O(M)` above probably means the length of the storage value, right?

gavofyork: yeah but you're right - that was a typo.

gavofyork:
@NikVolf feel free to approve otherwise :)

frame/society/src/lib.rs
```diff
@@ -823,8 +829,36 @@ decl_module! {
...
 			<Head<T, I>>::put(&founder);
 			<Founder<T, I>>::put(&founder);
+			Rules::<T, I>::put(T::Hashing::hash(&rules));
 			Self::deposit_event(RawEvent::Founded(founder));
 		}
+
+		/// Anull the founding of the society.
+		///
+		/// The dispatch origin for this call must be Signed, and the signing account must be both
+		/// the `Founder` and the `Head`. This implies that it may only be done when there is one
```
bkchr: If I understand the code correctly, the Head will only be updated after some period. Couldn't this lead to a state at which members are already added, but the founder is still the head?

gavofyork: candidates might be selected, but Head changes only with the content of Members afaik.

shawntabrizi: Yeah there seems to be some problems with this:

Once one or more additional members are added to society, someone else is guaranteed to become the Head. Once they are Head, there is no way for them to be removed from society. So the Founder can never become Head once there is a second member. There will always be two members.

There is also an issue that @bkchr describes. At genesis, if you set initial members for the society, Founder will also be set as the Head, and with non-zero members, they would be able to call this function. Additionally, if somehow the founder did become both head and founder again as the last member (with some updated logic), it is possible that a suspended member becomes unsuspended, and then you will have non-zero members which satisfy these conditions.

frame/society/src/lib.rs
```diff
@@ -823,8 +829,36 @@ decl_module! {
...
+		///
+		/// Total Complexity: O(1)
+		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedNormal(20_000)]
+		fn unfound(origin) {
+			let founder = ensure_signed(origin)?;
+			ensure!(Founder::<T, I>::get() == Some(founder.clone()), Error::<T, I>::NotFounder);
+			ensure!(Head::<T, I>::get() == Some(founder.clone()), Error::<T, I>::NotHead);
+
+			Members::<T, I>::kill();
```
shawntabrizi: Seems we should `kill` / `remove_all()` all storage items? Including votes, suspended members, suspended candidates, etc...

frame/society/src/tests.rs
```diff
@@ -42,11 +43,39 @@ fn founding_works() {
...
+		// 1 sets the founder to 20 this time
+		assert_ok!(Society::found(Origin::signed(1), 20, 100, vec![]));
+		// Bring in a new member...
+		assert_ok!(Society::bid(Origin::signed(10), 0));
+		run_to_block(4);
+		assert_ok!(Society::vote(Origin::signed(20), 10, true));
+		run_to_block(8);
+
+		// Unfounding won't work now, even though it's from 20.
+		assert_noop!(Society::unfound(Origin::signed(20)), Error::<Test, _>::NotHead);
```
shawntabrizi: As mentioned in the comment above, this new user will become the "head", and there is no way for them to lose being the head.

frame/society/src/lib.rs
gavofyork: certainly candidates.

frame/society/src/tests.rs
gavofyork: resolved below

frame/society/src/lib.rs
gavofyork: i don't see much need to remove anything else.

frame/society/src/lib.rs
gavofyork: > with some updated logic

future changes that break current assumptions are hardly a concern here.

PR #4672, opened 2020-01-18
Update bridge pallet to work with the (almost) lastest master

HCastano:
I've rebased the `hc-jp-bridge-module` branch onto `master` from last week-ish, and now in this PR I've make some changes to make the pallet compile. The two main things are using the new `decl_error!` changes introduced by Basti, and updating dependencies to use their 2.0.0 versions.

### DIFF
--- a/frame/bridge/src/lib.rs
+++ b/frame/bridge/src/lib.rs
@@ -42,6 +42,7 @@ use fg_primitives::{AuthorityId, AuthorityWeight};
 use sp_runtime::traits::Header;
 use state_machine::StorageProof;
 use frame_support::{
+	dispatch::{DispatchResult, DispatchError},
 	decl_error, decl_module, decl_storage,
 };
 use frame_system::{self as system, ensure_signed};
@@ -134,11 +135,11 @@ decl_error! {
 	}
 }
 
-impl<T: Trait> From<StorageError> for Error<T> {
+impl<T: Trait> From<StorageError> for DispatchError {
 	fn from(error: StorageError) -> Self {
 		match error {
-			StorageError::StorageRootMismatch => Error::<T>::StorageRootMismatch,
-			StorageError::StorageValueUnavailable => Error::<T>::StorageValueUnavailable,
+			StorageError::StorageRootMismatch => Error::<T>::StorageRootMismatch.into(),
+			StorageError::StorageValueUnavailable => Error::<T>::StorageValueUnavailable.into(),
 		}
 	}
 }
@@ -148,7 +149,7 @@ impl<T: Trait> Module<T> {
 		state_root: &T::Hash,
 		proof: StorageProof,
 		validator_set: &Vec<(AuthorityId, AuthorityWeight)>,
-	) -> Result<(), Error<T>> {
+	) -> DispatchResult {
 
 		let checker = <StorageProofChecker<<T::Hashing as sp_runtime::traits::Hash>::Hasher>>::new(
 			*state_root,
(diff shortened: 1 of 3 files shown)


frame/bridge/src/lib.rs
```diff
@@ -136,67 +139,84 @@ impl<T: Trait> Module<T> {
...
-	) -> Result<(), Error> {
+	) -> DispatchResult {
 
 		let checker = <StorageProofChecker<<T::Hashing as sp_runtime::traits::Hash>::Hasher>>::new(
 			*state_root,
 			proof.clone()
-		)?;
+		);
+
+		// TODO: Remove once I can impl From<StorageError> for DispatchError
```
HCastano: Ideally I would've like to have implemented this, but with the way `decl_error!` is written at the moment I'm not able to do this. If anybody has any better ideas than the `map_storage_err` function I'm using at the moment I'm all ears.

tomusdrw: Should be just
```rust
let checker = checker.map_err(Self::map_storage_err)?;
```
I don't think that it's so bad that it requires a `TODO` line in the code. Feel free to create an issue for this though.

frame/bridge/src/lib.rs
```diff
@@ -136,67 +139,84 @@ impl<T: Trait> Module<T> {
...
 
 		// By encoding the given set we should have an easy way to compare
 		// with the stuff we get out of storage via `read_value`
 		let encoded_validator_set = validator_set.encode();
-		let actual_validator_set = checker
-			.read_value(b":grandpa_authorities")?
-			.ok_or(Error::StorageValueUnavailable)?;
+
+		// TODO: Remove once I can impl From<StorageError> for DispatchError
+		let actual_validator_set = match checker.read_value(b":grandpa_authorities") {
```
tomusdrw: Same here:
```rust
let actual_validator_set = checker.read_value(MOVED_TO_CONSTANT).map_err(Self::map_storage_err)?
let actual_validator_set = actual_validator_set.ok_or(Error::<T>::StorageValueUnavailable))?;
```

frame/bridge/src/lib.rs
```diff
@@ -136,67 +139,84 @@ impl<T: Trait> Module<T> {
...
 
-		parent_hash = header.parent_hash();
-		if *parent_hash == ancestor_hash {
-			return Ok(())
-		}
+		Err(Error::<T>::InvalidAncestryProof.into())
 	}
 
-	Err(Error::InvalidAncestryProof)
+	// TODO: Remove once I can impl From<StorageError> for DispatchError
```
tomusdrw: IMHO the function is fine.
```suggestion
```

frame/bridge/src/lib.rs
```diff
@@ -136,67 +139,84 @@ impl<T: Trait> Module<T> {
...
+	fn map_storage_err(e: StorageError) -> DispatchError {
+		match e {
+			StorageError::StorageRootMismatch => Error::<T>::StorageRootMismatch.into(),
+			StorageError::StorageValueUnavailable => Error::<T>::StorageValueUnavailable.into(),
+		}
```
tomusdrw:
```suggestion
		}.into()
```
You could put `into` here to avoid duplication, but it's a minor thing.

frame/bridge/src/lib.rs
```diff
@@ -260,18 +281,21 @@ mod tests {
...
-		let backend = <InMemory<Blake2Hasher>>::from(vec![
-			(None, b":grandpa_authorities".to_vec(), Some(encoded_set)),
+		let backend = <InMemoryBackend<Blake2Hasher>>::from(vec![
+			(None, vec![(b":grandpa_authorities".to_vec(), Some(encoded_set))]),
 		]);
 		let root = backend.storage_root(std::iter::empty()).0;
 
 		// Generates a storage read proof
-		let proof = prove_read(backend, &[&b":grandpa_authorities"[..]]).unwrap();
+		let proof: StorageProof = prove_read(backend, &[&b":grandpa_authorities"[..]])
```
tomusdrw: same here.

PR #4677, opened 2020-01-20
ci: increase retention for logs of tests to 144 hours

gabreal:
increase test log file retention to 144h

cannot debug https://gitlab.parity.io/parity/substrate/-/jobs/356316/artifacts/file/cc589213_int_failure.log anymore


TriplEight:
Why 144 hours instead of `6 days`? And why 6 when we use 7 for build artifacts?

gabreal:
I changed it to 3 days, just being cautious. Afaics the size of the files is quite small in recent runs.

PR #4680, opened 2020-01-20
Move DummySpecialization to sc-network

tomaka:
Right now it is defined in `sc-network-test`, but I need to use it for tests within `sc-network`. We can avoid a dependency loop by moving it to `sc-network`.

cc'ing @rphmeier because there might be a conflict with #4665


rphmeier:
It'll conflict, but I'll just amend #4665 to remove it from there.

PR #4681, opened 2020-01-20
keep nominations after getting kicked with zero slash

### LINKED ISSUE #4636: Don't require re-activation of nominations when no slash occured
Currently, if a validator offends, then they are booted, possibly slashed and must seek renomination from each of their nominators. This isn't especially reasonable when a validator accidentally gets caught offline (possibly due to faulty software) and no slash actually happens. Instead, in this circumstance, they should be booted but not require renomination.

rphmeier:
fixes #4636 - I was wrong in my issue comment about the migration not being possible, due to prefixed map changes I wasn't aware of.

### DIFF
--- a/frame/staking/src/migration.rs
+++ b/frame/staking/src/migration.rs
@@ -20,12 +20,14 @@
 pub type VersionNumber = u32;
 
 // the current expected version of the storage
-pub const CURRENT_VERSION: VersionNumber = 1;
+pub const CURRENT_VERSION: VersionNumber = 2;
 
+/// The inner logic of migrations.
 #[cfg(any(test, feature = "migrate"))]
-mod inner {
+pub mod inner {
 	use crate::{Store, Module, Trait};
-	use frame_support::{StorageLinkedMap, StorageValue};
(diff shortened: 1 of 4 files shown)


frame/staking/src/migration.rs
```diff
@@ -60,6 +62,55 @@ mod inner {
...
+			prior: Vec<EraIndex>,
+		}
+
+		if *version != 1 { return }
+		*version += 1;
+
+		let prefix = <Module<T> as Store>::SlashingSpans::final_prefix();
+		let mut current_key = prefix.to_vec();
+		loop {
+			let next_key = sp_io::storage::next_key(&current_key[..])
```
gavofyork:
```suggestion
			let maybe_next_key = sp_io::storage::next_key(&current_key[..])
```

frame/staking/src/migration.rs
```diff
@@ -60,6 +62,55 @@ mod inner {
...
+				.filter(|v| v.starts_with(&prefix[..]));
+
+			match next_key {
```
gavofyork:
```suggestion
			match maybe_next_key {
```

PR #4683, opened 2020-01-20
Make debug builds more usable

bkchr:
This pr makes debug builds more usable in terms of `cargo run -- --dev`.

1. `--dev` activates `--execution native`, iff `--execution` is not
given or no sub `--execution-*` is given.
2. It was probably a mistake to compile WASM in debug for a debug build.
So, we now build the WASM binary always as `release` (if not requested
differently by the user). So, we trade compilation time for a better
debug experience.

### DIFF
--- a/bin/node-template/runtime/build.rs
+++ b/bin/node-template/runtime/build.rs
@@ -19,7 +19,7 @@ use wasm_builder_runner::{build_current_project_with_rustflags, WasmBuilderSourc
 fn main() {
 	build_current_project_with_rustflags(
 		"wasm_binary.rs",
-		WasmBuilderSource::Crates("1.0.8"),
+		WasmBuilderSource::Crates("1.0.9"),
 		// This instructs LLD to export __heap_base as a global variable, which is used by the
 		// external memory allocator.
 		"-Clink-arg=--export=__heap_base",
--- a/client/cli/src/lib.rs
+++ b/client/cli/src/lib.rs
@@ -40,6 +40,7 @@ use sc_network::{
 	},
 };
 use sp_core::H256;
+use execution_strategy::*;
 
 use std::{
 	io::{Write, Read, Seek, Cursor, stdin, stdout, ErrorKind}, iter, fmt::Debug, fs::{self, File},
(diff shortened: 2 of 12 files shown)


client/cli/src/lib.rs
```diff
@@ -942,8 +953,14 @@ pub fn fill_import_params<C, G, E>(
...
 
+	let execution = if is_dev {
+		cli.execution_strategies.execution.or(Some(ExecutionStrategy::Native))
+	} else {
+		cli.execution_strategies.execution
+	};
+
 	let exec = &cli.execution_strategies;
-	let exec_all_or = |strat: ExecutionStrategy| exec.execution.unwrap_or(strat).into();
+	let exec_all_or = |strat: ExecutionStrategy| execution.unwrap_or(strat).into();
```
tomusdrw: This will default to `execution` even if `strat` is ovewritten in CLI, I thought you wanted to use `strat` and only default to `execution`, no?

andresilva: Seems like we need to test whether exec was explicitly set or if it's a default as part of `is_dev` being true.

client/cli/src/lib.rs
```diff
@@ -943,13 +955,22 @@ pub fn fill_import_params<C, G, E>(
 	config.wasm_method = cli.wasm_method.into();
 
 	let exec = &cli.execution_strategies;
-	let exec_all_or = |strat: ExecutionStrategy| exec.execution.unwrap_or(strat).into();
+	let exec_all_or = |strat: ExecutionStrategy, default: ExecutionStrategy| {
+		if strat == default && is_dev {
```
tomusdrw: We don't really differentiate between flag being provided at the CLI level or not, we would need to switch to `Options`, but it complicates help and default values, so it's most likely not worthy at all.

It seems to do what it says on the tin now.

We could maybe simplify the code a bit to:
```rust
exec.execution.unwrap_or(if is_dev && strat == default {
  ExecutionStrategy::Native
} else {
  strat
})
```

xlc:
I use this for development: `SKIP_WASM_BUILD= cargo run -- --dev --execution=native -lruntime=debug`. This saves the `--execution=native` part. Can we also have runtime logs enabled so we can see errors from failed extrinsics?

bkchr:
> I use this for development: `SKIP_WASM_BUILD= cargo run -- --dev --execution=native -lruntime=debug`. This saves the `--execution=native` part. Can we also have runtime logs enabled so we can see errors from failed extrinsics?

I'm not sure, I looked at the output and it clearly changes between both. My intention of this pr was to make `cargo run -- --dev` working and preventing to have people being stuck over and over again.
For now I'm against changing the log filter. You can test a lot of stuff with the `--dev` chain, not just transactions and I see people coming and wanting to enable x or y logging. Not sure.

xlc:
The previous behavior was by default printing runtime logs including extrinsic failing messages and it was disabled by some logger changes and people were wondering why they can no longer see the logging messages from runtime. Those logs are not useful for a normal full node but super useful for developments purpose.

bin/node-template/runtime/build.rs
```diff
@@ -19,7 +19,7 @@ use wasm_builder_runner::{build_current_project_with_rustflags, WasmBuilderSourc
 fn main() {
 	build_current_project_with_rustflags(
 		"wasm_binary.rs",
-		WasmBuilderSource::Crates("1.0.8"),
+		WasmBuilderSource::Crates("1.0.9"),
```
xlc: Do we also need to bump this on other substrate node template based projects?

bkchr: When you want to always have release" builds, yes.

bkchr:
> The previous behavior was by default printing runtime logs including extrinsic failing messages and it was disabled by some logger changes and people were wondering why they can no longer see the logging messages from runtime. Those logs are not useful for a normal full node but super useful for developments purpose.

Yeah these logs were disabled because of people complaining about the noise in the output when syncing :) FWIW, I will make sure that we document this.

PR #4684, opened 2020-01-20
client/finality-grandpa/communication: Add doc comment for PeerReport

mxinden:
Applying review suggestion from https://github.com/paritytech/substrate/pull/4661/files#r368046058.

@rphmeier I think the members of the struct are self-explanatory. Let me know if you still want them documented.


rphmeier:
> I think the members of the struct are self-explanatory. Let me know if you still want them documented.

I won't push for it, but I have a preference for documenting all public items because the auto-generated docs look better and it's a good habit to keep.

PR #4685, opened 2020-01-20
contracts: New contract events + unconfusions

Robbepop:
Improves contracts PALLET by

- adds two new events: `Restored` and `Evicted`
    - `Event::Evicted(AccountId, bool)`
        1. `contract: AccountID`: The account ID of the evicted contract.
        1. `tombstone: bool`: True if the evicted contract left behind a tombstone.
    - `Event::Restored(AccountId, AccountId, Hash, Balance, bool)`
        1. `donor: AccountId`: The Account ID of the restoring contract
        1. `dest: AccountId`: The account ID of the restored contract
        1. `code_hash: Hash`: The code hash of the restored contract
        1. `rent_allowance: Balance`: The initial rent allowance of the restored contract
        1. `success: bool`: If restoration was successful
- rename `Event::Contract` to `Event::ContractExecution` so that it is clear that the event was triggered by a contract execution
- improving error message that have previously be shown to confuse users (`"during execution"`)

### DIFF
--- a/frame/contracts/src/exec.rs
+++ b/frame/contracts/src/exec.rs
@@ -782,7 +782,7 @@ where
 	fn deposit_event(&mut self, topics: Vec<T::Hash>, data: Vec<u8>) {
 		self.ctx.deferred.push(DeferredAction::DepositEvent {
 			topics,
-			event: RawEvent::Contract(self.ctx.self_account.clone(), data),
+			event: RawEvent::ContractExecution(self.ctx.self_account.clone(), data),
 		});
 	}
 
--- a/frame/contracts/src/exec.rs
+++ b/frame/contracts/src/exec.rs
@@ -356,7 +356,15 @@ where
 		// Assumption: pay_rent doesn't collide with overlay because
 		// pay_rent will be done on first call and dest contract and balance
 		// cannot be changed before the first call
-		let contract_info = rent::pay_rent::<T>(&dest);
+		let (rent_outcome, contract_info) = rent::pay_rent::<T>(&dest);
+
+		// Deposit an event to indicate contract eviction.
+		if rent_outcome == rent::RentOutcome::Evicted {
+			self.deferred.push(DeferredAction::DepositEvent {
+				event: RawEvent::Evicted(dest.clone()),
+				topics: Vec::new(),
+			});
+		}
 
 		// Calls to dead contracts always fail.
 		if let Some(ContractInfo::Tombstone(_)) = contract_info {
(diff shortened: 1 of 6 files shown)


frame/contracts/src/lib.rs
```diff
@@ -671,6 +671,7 @@ decl_module! {
 			// If poking the contract has lead to eviction of the contract, give out the rewards.
 			if rent::try_evict::<T>(&dest, handicap) == rent::RentOutcome::Evicted {
 				T::Currency::deposit_into_existing(&rewarded, T::SurchargeReward::get())?;
+				Self::deposit_event(RawEvent::Evicted(dest));
```
thiolliere: A contract can be evicted when it tries to pay its rent when it is called. I think this is missing, this should happen in exec.rs when it calls pay_rent.

Robbepop: Okay thank you for telling me. I saw this and wasn't sure if that's actually the case.

frame/contracts/src/wasm/mod.rs
```diff
@@ -1430,7 +1430,7 @@ mod tests {
 				MockExt::default(),
 				&mut gas_meter
 			),
-			Err(ExecError { reason: DispatchError::Other("during execution"), buffer: _ })
+			Err(ExecError { reason: DispatchError::Other("contract trapped during execution"), buffer: _ })
```
pepyakin: Can we wrap this and the line below, since they exceed 100 characters soft-limit?

Robbepop: done

frame/contracts/src/lib.rs
Robbepop: Should be implemented now. Please confirm correctness.

frame/contracts/src/exec.rs
```diff
@@ -356,7 +356,15 @@ where
...
-		let contract_info = rent::pay_rent::<T>(&dest);
+		let (rent_outcome, contract_info) = rent::pay_rent::<T>(&dest);
+
+		// Deposit an event to indicate contract eviction.
+		if rent_outcome == rent::RentOutcome::Evicted {
+			self.deferred.push(DeferredAction::DepositEvent {
+				event: RawEvent::Evicted(dest.clone()),
+				topics: Vec::new(),
+			});
+		}
```
thiolliere: but you can't used deferred field to store the Evicted event because the removal of a contract when it is called happens directly, not in the overlay.

i.e. if contract A call B and B call C and C is evicted and B trap with error then the deferred values are not merged to A.

But what you can simply do is actually make the deposit of the event in pay_rent as it is direct eviction.

Robbepop: Did the requested changes and also noticed that it might be handy if the event would provide information about whether a tombstone has been created or not.

PR #4686, opened 2020-01-20
executor: Migrate wasmtime backend to a high-level API

### LINKED ISSUE #4599: wasmtime implementation violates aliasing requirements
When wasmtime handles a call to host functions it uses a mutable pointer to the contents of `FunctionExecutorState` and linear memory. The problem arises when using the sandbox wasmtime reenters into the host handling routines and obtains the mutable reference to the same items again.

pepyakin:
Fixes #4599

This PR is basically a refactoring that migrates the code from wasmtime-runtime, wasmtime-jit et al to the now stable `wasmtime` crate. The crate is now sprinkled with comments. Hopefully, one now doesn't have to have a black belt in unsafejitsu to deal with this crate. 

Note that this PR relies on some fixes on the wasmtime side, hence the git reference.

This PR should not introduce any significant functional changes and should have the same behavior.

<hr>

TODO:

- [x] Verify that it syncs on Kusama.
- [x] Provide a companion PR for Polkadot
- [x] Update to upstream version of wasmtime


client/executor/wasmtime/src/state_holder.rs
```diff
@@ -0,0 +1,86 @@
...
+	/// During the execution of the provided function `with_context` will be callable.
+	pub fn init_state<R, F>(&self, state: &mut HostState, f: F) -> R
+	where
+		F: FnOnce() -> R,
+	{
+		// We need an `Option` here since Rust will doesn't allow to assign a value for the first
+		// time from within a closure.
+		let mut ret: Option<R> = None;
+
+		take_mut::take(state, |state| {
```
bkchr: Hmm, I think that will not work. It internally catches panics and aborts the process afterwards, but we can not abort the process.

pepyakin: Ah, good catch!

pepyakin:
Ok, it seems that we are having some more fundamential problems here...

client/executor/wasmtime/src/host.rs
```diff
@@ -0,0 +1,356 @@
...
+	/// Constructs a new `HostState`.
+	pub fn new(allocator: FreeingBumpHeapAllocator, instance: InstanceWrapper) -> Self {
+		HostState {
+			sandbox_store: RefCell::new(sandbox::Store::new()),
+			allocator: RefCell::new(allocator),
+			instance,
+		}
+	}
+
+	pub fn into_instance(self) -> InstanceWrapper {
```
tomusdrw: missing docs?

client/executor/wasmtime/src/imports.rs
```diff
@@ -0,0 +1,225 @@
...
+				wasm instance is always called initializing context;
+				therefore host_ctx cannot be None;
+				qed
+				"
+			);
+			// `into_value` panics if it encounters a value that doesn't fit into the values
+			// available in substrate.
+			//
+			// This, however, cannot happen since the signature of this function is created from
+			// a `dyn Function` signature of which cannot have a non substrate value by definition.
```
tomusdrw: What if the values that we support change over time? Would the documentation above still be valid?

client/executor/wasmtime/src/instance_wrapper.rs
```diff
@@ -0,0 +1,245 @@
...
+		self.memory.data_size() as u32
+	}
+
+	/// Reads `__heap_base: i32` global variable and returns it.
+	///
+	/// If it doesn't exist, not a global or of not i32 type returns an error.
+	pub fn extract_heap_base(&self) -> Result<u32> {
+		let heap_base_export = self
+			.instance
+			.get_export("__heap_base")
```
tomusdrw: Should stuff like `__heap_base` rather be declared as some constants? Usually I prefer that over random strings sprinkled all over, at least you can maintain some convention, even if the constant is only used in one place.

client/executor/wasmtime/src/state_holder.rs
```diff
@@ -0,0 +1,86 @@
...
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use crate::host::{HostContext, HostState};
+use std::cell::RefCell;
+use std::rc::Rc;
+
+/// A common place to store a reference to the `HostState`.
+///
+/// This structure is passed into each host function handler and retained in the implementation of
+/// `WasmRuntime`. Whenever, a call into a runtime method is initiated the host state is populated
```
tomusdrw:
```suggestion
/// `WasmRuntime`. Whenever a call into a runtime method is initiated, the host state is populated
```

client/executor/wasmtime/src/instance_wrapper.rs
pepyakin: That would also give a place to leave some docs!

client/executor/wasmtime/src/imports.rs
```diff
@@ -0,0 +1,225 @@
...
+/// Converts a `Val` into a substrate runtime interface `Value`.
+///
+/// Panics if the given value doesn't have a corresponding variant in `Value`.
+fn into_value(val: Val) -> Value {
+	match val {
+		Val::I32(v) => Value::I32(v),
+		Val::I64(v) => Value::I64(v),
+		Val::F32(f_bits) => Value::F32(f_bits),
+		Val::F64(f_bits) => Value::F64(f_bits),
+		_ => panic!(),
```
DemiMarie-parity: Needs a better panic message

client/executor/wasmtime/src/instance_wrapper.rs
```diff
@@ -0,0 +1,245 @@
...
+	table: Option<Table>,
+}
+
+impl InstanceWrapper {
+	/// Wrap the given `Instance`.
+	///
+	/// # Safety
+	///
+	/// This wrapper requires exclusive ownership over the given instance and all its components.
+	/// Since `Instance` is a `Clone` we cannot use typesystem to prevent this, hence unsafe.
```
DemiMarie-parity: We can (and should) use `RefCell` or `Mutex` for this.

client/executor/wasmtime/src/instance_wrapper.rs
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	/// Write data to a slice of memory.
+	///
+	/// Returns an error if the write would go out of the memory bounds.
+	pub fn write_memory_from(&self, address: Pointer<u8>, data: &[u8]) -> Result<()> {
+		unsafe {
+			// This should be safe since we don't grow up memory while caching this reference and
+			// we give up the reference before returning from this function.
+			let memory = self.memory_as_slice_mut();
```
DemiMarie-parity: We need to ensure that `InstanceWrapper` is `!Sync` for this to be safe.

client/executor/wasmtime/src/instance_wrapper.rs
```diff
@@ -0,0 +1,245 @@
...
+	///
+	/// Returns `Err` in case memory cannot be allocated. Refer to the allocator documentation
+	/// to get more details.
+	pub fn allocate(
+		&self,
+		allocator: &mut sc_executor_common::allocator::FreeingBumpHeapAllocator,
+		size: WordSize,
+	) -> Result<Pointer<u8>> {
+		unsafe {
+			// This should be safe since we don't grow up memory while caching this reference and
```
DemiMarie-parity: Ditto

client/executor/wasmtime/src/instance_wrapper.rs
```diff
@@ -0,0 +1,245 @@
...
+	/// Deallocate the memory pointed by the given pointer.
+	///
+	/// Returns `Err` in case the given memory region cannot be deallocated.
+	pub fn deallocate(
+		&self,
+		allocator: &mut sc_executor_common::allocator::FreeingBumpHeapAllocator,
+		ptr: Pointer<u8>,
+	) -> Result<()> {
+		unsafe {
+			// This should be safe since we don't grow up memory while caching this reference and
```
DemiMarie-parity: Ditto

client/executor/wasmtime/src/instance_wrapper.rs
```diff
@@ -0,0 +1,245 @@
...
+		}
+	}
+
+	/// Returns linear memory of the wasm instance as a slice.
+	///
+	/// # Safety
+	///
+	/// Wasmtime doesn't provide comprehensive documentation about the exact behavior of the data
+	/// pointer. If a dynamic style heap is used the base pointer of the heap can change. Since
+	/// growing, we cannot guarantee the lifetime of the returned slice reference.
```
DemiMarie-parity: A better approach would be for this to be a function taking a callback, such that the reference cannot escape the callback.

client/executor/wasmtime/src/instance_wrapper.rs
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	/// Returns linear memory of the wasm instance as a slice.
+	///
+	/// # Safety
+	///
+	/// See `[memory_as_slice]`. In addition to those requirements, since a mutable reference is
+	/// returned it must be ensured that only one mutable and no shared references to memory exists
+	/// at the same time.
+	unsafe fn memory_as_slice_mut(&self) -> &mut [u8] {
```
DemiMarie-parity: Ditto

client/executor/wasmtime/src/runtime.rs
```diff
@@ -87,282 +64,108 @@ pub fn create_instance(
...
-		ActionOutcome::Returned { values } => match values.as_slice() {
-			[RuntimeValue::I64(retval)] => unpack_ptr_and_len(*retval as u64),
-			_ => return Err(Error::InvalidReturn),
-		}
-		ActionOutcome::Trapped { message } => return Err(trap_error.unwrap_or_else(
-			|| format!("Wasm execution trapped: {}", message).into()
-		)),
+		// This is safe since the requirement that `InstanceWrapper` own the instance exclusively
+		// is held.
+		InstanceWrapper::new(instance)?
```
DemiMarie-parity: A better `InstanceWrapper` API would avoid the need for this to be unsafe.

client/executor/wasmtime/src/instance_wrapper.rs
pepyakin: yeah i think it is worth specifying explicitly.

client/executor/wasmtime/src/runtime.rs
pepyakin: Please note that the decision using here `unsafe` (that spans two modules) was completely deliberate. The other option - constructing the instance within the `InstanceWrapper::new` - thus avoiding unsafe here of course was also considered. However, I settled on the present approach to make the invariant more visible (if not screaming). In the other approach there won't be a suitable place to put an unsafe block.

client/executor/wasmtime/src/instance_wrapper.rs
pepyakin: I am not sure if I agree: how would they help from violating the invariant. It would still be possible to clone the memory instance from within this module and create aliases, right?

client/executor/wasmtime/src/imports.rs
pepyakin: I guess no. But I am not sure what would be a better way?

client/executor/wasmtime/src/instance_wrapper.rs
pepyakin: I think it would be a good idea to address this in a follow-up PR!

tomusdrw: Fine with me!

client/executor/wasmtime/src/imports.rs
```diff
@@ -0,0 +1,327 @@
...
+		if at_least > max {
+			return Err(WasmError::Other(format!(
+				"incremented number of pages by heap_pages (total={}) is more than maximum requested\
+				by the runtime wasm module {}",
+				at_least,
+				max,
+			)));
+		}
+	}
+	let limits = Limits::at_least(at_least);
```
bkchr:
```suggestion
	let limits = Limits::new(at_least, at_least);
```

We don't want to have the memory grow above the set heap pages?

client/executor/wasmtime/src/imports.rs
```diff
@@ -0,0 +1,327 @@
...
+			return Err(WasmError::Other(format!(
+				"host doesn't provide any imports from non-env module: {}:{}",
+				import_ty.module(),
+				import_ty.name()
+			)));
+		}
+
+		let resolved = match import_ty.name() {
+			"memory" => {
+				memory_import_index = Some(externs.len());
```
bkchr: Should we error when a second `memory` instance is requested?

client/executor/wasmtime/src/state_holder.rs
```diff
@@ -0,0 +1,77 @@
...
+	pub fn empty() -> StateHolder {
+		StateHolder {
+			state: Rc::new(RefCell::new(None)),
+		}
+	}
+
+	/// Provide `HostState` for the runtime method call and execute the given function `f`.
+	///
+	/// During the execution of the provided function `with_context` will be callable.
+	pub fn init_state<R, F>(&self, state: HostState, f: F) -> (R, HostState)
```
bkchr:
```suggestion
	pub fn init_state_and_execute<R, F>(&self, state: HostState, f: F) -> (R, HostState)
```

Something like this? After the function is executed, the state is not really inited, it is reset again. I think the function name doesn't reflects this.

client/executor/wasmtime/src/imports.rs
pepyakin: I am not sure!

First, it won't be possible atm, since if the module has 2 memory imports then it is a not valid module. Second, as soon as multiple memories are supported then it is OK (although useless) to import the same memory twice.

client/executor/wasmtime/src/state_holder.rs
pepyakin: Hm, then maybe `with_init_state`?

client/executor/wasmtime/src/imports.rs
pepyakin: This is what wasmi does so yeah. Although I wonder if we should reconsider this!

bkchr: So when we use this for validating a parachain, we only want that the validation gets 1GB of memory and can not use more. I think that is something we should enforce/be able to enforce.

client/executor/wasmtime/src/state_holder.rs
bkchr: Yeah sounds good!

PR #4687, opened 2020-01-20
Make dummy crypto app-crypto-wrappable.

tomusdrw:
Expose `dummy` crypto for runtime use as well, and allow it to be wrapped by `app_crypto!` macro.
This makes it possible to use it for tests where we pass mocks.

Mostly useful in cases where we don't really care about particular accounts, and only want some kind of crypto to instantiate properly `system::offchain` extensions (like `CreateTransaction` test).

Marking as a draft, until we figure out if it solves all the issues when writing tests for offchain workers that submit signed transactions.

### DIFF
--- a/primitives/application-crypto/src/lib.rs
+++ b/primitives/application-crypto/src/lib.rs
@@ -37,6 +37,24 @@ pub use sp_std::{ops::Deref, vec::Vec};
 
 pub mod ed25519;
 pub mod sr25519;
+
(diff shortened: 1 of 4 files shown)


bkchr:
Status?

tomusdrw:
Erm :flushed:. I forgot about this completely, didn't hear back from anyone who requested this, let's close for now and revisit if ever needed again.

PR #4688, opened 2020-01-20
Pass an executor through the Configuration

tomaka:
Hopefully fixes the performance issues that Polkadot was having.

One issue that we have in master right now is that all the tasks go through this path:

- https://github.com/paritytech/substrate/blob/a90c72e6d7088d0473670bf74253c65791411589/client/service/src/lib.rs#L334
- https://github.com/paritytech/substrate/blob/a90c72e6d7088d0473670bf74253c65791411589/client/service/src/lib.rs#L339-L341

In other words, everything is single-threaded (except the networking).

In this PR, we now pass a "tasks executor" through the `Configuration` struct.
It is set up to be a tokio runtime, or the `spawn_local` function of `wasm_bindgen_futures`, ~~or a threads pool by default, or local polling if all else fails~~ EDIT: removed fallbacks after review.

**Important**: this executor is stored within the `Service`, and all the components of Substrate should then use the `Service` (through `spawn_task` or `spawn_task_handle` or `spawn_essential_task`) to create tasks.

### DIFF
--- a/client/service/src/lib.rs
+++ b/client/service/src/lib.rs
@@ -95,6 +95,8 @@ pub struct Service<TBl, TCl, TSc, TNetStatus, TNet, TTxPool, TOc> {
 	to_spawn_tx: mpsc::UnboundedSender<Pin<Box<dyn Future<Output = ()> + Send>>>,
 	/// Receiver for futures that must be spawned as background tasks.
 	to_spawn_rx: mpsc::UnboundedReceiver<Pin<Box<dyn Future<Output = ()> + Send>>>,
+	/// How to spawn background tasks.
+	tasks_executor: Option<Box<dyn Fn(Pin<Box<dyn Future<Output = ()> + Send>>) + Send>>,
 	/// List of futures to poll from `poll`.
 	/// If spawning a background task is not possible, we instead push the task into this `Vec`.
 	/// The elements must then be polled manually.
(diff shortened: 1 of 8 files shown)


client/service/src/lib.rs
```diff
@@ -322,16 +324,10 @@ impl<TBl: Unpin, TCl, TSc: Unpin, TNetStatus, TNet, TTxPool, TOc> Future for
...
-			let executor = tokio_executor::DefaultExecutor::current();
-			use futures01::future::Executor;
-			if let Err(err) = executor.execute(task_to_spawn.unit_error().compat()) {
-				debug!(
-					target: "service",
-					"Failed to spawn background task: {:?}; falling back to manual polling",
-					err
-				);
-				this.to_poll.push(Box::pin(err.into_future().compat().map(drop)));
+			if let Some(tasks_executor) = this.tasks_executor.as_ref() {
```
bkchr: Do we still need this logic? The builder creates a thread builder anyway and browser gives us a spawner as well. This sounds like we can remove the option here?

tomaka: If the user leaves `None` in the `Configuration`, we try to create a `ThreadPool` and this is the path that is reached if spawning the threads failed.

In practice, without this code we will get a panic in the browser if the user forgets to tweak the configuration. I guess it's indeed debatable whether it's worth the additional code.

tomaka:
I'm trying of create a version of Polkadot using this branch to see if it indeed fixes the performance issue.
(that's not a blocker for merging though, I think this change is beneficial nonetheless)

client/service/src/lib.rs
bkchr: I think we should go the way of checking at `Service` creation that a `task_executor` is provided and otherwise an error is returned. I know that is not the best solution, but IMHO the best we can do for now.

expenses: Agreed with @bkchr.

tomaka:
We deployed this PR to both a Kusama validator and a Kusama sentry node, and CPU usage seems to be mostly the same if not lower than version 0.7.18 (that includes the revert to libp2p 0.13).

bkchr:
@tomaka service tests needs to be updated.

PR #4633, opened 2020-01-15
Remove more instances of futures01

expenses:
Followup to #4447.

PR #4691, opened 2020-01-20
client/finality-grandpa: Make round_communication use bounded channel

mxinden:
`round_communication` returns a `Sink` and a `Stream` for outgoing and
incoming messages. The messages send into the `Sink` are forwarded down
to the network as well as send back into the `Stream` to ensure the node
processes its own messages.

So far, to send messages into the `Sink` back into the `Stream`, an
unbounded channel was used. This patch updates `round_communication` and
`OutgoingMessages` to use a bounded channel.

This is part of a greater effort to reduce the number of owners of
components within `finality-grandpa` and `network` as well as to reduce
the amount of unbounded channels. For details see d4fbb897c and
f0c18520a.

### DIFF
--- a/client/finality-grandpa/src/communication/mod.rs
+++ b/client/finality-grandpa/src/communication/mod.rs
@@ -29,10 +29,10 @@
 
 use futures::prelude::*;
 use futures03::{
-	sink::Sink as Sink03,
 	channel::mpsc as mpsc03,
 	compat::Compat,
-	future::{Future as Future03},
+	future::{Future as Future03, ready},
+	sink::Sink as Sink03,
 	stream::{Stream as Stream03, StreamExt},
 };
 use log::{debug, trace};
@@ -298,13 +298,13 @@ impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
 				match decoded {
 					Err(ref e) => {
 						debug!(target: "afg", "Skipping malformed message {:?}: {}", notification, e);
-						return futures03::future::ready(None);
+						return ready(None);
 					}
 					Ok(GossipMessage::Vote(msg)) => {
 						// check signature.
 						if !voters.contains_key(&msg.message.id) {
 							debug!(target: "afg", "Skipping message from unknown voter {}", msg.message.id);
-							return futures03::future::ready(None);
+							return ready(None);
 						}
 
 						match &msg.message.message {
@@ -331,11 +331,11 @@ impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
 							},
 						};
 
-						futures03::future::ready(Some(msg.message))
+						ready(Some(msg.message))
 					}
 					_ => {
 						debug!(target: "afg", "Skipping unknown message type");
-						return futures03::future::ready(None);
+						return ready(None);
 					}
 				}
 			});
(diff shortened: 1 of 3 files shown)


mxinden:
> We need to ensure that we don’t create deadlocks.

Correct. The channel being bounded in this patch is used to send packets back to our self. Thus we depend on us reading from the channel.

tomaka:
Let's maybe merge https://github.com/paritytech/substrate/pull/4612 first?

DemiMarie-parity:
@mxinden We need to ensure that we don’t buffer an unbounded amount of data.

andresilva:
I agree with the theory behind this change. In practice though I think this is bound to cause more problems than it aims to solve. In particular because AFAIK going OOM due to these unbounded channels has not been a problem so far. I'd appreciate that this is properly tested before it is rolled out on a live network to avoid surprises (maybe by deploying to flaming fir along with #4612).

mxinden:
> @mxinden We need to ensure that we don’t need to buffer an unbounded amount of data.

@DemiMarie-parity I don't understand your comment. Yes, we need to make sure to not buffer unbounded amount of data. That is what the effort behind these pull requests tries to achieve. In the end you can see an unbounded channel as an unbounded buffer.

mxinden:
I would prefer not to block this pull request on https://github.com/paritytech/substrate/pull/4612 given that the latter is quite large and given that it is unclear when it will be merged.

> so that we don't need to mix futures versions.

We already mix both futures01 and futures03 within `client/finality-grandpa`.

The changeset in this pull request is not large and I am happy to resolve the merge conflicts with https://github.com/paritytech/substrate/pull/4612 (already half done locally) once this one merged.

@tomaka @andresilva @expenses let me know if this would be ok for you.

(I will look into https://github.com/paritytech/substrate/pull/4612 and https://github.com/paritytech/substrate/pull/4633 next and try to help out.)

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -290,22 +291,20 @@ impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
...
-				decoded.ok()
-			})
-			.and_then(move |msg| {
-				match msg {
-					GossipMessage::Vote(msg) => {
+
+				match decoded {
+					Err(ref e) => {
+						debug!(target: "afg", "Skipping malformed message {:?}: {}", notification, e);
+						return ready(None);
```
andresilva: I'd prefer if this read `future::ready` instead. Wasn't immediately obvious to me.

andresilva:
> We already mix both futures01 and futures03 within client/finality-grandpa.

After #4612 we don't, which is what I meant. I'm OK with not blocking this PR though.

client/finality-grandpa/src/communication/mod.rs
rphmeier: I agree. As a rule of thumb, importing free functions into the local namespace hurts readability.

`module::function` lets us know where to look for the function without searching the imports.
`function` lets us know that the function is in the current module.

mxinden: Good point. Fixed with https://github.com/paritytech/substrate/pull/4691/commits/6858319a169babd168871414883bbb969cc8c32d.

mxinden:
Thanks for the reviews and help!

@rphmeier any further comments from your side?

Issue #4663, opened 2020-01-17
CLI's version uses static str which prevents formatting


cecton:
Original comment about the source of the issue: https://github.com/paritytech/cumulus/pull/42#discussion_r367934842

In this particular case we want to add text in the description that shows how to use the command line to pass parameters to the relay chain and the para chain. To do that it would be best if we know the name of the executable.

Unfortunately Version takes static str so it is not possible to format at runtime.

cc @bkchr

bkchr:
The problem I see is that everyone who implements a collator will need to replicate this piece of code. This should not be required.

cecton:
We could/should use https://doc.rust-lang.org/std/env/fn.current_exe.html

But this is not possible because app_dirs requires &'static str.

I believe we can circumvent this issue by using lazy_static.

pscott:
@cecton Has this issue been addressed? Trying to find the initial code, but github is giving me [this error](https://github.com/paritytech/cumulus/pull/42/files/7dfecb1634876a4c132af25a7da4c67d8e85e52e#diff-2485b6feddda62972bc63aed70f55e5f), and I can't find the code in the current repo... 

It looks like it has been solved here:

https://github.com/paritytech/cumulus/blob/05a83e8bdbcf4c216e8632431aa49dd996ce4c3f/test/parachain/src/command.rs#L46-L51

but I'm not sure at all ^^

cecton:
@pscott the stuff that was blocking to make it work is fixed (app_dirs). So it is now possible to convert all those `&'static str` to String instead AND to use `current_exe` instead of having to specify one.

To summarize:
1. If we convert the `&'static str` to String we can customize the description more easily
2. Having to specify the `executable_name` is a constraint introduced by the `app_dirs` dependency and it is less flexible (and more boilerplate) compared to using the `current_exe()`

You can pick the task if you want. I believe it should go without surprise but you can ping me if you have any issue.

PR #4692, opened 2020-01-20
CLI API refactoring and improvement

cecton:
It changes the way we extended the CLI functionalities of substrate to allow more flexibility. (If this was not clear, here is another version: it changes the `sc_cli` API to allow more flexibility).

This touches a few important things:
 - the startup of the async task with tokei:
    This was in node and node-template and I moved it to substrate. The idea is to have 1 time the code that handles unix signals (SIGTERM and SIGINT) properly. It is however possible to make this more generic to wait for a future instead and provide only a helper for the basic handling of SIGTERM and SIGINT.
 - deprecates the use of the CLI without subcommand to run the node, the user should now use the subcommand "run" but the old behavior still work and will display a warning:
    The fact that running the node is not a subcommand makes it difficult with structopt to extend properly. Using a proper subcommand for running the node simplifies the code and can allow a better separation of global arguments vs subcommand arguments
 - increased the version of structopt and tokei
 - no more use of structopt internal's API
 - less use of generics

Impacts to #4663: that ticket shouldn't be started before this one is finished. I already know the code so I know what will (not) work for the implementation. It's more complicated than it looks.

Related to #4643 and https://github.com/paritytech/cumulus/pull/42: the implementation of "into_configuration" and "get_config" are similar but with better flexibility so it is now possible in cumulus to have the command-line arguments only of the run command for polkadot if we want

Related to https://github.com/paritytech/cumulus/issues/24 and https://github.com/paritytech/cumulus/issues/34 : it will now be possible to make a configuration struct for polkadot with some overrides of the default parameters much more easily.

> - [ ] Your PR adheres [the style guide](https://github.com/paritytech/polkadot/wiki/Style-Guide)

There is a bit of clean-up to do but the formatting should be correct. The clean-up is more in the name of the generics...

> - [ ] You bumped the runtime version if there are breaking changes in the **runtime**.

I'm not sure. Can the reviewer help me on that and check?

> - [ ] You updated any rustdocs which may have changed

I still need to add doc to the new exposed functions.

> - [ ] Has the PR altered the external API or interfaces used by Polkadot? Do you have the corresponding Polkadot PR ready?

Definitely alternates the CLI API. I updated node and node-template here but cumulus and polkadot will need to be updated too.


bkchr:
>deprecates the use of the CLI without subcommand to run the node, the user should now use the subcommand "run" but the old behavior still work and will display a warning:
The fact that running the node is not a subcommand makes it difficult with structopt to extend properly. Using a proper subcommand for running the node simplifies the code and can allow a better separation of global arguments vs subcommand arguments

This is intended behavior and was discussed already before. If the user does not specifies any `subcommand`, it is automatically `run`. And `run` should not be accepted as valid sub command.

client/cli/src/lib.rs
```diff
@@ -177,461 +179,393 @@ fn is_node_name_valid(_name: &str) -> Result<(), &str> {
...
+	new_light: F2,
+	new_full: F3,
+	spec_factory: F,
+	builder: F4,
+	impl_name: &'static str,
+	version: &VersionInfo,
+) -> error::Result<()>
+where
+	F: FnOnce(&str) -> Result<Option<ChainSpec<G, E>>, String>,
+	F2: FnOnce(Configuration<G, E>) -> Result<T1, sc_service::error::Error>,
```
tomaka: I'm really not a fan of this callback style where we're hard-coding the fact that `new_full` and `new_light` take a configuration and return a service.

The reason for not using callbacks originally is so that users can straight up do a function call, which to me is much more intuitive than having to pass a closure that is then called by something else.

When later we want to pass another parameter to `new_full`, what are we going to do? Pass this parameter to `run` so that it is then passed to the callback?

Yes this PR reduces code duplication, but IMO it's not a good thing here.

tomaka:
I believe this change is incompatible with #4688, as we have to create the tokio runtime and modify the `Configuration` before creating the `Service`.

client/cli/src/lib.rs
cecton: But... but... ☝️  It was already a closure/callback before. I didn't change that ❓ what do you mean❓  https://github.com/paritytech/substrate/blob/master/bin/node/cli/src/cli.rs#L101

cecton:
> I believe this change is incompatible with #4688, as we have to create the tokio runtime and modify the `Configuration` before creating the `Service`.

I think you're right. It's also useful for cumulus because we have to start 2 nodes so we need a better control over the runtime. I will try to improve this.

client/cli/src/lib.rs
```diff
@@ -177,461 +179,393 @@ fn is_node_name_valid(_name: &str) -> Result<(), &str> {
...
+	F3: FnOnce(Configuration<G, E>) -> Result<T2, sc_service::error::Error>,
+	F4: FnOnce(Configuration<G, E>) -> Result<T3, sc_service::error::Error>,
+	G: RuntimeGenesis,
+	E: ChainSpecExtension,
+	T1: AbstractService + std::marker::Unpin,
+	T2: AbstractService + std::marker::Unpin,
+	T3: ServiceBuilderCommand<Block = T5> + std::marker::Unpin,
+	T5: sp_runtime::traits::Block + Debug,
+	<<<T5 as sp_runtime::traits::Block>::Header as sp_runtime::traits::Header>::Number as
```
expenses: `<<<<T5 as BlockT>::Header as HeaderT>::Number as std::str::FromStr>::Error: std::fmt::Debug` is probably cleaner

client/cli/src/lib.rs
```diff
@@ -177,461 +179,393 @@ fn is_node_name_valid(_name: &str) -> Result<(), &str> {
...
+	new_full: F3,
+	version: &VersionInfo,
+) -> error::Result<()>
+where
+	F2: FnOnce(Configuration<G, E>) -> Result<T1, sc_service::error::Error>,
+	F3: FnOnce(Configuration<G, E>) -> Result<T2, sc_service::error::Error>,
+	G: RuntimeGenesis,
+	E: ChainSpecExtension,
+	T1: AbstractService + std::marker::Unpin,
+	T2: AbstractService + std::marker::Unpin,
```
expenses: `Unpin` is imported by default

client/cli/Cargo.toml
```diff
@@ -16,7 +16,7 @@ time = "0.1.42"
 ansi_term = "0.12.1"
 lazy_static = "1.4.0"
 app_dirs = "1.2.1"
-tokio = "0.2.1"
+tokio = { version = "0.2.9", features = [ "full" ] }
```
expenses: You should probably only import the features you need.

client/cli/src/lib.rs
```diff
@@ -177,461 +179,393 @@ fn is_node_name_valid(_name: &str) -> Result<(), &str> {
...
-	/// Command to check a block.
-	CheckBlock(CheckBlock<'a>),
-	/// Command ready to purge the chain.
-	PurgeChain(ParseAndPreparePurge<'a>),
-	/// Command ready to revert the chain.
-	RevertChain(ParseAndPrepareRevert<'a>),
-	/// An additional custom command passed to `parse_and_prepare`.
-	CustomCommand(CC),
-}
+pub fn run<F, G, E, F2, F3, F4, T1, T2, T3, T5>(
```
expenses: You make these traits a little more descriptive, like the way they are in `Service`.

client/cli/src/lib.rs
tomaka: Oh, just because it's in master doesn't mean I like it :sweat_smile: 
Of course that's just my opinion, but as the one who has done the previous refactoring here, I was trying to remove these callbacks-like structure. The one you linked to was hard to remove at the time, for a reason I can't recall.

client/cli/src/lib.rs
```diff
@@ -177,461 +171,397 @@ fn is_node_name_valid(_name: &str) -> Result<(), &str> {
...
+	info!("Roles: {}", display_role(&config));
+	match config.roles {
+		ServiceRoles::LIGHT => run_service_until_exit(
+			new_light(config)?,
+		),
+		_ => run_service_until_exit(
+			new_full(config)?,
+		),
+	}
 }
```
cecton: > > I believe this change is incompatible with #4688, as we have to create the tokio runtime and modify the `Configuration` before creating the `Service`.
> 
> I think you're right. It's also useful for cumulus because we have to start 2 nodes so we need a better control over the runtime. I will try to improve this.

@tomaka I'm making a thread here in the code to make it easier to track on GitHub.

This is what I had in mind to solve this problem: offer the user to match the subcommand with RunCmd so you can add a match branch for your custom implementation and let it fallback to the default for the rest.

Here is an example (based on bin/node-template/cli):
```rust
	match &core_params {
		CoreParams::RunCmd(cmd_params) => {
			let config = get_config(....);
			total_control_on_how_you_run_your_node();
		},
		params => sc_cli::run(
			params,
			service::new_light,
			service::new_full,
			load_spec,
			|config: Config<_, _>| Ok(new_full_start!(config).0),
			"substrate-node",
			&version,
		},
	}
```

So for RunCmd it will run your own tokio runtime while for the other commands it will use the defaults.

client/cli/src/lib.rs
cecton: haha :grin: ok! Actually I just tried to remove the callback and use a trait instead (so you can override functions in the trait) but it's actually very complicated because of the return types: `new_full` and `new_start` return 2 different things in bin/node: one is Service and the other is impl AbstractService...

I couldn't find a solution to that but in any case: I'm trying also to reduce the complexity of the signatures.

cecton: > If later we want to pass another parameter to new_full, what are we going to do? Pass this parameter to run so that it is then passed to the callback?

https://github.com/paritytech/substrate/pull/4692#discussion_r368922595

client/cli/src/lib.rs
cecton: Yes :grin: this was temporary. Check again, I removed the numbers

client/cli/src/lib.rs
tomaka: The problem to me, is: this `RunCmd` situation is one example of one of the issues that could arise from hiding the behaviour inside of that `sc_cli::run` function.

The reason why I'm not a fan of this specific change is that we're locking ourselves inside of that specific use case, and if we ever need more control then we will start piling up hacks. From my experience, having control over what happens taken away from users is exactly how a source code starts having hacks.

cecton: Do you call this a hack?
```rust
match args {
    Variant1 => process1(),
    Variant2 => process2(),
    other => catch_all(),
}
```

I see `sc_cli::run` as a helper function for a catch-all argument of a match arm. If we don't do that, every CLI must implements every variant of substrate even if they don't give a care.

I think if you compare this:
```rust
	match subcommand {
		Cli::SubstrateCli(cli) => sc_cli::run(
			cli,
			service::new_light,
			service::new_full,
			load_spec,
			|config: Config<_, _>| Ok(new_full_start!(config).0),
			"substrate-node",
			&version,
		),
		Cli::Factory(cli_args) => {
			...
		},
	}
```
vs that:
```rust
	match parse_and_prepare::<CustomSubcommands, NoCustom, _>(&version, "substrate-node", args) {
		ParseAndPrepare::Run(cmd) => cmd.run(load_spec, exit,
		|exit, _cli_args, _custom_args, config: Config<_, _>| {
			info!("{}", version.name);
			info!("  version {}", config.full_version());
			info!("  by Parity Technologies, 2017-2019");
			info!("Chain specification: {}", config.chain_spec.name());
			info!("Node name: {}", config.name);
			info!("Roles: {}", display_role(&config));
			let runtime = RuntimeBuilder::new().name_prefix("main-tokio-").build()
				.map_err(|e| format!("{:?}", e))?;
			match config.roles {
				ServiceRoles::LIGHT => run_until_exit(
					runtime,
					service::new_light(config)?,
					exit
				),
				_ => run_until_exit(
					runtime,
					service::new_full(config)?,
					exit
				),
			}
		}),
		ParseAndPrepare::BuildSpec(cmd) => cmd.run::<NoCustom, _, _, _>(load_spec),
		ParseAndPrepare::ExportBlocks(cmd) => cmd.run_with_builder(|config: Config<_, _>|
			Ok(new_full_start!(config).0), load_spec, exit),
		ParseAndPrepare::ImportBlocks(cmd) => cmd.run_with_builder(|config: Config<_, _>|
			Ok(new_full_start!(config).0), load_spec, exit),
		ParseAndPrepare::CheckBlock(cmd) => cmd.run_with_builder(|config: Config<_, _>|
			Ok(new_full_start!(config).0), load_spec, exit),
		ParseAndPrepare::PurgeChain(cmd) => cmd.run(load_spec),
		ParseAndPrepare::RevertChain(cmd) => cmd.run_with_builder(|config: Config<_, _>|
			Ok(new_full_start!(config).0), load_spec),
		ParseAndPrepare::CustomCommand(CustomSubcommands::Factory(cli_args)) => {
			...
		}
	}
```

You can clearly see it became easier to use substrate. I'm skipping the types but you can see the number of generics needed for the custom command, this is also a big part of the benefit. I'm also skipping the IntoExit that must be implemented but here it is provided automatically if you don't need something special.

And most of the steps inside `sc_cli::run` can be imported separately if needed (the signal handling and `get_config`)

bkchr: I think we should go use some sort of mixed solution here.

- Create a new crate "opinionated-cli" (Find a better name :D)
- This new crates does all these repetitive stuff like `match parse()` blah.
- Polkadot/Substrate node can use this.
- If someone wants to go the extra mile, they can just use `sc-cli` and do it by hand. 

TLDR: Keep the interface of `sc-cli` as generic as possible, but move all the repetitive/common stuff into own crate.

tomaka: Your code is not a hack. I'm saying that your code will lead us to being locked in a specific use case, which may cause problems in the future if something doesn't fit exactly in one of the three provided ways to create a service.

You're also comparing apples vs oranges. I'm obviously in favour of removing `IntoExit` and all. My grief is only about this `run` function.

I know it's a bad argumentation, but I'm mostly drawing this out of experience because this `run` function looks like the kind of code that usually causes the most trouble a couple months later.

cecton: > TLDR: Keep the interface of sc-cli as generic as possible, but move all the repetitive/common stuff into own crate.

That's fine by me :woman_shrugging: 

> I'm saying that your code will lead us to being locked in a specific use case, which may cause problems in the future if something doesn't fit exactly in one of the three provided ways to create a service.

I trust your experience on this don't worry. I'm just trying to understand and find solutions. I guess I don't 100% understand your point yet. This PR fixes most of the issues/limitations I encountered when I worked on cumulus CLI (as a user of substrate), it's really an attempt to make it also more flexible than what it is right now. That is why I'm exposing RunCmd now (which wasn't exposed before).

Does the solution of @bkchr sound good? I don't think it would be bad to keep client/cli as basic/simple as possible and provide opinionated helpers in another module (which would provide signal handling and running most of the commands). (Or to its own crate)

What do you think?

bin/node-template/src/cli.rs
```diff
@@ -1,58 +1,49 @@
...
-
-	Ok(())
+	type Config<A, B> = Configuration<A, B>;
+
+	let args: Vec<_> = args.collect();
+	let opt = sc_cli::from_iter::<Cli, _>(args.clone(), &version);
+	let subcommand = opt.subcommand.unwrap_or_else(|| {
+		let opt = sc_cli::from_iter::<RunCmd, _>(args, &version);
+
+		eprintln!(
```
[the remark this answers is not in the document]
cecton: Is it like the tab vs space debate or do I have a word to say? :confounded:

bin/node-template/Cargo.toml
```diff
@@ -36,6 +35,7 @@ sc-client = { version = "2.0.0", path = "../../client/" }
 node-template-runtime = { version = "2.0.0", path = "runtime" }
 sp-runtime = { version = "2.0.0", path = "../../primitives/runtime" }
 sc-basic-authority = { path  = "../../client/basic-authorship" }
+structopt = "=0.3.8"
```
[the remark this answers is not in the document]
cecton: Very good point :grin: 

I was thinking: don't you think it would be best that the users use structopt from a public export in `sc_cli` like structopt does for clap?

I mean:
```rust
pub structopt::StructOpt;
```

client/cli/src/lib.rs
bkchr: For me this sounds okay and as said before, I would move the opinionated into its own crate. To have a better distinction.

bin/node-template/Cargo.toml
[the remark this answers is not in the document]
bkchr: Yeah we can also re-export, that is up to you :)

client/cli/src/lib.rs
tomaka: I like the solution for the fact that we're giving the choice.

> Create a new crate "opinionated-cli" (Find a better name :D)

But why move the opinionated stuff into a separate crate? Why not put it in `sc-cli`?

I think it's great if `sc-cli` exposes both `parse_and_execute` and `run`, and you can choose which API you want.

bkchr: Yeah or that way, I thought that it would make the distinction more clear.

tomaka: Discarding my concerns after discussing in DM.
The concern is valid, but it's not as bad as I thought and there's no point in trying to have a stable API of sc-cli yet.

arkpar:
Looks good, except for deprecating default command.

client/cli/src/lib.rs
cecton: I added 2 important things:

1. A change on the way Configuration is created. It can be now created with `Configuration::default` which can then be passed in argument to the run functions. The idea is to solve the problem https://github.com/paritytech/cumulus/issues/24 by allowing the user to change the base configuration **before** the command line arguments gets loaded.

Turns out that after doing that, a lot of constraints in the code disappeared and I was able to remove a lot of line of codes again.

The only new constraint added here is that `chain_spec` on Configuration is now an `Option`.

2. I moved the different command actions to their own function again like it was with `ParseAndPrepare`. The only difference is that I use the structopt directly (which is simpler and allows the user to do the exact same thing than ParseAndPrepare before).

client/cli/src/params.rs
```diff
@@ -578,74 +594,12 @@ pub struct RunCmd {
...
-}
-
 /// Wrapper for exposing the keyring test accounts into the Cli.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, StructOpt)]
 pub struct Keyring {
 	pub account: Option<sp_keyring::Sr25519Keyring>,
 }
 
-impl StructOpt for Keyring {
```
bkchr: Hmm, this is another change I don't really want to see. `--alice` is much shorter than `--account alice`

client/cli/src/params.rs
```diff
@@ -578,74 +594,12 @@ pub struct RunCmd {
 	pub password_filename: Option<PathBuf>
 }
 
-/// Stores all required Cli values for a keyring test account.
-struct KeyringTestAccountCliValues {
```
bkchr: You are removing quite a lot of functionality without even delivering an alternative here.

client/cli/src/params.rs
cecton: Oops, that was a mistake. I will change that

client/cli/src/params.rs
cecton: This is just for the `--alice` or is there something else with this? This struct is private.

bkchr: This is used by `--alice`, `--bob`, etc. These `conflicts_with` are also important.

cecton:
@bkchr  @expenses @tomaka I have updated the branch (pulled the changes one by one to be sure they were all correctly implemented as they were intended, especially #4688). I'm fixing the various issues in the CI now. Can you all check that everything is fine in the code and validate your review? :pray:  Thanks!!

cecton:
(let me click on the squash-merge button and modify the message so I can remove all the wips :grin: )

cecton:
ping @bkchr @expenses @tomaka

expenses:
LGTM, but I don't know the cli stuff that well.

cecton:
> LGTM, but I don't know the cli stuff that well.

Np! Can you removed the "request changes" then? I will wait for the feedback of the others

bin/node/cli/Cargo.toml
```diff
@@ -30,7 +30,7 @@ hex-literal = "0.2.1"
 jsonrpc-core = "14.0.3"
 log = "0.4.8"
 rand = "0.7.2"
-structopt = "=0.3.7"
+structopt = "0.3.8"
```
tomaka: This was pinned because of https://github.com/paritytech/substrate/pull/4509
I don't know anything about that issue, but did we stop depending on structopt's internals?

client/cli/src/lib.rs
```diff
@@ -25,47 +25,47 @@ mod params;
...
+		NetworkConfiguration, TransportConfig, NonReservedPeerMode,
 	},
 };
-use sp_core::H256;
-use execution_strategy::*;
 
 use std::{
-	io::{Write, Read, Seek, Cursor, stdin, stdout, ErrorKind}, iter, fmt::Debug, fs::{self, File},
-	net::{Ipv4Addr, SocketAddr}, path::{Path, PathBuf}, str::FromStr, pin::Pin, task::Poll
+	io::{Write}, iter, fmt::Debug, fs,
```
tomaka:
```suggestion
	io::Write, iter, fmt::Debug, fs,
```

client/cli/src/lib.rs
```diff
@@ -161,637 +127,200 @@ fn base_path(
...
-///
-/// `RP` are custom parameters for the run command. This needs to be a `struct`! The custom
-/// parameters are visible to the user as if they were normal run command parameters. If no custom
-/// parameters are required, `NoCustom` can be used as type here.
-pub fn parse_and_prepare<'a, CC, RP, I>(
-	version: &'a VersionInfo,
-	impl_name: &'static str,
-	args: I,
-) -> ParseAndPrepare<'a, CC, RP>
+/// Gets the struct from any iterator such as a `Vec` of your making.
```
tomaka: Needs better documentation. It doesn't even mention that it's for parsing command-line arguments.

client/cli/src/lib.rs
```diff
@@ -161,637 +127,200 @@ fn base_path(
...
-		let start = std::time::Instant::now();
-		let check = builder(config)?
-			.check_block(block_id);
-		let mut runtime = tokio::runtime::Runtime::new().unwrap();
-		runtime.block_on(check)?;
-		println!("Completed in {} ms.", start.elapsed().as_millis());
-		Ok(())
-	}
-}
+/// A helper function that:
```
tomaka:
```suggestion
/// A helper function that initializes and runs any of the command variants of `CoreParams`.
```

client/cli/src/lib.rs
```diff
@@ -161,637 +127,200 @@ fn base_path(
...
+/// Initialize substrate and its configuration
+///
+/// This method:
+///
+/// 1.  set the panic handler
+/// 2.  raise the FD limit
+/// 3.  initialize the logger
+/// 4.  update the configuration provided with the chain specification, config directory,
+///     information (version, commit), database's path, boot nodes and telemetry endpoints
+pub fn init<G, E, F>(
```
tomaka: I'm not a fan of this API as it's way too error-prone, but I guess we can improve it later.
As an example of how it's error-prone, I have absolutely no idea whether you forgot to set a field of the configuration in this function except by carefully reviewing. If a new field is added, it is totally possible that one forgets to update this function.

client/cli/src/lib.rs
```diff
@@ -161,637 +127,200 @@ fn base_path(
...
+	Ok(())
 }
 
-/// Create a `NodeKeyConfig` from the given `NodeKeyParams` in the context
-/// of an optional network config storage directory.
-fn node_key_config<P>(params: NodeKeyParams, net_config_dir: &Option<P>)
-	-> error::Result<NodeKeyConfig>
+/// Run the node
+///
+/// Run the light node if the role is "light", otherwise run the full node.
```
tomaka:
```suggestion
/// Builds and runs either a full or a light node, depending on the `role` within the `Configuration`.
```

client/cli/src/lib.rs
```diff
@@ -869,18 +398,18 @@ fn input_keystore_password() -> Result<String, String> {
...
 
 /// Use in memory keystore config when it is not required at all.
-fn fill_config_keystore_in_memory<C, G, E>(config: &mut sc_service::Configuration<C, G, E>)
+fn fill_config_keystore_in_memory<G, E>(config: &mut sc_service::Configuration<G, E>)
 	-> Result<(), String>
 {
 	match &mut config.keystore {
-		cfg @ KeystoreConfig::None => { *cfg = KeystoreConfig::InMemory; Ok(()) },
-		_ => Err("Keystore config specified when it should not be!".into()),
+		 cfg @ KeystoreConfig::None => { *cfg = KeystoreConfig::InMemory; Ok(()) },
```
tomaka:
```suggestion
		cfg @ KeystoreConfig::None => { *cfg = KeystoreConfig::InMemory; Ok(()) },
```

client/cli/src/lib.rs
```diff
@@ -869,18 +398,18 @@ fn input_keystore_password() -> Result<String, String> {
...
+		 _ => Err("Keystore config specified when it should not be!".into()),
```
tomaka:
```suggestion
		_ => Err("Keystore config specified when it should not be!".into()),
```

client/cli/src/node_key.rs
```diff
@@ -0,0 +1,225 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use sc_network::{
```
tomaka:
```suggestion
use sc_network::config::NodeKeyConfig;
```

The `self` isn't necessary.

client/cli/src/node_key.rs
```diff
@@ -0,0 +1,225 @@
...
+
+/// The maximum number of characters for a node name.
+const NODE_NAME_MAX_LENGTH: usize = 32;
+
+/// The file name of the node's Ed25519 secret key inside the chain-specific
+/// network config directory, if neither `--node-key` nor `--node-key-file`
+/// is specified in combination with `--node-key-type=ed25519`.
+const NODE_KEY_ED25519_FILE: &str = "secret_ed25519";
+
+pub fn generate_node_name() -> String {
```
tomaka: What does this function do in this module? It seems out of place.

client/cli/src/runtime.rs
```diff
@@ -0,0 +1,127 @@
...
+use futures::{Future, future, future::FutureExt};
+use futures::select;
+use futures::pin_mut;
+use sc_service::{AbstractService, Configuration};
+use crate::error;
+use crate::informant;
+
+async fn main<F, E>(func: F) -> Result<(), Box<dyn std::error::Error>>
+where
+	F: Future<Output = Result<(), E>> + future::FusedFuture + Unpin,
```
tomaka:
```suggestion
	F: Future<Output = Result<(), E>> + future::FusedFuture,
```

client/cli/src/runtime.rs
```diff
@@ -0,0 +1,127 @@
...
+	use tokio::signal::unix::{signal, SignalKind};
+
+	let mut stream_int = signal(SignalKind::interrupt())?;
+	let mut stream_term = signal(SignalKind::terminate())?;
+
+	let t1 = stream_int.recv().fuse();
+	let t2 = stream_term.recv().fuse();
+	let mut t3 = func;
+
+	pin_mut!(t1, t2);
```
tomaka:
```suggestion
	pin_mut!(t1, t2, t3);
```

client/cli/src/runtime.rs
```diff
@@ -0,0 +1,127 @@
...
+
+	Ok(())
+}
+
+async fn run_service_with_informant<T>(service: T) -> error::Result<()>
+where
+	T: AbstractService + Unpin,
+{
+	let informant_future = informant::build(&service);
+	let handle = tokio::spawn(informant_future);
```
tomaka: This is okay-ish because this entire module uses tokio. Ideally I'd prefer to pass an abstraction over an executor by parameter to this function.

client/service/src/config.rs
```diff
@@ -146,19 +144,16 @@ pub enum DatabaseConfig {
 	Custom(Arc<dyn KeyValueDB>),
 }
 
-impl<C, G, E> Configuration<C, G, E> where
-	C: Default,
-	G: RuntimeGenesis,
-	E: Extension,
+impl<G, E> Default for Configuration<G, E>
```
tomaka:
```suggestion
impl<G, E> Default for Configuration<G, E> {
```

bin/node/cli/Cargo.toml
cecton: Yes it was because we were using internals of structopt. Part of my refactoring goal was also to remove that need and use only the public API so it can be upgraded easily.

client/cli/src/lib.rs
cecton: Will improve that. I copy-pasted it from structopt's doc

client/cli/src/lib.rs
cecton: Most defaults actually come from `Configuration::default()` but the problem is that some parameters need to be present to run a node (like `chain_spec` but others too) and that should fail with an assert somehow.

If a new field is added, one will have to update `Configuration::default()` with a default value. But if that value can only be provided after parsing, then it falls into the same problem of `chain_spec` and the other parameters that are Option while they really are important for running.

I don't see how to improve this at the moment.

client/cli/src/node_key.rs
cecton: True! Will move it back! I tried to split a bit because `lib.rs` was too dense

client/cli/src/runtime.rs
cecton: Actually you can set the `task_executor` yourself and don't use `run_until_exit` at all. That's what is happening with WASM.

tomaka: What I mean is that `run_service_with_informant` is here an implementation detail of `run_service_until_exit`, and `run_service_until_exit` uses tokio, so it's ok that `run_service_with_informant` relies on tokio being there.

However, if someone is to mark `run_service_with_informant` as `pub`, then we would have to pass an executor as parameter.

My reasoning is basically: we could do either this now in order to avoid future accidents, or not care.

cecton: I'm fixing that now but I just realized `tasks_executor` is not proper English. Correct me if I'm wrong but I think it's `task_executor`. Maybe I should fix that now so we're done with it.

client/cli/src/runtime.rs
```diff
@@ -86,7 +86,7 @@ where
 	T: AbstractService + Unpin,
 {
 	let informant_future = informant::build(&service);
-	let handle = async { informant_future.await };
+	async { informant_future.await };
```
tomaka: That should be just:
```suggestion
	informant_future.await;
```

cecton: Maybe the easiest would be to simply use a `join!`

cecton: I've removed `run_service_with_informant` and just put the code in `run_service_until_exit`

cecton:
> [...] or that there's a state mismatch in the configuration (e.g. sentry=true and role=light).

That can be done at `structopt` level since it's use input. Structopt and clap are really good for this and it will prevent de program to run with inconsistencies.

```
	/// Enable sentry mode.
	///
	/// The node will be started with the authority role and participate in
	/// consensus tasks as an "observer", it will never actively participate
	/// regardless of whether it could (e.g. keys are available locally). This
	/// mode is useful as a secure proxy for validators (which would run
	/// detached from the network), since we want this node to participate in
	/// the full consensus protocols in order to have all needed consensus data
	/// available to relay to private nodes.
	#[structopt(
		long = "sentry",
		conflicts_with_all = &[ "validator" ]
	)]
	pub sentry: bool,

	/// Experimental: Run in light client mode.
	#[structopt(long = "light")]
	pub light: bool,
```

So those 2 flags should conflict, right? I can just add a `conflict_with` here.

tomaka:
> That can be done at structopt level since it's use input. Structopt and clap are really good for this and it will prevent de program to run with inconsistencies.

Even better if it's done in the type system for the `Configuration` type, so that code that takes a `Configuration` as input don't have to awkwardly do something like: 

```rust
if config.light && config.sentry {
    panic!("Some code at the other side of the code base guarantees (or hopefully still does) that this cannot happen");
}
```

cecton:
> Even better if it's done in the type system for the `Configuration` type, so that code that takes a `Configuration` as input don't have to awkwardly do something like:
> 
> ```rust
> if config.light && config.sentry {
>     panic!("Some code at the other side of the code base guarantees (or hopefully still does) that this cannot happen");
> }
> ```

You need the 2 validations.
1. a user validation with a user friendly message (that's clap)
2. a hard validation (ideally with types so it doesn't compile) later on

That's why I don't like much having so many `Option` in the Configuration but right now I think it would be too much to change that. If I help on ServiceBuilder I will probably change that though. Maybe with some asserts like you proposed (ideally with types).

cecton:
I have managed to get a better implementation for the run command not being a subcommand. It's now working as you would expect with structopt (with `#[structopt(subcommand)]` and `subcommand: Option<Subcommand>`. It also looks better in the bin/node and node-template. I already prepared a polkadot branch: https://github.com/paritytech/polkadot/pull/805 (but I will need to update it again).

client/cli/src/params.rs
```diff
@@ -530,8 +544,36 @@ pub struct RunCmd {
 	pub pool_config: TransactionPoolParams,
 
 	#[allow(missing_docs)]
```
bkchr: There was documentation for each of these flags before. Please readd.

client/cli/src/params.rs
```diff
@@ -881,112 +869,324 @@ pub enum CoreParams<CC, RP> {
...
+	where
+		B: FnOnce(Configuration<G, E>) -> Result<BC, sc_service::error::Error>,
+		G: RuntimeGenesis,
+		E: ChainSpecExtension,
+		BC: ServiceBuilderCommand<Block = BB> + Unpin,
+		BB: sp_runtime::traits::Block + Debug,
+		<<<BB as BlockT>::Header as HeaderT>::Number as std::str::FromStr>::Err: std::fmt::Debug,
+		<BB as BlockT>::Hash: std::str::FromStr,
+	{
+		assert!(config.chain_spec.is_some(), "chain_spec must be present before continuing");
```
bkchr: Again some `assert!` because of the `chain_spec` change in `Configuration`.

client/cli/src/lib.rs
```diff
@@ -161,637 +144,227 @@ fn base_path(
...
-			Pin::new(&mut import_fut).poll(cx)
-		});
-
-		let mut runtime = tokio::runtime::Runtime::new().unwrap();
-		runtime.block_on(fut)?;
-		Ok(())
-	}
-}
+/// A helper function that initializes and runs the node
+pub fn run<F, G, E, FNL, FNF, SL, SF>(
```
bkchr: Why do we have a function that calls the `run` method of `RunCmd`. Not sure why we not call it directly.

client/cli/src/lib.rs
```diff
@@ -161,637 +144,227 @@ fn base_path(
...
-		NodeKeyType::Ed25519 =>
-			params.node_key.as_ref().map(parse_ed25519_secret).unwrap_or_else(||
-				Ok(params.node_key_file
-					.or_else(|| net_config_file(net_config_dir, NODE_KEY_ED25519_FILE))
-					.map(sc_network::config::Secret::File)
-					.unwrap_or(sc_network::config::Secret::New)))
-				.map(NodeKeyConfig::Ed25519)
+	info!("{}", version.name);
+	info!("  version {}", config.full_version());
+	info!("  by {}, 2017-2020", version.author);
```
bkchr: When a project started in 2020, this wouldn't be correct :P 

And what happens in 2021?

bin/node/cli/src/command.rs
```diff
@@ -0,0 +1,67 @@
+use sc_cli::{VersionInfo, error};
```
bkchr: Copyright header!

bin/node/cli/src/command.rs
```diff
@@ -0,0 +1,67 @@
...
+use sc_service::{Roles as ServiceRoles};
+use node_transaction_factory::RuntimeAdapter;
+use crate::{Cli, service, ChainSpec, load_spec, Subcommand};
```
bkchr: The line below could be included as well.

bin/node-template/src/command.rs
```diff
@@ -0,0 +1,32 @@
...
+use crate::chain_spec;
+use crate::cli::Cli;
+
+/// Parse and run command line arguments
+pub fn run(version: VersionInfo) -> error::Result<()>
+{
+	let opt = sc_cli::from_args::<Cli>(&version);
+
+	let mut config = sc_service::Configuration::default();
+	config.impl_name = "node-template";
```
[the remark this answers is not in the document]
cecton: There is a default `impl_name`

client/service/src/config.rs
```diff
@@ -216,11 +206,20 @@ impl<C, G, E> Configuration<C, G, E> {
...
 			path
 		})
 	}
+
+	/// Return a reference to the `ChainSpec` of this `Configuration`.
+	///
+	/// ### Panics
+	///
+	/// This method panic if the `chain_spec` is `None`
+	pub fn expect_chain_spec(&self) -> &ChainSpec<G, E> {
```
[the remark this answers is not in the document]
cecton: The idea is to be able to make a `Configuration` before it being modified by the commandline arguments. For that I need defaults and `chain_spec`, like some other fields, are `Option` even though they are required for running a node. IT's not ideal. I think it would be best to have some kind of intermediate object, one that has the `Option`s that can be converted to one that doesn't have and it would be fail-able at that point.

But I don't want to do a complete refactoring. One step at a time. I'm already waiting for too long with this one and I get conflicts every day.

client/cli/src/runtime.rs
```diff
@@ -0,0 +1,116 @@
...
+	let mut stream_term = signal(SignalKind::terminate())?;
+
+	let t1 = stream_int.recv().fuse();
+	let t2 = stream_term.recv().fuse();
+	let t3 = func;
+
+	pin_mut!(t1, t2, t3);
+
+	select! {
+		_ = t1 => println!("Caught SIGINT"),
```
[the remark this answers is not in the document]
cecton: You want eprintln?

client/cli/src/lib.rs
cecton: It might not be useful anymore indeed. Right now it's only helper to initialize & run the command. Should I delete?

client/cli/src/lib.rs
cecton: The code moved. There is no support for the date in `VersionInfo`. I see only 2 possibilities to fix this:

1. delete `run_node` and let the caller do this manually: this is okay since `run_node` doesn't do  much anymore, only starts the service if light or full depending on the role
2. delete the date from the info

What do you want? (or you have another idea?)

client/cli/src/params.rs
cecton: That's intended. I close this discussion in favor of https://github.com/paritytech/substrate/pull/4692#discussion_r372049016

utils/browser/src/lib.rs
```diff
@@ -34,25 +34,24 @@ pub use console_log::init_with_level as init_console_log;
...
 where
-	C: Default,
 	G: RuntimeGenesis,
 	E: Extension,
 {
 	let name = chain_spec.name().to_string();
 
 	let transport = ExtTransport::new(transport);
-	let mut config = Configuration::default_with_spec_and_base_path(chain_spec, None);
+	let mut config = Configuration::default();
```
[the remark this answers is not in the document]
cecton: I don't see why it is a probleme. You can leave it to the default?

bin/node-template/src/command.rs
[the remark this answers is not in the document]
cecton: Closing in favor of https://github.com/paritytech/substrate/pull/4692#discussion_r372049016

client/service/src/config.rs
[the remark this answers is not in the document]
cecton: This will allow to solve: https://github.com/paritytech/cumulus/issues/24

Either `chain_spec` is not an `Option` but then we need to parse the command line argument to provide a sensible default (which prevents modification of the configuration before parsing the command line), either `chain_spec` is an `Option` but then it means running the node can fail with a runtime error if the developer didn't fill all the required fields in `Configuration`.

Note also that `chain_spec` will not be alone in this situation since `config_dir` has the same issue (if you don't call `init`, it will fail somehow) https://github.com/paritytech/substrate/blob/cecton-cli-refactoring/client/cli/src/lib.rs#L308 and `task_executor` https://github.com/paritytech/substrate/blob/cecton-cli-refactoring/client/service/src/builder.rs#L1152

If I try to revisit and improve the `ServiceBuilder` I will probably fix this. But I think this PR is already big enough.

client/cli/src/runtime.rs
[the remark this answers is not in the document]
tomaka: More like: why is anything printed at all?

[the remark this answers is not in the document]
cecton: Because usually that's what you do when you catch a signal?

Anyway. Do I remove it then?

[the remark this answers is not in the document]
cecton: Personally I prefer the messages that mednafen displays for SIGTERM and SIGINT:

* SIGTERM: "MUST TERMINATE ALL HUMANS"
* SIGINT: "How dare you interrupt me!"

But really I want to move on because I'm getting new conflicts every day. So I can remove it if that's what you want.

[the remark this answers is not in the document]
cecton: New merge conflicts... can we finish this review? :pray:

client/cli/src/runtime.rs
```diff
@@ -0,0 +1,116 @@
...
+use sc_service::{AbstractService, Configuration};
+use crate::error;
+use crate::informant;
+
+async fn main<F, E>(func: F) -> Result<(), Box<dyn std::error::Error>>
+where
+	F: Future<Output = Result<(), E>> + future::FusedFuture,
+	E: 'static + std::error::Error,
+{
+	use tokio::signal::unix::{signal, SignalKind};
```
expenses: Using this would mean that substrate wouldn't compile on windows, but I'm not sure we care about that.

cecton: I thought it was not supported but now that I checked the doc, they do mention building substrate entirely on Windows.

cecton: I just pushed a fixed for that

client/cli/src/runtime.rs
```diff
@@ -0,0 +1,138 @@
...
+	select! {
+		_ = t1 => println!("Caught SIGINT"),
+		_ = t2 => println!("Caught SIGTERM"),
+		res = t3 => res?,
+	}
+
+	Ok(())
+}
+
+#[cfg(target_family = "windows")]
```
tomaka:
```suggestion
#[cfg(not(unix))]
```

Let's stay as platform-independent as possible.

I'm also not sure if there's a point in catching specifically SIGINT and SIGTERM on Unix rather than using `tokio::signal::ctrlc` everywhere.

cecton: On Linux background services are usually stopped with SIGTERM (example here with system.d https://www.freedesktop.org/software/systemd/man/systemd.kill.html )

I believe it's best to keep both: SIGINT and SIGTERM for Unix, and ctrlc for the others.

(Btw, on tokio they use: `#[cfg(unix)]` and `#[cfg(windows)]` which are shortcuts for what I did.
https://docs.rs/tokio/0.2.9/src/tokio/signal/ctrl_c.rs.html#50-53 )

client/cli/src/lib.rs
cecton: option 2 applied

client/cli/src/lib.rs
bkchr: Yes, I think all of it could be moved into the `run` function of `RunCmd` itself.

cecton: It's better to keep it for consistency with `run_subcommand`. The user can use `sc_cli::run` and `sc_cli::run_subcommand` depending on the arm in the match. It doesn't hurt to have it so I will keep it.

If you think it's a bad decision and it's best to delete it tell me and I will update the code.

client/cli/src/lib.rs
bkchr: 3. Add `copyright_start_year` or similar to `VersionInfo`. The second year can just be `current_year`.

client/cli/src/runtime.rs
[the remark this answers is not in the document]
cecton: It's really fine as it is now. If you think it's a bad decision just tell me what to do exactly and I will change it.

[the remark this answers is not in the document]
bkchr: Delete it please.

utils/browser/src/lib.rs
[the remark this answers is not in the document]
bkchr: The `impl_name` is send to the telemetry. You can leave it to the default that is right, but not a nice interface.

client/service/src/config.rs
```diff
@@ -191,16 +184,13 @@ impl<C, G, E> Configuration<C, G, E> where
 			tracing_targets: Default::default(),
 			tracing_receiver: Default::default(),
 		};
-		configuration.network.boot_nodes = configuration.chain_spec.boot_nodes().to_vec();
```
bkchr: Where is this done now?

client/service/src/config.rs
```diff
@@ -191,16 +184,13 @@ impl<C, G, E> Configuration<C, G, E> where
...
-
-		configuration.telemetry_endpoints = configuration.chain_spec.telemetry_endpoints().clone();
```
bkchr: Same.

primitives/keyring/src/sr25519.rs
```diff
@@ -113,6 +113,33 @@ impl From<Keyring> for sp_runtime::MultiSigner {
 	}
 }
 
+#[derive(Debug)]
+pub struct ParseKeyringError;
```
bkchr: Do we still need these lines here?

client/cli/src/params.rs
```diff
@@ -532,9 +543,37 @@ pub struct RunCmd {
 	#[structopt(flatten)]
 	pub pool_config: TransactionPoolParams,
 
-	#[allow(missing_docs)]
-	#[structopt(flatten)]
-	pub keyring: Keyring,
+	/// Shortcut for `--name alice --validator` with session keys for `alice` added to keystore.
```
bkchr:
```suggestion
	/// Shortcut for `--name Alice --validator` with session keys for `Alice` added to keystore.
```

client/cli/src/params.rs
```diff
@@ -532,9 +543,37 @@ pub struct RunCmd {
...
+	#[structopt(long, conflicts_with_all = &["bob", "charlie", "dave", "eve", "ferdie", "one", "two"])]
+	pub alice: bool,
+
+	/// Shortcut for `--name bob --validator` with session keys for `bob` added to keystore.
```
bkchr:
```suggestion
	/// Shortcut for `--name Bob --validator` with session keys for `Bob` added to keystore.
```

client/cli/src/params.rs
```diff
@@ -532,9 +543,37 @@ pub struct RunCmd {
...
+	#[structopt(long, conflicts_with_all = &["alice", "charlie", "dave", "eve", "ferdie", "one", "two"])]
+	pub bob: bool,
+
+	/// Shortcut for `--name charlie --validator` with session keys for `charlie` added to keystore.
```
bkchr:
```suggestion
	/// Shortcut for `--name Charlie --validator` with session keys for `Charlie` added to keystore.
```

client/cli/src/params.rs
```diff
@@ -532,9 +543,37 @@ pub struct RunCmd {
...
+	#[structopt(long, conflicts_with_all = &["alice", "bob", "dave", "eve", "ferdie", "one", "two"])]
+	pub charlie: bool,
+
+	/// Shortcut for `--name dave --validator` with session keys for `dave` added to keystore.
```
bkchr:
```suggestion
	/// Shortcut for `--name Dave --validator` with session keys for `Dave` added to keystore.
```

client/cli/src/params.rs
```diff
@@ -532,9 +543,37 @@ pub struct RunCmd {
...
+	#[structopt(long, conflicts_with_all = &["alice", "bob", "charlie", "eve", "ferdie", "one", "two"])]
+	pub dave: bool,
+
+	/// Shortcut for `--name eve --validator` with session keys for `eve` added to keystore.
```
bkchr:
```suggestion
	/// Shortcut for `--name Eve --validator` with session keys for `Eve` added to keystore.
```

client/cli/src/params.rs
```diff
@@ -532,9 +543,37 @@ pub struct RunCmd {
...
+	#[structopt(long, conflicts_with_all = &["alice", "bob", "charlie", "dave", "ferdie", "one", "two"])]
+	pub eve: bool,
+
+	/// Shortcut for `--name ferdie --validator` with session keys for `ferdie` added to keystore.
```
bkchr:
```suggestion
	/// Shortcut for `--name Ferdie --validator` with session keys for `Ferdie` added to keystore.
```

client/cli/src/params.rs
```diff
@@ -532,9 +543,37 @@ pub struct RunCmd {
...
+	#[structopt(long, conflicts_with_all = &["alice", "bob", "charlie", "dave", "eve", "one", "two"])]
+	pub ferdie: bool,
+
+	/// Shortcut for `--name one --validator` with session keys for `one` added to keystore.
```
bkchr:
```suggestion
	/// Shortcut for `--name One --validator` with session keys for `One` added to keystore.
```

client/cli/src/params.rs
```diff
@@ -532,9 +543,37 @@ pub struct RunCmd {
...
+	#[structopt(long, conflicts_with_all = &["alice", "bob", "charlie", "dave", "eve", "ferdie", "two"])]
+	pub one: bool,
+
+	/// Shortcut for `--name two --validator` with session keys for `two` added to keystore.
```
bkchr:
```suggestion
	/// Shortcut for `--name Two --validator` with session keys for `Two` added to keystore.
```

client/cli/src/params.rs
```diff
@@ -884,112 +869,323 @@ pub enum CoreParams<CC, RP> {
...
+				Some('y') | Some('Y') => {},
+				_ => {
+					println!("Aborted");
+					return Ok(());
+				},
+			}
 		}
+
+		match fs::remove_dir_all(&db_path) {
+			Result::Ok(_) => {
```
bkchr:
```suggestion
			Ok(_) => {
```

What? :O

client/cli/src/params.rs
```diff
@@ -884,112 +869,323 @@ pub enum CoreParams<CC, RP> {
...
+				println!("{:?} removed.", &db_path);
+				Ok(())
+			},
+			Result::Err(ref err) if err.kind() == io::ErrorKind::NotFound => {
```
bkchr:
```suggestion
			Err(ref err) if err.kind() == io::ErrorKind::NotFound => {
```

client/cli/src/params.rs
```diff
@@ -884,112 +869,323 @@ pub enum CoreParams<CC, RP> {
...
+				eprintln!("{:?} did not exist.", &db_path);
+				Ok(())
+			},
+			Result::Err(err) => Result::Err(err.into())
```
bkchr:
```suggestion
			Err(err) => Result::Err(err.into())
```

client/service/src/config.rs
cecton: https://github.com/paritytech/substrate/pull/4692/files#diff-ed4918dd0da951e344e205c78f972484R312

client/service/src/config.rs
cecton: https://github.com/paritytech/substrate/pull/4692/files#diff-ed4918dd0da951e344e205c78f972484R313

primitives/keyring/src/sr25519.rs
cecton: No but it's good to have it if someone wants to make a CLI and use `--account` instead of the `--alice` and the others. I have the idea that sc-cli is a crate for the user to help build the CLI so it's good to be generic.

client/cli/src/params.rs
cecton: It's not me :innocent: I just moved things around

utils/browser/src/lib.rs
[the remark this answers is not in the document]
cecton: I understand. It doesn't bother me as it is now. But if you have a proposition for a nice interface I can change this.

There are also the fields for the version, commit and probably others. Having a simple `default()` is better for the tests though.

client/cli/src/runtime.rs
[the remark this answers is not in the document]
cecton: done

client/cli/src/lib.rs
cecton: OK. but to get the current year I will have to introduce chrono

cecton: done

client/cli/src/lib.rs
cecton: I think we wrote at the same time :grin: 

> Yes, I think all of it could be moved into the run function of RunCmd itself.

No it's not meant to be this way. Initialization must be separated from running the command so the user can change the whole logic with little effort. I might be wrong but here is the logic:

If the run command do the initialization for you than the user cannot insert something between the execution of the command and the initialization. The initialization does update the config.

Though I must say that none of our usecases are like that anymore (node, node-template, polkadot). Except maybe in cumulus where we will have to run 2 nodes but initialize only once.

(It also offers the user the opportunity to initialize without using the initializer but run the built-in command. This could be useful.)

cecton:
(Yet another conflict resolved. Also something that was already fixed with this PR....)

Let's finish this PR already please. What remains to get your approval?

client/cli/src/lib.rs
```diff
@@ -161,637 +147,227 @@ fn base_path(
...
-		// for backwards compatibility reasons.
-		let (exit_send, exit_recv) = std::sync::mpsc::channel();
-		let exit = exit.into_exit();
-		std::thread::spawn(move || {
-			block_on(exit);
-			let _ = exit_send.send(());
-		});
+	let mut full_version = sc_service::config::full_version_from_strs(
+		version.version,
+		version.commit
```
```suggestion
		version.commit,
```

client/cli/src/lib.rs
bkchr: @cecton Not sure why you ignore this?

client/cli/src/lib.rs
bkchr: @cecton ???

client/cli/src/lib.rs
bkchr: @cecton ???

client/service/src/config.rs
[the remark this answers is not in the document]
bkchr: Please create an issue for this!

bkchr:
> (Yet another conflict resolved. Also something that was already fixed with this PR....)
> 
> Let's finish this PR already please. What remains to get your approval?

Before asking such question, I expect that every review feedback is either worked in or a comment is added why not.

Found at least 3 comments that you ignored.

cecton:
> > What remains to get your approval?
> 
> Before asking such question, I expect that every review feedback is either worked in or a comment is added why not.
> 
> Found at least 3 comments that you ignored.

Obviously I didn't ignore them (I'm not "that" evil) but rather I didn't **see** them. I check on github + on the emails for the hidden ones but I guess it is possible I missed some. :pray: thanks for your review

client/cli/src/lib.rs
cecton: Sorry didn't see it for some reason. It appears in the PR files tab but not in the summary tab.

tomaka:
I haven't taken a look at the changes since my previous review, but it LGTM at the time, so +1 for merging.

client/service/src/config.rs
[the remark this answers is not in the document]
cecton: Created #4776

Issue #4420, opened 2019-12-17
Allow OffchainWorker Local Storage (DB) set/get over RPC.


tomusdrw:
The RPC API should pretty much directly reassemble the `sp_io` local storage API.

This will allow setting up some secrets via RPC or even enabling/disabling some workers (requires the worker code to support that obviously).

Also allows investigating the state of the worker on running node.

xlc:
This will also solves #3912

xlc:
Without #2303, this could be a DOS vector as anyone have access to this RPC will be able to insert data to fill the disk of the node.

tomusdrw:
Implemented in #4694

PR #4694, opened 2020-01-21
RPC api for offchain storage

ark930:
@tomusdrw 
This RP is for RPC api to set/get offchain local storage, See aslo https://github.com/paritytech/substrate/issues/4420.

### DIFF
--- a/client/offchain/src/lib.rs
+++ b/client/offchain/src/lib.rs
@@ -51,7 +51,7 @@ pub use sp_offchain::{OffchainWorkerApi, STORAGE_PREFIX};
 /// An offchain workers manager.
 pub struct OffchainWorkers<Client, Storage, Block: traits::Block> {
 	client: Arc<Client>,
-	pub db: Storage,
+	db: Storage,
 	_block: PhantomData<Block>,
 	thread_pool: Mutex<ThreadPool>,
 }
--- a/client/offchain/src/lib.rs
+++ b/client/offchain/src/lib.rs
@@ -51,7 +51,7 @@ pub use sp_offchain::{OffchainWorkerApi, STORAGE_PREFIX};
 /// An offchain workers manager.
 pub struct OffchainWorkers<Client, Storage, Block: traits::Block> {
 	client: Arc<Client>,
-	db: Storage,
+	pub db: Storage,
 	_block: PhantomData<Block>,
 	thread_pool: Mutex<ThreadPool>,
 }
--- a/client/rpc-api/src/offchain/error.rs
+++ b/client/rpc-api/src/offchain/error.rs
@@ -1,4 +1,4 @@
-// Copyright 2017-2020 Parity Technologies (UK) Ltd.
+// Copyright 2020 Parity Technologies (UK) Ltd.
 // This file is part of Substrate.
 
 // Substrate is free software: you can redistribute it and/or modify
--- a/client/rpc-api/src/offchain/error.rs
+++ b/client/rpc-api/src/offchain/error.rs
@@ -1,0 +1,51 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
+// This file is part of Substrate.
+
+// Substrate is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Offchain RPC errors.
+
+use jsonrpc_core as rpc;
+
+/// Offchain RPC Result type.
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// Offchain RPC errors.
+#[derive(Debug, derive_more::Display, derive_more::From)]
+pub enum Error {
+	/// Unavailable storage kind error.
+	#[display(fmt="This storage kind is not available yet.")]
+	UnavailableStorageKind,
+}
+
+impl std::error::Error for Error {
+	fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+		None
+	}
+}
+
+/// Base error code for all offchain errors.
+const BASE_ERROR: i64 = 5000;
+
+impl From<Error> for rpc::Error {
+	fn from(e: Error) -> Self {
+		match e {
+			Error::UnavailableStorageKind => rpc::Error {
+				code: rpc::ErrorCode::ServerError(BASE_ERROR + 1),
+				message: "This storage kind is not available yet" .into(),
+				data: None,
+			},
+		}
+	}
+}
--- a/client/rpc-api/src/offchain/mod.rs
+++ b/client/rpc-api/src/offchain/mod.rs
@@ -20,7 +20,7 @@ pub mod error;
 
 use jsonrpc_derive::rpc;
 use self::error::Result;
-use sp_core::offchain::StorageKind;
+use sp_core::{Bytes, offchain::StorageKind};
 
 pub use self::gen_client::Client as OffchainClient;
 
@@ -29,9 +29,9 @@ pub use self::gen_client::Client as OffchainClient;
 pub trait OffchainApi {
 	/// Set offchain local storage under given key and prefix.
 	#[rpc(name = "offchain_localStorageSet")]
-	fn set_local_storage(&self, kind: StorageKind, key: Vec<u8>, value: Vec<u8>) -> Result<()>;
+	fn set_local_storage(&self, kind: StorageKind, key: Bytes, value: Bytes) -> Result<()>;
 
 	/// Get offchain local storage under given key and prefix.
 	#[rpc(name = "offchain_localStorageGet")]
-	fn get_local_storage(&self, kind: StorageKind, key: Vec<u8>) -> Result<Option<Vec<u8>>>;
+	fn get_local_storage(&self, kind: StorageKind, key: Bytes) -> Result<Option<Bytes>>;
 }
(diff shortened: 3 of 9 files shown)


client/rpc-api/src/chain/mod.rs
```diff
@@ -54,6 +55,23 @@ pub trait ChainApi<Number, Hash, Header, SignedBlock> {
 	#[rpc(name = "chain_getFinalizedHead", alias("chain_getFinalisedHead"))]
 	fn finalized_head(&self) -> Result<Hash>;
 
+	/// Set offchain storage under given key and prefix.
+	#[rpc(name = "chain_setOffchainStorage")]
```
tomusdrw:
```suggestion
	#[rpc(name = "offchain_localStorageSet")]
```

I'd suggest to keep the naming from internal modules, i.e. `Local Storage` and verb at the end.
Also IMHO it deserves a separate namespace in the RPC (i.e. not `chain_`, but rather `offchain_`) and a separate file for this.

client/rpc-api/src/chain/mod.rs
```diff
@@ -54,6 +55,23 @@ pub trait ChainApi<Number, Hash, Header, SignedBlock> {
...
+	fn set_offchain_storage(
+		&self,
+		prefix: StorageKey,
```
tomusdrw: Could we rather keep the enum instead of allowing any kind of prefix directly? I think this will get sorted as well if you switch to use `OffchainWorkers` structure directly instead of going through `Client` to low-level `OffchainStorage` db.

client/src/client.rs
```diff
@@ -1265,6 +1266,25 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 		self.backend.blockchain().justification(*id)
 	}
 
+	/// Set offchain storage under given key and prefix.
+	pub fn set_offchain_storage(&self, prefix: &[u8], key: &[u8], value: &[u8]) -> sp_blockchain::Result<()> {
```
tomusdrw: I'd prefer to avoid going through the `Client` and use `OffchainWorkers` or `OffchainStorage` directly in the RPC.

client/rpc-api/src/chain/mod.rs
ark930: You mean prefix must be StorageKind type, right?

client/src/client.rs
ark930: Ok, I'll try to use OffchainWorkers.

tomusdrw:
@ark930 sad to see it being closed :disappointed: Is it just a mistake? Is there any way I could help you finish it up?

ark930:
> @ark930 sad to see it being closed 😞 Is it just a mistake? Is there any way I could help you finish it up?

I just revert commit and re-implement Offchain RPC Api by high level `OffchainStorage`.

primitives/core/src/offchain/mod.rs
```diff
@@ -50,6 +52,7 @@ pub trait OffchainStorage: Clone + Send + Sync {
 
 /// A type of supported crypto.
 #[derive(Clone, Copy, PartialEq, Eq, Encode, Decode, RuntimeDebug, PassByEnum)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```
tomusdrw:
```suggestion
#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
```
You can do this to avoid adding extra, feature-gated import line.

client/offchain/src/lib.rs
```diff
@@ -51,7 +51,7 @@ pub use sp_offchain::{OffchainWorkerApi, STORAGE_PREFIX};
 /// An offchain workers manager.
 pub struct OffchainWorkers<Client, Storage, Block: traits::Block> {
 	client: Arc<Client>,
-	db: Storage,
+	pub db: Storage,
```
tomusdrw: Would be best to avoid making it public.

ark930:
> [clabot:check]

client/rpc/src/offchain/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+pub struct Offchain<T: OffchainStorage> {
+	/// Offchain storage
+	storage: Arc<Mutex<T>>,
+}
+
+impl<T: OffchainStorage> Offchain<T> {
+	/// Create new instance of Offchain API.
+	pub fn new(storage: T) -> Self {
+		Offchain {
+			storage: Arc::new(Mutex::new(storage)),
```
tomusdrw: A `RwLock` should suffice here.

client/rpc-api/src/offchain/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+use sp_core::offchain::StorageKind;
+
+pub use self::gen_client::Client as OffchainClient;
+
+/// Substrate offchain RPC API
+#[rpc]
+pub trait OffchainApi {
+	/// Set offchain local storage under given key and prefix.
+	#[rpc(name = "offchain_localStorageSet")]
+	fn set_local_storage(&self, kind: StorageKind, key: Vec<u8>, value: Vec<u8>) -> Result<()>;
```
tomusdrw: This all should be `Bytes` instead of `Vec<u8>` otherwise we will get an array of numbers instead of hex-string.

ark930:
> I was thinking about using the entire `OffchainWorkers` struct, as it's already in `Arc`. Cloning `offchain::LocalStorage` works fine, so I'm fine keeping it as-is. Few small things, but looks good

I try to use `OffchainWorkers`, but the complier reminder me `OffchainWorkers`'s db is not mutable when I call `set` method of `OffchainStorage`. How could I resolve this ploblem?

client/rpc-api/src/offchain/error.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2020 Parity Technologies (UK) Ltd.
```

client/rpc-api/src/offchain/mod.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2020 Parity Technologies (UK) Ltd.
```

client/rpc/src/offchain/mod.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2020 Parity Technologies (UK) Ltd.
```

client/rpc/src/offchain/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+		}
+	}
+}
+
+impl<T: OffchainStorage + 'static> OffchainApi for Offchain<T> {
+	/// Set offchain local storage under given key and prefix.
+	fn set_local_storage(&self, kind: StorageKind, key: Bytes, value: Bytes) -> Result<()> {
+		let prefix = match kind {
+			StorageKind::PERSISTENT => sp_offchain::STORAGE_PREFIX,
+			StorageKind::LOCAL => return Err(Error::UnavailableStorageKind),
```
bkchr: Should this be supported in the future? If not, I don't understand why we take `kind` here.

client/rpc/src/offchain/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+		};
+		self.storage.write().set(prefix, &*key, &*value);
+		Ok(())
+	}
+
+	/// Get offchain local storage under given key and prefix.
+	fn get_local_storage(&self, kind: StorageKind, key: Bytes) -> Result<Option<Bytes>> {
+		let prefix = match kind {
```
bkchr: Same question.

client/rpc/src/offchain/tests.rs
```diff
@@ -0,0 +1,36 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2020 Parity Technologies (UK) Ltd.
```

client/service/src/builder.rs
```diff
@@ -1046,13 +1046,26 @@ ServiceBuilder<
...
 			let system = system::System::new(system_info, system_rpc_tx.clone());
 
-			sc_rpc_server::rpc_handler((
-				state::StateApi::to_delegate(state),
-				chain::ChainApi::to_delegate(chain),
-				author::AuthorApi::to_delegate(author),
-				system::SystemApi::to_delegate(system),
-				rpc_extensions.clone(),
-			))
+			match offchain_storage.clone() {
```
bkchr:
```suggestion
			match offchain_storage {
```

client/service/src/builder.rs
```diff
@@ -1046,13 +1046,26 @@ ServiceBuilder<
...
+				Some(storage) => {
+					let offchain = sc_rpc::offchain::Offchain::new(storage);
+					sc_rpc_server::rpc_handler((
```
bkchr: @tomusdrw it would be nice if `rpc_handlers` would support `Option<>`, so we don't have this repetition here.

client/rpc/src/offchain/mod.rs
tomusdrw: Yes, it mirrors the API available for offchain workers. Implementation could be done on top of #3263

client/service/src/builder.rs
tomusdrw: Done: https://github.com/paritytech/jsonrpc/pull/535

client/rpc/src/offchain/mod.rs
bkchr: What is the difference between local and persistent?

tomusdrw: Fork awarenes. `LOCAL` entries should be removed in case there is an reorg, while `PERSISTENT` should stay even if reorg happened (i.e. we see results produced by offchain workers that run on a different fork). The second one is useful if you want to make sure to avoid duplication of external "side effects".

client/service/src/builder.rs
bkchr: @ark930 please apply this as well.

xlc:
Some one need to document this as an unsafe RPC https://github.com/paritytech/substrate/wiki/Public-RPC

PR #4695, opened 2020-01-21
Create Benchmarking Setup for Identity Pallet

shawntabrizi:
This is an in-progress PR for creating a benchmarking setup for the Identity pallet.

Likely lots of this will be manually written for now, and then automated with macros later.


marcio-diaz:
I'm trying to build some kind of push button framework for benchmarking: https://github.com/paritytech/substrate/pull/4614. I would love to have your feedback :).

primitives/runtime/src/traits.rs
```diff
@@ -1263,6 +1263,18 @@ pub trait BlockIdTo<Block: self::Block> {
 	) -> Result<Option<NumberFor<Block>>, Self::Error>;
 }
 
+/// The module benchmarking trait.
+pub trait Benchmarking {
+	/// Type for the results of benchmarks.
+	type BenchmarkResults;
```
gavofyork: this stuff probably isn't necessary.

frame/identity/src/benchmarking.rs
```diff
@@ -0,0 +1,149 @@
+use super::*;
```
kianenigma: license header.

frame/identity/src/benchmarking.rs
```diff
@@ -0,0 +1,149 @@
...
+	type ForceOrigin = EnsureSignedBy<Two, u64>;
+}
+type System = frame_system::Module<Benchmark>;
+type Balances = pallet_balances::Module<Benchmark>;
+type Identity = Module<Benchmark>;
+
+pub type IdentityCall = crate::Call<Benchmark>;
+
+pub mod set_identity {
+	use super::*;
```
kianenigma:
```suggestion
use super::*;
```

frame/identity/src/lib.rs
```diff
@@ -862,14 +864,59 @@ decl_module! {
...
+				// Select the mid value for all the other components.
+				let c: Vec<(BenchmarkParameter, u32)> = components.iter()
+					.map(|(n, l, h)|
+						(*n, if n == name { component_value } else { (h - l) / 2 + l })
+					).collect();
+
+				for _r in 0..Self::REPEATS {
+					let instance = benchmarking::set_identity::instance(&c);
+					let start = sp_io::offchain::timestamp();
+					instance.dispatch(Some(0).into());
```
kianenigma: Which storage type does this use?

frame/identity/src/lib.rs
```diff
@@ -862,14 +864,59 @@ decl_module! {
...
+					let elapsed = finish.0 - start.0;
+					results.push((c.clone(), elapsed));
+				}
+			}
+		}
+		return results;
+	}
+}
+
+sp_api::decl_runtime_apis! {
```
kianenigma: I think we usually place the apis outside of the runtime? but maybe it is not a big deal.

thiolliere: Indeed I think at the end we should have only one runtime_api function which is `run_benchmarks(module_to_run: Vec<ModuleInde>)` or something like that

frame/identity/src/benchmarking.rs
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/// Assumes externalities are set up with a mutable state.
+	///
+	/// Panics if `component_name` isn't from `set_identity::components` or `component_value` is out of
+	/// the range of `set_identity::components`.
+	///
+	/// Sets up state randomly and returns a randomly generated `set_identity` with sensible (fixed)
+	/// values for all complexity components except those mentioned in the identity.
+	pub fn instance(components: &[(BenchmarkParameter, u32)]) -> Call
```
gavofyork: This should be able to work with the node's runtime.

frame/identity/src/benchmarking.rs
```diff
@@ -0,0 +1,151 @@
...
+	{
+		// Add r registrars
+		let r = components.iter().find(|&c| c.0 == BenchmarkParameter::R).unwrap();
+		assert_eq!(Balances::set_balance(Origin::ROOT, 1u64.into(), 1_000_000_000_000, 0), Ok(()));
+		for i in 0..r.1 {
+			assert_eq!(Balances::set_balance(Origin::ROOT, i.into(), 1_000_000_000_000, 0), Ok(()));
```
gavofyork: `i.into()` isn't a random account

frame/identity/src/lib.rs
kianenigma: I was working on this and it should be placed in a std sub-crate of the frame module. See how system is defining some extra runtime api for its rpc.

shawntabrizi:
Switching to a branch _not_ in a fork.

PR #4697, opened 2020-01-21
Inspection extension to node CLI

tomusdrw:
This PR introduces a new `node-inspect` CLI extension and adds a sub-command to default `node` binary.

The idea is to be able to inspect (print details of) blocks and extrinsics from either:
1. existing Database (without running the node) or 
2. separate SCALE-encoded hex string.

The data is decoded with current (native) version of the runtime that's built into the client.

I call it an "extension", cause it should be easy to intergrate with other chains/binaries if needed. Also the way we print blocks and extrinsics is fully customisable, by default we print simple `Debug` details.

Example:
```
$ ./target/release/substrate inspect --dev --pruning=archive block 0260daa981295a56bf2e59f6ce5faa47489067372ace3116f7b3272fd8d33958
Header:
Header { parent_hash: 0x71fb94a50e7482581a4678041c90ad97c99fd25e7a930f933493bf22e22bd1f2, number: 7, state_root: 0xf5839bbb8e772d235f3f945c7777de2f15b03ea16f68f858759f6b170a7c6ad4, extrinsics_root: 0x9d23d5768c4ba48308a3474e93bcfbb28c9442cec015497aac79b39b8a046776, digest: Digest { logs: [DigestItem::PreRuntime([66, 65, 66, 69], [2, 0, 0, 0, 0, 254, 88, 98, 31, 0, 0, 0, 0]), DigestItem::Seal([66, 65, 66, 69], [38, 85, 154, 128, 86, 117, 164, 127, 144, 251, 230, 208, 86, 247, 181, 192, 105, 76, 134, 208, 22, 232, 6, 166, 2, 79, 191, 52, 76, 107, 125, 42, 134, 127, 5, 37, 62, 68, 59, 191, 73, 199, 254, 133, 217, 183, 178, 161, 66, 24, 166, 26, 186, 154, 139, 34, 10, 115, 158, 78, 174, 28, 56, 138])] } }
Bytes: 71fb94a50e7482581a4678041c90ad97c99fd25e7a930f933493bf22e22bd1f21cf5839bbb8e772d235f3f945c7777de2f15b03ea16f68f858759f6b170a7c6ad49d23d5768c4ba48308a3474e93bcfbb28c9442cec015497aac79b39b8a046776080642414245340200000000fe58621f000000000542414245010126559a805675a47f90fbe6d056f7b5c0694c86d016e806a6024fbf344c6b7d2a867f05253e443bbf49c7fe85d9b7b2a14218a61aba9a8b220a739e4eae1c388a08280403000b90e082c86f0110040e0010
Extrinsics (2)
- 0:
 UncheckedExtrinsic(None, Call::Timestamp(set(1579617018000,)))
 Bytes: 280403000b90e082c86f01
- 1:
 UncheckedExtrinsic(None, Call::FinalityTracker(final_hint(4,)))
 Bytes: 10040e0010
```


bin/node/cli/src/cli.rs
```diff
@@ -138,6 +146,13 @@ pub fn run<I, T, E>(args: I, exit: E, version: sc_cli::VersionInfo) -> error::Re
 		ParseAndPrepare::PurgeChain(cmd) => cmd.run(load_spec),
 		ParseAndPrepare::RevertChain(cmd) => cmd.run_with_builder(|config: Config<_, _>|
 			Ok(new_full_start!(config).0), load_spec),
+		ParseAndPrepare::CustomCommand(CustomSubcommands::Inspect(cmd)) => cmd
+			.run_with_builder(|config: Config<_, _>| {
+				let client = sc_service::new_full_client::<
+					node_runtime::Block, node_runtime::RuntimeApi, node_executor::Executor, _, _, _,
+				>(&config)?;
+				Ok(node_inspect::Inspector::<node_runtime::Block>::new(client))
+			}, load_spec, &version),
```
NikVolf: each argument on the new line?

bin/node/inspect/src/cli.rs
```diff
@@ -0,0 +1,244 @@
...
+		/// representing SCALE encoding of an extrinsic.
+		#[structopt(value_name = "BLOCK:INDEX or BYTES")]
+		input: String,
+	},
+}
+
+impl InspectCmd {
+	/// Run the inspect command, passing an inspector buuilder,
+	/// spec_factory and software version.
+	pub fn run_with_builder<C, G, E, B, P>(
```
NikVolf: I think you are missing `fill_config_keystore_in_memory` here :)

arkpar:
Is it that different from `substrate export-blocks --json` ?

tomusdrw:
@arkpar There are few reasons I've decided to make it a separate subcommand:
1. Export blocks can only be used for in-chain blocks. I would like to be able to inspect Blocks or Extrinsics that are not part of the chain. For instance to understand why they were rejected by the pool or the client.
2. Export block is using raw data / opaque extrinsics, so the best you get about extrinsics is just it's raw bytes, which doesn't solve my issues at all. I'm aware that current implementation is always limited to the native runtime (i.e. you can't decode old extrinsics if the format changed), but I'm confident that it would still be useful.

My main motivation was extrinsics inspection by giving the encoding, I've added support for things that are part of the chain (by hash or index) and blocks because it just seemed easy, but I'm happy to limit the functionality if you feel that the overlap is not acceptable.

gavofyork:
another merge needed.

bin/node/cli/Cargo.toml
```diff
@@ -100,6 +101,7 @@ tempfile = "3.1.0"
 build-script-utils = { version = "2.0.0", package = "substrate-build-script-utils", path = "../../../utils/build-script-utils" }
 structopt = { version = "0.3.8", optional = true }
 node-transaction-factory = { version = "0.8.0", optional = true, path = "../transaction-factory" }
+node-inspect = { version = "0.8.0", optional = true, path = "../inspect" }
```
bkchr: Why do we need it as build dependency?

bin/node/inspect/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+			}
+		}
+
+		let ext = match input {
+			cli::ExtrinsicAddress::Block(block, index) => {
+				let block = self.get_block(block)?;
+				block.extrinsics()
+					.get(index)
+					.cloned()
+					.ok_or_else(|| Error::NotFound(format!("Not found TODO")))?
```
bkchr: ?

bin/node/cli/Cargo.toml
tomusdrw: I was puzzled too, but it seems that in the `build.rs` we are building some bash-completion support and it requires duplicating all the deps as build deps.

bin/node/inspect/src/lib.rs
tomusdrw: :see_no_evil:

bin/node/cli/Cargo.toml
bkchr: Okay :woman_shrugging:

gnunicorn:
CI failing. Maybe an update to master fixes it?

PR #4700, opened 2020-01-21
grandpa: filter some telemetry events on larger voter sets

andresilva:
The GRANDPA client sends telemetry events upon receiving a new vote or commit. These events are then used to produce a visualization of the round state. With a high number of authorities, say 150, every node will be sending ~300 telemetry events per round, for a total of ~45000 events. Currently on the Kusama network, a round usually takes less than a second to complete. @arkpar mentioned that formatting of these telemetry events showed up in the memory profiler. With a validator set of that size the visualization UI is also unusable, therefore we will filter sending these telemetry events if the validator set size is bigger than 10.


bkchr:
We should have an issue to fix these telemetry macros.

PR #4703, opened 2020-01-21
Support `u128`/`i128` in runtime interface

bkchr:
This implements support for `u128`/`i128` as parameters/return value in
runtime interfaces. As we can not pass them as identity, as for the
other primitives types, we pass them as an pointer to an `[u8; 16]` array.

### DIFF
--- a/primitives/runtime-interface/src/impls.rs
+++ b/primitives/runtime-interface/src/impls.rs
@@ -494,7 +494,7 @@ macro_rules! for_u128_i128 {
 		#[cfg(not(feature = "std"))]
 		impl FromFFIValue for $type {
 			fn from_ffi_value(arg: u32) -> $type {
-				unsafe { mem::transmute::<[u8; 16], $type>(<[u8; 16]>::from_ffi_value(arg)) }
+				<$type>::from_le_bytes(<[u8; mem::size_of::<$type>()]>::from_ffi_value(arg))
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


primitives/runtime-interface/src/impls.rs
```diff
@@ -471,3 +471,56 @@ impl<T: sp_wasm_interface::PointerType> IntoFFIValue for Pointer<T> {
...
+
+		#[cfg(feature = "std")]
+		impl FromFFIValue for $type {
+			type SelfInstance = $type;
+
+			fn from_ffi_value(context: &mut dyn FunctionContext, arg: u32) -> Result<$type> {
+				let data = context.read_memory(Pointer::new(arg), 16)?;
+				let mut res = [0u8; 16];
+				res.copy_from_slice(&data);
+				Ok(unsafe { mem::transmute::<[u8; 16], $type>(res) })
```
pepyakin: 1. I wonder if this is safe? The alignment of u128 might be 16 and I am not sure sure if we can say anything regarding this about `res`. 
2. Does it assume that wasm and host have the same endianness?

I think this can be swapped with `u128::from_le_bytes` here instead of transmute?

bkchr: 1. What? Why can we not say the same about `res`? It is as well 16 bytes. (Looking at the rust source code for `from_le_bytes` it does exactly the same transmute)
2. Yes it assumes it. Please don't start again this xD As it is no primitive type, I probably also need to call `to_le_bytes` on the host?

bkchr: But good catches! I searched yesterday evening for `*_let_bytes`, but failed for some reason :woman_shrugging:

PR #4715, opened 2020-01-22
Add fool protection and comment construct_block

pepyakin:
Instead of getting a transaction root mismatch at the end of the block execution, you will get a nice panic message, which might save you some nerves and time units.

### DIFF
--- a/bin/node/executor/tests/common.rs
+++ b/bin/node/executor/tests/common.rs
@@ -21,7 +21,7 @@ use sp_core::{
 	Blake2Hasher, NeverNativeValue, NativeOrEncoded,
 	traits::CodeExecutor,
 };
-use sp_runtime::traits::Header as HeaderT;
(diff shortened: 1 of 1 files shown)


bin/node/executor/tests/common.rs
```diff
@@ -126,14 +130,20 @@ pub fn construct_block(
...
 			"BlockBuilder_apply_extrinsic",
-			&i.encode(),
+			&extrinsic.encode(),
 			true,
 			None,
-		).0.unwrap();
+		).0.expect("application of an extrinsic failed").into_encoded();
+		match ApplyExtrinsicResult::decode(&mut &r[..]).expect("apply result deserialization failed") {
+			Ok(_) => {},
+			Err(e) => panic!("panic while applying extrinsic: {:?}", e),
```
bkchr:
```suggestion
			Err(e) => panic!("Applying extrinsic failed: {:?}", e),
```

PR #4718, opened 2020-01-23
add state_getKeysPaged

### LINKED ISSUE #4717: RPC method state_nextKey
Currently we have RPC `state_getKeys` which returns all keys with a same prefix, this may cause trouble for example trying to iterate all the account free balances.

Can we have `state_nextKey` that is basically `sp_io::storage::next_key` so SDK can control the iteration process.

xlc:
Closes #4717

Should also have `state_getNextChildKey`, but it can happen in later PR.


client/rpc/src/state/state_light.rs
```diff
@@ -199,6 +199,14 @@ impl<Block, F, B, E, RA> StateBackend<B, E, Block, RA> for LightState<Block, F,
 		Box::new(result(Err(client_err(ClientError::NotAvailableOnLightClient))))
 	}
 
+	fn storage_next_key(
+		&self,
+		_block: Option<Block::Hash>,
+		_key: StorageKey,
+	) -> FutureResult<Option<StorageKey>> {
+		Box::new(result(Err(client_err(ClientError::NotAvailableOnLightClient))))
```
xlc: Not sure if it is possible to implement this on light client

NikVolf: @svyatonik ?

svyatonik: `NotAvailableOnLightClient` should be enough for now.

xlc:
Ping @bkchr @tomusdrw

client/rpc-api/src/state/mod.rs
```diff
@@ -39,10 +39,20 @@ pub trait StateApi<Hash> {
 	#[rpc(name = "state_call", alias("state_callAt"))]
 	fn call(&self, name: String, bytes: Bytes, hash: Option<Hash>) -> FutureResult<Bytes>;
 
-	/// Returns the keys with prefix, leave empty to get all the keys
+	/// Returns the keys with prefix, leave empty to get all the keys.
```
tomusdrw: Shall we deprecate this method then? I imagine adding deprecation notice in the code and printing a warning when the method is used (but at most once a minute or so).

client/src/client.rs
```diff
@@ -236,12 +236,40 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
 	}
 
+	/// Paged version of `storage_keys`. Only return up to `count` keys after `start_key` or `prefix`.
+	pub fn storage_keys_paged(
+		&self,
+		id: &BlockId<Block>,
+		prefix: &StorageKey,
+		count: u32,
+		start_key: &Option<StorageKey>
+	) -> sp_blockchain::Result<Vec<StorageKey>> {
```
tomusdrw: Why not return an iterator here and allocate on the caller side (when we know that `count` is within reasonable bounds)?

client/src/client.rs
```diff
@@ -236,12 +236,40 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
 		let keys = self.state_at(id)?.keys(&key_prefix.0).into_iter().map(StorageKey).collect();
 		Ok(keys)
 	}
 
+	/// Paged version of `storage_keys`. Only return up to `count` keys after `start_key` or `prefix`.
+	pub fn storage_keys_paged(
+		&self,
+		id: &BlockId<Block>,
+		prefix: &StorageKey,
+		count: u32,
```
tomusdrw: I think it should be `usize` in this layer already.
```suggestion
		count: usize,
```

client/src/client.rs
```diff
@@ -236,12 +236,40 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
+		let state = self.state_at(id)?;
+		let mut result = Vec::<StorageKey>::with_capacity(count as usize);
+		let start_key = start_key.as_ref().unwrap_or(prefix);
+		let prefix = &prefix.0[..];
+
+		let mut current_key = start_key.0.clone();
+		while result.len() < (count as usize) {
+			let maybe_next_key = state.next_storage_key(&current_key[..])
+				.map_err(|e| sp_blockchain::Error::from_state(Box::new(e)))?
+				.filter(|v| v.starts_with(prefix));
```
tomusdrw: I think this condition should be more strict. `next_storage_key` will return values that are ordered, so when we find first one that doesn't share the prefix we should just quit right away.
Current implementation may traverse entire trie if `count` is greater than than the number of keys that share the same prefix, no?

Also have you considered using `self.state_at(id)?.keys(&key_prefix.0)` and only paging from there? How is that method implemented (I guess it allocates currently), perhaps we could just alter it to do what we want (i.e. return an iterator over prefixed fields instead of allocating everything right away).

client/rpc-api/src/state/mod.rs
xlc: printing a warning on server wouldn't be helpful unless we modify the response to indicate this method is deprecated. better do it on SDK side?

client/src/client.rs
xlc: It does break immediately when finding first result that does not start with prefix. the `filter` call is on `Option` not `Vec`.

But modify `keys` to make it return an iterator sounds a better approach, will do this way.

client/rpc/src/state/mod.rs
```diff
@@ -244,6 +255,27 @@ impl<B, E, Block, RA> StateApi<Block::Hash> for State<B, E, Block, RA>
...
 
+	fn storage_keys_paged(
+		&self,
+		prefix: StorageKey,
+		count: u32,
+		start_key: Option<StorageKey>,
+		block: Option<Block::Hash>,
+	) -> FutureResult<Vec<StorageKey>> {
+		if count > STORAGE_KEYS_PAGED_MAX_COUNT {
+			return Box::new(result(Err(client_err(
```
bkchr: I want a variant for this in the `Error` enum and have a fixed error code assigned. This makes it possible for the other side to react to this without parsing string errors.

client/rpc-api/src/state/mod.rs
```diff
@@ -39,10 +39,21 @@ pub trait StateApi<Hash> {
...
+	/// DEPRECATED: Please use `state_getKeysPaged` with proper paging support.
+	/// Returns the keys with prefix, leave empty to get all the keys.
 	#[rpc(name = "state_getKeys")]
 	fn storage_keys(&self, prefix: StorageKey, hash: Option<Hash>) -> FutureResult<Vec<StorageKey>>;
 
+	/// Returns the keys with prefix with pagination support. Max count is 10 000.
+	#[rpc(name = "state_getKeysPaged", alias("state_getKeysPagedAt"))]
+	fn storage_keys_paged(
+		&self,
+		prefix: StorageKey,
```
bkchr: If the prefix is optional, I would say that we make it clear by using an `Option`.

xlc: prefix is not optional. it can be an empty vec which iterates all keys (as they all prefixed with a zero-length key)

client/rpc/src/state/mod.rs
xlc: What is the best place for me to add this new variant? Add a new `enum Error` in this file?

bkchr: Add it to the already existing error enum for state.

client/rpc-api/src/state/mod.rs
bkchr: I'm not sure what you mean by "zero-length key". Such a construct does not exists. The one that is doing the prefix match is you with `starts_with` that returns `true` for the empty slice. Api wise `Option` is much cleaner. That `None` and `Some("")` will return the same values is okay, but `Option` makes the api much better.

client/rpc-api/src/state/error.rs
```diff
@@ -41,6 +41,14 @@ pub enum Error {
 		/// Details of the error message.
 		details: String,
 	},
+	/// Provided count exceeds maximum value.
+	#[display(fmt = "count exceeds maximum value. value: {}, max: {}", value, max)]
```
bkchr: Looks good! But please also update below the `From<Error> for rpc::Error` and just use `BASE_ERROR + 2`

client/rpc/src/state/mod.rs
```diff
@@ -244,6 +255,24 @@ impl<B, E, Block, RA> StateApi<Block::Hash> for State<B, E, Block, RA>
...
+	) -> FutureResult<Vec<StorageKey>> {
+		if count > STORAGE_KEYS_PAGED_MAX_COUNT {
+			return Box::new(result(Err(
+				Error::InvalidCount {
+					value: count,
+					max: STORAGE_KEYS_PAGED_MAX_COUNT,
+				}
+			)));
+		}
+		self.backend.storage_keys_paged(block, prefix.unwrap_or_else(|| StorageKey(vec![])), count, start_key)
```
bkchr: Please forward the `prefix` option to the iterator directly.

PR #4719, opened 2020-01-23
Add `system_time` host function

shawntabrizi:
This PR introduces a new host function which allows the runtime to get the `SystemTime` from the node.

The intention is to use this to create benchmarking timers within the runtime.

I have placed this in the `misc` host functions, so it is accessible as:

```
sp_io::misc::system_time()
````

And returns the current time in nanoseconds using `u128`.

### DIFF
--- a/primitives/io/src/lib.rs
+++ b/primitives/io/src/lib.rs
@@ -33,7 +33,6 @@ use sp_std::vec::Vec;
 use sp_std::ops::Deref;
 
 #[cfg(feature = "std")]
-#[cfg(feature = "std")]
 use sp_core::{
 	crypto::Pair,
 	traits::{KeystoreExt, CallInWasmExt},
--- a/primitives/io/src/lib.rs
+++ b/primitives/io/src/lib.rs
@@ -775,9 +775,9 @@ pub trait Logging {
 /// Interface that provides functions for benchmarking the runtime.
 #[runtime_interface]
 pub trait Benchmarking {
-		/// Get the current system time in nanoseconds.
+		/// Get the number of nanoseconds passed since the UNIX epoch
 		///
-		/// WARNING! This is a non-determinisic call. Do not use this within
+		/// WARNING! This is a non-deterministic call. Do not use this within
 		/// consensus critical logic.
 		fn current_time() -> u128 {
 			std::time::SystemTime::now().duration_since(std::time::SystemTime::UNIX_EPOCH)
--- a/primitives/io/src/lib.rs
+++ b/primitives/io/src/lib.rs
@@ -930,7 +930,6 @@ pub type SubstrateHostFunctions = (
 	logging::HostFunctions,
 	sandbox::HostFunctions,
 	crate::trie::HostFunctions,
-	benchmarking::HostFunctions,
 );
 
 #[cfg(test)]
--- a/primitives/io/src/lib.rs
+++ b/primitives/io/src/lib.rs
@@ -33,7 +33,6 @@ use sp_std::vec::Vec;
 use sp_std::ops::Deref;
 
 #[cfg(feature = "std")]
-use std::time;
 #[cfg(feature = "std")]
 use sp_core::{
 	crypto::Pair,
--- a/primitives/io/src/lib.rs
+++ b/primitives/io/src/lib.rs
@@ -930,6 +930,7 @@ pub type SubstrateHostFunctions = (
 	logging::HostFunctions,
 	sandbox::HostFunctions,
 	crate::trie::HostFunctions,
+	benchmarking::HostFunctions,
 );
 
 #[cfg(test)]
--- a/primitives/io/src/lib.rs
+++ b/primitives/io/src/lib.rs
@@ -34,7 +34,6 @@ use sp_std::ops::Deref;
 
 #[cfg(feature = "std")]
 use std::time;
-
 #[cfg(feature = "std")]
 use sp_core::{
 	crypto::Pair,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: primitives/io/src/lib.rs [hunk 1]
```diff
@@ -32,6 +32,9 @@ use sp_std::vec::Vec;
 #[cfg(feature = "std")]
 use sp_std::ops::Deref;
 
+#[cfg(feature = "std")]
```
bkchr:
```suggestion
```

primitives/io/src/lib.rs
```diff
@@ -32,6 +32,9 @@ use sp_std::vec::Vec;
...
+use std::time;
```
bkchr:
```suggestion
```

primitives/io/src/lib.rs
```diff
@@ -32,6 +32,9 @@ use sp_std::vec::Vec;
...
+
```
bkchr:
```suggestion
```

primitives/io/src/lib.rs
```diff
@@ -370,6 +373,13 @@ pub trait Misc {
 			.call_in_wasm(wasm, "Core_version", &[], &mut ext)
 			.ok()
 	}
+
+	/// Get the current system time in nanoseconds.
+	fn system_time() -> u128 {
+		time::SystemTime::now().duration_since(time::SystemTime::UNIX_EPOCH)
```
bkchr:
```suggestion
		std::time::SystemTime::now().duration_since(time::SystemTime::UNIX_EPOCH)
```

primitives/io/src/lib.rs
```diff
@@ -370,6 +373,13 @@ pub trait Misc {
 			.call_in_wasm(wasm, "Core_version", &[], &mut ext)
 			.ok()
 	}
+
+	/// Get the current system time in nanoseconds.
+	fn system_time() -> u128 {
```
bkchr: Are we going to support the benchmark, whatever they called macros from within the runtime? Or is this just for the benchmark purposes?

Nevertheless, I want some big exclamation marks in the documentation of this function. The user needs to be aware that this should not be used for anything that touches the storage, as otherwise we get storage root mismatches.

bkchr:
```suggestion
	fn current_time() -> u128 {
```

pepyakin: If I understood correctly, @gavofyork asked for this function to be available only behind a flag and it seems here it is available always. Not sure how strong that request was but still worth asking I guess?

primitives/io/src/lib.rs
```diff
@@ -370,6 +373,13 @@ pub trait Misc {
 			.call_in_wasm(wasm, "Core_version", &[], &mut ext)
 			.ok()
 	}
+
+	/// Get the current system time in nanoseconds.
```
pepyakin:
```suggestion
	/// Get the number of nanoseconds passed since the UNIX epoch.
```

primitives/io/src/lib.rs
shawntabrizi: It should be for benchmarking purposes only. I can add more documentation.

primitives/io/src/lib.rs
pepyakin: However, I wonder why we use `SystemTime` here? System time is known to be non-monotonic and not consistent and usually is not used for benchmarking (which is the only purpose of this host function, right?). Can we use some other clock like uptime based or something similar?

primitives/io/src/lib.rs
```diff
@@ -916,6 +930,7 @@ pub type SubstrateHostFunctions = (
 	logging::HostFunctions,
 	sandbox::HostFunctions,
 	crate::trie::HostFunctions,
+	benchmarking::HostFunctions,
```
bkchr:
```suggestion
```

Should not be enabled by default.

primitives/io/src/lib.rs
shawntabrizi: Well, @pepyakin, we know that using [`arch::_rdtsc`](https://doc.rust-lang.org/core/arch/x86_64/fn._rdtsc.html) has its own issues only working on certain architectures.

The question really boils down to: is `SystemTime` good enough? I think we will need to use it to know?

I dont think anyone should feel this commit is us committing to anything. I just want to make this function available so others like @mattrutherford can also start doing some benchmarking work.

primitives/io/src/lib.rs
shawntabrizi: It is now disabled by default per the latest suggestion by Basti

primitives/io/src/lib.rs
pepyakin: Agreed, we can iterate later

shawntabrizi:
Because this is hidden in @pepyakin's resolved conversation, just wanted to say again here.

This is just one quick idea to start adding additional functionality to the runtime which enables us to benchmark operations.

This function is not intended to be exposed by default, and is not meant to be supported in runtime logic. It is simply meant to act as a way for us to time thing while _within_ wasm, but by running very specific benchmarking code which only executes while we are benchmarking.

We can iterate, remove, or whatever this later, but I wanted to merge this to master so that other parallel efforts for benchmarking can all create around a single API for now.

shawntabrizi:
Closing this for now in favor of using the existing offchain workers host function: https://github.com/paritytech/substrate/blob/8fe1161aacc13abda524adc4d9779b1c05522ae0/primitives/io/src/lib.rs#L582

burdges:
An actual time oracle for the runtime could probably be implemented via the relative time story, i.e. all relay chain block producers would declare their view of the time, and the relative time clock consensus system would prevent silly times from being built upon or finalized.

Issue #4696, opened 2020-01-21
Add Endpoint to Retrieve Info About Keystore


joepetrowski:
**Problem:** Validators often run multiple clients of which only one is their actual validator. For example, they may have several sentries for each validator, and multiple validators may share sentries.

Validators want a way to query their node's keystore to see if it contains certain keys. This will probably take the form of an `author` RPC endpoint.

**Input:** The session keys for a given validator. (Can be obtained with `nextKeys(stash: AccountId)`)

**Output:** `true` iff the node contains the private keys associated with the public session keys. `false` otherwise.

PR #4720, opened 2020-01-23
Introduces `author_hasKey` and `author_hasSessionKeys` rpc endpoints

bkchr:
Both endpoints can be used to check if a key is present in the keystore.

- `hasKey` works on with an individual public key and key type. It
checks if a private key for the given combination exists in the
keystore.
- `hasSessionKeys` works with the full encoded session key blob stored
on-chain in `nextKeys`. This requires that the given blob can be decoded
by the runtime. It will return `true`, iff all public keys of the
session key exist in the storage.

Fixes: https://github.com/paritytech/substrate/issues/4696


client/rpc-api/src/author/error.rs
```diff
@@ -54,6 +54,9 @@ pub enum Error {
 	/// Some random issue with the key store. Shouldn't happen.
 	#[display(fmt="The key store is unavailable")]
 	KeyStoreUnavailable,
+	/// Invalid session keys encoding.
+	#[display(fmt="The session keys are not encoded correctly")]
```
NikVolf:
```suggestion
	#[display(fmt="Session keys encoded incorrectly")]
```

client/rpc/src/author/mod.rs
```diff
@@ -112,6 +112,23 @@ where
 		).map(Into::into).map_err(|e| Error::Client(Box::new(e)))
 	}
 
+	fn has_session_keys(&self, session_keys: Bytes) -> Result<bool> {
+		let best_block_hash = self.client.chain_info().best_hash;
+		let keys = self.client.runtime_api().decode_session_keys(
+			&generic::BlockId::Hash(best_block_hash),
+			session_keys.to_vec(),
+		)
+		.map_err(|e| Error::Client(Box::new(e)))?
```
NikVolf: no identation

NikVolf:
This really can save some time bootstrapping node/network

client/keystore/src/lib.rs
```diff
@@ -97,25 +98,22 @@ impl Store {
...
 	}
 
 	/// Insert the given public/private key pair with the given key type.
 	///
 	/// Does not place it into the file system store.
-	fn insert_ephemeral_pair<Pair: PairT>(&mut self, pair: &Pair, key_type: KeyTypeId) {
+	fn insert_ephemeral_pair<Pair: PairT>(&mut self, pair: &Pair, seed: &str, key_type: KeyTypeId) {
 		let key = (key_type, pair.public().to_raw_vec());
-		self.additional.insert(key, pair.to_raw_vec());
+		self.additional.insert(key, seed.into());
```
tomusdrw: The docs for `additional` say it should be `raw public key`, but here it seems we just put `seed` in.

Also we can just put bulshit, cause we don't check that the `pair` is derived from that `seed`, do we?

bkchr: The docs say: `(KeyTypeId, Raw public key)` -> `Key phrase/seed`
                                      Key                            ->       Value

(the key is build above in exactly this way.

PR #4722, opened 2020-01-23
[contracts] Add ext_tombstone_deposit

Robbepop:
This add the `ext_tombstone_deposit` to the contract host API to make it possible for smart contracts to query the amount of balance they need to create a tombstone upon eventual eviction.

This API is required since without it users cannot determine precisely what `rent_allowance` they should minimal choose in order to be guaranteed to create a tombstone upon contract eviction.

### DIFF
--- a/frame/contracts/src/wasm/runtime.rs
+++ b/frame/contracts/src/wasm/runtime.rs
@@ -623,11 +623,17 @@ define_env!(Env, <E: Ext>,
 		Ok(())
 	},
 
-	// Stores the balance required to create a tombstone upon contract eviction
-	// into the scratch buffer.
+	// Stores the tombstone deposit into the scratch buffer.
 	//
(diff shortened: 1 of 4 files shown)


frame/contracts/src/wasm/runtime.rs
```diff
@@ -623,6 +623,17 @@ define_env!(Env, <E: Ext>,
 		Ok(())
 	},
 
+	// Stores the balance required to create a tombstone upon contract eviction
+	// into the scratch buffer.
```
pepyakin: We might want to expand that this deposit is on top of the existential deposit.
In other words, in order for a contract to leave a tombstone the balance of the contract should not go below the sum of existential deposit and the tombstone deposit. This sum is commonly referred as subsistence threshold in code.

Robbepop: done

PR #4724, opened 2020-01-23
Add child-trie roots to cache

cheme:
This PR add child trie root to storage cache.
Previously only the child trie value where cached and on cache miss we would have query for a the child trie root without using the cache resulting to multiple avoidable query of child trie root.

ChildInfo is stored in storage cache similarly to a storage key, but it does not use a lru cache but a simple hash map. The child info content is referenced through rc from every child storage item.
A better design would be to manage lru cache globally among hash maps and have a lrumap per child trie, storing the child info and lru map in a child trie parent map similarilly to what is done in state-machine (`HashMap<ChildKey, (LruMap<Key, Option<Value>>, Option<ChildInfo>)>`). This would mean implementing a lru shared among multiple map, which is out of the scope of this pr. 

@Stefie also find a bug or unexpected behavior where using a wrong child info on the rpc query could return the value of the right child info. This is rather harmless, but could lead to confusion.

Therefore this PR also add a validation of the child info (through a call to method 'valid_as') and return an empty result if the child info does not match the right one.

Child infos map get clean every 100 sync.

I also did change the way state overlay was storing cached children root, previously the calculated root were stored in top storage, this would have lead to problem in case we fetch root after this store.


cheme:
Putting back in progress, I am working on a different approach where state-machine manage multiple trie backend (one per child trie), which seems a better basis for the state overlay part of this pr.

gnunicorn:
Closing as part of a general PR clean up.

---

This PR hasn't seen activity in quite a while (>30 days), thus I am closing this for now. This is not a comment on the approach, code quality or whether we like it or not and it would get merged eventually, but it won't get merged as is and seems to be stale. You are welcome to reopen or raise another PR about this any time later.

Issue #4704, opened 2020-01-22
Emit even on runtime upgrade


xlc:
Can we have an event on `set_code` / `set_code_without_checks`? I would like to know in which block the new runtime was applied.

PR #4726, opened 2020-01-24
Deposit event on `frame-system::set_code`

bkchr:
Closes: https://github.com/paritytech/substrate/issues/4704


frame/system/src/lib.rs
```diff
@@ -1688,6 +1693,11 @@ mod tests {
 				RawOrigin::Root.into(),
 				substrate_test_runtime_client::runtime::WASM_BINARY.to_vec(),
 			).unwrap();
+
+			assert_eq!(
+				System::events()
```
thiolliere:
```suggestion
				System::events(),
```

PR #4730, opened 2020-01-24
Cursed implementation of allowing missing imports on wasmtime

pepyakin:
This is a quick and dirty implementation of the wasmtime counterpart of https://github.com/paritytech/substrate/pull/4550. This is aimed to unblock polkadot progress while we are waiting #4686.

The idea is that we employ parity-wasm to extract all the imports and do the resolving ourselves, then we collect all the missing function imports and create modules and stubs for them.

### DIFF
--- a/client/executor/wasmtime/src/runtime.rs
+++ b/client/executor/wasmtime/src/runtime.rs
@@ -140,10 +140,6 @@ impl MissingFunctionStubs {
 			sig,
 		});
 	}
-
-	fn stubs(&self, module: &str) -> &[MissingFunction] {
-		self.stubs.get(module).map(|stubs| &*stubs as &[_]).unwrap_or(&[])
-	}
 }
 
 fn scan_missing_functions(
@@ -153,7 +149,6 @@ fn scan_missing_functions(
 	let isa = target_isa()?;
 	let call_conv = isa.default_call_conv();
 
-	// TODO:
 	let module = parity_wasm::elements::Module::from_bytes(code)
 		.map_err(|e| WasmError::Other(format!("cannot deserialize error: {}", e)))?;
 
@@ -184,12 +179,10 @@ fn scan_missing_functions(
 		};
 
 		if import_entry.module() == "env" {
-			if host_functions
+			if let Some(hf) = host_functions
 				.iter()
 				.find(|hf| hf.name() == import_entry.field())
-				.is_some()
 			{
-				// TODO: Check the signature.
 				continue;
 			}
 		}
@@ -246,7 +239,7 @@ fn create_compiled_unit(
 	let compiler = new_compiler(compilation_strategy)?;
 
 	let mut missing_functions_stubs = if allow_missing_func_imports {
-		dbg!(scan_missing_functions(code, host_functions)?)
+		scan_missing_functions(code, host_functions)?
 	} else {
 		// If there are in fact missing functions they will be detected at the instantiation time
 		// and the module will be rejected.
@@ -256,13 +249,13 @@ fn create_compiled_unit(
 	let env_missing_functions = missing_functions_stubs.stubs.remove("env").unwrap_or_else(|| Vec::new());
 	context.name_instance(
 		"env".to_owned(),
-		instantiate_env_module(global_exports, compiler, host_functions, &env_missing_functions)?
+		instantiate_env_module(global_exports, compiler, host_functions, env_missing_functions)?
 	);
 
 	for (module, missing_functions_stubs) in missing_functions_stubs.stubs {
 		let compiler = new_compiler(compilation_strategy)?;
 		let global_exports = context.get_global_exports();
-		let instance = instantiate_env_module(global_exports, compiler, &[], &env_missing_functions)?;
+		let instance = instantiate_env_module(global_exports, compiler, &[], missing_functions_stubs)?;
 		context.name_instance(module, instance);
 	}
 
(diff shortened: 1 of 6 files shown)


client/executor/wasmtime/src/runtime.rs
```diff
@@ -115,9 +121,98 @@ pub fn create_instance(
...
+		let func_ty = match import_entry.external() {
+			parity_wasm::elements::External::Function(func_ty_idx) => {
+				let parity_wasm::elements::Type::Function(ref func_ty) =
+					types.get(*func_ty_idx as usize).ok_or_else(|| {
+						WasmError::Other(format!("corrupted module, type out of bounds"))
+					})?;
+				func_ty
+			}
+			_ => {
+				// The executor doesn't provide any non function imports!
```
bkchr: When we switch to import memory, will this not "break" here? Why not ignore non function imports?

client/executor/wasmtime/src/runtime.rs
```diff
@@ -115,9 +121,98 @@ pub fn create_instance(
...
+				return Err(WasmError::Other(format!(
+					"{}:{} is not a function",
+					import_entry.module(),
+					import_entry.field()
+				)));
+			}
+		};
+		let signature = convert_parity_wasm_signature(func_ty);
+
+		if import_entry.module() == "env" {
```
bkchr: Can we have function exports that don't have the `module` `env`?

client/executor/wasmtime/src/runtime.rs
```diff
@@ -115,9 +121,98 @@ pub fn create_instance(
...
+				.iter()
+				.find(|hf| hf.name() == import_entry.field())
+			{
+				if signature == hf.signature() {
+					continue;
+				}
+			}
+		}
+
+		// This import is not a function, not from env module, or doesn't have a corresponding host
```
bkchr:
```suggestion
		// This import is a function, not from env module, or doesn't have a corresponding host
```

client/executor/wasmtime/src/runtime.rs
```diff
@@ -115,9 +121,98 @@ pub fn create_instance(
...
+		// function, or the signature is not matching.
+		let sig = cranelift_ir_signature(signature, &call_conv);
+
+		missing_functions.insert(
+			import_entry.module().to_string(),
+			import_entry.field().to_string(),
+			sig
```
bkchr:
```suggestion
			sig,
```

client/executor/wasmtime/src/runtime.rs
```diff
@@ -130,8 +225,27 @@ fn create_compiled_unit(
...
+	} else {
+		// If there are in fact missing functions they will be detected at the instantiation time
+		// and the module will be rejected.
+		MissingFunctionStubs::new()
+	};
+
+	let env_missing_functions = missing_functions_stubs.stubs.remove("env").unwrap_or_else(|| Vec::new());
+	context.name_instance(
+		"env".to_owned(),
+		instantiate_env_module(global_exports, compiler, host_functions, env_missing_functions)?
```
bkchr:
```suggestion
		instantiate_env_module(global_exports, compiler, host_functions, env_missing_functions)?,
```

client/executor/wasmtime/src/runtime.rs
```diff
@@ -130,8 +225,27 @@ fn create_compiled_unit(
...
+
+	let mut missing_functions_stubs = if allow_missing_func_imports {
+		scan_missing_functions(code, host_functions)?
+	} else {
+		// If there are in fact missing functions they will be detected at the instantiation time
+		// and the module will be rejected.
+		MissingFunctionStubs::new()
+	};
+
+	let env_missing_functions = missing_functions_stubs.stubs.remove("env").unwrap_or_else(|| Vec::new());
```
bkchr: Can we not do this in the loop below and just check:
```
if module == env {
    host_functions
} else {
    &[]
}
```

client/executor/wasmtime/src/runtime.rs
```diff
@@ -231,6 +346,26 @@ fn instantiate_env_module(
 		finished_functions.push(trampoline);
 	}
 
+	for MissingFunction { name, sig } in missing_functions_stubs {
+		let sig = translate_signature(
+			sig,
+			pointer_type
```
bkchr:
```suggestion
			pointer_type,
```

client/executor/wasmtime/src/runtime.rs
pepyakin: Ah yeah you are right, I changed the condition but didn't change the comment.

client/executor/wasmtime/src/runtime.rs
pepyakin: Yeah, we can! If I remember correctly, in Rust you'd use `#[link(wasm_import_module = "non_env")]` for that.

client/executor/wasmtime/src/runtime.rs
pepyakin: The main point of doing that is not in host functions but rather in creating a new instance and giving it a name for `non-env` modules.

client/executor/wasmtime/src/runtime.rs
pepyakin: Yeah, it will bail here! We can remove this in rely it to fail at the instantiation time.

PR #4733, opened 2020-01-24
pallet-contracts: Refactor and comment `rent` module.

pepyakin:
This change disentangles the rent module and hopefully makes it easier to understand. The idea behind this change is to separate making a decision about rent with actually performing it. This separation will come in handy in the following PRs.

No functional changes are expected (no spec_version bump is required, only impl).


frame/contracts/src/rent.rs
```diff
@@ -14,64 +14,87 @@
...
+			ExistenceRequirement::KeepAlive,
+		) {
+			// This should never fail. However, let's err on the safe side.
+			T::RentPayment::on_unbalanced(imbalance);
+		}
+	}
+}
+
+enum Verdict<T: Trait> {
+	Exempt,
```
NikVolf: This one does not deserve doc comment?

pepyakin: Hah, before wrapping up I looked at it and thought that it's pretty clear.

Now, however, I don't think so )

frame/contracts/src/rent.rs
```diff
@@ -127,75 +147,140 @@ fn try_evict_or_and_pay_rent<T: Trait>(
...
+	let verdict = consider_case::<T>(
+		account,
+		current_block_number,
+		handicap,
+		&alive_contract_info,
+	);
+
+	// Enact the verdict only if the contract gets removed.
+	match verdict {
+		Verdict::Kill => {
```
NikVolf:
```suggestion
		Verdict::Kill | Verdict::Evict { .. } => {
```

frame/contracts/src/rent.rs
pepyakin: Done!

frame/contracts/src/rent.rs
pepyakin: Done! (Although I didn't use your suggestion since the arm must be removed as well)

bin/node/runtime/src/lib.rs
```diff
@@ -80,8 +80,8 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// and set impl_version to equal spec_version. If only runtime
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
-	spec_version: 205,
-	impl_version: 205,
+  spec_version: 205,
```
pepyakin: Github merge-conflict editor changes tabs to spaces?

frame/contracts/src/rent.rs
```diff
@@ -127,75 +151,136 @@ fn try_evict_or_and_pay_rent<T: Trait>(
...
+///
+/// `alive_contract_info` should be from the same address as `account`.
+fn enact_verdict<T: Trait>(
+	account: &T::AccountId,
+	alive_contract_info: AliveContractInfo<T>,
+	current_block_number: T::BlockNumber,
+	verdict: Verdict<T>,
+) -> Option<ContractInfo<T>> {
+	match verdict {
+		Verdict::Exempt => return Some(ContractInfo::Alive(alive_contract_info)),
```
thiolliere: I just realized that now but if a contract is exempted then its deduct block doesn't get updated. Thus if one contract is exempted because it has enough funds to cover its storage and then suddenly the fee_per_block is more than 0 then it will have to pay the rent of this from the last time he actually wasn't exempted up to the current block. I think this is wrong.

bin/node/runtime/src/lib.rs
thiolliere:
```suggestion
	spec_version: 205,
```

frame/contracts/src/rent.rs
pepyakin: Good catch!

I think we need to add some tests directly in this module to exercise conditions like this...

frame/contracts/src/rent.rs
```diff
@@ -14,64 +14,91 @@
...
+	/// For example, it already paid its rent in the current block, or it has enough deposit for not
+	/// paying rent at all.
+	Exempt,
+	/// Funds dropped below the subsistence deposit.
+	///
+	/// Remove the contract along with it's storage.
+	Kill,
+	/// The contract cannot afford payment within its rent budget so it gets evicted. However,
+	/// because its balance is greater than the subsistence threshold it leaves a tombstone.
+	Evict {
```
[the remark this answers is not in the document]
pepyakin: Yes, they would.

I think the situation is a bit different from Ethereum:

1. tombstone is small and of fixed size,
1. we assume inflationary model. That means that having a deposit is not free, inflation will destroy some value with time.
1. in Ethereum one cannot do much with refunds in gas (not supposed to? modulo GasToken).

frame/contracts/src/rent.rs
```diff
@@ -14,64 +14,91 @@
...
+			T::RentPayment::on_unbalanced(imbalance);
+		}
+	}
+}
+
+enum Verdict<T: Trait> {
+	/// The contract is exempted from paying rent.
+	///
+	/// For example, it already paid its rent in the current block, or it has enough deposit for not
+	/// paying rent at all.
```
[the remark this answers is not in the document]
pepyakin: Yep, exactly. This is a mechanism for making lives of socially valuable contracts easier. However, it is not free-free. Check out my comment on the tombstones.

frame/contracts/src/rent.rs
```diff
@@ -127,75 +151,136 @@ fn try_evict_or_and_pay_rent<T: Trait>(
...
-		});
-
-		<ContractInfoOf<T>>::insert(account, &contract_info);
+			<Module<T>>::deposit_event(RawEvent::Evicted(account.clone(), true));
+			Some(tombstone_info)
+		}
+		Verdict::Charge { amount } => {
+			let contract_info = ContractInfo::Alive(AliveContractInfo::<T> {
+				rent_allowance: alive_contract_info.rent_allowance - amount.peek(),
+				deduct_block: current_block_number,
```
thiolliere: Hmm also this is tricky, if consider the case for a past block (like we do when there is handicap) then the verdict is for a past block. But the code is fine as the charge is never done for a case considered in the past.

frame/contracts/src/rent.rs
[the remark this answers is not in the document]
Robbepop: We could theoretically even collect taxes of storage fee for tombstone contracts, right?
The fee would definitely be small enough for them to potentially last for an extended amount of time (e.g. weeks or month) with very small additional offset above existential deposit + tombstone deposit.

PR #4734, opened 2020-01-24
Transaction queue tests update 

NikVolf:
This also adds some tests for maintaince


client/transaction-pool/src/tests/mod.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018-2020 Parity Technologies (UK) Ltd.
+// Copyright 2020 Parity Technologies (UK) Ltd.
```
tomusdrw: I think should stay 2018-2020, but I don't care enough :)

```suggestion
// Copyright 2018-2020 Parity Technologies (UK) Ltd.
```

client/transaction-pool/src/tests/api.rs
```diff
@@ -0,0 +1,178 @@
...
+pub struct ChainState {
+	pub block_by_number: HashMap<BlockNumber, Vec<Extrinsic>>,
+	pub block_by_hash: HashMap<Hash, Vec<Extrinsic>>,
+	pub header_by_number: HashMap<BlockNumber, Header>,
+	pub nonces: HashMap<AccountId, u64>,
+	pub invalid_hashes: HashSet<Hash>,
+}
+
+pub struct TestApi {
+	pub valid_modifier: RwLock<Box<dyn Fn(&mut ValidTransaction) + Send + Sync>>,
```
tomusdrw: Maybe better add a method to alter that and keep all the fields private? Now it's a bit inconsistent.

client/transaction-pool/src/tests/api.rs
```diff
@@ -0,0 +1,178 @@
...
+	chain: RwLock<ChainState>,
+	validation_requests: RwLock<Vec<Extrinsic>>,
+}
+
+impl TestApi {
+	pub fn default() -> Self {
```
tomusdrw: Should be `Default` impl, no?

client/transaction-pool/src/tests/api.rs
```diff
@@ -0,0 +1,178 @@
...
+		self.chain.write().invalid_hashes.insert(
+			Self::hash_and_length_inner(xts).0
+		);
+	}
+
+	pub fn validation_requests(&self) -> Vec<Extrinsic> {
+		self.validation_requests.read().clone()
+	}
+
+	pub fn inc_nonce(&self, account: AccountId) {
```
tomusdrw: I think I'd prefer `increment_nonce`, but not super strong on it.

client/transaction-pool/src/tests/api.rs
NikVolf: I renamed it to empty, there was a semantic change there also

client/transaction-pool/src/tests/mod.rs
NikVolf: It's just GitHub thinks the file moved, while it is in fact a new file completely

PR #4737, opened 2020-01-26
WASM runtime switch to import memory

bkchr:
Up to now runtimes have exported their memory. To unify it with
sandboxing, this pr switches runtimes to import memory as well.

From a functional perspective, exporting/importing memory makes no
difference to the runtime.

To provide backwards compatibility, WASM exported memory is still supported.


client/executor/runtime-test/src/lib.rs
```diff
@@ -56,7 +61,7 @@ sp_core::wasm_export_functions! {
 
 	fn test_empty_return() {}
 
-	fn test_exhaust_heap() -> Vec<u8> { Vec::with_capacity(16777216) }
+	fn test_exhaust_heap() -> Vec<u8> { Vec::with_capacity(18 * 64 * 1024) }
```
pepyakin: The new value is `1179648`. I wonder why would we need such change?

primitives/runtime-interface/test/src/lib.rs
```diff
@@ -41,7 +41,7 @@ fn call_wasm_method<HF: HostFunctionsT>(method: &str) -> TestExternalities {
 		sc_executor::WasmExecutionMethod::Interpreted,
 		&mut ext_ext,
 		&WASM_BINARY[..],
-		8,
+		1024,
```
pepyakin: I wonder why we need this change?

client/executor/wasmtime/src/runtime.rs
```diff
@@ -32,42 +32,38 @@ use sc_executor_common::{
...
 impl WasmRuntime for WasmtimeRuntime {
 	fn update_heap_pages(&mut self, heap_pages: u64) -> bool {
-		match heap_pages_valid(heap_pages, self.max_heap_pages) {
-			Some(heap_pages) => {
-				self.heap_pages = heap_pages;
-				true
-			}
-			None => false,
-		}
+		self.heap_pages as u64 == heap_pages
```
pepyakin: I wish I landed the chance that removes `update_heap_pages` separately...

bkchr: Yeah, we can do this later :)

client/executor/runtime-test/src/lib.rs
bkchr: Ahh, I can probably revert this. Do you remember this question I had about initial pages?

I first had an implementation that used memory pages == heap pages, but now it is memory pages == intial + heap pages.

Tdlr, I revert this.

primitives/runtime-interface/test/src/lib.rs
bkchr: Some misunderstanding from me, I revert this as well.

client/executor/wasmtime/src/runtime.rs
pepyakin: Yep, sure, I still hope my PR lands eventually lol

PR #4740, opened 2020-01-27
Adds fork-awareness and finalization notifications to transaction pool watchers.

### LINKED ISSUE #4438: Transaction Status should differentiate between InBlock and Finalized.

seunlanlege:
This PR augments `MaintainedTransactionPool` trait to take a `ChainEvents` enum for passing finalization notifications. This, in turn, enables the `watch_extrinsic` rpc method to track actual finalization events emitted by grandpa.

closes #4438

### DIFF
--- a/client/finality-grandpa/src/observer.rs
+++ b/client/finality-grandpa/src/observer.rs
@@ -57,7 +57,6 @@ impl<'a, Block: BlockT, B, E, RA> finality_grandpa::Chain<Block::Hash, NumberFor
 	}
 }
 
-/// TODO: augument this fn to take an Option<Sink> and an exectutor for sending out notifications
 fn grandpa_observer<B, E, Block: BlockT, RA, S, F>(
 	client: &Arc<Client<B, E, Block, RA>>,
 	authority_set: &SharedAuthoritySet<Block::Hash, NumberFor<Block>>,
@@ -109,7 +108,6 @@ fn grandpa_observer<B, E, Block: BlockT, RA, S, F>(
 		};
 
 		if let Some(_) = validation_result.ghost() {
-			// TODO: send finalization notifications here!
 			let finalized_hash = commit.target_hash;
 			let finalized_number = commit.target_number;
 
--- a/client/finality-grandpa/src/observer.rs
+++ b/client/finality-grandpa/src/observer.rs
@@ -57,6 +57,7 @@ impl<'a, Block: BlockT, B, E, RA> finality_grandpa::Chain<Block::Hash, NumberFor
 	}
 }
 
+/// TODO: augument this fn to take an Option<Sink> and an exectutor for sending out notifications
 fn grandpa_observer<B, E, Block: BlockT, RA, S, F>(
 	client: &Arc<Client<B, E, Block, RA>>,
 	authority_set: &SharedAuthoritySet<Block::Hash, NumberFor<Block>>,
@@ -108,6 +109,7 @@ fn grandpa_observer<B, E, Block: BlockT, RA, S, F>(
 		};
 
 		if let Some(_) = validation_result.ghost() {
+			// TODO: send finalization notifications here!
 			let finalized_hash = commit.target_hash;
 			let finalized_number = commit.target_number;
 
--- a/client/service/src/builder.rs
+++ b/client/service/src/builder.rs
@@ -883,10 +883,11 @@ ServiceBuilder<
 			let is_validator = config.roles.is_authority();
 
 			let (import_stream, finality_stream) = (
-				client.import_notification_stream().map(|n| ChainEvent::Canonical {
+				client.import_notification_stream().map(|n| ChainEvent::NewBlock {
 					id: BlockId::Hash(n.hash),
 					header: n.header,
 					retracted: n.retracted,
+					is_new_best: n.is_new_best,
 				}),
 				client.finality_notification_stream().map(|n| ChainEvent::Finalized {
 					hash: n.hash
@@ -899,9 +900,9 @@ ServiceBuilder<
 						ChainEvent::NewBlock { ref header, is_new_best, .. } => {
 							let offchain = offchain.as_ref().and_then(|o| o.upgrade());
 							match offchain {
-								Some(offchain) if notification.is_new_best => {
+								Some(offchain) if is_new_best => {
 									let future = offchain.on_block_imported(
-										&notification.header,
+										&header,
 										network_state_info.clone(),
 										is_validator,
 									);
@@ -913,7 +914,7 @@ ServiceBuilder<
 								Some(_) => log::debug!(
 									target: "sc_offchain",
 									"Skipping offchain workers for non-canon block: {:?}",
-									notification.header,
+									header,
 								),
 								_ => {},
 							}
--- a/client/service/src/builder.rs
+++ b/client/service/src/builder.rs
@@ -896,30 +896,27 @@ ServiceBuilder<
 			let events = futures::stream::select(import_stream, finality_stream)
 				.for_each(move |event| {
 					// offchain worker is only interested in block import events
-					match event {
-						ChainEvent::NewBlock { ref header, is_new_best, .. } => {
-							let offchain = offchain.as_ref().and_then(|o| o.upgrade());
-							match offchain {
-								Some(offchain) if is_new_best => {
-									let future = offchain.on_block_imported(
-										&header,
-										network_state_info.clone(),
-										is_validator,
-									);
-									let _ = to_spawn_tx_.unbounded_send((
-										Box::pin(future),
-										From::from("offchain-on-block"),
-									));
-								},
-								Some(_) => log::debug!(
+					if let ChainEvent::NewBlock { ref header, is_new_best, .. } = event {
+						let offchain = offchain.as_ref().and_then(|o| o.upgrade());
+						match offchain {
+							Some(offchain) if is_new_best => {
+								let future = offchain.on_block_imported(
+									&header,
+									network_state_info.clone(),
+									is_validator,
+								);
+								let _ = to_spawn_tx_.unbounded_send((
+									Box::pin(future),
+									From::from("offchain-on-block"),
+								));
+							},
+							Some(_) => log::debug!(
 									target: "sc_offchain",
 									"Skipping offchain workers for non-canon block: {:?}",
 									header,
 								),
-								_ => {},
-							}
-						},
-						_ => {}
+							_ => {},
+						}
 					};
 
 					let txpool = txpool.upgrade();
--- a/client/service/src/builder.rs
+++ b/client/service/src/builder.rs
@@ -886,7 +886,6 @@ ServiceBuilder<
 				client.import_notification_stream().map(|n| ChainEvent::Canonical {
 					id: BlockId::Hash(n.hash),
 					header: n.header,
-					is_new_best: n.is_new_best,
 					retracted: n.retracted,
 				}),
 				client.finality_notification_stream().map(|n| ChainEvent::Finalized {
@@ -895,41 +894,40 @@ ServiceBuilder<
 			);
 			let events = futures::stream::select(import_stream, finality_stream)
 				.for_each(move |event| {
-					let txpool = txpool.upgrade();
+					// offchain worker is only interested in block import events
+					match event {
+						ChainEvent::NewBlock { ref header, is_new_best, .. } => {
+							let offchain = offchain.as_ref().and_then(|o| o.upgrade());
+							match offchain {
+								Some(offchain) if notification.is_new_best => {
+									let future = offchain.on_block_imported(
+										&notification.header,
+										network_state_info.clone(),
+										is_validator,
+									);
+									let _ = to_spawn_tx_.unbounded_send((
+										Box::pin(future),
+										From::from("offchain-on-block"),
+									));
+								},
+								Some(_) => log::debug!(
+									target: "sc_offchain",
+									"Skipping offchain workers for non-canon block: {:?}",
+									notification.header,
+								),
+								_ => {},
+							}
+						},
+						_ => {}
+					};
 
+					let txpool = txpool.upgrade();
 					if let Some(txpool) = txpool.as_ref() {
-						let future = txpool.maintain(&event);
+						let future = txpool.maintain(event);
 						let _ = to_spawn_tx_.unbounded_send((
 							Box::pin(future),
 							From::from("txpool-maintain")
 						));
-					}
-
-					// offchain worker is only interested in block import events
-					let (header, is_new_best) = match event {
-						ChainEvent::Canonical { header, is_new_best, .. } => (header, is_new_best),
-						_ => return ready(())
-					};
-
-					let offchain = offchain.as_ref().and_then(|o| o.upgrade());
-					match offchain {
-						Some(offchain) if is_new_best => {
-							let future = offchain.on_block_imported(
-								&header,
-								network_state_info.clone(),
-								is_validator,
-							);
-							let _ = to_spawn_tx_.unbounded_send((
-									Box::pin(future),
-									From::from("offchain-on-block"),
-							));
-						},
-						Some(_) => log::debug!(
-							target: "sc_offchain",
-							"Skipping offchain workers for non-canon block: {:?}",
-							notification.header,
-						),
-						_ => {},
 					}
 
 					ready(())
--- a/client/service/src/builder.rs
+++ b/client/service/src/builder.rs
@@ -51,7 +51,7 @@ use std::{
 use wasm_timer::SystemTime;
 use sysinfo::{get_current_pid, ProcessExt, System, SystemExt};
 use sc_telemetry::{telemetry, SUBSTRATE_INFO};
-use sp_transaction_pool::MaintainedTransactionPool;
+use sp_transaction_pool::{MaintainedTransactionPool, ChainEvent};
 use sp_blockchain;
 use grafana_data_source::{self, record_metrics};
 
@@ -882,26 +882,40 @@ ServiceBuilder<
 			let network_state_info: Arc<dyn NetworkStateInfo + Send + Sync> = network.clone();
 			let is_validator = config.roles.is_authority();
 
-			let events = client.import_notification_stream()
-				.for_each(move |notification| {
+			let (import_stream, finality_stream) = (
+				client.import_notification_stream().map(|n| ChainEvent::Canonical {
+					id: BlockId::Hash(n.hash),
+					header: n.header,
+					is_new_best: n.is_new_best,
+					retracted: n.retracted,
+				}),
+				client.finality_notification_stream().map(|n| ChainEvent::Finalized {
+					hash: n.hash
+				})
+			);
+			let events = futures::stream::select(import_stream, finality_stream)
+				.for_each(move |event| {
 					let txpool = txpool.upgrade();
 
 					if let Some(txpool) = txpool.as_ref() {
-						let future = txpool.maintain(
-							&BlockId::hash(notification.hash),
-							&notification.retracted,
-						);
+						let future = txpool.maintain(&event);
 						let _ = to_spawn_tx_.unbounded_send((
 							Box::pin(future),
 							From::from("txpool-maintain")
 						));
 					}
 
+					// offchain worker is only interested in block import events
+					let (header, is_new_best) = match event {
+						ChainEvent::Canonical { header, is_new_best, .. } => (header, is_new_best),
+						_ => return ready(())
+					};
+
 					let offchain = offchain.as_ref().and_then(|o| o.upgrade());
 					match offchain {
-						Some(offchain) if notification.is_new_best => {
+						Some(offchain) if is_new_best => {
 							let future = offchain.on_block_imported(
-								&notification.header,
+								&header,
 								network_state_info.clone(),
 								is_validator,
 							);
--- a/client/transaction-pool/graph/src/listener.rs
+++ b/client/transaction-pool/graph/src/listener.rs
@@ -17,29 +17,30 @@
 
 use std::{
 	collections::HashMap,
-	fmt,
 	hash,
 };
 use serde::Serialize;
-use crate::watcher;
-use sp_runtime::traits;
+use crate::{watcher, ChainApi, BlockHash};
 use log::{debug, trace, warn};
+use sp_runtime::traits;
 
 /// Extrinsic pool default listener.
-pub struct Listener<H: hash::Hash + Eq, H2> {
-	watchers: HashMap<H, watcher::Sender<H, H2>>
+pub struct Listener<H: hash::Hash + Eq, C: ChainApi> {
+	watchers: HashMap<H, watcher::Sender<H, BlockHash<C>>>,
+	finality_watchers: HashMap<BlockHash<C>, Vec<watcher::Sender<H, BlockHash<C>>>>,
 }
 
-impl<H: hash::Hash + Eq, H2> Default for Listener<H, H2> {
+impl<H: hash::Hash + Eq, C: ChainApi> Default for Listener<H, C> {
 	fn default() -> Self {
 		Listener {
 			watchers: Default::default(),
+			finality_watchers: Default::default(),
 		}
 	}
 }
 
-impl<H: hash::Hash + traits::Member + Serialize, H2: Clone + fmt::Debug> Listener<H, H2> {
-	fn fire<F>(&mut self, hash: &H, fun: F) where F: FnOnce(&mut watcher::Sender<H, H2>) {
+impl<H: hash::Hash + traits::Member + Serialize, C: ChainApi> Listener<H, C> {
+	fn fire<F>(&mut self, hash: &H, fun: F) where F: FnOnce(&mut watcher::Sender<H, BlockHash<C>>) {
 		let clean = if let Some(h) = self.watchers.get_mut(hash) {
 			fun(h);
 			h.is_done()
@@ -55,7 +56,7 @@ impl<H: hash::Hash + traits::Member + Serialize, H2: Clone + fmt::Debug> Listene
 	/// Creates a new watcher for given verified extrinsic.
 	///
 	/// The watcher can be used to subscribe to lifecycle events of that extrinsic.
-	pub fn create_watcher(&mut self, hash: H) -> watcher::Watcher<H, H2> {
+	pub fn create_watcher(&mut self, hash: H) -> watcher::Watcher<H, BlockHash<C>> {
 		let sender = self.watchers.entry(hash.clone()).or_insert_with(watcher::Sender::default);
 		sender.new_watcher(hash)
 	}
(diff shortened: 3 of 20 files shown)


primitives/runtime/src/traits.rs
```diff
@@ -614,6 +614,8 @@ pub type HasherFor<B> = <HashFor<B> as Hash>::Hasher;
 pub type HashFor<B> = <<B as Block>::Header as Header>::Hashing;
 /// Extract the number type for a block.
 pub type NumberFor<B> = <<B as Block>::Header as Header>::Number;
+/// Extract the hash type for a block
+pub type BlockHashFor<B> = <B as Block>::Hash;
```
NikVolf: @andresilva, and you said it is not needed

NikVolf:
Why don't we reuse maintaince api and pass finalisation events there?

seunlanlege:
> Why don't we reuse maintaince api and pass finalisation events there?

not entirely sure how that would work, can you elaborate?

NikVolf:
> > Why don't we reuse maintaince api and pass finalisation events there?
> 
> not entirely sure how that would work, can you elaborate?

We just extend `maintain` function arguments to accept enum here 

https://github.com/paritytech/substrate/blob/8bb1811e31e000ec442a8f6ddc1d51ca51fa4e2f/client/transaction-pool/src/lib.rs#L292

(what is now passed becomes `ChainEvent::Canonical(id, retracted)`, then we also add `ChainEvent::Finalized(id)`)

and handle latter in `maintain` function body
and we won't need dedicated thread for just fetching and feeding finalisation events to the transaction pool

@tomusdrw might have an opinion about that also

tomusdrw:
I like @NikVolf's approach a lot, we really shouldn't introduce any threads to transaction pool, piggybacking on the `maintain` method makes a lot of sense and the finality notification is simply an async stream, so instead of spawning any threads we should just spawn a background task on the runtime, but I think this can be avoided too and we can just re-use the same background task we have for block notifications (possibly doing a `select!` there).

client/transaction-pool/src/lib.rs
```diff
@@ -267,95 +265,107 @@ impl<N: Clone + Copy + SimpleArithmetic> RevalidationStatus<N> {
...
-
-			if next_action.revalidate {
-				if let Err(e) = pool.revalidate_ready(&id, next_action.revalidate_amount).await {
-					log::warn!("Revalidate ready failed {:?}", e);
-				}
+			ChainEvent::Finalized { hash } => {
+				let hash = hash.clone();
+				let pool = self.pool().clone();
+				async move {
+					pool.finalized(&hash)
```
NikVolf: but this is sync method, why you even need async block and cloning here?
`just do Box::pin(futures::future::Ready(()))` ?

client/transaction-pool/src/lib.rs
```diff
@@ -267,95 +265,107 @@ impl<N: Clone + Copy + SimpleArithmetic> RevalidationStatus<N> {
...
-						.filter(|tx| tx.is_signed().unwrap_or(true));
-
-					resubmit_transactions.extend(block_transactions);
-				}
-				if let Err(e) = pool.submit_at(&id, resubmit_transactions, true).await {
-					log::debug!(target: "txpool",
-						"[{:?}] Error re-submitting transactions: {:?}", id, e
-					)
-				}
+	fn maintain(&self, event: &ChainEvent<Self>) -> Pin<Box<dyn Future<Output=()> + Send>> {
```
NikVolf: maybe pass `event` by-value?

client/service/src/builder.rs
```diff
@@ -874,22 +874,35 @@ ServiceBuilder<
...
+			let events = futures::stream::select(import_stream, finality_stream)
+				.for_each(move |event| {
 					let txpool = txpool.upgrade();
 
 					if let Some(txpool) = txpool.as_ref() {
-						let future = txpool.maintain(
-							&BlockId::hash(notification.hash),
-							&notification.retracted,
-						);
+						let future = txpool.maintain(&event);
```
seunlanlege: @NikVolf can't pass by value because the event will be reused by the off-chain worker.

NikVolf: If yo want, you can do (since you reuse header only)
```rust
enum ChainEvent<'a> { 
  Canonical: { 
     header: &'a Header, 
     ... 
  },
  Finalized { ... }
}
```

seunlanlege: So previously the trait took references to its arguments, I think there's nothing wrong with leaving it as it is. If other systems want to consume the `ChainEvent` stream they would also take references instead of by value which would lead to unnecessary cloning.

seunlanlege:
I would like to write an integration test for this using manual-seal, are there any examples of the `ServiceBuilder` used in tests?

client/transaction-pool/src/lib.rs
```diff
@@ -261,95 +259,104 @@ impl<N: Clone + Copy + SimpleArithmetic> RevalidationStatus<N> {
...
+		match event {
+			ChainEvent::Canonical {
```
tomusdrw: Please keep it in one line to avoid indenting the main part of the code so much, i.e.
```rust
match event {
  ChainEvent::Canonical { id, retracted, ... } => {
    let id = id.clone();
   }  
}
```

client/transaction-pool/src/testing/pool.rs
```diff
@@ -157,8 +167,20 @@ fn should_prune_old_during_maintenance() {
 	assert_eq!(pool.status().ready, 1);
 
 	pool.api.push_block(1, vec![xt.clone()]);
-
-	block_on(pool.maintain(&BlockId::number(1), &[]));
+	let header = Header {
```
tomusdrw: maybe use the `header` helper function added above?

primitives/transaction-pool/src/pool.rs
```diff
@@ -217,11 +219,29 @@ pub trait TransactionPool: Send + Sync {
 	fn ready_transaction(&self, hash: &TxHash<Self>) -> Option<Arc<Self::InPoolTransaction>>;
 }
 
+/// Events that the transaction pool listens for.
+pub enum ChainEvent<TP: TransactionPool> {
+	/// New blocks have been added to the chain
+	Canonical {
```
tomusdrw: I think it should be just called `NewBlock`, as afair it gets triggered for every block (not only canonical ones). The doc above seems to confirm that as well.

primitives/runtime/src/traits.rs
tomusdrw: Isn't that the same as `HashFor`?

primitives/transaction-pool/src/pool.rs
```diff
@@ -100,6 +100,8 @@ pub enum TransactionStatus<Hash, BlockHash> {
 	/// Transaction has been included in block with given hash.
 	#[serde(rename = "finalized")] // See #4438
 	InBlock(BlockHash),
+	/// Transaction has been finalized by a finality-gadget, e.g GRANDPA
```
tomusdrw: I think it doesn't have to be necessarily finality gadget (as the consensus might have finality builtin and not rely on a gadget), but the comment is fine, as the majority of the cases it's going to be a gadget.

primitives/transaction-pool/src/pool.rs
```diff
@@ -122,7 +124,7 @@ pub type BlockHash<P> = <<P as TransactionPool>::Block as BlockT>::Hash;
 /// Transaction type for a pool.
 pub type TransactionFor<P> = <<P as TransactionPool>::Block as BlockT>::Extrinsic;
 /// Type of transactions event stream for a pool.
-pub type TransactionStatusStreamFor<P> = TransactionStatusStream<TxHash<P>, BlockHash<P>>;
+pub type TransactionStatusStreamFor<P, BH> = TransactionStatusStream<TxHash<P>, BH>;
```
tomusdrw: How is that better?

primitives/transaction-pool/src/pool.rs
```diff
@@ -217,11 +219,29 @@ pub trait TransactionPool: Send + Sync {
...
+}
+
 /// Trait for transaction pool maintenance.
-pub trait MaintainedTransactionPool : TransactionPool {
+pub trait MaintainedTransactionPool: TransactionPool {
 	/// Perform maintenance
-	fn maintain(&self, block: &BlockId<Self::Block>, retracted: &[BlockHash<Self>])
-		-> Pin<Box<dyn Future<Output=()> + Send>>;
+	fn maintain(&self, event: &ChainEvent<Self>) -> Pin<Box<dyn Future<Output=()> + Send>>
+		where Self: Sized;
```
tomusdrw: Why do you need `Self: Sized`?

client/transaction-pool/src/testing/pool.rs
```diff
@@ -173,9 +195,21 @@ fn should_revalidate_during_maintenance() {
 	assert_eq!(pool.status().ready, 2);
 	assert_eq!(pool.api.validation_requests().len(), 2);
 
+	let header = Header {
```
tomusdrw: Same here, consider using the helper function you just added.

client/finality-grandpa/src/observer.rs
```diff
@@ -57,6 +57,7 @@ impl<'a, Block: BlockT, B, E, RA> finality_grandpa::Chain<Block::Hash, NumberFor
 	}
 }
 
+/// TODO: augument this fn to take an Option<Sink> and an exectutor for sending out notifications
```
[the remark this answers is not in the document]
tomusdrw: Or is it really needed? Isn't `finalization_stream` from client not enough?

client/transaction-pool/graph/src/listener.rs
```diff
@@ -101,8 +102,24 @@ impl<H: hash::Hash + traits::Member + Serialize, H2: Clone + fmt::Debug> Listene
 	}
 
 	/// Transaction was pruned from the pool.
-	pub fn pruned(&mut self, header_hash: H2, tx: &H) {
-		debug!(target: "txpool", "[{:?}] Pruned at {:?}", tx, header_hash);
-		self.fire(tx, |watcher| watcher.in_block(header_hash))
+	pub fn pruned(&mut self, block_hash: BlockHash<C>, tx: &H) {
+		debug!(target: "txpool", "[{:?}] Pruned at {:?}", tx, block_hash);
+		self.fire(tx, |watcher| watcher.in_block(block_hash.clone()));
```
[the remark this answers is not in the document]
tomusdrw: Okay, I just saw that you've changed that `InBlock` does not cause the watcher to return `true` in `is_done` function, which I believe is ok, but to be honest I wouldn't mind simplifying to the above code that I proposed.

primitives/transaction-pool/src/pool.rs
seunlanlege: You get a compiler error without it. I think because `ChainEvent<Self>` may potentially be a recursive type, adding that the trait has a known size at compile time prevents this.

primitives/runtime/src/traits.rs
NikVolf: It is the same as `B::Hash` actually

client/transaction-pool/graph/src/pool.rs
```diff
@@ -88,6 +88,15 @@ pub trait ChainApi: Send + Sync {
...
 	fn block_body(&self, at: &BlockId<Self::Block>) -> Self::BodyFuture;
+
+	/// Returns block header
+	fn block_header(&self, at: BlockId<Self::Block>) -> Result<
+		Option<<Self::Block as traits::Block>::Header>,
+		Self::Error
+	>;
+
+	/// Returns the hash of the last finalized block
+	fn last_finalized(&self) -> BlockHash<Self>;
```
tomusdrw: Seems unused?

client/transaction-pool/graph/src/pool.rs
```diff
@@ -438,6 +447,16 @@ impl<B: ChainApi> Pool<B> {
 		(hash, validity)
 	}
 
+	/// Notify all watchers that transactions in the block with hash been finalized
```
```suggestion
	/// Notify all watchers that transactions in the block with given hash have been finalized
```

client/transaction-pool/graph/src/pool.rs
```diff
@@ -438,6 +447,16 @@ impl<B: ChainApi> Pool<B> {
...
+	pub async fn finalized(&self, block_hash: BlockHash<B>) -> Result<(), B::Error> {
```
tomusdrw: There is already a method called `on_broadcasted`, I think it would be good to keep the convention and call this one `on_block_finalized`
```suggestion
	pub async fn on_block_finalized(&self, block_hash: BlockHash<B>) -> Result<(), B::Error> {
```

Basically to separate functions that control internal state of the pool from functions that are just handlers for some external events.

client/transaction-pool/graph/src/pool.rs
```diff
@@ -438,6 +447,16 @@ impl<B: ChainApi> Pool<B> {
...
+		self.validated_pool.finalized(block_hash).await
+	}
+
+	/// Notify the listener of retracted blocks
+	pub fn retracted(&self, block_hash: &BlockHash<B>) {
```
tomusdrw: Same here.

```suggestion
	pub fn on_block_retracted(&self, block_hash: &BlockHash<B>) {
```

client/transaction-pool/graph/src/pool.rs
```diff
@@ -438,6 +447,16 @@ impl<B: ChainApi> Pool<B> {
...
+		self.validated_pool.retracted(block_hash)
+	}
+
 	/// Get ready transaction by hash, if it present in the pool.
 	pub fn ready_transaction(&self, hash: &ExHash<B>) -> Option<TransactionFor<B>> {
 		self.validated_pool.ready_by_hash(hash)
```
tomusdrw: I really start to hate the fact that we have to duplicate and delegate so many methods. Perhaps better to just expose `validated_pool` via a getter and call the methods directly?
```
pub fn validated_pool(&self) -> &ValidatedPool {
  &self.validated_pool
}
```
and call via:
```
pool.validated_pool().ready_transaction(hash)
```

client/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -65,11 +64,12 @@ pub type ValidatedTransactionFor<B> = ValidatedTransaction<
 pub(crate) struct ValidatedPool<B: ChainApi> {
 	api: Arc<B>,
 	options: Options,
-	listener: RwLock<Listener<ExHash<B>, BlockHash<B>>>,
+	listener: RwLock<Listener<ExHash<B>, B>>,
 	pool: RwLock<base::BasePool<
 		ExHash<B>,
 		ExtrinsicFor<B>,
 	>>,
+	last_finalized_hash: Arc<Mutex<BlockHash<B>>>,
```
tomusdrw: Why is it an `Arc<Mutex` and not just `RwLock`?

client/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -501,14 +502,53 @@ impl<B: ChainApi> ValidatedPool<B> {
...
+			if last_finalized_hash == Default::default() {
+				// self.last_finalized_hash is created with a default value
+				break
+			}
+
+			header = self.api.block_header(BlockId::Hash(header.parent_hash().clone()))?
+				.expect("parent block should have block header; qed");
+		}
+
+		*self.last_finalized_hash.lock() = block_hash;
```
tomusdrw: This creates a race condition. It would probably best to fire this kind of notifications from a single-threaded environment or at least we need to fix the logic here.

client/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -501,14 +502,53 @@ impl<B: ChainApi> ValidatedPool<B> {
...
+			let tx_hashes = self.api.block_body(&BlockId::Hash(hash.clone())).await?
+				.expect("fetched block header should have block body; qed")
+				.into_iter()
+				.map(|tx| self.api.hash_and_length(&tx).0)
+				.collect::<Vec<_>>();
+
+			// notify the watcher that these extrinsics have been finalized
+			self.listener.write().finalized(&hash, &tx_hashes);
+
+			if last_finalized_hash == Default::default() {
```
tomusdrw: This check does not depend on the loop at all, should be moved somewhere outside.

client/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -501,14 +502,53 @@ impl<B: ChainApi> ValidatedPool<B> {
...
+
+	/// Notify all watchers that transactions in the block with hash have been finalized
+	pub async fn finalized(&self, block_hash: BlockHash<B>) -> Result<(), B::Error> {
+		debug!(target: "txpool", "Attempting to notify watchers of finalization for {}", block_hash);
+		let last_finalized_hash = *self.last_finalized_hash.lock();
+
+		let mut header = self.api.block_header(BlockId::Hash(block_hash.clone()))?
+			.expect("finalized block should have block header; qed");
+
+		while header.hash() != last_finalized_hash {
```
tomusdrw: This loop will exit with an error in case there is no path between `header` and `last_finalized_hash`.

We have a helper: `sp_blockchain::tree_route` that will nicely tell you all the blocks on the route between any two of them. The current logic does not handle re-orgs correctly or cases where there is no common block at all.

client/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -501,14 +502,53 @@ impl<B: ChainApi> ValidatedPool<B> {
...
+
+			// notify the watcher that these extrinsics have been finalized
+			self.listener.write().finalized(&hash, &tx_hashes);
+
+			if last_finalized_hash == Default::default() {
+				// self.last_finalized_hash is created with a default value
+				break
+			}
+
+			header = self.api.block_header(BlockId::Hash(header.parent_hash().clone()))?
```
tomusdrw: In case we don't have a common block with `last_finalized_hash`, we may end up going all the way down to genesis.

rphmeier:
Probably off-topic, but it is confusing to call these finalization events. Finality by definition shouldn't fork. I suggest a rename to `inclusion` events at some point, because these events are about the times where a transaction is included in the chain.

seunlanlege:
@rphmeier yeah sorry about the confusion, meant `fork-awareness and finalization` notifications.

Block inclusion events where already being tracked by the txpool. This PR allows the txpool to notify watchers when a fork occurs and txs that were previously included in a block have been resubmitted to the pool, as well as finality notifications.

client/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -521,14 +531,55 @@ impl<B: ChainApi> ValidatedPool<B> {
 	pub fn status(&self) -> PoolStatus {
 		self.pool.read().status()
 	}
+
+	/// Notify all watchers that transactions in the block with hash have been finalized
+	pub async fn on_block_finalized(&self, block_hash: BlockHash<B>) -> Result<(), B::Error> {
```
tomusdrw: Can we be sure this method is not called concurrently? Since it's an async task it has a racy behavior that may cause the same work being done multiple times (if we receive two finalization events in a short succession).
While I think the notifications won't be sent twice (since the watchers are removed on finalisation), I'd still fix the method just to be safe. My proposal is to update `last_finalized_hash` right away (and switch to `Mutex` or some atomic cell type), i.e.:
```
let last_finalized = mem::replace(&mut self.last_finalized_hash.lock(), block_hash);
```

client/transaction-pool/graph/src/watcher.rs
```diff
@@ -85,26 +85,36 @@ impl<H: Clone, H2: Clone> Sender<H, H2> {
...
+
+	/// Extrinsic has been finalized by a finality gadget.
+	pub fn finalized(&mut self) {
+		self.send(TransactionStatus::Finalized);
+		self.is_finalized = true;
+	}
+
+	/// The block this extrinsic was included in has been retracted
+	pub fn retracted(&mut self) {
+		self.send(TransactionStatus::Retracted);
```
tomusdrw: Should include block hash in case you receive events out of order.

client/transaction-pool/graph/src/listener.rs
```diff
@@ -101,8 +101,26 @@ impl<H: hash::Hash + traits::Member + Serialize, H2: Clone + fmt::Debug> Listene
 	}
 
 	/// Transaction was pruned from the pool.
-	pub fn pruned(&mut self, header_hash: H2, tx: &H) {
-		debug!(target: "txpool", "[{:?}] Pruned at {:?}", tx, header_hash);
-		self.fire(tx, |watcher| watcher.in_block(header_hash))
+	pub fn pruned(&mut self, block_hash: BlockHash<C>, tx: &H) {
+		debug!(target: "txpool", "[{:?}] Pruned at {:?}", tx, block_hash);
+		self.fire(tx, |s| s.in_block(block_hash.clone()));
+		self.finality_watchers.entry(block_hash).or_default().push(tx.clone());
```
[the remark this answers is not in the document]
tomusdrw: I'm afraid that we may end up with `finality_watchers` growing forever, I'd really like to prevent that and only store a fixed amount of blocks here (say 512 for now).
I think we should introduce another status update that would be send to watchers in such case (`FinalityTimeout`).
Roughly:
```
while self.finality_watchers.len() > 512 {
  if let Some((hash, txs)) = self.finality_watchers.pop_front() {
    for tx in txs {
      self.fire(tx, |s| s.finality_timeout(hash));
    }
  }
}
```

Also `finality_watchers` should switch to be a `LinkedHashMap` and `timeout` event should be causing the watcher to be dropped.

tomusdrw:
BTW have you checked if the `finality_notification_stream` does not have the logic you are implementing here? I.e. that it does return ALL hashes on the route between last finalised and newly finalised block? Much of the code would then be redudant and to be honest I'd expect that method to do that already. CC @andresilva ?

andresilva:
@tomusdrw you are correct, you should get notifications for all intermediary blocks from last finalized to new finalized.

andresilva:
https://github.com/paritytech/substrate/blob/master/client/src/client.rs#L1216

This is the code responsible for dispatching finality notifications.

seunlanlege:
> @tomusdrw you are correct, you should get notifications for all intermediary blocks from last finalized to new finalized.

So I guess, there's no need to track `last_finalized_block` and fetching `TreeRoute`. :cry:

client/transaction-pool/graph/src/listener.rs
```diff
@@ -101,8 +101,26 @@ impl<H: hash::Hash + traits::Member + Serialize, H2: Clone + fmt::Debug> Listene
...
+		if let Some(hashes) = self.finality_watchers.remove(block_hash) {
+			for hash in hashes {
+				self.fire(&hash, |s| s.retracted())
+			}
+		}
+	}
+
+	/// Notify all watchers that transactions have been finalized
+	pub fn finalized(&mut self, block_hash: &BlockHash<C>, txs: &[H]) {
+		self.finality_watchers.remove(block_hash);
```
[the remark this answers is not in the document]
seunlanlege: > Should we join the two together?

the hashes in `finality_watchers` are aready in `txs`, reason we ignore it is to notify potentially retracted txs that made it into a new block.

tomusdrw:
> So I guess, there's no need to track last_finalized_block and fetching TreeRoute. cry

Yeah, I'm pretty sure I've mentioned that it's best to double check what kind of events `finalisation_stream` returns. Sorry if I wasn't explicit enough, now looking at the code you started writing I've noticed that it's really non trivial, and most likely is already implemented somewhere, hence wanted to double check.

seunlanlege:
Ping @tomusdrw, this should be ready to be merged, wdyt?

client/transaction-pool/graph/src/listener.rs
```diff
@@ -16,30 +16,34 @@
...
 /// Extrinsic pool default listener.
-pub struct Listener<H: hash::Hash + Eq, H2> {
-	watchers: HashMap<H, watcher::Sender<H, H2>>
+pub struct Listener<H: hash::Hash + Eq, C: ChainApi> {
+	watchers: HashMap<H, watcher::Sender<H, BlockHash<C>>>,
+	finality_watchers: LinkedHashMap<BlockHash<C>, Vec<H>>,
 }
 
-impl<H: hash::Hash + Eq, H2> Default for Listener<H, H2> {
+/// Maximum number of watchers awaiting finality at any time.
```
tomusdrw:
```suggestion
/// Maximum number of blocks awaiting finality at any time.
```

client/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -521,14 +529,35 @@ impl<B: ChainApi> ValidatedPool<B> {
...
+		debug!(target: "txpool", "Attempting to notify watchers of finalization for {}", block_hash);
+		// fetch all extrinsic hashes
+		let tx_hashes = self.api.block_body(&BlockId::Hash(block_hash.clone())).await?
+			.expect("fetched block header should have block body; qed")
```
tomusdrw: This proof is most likely valid, but I'd prefer to be on the safe side here and avoid crashing if we ever end up with a hash that does not have a body in DB (for whatever reason).

Could you change that to `if let`?

client/transaction-pool/graph/src/watcher.rs
```diff
@@ -85,26 +85,41 @@ impl<H: Clone, H2: Clone> Sender<H, H2> {
...
 
 	/// Extrinsic has been included in block with given hash.
-	pub fn in_block(&mut self, hash: H2) {
+	pub fn in_block(&mut self, hash: BH) {
 		self.send(TransactionStatus::InBlock(hash));
-		self.finalized = true;
+	}
+
+	/// Extrinsic has been finalized by a finality gadget.
+	pub fn finalized(&mut self, hash: &BH) {
```
tomusdrw:
```suggestion
	pub fn finalized(&mut self, hash: BH) {
```
to avoid cloning

client/transaction-pool/graph/src/watcher.rs
```diff
@@ -85,26 +85,41 @@ impl<H: Clone, H2: Clone> Sender<H, H2> {
...
+		self.send(TransactionStatus::Finalized(hash.clone()));
+		self.is_finalized = true;
+	}
+
+	/// The block this extrinsic was included in has been retracted
+	pub fn finality_timeout(&mut self, hash: &BH) {
```
tomusdrw:
```suggestion
	pub fn finality_timeout(&mut self, hash: BH) {
```
ditto

client/transaction-pool/graph/src/watcher.rs
```diff
@@ -85,26 +85,41 @@ impl<H: Clone, H2: Clone> Sender<H, H2> {
...
+		self.send(TransactionStatus::FinalityTimeout(hash.clone()));
+	}
+
+	/// The block this extrinsic was included in has been retracted
+	pub fn retracted(&mut self, hash: &BH) {
```
tomusdrw:
```suggestion
	pub fn retracted(&mut self, hash: BH) {
```
and here. It's better to make the caller aware that you need an owned hash, than to pretend you are fine with a reference and then clone internally.

client/transaction-pool/graph/src/watcher.rs
```diff
@@ -85,26 +85,41 @@ impl<H: Clone, H2: Clone> Sender<H, H2> {
...
+
+	/// Extrinsic has been finalized by a finality gadget.
+	pub fn finalized(&mut self, hash: &BH) {
+		self.send(TransactionStatus::Finalized(hash.clone()));
+		self.is_finalized = true;
+	}
+
+	/// The block this extrinsic was included in has been retracted
+	pub fn finality_timeout(&mut self, hash: &BH) {
+		self.send(TransactionStatus::FinalityTimeout(hash.clone()));
```
tomusdrw: Should set `is_finalized = true`

primitives/transaction-pool/src/pool.rs
```diff
@@ -217,11 +224,30 @@ pub trait TransactionPool: Send + Sync {
...
+		/// Header of the just imported block
+		header: B::Header,
+		/// List of retracted blocks ordered by block number.
+		retracted: Vec<B::Hash>,
+	},
+	/// An existing block has been finalzied.
+	Finalized {
+		/// Hash of just finalized block
+		hash: B::Hash,
+	}
```
tomusdrw:
```suggestion
	},
```

seunlanlege:
Ping @tomusdrw, :confetti_ball:

PR #4743, opened 2020-01-27
Add decl_error to template module

JoshOrndorff:
This PR demonstrates the `decl_error!` macro in the TemplateModule. The big changes are:
* Adds a decl_error block
* Adds a dispatchable call which throws errors under appropriate circumstances
* Updates language from module to pallet
* Adds a test to ensure the proper error is thrown in a particular case.

### DIFF
--- a/bin/node-template/runtime/src/template.rs
+++ b/bin/node-template/runtime/src/template.rs
@@ -29,6 +29,26 @@ decl_storage! {
 	}
 }
 
+// The pallet's events
+decl_event!(
+	pub enum Event<T> where AccountId = <T as system::Trait>::AccountId {
+		/// Just a dummy event.
+		/// Event `Something` is declared with a parameter of the type `u32` and `AccountId`
+		/// To emit this event, we call the deposit funtion, from our runtime funtions
+		SomethingStored(u32, AccountId),
+	}
+);
+
+// The pallet's errors
+decl_error! {
+	pub enum Error for Module<T: Trait> {
+		/// Value was None
+		NoneValue,
+		/// Value reached maximum and cannot be incremented further
+		StorageOverflow,
+	}
+}
+
 // The pallet's dispatchable functions.
 decl_module! {
 	/// The module declaration.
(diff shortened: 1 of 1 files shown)


bin/node-template/runtime/src/template.rs
```diff
@@ -52,25 +56,52 @@ decl_module! {
...
+		/// Just a dummy event.
+		/// Event `Something` is declared with a parameter of the type `u32` and `AccountId`
+		/// To emit this event, we call the deposit funtion, from our runtime funtions
 		SomethingStored(u32, AccountId),
 	}
 );
 
-/// tests for this module
+// The pallet's errors
+decl_error! {
```
kianenigma: better to put it above decl_module

bin/node-template/runtime/src/template.rs
```diff
@@ -52,25 +56,52 @@ decl_module! {
 			Self::deposit_event(RawEvent::SomethingStored(something, who));
 			Ok(())
 		}
+
+		/// Another dummy entry point.
+		/// takes no parameters, attempts to increment storage value, and possibly throws an error
+		pub fn cause_error(origin) -> dispatch::DispatchResult {
+			// TODO: You only need this if you want to check it was signed.
```
kianenigma: is this intentional? else better to say `NOTE:` as `TODO` can be confusing with leftover work from a PR

bin/node-template/runtime/src/template.rs
```diff
@@ -29,17 +29,21 @@ decl_storage! {
 	}
 }
 
-// The module's dispatchable functions.
+// The pallet's dispatchable functions.
 decl_module! {
 	/// The module declaration.
 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		// Initializing errors
+		// this is needed only if you are using errors in your pallet
```
bkchr: Maybe also say that you need this to expose the error in the Metadata.

bin/node-template/runtime/src/template.rs
bkchr: You probably always need the check. However, if you want to check for non-signed you should use `ensure_none! `

bin/node-template/runtime/src/template.rs
JoshOrndorff: Okay, I'll do that. I'm just curious why that is preferred.

bin/node-template/runtime/src/template.rs
JoshOrndorff: You mean in `construct_runtime!`? I just realized I didn't actually do that in the node template. I'll add it there too.

shawntabrizi: What @bkchr is saying is that adding `type Error = Error<T>` also tells `decl_module` to generate the right stuff to expose the Error enum in your metadata. No need to change code, simply update the comment to include that this line also exposes information in the metadata.

PR #4744, opened 2020-01-28
Remove unnecessary duplication in generic asset create

hoani:
The generic asset create extrinsic duplicated code that was already in `create_asset`.
This is a simple refactoring to make generic asset a tiny bit DRY-er.
No functional changes made.


hoani:
[clabot:check]

PR #4745, opened 2020-01-28
Make sure destroy_state is called in case of errors.

tomusdrw:
This PR makes sure that `backend.destroy_state` is being called even if the call returns an error.
Seems that the consequences could be unsynchronized cache or incorrect usage reporting (so nothing super bad afaiu).

Long term solution would rather be to have a guard that does that on `Drop`, but I've heard that @arkpar might be investigating this already.

### DIFF
--- a/client/src/call_executor.rs
+++ b/client/src/call_executor.rs
@@ -181,7 +181,7 @@ where
 			let _lock = self.backend.get_import_lock().read();
 			self.backend.destroy_state(state)?;
 		}
-		result
+		result.map_err(Into::into)
 	}
 
 	fn runtime_version(&self, id: &BlockId<Block>) -> sp_blockchain::Result<RuntimeVersion> {
--- a/client/src/call_executor.rs
+++ b/client/src/call_executor.rs
@@ -89,12 +89,12 @@ where
 		).execute_using_consensus_failure_handler::<_, NeverNativeValue, fn() -> _>(
 			strategy.get_manager(),
 			None,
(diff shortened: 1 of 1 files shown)


arkpar:
Unsynced cache is not a big deal indeed. Reads won't be propagated to cache and that's it.

client/src/call_executor.rs
```diff
@@ -173,18 +176,19 @@ where
...
 			.with_storage_transaction_cache(storage_transaction_cache.as_mut().map(|c| &mut **c))
 			.execute_using_consensus_failure_handler(execution_manager, native_call)
-		}?;
+		};
 		{
 			let _lock = self.backend.get_import_lock().read();
 			self.backend.destroy_state(state)?;
 		}
-		Ok(result)
+		Ok(result?)
```
bkchr:
```suggestion
		result
```

client/src/call_executor.rs
```diff
@@ -173,18 +176,19 @@ where
...
 			.with_storage_transaction_cache(storage_transaction_cache.as_mut().map(|c| &mut **c))
 			.execute_using_consensus_failure_handler(execution_manager, native_call)
-		}?;
+		};
 		{
 			let _lock = self.backend.get_import_lock().read();
 			self.backend.destroy_state(state)?;
 		}
-		Ok(result)
+		result
```
bkchr:
```suggestion
		result.map_err(Into::into)
```

PR #4748, opened 2020-01-28
Add SS58 network identifier for Centrifuge

philipstanislaus:
Thank you for your work on substrate!

As discussed in the Substrate Builder's Program call, we would love to use our own network identifier for Centrifuge Chain. 

There are multiple reasons for that, the main one is to prevent reusage across networks, decreasing the value at risk if a private key is lost. Also it allows us to easily identify Centrifuge addresses, which will start with `4b` to `4h`. 

Looking forward to hear your feedback!


philipstanislaus:
[clabot:check]

bkchr:
Please merge master to fix the conflict.

philipstanislaus:
Done @bkchr

PR #4754, opened 2020-01-28
pallet-contracts: Rent projection RPC

pepyakin:
This change adds a new RPC endpoint which allows a user to query the current state of a given contract.

State rent is enacted lazy. That is, if a contract runs out of funds it becomes inaccessible, but it is not removed from the chain storage right away. Instead, it will be removed on the first access: a call or e.g. or poking with `claim_surcharge`.

This however poses a problem from an external users of the chain. If an external user wants to know if a given contract is usable or not, the user will have to do the calculations by themselves.

While it might be possible to work around this issue by piggybacking on some other call, we introduce a new RPC call here: `contracts_rentProjection`. This call doesn't can tell the status of the given contract, it will also return the projected block number of eviction. Such functionality might be useful in creation of crawlers that would call `claim_surcharge` for the nearly expiring contracts.


frame/contracts/common/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+	/// The specified contract is a tombstone and thus cannot have any storage.
+	IsTombstone,
+}
+
+pub type RentProjectionResult<BlockNumber> = Result<RentProjection<BlockNumber>, GetStorageError>;
+
+#[derive(Eq, PartialEq, codec::Encode, codec::Decode, sp_runtime::RuntimeDebug)]
+pub enum RentProjection<BlockNumber> {
+	/// Eviction is projected to happen at the specified block number.
+	EvictionDate(BlockNumber),
```
NikVolf: `EvictionBlock` ?

frame/contracts/src/rent.rs
```diff
@@ -14,102 +14,152 @@
...
+/// Returns amount of funds available to consume by rent mechanism.
+///
+/// Rent mechanism cannot consume more than `rent_allowance` set by the contract and it cannot make
+/// the balance lower than [`subsistence_threshold`].
+fn rent_budget<T: Trait>(
+	balance: &BalanceOf<T>,
+	contract: &AliveContractInfo<T>,
+	subsistence_threshold: &BalanceOf<T>,
+) -> BalanceOf<T> {
+	let rent_allowed_to_charge = *balance - *subsistence_threshold;
```
tomusdrw: Is it always true that `balance > subsitence_threshold`? It definitely makes sense, unless it's being called in cases when we know the account will be dropped, but still for some reason call this function. Perhaps better to be safe here and do saturated subtraction?

frame/contracts/src/rent.rs
```diff
@@ -14,102 +14,152 @@
...
+	<BalanceOf<T>>::min(contract.rent_allowance, rent_allowed_to_charge)
+}
 
+/// Consider the case for rent payment of the given account and returns a `Verdict`.
+///
+/// The `current_block_number` must be equal to the current block number. Use `handicap` do
```
tomusdrw: > The `current_block_number` must be equal to the current block number.

Sounds pretty obvious :D

frame/contracts/src/rent.rs
```diff
@@ -14,102 +14,152 @@
...
+	let subsistence_threshold = subsistence_threshold::<T>();
 	if balance < subsistence_threshold {
 		// The contract cannot afford to leave a tombstone, so remove the contract info altogether.
-		<ContractInfoOf<T>>::remove(account);
-		child::kill_storage(&contract.trie_id, contract.child_trie_unique_id());
-		<Module<T>>::deposit_event(RawEvent::Evicted(account.clone(), false));
-		return (RentOutcome::Evicted, None);
+		return Verdict::Kill;
 	}
+	let rent_budget = rent_budget::<T>(&balance, contract, &subsistence_threshold);
```
tomusdrw: Ah, here we make sure that it's actually the case that `rent_budget` is being called with `balance >= subsistence_threshold`, maybe move the check within the function (and return `Option`) to make that more colocated?

frame/contracts/src/rent.rs
```diff
@@ -127,75 +177,200 @@ fn try_evict_or_and_pay_rent<T: Trait>(
...
+		account,
+		current_block_number,
+		Zero::zero(),
+		&alive_contract_info,
+	);
+	enact_verdict(account, alive_contract_info, current_block_number, verdict)
 }
 
-/// Evict the account if it should be evicted at the given block number.
+/// Process a snitch that a contract under the given address should be evicted.
```
tomusdrw: Is `snitch` commonly used in that file? What's the meaning in that context?

frame/contracts/src/rent.rs
```diff
@@ -127,75 +177,200 @@ fn try_evict_or_and_pay_rent<T: Trait>(
...
+/// returned projection is relevent for the current block, i.e. it is as if the contract was
+/// accessed at the beginning of the current block. Returns `None` in case if the contract was
+/// evicted before or as a result of the rent collection.
+///
+/// The returned value is only an estimation. It doesn't take into account any top ups, changing the
+/// rent allowance, or any problems coming from withdrawing the dues.
+///
+/// NOTE that this is not a side-effect free function! It will actually collect rent and then
+/// compute the projection.
+pub fn compute_rent_projection<T: Trait>(
```
tomusdrw: Maybe call it `compute_and_collect_rent`?

pepyakin: `compute_and_collect_rent_projection` you mean? Yeah, that sounds good.

frame/contracts/src/rent.rs
pepyakin: Ah, I just realized that I meant not "a snitch", but rather a thing that a snitch is doing!

But this is from [this](https://github.com/paritytech/substrate/pull/4733) PR...

frame/contracts/src/rent.rs
pepyakin: Ahh, good point. I will rephrase.

This is also from https://github.com/paritytech/substrate/pull/4733 . I will file a separate PR to address this.

frame/contracts/common/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "pallet-contracts-common"
```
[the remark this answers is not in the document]
pepyakin: Yeah that was a tough one.

We have all our primitives living in `/primitives` but if we share some piece of code they live in a common package. Or used to. Now, it seems that we do not use this convention anymore.

Do you think I should move it to `/primitives` as well? Otherwise it would be kind of weird that it is kind of primitives but not in primitives dir?

frame/contracts/rpc/src/lib.rs
```diff
@@ -132,6 +132,13 @@ pub trait ContractsApi<BlockHash, AccountId, Balance> {
 		key: H256,
 		at: Option<BlockHash>,
 	) -> Result<Option<Bytes>>;
+
+	#[rpc(name = "contracts_rentProjection")]
```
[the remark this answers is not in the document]
pepyakin: Ooops!

frame/contracts/src/rent.rs
pepyakin: ah, yeah, makes sense!

bin/node/runtime/src/lib.rs
```diff
@@ -744,7 +744,7 @@ impl_runtime_apis! {
 		}
 	}
 
-	impl pallet_contracts_rpc_runtime_api::ContractsApi<Block, AccountId, Balance> for Runtime {
+	impl pallet_contracts_rpc_runtime_api::ContractsApi<Block, AccountId, Balance, BlockNumber> for Runtime {
```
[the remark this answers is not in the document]
pepyakin: But how do I access it from the declaration of `ContractsApi`. The `Block` type doesn't figure there... or does it?

frame/contracts/src/rent.rs
pepyakin: ah, no, we compute rent projection but collect rent. That the reason why I didn't like `compute_and_collect_rent`, but `compute_and_collect_rent_projection` is no better.

frame/contracts/common/Cargo.toml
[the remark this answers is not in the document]
tomusdrw: It seems we only have `consensus-common` and `executor-common`, The first one seems like a good name for me, cause it's the common part between multiple consensus algorithms. `primitive` to me seems more like a thing that is defining the interface/interoperation between two separate modules. And I think this is exactly how that crate is being used.

However I don't think it should be moved to `/primitives`, that directory is special  as it's the kind of primitive types that are defining (currently virtual) repository/logical boundaries between `/frame`, `/client` and `node`. The RPC is a bit tricky, cause it will always depend on `/frame` and `/primitives`, I'm actually totally fine with keeping it within `/frame`, hence the `contract-primitives` could stay in `/frame` as well.

frame/contracts/rpc/src/lib.rs
```diff
@@ -149,15 +156,15 @@ impl<C, B> Contracts<C, B> {
 		}
 	}
 }
-
-impl<C, Block, AccountId, Balance> ContractsApi<<Block as BlockT>::Hash, AccountId, Balance>
-	for Contracts<C, Block>
+impl<C, Block, AccountId, Balance, BlockNumber>
+	ContractsApi<<Block as BlockT>::Hash, BlockNumber, AccountId, Balance> for Contracts<C, Block>
```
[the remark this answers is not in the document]
tomusdrw: Should be still implemented. Now the `BlockNumber` could in theory be different than `<Block as BlockT>::Number` which would be quite weird.

frame/contracts/src/rent.rs
tomusdrw: Right, I see what you mean. I guess `collect_rent_and_get_projection` would be the right name, but it becomes quite longish.

Now it seems we have a inconspicuous name for a function with quite important side effects. I think we could flip it and have a more side-effecty name that happens to return some additional value. What I have in mind is `fn collect_rent() -> FutureRentProjection` function to emphasize on the side effect part, not the return value part.

frame/contracts/rpc/src/lib.rs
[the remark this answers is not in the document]
pepyakin: Ah, sorry! Initially I assumed that this is directly related to [this](https://github.com/paritytech/substrate/pull/4754#discussion_r372533662), but in fact this is a legit concern.

frame/contracts/common/Cargo.toml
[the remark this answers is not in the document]
pepyakin: Thanks for the explaintations, really helpful!

frame/contracts/src/rent.rs
pepyakin: Yeah, but the problem that we already have `collect_rent` which is actually used in on-chain logic...

This one is used only for the RPC, the difference from `collect_rent` is that this one returns additional information which is useless in the on-chain version.

bin/node/runtime/src/lib.rs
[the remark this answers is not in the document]
pepyakin: I actually had this idea already, but I dismissed it: I have **big** problems with items coming out of nowhere, i.e. when the developer didn't name them explicitly.

In my opinion the items should come either from imports, defined in the current file, or at least be explicitly named by the developer and must be easily greppable in the same file. This IMO eases reviews and also improves onboarding experience for new contributors, since this way it will be always clear where all the items are coming from.

However, I actually just took a shot to this for self-education and got biten by some nonsensical error.

<details>

```
error[E0053]: method `ContractsApi_rent_projection_runtime_api_impl` has an incompatible type for trait
   --> bin/node/runtime/src/lib.rs:806:1
    |
656 | / impl_runtime_apis! {
657 | |     impl sp_api::Core<Block> for Runtime {
658 | |         fn version() -> RuntimeVersion {
659 | |             VERSION
...   |
805 | |     }
806 | | }
    | | ^
    | | |
    | |_expected associated type, found u32
    |   in this macro invocation
    |
    = note: expected type `fn(&RuntimeApiImpl<__SR_API_BLOCK__, RuntimeApiImplCall>, &sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::BlockId<__SR_API_BLOCK__>, REWARD_CURVE::_sp_runtime::sp_application_crypto::sp_core::ExecutionContext, std::option::Option<REWARD_CURVE::_sp_runtime::AccountId32>, std::vec::Vec<u8>) -> std::result::Result<REWARD_CURVE::_sp_runtime::sp_application_crypto::sp_core::NativeOrEncoded<std::result::Result<pallet_contracts_primitives::RentProjection<<<__SR_API_BLOCK__ as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::BlockT>::Header as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::HeaderT>::Number>, pallet_contracts_primitives::ContractAccessError>>, <RuntimeApiImplCall as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::CallApiAt<__SR_API_BLOCK__>>::Error>`
               found type `fn(&RuntimeApiImpl<__SR_API_BLOCK__, RuntimeApiImplCall>, &sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::BlockId<__SR_API_BLOCK__>, REWARD_CURVE::_sp_runtime::sp_application_crypto::sp_core::ExecutionContext, std::option::Option<REWARD_CURVE::_sp_runtime::AccountId32>, std::vec::Vec<u8>) -> std::result::Result<REWARD_CURVE::_sp_runtime::sp_application_crypto::sp_core::NativeOrEncoded<std::result::Result<pallet_contracts_primitives::RentProjection<u32>, pallet_contracts_primitives::ContractAccessError>>, <RuntimeApiImplCall as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::CallApiAt<__SR_API_BLOCK__>>::Error>`
    = note: consider constraining the associated type `<<__SR_API_BLOCK__ as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::BlockT>::Header as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::HeaderT>::Number` to `u32` or calling a method that returns `<<__SR_API_BLOCK__ as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::BlockT>::Header as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::HeaderT>::Number`
    = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html

   --> bin/node/runtime/src/lib.rs:656:1
    |
656 | / impl_runtime_apis! {
657 | |     impl sp_api::Core<Block> for Runtime {
658 | |         fn version() -> RuntimeVersion {
659 | |             VERSION
...   |
805 | |     }
806 | | }
    | | ^
    | | |
    | | expected u32, found associated type
    | |_expected `std::result::Result<REWARD_CURVE::_sp_runtime::sp_application_crypto::sp_core::NativeOrEncoded<std::result::Result<pallet_contracts_primitives::RentProjection<u32>, pallet_contracts_primitives::ContractAccessError>>, <RuntimeApiImplCall as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::CallApiAt<__SR_API_BLOCK__>>::Error>` because of return type
    |   in this macro invocation
    |
    = note: expected type `std::result::Result<REWARD_CURVE::_sp_runtime::sp_application_crypto::sp_core::NativeOrEncoded<std::result::Result<pallet_contracts_primitives::RentProjection<u32>, _>>, _>`
               found type `std::result::Result<REWARD_CURVE::_sp_runtime::sp_application_crypto::sp_core::NativeOrEncoded<std::result::Result<pallet_contracts_primitives::RentProjection<<<__SR_API_BLOCK__ as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::BlockT>::Header as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::HeaderT>::Number>, _>>, _>`
    = note: consider constraining the associated type `<<__SR_API_BLOCK__ as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::BlockT>::Header as sp_api_hidden_includes_IMPL_RUNTIME_APIS::sp_api::HeaderT>::Number` to `u32`
    = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html
```
</details>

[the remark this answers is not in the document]
tomusdrw: Fair enough! I think this approach is a bit superfluous but indeed much less confusing.

frame/contracts/src/rent.rs
tomusdrw: I see, thanks for explanations, sounds good to leave it as is then. Perhaps a comment indicating that the function is used only for `RuntimeApi` calls, which would not persist the changes anyway would be good.

frame/contracts/src/rent.rs
pepyakin: https://github.com/paritytech/substrate/pull/4790

frame/contracts/src/rent.rs
pepyakin: https://github.com/paritytech/substrate/pull/4790

PR #4756, opened 2020-01-29
Add `get_global` for `Sandbox`

bkchr:
This pr adds `get_global` to retrieve a `global` variable from an
instantiated sandbox wasm blob.


primitives/wasm-interface/src/lib.rs
```diff
@@ -298,6 +331,11 @@ pub trait Sandbox {
 		raw_env_def: &[u8],
 		state: u32,
 	) -> Result<u32>;
+
+	/// Get a global variable with `name` from the given `instance_idx` instance.
```
pepyakin:
```suggestion
	/// Get the value of a global variable with `name` from the given `instance_idx` instance.
```

bkchr:
Please not merge on green, I will do the rename to `get_global_val`.

PR #4759, opened 2020-01-29
Revamp some of the weights

gavofyork:


frame/support/src/weights.rs
```diff
@@ -179,25 +179,37 @@ pub enum SimpleDispatchInfo {
...
 	MaxNormal,
 	/// A normal dispatch with no weight.
-	FreeNormal,
+	///
+	/// WARNING: This causes a dispatchable call (i.e. transaction type) to be placed in the chain
+	/// an arbitrary number of times by any account (even a non-existence account) for no cost. IT
+	/// IS A DOS VECTOR. To use it correctly, there must be customised transaction validation logic
+	/// which limits inclusion of the transaction in some other way. If you do not do this, then
+	/// your chain WILL BE INSECURE.
+	InsecureFreeNormal,
```
xlc: FreeNormal still have pays_fee = true so it is safe. i.e. it still needs the base fee + bytes fee. Only `FreeOperational` have pays_fee = false, which makes it insure.

frame/support/src/weights.rs
```diff
@@ -213,11 +225,11 @@ impl<T> PaysFee<T> for SimpleDispatchInfo {
 		match self {
 			SimpleDispatchInfo::FixedNormal(_) => true,
 			SimpleDispatchInfo::MaxNormal => true,
-			SimpleDispatchInfo::FreeNormal => true,
+			SimpleDispatchInfo::InsecureFreeNormal => true,
 
 			SimpleDispatchInfo::FixedOperational(_) => true,
 			SimpleDispatchInfo::MaxOperational => true,
-			SimpleDispatchInfo::FreeOperational => false,
+			SimpleDispatchInfo::InsecureFreeOperational => false,
```
xlc: Only this one is insecure because it have pays_fee = false

frame/support/src/weights.rs
```diff
@@ -178,26 +177,23 @@ pub enum SimpleDispatchInfo {
 	FixedNormal(Weight),
 	/// A normal dispatch with the maximum weight.
 	MaxNormal,
-	/// A normal dispatch with no weight.
+	/// A normal dispatch with no weight. Base and bytes fees still need to be paid.
 	FreeNormal,
```
bkchr:
```suggestion
	InsecureFreeNormal,
```

PR #4760, opened 2020-01-29
LinkedMap doesn't implement StoragePrefixedMap

thiolliere:
StoragePrefixedMap means a storage has all its value after a prefix
.
But for the linkedmap the value is different from `#value_type` as the actual value stored is `#values_type++Option<previous_key>++Option<next_key>`
and also there is a head key which is after another prefix.

This creates 2 bugs:
* remove_all doesn't remove the head.
* translate actually removes all `Option<previous_key>++Option<next_key>`, breaking the linkedmap.

We should implement those functions in storagelinkedmap directly.

### DIFF
--- a/frame/support/procedural/src/storage/storage_struct.rs
+++ b/frame/support/procedural/src/storage/storage_struct.rs
@@ -167,18 +167,6 @@ pub fn decl_and_impl(scrate: &TokenStream, def: &DeclStorageDefExt) -> TokenStre
 				);
 
 				quote!(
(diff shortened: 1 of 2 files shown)


thiolliere:
> We should have prefixed the `Head` as well.^^

currently head is at `twox128(moduleprefix)++twox128("HeadOf"++storageprefix)` do you mean putting the head after `twox128(moduleprefix)++twox128(storageprefix)` ? then the iteration would have been broken. I think better to keep the head in a separated prefix, in my mind a storage linked map is a combination of a storage value which hold the head and a storage map which hold the value+linkage.

PR #4766, opened 2020-01-29
client/finality-grandpa/src/observer: Poll NetworkBridge

mxinden:
Patch d4fbb897c reintegrated the periodic neighbor packet worker, by
moving its logic into a `Future` implementation on `NetworkBridge` which
needs to be polled by its upper layer.

Polling by the upper layer was implemented within the `Future`
implementation of the `VoterWork` struct but not within the `Future`
implementation of the `ObserverWork` struct. This patch adds polling of
the `NetworkBridge` to the latter.

Without this bugfix nodes running as a Grandpa Observer don't report
reputation changes to the underlying network and don't send neighbor
packets.

### DIFF
--- a/client/finality-grandpa/src/observer.rs
+++ b/client/finality-grandpa/src/observer.rs
@@ -336,10 +336,8 @@ where
 {
 	type Output = Result<(), Error>;
 
-	fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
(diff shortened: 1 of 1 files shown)


mxinden:
> Issue for the regression test for this? :)

Thanks for pushing for this @NikVolf . Regression test pull request: https://github.com/paritytech/substrate/pull/4778

PR #4767, opened 2020-01-29
client/network-gossip: Integrate GossipEngine tasks into Future impl

mxinden:
`GossipEngine` spawns two tasks, one for a periodic tick, one to forward
messages from the network to subscribers. These tasks hold an `Arc` to a
`GossipEngineInner`.

To reduce the amount of shared ownership (locking) this patch integrates
the two tasks into a `Future` implementation on the `GossipEngine`
struct. This `Future` implementation can now be called from a single
owner, e.g. the `finality-grandpa` `NetworkBridge`.

As a side effect this removes the requirement on the `network-gossip`
crate to spawn tasks and thereby removes the requirement on the
`finality-grandpa` crate to spawn any tasks.

This is part of a greater effort to reduce the number of owners of
components within `finality-grandpa`, `network` and `network-gossip` as
well as to reduce the amount of unbounded channels. For details see
d4fbb897c, f0c18520a and 5afc7777e.

Once this simplification lands we can get rid of the construct of a
`GossipEngine` holding an `Arc<Mutex<GossipEngineInner>>`,
given that `GossipEngine` only has a single owner.


bin/node-template/src/service.rs
```diff
@@ -160,7 +160,6 @@ pub fn new_full<C: Send + Default + 'static>(config: Configuration<C, GenesisCon
 				grandpa_link,
 				service.network(),
 				service.on_exit(),
-				service.spawn_task_handle(),
```
mxinden: None of `finality-grandpa` nor `network-gossip` need an executor to spawn tasks after this patch. Thus the logic passing said executor to those crates is removed in `bin/node`, `bin/node-template`, `client/service` and a lot of tests.

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -444,6 +442,12 @@ impl<B: BlockT, N: Network<B>> Future for NetworkBridge<B, N> {
 			}
 		}
 
+		match self.gossip_engine.poll_unpin(cx) {
```
mxinden: The `GossipEngine` needs to be polled by the `NetworkBridge`.

client/finality-grandpa/src/observer.rs
```diff
@@ -351,24 +346,24 @@ where
...
 			Poll::Ready(Some(command)) => {
 				// some command issued externally
-				this.handle_voter_command(command)?;
+				self.handle_voter_command(command)?;
 				cx.waker().wake_by_ref();
 			}
 		}
 
-		Poll::Pending
+		Future::poll(Pin::new(&mut self.network), cx)
```
mxinden: Bug originally fixed in https://github.com/paritytech/substrate/pull/4766.

client/network-gossip/src/bridge.rs
```diff
@@ -36,101 +37,42 @@ pub struct GossipEngine<B: BlockT> {
 struct GossipEngineInner<B: BlockT> {
 	state_machine: ConsensusGossip<B>,
 	network: Box<dyn Network<B> + Send>,
+	periodic_maintenance_interval: futures_timer::Delay,
+	network_event_stream: Compat01As03<Box<dyn Stream01<Error = (), Item = Event> + Send>>,
+	engine_id: ConsensusEngineId,
```
mxinden: Once this lands there is no need for `GossipEngineInner` anymore, thus the `engine_id` does not need to be duplicated either.

client/network-gossip/src/bridge.rs
```diff
@@ -36,101 +37,42 @@ pub struct GossipEngine<B: BlockT> {
...
+			network_event_stream: Compat01As03::new(network_event_stream),
+			engine_id,
 		}));
 
 		let gossip_engine = GossipEngine {
 			inner: inner.clone(),
 			engine_id,
 		};
 
-		let res = executor.spawn({
```
mxinden: These are the two tasks previously being spawned. You can find the new `Future` implementation further below.

client/network-gossip/src/bridge.rs
```diff
@@ -222,6 +164,59 @@ impl<B: BlockT> GossipEngine<B> {
...
+	}
+}
+
+impl<B: BlockT> Future for GossipEngineInner<B> {
+	type Output = ();
+
+	fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+		let this = &mut *self;
+
+		while let Poll::Ready(Some(Ok(event))) = this.network_event_stream.poll_next_unpin(cx) {
```
mxinden: This is copy and pasted from the previously spawned logic. In the long run I think we should at least log errors on this channel.

mxinden:
:stop_sign: I have not yet created the corresponding Polkadot pull-request. But in the meantime I think this is worth taking a look. :stop_sign:

tomaka:
Sorry for the delay.
What about entirely adding a `GossipEngineInner::next_event()` async function instead of manually implementing `Future`?

Something like:

```rust
impl<B: BlockT> Future for GossipEngine<B> {
	type Output = ();

	fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
		self.inner.lock().run().poll_unpin(cx)
	}
}

impl<B: BlockT> GossipEngineInner<B> {
    async fn run(&mut self) { ... }
}
```

The `async` function should then however not use any local variable and only operate on the fields of `self`.

mxinden:
I talked with @tomaka in regards to the suggestion above:

Using an `async` function without local variables would on the one hand

- Reduce boiler plate code (`impl Future for [...]`)
- Not involve any pinning

on the other hand it would make the following two constraints previously enforced by the compiler only enforced by code comments.
- Not being allowed to use local variables within the `async` function.
- Not being allowed to depend on the fact that the generated Generator continues where it last stopped.  

I would prefer to stick with the `poll` function style.

PR #4768, opened 2020-01-29
Fix warnings.

tomusdrw:
There is one left that I don't really know where it's coming from. My suspicion is that it's the `codec` derive code that generates a code with this unused variable. That should probably be fixed in upstream `codec` crate.

```rust
warning: unused variable: `message`
   --> primitives/runtime/src/lib.rs:376:3
    |
367 |       /// A custom error in a module
    |  _____-
368 | |     Module {
369 | |         /// Module index, matching the metadata module index
370 | |         index: u8,
...   |
375 | |         #[allow(unused_variables)] // most likely caused by codec(skip)?
376 | |         message: Option<&'static str>,
    | |         ^^^^^^-
    | |_______________|
    |                 help: try ignoring the field: `message: _`
    |
    = note: `#[warn(unused_variables)]` on by default

```

### DIFF
--- a/bin/node-template/src/cli.rs
+++ b/bin/node-template/src/cli.rs
@@ -90,7 +90,9 @@ where
 
 	let _ = exit_send.send(());
 
-	runtime.block_on(handle);
+	if let Err(e) = runtime.block_on(handle) {
+		log::error!("Error running node: {:?}", e);
+	}
 
 	match service_res {
 		Either::Left((res, _)) => res.map_err(error::Error::Service),
--- a/bin/node/cli/src/cli.rs
+++ b/bin/node/cli/src/cli.rs
@@ -200,7 +200,9 @@ where
 
 	let _ = exit_send.send(());
 
-	runtime.block_on(handle);
+	if let Err(e) = runtime.block_on(handle) {
+		log::error!("Error running node: {:?}", e);
+	}
 
 	match service_res {
 		Either::Left((res, _)) => res.map_err(error::Error::Service),
--- a/bin/node-template/src/service.rs
+++ b/bin/node-template/src/service.rs
@@ -11,7 +11,6 @@ use sc_executor::native_executor_instance;
 pub use sc_executor::NativeExecutor;
 use sp_consensus_aura::sr25519::{AuthorityPair as AuraPair};
 use grandpa::{self, FinalityProofProvider as GrandpaFinalityProofProvider};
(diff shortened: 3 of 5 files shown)


bkchr:
The missing one is already fixed by this one: https://github.com/paritytech/substrate/pull/4756 :)

bin/node-template/src/cli.rs
```diff
@@ -90,7 +90,9 @@ where
 
 	let _ = exit_send.send(());
 
-	runtime.block_on(handle);
+	if let Err(e) = runtime.block_on(handle) {
```
bkchr:
```suggestion
	let _ = runtime.block_on(handle);
```

As it is only the informant, we can ignore what happens to the task IMHO.

bin/node-template/src/cli.rs
```diff
@@ -90,7 +90,9 @@ where
...
+		log::error!("Error running node: {:?}", e);
```
bkchr:
```suggestion
```

bin/node-template/src/cli.rs
```diff
@@ -90,7 +90,9 @@ where
...
+	}
```
bkchr:
```suggestion
```

bin/node/cli/src/cli.rs
```diff
@@ -200,7 +200,9 @@ where
 
 	let _ = exit_send.send(());
 
-	runtime.block_on(handle);
+	if let Err(e) = runtime.block_on(handle) {
```
bkchr:
```suggestion
	let _ = runtime.block_on(handle);
```

bin/node/cli/src/cli.rs
```diff
@@ -200,7 +200,9 @@ where
...
+		log::error!("Error running node: {:?}", e);
```
bkchr:
```suggestion
```

bin/node/cli/src/cli.rs
```diff
@@ -200,7 +200,9 @@ where
...
+	}
```
bkchr:
```suggestion
```

bin/node-template/src/cli.rs
tomusdrw: It's the `exit` handle as well. I think it's better to log something than to just swallow the error.

bkchr: `let service_res = runtime.block_on(select(service, exit));`

`exit` is also taken here ;)

PR #4769, opened 2020-01-29
Fix miscellaneous compile warnings

expenses:


bkchr:
We already have https://github.com/paritytech/substrate/pull/4768, not sure which one we want to take :D

expenses:
let's go with that one then.

PR #4771, opened 2020-01-29
Refactor im-online and print more debug info.

tomusdrw:
Marking as a draft, as I'm still testing it works correctly on dev chain / local networks.

Main goal of the changes:
1. Make it easier to debug what happens (print more debug info, even when running in WASM).
2. Handle all local keys separately - separate db state & debug messages, previously we were aggregating.
3. Retry heartbeat if the previous one was not included.
4. Simplify logic when it should be running, previous code could even stall if invalid data was found in the DB.

As a side-effect I've also introduced `StorageValueRef` high-level wrapper for the offchain worker local storage API to simplify the logic there.


frame/im-online/src/lib.rs
```diff
@@ -129,37 +130,64 @@ pub mod ed25519 {
...
+	pub sent_at: BlockNumber,
+}
+
+impl<BlockNumber: PartialEq + SimpleArithmetic + Copy> HeartbeatStatus<BlockNumber> {
+	/// Returns true if heartbeat has been recently sent.
+	///
+	/// Note this function will return `true` iff:
+	/// 1. the requested block number (`heartbeat_at`) is the same
+	/// 2. the heartbeat has been sent recently
+	fn is_recent(&self, heartbeat_at: BlockNumber, now: BlockNumber) -> bool {
```
gavofyork: could do with a description of the parameters as it's not immediately obvious what `heartbeat_at` means in the context of repeated heartbeat sending attempts.

frame/im-online/src/lib.rs
```diff
@@ -365,156 +410,130 @@ impl<T: Trait> Module<T> {
...
-			}
-		} else {
-			debug::native::debug!(
-				target: "imonline",
-				"Skipping gossip at: {:?} >= {:?} || {:?}",
-				next_gossip,
-				now,
-				if not_yet_gossipped { "not gossipped" } else { "gossipped" }
-			);
+	pub(crate) fn offchain(block_number: T::BlockNumber)
```
gavofyork: could do with a better name.

frame/im-online/src/lib.rs
```diff
@@ -225,8 +253,8 @@ decl_event!(
 
 decl_storage! {
 	trait Store for Module<T: Trait> as ImOnline {
-		/// The block number when we should gossip.
-		GossipAt get(fn gossip_at): T::BlockNumber;
+		/// The block number when we should send heartbeat.
```
gavofyork: this doesn't make sense. surely there are many blocknumbers when we "should" (or rather "could") be sending a heartbeat.

i guess there is one block number in each session before which we would not expect to send a heartbeat, but there's no mention here of "after which" or "current session".

frame/im-online/src/lib.rs
```diff
@@ -365,156 +410,130 @@ impl<T: Trait> Module<T> {
...
-					authority_index,
-					block_number
-				);
-				continue;
-			}
 
+	fn send_heartbeat(
+		authority_index: u32,
+		key: T::AuthorityId,
+		heartbeat_at: T::BlockNumber,
```
gavofyork: `heartbeat_at` here seems to be used to identify which session a heartbeat is being sent for. if this is the case, i would just use `session_index` instead which is much clearer and doesn't accidentally encode other information like the delay period before which a session's heartbeat should be sent.

frame/im-online/src/lib.rs
```diff
@@ -365,156 +410,130 @@ impl<T: Trait> Module<T> {
...
+			key.extend(authority_index.encode());
+			key
+		};
+		let storage = StorageValueRef::persistent(&key);
+		let res = storage.mutate(|status: Option<Option<HeartbeatStatus<T::BlockNumber>>>| {
+			// Check if there is already a lock for that particular block.
+			// This means that the heartbeat has already been sent, and we are just waiting
+			// for it to be included. However if it doesn't get included for INCLUDE_THRESHOLD
+			// we will re-send it.
+			match status {
```
gavofyork: this block could be rewritten as a single match evaluating to the `Err` on the first arm and the `Ok` on the second arm. there doesn't seem to be a need for the `return`.

frame/im-online/src/lib.rs
```diff
@@ -365,156 +423,128 @@ impl<T: Trait> Module<T> {
...
-			Some(last) => {
-				let worker_status: WorkerStatus<T::BlockNumber> = Decode::decode(&mut &last[..])
-					.map_err(|_| OffchainErr::DecodeWorkerStatus)?;
-
-				let was_aborted = !worker_status.done && worker_status.gossipping_at < now;
+		f: impl FnOnce() -> OffchainResult<T, R>,
+	) -> OffchainResult<T, R> {
+		let key = {
+			let mut key = DB_PREFIX.to_vec();
+			key.extend(authority_index.encode());
```
tomusdrw: A heads up here that this will pollute local storage a bit, since the keys are based on `authority_index` and we don't really remove the locks at any point in the future.
However this should always be bounded to the maximal size of validator set (which I believe has some reasonable values anyway). So I don't think it's necessary to iterate and clear it, if we decide it is, it can be easily added retroactively.

tomusdrw:
Should be ready to review now. I've run a dev chain and a local chain with two validators and it seems it's working fine. Would be cool to check on a bigger network though.

Issue #4611, opened 2020-01-13
Per-pallet child tries


gavofyork:
There should be a variant of `decl_storage` that stores all items in a child-trie rather than the main trie.

Storage format/hashing should be equivalent, except that there should be no module prefix.

Additionally, there should be migration code allowing for modules using the existing system (e.g. Kusama) to migrate their modules' storage over to per-pallet child tries during an upgrade.

thiolliere:
What should be the childinfo used here ? does the module should be the childtrie at the path `twox_64("ModulePrefix")` ? Or is this too weak.
Because this path must not be the start of any other childtrie. as its deletion will delete all database key starting with this prefix, no ? (isn't it @cheme  ?)

cheme:
Yes, child trie prefix in addition to being unique need to avoid starting with another existing child prefix (to avoid making assumption on the keys used by the module). 

So the child trie prefix got the collision resistance of the smaller prefix use, at this point contract allows 256 bit prefix.

So having `twox_64("ModulePrefix")` lowers the collision resistance of other child trie prefix to 64bit. (considering random key usage or batch deletion by child trie prefix).
Seems a bit small (there may be a middle ground with 256, but 64 definitely seems small).

So I would keep something that is at least 128bit (not sure if enough), `twox` would be fine, no need for a crypto resistant hash here as we use some static input in this case. But this prefix should be a constant, so using a crypto resistant hash function got no disadvantage, therefore I would probably use the same hash method as for contract.

But if we consider that pallet modules using those macros cannot use random key, IIRC they are already using some twox hashes over field id, if smaller field prefix is a twox_64 then we can use a hash that is "targetted_collision_resistance_size - 64".

If we don't implement batch deletion and consider pallet module cannot write random key, then it should be fine to just use twox_64, it relies on the the same assumption as for module prefix definition (key used internally in rocksdb will be fairly similar).
So maybe it could be more reasonable to forget this deletion implementation for current child trie  implementation and target batch child trie deletion for a possible future implementation of child trie with managed unique id.

thiolliere:
also we could do more precise stuff such as module could be at `00++twox_64("ModulePrefix")` and otherhash that needs more bytes at `01++hash of size 128`, `10++ hash of size 256`.

cheme:
yes, that seems like a possibility, that could even be different child_trie kinds.

thiolliere:
[...]

PR #4773, opened 2020-01-29
Storage reorganize

thiolliere:
This is an attemp to make frame-support::storage more clear. Multiple things has been done:

1- first commit
  * unhashed and hashed module are renamed top and top_hashed, as they a tools to abstract of top trie, child module keep its name child.
  * generator::Storage* and its corresponding storage trait have been merged, so now there is only one trait for each Storage trait implemented by decl_storage.

2- second commit
  * StorageMap have both generic types and associated types, I removed all generic to only use associated one. Basically `StorageMap<K, V> { type Query; ..}` is now `StorageMap { type Key; type Value; type Query; ... }`. This should ease understanding of the structure but also make easier to build user function on top of them:
     instead of having to declate `fn mywrapper<K: FullCodec, V: FullCodec, S: StorageMap<K, V>>(...) {...}` one can simply write `fn mywrapper<S: StorageMap>(..) {...}` 

    downside: we can no longer implement `StorageValue<OneType>` and `StorageValue<AnotherType>` on the same structure.

3 - third commit
  * I just rename all `storage_*_final_key` to `top_trie_key`. After implementation of https://github.com/paritytech/substrate/issues/4611 this method will be changed to either being only callable when used on top trie (using a where clause) or returning an enum with two variant, one for top trie and one for child trie.

Note: I usually here people not understanding what is our macro magic but for storages we only implement a few trait which are mostly define out of the macro. Maybe this structure will ease understanding of what is implemented on the structure.


thiolliere:
I close to something with only the most useful changes

PR #4775, opened 2020-01-30
Remove --release from check_for_exit.sh

cecton:
I don't see any good reason to build with --release to test this but this step takes 8-8:30 min on the CI and I suspect we would save some time by using the debug build instead.

### DIFF
--- a/.maintain/check_for_exit.sh
+++ b/.maintain/check_for_exit.sh
@@ -4,8 +4,8 @@
 
 set -e
 
-cargo build --release
-./target/release/substrate --dev &
+cargo build
+./target/debug/substrate --dev &
 PID=$!
 
 # Let the chain running for 60 seconds


bkchr:
`9 minutes 49 seconds` :woman_shrugging:

cecton:
I'm currently testing on GitLab hoping to see the difference. But it didn't make much difference. There are a lot of variation. 8min was for an older version, right now it's taking up to 12min.

I'm not sure if we will see any improvement at all but it doesn't seem to hurt and it does make sense that a debug build should go faster. Maybe sccache is doing its job but then 9min to pull artifacts seem long.

cecton:
I ran 2 builds at the same time but they are not exactly on the same commit I think.

RELEASE: ~12min https://gitlab.parity.io/parity/substrate/-/jobs/368362
DEBUG: ~7min https://gitlab.parity.io/parity/substrate/-/jobs/368361

In this particular time frame it worked much better but that could be due to so many reasons.

PR #4781, opened 2020-01-30
Revalidation queue for transaction pool

NikVolf:
Probably last patch for transaction pool before major refactoring.

This changes how chain tip is processed in the transaction pool. For full pool, all pending transactions are sent to the revalidation queue and revalidated in background - oldest first. This should allow purge invalid transaction much faster and maintain more actual chain tip.

Revalidation queue is always async, but for light client it does not spawn background worker.


client/transaction-pool/src/revalidation.rs
```diff
@@ -0,0 +1,346 @@
...
+			from_queue,
+			tick_timeout: Box::pin(interval(BACKGROUND_REVALIDATION_INTERVAL)),
+			block_ordered: Default::default(),
+			members: Default::default(),
+			best_block: Zero::zero(),
+		}
+	}
+
+	fn prepare_batch(&mut self) -> Vec<ExHash<Api>> {
+		let mut queued_exts = Vec::new();
```
tomusdrw: This function seems to do what it says, but I think I'd try to implement it in a bit more straight forward way (I know it's not simple though).
My attempt below (but haven't really tested if it passes the borrow checker):

```rust
let mut queued = Vec::new();
let mut left = BACKGROUND_REVALIDATION_BATCH_SIZE;

while left > 0 {
   let first_block = self.block_ordered.iter().next().cloned() {
     Some(bn) => bn,
     None => break;
   };
   let mut block_drained = false;
   if let Some(extrinsics) = self.block_ordered.get_mut(&first_block) {
       let to_queue = extrinsics.iter().take(left).cloned().collect::<Vec<_>>();
       if to_queue.len() == extrinsics.len() {
          block_drained = true;
       } else {
           for xt in &to_queue {
              extrinsics.remove(xt);
           }
       }
       left -= to_queue.len();
       queued.extend(&mut to_queue);
   }
  
   if block_drained {
     self.block_ordered.remove(&first_block);
   }
}
```
Perhaps we should use `BTreeSet` instead of `HashSet` to allow less iteration (having `split_off` would help a bit).
The issue might be though that it's processed in an ordered fashion.

client/transaction-pool/src/revalidation.rs
```diff
@@ -0,0 +1,346 @@
...
+
+		queued_exts
+	}
+
+	fn push(&mut self, worker_payload: WorkerPayload<Api>) {
+		// we don't add something that already scheduled for revalidation
+		let transactions = worker_payload.transactions;
+		let block_number = worker_payload.at;
+
+		for ext_hash in transactions.into_iter() {
```
tomusdrw:
```suggestion
                let transactions = transactions.into_iter().filter(|hash| !self.members.contains_key(&hash));
		for ext_hash in transactions {
		    ...
		}
```
Or just remove `into_iter` from this line if you prefer to keep `contains_key` within the loop.

client/transaction-pool/src/revalidation.rs
```diff
@@ -0,0 +1,346 @@
...
+
+					let next_batch = this.prepare_batch();
+					let batch_revalidate_fut =
+						batch_revalidate(this.pool.clone(), this.api.clone(), this.best_block, next_batch);
+
+					futures::pin_mut!(batch_revalidate_fut);
+					match batch_revalidate_fut.poll_unpin(cx)
+					{
+						Poll::Ready(_) => continue,
+						_ => break,
```
tomusdrw: That looks incorrect. We only `poll` `batch_revalidate_fut` once and drop it first time it returns `Pending`? Can't we use `async/await ` for the entire revalidation logic instead?

client/transaction-pool/src/revalidation.rs
```diff
@@ -0,0 +1,346 @@
...
+			),
+		}
+	}
+
+	/// Offload some revalidation to the queue.
+	///
+	/// If queue configured with background worker, this will return immediately.
+	/// If queue configured without background worker, this will resolve after
+	/// revalidation is actually done.
+	pub async fn offload(&self, at: NumberFor<Api>, transactions: Vec<ExHash<Api>>) {
```
tomusdrw: I'd call it `revalidate_later`, `offload` seems pretty generic and it's hard to understand what we off-load without a lot of context.

client/transaction-pool/src/revalidation.rs
```diff
@@ -0,0 +1,346 @@
...
+		Self {
+			api,
+			pool,
+			background: None,
+		}
+	}
+
+	/// New revalidation queue with background worker.
+	pub fn new_background(api: Arc<Api>, pool: Arc<Pool<Api>>) -> Self {
+		let spawner = futures::executor::ThreadPool::builder()
```
tomusdrw: Why do we need a separate background thread for full client but not for the light client? Can't we just spawn it on the existing thread pool?

client/transaction-pool/src/revalidation.rs
tomaka: This would be ok if the future of `batch_revalidate` didn't hold any state. Unfortunately that's not the case.
Being unfamiliar with the txpool, I must admit that I don't have context for the logic here.

+1 for rewriting with async/await anyway, as it's way less confusing.

client/transaction-pool/src/revalidation.rs
NikVolf: Which existing thread pool do you mean?

tomusdrw: Whatever futures runtime the `service` is using under the hood.

client/transaction-pool/src/revalidation.rs
NikVolf: Don't see much differences, but no problem. `first_block` probably should clone just key, not key-value.

client/transaction-pool/graph/src/pool.rs
```diff
@@ -170,40 +171,18 @@ impl<B: ChainApi> Pool<B> {
...
-
-		let now = Instant::now();
-		let revalidated_transactions = self.verify(at, ready, false).await?;
-		log::debug!(target: "txpool",
-			"Re-verified transactions, took {} ms. Resubmitting.",
-			now.elapsed().as_millis()
-		);
-
-		let now = Instant::now();
+		revalidated_transactions: HashMap<ExHash<B>, ValidatedTransactionFor<B>>,
```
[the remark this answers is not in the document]
NikVolf: `resubmit` takes HashMap.

client/transaction-pool/src/revalidation.rs
NikVolf: Indeed much cleaner with async/await, sorry haven't pushed myself to do it initially :)

client/transaction-pool/src/revalidation.rs
```diff
@@ -0,0 +1,346 @@
...
+			},
+			Ok(Err(TransactionValidityError::Unknown(err))) => {
+				// skipping unknown, they might be pushed by valid or invalid transaction
+				// when latter resubmitted.
+				log::debug!(target: "txpool", "[{:?}]: Unknown during revalidation: {:?}", ext_hash, err);
+			},
+			Ok(Ok(validity)) => {
+				let data = ext.data.clone();
+				let (_, bytes) = api.hash_and_length(&data);
+				revalidated.insert(
```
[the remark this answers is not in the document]
NikVolf: Well I'd like not to increase api surface, while this is just a simple struct instantiation

client/service/src/builder.rs
```diff
@@ -890,7 +894,7 @@ ServiceBuilder<
 						);
 						let _ = to_spawn_tx_.unbounded_send((
 							Box::pin(future),
-							From::from("txpool-maintain")
+							"txpool-maintain".into(),
```
tomaka: Why change these?

client/transaction-pool/src/testing/pool.rs
```diff
@@ -176,6 +179,10 @@ fn should_revalidate_during_maintenance() {
 	pool.api.push_block(1, vec![xt1.clone()]);
 
 	block_on(pool.maintain(&BlockId::number(1), &[]));
+
+	// maintaince is in background
+	block_on(futures_timer::Delay::new(std::time::Duration::from_millis(10)));
```
tomaka: This smells like it will become a flaky test.

client/service/src/builder.rs
NikVolf: I thought everyone would prefer shorter version
I'll change back, I myself don't really have preference

client/transaction-pool/src/testing/pool.rs
NikVolf: There is a fixed timeout just for tests, so it shouldn't be

client/transaction-pool/src/revalidation.rs
```diff
@@ -0,0 +1,312 @@
...
+
+impl<Api: ChainApi> Unpin for RevalidationWorker<Api> {}
+
+fn interval(duration: Duration) -> impl Stream<Item=()> + Unpin {
+	unfold((), move |_| {
+		Delay::new(duration).map(|_| Some(((), ())))
+	}).map(drop)
+}
+
+/// Actual async revalidation batch.
```
tomusdrw: Please elaborate, please at least attempt to roughly define the process of what `async revalidation batch` is supposed to be.

client/transaction-pool/src/revalidation.rs
```diff
@@ -0,0 +1,312 @@
...
+				.or_insert_with(|| {
+					let mut bt = HashSet::new();
+					bt.insert(ext_hash.clone());
+					bt
+				});
+			self.members.insert(ext_hash.clone(), block_number);
+		}
+	}
+
+	pub async fn run(mut self, from_queue: mpsc::UnboundedReceiver<WorkerPayload<Api>>) {
```
tomusdrw: Docs please, rough description of what this task is doing would be really nice.

client/transaction-pool/src/revalidation.rs
```diff
@@ -0,0 +1,312 @@
...
+
+	/// Queue some transaction for later revalidation.
+	///
+	/// If queue configured with background worker, this will return immediately.
+	/// If queue configured without background worker, this will resolve after
+	/// revalidation is actually done.
+	pub async fn revalidate_later(&self, at: NumberFor<Api>, transactions: Vec<ExHash<Api>>) {
+		if let Some(ref to_worker) = self.background {
+			to_worker.unbounded_send(WorkerPayload { at, transactions })
+				.expect("background task is never dropped");
```
tomusdrw: There is no way you can prove that from the context of this worker. The future is handed outside of this function and you have no idea what happens with it.

client/transaction-pool/src/testing/pool.rs
tomusdrw: Is there a way to avoid relying on timeouts/intervals in tests? Like manually driving the process maybe?
If there is no way the duration should be dependent on the actual constant and not assume that we have 5 there so it should be 10 here.

client/transaction-pool/src/testing/pool.rs
```diff
@@ -213,12 +220,80 @@ fn should_not_retain_invalid_hashes_from_retracted() {
 	pool.api.add_invalid(&xt);
 
 	block_on(pool.maintain(&BlockId::number(1), &[retracted_hash]));
+
+	// maintaince is in background
```
tomusdrw:
```suggestion
	// maintenance is in background
```

client/transaction-pool/src/testing/pool.rs
```diff
@@ -213,12 +220,80 @@ fn should_not_retain_invalid_hashes_from_retracted() {
...
+	block_on(futures_timer::Delay::new(std::time::Duration::from_millis(10)));
```
tomusdrw: Can't we just call something directly to trigger it instead?

client/transaction-pool/src/testing/pool.rs
```diff
@@ -213,12 +220,80 @@ fn should_not_retain_invalid_hashes_from_retracted() {
...
+		futures::executor::block_on_stream(watcher1).collect::<Vec<_>>(),
+		vec![TransactionStatus::Ready, TransactionStatus::InBlock(Hash::zero())],
+	);
+	assert_eq!(
+		futures::executor::block_on_stream(watcher2).collect::<Vec<_>>(),
+		vec![TransactionStatus::Ready, TransactionStatus::InBlock(Hash::zero())],
+	);
+}
+
 fn can_track_heap_size() {
```
tomusdrw: Missing `#[test]` so it doesn't really do anything.

NikVolf: It's just a merge bug

client/transaction-pool/src/lib.rs
```diff
@@ -102,24 +105,40 @@ impl<PoolApi, Block> BasicPool<PoolApi, Block>
...
+				let (queue, background) = revalidation::RevalidationQueue::new_background(pool_api.clone(), pool.clone());
+				(queue, Some(background))
+			},
+		};
 		BasicPool {
-			api: cloned_api,
-			pool: Arc::new(sc_transaction_graph::Pool::new(options, pool_api)),
+			api: pool_api,
+			pool,
+			background_task: background_task.into(),
```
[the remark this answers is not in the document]
NikVolf: Well, it will in turn require `ServiceBuilder` refactoring - it spawns tasks later, when subsystems such as txpool already instantiated.

client/transaction-pool/src/lib.rs
```diff
@@ -51,6 +52,8 @@ pub struct BasicPool<PoolApi, Block>
 	pool: Arc<sc_transaction_graph::Pool<PoolApi>>,
 	api: Arc<PoolApi>,
 	revalidation_strategy: Arc<Mutex<RevalidationStrategy<NumberFor<Block>>>>,
```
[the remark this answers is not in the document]
NikVolf: The other mutex is taken only once in application lifecycle.

client/transaction-pool/src/revalidation.rs
NikVolf: It is not from the context of worker, it is the other part, which is sending.
Worker is the receiving part, and is a never-ending future.

But I agree that we can avoid panic here anyway

client/transaction-pool/src/lib.rs
```diff
@@ -198,6 +217,10 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
 	fn ready_transaction(&self, hash: &TxHash<Self>) -> Option<Arc<Self::InPoolTransaction>> {
 		self.pool.ready_transaction(hash)
 	}
+
+	fn background_task(&self) -> Option<Pin<Box<dyn Future<Output = ()> + Send>>> {
```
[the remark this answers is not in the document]
NikVolf: You allowed to call it any number of times, it will return the actual task only once. I don't see other way to do it within the current `ServiceBuilder`

tomaka:
Ideally no background thread would be used, all the background tasks would be spawned by the service in order to be given names and to be diagnosable by the `futures-diagnose` library.

client/transaction-pool/graph/src/pool.rs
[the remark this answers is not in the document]
NikVolf: I am not sure, I can't touch everything in one PR, so I rely on something that was there before me :)

client/transaction-pool/src/lib.rs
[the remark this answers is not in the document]
NikVolf: Refactored

client/transaction-pool/src/lib.rs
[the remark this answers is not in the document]
NikVolf: And refactored done

client/transaction-pool/src/testing/pool.rs
NikVolf: This will require bunch of confusing code that will exists only for 1 line in 2 tests. I would prefer not to.

client/transaction-pool/src/testing/pool.rs
NikVolf: added const * 2 factor

client/transaction-pool/src/testing/pool.rs
NikVolf: There should be no issues with this, really

It is not like we wait for something that _approximately_ finishes in 10ms. We know for sure it will be done in 5ms, so no flakiness here.

NikVolf:
> lgtm! @NikVolf did you have a chance to make sure it works correctly on a local chain or sth? Might be good to use @shawntabrizi script to populate the pool and see how it performs.

Yep, will do

NikVolf:
> lgtm! @NikVolf did you have a chance to make sure it works correctly on a local chain or sth? Might be good to use @shawntabrizi script to populate the pool and see how it performs.

I've checked intermediate versions with bench tool, will also check final one

NikVolf:
Checked with highload of 10000tx again, all ok!

PR #4782, opened 2020-01-30
Cleanup transaction pool deps

bkchr:


tomusdrw:
It seems that some elements of `testing.rs` rely on `substrate-test-runtime-client`. Perhaps we should make it optional and hide `testing` under a feature.

bkchr:
I'm already on it. I just had run `cargo test` locally..

test-utils/runtime/transaction-pool/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "substrate-test-runtime-transaction-pool"
+version = "2.0.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+license = "GPL-3.0"
+
```
gnunicorn:
```suggestion
publish = false

```
?

Dunno, haven't really settled on publishing or not publishing the test-crates yet ...

bkchr: The other crates currently also don't have it. And this it not really a test crate, more a "test-helper". So, something that can be reused and could be published IMHO.

PR #4784, opened 2020-01-31
fork-tree: add support for find_node_mut_where

sorpaas:
(Split out of #4600, for easier review/merge.)

This adds support to get mutable references of `Node` in a `ForkTree`, by using a helper method `find_node_index_where` that returns the indexes to access the node. Mutable reference is needed for consensus such as Sassafras, because we need to push VRF proofs after the epoch data is created.

### DIFF
--- a/utils/fork-tree/src/lib.rs
+++ b/utils/fork-tree/src/lib.rs
@@ -635,9 +635,9 @@ mod node_implementation {
 		/// The given function `is_descendent_of` should return `true` if the second hash (target)
 		/// is a descendent of the first hash (base).
 		///
-		/// The returned indices are from last to first, meaning the last is the least significant
-		/// child, and the first is the most significant child. An empty list means that the
-		/// current node is the result.
+		/// The returned indices are from last to first. The earliest index in the traverse path
+		/// goes last, and the final index in the traverse path goes first. An empty list means
+		/// that the current node is the result.
 		pub fn find_node_index_where<F, P, E>(
 			&self,
 			hash: &H,
--- a/utils/fork-tree/src/lib.rs
+++ b/utils/fork-tree/src/lib.rs
@@ -264,7 +264,7 @@ impl<H, N, V> ForkTree<H, N, V> where
 	{
 		// search for node starting from all roots
 		for root in self.roots.iter_mut() {
-			let node = root.find_node_mut_where(hash, number, is_descendent_of, predicate)?;
+			let node = root.find_node_where_mut(hash, number, is_descendent_of, predicate)?;
 
 			// found the node, early exit
 			if let FindOutcome::Found(node) = node {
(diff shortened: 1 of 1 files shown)


utils/fork-tree/src/lib.rs
```diff
@@ -250,6 +250,31 @@ impl<H, N, V> ForkTree<H, N, V> where
 		Ok(None)
 	}
 
+	/// Same as `find_node_where`, but returns mutable reference.
```
bkchr:
```suggestion
	/// Same as [`find_node_where`](Self::find_node_where), but returns mutable reference.
```

utils/fork-tree/src/lib.rs
```diff
@@ -655,14 +684,78 @@ mod node_implementation {
...
+				},
+			}
+		}
+
+		/// Find a node in the tree that is the deepest ancestor of the given
+		/// block hash which also passes the given predicate, backtracking
+		/// when the predicate fails.
+		/// The given function `is_descendent_of` should return `true` if the second hash (target)
+		/// is a descendent of the first hash (base).
+		pub fn find_node_mut_where<F, P, E>(
```
bkchr:
```suggestion
		pub fn find_node_where_mut<F, P, E>(
```

utils/fork-tree/src/lib.rs
```diff
@@ -250,6 +250,31 @@ impl<H, N, V> ForkTree<H, N, V> where
 		Ok(None)
 	}
 
+	/// Same as `find_node_where`, but returns mutable reference.
+	pub fn find_node_mut_where<F, E, P>(
```
bkchr:
```suggestion
	pub fn find_node_where_mut<F, E, P>(
```

utils/fork-tree/src/lib.rs
```diff
@@ -609,16 +634,17 @@ mod node_implementation {
 		/// when the predicate fails.
 		/// The given function `is_descendent_of` should return `true` if the second hash (target)
 		/// is a descendent of the first hash (base).
-		// FIXME: it would be useful if this returned a mutable reference but
-		// rustc can't deal with lifetimes properly. an option would be to try
-		// an iterative definition instead.
-		pub fn find_node_where<F, P, E>(
+		///
+		/// The returned indexes are from last to first, meaning the last is the least significant
```
rphmeier:
```suggestion
		/// The returned indices are from last to first, meaning the last is the least significant
```

rphmeier: it's not clear what "least significant" means in this comment, but from reading the code i take it to be the earlier index in the traversal path.

sorpaas: I updated the docs just calling it the "earliest" and "final" index.

rphmeier:
Thank you Wei!

PR #4785, opened 2020-01-31
Refactor epoch changes to a separate crate

sorpaas:
(Split off from #4600)

This refactors `EpochChanges` to be in a separate crate, so that it can be reused in other engines, such as Sassafras.


primitives/consensus/babe/src/digests.rs
```diff
@@ -41,7 +41,7 @@ use sp_std::vec::Vec;
 /// (VRF based) and to a secondary (slot number based).
 #[cfg(feature = "std")]
 #[derive(Clone, Debug)]
-pub enum BabePreDigest {
+pub enum PreDigest {
```
sorpaas: There are some naming changes. This is opinioned, but the rationale is that:

* `digest` mod is renamed to `digests`. This is to keep it consistent with `inherents` mod.
* `BabePreDigest` is renamed to `PreDigest`. We usually only use one engine at a time, and thus it is not common to have naming collisions of multiple `PreDigest`. Plus, it keeps it consistent with `NextEpochDescriptor`.

client/consensus/epochs/src/lib.rs
```diff
@@ -1,33 +1,10 @@
...
-//!
-//! This exposes the `SharedEpochChanges`, which is a wrapper around a
-//! persistent DAG superimposed over the forks of the blockchain.
-
-use std::sync::Arc;
-use sp_consensus_babe::{Epoch, SlotNumber, NextEpochDescriptor};
-use fork_tree::ForkTree;
-use parking_lot::{Mutex, MutexGuard};
-use sp_runtime::traits::{Block as BlockT, NumberFor, One, Zero};
+use std::{sync::Arc, ops::Add};
```
andresilva: Missing copyright header.

sorpaas:
@andresilva Can you take another look?

PR #4790, opened 2020-01-31
Small follow-up docs improvements in rent module

pepyakin:
This is follow-up to #4754 and #4733.

Runtime version doesn't need to be updated.

### DIFF
--- a/frame/contracts/src/rent.rs
+++ b/frame/contracts/src/rent.rs
@@ -132,7 +132,7 @@ fn rent_budget<T: Trait>(
 
 /// Consider the case for rent payment of the given account and returns a `Verdict`.
 ///
-/// Use `handicap` if case you want  change the reference block number. (Check out the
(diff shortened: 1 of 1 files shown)


frame/contracts/src/rent.rs
```diff
@@ -130,8 +132,8 @@ fn rent_budget<T: Trait>(
 
 /// Consider the case for rent payment of the given account and returns a `Verdict`.
 ///
-/// The `current_block_number` must be equal to the current block number. Use `handicap` do
-/// change the reference block number. (See `snitch_contract_should_be_evicted` for more details).
+/// Use `handicap` if case you want  change the reference block number. (Check out the
```
thiolliere:
```suggestion
/// Use `handicap` in case you want  change the reference block number. (Check out the
```

thiolliere: (also I usually prefer smaller sentence: See `..` function for more details)

PR #4791, opened 2020-01-31
Added GetCallMetadata for extrinsic calls.

gautamdhameja:
Trait to get function name and pallet name for the extrinsic calls.

### DIFF
--- a/frame/support/src/dispatch.rs
+++ b/frame/support/src/dispatch.rs
@@ -28,6 +28,7 @@ pub use crate::weights::{
 	TransactionPriority, Weight, WeighBlock, PaysFee,
 };
 pub use sp_runtime::{traits::Dispatchable, DispatchError, DispatchResult};
+pub use crate::traits::{CallMetadata, GetCallMetadata};
 
 /// A type that cannot be instantiated.
 pub enum Never {}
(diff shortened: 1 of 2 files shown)


frame/support/src/dispatch.rs
```diff
@@ -1302,42 +1303,52 @@ macro_rules! decl_module {
...
+								($( $param_name, )*)
+							);
+							$crate::dispatch::DispatchInfo { 
+								weight, 
+								class, 
+								pays_fee,
+							}
+						},
+					)*
+					_ => unreachable!("__PhantomItem should never be used."),
```
thiolliere: If the last variant is only the phantom item then we should be able to match it directly `$call_type::__PhantomItem(_, _)`

frame/support/src/dispatch.rs
```diff
@@ -1302,42 +1303,52 @@ macro_rules! decl_module {
...
+		impl<$trait_instance: $trait_name $(<I>, $instance: $instantiable)?> $crate::dispatch::GetCallMetadata
+			for $call_type<$trait_instance $(, $instance)?> where $( $other_where_bounds )*
+		{
+			fn get_call_metadata(&self) -> $crate::dispatch::CallMetadata {
+				match *self {
+					$(
+						$call_type::$fn_name( $( ref $param_name ),* ) => {
+							let function_name = stringify!($fn_name);
+							// We assign the pallet_name as None as it is not available here.
+							// It is assigned a value where we call this function in impl_outer_dispatch.
```
thiolliere: ok or we could provide two traits one for the outer call one for the module call

frame/support/src/dispatch.rs
```diff
@@ -2084,4 +2107,11 @@ mod tests {
 		assert_eq!(<Test as WeighBlock<u32>>::on_finalize(2), 10);
 		assert_eq!(<Test as WeighBlock<u32>>::on_finalize(3), 0);
 	}
+
+	#[test]
+	fn call_metadata() {
+		let metadata = Call::<TraitImpl>::aux_3().get_call_metadata();
+		let expected = CallMetadata { function_name: "aux_3".into(), pallet_name: None };
+		assert_eq!(metadata, expected);
+	}
```
thiolliere: we could also do a test for the outer_call metadata, maybe in frame/support/src/lib.rs there is a test for metadata we can use impl_outer_dispatch and test it there 🤔

frame/support/src/dispatch.rs
bkchr: Yeah, I'm with @thiolliere

frame/support/src/dispatch.rs
```diff
@@ -1302,42 +1303,52 @@ macro_rules! decl_module {
...
+							let function_name = stringify!($fn_name);
+							// We assign the pallet_name as None as it is not available here.
+							// It is assigned a value where we call this function in impl_outer_dispatch.
+							$crate::dispatch::CallMetadata { 
+								function_name,
+								pallet_name: None,
+							}
+						},
+					)*
+					_ => unreachable!("__PhantomItem should never be used."),
```
bkchr: The same as @thiolliere proposed above with matching the `PhantomItem` directly.

frame/support/src/traits.rs
```diff
@@ -802,3 +803,17 @@ pub trait ModuleToIndex {
...
+#[derive(Clone, Eq, PartialEq, Default, RuntimeDebug)]
+/// The function and pallet name of the Call.
+pub struct CallMetadata {
+	/// Name of the function.
+	pub function_name: &'static str,
+	/// Name of the pallet to which the function belongs.
+	pub pallet_name: Option<&'static str>,
+}
+
+pub trait GetCallMetadata {
```
bkchr: Trait should also have some documentation.

frame/support/src/dispatch.rs
```diff
@@ -1302,42 +1303,47 @@ macro_rules! decl_module {
...
+		}
 
+		// Implement GetCallName for the Call.
+		impl<$trait_instance: $trait_name $(<I>, $instance: $instantiable)?> $crate::dispatch::GetCallName
+			for $call_type<$trait_instance $(, $instance)?> where $( $other_where_bounds )*
+		{
+			fn get_call_name(&self) -> &'static str {
+				match *self {
+					$(
+						$call_type::$fn_name( $( ref $param_name ),* ) => {
```
thiolliere:
```suggestion
						$call_type::$fn_name( $( ref $param_name ),* ) => {
```
```suggestion
						$call_type::$fn_name(..) => {
```

frame/support/src/dispatch.rs
```diff
@@ -1302,42 +1303,47 @@ macro_rules! decl_module {
...
+							let function_name = stringify!($fn_name);
```
bkchr:
```suggestion
							stringify!($fn_name)
```

frame/support/src/dispatch.rs
```diff
@@ -1302,42 +1303,47 @@ macro_rules! decl_module {
...
+							function_name
```
bkchr:
```suggestion
```

frame/support/src/traits.rs
```diff
@@ -800,3 +801,24 @@ pub trait ModuleToIndex {
...
+
+/// Gets the function name of the Call.
+pub trait GetCallName {
+	/// Return the function name of the Call.
+	fn get_call_name(&self) -> &'static str;
+}
+
+/// Gets the metadata for the Call - function name and pallet name.
+pub trait GetCallMetadata {
+	/// Return a `CallMetadata`, containing function and pallet name of the Call.
```
```suggestion
	/// Return a [`CallMetadata`], containing function and pallet name of the Call.
```

frame/support/src/traits.rs
```diff
@@ -800,3 +801,24 @@ pub trait ModuleToIndex {
 impl ModuleToIndex for () {
 	fn module_to_index<M: 'static>() -> Option<usize> { Some(0) }
 }
+
+#[derive(Clone, Eq, PartialEq, Default, RuntimeDebug)]
```
```suggestion
```

frame/support/src/traits.rs
```diff
@@ -800,3 +801,24 @@ pub trait ModuleToIndex {
...
+/// The function and pallet name of the Call.
+pub struct CallMetadata {
```
```suggestion
#[derive(Clone, Eq, PartialEq, Default, RuntimeDebug)]
pub struct CallMetadata {
```

frame/support/src/dispatch.rs
gautamdhameja: This change did not compile. Not sure if I am missing something.

frame/support/src/dispatch.rs
```diff
@@ -1302,42 +1303,46 @@ macro_rules! decl_module {
...
 
+		// Implement GetCallName for the Call.
+		impl<$trait_instance: $trait_name $(<I>, $instance: $instantiable)?> $crate::dispatch::GetCallName
+			for $call_type<$trait_instance $(, $instance)?> where $( $other_where_bounds )*
+		{
+			fn get_call_name(&self) -> &'static str {
+				match *self {
+					$(
+						$call_type::$fn_name( $( ref $param_name ),* ) => {
+							stringify!($fn_name)
```
bkchr:
```suggestion
							// Don't generate any warnings for unused variables
							let _ = ( $( $param_name ),* );
							stringify!($fn_name)
```

PR #4792, opened 2020-01-31
Increase transaction pool default limits

NikVolf:
Also improve logging in authorship/txpool

### DIFF
--- a/client/transaction-pool/graph/src/validated_pool.rs
+++ b/client/transaction-pool/graph/src/validated_pool.rs
@@ -150,9 +150,16 @@ impl<B: ChainApi> ValidatedPool<B> {
 		let future_limit = &self.options.future;
 
 		debug!(target: "txpool", "Pool Status: {:?}", status);
-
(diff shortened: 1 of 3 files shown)


client/transaction-pool/graph/src/validated_pool.rs
```diff
@@ -150,9 +150,16 @@ impl<B: ChainApi> ValidatedPool<B> {
...
 
 		debug!(target: "txpool", "Pool Status: {:?}", status);
-
 		if ready_limit.is_exceeded(status.ready, status.ready_bytes)
-			|| future_limit.is_exceeded(status.future, status.future_bytes) {
+			|| future_limit.is_exceeded(status.future, status.future_bytes)
+		{
+			debug!(
+				target: "txpool",
+				"Enforcing limits ({}/{}b ready, {}/{}b future",
```
gavofyork: I think this would be better formatted as KB, no?

NikVolf: sure, but it's a technical trace

PR #4793, opened 2020-01-31
Adding Serde Deserialisation to UncheckExtrinsics

gnunicorn:
to be able to use the ChainApi RPC from the client side. Of course guarded behind `std`.


primitives/runtime/src/generic/unchecked_extrinsic.rs
```diff
@@ -274,6 +274,18 @@ impl<Address: Encode, Signature: Encode, Call: Encode, Extra: SignedExtension> s
 	}
 }
 
+#[cfg(feature = "std")]
+impl<'a, Address: Decode, Signature: Decode, Call: Decode, Extra: SignedExtension> serde::Deserialize<'a>
+	for UncheckedExtrinsic<Address, Call, Signature, Extra> {
```
bkchr:
```suggestion
	for UncheckedExtrinsic<Address, Call, Signature, Extra>
{
```

PR #4794, opened 2020-01-31
Remove ambiguous into_iter

thiolliere:
Fix the compiler warning:
```
warning: this method call currently resolves to `<&[T; N] as IntoIterator>::into_iter` (due to autoref coercions), but that might change in the future when `IntoIterator` impls for arrays are added.
   --> client/cli/src/node_key.rs:169:28
    |
169 |             NodeKeyType::variants().into_iter().try_for_each(|t| {
    |                                     ^^^^^^^^^ help: use `.iter()` instead of `.into_iter()` to avoid ambiguity: `iter`
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #66145 <https://github.com/rust-lang/rust/issues/66145>

```

Note that all instance of this code are in test so that doesn't even matter to right that this could be improved in the future once into_iter for array is implemented.

### DIFF
--- a/client/api/src/notifications.rs
+++ b/client/api/src/notifications.rs
@@ -323,7 +323,7 @@ mod tests {
 			let child_filters = Some([
 				(StorageKey(vec![4]), None),
 				(StorageKey(vec![5]), None),
-			].into_iter().cloned().collect());
+			].iter().cloned().collect());
 			StorageChangeSet {
 				changes: Arc::new(changes.0),
 				child_changes: Arc::new(changes.1),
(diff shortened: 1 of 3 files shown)


client/api/src/notifications.rs
```diff
@@ -323,7 +323,7 @@ mod tests {
 			let child_filters = Some([
 				(StorageKey(vec![4]), None),
 				(StorageKey(vec![5]), None),
-			].into_iter().cloned().collect());
+			].iter().cloned().collect());
```
niklasad1:
```suggestion
			].iter().copied().collect());
```

PR #4795, opened 2020-01-31
client/finality-grandpa: Add regression test for observer polling network (was #4778)

gavofyork:
Same as #4778, except with nits fixed.


mxinden:
Thanks @gavofyork.

Issue #4641, opened 2020-01-16
Thread 'event.loop0' panicked at 'Expected always-serializable type; qed: Error("u128 is not supported", line: 0, column: 0)', src/libcore/result.rs:1165


IPSE-TEAM:
Version: 2.0.0-401c373d-x86_64-macos

stack backtrace:
   0: backtrace::backtrace::trace
   1: backtrace::capture::Backtrace::new
   2: sp_panic_handler::set::{{closure}}
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::continue_panic_fmt
   5: rust_begin_unwind
   6: core::panicking::panic_fmt
   7: core::result::unwrap_failed
   8: <futures::future::map::Map<A,F> as futures::future::Future>::poll
   9: <futures::future::map_err::MapErr<A,F> as futures::future::Future>::poll
  10: <futures::future::either::Either<A,B> as futures::future::Future>::poll
  11: <futures::future::lazy::Lazy<F,R> as futures::future::Future>::poll
  12: futures::future::chain::Chain<A,B,C>::poll
  13: <futures::future::then::Then<A,B,F> as futures::future::Future>::poll
  14: <futures::future::map::Map<A,F> as futures::future::Future>::poll
  15: <futures::future::either::Either<A,B> as futures::future::Future>::poll
  16: <futures::future::map::Map<A,F> as futures::future::Future>::poll
  17: <futures::future::select::Select<A,B> as futures::future::Future>::poll
  18: <futures::future::map::Map<A,F> as futures::future::Future>::poll
  19: <futures::future::map_err::MapErr<A,F> as futures::future::Future>::poll
  20: futures::task_impl::std::set
  21: std::panicking::try::do_call
  22: __rust_maybe_catch_panic
  23: tokio_threadpool::task::Task::run
  24: tokio_threadpool::worker::Worker::run_task
  25: tokio_threadpool::worker::Worker::run
  26: tokio_reactor::with_default
  27: tokio::runtime::threadpool::builder::Builder::build::{{closure}}
  28: std::thread::local::LocalKey<T>::with
  29: std::thread::local::LocalKey<T>::with
  30: std::sys_common::backtrace::__rust_begin_short_backtrace
  31: std::panicking::try::do_call
  32: __rust_maybe_catch_panic
  33: core::ops::function::FnOnce::call_once{{vtable.shim}}
  34: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
[...]

PR #4798, opened 2020-02-01
Reorganize frame support storage

thiolliere:
I tried to reorganize a bit storage module in frame-support so that it is more understandable.
Note that it should mostly not break any code but print a lot of warning for deprecated functions.

### Done:

* unhashed and hashed module are renamed top and top_hashed, as they a tools to abstract of top trie, child module keep its name child.
  (unhashed and hashed still exists but deprecated)
* generator::Storage* and its corresponding storage trait have been merged, so now there is only one trait for each Storage trait implemented by decl_storage.
  though the generator is now reexported where the other trait has been removed.
  note: I feel like exposing the full trait directly will make it more understandable about the underlying implementation.
* `storage_map_final_key` and `hashed_key_for` functions has been deprecated for top_trie_key. This is because we want to make use of childtries in decl_storage in the future, and having such functions which doesn't make clear where the value is from will I think lead to confusion. (#4641 )

### Note:

* so nothing should break but user should expect to see a lot of deprecated warnings.
* maybe the naming of top for the top storage is debatable, though it is already use for instance here: https://substrate.dev/rustdocs/master/sp_runtime/struct.Storage.html
* one of the goal for making use of top_trie_key is that we could still support it under a where condition where we generalize with childtries storage.


bkchr:
When we will make use of child storage in modules, why would the `key` be different between the normal storage and the child storage? (I'm not speaking about the key that we will use to access the backend) I would expect that we use the same key and just use `child_*` to access the storage.

Naming it top storage doesn't make it less confusing IMHO.

thiolliere:
> When we will make use of child storage in modules, why would the `key` be different between the normal storage and the child storage? (I'm not speaking about the key that we will use to access the backend) I would expect that we use the same key and just use `child_*` to access the storage.
> 
> Naming it top storage doesn't make it less confusing IMHO.

I was thinking it provides more type guarantee, let's say in the code one do call `MyStorage::hashed_key()` and upgrade its module to use childtries instead of top trie then if they forget to upgrade this usage of `MyStorage::hashed_key()` their logic is potentially broken.

So my idea was to provide a `fn trie_key() -> TrieKey` with `enum TrieKey { TopTrie(Vec<u8>), ChildTrie(...) }` and keep and rewrite the function top_trie_key as such:
```
trait StorageMap {
  type StorageSpace;
  fn trie_key() -> TrieKey;
  fn top_trie_key() -> Vec<u8> where MyStorageSpace: TopTrieStorageSpace;
}
```
or something similar

Thus usage of top_trie_key will only be available for Storage on top trie, making the upgrade of a module from top trie to child trie less error prone.

Though maybe I overthink type safety there, I can be ok this to be rejected.

thiolliere:
I close it as it needs some other thoughts and can't be done before 2.0

thiolliere:
Currently generators are hidden from the doc and the internal doc of frame/support/src/storage/generator says

>  This is internal api and is subject to change.

It make sense to me that people should not use the associated type storage_prefix directly, as this associated type can be removed once we switch to compile time prefix hash.

But on the other hand it make people confuse about what those generator are, decl_storage refers to them, and we actually use it in `./utils/frame/rpc/support/src/lib.rs` for example.

Is it better to make generator module not hidden anymore ? I'm really mixed with the situation, I feel like this private API is difficult for user to deal with.

Issue #4800, opened 2020-02-01
Test fails on Windows


Swader:
Post-build I run `cargo test --all` and get this result:

```
   Compiling frame-support-test v2.0.0 (C:\Users\Bruno\repos\substrate\frame\support\test)
error: linking with `link.exe` failed: exit code: 1120
  |
[...]

PR #4802, opened 2020-02-01
Don't attempt to upgrade balances module on chain start

gavofyork:
Also,
- introduce vesting to node; and
- debug message for upgrades.


Swader:
This does not fix #4800

PR #4803, opened 2020-02-01
Additional RPC for dumping all main storage key pairs under a prefix

gavofyork:


bkchr:
Isn't that a duplicate of https://github.com/paritytech/substrate/pull/4718 ?

xlc:
Without paging, this can easily causing trouble when someone want to request the whole state.

I think this should be a follow up of #4718 to introduce a variant that also return values along with keys.

tomusdrw:
Agree with @xlc, the output would be huge and can easily kill the node. Perhaps we could put such things to a separate `debug` namespace and only allow them to be exposed over `localhost`. Even though I feel that we should not endorse developers to use them for building UIs or dapps, cause I feel it will backfire in the (near) future.

gavofyork:
This isn't designed for general use. This was a piece of functionality I actually *needed* in order to properly generate a migration test and that I could conceivably need again. I could have alternatively placed an additional CLI mode, but this is quite a lot lighter weight. Introducing paging would have resulted in more complex, less optimal code with a greater amount of effort on the part of the user (i.e. me) in driving it as I would have had to faff around scripting it to fetch and merge all pages.

Refusing to include a useful RPC just because it might conceivably be misused is dumb. We're not creating children's toys here. Of course if it turns out to be too great a foot-gun, then it can perhaps be hidden behind a `--expensive-rpcs` flag or something, but that's its own issue.

gavofyork:
> Isn't that a duplicate of #4718 ?

No. Quite the opposite, in fact.

client/src/client.rs
```diff
@@ -235,13 +235,33 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
+	/// Given a `BlockId` and a key prefix, return the matching child storage keys and values in that block.
+	pub fn storage_pairs(&self, id: &BlockId<Block>, key_prefix: &StorageKey)
+		-> sp_blockchain::Result<Vec<(StorageKey, StorageData)>>
+	{
+		let state = self.state_at(id)?;
+		let keys = state
+			.keys(&key_prefix.0)
+			.into_iter()
+			.map(|k| {
+				let d = state.storage(&k).unwrap_or(None).unwrap_or(vec![]);
```
bkchr:
```suggestion
				let d = state.storage(&k).ok().unwrap_or_default();
```

gavofyork: that doesn't do the same thing, does it?

client/src/client.rs
```diff
@@ -235,13 +235,33 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
+	/// Given a `BlockId` and a key prefix, return the matching child storage keys and values in that block.
+	pub fn storage_pairs(&self, id: &BlockId<Block>, key_prefix: &StorageKey)
+		-> sp_blockchain::Result<Vec<(StorageKey, StorageData)>>
+	{
+		let state = self.state_at(id)?;
+		let keys = state
+			.keys(&key_prefix.0)
+			.into_iter()
+			.map(|k| {
+				let d = state.storage(&k).ok().unwrap_or_default();
```
bkchr:
```suggestion
				let d = state.storage(&k).ok().flatten().unwrap_or_default();
```

client/src/client.rs
bkchr: Yeah, I had forgotten a `.flatten()` :( Sorry. FWIW, I also merged master now.

xlc:
I guess this will be another "unsafe" RPC to be exposed publicly. Some docs here? https://github.com/paritytech/substrate/wiki/Public-RPC

tomusdrw:
> Refusing to include a useful RPC just because it might conceivably be misused is dumb. We're not creating children's toys here. It is not the job of developers to second-guess the RPC-user and force them to jump through hoops of tiny RPC calls in order to allow them to do their job.

Sure thing, I didn't try to block the functionality, but proposed two alternative solutions that would make it less of a footgun (or easier to communicate that it's unsafe):
1. Either paged approach (like `getKeysPaged)
2. Or separate namespace (`debug_`)

If we include it in current form we will in future either have to deprecate it and ask everyone to move or have inconsistent set of APIs within a namespace depending on a particular flag (filtering the entire namespace is easier and I think more clear regarding safeness/unsafeness).

gavofyork:
> Either paged approach (like `getKeysPaged)

As I explained above, I do not consider that an option.

I'm ok with the idea of prefixing it with `unstable_`, `expensive_` or some other label indicative of its use or that it may yet change.

gavofyork:
> I guess this will be another "unsafe" RPC to be exposed publicly. Some docs here (when this is merged)? https://github.com/paritytech/substrate/wiki/Public-RPC

Done.

PR #4804, opened 2020-02-01
Avoid challenging those that can't be suspended anyway

gavofyork:


frame/society/src/lib.rs
```diff
@@ -1536,18 +1536,22 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
-			let seed = T::Randomness::random(phrase);
-			// seed needs to be guaranteed to be 32 bytes.
-			let seed = <[u8; 32]>::decode(&mut TrailingZeroInput::new(seed.as_ref()))
-				.expect("input is padded with zeroes; qed");
-			let mut rng = ChaChaRng::from_seed(seed);
-			let chosen = pick_item(&mut rng, &members).expect("exited if members empty; qed");
-
-			<Defender<T, I>>::put(&chosen);
-			Self::deposit_event(RawEvent::Challenged(chosen.clone()));
+			if members.len() > 2 {
```
bkchr: I think having a comment explaining that `Head` and `Founder` can not be removed/suspended would be helpful here.

PR #4806, opened 2020-02-02
MallocSizeOf for runtime primitives

NikVolf:
Opening mostly to discuss if such bounds on basic traits is acceptable. Actual types can trivially implement memory-tracking `MallocSizeOf` (we maintain mozilla clone of this trait in `parity-common`) or already implement it (like `H256`, `Vec<u8>`, etc.)

pros
`+` will allow track memory of all subsystems (transaction pool, networking, consensus)

cons
`-` more code

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1107,6 +1107,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "ethbloom"
+version = "0.8.1"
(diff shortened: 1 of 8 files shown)


bkchr:
When it allows to track the memory usage, I think it is a good idea. We really need to have something to track the memory-usage of the individual systems.

PR #4807, opened 2020-02-02
babe: pass epoch data via intermediates

sorpaas:
Fixes https://github.com/paritytech/substrate/pull/4652#issuecomment-577193918

This passes the epoch data via block import intermediates, avoiding a second call to `epoch_changes.epoch_for_child_of`. ~~It does require encoding/decoding of the data when passing, which, under the current design of `BlockImportParams`, I don't think is avoidable.~~ (Fixed in #4809)


client/consensus/babe/src/lib.rs
```diff
@@ -963,20 +985,29 @@ impl<B, E, Block, I, RA, PRA> BlockImport<Block> for BabeBlockImport<B, E, Block
...
-					babe_err(Error::<Block>::BlockNotValid(hash)).into()
-				))?;
-
+			let intermediate = block.take_intermediate::<BabeIntermediate>(
+				INTERMEDIATE_KEY
+			)?;
+
+			let epoch = match intermediate.epoch {
+				Some(epoch) => epoch,
+				None => {
```
sorpaas: This branch is needed because when proposing blocks, it won't invoke the verifier. In those cases the epoch must be re-fetched.

sorpaas:
Hmm looks like there's an error in tests. Let me debug.

bkchr:
> When authoring blocks we still fetch epoch data from the slot worker in order to author the block. We could change `block_import_params` to also take `EpochData`, which we could then use to populate the intermediate. We could then change `BabeIntermediate` to not use `Option`. WDYT?

Sounds like a good idea to me. :+1:

sorpaas:
Integration test fixed!

sorpaas:
PR is ready, but I want to merge it after https://github.com/paritytech/substrate/pull/4785 is merged. So marking it `in-progress` for now.

PR #4811, opened 2020-02-03
Node template folders restructuring

jimmychu0807:
While I was working on the offchain worker project, I am figuring what's the best way to store node, runtime, and the price-fetch pallet. I realized node to runtime is a 1-to-1 relationship. Runtime to pallet is a 1-to-many relationship. The reason we put runtime into a separate folder is because it requires a different compilation method from node. We need a build.rs to compile runtime to both native and wasm.

I propose restructuring the node-template to the following (with @gnunicorn dir structure suggestion):
Also taking @riusricardo suggestion of putting mock and tests in a single file tests.rs.

```
node
  // this is the node, main entry point. 
  // folder renamed from `src` to `node` to make its meaning clearer
  L main.rs
  L ...
  L runtime
    // this is our runtime
    L src
      L lib.rs
    L build.rs
    L Cargo.toml
pallets
  L template
    // this is our `template` pallet 
    L src
      L lib.rs
      L tests.rs
    L Cargo.toml
Cargo.toml
```

The advantage are:

* It is clear to dev that the node, runtime, and pallets are "separated".
* Tests are separated from the main code.
* You can easily extend this structure to write multiple pallets.
* We are also introducing the structure of substrate main codebase to dev early on and they can familiarize to this structure.
* With this structure, I think we can phase out `substrate-pallet-template`, as the `pallet-template` inside `/pallets` is already a good demonstration of how pallets are stored and included in the outer node and runtime. With its standalone code, tests & mock runtime, and its Cargo.toml, one just need to take this folder out and it is a good pallet to upload to git / cargo crate.

I am aware this will cause some documentation change. This PR can be merge after all code review **AND** after devEx team have updated the docs/tutorial about the node template. But I feel this is a better way to structure node template in the long run.

Context and discussion:
- https://github.com/substrate-developer-hub/substrate-node-template/pull/22  (this is also how the downstream node-template will look like)
- paritytech/substrate#3769


bin/node-template/pallets/template/src/tests.rs
```diff
@@ -0,0 +1,75 @@
+// Tests to be written here
+#![cfg(test)]
```
bkchr:
```suggestion
```

bin/node-template/pallets/template/src/lib.rs
```diff
@@ -1,16 +1,19 @@
...
 /// If you remove this file, you can remove those references
 
-
 /// For more guidance on Substrate modules, see the example module
 /// https://github.com/paritytech/substrate/blob/master/frame/example/src/lib.rs
 
 use frame_support::{decl_module, decl_storage, decl_event, decl_error, dispatch};
 use system::ensure_signed;
 
+mod tests;
```
bkchr:
```suggestion
#[cfg(test)]
mod tests;
```

bin/node-template/Cargo.toml
```diff
@@ -2,13 +2,13 @@
...
 authors = ["Anonymous"]
-build = "build.rs"
 edition = "2018"
 license = "Unlicense"
+build = "node/build.rs"
 
 [[bin]]
 name = "node-template"
-path = "src/main.rs"
+path = "node/main.rs"
```
bkchr: Why do we not follow default rust project structure?

jimmychu0807: The reason is to make our intention clear. `/src` is where rust developers would quickly focus on and regarded as their main workspace. But in node template, I believe node is not their focus, but `/pallets` should be. We have a `pallet-template` to demonstrate what a simplest pallet would look like. They can play with this pallet template, or extend to write additional pallets and integrate it in the runtime.

Also this new naming make it clear there are `node`, `runtime` that is under node, and (possibly) multiple pallets.

I will confirm on the release script.

bkchr: Node needs also to be modified when you change something on the chain spec for example.

The structure I imagine:
node-template
   - node
       - src
       - Cargo.toml
   - runtime
       - src 
       - Cargo.toml
       - build.rs
   - pallets
       - template
          - src
   - Cargo.toml 

People already always complain that we don't follow default Rust etc and in this case  don't see the reason to not follow default Rust.

jimmychu0807: Cool. Working on the release script so it replace the path in pallet template

jimmychu0807:
1. This is the final structure inside `bin/node-template`:

```
Cargo.toml // this is a virtual manifest, containing workspaces node, runtime, and pallet-template
node
  L src
    L ...
  L build.rs
  L Cargo.toml
runtime
  L src
    L ...
  L build.rs
  L Cargo.toml
pallets
  L template
    L src
      L lib.rs
      L mock.rs
      L tests.rs
    L Cargo.toml
```

2. `tests.rs` in pallet-template is separated to `mock.rs` and `tests.rs`

3. Updated `node-template-release.sh` to create the topmost `Cargo.toml` and tested locally that cut a new node-template out.

See if this is good. 

@gnunicorn  @shawntabrizi  @bkchr

jimmychu0807:
Basti, I have an error in the CI test

```
 ERROR: failed to execute cargo: No such file or directory (os error 2)
 test ui ... FAILED
 failures:
 ---- ui stdout ----
 thread 'ui' panicked at 'tests failed', /ci-cache/substrate/cargo/test-linux-stable/registry/src/github.com-1ecc6299db9ec823/trybuild-1.0.21/src/run.rs:38:13
```

Could you point me on what I should do to fix this error?

bkchr:
> Basti, I have an error in the CI test
> 
> ```
>  ERROR: failed to execute cargo: No such file or directory (os error 2)
>  test ui ... FAILED
>  failures:
>  ---- ui stdout ----
>  thread 'ui' panicked at 'tests failed', /ci-cache/substrate/cargo/test-linux-stable/registry/src/github.com-1ecc6299db9ec823/trybuild-1.0.21/src/run.rs:38:13
> ```
> 
> Could you point me on what I should do to fix this error?

This was some spurious CI failure, I restarted the job.

PR #4812, opened 2020-02-03
CLI improvements & fixes

cecton:
These are a few changes I missed during the refactoring.

1. Initialization issue and boilerplate

    Most importantly: part of the `Configuration` initialization was done in `sc_cli::init`. This means the user can not benefit from this initialization boilerplate if they have multiple `Configuration` since `sc_cli::init` can only be called once.

2. Boilerplate for `VersionInfo` and `Configuration`

    I'm also answering to the critic of @bkchr on the initialization using version: https://github.com/paritytech/substrate/pull/4692/files/bea809d4c14a2ede953227ac885e3b3f9771c548#r372047238 This will allow initializing a `Configuration` and provide the version by default.

3. Loading the `chain_spec` explicitly

    In the past it was done automatically but in some cases we want to delay this. I moved the code to `Configuration.load_spec()` so it can be called later on. `chain_spec` can also be written directly to the `Configuration` without using this `load_spec` helper.

4. ~~Make `parse_address` public~~

    ~~This change is necessary to allow the user to change the default addresses of `rpc_http`, `rpc_ws` and `grafana_port`.~~

    [edit: I don't need that!]

5. Fixing issue that prevents the user to override the port

    In the refactoring I introduced a bug by mistake that could potentially prevent the CLI user to override the ports if defaults where provided for these ports (only on cumulus).

6. Change task_executor from Box to Arc

    This is useful for cumulus where we have 2 nodes with 2 separate Configuration that need to spawn tasks to the same runtime.

7. Renamed TasksExecutorRequired to TaskExecutor

    For consistency.

This is related to https://github.com/paritytech/cumulus/issues/24

This is the continuation (and hopefully the end of) #4692


client/cli/src/lib.rs
```diff
@@ -35,6 +35,7 @@ use sc_service::{
 	RuntimeGenesis, ChainSpecExtension, PruningMode, ChainSpec,
 	AbstractService, Roles as ServiceRoles,
 };
+pub use sc_service::config::{VersionInfo};
```
```suggestion
pub use sc_service::config::VersionInfo;
```

client/cli/src/lib.rs
```diff
@@ -79,26 +80,18 @@ const DEFAULT_KEYSTORE_CONFIG_PATH : &'static str =  "keystore";
...
-	/// Executable file author.
-	pub author: &'static str,
-	/// Support URL.
-	pub support_url: &'static str,
-	/// Copyright starting year (x-current year)
-	pub copyright_start_year: i32,
-}
+#[cfg(test)]
+#[doc(hidden)]
+pub const TEST_VERSION_INFO: &'static VersionInfo = &VersionInfo {
```
bkchr: Why is that not defined in the tests mod below?

client/service/src/config.rs
```diff
@@ -28,6 +28,27 @@ use sp_core::crypto::Protected;
...
+	/// Executable file description.
+	pub description: &'static str,
+	/// Executable file author.
+	pub author: &'static str,
+	/// Support URL.
+	pub support_url: &'static str,
+	/// Copyright starting year (x-current year)
+	pub copyright_start_year: i32,
+}
+
```
cecton: This is moved here because I need it in the `Configuration::new()` and actually this is fine. Part of the logic that determine the version, OS, arch... was already here. To keep the original API I re-export `VersionInfo` in `sc_cli` too,

client/cli/src/lib.rs
```diff
@@ -79,26 +80,18 @@ const DEFAULT_KEYSTORE_CONFIG_PATH : &'static str =  "keystore";
...
+	name: "node-test",
+	version: "0.1.0",
+	commit: "some_commit",
+	executable_name: "node-test",
+	description: "description",
+	author: "author",
+	support_url: "http://example.org",
+	copyright_start_year: 2020,
+};
```
cecton: I believe this is going to be useful more than once so I put it there.

client/cli/src/lib.rs
cecton: I believe it can be used in other tests in other crates. I thought it would be a good idea. What do you think?

cecton: (I plan to write some tests for the CLI)

client/cli/src/lib.rs
```diff
@@ -625,16 +613,16 @@ where
 			}
 		});
 
-	if config.rpc_http.is_none() {
+	if config.rpc_http.is_none() || cli.rpc_port.is_some() {
```
bkchr: Why? Isn't the port setting the `rpc_http`?

client/cli/src/lib.rs
```diff
@@ -625,16 +613,16 @@ where
...
 		config.rpc_http = Some(parse_address(&format!("{}:{}", rpc_interface, 9933), cli.rpc_port)?);
 	}
-	if config.rpc_ws.is_none() {
+	if config.rpc_ws.is_none() || cli.ws_port.is_some() {
 		let ws_interface: &str = interface_str(cli.ws_external, cli.unsafe_ws_external, cli.validator)?;
 		config.rpc_ws = Some(parse_address(&format!("{}:{}", ws_interface, 9944), cli.ws_port)?);
 	}
 
-	if config.grafana_port.is_none() {
+	if config.grafana_port.is_none() || cli.grafana_port.is_some() {
```
bkchr: Same.

client/cli/src/lib.rs
```diff
@@ -815,4 +804,60 @@ mod tests {
...
+		update_config_for_running_node(&mut config, cli, TEST_VERSION_INFO).unwrap();
+
+		assert!(config.config_dir.is_some());
+		assert!(config.database.is_some());
+		if let Some(DatabaseConfig::Path { ref cache_size, .. }) = config.database {
+			assert!(cache_size.is_some());
+		} else {
+			panic!("invalid config.database variant");
+		}
+		assert_ne!(config.name, "");
```
bkchr: Either check for the expected name or: 
```suggestion
		assert!(!config.name.is_empty());
```

client/cli/src/lib.rs
cecton: On `Configuration` yes but here it's on `RunCmd` (cli). CLI is the user input and if it is provided it must override the one configured (if there is one)

bin/node/cli/tests/running_the_node_and_interrupt.rs
```diff
@@ -0,0 +1,22 @@
...
+	sleep(Duration::from_secs(30));
+	assert!(cmd.try_wait().unwrap().is_none(), "the process should still be running");
+	unsafe { kill(cmd.id() as i32, SIGINT) };
+	assert!(cmd.wait().unwrap().success(), "the process must exit gracefully after SIGINT");
+
+	let mut cmd = Command::new(cargo_bin("substrate")).spawn().unwrap();
+	sleep(Duration::from_secs(30));
+	assert!(cmd.try_wait().unwrap().is_none(), "the process should still be running");
+	unsafe { kill(cmd.id() as i32, SIGTERM) };
+	assert!(cmd.wait().unwrap().success(), "the process must exit gracefully after SIGTERM");
```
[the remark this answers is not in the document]
cecton: Right, I didn't see it, thanks!

bin/node/cli/tests/running_the_node_and_interrupt.rs
```diff
@@ -1,22 +1,55 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2020 Parity Technologies (UK) Ltd.
```

Didn't know that you can time travel :D

bin/node/cli/Cargo.toml
```diff
@@ -96,7 +96,7 @@ sc-service-test = { version = "2.0.0", path = "../../../client/service/test" }
 futures = "0.3.1"
 tempfile = "3.1.0"
 assert_cmd = "0.12"
-libc = "0.2"
+nix = "0.17"
```
bkchr: Don't we need to make this an optional dependency that is only activated for unix?

cecton: I asked myself the same question but I saw this in the source code:

https://docs.rs/nix/0.17.0/src/nix/lib.rs.html#6

The whole lib is in unix only unless I'm mistaking

cecton: I will check on Windows to be sure

cecton: Confirmed.

PR #4814, opened 2020-02-03
Fix set of earliest unapplied slashes on offence

thiolliere:
When a new offence is created is update the earliest unapplied storage only if none exist.

But this is used by apply_unapplied_slashed
https://github.com/paritytech/substrate/blob/9195fac755df57e18ed59fbd358ee528d2a52d0e/frame/staking/src/lib.rs#L1460-L1474

Thus an offence previous to `current_era - slash_defer_duration` will never be applied.

### DIFF
--- a/frame/staking/src/lib.rs
+++ b/frame/staking/src/lib.rs
@@ -1765,9 +1765,7 @@ impl <T: Trait> OnOffenceHandler<T::AccountId, pallet_session::historical::Ident
 		};
 
 		<Self as Store>::EarliestUnappliedSlash::mutate(|earliest| {
-			if earliest.is_none() {
-				*earliest = Some(era_now)
-			}
+			*earliest = Some(earliest.unwrap_or(era_now).min(era_now));
 		});
 
 		let slash_defer_duration = T::SlashDeferDuration::get();
--- a/bin/node/runtime/src/lib.rs
+++ b/bin/node/runtime/src/lib.rs
@@ -79,7 +79,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// and set impl_version to 0. If only runtime
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
-	spec_version: 212,
+	spec_version: 213,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 };


frame/staking/src/lib.rs
```diff
@@ -1765,9 +1765,7 @@ impl <T: Trait> OnOffenceHandler<T::AccountId, pallet_session::historical::Ident
 		};
 
 		<Self as Store>::EarliestUnappliedSlash::mutate(|earliest| {
-			if earliest.is_none() {
-				*earliest = Some(era_now)
-			}
+			*earliest = Some(earliest.unwrap_or(era_now).min(era_now));
```
rphmeier: The previous logic seems correct - the era hasn't moved backwards since the last time `EarliestUnappliedSlash` was set and we have `apply_unapplied_slashes` for moving `EarliestUnappliedSlash` forward.

rphmeier: You could rewrite this block as

```rust
*earliest = Some(match *earliest {
    None => era_now,
    Some(prev) => prev,
})
```

which is the same as the previous logic.

rphmeier:
> Can we get a test? :D

I don't think there is anything to test, because as far as I see there is no bug fixed or change in behavior

thiolliere:
hmm implementing the test got me understanding the code better:

When an offence is received it will apply at `current_era + defer_slash_duration` even if the slash is in session in the very past it will still apply in `era_of_on_offence_call + defer_slash_duration` not the `era_of_slash_session + defer_slash_duration`.

Then I'm not sure I understand why this `EarliestUnappliedSlash` storage is needed. the slashes to apply are always the slashes of `current_era.checked_sub(slash_defer_duration + 1)`

it seems I can write this:
```diff
diff --git a/frame/staking/src/lib.rs b/frame/staking/src/lib.rs
index 7446a6c39..441179876 100644
--- a/frame/staking/src/lib.rs
+++ b/frame/staking/src/lib.rs
@@ -1460,17 +1460,12 @@ impl<T: Trait> Module<T> {
        /// Apply previously-unapplied slashes on the beginning of a new era, after a delay.
        fn apply_unapplied_slashes(current_era: EraIndex) {
                let slash_defer_duration = T::SlashDeferDuration::get();
-               <Self as Store>::EarliestUnappliedSlash::mutate(|earliest| if let Some(ref mut earliest) = earliest {
-                       let keep_from = current_era.saturating_sub(slash_defer_duration);
-                       for era in (*earliest)..keep_from {
-                               let era_slashes = <Self as Store>::UnappliedSlashes::take(&era);
-                               for slash in era_slashes {
-                                       slashing::apply_slash::<T>(slash);
-                               }
+               if let Some(era) = current_era.checked_sub(slash_defer_duration + 1) {
+                       let era_slashes = <Self as Store>::UnappliedSlashes::take(&era);
+                       for slash in era_slashes {
+                               slashing::apply_slash::<T>(slash);
                        }
-
-                       *earliest = (*earliest).max(keep_from)
-               })
+               }
        }
 
        /// Select a new validator set from the assembled stakers and their role preferences.
@@ -1764,10 +1759,6 @@ impl <T: Trait> OnOffenceHandler<T::AccountId, pallet_session::historical::Ident
                        }
                };
 
-               <Self as Store>::EarliestUnappliedSlash::mutate(|earliest| {
-                       *earliest = Some(earliest.unwrap_or(era_now).min(era_now));
-               });
-
                let slash_defer_duration = T::SlashDeferDuration::get();
 
                for (details, slash_fraction) in offenders.iter().zip(slash_fraction) {
```

But ok.

rphmeier:
> But maybe it allows to change dynamically the defer duration

Yup!

PR #4817, opened 2020-02-03
Bump `parking_lot`

DemiMarie-parity:
It’s a major version bump, but doesn’t actually break anything.


pepyakin:
I believe that check-runtime here is moot since this PR cannot change the runtime behavior. Merging.

PR #4819, opened 2020-02-04
Fix flaky messed signature test

bkchr:
When messing with the signature, we need to make sure that we acutally
mess-up the signature. As the generated private/public key is random,
the signature is random as well. It can happen that `bytes[0] ==
bytes[2]` which makes the test fail. We fix this problem by just
inverting the bytes at `0` and `2`.


NikVolf:
it fails one in 2^64? :)

NikVolf:
Ah, ok, sorry. I thought signature parts were `u64`, that's why I decided to not care about those probabilities myself.

bkchr:
I just saw it failing :) I did not think about that this could be the same values while merging the initial pr as well.

Issue #2052, opened 2019-03-20
Refactor fees into separate SRML module to avoid hack


gavofyork:
From srml/balances module:

> Somewhat ugly hack in order to gain access to module's `increase_total_issuance_by`
> using only the Subtrait (which defines only the types that are not dependent
> on Positive/NegativeImbalance). Subtrait must be used otherwise we end up with a
> circular dependency with Trait having some types be dependent on PositiveImbalance<Trait>
> and PositiveImbalance itself depending back on Trait for its Drop impl (and thus
> its type declaration).
>
> This works as long as `increase_total_issuance_by` doesn't use the Imbalance
> types (basically for charging fees).
>
> This should eventually be refactored so that the three type items that do
> depend on the Imbalance type (TransactionPayment, TransferPayment, DustRemoval)
> are placed in their own SRML module.

PR #4820, opened 2020-02-04
Composite accounts

gavofyork:
This brings in the final element of my balance refactorings, namely to avoid the (pessimising) spreading of typically co-accessed account data between multiple trie keys. This is particularly annoying for nonce and balance, which for all normal transactions are always mutated and yet are split between two separate trie items.

This provides an extensible framework, via a trait `StoredMap`, which allows an item to be stored *somewhere*, without specifying whether it's stored in its own storage item or in a shared composite storage item with other (probably related) values. It is used by the system module to allow for arbitrary data to be stored alongside the account nonce.

Concretely, this merges the balances of an account with the nonce, reducing the number of trie accesses for a typical transfer from 3 to 2.

Until we know the exact tradeoff between bulking up a single key with lots of data that may possibly not tend to be co-accessed versus having multiple, smaller, trie keys, then we'll keep it lean, but benchmarking typical use-cases with "fat account items" might lead to further optimisations.

## TODO
- [x] Indices should not use chunking.
- [x] Indices should hold a deposit.
- [x] Reclaiming an index should not be possible before removal.
- [x] Migration code.
- [x] Remove index lookup (in Kusama/Polkadot).
- [x] Test migration on live Kusama state.

## Additionally...

In addition to this, it makes the system of account indices opt-in; this means that accounts need not have indexes despite being funded. Two calls for registering indices are provided, as well as a third call for giving an index up and a forth for administrating indices by root.

The upshot to this is that there no longer needs to be a separate `AccountCreationFee`. All transfers work in the same way.

^^^ CC @jacogr @pepyakin 

Indices were originally brought in to provide for a means of enumerating all accounts (e.g. for migrations). This is no longer needed since we improved the storage interface. Their other usage was for minimising the size of a transaction, at the cost of an additional storage lookup. This additional looking is in reality an extra cost that must be borne out through conditional weighting, increasing code complexity and attack surface. The reality is that the extra bytes fee for putting an normal address is likely smaller than the fee for the extra lookup, meaning it will never be used. Its third usage is in terms of usability, allowing users to pass around smaller addresses. This is slightly dangerous in the previous usage, due to address reclaiming, and needs careful UI provisions. The cost, as it has been implemented, is an ever-growing state or significant fees.

This is mitigated in the current PR by having a single storage entry per index and a deposit, where the deposit can be reclaimed.

## Upgrading Pallets & Runtimes

If you're using `decl_event` in your tests, you'll now need to include `system<T>` as an item, so
`decl_event!(my_pallet<T>);` would become `decl_event!(system<T>, my_pallet<T>);`.

`system::Trait` is now the place to put any `OnNewAccount` or `OnReapAccount` hooks; these will need moving from `balances::Trait`. You should include *at least* the `Balances` handler in the reap account, so:

```
impl frame_system::Trait for Runtime {
    ...
    type OnNewAccount = (...);
    type OnReapAccount = (Balances, ...);
    ...
}
```

Additionally, there's a new item in `system::Trait` called `AccountData` and a new item in `balances::Trait` called `AccountStore`. These can be wired together in order to let system store your account balances, which will be more efficient than balances storing them itself:

```
impl frame_system::Trait for Runtime {
    ...
    type AccountData = pallet_balances::AccountData<Balance>; // Or whatever your balance type is (e.g. u64 in tests)
    ...
}
impl pallet_balances::Trait for Runtime {
    ...
    type AccountStore = System;
    ...
}
```

Several config trait items are gone, they can be removed:

- `pallet_balances::Trait::OnNewAccount` (now in `frame_system`)
- `pallet_balances::Trait::OnReapAccount` (now in `frame_system`)
- `pallet_balances::Trait::OnFreeBalanceZero` (just use `OnReapAccount`)
- `pallet_balances::Trait::CreationFee` (removed)
- `pallet_balances::Trait::TransferFee` (removed)


frame/society/src/mock.rs
```diff
@@ -82,11 +81,9 @@ impl pallet_balances::Trait for Test {
 	type Balance = u64;
 	type OnNewAccount = ();
 	type Event = ();
-	type TransferPayment = ();
 	type DustRemoval = ();
 	type ExistentialDeposit = ExistentialDeposit;
-	type CreationFee = CreationFee;
-	type OnReapAccount = System;
+type OnReapAccount = System;
```
jacogr: whitespace.

jacogr:
Looks good from a functionality perspective to me - did a couple of transfers and messages, the nonce/balances adjust correctly under system.account (as expected).

gavofyork:
Can also confirm this is tested on a recent Kusama state and the migration went through as expected.

frame/indices/src/lib.rs
```diff
@@ -82,36 +74,145 @@ decl_event!(
...
 		<T as Trait>::AccountIndex
 	{
+		/// A previously used account index was re-assigned.
+		IndexAssigned(AccountId, AccountIndex),
 		/// A new account index was assigned.
-		///
-		/// This event is not triggered when an existing index is reassigned
-		/// to another `AccountId`.
-		NewAccountIndex(AccountId, AccountIndex),
+		IndexFreed(AccountIndex),
```
joepetrowski: Docs need updating on this event.

frame/support/src/storage/generator/map.rs
```diff
@@ -123,6 +123,52 @@ impl<K: FullEncode, V: FullCodec, G: StorageMap<K, V>> storage::StorageMap<K, V>
 		ret
 	}
 
+	fn mutate_exists<KeyArg: EncodeLike<K>, R, F: FnOnce(&mut Option<V>) -> R>(key: KeyArg, f: F) -> R {
+		let final_key = Self::storage_map_final_key(key);
+		let mut val = unhashed::get(final_key.as_ref());
+
+		let ret = f(&mut val);
+		match val {
+			Some(ref val) => unhashed::put(final_key.as_ref(), &val.borrow()),
```
thiolliere: I see this borrow is on the above code but this is not needed, val is actually the type we want to store.
```suggestion
			Some(ref val) => unhashed::put(final_key.as_ref(), &val),
```

frame/support/src/storage/generator/map.rs
```diff
@@ -123,6 +123,52 @@ impl<K: FullEncode, V: FullCodec, G: StorageMap<K, V>> storage::StorageMap<K, V>
...
+		key: KeyArg,
+		f: F
+	) -> Result<R, E> {
+		let final_key = Self::storage_map_final_key(key);
+		let mut val = G::from_optional_value_to_query(unhashed::get(final_key.as_ref()));
+
+		let ret = f(&mut val);
+		if ret.is_ok() {
+			match G::from_query_to_optional_value(val) {
+				Some(ref val) => unhashed::put(final_key.as_ref(), &val.borrow()),
```
thiolliere: same if i'm not wrong
```suggestion
				Some(ref val) => unhashed::put(final_key.as_ref(), &val),
```

frame/support/src/storage/generator/map.rs
```diff
@@ -123,6 +123,52 @@ impl<K: FullEncode, V: FullCodec, G: StorageMap<K, V>> storage::StorageMap<K, V>
...
+		key: KeyArg,
+		f: F
+	) -> Result<R, E> {
+		let final_key = Self::storage_map_final_key(key);
+		let mut val = unhashed::get(final_key.as_ref());
+
+		let ret = f(&mut val);
+		if ret.is_ok() {
+			match val {
+				Some(ref val) => unhashed::put(final_key.as_ref(), &val.borrow()),
```
thiolliere:
```suggestion
				Some(ref val) => unhashed::put(final_key.as_ref(), &val),
```

frame/system/src/lib.rs
```diff
@@ -879,12 +898,118 @@ impl<T: Trait> Module<T> {
...
+	pub fn kill_account(who: T::AccountId) {
+		if Account::<T>::contains_key(&who) {
+			Account::<T>::remove(&who);
+			Self::on_killed_account(who);
+		}
+	}
+
+	/// Do anything that needs to be done after an account has been killed.
+	fn on_killed_account(who: T::AccountId) {
+		T::OnReapAccount::on_reap_account(&who);
```
```suggestion
		T::OnKilledAccount::on_reap_account(&who);
```

Probably best to update the name across the board here.

frame/system/src/lib.rs
```diff
@@ -879,12 +898,118 @@ impl<T: Trait> Module<T> {
...
+	}
+	fn mutate<R>(k: &T::AccountId, f: impl FnOnce(&mut T::AccountData) -> R) -> R {
+		let existed = Account::<T>::contains_key(k);
+		let r = Account::<T>::mutate(k, |a| f(&mut a.1));
+		if !existed {
+			Self::on_created_account(k.clone());
+		}
+		r
+	}
+	fn mutate_exists<R>(k: &T::AccountId, f: impl FnOnce(&mut Option<T::AccountData>) -> R) -> R {
```
shawntabrizi: Reminder to do a second review of this

bin/node-template/runtime/src/lib.rs
```diff
@@ -244,12 +249,12 @@ construct_runtime!(
 		NodeBlock = opaque::Block,
 		UncheckedExtrinsic = UncheckedExtrinsic
 	{
-		System: system::{Module, Call, Storage, Config, Event},
+		System: system::{Module, Call, Config, Storage, Event<T>},
 		Timestamp: timestamp::{Module, Call, Storage, Inherent},
 		Aura: aura::{Module, Config<T>, Inherent(Timestamp)},
 		Grandpa: grandpa::{Module, Call, Storage, Config, Event},
-		Indices: indices,
```
shawntabrizi: Let's remove this feature from the macro please.

Will follow up.

shawntabrizi: https://github.com/paritytech/substrate/issues/3001

frame/balances/src/lib.rs
```diff
@@ -376,11 +357,9 @@ decl_storage! {
 		///
 		/// NOTE: THIS MAY NEVER BE IN EXISTENCE AND YET HAVE A `total().is_zero()`. If the total
 		/// is ever zero, then the entry *MUST* be removed.
-		pub Account get(fn account)
-			build(|config: &GenesisConfig<T, I>| config.balances.iter()
-				.map(|&(ref who, free)| (who.clone(), AccountData { free, .. Default::default() }))
-				.collect::<Vec<_>>()
-			): map hasher(blake2_256) T::AccountId => AccountData<T::Balance>;
+		///
+		/// NOTE: This is only used in the case that this module is used to store balances.
```
shawntabrizi: So you could specify Balances as implementing the `AccountStore` trait in the pallet configuration, and account balances would be kept here and act like before?

shawntabrizi: Am I missing something? Balances should implement StoredMap?

frame/balances/src/lib.rs
```diff
@@ -626,87 +606,110 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
-		Locks::<T, I>::remove(who);
-		Account::<T, I>::remove(who);
-		T::OnReapAccount::on_reap_account(who);
-		Self::deposit_event(RawEvent::ReapedAccount(who.clone(), dust));
+	/// NOTE: Doesn't do any preparatory work for creating a new account, so should only be used
+	/// when it is known that the account already exists.
+	///
+	/// NOTE: LOW-LEVEL: This will not attempt to maintain total issuance. It is expected that
+	/// the caller will do this.
+	fn try_mutate_account<R, E>(
```
shawntabrizi: I would want to double check all this logic and the interactions with the `AccountStore` impl in system. It looks right, but is not super easy to follow tbh

frame/balances/src/lib.rs
```diff
@@ -722,6 +725,12 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
 	}
 }
 
+impl<T: Trait<I>, I: Instance> OnReapAccount<T::AccountId> for Module<T, I> {
+	fn on_reap_account(who: &T::AccountId) {
+		Locks::<T, I>::remove(who);
```
```suggestion
		Locks::<T, I>::remove(who);
		Account::<T, I>::remove(who);
```

I think technically we need to remove `Account` too.

frame/balances/src/lib.rs
```diff
@@ -885,9 +894,8 @@ mod imbalances {
 // its type declaration).
 // This works as long as `increase_total_issuance_by` doesn't use the Imbalance
 // types (basically for charging fees).
-// This should eventually be refactored so that the three type items that do
-// depend on the Imbalance type (TransferPayment, DustRemoval)
-// are placed in their own SRML module.
+// This should eventually be refactored so that the type item that
+// depends on the Imbalance type (DustRemoval) is placed in its own SRML module.
```
shawntabrizi: Will we address https://github.com/paritytech/substrate/issues/2052 in the near future?

frame/balances/src/lib.rs
```diff
@@ -1148,59 +1109,74 @@ impl<T: Trait<I>, I: Instance> Currency<T::AccountId> for Module<T, I> where
...
-		};
-		account.free = value;
-
-		// If the balance is too low, then the account is reaped.
-		// Free balance can never be less than ED. If that happens, it gets reduced to zero
-		// and the account information relevant to this subsystem is deleted (i.e. the
-		// account is reaped).
-		let outcome = Self::set_account(who, &account, &old_account);
-		(imbalance, outcome)
+			ensure!(value >= ed || !account.free.is_zero(), ());
```
shawntabrizi: This logic is not right anymore I think.

We should check that value + existing reserved value >= ed.

Also, why does this fail now if `account.free.is_zero()`. We can't call this function on account with some free balance already?

frame/balances/src/lib.rs
```diff
@@ -1148,59 +1109,74 @@ impl<T: Trait<I>, I: Instance> Currency<T::AccountId> for Module<T, I> where
...
+			-> Result<Self::NegativeImbalance, DispatchError>
+		{
+			let new_free_account = account.free.checked_sub(&value)
+				.ok_or(Error::<T, I>::InsufficientBalance)?;
 
-		// bail if not yet created and this operation wouldn't be enough to create it.
-		if value < ed && account.total().is_zero() { return Self::PositiveImbalance::zero() }
+			// bail if we need to keep the account alive and this would kill it.
+			let ed = T::ExistentialDeposit::get();
+			let would_kill = new_free_account < ed && account.free >= ed;
```
shawntabrizi: We need to look at total balance for all the ED checks.

frame/balances/src/lib.rs
```diff
@@ -1270,47 +1238,41 @@ impl<T: Trait<I>, I: Instance> ReservableCurrency<T::AccountId> for Module<T, I>
...
 	fn repatriate_reserved(
 		slashed: &T::AccountId,
 		beneficiary: &T::AccountId,
 		value: Self::Balance,
-	) -> result::Result<Self::Balance, DispatchError> {
+		status: Status,
+	) -> Result<Self::Balance, DispatchError> {
 		if value.is_zero() { return Ok (Zero::zero()) }
 
 		if slashed == beneficiary {
```
shawntabrizi: This needs to take into account `status`

Cannot make a reasonable multiline github suggestion but:

`if status == Status::Free` then do the logic below

`if status == Status::Reserved` then do nothing

frame/balances/src/lib.rs
shawntabrizi: It seems StorageAccountShim does the magic here per the "local test". If this is right, you can resolve this conversation.

frame/elections-phragmen/src/lib.rs
```diff
@@ -1424,7 +1423,7 @@ mod tests {
 
 			assert_ok!(Elections::report_defunct_voter(Origin::signed(5), 3));
 			assert_eq!(
-				System::events()[1].event,
+				System::events()[7].event,
```
shawntabrizi: There is a better way to match these events I think...

frame/support/procedural/src/construct_runtime/mod.rs
```diff
@@ -299,14 +325,10 @@ fn decl_outer_event_or_origin<'a>(
...
-		DeclOuterKind::Event => quote!(#scrate::impl_outer_event!),
-		DeclOuterKind::Origin => quote!(#scrate::impl_outer_origin!),
-	};
-	let enum_name = Ident::new(kind_str.as_str(), Span::call_site());
+
 	Ok(quote!(
-		#macro_call {
-			pub enum #enum_name for #runtime_name where system = #system_name {
+		#scrate::impl_outer_event! {
+			pub enum Event for #runtime_name {
```
shawntabrizi: Why is `Event` now a hard-coded name?

frame/balances/src/lib.rs
shawntabrizi: Would be good to search for all usage of ED, and check it is using total balance for comparison

frame/balances/src/lib.rs
gavofyork: it's a non-trivial refactor and what we have works, so probably not.

frame/balances/src/lib.rs
gavofyork: done.

frame/support/procedural/src/construct_runtime/mod.rs
gavofyork: i think this was @bkchr 's change...

gavofyork:
logic errors fixed.

frame/support/procedural/src/construct_runtime/mod.rs
bkchr: It was always named `Event` in `construct_runtime!`. Before this change, the method could be used either for `Origin` or for `Event`. Now these are split and I hard coded the name.

frame/indices/src/lib.rs
```diff
@@ -82,36 +74,145 @@ decl_event!(
...
+		///
+		/// Emits `IndexAssigned` if successful.
+		///
+		/// # <weight>
+		/// - `O(1)`.
+		/// - One storage mutation (codec `O(1)`).
+		/// - One reserve operation.
+		/// - One event.
+		/// # </weight>
+		fn claim(origin, index: T::AccountIndex) {
```
shawntabrizi: This function allows a single account to claim multiple indices. Is that intended?

frame/indices/src/lib.rs
```diff
@@ -82,36 +74,145 @@ decl_event!(
...
+		/// - `O(1)`.
+		/// - One storage mutation (codec `O(1)`).
+		/// - One transfer operation.
+		/// - One event.
+		/// # </weight>
+		fn transfer(origin, new: T::AccountId, index: T::AccountIndex) {
+			let who = ensure_signed(origin)?;
+			ensure!(who != new, Error::<T>::NotTransfer);
+
+			Accounts::<T>::try_mutate(index, |maybe_value| {
```
shawntabrizi: How does `try_mutate` work in terms of the lower storage layers?

For example, your first line in here is:

```
let (account, amount) = maybe_value.take().ok_or(Error::<T>::NotAssigned)?;
```

This is one storage read, and if this fails, we need not do any more logic. If `try_mutate` also invokes a storage read before any of this logic is run I would say that this method of writing the function possibly invokes an additional storage read in situations when it would not be necessary.

frame/indices/src/lib.rs
```diff
@@ -82,36 +74,145 @@ decl_event!(
...
+				let (account, amount) = maybe_value.take().ok_or(Error::<T>::NotAssigned)?;
+				ensure!(&account == &who, Error::<T>::NotOwner);
+				*maybe_value = Some((new.clone(), amount));
```
shawntabrizi: Repatriate reserved will return any amount that was not able to be repatriated.

We should use that value to put the correct amount of reserved funds this module manages versus copying the amount straight over and swallowing the result of repatriate reserved

frame/system/src/lib.rs
```diff
@@ -435,29 +292,43 @@ type EventIndex = u32;
 
 decl_storage! {
 	trait Store for Module<T: Trait> as System {
-		/// Extrinsics nonce for accounts.
-		pub AccountNonce get(fn account_nonce): map hasher(blake2_256) T::AccountId => T::Index;
+		/// The full account information for a particular account ID.
+		// TODO: should be hasher(twox64_concat) - will need staged migration
+		// TODO: should not including T::Index (the nonce)
```
```suggestion
		// TODO: should not including T::Index (the nonce)
		// https://github.com/paritytech/substrate/issues/4917
```

frame/system/src/lib.rs
```diff
@@ -435,29 +292,43 @@ type EventIndex = u32;
...
 		ExtrinsicCount: Option<u32>;
+
 		/// Total weight for all extrinsics put together, for the current block.
 		AllExtrinsicsWeight: Option<Weight>;
+
 		/// Total length (in bytes) for all extrinsics put together, for the current block.
 		AllExtrinsicsLen: Option<u32>;
+
 		/// Map of block numbers to block hashes.
+		// TODO: should be hasher(twox64_concat) - will need one-off migration
```
```suggestion
		// TODO: should be hasher(twox64_concat) - will need one-off migration
		// https://github.com/paritytech/substrate/issues/4917
```

frame/system/src/lib.rs
```diff
@@ -499,6 +370,149 @@ decl_storage! {
...
+		FailedToExtractRuntimeVersion,
+	}
+}
+
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		type Error = Error<T>;
+
+		/// A big dispatch that will disallow any other transaction to be included.
+		// TODO: this must be preferable available for testing really (not possible at the moment).
```
shawntabrizi: What does this TODO mean?

frame/system/src/lib.rs
shawntabrizi: https://github.com/paritytech/substrate/issues/4917

frame/system/src/lib.rs
shawntabrizi: https://github.com/paritytech/substrate/issues/4917

frame/indices/src/lib.rs
gavofyork: yes.

frame/system/src/lib.rs
gavofyork: i don't really know for sure, but i would read it as "This should *only* be available for testing, rather than in general usage, but that's not possible (since it's within the decl_module macro)"

frame/staking/src/lib.rs
```diff
@@ -761,9 +760,6 @@ decl_storage! {
 
 		/// The earliest era for which we have a pending, unapplied slash.
 		EarliestUnappliedSlash: Option<EraIndex>;
-
-		/// The version of storage for upgrade.
-		StorageVersion: u32;
```
thiolliere: I suspect the storage version hasn't been removed from storage in kusama

thiolliere: But if I query on kusama the key `twox128("Staking") ++ twox128("StorageVersion") == 0x5f3e4907f716ac89b6347d15ececedca308ce9615de0775a82f8a94dc3d285a1` then I get no value, and I can't understand why yet.

PR #4821, opened 2020-02-04
pallet-evm: log created address

sorpaas:
cc @danforbes

Currently it's not trivial to fetch the created address after contract creation. This PR adds `Event::Created(H160)` so that the address is apparent.

### DIFF
--- a/frame/evm/src/lib.rs
+++ b/frame/evm/src/lib.rs
@@ -174,6 +174,8 @@ decl_event! {
 	pub enum Event {
 		/// Ethereum events from contracts.
 		Log(Log),
+		/// A contract has been created at given address.
+		Created(H160),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


frame/evm/src/lib.rs
```diff
@@ -361,6 +364,7 @@ decl_module! {
 
 			let (values, logs) = executor.deconstruct();
 			backend.apply(values, logs, true);
+			Module::<T>::deposit_event(Event::Created(create_address));
```
bkchr: When `res` is an error, didn't the creation failed and we should not send the event in that case?

danforbes:
[Works like a charm](https://github.com/danforbes/substrate-evm/commit/00b4983015894f89cda266b17a9e010223b45057#diff-5e0cdf65ccf1fba231d416508e7ba9bbR34)...thanks so much!

PR #4822, opened 2020-02-04
Subsystems memory tracking: 1. Transaction pool

NikVolf:
There will be extra some bounds on runtime primitives, as proposed in #4806 

If some project uses exotic types for `OpaqueExtrinsic`/`Hash`, they might need to implement `MallocSizeOf` for it also


client/cli/src/informant.rs
```diff
@@ -40,6 +41,11 @@ pub fn build(service: &impl AbstractService) -> impl futures::Future<Output = ()
 			} else {
 				trace!(target: "usage", "Usage statistics not displayed as backend does not provide it")
 			}
+			trace!(
+				target: "usage",
+				"Subsystems memory [txpool: {}]",
```
tomusdrw: What unit does it print the value in? I'd also simplify it to:
`[txpool] X Kbytes` or sth` as the `usage` target will give us this kind of details already.

client/transaction-pool/graph/src/future.rs
```diff
@@ -243,3 +244,31 @@ impl<Hash: hash::Hash + Eq + Clone, Ex> FutureTransactions<Hash, Ex> {
 		self.waiting.values().fold(0, |acc, tx| acc + tx.transaction.bytes)
 	}
 }
+
+#[cfg(test)]
+mod tests {
+
```
tomusdrw:
```suggestion
```

client/transaction-pool/graph/src/ready.rs
```diff
@@ -44,6 +44,15 @@ pub struct TransactionRef<Hash, Ex> {
 	pub insertion_id: u64,
 }
 
+impl<Hash, Ex> parity_util_mem::MallocSizeOf for TransactionRef<Hash, Ex>
```
tomusdrw: Derive doesn't work due to extra bounds on `Hash` and `Ex`, right?

client/transaction-pool/graph/src/ready.rs
```diff
@@ -74,7 +83,7 @@ impl<Hash, Ex> PartialEq for TransactionRef<Hash, Ex> {
 }
 impl<Hash, Ex> Eq for TransactionRef<Hash, Ex> {}
 
-#[derive(Debug)]
+#[derive(Debug, parity_util_mem::MallocSizeOf)]
```
tomusdrw: I'd really just import it via `use parity_util_mem::MallocSizeOf` cause it's used quite frequently in this file.

client/transaction-pool/graph/src/ready.rs
```diff
@@ -116,6 +125,19 @@ pub struct ReadyTransactions<Hash: hash::Hash + Eq, Ex> {
 	best: BTreeSet<TransactionRef<Hash, Ex>>,
 }
 
+impl<Hash: hash::Hash + Eq, Ex> parity_util_mem::MallocSizeOf for ReadyTransactions<Hash, Ex>
```
tomusdrw: Would it be possible to fix the derive to allow custom bounds (like in serde?) Cause it's really superfluous and can get out of sync when someone adds a field to the original struct easily.

frame/system/src/lib.rs
```diff
@@ -171,12 +171,12 @@ pub trait Trait: 'static + Eq + Clone {
...
 	type BlockNumber:
 		Parameter + Member + MaybeSerializeDeserialize + Debug + MaybeDisplay + SimpleArithmetic
-		+ Default + Bounded + Copy + sp_std::hash::Hash + sp_std::str::FromStr;
+		+ Default + Bounded + Copy + sp_std::hash::Hash + sp_std::str::FromStr + MaybeMallocSizeOf;
 
 	/// The output of the `Hashing` function.
 	type Hash:
 		Parameter + Member + MaybeSerializeDeserialize + Debug + MaybeDisplay + SimpleBitOps + Ord
-		+ Default + Copy + CheckEqual + sp_std::hash::Hash + AsRef<[u8]> + AsMut<[u8]>;
+		+ Default + Copy + CheckEqual + sp_std::hash::Hash + AsRef<[u8]> + AsMut<[u8]> + MaybeMallocSizeOf;
```
tomusdrw: Hmm, that's a pretty strong requirement to put into the runtime imho, requires bringing in a separate (parity-specific) lib to implement properly. Can't we add `where` clauses instead?

NikVolf: Unless some exotic Hash type is used, it is implemented by default within `parity-util-mem` (all H*, integers, fixed arrays, `Vec<u8>` etc. are covered)

client/cli/src/informant.rs
NikVolf: For technical traces invisible to user, I'd always put simple bytes

tomusdrw: Let's add `bytes` then, although I feel it would be more readable as `KB`

client/transaction-pool/graph/src/ready.rs
NikVolf: It does work, actually

client/transaction-pool/graph/src/ready.rs
NikVolf: Actually, I was able to remove this custom impl almost everywhere

But custom bound would be useful later, yes, we'll add it in `parity-common`

client/transaction-pool/graph/src/ready.rs
NikVolf: no longer the case

frame/system/src/lib.rs
NikVolf: And we have `MaybeSerializeDeserialize` there,, so using exotic Hash will require some work anyway.

client/cli/src/informant.rs
```diff
@@ -40,6 +41,11 @@ pub fn build(service: &impl AbstractService) -> impl futures::Future<Output = ()
 			} else {
 				trace!(target: "usage", "Usage statistics not displayed as backend does not provide it")
 			}
+			trace!(
```
bkchr: I think this should go to grafana to have nice visualizations.

bkchr: A follow up could be that we integrate some sort of service that sends these memory usage infos to grafana. Each subsystem that can track its own memory usage could report this to this service.

bkchr: (By service I mean something that runs inside of Substrate)

primitives/runtime/src/generic/unchecked_extrinsic.rs
```diff
@@ -44,6 +44,18 @@ where
...
 }
 
+#[cfg(feature = "std")]
+impl<Address, Call, Signature, Extra> parity_util_mem::MallocSizeOf
+	for UncheckedExtrinsic<Address, Call, Signature, Extra>
+where
+	Extra: SignedExtension
+{
+	fn size_of(&self, _ops: &mut parity_util_mem::MallocSizeOfOps) -> usize {
+		// Exists only in runtime (TODO: is it?)
```
bkchr:
```suggestion
```

And yeah, it should only be used in the runtime :)

primitives/runtime/src/lib.rs
```diff
@@ -639,6 +639,14 @@ macro_rules! assert_eq_error_rate {
 #[derive(PartialEq, Eq, Clone, Default, Encode, Decode)]
 pub struct OpaqueExtrinsic(pub Vec<u8>);
 
+#[cfg(feature = "std")]
+impl parity_util_mem::MallocSizeOf for OpaqueExtrinsic
```
bkchr:
```suggestion
impl parity_util_mem::MallocSizeOf for OpaqueExtrinsic {
```

primitives/runtime/src/lib.rs
```diff
@@ -639,6 +639,14 @@ macro_rules! assert_eq_error_rate {
...
+{
```
bkchr:
```suggestion
```

primitives/runtime/src/testing.rs
```diff
@@ -161,6 +161,7 @@ pub struct Header {
 	/// Merkle root of block's extrinsics
 	pub extrinsics_root: H256,
 	/// Digest items
+	#[ignore_malloc_size_of = "TODO"]
```
bkchr: What will this `TODO` do?

primitives/runtime/src/testing.rs
```diff
@@ -300,6 +303,15 @@ impl<'a, Xt> Deserialize<'a> for Block<Xt> where Block<Xt>: Decode {
 #[derive(PartialEq, Eq, Clone, Encode, Decode)]
 pub struct TestXt<Call, Extra>(pub Option<(u64, Extra)>, pub Call);
 
+#[cfg(feature = "std")]
+impl<Call, Extra> parity_util_mem::MallocSizeOf for TestXt<Call, Extra>
+{
+	fn size_of(&self, _ops: &mut parity_util_mem::MallocSizeOfOps) -> usize {
+		// Non-opaque extrinsics always 0 (TODO: check if needed)
```
bkchr:
```suggestion
		// Only used in tests, size doesn't matter
```

Or something similar.

primitives/runtime/src/testing.rs
```diff
@@ -300,6 +303,15 @@ impl<'a, Xt> Deserialize<'a> for Block<Xt> where Block<Xt>: Decode {
 #[derive(PartialEq, Eq, Clone, Encode, Decode)]
 pub struct TestXt<Call, Extra>(pub Option<(u64, Extra)>, pub Call);
 
+#[cfg(feature = "std")]
+impl<Call, Extra> parity_util_mem::MallocSizeOf for TestXt<Call, Extra>
```
bkchr:
```suggestion
impl<Call, Extra> parity_util_mem::MallocSizeOf for TestXt<Call, Extra> {
```

primitives/runtime/src/testing.rs
```diff
@@ -300,6 +303,15 @@ impl<'a, Xt> Deserialize<'a> for Block<Xt> where Block<Xt>: Decode {
...
+{
```
bkchr:
```suggestion
```

primitives/runtime/src/testing.rs
```diff
@@ -220,10 +221,12 @@ impl<'a> Deserialize<'a> for Header {
 }
 
 /// An opaque extrinsic wrapper type.
-#[derive(PartialEq, Eq, Clone, Debug, Encode, Decode)]
+#[derive(PartialEq, Eq, Clone, Debug, Encode, Decode, parity_util_mem::MallocSizeOf)]
 pub struct ExtrinsicWrapper<Xt>(Xt);
 
-impl<Xt> traits::Extrinsic for ExtrinsicWrapper<Xt> {
+impl<Xt> traits::Extrinsic for ExtrinsicWrapper<Xt>
+where Xt: parity_util_mem::MallocSizeOf
```
bkchr:
```suggestion
	where Xt: parity_util_mem::MallocSizeOf
```

test-utils/runtime/src/lib.rs
```diff
@@ -114,6 +114,8 @@ pub enum Extrinsic {
 	ChangesTrieConfigUpdate(Option<ChangesTrieConfiguration>),
 }
 
+parity_util_mem::malloc_size_of_is_0!(Extrinsic); // non-opaque extrinisic does not need this
```
bkchr: Why don't we use this macro for the other ones where we return `0`?

NikVolf: Indeed, I thought it does not support generics, but it now does, thanks!

primitives/runtime/src/testing.rs
NikVolf: no longer required

test-utils/runtime/src/lib.rs
NikVolf: Looks like it is not properly supported though
And it also adds bounds on generic arguments which are not required in this case, so I won't use `malloc_size_of_is_0!` for now, but I'll add another macro later.

frame/system/src/lib.rs
NikVolf: @bkchr what's your view on this?
@tomusdrw are you ok with an argument?

test-utils/runtime/src/lib.rs
NikVolf: `malloc_size_of_is_0!` is upgraded, and one case covered

frame/system/src/lib.rs
bkchr: For me this is okay.

client/cli/src/informant.rs
bkchr: Grafana should be done in this pr, IMHO.

There should be some macro to call. @expenses

NikVolf: `record_metrics!`?

pretty high level stuff is only written there though

NikVolf: Done

bkchr: Ahh okay, I thought that you can call this macro from everywhere, like the telemetry. But seems to be okay for now.

Issue #1252, opened 2018-12-11
Compile the wasm runtime with a stable toolchain


pepyakin:
We require nightly compiler for building the wasm runtime. The main reason for that is several important features was available only on the nightly compilers.

However, as the time progresses more and more features are promoted to stable.

- [x] alloc crate ([tracking issue](https://github.com/rust-lang/rust/issues/27783), [RFC](https://github.com/rust-lang/rfcs/pull/2480), [PR](https://github.com/rust-lang/rust/pull/59675))
- [x] `std::arch::wasm32::unreachable` ([tracking issue](https://github.com/rust-lang/rust/issues/61119))
- [x] `core_intrinsics` ([trie/hash-db](https://github.com/paritytech/trie/blob/283bf5ccf2d251bf2aa9cb51b42d3e728cc88a8f/hash-db/src/lib.rs#L17), not sure why)
- [x] waiting on `alloc_error_handler` becoming panic by default ([tracking issue](https://github.com/rust-lang/rust/issues/66741), [formelly this](https://github.com/rust-lang/rust/issues/51540))

pepyakin:
The alloc crate is on the course of stabilization: [PR](https://github.com/rust-lang/rust/pull/59675)

thiolliere:
alloc crate is stabilized by the way

tomaka:
The biggest offenders for this are actually not in Substrate, but dependencies such as `elastic-array` or `hash-db`.
There seems to be way more than just `std::arch::wasm32::unreachable` missing.

pepyakin:
- One dependency that still uses features is `safe_mix`. It does so needlessly and actually is already fixed on master (not published yet though).
- `core::arch::wasm32::unreachable` was landed and it allowed us to get rid of `core_intrinsics`, see #4823.
- `alloc_handler_error` is a real problem. We cannot work it around since `liballoc` requires its presence if used. There are some activity in discussion with some unhappy splats coming out of it, so 🤷‍♂

athei:
What about dropping `#[no_std]`?  `wasm32-unknown-unknown` has a std target. It sounds weird that an `unknown`  has  std support but memory allocation is standardized in wasm so it can support it even without an explicit operating system. Any file system operation won't work of course (i.e panic). But that will make the runtime compile with stable.

Given that using nightly creates a a lot of friction and support overhead that might worth a shot.

xlc:
[...]

PR #4823, opened 2020-02-04
Get rid of in-substrate usages of `core_intrinsics` feature

pepyakin:
Removes all instances of usage `core_intrinsics`. In two cases they were really useless. In the third case, we could substitute it with a call to `core::arch::wasm32::unreachable` which is stable.

One little step towards https://github.com/paritytech/substrate/issues/1252

The behavior hasn't changed, thus no bumping of runtime is required.

### DIFF
--- a/primitives/application-crypto/Cargo.toml
+++ b/primitives/application-crypto/Cargo.toml
@@ -21,5 +21,9 @@ std = [ "full_crypto", "sp-core/std", "codec/std", "serde", "sp-std/std", "sp-io
 # or Intel SGX.
 # For the regular wasm runtime builds this should not be used.
 full_crypto = [
-    "sp-core/full_crypto"
+    "sp-core/full_crypto",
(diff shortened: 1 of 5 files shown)


pepyakin:
The CI failure implies that the nightly used is rather old.

bkchr:
But why does it work for the main build where we build the wasm blobs? Hmm.

pepyakin:
My hypothesis is that the `gitlab-test-full-crypto-feature` job uses an outdated nightly. I am not sure though, @gabreal might be able to confirm this.

pepyakin:
Ok, failure is in fact legit: `core::arch::wasm32` is only available for `wasm32` and for some reason we compile `sr-io` for `no_std` but not `wasm32` (probably, SGX?).

I think I am inclined to disable panic_handler and alloc_error_handler since they are disabled for those use-cases anyway.

PR #4827, opened 2020-02-04
Manage child trie content independently

cheme:
This PR is a refactoring of child trie transaction.

Before this PR as single transaction of encoded trie node was produced by state-machine, packing the node from the top trie and the child trie without distinction.
As a result, prefixing rocksdb key by a unique id for child trie was done in state machine crate.

This PR move those key prefixing to client/db/src/lib.rs and split all transaction content between child trie.

We move from a single hashmap of key value to a btreemap of hashmaps of key values, but there will be a less data redundancy.

### Main changes:

- all code prepending unique id for child trie is now in client/db/src/lib.rs, there is no longer KeySpacedDB plugged in an artificial way in state machine.
- child info has been rework by removing 'OwnedChildInfo', at some point I did implement a borrow implementation (this commit removing it is best to check impl : https://github.com/cheme/substrate/commit/0d45d8559a46d297311574c1304245e371ea7ee4 ) but it was probably bad design.
- top trie can be seen as a child trie with empty unique id. To keep thing sane this change do not touch directly state-machine where we keep a top_storage and dedicated/duplicated api: that way things do not change much except that child api can now redirect to parent api when it is safe (currently no runtime can call a empty unique id so this is only changing rpc behavior in a way that does not seem problematic). (see ext.rs).
- A hasher trait with constant is added, this should be moved to hashdb crate (then it will cover all memorydb instance, not only this specific state machine case), but seems fine here for the time (branch with hashdb possible change : https://github.com/cheme/trie/tree/const_empty ).

<del>
- change of format for journal of block change and pruning  cc @arkpar :
  - the new format just use a different prefix, so if any change happen we would just change this prefix again. Alternate design would be to include a versioning information in the encoded journal.
  - the old format is still in use, this costs a double query to fetch old journal. I do not think this is very problematic in the journal case, an alternate way of doing thing would be to migrate all existing journals. I am not sure if it is useful at this point (would need rencode/rewrite of everything to avoid a double query), maybe it is fast, though it should need some testing could be done later.
</del>

- change trie is not using this split: the code do not need key isolation as it is prefixed to avoid any key collision conflict. So in this part of code everything run in top trie.
- proof is still running without child trie handle, everything is in a top trie, if at some point we got different child trie type the split will be needed. It should be possible to split proof, this would be useful for proof compaction (at this point compacting proof of a `call` will require to isolate each trie payload by key prefix, that is far from ideal). I think this should switch, but it will break light client proof and cumulus witness, so I keep this question out of the scope of this PR. CC @bkchr @jimpo

### Minor changes

- full storage root can return child_infos for technical reason.

Even if this PR mainly touches internals, it requires changes to polkadot (https://github.com/cheme/polkadot-1/commit/b94eb4696aea3ef0b33924220f605fe7d42f7b5b , pr will be created later).


primitives/state-machine/src/ext.rs
```diff
@@ -520,20 +548,10 @@ where
...
+					self.backend.child_storage_root(storage_key, child_info, delta)
 				};
 
 				let root = root.encode();
-				// We store update in the overlay in order to be able to use 'self.storage_transaction'
-				// cache. This is brittle as it rely on Ext only querying the trie backend for
-				// storage root.
-				// A better design would be to manage 'child_storage_transaction' in a
-				// similar way as 'storage_transaction' but for each child trie.
-				if is_empty {
```
cheme: Note, this part of code is replaced by a caching in transaction_child_storage_root in storage_transaction_cache, this is more correct than the previous code that did rely on assuming root is always queried on the backend trie when accessing child content.

client/db/src/lib.rs
```diff
@@ -672,8 +681,16 @@ impl<Block: BlockT> sc_state_db::NodeDb for StorageDb<Block> {
 	type Error = io::Error;
 	type Key = [u8];
 
-	fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, Self::Error> {
-		self.db.get(columns::STATE, key).map(|r| r.map(|v| v.to_vec()))
+	fn get(&self, child_info: &ChildInfo, key: &[u8]) -> Result<Option<Vec<u8>>, Self::Error> {
+		if child_info.is_top_trie() {
```
bkchr: Okay, this is getting weird IMHO.

We pass a `ChildInfo` that says if we should access the top trie?

Some out of the context question, is top trie the wording we use? I still find it confusing to speak from top/child trie.

client/db/src/lib.rs
```diff
@@ -672,8 +681,16 @@ impl<Block: BlockT> sc_state_db::NodeDb for StorageDb<Block> {
...
+			self.db.get(columns::STATE, key)
+		} else {
+			let keyspace = child_info.keyspace();
+			let mut key_buffer = vec![0; keyspace.len() + key.len()];
```
bkchr: Can we not reuse this `key_buffer`?

client/db/src/lib.rs
```diff
@@ -1102,26 +1124,29 @@ impl<Block: BlockT> Backend<Block> {
 			}
 
 			let finalized = if operation.commit_state {
-				let mut changeset: sc_state_db::ChangeSet<Vec<u8>> = sc_state_db::ChangeSet::default();
+				let mut changesets = ChildrenMap::<sc_state_db::ChangeSet<Vec<u8>>>::default();
 				let mut ops: u64 = 0;
 				let mut bytes: u64 = 0;
```
bkchr: I think we should do: 
```suggestion
				let mut bytes = 0;
```

And just do the `bytes as u64` once when we call this `tally_writes`.

client/db/src/lib.rs
bkchr: Can we not pass some `enum` as `info`?

client/db/src/lib.rs
```diff
@@ -1307,11 +1332,32 @@ impl<Block: BlockT> Backend<Block> {
...
 
 fn apply_state_commit(transaction: &mut DBTransaction, commit: sc_state_db::CommitSet<Vec<u8>>) {
-	for (key, val) in commit.data.inserted.into_iter() {
-		transaction.put(columns::STATE, &key[..], &val);
-	}
-	for key in commit.data.deleted.into_iter() {
-		transaction.delete(columns::STATE, &key[..]);
+	let mut key_buffer = Vec::new();
+	for child_data in commit.data.into_iter() {
+		if child_data.0.is_top_trie() {
```
bkchr: Again.

client/db/src/lib.rs
```diff
@@ -1307,11 +1332,32 @@ impl<Block: BlockT> Backend<Block> {
...
+			for (key, val) in child_data.1.inserted.into_iter() {
+				transaction.put(columns::STATE, &key[..], &val);
+			}
+			for key in child_data.1.deleted.into_iter() {
+				transaction.delete(columns::STATE, &key[..]);
+			}
+		} else {
+			let keyspace = child_data.0.keyspace();
+			let keyspace_len = keyspace.len();
+			key_buffer.resize(keyspace_len, 0);
```
bkchr: There should exists a function for this. Better would be a type `ChildKey` or whatever.

```
impl ChildKey {
    fn new(keyspace, key)
    fn update_key()
    fn update_space_and_key()
}
```

Something in this direction.

client/src/cht.rs
```diff
@@ -23,11 +23,10 @@
 //! root has. A correct proof implies that the claimed block is identical to the one
 //! we discarded.
 
-use hash_db;
 use codec::Encode;
 use sp_trie;
 
-use sp_core::{H256, convert_hash};
+use sp_core::{H256, convert_hash, self};
```
bkchr:
```suggestion
use sp_core::{H256, convert_hash};
```

We use using Rust 2018!

client/db/src/lib.rs
cheme: About top trie wording, this is rather used in state-machine, but it may not be the best wording, maybe `root_trie` will be better?

ChildInfo is an enum, I just did not want to create a special one for the top trie, the reason for that is that technically the top trie is the child trie without a unique id, it does not have a specific behavior regarding the storage. So this check is a small optimization case, running it with the other condition works.

Giving it more thought, top trie mainly differs by the fact that it does not attach to a parent trie: in state-machine I do some redirection that repercute child calls to parent if it is top_trie, those are bad as it ignores the fact that we give a storagekey where the root will not be attached.
I will switch them to fail unless the parent key is empty.

client/db/src/lib.rs
cheme: I tried to a bit, but this means inner mutability, do you think it is worth it?
(I also did try to switch the primitive to get(&mut self... but this was a bit to impacting, I could try a bit more).

client/db/src/lib.rs
bkchr: Yeah, I know that the difference is not that big. But I still think that having everywhere the distinction and than we use `ChildInfo`. Maybe we could just rename this to `TrieInfo`?

Regarding `root_trie`, IMHO that sounds better.

cheme: Yes `Child` is not very happy naming, `Trie` is not either because it can potentially be a binary tree or any kind of structure that maintain a state with data.
So I could propose a renaming to 'StateInfo', and replace 'root_trie' by 'root_state'.
I really like it, but it propagates to the whole code base rather quickly (removing every 'child' to make 'child' function the default case function and default case function the 'root' specific function), that can be a pretty big move in term of documentation.

arkpar:
I don't see why `state-db` should be concerned with any of the trie stuff, including child tries. It's a totally different abstraction, which is a journaled and versioned storage for state backend. It does not care how state is mapped to key-value nodes. The only thing that it requires is that the same key is used only inserted once per state. The key prefixing was introduced to support that requirement in the first place. 
Would it be possible to rework this so that all the prefixing is done before inserting to `state-db`?

cheme:
> I don't see why state-db should be concerned with any of the trie stuff, including child tries

I agree, `ChildTrie` is a rather bad name in this case, maybe it should be renamed to `State` or `StateContent` or something that indicates it is not necessarily a Trie, and Child is not very usefull if you consider the top root as a child trie that does attach its short state to block header instead of another state.

> is a journaled and versioned storage for state backend.

After this PR this is the same but for multiple state backends.

> It does not care how state is mapped to key-value nodes.

This PR breaks this definition by having multiple states.
For child trie I still want to have system managed unique identifier (from an incremental id), which requires using a different state column.
This PR does allow that easily, otherwise this would need multiple instance of `state-db` which does not sounds right to me because each instance will share the same chain management logic, so it would be easier to have a same `state-db` managing two different states.
(or migrate every state key value to use prefixing, but then it seems sad that we cannot dedicate an unprefixed column to a big state).
Furthermore the question of key isolation between child tries is way simpler if you say that the keyvaluedb got the ability to manage subcollections (on rocksdb by prefixing the keys in a dedicated column), then the two places where the prefixing is done in the pr could move to rocksdb specific implementation later, I like this design.
So in those case considering one state-db per child trie would be a way to do that,  I am not sure if multiplicating the number of journal instances makes sense. But yes it is not strictly required as long as we only put things in the same state column.

> The only thing that it requires is that the same key is used only inserted once per state. The key prefixing was introduced to support that requirement in the first place.

Is it? I was thinking reference counting was handling that, and the prefixing was only to avoid collision at rocksdb level, so prefixing only before putting content in rocksdb would have save a little memory at the cost of splitting the storage.

> Would it be possible to rework this so that all the prefixing is done before inserting to state-db?

Currently the prefixing was reduced to two places in client db:
- StorageDb `get`: this stays true, but we need to do it before accessing state-db too (should be rather straight forward in `Storage` implementation for `StorageDB`).
- `apply_state_commit`: this would not apply any more, and the prefixing will be in `try_commit_operation` befor calling state-db `insert-block` function (that is where there could be usage of others state_db instance).

So yes, even if it is not really the direction I was thinking of, it is fairly doable. It will revert to bigger journals, but finally it may be a good point to make this pr touch less crates.
I will do that tomorrow.

arkpar:
> This PR breaks this definition by having multiple states.
> For child trie I still want to have system managed unique identifier (from an incremental id), which requires using a different state column.

If you are talking about using an additional column family for ALL child trees, I don't see much point in that. There's no clear benefit as you still can't manage tries individually. As for using a column family per tree - rocksdb is not really designed for dynamic addition/removal of column families.

Could you explain a bit how system managed unique id would work?

> Furthermore the question of key isolation between child tries is way simpler if you say that the keyvaluedb got the ability to manage subcollections (on rocksdb by prefixing the keys in a dedicated column), then the two places where the prefixing is done in the pr could move to rocksdb specific implementation later, I like this design.

I doubt we'll have any support for separate collections in the storage backend. It was always designed to be a simple key-value store. The partitioning (prefixing) is done by a higher level code, which gives it more flexibility. So I'd suggest leaving a single instance of `state_db`, which is just a key value store. The higher level would manage ids and prefixes. IMO IDs should come naturally from the higher level code. In fact they should be passed down from the runtime. 

> > The only thing that it requires is that the same key is used only inserted once per state. The key prefixing was introduced to support that requirement in the first place.
> 
> Is it? I was thinking reference counting was handling that, and the prefixing was only to avoid collision at rocksdb level, so prefixing only before putting content in rocksdb would have save a little memory at the cost of splitting the storage.
> 

Reference counting is used to track the same keys introduced in different forks, but no in the same state or its ancestors. To track references through the entire history we'd need to have persistent counters in the DB and that would kill performance.

cheme:
I did remove changes to `state-db` crate, this means that we stay with the fatest old journal format, but we do not have to worry about the change (even if it was already tested on kusama).
I was a bit reluctant to do this change, but finally it works quite well :+1:.
The problematic of using subcollection at key value db level could be tackled when or if it becomes needed.

cheme:
> Could you explain a bit how system managed unique id would work?

It uses unique identifier for each child trie, those are just sequential values.
The prefixing is done over a encodable/decodable serialization of this identifier, which means we do not have key conflict (but the scheme needs to apply to all content to be collision free, so it is not compatible with the current state column where top trie is not prefixed).
That way we got more compact prefix.
The counter is never on state, it is implementation strategy for keyvaluedb (mapping storage key counter at kvdb) or some 'higher level'.
The thing that needs a bit of work is that the mapping storage key - child counter needs to be fork aware so it needs some the kind of storage that was put in place in previously dropped pr.
An alternate design is to put the counter on state, but it is technical implementation details and have nothing to do there.

> I doubt we'll have any support for separate collections in the storage backend. It was always designed to be a simple key-value store.

Sometime rocksdb is not the best for some problematic, so having a trait that give more possibility and allows some alternate storage design could sound good. But yes does not have to be keyvaluedb trait modification just some storage trait allowing different implementation per sub-collection.
I mean, for instance if my child trie is a length bounded collection, I could certainly store thing as a versioned binary blob rather than storing the internal trie structure in a db. It really gives flexibility over the way the backend is implemented.

> The partitioning (prefixing) is done by a higher level code, which gives it more flexibility

I am not sure I understand this point.

> Reference counting is used to track the same keys introduced in different forks, but no in the same state or its ancestors.

Ah, right, my mistake. Isn't it doing the job still (state-db level only of cours)?

>  To track references through the entire history we'd need to have persistent counters in the DB and that would kill performance.

:+1: That was my point of view, I was only speaking about the `state-db` level.

arkpar:
> 
> Ah, right, my mistake. Isn't it doing the job still (state-db level only of cours)?
> 

state-db tracks references for forks in memory. It can do that because forks are relatively short and we can use reference counting for non-canonicalized window only. If some canonical state inserts a value in block N and adds another reference at N + 1000000, state-db would not be able to track that.

cheme:
> state-db tracks references for forks in memory. It can do that because forks are relatively short and we can use reference counting for non-canonicalized window only. If some canonical state inserts a value in block N and adds another reference at N + 1000000, state-db would not be able to track that.

I think we are on the same page.
State-db rc does avoid state collision as long as content is in state-db, so we can store unprefixed key as I did initially  (prefix being store only once per child trie in the journal).
The prefixing being done on a per child trie basis when we got out of memory limited state-db at block canonicalization.

(there is not this code anymore, even if I wanted to bring it back, which is not the case right now, I would probably do it in another pr).

cheme:
Putting back in progress as #4857 will conflict a lot with this PR and got a higher priority.

client/state-db/src/pruning.rs
```diff
@@ -155,16 +238,29 @@ impl<BlockHash: Hash, Key: Hash> RefWindow<BlockHash, Key> {
...
+		)).collect();
+
+		let journal_record = JournalRecordV1 {
 			hash: hash.clone(),
 			inserted,
 			deleted,
 		};
 		let block = self.pending_number + self.death_rows.len() as u64;
-		let journal_key = to_journal_key(block);
+		let journal_key = to_old_journal_key(block);
```
cheme: incorrect should be new.

cheme:
```suggestion
		let journal_key = to_journal_key_v1(block);
```

arkpar:
In #5769 I've made prefixing optional for the database that does not require them.

`PrefixedMemoryDB` that currently acts as the trie transaction does indeed mix keys and prefixes right away. A better implementation would be to keep them separated and let the `client-db` decide whether to use prefixes or not. I would not bother changing that though, we'll probably drop support for rocksdb at some point, or implement reference counting at the lower level. The long term plan is to remove rocksdb and prefixing all together.

cheme:
> In #5769 I've made prefixing optional for the database that does not require them.

So the way it works in #5769 is that you switch at client level the storage accessed by the state machine when querying.

And you also need to probably use a MemoryDB overlay in state machine rather than a PrefixedMemory one (actually you drop of additional content from key in client, which also work but is a bit more costy but code is way more readable than having an additional types everywhere). Ok just realize that it was in your comment :)

My first version of this PR (the one where child info was put in the journals), acts rather similarly by adding the child trie prefix at the place where #5769 is removing it (but that is can not achieve what #5769 does because the trie prefixes are already here).

So the question at this point is 'is this PR of any use' (I am talking from the perspective of the first version where child info was written in journals).
The goods 
- separation of child trie contents: this is not really necessary at this point but if at any time we want a child trie with different hash or a child trie storage that does not work as a merkle trie, or even using different db column depending on child trie (different hash length on a db that do not support that, or for performance purpose), this can prove usefull. The point is that it is a bit of refactoring.

The bads
- btreemap everywhere, new journal storage format.

TLDR; this PR split all data payload in order to allows the db to use child trie definition, and more generally allow db specific process for child trie as in #5280.

So even if we decide to drop #5280, this could still be interesting, (if dropped, I do not know if it will be easy to bring back to life).
There is also some code improvement that can be extracted (but I see some similar thing in #5769 regarding ephemeral).

It could make sense to drop those two pr (just wish we did this when they were drafted and got no real alternative for the rent use case).

arkpar:
> separation of child trie contents: this is not really necessary at this point but if at any time we want a child trie with different hash or a child trie storage that does not work as a merkle trie, or even using different db column depending on child trie (different hash length on a db that do not support that, or for performance purpose), this can prove usefull. The point is that it is a bit of refactoring.

I would not complicate things now because of this potential future implementation. If at some point separate storage for child trees would be required, some changes to the backend would still be needed. E.g. rocksdb does not really support dynamic columns, and neither does parity-db at this point. Also, it might make sense to reuse the existing code, rather than extend it. I.e have a separate instance of `StateDb` for each chile tree.

cheme:
> have a separate instance of StateDb for each chilie tree.

Yes, that is a different way to do it, then state-machine is the only abstraction that cannot in my opinion be split (need to share info between child trie).
The glue between top trie and child trie is mainly `full_storage_root`, and it is in Backend (for proof we consume backend).
But if we consider moving it upward in externality and create a new ProofProcessing ext (currently proof are run over TrieStorageBackend Backend impl).
We would have one Ext containing multiple overlay change and multiple backend with possibly Statedbs as a backends.
That seems like a bigger step from existing code (the code internally was managing child trie content into overlaydb and got a single backend using multiple child trie as needed).
Statedb being directly use as a backend then we could have one statedb by child trie.
It looks like a lot more code change, but indeed better architecture, I remember trying to go in this direction and being blocked (probably on the proof code, but I feel like it is doable).

In #4938 I did rollback these pr change and there was a single place where I needed to keep using child_info as storage input and it was `TrieBackendStorage`. This is because it is the backend for proof production and I needed to split proof payload for compaction. Doing with multiple backend and multiple proof recorder seems possible (in #4938 I kept the possibility to use a flat proof recorder for perf but it does not make much sense, but generally #4938 is allowing too many configuration but at this point it is mainly for testing the proof compaction).

gnunicorn:

cheme:
Closing this as it all depends on replies about question in https://github.com/paritytech/substrate/pull/5280#issuecomment-625144600 and there is no use in keeping two pr frozen behind this, will reopen if 5280 approach get some kind of positive feedback and we proceed this way.

PR #4828, opened 2020-02-04
Pause Kademlia if too many connections

tomaka:
Substrate currently opens **a lot** of TCP connections.
I'm going to properly investigate why tomorrow, but here's a small PR that makes sense to me: we should stop the discovery process if we have a lot of existing connections.

### DIFF
--- a/client/network/src/discovery.rs
+++ b/client/network/src/discovery.rs
@@ -87,6 +87,8 @@ pub struct DiscoveryBehaviour<TSubstream> {
 	/// If false, `addresses_of_peer` won't return any private IPv4 address, except for the ones
 	/// stored in `user_defined`.
 	allow_private_ipv4: bool,
+	/// Number of active connections over which we interrupt the discovery process.
+	discovery_only_if_under_num: u64,
 }
 
 impl<TSubstream> DiscoveryBehaviour<TSubstream> {
@@ -98,6 +100,7 @@ impl<TSubstream> DiscoveryBehaviour<TSubstream> {
 		user_defined: Vec<(PeerId, Multiaddr)>,
 		enable_mdns: bool,
 		allow_private_ipv4: bool,
+		discovery_only_if_under_num: u64,
 	) -> Self {
 		if enable_mdns {
 			#[cfg(target_os = "unknown")]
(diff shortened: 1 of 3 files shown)


client/network/src/service.rs
```diff
@@ -232,6 +232,7 @@ impl<B: BlockT + 'static, S: NetworkSpecialization<B>, H: ExHashT> NetworkWorker
 					TransportConfig::MemoryOnly => false,
 					TransportConfig::Normal { allow_private_ipv4, .. } => allow_private_ipv4,
 				},
+				u64::from(params.network_config.out_peers) * 2,
```
arkpar: I think this should be a a fixed number of additional discovery connections, rather than a x2. Once we reach the target number of peers, it is fine for discovery to be less aggressive.

client/network/src/discovery.rs
```diff
@@ -406,6 +418,10 @@ where
 				NetworkBehaviourAction::GenerateEvent(event) => {
 					match event {
 						MdnsEvent::Discovered(list) => {
+							if self.num_connections < self.discovery_only_if_under_num {
```
twittner:
```suggestion
							if self.num_connections >= self.discovery_only_if_under_num {
```
?

arkpar:
> > Substrate currently opens a lot of TCP connections.
> 
> What is "a lot" in terms of number of established connections to number of (online) peers? 

Devops reported ~1500 ESTABLISHED TCP connections on the validator nodes. 

Regardless, there should be a hard limit on both:
1.  Maximum number of TCP connections the node can make (no matter if these are for discovery or anything else) 
2. The rate of new connections per seconds.

client/network/src/discovery.rs
```diff
@@ -406,6 +418,10 @@ where
 				NetworkBehaviourAction::GenerateEvent(event) => {
 					match event {
 						MdnsEvent::Discovered(list) => {
+							if self.num_connections >= self.discovery_only_if_under_num {
```
andresilva: Not familiar with this code so might be a stupid question. If we already discovered the node why not keep it? Kademlia and sub protocol account for the peer limits separately right?

tomaka: I guess it doesn't really matter for mDNS.
I applied the change to mDNS as well because the variable name is about stopping discovery altogether, and not just Kademlia.

client/network/src/service.rs
tomaka: I changed it for `out_peers + 15`. The `15` is a bit magic, but I think this option is a bit too specific to warrant a field in the configuration.

PR #4831, opened 2020-02-05
Remove support for secio

tomaka:
It's always been intended that we deprecate secio in favour of noise. Noise has been working in production for the past six months to a year, without causing any issue. We can say it's working.

Let's remove secio before we ship Substrate 2.0, in order to avoid having to make a networking breaking change afterwards.

This unfortunately breaks the browser stuff (cc @expenses). The intent is to restore it as soon as possible by doing https://github.com/libp2p/rust-libp2p/issues/1414.

### DIFF
--- a/client/network/src/transport.rs
+++ b/client/network/src/transport.rs
@@ -17,12 +17,10 @@
 use futures::prelude::*;
 use libp2p::{
 	InboundUpgradeExt, OutboundUpgradeExt, PeerId, Transport,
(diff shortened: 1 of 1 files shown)


arkpar:
Is there a libp2p cargo feature to remove compile time dependency as well?

tomaka:
> Is there a libp2p cargo feature to remove compile time dependency as well?

Unfortunately no.
cc https://github.com/libp2p/rust-libp2p/issues/1364

expenses:
😢 👍

PR #4832, opened 2020-02-05
Join verifications in batch async/await verification 

NikVolf:
@marcio-diaz reported that it might be slow for his use cases

### DIFF
--- a/client/transaction-pool/graph/src/pool.rs
+++ b/client/transaction-pool/graph/src/pool.rs
@@ -378,8 +378,13 @@ impl<B: ChainApi> Pool<B> {
 		let block_number = self.resolve_block_number(at)?;
 		let mut result = HashMap::new();
 
-		for xt in xts {
-			let (hash, validated_tx) = self.verify_one(at, block_number, xt, force).await;
+		for (hash, validated_tx) in
+			futures::future::join_all(
+				xts.into_iter()
+					.map(|xt| self.verify_one(at, block_number, xt, force))
+			)
+			.await
+		{
 			result.insert(hash, validated_tx);
 		}
 


client/transaction-pool/graph/src/pool.rs
```diff
@@ -378,8 +378,13 @@ impl<B: ChainApi> Pool<B> {
...
 
-		for xt in xts {
-			let (hash, validated_tx) = self.verify_one(at, block_number, xt, force).await;
+		for (hash, validated_tx) in
+			futures::future::join_all(
+				xts.into_iter()
+					.map(|xt| self.verify_one(at, block_number, xt, force))
+			)
+			.await
+		{
```
marcio-diaz: Maybe use another `map` instead of `for`.

TBH, I'm not sure this fixes the issue. I tried exactly this change and it looked still sequential.

NikVolf: `join_all` is parallel afaik, so

marcio-diaz: Yup, but I'm not sure that what is inside `verify_one` makes it parallel. 
Anyway, I'll try to make a minimal version and check.

rphmeier: The `join_all(...).await` will wait for all to be joined before proceeding, but that's probably fine here.
If you want to do an iteration of the loop as soon as any are ready, use `FuturesUnordered`.

rphmeier: Since each iteration of the loop is so light it probably doesn't matter much, the main thing is to start all the work at once.

NikVolf: You mean use `FuturesUnordered` to start first futures while we are still iterating and creating next ones? I don't think we require this here, yep.

rphmeier: I mean that with `FuturesUnordered` you can process the first result before all have completed, with join you wait for all to complete. Makes no difference here though

client/transaction-pool/graph/src/pool.rs
```diff
@@ -378,8 +378,13 @@ impl<B: ChainApi> Pool<B> {
 		let block_number = self.resolve_block_number(at)?;
 		let mut result = HashMap::new();
 
-		for xt in xts {
-			let (hash, validated_tx) = self.verify_one(at, block_number, xt, force).await;
+		for (hash, validated_tx) in
+			futures::future::join_all(
```
tomusdrw: tbh I'd prefer to assign the result to a separate variable to avoid this super long `for` block :) But it's a minor style thing.

tomusdrw:
@marcio-diaz Would you be able to run your thing and collect some logs (`-ltxpool=trace`), maybe we can add some more to make sure it is running in parallel.

PR #4835, opened 2020-02-05
node: disable grandpa automatic finality fallback

andresilva:
The current implementation of the finality tracker can only track finality using a window of the last 100 blocks, with 3s block times this means that any GRANDPA stall longer than 5 minutes triggers a forced authority set change. This feature is currently unused in Polkadot and we don't have plans to use the automatic fallback anytime soon (it would require changing the finality tracker implementation to efficiently allow tracking longer windows), and instead we plan on implementing a way to manually trigger a forced change (which on Kusama/Polkadot would happen through democracy). Having this enabled makes it harder to diagnose any finality issues introduced by recent changes / test builds on Flaming Fir since the authority set will be forcefully changed after any temporary stall.

I think we should reset flaming fir and redeploy it after this is merged. The reason being that in the current state it already went through some forced authority set changes (which kills light clients right now). Alternatively we can just do a runtime upgrade.

### DIFF
--- a/bin/node/runtime/src/lib.rs
+++ b/bin/node/runtime/src/lib.rs
@@ -476,7 +476,7 @@ parameter_types! {
 }
 
 impl pallet_finality_tracker::Trait for Runtime {
-	type OnFinalizationStalled = Grandpa;
+	type OnFinalizationStalled = ();
 	type WindowSize = WindowSize;
 	type ReportLatency = ReportLatency;
 }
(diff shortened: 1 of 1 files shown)


bkchr:
@ddorgan can you take care of resetting FF?

andresilva:
I talked to David earlier and we decided not to reset for now. Let's just push a runtime upgrade for now and reset later if needed.

PR #4836, opened 2020-02-05
ci: enable build for pre-tags

gabreal:
https://github.com/paritytech/devops/issues/419

building and publishing rustdocs is requested to happen for pre-tags as the distance to official tagged releases is too long.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -525,7 +525,7 @@ publish-gh-doc:
       insteadOf = "https://github.com/"
       EOC
     - unset GITHUB_TOKEN
-    - git clone https://github.com/substrate-developer-hub/rustdocs.git
+    - git clone --depth 1 https://github.com/substrate-developer-hub/rustdocs.git
     - rsync -ax --delete ./crate-docs/ ./rustdocs/${CI_COMMIT_REF_NAME}/
     - cd ./rustdocs; git add .
     - git commit -m "update rustdoc ${CI_COMMIT_REF_NAME}"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -65,6 +65,7 @@ variables:
   only:
     - master
     - /^v[0-9]+\.[0-9]+.*$/        # i.e. v1.0, v2.1rc1
+    - /^pre-v[0-9]+\.[0-9]+-[0-9a-f]+$/
     - web
 
 


gabreal:
the https://github.com/substrate-developer-hub/rustdocs.git repository has **4.1Gb** now. @shawntabrizi can we host it maybe differently? ci jobs are taking more and more time to check it out every time a tag or master is being build.

JoshOrndorff:
Does this change apply to tags? or just branches?

Regarding the size of https://github.com/substrate-developer-hub/rustdocs, I notice there are both `v1.0` and `v1.0.0` directories. Is that intentional?

gabreal:
> Does this change apply to tags? or just branches?

https://docs.gitlab.com/ee/ci/yaml/#onlyexcept-basic :
"only defines the names of branches and tags for which the job will run."

> Regarding the size of https://github.com/substrate-developer-hub/rustdocs, I notice there are both `v1.0` and `v1.0.0` directories. Is that intentional?

`v1.0` are the docs from substrate `v1.0` branch, `v1.0.0` is from the run of the tag.

bkchr:
> the https://github.com/substrate-developer-hub/rustdocs.git repository has **4.1Gb** now. @shawntabrizi can we host it maybe differently? ci jobs are taking more and more time to check it out every time a tag or master is being build.

Can we not "prune" aka delete old revisions? 

And could CI not only fetch the HEAD?

gabreal:
`.git` folder has `1.6Gb`
entire repo `4.7Gb`

shawntabrizi:
> the https://github.com/substrate-developer-hub/rustdocs.git repository has **4.1Gb** now. @shawntabrizi can we host it maybe differently? ci jobs are taking more and more time to check it out every time a tag or master is being build.

I am open to pretty much any suggestion. We only set things up this way because it made sense at the time for what we were looking to do. If the scope of the repo is changing, process can as well.

JoshOrndorff:
Thanks for the link to the gitlab docs. This solves https://github.com/paritytech/devops/issues/419 so I've approved.

I've moved the discussion about the size of the rustdocs repo to an issue there and tagged you all. That issue could also be moved to devops repo if that makes more sense.

Thank for writing this @gabreal !

JoshOrndorff:
@gabreal I pushed this tag about an hour ago https://github.com/paritytech/substrate/releases/tag/pre-v2.0-3e65111 but I don't see a corresponding directory at https://github.com/substrate-developer-hub/rustdocs . Is there something else I need to do to trigger the build?

bkchr:
The tag probably needs to include the commit of this pr.

gabreal:
exactly. the tag is already on gitlab: https://gitlab.parity.io/parity/substrate/-/tags
it just doesn't contain this pr.

JoshOrndorff:
Oh I see. That makes sense. Is there anything we can do manually to build the rustdocs for 3e65111?

gabreal:
one could build and push them manually like running the task manually. triggering that for the ci is more complicated. one would have to tweak ci pipelines on a dedicated branch forking off from that tag and manually set the publishing path.

JoshOrndorff:
Just to be sure I'm understanding, I could checkout out the correct Substrate commit, run `cargo doc` locally, and manually commit those docs to the [hosted rustdocs repo](https://github.com/substrate-developer-hub/rustdocs)?

That sounds easy enough. I'll give it a go.

Issue #4826, opened 2020-02-04
Possible Allocation Error: `AllocatorOutOfSpace`


shawntabrizi:
When I was trying to benchmark the runtime, I ran into an issue where a loop was causing `AllocatorOutOfSpace` errors.

You can reproduce this error by checking out this commit of Substrate:

https://github.com/paritytech/substrate/tree/7bd0ae11cdff291b805c9633f184e59f49d17e43

Run a node with `./target/release/substrate --dev --execution=Wasm --wasm-execution=Compiled`

Then call the RPC `state_call` with `IdentityBenchmarks_run_benchmarks`

You will notice that the processes starts, completes a few loops, then returns the alocator out of space error.

We added some print statements around these error points and noticed a consistent growth in `item_size` which ultimately causes this error.

To solve this problem, we can reduce the size of the loop (before we hit a memory error), or we can use `native` execution.

Let me know if I can provide any more details.

shawntabrizi:
Still facing this issue actually on the latest stuff. Large loops do not seem to be good here:

```
====================

Version: 2.0.0-5ee888f90-x86_64-macos

stack backtrace:
   0: backtrace::backtrace::trace
   1: backtrace::capture::Backtrace::new
   2: sp_panic_handler::set::{{closure}}
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::continue_panic_fmt
   5: rust_begin_unwind
   6: core::panicking::panic_fmt
   7: core::result::unwrap_failed
   8: <<sp_io::allocator::HostFunctions as sp_wasm_interface::HostFunctions>::host_functions::ExtAllocatorMallocVersion1 as sp_wasm_interface::Function>::execute
   9: std::panicking::try::do_call
  10: __rust_maybe_catch_panic
  11: sc_executor_wasmtime::trampoline::stub_fn


Thread 'main' panicked at 'Failed to allocate memory: "Allocator ran out of space"', src/libcore/result.rs:1165

This is a bug. Please report it at:

        https://github.com/paritytech/substrate/issues/new

Error: Service(Other("Error executing runtime benchmark: FunctionExecution(\"ext_allocator_malloc_version_1\", \"Failed to allocate memory: \\\"Allocator ran out of space\\\"\")"))

**Comment by kianenigma** [comment]
Similar issue:

```
====================

Version: 2.0.0-2ed0a1d90-x86_64-macos

stack backtrace:
   0: backtrace::backtrace::trace
[...]

PR #4837, opened 2020-02-05
Fix memory leak in runtime interface

bkchr:
We used `slice::from_raw_parts` in runtime-interface which did not free
the memory afterwards. This pr changes it to `Vec::from_raw_parts` to
make sure `drop` is called properly and the values are freed.

### DIFF
--- a/primitives/runtime-interface/src/impls.rs
+++ b/primitives/runtime-interface/src/impls.rs
@@ -38,9 +38,6 @@ use sp_std::{any::TypeId, mem, vec::Vec};
 #[cfg(feature = "std")]
 use sp_std::borrow::Cow;
 
-#[cfg(not(feature = "std"))]
-use sp_std::{slice, boxed::Box};
-
 // Make sure that our assumptions for storing a pointer + its size in `u64` is valid.
 #[cfg(all(not(feature = "std"), not(feature = "disable_target_static_assertions")))]
 assert_eq_size!(usize, u32);
@@ -195,12 +192,12 @@ impl<T: 'static + Decode> FromFFIValue for Vec<T> {
 	fn from_ffi_value(arg: u64) -> Vec<T> {
 		let (ptr, len) = unpack_ptr_and_len(arg);
 		let len = len as usize;
+		let data = unsafe { Vec::from_raw_parts(ptr as *mut u8, len, len) };
 
 		if TypeId::of::<T>() == TypeId::of::<u8>() {
-			unsafe { mem::transmute(Vec::from_raw_parts(ptr as *mut u8, len, len)) }
+			unsafe { mem::transmute(data) }
 		} else {
-			let slice = unsafe { slice::from_raw_parts(ptr as *const u8, len) };
-			Self::decode(&mut &slice[..]).expect("Host to wasm values are encoded correctly; qed")
+			Self::decode(&mut &data[..]).expect("Host to wasm values are encoded correctly; qed")
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


primitives/runtime-interface/test-wasm/src/lib.rs
```diff
@@ -39,6 +39,11 @@ pub trait TestApi {
 		data
 	}
 
+	/// Returns 16kb data.
+	fn return_16kb() -> Vec<u32> {
+		vec![0; 4 * 1024]
```
pepyakin: Can we actually use `u8` here with length `16 * 1024` ? : )

primitives/runtime-interface/src/impls.rs
```diff
@@ -195,12 +192,12 @@ impl<T: 'static + Decode> FromFFIValue for Vec<T> {
 	fn from_ffi_value(arg: u64) -> Vec<T> {
 		let (ptr, len) = unpack_ptr_and_len(arg);
 		let len = len as usize;
+		let data = unsafe { Vec::from_raw_parts(ptr as *mut u8, len, len) };
```
pepyakin: Not related to this PR thus not to address here, but is there a guarantee that `len` > 0?

shawntabrizi:
Closes https://github.com/paritytech/substrate/issues/4826

primitives/runtime-interface/test-wasm/src/lib.rs
bkchr: `Vec<u8>` uses a different code path internally that doesn't leak memory before. This code path uses `SCALE` to pass the data.

primitives/runtime-interface/src/impls.rs
bkchr: I will add a check, but I think it should be okay without it as well: https://doc.rust-lang.org/1.4.0/src/alloc/raw_vec.rs.html#424

primitives/runtime-interface/test-wasm/src/lib.rs
pepyakin: Ah good point! Worth adding a comment?

primitives/runtime-interface/src/impls.rs
pepyakin: Ah, no, I more thought about that with zero len often comes a zero pointer

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5d9e82f1097c43c090f73e1c37ceeb22

I think in order to avoid this pitfall and also instead of trying to ensure that `ptr` is actually has a proper alignment for `T` it is just easier to explicitly handle the zero length case.

bkchr: Yeah it is handled now, but WTF. Why does `reconstruct` returns a `None`?

primitives/runtime-interface/test-wasm/src/lib.rs
bkchr: Done

primitives/runtime-interface/src/impls.rs
pepyakin: `Vec` can be thought as if it has a `data_ptr: NonNull` field. If you actually put null there it is UB.

That demo consistently exhibits such a behavior, because Rust uses layout optimization for enums 
based on illegal values. That's why `ptr::null()` maps to `None` in `Option<NonNull<T>>`.

bkchr: Interesting :)

Issue #4776, opened 2020-01-30
Service's Configuration has optional fields that shouldn't be optional


cecton:
Following the discussion on: https://github.com/paritytech/substrate/pull/4692/files/bea809d4c14a2ede953227ac885e3b3f9771c548#r372049016

There are 2 conflicting things we would like to have:
 * Running a node requires a Configuration and some of the fields are not optional but they are defined as Option
 * We should be able to define defaults to the Configuration before the command line get parsed. Therefore some of the fields need to be Option

> The idea is to be able to make a Configuration before it being modified by the commandline arguments. For that I need defaults and chain_spec, like some other fields, are Option even though they are required for running a node.

> It's not ideal. I think it would be best to have some kind of intermediate object, one that has the Options that can be converted to one that doesn't have and it would be fail-able at that point.

TimvanScherpenzeel:
I was going to make a new ticket but I think my issue is possible related to this ticket and known.

It is currently not possible to disable the `grafana data source server` by setting the `grafana_port` to `None` as suggested in:

https://github.com/paritytech/substrate/blob/26e37ff685819646633f5246419c03c02df81dff/client/service/src/config.rs#L75-L76

This appears to be because of:

https://github.com/paritytech/substrate/blob/26e37ff685819646633f5246419c03c02df81dff/client/cli/src/lib.rs#L637-L642

The `grafana data source server` should not be a strict requirement to run a node right?

Same goes for `rpc_http` and `rpc_ws` (where setting `None` doesn't actually disable them):

https://github.com/paritytech/substrate/blob/26e37ff685819646633f5246419c03c02df81dff/client/cli/src/lib.rs#L628-L635

cecton:
Is it new? It doesn't look related to my change, the code was there before:

https://github.com/paritytech/substrate/blob/1ed1cc542110aaaee9cf0de87a7c256a1de407c6/client/cli/src/lib.rs#L1072-L1076

TimvanScherpenzeel:
It is not directly related to your latest changes I believe but I thought it might relate to this ticket. If it does not I’ll gladly make a new ticket specifically for the issue that I’m facing.

cecton:
No no I think you're right, it's definitely related to this ticket. I was just afraid I broke something.

The comment really is an indication that it was meant to be an option but clearly it's never None when running a node.

cecton:
[...]

PR #4842, opened 2020-02-06
Add tests & Service's Configuration has optional fields that shouldn't be optional

cecton:
Related to #4776 
Related to https://github.com/paritytech/polkadot/pull/832

To summarize the changes:
1. I did not manage to validate with types the service's Configuration. But I did reduce the possibility of errors by moving all the "fill" functions to their respective structopts
2. I split params.rs to multiple modules: one module params for just CLI parameters and one module commands for CLI subcommands (and RunCmd). Every command and params are in their own file so things are grouped better together and easier to remove
3. I removed the run and run_subcommand helpers as they are not helping much anymore. Running a command is always a set of 3 commands: 1. init 2. update config 3. run. This still allow the user to change the config before arguments get parsed or right after.
4. I added tests for all subcommands.
5. ~I updated some feature flags with the new benchmark subcommands because client/cli couldn't build anymore. So I added a flag rocksdb~ [edit: it has been fixed in master and this is no longer necessary so I removed it]

Overall the aim is to improve the situation with the Configuration and the optional parameters, add tests, make the API more consistent and simpler.


bin/node/cli/tests/build_spec_works.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::{process::{Command, Stdio}, fs, path::PathBuf};
+
+mod common;
+
+#[test]
+fn build_spec_works() {
+	let base_path = "build_spec_test";
+
+	let _ = fs::remove_dir_all(base_path);
+	let status = Command::new(cargo_bin("substrate"))
```
bkchr: I think having some checks that the stuff we expect is happening, would be nice.

Like that `build-spec` is a valid json and contains some information.

cecton: Agree! I can do the valid JSON. But what information should I check?

client/cli/src/lib.rs
```diff
@@ -215,487 +133,26 @@ where
...
-	BC: ServiceBuilderCommand<Block = BB> + Unpin,
-	BB: sp_runtime::traits::Block + Debug,
-	<<<BB as BlockT>::Header as HeaderT>::Number as std::str::FromStr>::Err: std::fmt::Debug,
-	<BB as BlockT>::Hash: std::str::FromStr,
-{
-	let shared_params = subcommand.get_shared_params();
-	init(shared_params, version)?;
-	init_config(&mut config, shared_params, version, spec_factory)?;
-	subcommand.run(config, builder)
-}
```
cecton: @bkchr  since you're here already I'm showing you this.

As you can see I'm dropping all of this. I know you said it would be best to have to functions directly on the commands and that's what I did. Every command now has a `update_config` and a `run` functions. They are separated to allow the developer to force/override things after parsing the command line. Check in bin/node/cli/src/commands.rs

client/cli/src/lib.rs
```diff
@@ -174,14 +84,22 @@ where
 		.name(version.executable_name)
 		.author(version.author)
 		.about(version.description)
-		.version(full_version.as_str());
+		.version(full_version.as_str())
+		.settings(&[
+			AppSettings::GlobalVersion,
+			AppSettings::ArgsNegateSubcommands,
+			AppSettings::SubcommandsNegateReqs,
+		]);
```
cecton: @bkchr I'm putting back this here.

I didn't want to have them in sc_cli because I thought it would be best if they are a choice of the dev who makes the app that uses sc_cli. But I noticed in polkadot and cumulus that this was error prone especially since now that RunCmd is not a variant of the subcommands anymore.

So I added more doc on the subject on this function.

This is still flexible because the dev can choose to not use this helper and use structopt's from_iter directly if they want.

client/cli/src/lib.rs
```diff
@@ -19,127 +19,34 @@
...
-	SharedParams, ImportParams, ExecutionStrategy, Subcommand, RunCmd, BuildSpecCmd,
-	ExportBlocksCmd, ImportBlocksCmd, CheckBlockCmd, PurgeChainCmd, RevertCmd,
-	BenchmarkCmd,
-};
-pub use traits::GetSharedParams;
-use app_dirs::{AppInfo, AppDataType};
+pub use params::*;
+pub use commands::*;
+pub use arg_enums::*;
+pub use error::*;
```
cecton: @bkchr a lot more things are now directly accessible without having to go into the submodules. But a lot more things is also exposed (more than before). For example, all the arg_enums are now exposed. This allow another crate to manipulate them if they want (or re-use them). Anyway, all structopt should be recursively public I think

bin/node/cli/src/command.rs
```diff
@@ -14,46 +14,49 @@
...
 
-			sc_cli::fill_config_keystore_in_memory(&mut config)?;
+			config.use_in_memory_keystore()?;
 
 			match ChainSpec::from(config.expect_chain_spec().id()) {
 				Some(ref c) if c == &ChainSpec::Development || c == &ChainSpec::LocalTestnet => {},
-				_ => panic!("Factory is only supported for development and local testnet."),
+				_ => return Err(
+					"Factory is only supported for development and local testnet.".into()
+				),
```
cecton: This was panicking with a nice stacktrace in the console. I don't know why. It seems to me it's a normal user error so I changed it. No more stacktrace but the error message is still explicit enough imho.

bin/node-template/node/src/command.rs
```diff
@@ -15,32 +15,35 @@
...
+		None => {
+			opt.run.init(&version)?;
+			opt.run.update_config(&mut config, chain_spec::load_spec, &version)?;
+			opt.run.run(
+				config,
+				service::new_light,
+				service::new_full,
+				&version,
+			)
+		},
```
cecton: This should be way less error prone than before.

Before: I had put opt.run.shared_params in the subcommand's init and it was a bug.

client/service/src/benchmark.rs
```diff
@@ -0,0 +1,69 @@
+#![cfg(feature = "rocksdb")]
```
cecton: This whole file and the whole benchmark command doesn't work without the feature rocksdb in sc_service. So I added the same rocksdb flag for sc_cli so I can compile it alone (with or without rocksdb) and it doesn't break.

cecton:
Now also fixes https://github.com/paritytech/substrate/issues/4888

seunlanlege:
> I split params.rs to multiple modules:

Oh no :scream:

cecton:
> > I split params.rs to multiple modules:
> 
> Oh no scream , already did the same thing here #4954
> 
> Rebasing is going to be a nightmare

:rofl: sorry

bin/node/cli/tests/factory.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use assert_cmd::cargo::cargo_bin;
+use std::process::{Command, Stdio};
+use tempfile::tempdir;
+
+mod common;
+
+#[test]
+fn factory_works() {
+	let base_path = tempdir().unwrap();
```
NikVolf: better always use `.expect("<test part that failed>")` - may save some time for future generation debugging test failures

bin/node/cli/tests/factory.rs
```diff
@@ -0,0 +1,39 @@
...
+
+	let status = Command::new(cargo_bin("substrate"))
+		.stdout(Stdio::null())
+		.args(&["factory", "--dev", "-d", base_path.path().to_str().unwrap()])
```
NikVolf:
```suggestion
		.args(&["factory", "--dev", "-d", base_path.path().to_string_lossy()])
```

or `expect`..

cecton: Or even better: let it convert to OsStr by itself :grin: (check the code)

bin/node-template/node/src/command.rs
```diff
@@ -15,32 +15,35 @@
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
 use sp_consensus_aura::sr25519::{AuthorityPair as AuraPair};
-use sc_cli::{VersionInfo, error};
+use sc_cli::{VersionInfo, self};
```
bkchr: Self not required.

bin/node-template/node/src/main.rs
```diff
@@ -7,10 +7,10 @@ mod service;
 mod cli;
 mod command;
 
-pub use sc_cli::{VersionInfo, error};
+pub use sc_cli;
```
bkchr: Not sure why you want to public re-export something from a binary.

bin/node/cli/bin/main.rs
```diff
@@ -18,10 +18,10 @@
 
 #![warn(missing_docs)]
 
-use sc_cli::VersionInfo;
+use sc_cli;
```
bkchr: Not required.

bin/node/cli/tests/build_spec_works.rs
```diff
@@ -0,0 +1,41 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+#![cfg(unix)]
```
gnunicorn: Why would this not run on windows? I would expect the test to also run on windows, if not we should probably file an issue for it and reference it here as a FIXME...

If not, I suggest removing this line

```suggestion
```

client/cli/src/arg_enums.rs
```diff
@@ -14,10 +14,74 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
+// NOTE: we allow missing docs here because arg_enum! creates the function variants without doc
```
gnunicorn: maybe we should file a bug with clap about this? Sounds like a reasonable ask...

client/cli/src/commands/runcmd.rs
```diff
@@ -0,0 +1,737 @@
...
+	///
+	/// Default is local. Note: not all RPC methods are safe to be exposed publicly. Use a RPC proxy
+	/// server to filter out dangerous methods. More details: https://github.com/paritytech/substrate/wiki/Public-RPC.
+	/// Use `--unsafe-ws-external` to suppress the warning if you understand the risks.
+	#[structopt(long = "ws-external")]
+	pub ws_external: bool,
+
+	/// Listen to all Websocket interfaces.
+	///
+	/// Same as `--ws-external`.
```
gnunicorn:
```suggestion
	/// Same as `--ws-external` but doesn't warn you about it.
```

client/service/src/benchmark.rs
bkchr: This whole benchmark command was moved to its own crate `utils/frame/benchmarking-cli`

client/cli/src/params/transaction_pool_params.rs
```diff
@@ -0,0 +1,50 @@
...
+	#[structopt(long = "pool-kbytes", value_name = "COUNT", default_value = "20480")]
+	pub pool_kbytes: usize,
+}
+
+impl TransactionPoolParams {
+	/// Fill the given `PoolConfiguration` by looking at the cli parameters.
+	pub fn update_config<G, E>(
+		&self,
+		config: &mut Configuration<G, E>,
+	) -> error::Result<()> {
```
bkchr:
```suggestion
	) -> crate::error::Result<()> {
```

client/cli/src/params/transaction_pool_params.rs
```diff
@@ -0,0 +1,50 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use structopt::StructOpt;
+use sc_service::Configuration;
+
+use crate::error;
```
bkchr:
```suggestion
```

client/cli/src/params/transaction_pool_params.rs
```diff
@@ -0,0 +1,50 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+use structopt::StructOpt;
+use sc_service::Configuration;
+
```
bkchr:
```suggestion
```

client/cli/src/params/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+pub struct BlockNumber(String);
+
+impl FromStr for BlockNumber {
+	type Err = String;
+
+	fn from_str(block_number: &str) -> Result<Self, Self::Err> {
+		if block_number.chars().any(|d| !d.is_digit(10)) {
+			Err(format!(
+				"Invalid block number: {}, expected decimal formatted unsigned integer",
+				block_number
```
bkchr:
```suggestion
				block_number,
```

client/cli/src/params/mod.rs
```diff
@@ -0,0 +1,65 @@
+// Copyright 2018-2020 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2020 Parity Technologies (UK) Ltd.
```

client/cli/src/arg_enums.rs
cecton: Good point, will do!

bin/node/cli/tests/build_spec_works.rs
cecton: Testing on Windows was complicated but you're right that this test should work on Windows. I will create a ticket for that as I don't have time to test.

client/cli/src/node_key.rs
```diff
@@ -1,209 +0,0 @@
-// Copyright 2017-2020 Parity Technologies (UK) Ltd.
```
bkchr: Why isn't this file removed?

bin/node/cli/tests/build_spec_works.rs
cecton: Created #5018

client/cli/src/arg_enums.rs
cecton: Issue created: https://github.com/clap-rs/clap/pull/1702

bin/node-template/node/src/command.rs
```diff
@@ -15,32 +15,35 @@
...
-		None => sc_cli::run(
-			config,
-			opt.run,
-			service::new_light,
-			service::new_full,
-			chain_spec::load_spec,
-			&version,
-		)
+		Some(subcommand) => {
+			subcommand.init(&version)?;
```
bkchr: Why aren't `init` and `update_config` not done inside of `run()`? Someone could just call `run()` without calling the other methods before and it would be incorrect.

cecton:
> Assuming the params and commands didn't change other than being split up, because I only gave them a rough glance ...

Yes well it was complicated because a lot of changes have been pushed to master, especially yesterday.

TL;DR: I'm very confident nothing has been lost.

Long story: I'm experimenting a new workflow where I merge the most non-conflicting commits to my branch to update it and then I create a single merge commit with the conflicting commit where I solve the issue independently.

Because of that, I look at the commit carefully and I replicate the changes in my files properly. It's a bit like what happen with rebase (resolving conflicts commit by commit) except that it's merging instead of rebasing.

> Thanks for all the tests!

You're welcome! :hugs: I have hope to improve a lot `sc_cli` to a point where it would be very stable and intuitive.

bin/node/cli/src/command.rs
```diff
@@ -14,33 +14,36 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
-use sc_cli::{VersionInfo, error};
+use sc_cli::{VersionInfo, self};
```
bkchr:
```suggestion
use sc_cli::VersionInfo;
```

Rust 2018 automatically makes crates usable everywhere without importing them explicitly.

bin/node/cli/tests/build_spec_works.rs
```diff
@@ -0,0 +1,37 @@
...
+use std::process::Command;
+use tempfile::tempdir;
+
+#[test]
+fn build_spec_works() {
+	let base_path = tempdir().expect("could not create a temp dir");
+
+	let output = Command::new(cargo_bin("substrate"))
+		.args(&["build-spec", "--dev", "-d"])
+		.arg(base_path.path().as_os_str())
```
bkchr:
```suggestion
		.arg(base_path.path())
```

bin/node/cli/tests/common.rs
```diff
@@ -32,3 +39,25 @@ pub fn wait_for(child: &mut Child, secs: usize) -> Option<ExitStatus> {
 
 	None
 }
+
+pub fn run_command_for_a_while(base_path: &Path, dev: bool) {
```
bkchr: Missing docs and "a while" is a very vague description :D

bin/node-template/node/src/command.rs
cecton: Because for cumulus we want to be able to modify the config before running the node.

We could have another "run" command that would do all 3 commands but we need to keep the existing 3.

Since a helper would simply call init & update_config & run, I thought it wouldn't be super helpful and I have not added it.

bin/node/cli/src/command.rs
cecton: noted, thanks!

bin/node/cli/tests/common.rs
cecton: I added some doc with more precise description

bin/node-template/node/src/command.rs
cecton: > I removed the run and run_subcommand helpers as they are not helping much anymore. Running a command is always a set of 3 commands: 1. init 2. update config 3. run. This still allow the user to change the config before arguments get parsed or right after.

:scream: you didn't read the PR description

bkchr: Code is law :P 

Yeah it is okay for now. However it could happen that the user just calls `run()` (because nothing prevents them) and it will break.

bkchr: And this "pattern" is a let's repeat some code "pattern" :D

bin/node/cli/tests/common.rs
```diff
@@ -14,7 +14,14 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::{process::{Child, ExitStatus}, thread, time::Duration};
+#![cfg(unix)]
+#![allow(dead_code)]
```
bkchr: What is dead inside here? :P

cecton: I don't know. I think common.rs is executed on its own...

cecton: So everything looks like dead code when it's compiled

client/cli/src/arg_enums.rs
cecton: It's merged. At some point we will be able to remove that arg_enums module ...

PR #4843, opened 2020-02-06
pow: re-add support for algorithms where only linear verification is possible

sorpaas:
This restores the API of PoW engine after https://github.com/paritytech/substrate/pull/4652. No-parent verification now again becomes optional (called `preliminary_verify`). This is needed because for some algorithms only linear verification is possible (for PoW algorithms that just relies on parent block information). Also fixed some unused generic parameters.

### DIFF
--- a/client/consensus/pow/src/lib.rs
+++ b/client/consensus/pow/src/lib.rs
@@ -63,8 +63,8 @@ pub enum Error<B: BlockT> {
 	HeaderUnsealed(B::Hash),
 	#[display(fmt = "PoW validation error: invalid seal")]
 	InvalidSeal,
-	#[display(fmt = "PoW validation error: invalid difficulty")]
-	InvalidDifficulty,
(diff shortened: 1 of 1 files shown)


client/consensus/pow/src/lib.rs
```diff
@@ -450,20 +467,17 @@ pub fn register_pow_inherent_data_provider(
...
+	block_import: BoxBlockImport<B, Transaction>,
 	algorithm: Algorithm,
 	inherent_data_providers: InherentDataProviders,
 ) -> Result<
-	PowImportQueue<B, sp_api::TransactionFor<C, B>>,
+	PowImportQueue<B, Transaction>,
 	sp_consensus::Error
 > where
 	B: BlockT,
-	C: ProvideRuntimeApi<B> + HeaderBackend<B> + BlockOf + ProvideCache<B> + AuxStore,
```
NikVolf: I wonder how many of those generic bounds never get deleted elsewhere

PR #4844, opened 2020-02-06
WIP: Test: does cargo update work?

cecton:


bkchr:
What is the purpose?

cecton:
Apparently some dependency upgrade break one of the test: `should_cleanup_subscribers_if_dropped`

bkchr:
@tomusdrw any idea?

tomusdrw:
`futures` seems like an obvious suspect, perhaps the `block_on_stream` behavior changed? (Checked the code but it doesn't seem so). It looks that even though we dropped `mpsc::UnboundedReceiver` doing a `unbounded_send` does not result in an `Err` :fearful:

bkchr:
Probably some flaky test. Needs fixing, but yeah this pr can be closed.

cecton:
@bkchr it's not flaky. I thought so in the beginning but it's not. We can restart the CI to make sure if you want.

Also if we can't do `cargo update` without breaking the CI we are screwed. This is something we need to fix in one way or another.

I believe this is due to a change of behavior in one of our dependency and they didn't make a major release for it.

bkchr:
Yeah okay, I'm not against fixing it. However we have enough open prs. If you fix it, you can create a pr and we merge it. Normally you should also not call `cargo update`.

cecton:
> Normally you should also not call cargo update.
Why?

bkchr:
What are you doing in here? You can just compile locally. You don't need the CI for this.

If you need some crate updated, the best would be to pin the crate to the requested version in a `Cargo.toml` and if you need to update some package `cargo update -p` is what you should use.

bkchr:
Blindly updating all packages brings us into these troubles you see here.

cecton:
> Blindly updating all packages brings us into these troubles you see here.

I guess so... but updating them one by one manually is a tedious work. As you said, we already have so many PR.

I like to think that our dependencies will follow the semantic versioning as best as they can and from time to time we will have an issue like this one when some dependency changed a behavior.

I've been working on substrate for 1 month and this is the first time it happens. At this rate we will fix this 12 times a year but all our dependencies will be more up-to-date.

bkchr:
Updating them here in the `Cargo.lock` does not brings that much for projects building on Substrate :woman_shrugging: If you include a crate, you don't include its `Cargo.lock`. The only way to solve this properly is to update the `Cargo.toml`'s

cecton:
I know I'm just localizing the breaking dependency precisely and then I will put a `=` in front of our dependency that have a behavior change.

bkchr:
You are wasting ci resources.

cecton:
I'll finish testing locally and push at the end then

cecton:
Futures 0.3.4 has been released today and doesn't have this issue. We can cargo update again

cecton:
```
 442 |         secp256k1::sign(&message, &self.secret).into()
    |                    ^^^^ not found in `secp256k1`
```

@dvdplm I think you changed the API and didn't change the major of the version of libsecp256k1

primitives/trie/Cargo.toml
```diff
@@ -15,13 +15,13 @@ harness = false
 codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false }
 sp-std = { version = "2.0.0", default-features = false, path = "../std" }
 hash-db = { version = "0.15.2", default-features = false }
-trie-db = { version = "0.19.2", default-features = false }
-trie-root = { version = "0.15.2", default-features = false }
-memory-db = { version = "0.18.1", default-features = false }
+trie-db = { version = "0.20.0", default-features = false }
```
bkchr: Did you test this with Kusama? We already had trie updates breaking Kusama.

CC @cheme

cheme: If I checked correctly this is just an update to parity-util-mem, so that should be fine.
But at the same time, testing with kusama don't take much time usually (I mean if there was actual change to trie I would try to resync from scratch which is a few hours (maybe a bit more)).

bkchr: I have seen that this also includes some prefixed key iterator?

cheme: Oh yes. This iterator is not in use anywhere, it is new (would need to PR a change in substrate, and this will not be backward compatible regarding proof production (cumulus :eyes:).

bkchr: Okay :P On your responsibility :D

cecton: There is only my bump of version with this update and a new iterator function, it's fine.

cecton: https://github.com/paritytech/trie/pull/78
https://github.com/paritytech/trie/pull/79

cecton: Oh my page didnt refresh and I answered to old stuff. Nevermind.

cecton:
Something I miss on GitLab: PR that starts with WIP cannot be merged....... cool features to avoir having WIP in the history

bkchr:
We merge in Github, Gitlab can not do anything.

cecton:
I mean miss this feature in GitLab: https://about.gitlab.com/blog/2016/01/08/feature-highlight-wip/

PR #4845, opened 2020-02-06
Enable trace timings logs for transaction factory

marcio-diaz:
It'd be used to measure a couple of things like import times.


client/cli/src/params.rs
```diff
@@ -113,6 +113,20 @@ pub struct SharedParams {
 	/// Sets a custom logging filter.
 	#[structopt(short = "l", long = "log", value_name = "LOG_PATTERN")]
 	pub log: Option<String>,
+
+	/// Comma separated list of targets for tracing
+	#[structopt(long = "tracing-targets", value_name = "TARGETS")]
+	pub tracing_targets: Option<String>,
```
bkchr: For what do we need to have `tracing-targets` when I for example call `purge-chain`?

This does not makes any sense at all.

bkchr: This should have gone into `ImportParams`.

Issue #4849, opened 2020-02-07
purge-chain doesn't work on node-template


xlc:
`node-template (master) $ cargo run -- purge-chain --dev -y`

```
====================

Version: 2.0.0-f3742e70b-x86_64-macos

stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /Users/xiliangchen/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.43/src/backtrace/libunwind.rs:86
      backtrace::backtrace::trace_unsynchronized
             at /Users/xiliangchen/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.43/src/backtrace/mod.rs:66
   1: backtrace::backtrace::trace
             at /Users/xiliangchen/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.43/src/backtrace/mod.rs:53
   2: backtrace::capture::Backtrace::create
             at /Users/xiliangchen/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.43/src/capture.rs:164
   3: backtrace::capture::Backtrace::new
             at /Users/xiliangchen/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.43/src/capture.rs:128
   4: sp_panic_handler::panic_hook
             at primitives/panic-handler/src/lib.rs:148
   5: sp_panic_handler::set::{{closure}}
             at primitives/panic-handler/src/lib.rs:58
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:468
   7: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:373
   8: rust_begin_unwind
             at src/libstd/panicking.rs:302
   9: std::panicking::begin_panic
  10: std::panicking::begin_panic
  11: core::option::Option<T>::expect
             at /rustc/73528e339aae0f17a15ffa49a8ac608f50c6cf14/src/libcore/option.rs:345
  12: sc_service::config::Configuration<G,E>::expect_database
             at /Users/xiliangchen/projects/substrate/client/service/src/config.rs:255
  13: sc_cli::params::PurgeChainCmd::run
             at /Users/xiliangchen/projects/substrate/client/cli/src/params.rs:1128
  14: sc_cli::params::Subcommand::run
             at /Users/xiliangchen/projects/substrate/client/cli/src/params.rs:911
[...]

PR #4851, opened 2020-02-07
Fix CLI setup again

bkchr:
We need to set `config_dir` and `database_path` for almost every
command.
This fixes `purge-chain` and also adds a test to make sure we don't
break it again.

Closes: https://github.com/paritytech/substrate/issues/4849

### DIFF
--- a/client/cli/src/lib.rs
+++ b/client/cli/src/lib.rs
@@ -233,7 +233,8 @@ where
 	SL: AbstractService + Unpin,
 	SF: AbstractService + Unpin,
 {
-	init(&mut config, &run_cmd.shared_params, version, spec_factory)?;
+	init(&run_cmd.shared_params, version)?;
+	init_config(&mut config, &run_cmd.shared_params, version, spec_factory)?;
 	run_cmd.run(config, new_light, new_full, version)
 }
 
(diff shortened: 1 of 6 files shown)


client/cli/src/lib.rs
```diff
@@ -281,6 +288,21 @@ pub fn init(shared_params: &SharedParams, version: &VersionInfo) -> error::Resul
...
+
+	if config.database.is_none() {
+		config.database = Some(DatabaseConfig::Path {
+			path: config
+				.in_chain_config_dir(DEFAULT_DB_CONFIG_PATH)
+				.expect("We provided a base_path/config_dir."),
+			cache_size: None,
+		});
+	}
+
```
cecton: This is breaking my changes. My motivation is to remove all the configuration setup from the `init()` method.

`init()` must be called only once. In cumulus we have multiple `Configuration` and they all need to be properly set up. This is not possible if you put config setup logic in this `init_method` because we won't want to call a second time `init` which will call `init_logger` again.

You need to create a private method `fill_config_dir_and_database_path` that will be called in `update_config_for_running_node` and all the `run()` commands that needs it in `client/cli/src/lib.rs`.

bkchr: I have done it in a similar way. But that just shows us the fucked up design of the CLI.

PR #4854, opened 2020-02-07
Add trait to get module and call names.

marcio-diaz:
This add a couple of traits to get module and call names for each module:

- `Call::all_modules()` => this gives something like ["Balances", "Identity", ...]
- `Call::all_calls("Balances".into())` => this gives something like ["transfer", "set_balance", ...]

I'll use this for benchmarking, but maybe it's useful somewhere else.


frame/support/src/benchmarking.rs
```diff
@@ -0,0 +1,24 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+pub trait Module {
+	fn all_modules() -> &'static [&'static str];
+	fn all_calls(module: &str) -> &'static [&'static str];
+}
+
+pub trait Call {
```
bkchr: I think this can be added to https://crates.parity.io/frame_support/traits/trait.GetCallName.html

frame/support/src/benchmarking.rs
```diff
@@ -0,0 +1,24 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+pub trait Module {
```
bkchr: And this to https://crates.parity.io/frame_support/traits/trait.GetCallMetadata.html

marcio-diaz: Moved.

frame/support/src/traits.rs
```diff
@@ -819,12 +819,18 @@ pub struct CallMetadata {
...
+	fn get_call_names() -> &'static [&'static str];
 	/// Return the function name of the Call.
 	fn get_call_name(&self) -> &'static str;
 }
 
 /// Gets the metadata for the Call - function name and pallet name.
 pub trait GetCallMetadata {
+	/// Return all module names.
+	fn get_module_names() -> &'static [&'static str];
+	/// Return all call names for this module.
```
```suggestion
	/// Return all function names for the given `module`.
```

bkchr:
Please bump the `impl_version`.

PR #4855, opened 2020-02-07
Refactor and document allocator

pepyakin:
A little bit belated refactoring of the allocator crate.

Overview of changes:

- A concept of `Order` and a separate type is introduced for it. The term Order was inspired by [this](https://www.kernel.org/doc/gorman/html/understand/understand009.html).
- Instead of juglling around with u32 for specifying the pointer to the next item, a separate enum `Link = Ptr(u32) | Null` is introduced. The end of linked list is still ultimately encoded as a `u32::max_value`.
- Prefix is now referred as a Header. A separate enum is introduced as a `Header = Free(Link) | Occupied(Order)`. Apart from that, the in-memory encoding is changed so that it is now it is possible to tell if the header encodes a free or occupied variant, which should aid in detecting memory corruptions.
- Introduced the `Memory` trait to abstract different ways of accessing memory. Might come in handy for the case if we wanted to reuse the allocator code for implementing the substrate runtime interface for a sandboxed module.

This PR (at the moment of writing) is reviewable on commit by commit basis. The changes to API should be backwards compatible.

### DIFF
--- a/primitives/allocator/src/freeing_bump.rs
+++ b/primitives/allocator/src/freeing_bump.rs
@@ -63,8 +63,6 @@ const N: usize = 22;
 const MAX_POSSIBLE_ALLOCATION: u32 = 16777216; // 2^24 bytes
 const MIN_POSSIBLE_ALLOCATION: u32 = 8;
 
-const EMPTY_MARKER: u32 = u32::max_value();
-
 // Each pointer is prefixed with 8 bytes, which identify the list index
 // to which it belongs.
 const PREFIX_SIZE: u32 = 8;
@@ -145,13 +143,44 @@ impl Order {
 	}
 }
 
+/// A marker for denoting the end of the linked list.
+const EMPTY_MARKER: u32 = u32::max_value();
+
+/// A link between headers in the free list.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum Link {
+	/// Null, denotes that there is no next element.
+	Null,
+	/// Link to the next element represented as a pointer to the a header.
+	Ptr(u32),
+}
+
+impl Link {
+	/// Creates a link from raw value.
+	fn from_raw(raw: u64) -> Self {
+		debug_assert_eq!(raw as u32 as u64, raw);
+		if raw as u32 != EMPTY_MARKER {
+			Self::Ptr(raw as u32)
+		} else {
+			Self::Null
+		}
+	}
+
+	/// Converts this link into a raw u64.
+	fn into_raw(self) -> u64 {
+		match self {
+			Self::Null => EMPTY_MARKER as u64,
+			Self::Ptr(ptr) => ptr as u64,
+		}
+	}
+}
 
 /// An implementation of freeing bump allocator.
 ///
 /// Refer to the module-level documentation for further details.
 pub struct FreeingBumpHeapAllocator {
 	bumper: u32,
-	heads: [u32; N],
+	heads: [Link; N],
 	ptr_offset: u32,
 	total_size: u32,
 }
@@ -169,7 +198,7 @@ impl FreeingBumpHeapAllocator {
 
 		FreeingBumpHeapAllocator {
 			bumper: 0,
-			heads: [EMPTY_MARKER; N],
+			heads: [Link::Null; N],
 			ptr_offset,
 			total_size: 0,
 		}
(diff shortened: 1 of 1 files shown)


primitives/allocator/src/freeing_bump.rs
```diff
@@ -93,6 +81,219 @@ macro_rules! trace {
...
+			Ok(Self(order))
+		} else {
+			Err(error("invalid order"))
+		}
+	}
+
+	/// Compute the order by the given size
+	///
+	/// The size is clamped, so that the following holds:
+	//
```
gavofyork:
```suggestion
	///
```

primitives/allocator/src/freeing_bump.rs
```diff
@@ -93,6 +81,219 @@ macro_rules! trace {
...
+	// MIN_POSSIBLE_ALLOCATION <= size <= MAX_POSSIBLE_ALLOCATION
```
gavofyork:
```suggestion
	/// MIN_POSSIBLE_ALLOCATION <= size <= MAX_POSSIBLE_ALLOCATION
```

primitives/allocator/src/freeing_bump.rs
```diff
@@ -93,6 +81,219 @@ macro_rules! trace {
...
+/// A header of an allocation.
+///
+/// The header is encoded in memory as follows.
+///
+/// ## Free header
+///
+/// ```ignore
+/// 64             32                  0
+//  +--------------+-------------------+
+/// |            0 | next element link |
```
gavofyork: doesn't this end up using twice as much memory for the header?

pepyakin: Compared to the former version? No, before the refactoring the header (né `PREFIX`) was also 8 bytes but was only using only the lower 32 bits. In this PR we just start using the upper ones.

That said, I think there are definitely an area of improvement. For example we can leverage the fact that we store either an order, which doesn't require a lot of bits, and a pointer, which is aligned to 8 therefore we can make use of 3 lower bits, thus decreasing the minimal size required for the header to 32 bits without a lot of hassle (we would need to preserve alignment guarantees though).

I didn't do anything alike to keep this PR simple and to make it easier to review.

primitives/allocator/src/freeing_bump.rs
```diff
@@ -93,6 +81,219 @@ macro_rules! trace {
...
+		} else {
+			Err(error("invalid order"))
+		}
+	}
+
+	/// Compute the order by the given size
+	///
+	/// The size is clamped, so that the following holds:
+	///
+	/// MIN_POSSIBLE_ALLOCATION <= size <= MAX_POSSIBLE_ALLOCATION
```
bkchr:
```suggestion
	/// `MIN_POSSIBLE_ALLOCATION <= size <= MAX_POSSIBLE_ALLOCATION`
```

primitives/allocator/src/freeing_bump.rs
```diff
@@ -93,6 +81,219 @@ macro_rules! trace {
...
+		let order = power_of_two_size.trailing_zeros() - MIN_POSSIBLE_ALLOCATION.trailing_zeros();
+
+		Ok(Self(order))
+	}
+
+	/// Returns the corresponding size for this order.
+	///
+	/// Note that it is always a power of two.
+	fn size(&self) -> u32 {
+		1 << MIN_POSSIBLE_ALLOCATION.trailing_zeros() << self.0
```
bkchr:
```suggestion
		MIN_POSSIBLE_ALLOCATION << self.0
```

primitives/allocator/src/freeing_bump.rs
```diff
@@ -93,6 +81,219 @@ macro_rules! trace {
...
+	Null,
+	/// Link to the next element represented as a pointer to the a header.
+	Ptr(u32),
+}
+
+impl Link {
+	/// Creates a link from raw value.
+	fn from_raw(raw: u32) -> Self {
+		if raw != EMPTY_MARKER {
+			Self::Ptr(raw as u32)
```
bkchr:
```suggestion
			Self::Ptr(raw)
```

primitives/allocator/src/freeing_bump.rs
```diff
@@ -122,128 +321,127 @@ impl FreeingBumpHeapAllocator {
...
-		Ok(Pointer::new(self.ptr_offset + ptr + PREFIX_SIZE))
+		Ok(Pointer::new(header_ptr + HEADER_SIZE))
 	}
 
 	/// Deallocates the space which was allocated for a pointer.
 	///
 	/// # Arguments
 	///
-	/// - `mem` - a slice representing the linear memory on which this allocator operates.
+	/// - `mem` - a slice representing the linear memory on which this allocator oper®®ates.
```
bkchr:
```suggestion
	/// - `mem` - a slice representing the linear memory on which this allocator operates.
```

pepyakin: wait wat???

bkchr: IDK :D

Issue #4709, opened 2020-01-22
Introduce habits of prefixing child storage with the hash of module name.


thiolliere:
Similarly to what has been done for storages in the main trie, we should introduce the habits of prefixing the child storage key with the hash of the module name so we can distinguish who this child storage is belonging to.

Currently in substrate only contracts uses child storage and in polkadot crowdfund. My proposal is to make all contract childtries after having the prefix `twox_64("contracts-child-tries")` and all childtries of crowdfund having the prefix `twox_64("crowdfund-child-tries")`.

Note that this change works well with https://github.com/paritytech/substrate/issues/4611
As if a pallet choose to be inside a childtrie then we would have in the pallet in the childtrie `twox_64("PalletPrefix")` and all its childtries prefixed with `twox_64("PalletPrefix-child-tries")`.

I'm not sure about the hash there is seems someone could craft a pallet prefix so that it collides with other as twox64 is weak.

thiolliere:
I'm not sure about how unique_id of child trie should be choosen, this is important for https://github.com/paritytech/substrate/issues/4611 and https://github.com/paritytech/substrate/issues/4707

It seems if we want to avoid having some collision between childtrie ensuring a length of 256 bit is a good decision.

The unique_id used by childtries will prefix the database key (nodes hash) (different from the key in the trie which could be named path).
If you use "abc" as unique_id all the database key used by your child trie will have the prefix "abc".
On batch deletion (which is not implemented but one of the goal of childtries) the client will remove all database key starting with this prefix, thus no other key must start with the same prefix.

So a collision can happen:
* Between two childtries if one unique_id is the start of another one.
  Thus if we use just `twox_64("module_foo")` for the unique_id of the child trie and at the same time keep contracts using blake2_256 of some stuff, if contracts user can generate the contract address they want then they just need to attack the first 64bit and in case the module foo removes its child trie then the contract will be removed as well (So the security of the hash blake2_256 can actually attack by only crafting the first 64bit).
* Between a childtrie and the top trie.
  top trie node (and childtrie node as well) are stored in the database actually not at the key `hash(node)` but actually a combination of this hash of itself and the path to access this node. (I don't know details). But knowing this if somebody successfully create some nodes with a key that start with some unique_id then this nodes will be removed if the childtrie is removed.


Considering those two attacks 
From those two kind of collision it seems to me we should always use a unique_id of length at least 256, and thus attacker must attack 256bit. and thus we can use blake2_256 of untrusted input.

[...]

PR #4857, opened 2020-02-07
Child trie api changes BREAKING

### LINKED ISSUE #2325: Refactor redundancy of TrieIdGenerate
At the moment the TrieId is required to return a `:child_storage:` prefix from the trie id it generated. This is not ideal because a trie_id is stored along with every instance of an account, and that means that we have extra 15 bytes of overhead.

We initially came with this approach because we needed something done quick and we were afraid of adding this prefix manually on every access to the child storage because it is more error-prone. 

The ideal solution would be to perform access to the child storage via some common abstraction that will handle all this subtleties.

cheme:
The child trie api got some questioning recently, so this pr try to make things cleaner and easier to specify.

- Allows child trie to be build from their storage key only. This relies on two condition
  - storage_key of the child trie is a unique id strong enough to avoid hash collision.
  - storage_key of the child trie should only be use once for the child trie (eg if the child trie is deleted, it cannot be created again at the same place).

Contract code falls in this category, polkadot crowdfund to, I think child trie module do also (cc @thiolliere).

- Changes the way `StorageKey` is define to never include ':child_storage:default:' and include it only when needed (when reading or updating parent root).

- Separates child trie host function into their own runtime api (so future child trie of different kind could use their own).

So this breaks rpc, with this pr rpc would use storage_key without its prefix and no more child_info and child type fields. cc @Stefie
This PR also do break existing runtime by removing the child info and  child type too.

This also changes `TrieIdGenerator` of contract pallet to stop including the child prefixes for the key, the changes needed are in the pr cc @pepyakin  Closes #2325.

So for other kind of child trie new rpc api and runtime host function will be needed.

polkadot companion: https://github.com/paritytech/polkadot/pull/950


primitives/externalities/src/lib.rs
```diff
@@ -47,7 +47,7 @@ pub trait Externalities: ExtensionStore {
 	/// Returns an `Option` that holds the SCALE encoded hash.
 	fn child_storage_hash(
 		&self,
-		storage_key: ChildStorageKey,
+		storage_key: &[u8],
```
Stefie: Is there a reason why you're using `&[u8]` here instead of `StorageKey` from `use sp_core::storage::{StorageKey};`? Can the `StorageKey` be of any/ variable length here?

cheme: Yest storage can can be of any/variable length, using a slice here is possible as the storage key does not change, I could have use '&ChildStorageKey' instead, but that is consistent with the way `key` parameter is used.

Note that I am in the process of doing another pass of refactoring, simplifying a lot more, I hope it will be finished this evening :)

Stefie:
> So this breaks rpc, with this pr rpc would use storage_key without its prefix and default to the child_type 0 with an empty child_info description. cc @Stefie

cheme:
This PR is now ready for review.
It does not contain logic change except for the prefixing of the storage key that is now done before writing and reading root in parent trie only.
The main change is mostly reverting back to some simple rpc and simplier runtime host function.

client/rpc/src/state/state_full.rs
```diff
@@ -308,60 +308,57 @@ impl<B, E, Block, RA> StateBackend<B, E, Block, RA> for FullState<B, E, Block, R
 	fn child_storage_keys(
 		&self,
 		block: Option<Block::Hash>,
-		child_storage_key: StorageKey,
```
Stefie: The naming seems to not be consistent between [`state_light.rs`](https://github.com/paritytech/substrate/pull/4857/files#diff-fcaa37ac7f8e7bb8e37ca69b7924809aR252), `state_full.rs` and [`mod.rs`](https://github.com/paritytech/substrate/pull/4857/files#diff-3d66ab185ffb08d24bb836f41283bc86R306).
You changed `child_storage_key` to `storage_key` here, but kept `child_storage_key` in the other two files (in all child_storage_* functions).

primitives/trie/src/lib.rs
```diff
@@ -211,15 +211,13 @@ pub fn read_trie_value_with<
 
 /// Determine the default child trie root.
```
thiolliere: default here confuse with default child trie type, maybe the doc should specify that this is for empty child tries, or this functions could be renamed empty_child_trie_root

primitives/storage/src/lib.rs
```diff
@@ -126,180 +129,200 @@ pub mod well_known_keys {
...
+/// top trie being a child trie with no keyspace and no storage key.
+/// Its keyspace is the variable (unprefixed) part of its storage key.
+/// It shares its trie nodes backend storage with every other
+/// child trie, so its storage key needs to be a unique id
+/// that will be use only once.
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "std", derive(PartialEq, Eq, Hash, PartialOrd, Ord))]
-pub struct OwnedChildTrie {
-	/// See `ChildTrie` reference field documentation.
+pub struct ChildTrie {
```
thiolliere: I actually kind of have hard time understanding what is a non default child.
If a non default childtrie is a childtrie then this name is confusing.

primitives/storage/src/lib.rs
```diff
@@ -62,7 +62,10 @@ pub struct StorageChild {
 pub struct Storage {
 	/// Top trie storage data.
 	pub top: StorageMap,
-	/// Children trie storage data by storage key.
+	/// Children trie storage data.
+	/// The key does not including prefix, for the `default`
+	/// trie kind, so this is exclusively for the `ChildType::ParentKeyId`
+	/// tries.
 	pub children: std::collections::HashMap<Vec<u8>, StorageChild>,
```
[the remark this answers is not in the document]
cheme: Yes, but I also have a mix feeling calling it 'default', if at some point the default change.
Considering it can be rename if the default change, I would also think it can be rename when a second child type appears. But that is probably not a good idea as it will probably be forgotten, so 'children_default' or 'default' could be fine.
I like 'default' as it is shorter but yes, default_children is probably better.

client/api/src/light.rs
```diff
@@ -82,11 +82,6 @@ pub struct RemoteReadChildRequest<Header: HeaderT> {
 	pub header: Header,
 	/// Storage key for child.
 	pub storage_key: Vec<u8>,
```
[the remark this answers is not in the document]
cheme: I think the first option is good (I obviously prefer second option as I wrote it first but the point of this PR is to avoid the general api and switch to the specific one), will need an update on polkadot js too.

client/chain-spec/src/chain_spec.rs
```diff
@@ -76,13 +76,13 @@ impl<G: RuntimeGenesis, E> BuildStorage for ChainSpec<G, E> {
 			Genesis::Runtime(gc) => gc.build_storage(),
 			Genesis::Raw(RawGenesis { top: map, children: children_map }) => Ok(Storage {
 				top: map.into_iter().map(|(k, v)| (k.0, v.0)).collect(),
-				children: children_map.into_iter().map(|(sk, child_content)| {
+				children: children_map.into_iter().map(|(storage_key, child_content)| {
```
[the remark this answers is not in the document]
cheme: I agree on all your point, my first idea was simply new type of child trie then new field in genesis raw for the new child trie.
So I guess renaming 'children' to 'children_default' would be what makes more sense (it does not break kusama genesis because the json do not name the field (it is using an old format that is accepted (array as a map that was something I noticed last time I change the json)).
But adding new field is certainly breaking as the command for generating genesis does explicitely name the field in the json (I will need to see if there is a way to force the array syntax).

client/network/src/protocol/schema/light.v1.proto
```diff
@@ -73,13 +73,8 @@ message RemoteReadChildRequest {
...
 	// Child Storage key.
 	bytes storage_key = 3;
-	// Child trie source information.
-	bytes child_info = 4;
-	/// Child type, its required to resolve `child_info`
-	/// content and choose child implementation.
-	uint32 child_type = 5;
 	// Storage keys.
-	repeated bytes keys = 6;
+	repeated bytes keys = 4;
```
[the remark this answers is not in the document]
cheme: I think it is only breaking light client, so this requires a client update, but there is no child trie in use in kusama so I chose to change it.
On the other hand it is not a problem to keep 6 as index, I will change that.

primitives/io/src/lib.rs
```diff
@@ -89,30 +76,6 @@ pub trait Storage {
...
-		child_definition: &[u8],
-		child_type: u32,
-		key: &[u8],
-	) -> Option<Vec<u8>> {
-		let storage_key = child_storage_key_or_panic(child_storage_key);
-		let child_info = ChildInfo::resolve_child_info(child_type, child_definition)
-			.expect("Invalid child definition");
-		self.child_storage(storage_key, child_info, key).map(|s| s.to_vec())
-	}
-
```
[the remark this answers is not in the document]
cheme: Kusama has the crowdfund module using child trie, but it has not been linked to the runtime so it is ok.
But the Edgeware case is more preoccupying to me, I will restore the old IO as deprecated for now.

primitives/state-machine/src/basic.rs
```diff
@@ -261,20 +254,19 @@ impl Externalities for BasicExternalities {
 
 	fn storage_root(&mut self) -> Vec<u8> {
 		let mut top = self.inner.top.clone();
-		let keys: Vec<_> = self.inner.children.keys().map(|k| k.to_vec()).collect();
+		let keys: Vec<_> = self.inner.children.iter().map(|(_k, v)| {
+			(v.child_info.prefixed_storage_key(), v.child_info.clone())
```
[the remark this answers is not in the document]
cheme: I will rename the variable, this is used to update top trie, so we need to include the prefix.

primitives/state-machine/src/changes_trie/build.rs
```diff
@@ -138,7 +138,7 @@ fn prepare_extrinsics_input_inner<'a, B, H, Number>(
 		Number: BlockNumber,
 {
 	let (committed, prospective, child_info) = if let Some(sk) = storage_key.as_ref() {
-		let child_info = changes.child_info(sk).cloned();
+		let child_info = changes.default_child_info(sk).cloned();
```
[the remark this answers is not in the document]
cheme: The idea here was to rename the method to make it clear that we manage child info for a default trie, the point is that child trie do not have to relate the type of the trie (it only manage key values (and put them in a trie but the child trie does not have to relate to the source trie for the delta of kv)).
Note that it manipulates prefixed keys after latest commit, which was needed.

primitives/storage/src/lib.rs
cheme: I am renaming it, I with this one was private, but I would need to put the parent enum in a wrapping type, not sure if worth it.

primitives/io/src/lib.rs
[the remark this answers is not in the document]
cheme: While restoring the old api, I realize it was a broken design: using a child type we cannot add new child type later as it will break the past behavior, so even if I do not like all those change, this is good one (runtime level, I am really bugged by the rpc still).

cheme:
Renaming rpc was a bit ugly, so in the end I bring back the child_type parameter to the rpc.

cheme:
I just did realize that even if the deprecated runtime is kept, for chain using contract, an data migration is needed to remove the prefixes from the contract ids (because of solving #2325).

client/api/src/light.rs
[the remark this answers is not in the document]
thiolliere: hmm but now you have storage_key and child_type, shouldn't it be child_info and child_type ? I mean if child_type is not `ParentKeyId` then there might be information missing to be able to answer the request, no ?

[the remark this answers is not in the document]
cheme: It comes to the idea that there is no way information would not be accessible in state.
Here child type is only a way to avoid information redundancy in the storage key and not a way to have additional information in the api (as would be child_info).
So with this api child_info, when needed, should be resolve from the system (even if it can not be easy to implement). So this makes the previous default child type impossible to implement, and force a managed unique id approach in this case (from a specification perspective it was not really simple to legitimate child_info, only argument was that managed unique id did not fit the timeline but considering the current definition of child trie can be only the parentkeyid one, this argument does not really hold).

There is still a case where I find that it is not really good, that would be for child trie that would not be at a location dependant on their type: then the child type would be redundant and would probably use a single value value for this kind of trie and become redundant. But accepting child trie at random location is probably a big redesign with new design concern and a dedicated api could make sense then (and I do not think there is need for such trie).

client/network/src/protocol/schema/light.v1.proto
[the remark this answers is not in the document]
thiolliere: ok, both way are fine to me then

primitives/io/src/lib.rs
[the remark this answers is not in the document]
cheme: Actually previous version was correct, we cannot break pas behavior as there was no child trie of the new type existing, so api with child type is right as is api where storage key type is defined by prefix. I will probably switch everything back to prefixed storage key.

primitives/storage/src/lib.rs
```diff
@@ -126,180 +129,201 @@ pub mod well_known_keys {
...
-	data: &'a[u8],
-}
-
-/// Owned version of default child trie `ChildTrie`.
+/// It uses the same default implementation as the top trie,
+/// top trie being a child trie with no keyspace and no storage key.
+/// Its keyspace is the variable (unprefixed) part of its storage key.
+/// It shares its trie nodes backend storage with every other
+/// child trie, so its storage key needs to be a unique id
+/// that will be use only once.
```
thiolliere: > so its storage key needs to be a unique id that will be use only once.

I'm not sure uniqueness is enough, if one child trie unique id is "abc" and another is "abcd" then both are unique, however removing "abc" will "abcd" when batch deletion is implemented if I'm correct.

I try to discuss it here https://github.com/paritytech/substrate/issues/4709

primitives/storage/src/lib.rs
```diff
@@ -126,180 +129,201 @@ pub mod well_known_keys {
...
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "std", derive(PartialEq, Eq, Hash, PartialOrd, Ord))]
-pub struct OwnedChildTrie {
-	/// See `ChildTrie` reference field documentation.
+pub struct ChildTrieParentKeyId {
+	/// Data is the full prefixed storage key.
```
thiolliere: If I read correctly throughout this module data is considered without prefix.
```suggestion
	/// Data is the storage key without prefix.
```

cheme:
Ok, I did switch back to only storage key for rpc, but with child type prefix.
The runtime is still using a set of function for default child type only (so using storage key without prefix).

cheme:
I started implemented a different api in https://github.com/cheme/substrate/commit/9ad968051ccc548c30240d4d70d9df0377974ed8 , I am stopping it at this point as it is more work than I believed and it was just to test (it would only start looking nice only after changing also backend trait and using a slab in overlay_change).

client/rpc-api/src/state/mod.rs
```diff
@@ -77,8 +77,6 @@ pub trait StateApi<Hash> {
 	fn child_storage_keys(
 		&self,
 		child_storage_key: StorageKey,
-		child_info: StorageKey,
```
cheme: https://github.com/paritytech/substrate/pull/5280#discussion_r394284782

cheme:
Latest commit restore existing rpc interface and add a new child state rpc api.
I did not change 'rpc_methods' version as it is only new methods. I would have like to remove the restored method in a version 2 of the rpc api, but this is out of scope of this PR as it would probably requires significative change in json rpc, I am thinking some new derive attribute `deprecated(2)` and `since(2)` (or something hardcoded changing version is really needed in this case).
cc @tomusdrw , @jacogr

jacogr:
@cheme You can actually remove the old methods completly. No need to support them and keep them around. The middleware (e.. APIs) can detect what the node has and use those. As long as the rename is there, the old can go the way of the dodo.

cheme:
Ok, so it is breaking, but we can detect it. Seems better to me, I will remove them, thanks.

test-utils/client/src/lib.rs
```diff
@@ -129,17 +129,17 @@ impl<Block: BlockT, Executor, Backend, G: GenesisInit> TestClientBuilder<Block,
...
 		mut self,
+		child_info: &ChildInfo,
 		key: impl AsRef<[u8]>,
-		child_key: impl AsRef<[u8]>,
-		child_info: ChildInfo,
 		value: impl AsRef<[u8]>,
 	) -> Self {
-		let entry = self.child_storage_extension.entry(key.as_ref().to_vec())
+		let storage_key = child_info.storage_key();
+		let entry = self.child_storage_extension.entry(storage_key.to_vec())
```
thiolliere: would be good to add in the doc of child_storage_extension field that it is supposed to be the key without prefix.

cheme:
> The fact that both ChildInfo::storage_key and ChildInfo::prefixed_storage_key return the same type (byte slice kind of) makes it difficult to follow

True, I will try to do something about it tomorrow.

cheme:
@thiolliere , I added a PrefixedStorageKey type in https://github.com/paritytech/substrate/pull/4857/commits/1b2a30cd9938ee776ebe441279374d88e8e81a27 , let me know if you think it is better.

primitives/storage/src/lib.rs
```diff
@@ -32,6 +34,67 @@ pub struct StorageKey(
...
+	fn deref_mut(&mut self) -> &mut Vec<u8> {
+		&mut self.0
+	}
+}
+
+impl PrefixedStorageKey {
+	/// Create a prefixed storage key from its byte array
+	/// representation.
+	/// Returns `None` on unknown prefix.
+	pub fn new(inner: Vec<u8>) -> Option<Self> {
```
thiolliere: I don't know how much it worth checking that it is valid here even though a PrefixedStorageKey can be invalid.

client/network/src/protocol.rs
```diff
@@ -312,14 +312,14 @@ impl<'a, B: BlockT> LightDispatchNetwork<B> for LightDispatchIn<'a> {
 		who: &PeerId,
 		id: RequestId,
 		block: <B as BlockT>::Hash,
-		storage_key: Vec<u8>,
+		storage_key: PrefixedStorageKey,
 		keys: Vec<Vec<u8>>,
 	) {
 		let message: Message<B> = message::generic::Message::RemoteReadChildRequest(
 			message::RemoteReadChildRequest {
```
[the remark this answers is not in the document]
cheme: In fact it was just to  avoid defining a new type in the protobuf schema, but it is the content of with the prefix that is in use.

primitives/storage/src/lib.rs
```diff
@@ -264,7 +309,7 @@ impl ChildType {
 
 	/// Transform a prefixed key into a tuple of the child type
 	/// and the unprefixed representation of the key.
-	pub fn from_prefixed_key<'a>(storage_key: &'a [u8]) -> Option<(Self, &'a [u8])> {
+	pub fn from_prefixed_key<'a>(storage_key: &'a PrefixedStorageKey) -> Option<(Self, &'a [u8])> {
```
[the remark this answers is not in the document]
cheme: Yes that is mainly for receiving from rpc.

client/rpc-api/src/child_state/mod.rs
```diff
@@ -32,7 +32,7 @@ pub trait ChildStateApi<Hash> {
 	#[rpc(name = "childstate_getKeys")]
 	fn storage_keys(
 		&self,
-		child_storage_key: StorageKey,
+		child_storage_key: PrefixedStorageKey,
```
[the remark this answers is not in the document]
cheme: Yes the type is meant to be strictly informative, the guaranty that the input is valid is meant to be only for ChildInfo.

primitives/storage/src/lib.rs
cheme: I think you are right, I am switching that.

primitives/storage/src/lib.rs
```diff
@@ -32,6 +34,51 @@ pub struct StorageKey(
 	pub Vec<u8>,
 );
 
+/// Storage key of a child trie, it contains the prefix to the key.
```
thiolliere:
```suggestion
/// Storage key of a child trie, it contains the prefix to the key.
/// There is no guarantee that it holds a valid child trie storage key.
```

thiolliere:
OK looks good to me, I can do a final review but it correct to me.

cheme:
> In the future, to make it more reviewable, maybe we should move out of the PR everything that can be, like renames and some changes.

Yes, I will keep that in mind :+1: , I will try to merge master and update the polkadot branch (and open pr and finally try this new CI feature) later today.

.maintain/gitlab/check_polkadot_companion_build.sh
```diff
@@ -57,7 +57,7 @@ then
   # get the last reference to a pr in polkadot
   pr_data="$(curl -sSL -H "${github_header}" -s ${github_api_substrate_pull_url}/${CI_COMMIT_REF_NAME})"
   pr_ref="$(echo $pr_data | grep -Po '"ref"\s*:\s*"\K(?!master)[^"]*')"
-  pr_body="$(echo $pr_data | sed -n -r 's/^[[:space:]]+"body": (".*")[^"]+$/\1/p')"
+  pr_body="$(echo "$pr_data" | sed -n -r 's/^[[:space:]]+"body": (".*")[^"]+$/\1/p')"
```
cheme, line 60:
```suggestion
  pr_body="$(echo "${pr_data}" | sed -n -r 's/^[[:space:]]+"body": (".*")[^"]+$/\1/p')"
```

gavofyork:
Companion not building, @cheme

gnunicorn:
merge master should fix the breaking CI test

primitives/io/src/lib.rs
```diff
@@ -128,26 +99,81 @@ pub trait Storage {
...
+		child_type: u32,
+		key: &[u8],
+	) -> Option<Vec<u8>> {
+		deprecated_storage_key_prefix_check(storage_key);
+		let child_info = ChildInfo::resolve_child_info(child_type, child_definition)
+			.expect("Invalid child definition");
+		self.child_storage(&child_info, key).map(|s| s.to_vec())
+	}
+
+	/// Deprecated, please use dedicated runtime apis.
```
NikVolf, line 163:
```suggestion
	/// Read child key.
	/// 
	/// Deprecated, please use dedicated runtime apis (`sp_io::default_child_storage::get`).
```

primitives/io/src/lib.rs
```diff
@@ -157,159 +183,219 @@ pub trait Storage {
...
 
-	/// Set `key` to `value` in the storage.
-	fn set(&mut self, key: &[u8], value: &[u8]) {
-		self.set_storage(key.to_vec(), value.to_vec());
-	}
-
-	/// Set `key` to `value` in the child storage denoted by `child_storage_key`.
-	///
-	/// See `child_get` for common child api parameters.
+	/// Deprecated, please use dedicated runtime apis.
```
NikVolf, line 186: Same, reference to the replacement for the deprecated api needed

primitives/io/src/lib.rs
```diff
@@ -157,159 +183,219 @@ pub trait Storage {
...
+		self.next_child_storage_key(&child_info, key)
+	}
+
+}
+
+
+/// Interface for accessing the child storage for default child trie,
+/// from within the runtime.
+#[runtime_interface]
+pub trait DefaultChildStorage {
```
NikVolf, line 291: Can we call it `ChildStorge` and not `DefaultChildStorage` ?
Is another type of child storage planned in `sp-io` ?

primitives/io/src/lib.rs
```diff
@@ -157,159 +183,219 @@ pub trait Storage {
...
+		let child_info = ChildInfo::resolve_child_info(child_type, child_definition)
+			.expect("Invalid child definition");
+		self.next_child_storage_key(&child_info, key)
+	}
+
+}
+
+
+/// Interface for accessing the child storage for default child trie,
+/// from within the runtime.
```
NikVolf, line 289:
```suggestion
/// Child storage interface.
/// 
/// Interface for accessing the child storage for default child trie,
/// from within the runtime.
```

And in general, good doc comments format which then renders nicely in html is

```rust
/// Header
/// 
/// Detailed description
```

(plenty of this requires some love in `sp-io`)

primitives/io/src/lib.rs
```diff
@@ -128,26 +99,81 @@ pub trait Storage {
...
+	fn changes_root(&mut self, parent_hash: &[u8]) -> Option<Vec<u8>> {
+		self.storage_changes_root(parent_hash)
+			.expect("Invalid `parent_hash` given to `changes_root`.")
+	}
+
+	/// Get the next key in storage after the given one in lexicographic order.
+	fn next_key(&mut self, key: &[u8]) -> Option<Vec<u8>> {
+		self.next_storage_key(&key)
+	}
+
```
NikVolf, line 147:
```suggestion
```

primitives/io/src/lib.rs
```diff
@@ -157,159 +183,219 @@ pub trait Storage {
...
+		child_definition: &[u8],
+		child_type: u32,
+		key: &[u8],
+	) -> Option<Vec<u8>> {
+		deprecated_storage_key_prefix_check(storage_key);
+		let child_info = ChildInfo::resolve_child_info(child_type, child_definition)
+			.expect("Invalid child definition");
+		self.next_child_storage_key(&child_info, key)
+	}
+
```
NikVolf, line 284:
```suggestion
```

primitives/io/src/lib.rs
```diff
@@ -128,26 +99,81 @@ pub trait Storage {
...
+	/// Get the next key in storage after the given one in lexicographic order.
+	fn next_key(&mut self, key: &[u8]) -> Option<Vec<u8>> {
+		self.next_storage_key(&key)
+	}
+
+
+	/// Deprecated, please use dedicated runtime apis.
+	fn child_get(
```
bkchr, line 150: Is this code used somewhere? I don't think so? 

So please remove it, the same goes for all deprecated unused functions.

primitives/io/src/lib.rs
cheme: Previous api was using a child type as parameter, but I realize that adding a new child type breaks expectation for the method (it changes behavior and would require a new version of the method, so at the end it seems easier to have an api per child type).

> Is another another type of child storage planned in sp-io ?

I hope so, that was core component of initial design (the type was initially fetch from storage key and for this type of child storage the prefixed version of the storage key still include the type).

primitives/io/src/lib.rs
cheme: I did remove them first, but restore them back to retain compatibility with edgeware chain, but I would also really prefer to remove it (and the associated code).
Think we can break that?

primitives/io/src/lib.rs
cheme: That's good advice, thanks.

primitives/io/src/lib.rs
bkchr: Yeah, if we not do it now, we need to keep this code forever.

arkpar:
Changes in `client/network/src/protocol/message.rs` probably require protocol version bump and some code to support backward compatibility

arkpar:
This might be a good time to rename "child" to "subtree" or "substorage". Please vote if you think this is a good idea.

cheme:
> This might be a good time to rename "child" to "subtree" or "substorage". Please vote if you think this is a good idea.

My preference goes toward «substorage» (do not have to be a tree, just some storage with a proof system), I could also propose «substate». 
This would be a big renaming but I think it will be worth it.

primitives/io/src/lib.rs
cheme: I did remove the code (it messed up my last master merge and it is in latest master merge commit).

cheme:
> Changes in client/network/src/protocol/message.rs probably require protocol version bump and some code to support backward compatibility

https://github.com/paritytech/substrate/pull/4857/commits/619b454c3ccf7b5d35bfaf8eea3df00084ee1272 is trying to do that, note that it assumes input data is of the right size (I found no other way to remove possible false positive when decoding with the latest encoding without changing to much the way things are encoded).
@arkpar, I do not have code compatibility on the light client handler either, reasoning was that the child trie api should not be use at the moment (one of the latest change remove the sp-io compatibility there). I observe protobuf version is still 1 so it may not be needed (not sure how much support for light client is currently needed).

arkpar:
That's a good point. If it is not used there's no need to bother with compatibility indeed. So `619b454` could be reverted I guess.  If not, you could implement custom `Decode` that would try to decode as previous version and then as new version if it fails. Like here: https://github.com/paritytech/substrate/blob/master/client/network/src/protocol/message.rs#L349

cheme:
Hum custom decoder looks like a better idea indeed (or at least put everything in RemoteReadChildRequest), but I am reverting it right now.

gavofyork:
Needs resolving @cheme 

Is there any reason we shouldn't merge this before 2.0?

cheme:
> Is there any reason we shouldn't merge this before 2.0?

It is breaking some compatibility so I would think it is better if merge before 2.0. I am resolving it.

gnunicorn:
@cheme this needs one more conflict resolution and can be merged.

PR #4858, opened 2020-02-07
Change prefixed trie iterator to be limited to prefix. BREAKING

cheme:
Currently the iteration based on prefix stops at the first key that do not contain a prefix.
This PR change this behavior to stop the iteration at the last key start with the prefix.

This will change all proofs that uses the iteration by prefix, they will contain a few less node (generally one, can be a bit more depending on the depth of next key value that did not start with the prefix).
Therefore this can breaks 'call' light client proof (but it is not use in polkadot or substrate test runtime).
It can also break cumulus witness for runtime using prefixed key iteration (cc @bkchr), but I do not think there is any at this point.

Generally this PR is not a must have, and keeping old iteration could be fine (not sure which version is easier to specify in polkadot specs).

Technical details and tests are at https://github.com/paritytech/trie/pull/39


bkchr:
We currently have prefixed iteration. However by breaking you mean if we had something released?

cheme:
Yes breaking for something released where proof of call to this prefixed iteration is in use.

I did not see any light client rpc using it directly so it will be only the proof for 'call' of extrinsic, but I did not see use of this prefixed iter for something else than ext clear_prefix and clear_child_prefix which seem used only in kill_prefix, used in double map remove prefix, used in imonline event on_before_session_ending, evm module, staking new_session (non exposed I think) through slashing clear_era_metadata).
So the call proof for light client should only be broken for the evm module (in case such proof is used).
Regarding cumulus this can obviously be in the block transition, so this will break all existing cumulus witness using imonline, evm or staking (friday I overlooked this).

About the testing, there does not seems to be existing tests for this, on the other hand I did write new ones in https://github.com/paritytech/trie/pull/39 , I could probably bring them to substrate to, but I am not sure if it is needed.

cheme:
I missed a really important call: polkadot `set_heads`, this could be in theory used in a light client 'call' proof, so this is breaking in the sense that we need a client update.
I am switching the tags, I doubt it is a problem in practice, but it is actually breaking over light client call proof of polkadot for `set_heads` extrinsics.

bkchr:
Did you tried to sync kusama?

cheme:
No, this shouldn't break consensus. But it is a small effort to do, I cannot access proper machine today, but could do tomorrow.

bkchr:
> No, this shouldn't break consensus. But it is a small effort to do, I cannot access proper machine today, but could do tomorrow.

Yeah that is okay! Just report back :)

cheme:
Reporting back, did synch kusuma without issue, as expected :)

bkchr:
Ty!

PR #4860, opened 2020-02-07
offchain: Upgrade hyper to 0.13, which uses tokio 0.2

### LINKED ISSUE #4647: offchain::http_response_wait() fatal executor error: SpawnError { is_shutdown: true }
When I use `offchain` methods:

```
let id: HttpRequestId = sp_io::offchain::http_request_start(method, uri, &[0]).unwrap();
let deadline = sp_io::offchain::timestamp().add(Duration::from_millis(100_000));

match sp_io::offchain::http_response_wait(&[id], Some(deadline))[0] {
	HttpRequestStatus::Finished(200) => (),
	_ => {
		return Err("Request failed");
	}
}
```
it will  return `HttpRequestStatus::IoError`,  runtime logs are:

```
main WARN hyper::common::exec  executor error: SpawnError { is_shutdown: true }
main WARN hyper::client  error spawning critical client task: executor failed to spawn task: tokio::spawn failed (is a tokio runtime running this future?)
```

Substrate Version: commit 39db16e213d92db4f0bffc9dae715ad9c59d2a0c

Is there something wrong with `offchain http_request` or just not the right methods I used ?

Thanks.

Xanewok:
Fixes the regressed offchain HTTP API by upgrading hyper to use tokio 0.2.

While an integration test for offchain HTTP API [exists](https://github.com/paritytech/substrate/blob/b02a97129286a8453ef78f58963b72c7ae47aef6/client/executor/src/integration_tests/mod.rs#L469), it [uses mocks in `TestOffchainExt`](https://github.com/paritytech/substrate/blob/6e0a4877afbe2309e106b84333bb70200531b604/primitives/core/src/offchain/testing.rs#L199-L304), so I decided against duplicating an another harness with actual HTTP.

I imagine a better approach would be to set up most of the runtime and test the HTTP from offchain workers directly but didn't want to block the fix here, let me know if I should work on setting that up as well.

Tested manually on https://github.com/Xanewok/substrate/tree/offchain-worker-example with `cargo run --bin node-template -- --dev`.

Fixes #4647.

### DIFF
--- a/client/offchain/Cargo.toml
+++ b/client/offchain/Cargo.toml
@@ -11,7 +11,6 @@ bytes = "0.5"
 sc-client-api = { version = "2.0.0", path = "../api" }
 sp-api = { version = "2.0.0", path = "../../primitives/api" }
 fnv = "1.0.6"
-futures01 = { package = "futures", version = "0.1" }
 futures = "0.3.1"
 futures-timer = "3.0.1"
 log = "0.4.8"
@@ -34,7 +33,7 @@ hyper-rustls = "0.19"
 sc-client-db = { version = "0.8", default-features = true, path = "../db/" }
 env_logger = "0.7.0"
 substrate-test-runtime-client = { version = "2.0.0", path = "../../test-utils/runtime/client" }
-tokio = "0.1.22"
+tokio = "0.2"
 sc-transaction-pool = { version = "2.0.0", path = "../../client/transaction-pool" }
 sp-transaction-pool = { version = "2.0.0", path = "../../primitives/transaction-pool" }
 
--- a/client/offchain/Cargo.toml
+++ b/client/offchain/Cargo.toml
@@ -7,7 +7,7 @@ authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
 [dependencies]
-bytes = "0.4.12"
+bytes = "0.5"
 sc-client-api = { version = "2.0.0", path = "../api" }
 sp-api = { version = "2.0.0", path = "../../primitives/api" }
 fnv = "1.0.6"
@@ -27,8 +27,8 @@ sc-network = { version = "0.8", path = "../network" }
 sc-keystore = { version = "2.0.0", path = "../keystore" }
 
 [target.'cfg(not(target_os = "unknown"))'.dependencies]
-hyper = "0.12.35"
-hyper-rustls = "0.17.1"
+hyper = "0.13.2"
+hyper-rustls = "0.19"
 
 [dev-dependencies]
 sc-client-db = { version = "0.8", default-features = true, path = "../db/" }
--- a/client/offchain/src/api/http.rs
+++ b/client/offchain/src/api/http.rs
@@ -688,31 +688,36 @@ impl fmt::Debug for HttpWorkerRequest {
 
 #[cfg(test)]
 mod tests {
+	use core::convert::Infallible;
 	use crate::api::timestamp;
 	use super::http;
-	use futures::prelude::*;
-	use futures01::Future as _;
 	use sp_core::offchain::{HttpError, HttpRequestId, HttpRequestStatus, Duration};
 
 	// Returns an `HttpApi` whose worker is ran in the background, and a `SocketAddr` to an HTTP
 	// server that runs in the background as well.
 	macro_rules! build_api_server {
 		() => {{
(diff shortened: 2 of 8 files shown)


Xanewok:
Forgot to adapt the test code as well, let me fix it up real quick.

bkchr:
I think it would be nice to have a real test. I think if we use the test runtime and create some server for the test. You would connect to this server from the offchain worker.

client/offchain/src/api/http.rs
```diff
@@ -688,31 +688,38 @@ impl fmt::Debug for HttpWorkerRequest {
...
 	use super::http;
-	use futures::prelude::*;
-	use futures01::Future as _;
 	use sp_core::offchain::{HttpError, HttpRequestId, HttpRequestStatus, Duration};
 
 	// Returns an `HttpApi` whose worker is ran in the background, and a `SocketAddr` to an HTTP
 	// server that runs in the background as well.
 	macro_rules! build_api_server {
 		() => {{
+			macro_rules! tokio_run {
```
tomusdrw: Not sure if the number of repetitions is worth a macro. We could either inline it or just use a closure?

Xanewok: We can't use closures since the two futures differ in concrete types (this would require a generic closure, it's what I tried originally) - I can inline it or use a helper function :+1:

Xanewok:
Wasm build fails on `cargo build --target=wasm32-unknown-unknown -p sc-consensus-aura` due to pulling in optional `net2` dependency of hyper but I'm opting out of default features... Hopefully it's not feature conflating Cargo bug; will investigate.

client/offchain/src/api/http.rs
Xanewok: Used a generic helper function.

Xanewok:
If I were to guess, I'd say it's the new `hyper-rustls` here (by default) including `runtime` feature for `hyper` but it seems that the 0.13 version is also pulled in wasm in https://github.com/paritytech/substrate/blob/a9680b40469c8180e5b6b083c01dc04f69e4037c/utils/grafana-data-source/Cargo.toml#L11
hence we unfortunately conflate features and pull in the runtime to wasm accidentally.

Cargo features :angry:

gavofyork:
web-wasm failing?

client/offchain/Cargo.toml
```diff
@@ -27,14 +26,14 @@ sc-network = { version = "0.8", path = "../network" }
 sc-keystore = { version = "2.0.0", path = "../keystore" }
 
 [target.'cfg(not(target_os = "unknown"))'.dependencies]
-hyper = "0.12.35"
-hyper-rustls = "0.17.1"
+hyper = { version = "0.13.1", default-features = false, features = ["stream"] }
```
tomaka: I think you're missing features here anyway, as `hyper` needs the `tcp` or `runtime` feature in order to use tokio.

client/offchain/Cargo.toml
```diff
@@ -27,14 +26,14 @@ sc-network = { version = "0.8", path = "../network" }
...
+hyper-rustls = "0.19"
 
 [dev-dependencies]
 sc-client-db = { version = "0.8", default-features = true, path = "../db/" }
 env_logger = "0.7.0"
 substrate-test-runtime-client = { version = "2.0.0", path = "../../test-utils/runtime/client" }
-tokio = "0.1.22"
+tokio = "0.2"
```
tomaka: Have you tried putting that in `target.cfg(not(unknown)).dev-dependencies`?

client/offchain/Cargo.toml
Xanewok: You're right, I originally copied it from `grafana-data-source` but forgot to change it, meanwhile `hyper-rustls` enabled the `hyper/runtime` via https://github.com/ctz/hyper-rustls/blob/f51987a477af6d104783dee3b354e3db244f61a4/Cargo.toml#L29.

I'll fix it to pull all default features here for clarity.

utils/grafana-data-source/Cargo.toml
```diff
@@ -20,4 +18,6 @@ futures-timer = "3.0.1"
 derive_more = "0.99"
 
 [target.'cfg(not(target_os = "unknown"))'.dependencies]
+hyper = { version = "0.13.1", default-features = false, features = ["stream"] }
```
tomaka: Nit: many people just don't care, but I try to enforce alphabetical ordering.

PR #4863, opened 2020-02-09
Alter usage collection a bit

NikVolf:
Should properly return some usage info even after a short time has passed since service start

### DIFF
--- a/client/db/src/lib.rs
+++ b/client/db/src/lib.rs
@@ -699,7 +699,7 @@ impl<Block: BlockT> sp_state_machine::Storage<HasherFor<Block>> for DbGenesisSto
 /// Used as inner structure under lock in `FrozenForDuration`.
 struct Frozen<T: Clone> {
 	at: std::time::Instant,
-	value: T,
+	value: Option<T>,
 }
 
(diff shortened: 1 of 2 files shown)


client/db/src/lib.rs
```diff
@@ -713,22 +713,24 @@ pub(crate) struct FrozenForDuration<T: Clone> {
...
 			duration,
-			value: Frozen { at: std::time::Instant::now(), value: initial }.into(),
+			value: Frozen { at: std::time::Instant::now(), value: None }.into(),
 		}
 	}
 
 	fn take_or_else<F>(&self, f: F) -> T where F: FnOnce() -> T {
-		if self.value.read().at.elapsed() > self.duration {
-			let mut write_lock = self.value.write();
+		let read_lock = self.value.upgradable_read();
```
bkchr: Upgradeable read locks are exclusive as well. So there could exists readers at the same time, but no concurrent other upgradable read lock. So, please just use a write lock directly.

NikVolf: Right, I'll just replace with mutex then

PR #4864, opened 2020-02-09
Fix vesting logic

gavofyork:
Previously, the `vesting_balance` was used to deduce if a vest was finished. This is problematic since it will be clamped at `free_balance` and thus if `free_balance` happens to be temporarily zero then (e.g. because the vesting schedule is placed on an account before the funds are transferred as in `claims.rs`) then the schedule will be insta-dropped.

This checks with the underlying vesting schedule, ignoring the free balance and will only drop if the schedule is actually over.

### DIFF
--- a/frame/vesting/src/lib.rs
+++ b/frame/vesting/src/lib.rs
@@ -52,7 +52,7 @@ use codec::{Encode, Decode};
 use sp_runtime::{DispatchResult, RuntimeDebug, traits::{
 	StaticLookup, Zero, SimpleArithmetic, MaybeSerializeDeserialize, Saturating, Convert
 }};
-use frame_support::{decl_module, decl_event, decl_storage, ensure, decl_error};
+use frame_support::{decl_module, decl_event, decl_storage, decl_error};
 use frame_support::traits::{
 	Currency, LockableCurrency, VestingSchedule, WithdrawReason, LockIdentifier
 };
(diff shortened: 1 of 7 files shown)


frame/vesting/src/lib.rs
```diff
@@ -249,6 +251,10 @@ impl<T: Trait> VestingSchedule<T::AccountId> for Module<T> where
 	/// If there already exists a vesting schedule for the given account, an `Err` is returned
 	/// and nothing is updated.
 	///
+	/// On success, a linearly reducing amount of funds will be locked. In order to realise any
```
bkchr: These docs should be added to the trait definition :) 

And I think it would be good to document that this doesn't creates any currency.

gavofyork: i'll put it in if you like, but nothing about the function name or existing docs state anything about adding "currency" or "balance". the name is "add_vesting_schedule", implying a vesting schedule is added, not any funds.

gavofyork:
> Isn't the implementation of `vesting_balance` still invalid?
> 
> Shouldn't it return what currently is stored in the lock for `vesting`?

No, vesting balance is correct.

In general the amount locked for vesting (`unvested`) will be at least the `free_balance`, and as such the amount of funds that will eventually vest (`vesting_balance`) is equal to `unvested`. However, in the case that `unvested` is greater, then the amount of funds that will *eventually* vest is not that but `free_balance`. In general then it's the minimum of the two.

frame/vesting/src/lib.rs
bkchr: Yeah that is correct, I'm probably just confused of the old implementation.

However, it would be nice to put your doc update to the trait definition.

PR #4865, opened 2020-02-09
Full block import benchmark

NikVolf:
`$ cargo bench -p node-testing import`

This isolates block execution and benches under worst programmable (not imaginable) conditions  - no cache due to random transfer destinations.

What it does is first generates block full of randomized transactions (so that cache never should be hit when importing it), and then measures many imports of such a block on new fresh database every time.

Also there is a small profiling "benchmark" (not actually benchmark) to help profiling in perf/instruments (it makes pause right before and right after actual _interesting_ work is done, so the right time interval can be selected in instruments). This is just there until better tooling for sending signals to perf/instruments somewhere available.

### DIFF
--- a/bin/node/testing/benches/import.rs
+++ b/bin/node/testing/benches/import.rs
@@ -1,0 +1,390 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
+// This file is part of Substrate.
+
+// Substrate is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Block import benchmark.
+//!
+//! This benchmark is expected to measure block import operation of
+//! some full block.
+//!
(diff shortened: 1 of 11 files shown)


bin/node/testing/benches/import.rs
```diff
@@ -0,0 +1,504 @@
...
+//!
+//! As we also want to protect against cold-cache attacks, this
+//! benchmark should not rely on any caching (except those that
+//! DO NOT depend on user input). Thus block generation should be
+//! based on randomized operation.
+//!
+//! This is supposed to be very simple benchmark and is not subject
+//! to much configuring - just block full of randomized transactions.
+//! It is not supposed to measure runtime modules weight correctness
+//! (there is a dedicated benchmarking mode for this).
```
marcio-diaz: Maybe delete this sentence, it doesn't seem necessary.

marcio-diaz:
FYI, I got these results with the maximum of 751 transfers:
- 2.43 seconds, wasm import.
- 173 milliseconds, native import.

primitives/keyring/src/sr25519.rs
```diff
@@ -86,7 +86,7 @@ impl Keyring {
 	pub fn public(self) -> Public {
 		self.pair().public()
 	}
-
+
```
marcio-diaz:
```suggestion

```

arkpar:
Would be cool to be able to set db cache settings and pruning mode on the CLI as well.

NikVolf:
> Would be cool to be able to set db cache settings and pruning mode on the CLI as well.

Do want to try some small values of it?
Because with just 1 block it never hits reasonable limits anyway.

arkpar:
WIth pruning enabled we end up writing an additional journal to the DB. Cache size is not that important I guess.

NikVolf:
@arkpar I will add it shortly

PR #4869, opened 2020-02-09
Export crypto_full feature in primitives/finality-grandpa

h4x3rotab:
Trivial change. Exporting `crypto_full` can make it easier to build a light client in Intel SGX (`no_std`) environment.


bkchr:
You don't need this. If you want to use the `full-crypto` feature downstream, just add `sp-core` as dependency and enable the `full-crypto` feature. Cargo will enable the feature for every usage of `sp-core`.

I don't want to clutter all the `Cargo.toml` files if not needed.

PR #4871, opened 2020-02-10
Add timing trace for import block.

marcio-diaz:
I think we should trace the timing for import block.
I'm not sure this is the best place though (another option may be to log in `import_batch_of_blocks`?).

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -1599,6 +1599,9 @@ impl<B, E, Block, RA> sp_consensus::BlockImport<Block> for &Client<B, E, Block, 
 		mut import_block: BlockImportParams<Block, backend::TransactionFor<B, Block>>,
 		new_cache: HashMap<CacheKeyId, Vec<u8>>,
 	) -> Result<ImportResult, Self::Error> {
+		let span = tracing::span!(tracing::Level::DEBUG, "import_block");
+		let _enter = span.enter();
+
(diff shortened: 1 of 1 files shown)


client/src/client.rs
```diff
@@ -1599,6 +1599,9 @@ impl<B, E, Block, RA> sp_consensus::BlockImport<Block> for &Client<B, E, Block,
 		mut import_block: BlockImportParams<Block, backend::TransactionFor<B, Block>>,
 		new_cache: HashMap<CacheKeyId, Vec<u8>>,
 	) -> Result<ImportResult, Self::Error> {
+		let span = tracing::span!(tracing::Level::DEBUG, "import_block");
```
bkchr: Isn't there a macro for doing this?

CC @mattrutherford

mattrutherford: actually no, there isn't at the moment.

mattrutherford:
Perhaps we can add to the `values` of the trace: 

- the hash as a means to identify the block 
- the result (maybe also including `ImportedAux`)

marcio-diaz:
> Perhaps we can add to the `values` of the trace:
> 
>     * the hash as a means to identify the block
> 
>     * the result (maybe also including `ImportedAux`)

I'm afraid that may have some performance penalty, so I'm leaving like this for now.

PR #4872, opened 2020-02-10
Export GRANDPA AuthorityPair when full_crypto is enabled

h4x3rotab:
Exporting GRANDPA `AuthorityPair` can make it easier to build a light client in Intel SGX (no_std) environment.

This PR is a re-open of #4869. In the previous PR I forgot to add the AuthorityPair change.

### DIFF
--- a/primitives/finality-grandpa/Cargo.toml
+++ b/primitives/finality-grandpa/Cargo.toml
@@ -23,3 +23,6 @@ std = [
 	"sp-api/std",
 	"sp-runtime/std",
 ]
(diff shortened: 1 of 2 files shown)


bkchr:
Will you only need this for Grandpa? Or also for Babe or similar?

h4x3rotab:
Good point. Right now Grandpa is the only one I'm using but I may need it for other consensus modules as well. Do you think I should change them at once?

h4x3rotab:
Thanks!

PR #4874, opened 2020-02-10
Update trie-db to the latest

cecton:
This will remove the duplicate dependencies on trie-db and trie-root

```
[0] [09:52:55] ~/r/substrate cecton-update-trie-db > rg trie-root
Cargo.lock
3419: "trie-root 0.15.2",
7199: "trie-root 0.15.2",
7266: "trie-root 0.16.0",
8151: "trie-root 0.16.0",
8183:name = "trie-root"
8192:name = "trie-root"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -472,9 +472,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.1.2"
+version = "3.2.0"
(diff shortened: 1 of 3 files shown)


bkchr:
Did you sync Kusama with this?

cecton:
polkadot is already using the lastest version. Only the state-machine here were still using the old one.

PR #4875, opened 2020-02-10
Don't expose `Benchmarking` host functions by default

bkchr:
Instead of exposing the benchmarking functions to each Substrate runtime by default, it must be explicitly enabled in the  `native_executor_instance!` macro call.

If the functions are not exported, but requested by the runtime, the instantiation will fail.

### DIFF
--- a/bin/node/executor/src/lib.rs
+++ b/bin/node/executor/src/lib.rs
@@ -25,5 +25,6 @@ use sc_executor::native_executor_instance;
 native_executor_instance!(
 	pub Executor,
 	node_runtime::api::dispatch,
-	node_runtime::native_version
+	node_runtime::native_version,
+	sp_io::benchmarking::HostFunctions,
 );
--- a/primitives/io/src/lib.rs
+++ b/primitives/io/src/lib.rs
@@ -774,25 +774,25 @@ pub trait Logging {
 /// Interface that provides functions for benchmarking the runtime.
 #[runtime_interface]
 pub trait Benchmarking {
-		/// Get the number of nanoseconds passed since the UNIX epoch
-		///
-		/// WARNING! This is a non-deterministic call. Do not use this within
-		/// consensus critical logic.
-		fn current_time() -> u128 {
-			std::time::SystemTime::now().duration_since(std::time::SystemTime::UNIX_EPOCH)
-				.expect("Unix time doesn't go backwards; qed")
-				.as_nanos()
-		}
+	/// Get the number of nanoseconds passed since the UNIX epoch
+	///
+	/// WARNING! This is a non-deterministic call. Do not use this within
+	/// consensus critical logic.
+	fn current_time() -> u128 {
+		std::time::SystemTime::now().duration_since(std::time::SystemTime::UNIX_EPOCH)
+			.expect("Unix time doesn't go backwards; qed")
+			.as_nanos()
+	}
 
-		/// Reset the trie database to the genesis state.
-		fn wipe_db(&mut self) {
-			self.wipe()
-		}
+	/// Reset the trie database to the genesis state.
+	fn wipe_db(&mut self) {
+		self.wipe()
+	}
 
-		/// Commit pending storage changes to the trie database and clear the database cache.
-		fn commit_db(&mut self) {
-			self.commit()
-		}
+	/// Commit pending storage changes to the trie database and clear the database cache.
+	fn commit_db(&mut self) {
+		self.commit()
+	}
 }
 
 /// Wasm-only interface that provides functions for interacting with the sandbox.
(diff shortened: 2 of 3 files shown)


bin/node/executor/src/lib.rs
```diff
@@ -25,5 +25,6 @@ use sc_executor::native_executor_instance;
 native_executor_instance!(
 	pub Executor,
 	node_runtime::api::dispatch,
-	node_runtime::native_version
+	node_runtime::native_version,
+	sp_io::benchmarking::HostFunctions,
```
tomusdrw: It's still exposed, but only in `node`, isn't it? Why not use `Capabilities` and `Externalities` extensions for this?

primitives/io/src/lib.rs
```diff
@@ -774,25 +774,25 @@ pub trait Logging {
...
-		fn current_time() -> u128 {
-			std::time::SystemTime::now().duration_since(std::time::SystemTime::UNIX_EPOCH)
-				.expect("Unix time doesn't go backwards; qed")
-				.as_nanos()
-		}
+	/// Get the number of nanoseconds passed since the UNIX epoch
+	///
+	/// WARNING! This is a non-deterministic call. Do not use this within
+	/// consensus critical logic.
+	fn current_time() -> u128 {
```
tomusdrw: This is pretty much the same as `offchain::now`

bin/node/executor/src/lib.rs
bkchr: This is not an extension, these are the actual host functions that will be exposed by the wasm executor.

So, what you want will not work.

primitives/io/src/lib.rs
bkchr: Not sure. Probably @shawntabrizi knows the reasoning behind this.

shawntabrizi: Mainly I needed nanoseconds and did not want to touch the offchain worker stuff. Is there some downside to exposing my own host function?

Especially given the fact that we will eventually hide all these host functions behind a feature flag?

bin/node/executor/src/lib.rs
tomusdrw: I'm not sure I get how these functions here differ from extensions. To best of my knowledge, extensions are always exposed, but are not always callable. This here will also always expose the bechmarking functions and make them always callable, but only when you are using `node/executor`. So it means benchmarking functions are going to be available even when you import blocks (running `node` obviously). To me it doesn't sound like it's superior to extensions then unless we actually want to measure real import time / extrinsic dispatch. but even for this I'm sure we could come up with API that is deterministic (but just have non-deterministic side effects, like printing timing log lines).

primitives/io/src/lib.rs
tomusdrw: Fair enough, I wanted to avoid duplicating functions (especially if you decide to switch to capabilities anyway). Increased precision and limited usage scope is a good argument to keep it as is though.

bin/node/executor/src/lib.rs
bkchr: This operates on a level below `Extensions`. Let's explain at the example of offchain runtime interfaces. We always expose the offchain runtime interfaces to the runtime, this is required because the same runtime will be used to execute the offchain worker. However, when you call an offchain function from inside the block production/import it will fail because we did not give it the extension. However, that means that the function is callable.

Now to this change here. A runtime can still call these benchmarking functions, that is right. However, by default it will fail at instantiation of the runtime, because these functions are not given to the wasm-executor and thus, they are not exposed to the runtime. If you try to do this, you will get an error that complains about missing functions bla. 

However, as we need the benchmarking functionality in the Substrate node, we instruct the `Executor` to also expose the benchmarking functions. So, it will not fail at instantiation of the runtime.

tomusdrw: Yup, but it means that it's still possible to build a runtime that uses this functions, compile it and if it's going to be run on the correct executor (default `node/executor` here) it will just work fine, exposing non-deterministic functions to the state transition function.

To me this is going to set a precedence of having non-deterministic functions callable, with only a warning in the docs. I was hoping we can do better than this to protect runtime devs, but it seems I'm in minority, and since we're not creating children's toys here, I won't oppose.

bkchr: If you tell me how? :D 

It will run on the Substrate node, but that is anyway some special node that we use for benchmarking etc. When you create your own node, you should not expose this functionality to the executor! Any executor created by `native_executor_instance` will fail to instantiate your runtime when you try to call one of these functions.

We will also improve here in the node with a benchmarking feature or similar. However, this requires some changes to `wasm-builder` and `sp-api` to make it work correctly.

PR #4877, opened 2020-02-10
Co-access some system storage items

gavofyork:
This is just a bit of low-hanging fruit that merges a few co-accessed storage items. There's really no need to merge unless/until benchmarks imply that it actually speeds things up.

CC @NikVolf


NikVolf:
```
import block/Native     time:   [244.93 ms 251.98 ms 260.61 ms]                              
                        change: [-27.332% -16.345% -3.9543%] (p = 0.02 < 0.05)
                        Performance has improved.
```

though I'm still not sure that this benchmark is totally reliable, give me some more time to test it in different scenarios before using it as argument

NikVolf:
Ok, I did some work to reduce benchmarking variance and getting little difference (1-3%) now.

I will do a CI job update so that results will be seen on independent machine without any bias. This will also allow to spot any variance by just restarting the job few times on the same example and work on this as well.

gavofyork:
ok - if it's definitely better, then we can merge.

NikVolf:
Sorry to disappoint, but I can't say it is definitely better -- updated benchmark running on dedicated VPS (with almost no variance) shows no significant difference

PR #4878, opened 2020-02-10
Run offchain workers at hash, not number.

tomusdrw:
Current execution of offchain workers is potentially racey with block (fork) imports. We run the workers at specific block NUMBER, while we should be running at specific block HASH.

This PR changes the execution to be bound to a specific hash and adds a bunch of more detailed logging to `UnknownBlock` errors that are observed in the wild.

```
2020-02-10 03:14:37.295 import-queue-worker-0 TRACE babe  Ignoring digest not meant for us
2020-02-10 03:14:37.297 import-queue-worker-0 TRACE db  Canonicalize block #979162 (0x1fe9739db9a2f21edc075394a166e631d20296dad1d8757e01ab98f79f1506da)
2020-02-10 03:14:37.297 import-queue-worker-0 DEBUG db  DB Commit 0x41222a919b0d1fdc868f6566d8093f5e84f671b49d9bfe8dfbf9d88647087518 (983258), best = false
2020-02-10 03:14:37.297 import-queue-worker-0 TRACE forks  First available: None (0), Last canon: None (0), Best forks: []
2020-02-10 03:14:37.297 import-queue-worker-0 TRACE sync  Block imported successfully Some(983258) (0x4122…7518)
2020-02-10 03:14:37.297 main-tokio- ERROR sc_offchain  Unsupported Offchain Worker API version: (Err(UnknownBlock("BlockId::Number(983258)")), Err(UnknownBlock("BlockId::Number(983258)"))). Consider turning off offchain workers if they
are not part of your runtime..
```
The error gets triggered when we import a block that is not a new best block. The state is not available for such blocks, and the offchain worker is failing.

Because of that in this PR I've added additional check to only run OW when the block is new best.

### DIFF
--- a/client/db/src/utils.rs
+++ b/client/db/src/utils.rs
@@ -320,7 +320,9 @@ pub fn require_header<Block: BlockT>(
 	id: BlockId<Block>,
 ) -> sp_blockchain::Result<Block::Header> {
 	read_header(db, col_index, col, id)
-		.and_then(|header| header.ok_or_else(|| sp_blockchain::Error::UnknownBlock(format!("{}", id))))
+		.and_then(|header| header.ok_or_else(||
+			sp_blockchain::Error::UnknownBlock(format!("Require header: {}", id))
+		))
 }
 
 /// Read meta from the database.
--- a/client/service/src/builder.rs
+++ b/client/service/src/builder.rs
@@ -913,7 +913,7 @@ ServiceBuilder<
 						Some(_) => log::debug!(
 							target: "sc_offchain",
 							"Skipping offchain workers for non-canon block: {:?}",
-							notification.header
+							notification.header,
 						),
 						_ => {},
 					}
--- a/client/service/src/builder.rs
+++ b/client/service/src/builder.rs
@@ -907,7 +907,7 @@ ServiceBuilder<
 							);
 							let _ = to_spawn_tx_.unbounded_send((
 									Box::pin(future),
-									From::from("offchain-on-block")
+									From::from("offchain-on-block"),
 							));
 						},
 						Some(_) => log::debug!(
--- a/client/service/src/builder.rs
+++ b/client/service/src/builder.rs
@@ -903,7 +903,7 @@ ServiceBuilder<
 							let future = offchain.on_block_imported(
 								&notification.header,
 								network_state_info.clone(),
-								is_validator
+								is_validator,
 							);
 							let _ = to_spawn_tx_.unbounded_send((
 									Box::pin(future),
--- a/client/service/src/builder.rs
+++ b/client/service/src/builder.rs
@@ -898,16 +898,24 @@ ServiceBuilder<
 					}
 
 					let offchain = offchain.as_ref().and_then(|o| o.upgrade());
-					if let Some(offchain) = offchain {
-						let future = offchain.on_block_imported(
-							&notification.header,
-							network_state_info.clone(),
-							is_validator
-						);
-						let _ = to_spawn_tx_.unbounded_send((
-							Box::pin(future),
-							From::from("offchain-on-block")
-						));
+					match offchain {
+						Some(offchain) if notification.is_new_best => {
+							let future = offchain.on_block_imported(
+								&notification.header,
+								network_state_info.clone(),
+								is_validator
+							);
+							let _ = to_spawn_tx_.unbounded_send((
+									Box::pin(future),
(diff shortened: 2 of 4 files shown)


client/db/src/utils.rs
```diff
@@ -320,7 +320,9 @@ pub fn require_header<Block: BlockT>(
 	id: BlockId<Block>,
 ) -> sp_blockchain::Result<Block::Header> {
 	read_header(db, col_index, col, id)
-		.and_then(|header| header.ok_or_else(|| sp_blockchain::Error::UnknownBlock(format!("{}", id))))
+		.and_then(|header| header.ok_or_else(||
+			sp_blockchain::Error::UnknownBlock(format!("Require header: {}", id))
```
NikVolf: Isn't it supposed to be just block id in this error?
Looks strange, is it `UnknownBlock` or "Require header" error now?

client/service/src/builder.rs
```diff
@@ -898,16 +898,24 @@ ServiceBuilder<
...
+							);
+							let _ = to_spawn_tx_.unbounded_send((
+									Box::pin(future),
+									From::from("offchain-on-block")
+							));
+						},
+						Some(_) => log::debug!(
+							target: "sc_offchain",
+							"Skipping offchain workers for non-canon block: {:?}",
+							notification.header
```
NikVolf:
```suggestion
							notification.header,
```

client/service/src/builder.rs
```diff
@@ -898,16 +898,24 @@ ServiceBuilder<
...
-						let _ = to_spawn_tx_.unbounded_send((
-							Box::pin(future),
-							From::from("offchain-on-block")
-						));
+					match offchain {
+						Some(offchain) if notification.is_new_best => {
+							let future = offchain.on_block_imported(
+								&notification.header,
+								network_state_info.clone(),
+								is_validator
```
NikVolf:
```suggestion
								is_validator,
```

primitives/blockchain/src/backend.rs
```diff
@@ -59,19 +59,23 @@ pub trait HeaderBackend<Block: BlockT>: Send + Sync {
 
 	/// Get block header. Returns `UnknownBlock` error if block is not found.
 	fn expect_header(&self, id: BlockId<Block>) -> Result<Block::Header> {
-		self.header(id)?.ok_or_else(|| Error::UnknownBlock(format!("{}", id)))
+		self.header(id)?.ok_or_else(|| Error::UnknownBlock(format!("Expect header: {}", id)))
```
NikVolf: Semantics is unclear, maybe provide more explanation in the error message, just leave id here there, or create new error enum variant?

client/db/src/utils.rs
tomusdrw: It's `UnknownBlock` in the context of `Require header`. We use the same error type in so many places that it's hard to understand which place triggers the error currently. So I wanted to augment the error messages, without affecting the semantics.

primitives/blockchain/src/backend.rs
tomusdrw: Not sure what exactly you expect, `UnknownBlock` error is used pretty commonly everywhere, I wanted to add some extra context to the error message so that we have more info tracing where it comes from. Such `UnknownBlock` errors are constructed in other places in the codebase already (giving some extra description).

Unless you mean to just change `"Expect header" (the function name) to something more like: "Missing header when calling `expect_header`", is that what you mean?

NikVolf: Yeah, roughly.
Ok, so
`UnknownBlock(<Specifying which block>)` - is a clear semantics
`UnknownBlock(<Explaing situation that lead to that error>)` - is a clear semantics
`UnknownBlock(<Expect header?/Require header?>)` - not so much clear

 But not strong on this, especially if it is just for tracking error origin/tracing.

client/service/src/builder.rs
```diff
@@ -898,16 +898,24 @@ ServiceBuilder<
...
+					match offchain {
+						Some(offchain) if notification.is_new_best => {
+							let future = offchain.on_block_imported(
+								&notification.header,
+								network_state_info.clone(),
+								is_validator,
+							);
+							let _ = to_spawn_tx_.unbounded_send((
+									Box::pin(future),
+									From::from("offchain-on-block")
```
NikVolf:
```suggestion
									From::from("offchain-on-block"),
```

PR #4859, opened 2020-02-07
epoch_changes: functions to fetch by reference and mutable reference

sorpaas:
Add function `epoch_for_child_of_ref` and `epoch_for_child_of_mut` to allow fetching values by reference and mutable reference, avoiding copying, and enabling writing data back to fork tree.

PR #4881, opened 2020-02-10
epoch-changes: separate epoch header with epoch data

sorpaas:
The intention of this PR is to make it easier to access and modify values within epoch data. This is currently not possible, because `EpochChanges` now only returns a cloned `Epoch`.

* `ForkTree` now only stores an `EpochHeader` as its data field.
* The rest of the epoch data is moved to a `BTreeMap` on `EpochChanges::epochs`.
* When accessing an `EpochChanges` structure, the caller first uses an expensive operation to obtain an "epoch descriptor". This descriptor pin-point an epoch data. The caller can then use the descriptor to access the actual epoch data repetitively, which is a fast `BTreeMap` get operation. This fits nicely with block import's "intermediates" without lifetime issues.

Note that this change will break existing BABE nodes, due to the changes in `EpochChanges` structure. It does not break consensus, but existing nodes will have to resync. An alternative is #4859, but that one has few disadvantages, notably any engines that require mutating epoch data cannot take advantage of block import intermediates.


sorpaas:
> How are ForkTrees storing old BABE data migrated?

If resync is too much to ask, we may need a migration script (to convert the old `EpochChanges` data to new format).

> How do you protect against races with pruning when fetching by epoch descriptor.

I think that's nearly impossible to happen, because we prune only finalized epochs, which, by consensus algorithm, has little chance to race. Even if it does race there won't be any harms -- the epoch fetch operation fails (returns `None`) and it causes a block import to fail, but nothing else.

> I'm not sure I understand the motivation for the split either.

The current design of `EpochChanges` does not allow changing epoch data after an epoch is created. This is a function needed for engines like Sassafras. We can either do it like in https://github.com/paritytech/substrate/pull/4859 or like in this PR. If we do it like in #4859, we cannot take advantage of `intermediates` and must fetch epochs twice during both block import and verifier. If we do it like in this PR, we can fetch the descriptor, put it in intermediates, and use the descriptor as a handle to modify epoch data.

bkchr:
Any update?

CC @andresilva

sorpaas:
Let me fix merge conflicts and also get the migration script within this PR.

sorpaas:
Added migration script for the new epoch changes definition.

client/consensus/slots/src/lib.rs
```diff
@@ -194,7 +195,11 @@ pub trait SimpleSlotWorker<B: BlockT> {
 
 		let authorities_len = self.authorities_len(&epoch_data);
 
-		if !self.force_authoring() && self.sync_oracle().is_offline() && authorities_len > 1 {
+		if authorities_len.is_none() ||
+			(!self.force_authoring() &&
+			 self.sync_oracle().is_offline() &&
+			 authorities_len.map(|a| a > 1).unwrap_or(true))
```
andresilva:
```suggestion
			 authorities_len.map(|a| a > 1).unwrap_or(false))
```

andresilva: I think we can remove initial `authorities_len.is_none()` check if we do this instead.

client/consensus/epochs/src/lib.rs
```diff
@@ -67,97 +69,209 @@ impl<'a, H, Block> IsDescendentOfBuilder<Block::Hash>
 }
 
 /// Epoch data, distinguish whether it is genesis or not.
+///
+/// Once an epoch is created, it must have a known `start_slot` and `end_slot`, which cannot be
+/// changed. Consensus engine can modify and other data into the epoch, if needed.
```
andresilva:
```suggestion
/// changed. Consensus engine may modify any other data in the epoch, if needed.
```

client/consensus/epochs/src/lib.rs
```diff
@@ -67,97 +69,209 @@ impl<'a, H, Block> IsDescendentOfBuilder<Block::Hash>
...
 pub trait Epoch {
 	/// Descriptor for the next epoch.
 	type NextEpochDescriptor;
 	/// Type of the slot number.
-	type SlotNumber: Ord;
+	type SlotNumber: Ord + Copy;
 
+	/// The starting slot of the epoch.
+	fn start_slot(&self) -> Self::SlotNumber;
+	/// The end slot of the epoch.
```
andresilva: > /// Produce the "end slot" of the epoch. This is NOT inclusive to the epoch,
> /// i.e. the slots covered by the epoch are `self.start_slot() .. self.end_slot()`. 

Do you think this is not relevant for the generic epoch logic below? The pruning logic relies on this IIRC.

client/consensus/epochs/src/lib.rs
```diff
@@ -277,20 +508,19 @@ impl<Hash, Number, Epoch> EpochChanges<Hash, Number, Epoch> where
 
 		if parent_number == Zero::zero() {
 			// need to insert the genesis epoch.
-			let genesis_epoch = make_genesis(slot_number);
-			return Ok(Some(ViableEpoch::Genesis(UnimportedGenesisEpoch(genesis_epoch))));
+			return Ok(Some(ViableEpochDescriptor::UnimportedGenesis(slot_number)))
```
andresilva:
```suggestion
			return Ok(Some(ViableEpochDescriptor::UnimportedGenesis(slot_number)));
```

client/consensus/babe/src/aux_schema.rs
```diff
@@ -25,10 +25,11 @@ use sc_client_api::backend::AuxStore;
 use sp_blockchain::{Result as ClientResult, Error as ClientError};
 use sp_runtime::traits::Block as BlockT;
 use sp_consensus_babe::BabeBlockWeight;
-use sc_consensus_epochs::{EpochChangesFor, SharedEpochChanges};
+use sc_consensus_epochs::{EpochChangesFor, SharedEpochChanges, migration::EpochChangesForV0};
 use crate::Epoch;
 
-const BABE_EPOCH_CHANGES: &[u8] = b"babe_epoch_changes";
+const BABE_EPOCH_CHANGES_V0: &[u8] = b"babe_epoch_changes";
```
andresilva: In GRANDPA code we are keeping the data under the same key, but instead we keep track of data version in a different key (https://github.com/paritytech/substrate/blob/master/client/finality-grandpa/src/aux_schema.rs#L37). I think this is a bit tidier but wouldn't block on this.

rphmeier:
LGTM after Andre's review comments. A separate key for version in storage would make it better for doing further migrations in the future.

client/consensus/babe/src/aux_schema.rs
```diff
@@ -52,14 +53,28 @@ fn load_decode<B, T>(backend: &B, key: &[u8]) -> ClientResult<Option<T>>
...
-				"Creating empty BABE epoch changes on what appears to be first startup."
-			);
-			SharedEpochChanges::<Block, Epoch>::default()
-		});
+				  "V1 version of BABE epoch changes does not exist, trying to load V0.");
+			match load_decode::<_, EpochChangesForV0<Block, Epoch>>(
+				backend,
+				BABE_EPOCH_CHANGES_V0
+			)? {
+				Some(v0) => Arc::new(Mutex::new(v0.migrate())),
```
rphmeier: I'd like a test that this code path works as expected.

client/consensus/epochs/src/lib.rs
sorpaas: Checking https://wiki.parity.io/Substrate-Style-Guide I think we encourage not having trailing `;` for `return` statement.

> * Avoid trailing ;s where unneeded.
> ```
> if condition {
> 	return 1    // <-- no ; here
> }
> ```

So I think maybe we should keep the current.

utils/fork-tree/src/lib.rs
```diff
@@ -93,41 +93,77 @@ impl<H, N, V> ForkTree<H, N, V> where
...
 
-		if let Some(root) = new_root {
-			let mut root = root.clone();
+		if let Some(mut root_index) = new_root_index {
+			let old_roots = std::mem::replace(&mut self.roots, Vec::new());
+
+			let mut root = None;
+			let mut cur_children = Some(old_roots);
+
+			while let Some(cur_index) = root_index.pop() {
```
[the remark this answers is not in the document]
sorpaas: Yeah. I realized that the previous logic of the removal list was flawed, just fixed it and added some tests. So it's all top-level non-canon `Node`, with `children` field with whatever is still being kept removed from the list.

utils/fork-tree/src/lib.rs
```diff
@@ -93,41 +93,77 @@ impl<H, N, V> ForkTree<H, N, V> where
...
+			let root_children = std::mem::replace(&mut root.children, Vec::new());
+			let mut is_first = true;
+
+			for child in root_children {
+				if is_first &&
+					(child.number == *number && child.hash == *hash ||
+					 child.number < *number && is_descendent_of(&child.hash, hash).unwrap_or(false))
+				{
+					root.children.push(child);
+					is_first = false;
```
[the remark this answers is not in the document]
sorpaas: Looks like this is done!

client/consensus/epochs/src/lib.rs
sorpaas: This is still relevant to the generic epoch logic, I think. However, I don't believe we have any special logic for genesis `end_slot`. It should be treated the same as any other epochs.

I'll add back the comments.

sorpaas:
Fixing some weird issues for tests. Marking in-progress until I'm finished.

sorpaas:
Fixed the issue and added migration tests.

rphmeier:
Looks good! Once you can confirm that you've tested out the migration code on a synced Kusama node we should go ahead and merge this.

sorpaas:
Tested full sync of Kusama with migration at block ~0.5m. Not detecting any issues.

PR #4883, opened 2020-02-10
Allow to distinguish out of gas from other traps

### LINKED ISSUE #2584: Is it possible to tell if a contract execution error is caused by out of gas?
All I can see now from node terminal is `Runtime: during execution` and nothing much on sdk side.

What does this `during execution` actually mean? Additional println give me `Err(Trap(Trap { kind: Host(DummyHostError) }))`, which is still not very helpful.

Does `DummyHostError` / `during execution` already mean out of gas or there are other conditions to trigger this?

@pepyakin

athei:
**Description**
When a contract encounters a runtime error a wasm trap is
triggered and the execution is halted. Currently, no matter
what was the cause for the trap it is always reported as:
DispatchError::Other("contract trapped during execution").

However, the trap that is triggered if a contract exhausts
its gas budget is particulary interesting. Therefore we add
a seperate error message for this cause:
DispatchError::Other("ran out of gas during contract execution").

A test is added hat executes a contract that never terminates.
Therefore it always exhausts is gas budget.

 fixes #2584

**Open Issues**
* ~~I am not sure whether to bump `impl_version`or `spec_version`.~~

_spec_version should be bumped_

* ~~The field of the `Other` error variant (the error string) seems not to be passed through to the PolkadotAppsUI which renders this improvement rather unsatisfying:~~
extrinsic_failed

_This is OK for now_

### DIFF
--- a/frame/contracts/src/tests.rs
+++ b/frame/contracts/src/tests.rs
@@ -722,6 +722,51 @@ fn dispatch_call_not_dispatched_after_top_level_transaction_failure() {
 	});
 }
 
+const CODE_RUN_OUT_OF_GAS: &str = r#"
+(module
+	(func (export "call")
+		(loop $inf (br $inf)) ;; just run out of gas
+		(unreachable)
+	)
+	(func (export "deploy"))
+)
+"#;
+
+#[test]
+fn run_out_of_gas() {
+	let (wasm, code_hash) = compile_module::<Test>(CODE_RUN_OUT_OF_GAS).unwrap();
+
+	ExtBuilder::default()
+		.existential_deposit(50)
+		.build()
+		.execute_with(|| {
+			Balances::deposit_creating(&ALICE, 1_000_000);
+
+			assert_ok!(Contract::put_code(Origin::signed(ALICE), 100_000, wasm));
+
+			assert_ok!(Contract::instantiate(
+				Origin::signed(ALICE),
+				100,
+				100_000,
+				code_hash.into(),
+				vec![],
+			));
+
+			// Call the contract with a fixed gas limit. It must run out of gas because it just
+			// loops forever.
+			assert_err!(
+				Contract::call(
+					Origin::signed(ALICE),
+					BOB, // newly created account
+					0,
+					1000,
+					vec![],
+				),
+				"ran out of gas during contract execution"
+			);
+		});
+}
+
 const CODE_SET_RENT: &str = r#"
 (module
 	(import "env" "ext_dispatch_call" (func $ext_dispatch_call (param i32 i32)))
--- a/frame/contracts/src/tests.rs
+++ b/frame/contracts/src/tests.rs
@@ -742,9 +742,9 @@ fn run_out_of_gas() {
 		.execute_with(|| {
 			Balances::deposit_creating(&ALICE, 1_000_000);
 
-			assert_ok!(Contract::put_code(Origin::signed(ALICE), 100_000, wasm));
+			assert_ok!(Contracts::put_code(Origin::signed(ALICE), 100_000, wasm));
 
-			assert_ok!(Contract::instantiate(
+			assert_ok!(Contracts::instantiate(
 				Origin::signed(ALICE),
 				100,
 				100_000,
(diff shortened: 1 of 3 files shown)


athei:
[clabot:check]

frame/contracts/src/tests.rs
```diff
@@ -681,6 +681,50 @@ fn dispatch_call_not_dispatched_after_top_level_transaction_failure() {
...
+
+	ExtBuilder::default().existential_deposit(50).build().execute_with(|| {
+		Balances::deposit_creating(&ALICE, 1_000_000);
+
+		assert_ok!(Contract::put_code(Origin::signed(ALICE), 100_000, wasm));
+
+		assert_ok!(Contract::instantiate(
+			Origin::signed(ALICE),
+			100,
+			100_000,
```
NikVolf: is it gas for a call?
maybe put less here so that it will stop infinite loop above a bit faster?

frame/contracts/src/tests.rs
```diff
@@ -681,6 +681,50 @@ fn dispatch_call_not_dispatched_after_top_level_transaction_failure() {
...
+		(loop $inf (br $inf)) ;; just run out of gas
+		(unreachable)
+	)
+	(func (export "deploy"))
+)
+"#;
+
+#[test]
+fn run_out_of_gas() {
+
```
pepyakin:
```suggestion
```

frame/contracts/src/wasm/runtime.rs
```diff
@@ -300,7 +318,7 @@ define_env!(Env, <E: Ext>,
 	//
 	// - amount: How much gas is used.
 	gas(ctx, amount: u32) => {
-		charge_gas(&mut ctx.gas_meter, ctx.schedule, RuntimeToken::Explicit(amount))?;
+		charge_gas(&mut ctx.gas_meter, ctx.schedule, &mut ctx.special_trap, RuntimeToken::Explicit(amount))?;
```
pepyakin: This (an a few other) line exceeds the soft limit of 100. It would be great if we wrapped this line, like

```rust
charge_gas(
    &mut ctx.gas_meter,
    ctx.schedule,
    &mut ctx.special_trap,
    RuntimeToken::Explicit(amount)
)?;
```

frame/contracts/src/wasm/runtime.rs
```diff
@@ -39,6 +39,7 @@ const TRAP_RETURN_CODE: u32 = 0x0100;
 enum SpecialTrap {
 	/// Signals that trap was generated in response to call `ext_return` host function.
 	Return(Vec<u8>),
+	OutOfGas,
```
pepyakin: Would be useful to leave some words about this variant.

frame/contracts/src/tests.rs
athei: You are right. We can save some cycles here. No need to make it run longer. The tighter gas limit should be on the `::call` and not on the `::instantiate``though.

frame/contracts/src/wasm/runtime.rs
athei: I will look into the overlong lines.

frame/contracts/src/wasm/runtime.rs
athei: Yes. Makes sense.

athei:
Any idea about the Open Issues part in my PR description?

pepyakin:
Ah yeah sorry. Yes, since you change the runtime behavior you need to bump the spec version as well. Otherwise, clients with older native runtime compiled in will execute that stale code rather than the wasm version under some circumstances.

The second concern is valid, but I believe we cannot do anything with it atm.

athei:
> The second concern is valid, but I believe we cannot do anything with it atm.

I guess the change would be to the javascript? I do not even see the extrinsic error on the node console. Is this intended or do I change the loglevel? The only way that I see that the change is working is that the test succeeds.

pepyakin:
I think yeah you need to change the log level, try `RUST_LOG=runtime=trace`.

frame/contracts/src/tests.rs
athei: I  reduced it to 1000. Enough to pay the base call price.

frame/contracts/src/wasm/runtime.rs
athei: Removed the long lines I added / mofified.

frame/contracts/src/wasm/runtime.rs
athei: Added a comment

xlc:
This fixes https://github.com/paritytech/substrate/issues/2584

PR #4884, opened 2020-02-11
Add a sub command to generate a node key file

hbakkum-dotstar:
This PR adds a new subcommand to subkey, 'generate-node-key' which can be used to generate a new libp2p secret key, save it to file in the format required by a substrate node and also print its peer ID.


tomaka:
> is that something that is useful?

Yes, if you're setting up a bootstrap nodes you need to know your `PeerId`. Right now the only way is to start the node and read what the log messages say.

PR #4885, opened 2020-02-11
Adds `with_pair!` macro to application-crypto

bkchr:
This macro will "generate" the given code only when the crypto pair is
available. So, when either the `std` or the `full_crypto` feature is
enabled.

CC @h4x3rotab


h4x3rotab:
Looks good to me. With this change I can just enable "sp-application-crypto/full_crypto" to use the key pair types in SGX env. Thanks!

h4x3rotab:
CI is complaining about not bumping the runtime impl version. I did't see why the wasm build differs though.

bkchr:
> CI is complaining about not bumping the runtime impl version. I did't see why the wasm build differs though.

That is just a false alarm and can be ignored :)

Issue #4861, opened 2020-02-08
Create macros to simplify benchmarking.


shawntabrizi:
The benchmarking setup introduced by https://github.com/paritytech/substrate/pull/4818 has a lot of boilerplate code which can be improved through the use of macros.

## `decl_module`

The `decl_module` macro should introduce a `#[benchmark(...)]` tag which can be placed over extrinsics. This can include a list of letters [A-Z] which represents parameters to test, and a range over which we should test them. Example:

```rust
#[benchmark([A, 1, 10], [B, 5, 50])]
fn transfer(origin, who: T::AccountId, amount: T::Balance) {
   ...
}
```

All extrinsics which have this attribute will:
* Be added to an Enum which will implement `BenchmarkingSetup` so that we can select which benchmark to run through the CLI.
* Have a lookup function which maps the string name of the function to its appropriate BenchmarkingSetup enum.

Adding the `#[benchmark(...)]` tag should also create an object that can be exposed into the `construct_runtime!` macro, so that we can know this pallet should expose it's benchmarking setup.

## `construct_runtime!`

The `construct_runtime!` macro should then implement the `Benchmark` runtime api, creating the `dispatch_benchmark` function and populating it with Pallets that have exposed benchmarking.

It would simply create a function which maps the appropriate strings to the correct `run_benchmark` function.

```rust
match module.as_slice() {
	b"pallet-identity" | b"identity" => Identity::run_benchmark(extrinsic, steps, repeat).ok(),
	_ => return None,
}

marcio-diaz:
I had the same idea a while ago, and give it some thinking.
Somehow I ended up concluding that writing a trait for each input type was better approach: 

- Pros: it's an order of magnitude less work (~100 tx vs ~10 meaningful types).
- Cons: less precision?

I'm looking to improve my macro skills, so I can give it a try if the macro gods are busy.

shawntabrizi:
@marcio-diaz Traits where applicable are def the way to go, but in terms of writing macros, I want to reduce the following redundant code (from my initial PR):

```rust
// The list of available benchmarks for this pallet.
enum SelectedBenchmark {
	AddRegistrar,
	SetIdentity,
	SetSubs,
	ClearIdentity,
	RequestJudgement,
	CancelRequest,
	SetFee,
	SetAccountId,
	SetFields,
	ProvideJudgement,
	KillIdentity,
}

// Allow us to select a benchmark from the list of available benchmarks.

[...]

Issue #4873, opened 2020-02-10
Hide benchmarking pipeline behind a feature flag


shawntabrizi:
By default, a Substrate node should not expose the benchmarking pipeline added by https://github.com/paritytech/substrate/pull/4818

Instead, a user should have to compile their node using a specific flag like `--features=runtime_benchmarks`.

Without this feature flag, we should hide:

* The bechmark host functions (to prevent misuse within the runtime environment)
* The benchmark runtime api (they wont work)
* The benchmark runtime code (to keep wasm code size small)

PR #4891, opened 2020-02-11
Benchmark Timestamp Pallet

marcio-diaz:
Contains the following changes:
- Starts the benchmark of timestamp pallet. (Am I done?)
- Add macro to remove some duplicated code (part of #4861, I'll keep improving it).
- Use macro in Identity pallet benchmark.


frame/timestamp/Cargo.toml
```diff
@@ -9,6 +9,7 @@ license = "GPL-3.0"
 serde = { version = "1.0.101", optional = true }
 codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false, features = ["derive"] }
 sp-std = { version = "2.0.0", default-features = false, path = "../../primitives/std" }
+sp-io = { version = "2.0.0", default-features = false, path = "../../primitives/io" }
```
kianenigma: Would be good if you and @shawntabrizi make imports that are needed for benchmakring be feature-gated sooner rather than later.

frame/timestamp/src/benchmarking.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use super::*;
+
+use sp_std::prelude::*;
+
+use frame_system::RawOrigin;
+use sp_runtime::{BenchmarkResults, BenchmarkParameter, selected_benchmark};
+use sp_runtime::traits::{Benchmarking, BenchmarkingSetup, Dispatchable};
+
+// Benchmark `set` extrinsic.
```
kianenigma:
```suggestion
/// Benchmark `set` extrinsic.
```

frame/timestamp/src/benchmarking.rs
```diff
@@ -0,0 +1,80 @@
...
+struct Set;
+impl<T: Trait> BenchmarkingSetup<T, crate::Call<T>, RawOrigin<T::AccountId>> for Set {
```
kianenigma:
```suggestion
impl<T: Trait> BenchmarkingSetup<T, Call<T>, RawOrigin<T::AccountId>> for Set {
```

should be imported via `import super::*`?

primitives/runtime/src/traits.rs
```diff
@@ -1349,6 +1349,47 @@ pub trait BenchmarkingSetup<T, Call, RawOrigin> {
...
+/// struct Transfer;
+/// impl BenchmarkingSetup for Transfer { ... }
+///
+/// struct SetBalance;
+/// impl BenchmarkingSetup for SetBalance { ... }
+///
+/// selected_benchmark!(Transfer, SetBalance);
+/// ```
+#[macro_export]
+macro_rules! selected_benchmark {
```
kianenigma:
```suggestion
macro_rules! decl_benchmark {
```

why not? :D

frame/timestamp/src/benchmarking.rs
```diff
@@ -0,0 +1,80 @@
...
+			_ => return Err("Could not find extrinsic."),
+		};
+		
+		// Warm up the DB
+		sp_io::benchmarking::commit_db();
+		sp_io::benchmarking::wipe_db();
+
+		// results go here
+		let mut results: Vec<BenchmarkResults> = Vec::new();
+		// Run the benchmark `repeat` times.
```
kianenigma:
```suggestion
		// Run the benchmark `repeat` times.
```

I am going to recommend some of these docs to be removed since I think over-documentation is also a problem. Probably thee come from Identity which was the first one and it was a bit overly-documented..

frame/timestamp/src/benchmarking.rs
```diff
@@ -0,0 +1,80 @@
...
+		for _r in 0..repeat {
+			// Set up the externalities environment for the setup we want to benchmark.
+			let (call, caller) = <SelectedBenchmark as BenchmarkingSetup<T, crate::Call<T>, RawOrigin<T::AccountId>>>::instance(&selected_benchmark, &[])?;
```
kianenigma:
```suggestion
			let (call, caller) = <SelectedBenchmark as BenchmarkingSetup<T, Call<T>, RawOrigin<T::AccountId>>>::instance(&selected_benchmark, &[])?;
```

primitives/runtime/src/traits.rs
```diff
@@ -1349,6 +1349,47 @@ pub trait BenchmarkingSetup<T, Call, RawOrigin> {
...
+		// The list of available benchmarks for this pallet.
+		enum SelectedBenchmark {
+			$( $bench, )*
+		}
+
+		// Allow us to select a benchmark from the list of available benchmarks.
+		impl<T: Trait> BenchmarkingSetup<T, crate::Call<T>, RawOrigin<T::AccountId>> for SelectedBenchmark {
+			fn components(&self) -> Vec<(BenchmarkParameter, u32, u32)> {
+				match self {
+					$( Self::$bench => <$bench as BenchmarkingSetup<T, crate::Call<T>, RawOrigin<T::AccountId>>>::components(&$bench), )*
```
kianenigma: I'm a bit of a novice with this stuff myself, but I think this is now assuming that this macro is called in a context where `crate::Call<T>` is something. I think we have better workaround for these in our `decl_xxx` macros.

primitives/runtime/src/traits.rs
kianenigma: (given the context, `decl_` doesn't make sense. nvmd)

marcio-diaz: Sounds too cool for the little macro, hehe. I'll improve it and then rename it.

frame/timestamp/Cargo.toml
shawntabrizi: That is being tracked here: https://github.com/paritytech/substrate/issues/4873

frame/timestamp/src/benchmarking.rs
shawntabrizi: Dumb question, whats wrong with "over" docmentation?

frame/timestamp/src/benchmarking.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use frame_system::RawOrigin;
+use sp_runtime::{BenchmarkResults, BenchmarkParameter, selected_benchmark};
+use sp_runtime::traits::{Benchmarking, BenchmarkingSetup, Dispatchable};
+
+// Benchmark `set` extrinsic.
+struct Set;
+impl<T: Trait> BenchmarkingSetup<T, crate::Call<T>, RawOrigin<T::AccountId>> for Set {
+	fn components(&self) -> Vec<(BenchmarkParameter, u32, u32)> {
+		vec![]
```
shawntabrizi: Might as well set a range of now values to be iterated over.

primitives/runtime/src/traits.rs
marcio-diaz: Mmm, I think there are quite a few things wrong with this macro. I'll try next to put the whole thing inside a macro.

frame/timestamp/src/benchmarking.rs
kianenigma: Reduced readability. I also don't recall the fancy names of software engineering literature nowadays, but there is a thing with the context of _a good could should be documented via the code itself_. i.e. it shouldn't need that much documentation. 

Either way, this might be a bit of a personal taste, but IMO strongly: documentation is not like _the more the better_. 

Yes, no documentation is also bad. I am not denying this.

kianenigma: i.e. 

```
// Wipe the DB back to the genesis state.
sp_io::benchmarking::wipe_db();
```

the name of the function here is really enough. Even if not, the reader should go to the definition of the function to read it. Explaining what a --for example-- function does at __each and every call site__ is too much.

kianenigma: (I hope I have convinced you enough :D)

shawntabrizi: I disagree at least with the very specific thing you pointed out.

It is not clear that `wipe_db` brings the runtime state back to genesis versus _actually_ wiping the DB clean.

Usually I let to tell a story with my code, and I like to make sure that story is told in words too. Sometimes the words mean one thing but the code does another. Writing both is an easy way to allow others to catch errors.

bkchr: Code should be as much as possible self explanational. If you need to document each line, there is something wrong with your code. (there are exceptions, like when you do really hard shit ^^)

However, saying that you iterate x times is obvious and is just unnecessary.

For the `wipe_db` function, the name is probably already chosen incorrectly. If it resets to genesis, it should not be called `wipe`. It could be `reset_db` (tells you that we reset to the old state) or even more expressive `reset_db_to_genesis` or just `reset_to_genesis`. This makes it directly clear what happens. For `wipe_db` you now always need to repeat yourself.

kianenigma: EOD, there's a bit of personal preference into it. I am more leaned toward minimising inline doc. Fair enough and thanks to both for the good discussion 👍

PR #4892, opened 2020-02-11
Add command-line flag to enable yamux flow control.

twittner:
We never enabled proper flow-control for yamux streams which may cause stream buffers to exceed their configured limit when the stream producer outpaces the stream consumer. By switching the window update mode to on-read, producers will only receive more sending credit when all data has been consumed from the stream buffer. Using this option creates backpressure on producers. However depending on the protocol there is a risk of deadlock if both endpoints concurrently attempt to send more data than they have credit for and neither side reads before finishing their writes. To facilitate proper testing, this PR adds a command-line flag `use-yamux-flow-control`.

In addition, this PR enables another yamux option -- `lazy_open` -- which prevents the sending of an initial control frame as soon as an endpoint opens a stream, but instead only sets the SYN flag on the first data frame that is being sent eventually. The only good reason that comes to mind, why one would want to send an initial control frame is to grant the remote side more than the default credit which it then can immediately make use of. Since all our streams start with a protocol negotiation where only small protocol names are exchanged, the default credit of 256KiB is more than enough and it is preferable to save the extra, unnecessary frame.

### DIFF
--- a/client/cli/src/params.rs
+++ b/client/cli/src/params.rs
@@ -238,6 +238,21 @@ pub struct NetworkConfigurationParams {
 	#[allow(missing_docs)]
 	#[structopt(flatten)]
 	pub node_key_params: NodeKeyParams,
+
+	/// Enable flow control for yamux streams.
+	///
+	/// By default this is not enabled. When enabled, yamux ensures that senders
+	/// do not overwhelm receivers. Depending on the protocol, there is a risk
+	/// of deadlock, namely if both endpoints want to send data larger than the
+	/// receiver's window and they do not read before finishing their writes.
+	///
+	/// Use this mode only if you are sure that this will never happen, i.e. if
+	/// endpoints A and B never write at the same time, *or* endpoints A and B
+	/// write at most n frames concurrently such that the sum of the frame
+	/// lengths is less or equal to the available credit of A and B
+	/// respectively (which by default is 256 KiB).
+	#[structopt(long = "use-yamux-flow-control")]
+	pub use_yamux_flow_control: bool,
 }
 
 arg_enum! {
--- a/client/cli/src/params.rs
+++ b/client/cli/src/params.rs
@@ -239,18 +239,7 @@ pub struct NetworkConfigurationParams {
 	#[structopt(flatten)]
 	pub node_key_params: NodeKeyParams,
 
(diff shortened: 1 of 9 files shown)


client/cli/src/params.rs
```diff
@@ -238,6 +238,21 @@ pub struct NetworkConfigurationParams {
 	#[allow(missing_docs)]
 	#[structopt(flatten)]
 	pub node_key_params: NodeKeyParams,
+
+	/// Enable flow control for yamux streams.
+	///
+	/// By default this is not enabled. When enabled, yamux ensures that senders
+	/// do not overwhelm receivers. Depending on the protocol, there is a risk
```
bkchr: Where do we write to a socket in a synchronous matter without reading? That doesn't sound like a proper async implementation.

tomaka: We don't. I'm not sure whether it's a good idea to actually tell that to users.
Rather than go into details, maybe the documentation should just say "This is an experimental feature that will be turned on by default once it's been tested.".

bkchr: Sounds like a good idea, could you change this @twittner?

twittner: I can certainly replace the comment with an opaque wording if that is what you want. However I do not understand the reference to "synchronous write" here. The case can simply occur if both ends write concurrently more than their credit allows for and when done in  async/await style (e.g. `stream.write_all(...).await`)  there is no manual polling where they would do an interleaved read operation after a `Poll::Pending`.

bkchr: `stream.write_all(...).await` the same behavior would appear for any TCP socket.

bkchr: So, nothing special for yamux here.

twittner: But what do you mean by "doesn't sound like a proper async implementation"?

bkchr: It was me being stupid. I'm still not familiar with async/await and did not thought about `write_all().await` while that still could have happened with "old" futures as well. So yeah, me just being stupid.

twittner: Thanks for clarifying. I thought I may have misunderstood something.

bkchr: Yeah sorry!

Issue #3798, opened 2019-10-11
Improve test suite for runtime modules. should be tested with non null ExistentialDeposit


thiolliere:
I found an issue in treasury due to the test suite having ExistentialDeposit to 0 https://github.com/paritytech/substrate/pull/3797

if you grep to other modules there is:
```
./srml/utility/src/lib.rs:111:		pub const ExistentialDeposit: u64 = 0;
./srml/democracy/src/lib.rs:1027:		pub const ExistentialDeposit: u64 = 0;
./srml/elections/src/mock.rs:59:	pub const ExistentialDeposit: u64 = 0;
./srml/scored-pool/src/mock.rs:52:	pub const ExistentialDeposit: u64 = 0;
./srml/example/src/lib.rs:679:		pub const ExistentialDeposit: u64 = 0;
./srml/executive/src/lib.rs:357:		pub const ExistentialDeposit: u64 = 0;
./srml/council/src/lib.rs:127:		pub const ExistentialDeposit: u64 = 0;
```

I think test suite should take into account existential deposit in general this is more relevant and probably catch more bug.

Actually I think all parameter types in
```rust
	parameter_types! {
		pub const ExistentialDeposit: u64 = 0;
		pub const TransferFee: u64 = 0;
		pub const CreationFee: u64 = 0;
		pub const TransactionBaseFee: u64 = 0;
		pub const TransactionByteFee: u64 = 0;
	}
```

should be set to a value different from 0 in general (except in some precise test obviously)

kianenigma:
related: https://github.com/paritytech/substrate/pull/1966

kianenigma:
As I also stated here [here](https://github.com/paritytech/substrate/pull/3797#issuecomment-541635501), it seems like the general situation is that the current balances module is really based on having a positive ed. There are some notes here:

- the semantics of ed is defined in a way that ed = 0 doesn't make much sense. One simple approach is to ban ed = 0 and just call it a day.
- allow ed to be shut-down aka. ed = 0 but it looks like some modules (namely Indices and treasury) need further investigation regarding how they interact with this feature.

PR #4894, opened 2020-02-11
Do not allow zero Existential Deposit when using Balances

shawntabrizi:
Fixes https://github.com/paritytech/substrate/issues/3798

The balances module uses comparison operators within it's logic which assume that the existential deposit is some non-zero value.

However, many of our tests use ED of 0 which may result in unexpected behaviors when actually deploying to a real environment.

Here we add a check that the ED of the balances module is non-zero in the construction of the genesis to prevent potential issues on live chains where they may try to set the value to 0.

Additionally, we update all Pallets to have an ED > 0 for their tests.


frame/society/src/lib.rs
```diff
@@ -788,7 +788,7 @@ decl_module! {
 			let mut payouts = <Payouts<T, I>>::get(&who);
 			if let Some((when, amount)) = payouts.first() {
 				if when <= &<system::Module<T>>::block_number() {
-					T::Currency::transfer(&Self::payouts(), &who, *amount, KeepAlive)?;
+					T::Currency::transfer(&Self::payouts(), &who, *amount, AllowDeath)?;
```
shawntabrizi: @gavofyork: ping on this change

The `payouts` transfer allow death in all the other parts of the code. I assume this was just missed.

A user _should_ be able to get paid out using the full funds of the `payouts` account.

shawntabrizi: This fixes a test which was failing now that ED is non-zero, and we weren't allowing a full payout.

frame/balances/src/lib.rs
```diff
@@ -394,6 +394,10 @@ decl_storage! {
 		config(balances): Vec<(T::AccountId, T::Balance)>;
 		// ^^ begin, length, amount liquid at genesis
 		build(|config: &GenesisConfig<T, I>| {
+			assert!(
+				<T as Trait<I>>::ExistentialDeposit::get() > Zero::zero(),
+				"the existential deposit should be non-zero."
```
bkchr:
```suggestion
				"The existential deposit should be greater than zero."
```

PR #4895, opened 2020-02-11
Lazy reaping

gavofyork:
This mostly removes the functionality of `OnReapAccount` and introduces reference-counting on accounts to ensure that data structures referring to an account whose economic defence is the account's ED must be cleaned up before the ED is liquidated.

It's a followup/alternative to #4820


frame/system/src/lib.rs
```diff
@@ -513,6 +535,17 @@ decl_module! {
 			ensure_root(origin)?;
 			storage::unhashed::kill_prefix(&prefix);
 		}
+
+		/// Kill the sending account, assuming there are no references outstanding and the composite
+		/// data is equal to its default value.
+		#[weight = SimpleDispatchInfo::FixedOperational(25_000)]
+		fn suicide(origin) {
```
```suggestion
		fn terminate(origin) {
```

I think we should learn from the Ethereum world and avoid the use of this word :/

The Contracts folks will also be using the term terminate as self-destruct for smart contracts.

frame/system/src/lib.rs
```diff
@@ -633,13 +666,55 @@ impl Default for InitKind {
...
+}
+
 impl<T: Trait> Module<T> {
 	/// Deposits an event into this block's event record.
 	pub fn deposit_event(event: impl Into<T::Event>) {
 		Self::deposit_event_indexed(&[], event.into());
 	}
 
-		/// Deposits an event into this block's event record adding this event
+	/// Increment the reference counter on an account. `_context` is a RefContext value identifying
```
```suggestion
	/// Increment the reference counter on an account. `_context` is a RefContext value identifying
```

`_context`? Docs are stale?

frame/system/src/lib.rs
```diff
@@ -944,53 +1017,51 @@ impl<T: Trait> Happened<T::AccountId> for CallKillAccount<T> {
...
+		Account::<T>::get(k).data
 	}
 	fn is_explicit(k: &T::AccountId) -> bool {
 		Account::<T>::contains_key(k)
 	}
-	fn insert(k: &T::AccountId, t: T::AccountData) {
+	fn insert(k: &T::AccountId, data: T::AccountData) {
 		let existed = Account::<T>::contains_key(k);
-		Account::<T>::insert(k, (T::Index::default(), t));
+		Account::<T>::insert(k, AccountInfo{ nonce: Default::default(), refcount: 0, data });
```
shawntabrizi: Why is refcount returning to zero here?

The goal is just to insert some account data onto the user?

We should use the existing ref count...

frame/staking/src/lib.rs
```diff
@@ -1254,9 +1259,22 @@ decl_module! {
...
 		}
+
+		/// Remove all data structure concerning a staker/stash once its balance is zero.
+		/// This is essentially equivalent to `withdraw_unbonded` except it can be called by anyone
+		/// and the target `stash` must have no funds left.
+		///
+		/// This can be called from any origin.
+		///
+		/// - `stash`: The stash account to reap. Its balance must be zero.
+		fn reap_stash(_origin, stash: T::AccountId) {
```
shawntabrizi: How is this not an attack vector for a spam/ddos attack?

frame/recovery/src/lib.rs
```diff
@@ -160,8 +160,11 @@ use codec::{Encode, Decode};
 use frame_support::{
 	decl_module, decl_event, decl_storage, decl_error, ensure,
 	Parameter, RuntimeDebug,
-	weights::{GetDispatchInfo, SimpleDispatchInfo, FunctionOf},
-	traits::{Currency, ReservableCurrency, Get, OnReapAccount, BalanceStatus},
+	weights::{
+		GetDispatchInfo, PaysFee, DispatchClass, ClassifyDispatch, Weight, WeighData,
```
shawntabrizi: I think these deps can be cleaned up. Check for warnings

frame/balances/src/lib.rs
```diff
@@ -721,14 +731,20 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
 
-impl<T: Trait<I>, I: Instance> OnReapAccount<T::AccountId> for Module<T, I> {
-	fn on_reap_account(who: &T::AccountId) {
-		Locks::<T, I>::remove(who);
-		Account::<T, I>::remove(who);
+		let existed = Locks::<T, I>::contains_key(who);
+		if locks.is_empty() {
+			Locks::<T, I>::remove(who);
+			if existed {
+				// TODO: use Locks::<T, I>::hashed_key
```
```suggestion
				// TODO: use Locks::<T, I>::hashed_key
				// https://github.com/paritytech/substrate/issues/4969
```

frame/contracts/src/lib.rs
```diff
@@ -941,6 +941,9 @@ decl_storage! {
 	}
 }
 
+// TODO: this should be removed in favour of a self-destruct contract host function allowing the
+// contract to delete all storage and the `ContractInfoOf` key and transfer remaining balance to
+// some other account. As it stands, it's an economic insecurity on any smart-contract chain.
```
shawntabrizi:
```suggestion
// some other account. As it stands, it's an economic insecurity on any smart-contract chain.
// https://github.com/paritytech/substrate/issues/4952
```

frame/staking/src/lib.rs
gavofyork: how could it be?

shawntabrizi: What stops anyone from calling this function as many times as they want, not incur a fee, and force the blockchain to do some storage reads and other calculations?

frame/democracy/src/lib.rs
```diff
@@ -818,6 +866,30 @@ decl_module! {
...
+
+		/// Become a proxy.
+		///
+		/// This must be called prior to a later `activate_proxy`.
+		///
+		/// Origin must be a Signed.
+		///
+		/// - `target`: The account whose votes will later be proxied.
+		///
+		/// `close_proxy` must be called before the account can be destructed.
```
```suggestion
		/// `close_proxy` must be called before the account can be destroyed.
```

frame/staking/src/lib.rs
gavofyork: it's not a free transaction for anyone other than the block producer (it is for them, but if they want to waste their block, they'll incur an according opportunity cost).

frame/system/src/lib.rs
gavofyork: that's the goal, yes.

gavofyork: this looks indeed bad. good catch 👍

shawntabrizi:
```suggestion
		Account::<T>::mutate(k, |a| a.data = data);
```

This is what you want right?

NikVolf:
Sorry for spam, testing the bot, no more noise from me 🤓

NikVolf:
/bench

And few times more to see the variance, sorry again!

NikVolf:
/bench

Little more spam.. Still have to calibrate :)

frame/system/src/lib.rs
shawntabrizi: I am going to commit this suggestion.

bin/node/runtime/src/lib.rs
```diff
@@ -131,7 +131,7 @@ impl frame_system::Trait for Runtime {
 	type ModuleToIndex = ModuleToIndex;
 	type AccountData = pallet_balances::AccountData<Balance>;
 	type OnNewAccount = ();
-	type OnReapAccount = (Balances, Staking, Contracts, Session, Recovery);
+	type OnReapAccount = ();
```
kianenigma: from the other PR: 

> system::Trait is now the place to put any OnNewAccount or OnReapAccount hooks; these will need moving from balances::Trait. You should include _at least_ the Balances handler in the reap account

frame/balances/src/lib.rs
```diff
@@ -609,7 +610,16 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
-			put_storage_value(b"System", b"Account", &hash, (nonce, balances));
+			let mut refs: system::RefCount = 0;
+			// The items in Kusama that would result in a ref count being incremented.
+			if have_storage_value(b"Democracy", b"Proxy", &hash) { refs += 1 }
+			// We skip Recovered since it's being replaced anyway.
+			let mut prefixed_hash = twox_64(&b":session:keys"[..]).to_vec();
+			prefixed_hash.extend(&b":session:keys"[..]);
+			prefixed_hash.extend(&hash[..]);
+			if have_storage_value(b"Session", b"NextKeys", &prefixed_hash) { refs += 1 }
+			if have_storage_value(b"Staking", b"Bonded", &hash) { refs += 1 }
```
kianenigma: I understand that we don't have an easy and better choice here, but maybe we should document the rules of these references somewhere in a very strict way. The pattern in staking/session of `inc_ref`/`dec_red` reminds me of `malloc`/`free` and such, paradigms that we probably want to avoid.

bin/node/runtime/src/lib.rs
shawntabrizi: Only if you store your balances in the balances module I think.

If it is stored in the composite account, then you wont need to.

let me double check

frame/balances/src/lib.rs
```diff
@@ -1283,6 +1301,17 @@ impl<T: Trait<I>, I: Instance> ReservableCurrency<T::AccountId> for Module<T, I>
 	}
 }
 
+/// Implement `OnReapAccount` to remove the local account, if using local account storage.
+///
+/// NOTE: You probably won't need to use this! This only needs to be "wired in" to System module
+/// if you're using the local balance storage. **If you're using the composite system account
+/// storage (which is the default in most examples and tests) then there's no need.**
+impl<T: Trait<I>, I: Instance> OnReapAccount<T::AccountId> for Module<T, I> {
```
kianenigma: OK I see why https://github.com/paritytech/substrate/pull/4895/files#r382572970

frame/balances/src/lib.rs
shawntabrizi: Are these the kind of rules would you be looking for?

If a module wants to "keep alive" an account until it is done with it, it should put a reference counter on it.

When it is done with it, it should remove that reference counter.

It should only happen once each way. (But I guess there is nothing that would stop a module from managing multiple reference counters in both directions.)

frame/recovery/src/mock.rs
```diff
@@ -80,7 +80,7 @@ impl frame_system::Trait for Test {
 	type ModuleToIndex = ();
 	type AccountData = pallet_balances::AccountData<u128>;
 	type OnNewAccount = ();
-	type OnReapAccount = (Balances, Recovery);
+	type OnReapAccount = Balances;
```
kianenigma: Just a mock, but still raises the question: The balances is declared in this dummy runtime to use the system as storage. Does it need to be on then `OnReapAccount ` then?

shawntabrizi: You are probably right here. let me look

frame/session/src/lib.rs
```diff
@@ -612,10 +629,30 @@ impl<T: Trait> Module<T> {
...
+		Ok(())
+	}
+
+	/// Perform the set_key operation, checking for duplicates. Does not set `Changed`.
+	///
+	/// The old keys for this validator are returned, or `None` if there were none.
+	///
+	/// This does not ensure that the reference counter in system is incremented appropriately, it
+	/// must be done by the caller or the keys will be leaked in storage.
+	fn inner_set_keys(who: &T::ValidatorId, keys: T::Keys) -> Result<Option<T::Keys>, DispatchError> {
```
kianenigma: based on the doc maybe such functions should be called `xxx_unchecked` or sth that indicates the caller must be careful

frame/staking/src/lib.rs
kianenigma: Could use a reward/slash mechanism?

frame/system/src/lib.rs
```diff
@@ -935,7 +987,13 @@ impl<T: Trait> Happened<T::AccountId> for CallOnCreatedAccount<T> {
 pub struct CallKillAccount<T>(PhantomData<T>);
 impl<T: Trait> Happened<T::AccountId> for CallKillAccount<T> {
 	fn happened(who: &T::AccountId) {
-		Module::<T>::kill_account(who.clone());
+		if Account::<T>::contains_key(who) {
+			let account = Account::<T>::take(who);
+			if account.refcount > 0 {
+				sp_runtime::print("WARNING: Referenced account deleted. This is probably a bug.");
```
kianenigma: Looks important, would use `frame_support::debug` stuff with `target: System` so that it is filter-able.

frame/recovery/src/mock.rs
kianenigma: This was misleading then `type AccountStore = System;`

shawntabrizi: Ah I think I might be wrong then...

shawntabrizi: Yes, I removed them. thanks!

frame/system/src/lib.rs
shawntabrizi: good idea

frame/staking/src/lib.rs
shawntabrizi: probably something to follow up with. I dont see a trivial place to introduce a reward atm

NikVolf:
/bench

Some final spam :)

gavofyork:
concerning that the benchmark is consistently reporting that this branch is slower.

what is it actually benchmarking? this should be waaay faster (and i believe @shawn already showed that it was).

NikVolf:
> concerning that the benchmark is consistently reporting that this branch is slower.
> 
> what is it actually benchmarking? this should be waaay faster (and i believe @shawn already showed that it was).

Well, it is just 2-3% (it is inside benchmark variance, so don't pay much attention, until it shows > 5-6% difference)

This benchmark is just doing import of block with 100 random transfers, and I believe code path of this pr is not touched, I mostly tested the bot on pr more or less related to performance.

I will be happy to add benchmark that should show improvements on this, btw.

frame/balances/src/lib.rs
hirschenberger: @shawntabrizi Hey, did you mean using `hashed_key` like that to detect a change more efficiently?

```rust
...
		let old_key = Locks::<T, I>::hashed_key_for(who);
		if locks.is_empty() {
			Locks::<T, I>::remove(who);
			// account existed
			if old_key != Locks::<T, I>::hashed_key_for(who) {
				// TODO: use Locks::<T, I>::hashed_key
				// https://github.com/paritytech/substrate/issues/4969
				system::Pallet::<T>::dec_consumers(who);
			}
		} else {
			Locks::<T, I>::insert(who, bounded_locks);
			// account didn't exist
			if old_key == Locks::<T, I>::hashed_key_for(who) {
				if system::Pallet::<T>::inc_consumers(who).is_err() {
					// No providers for the locks. This is impossible under normal circumstances
					// since the funds that are under the lock will themselves be stored in the
					// account and therefore will need a reference.
```

PR #4898, opened 2020-02-12
serialize partial_fee into string

### LINKED ISSUE #4897: Transaction payment API cannot handle fee that is too high
https://github.com/paritytech/substrate/blob/00a400f82539e2f78e8ddbcd98aea512c87c5f3c/frame/transaction-payment/rpc/runtime-api/src/lib.rs#L54

This is causing issue for us.

On surface u64 supports value up to ~10^19, which is enough for most of the chains even if their token have 18 decimals (which we do).

But we are having issue that because the number is serialized as a number literal in JSON and `BN.js` will throw if the value is more than `Number.MAX_SAFE_INTEGER` which is `(2^53 -1)`. So the real save range that polkadot.js can handle is only up to ~10^16. That is a much more limited range.

We should instead serialize the `partial_fee` into a string, rather number. This will also allow it support u128 type or whatever the type `Balance` is.

xlc:
Fixes #4897 

Not sure if I need to an implementation with `deserialize_with`

This is kind of a breaking change but polkadot.js should automatically handle string to number conversion.

### DIFF
--- a/bin/node/runtime/src/lib.rs
+++ b/bin/node/runtime/src/lib.rs
@@ -82,7 +82,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
 	spec_version: 216,
-	impl_version: 1,
+	impl_version: 2,
 	apis: RUNTIME_API_VERSIONS,
 };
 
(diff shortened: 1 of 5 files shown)


bkchr:
The JS documentation clearly says:
```
The Number.MAX_SAFE_INTEGER constant represents the maximum safe integer in JavaScript (253 - 1).

For larger integers, consider using BigInt.
```

JSON is also not putting any limit on the maximum size of the integer.

You should probably switch your parser to support big ints.

bkchr:
@tomusdrw what do you think?

xlc:
@tomusdrw done

PR #4902, opened 2020-02-12
A Pallet for Benchmarking Common Runtime Operations

shawntabrizi:
This is a pallet which is meant purely to benchmark individual operations commonly performed throughout the runtime. It serves no logical purpose, and probably this PR should _not_ be merged in where the Benchmark pallet is included in the runtime.

Also introduces a small update to the `benchmark!` macro to allow for custom functions to be inserted directly into the string matching logic from the CLI so that fully custom benchmarks can be written parallel to the pre-designed benchmarking framework.

This is useful for timing non-extrinsic functions as is done here for `current_time` and `storage_root`.

Question: Is there a way to place a feature flag over this in the runtime?


frame/benchmark/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+		}
+
+		/// Insert a value into a map.
+		pub fn insert_map(_origin, repeat: u32) {
+			for r in 0..repeat {
+				MyMap::insert(r, r);
+			}
+		}
+
+		/// Insert a value into a map.
```
marcio-diaz: fix comments of contains*

frame/benchmark/src/benchmarking.rs
```diff
@@ -0,0 +1,465 @@
...
+	InsertMap,
+	ContainsKeyMap,
+	ContainsKeyHashMap,
+	ContainsKeyHashMap2,
+	RemovePrefix,
+	AddMemberList,
+	AppendMemberList
+}
+
+impl<T: Trait> Benchmarking<BenchmarkResults> for Module<T> {
```
marcio-diaz: Probably you don't need this function with the new version of `selected_benchmark!`

frame/benchmark/src/benchmarking.rs
```diff
@@ -0,0 +1,465 @@
...
+		let max_size = <ContainsKeyHashMap as BenchmarkingSetup<T, _, _>>::components(&ContainsKeyHashMap)[0].2;
+
+		sp_std::if_std! {
+			println!("{:?}", max_size);
+		}
+
+		for i in 0..n {
+			let hash = sp_io::hashing::twox_128(&i.encode()).into();
+			let num = u128::from_le_bytes(hash);
+			if i <= n {
```
marcio-diaz: I think this is always true.

shawntabrizi: yes, but I was just copying code from when it was always 0..1000.

frame/benchmark/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "pallet-benchmark"
+version = "2.0.0-alpha.2"
```
kianenigma:
```suggestion
version = "2.0.0-alpha.3"
```

frame/benchmark/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+license = "GPL-3.0"
+
+[dependencies]
+serde = { version = "1.0.101", optional = true }
+codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false, features = ["derive"] }
+enumflags2 = { version = "0.6.2" }
```
kianenigma:
```suggestion
```

frame/benchmark/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+sp-std = { version = "2.0.0-alpha.2", default-features = false, path = "../../primitives/std" }
```
kianenigma:
```suggestion
sp-std = { version = "2.0.0-alpha.3", default-features = false, path = "../../primitives/std" }
```

kianenigma: and all the rest.

frame/benchmark/src/benchmarking.rs
```diff
@@ -0,0 +1,159 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Identity pallet benchmarking.
```
kianenigma:
```suggestion
//! Benchmarking of common operations in runtime. 
```

frame/benchmark/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+	}
+}
+
+// The pallet's dispatchable functions.
+decl_module! {
+	/// The module declaration.
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		// Initializing errors
+		// this includes information about your errors in the node's metadata.
+		// it is needed only if you are using errors in your pallet
```
kianenigma: Some of these docs are probably pulled in unintentionally from https://github.com/paritytech/substrate/blob/5a2824d9b20f151466d92f20c767141c30a35b65/bin/node-template/pallets/template/src/lib.rs#L65 and such

frame/benchmark/src/lib.rs
```diff
@@ -0,0 +1,155 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+/// A runtime module template with necessary imports
```
kianenigma: Needs a minimal but proper module doc.

frame/benchmark/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+		fn deposit_event() = default;
+
+		/// Do nothing.
+		pub fn do_nothing(_origin, input: u32) {
+			if input > 0 {
+				return Ok(());
+			}
+		}
+
+		/// Read a value from storage value `repeat` number of times.
```
kianenigma: I can't read the benchmark macro and get it -- so some comments here: Is it considered that this will be fetched from the Overlay from the second time, onward? Or we actually want to benchamrk assuming that there is an overlay?

frame/benchmark/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+		pub fn read_value(_origin, repeat: u32) {
+			for _ in 0..repeat {
+				MyValue::get();
+			}
+		}
+
+		/// Put a value into a storage value.
+		pub fn put_value(_origin, repeat: u32) {
```
kianenigma: likewise.

frame/benchmark/src/lib.rs
kianenigma: basically my comment is that maybe the loop must be in the call site so you can wipe the overlay in between.

frame/benchmark/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+sp-io = { version = "2.0.0-alpha.2", default-features = false, path = "../../primitives/io" }
+sp-runtime = { version = "2.0.0-alpha.2", default-features = false, path = "../../primitives/runtime" }
+frame-support = { version = "2.0.0-alpha.2", default-features = false, path = "../support" }
+frame-system = { version = "2.0.0-alpha.2", default-features = false, path = "../system" }
+frame-benchmarking = { version = "2.0.0-alpha.2", default-features = false, path = "../benchmarking" }
+
+[dev-dependencies]
```
kianenigma:
```suggestion
```

frame/benchmark/src/lib.rs
shawntabrizi: Yes, it is considered fetched from the overlay the second time around.

To benchmark multiple "first-time-calls", we use the repeat feature at the benchmark macro level.

kianenigma: Okay just wanted to make sure we are aware of this.

frame/benchmark/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+
+		// Append user to the list.
+		pub fn append_member_list(origin) {
+			let who = ensure_signed(origin)?;
+			MyMemberList::<T>::append(&[who])?;
+		}
+
+		// Append user to the list.
+		pub fn encode_accounts(_origin, accounts: Vec<T::AccountId>) {
+			let _bytes = accounts.encode();
```
bkchr: Are you sure that isn't optimized away?

frame/benchmark/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		// Append user to the list.
+		pub fn decode_accounts(_origin, bytes: Vec<u8>) {
+			let _accounts: Vec<T::AccountId> = Decode::decode(&mut bytes.as_slice()).map_err(|_| "Could not decode")?;
```
bkchr: Same question.

frame/benchmark/src/lib.rs
shawntabrizi: We saw the jumps at powers of two remember?
https://www.shawntabrizi.com/substrate-graph-benchmarks/?p=pallet-benchmark&e=encode_accounts

Is there some trivial syntax I could add to make you feel better here?

frame/benchmark/src/lib.rs
shawntabrizi: same general answer, the benchmark gave good results afaik: https://www.shawntabrizi.com/substrate-graph-benchmarks/?p=pallet-benchmark&e=decode_accounts

frame/benchmark/src/lib.rs
bkchr: We could maybe add some check
```
if bytes.empty() {
    panic!("Will never happen, but the compiler can not know this");
}
```

shawntabrizi: turned it into a print statement

gnunicorn:
@shawntabrizi please merge master and resolve the conflict.

gnunicorn:
@shawntabrizi looks like something went wrong, CI tests failing.

PR #4903, opened 2020-02-12
Adds a test to ensure that we clear the heap between calls into runtime

bkchr:
The tests shows that we currently not clearing the heap in wasmtime.
For now we don't run the test for wasmtime.


client/executor/common/src/wasm_runtime.rs
```diff
@@ -34,4 +34,7 @@ pub trait WasmRuntime {
 
 	/// Call a method in the Substrate runtime by name. Returns the encoded result on success.
 	fn call(&mut self, method: &str, data: &[u8]) -> Result<Vec<u8>, Error>;
+
+	/// Get the value from a global with the given `name`.
+	fn get_global_val(&mut self, name: &str) -> Result<Option<Value>, Error>;
```
arkpar: We tend not use `get_` prefix. Rename it to `global_val` perhaps?

arkpar:
Is this really intended behaviour? I'd expect runtime instance to keep all of its state between calls. This might not be what we want in substrate, but it makes more sense for wasmi in general. It would make more sense to have an explicit `clear_heap()` or even `reset()` that would reset not only heap, but global variables as well.

bkchr:
> Is this really intended behaviour? I'd expect runtime instance to keep all of its state between calls. This might not be what we want in substrate, but it makes more sense for wasmi in general. It would make more sense to have an explicit `clear_heap()` or even `reset()` that would reset not only heap, but global variables as well.

Yes that is intended behavior, we clear the memory between calls.

PR #4517, opened 2020-01-02
Offchain Phragmén BREAKING.

kianenigma:
Fix https://github.com/paritytech/substrate/issues/4410

---

Currently, Kusama is suffering from severe stalls at the end of each era. Seemingly validators keeep missing their slot, which makes whatever we do in `new_era` a suspect. Phragmén election is one of them. I have made some attempts (which I hope to finish soon as reusable tools) to benchmark the wasm code but hadn't got any decisive conclusion yet. Nonetheless, using https://github.com/kianenigma/offline-phragmen I could verify that the current Kusama network (~200 candidates, ~600 voters) takes around 0.5 seconds just to compute the election result in _native_. In wasm, and taking the overlooked overhead of state lookup, hashing and decoding, it is enough to prioritise offchain phragmen as it is obviously hindering the block's limited 2 seconds. 

In the offchain model: 
- A certain number of blocks before the era is expected to end, the staking module enables a flag that allows for solutions to be submitted. 
- All validators who opt to run the offchain code start computing phragmen and submit the solution back to the chain.
- Any other account can also submit a solution. 
- The staking module evaluates and stores only the best result. 
- If for any reason, at the end of the era, no solutions have been submitted, we fallback to the current system: on-chain phragmen. 

#### In this PR 

- [x] A way for staking to know when era ends.
- [x] Main skeleton of choosing between on/off chain solutions
- [x] Signing the transaction from offchain @tomusdrw (waiting for #4200)
- [x] Solution _reduction_
- [x] Solution encoding and decoding
- [x] Solution evaluation
- [x] Allow submission from also unsigned (for validators)
- [x] clean TODOS
- [x] A metric ton of tests. 
- [x] Real testing on dev chain. 
- [x] Benchmark the runtime side of things.
- [x] Update hashes if needed https://github.com/paritytech/substrate/pull/5226/files

##### In follow ups:

- ~~Staking should freeze and not allow any changes to sensitive data when election window is open.~~ (already done) 
- Clean the reduce code. Add index cache (+ other optimizations) Add benchmarks and keep track of performance. This is not critical as that code will run offchain. 
- Run `equalize()` again.
-  rewards? punishment? 

### Discussions

- The key that validators sign the transaction with would be the ephemeral session key AFAIK, how would they pay for it? **Update: Unsigned support added**
- Maybe disable equalisation and only do seq-phragmen in fallback mode. **Update: already done**
[...]

PR #4904, opened 2020-02-12
Per-things trait.

kianenigma:
Makes `Per-XXX` implement a trait for their common functionality. 

Used in `sp-phragmen`, so henceforth the caller of Phragmén can specify their desired accuracy of results. This is needed for https://github.com/paritytech/substrate/pull/4517.  Making a separate PR for the sake of easy review.


primitives/arithmetic/src/per_things.rs
```diff
@@ -71,48 +114,73 @@ macro_rules! implement_per_thing {
...
+			///
+			/// This can be called at compile time.
+			pub const fn from_parts_const(parts: $type) -> Self {
+				Self([parts, $max][(parts > $max) as usize])
+			}
+
+			/// Converts a percent into `Self`. Equal to `x / 100`.
+			///
+			/// This can be created at compile time.
+			pub const fn from_percent_const(x: $type) -> Self {
```
bkchr: Why not just call it `from_percent`? The compiler should be able to pick the correct one.

primitives/arithmetic/src/per_things.rs
```diff
@@ -71,48 +114,73 @@ macro_rules! implement_per_thing {
...
+				$name(part as Self::Inner)
+			}
+		}
+
+		/// Implement const functions
+		impl $name {
+			/// From an explicitly defined number of parts per maximum of the type.
+			///
+			/// This can be called at compile time.
+			pub const fn from_parts_const(parts: $type) -> Self {
```
bkchr: Same.

primitives/arithmetic/src/traits.rs
```diff
@@ -28,17 +28,9 @@ use sp_std::ops::{
...
 };
 
-/// A meta trait for arithmetic.
-///
-/// Arithmetic types do all the usual stuff you'd expect numbers to do. They are guaranteed to
-/// be able to represent at least `u32` values without loss, hence the trait implies `From<u32>`
-/// and smaller ints. All other conversions are fallible.
-pub trait SimpleArithmetic:
+/// A meta trait for arithmetic type operations, regardless of any limitation on size.
+pub trait BaseArithmetic:
```
bkchr: These names

primitives/arithmetic/src/traits.rs
```diff
@@ -65,8 +53,55 @@ impl<T:
...
 	Saturating + PartialOrd<Self> + Ord + Bounded +
 	HasCompact + Sized
+> BaseArithmetic for T {}
+
+/// A meta trait for arithmetic.
+///
+/// Arithmetic types do all the usual stuff you'd expect numbers to do. They are guaranteed to
+/// be able to represent at least `u32` values without loss, hence the trait implies `From<u32>`
+/// and smaller ints. All other conversions are fallible.
+pub trait SimpleArithmetic:
```
bkchr: get out

primitives/arithmetic/src/traits.rs
```diff
@@ -65,8 +53,55 @@ impl<T:
...
+	From<u8> + From<u16> + From<u32> + TryInto<u8> + TryInto<u16> + TryInto<u32> + TryFrom<u64> +
+	TryInto<u64> + TryFrom<u128> + TryInto<u128> + TryFrom<usize> + TryInto<usize> +
+	UniqueSaturatedInto<u8> + UniqueSaturatedInto<u16> + UniqueSaturatedInto<u32> +
+	UniqueSaturatedFrom<u64> + UniqueSaturatedInto<u64> + UniqueSaturatedFrom<u128> +
+	UniqueSaturatedInto<u128>
 > SimpleArithmetic for T {}
 
+/// Same as [`SimpleArithmetic`] for integer types of any size, without any assumption of their size
+/// being at least `u32`.
+pub trait MiniArithmetic:
```
bkchr: of control.

primitives/arithmetic/src/traits.rs
bkchr: So, `SimpleArithmetic` is already somewhat not that expressive. But splitting this now into 3 traits who all are doing basically the same. 

We should up with something better.

1. Have one trait, probably just called `Arithmetic`.
2. `MiniArithmetic` is just a subset of `SimpleArithmetic` or? We probably can just merge `Arithmetic` and `MiniArithmetic`.
3. `SimpleArithmetic` should get some better name, IDK :D The name should give a clear indication what the difference of the trait is.

kianenigma: 1. Agree that the `Simple` prefix is not cool. 

2. Mini Arithmetic is _not_ a subset. it is actually a superset logically, since it has the same operations but can be any size, while Simple Arithmetic can be u32 or bigger. But this is not visible trait-wise. Neither can be composed of the other one, at least in the current format. 

Hence, I created the third one:

- `BasicArithmetic` is just operations with no assumption on size
- `Mini` and `Simple` can go on an _compose_ it + add limitation on size. 

Aside from the fact that `Simple` is not a nice prefix, I think this is quite nice. dk what you would say it is out of control.

bkchr: 2. Yeah, I actually meant it the way you described it :D 

`Mini` isn't a nice prefix as well :P Maybe `U8Arithmetic` + `U32Arithmetic`? To make it clear on reading?

kianenigma: ahh and the fact that the prefix `Mini` also doesn't make much sense anymore :D 

How about this 

`RuntimeArithmetic` for the one which is u32 and above, since we also use it for blocknumber etc.
`Arithmetic` for the one which is generic on size.

primitives/arithmetic/src/traits.rs
```diff
@@ -65,7 +53,54 @@ impl<T:
...
 	HasCompact + Sized
-> SimpleArithmetic for T {}
+> BaseArithmetic for T {}
+
+/// A meta trait for arithmetic.
+///
+/// Arithmetic types do all the usual stuff you'd expect numbers to do. They are guaranteed to
+/// be able to represent at least `u32` values without loss, hence the trait implies `From<u32>`
+/// and smaller ints. All other conversions are fallible.
+pub trait RuntimeArithmetic:
```
bkchr: I think we should call it `AtLeast32Bit`. The name `RuntimeArithmetic` doesn't make sense here, as this is nothing we enforce in the runtime.

primitives/phragmen/src/lib.rs
```diff
@@ -272,62 +277,68 @@ pub fn elect<AccountId, Balance, FS, C>(
...
 							// scores, hence MUST have the same denominator.
 							Zero::zero()
 						}
 					}
 				};
-				// safer to .min() inside as well to argue as u32 is safe.
-				let per_thing = Perbill::from_parts(
-					per_bill_parts.min(Perbill::accuracy().into()) as u32
-				);
+				let per_thing = R::from_parts(per_bill_parts);
```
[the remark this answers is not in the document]
kianenigma: `from_parts` will do a `min` internally.

primitives/arithmetic/src/traits.rs
```diff
@@ -65,7 +53,54 @@ impl<T:
...
+	From<u8> + From<u16> + From<u32> + TryInto<u8> + TryInto<u16> + TryInto<u32> + TryFrom<u64> +
+	TryInto<u64> + TryFrom<u128> + TryInto<u128> + TryFrom<usize> + TryInto<usize> +
+	UniqueSaturatedInto<u8> + UniqueSaturatedInto<u16> + UniqueSaturatedInto<u32> +
+	UniqueSaturatedFrom<u64> + UniqueSaturatedInto<u64> + UniqueSaturatedFrom<u128> +
+	UniqueSaturatedInto<u128>
+> RuntimeArithmetic for T {}
+
+/// Same as [`RuntimeArithmetic`] for integer types of any size, without any assumption of their size
+/// being at least `u32`.
+pub trait AnyArithmetic:
```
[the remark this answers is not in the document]
kianenigma: Let me see if I get you right: 

What you recommend is putting all the `TryInto<_>`, `TryFrom<_>`, `UniqueSaturatedInto<_>` and `UniqueSaturatedFrom<_>`s in `Base` for all the 6 major integer types, since we don't know the size there, right? and then we can say `AtLeast32Bit: Base + From<u8> + From<u16> + From<u32>`.

The only minor problem with that would be that some of the traits would be duplicate and don't really make sense. I.e. a type ends up being `From<u32> + TryFrom<u32>`. Is this cool?

primitives/arithmetic/src/traits.rs
kianenigma: `32BitArithmetic`?

bkchr: That is not right, because we are not doing `32bit` arithmetic here :D We just add `From<u32>` and `From<u16>`. That does say anything about the arithmetic.

primitives/arithmetic/src/traits.rs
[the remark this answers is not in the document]
bkchr: 1. Yeah base takes all these trait bounds and I also would give it `From<u8>` (there is no datatype that is smaller than u8)
2. Every type implements `From<Self>` and also 
```
impl<T, U> TryFrom<U> for T where
    U: Into<T>, 
```
   This means that every type automatically implements `TryFrom` for all types it implements `From` for.

primitives/phragmen/src/lib.rs
```diff
@@ -272,62 +277,68 @@ pub fn elect<AccountId, Balance, FS, C>(
...
+							let parts = multiply_by_rational(
 								desired_scale,
 								e.load.n(),
 								n.load.n(),
-							).unwrap_or(Bounded::max_value())
+							)
+							// If result cannot fit in u128. Not much we can do about it.
+							.unwrap_or(Bounded::max_value());
+
+							<R::Inner as TryFrom<ExtendedBalance>>::try_from(parts)
```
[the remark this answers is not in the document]
kianenigma: it is 👍

primitives/arithmetic/src/traits.rs
kianenigma: I am not fully happy with `AtLeast32Bit ` but we'll go with it for now since I cannot come up with any other name :D 🤷🏻‍♂️

bkchr: Yeah me neither :D But I also don't have any better idea :(

Issue #4899, opened 2020-02-12
Staging chain panic


everhusk:
Clone the master branch and run the chain in staging and it gives the following error:

2020-02-11 21:03:48 Substrate Node
2020-02-11 21:03:48   version 2.0.0-00a400f82-x86_64-macos
2020-02-11 21:03:48   by Parity Technologies <admin@parity.io>, 2017-2020
2020-02-11 21:03:48 Chain specification: Staging Testnet
2020-02-11 21:03:48 Node name: rebel-idea-8685
2020-02-11 21:03:48 Roles: FULL

====================

Version: 2.0.0-00a400f82-x86_64-macos

stack backtrace:
   0: backtrace::backtrace::trace
   1: backtrace::capture::Backtrace::new
   2: sp_panic_handler::set::{{closure}}
   3: std::panicking::rust_panic_with_hook
   4: rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::slice::slice_index_len_fail
   7: node_cli::chain_spec::testnet_genesis
   8: node_cli::chain_spec::staging_testnet_config_genesis
   9: core::ops::function::Fn::call
  10: <sc_chain_spec::chain_spec::ChainSpec<G,E> as sp_runtime::BuildStorage>::build_storage
  11: sc_client::client::Client<B,E,Block,RA>::new
  12: sc_client_db::new_client
  13: sc_service::builder::ServiceBuilder<(),(),TGen,TCSExt,(),(),(),(),(),(),(),(),(),()>::new_full
  14: node_cli::service::new_full
  15: sc_cli::runtime::run_service_until_exit
  16: sc_cli::run_node
  17: sc_cli::params::RunCmd::run
  18: sc_cli::run
  19: node_cli::command::run
  20: substrate::main
  21: std::rt::lang_start::{{closure}}
  22: std::panicking::try::do_call
  23: __rust_maybe_catch_panic
  24: std::rt::lang_start_internal
  25: main

Thread 'main' panicked at 'index 3 out of range for slice of length 1', src/libcore/slice/mod.rs:2674

This is a bug. Please report it at:

	https://github.com/paritytech/substrate/issues/new

PR #4906, opened 2020-02-12
Fix chain-spec and make sure it does not breaks again

bkchr:
Fixes: https://github.com/paritytech/substrate/issues/4899


bin/node/cli/src/chain_spec.rs
```diff
@@ -295,7 +295,8 @@ pub fn testnet_genesis(
 		pallet_membership_Instance1: Some(Default::default()),
 		pallet_treasury: Some(Default::default()),
 		pallet_society: Some(SocietyConfig {
-			members: endowed_accounts[0..3].to_vec(),
+			members: endowed_accounts.iter().cloned()
```
kianenigma: could work a bit nicer with `take()`?

bkchr: Good idea! :D

PR #4907, opened 2020-02-12
pallet-evm: add support for transaction-level create2

sorpaas:
This allows CREATE2 to be directly called as transactions. Previously this must be done through factory contracts.


shawntabrizi:
Is it possible to start writing tests for the EVM pallet?

frame/evm/src/lib.rs
```diff
@@ -378,6 +390,76 @@ decl_module! {
 
 			ret.map_err(Into::into)
 		}
+
+		/// Issue an EVM create2 operation.
+		#[weight = WeightForCallCreate]
```
kianenigma: You can use `weight::FunctionOf` here to make your life easier :D

sorpaas: Fixed!

sorpaas:
@shawntabrizi Let me do it in a separate PR.

Issue #4587, opened 2020-01-10
Deferred Service Builder


gnunicorn:
Right now, the `ServiceBuilder` creates the client, backend and all else right at the beginning and executes the configured functions directly. This has some implicatations, we do not want:
1.  you can out the ServiceBuilder into an unexpected/unrecommended State by calling a service method multiple times with different callbacks. As these callbacks typically have side-effects, this can lead to inconsistent and unreliable cases–even to race conditions if, for example, an authoring mechanism was initiated.
1. some of these functions actually conflict and shouldn''t both be called. However right now, we have no way of preventing that. 
1. It is up to the user to call the functions in the right order, which is unreliable and an unnecessary foodgun – especially if we had to change that order.
1. We can't provide sane defaults – like the longest chain selector or the light client transaction pool, both are the same for all known usages of our node-template – without requiring to the do 1 and cause unexpected side effects.
1. Certain aspects, like the transaction pool or externality extensions, are actually created during client creation and thus have a hacky API so we can change them after the fact, see #4583.

Instead the pattern should only keep the references to the `fn`s and execute all in one go on `.build` only. This would allow us to provide defaults, enforce the order in which set up takes place and would ensure any `fn` is only called once.

Further more this might allow us to clean up the api, so you can have one global builder and then call `build_light` and `build_full` on them rather than having to specify two builder...

tomaka:
The original idea behind the `ServiceBuilder` is that the callbacks would be completely removed.
It doesn't make sense to me that the user has to call a function and pass a closure, only for this closure to be immediately called by that function and that's it.

After all, the parameters that the builder passes to the stack are very arbitrary. If for instance you want to use an implementation of a `TxPool` that requires a reference to the network, you would need to modify the `ServiceBuilder` to pass the `NetworkService` to the closure that creates the `TxPool`.

I'm not sure how feasible that is, but my original idea was to create everything on the stack, and then pass that to the `ServiceBuilder` as values. Without having any closure/callback involved.

cecton:
This is mine now

![Cat owning food](https://yogizilla.files.wordpress.com/2014/04/there-now-its-mine-cat-meme-imgur.png)

cecton:
I got an idea. Let me know if that sounds good:

- I realized that what the ServiceBuilder will do after the change is: a. create a struct with whatever you pass to it; b. provide functions to add all the builders; c. and finally allow you to build() the service or client and the queue (which is the relevant part).
[...]

PR #4909, opened 2020-02-12
Open one substream for each notifications protocol

tomaka:
In #4284 we introduced a new API for the networking crate, where one can register a "notifications protocol". The implementation, however, was simply translating notifications into old messages and sending them the same way as before.

This PR is the next step.
In addition to the current unique substream with the remote, we now also try to open a separate substream for each notifications protocol that has been registered.
If the remote doesn't support the new protocol, we don't care. If the remote *does* support the new protocol, then we use it to send our notifications.

(Note: connections that are already open when a new protocol is registered will not use the new protocol. This really should be fixed eventually by forbidding registering new protocols after the initialization process (cc #4587 ).)

Right now the only protocol concerned is GrandPa. I went with `/paritytech/grandpa/1` for the protocol name.
cc'ing @rphmeier @andresilva 

## New protocol description

The new substreams have the following properties:

- They are unidirectional w.r.t notifications. If A wants to send notifications to B, it needs to open a substream. If then B wants to send notifications to A, then it needs to open its own substream for that. It is encouraged, but not mandatory, for B to open a substream with A if A opens one with B. The reason is to avoid collisions if both sides want to open a substream at the same time.
- After A opens a substream with B, it starts by sending an initial message. B then needs to send back a handshake message. A can then start sending notifications. B can refuse the substream (if it is full) by closing the substream without replying back. This handshake isn't accessible from the public API yet.
- It is a protocol error to open another substream for a given protocol while one already exists.

## Backwards compatibility

This PR is fully backwards compatible with the current network.
Also note that if we receive a notification the "old way", we still accept it even if the new protocol is in use.

Right now the legacy protocol is still the only way to perform syncing and light client requests, and therefore we make the presence of the legacy substream act as "the authority".
In other words, we are considered "connected" to a node if and only if we have a legacy substream open with that node. The new substreams exist in parallel but are optional. Also note that it is not possible to open a new substream without opening the legacy substream as well, but this will change in the future.

However **nodes that have this PR will immediately talk to each other using the new protocol**. It should therefore be strongly tested before it is deployed "for real" on Kusama.


tomaka:
I just remembered that at the moment GRANDPA uses the legacy substream handshake to determine whether the remote is a full node or an authority.

This "role" should instead be passed as part of the handshake. Before we merge, I will expose this handshake API and make it pass the role for GRANDPA.

client/network/src/protocol/generic_proto/upgrade/notifications.rs
```diff
@@ -0,0 +1,617 @@
...
+	/// Remote doesn't process our messages quickly enough.
+	///
+	/// > **Note**: This is not necessarily the remote's fault, and could also be caused by the
+	/// >			local node sending data too quickly. Properly doing back-pressure, however,
+	/// > 			would require a deep refactoring effort in Substrate as a whole.
+	Clogged,
+}
+
+/// Reads a variable-length integer from the `socket`.
+pub async fn read_varint(socket: &mut (impl AsyncRead + Unpin)) -> Result<usize, io::Error> {
```
twittner: I have revived https://github.com/paritytech/unsigned-varint/pull/16 as https://github.com/paritytech/unsigned-varint/pull/24 to directly support this use case in unsigned-varint.

client/network/src/protocol/generic_proto/handler/group.rs
```diff
@@ -0,0 +1,523 @@
...
+
+/// Implements the `IntoProtocolsHandler` trait of libp2p.
+///
+/// Every time a connection with a remote starts, an instance of this struct is created and
+/// sent to a background task dedicated to this connection. Once the connection is established,
+/// it is turned into a [`NotifsHandler`].
+///
+/// See the documentation at the module level for more information.
+pub struct NotifsHandlerProto {
+	/// Prototypes for handlers for ingoing substreams.
```
mxinden:
```suggestion
	/// Prototypes for handlers for incoming substreams.
```

client/network/src/protocol/generic_proto/handler/group.rs
```diff
@@ -0,0 +1,523 @@
...
+
+	/// Prototype for handler for backwards-compatibility.
+	legacy: LegacyProtoHandlerProto,
+}
+
+/// The actual handler once the connection has been established.
+///
+/// See the documentation at the module level for more information.
+pub struct NotifsHandler {
+	/// Handlers for ingoing substreams.
```
mxinden:
```suggestion
	/// Handlers for incoming substreams.
```

client/finality-grandpa/src/communication/mod.rs
```diff
@@ -185,7 +186,7 @@ impl<B: BlockT, N: Network<B>> NetworkBridge<B, N> {
 		);
 
 		let validator = Arc::new(validator);
-		let gossip_engine = GossipEngine::new(service.clone(), GRANDPA_ENGINE_ID, validator.clone());
+		let gossip_engine = GossipEngine::new(service.clone(), GRANDPA_ENGINE_ID, GRANDPA_PROTOCOL_NAME, validator.clone());
```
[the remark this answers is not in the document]
tomaka: It is also used in non-networking parts of the code.

PR #4910, opened 2020-02-12
Cache CARGO_TARGET_DIR

TriplEight:
`CARGO_TARGET_DIR` dir is now shared and available for the same job on the same branch.
This drastically speeds up the CI enabling to reuse build artifacts from a previous job run. As on your local machine.
`CARGO_INCREMENTAL=0` is not needed any more since now we are able to benefit from incremental compilations. Previously it was turned to 0 to get advantage caching those files with `sccache`.
- scheme is `/ci-cache/${CI_PROJECT_NAME}/targets/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}` is a compulsory measure because of both security (to not allow someone to spoil your cache) and to avoid data-race conditions which are very much a thing there if you try accessing the same cache from multiple sides.
- first CI_JOB_NAME of the new branch on a runner will hardlink to the same CI_JOB_NAME on the `master` branch and reuse a big chunk of cache from it. This does not compromise the security of the cache, saving more than 60% of disk space and time for the first build. The only downside is that the data race with `master` can happen, but this is unlikely and quite easily fixable. 
- CI servers periodically clean up old files, FIFO.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -17,8 +17,7 @@ variables:
   GIT_STRATEGY:                    fetch
   GIT_DEPTH:                       100
   CARGO_HOME:                      "/ci-cache/${CI_PROJECT_NAME}/cargo/${CI_JOB_NAME}"
-  SCCACHE_DIR:                     "/ci-cache/${CI_PROJECT_NAME}/sccache"
-  CARGO_INCREMENTAL:               0
+  CARGO_TARGET_DIR:                "/ci-cache/${CI_PROJECT_NAME}/targets/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
   CI_SERVER_NAME:                  "GitLab CI"
   DOCKER_OS:                       "debian:stretch"
   ARCH:                            "x86_64"
@@ -44,6 +43,8 @@ variables:
     - rustup show
     - cargo --version
     - sccache -s
+    - mkdir -p ${CARGO_HOME}; touch ${CARGO_HOME}/config
+    - mkdir -p ${CARGO_TARGET_DIR}
   only:
     - master
     - /^v[0-9]+\.[0-9]+.*$/        # i.e. v1.0, v2.1rc1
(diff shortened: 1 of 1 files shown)


NikVolf:
There should be a way to clean it for a developer.

Sometimes things go wrong with artifacts :/

bkchr:
> - first pipeline on the branch builds the cache from scratch, so it's beneficial for consequent builds only.

Hmm, is there no way to use the cache from another branch for the first pipeline on a new branch?

TriplEight:
> Hmm, is there no way to use the cache from another branch for the first pipeline on a new branch?
Yes, there's a way and right now I'm working on it.

TriplEight:
> There should be a way to clean it for a developer.
> Sometimes things go wrong with artifacts :/

Of course, it can happen, but not very often. 
Just temporarily add `cargo clean` at the start of the job you want or to the template job https://github.com/paritytech/substrate/blob/master/.gitlab-ci.yml#L40

gabreal:
>     * the `master` branch and reuse a big chunk of cache from it. This does not compromise the security of the cache, saving more than 60% of disk space and time for the first build.

Can you explain further why not? Since any pr will have access to the master branches build fragments it could also write to them.

bkchr:
> > ```
> > * the `master` branch and reuse a big chunk of cache from it. This does not compromise the security of the cache, saving more than 60% of disk space and time for the first build.
> > ```
> 
> Can you explain further why not? Since any pr will have access to the master branches build fragments it could also write to them.

@TriplEight ?

TriplEight:
This research is still undergoing.
Hard linking did not work out as `cargo` instead of rewriting writes to the same files (inodes).

We are about to test it with `btrfs` and `zfs`, it should work out.

In the meanwhile, we can prepopulate the cache with a simple copy of any other up to date branch on the machine, not necessarily from `master`.

bkchr:
I think with the current problems, this can be closed for now.

TriplEight:
it has nothing to do with `CARGO_HOME` caching bug, but can be closed for now.

PR #4911, opened 2020-02-12
Implement multi account module

### LINKED ISSUE #4834: Implement multisigs with dynamic signatories
The current multisig implementation is a great start, but it derives the AccountId from a fixed set of signatories and a threshold. 

That design has a couple of limitations, e. g.: 

1. When using multisigs for token holdings and changing signatories or the threshold, the tokens need to be transferred to a new AccountId, which might be a taxable event, depending on jurisdiction. Even if it's not, it complicates communication with accountants/tax advisors.
2. When using multisigs for staking and changing signatories or the threshold, one needs to unbond, transfer tokens and stake again. That can lead to months of lost staking rewards, depending on the unbonding time. 
3. When using multisigs for voting and changing signatories or the threshold, one needs to remove votes, transfer tokens and vote again, which is cumbersome.

These limitations could be addressed by changing the design of multisigs to allow dynamic threshold and signatories, where members are stored on-chain.

Is there an interest from other teams on such an implementation? 
Are there any trade-offs I'm missing here?

Many thanks!

philipstanislaus:
# Multi Account Module

A module with multisig functionality with a static `AccountId` and a dynamic `threshold` and set of `signatories`.

## Overview

This module contains functionality to create, update and remove multi accounts, identified by a deterministicallygenerated account ID, as well as (potentially) stateful multisig dispatches. 

This module is closely implemented after the multisig functionality of the utility pallet. The main difference is that this module stores static account ids on chain which are not changing when the threshold or signatories are updated. 

This module allows multiple signed origins (accounts) to coordinate and dispatch calls from a previously created multi account origin. When dispatching calls, the threshold defined in the multi account defines the number of accounts from the multi account signatory set that must approve it. In the case that the threshold is just one then this is a stateless operation. This is useful for multisig wallets where cryptographic threshold signatures are not available or desired or where a dynamic set of signatories with a static account ID is desired.

The motivation for this module is that a deterministically created account ID from threshold and a set of
signatories as in the utility pallet has certain downsides:

1. When using multisigs for token holdings and changing signatories or the threshold, the tokens need to be transferred to a new account ID, which might be a taxable event, depending on jurisdiction. Even if it's not, it complicates communication with accountants/tax advisors.
2. When using multisigs for staking and changing signatories or the threshold, the account needs to unbond, transfer tokens and re-stake again. That can lead to months of lost staking rewards, depending on the unbonding time.
3. When using multisigs for voting and changing signatories or the threshold, the accounts needs to unvotes, transfer tokens and vote again, which is cumbersome.

The downside of using this multi account module is that it is slightly more complex than the utility multisigs and that it introduces additional state and the requirement of creating multi accounts before using multisigs.

#### For multi account management

* `create` - Create a new multi account with a threshold and given signatories.
* `update` - Update an existing multi account with a new threshold and new signatories.
* `remove` - Remove an existing multi account.

#### For multisig dispatch

* `call` - Approve and if possible dispatch a call from a multi account origin.
* `approve` - Approve a call from a multi account origin.
* `cancel` - Cancel a call from a multi account origin.

## Edge Cases

Update:
Multi account thresholds and signatories can only be updated and removed if all active/ongoing multisig operations have been cancelled. That eliminates all of the previous edge cases outlined below. To prevent an outgoing multi account member to block updates/removals, the multi account itself can cancel any active multisig.

| | ~~Approve/dispatch multisig~~ | ~~Cancel multisig~~ |
| ---- | ---- | ---- |
| ~~1. Multisig owner no longer in signatories~~ | ~~Only current signatories should count for approvals~~ | ~~Owner should still be able to cancel~~ |~~~~
| ~~2. Multisig approver no longer in signatories~~ | ~~Only current signatories should count for approvals~~ | ~~Owner should be able to cancel~~ |~~~~
| ~~3. Multi account removed~~ | ~~Approvals/dispatch should not be possible anymore~~ | ~~Owner should still be able to cancel~~ |

1. ~~Multi account changed: Multisig owner no longer in signatories~~

    - [x] ~~The owner should still be able to cancel the multisig, even after he left the signatories of the multi account, to get back their deposit.~~
    - [x] ~~The multisig can still be approved and dispatched, even if the owner is no longer one of the signatories, as long as the number of approvals through the current signatories is >= the threshold. The original owner should get back their deposit.~~

2. ~~Multi account changed: Multisig approver no longer in signatories~~

    - [x] ~~At the time of dispatch, only approvals from current signatories should be counted, not from signatories that are no longer part of the multi account. (Can be done either by filtering the approval list immediately before the dispatch, or by updating all the approvals of all multisigs when updating the multi account's signatories. The former is prefrrable, since that search is bounded by the signatory size, while updating all multisigs is unbounded.).~~
    - [x] ~~The multisig can still be approved and dispatched, even if one of the approvers is no longer in the signatory set, as long as the number of approvals through the current signatories is >= the threshold. ~~

3. ~~Multi account removed~~

    - [x] ~~Multisigs should not be allowed to be approved and/or dispatched after a multi account has been approved.~~
    - [x] ~~Multisigs should still be allowed to be cancelled by the owner, even if a multi account has been removed.~~

## Questions

- [ ] Are there any security implications of using just a string and a nonce for multi account id generation? If yes, how can this be improved?
- [x] Is a PR to https://github.com/polkadot-js/api necessary (cc @jacogr)?

## Misc

Closes #4834.


jacogr:
No PR there needed. But if accepted, there will be some work to add the required types for proper support. (Help there will be appreciated, although not required)

frame/utility/src/lib.rs
```diff
@@ -609,7 +609,7 @@ impl<T: Trait> Module<T> {
 	///
 	/// NOTE: `who` must be sorted. If it is not, then you'll get the wrong answer.
 	pub fn multi_account_id(who: &[T::AccountId], threshold: u16) -> T::AccountId {
-		let entropy = (b"modlpy/utilisuba", who, threshold).using_encoded(blake2_256);
+		let entropy = (b"modlpy/multisuba", who, threshold).using_encoded(blake2_256);
```
philipstanislaus: This breaks existing functionality. Is this a typo? It's the same prefix that is used in `sub_account_id` in line 602.

philipstanislaus:
Added a few notes on missing tests/edge cases above.

frame/utility/src/lib.rs
joshua-mir: if there are any multisigs deployed on Kusama, changing this value would make them completely inaccessible, so I would personally revert this change even if it is a typo.

philipstanislaus: Removed this breaking change from the PR in 3f2b74370c393fabe1d2f96619d5b998eb3df2f6

philipstanislaus:
Added tests and handling of all the edge cases described above.

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+for MultiPassthrough<Call, AccountId, Timepoint>
+{
+	fn weigh_data(&self, (_, _, call): (&AccountId, &Timepoint, &Box<Call>)) -> Weight {
+		call.get_dispatch_info().weight
+	}
+}
+impl<Call: GetDispatchInfo, AccountId, Timepoint> ClassifyDispatch<(&AccountId, &Timepoint, &Box<Call>)>
+for MultiPassthrough<Call, AccountId, Timepoint>
+{
+	fn classify_dispatch(&self, (_, _, call): (&AccountId, &Timepoint, &Box<Call>))
```
shawntabrizi: All of this passthrough stuff can be updated with `FunctionOf` in the weight. You can check out master for the latest here.

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+			threshold: u16,
+			other_signatories: Vec<T::AccountId>
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			ensure!(threshold >= 1, Error::<T>::ZeroThreshold);
+			let max_sigs = T::MaxSignatories::get() as usize;
+			ensure!(!other_signatories.is_empty(), Error::<T>::TooFewSignatories);
+			ensure!(threshold as usize <= other_signatories.len() + 1, Error::<T>::TooFewSignatories);
+			ensure!(other_signatories.len() < max_sigs, Error::<T>::TooManySignatories);
+			ensure!(other_signatories.len() < u32::max_value() as usize, Error::<T>::TooManySignatories);
```
shawntabrizi: This check is useless?

The line above already confirms this is true.

philipstanislaus:
> IIUC this would no longer be breaking at all if you don't increment spec version. as it exists atm, this PR doesn't touch the runtime at all so incrementing the runtime spec version is not necessary.
> 
> I would put the integration into the substrate runtime into a separate PR if you find it necessary. otherwise I would remove the change to lib.rs

Thanks, I just removed the spec version increase. This makes a check fail though.

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+			threshold: u16,
+			signatories: Vec<T::AccountId>
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			ensure!(threshold >= 1, Error::<T>::ZeroThreshold);
+			let max_sigs = T::MaxSignatories::get() as usize;
+			ensure!(signatories.len() >= 2, Error::<T>::TooFewSignatories);
+			ensure!(threshold as usize <= signatories.len(), Error::<T>::TooFewSignatories);
+			ensure!(signatories.len() <= max_sigs, Error::<T>::TooManySignatories);
+			ensure!(signatories.len() <= u32::max_value() as usize, Error::<T>::TooManySignatories);
```
shawntabrizi: same comment

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+		/// # </weight>
+		#[weight = <SigsLen<T::AccountId>>::new()]
+		fn update(origin,
+			threshold: u16,
+			signatories: Vec<T::AccountId>
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			ensure!(threshold >= 1, Error::<T>::ZeroThreshold);
+			let max_sigs = T::MaxSignatories::get() as usize;
+			ensure!(signatories.len() >= 2, Error::<T>::TooFewSignatories);
```
shawntabrizi: Why must it be `>= 2`?

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+		/// - Storage: inserts one item, value size bounded by `MaxSignatories`, with a
+		///   deposit taken for its lifetime of
+		///   `MultiAccountDepositBase + signatories.len() * MultiAccountDepositFactor`.
+		/// # </weight>
+		#[weight = <SigsLen<T::AccountId>>::new()]
+		fn update(origin,
+			threshold: u16,
+			signatories: Vec<T::AccountId>
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
```
shawntabrizi: You never verify `who` here. This means anyone can update the multi-account.

This call should come from the multi-account itself, and it should be checked.

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+	fn ensure_sorted(signatories: Vec<T::AccountId>) -> Result<Vec<T::AccountId>, DispatchError> {
+		let mut maybe_last = None;
+		for item in signatories.iter() {
+			if let Some(last) = maybe_last {
+				ensure!(last < item, Error::<T>::SignatoriesOutOfOrder);
+			}
+			maybe_last = Some(item);
+		}
+		Ok(signatories)
+	}
```
shawntabrizi: There is a more idiomatic way to do this:

https://github.com/paritytech/substrate/blob/master/frame/recovery/src/lib.rs#L632

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+			signatories: Vec<T::AccountId>
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			ensure!(threshold >= 1, Error::<T>::ZeroThreshold);
+			let max_sigs = T::MaxSignatories::get() as usize;
+			ensure!(signatories.len() >= 2, Error::<T>::TooFewSignatories);
+			ensure!(threshold as usize <= signatories.len(), Error::<T>::TooFewSignatories);
+			ensure!(signatories.len() <= max_sigs, Error::<T>::TooManySignatories);
+			ensure!(signatories.len() <= u32::max_value() as usize, Error::<T>::TooManySignatories);
+			let signatories = Self::ensure_sorted(signatories.clone())?;
```
shawntabrizi:
```suggestion
			let signatories = Self::ensure_sorted_and_unique(signatories.clone())?;
```

Would clarify the function name

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+
+			if let Some(multi_account) = <MultiAccounts<T>>::get(&who) {
```
shawntabrizi: This can be written a little nicer

```suggestion
			let multi_account = <MultiAccounts<T>>::get(&who).ok_or(Error::<T>::MultiAccountNotFound)?;
```

Then you don't have to wrap everything in a giant if statement.

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+		///   signatories. `S` is capped by `MaxSignatories`, with weight being proportional.
+		/// - I/O: 1 mutate `O(S)`.
+		/// - One event.
+		/// - Storage: removes one item, value size bounded by `MaxSignatories`.
+		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
+		fn remove(origin) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			if let Some(multi_account) = <MultiAccounts<T>>::get(&who) {
```
shawntabrizi: same comment here

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+				let _ = T::Currency::unreserve(&multi_account.depositor, multi_account.deposit);
+
+				<MultiAccounts<T>>::remove(&who);
```
shawntabrizi: You should also clean any pending multi-sig operations.

something like:

```
<Multisigs<T>>::remove_prefix(&who);
```

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+		fn call(origin,
+			multi_account_id: T::AccountId,
+			maybe_timepoint: Option<Timepoint<T::BlockNumber>>,
+			call: Box<<T as Trait>::Call>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			if let Some(multi_account) = <MultiAccounts<T>>::get(&multi_account_id) {
+				// ensure that the origin is a signatory of the multi account
+				if let Err(_) = multi_account.signatories.binary_search(&who) {
```
shawntabrizi: User should provide the index in the vector where their account is, and you search a single element to see they are there.

So we change a binary search to a single value lookup.

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+				ensure!(item != &who, Error::<T>::SenderInSignatories);
+				index += 1;
+			}
+			maybe_last = Some(item);
+		}
+		signatories.insert(index, who);
+		Ok(signatories)
+	}
+
+	/// Remove approvals from account IDs that are no longer in the set of signatories
```
shawntabrizi: This looks very heavy. You are iterating over all approvals, and then doing a binary search for each of those.

Rather, you should make it so that the list of signatories cannot change if there are pending multisig operations.

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+		///   signatories. `S` is capped by `MaxSignatories`, with weight being proportional.
+		/// - One encode & hash, both of complexity `O(S)`.
+		/// - I/O: 1 read `O(S)`, 1 mutate `O(S)`.
+		/// - One event.
+		/// - Storage: inserts one item, value size bounded by `MaxSignatories`, with a
+		///   deposit taken for its lifetime of
+		///   `MultiAccountDepositBase + signatories.len() * MultiAccountDepositFactor`.
+		/// # </weight>
+		#[weight = <SigsLen<T::AccountId>>::new()]
+		fn update(origin,
```
shawntabrizi: I made a comment below, but the assumptions in this pallet can be reduced if you ensure that there are no existing multisig operations when update is called.

You can choose to have this function fail or you can chose to automatically delete any existing operations.

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+		/// # </weight>
+		#[weight = <MultiPassthrough<<T as Trait>::Call, T::AccountId, Option<Timepoint<T::BlockNumber>>>>::new()]
+		fn call(origin,
+			multi_account_id: T::AccountId,
+			maybe_timepoint: Option<Timepoint<T::BlockNumber>>,
+			call: Box<<T as Trait>::Call>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			if let Some(multi_account) = <MultiAccounts<T>>::get(&multi_account_id) {
```
shawntabrizi: Same comment here

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+					Err(Error::<T>::NotSignatory)?
+				}
+
+				let call_hash = call.using_encoded(blake2_256);
+
+				if let Some(mut m) = <Multisigs<T>>::get(&multi_account_id, call_hash) {
+					let timepoint = maybe_timepoint.ok_or(Error::<T>::NoTimepoint)?;
+					ensure!(m.when == timepoint, Error::<T>::WrongTimepoint);
+
+					let valid_approvals = Self::filter_approvals(m.approvals.clone(),
```
shawntabrizi: pass by reference

shawntabrizi: But I suggested in general to remove the `filter_approvals` logic

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+		fn approve(origin,
+			multi_account_id: T::AccountId,
+			maybe_timepoint: Option<Timepoint<T::BlockNumber>>,
+			call_hash: [u8; 32],
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			if let Some(multi_account) = <MultiAccounts<T>>::get(&multi_account_id) {
+				// ensure that the origin is a signatory of the multi account
+				if let Err(_) = multi_account.signatories.binary_search(&who) {
```
shawntabrizi: same comment about user passing an index

frame/multi-account/src/lib.rs
shawntabrizi: You can't really delete any existing operations, so you should fail here.

frame/multi-account/src/lib.rs
philipstanislaus: Good point, removed

frame/multi-account/src/lib.rs
philipstanislaus: The `create` function requires at least 1 other signatory, meaning each multi account requires 2 or more sigantories. Likewise, we require an update to set at least 2 signatories. This is in line with `as_multi` in the `utility` pallet. 

I agree though that we would allow more cases if we allow 1 signatory (proxy accounts that can be extendend later etc...). What do you think?

frame/multi-account/src/lib.rs
philipstanislaus: Great pattern, thanks!

frame/multi-account/src/lib.rs
philipstanislaus: Not sure I agree, that makes the interface much harder for users. This is implemented the same as https://github.com/paritytech/substrate/blob/master/frame/utility/src/lib.rs#L340 right now.

frame/multi-account/src/lib.rs
philipstanislaus: I like the approach you are suggesting, but it means that a former member can block it's removal by not removing a pending multisig operation.

In reality, there is an upper through `MaxSignatories`. So the upper boundary here is O(N * log(N)), where N is in reality most likely <= 100. 

An attack could be constructed by changing the signatories before the threshold is reached, so the approvals and signatories increase – but there is an economic cost to changing the signatories.

An alternative approach would be to clear all pending multisig operations on a signatory update, but finding all approvals at that stage is even more expensive. 

The logic here can be improved by first building a set and looking the signatories up on the set, but since N <= 100, it could be more expensive to build the set than to just do a binary search. 

Let me know what you think :)

frame/multi-account/src/lib.rs
philipstanislaus: See discussion above

frame/multi-account/src/lib.rs
philipstanislaus: See discussion above

frame/multi-account/src/lib.rs
philipstanislaus: Love that, thanks! 👍

frame/multi-account/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+				.expect("multi account indices will never reach 2^64 before the death of the universe; qed");
+
+			let id = Self::multi_account_id(multi_account_index);
+
+			let deposit = T::MultiAccountDepositBase::get()
+				+ T::MultiAccountDepositFactor::get() * (other_signatories.len() as u32 + 1).into();
+
+			MultiAccountIndex::put(multi_account_index);
+
+			T::Currency::reserve(&who, deposit)?;
```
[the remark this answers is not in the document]
philipstanislaus: Great find, changed. Thanks!

frame/multi-account/src/lib.rs
philipstanislaus: Awesome, changed. Thanks for pointing this out!

frame/multi-account/src/lib.rs
shawntabrizi: It will probably get changed then. In general, its not about "harder for users to use". The JS API can take care of these things where needed. If we can reduce on-chain computation, this means we can reduce the fees we charge to users and increase number of txs included in a single block. This is both much more important. (imo)

frame/multi-account/src/lib.rs
shawntabrizi: The problem is you are doing this with every single vote on a multi-sig, even when nothing has been changed. It certainly can be made more efficient. I am not sure what approach you want to take, but the one you have chosen here I don't think I would agree with.

I personally would not allow the multi-sig to change owners until all open multi-sig operations are closed. You can create a function where the multi-sig can close any operation on behalf of the original caller.

frame/multi-account/src/lib.rs
shawntabrizi: To be honest it is probably not that bad to keep the same, but I would guess a round of runtime optimization would change it to the other pattern anyway.

philipstanislaus: Makes sense. Happy to change it if this becomes a blocker for getting this merged. Thanks!

frame/multi-account/src/lib.rs
philipstanislaus: @shawntabrizi that's a good alternative. 

I like the simplicity of what you are suggesting – it makes the whole logic much simpler and removes all the edge cases that need to be considered.

Just changed the code, documentation and tests accordingly – would appreciate your feedback.

Many thanks!

philipstanislaus:
We have created a dedicated repository for this pallet: https://github.com/centrifuge/substrate-pallet-multi-account 

I keep this PR open though in the hope that it will still be merged – to be audited and reviewed as a part of Substrate. 

Thanks for all the feedback so far, that helped a lot!

shawntabrizi:
@philipstanislaus Thanks for your PR.

As we discussed offline, this PR will not get merged into Substrate. Unfortunately, every pallet in our repository adds additional overhead to the Parity team in terms of maintenance and support. Especially given the fact we will not use this pallet in any of our core projects (Substrate Node, Polkadot/Kusama Node), it is even more likely that some breaking change/bug is introduced and we miss it. This would be bad for everyone involved.

We encourage all Substrate developers to create new and useful pallets, but contribute them to the community just like any open-source project. Parity does not intend to be the only place where pallets are found. We hope in the long term, we our pallets are just one choice in pool of many options for different runtime developers.

I hope the review I gave was sufficient for your needs, and I hope that you will include this pallet into your runtime and as a part of an open source repository so others can use it.

Thanks, and feel free to follow up with me about any concerns you might have!

philipstanislaus:
@shawntabrizi understood, thanks for the context and especially your valuable feedback in the process. 

We improved a few things with the multi account pallet and performed an internal audit as well (3 other engineers were involved in it). It is used in our chain right now. 

Here is the link again for anyone that would like to use it: https://github.com/centrifuge/substrate-pallet-multi-account

PR #4912, opened 2020-02-13
Small bench macro

marcio-diaz:
Another iteration over the bench macro that I did yesterday.
(Pushing in case someone wants to use it, probably will be subsumed later on by one written by Gav)


marcio-diaz:
superseded

Issue #4908, opened 2020-02-12
Optimize signature verification


NikVolf:
1. During block import, they can be split in advance and verified in parallel.
2. During block construction, the fact that transaction previously was when imported in transaction pool can be cached.

(sometime maybe) 3. Some networks that do only something like staking might end up with a lot of signatures from the same private key. Those can be batch-verified.

tomusdrw:
I think `sr` also supports batch-verification of multiple signatures (which is faster than verifying them one-by-one), might be a good idea to incorporate that too. I don't know the details though and if it's already implemented in schnorrkel already, but for sure @burdges will know both :)

burdges:
You can even batch verify with a futures API like https://www.zfnd.org/blog/futures-batch-verification/

There are `verify_batch` fns for both [ed25519](https://docs.rs/ed25519-dalek/1.0.0-pre.1/ed25519_dalek/fn.verify_batch.html) and [schnorrkel](https://docs.rs/schnorrkel/0.8.5/schnorrkel/sign/fn.verify_batch.html).  Also, I made sure calls to [`schnorrkel::Signature::from_bytes`](https://docs.rs/schnorrkel/0.8.5/schnorrkel/sign/struct.Signature.html#method.from_bytes) always fail on ed25519 signatures, so one can write a `verify_batch` for `AnySignature` by assigning the signatures that deserialize as schnorkel to the schnorkel side, and the others to the ed25519 side, and then calls the appropriate `verify_batch` for each, like
```
pub struct AnySignedBatch<'m> {
    sr_t: Vec<merlin::Transcript>,
    sr_sigs: Vec<schnorrkel::Signature>,
    sr_pks: Vec<schnorkel::PublicKey>,
    ed_m: Vec<Cow<'m,[u8]>>,
    ed_sigs: Vec<ed25519_dalek::Signature>,
    ed_pks: Vec<ed25519_dalek::PublicKey>,
}

impl<'m> AnySignedBatch<'m> {
    pub fn add(&mut self, msg: &[u8], signature: [u8; 64], publickey: [u8; 32]) -> Result<()> {
        if Ok(sr_sig) = ::schnorrkel::Signature::from_bytes(signature) {
              let sr_pk = ::schnorrkel::PublicKey::from_bytes(publickey) ?;
              // Do substrate's thing for transcripts
              // Push them on the first three Vecs
        } else if Ok(ed_sig) = ::ed25519_dalek::Signature::from_bytes(signature) {
[...]

Issue #4682, opened 2020-01-20
Wasmtime spends too much time growing memory


NikVolf:
During the block execution/construction with ~500-600 transactions, about 50% of time is spent on growing memory inside wasmtime.

(this method https://github.com/bytecodealliance/wasmtime/blob/815576edc5ab00da8c6b67916bf7e9f8af1504d2/crates/runtime/src/instance.rs#L632)

bkchr:
And wasmi behaves differently? I assume you test `remark`?

NikVolf:
I am not sure about wasmi, it is much slower anyway.

I tested transfers, but only between several fixed accounts, so should be not much difference than just remark.

bkchr:
By growing memory you mean allocating for something inside the runtime or really for some structures inside `wasmtime`?

NikVolf:
Sorry, I probably gave too little context

https://github.com/bytecodealliance/wasmtime/blob/815576edc5ab00da8c6b67916bf7e9f8af1504d2/crates/runtime/src/instance.rs#L632

Exactly this method takes up to 50% time

pepyakin:
Yeah, thanks for the filling the issue! It is on my radar already.

pepyakin:
I think the issue should be fixed by now.

PR #4916, opened 2020-02-13
Build block without checking signatures

NikVolf:
- we now assume that transactions come from the source that checks signatures before building the block

- this adds another function to block building runtime-api, called `apply_trusted_extrinsic`, which will skip signature checking.

- `Checkable` trait augmented so that `check` could skip signature verification part.

speculative part of #4908


frame/executive/src/lib.rs
```diff
@@ -272,9 +294,11 @@ where
 		uxt: Block::Extrinsic,
 		encoded_len: usize,
 		to_note: Option<Vec<u8>>,
+		check_signature: bool,
 	) -> ApplyExtrinsicResult {
 		// Verify that the signature is good.
-		let xt = uxt.check(&Default::default())?;
+		let xt = if check_signature { uxt.check(&Default::default())? }
```
bkchr: Should all split into individual lines.

frame/executive/src/lib.rs
```diff
@@ -272,9 +294,11 @@ where
 		uxt: Block::Extrinsic,
 		encoded_len: usize,
 		to_note: Option<Vec<u8>>,
+		check_signature: bool,
```
bkchr: Maybe an enum to be more explicit?

```
enum CheckSignature {
    Yes,
    No,
}
```

frame/executive/src/lib.rs
```diff
@@ -144,10 +150,14 @@ impl<
 		WeighBlock<System::BlockNumber>,
 > Executive<System, Block, Context, UnsignedValidator, AllModules>
 where
-	Block::Extrinsic: Checkable<Context> + Codec,
+	Block::Extrinsic: Checkable<Context> + UnsafeConvert<Context> + Codec,
```
bkchr:
```suggestion
	Block::Extrinsic: Checkable<Context> + UnsafeConvert<Context, UnsafeResult = CheckedOf<Block::Extrinsic, Context>> + Codec,
```

Than you should be able to remove the other bounds below.

frame/executive/src/lib.rs
```diff
@@ -118,10 +120,14 @@ impl<
 		WeighBlock<System::BlockNumber>,
 > ExecuteBlock<Block> for Executive<System, Block, Context, UnsignedValidator, AllModules>
 where
-	Block::Extrinsic: Checkable<Context> + Codec,
+	Block::Extrinsic: Checkable<Context> + UnsafeConvert<Context> + Codec,
```
bkchr: Same

primitives/runtime/src/traits.rs
```diff
@@ -652,6 +652,18 @@ pub trait BlindCheckable: Sized {
...
 
+/// Unsafe convert.
+///
+/// Similar to `Checkable`, but without an actual check. "Unsafe" means that user is encouraged
+/// to look in individual implementation of this trait for actual assumption which could be unsafe.
+pub trait UnsafeConvert<Context>: Sized {
+	/// Returned type of unsafe convert.
+	type UnsafeResult;
+
+	/// Convert self to .
```
marcio-diaz: to?

frame/executive/src/lib.rs
```diff
@@ -252,16 +262,28 @@ where
 	///
 	/// This doesn't attempt to validate anything regarding the block, but it builds a list of uxt
 	/// hashes.
+	/// This also does not check transaction signature and assumes `uxt` passed is originated
```
marcio-diaz: What is implying the `true`?

primitives/runtime/src/traits.rs
```diff
@@ -652,6 +652,18 @@ pub trait BlindCheckable: Sized {
 	fn check(self) -> Result<Self::Checked, TransactionValidityError>;
 }
 
+/// Unsafe convert.
+///
+/// Similar to `Checkable`, but without an actual check. "Unsafe" means that user is encouraged
+/// to look in individual implementation of this trait for actual assumption which could be unsafe.
+pub trait UnsafeConvert<Context>: Sized {
```
bkchr: Yeah, I think we should come up with a better idea/name for this^^

client/block-builder/src/lib.rs
```diff
@@ -153,7 +153,7 @@ where
 			})
 		} else {
 			self.api.map_api_result(|api| {
-				match api.apply_extrinsic_with_context(
+				match api.apply_trusted_extrinsic_with_context(
```
bkchr: We can not really know at this place if the extrinsic is trusted ^^

bkchr:
I think in general it looks like a direction we can follow.

Have you done any benchmarks?

@gavofyork WDYT?

frame/executive/src/lib.rs
```diff
@@ -255,13 +265,25 @@ where
 	pub fn apply_extrinsic(uxt: Block::Extrinsic) -> ApplyExtrinsicResult {
```
bkchr: We should have some test that ensures that this denies an extrinsic with an invalid signature and the new one applies it.

frame/executive/src/lib.rs
tomusdrw: Why not pass extra parameters to `Checkable` instead? An extra trait `UnsafeConvert` seems to complicate everything more than help imho.

bkchr: Yeah, I thought the same, but I'm not 100% sure.

NikVolf: Well it was my first lazy thought, but because it is called "Checkable", it is supposed to check, not to skip check :)

bkchr: Yeah that would be my mental model as well, however a `skip_check` parameter would a little bit cleaner.

client/block-builder/src/lib.rs
NikVolf: > Have you done any benchmarks?

Haven't been doing any benchmarks for block construction so far, only for block import.
But I think signature verification should not be a bottleneck there (since it is always native?), so do not expect much gain.

NikVolf:
@bkchr 
> Have you done any benchmarks?

Haven't been doing any benchmarks for block construction so far, only for block import.
But I think signature verification should not be a bottleneck there (since it is always native?), so do not expect much gain. When profiling last time, wasmtime was a bottleneck, and it's memory growing.

client/block-builder/src/lib.rs
NikVolf: I thought we always use trusted for block construction from this pr on?

bkchr:
> @bkchr
> 
> > Have you done any benchmarks?
> 
> Haven't been doing any benchmarks for block construction so far, only for block import.
> But I think signature verification should not be a bottleneck there (since it is always native?), so do not expect much gain. When profiling last time, wasmtime was a bottleneck, and it's [memory growing](https://github.com/paritytech/substrate/issues/4682).

If that isn't the bottleneck, why do we even try to optimize it? :D 

(The issue should not be rechecked, as we landed some optimizations to wasmtime)

NikVolf:
> If that isn't the bottleneck, why do we even try to optimize it? :D

Well, it can become bottleneck, once block construction will approach native speed :)
If you recall, I wanted to start with batch verification for import, where signature verification takes ~20% time, and there are benchmarks.

bkchr:
Yeah, I trust you with these results on block import. I just was curios if we see any improvement on block production :) But as we are current not bound by it, benchmark results will not be that helpful.

burdges:
I'd expect signature verification code should actually run native, so the runtime WASM code would check the block normally, but substrate would batch the signature verification itself, and then call `verify_batch` before finishing.

If signatures come from untrusted sources, like in block production, then you can use batch verification, but it requires either a traitor tracing algorithm, or just rerunning the construction with individual verification.  We love batch verification in block verification because someone always spends some limited resource like VRF wins producing that block. 

It's possible for this batch verification to become a mild DoS vector for block production though:  An adversary fills a block producer's transaction pool with ECDSA ecRecover calls, which run slow and cannot be batched, and a single invalid Ed25519 signature, which we batch until the end.   Now the block producers builds their block checking each good slow ECDSA ecRecover along the way, but they queue the one bad Ed25519 until then end, at which point the block construction fails, and they must start over verifying each transaction along the way.  This results in twice the computational work.

bkchr:
We verify each transaction before it goes into the transaction pool and can be used for block production. So, you can not slow down the block production. You can Ddos the node with invalid transaction, but it will probably just ban you.

burdges:
I see.  If different runtimes handle their transaction pool differently then you need the runtime to make the choice about whether it trusts signature or not, and it cannot be decided purely by substrate, yes?  As you're already discussing, any interface to these pre-verified signature should be missuse resistant so that parachain authors do not accidentally use it in the wrong place. 

You could still have three or four modes for signature verification from substrate's side:
 - individual verification - our current approach but maybe still useful for bench marking
 - batch verified signatures - 40% faster than individual verification
 - all signatures pre-verified - requires care to avoid parachain authors never missuse this
 - cached signature verification - faster than batch verification, also more missuse resistant and slower than pre-verified, but requires nasty cache management. 

We could maybe even disable pre-verified entirely on the substrate site for polkadot validators running parachain runtimes?

NikVolf:
> I see. If different runtimes handle their transaction pool differently then you need the runtime to make the choice about whether it trusts signature or not, and it cannot be decided purely by substrate, yes? 

Currently block proposing uses only transaction pool as interface, and transaction pool is always checking signatures.

But it is true, in general, block proposer interface to transaction source needs to be altered, simplified, and made more general, and then this current assumption will no longer hold. But by this time we will have batch verification - same as for block import already.

> cached signature verification - faster than batch verification, also more missuse resistant and slower than pre-verified, but requires nasty cache management.

I am in favour of this approach actually, for block proposing. Cache management seems rather trivial.

client/block-builder/src/lib.rs
bkchr: The blockbuilder is just a wrapper around the `BlockBuilder` runtime api and you still have the `apply_extrinsic` function. So, either remove `apply_extrinsic` or extend the `BlockBuilder` struct with a function to call this funciton.

NikVolf: But there is a compatibility call above to `apply_extrinsic` (or I misunderstand how versioning works)

NikVolf:
@bkchr @tomusdrw addressed suggestions, thanks!

primitives/runtime/src/testing.rs
```diff
@@ -313,24 +370,40 @@ impl<Call, Extra> Serialize for TestXt<Call, Extra> where TestXt<Call, Extra>: E
...
 	type Checked = Self;
-	fn check(self, _: &Context) -> Result<Self::Checked, TransactionValidityError> { Ok(self) }
+	fn check(self, signature: CheckSignature, _: &Context) -> Result<Self::Checked, TransactionValidityError> {
+		match self.validity {
+			TestValidity::Valid => Ok(self),
+			TestValidity::SignatureInvalid(e) =>
+				if let CheckSignature::No = signature {
+					Ok(self)
+				}
+				else {
```
bkchr:
```suggestion
				} else {
```

primitives/runtime/src/testing.rs
```diff
@@ -313,24 +370,40 @@ impl<Call, Extra> Serialize for TestXt<Call, Extra> where TestXt<Call, Extra>: E
...
 impl<Call: Codec + Sync + Send, Context, Extra> Checkable<Context> for TestXt<Call, Extra> {
 	type Checked = Self;
-	fn check(self, _: &Context) -> Result<Self::Checked, TransactionValidityError> { Ok(self) }
+	fn check(self, signature: CheckSignature, _: &Context) -> Result<Self::Checked, TransactionValidityError> {
+		match self.validity {
+			TestValidity::Valid => Ok(self),
+			TestValidity::SignatureInvalid(e) =>
+				if let CheckSignature::No = signature {
+					Ok(self)
+				}
```
bkchr:
```suggestion
```

primitives/runtime/src/generic/mod.rs
```diff
@@ -39,6 +39,15 @@ pub use self::digest::{
 use crate::codec::Encode;
 use sp_std::prelude::*;
 
+#[derive(PartialEq, Eq, Clone, Copy)]
+/// Perform singature check.
```
bkchr:
```suggestion
/// Perform singature check.
#[derive(PartialEq, Eq, Clone, Copy)]
```

primitives/runtime/src/generic/mod.rs
```diff
@@ -39,6 +39,15 @@ pub use self::digest::{
 use crate::codec::Encode;
 use sp_std::prelude::*;
 
+#[derive(PartialEq, Eq, Clone, Copy)]
```
bkchr:
```suggestion
```

primitives/block-builder/src/lib.rs
```diff
@@ -58,6 +58,11 @@ sp_api::decl_runtime_apis! {
 		/// Returns an inclusion outcome which specifies if this extrinsic is included in
 		/// this block or not.
 		fn apply_extrinsic(extrinsic: <Block as BlockT>::Extrinsic) -> ApplyExtrinsicResult;
+		/// Apply the given extrinsic.
+		///
+		/// Same as `apply_extrinsic`, but assumes that signature check has been already performed
```
bkchr:
```suggestion
		/// Same as `apply_extrinsic`, but skips signature verification.
```

primitives/block-builder/src/lib.rs
```diff
@@ -58,6 +58,11 @@ sp_api::decl_runtime_apis! {
...
+		/// elsewhere.
```
bkchr:
```suggestion
```

frame/executive/src/lib.rs
```diff
@@ -272,9 +286,13 @@ where
 		uxt: Block::Extrinsic,
 		encoded_len: usize,
 		to_note: Option<Vec<u8>>,
+		check_signature: CheckSignature,
 	) -> ApplyExtrinsicResult {
 		// Verify that the signature is good.
-		let xt = uxt.check(&Default::default())?;
+		let xt = uxt.check(
+			check_signature,
+			&Default::default()
```
bkchr:
```suggestion
			&Default::default(),
```

frame/executive/src/lib.rs
```diff
@@ -255,13 +257,25 @@ where
...
+	/// Same as `apply_extrinsic`, but without signature checks.
+	/// This does not check transaction signature and assumes `uxt` passed is originated
+	/// from the source that validates transactions before applying (for example, transaction pool).
+	pub fn apply_trusted_extrinsic(uxt: Block::Extrinsic) -> ApplyExtrinsicResult {
+		let encoded = uxt.encode();
+		let encoded_len = encoded.len();
+		Self::apply_extrinsic_with_len(uxt, encoded_len, Some(encoded), CheckSignature::No)
+	}
+
+
```
bkchr:
```suggestion
```

frame/executive/src/lib.rs
```diff
@@ -255,13 +257,25 @@ where
...
 		let encoded = uxt.encode();
 		let encoded_len = encoded.len();
-		Self::apply_extrinsic_with_len(uxt, encoded_len, Some(encoded))
+		Self::apply_extrinsic_with_len(uxt, encoded_len, Some(encoded), CheckSignature::Yes)
 	}
 
+	/// Apply extrinsic outside of the block execution function.
+	///
+	/// Same as `apply_extrinsic`, but without signature checks.
+	/// This does not check transaction signature and assumes `uxt` passed is originated
```
bkchr: I think these both sentences can be removed.

frame/executive/src/lib.rs
```diff
@@ -255,13 +257,25 @@ where
 	pub fn apply_extrinsic(uxt: Block::Extrinsic) -> ApplyExtrinsicResult {
 		let encoded = uxt.encode();
 		let encoded_len = encoded.len();
-		Self::apply_extrinsic_with_len(uxt, encoded_len, Some(encoded))
+		Self::apply_extrinsic_with_len(uxt, encoded_len, Some(encoded), CheckSignature::Yes)
 	}
 
+	/// Apply extrinsic outside of the block execution function.
+	///
+	/// Same as `apply_extrinsic`, but without signature checks.
```
bkchr:
```suggestion
	/// Same as `apply_extrinsic`, but skips signature verification.
```

client/block-builder/src/lib.rs
```diff
@@ -121,11 +121,23 @@ where
...
+	/// This will ensure the extrinsic can be validly executed (by executing it).
 	pub fn push(&mut self, xt: <Block as BlockT>::Extrinsic) -> Result<(), ApiErrorFor<A, Block>> {
+		self.push_internal(xt, false)
+	}
+
+	/// Push onto the block's list of extrinsics.
+	///
+	/// This will treat incoming extrinsic `xt` as untrusted and perform additional checks
+	/// (currenty checking signature).
+	pub fn push_untrusted(&mut self, xt: <Block as BlockT>::Extrinsic) -> Result<(), ApiErrorFor<A, Block>> {
```
bkchr: Why do we switch the behavior here? I think it should be also called `push` and `push_trusted`.

client/block-builder/src/lib.rs
```diff
@@ -153,11 +165,20 @@ where
 			})
 		} else {
 			self.api.map_api_result(|api| {
-				match api.apply_extrinsic_with_context(
-					block_id,
-					ExecutionContext::BlockConstruction,
-					xt.clone(),
-				)? {
+				let apply_result = match do_untrusted_checks {
```
bkchr: `match bool` => `if bool`

client/block-builder/src/lib.rs
```diff
@@ -153,11 +165,20 @@ where
 			})
 		} else {
 			self.api.map_api_result(|api| {
-				match api.apply_extrinsic_with_context(
```
bkchr: You need to bump the version of the api and make sure that `push_trusted` calls `apply_extrinsic` for `version < 5`.

client/block-builder/src/lib.rs
NikVolf: Thought `push` would be defaulted to `_trusted`!
Changed

client/block-builder/src/lib.rs
```diff
@@ -151,13 +163,30 @@ where
 					Err(e) => Err(ApplyExtrinsicFailed::from(e).into()),
 				}
 			})
-		} else {
-			self.api.map_api_result(|api| {
-				match api.apply_extrinsic_with_context(
+		}  else {
```
bkchr:
```suggestion
		} else {
```

NikVolf:
@gnunicorn I am not sure if runtime impl/spec version should be bumped here

gnunicorn:
[doesn't this mean the API of the runtime](https://github.com/paritytech/substrate/pull/4916/files#diff-22dd0ecfa9822d47f43e68594e5f64e0R689) is different now? Then `spec_version` should have a bump – and usually we bump `impl_version` along with it...

NikVolf:
What could be invoked before hasn't changed, but I will just  bump `spec_veriosn` anyway... :)
`impl_version` should be zeroed afaik, not bumped along with it

Issue #4689, opened 2020-01-20
Add remote signing to substrate client


wpank:
The following proposes the addition of remote signing functionality within the substrate client. 

# Context

Security of Proof of Stake networks lie within the hands of validators - without the security these entities provide, the whole system falls apart. The responsibility of a validator is to operate stable, reliable, consistent, and secure operations of their nodes. This responsibility also includes managing their signing keys, keys that let the network know _they_ were the ones that verified that the activity they put on the network is non-byzantine.

As a validator, the current paradigm of storing hot session keys in the client leaves much to be desired in terms of security. Although session keys cannot lead to direct access of funds, a compromise of the validator host (and the session keys within it) can lead to a complete loss of funds for a validator and the funds of those nominating them. Furthermore, there is a greedy incenctive to compromise these keys, as up to 10% of the slash can get rewarded to those who report it. While key rotation helps mitigate this to an extent, a more elegant solution of key storage and signing will be required in the long run.

Separating out the storage and signing interface of session keys from the validator host client would allow validators to create more robust and flexible operations, while providing additional layers of defense against possible attack vectors. A full compromise of the validator host shouldn't enable conditions where the validator can be slashed. Separating out the storage of session keys would mean adding the ability to have a remote signing interface, which gives a flexible means of having a remote signing server - one which ideally has double signing protection and HSM, TEE, Ledger, and TSS support. This addition increases the cost of compromising validator operations, something that creates a more resilient and secure network in the long run. 

## Remote Signing Server

The following proposes the approaches of one remote signing server, although the interfaces exposed by the substrate client should allow for multiple implementations to exist. The signing server proposed here would live as a rust module in a separate repository - these considerations are for reference and context.

A remote signing server should be flexible to account for a diversity of key management approaches, including TEE, HSM, cloud HSM, Ledger, and encrypted software based key storage. Additionally, the remote signing server should be able to support multiple substrate based chains. This essentially acts as a single API for all key management and signing.

#### Approach

The signing server should run as a separate process on a physical on-premise host, although cloud based should be considered as well (although is less preferred).

[...]

PR #4925, opened 2020-02-14
Introduce `sign_with` method in keystore

rakanalh:
This PR is part of the effort on implementing [Remote signing](https://github.com/paritytech/substrate/issues/4689).

The change introduces `sign_with` method where the caller passes the public key and the key type ID. The pair matching against those parameters is done within the store rather than outside of it so that the call site isn't responsible about signing implementation details.


client/keystore/src/lib.rs
```diff
@@ -21,7 +21,7 @@
 use std::{collections::HashMap, path::PathBuf, fs::{self, File}, io::{self, Write}, sync::Arc};
 
 use sp_core::{
-	crypto::{KeyTypeId, Pair as PairT, Public, IsWrappedBy, Protected}, traits::BareCryptoStore,
+	crypto::{KeyTypeId, Pair as PairT, Public, IsWrappedBy, Protected, SupportedPublicKey, SignatureValue}, traits::BareCryptoStore,
```
gnunicorn: please wrap.

client/keystore/src/lib.rs
```diff
@@ -276,9 +276,35 @@ impl Store {
 		buf.push(key_type + key.as_str());
 		Some(buf)
 	}
+
+	/// Signs a message with the private key that matches
+	/// the public key passed.
+	fn _sign_with(&self, public: SupportedPublicKey, id: KeyTypeId, msg: &[u8]) -> std::result::Result<SignatureValue, String> {
```
gnunicorn: please wrap...

client/keystore/src/lib.rs
```diff
@@ -276,9 +276,35 @@ impl Store {
...
+			}
+		}
+	}
 }
 
 impl BareCryptoStore for Store {
+
+	/// Signs a message with the private key that matches
+	/// the public key passed.
+	fn sign_with(&self, public: SupportedPublicKey, id: KeyTypeId, msg: &[u8]) -> std::result::Result<SignatureValue, String> {
```
gnunicorn: please wrap

primitives/io/src/lib.rs
```diff
@@ -410,8 +413,10 @@ pub trait Crypto {
 		self.extension::<KeystoreExt>()
 			.expect("No `keystore` associated for the current context!")
 			.read()
-			.ed25519_key_pair(id, &pub_key)
-			.map(|k| k.sign(msg))
+			.sign_with((*pub_key).into(), id, msg)
+			.map(|sig| sig.try_into().ok())
```
gnunicorn: I just learned about the nicer `and_then` (instead of `map(..).flatten()` pattern) ...probably something to use here..

primitives/core/src/crypto.rs
```diff
@@ -529,6 +530,76 @@ pub trait Public: AsRef<[u8]> + AsMut<[u8]> + Default + Derive + CryptoType + Pa
...
+	fn to_raw_vec(&self) -> Vec<u8> { self.as_slice().to_vec() }
+
+	/// Return a slice filled with raw data.
+	fn as_slice(&self) -> &[u8] { self.as_ref() }
+}
+
+/// An enum which holds references to public keys which are
+/// supported for signing messages
+#[cfg(feature = "full_crypto")]
+pub enum SupportedPublicKey {
```
gnunicorn: Is it possible to use `AppKey` directly and drop this and the related enum in favor of it?

client/keystore/src/lib.rs
bkchr:
```suggestion
	fn sign_with(&self, public: SupportedPublicKey, id: KeyTypeId, msg: &[u8]) -> std::result::Result<SignatureValue, String> {
```

primitives/core/src/crypto.rs
bkchr: Can we not just follow the same strategy as before? Like `sign_with_ed25519` etc?

client/keystore/src/lib.rs
bkchr: This function should also return a `Future`.

primitives/core/src/crypto.rs
rakanalh: @gnunicorn and I have discussed this and here's a summary of our thoughts about this:

`AppPair`/`AppPublic` ... etc are defined in `sp-app-crypto` while the `BareCryptoStore` trait is defined in `sp-core`. The former crate depends on the latter so it's not a good idea to `use sp_app_crypto` inside `sp_core`. Therefore, the trait cannot use `AppPublic` as a parameter type in replacement of `SupportedPublicKey`.

primitives/core/src/traits.rs
```diff
@@ -74,6 +77,15 @@ pub trait BareCryptoStore: Send + Sync {
...
 	/// Returns `true` iff all private keys could be found.
 	fn has_keys(&self, public_keys: &[(Vec<u8>, KeyTypeId)]) -> bool;
+
+	/// Signs a message with the private key that matches
+	/// the public key passed.
+	fn sign_with(
+		&self,
+		kind: KeyKindId,
+		id: KeyTypeId,
+		msg: &[u8],
```
bkchr: context missing.

client/keystore/src/lib.rs
```diff
@@ -276,11 +277,59 @@ impl Store {
...
+					.sr25519_public_keys(id)
+					.pop()
+					.ok_or(String::from("sr25519 key not found"))?;
+				let key_pair: sr25519::Pair = self
+					.key_pair_by_type::<sr25519::Pair>(&public_key, id)
+					.map(Into::into)
+					.map_err(|e| e.to_string())?;
+				return Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec());
+			}
+			_ => Err(String::from("Key kind invalid")),
```
NikVolf:
```suggestion
			_ => Err(String::from("Key kind invalid"))
```

client/keystore/src/lib.rs
```diff
@@ -276,11 +277,59 @@ impl Store {
...
+		}
+	}
 }
 
 impl BareCryptoStore for Store {
+	/// Signs a message with the private key that matches
```
NikVolf: This doc should be in `BareCryptoStore` trait?

primitives/core/src/ecdsa.rs
```diff
@@ -95,6 +95,7 @@ impl Public {
 }
 
 impl TraitPublic for Public {
+	const KEY_KIND_ID: &'static str = "sr25519";
```
NikVolf: Why this is inside a trait directly and in ed25519 is defined through some const `ED25519_KIND_ID`?

primitives/core/src/testing.rs
```diff
@@ -131,6 +134,34 @@ impl crate::traits::BareCryptoStore for KeyStore {
 	fn has_keys(&self, public_keys: &[(Vec<u8>, KeyTypeId)]) -> bool {
 		public_keys.iter().all(|(k, t)| self.keys.get(&t).and_then(|s| s.get(k)).is_some())
 	}
+
+	fn sign_with(&self, kind: KeyKindId, id: KeyTypeId, msg: &[u8]) -> Result<Vec<u8>, String> {
+		match kind {
+			"ed25519" => {
```
NikVolf: Shouldn't it use const you define above?

primitives/core/src/traits.rs
rakanalh: As stated elsewhere, this is not the case anymore.

client/keystore/src/lib.rs
```diff
@@ -276,11 +277,59 @@ impl Store {
 		buf.push(key_type + key.as_str());
 		Some(buf)
 	}
+
+	/// Signs a message with the private key that matches
+	/// the public key passed.
+	fn _sign_with(
```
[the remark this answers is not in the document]
rakanalh: `sign_with` is defined in `BareCryptoStore` trait which is why we opted for underscoring this one in `KeyStore`

primitives/core/src/traits.rs
```diff
@@ -69,11 +66,60 @@ pub trait BareCryptoStore: Send + Sync {
 
 	/// Get the password for this store.
 	fn password(&self) -> Option<&str>;
-
+	/// Provided a list of (CryptoTypeId,[u8]) pairs, this would return
+	/// a filtered list of public keys which are supported by the keystore.
+	fn get_supported_keys(&self, id: KeyTypeId, keys: Vec<CryptoTypePublicPair>) -> Result<Vec<CryptoTypePublicPair>, String>;
```
NikVolf: we, along with Rust in general, avoid `get_` prefixes

primitives/core/src/traits.rs
```diff
@@ -69,11 +66,60 @@ pub trait BareCryptoStore: Send + Sync {
 
 	/// Get the password for this store.
 	fn password(&self) -> Option<&str>;
-
+	/// Provided a list of (CryptoTypeId,[u8]) pairs, this would return
```
NikVolf: First doc header, then description separated by empty line renders much better in html

```rust
/// List of supported keys.
/// 
/// Provided a list of (CryptoTypeId,[u8]) pairs, this would return 
/// a filtered list of public keys which are supported by the keystore.
```

primitives/core/src/testing.rs
```diff
@@ -131,6 +136,63 @@ impl crate::traits::BareCryptoStore for KeyStore {
...
+				return Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec());
+			}
+			sr25519::SR25519_CRYPTO_ID => {
+				let key_pair: sr25519::Pair = self
+					.sr25519_key_pair(id, &sr25519::Public::from_slice(key.1.as_slice()))
+					.map(Into::into)
+					.ok_or(String::from("sr25519 pair not found"))?;
+				return Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec());
+			}
+			_ => Err(String::from("Key kind invalid")),
```
NikVolf:
```suggestion
			_ => Err(String::from("Key kind invalid"))
```

client/keystore/src/lib.rs
```diff
@@ -276,11 +277,87 @@ impl Store {
...
+			}
+			sr25519::SR25519_CRYPTO_ID => {
+				let pub_key = sr25519::Public::from_slice(key.1.as_slice());
+				let key_pair: sr25519::Pair = self
+					.key_pair_by_type::<sr25519::Pair>(&pub_key, id)
+					.map(Into::into)
+					.map_err(|e| e.to_string())?;
+				return Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec());
+			}
+			_ => Err(String::from("Key kind invalid"))
```
gnunicorn:
```suggestion
			_ => Err(String::from("Key kind not supported"))
```

client/keystore/src/lib.rs
```diff
@@ -276,11 +277,87 @@ impl Store {
...
 	}
+
+	/// Signs a message with the private key that matches
+	/// the public key passed.
+	fn _sign_with(
+		&self,
+		id: KeyTypeId,
+		key: &CryptoTypePublicPair,
+		msg: &[u8],
+	) -> std::result::Result<Vec<u8>, String> {
```
gnunicorn: we already import `std`, so maybe we could just extend that import to also get `Result` and have a shorter version here and below? Preference ...

```suggestion
	) -> Result<Vec<u8>, String> {
```

client/keystore/src/lib.rs
```diff
@@ -276,11 +277,87 @@ impl Store {
...
+								.map(|keys| keys.iter().map(|k| k.to_raw_vec()).collect())
+								.map_err(|e| e.to_string())?;
+
+		Ok(keys.iter().filter_map(|k| {
+			match k.0 {
+				sr25519::SR25519_CRYPTO_ID if sr25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
+				ed25519::ED25519_CRYPTO_ID if ed25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
+				_ => None
+			}
+		}).cloned().collect::<Vec<_>>())
```
gnunicorn: do we really have to clone here? How come?

```suggestion
		}).collect::<Vec<_>>())
```

client/keystore/src/lib.rs
```diff
@@ -276,11 +277,87 @@ impl Store {
...
+		let mut keys: Vec<CryptoTypePublicPair> = vec![];
+		keys.extend(sr25519_existing_keys.iter()
+					.cloned()
+    				.map(|k| (sr25519::SR25519_CRYPTO_ID, k)));
+		keys.extend(ed25519_existing_keys.iter()
+    				.cloned()
+					.map(|k| (ed25519::ED25519_CRYPTO_ID, k))
+					.collect::<Vec<_>>());
+		Ok(keys)
+
```
gnunicorn: Maybe use `.chain` instead? And do that internal mapping already in the first iteration? Would save you a few vec allocations..


```rust

let ed_keys = self.public_keys_by_type::<ed25519::Public>(id)
	.map_err(|e| e.to_string())?
	.iter().map(|k| (ed25519::ED25519_CRYPTO_ID, k.to_raw_vec()));

let sr_keys = self.public_keys_by_type::<sr25519::Public>(id)
	.map_err(|e| e.to_string())?
	.iter().map(|k| (sr25519::SR25519_CRYPTO_ID, k.to_raw_vec()));

Ok(ed_keys.chain(sr_keys).collect::Vec<_>());
```

Wouldn't that do it?

primitives/core/src/crypto.rs
```diff
@@ -918,6 +925,13 @@ impl<'a> TryFrom<&'a str> for KeyTypeId {
 	}
 }
 
+/// An identifier for a specific cryptographic algorithm used by a key pair
+#[derive(Copy, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct CryptoTypeId(pub &'static str);
```
gnunicorn: Didn't we say, this should better be a bound sized `u8` so we can do `Encode/Decode` without Problems?
```suggestion
pub struct CryptoTypeId(pub [u8; 4]);
```

primitives/core/src/crypto.rs
```diff
@@ -918,6 +925,13 @@ impl<'a> TryFrom<&'a str> for KeyTypeId {
 	}
 }
 
+/// An identifier for a specific cryptographic algorithm used by a key pair
+#[derive(Copy, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
gnunicorn: see comment below.
```suggestion
#[derive(Copy, Clone, Encode, Decode, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
```

primitives/core/src/ecdsa.rs
```diff
@@ -36,10 +36,15 @@ use crate::{hashing::blake2_256, crypto::{Pair as TraitPair, DeriveJunction, Sec
...
 use serde::{de, Serializer, Serialize, Deserializer, Deserialize};
-use crate::crypto::{Public as TraitPublic, UncheckedFrom, CryptoType, Derive};
+use crate::crypto::{Public as TraitPublic, UncheckedFrom, CryptoType, Derive, CryptoTypeId};
 #[cfg(feature = "full_crypto")]
 use secp256k1::{PublicKey, SecretKey};
 
+/// A value which is passed along with the public key
+/// to define which cryptographic algorithm that key
+/// belongs to.
+pub const ECDSA_CRYPTO_ID: CryptoTypeId = CryptoTypeId("ecdsa");
```
gnunicorn:
```suggestion
pub const ECDSA_CRYPTO_ID: CryptoTypeId = CryptoTypeId(b"ecds");
```

primitives/core/src/ed25519.rs
```diff
@@ -38,10 +38,15 @@ use crate::crypto::{Pair as TraitPair, DeriveJunction, SecretStringError};
...
 use serde::{de, Serializer, Serialize, Deserializer, Deserialize};
-use crate::{crypto::{Public as TraitPublic, UncheckedFrom, CryptoType, Derive}};
+use crate::crypto::{Public as TraitPublic, UncheckedFrom, CryptoType, Derive, CryptoTypeId};
 use sp_runtime_interface::pass_by::PassByInner;
 use sp_std::ops::Deref;
 
+/// A value which is passed along with the public key
+/// to define which cryptographic algorithm that key
+/// belongs to.
+pub const ED25519_CRYPTO_ID: CryptoTypeId = CryptoTypeId("ed25519");
```
gnunicorn:
```suggestion
pub const ED25519_CRYPTO_ID: CryptoTypeId = CryptoTypeId(b"ed25");
```

primitives/core/src/sr25519.rs
```diff
@@ -54,6 +54,11 @@ use sp_runtime_interface::pass_by::PassByInner;
 #[cfg(feature = "full_crypto")]
 const SIGNING_CTX: &[u8] = b"substrate";
 
+/// A value which is passed along with the public key
+/// to define which cryptographic algorithm that key
+/// belongs to.
+pub const SR25519_CRYPTO_ID: CryptoTypeId = CryptoTypeId("sr25519");
```
gnunicorn:
```suggestion
pub const SR25519_CRYPTO_ID: CryptoTypeId = CryptoTypeId(b"sr25");
```

primitives/core/src/testing.rs
```diff
@@ -131,6 +136,63 @@ impl crate::traits::BareCryptoStore for KeyStore {
...
+			match k.0 {
+				sr25519::SR25519_CRYPTO_ID if sr25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
+				ed25519::ED25519_CRYPTO_ID if ed25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
+				_ => None
+			}
+		}).cloned().collect::<Vec<_>>())
+	}
+
+	fn get_keys(&self, id: KeyTypeId) -> Result<Vec<CryptoTypePublicPair>, String> {
+		let ed25519_existing_keys: Vec<Vec<u8>> = self.ed25519_public_keys(id).iter()
```
gnunicorn: see above

client/keystore/src/lib.rs
rakanalh: Result is already defined in the same file as an alias:
https://github.com/paritytech/substrate/blob/ab53994d74b0ad6983432050d540f834a8dc2c7f/client/keystore/src/lib.rs#L53

So i thought i would just use the full import path so that it's clear as to which one we're using here.

client/authority-discovery/src/lib.rs
```diff
@@ -211,10 +212,23 @@ where
...
 
-		for key in self.get_priv_keys_within_authority_set()?.into_iter() {
-			let signature = key.sign(&serialized_addresses);
+		let keys: Vec<CryptoTypePublicPair> = self.get_own_public_keys_within_authority_set()?
+			.into_iter()
+			.map(|k| (sr25519::SR25519_CRYPTO_ID, k.to_raw_vec()))
+			.collect();
+
+		let signatures = self.key_store
+							 .read()
```
bkchr: Whitespaces?

client/keystore/src/lib.rs
rakanalh: Without `cloned()`:
```
   --> client/keystore/src/lib.rs:320:6
    |
320 |           Ok(keys.iter().filter_map(|k| {
    |  ____________^
321 | |             match k.0 {
322 | |                 sr25519::SR25519_CRYPTO_ID if sr25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
323 | |                 ed25519::ED25519_CRYPTO_ID if ed25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
324 | |                 _ => None
325 | |             }
326 | |         }).collect::<Vec<_>>())
    | |______________________________^ expected tuple, found reference
    |
    = note: expected struct `std::vec::Vec<(sp_core::crypto::CryptoTypeId, std::vec::Vec<u8>)>`
               found struct `std::vec::Vec<&(sp_core::crypto::CryptoTypeId, std::vec::Vec<u8>)>`
```

client/keystore/src/lib.rs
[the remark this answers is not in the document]
bkchr: I know that the name exists in the trait, but that doesn't explain why this is with underscore. I also don't understand you not implement it directly in the trait implementation?

client/keystore/src/lib.rs
rakanalh: Done,,, also had to use `cloned` here

client/authority-discovery/src/lib.rs
rakanalh: Not sure what you mean here... the editor as well as github show the func calls vertically aligned. Also, the preceding indentation is all tabs except for 1 space which i assume is to fix the vertical alignment.

rakanalh: Removed the whitespace anyway

primitives/core/src/ecdsa.rs
```diff
@@ -36,10 +36,15 @@ use crate::{hashing::blake2_256, crypto::{Pair as TraitPair, DeriveJunction, Sec
 use crate::crypto::Ss58Codec;
 #[cfg(feature = "std")]
 use serde::{de, Serializer, Serialize, Deserializer, Deserialize};
-use crate::crypto::{Public as TraitPublic, UncheckedFrom, CryptoType, Derive};
+use crate::crypto::{Public as TraitPublic, UncheckedFrom, CryptoType, Derive, CryptoTypeId};
 #[cfg(feature = "full_crypto")]
 use secp256k1::{PublicKey, SecretKey};
 
+/// A value which is passed along with the public key
```
NikVolf: but this is a specific cryptographic algorithm, doc comment seems a bit off

rakanalh: Good point... will amend

client/authority-discovery/src/error.rs
```diff
@@ -46,4 +46,6 @@ pub enum Error {
 	EncodingDecodingScale(codec::Error),
 	/// Failed to parse a libp2p multi address.
 	ParsingMultiaddress(libp2p::core::multiaddr::Error),
+	/// Failed to map signature to public key
+	PublicKeyToSignatureMaping,
```
NikVolf: maybe `MissingSignature(index)` and put missing index which is requested below?

primitives/core/src/testing.rs
```diff
@@ -39,6 +42,24 @@ impl KeyStore {
...
 	}
+
+	fn sr25519_key_pair(&self, id: KeyTypeId, pub_key: &sr25519::Public) -> Option<sr25519::Pair> {
+		self.keys.get(&id)
+			.and_then(|inner|
+				inner.get(pub_key.as_slice())
+					.map(|s| sr25519::Pair::from_string(s, None).expect("`sr25519` seed slice is valid"))
+			)
+	}
+
```
gnunicorn:
```suggestion
```

primitives/core/src/testing.rs
```diff
@@ -131,6 +136,65 @@ impl crate::traits::BareCryptoStore for KeyStore {
...
+			match k.0 {
+				sr25519::SR25519_CRYPTO_ID if sr25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
+				ed25519::ED25519_CRYPTO_ID if ed25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
+				_ => None
+			}
+		}).cloned().collect::<Vec<_>>())
+	}
+
+	fn keys(&self, id: KeyTypeId) -> Result<Vec<CryptoTypePublicPair>, String> {
+		let ed25519_existing_keys: Vec<CryptoTypePublicPair> = self
```
gnunicorn: this still uses the less efficient multiple vec-allocation-code ... copy the code over from the other file?

client/keystore/src/lib.rs
```diff
@@ -276,8 +277,83 @@ impl Store {
...
+			.public_keys_by_type::<sr25519::Public>(id)
+			.map_err(|e| e.to_string())?;
+		let sr25519_existing_keys: Vec<Vec<u8>> = sr25519_existing_keys
+			.iter()
+			.map(|k| k.to_raw_vec())
+			.collect();
+
+		Ok(keys.iter().filter_map(|k| {
+			match k.0 {
+				sr25519::SR25519_CRYPTO_ID if sr25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
```
tomusdrw: This is pretty much quadratic complexity. I think this should be optimized into `Set` intersection. Currently if you pass a long list of `keys` and you have nothing in the keystore it's still linear, as we have to traverse every key, while having less keys in the keystore is a much more likely case - i.e. you pass a list of `1024` validators and you are just one of them.

tomusdrw: Given that we have `fn keys` below, let's try this:
```
let wanted_keys = keys.into_iter().collect::<HashSet<_>>();
let got_keys = self.keys()?.into_iter().collect::<HashSet<_>>();

Ok(wanted_keys.intersection(&got_keys).collect())
```

client/keystore/src/lib.rs
```diff
@@ -276,8 +277,83 @@ impl Store {
...
+				ed25519::ED25519_CRYPTO_ID if ed25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
+				_ => None
+			}
+		}).cloned().collect::<Vec<_>>())
+	}
+
+	fn keys(&self, id: KeyTypeId) -> std::result::Result<Vec<CryptoTypePublicPair>, String> {
```
tomusdrw: We should consider returning a `HashSet` here. Also can we please avoid `String` errors?

client/keystore/src/lib.rs
```diff
@@ -276,8 +277,83 @@ impl Store {
...
+			.map_err(|e| e.to_string())?;
+
+		let sr25519_existing_keys = self
+			.public_keys_by_type::<sr25519::Public>(id)
+			.map_err(|e| e.to_string())?;
+		let sr25519_existing_keys = sr25519_existing_keys
+			.iter().map(|k| (sr25519::SR25519_CRYPTO_ID, k.to_raw_vec()));
+
+		Ok(ed25519_existing_keys
+		   .iter().map(|k| (ed25519::ED25519_CRYPTO_ID, k.to_raw_vec()))
```
tomusdrw: Would be nice if both ed and sr has been consistent. BTW why don't we consider `ecdsa` here?

client/keystore/src/lib.rs
```diff
@@ -276,8 +277,83 @@ impl Store {
...
+		   .chain(sr25519_existing_keys)
+		   .collect())
+	}
+
+	fn sign_with(
+		&self,
+		id: KeyTypeId,
+		key: &CryptoTypePublicPair,
+		msg: &[u8],
+	) -> std::result::Result<Vec<u8>, String> {
```
tomusdrw: Stringified error again.

primitives/application-crypto/test/src/ed25519.rs
```diff
@@ -33,8 +33,5 @@ fn ed25519_works_in_runtime() {
 		.test_ed25519_crypto(&BlockId::Number(0))
 		.expect("Tests `ed25519` crypto.");
 
-	let key_pair = keystore.read().ed25519_key_pair(ED25519, &public.as_ref())
-		.expect("There should be at a `ed25519` key in the keystore for the given public key.");
-
-	assert!(AppPair::verify(&signature, "ed25519", &AppPublic::from(key_pair.public())));
+	assert!(AppPair::verify(&signature, "ed25519", &AppPublic::from(public)));
```
tomusdrw: Shouldn't that test also assert that the key was insterted to the keystore?

primitives/application-crypto/test/src/sr25519.rs
```diff
@@ -34,8 +34,5 @@ fn sr25519_works_in_runtime() {
 		.test_sr25519_crypto(&BlockId::Number(0))
 		.expect("Tests `sr25519` crypto.");
 
-	let key_pair = keystore.read().sr25519_key_pair(SR25519, public.as_ref())
-		.expect("There should be at a `sr25519` key in the keystore for the given public key.");
-
-	assert!(AppPair::verify(&signature, "sr25519", &AppPublic::from(key_pair.public())));
+	assert!(AppPair::verify(&signature, "sr25519", &AppPublic::from(public)));
```
tomusdrw: ditto

primitives/core/src/testing.rs
```diff
@@ -131,6 +135,62 @@ impl crate::traits::BareCryptoStore for KeyStore {
...
 		public_keys.iter().all(|(k, t)| self.keys.get(&t).and_then(|s| s.get(k)).is_some())
 	}
+
+	fn supported_keys(&self, id: KeyTypeId, keys: Vec<CryptoTypePublicPair>) -> Result<Vec<CryptoTypePublicPair>, String> {
+		let ed25519_existing_keys: Vec<Vec<u8>> = self.ed25519_public_keys(id).iter()
+								.map(|k| k.to_raw_vec()).collect();
+		let sr25519_existing_keys: Vec<Vec<u8>> = self.sr25519_public_keys(id).iter()
+								.map(|k| k.to_raw_vec()).collect();
+
+		Ok(keys.iter().filter_map(|k| {
```
tomusdrw:
```suggestion
		Ok(keys.into_iter().filter_map(|k| {
```

primitives/core/src/testing.rs
```diff
@@ -131,6 +135,62 @@ impl crate::traits::BareCryptoStore for KeyStore {
...
+			match k.0 {
+				sr25519::SR25519_CRYPTO_ID if sr25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
+				ed25519::ED25519_CRYPTO_ID if ed25519_existing_keys.contains(&k.1.to_vec()) => Some(k),
+				_ => None
+			}
+		}).cloned().collect::<Vec<_>>())
```
tomusdrw:
```suggestion
		}).collect::<Vec<_>>())
```

Or better just use set intersection again.

primitives/core/src/testing.rs
```diff
@@ -131,6 +135,62 @@ impl crate::traits::BareCryptoStore for KeyStore {
...
+	}
+
+	fn keys(&self, id: KeyTypeId) -> Result<Vec<CryptoTypePublicPair>, String> {
+		let ed25519_existing_keys = self.ed25519_public_keys(id);
```
tomusdrw: Should this be just a default implementation? Seems it's just going to be duplicated in every implementation?

primitives/core/src/traits.rs
```diff
@@ -69,11 +66,70 @@ pub trait BareCryptoStore: Send + Sync {
...
+	/// Sign with key
+	///
+	/// Signs a message with the private key that matches
+	/// the public key passed.
+	fn sign_with(
+		&self,
+		id: KeyTypeId,
+		key: &CryptoTypePublicPair,
+		msg: &[u8],
+	) -> std::result::Result<Vec<u8>, String>;
```
tomusdrw:
```suggestion
	) -> Result<Vec<u8>, String>;
```

primitives/core/src/traits.rs
```diff
@@ -69,11 +66,70 @@ pub trait BareCryptoStore: Send + Sync {
...
+	/// each key given that the key is supported.
+	///
+	/// Returns a list of `Option`s each representing the signature of each key.
+	/// None is return for non-supported keys.
+	fn sign_with_all(
+		&self,
+		id: KeyTypeId,
+		keys: Vec<CryptoTypePublicPair>,
+		msg: &[u8]
+	) -> Vec<Option<Vec<u8>>> {
```
tomusdrw: The return type feels pretty inconsistent with the rest of the methods. Why don't we return a `Result` here but rather opt in for `Option`, while we do the opposite for `sign_with`?

client/authority-discovery/src/lib.rs
```diff
@@ -211,10 +212,23 @@ where
...
+
+		let signatures = self.key_store
+							.read()
+							.sign_with_all(
+								key_types::AUTHORITY_DISCOVERY,
+								keys.clone(),
+								serialized_addresses.as_slice()
+							);
 
+		for (index, signature) in signatures.iter().enumerate() {
```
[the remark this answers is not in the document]
rakanalh: `sign_with_all` returns an `Option<Sig>` where none is for a non-supported key. I added a check there and changed the error to reflect this.

client/keystore/src/lib.rs
```diff
@@ -276,8 +277,83 @@ impl Store {
 }
 
 impl BareCryptoStore for Store {
+	fn supported_keys(
+		&self,
+		id: KeyTypeId,
+		keys: Vec<CryptoTypePublicPair>
+	) -> std::result::Result<Vec<CryptoTypePublicPair>, String> {
```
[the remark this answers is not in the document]
rakanalh: Using `String` for errors is common in `BareCryptoStore` where the function is defined.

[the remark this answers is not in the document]
rakanalh: I will update the code to remove String errors.

primitives/core/src/traits.rs
rakanalh: Would you prefer returning `Vec<Result>` instead?

client/keystore/src/lib.rs
rakanalh: Originally, the keystore did not support it. For example, there was neither `ecdsa_public_keys` nor `ecdsa_key_pair` functions in the keystore. @gnunicorn can you weigh in on why that is the case?

tomusdrw: If it's not supported in the current impl of keystore then fine, we don't need it here either:)

primitives/core/src/traits.rs
tomusdrw: I think it would be more consistent to do so. Or we could change `sign_with` to return `Option` i error is obvious enough. (Alternatively a `Result<_, ()>` to make it more semantically correct)

rakanalh: I got the error that i cannot use `_` in the signature, so i turned that into the full return type that `sign_with` returns.

primitives/core/src/traits.rs
```diff
@@ -69,11 +82,70 @@ pub trait BareCryptoStore: Send + Sync {
...
 	/// Checks if the private keys for the given public key and key type combinations exist.
 	///
 	/// Returns `true` iff all private keys could be found.
 	fn has_keys(&self, public_keys: &[(Vec<u8>, KeyTypeId)]) -> bool;
+
+	/// Sign with key
+	///
+	/// Signs a message with the private key that matches
+	/// the public key passed.
+	fn sign_with(
```
bkchr: It should be mentioned what this function returns and I think it would be easier to return the SCALE encoded signature (instead of the raw bytes).

client/authority-discovery/src/error.rs
```diff
@@ -46,4 +46,8 @@ pub enum Error {
 	EncodingDecodingScale(codec::Error),
 	/// Failed to parse a libp2p multi address.
 	ParsingMultiaddress(libp2p::core::multiaddr::Error),
+	/// Failed to sign using a specific public key
+	MissingSignature(usize),
+	/// Failed to sign using all public keys
+	SigningFailed
```
tomusdrw:
```suggestion
	SigningFailed,
```

client/keystore/src/lib.rs
```diff
@@ -276,8 +298,65 @@ impl Store {
...
+
+		let sr25519_existing_keys = self
+			.public_keys_by_type::<sr25519::Public>(id)
+			.map_err(|e| TraitError::from(e))?
+			.into_iter()
+			.map(|k| (sr25519::SR25519_CRYPTO_ID, k.to_raw_vec()));
+
+		Ok(ed25519_existing_keys
+			.chain(sr25519_existing_keys)
+			.collect::<HashSet<_>>())
```
tomusdrw:
```suggestion
			.collect())
```

This is redundant, since compiler will figure that out from the return type.

client/keystore/src/lib.rs
```diff
@@ -276,8 +298,65 @@ impl Store {
...
+		msg: &[u8],
+	) -> std::result::Result<Vec<u8>, TraitError> {
+		match key.0 {
+			ed25519::ED25519_CRYPTO_ID => {
+				let pub_key = ed25519::Public::from_slice(key.1.as_slice());
+				let key_pair: ed25519::Pair = self
+					.key_pair_by_type::<ed25519::Pair>(&pub_key, id)
+					.map(Into::into)
+					.map_err(|e| TraitError::from(e))?;
+				return Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec());
```
tomusdrw:
```suggestion
				Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec())
```

client/keystore/src/lib.rs
```diff
@@ -276,8 +298,65 @@ impl Store {
...
+			}
+			sr25519::SR25519_CRYPTO_ID => {
+				let pub_key = sr25519::Public::from_slice(key.1.as_slice());
+				let key_pair: sr25519::Pair = self
+					.key_pair_by_type::<sr25519::Pair>(&pub_key, id)
+					.map(Into::into)
+					.map_err(|e| TraitError::from(e))?;
+				return Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec());
```
tomusdrw:
```suggestion
				Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec())
```

We can drop `return` cause the entire expression is returned anyway.

primitives/core/src/testing.rs
```diff
@@ -131,6 +138,58 @@ impl crate::traits::BareCryptoStore for KeyStore {
...
+		id: KeyTypeId,
+		keys: Vec<CryptoTypePublicPair>
+	) -> std::result::Result<HashSet<CryptoTypePublicPair>, Error> {
+		let keys = keys.into_iter().collect::<HashSet<_>>();
+		let all_keys = self.keys(id)?;
+
+		Ok(keys.intersection(&all_keys).cloned().collect())
+	}
+
+	fn keys(&self, id: KeyTypeId) -> std::result::Result<HashSet<CryptoTypePublicPair>, Error> {
```
tomusdrw: As I mentioned earlier I think both of those implementations could be just default methods on the trait. They are currently identical.

primitives/core/src/traits.rs
```diff
@@ -16,16 +16,35 @@
...
+	collections::HashSet,
 	panic::UnwindSafe,
 	sync::Arc,
 };
 
 pub use sp_externalities::{Externalities, ExternalitiesExt};
 
+/// BareCryptoStore error
+#[derive(Debug)]
+pub enum Error {
```
tomusdrw: This `Error` is inside `traits` module, so it should have a more meaningful name, os that we don't have to rename it on import.
```suggestion
pub enum BareCryptoStoreError {
```

primitives/core/src/traits.rs
```diff
@@ -16,16 +16,35 @@
...
+	/// Public key type is not supported
+	KeyNotSupported,
+	/// Pair not found for public key and KeyTypeId
+	PairNotFound(String),
+	/// Validation error
+	ValidationError(String),
+	/// Keystore unavailable
+	Unavailable,
+	/// Programming errors
```
tomusdrw: Why is that needed? I think the convention is to rather call this variant `Other`

primitives/core/src/traits.rs
```diff
@@ -69,12 +85,12 @@ pub trait BareCryptoStore: Send + Sync {
 	/// Find intersection between provided keys and supported keys
 	///
 	/// Provided a list of (CryptoTypeId,[u8]) pairs, this would return
-	/// a filtered list of public keys which are supported by the keystore.
-	fn supported_keys(&self, id: KeyTypeId, keys: Vec<CryptoTypePublicPair>) -> Result<Vec<CryptoTypePublicPair>, String>;
+	/// a filtered set of public keys which are supported by the keystore.
+	fn supported_keys(&self, id: KeyTypeId, keys: Vec<CryptoTypePublicPair>) -> Result<HashSet<CryptoTypePublicPair>, Error>;
```
gnunicorn: @tomusdrw did you really mean returning `HashSet` here? I understood that comment as checking against a `hashset` internally, but not returning one. I am also wondering, if we want to keep an ordered return value ... just considering that many implementations would just use the first one (which is not deterministically possible with a `HashSet`), an external signer could express the preference in keys through the order. When returning a `HashSet` it can't anymmore, it is pure luck (or it must return a `HashSet` of len 1 to express that preference). I am not sure how important that is and whether this should be implied or maybe even explicitly said in the docs ... 

I feel 6 out 10, that a Vec is the more flexible solution...

primitives/core/src/traits.rs
```diff
@@ -40,9 +59,6 @@ pub trait BareCryptoStore: Send + Sync {
 		id: KeyTypeId,
 		seed: Option<&str>,
 	) -> Result<sr25519::Public, String>;
```
tomusdrw: We should log a separate issue to change these return types to use the new `Error` type as well.

primitives/core/src/traits.rs
```diff
@@ -40,9 +59,6 @@ pub trait BareCryptoStore: Send + Sync {
...
-	/// Returns the sr25519 key pair for the given key type and public key combination.
-	fn sr25519_key_pair(&self, id: KeyTypeId, pub_key: &sr25519::Public) -> Option<sr25519::Pair>;
-
 	/// Returns all ed25519 public keys for the given key type.
 	fn ed25519_public_keys(&self, id: KeyTypeId) -> Vec<ed25519::Public>;
```
tomusdrw: Should we remove these methods now? I find it weird that we now have two different set of APIs, returning different types but essentially doing the same thing. Happy to have it done in a follow up PR, don't want to block this one.

primitives/core/src/traits.rs
gnunicorn: Secondly just realised that for `sign_with` we only take a reference to `CryptoTypePublicPair`, while here we consume the entire type. Just thinking about the use case and am wondering, if we assumed we get the keys from on-chain storage, do we get them as references? If so, wouldn't the current API require an unneeded `clone`, while the following would not?

```suggestion
	fn supported_keys<'a>(&self, id: KeyTypeId, keys: Vec<&'a CryptoTypePublicPair>) -> Result<Vec<&'a CryptoTypePublicPair>, Error>;
```

:thinking: ?!?

primitives/core/src/traits.rs
```diff
@@ -69,12 +85,12 @@ pub trait BareCryptoStore: Send + Sync {
...
-	/// a filtered list of public keys which are supported by the keystore.
-	fn supported_keys(&self, id: KeyTypeId, keys: Vec<CryptoTypePublicPair>) -> Result<Vec<CryptoTypePublicPair>, String>;
+	/// a filtered set of public keys which are supported by the keystore.
+	fn supported_keys(&self, id: KeyTypeId, keys: Vec<CryptoTypePublicPair>) -> Result<HashSet<CryptoTypePublicPair>, Error>;
 	/// List all supported keys
 	///
-	/// Get a list of public keys the signer supports.
-	fn keys(&self, id: KeyTypeId) -> Result<Vec<CryptoTypePublicPair>, String>;
+	/// Returns a set of public keys the signer supports.
+	fn keys(&self, id: KeyTypeId) -> Result<HashSet<CryptoTypePublicPair>, Error>;
```
gnunicorn: (re my earlier comments, this would still return the entire type I think, because they are not bound to an incoming lifetime..?!?)

client/keystore/src/lib.rs
```diff
@@ -276,8 +298,65 @@ impl Store {
 }
 
 impl BareCryptoStore for Store {
+	fn supported_keys(
+		&self,
+		id: KeyTypeId,
+		keys: Vec<CryptoTypePublicPair>
+	) -> std::result::Result<HashSet<CryptoTypePublicPair>, TraitError> {
+		let keys = keys.into_iter().collect::<HashSet<_>>();
+		let all_keys = self.keys(id)?;
```
bkchr: I think it would be much faster if you directly check the filesystem etc for the incoming keys, instead of loading *all* keys and doing an intersection.

tomusdrw: Well, that's sort of the opposite what I suggested :) Aren't we going to have like thousands keys in `Vec` and then only couple in the storage? Also this can be optimized to not read from disk if needed.

primitives/core/src/traits.rs
tomusdrw: I didn't mean `HashSet` here, I suggested it might be useful for `fn keys` but at the same time we can just return `Vec`s everywhere and use `HashSet` internally everywhere (this would mean that `supported_keys` should not use `keys` and then convert from Vec to HashSet but rather copy the implementation and collect to a `HashSet`)

primitives/application-crypto/src/lib.rs
```diff
@@ -271,6 +271,8 @@ macro_rules! app_crypto_public_common {
 		}
 
 		impl $crate::Public for Public {
+			const CRYPTO_TYPE_ID: $crate::CryptoTypeId = $crate::CryptoTypeId($key_type.0);
```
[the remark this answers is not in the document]
rakanalh: `.0` from `$key_type` is a `[u8; 4]` which can be used to construct a new CryptoTypeId

primitives/core/src/traits.rs
rakanalh: Renamed to `Other`. The use case for this is converting keystore error to `BareKeyStoreError` where keystore has `Io` and `Json` errors that could be mapped to `Other` here.

primitives/core/src/traits.rs
```diff
@@ -69,11 +82,70 @@ pub trait BareCryptoStore: Send + Sync {
...
+		}
+		Err(BareCryptoStoreError::KeyNotSupported)
+	}
+
+	/// Sign with all keys
+	///
+	/// Provided a list of public keys, sign a message with
+	/// each key given that the key is supported.
+	///
+	/// Returns a list of `Option`s each representing the signature of each key.
```
mxinden: According to the signature it returns a list of `Result`'s.

client/authority-discovery/src/lib.rs
```diff
@@ -211,19 +211,39 @@ where
...
-		for key in self.get_priv_keys_within_authority_set()?.into_iter() {
-			let signature = key.sign(&serialized_addresses);
+		let keys: Vec<CryptoTypePublicPair> = self.get_own_public_keys_within_authority_set()?
+			.into_iter()
+			.map(Into::into)
+			.collect();
 
+		let signatures = self.key_store
+			.read()
+			.sign_with_all(
```
mxinden: :tada: great to not having to handle private keys here!

client/authority-discovery/src/error.rs
```diff
@@ -46,4 +46,8 @@ pub enum Error {
 	EncodingDecodingScale(codec::Error),
 	/// Failed to parse a libp2p multi address.
 	ParsingMultiaddress(libp2p::core::multiaddr::Error),
+	/// Failed to sign using a specific public key
+	MissingSignature(usize),
```
mxinden: I don't understand how the index would be useful for someone needing to debug this. Wouldn't something like the public key be more helpful?

client/authority-discovery/src/error.rs
```diff
@@ -46,4 +46,8 @@ pub enum Error {
...
+	/// Failed to sign using all public keys
+	SigningFailed,
```
mxinden: None of the error names above are postfixed with `Failed`. I don't care about the name but would prefer to stay consistent, e.g.:
```suggestion
	Signing,
```

client/authority-discovery/src/lib.rs
```diff
@@ -211,19 +211,39 @@ where
...
+				key_types::AUTHORITY_DISCOVERY,
+				keys.clone(),
+				serialized_addresses.as_slice(),
+			)
+			.map_err(|_| Error::SigningFailed)?;
+
+		for (index, (sign_result, key)) in signatures.iter().zip(keys).enumerate() {
 			let mut signed_addresses = vec![];
+
+			// sign_with_all returns Option<Signature> where the signature
```
mxinden: As far as I can tell it returns a `Result<_, _>`.

client/authority-discovery/src/lib.rs
```diff
@@ -211,19 +211,39 @@ where
...
+			// is None for a public key that is not supported.
+			// Verify that all signatures exist for all provided keys.
+			if sign_result.is_err() {
```
mxinden: I would prefer something like the following to not have an `unwrap`:

```suggestion
			let signature = sign_result.map_err(|_| Err(Error::MissingSignature(index)))?;
```

client/keystore/src/lib.rs
```diff
@@ -276,8 +298,65 @@ impl Store {
...
+			.into_iter()
+			.map(|k| (ed25519::ED25519_CRYPTO_ID, k.to_raw_vec()));
+
+		let sr25519_existing_keys = self
+			.public_keys_by_type::<sr25519::Public>(id)
+			.map_err(|e| TraitError::from(e))?
+			.into_iter()
+			.map(|k| (sr25519::SR25519_CRYPTO_ID, k.to_raw_vec()));
+
+		Ok(ed25519_existing_keys
```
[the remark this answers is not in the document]
rakanalh: I considered your suggestion. The code after this request becomes something like this:
```
/// Raw public keys
///
/// Returns a list of tuples of (key_type, public_key)
pub fn raw_public_keys(&self) -> Vec<(KeyTypeId, Vec<u8>)> {
    let mut public_keys: Vec<(KeyTypeId, Vec<u8>)> = self.additional.keys()
        .into_iter()
        .cloned()
        .collect();

    if let Some(path) = &self.path {
        for entry in fs::read_dir(&path)? {
            let entry = entry?;
            let path = entry.path();

            // skip directories and non-unicode file names (hex is unicode)
            if let Some(name) = path.file_name().and_then(|n| n.to_str()) {
                match hex::decode(name) {
                    Ok(ref hex) if hex.len() > 4 => {
                        let mut key_type: [u8; 4] = Default::default();
                        key_type.copy_from_slice(&hex[0..4]);
                        let key_type = KeyTypeId(key_type);
                        let public = hex[4..].to_vec();
                        public_keys.push((key_type, public));
                    }
                    _ => continue,
                }
            }
        }
    }

    public_keys
}

/// Get public keys of all stored keys that match the given key type.
pub fn public_keys_by_type<TPublic: Public>(&self, key_type: KeyTypeId) -> Result<Vec<TPublic>> {
    let raw_public_keys = self.raw_public_keys()?;
    Ok(raw_public_keys.into_iter().filter_map(|(ty, public)| {
        if ty == key_type {
            Some(TPublic::from_slice(&public[..]))
        } else {
            None
        }
    }).collect())
}
```

However, calling into `raw_public_keys` directly from `keys` means that i have to map `(KeyTypeId, Vec<u8>)`, which is what the above implementation shows, to `(CryptoTypeId, Vec<u8>)` which is the return type for `keys. I could not figure out a way to map KeyTypeId -> CryptoTypeId and i don't think there is a way as they're unrelated.

Unless you have a suggestion on the problem i just stated, i would keep the code as-is.

primitives/core/src/traits.rs
rakanalh: This is kinda related to my comment here:
https://github.com/paritytech/substrate/pull/4925/files#r384537219

What we can do is generalize them a bit by renaming those methods to something like `public_keys(CryptoTypeId, KeyTypeId)` but we're going to be left by functions such as `sr25519_generate_new` and similar functions which also need to somehow be generalized in the same fashion. This gets the scope of the PR a bit bigger though.

primitives/core/src/traits.rs
rakanalh: I had a look at how we use the codec crate... `Encode::encode` already returns a `Vec<u8>` as well. I am pretty sure i don't follow what you're suggesting. Is there code somewhere i can look into that does this?

client/authority-discovery/src/error.rs
rakanalh: Done,,, the enum variant should receive `CryptoTypePublicPair`

primitives/core/src/traits.rs
rakanalh: I understand the concern is about the order of keys passed to this function is not preserved when using a HashSet. As far as i can see, even if we end up using the hashset internally, we would still lose the order unless we re-sort the keys:
```
let intersection = keys.intersection(...).....;
let result: Vec<CryptoTypePublicPair> = keys.into_iter().filter(|key| intersection.contains(key));
```

Does this sound good?

client/keystore/src/lib.rs
```diff
@@ -276,8 +298,47 @@ impl Store {
...
+			}
+			sr25519::CRYPTO_ID => {
+				let pub_key = sr25519::Public::from_slice(key.1.as_slice());
+				let key_pair: sr25519::Pair = self
+					.key_pair_by_type::<sr25519::Pair>(&pub_key, id)
+					.map(Into::into)
+					.map_err(|e| TraitError::from(e))?;
+				Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec())
+			}
+			_ => Err(TraitError::KeyNotSupported)
```
tomusdrw: I think the error should contain the missing Id to make it easier to debug.

primitives/application-crypto/src/ed25519.rs
```diff
@@ -19,16 +19,35 @@
...
 use sp_std::vec::Vec;
+use sp_core::crypto::CryptoTypeId;
 
 pub use sp_core::ed25519::*;
 
+/// An identifier used to match public keys against ed25519 keys
+pub const CRYPTO_ID: CryptoTypeId = CryptoTypeId(*b"ed25");
+
 mod app {
+    use sp_core::crypto::{CryptoTypePublicPair, Public as TraitPublic};
```
tomusdrw: please reformat using tabs

client/rpc/src/author/tests.rs
```diff
@@ -21,7 +21,8 @@ use assert_matches::assert_matches;
 use codec::Encode;
 use sp_core::{
 	H256, blake2_256, hexdisplay::HexDisplay, testing::{ED25519, SR25519, KeyStore},
-	traits::BareCryptoStorePtr, ed25519, crypto::{Pair, Public},
+	traits::BareCryptoStorePtr, ed25519, sr25519,
+    crypto::{CryptoTypePublicPair, Pair, Public},
```
tomusdrw: spaces, please reformat.

primitives/application-crypto/src/ed25519.rs
```diff
@@ -19,16 +19,35 @@
...
+
 	crate::app_crypto!(super, ED25519);
 
 	impl crate::traits::BoundToRuntimeAppPublic for Public {
 		type Public = Self;
 	}
+
+    impl From<Public> for CryptoTypePublicPair {
+        fn from(key: Public) -> Self {
+            (&key).into()
```
tomusdrw: Nitpick: isn't there a way to avoid copying here? I.e. extract raw vec directly from `Public`?

primitives/application-crypto/src/sr25519.rs
```diff
@@ -19,16 +19,35 @@
...
 use sp_std::vec::Vec;
+use sp_core::crypto::CryptoTypeId;
 
 pub use sp_core::sr25519::*;
 
+/// An identifier used to match public keys against sr25519 keys
+pub const CRYPTO_ID: CryptoTypeId = CryptoTypeId(*b"sr25");
+
 mod app {
+    use sp_core::crypto::{CryptoTypePublicPair, Public as TraitPublic};
```
tomusdrw: spaces again.

primitives/application-crypto/src/sr25519.rs
```diff
@@ -19,16 +19,35 @@
...
+
 	crate::app_crypto!(super, SR25519);
 
 	impl crate::traits::BoundToRuntimeAppPublic for Public {
 		type Public = Self;
 	}
+
+    impl From<Public> for CryptoTypePublicPair {
+        fn from(key: Public) -> Self {
+            (&key).into()
```
tomusdrw: same here

primitives/application-crypto/test/src/ed25519.rs
```diff
@@ -33,8 +36,7 @@ fn ed25519_works_in_runtime() {
 		.test_ed25519_crypto(&BlockId::Number(0))
 		.expect("Tests `ed25519` crypto.");
 
-	let key_pair = keystore.read().ed25519_key_pair(ED25519, &public.as_ref())
-		.expect("There should be at a `ed25519` key in the keystore for the given public key.");
-
-	assert!(AppPair::verify(&signature, "ed25519", &AppPublic::from(key_pair.public())));
+	let supported_keys = keystore.read().keys(ED25519).unwrap();
+	assert!(true, supported_keys.contains(&public.into()));
```
tomusdrw:
```suggestion
	assert!(supported_keys.contains(&public.into()));
```

primitives/application-crypto/test/src/sr25519.rs
```diff
@@ -34,8 +37,7 @@ fn sr25519_works_in_runtime() {
 		.test_sr25519_crypto(&BlockId::Number(0))
 		.expect("Tests `sr25519` crypto.");
 
-	let key_pair = keystore.read().sr25519_key_pair(SR25519, public.as_ref())
-		.expect("There should be at a `sr25519` key in the keystore for the given public key.");
-
-	assert!(AppPair::verify(&signature, "sr25519", &AppPublic::from(key_pair.public())));
+	let supported_keys = keystore.read().keys(SR25519).unwrap();
+	assert!(true, supported_keys.contains(&public.into()));
```
tomusdrw:
```suggestion
	assert!(supported_keys.contains(&public.into()));
```

primitives/authority-discovery/src/lib.rs
```diff
@@ -21,8 +21,25 @@
 use sp_std::vec::Vec;
 
 mod app {
-	use sp_application_crypto::{app_crypto, key_types::AUTHORITY_DISCOVERY, sr25519};
+	use sp_application_crypto::{
+        CryptoTypePublicPair,
```
tomusdrw: even more spaces

primitives/core/src/crypto.rs
```diff
@@ -918,6 +920,20 @@ impl<'a> TryFrom<&'a str> for KeyTypeId {
...
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash, Encode, Decode)]
+pub struct CryptoTypeId(pub [u8; 4]);
+
+/// A type alias of CryptoTypeId & a public key
+#[derive(Debug, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash, Encode, Decode)]
+pub struct CryptoTypePublicPair(pub CryptoTypeId, pub Vec<u8>);
+
+impl sp_std::fmt::Display for CryptoTypePublicPair {
+    fn fmt(&self, f: &mut sp_std::fmt::Formatter) -> sp_std::fmt::Result {
+        write!(f, "{:#?}-{:x?}", self.0, self.1)
```
tomusdrw: Might be actually good to attempt to decode as string here and only fall back to hex when it fails (I mean `self.0`)

For `self.1` you should use `HexFormat` that will result in consistent formatting with the rest of the codebase.

primitives/application-crypto/src/sr25519.rs
```diff
@@ -19,16 +19,35 @@
 use crate::{RuntimePublic, KeyTypeId};
 
 use sp_std::vec::Vec;
+use sp_core::crypto::CryptoTypeId;
 
 pub use sp_core::sr25519::*;
 
+/// An identifier used to match public keys against sr25519 keys
+pub const CRYPTO_ID: CryptoTypeId = CryptoTypeId(*b"sr25");
```
tomusdrw: Why is it defined for the second time here? Can't we use the same thing that is in `sp_core`?

primitives/core/src/testing.rs
```diff
@@ -16,9 +16,15 @@
...
+use crate::{
+	crypto::{Pair, Public},
+	ed25519, sr25519,
+	traits::BareCryptoStoreError
+};
+#[cfg(feature = "std")]
+use std::collections::HashSet;
 
 /// Key type for generic Ed25519 key.
 pub const ED25519: KeyTypeId = KeyTypeId(*b"ed25");
```
tomusdrw: Just noticed that we already have `KeyTypeId`, why do we introduce `CryptoTypeId` then?

primitives/core/src/traits.rs
tomusdrw: Fine to do a clean-up in a second PR, but please create an issue for this after the PR is merged.

primitives/core/src/testing.rs
tomusdrw: Oh, I guess `KeyTypeId` is actually "purpose" or "application", right? Can we please rename it to make it clear? Also This constant should most likely be renamed to something like `TEST_APP: CryptoAppId = CryptoAppId(*b"test");` to avoid confusion with `CryptoTypeId`.

primitives/application-crypto/src/ed25519.rs
rakanalh: hmm, i am not sure i see the "copying" here. The purpose of this line is to call `from` from this block:
```
impl From<&Public> for CryptoTypePublicPair {
```
Which is below the one we're commenting on. Can you elabore on where the copy happens?

primitives/core/src/traits.rs
rakanalh: This is done.

primitives/core/src/traits.rs
gnunicorn: I don't understand the `keys.intersection()` part. [`Intersection`](https://doc.rust-lang.org/std/collections/hash_set/struct.Intersection.html) internally is just a wrapper type, that then becomes an iterator (for our usecase). Considering that we aren't needing it, it is probably even more efficient to just do it bare:

```rust
fn supported_keys<'a>(&self, id: KeyTypeId, keys: Vec<&'a CryptoTypePublicPair>) -> Result<Vec<&'a CryptoTypePublicPair>, Error> {
	let supported_keys = self.iter_keys(id).collect::<HashSet<_>>() //internal function that returns an iterator
	keys.into_iter().filter(|key| supported_keys.contains(key).collect::<Vec<_>>()
}
```

Considering that `iter_keys()` is an internal function similarly to what `.keys()` does now, but returns an iterator rather than the collected `Vec` (for performance). `.keys()` would use the same `iter_keys()` internally to collect the `Vec`. Arguably this could even be part of the trait in `std` ... but I dunno if that is a good pattern (to have different trait interfaces during `std`/`no_std`)

burdges: Why should order matter here?  We're never deterministic with signing anyways.

client/keystore/src/lib.rs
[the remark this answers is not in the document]
rakanalh: Addressed

primitives/core/src/traits.rs
rakanalh: Implemented @gnunicorn's comment about `HashSet`s but not including the `iter_keys` suggestion

client/keystore/src/lib.rs
```diff
@@ -263,8 +283,13 @@ impl Store {
 				if let Some(name) = path.file_name().and_then(|n| n.to_str()) {
 					match hex::decode(name) {
 						Ok(ref hex) if hex.len() > 4 => {
-							if &hex[0..4] != &key_type.0 { continue }
-							let public = TPublic::from_slice(&hex[4..]);
+							let mut key_type: [u8; 4] = Default::default();
```
gnunicorn: Why do we have to make this copy now that we didn't do before to be able to compare the id?

client/keystore/src/lib.rs
```diff
@@ -275,29 +300,23 @@ impl Store {
...
+		id: KeyTypeId
+	) -> std::result::Result<Vec<CryptoTypePublicPair>, TraitError> {
+		let raw_keys = self.raw_public_keys(id)?;
+		Ok(raw_keys.into_iter()
+		   .map(|k| vec![
+			   CryptoTypePublicPair(sr25519::CRYPTO_ID, k.clone()),
+			   CryptoTypePublicPair(ed25519::CRYPTO_ID, k.clone())
+		   ])
+		   .flatten()
+		   .collect())
```
gnunicorn: I feel like [`fold`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold) would be a bit nicer here, rather than allocating a bunch of 2-elem vectors for each items, just do
```suggestion
		   .fold(Vec::new(), |mut v, k| {
		   		v.push(CryptoTypePublicPair(sr25519::CRYPTO_ID, k.clone()));
		   		v.push(CryptoTypePublicPair(ed25519::CRYPTO_ID, k.clone()));
		   		v
		   	});
```

primitives/core/src/testing.rs
```diff
@@ -66,6 +66,20 @@ impl KeyStore {
 
 #[cfg(feature = "std")]
 impl crate::traits::BareCryptoStore for KeyStore {
+	fn keys(&self, id: KeyTypeId) -> Result<Vec<CryptoTypePublicPair>, BareCryptoStoreError> {
+		self.keys
+			.get(&id)
+			.map(|map| {
+				Ok(map.keys().map(|k| vec![
```
gnunicorn: see above.

client/keystore/src/lib.rs
rakanalh: I reverted to using `&hex[0..4] != id.0` instead of copying.

gnunicorn:
refs #4689

client/keystore/src/lib.rs
```diff
@@ -253,71 +296,106 @@ impl Store {
...
+		msg: &[u8],
+	) -> std::result::Result<Vec<u8>, TraitError> {
+		match key.0 {
+			ed25519::CRYPTO_ID => {
+				let pub_key = ed25519::Public::from_slice(key.1.as_slice());
+				let key_pair: ed25519::Pair = self
+					.key_pair_by_type::<ed25519::Pair>(&pub_key, id)
+					.map(Into::into)
+					.map_err(|e| TraitError::from(e))?;
+				Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec())
```
bkchr:
```suggestion
				Ok(key_pair.sign(msg).encode())
```

client/keystore/src/lib.rs
```diff
@@ -253,71 +296,106 @@ impl Store {
...
+			}
+			sr25519::CRYPTO_ID => {
+				let pub_key = sr25519::Public::from_slice(key.1.as_slice());
+				let key_pair: sr25519::Pair = self
+					.key_pair_by_type::<sr25519::Pair>(&pub_key, id)
+					.map(Into::into)
+					.map_err(|e| TraitError::from(e))?;
+				Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec())
```
bkchr:
```suggestion
				Ok(key_pair.sign(msg).encode())
```

client/keystore/src/lib.rs
```diff
@@ -253,71 +296,106 @@ impl Store {
...
+					.key_pair_by_type::<ed25519::Pair>(&pub_key, id)
+					.map(Into::into)
+					.map_err(|e| TraitError::from(e))?;
+				Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec())
+			}
+			sr25519::CRYPTO_ID => {
+				let pub_key = sr25519::Public::from_slice(key.1.as_slice());
+				let key_pair: sr25519::Pair = self
+					.key_pair_by_type::<sr25519::Pair>(&pub_key, id)
+					.map(Into::into)
```
bkchr: Why?

client/keystore/src/lib.rs
```diff
@@ -253,71 +296,106 @@ impl Store {
...
+		id: KeyTypeId,
+		key: &CryptoTypePublicPair,
+		msg: &[u8],
+	) -> std::result::Result<Vec<u8>, TraitError> {
+		match key.0 {
+			ed25519::CRYPTO_ID => {
+				let pub_key = ed25519::Public::from_slice(key.1.as_slice());
+				let key_pair: ed25519::Pair = self
+					.key_pair_by_type::<ed25519::Pair>(&pub_key, id)
+					.map(Into::into)
```
bkchr: Same

primitives/core/src/testing.rs
```diff
@@ -131,6 +154,42 @@ impl crate::traits::BareCryptoStore for KeyStore {
...
+				let key_pair: ed25519::Pair = self
+					.ed25519_key_pair(id, &ed25519::Public::from_slice(key.1.as_slice()))
+					.map(Into::into)
+					.ok_or(BareCryptoStoreError::PairNotFound("ed25519".to_owned()))?;
+				return Ok(<[u8; 64]>::from(key_pair.sign(msg)).to_vec());
+			}
+			sr25519::CRYPTO_ID => {
+				let key_pair: sr25519::Pair = self
+					.sr25519_key_pair(id, &sr25519::Public::from_slice(key.1.as_slice()))
+					.map(Into::into)
```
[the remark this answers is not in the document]
rakanalh: Probably leftovers from previous iterations... all removed

primitives/core/src/traits.rs
```diff
@@ -67,11 +82,78 @@ pub trait BareCryptoStore: Send + Sync {
...
 	///
 	/// Returns `true` iff all private keys could be found.
 	fn has_keys(&self, public_keys: &[(Vec<u8>, KeyTypeId)]) -> bool;
+
+	/// Sign with key
+	///
+	/// Signs a message with the private key that matches
+	/// the public key passed.
+	///
+	/// Returns the signature if key is found & supported,
```
bkchr:
```suggestion
	/// Returns the SCALE encoded signature if key is found & supported,
```

bkchr:
Before merging, I would like to test this on one of our Kusama validators.

rakanalh:
Opps, screwed something up... sorry about that and i'll fix it

ddorgan:
Deployed to one validator.

gnunicorn:
@ddorgan your silence means we are not seeing issues on that validator?!?

PR #4887, opened 2020-02-11
Add a crate to test the in-browser light client.

expenses:
At the moment, the wasm light client is at a high risk of being broken by incoming commits. This PR adds a simple test that makes sure the light client connects to peers and starts syncing. This requires #4880 to me merged first.

It's pretty messy, but I'm not sure if there's a great way to improve that.

PR #4880, opened 2020-02-10
Re-enable networking on wasm

expenses:
~~Waiting for a new version of libp2p to be published that contains https://github.com/libp2p/rust-libp2p/pull/1439.~~

Issue #4679, opened 2020-01-20
Network connection leak


arkpar:
Currently substrate is based on `libp2p 0.14.0-alpha.1` which has a few issues:

1. Network opens a lot of TCP connections. After running for a few hours on kusama the number of connections reported by `lsof` is over 2000.
2. Each connection preallocates at least 176 kbytes of memory for buffers in the noise protocol.
Which leads to memory usage in hundreds of megabytes.
3. Each connection spawns a short-lived thread, which leads to a leak in `sysinfo` crate.

Each of these should be fixed as soon as possible.  Substrate already has a limit for protocol connections. That limit must be enforced as a global limit for TCP connections, including Kademlia and all other protocols.

All buffers should be preallocated to some small size, like 1k. The rest should be allocated lazily.

twittner:
> 1. Network opens a lot of TCP connections. After running for a few hours on kusama the number of connections reported by `lsof` is over 2000.

If I am not mistaken most of those connections are not established yet, e.g.

```fish
> ls -l /proc/(pgrep polkadot)/fd | wc -l
1407
> lsof -n -p (pgrep polkadot) | rg ESTABLISHED | wc -l
185
> lsof -n -p (pgrep polkadot) | rg SYN_SENT | wc -l
1134
```

> 3. Each connection spawns a short-lived thread, which leads to a leak in `sysinfo` crate.

These threads are actually not as short lived as one would hope as for unreachable destinations each runs until the maximum number of initial SYN retries has been reached (cf. man 7 tcp). E.g. on a Linux system:

```
tcp_syn_retries (integer; default: 5; since Linux 2.2)
    The  maximum  number  of times initial SYNs for an active TCP
    connection attempt will be retransmitted. This value should not
    be higher than 255.  The default value is 5, which corresponds
    to approximately 180 seconds.
```

arkpar:
> If I am not mistaken most of those connections are not established yet, e.g.

This might be the case. However I've also put a counter of `NoiseOuput` instances that went to thousands after a few hours as well. It is not clear then why a connection that is not established yet requires preallocated buffers. And also why there are that many connection attempts in the first place.

> These threads are actually not as short lived as one would hope as for unreachable destinations each runs until the maximum number of initial SYN retries has been reached (cf. man 7 tcp). E.g. on a Linux system:
> 
> ```
> tcp_syn_retries (integer; default: 5; since Linux 2.2)
>     The  maximum  number  of times initial SYNs for an active TCP
[...]

PR #4928, opened 2020-02-14
Update libp2p to 0.16

tomaka:
Unlocks a few cool things, notably #4880 and #4887. cc'ing @expenses
Should also reduce the memory usage of connections, cc #4679 

Changes should be straight-forward.


expenses:
LGTM!

PR #4930, opened 2020-02-14
Remove rename for finalized event and add some docs.

tomusdrw:
Extra docs and removal of serde rename as a follow up for #4740, also see #4438

### DIFF
--- a/primitives/transaction-pool/src/pool.rs
+++ b/primitives/transaction-pool/src/pool.rs
@@ -98,7 +98,6 @@ pub enum TransactionStatus<Hash, BlockHash> {
 	/// The transaction has been broadcast to the given peers.
 	Broadcast(Vec<String>),
 	/// Transaction has been included in block with given hash.
-	#[serde(rename = "finalized")] // See #4438
 	InBlock(BlockHash),
 	/// The block this transaction was included in has been retracted.
 	Retracted(BlockHash),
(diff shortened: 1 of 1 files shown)


primitives/transaction-pool/src/pool.rs
```diff
@@ -67,16 +67,22 @@ impl PoolStatus {
...
-///		- `Finalized`
+///		- `InBlock`
 ///		- `Invalid`
 ///		- `Usurped`
 ///		- `Dropped`
+///	4. Re-entering the pool:
+///		- `Retracted`
+///	5. Block finalized:
+///		- `Finalized`
+///		- `FinalityTimeout`
```
bkchr: What is the timeout?

tomusdrw: After roughly 512 (a hardcoded constant; currently arbitrary, not based on any research) block pass with no finality we consider that a timeout. This is to prevent prevent leaking memory (and watchers) in case of either:
1. A bug in the code (like missing notification for a particular block)
2. Finality being temporarily broken (or just unavailable)

Note that you are free to re-subscribe for a particular hash after the timeout to wait for even longer (indefinitely)

bkchr: Yeah that is a good idea :) However, I think we should document it. Maybe at least say that `512` is a minimum?

tomusdrw: Done.

PR #4931, opened 2020-02-14
Allow customizing the NetworkBehaviour

tomaka:
Based on top of #4928 
Opening as a draft because it's not totally finished and because I'll rebase on top of master once #4928 is merged. You can review now, but I'm opening this PR more to show my intent and so we can try it in substrate-lfs.

Allows injecting an implementation of `NetworkBehaviour` into the networking code.

Contrary to `NetworkSpecialization`, this allows doing basically whatever you want: open new connections, kill connections, use connections that have been opened by other subsystems, and so on.

Note that this is an extremely powerful tool, but we'll very probably change the API of `NetworkBehaviour` in the future.

Also cc'ing @rphmeier since this can be useful for Polkadot.


tomaka:
After discussing with @gnunicorn, I'm going to close this.

This PR would expose the libp2p `NetworkBehaviour` trait in the public API of `sc-network`, and with the coming stabilization I don't want to be block future changes due to not being allowed to break this trait.

The purpose of this PR was to make it possible to customize the networking protocol from outside Substrate, but the notifications API and the future request-response API should be enough for that purpose.

PR #4932, opened 2020-02-14
Rename remaining occurences of SRML to FRAME

apopiak:
Rename remaining occurences of SRML to FRAME.


shawntabrizi:
@apopiak can you also search for changing `module` to `pallet` where appropriate?

Got to be a little careful cause `module` is correct for Rust modules.

apopiak:
> @apopiak can you also search for changing `module` to `pallet` where appropriate?
> 
> Got to be a little careful cause `module` is correct for Rust modules.

I'll have a look

apopiak:
@shawntabrizi Had a look and I have no idea how to systematically change that without combing through every location and checking so I'd like to put it in a separate PR at some other point.

apopiak:
> ```
> [...]
> frame/support/procedural/src/lib.rs
> 102:///   [`KeyFormat`](../srml_support/storage/generator/trait.KeyFormat.html) trait, which
> ```
> 
> oh actually that first one makes sense. Does the second one make sense?

that seems very outdated, good spot, thanks!

shawntabrizi:
I made a few quick updates, but I agree, module -> pallet is a bigger issue.

So many uses of the word module ingrained in the framework, for example:

```
/// The metadata of a runtime.
#[derive(Eq, Encode, PartialEq, RuntimeDebug)]
#[cfg_attr(feature = "std", derive(Decode, Serialize))]
pub struct RuntimeMetadataV11 {
	/// Metadata of all the modules.
	pub modules: DecodeDifferentArray<ModuleMetadata>,
	/// Metadata of the extrinsic.
	pub extrinsic: ExtrinsicMetadata,
}

/// The latest version of the metadata.
pub type RuntimeMetadataLastVersion = RuntimeMetadataV11;

/// All metadata about an runtime module.
#[derive(Clone, PartialEq, Eq, Encode, RuntimeDebug)]
#[cfg_attr(feature = "std", derive(Decode, Serialize))]
pub struct ModuleMetadata {
	pub name: DecodeDifferentStr,
	pub storage: Option<DecodeDifferent<FnEncode<StorageMetadata>, StorageMetadata>>,
	pub calls: ODFnA<FunctionMetadata>,
	pub event: ODFnA<EventMetadata>,
	pub constants: DFnA<ModuleConstantMetadata>,
	pub errors: DFnA<ErrorMetadata>,
}
```

No idea what the long term plan is.

cecton:
Do we already have an issue for module/pallet?

gnunicorn:
Runtime check is a false positive, suppose it can be ignored. Question, whether anything else needs addressing before it can be merged?

PR #4936, opened 2020-02-14
Refactor BlockImportParams to be non_exhaustive

sorpaas:
This marks `BlockImportParams` as `non_exhaustive`. From time to time we add new field into the struct. By making sure users use initialization function we avoid too much breakage.

### DIFF
--- a/bin/node/cli/src/service.rs
+++ b/bin/node/cli/src/service.rs
@@ -369,7 +369,7 @@ pub fn new_light(config: NodeConfiguration)
 
 #[cfg(test)]
 mod tests {
-	use std::{sync::Arc, collections::HashMap, borrow::Cow, any::Any};
+	use std::{sync::Arc, borrow::Cow, any::Any};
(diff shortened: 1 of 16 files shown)


bkchr:
Not sure that is the correct way we should go. I think the problem is that the struct already takes too much parameters. I think we should come up with a better solution, instead of adding more and more parameters.

sorpaas:
@bkchr Any suggestions?

My opinion is that because we have this block import chain, where we must have parameters passed down from the top-level verifier to bottom-level importer, this is unavoidable. Even if we split it down to multiple structs, we'd still in the end have to combine those small structs together back to a big one, because it has to be passed down.

bkchr:
Yeah, I'm not sure. Maybe the best to do this now with non exhaustive in regards of releasing Substrate. However, I still don't like it.

sorpaas:
Yeah maybe let's use non-exhaustive first and think about better ways in the mean time!

PR #4938, opened 2020-02-14
Allow usage of different proof formats

cheme:
This PR is integrating https://github.com/paritytech/trie/pull/44 to substrate.

This PR defines some Proof trait and allows choosing them in state machine.

The choice of proof and by extension state-machine backend, would move to client.
For this PR, I do restrict the scope to 'primitives' code.
Complete move to client can be found at https://github.com/cheme/substrate/tree/split_child_proof_master_state_trait .
Cht and change trie proof are also not covered in any branch.

New proof supported in the pr:

- Compact proof: we do not store the hashes that can be calculated from other encoded node in the proof (structure of proof is hold by ordering of nodes). 
In https://github.com/paritytech/trie/pull/81 one can run https://github.com/cheme/trie/blob/dedc9284a79d2158b9e06c4e4c837223e4a5b478/test-support/reference-trie/src/compare_proof_size.rs#L40 with different parameter to get an idea of the gain (lines containing 'standard', not 'hybrid').

- Query plan proof: a compact proof where we omit the values. Actually this proof cannot be use in state machine since it's input cannot be passed around and the proof cannot act as a backend. But we can still extract and check it as in test : https://github.com/cheme/substrate/blob/9c3c299a446a1dd19cf7f54a538906ae23369468/primitives/state-machine/src/lib.rs#L1306 . This proof is useful to check if a change occurs between two different state (IIUC was written for xcmp at the time).

The new proofs cannot be merged so new mechanisms for merging proof were added:
- proof now use an intermediate representation `ProofRaw` that can get merged and converted later to the final proof target.
- proof recorder backend needs to keep trace of roots queried, so when it is used multiple tiem as in the call executor we need to inject previous roots ('previous_input' in function 'from_previous_rec_state').

Noticable points:
- associated statemachine backend for recording and runing proof are not strictly needed as I got a single implementation here.
- Proofbackend is similar to record backend, but it is not bringing much complexity in the trait definition.
- Multiple proof implementation is leftover code from the previous version of this pr (single enum for all proofs), this is not really needed at this point and could be removed. I expect though, that we may see some use case for it, especially when two substrate chain want to check each other proofs: usual way is to use a wasm proof checker, but it could be nice to allow native verification over a subset of supported proof through this enum.

- Full variants for simple and compact proof are not strictly needed.

- The proof full recorder (see `into_partial_full_db`) is not strictly needed either.

- ChildrenProofMap and ChildInfoProof are not strictly needed with current supported child trie and can be replaced with ChildrenMap and ChildInfo.

- ProofInput is not using a trait and but is kept as an enum.
	- It can be simplified to roots only (the query plan test case use a different proof to register).
	- We could also choose to always register roots when using state machine and remove this input parameter, simplifying things a bit.

- Proposal from consensus common is storing encoded proof instead of non encoded proof and later encoding.

- Input for proof is an enum, to avoid complexifying to much the type definitions, but it could be an associated type.

- `FullForMerge` is storing by encoded hash, it avoids passing Hasher at some place, but could also use H as parameter (hasher ends up being necessary for compacting proof).

Previously this PR was using an enum for all supported proof type (https://github.com/paritytech/substrate/pull/4938/commits/d659b2c63cb183fbcc51dde06214b60720b9e6bc), and additional parameter was added to all api to chose proof kind.
This new version of the pr just define some proof trait and allow associating them to the state-machine backend in a safer way.


primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
+	pub fn can_use_as_partial_db(&self) -> bool {
+		match self {
+			StorageProofKind::KnownQueryPlanAndValues => false,
+			_ => true,
+		}
+	}
+
+	/// Return the best kind to use for merging later,
+	/// a boolean indicationg if merge should produce full proof.
+	pub fn mergeable_kind(&self) -> (Self, bool) {
```
tomusdrw, line 315: I feel like all methods in this type strongly suggest that the proof kinds should rather be separate types where we can rely on the typesystem to make sure they are used in correct way instead of doing a bunch of runtime checks for `input_kind` or mergeability etc.

primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
+
+	/// Contains trie roots used during proof processing.
+	/// Contains keys queried during the proof processing.
+	QueryPlan(ChildrenProofMap<(Vec<u8>, Vec<Vec<u8>>)>),
+}
+
+impl Input {
+	/// Update input with new content.
+	/// Return false on failure.
+	/// Fails when the input type differs, except for `None` input
```
tomusdrw, line 198: Please keep the form of the comment consistent either 3rd person or 1st person.

primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
+	/// that is always reassignable.
+	///
+	/// Not that currently query plan inputs are not mergeable
+	/// even if doable (just unimplemented).
```
tomusdrw, line 202: Please rephrase.

primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
+
+	/// Kind for `StorageProof::TrieSkipHashesFull`.
+	TrieSkipHashesFull = 127,
+}
+
+impl StorageProofKind {
+	/// Decode a byte value representing the storage kind.
+	/// Return `None` if the kind does not exists or is not allowed.
+	#[cfg(test)]
+	pub fn read_from_byte(encoded: u8) -> Option<Self> {
```
tomusdrw, line 145:
```suggestion
	pub fn from_byte(encoded: u8) -> Option<Self> {
```
We don't really read anything here.

primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
+				=> StorageProofKind::TrieSkipHashesForMerge,
+			x if x == StorageProofKind::TrieSkipHashesFull as u8
+				=> StorageProofKind::TrieSkipHashesFull,
+			_ => return None,
+		})
+	}
+
+	/// Decode a byte value representing the storage kind.
+	/// Return `None` if the kind does not exists or is not allowed.
+	#[cfg(not(test))]
```
tomusdrw, line 163: That's really confusing, why do the variatns exist in the first place if they can't be decoded?

primitives/storage/src/lib.rs
```diff
@@ -334,6 +388,11 @@ impl ChildTrieParentKeyId {
 	}
 }
 
+/// Map of child trie information stored by `ChildInfo`.
+pub type ChildrenMap<T> = BTreeMap<ChildInfo, T>;
+
+const DEFAULT_CHILD_TYPE_PARENT_PREFIX: &'static [u8] = b":child_storage:default:";
```
tomusdrw, line 394: Please move to `well_known_keys`.

primitives/state-machine/src/trie_backend_essence.rs
```diff
@@ -353,34 +430,57 @@ impl<H: Hasher> TrieBackendStorage<H> for Arc<dyn Storage<H>> {
...
+		if *key == self.essence.empty {
 			return Some([0u8].to_vec())
 		}
-		match self.storage.get(&key, prefix) {
+		let top;
+		let child_info = if let Some(child_info) = self.child_info {
+			child_info
+		} else {
+			top = ChildInfo::top_trie();
+			&top
```
tomusdrw, line 481: Nice trick for the lifetimes ;)

primitives/state-machine/src/proving_backend.rs
```diff
@@ -112,15 +115,42 @@ impl<'a, S, H> ProvingBackendRecorder<'a, S, H>
...
-pub type ProofRecorder<H> = Arc<RwLock<HashMap<<H as Hasher>::Out, Option<DBValue>>>>;
+/// A type that records all accessed trie nodes.
+pub enum ProofRecorder<H: Hasher> {
+	/// Records are separated by child trie, this is needed for
+	/// proof compaction.
+	Full(Arc<RwLock<ChildrenMap<RecordMapTrieNodes<H>>>>),
+	/// Single storage for all recoded nodes (as in
+	/// state db column).
+	/// That this variant exists only for performance
+	/// (on less map access than in `Full`), but is not strictly
```
tomusdrw, line 126: Why not store the `ChildInfo` alongside the values there? You can have an `Arc` or `Rc` there and it would still require only one lookup and wouldn't really consume more memory.

primitives/state-machine/src/proving_backend.rs
```diff
@@ -167,13 +258,27 @@ impl<'a, S: 'a + TrieBackendStorage<H>, H: 'a + Hasher> TrieBackendStorage<H>
 {
 	type Overlay = S::Overlay;
 
-	fn get(&self, key: &H::Out, prefix: Prefix) -> Result<Option<DBValue>, String> {
-		if let Some(v) = self.proof_recorder.read().get(key) {
-			return Ok(v.clone());
+	fn get(&self, child_info: &ChildInfo, key: &H::Out, prefix: Prefix) -> Result<Option<DBValue>, String> {
+		match &self.proof_recorder {
+			ProofRecorder::Flat(rec) => {
+				if let Some(v) = rec.read().get(key) {
```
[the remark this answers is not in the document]
tomusdrw: NVM, afaiu the `key` is prefixed anyway, so we only store extra `ChildInfo` there.

primitives/state-machine/src/overlayed_changes.rs
```diff
@@ -685,7 +691,7 @@ impl OverlayedChanges {
 		parent_hash: H::Out,
 		panic_on_storage_error: bool,
 		cache: &mut StorageTransactionCache<B::Transaction, H, N>,
-	) -> Result<Option<H::Out>, ()> where H::Out: Ord + Encode + 'static {
+	) -> Result<Option<H::Out>, ()> where H::Out: Ord + Decode + Encode + 'static {
```
tomusdrw, line 694:
```suggestion
	) -> Result<Option<H::Out>, ()> where H::Out: Ord + Codec + 'static {
```

primitives/state-machine/src/overlayed_changes.rs
```diff
@@ -569,7 +572,10 @@ impl OverlayedChanges {
 		changes_trie_state: Option<&ChangesTrieState<H, N>>,
 		parent_hash: H::Out,
 		mut cache: &mut StorageTransactionCache<B::Transaction, H, N>,
-	) -> Result<StorageChanges<B::Transaction, H, N>, String> where H::Out: Ord + Encode + 'static {
+	) -> Result<StorageChanges<B::Transaction, H, N>, String>
+		where
+			H::Out: Ord + Decode + Encode + 'static
```
tomusdrw, line 577:
```suggestion
			H::Out: Ord + Codec + 'static
```

primitives/state-machine/src/overlayed_changes.rs
```diff
@@ -558,7 +558,10 @@ impl OverlayedChanges {
 		changes_trie_state: Option<&ChangesTrieState<H, N>>,
 		parent_hash: H::Out,
 		mut cache: StorageTransactionCache<B::Transaction, H, N>,
-	) -> Result<StorageChanges<B::Transaction, H, N>, String> where H::Out: Ord + Encode + 'static {
+	) -> Result<StorageChanges<B::Transaction, H, N>, String>
+		where
+			H::Out: Ord + Decode + Encode + 'static
```
tomusdrw, line 563:
```suggestion
			H::Out: Ord + Codec + 'static
```

primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
+			StorageProofKind::TrieSkipHashes
+				| StorageProofKind::TrieSkipHashesFull
+				| StorageProofKind::TrieSkipHashesForMerge
+				| StorageProofKind::Full
+				| StorageProofKind::Flatten => InputKind::None,
+		}
+	}
+
+	/// Indicates what variant of proof recorder should be use.
+	pub fn need_register_full(&self) -> bool {
```
tomusdrw, line 276: This method name sounds like it's 3 random words, I think with a bit of care it can be rephrased to something a bit more self-explanatory, like;
`is_full_proof_recorder_needed()`

primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
+		}
+		true
+	}
+}
+
+/// Kind for a `Input` variant.
+pub enum InputKind {
+	/// `Input::None` kind.
+	None,
+
```
tomusdrw, line 239:
```suggestion
```
I don't feel like these extra lines add any value.

primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
+	QueryPlanWithValues(ChildrenProofMap<(Vec<u8>, Vec<(Vec<u8>, Option<Vec<u8>>)>)>),
+
+	/// Contains trie roots used during proof processing.
+	/// Contains keys queried during the proof processing.
+	QueryPlan(ChildrenProofMap<(Vec<u8>, Vec<Vec<u8>>)>),
+}
+
+impl Input {
+	/// Update input with new content.
+	/// Return false on failure.
```
tomusdrw, line 197: Why not return a `Result` instead? That's a clear pattern that has multiple syntactic benefits.

primitives/consensus/common/src/lib.rs
```diff
@@ -116,16 +117,40 @@ impl RecordProof {
 	/// Returns if `Self` == `Yes`.
 	pub fn yes(&self) -> bool {
```
tomusdrw, line 118: This method is probably not used anywhere now. I guess every time it's `yes()` you need the `kind()` anyway, so best to match on the latter.

cheme:
@tomusdrw , thanks for looking at it, I will try to clean that up tomorrow (before starting a refactoring of the PR).

> I feel however that we could gain way more clarity if a more type-safe approach was used for different storage proof kinds

I reach a similar conclusion a while back and put the pr back to `in progress` status, I am willing to do something that allows choosing the proof kind from client. (in this state the pr is ok for experimenting but it is a bit awkward).
I did work on this branch last week https://github.com/paritytech/substrate/compare/master...cheme:master-state-trait and will try to merge it in this pr after looking at your comments (the branch adds proof traits to state machine backend and allows defining state machine backend from bin (the branch does not change external api but move traits around a bit and may become a PR on its own but I want to try using it with this pr and test with a different trie implementation first)).
That is indeed probably more compatible to substrate as a framework (the enum storageproof may still be interesting though but using a trait seems better (but that is more change)).

tomusdrw:
> the branch adds proof traits to state machine backend and allows defining state machine backend from bin

I was thinking more about a half-way solution where we still have an enum, but the logic specific to different variants is split into structs and it's own modules. That way you don't necessarily need to change the high-level API, while lower level the type safety is enforced - i.e. after you match on a `kind` you start using type type-safe API.

primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
 	}
+}
+
+/// This encodes the full proof capabillity under
+/// legacy proof format.
+pub struct LegacyEncodeAdapter<'a>(pub &'a StorageProof);
 
+impl<'a> Encode for LegacyEncodeAdapter<'a> {
+	fn encode_to<T: CodecOutput>(&self, dest: &mut T) {
+		0u8.encode_to(dest);
```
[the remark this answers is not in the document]
cheme: That is a very questionable implementation that allows encoding any proof format under the legacy flat proof format, I will probably remove that from refactoring.

primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
+		} else {
+			self.1.read_byte()
+		}
+	}
+}
+
+impl Decode for LegacyDecodeAdapter {
+	fn decode<I: CodecInput>(value: &mut I) -> CodecResult<Self> {
+		let legacy = value.read_byte()?;
+		Ok(if legacy == 0 {
```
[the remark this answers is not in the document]
cheme: proof starting with 0 is the empty flatten proof if there is only one byte of content or the proof encoded afterward if there is more content, but will probably remove this encoding as it was here to facilitate transition, but even in the description of the PR I really started to question if it could be of any use.

primitives/trie/src/storage_proof.rs
```diff
@@ -98,12 +1028,175 @@ impl Iterator for StorageProofNodeIterator {
...
+}
+
+
+/// Container recording trie nodes.
+#[derive(Clone)]
+pub struct RecordMapTrieNodes<H: Hasher>(HashMap<H::Out, Option<DBValue>>);
+
+/// Container recording trie nodes and their encoded hash.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ProofMapTrieNodes(pub HashMap<Vec<u8>, DBValue>);
```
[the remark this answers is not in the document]
cheme: Actually I realize, this is a bit awkward, I am using Vec<u8> as encoded hash of encoded trie node (H::Out), probably did so to keep it compatible with different child trie hash format, but at the time we only got one that is using H::Out, so probably not worth encoding for that.

primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
+
+	// Technical variants
+
+	/// This is an intermediate representation that keep trace of
+	/// input and is therefore mergeable into compact representation.
+	/// Compatible with `TrieSkipHashes` and `TrieSkipHashesFull` proofs.
+	TrieSkipHashesForMerge(ChildrenProofMap<(ProofMapTrieNodes, Vec<u8>)>),
+
+	// Following variants are only for testing, they still can be use but
+	// decoding is not implemented.
```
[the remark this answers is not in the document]
cheme: In fact I was trying to say that these should not be use because the over variants are strictly better.

primitives/trie/src/storage_proof.rs
```diff
@@ -14,65 +14,990 @@
...
+			},
+			s => Ok(s),
+		}
 	}
 
-	/// Merges multiple storage proofs covering potentially different sets of keys into one proof
-	/// covering all keys. The merged proof output may be smaller than the aggregate size of the input
+	/// Run proof validation when the proof allows immediate
+	/// verification (`StorageProofKind::can_use_verify`).
+	pub fn verify<H: Hasher>(
```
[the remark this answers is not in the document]
cheme: This should become a different trait.

primitives/state-machine/src/trie_backend_essence.rs
```diff
@@ -85,8 +117,41 @@ impl<S: TrieBackendStorage<H>, H: Hasher> TrieBackendEssence<S, H> where H::Out:
...
+
+		let encoded_root = self.storage(child_info.prefixed_storage_key().as_slice())?;
+		if let Some(cache) = self.register_roots.as_ref() {
+			cache.write().insert(child_info.clone(), encoded_root.clone());
+		}
+
+		let root: Option<H::Out> = encoded_root
+			.and_then(|encoded_root| Decode::decode(&mut &encoded_root[..]).ok());
+
+		Ok(root)
```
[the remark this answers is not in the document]
cheme: I am not sure I understand this comment, `Ok(None)` is pretty usefull, it means there is no such child trie. Caching it avoid the storage query the next time we want to use such child trie.

primitives/state-machine/src/trie_backend_essence.rs
```diff
@@ -41,15 +43,45 @@ pub struct TrieBackendEssence<S: TrieBackendStorage<H>, H: Hasher> {
 	storage: S,
 	root: H::Out,
 	empty: H::Out,
+	/// If defined, we store encoded visited roots for top_trie and child trie in this
+	/// map. It also act as a cache.
+	pub register_roots: Option<RwLock<ChildrenMap<Option<StorageValue>>>>,
```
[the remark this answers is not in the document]
cheme: You think we should have dedicated function for acessing register_roots (operation is set and take basically)?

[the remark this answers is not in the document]
tomusdrw: Other fields are public, so yeah I think we should rather keep the struct consistent.

primitives/storage/src/lib.rs
cheme: It was my intention to not exposte it in `well_known_keys`, but you are not the first one that want it back (actually Basti restores it and I put it away again here). My rational is that we should check against the child info api (I did the change at some other place), having some magic constant is not what I like the best, but let's move it back (I am keeping the changed check though).

primitives/state-machine/src/proving_backend.rs
cheme: For the case where we use the Full variant we would need to split the content by child info afterward, I have no idea if it will be faster or not. Generally (as metioned in the PR description) I think this enum is not really needed, will see during refactoring if it can be dropped.

cheme:
The refactoring seems finish for this, PR, after fixing the oversized line breaks and going through the documentation, I will push the PR back to reviewable.
I already update the description and first feedbacks are welcome (code should not change much).

cheme:
I am pushing this PR back in reviewable state.
The refactoring changes the way proof are handled.
I am not entirely happy with followup code in https://github.com/cheme/substrate/tree/split_child_proof_master_state_trait that allows specifying proof type at client level, due to the number of change and general trait complexity.
But on the other hand it would allow a parachain to run on its own state backend (eg with a different trie or state).
Going back to a pr using globally `Multiple` (enum) proof in state could be a more straightforward way to allow using compact proof (would need to add a parameter to as_proof_backend), but it still feels like going in a wrong direction.

cheme:
Pr got updated with master as https://github.com/cheme/substrate/tree/split_child_proof_master_state_trait (generally this branch is rather interesting for its client and service related changes).

gnunicorn:

cheme:
Well I understand the reason to close pr for inactivity.
Just wanted to mention that I can open this again to anyone interested.
The reduction in size of proof as mentioned in the PR description being between :
trie size | nb value in proof | size compressed / size uncompressed 
100       |  1                         | 0.83
100       |  10                       | 0.23
100       |  100                     | 0.05
1000     |  1                         | 0.86
1000     |  10                       | 0.44
1000     |  100                     | 0.20
1000     |  1000                   | 0.03
10k       |  1                         | 0.84
10k       |  10                       | 0.56
10k       |  100                     | 0.32
10k       |  1000                   | 0.12
10k       |  10k                     | 0.02
100k     |  1                         | 0.87
100k     |  10                       | 0.63
100k     |  100                     | 0.45
100k     |  1000                   | 0.26
100k     |  10k                     | 0.09

So it all relates to the tree coverage for the proof.

Another closing note if someone want to use those proofs differently:
this PR did aim at bringing the proof choice and compression for a state at the client level and is only relevant when looking at the follow up branch: https://github.com/cheme/substrate/tree/split_child_proof_master_state_trait .

PR #4945, opened 2020-02-16
Database & keyring generation for benchmarks optimised

NikVolf:
- speeds up benchmarking dramatically (since keyring generation no longer done on every iteration)
- allows to reuse seed databases (just `clone`, and you get a copy with brand new directory)

### DIFF
--- a/bin/node/testing/benches/import.rs
+++ b/bin/node/testing/benches/import.rs
@@ -100,14 +100,12 @@ struct BenchKeyring {
 // This is prepared database with genesis and keyring
 // that can be cloned and then used for any benchmarking.
 struct BenchDb {
-	random_space: usize,
 	keyring: BenchKeyring,
 	directory_guard: Guard,
 }
 
 impl Clone for BenchDb {
 	fn clone(&self) -> Self {
-		let random_space = self.random_space;
 		let keyring = self.keyring.clone();
 		let dir = tempdir::TempDir::new("sub-bench").expect("temp dir creation failed");
 
@@ -132,14 +130,14 @@ impl Clone for BenchDb {
 			&fs_extra::dir::CopyOptions::new(),
 		).expect("Copy of seed database is ok");
 
-		BenchDb { keyring, directory_guard: Guard(dir), random_space }
+		BenchDb { keyring, directory_guard: Guard(dir) }
 	}
 }
 
 impl BenchDb {
 
-	fn new(random_space: usize) -> Self {
-		let keyring = BenchKeyring::new(random_space);
+	fn new(keyring_length: usize) -> Self {
+		let keyring = BenchKeyring::new(keyring_length);
 
 		let dir = tempdir::TempDir::new("sub-bench").expect("temp dir creation failed");
 		log::trace!(
@@ -150,7 +148,7 @@ impl BenchDb {
 		let (_client, _backend) = bench_client(dir.path(), Profile::Native, &keyring);
 		let directory_guard = Guard(dir);
 
-		BenchDb { keyring, random_space, directory_guard }
+		BenchDb { keyring, directory_guard }
 	}
 
 	fn generate_block(&mut self) -> Block {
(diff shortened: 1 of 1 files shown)


bin/node/testing/benches/import.rs
```diff
@@ -92,10 +92,169 @@ fn genesis(keyring: &BenchKeyring) -> node_runtime::GenesisConfig {
...
+			dir.path(),
+			&fs_extra::dir::CopyOptions::new(),
+		).expect("Copy of seed database is ok");
+
+		BenchDb { keyring, directory_guard: Guard(dir), random_space }
+	}
+}
+
+impl BenchDb {
+
```
bkchr:
```suggestion
```

bin/node/testing/benches/import.rs
```diff
@@ -92,10 +92,169 @@ fn genesis(keyring: &BenchKeyring) -> node_runtime::GenesisConfig {
...
+	fn new(random_space: usize) -> Self {
```
bkchr:
```suggestion
	fn new(account_num: usize) -> Self {
```

Or something similar, but `random_space` is such a random name and doesn't tell we what it does :D

bin/node/testing/benches/import.rs
```diff
@@ -92,10 +92,169 @@ fn genesis(keyring: &BenchKeyring) -> node_runtime::GenesisConfig {
...
 //     //endowed-user//N
+#[derive(Clone)]
 struct BenchKeyring {
 	accounts: BTreeMap<AccountId, sr25519::Pair>,
 }
 
+// This is prepared database with genesis and keyring
+// that can be cloned and then used for any benchmarking.
+struct BenchDb {
+	random_space: usize,
```
bkchr: Isn't that just `self.keyring.accounts.len()`?

bin/node/testing/benches/import.rs
```diff
@@ -92,10 +92,169 @@ fn genesis(keyring: &BenchKeyring) -> node_runtime::GenesisConfig {
...
+		let directory_guard = Guard(dir);
+
+		BenchDb { keyring, random_space, directory_guard }
+	}
+
+	fn generate_block(&mut self) -> Block {
+		let (client, _backend) = bench_client(
+			self.directory_guard.path(),
+			Profile::Wasm,
+			&self.keyring
```
bkchr:
```suggestion
			&self.keyring,
```

bin/node/testing/benches/import.rs
```diff
@@ -398,26 +428,20 @@ fn bench_block_import(c: &mut Criterion) {
...
-		(block, context.keep_db())
-	};
+	let mut bench_db = BenchDb::new(128);
+	let block = bench_db.generate_block();
 
 	log::trace!(
 		target: "bench-logistics",
 		"Seed database directory: {}",
-		guard.0.path().to_string_lossy(),
+		bench_db.path().to_string_lossy(),
```
bkchr:
```suggestion
		bench_db.path().display(),
```

bin/node/testing/benches/import.rs
```diff
@@ -92,10 +92,169 @@ fn genesis(keyring: &BenchKeyring) -> node_runtime::GenesisConfig {
...
+
+		let seed_dir = self.directory_guard.0.path();
+
+		log::trace!(
+			target: "bench-logistics",
+			"Copying seed db from {} to {}",
+			seed_dir.to_string_lossy(),
+			dir.path().to_string_lossy(),
+		);
+		let seed_db_files = std::fs::read_dir(seed_dir)
```
bkchr: So copying files is faster than recreating them? How do you make sure that rocksdb has flushed all data to disc at this point?

NikVolf: RocksDB is already dropped at this point, so should be flushed.

Copying seems faster than opening, letting genesis to instantiate everything and closing

bkchr: If RocksDB is dropped, it is good!

Issue #4915, opened 2020-02-13
How to prevent a Call to be dispatched indirectly


xlc:
Calls can be dispatch indirectly by for example `utility.as_sub`. But for indirect dispatched calls, our `SignedExtension` will not be able to detect those calls and perform additional validations before there enters tx pools.

For example, in our oracle module, only whitelisted users are allowed to submit `oracle.feed_value` and the tx fee is configured to be zero. We also want to ensure each whitelisted user can only submit up to one `oracle.feed_value` per block (so they cannot DOS the network permanently).

Because those tx are free, we need to ensure invalid txs are rejected which can be enforced by our `CheckOperator` signed extension.

But in our implementation, we are only able to handle calls that made to `oracle` module and validate them. If attacker use `utility.as_sub` to dispatch the call indirectly, the validate logic will be bypassed completely.

It is possible to modify our implementation to look for Calls from `utility` pallet, but it just does not scales and not a good solution.

The quick fix to prevent this issue is to be able to detect if a dispatchable is dispatched indirectly, and panic if it is, which will prevent the trouble tx to be added to tx pool.

On the other hand, many dispatchables are designed to be dispatched indirectly only (e.g. `system.set_code`). So we may want a better design to handle those cases:

- Calls that can be dispatched directly or indirectly (i.e. the current behavior)
- Calls that can only be dispatched indirectly  (e.g. `system.set_code`)
- Calls that can only be dispatched directly to avoid bypass additional validation logic.

The work on our side: https://github.com/laminar-protocol/open-runtime-module-library/pull/86

bkchr:
While you can use a `SignedExtension` for early rejection of invalid transactions. You should not use a `SignedExtension` to check the origin of a transaction. Not sure why you not check the `origin` inside of the function? What is the problem?

xlc:
Please let me clarify this.

I am presenting one issue:

It is possible to bypass validation logic in `validate` by dispatch the call indirectly (e.g. `utility.as_sub`).

For example, contracts pallet checks the transaction cannot exceed block gas limit in validate. The check won't be performed if the call is dispatched by `utility.as_sub`.
https://github.com/paritytech/substrate/blob/7d544ef396130426be1799729fbc9b9b22488c8e/frame/contracts/src/lib.rs#L1125

This basically prevents any to perform any security related logics in validate, which will be required if I want to reject certain tx from including into block / inserted into tx pool.

[...]

PR #4946, opened 2020-02-16
Add passthrough weight to Sudo

shawntabrizi:
This PR updates the weight function for `sudo` and `sudo_as` to use the weight of the call being made + 10,000 for the minimal additional computation it does.

Weight being correctly passed-through is important for block fullness calculation.

Passthrough logic is handled with `FunctionOf` which has been updated to support either raw values or closures for all of `Weight`, `DispatchClass`, and `PaysFee`.


xlc:
`pays_fee` cannot be false without additional checks. Because it means anyone can dispatch without fee, not just sudo key holder.
What you will want is only sudo key holder can dispatch this without paying fee, which unfortunately is harder than expected to implement correctly.

Related: #4915

shawntabrizi:
Ahh that makes sense. Thanks xlc :)

frame/sudo/src/lib.rs
```diff
@@ -214,3 +216,47 @@ decl_error! {
...
+	}
+}
+impl<Call: GetDispatchInfo> ClassifyDispatch<(&Box<Call>,)> for SudoPassthrough<Call> {
+	fn classify_dispatch(&self, (call,): (&Box<Call>,)) -> DispatchClass {
+		call.get_dispatch_info().class
+	}
+}
+impl<Call: GetDispatchInfo> PaysFee<(&Box<Call>,)> for SudoPassthrough<Call> {
+	fn pays_fee(&self, (call,): (&Box<Call>,)) -> bool {
+		call.get_dispatch_info().pays_fee
```
bkchr: Shouldn't this be true?

frame/sudo/src/lib.rs
```diff
@@ -214,3 +216,47 @@ decl_error! {
...
+	}
+}
+impl<Lookup, Call: GetDispatchInfo> ClassifyDispatch<(&Lookup, &Box<Call>)> for SudoAsPassthrough<Lookup, Call> {
+	fn classify_dispatch(&self, (_who, call): (&Lookup, &Box<Call>)) -> DispatchClass {
+		call.get_dispatch_info().class
+	}
+}
+impl<Lookup, Call: GetDispatchInfo> PaysFee<(&Lookup, &Box<Call>)> for SudoAsPassthrough<Lookup, Call> {
+	fn pays_fee(&self, (_who, call): (&Lookup, &Box<Call>)) -> bool {
+		call.get_dispatch_info().pays_fee
```
bkchr: Same?

frame/sudo/src/lib.rs
shawntabrizi: The way it is written, user only pays a fee if the underlying call would pay a fee. If someone called a "free" transaction, it would also be free.

Sounds like the right behavior to me?

bkchr: We just had the "problem" that we can not have calls that don't need to pay any fees. If we have such calls, it opens a ddos door. You could spam the system from accounts without any funds. The calls would be valid, but would fail at applying because the users aren't root. As @xlc said, this is a more complex topic and would require a `SignedExtension` to pay the fee back, if the origin is the sudo account.

shawntabrizi: @bkchr see the implementation in Util https://github.com/paritytech/substrate/blob/master/frame/utility/src/lib.rs#L192

Should that also then be updated? It seems the issue here is not with this part of the logic, but including extrinsics which would pay a fee.

bkchr: So, the system is currently not perfect. However, you could have the following:

- You have transaction `A` that doesn't pay any fees, but you protect this with a custom `SignedExtension`. This `SignedExtension` makes sure that `validate_transaction` fails when the origin for example is not correct.
- With the current propagation of `pays_fee`, a malicious user could create a `Sudo` transaction that wraps the `A` transaction and it would be free. The malicious user could start to spam the network with this `Sudo(A)` transaction.
- `SignedExtension` currently don't propagate the checks if your call wraps another call, which could you bring into this situation. So, the best for the moment is just to always pay fees.

kianenigma:
👎 

IMO if anything we should __deprecate__ all of these `PassThrough` and and use `FunctionOf`. I wanted to make an issue at some point for this. The only pity is that that `FunctionOf` needs to to define argument types. Which is still substantially less than all of this boilerplate. 

At leas these simple logics like 
- pay fee: true
- weight: param + 1000

are exactly why `FunctionOf` was created for.

kianenigma:
cc @pepyakin  ^^

shawntabrizi:
@kianenigma where can I read on the latest of FunctionOf?

kianenigma:
https://github.com/paritytech/substrate/blob/master/frame/support/src/weights.rs#L240

shawntabrizi:
@kianenigma I have updated this PR to use `FunctionOf`, which is certainly more ergonomic. However, it does not seem we can "passthrough" the dispatch classification.

Is this intentionally not allowed, or just a feature request?

shawntabrizi:
Updated code and docs

PR #4947, opened 2020-02-17
Benchmarking for Offchain Phragmén. 

kianenigma:
Adds some benchmarking and builds on top of https://github.com/paritytech/substrate/pull/4517.


frame/staking/Cargo.toml
```diff
@@ -20,6 +20,11 @@ pallet-session = { version = "2.0.0", features = ["historical"], path = "../sess
 pallet-authorship = { version = "2.0.0", default-features = false, path = "../authorship" }
 sp-application-crypto = { version = "2.0.0", default-features = false, path = "../../primitives/application-crypto" }
 
+# Needed for benchmarking stuff. TODO: this stuff must NOT be always compiled. use a feature.
+pallet-indices = { version = "2.0.0", default-features = false, path = "../indices" }
```
kianenigma: cc @DarkEld3r

bkchr: Could we not put the benchmarking into its own crate?

primitives/phragmen/compact/src/assignment.rs
```diff
@@ -187,6 +187,7 @@ pub(crate) fn assignment(
 
 				for _phragmen::Assignment { who, distribution } in assignments {
 					match distribution.len() {
+						0 => continue /* ignore and move on */,
```
marcio-diaz: Remove comment or explain better :P

frame/staking/src/offchain_election.rs
```diff
@@ -195,3 +158,91 @@ pub(crate) fn compute_offchain_election<T: Trait>() -> Result<(), OffchainElecti
...
+			let stake = stake_of(&a.who);
+			a.into_staked(stake, true)
+		})
+		.collect();
+
+	if do_reduce {
+		reduce(&mut staked);
+	}
+
+	// convert back to ratio assignment. This takes less space.
```
marcio-diaz:
```suggestion
	// Convert back to ratio assignment. This takes less space.
```

frame/staking/src/offchain_election.rs
```diff
@@ -195,3 +158,91 @@ pub(crate) fn compute_offchain_election<T: Trait>() -> Result<(), OffchainElecti
...
+			<usize as TryInto<ValidatorIndex>>::try_into(i).ok()
+		)
+	};
+	let stake_of = |who: &T::AccountId| -> ExtendedBalance {
+		<T::CurrencyToVote as Convert<BalanceOf<T>, u64>>::convert(
+			<Module<T>>::slashable_balance_of(who)
+		) as ExtendedBalance
+	};
+
+	// clean winners
```
marcio-diaz:
```suggestion
	// Clean winners.
```

frame/staking/Cargo.toml
kianenigma: We sure could. Question is, is it needed given 
https://github.com/paritytech/substrate/issues/4873

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,651 @@
...
+	fn instance(&self, components: &[(BenchmarkParameter, u32)])
+		-> Result<(Call<T>, RawOrigin<T::AccountId>), &'static str>
+	{
+		let num_validators = components.iter().find(|&c| c.0 == BenchmarkParameter::V).unwrap().1;
+		let num_nominators = components.iter().find(|&c| c.0 == BenchmarkParameter::N).unwrap().1;
+		let edge_per_voter = components.iter().find(|&c| c.0 == BenchmarkParameter::Z).unwrap().1;
+		let mode: BenchmarkingMode = BenchmarkingMode::WeakerSubmission;
+		let do_reduce: bool = true;
+		let to_elect: u32 = components.iter().find(|&c| c.0 == BenchmarkParameter::E).unwrap().1;
+
```
marcio-diaz: Mmm, I wonder if we should start building on top of the branch of Gav, it seems almost finished.

frame/staking/Cargo.toml
bkchr: As you know, features are broken in Rust/Cargo. I like to prevent their usage as much as possible. But yeah, as long as you don't have diverging code bases based on the feature, "it should be okay"

frame/staking/src/benchmarking.rs
kianenigma: Regardless of the benchmarking details, this also has one or two code updates that are needed in the base branch. So I'd prefer getting this stuff asap into the base.

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,703 @@
...
+static mut SEED: u64 = 999_666;
+static mut MODE: BenchmarkingMode = BenchmarkingMode::StrongerSubmission;
+
+type AddressOf<T> = Address<<T as frame_system::Trait>::AccountId, u32>;
+
+/// _r_andom number in _r_ange.
+fn rr(a: u32, b: u32) -> u32 {
+	// well, what do you expect? This is just benchmark code.
+	unsafe {
```
shawntabrizi: dumb question, what about this is unsafe?

kianenigma: Mutating a global static requires unsafe.

kianenigma:
@shawntabrizi I will try to keep this branch up to date for usability, but I don't see any point in merging it. The dependencies are ugly and the internals need to become compatible with the macro, which I am not sure if it is possible for now. So no need to rush into this.

kianenigma:
This is beyond saving. I will make a new PR soonish.

PR #4950, opened 2020-02-17
Fix description of --no-private-ipv4

### LINKED ISSUE #4919: Flag --no-private-ipv4 contains wrong description
When looking at the documentation produced by `--help`, the flag `--no-private-ipv4` reads

> Allow connecting to private IPv4 addresses (as specified in [RFC1918(https://tools.ietf.org/html/rfc1918)),unless the address was passed with `--reserved-nodes` or `--bootnodes`

This is wrong. As the name of the flag indicates, if set, it is **NOT** allowing to connect to private IPv4 addresses, unless they were passed with `--reserved-nodes` or `--bootnodes`.

That means, it does the exact opposite of what the current description says.

tomaka:
Fix #4919


gnunicorn:
please merge master to fix the failing test.

PR #4953, opened 2020-02-17
Utility/Recovery passthrough always pays a fee.

shawntabrizi:
This PR updates the logic around calculating the weight and fees for passthrough calls in the Utility pallet.

This makes it so that passthrough calls _always_ pay a fee to avoid allowing users to spam the network with potentially free transactions which cost computation.

Additionally we use the `FunctionOf` struct to standardize the way we calculate custom weight information. The calculations themselves (other than `pays_fee = true`) should be unchanged.


shawntabrizi:
@bkchr lets put this on ice, I will try to use @kianenigma new FunctionOf syntax and update all passthrough logic

PR #4956, opened 2020-02-17
Wasm import benchmark per block size 

NikVolf:
This shows non-linearity per block size of wasm block execution for now:

`$ cargo bench -p node-testing wasm_size_import`

Screenshot 2020-02-17 at 20 24 35


NikVolf:
Ops pr should have been updated to master brand before merge, will reopen

PR #4958, opened 2020-02-18
aura: remove unneeded SlotDuration struct and rename digest -> digests

sorpaas:
* `SlotDuration` defined in `aura` is redundant. This PR removes it.
* Renamed mod `digest` to `digests`. This is because we have `sp_consensus_babe::digests` which serves similar purposes, and this keeps it consistent. `digest` mod is not exposed so this won't break API.


client/consensus/aura/src/lib.rs
```diff
@@ -143,7 +121,7 @@ impl SlotCompatible for AuraSlotCompatible {
 
 /// Start the aura worker. The returned future should be run in a futures executor.
 pub fn start_aura<B, C, SC, E, I, P, SO, CAW, Error>(
-	slot_duration: SlotDuration,
+	slot_duration: SlotDuration<u64>,
```
tomusdrw: I guess the idea was that it's being used in the public API and you don't have to import `sc_consensus_slots` additionally. Maybe we should re-export it?

bkchr:
Still fails. Maybe just run everything local with `cargo test --all`

sorpaas:
Fixed!

PR #4962, opened 2020-02-18
Benchmark macro

### LINKED ISSUE #4861: Create macros to simplify benchmarking.

gavofyork:
Fixes #4861


kianenigma:
Will it still be possible, given this, to manually build the setup? There is a situation in https://github.com/paritytech/substrate/pull/4947 that you can see and it cannot be really tracked with this benchmarking setup. One of the parameters is 

```rust
Enum Mode {
   BenchMode1,
   BenchMode2, 
}
```

and stepping over this is not possible.

marcio-diaz:
> Will it still be possible, given this, to manually build the setup? There is a situation in #4947 that you can see and it cannot be really tracked with this benchmarking setup. One of the parameters is
> 
> ```rust
> Enum Mode {
>    BenchMode1,
>    BenchMode2, 
> }
> ```
> 
> and stepping over this is not possible.


> Will it still be possible, given this, to manually build the setup? There is a situation in #4947 that you can see and it cannot be really tracked with this benchmarking setup. One of the parameters is
> 
> ```rust
> Enum Mode {
>    BenchMode1,
>    BenchMode2, 
> }
> ```
> 
> and stepping over this is not possible.

I think you can make two benchmarks out of the enum and refactor the common parts.

frame/benchmarking/src/lib.rs
```diff
@@ -14,79 +14,397 @@
...
+					),*
+				]
+			}
 
-	/// Set up the storage, and prepare a call and caller to test in a single run of the benchmark.
-	fn instance(&self, components: &[(BenchmarkParameter, u32)]) -> Result<(Call, RawOrigin), &'static str>;
+			fn instance(&self, components: &[($crate::BenchmarkParameter, u32)])
+				-> Result<(crate::Call<T>, RawOrigin<T::AccountId>), &'static str>
+			{
+				let _benchmarks_seed = 0;
```
marcio-diaz:
```suggestion
```
`_benchmarks_seed` didn't work for me when trying to use it in `account()`.

gnunicorn:
So, the reports say that it consistently slightly improves on wasm but is sometimes a regression on native ... how do you want to proceed?

marcio-diaz:
> So, the reports say that it consistently slightly improves on wasm but is sometimes a regression on native ... how do you want to proceed?

Wrong PR :)? I'll merge once CI finish, other benchmarks depend on this.

PR #4973, opened 2020-02-18
[In Progress] Remove deprecated api

### LINKED ISSUE #4935: Remove all `deprecated` and `allow(deprecated)` APIs
Before 2.0 release, we should finally cut the cord on any APIs which we have marked deprecated.

DarkEld3r:
Closes #4935.

### DIFF
--- a/utils/wasm-builder-runner/src/lib.rs
+++ b/utils/wasm-builder-runner/src/lib.rs
@@ -315,6 +315,42 @@ impl WasmBuilderSource {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


utils/wasm-builder-runner/src/lib.rs
```diff
@@ -315,42 +315,6 @@ impl WasmBuilderSource {
 	}
 }
 
-/// Build the currently built project as WASM binary and extend `RUSTFLAGS` with the given rustflags.
```
bkchr: Please revert. This is a released crate.

PR #4974, opened 2020-02-18
babe: extract VRF into a separate crate and simplify pre-digest definition

sorpaas:
* Extract VRF into a separate crate, and use wrapper type to implement Encode/Decode.
* Simplify BABE pre-digest definition by removing `RawPreDigest`.


primitives/consensus/babe/src/digests.rs
```diff
@@ -18,178 +18,80 @@
...
+	/// Slot number
+	pub slot_number: SlotNumber,
+}
 
 /// A BABE pre-runtime digest. This contains all data required to validate a
 /// block and for the BABE runtime module. Slots can be assigned to a primary
 /// (VRF based) and to a secondary (slot number based).
-#[cfg(feature = "std")]
-#[derive(Clone, Debug)]
+#[derive(Clone, RuntimeDebug, Encode, Decode)]
```
andresilva: Is the binary encoding still the same compared to `RawPreDigest`? I think it is the same but just want to double check.

primitives/consensus/vrf/src/schnorrkel.rs
```diff
@@ -0,0 +1,208 @@
...
+#[cfg(feature = "std")]
+use std::{ops::{Deref, DerefMut}, convert::TryFrom};
+#[cfg(feature = "std")]
+use codec::EncodeLike;
+#[cfg(feature = "std")]
+use schnorrkel::{SignatureError, errors::MultiSignatureStage};
+
+#[cfg(feature = "std")]
+pub use schnorrkel::vrf::{VRF_PROOF_LENGTH, VRF_OUTPUT_LENGTH};
+
```
andresilva:
```suggestion

 /// The length of the VRF proof. 
```

primitives/consensus/vrf/src/schnorrkel.rs
```diff
@@ -0,0 +1,208 @@
...
+#[cfg(not(feature = "std"))]
+pub const VRF_PROOF_LENGTH: usize = 64;
+
```
andresilva:
```suggestion

 /// The length of the VRF output.
```

primitives/consensus/babe/src/digests.rs
```diff
@@ -18,178 +18,80 @@
...
+	/// Authority index
+	pub authority_index: AuthorityIndex,
+	/// Slot number
+	pub slot_number: SlotNumber,
+}
+
+/// A BABE secondary pre-runtime digest.
+#[derive(Clone, RuntimeDebug, Encode, Decode)]
+pub struct SecondaryPreDigest {
+	/// Authority index
```
andresilva:
```suggestion
	/// Authority index
	///
	/// This is not strictly necessary, since the secondary slots
	/// are assigned based on slot number and epoch randomness. But including
	/// it makes things easier for higher-level users of the chain data to
	/// be aware of the author of a secondary-slot block.
```

primitives/consensus/vrf/src/schnorrkel.rs
```diff
@@ -0,0 +1,208 @@
...
+#[cfg(not(feature = "std"))]
+pub const VRF_OUTPUT_LENGTH: usize = 32;
+
+#[derive(Clone, Eq, PartialEq, RuntimeDebug, Encode, Decode)]
```
andresilva: let's add some docs to the remaining public types.

primitives/consensus/babe/src/digests.rs
sorpaas: Yeah it's the same. The only difference of `RawPreDigest` and `PreDigest` was `VRFProof` and `VRFOutput`. The wrapped `VRFProof` and `VRFOutput` always encode/decode with the same format as `RawVRFProof` and `RawVRFOutput`.

primitives/consensus/vrf/src/schnorrkel.rs
sorpaas: Fixed!

sorpaas:
I realized that `Digest` and `RawDigest` are not exactly of the same functionality, because `Digest` contains additional VRF validations. As a result, the current approach of this PR might not work. I'm going to do a rework and change this to use a type alias + generic approach.

andresilva:
@sorpaas `Digest` requires that we have a valid `VRFProof` and `VRFOutput`, but don't we want to always enforce that? `RawPredigests` should be well-formed and this property should hold for the underlying byte slice.

sorpaas:
@andresilva I think we don't want to enforce that in runtime. Otherwise, both native and wasm will have to pull in some schnorrkel validation code, which will be inefficient. (If schnorrkel is pulled in native, then it has to be pulled in wasm as well because we want to ensure that they always return the same result.)

Let me know if I'm wrong, but my current understanding is that the runtime should treat VRFProof and VRFOutput purely as a black box of 32-byte array, and as a result not doing any additional validation of it.

andresilva:
Didn't think of the implication of using the VRF types in WASM.

PR #4975, opened 2020-02-18
Cargo.toml fixups for the release

gnunicorn:
which is bad and would prevent us from publishing. 

- [x] set missing fields in tomls
- [x] set versions to `-dev`
- [x] make `-test` non-publish
- [x] set licenses

all done thanks to [cargo unleash](https://crates.io/crates/cargo-unleash)


bin/node/cli/Cargo.toml
```diff
@@ -40,84 +42,84 @@ structopt = { version = "0.3.8", optional = true }
...
+sc-cli = { version = "0.8.0-dev", optional = true, path = "../../../client/cli" }
+frame-benchmarking-cli = { version = "2.0.0-dev", optional = true, path = "../../../utils/frame/benchmarking-cli" }
+node-transaction-factory = { version = "0.8.0-dev", optional = true, path = "../transaction-factory" }
+node-inspect = { version = "0.8.0-dev", optional = true, path = "../inspect" }
 
 # WASM-specific dependencies
 wasm-bindgen = { version = "0.2.57", optional = true }
 wasm-bindgen-futures = { version = "0.4.7", optional = true }
-browser-utils = { path = "../../../utils/browser", optional = true }
+browser-utils = { path = "../../../utils/browser", optional = true , version = "0.8.0-dev"}
```
gnunicorn: for cases like this, cargo unleash will impose a CI test in the future so that will never get merged!

PR #4976, opened 2020-02-18
fixes sc_transaction_pool::testing::fork_aware_finalization

seunlanlege:
I discovered the tx `from_dave` sometimes never re-enters the pool, erroring with `PriorityTooLow`. Simply increasing the nonce got rid of the flakiness. :ok:

### DIFF
--- a/client/transaction-pool/src/testing/pool.rs
+++ b/client/transaction-pool/src/testing/pool.rs
@@ -384,7 +384,7 @@ fn fork_aware_finalization() {
 	let mut canon_watchers = vec![];
 
 	let from_alice = uxt(Alice, 1);
-	let from_dave = uxt(Dave, 1);
+	let from_dave = uxt(Dave, 2);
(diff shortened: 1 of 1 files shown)


client/transaction-pool/src/testing/pool.rs
```diff
@@ -488,7 +488,7 @@ fn fork_aware_finalization() {
 
 	// block e1
 	{
-		let header = pool.api.push_block(5, vec![from_dave]);
+		let header = pool.api.push_block(5, vec![from_dave, from_bob]);
```
tomusdrw: `from_bob` should already be in the pool when `D2` is retracted, no? Can you add some assertions about the pool status after each block (i.e. how many transactions we expect to be in the pool).

seunlanlege: Yo @tomusdrw, this should be ready now

PR #4979, opened 2020-02-19
Added subscribe_all_heads RPC function

### LINKED ISSUE #4738: Expose all heads subscription option in subscribe new heads
As mentioned by @tomusdrw in polkadot-js issue  [#1744](https://github.com/polkadot-js/api/issues/1744#issuecomment-578663477)

Expose all new heads in subscribe new heads via parametrization or by creating new RPC call by disabling filter on [https://github.com/paritytech/substrate/blob/2bbdec44cfc2654e59f93a32f91b242bbd538e05/client/rpc/src/chain/mod.rs#L109](https://github.com/paritytech/substrate/blob/2bbdec44cfc2654e59f93a32f91b242bbd538e05/client/rpc/src/chain/mod.rs#L109)

jak-pan:
Added subscribe_all_heads RPC call, there was no way to get all blocks received by the node through RPC. This will help apps that analyze blockchain. I.E. Equivocation detection can now be done by apps connected to RPC. 

Also fixed updated comments and added test.

fixes #4738

### DIFF
--- a/client/rpc-api/src/chain/mod.rs
+++ b/client/rpc-api/src/chain/mod.rs
@@ -54,6 +54,18 @@ pub trait ChainApi<Number, Hash, Header, SignedBlock> {
 	#[rpc(name = "chain_getFinalizedHead", alias("chain_getFinalisedHead"))]
 	fn finalized_head(&self) -> Result<Hash>;
 
+	/// All head subscription
+	#[pubsub(subscription = "chain_allHead", subscribe, name = "chain_subscribeAllHeads")]
+	fn subscribe_all_heads(&self, metadata: Self::Metadata, subscriber: Subscriber<Header>);
+
+	/// Unsubscribe from all head subscription.
+	#[pubsub(subscription = "chain_allHead", unsubscribe, name = "chain_unsubscribeAllHeads")]
+	fn unsubscribe_all_heads(
+		&self,
+		metadata: Option<Self::Metadata>,
+		id: SubscriptionId,
+	) -> RpcResult<bool>;
+
 	/// New head subscription
 	#[pubsub(
 		subscription = "chain_newHead",
@@ -76,7 +88,7 @@ pub trait ChainApi<Number, Hash, Header, SignedBlock> {
 		id: SubscriptionId,
 	) -> RpcResult<bool>;
 
-	/// New head subscription
+	/// Finalized head subscription
 	#[pubsub(
 		subscription = "chain_finalizedHead",
 		subscribe,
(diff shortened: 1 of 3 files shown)


jak-pan:
[clabot:check]

client/rpc/src/chain/tests.rs
```diff
@@ -195,6 +195,34 @@ fn should_notify_about_latest_block() {
...
+
+	// assert initial head sent.
+	let (notification, next) = core.block_on(transport.into_future()).unwrap();
+	assert!(notification.is_some());
+	// assert notification sent to transport
+	let (notification, next) = core.block_on(next.into_future()).unwrap();
+	assert!(notification.is_some());
+	// no more notifications on this channel
+	assert_eq!(core.block_on(next.into_future()).unwrap().0, None);
+}
```
tomusdrw:
```suggestion
}

```

client/rpc-api/src/chain/mod.rs
```diff
@@ -54,6 +54,18 @@ pub trait ChainApi<Number, Hash, Header, SignedBlock> {
 	#[rpc(name = "chain_getFinalizedHead", alias("chain_getFinalisedHead"))]
 	fn finalized_head(&self) -> Result<Hash>;
 
+	/// All head subscription
+	#[pubsub(subscription = "chain_allHead", subscribe, name = "chain_subscribeAllHeads")]
```
tomusdrw: I'm not a big fan of the subscription name (`chain_allHead`) but don't really have any better ideas.

gnunicorn:
@tomusdrw still have to switch your review to approved or github will prevent us from merging.

tomusdrw:
@gnunicorn a tiny style issue & tests are required to be fixed first.

client/rpc-api/src/chain/mod.rs
jak-pan: I'm not a fan too, but I didn't want to break backward compatibility and wanted to keep it consistent.

andresilva: IMO it would be better if this method was `chain_newHead` and then the one below was named `chain_bestHead`, but since this rename would be a bit troublesome I also don't have a good suggestion.

jak-pan:
Are we waiting for anything here? @tomusdrw

tomusdrw:
@jak-pan it's good to go in, but we are stabilising 2.0 currently and any extra functionality is freezed. But I'm sure it's going to be merged as soon as possible.

gnunicorn:
we are still in alpha-phase and that isn't too involved. I'll merge it in.

PR #4983, opened 2020-02-19
Cleanup the exports of sc-network

tomaka:
Before we stabilize stuff, I thought I'd clean up a bit the API of `sc-network`.
This PR doesn't do anything except move code around between modules.

Most notably, everything that is only needed during the initialization process (for example the `TransactionPool` trait) has been moved into the `config` module.

I also moved everything related to `NetworkState` to a separate module.

A couple things have been `#[doc(hidden)]`'ed rather than removed:

- The `ProtocolConfig` struct contains only two fields and is used in `sc-network-test` only.
- The `message` module and its content are supposed to be implementation details and really aren't supposed to be exposed. Right now we create and destroy `ConsensusGossip` structs in `sc-network-gossip`, and we use the `BlockData` network message as the format to store blocks in when exporting/importing blocks in the CLI.

This PR is expected to go in pair with #4665, which will finish the clean-up.

### DIFF
--- a/client/finality-grandpa/src/finality_proof.rs
+++ b/client/finality-grandpa/src/finality_proof.rs
@@ -154,7 +154,7 @@ impl<B, Block: BlockT> FinalityProofProvider<B, Block>
 	}
 }
 
-impl<B, Block> sc_network::FinalityProofProvider<Block> for FinalityProofProvider<B, Block>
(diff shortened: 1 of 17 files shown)


tomaka:
There are a couple of methods that really aren't greatly designed (for example `add_reserved_peer` taking a `String`), so that's a best-effort PR.

gnunicorn:
I dunno why but the confirm squash and merge button doesn't work for me here...

PR #4985, opened 2020-02-19
Re-export PeerId from the peerset API

tomaka:
Nothing important, but makes it so that one doesn't need to import `libp2p` in order to use `sc-peerset`.

### DIFF
--- a/client/peerset/src/lib.rs
+++ b/client/peerset/src/lib.rs
@@ -21,11 +21,12 @@ mod peersstate;
 
 use std::{collections::{HashSet, HashMap}, collections::VecDeque};
 use futures::{prelude::*, channel::mpsc};
-use libp2p::PeerId;
 use log::{debug, error, trace};
 use serde_json::json;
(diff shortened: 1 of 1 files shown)


client/peerset/src/lib.rs
```diff
@@ -21,12 +21,13 @@ mod peersstate;
 
 use std::{collections::{HashSet, HashMap}, collections::VecDeque};
 use futures::{prelude::*, channel::mpsc};
-use libp2p::PeerId;
 use log::{debug, error, trace};
 use serde_json::json;
 use std::{pin::Pin, task::Context, task::Poll};
 use wasm_timer::Instant;
 
+pub use libp2p::PeerId;
```
andresilva:
```suggestion
/// A unique identifier for a peer in the network.
pub use libp2p::PeerId;
```

andresilva:
```suggestion
/// A unique identifier for a peer in the network.
pub use libp2p::PeerId;
```

PR #4988, opened 2020-02-19
Log kademlia errors when get/put record fails.

twittner:
The current approach makes it difficult to figure out what the underlying error was, that made the operation fail.

### DIFF
--- a/client/network/src/discovery.rs
+++ b/client/network/src/discovery.rs
@@ -391,7 +391,14 @@ impl NetworkBehaviour for DiscoveryBehaviour {
 
 								DiscoveryOut::ValueFound(results)
 							}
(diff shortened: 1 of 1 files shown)


client/network/src/discovery.rs
```diff
@@ -391,7 +391,14 @@ impl NetworkBehaviour for DiscoveryBehaviour {
 
 								DiscoveryOut::ValueFound(results)
 							}
+							Err(e@libp2p::kad::GetRecordError::NotFound{..}) => {
```
tomaka:
```suggestion
							Err(e @ libp2p::kad::GetRecordError::NotFound { .. }) => {
```

mxinden:
Thanks @twittner!

PR #4990, opened 2020-02-19
utils/proemtheus/src/lib: Log info that Grafana is removed

mxinden:
Log an info line to the console that the Grafana metric endpoint has been
removed in favor of the Prometheus endpoint.

@gautamdhameja suggested to add this log line on startup for people that either
didn't read the changelog or that are running on newest `master`.

What do people think? Let me know if this is to verbose. Once the Prometheus
endpoint is reasonably well known within the community we can remove the log
line.

### DIFF
--- a/utils/prometheus/src/lib.rs
+++ b/utils/prometheus/src/lib.rs
@@ -120,7 +120,8 @@ mod known_os {
 			.await
 			.map_err(|_| Error::PortInUse(prometheus_addr))?;
 
-		log::info!("Prometheus server started at {}", prometheus_addr);
+		log::info!("Prometheus server started at {}.", prometheus_addr);
+		log::info!("The Grafana endpoint has been removed in favor of the Prometheus endpoint.");
 
 		let service = make_service_fn(move |_| {
 			let registry = registry.clone();


bkchr:
It was never in a stable release. Not sure that we need this.

gautamdhameja:
Some of the Kusama validators have been using the Grafana endpoint, that's why I suggested adding a line to make them aware of the change. We can remove this line at a later stage.

bkchr:
We can just post to the channel and include it in the Polkadot changelog. Also I thought that this Prometheus endpoint can just be used with Grafana?

kianenigma:
Seems more like a release note type of thing to me as well.

gnunicorn:
@s3krit note for the 2.0.0 release notes!?!

bkchr:
It is only useful for Polkadot. Grafana did not exist for 1.0, so there is no reason to include it in any Substrate release notes.

mxinden:
> We can just post to the channel and include it in the Polkadot changelog. Also I thought that this Prometheus endpoint can just be used with Grafana?

@bkchr One can not *just* use it with Grafana. The previous Grafana endpoint was stateful through an in-process database, the new Prometheus endpoint is stateless. One needs to put a Prometheus Server in between Substrate and Grafana to persist the metrics.

---

Closing here in favor of a release note.

bkchr:
Ahh okay, I just always remembered you talking about it being compatible :D

PR #4993, opened 2020-02-19
Remove deprecated API

### LINKED ISSUE #4935: Remove all `deprecated` and `allow(deprecated)` APIs

DarkEld3r:
Remove deprecated api and dead code before 2.0 release.

Closes #4935.

### DIFF
--- a/client/block-builder/src/lib.rs
+++ b/client/block-builder/src/lib.rs
@@ -121,7 +121,7 @@ where
 			backend,
 		})
 	}
-	
+
 	/// Push onto the block's list of extrinsics.
 	///
 	/// This will ensure the extrinsic can be validly executed (by executing it).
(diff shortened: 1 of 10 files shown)


client/block-builder/src/lib.rs
```diff
@@ -141,60 +141,36 @@ where
...
-		if self
+		let use_trusted = skip_signature && self
 			.api
 			.has_api_with::<dyn BlockBuilderApi<Block, Error = ApiErrorFor<A, Block>>, _>(
 				block_id,
-				|version| version < 4,
-			)?
-		{
-			// Run compatibility fallback for v3.
-			self.api.map_api_result(|api| {
```
DarkEld3r: I'm not sure if this should be removed.

DarkEld3r:
`#[allow(dead_code)]` still remains in macros and tests.

frame/babe/src/mock.rs
```diff
@@ -108,5 +106,6 @@ pub fn new_test_ext(authorities: Vec<DummyValidatorId>) -> sp_io::TestExternalit
 	t.into()
 }
 
+#[allow(dead_code)]
 pub type System = frame_system::Module<Test>;
```
bkchr: If this is not used, remove it please.

PR #4995, opened 2020-02-19
Speedup import benchmark

NikVolf:
- remove wasm size benchmark for a while until figured how to skip setup
- reduce iterations

### DIFF
--- a/bin/node/testing/benches/import.rs
+++ b/bin/node/testing/benches/import.rs
@@ -152,6 +152,33 @@ struct Setup {
 	block: Block,
 }
 
+struct SetupIterator {
+	current: usize,
+	finish: usize,
+	multiplier: usize,
+}
+
+impl SetupIterator {
+	fn new(current: usize, finish: usize, multiplier: usize) -> Self {
+		SetupIterator { current, finish, multiplier }
+	}
+}
+
+impl Iterator for SetupIterator {
+	type Item = Setup;
+
+	fn next(&mut self) -> Option<Setup> {
+		if self.current >= self.finish { return None }
+
+		self.current += 1;
+
+		let size = self.current*self.multiplier;
+		let mut db = BenchDb::new(size);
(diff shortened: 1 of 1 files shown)


bin/node/testing/benches/import.rs
```diff
@@ -152,6 +152,33 @@ struct Setup {
...
+
+impl Iterator for SetupIterator {
+	type Item = Setup;
+
+	fn next(&mut self) -> Option<Setup> {
+		if self.current >= self.finish { return None }
+
+		self.current += 1;
+
+		let size = self.current*self.multiplier;
```
kianenigma:
```suggestion
		let size = self.current * self.multiplier;
```

kianenigma:
Can you explain a bit: 

- Creating the db lazily with an iterator looks nicer indeed. But we're going to create all of them in the range of (5, 15) anyhow, unless if there's a chance to exit early? in that case creating them ahead of time and lazily shouldn't make a different. 

- What is the main problem with benchmarks taking too much? we don't run them in the CI, nor should one expect them to be fast.

NikVolf:
Thanks @kianenigma 

> Can you explain a bit:
> 
> * Creating the db lazily with an iterator looks nicer indeed. But we're going to create all of them in the range of (5, 15) anyhow, unless if there's a chance to exit early? in that case creating them ahead of time and lazily shouldn't make a different.

Unfortunately, criterion does collect this iterator anyway, so it's useless now. I will workaround this in following work. Problem is that criterion still runs all those setup functions (which can take quite a while) before figuring out which benchmarks to actually perform (when they are filtered). Later we probably move each benchmark in separate crate because of that.

> * What is the main problem with benchmarks taking too much? we don't run them in the CI, nor should one expect them to be fast.

The problem is that benchmarks must run on one machine exclusively, so each one should be dedicated to one particular point of interest, and take as little time as possible to accomplish the goal with least variance.

PR #4998, opened 2020-02-20
Unsigned transactions should also note weight

kianenigma:
Currently we entirely miss the fact that [these too can be heavy and their weight should be noted](https://github.com/paritytech/substrate/pull/4517/files#diff-32ce5961608780655bf3e00b2cc07d59R1179).

### DIFF
--- a/frame/executive/src/lib.rs
+++ b/frame/executive/src/lib.rs
@@ -541,7 +541,7 @@ mod tests {
 		pallet_balances::GenesisConfig::<Runtime> {
 			balances: vec![(1, 211)],
 		}.assimilate_storage(&mut t).unwrap();
-		let xt = sp_runtime::testing::TestXt::new_signed(sign_extra(1, 0, 0), Call::Balances(BalancesCall::transfer(2, 69)));
+		let xt = TestXt::new_signed(sign_extra(1, 0, 0), Call::Balances(BalancesCall::transfer(2, 69)));
 		let weight = xt.get_dispatch_info().weight as u64;
 		let mut t = sp_io::TestExternalities::new(t);
 		t.execute_with(|| {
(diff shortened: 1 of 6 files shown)


kianenigma:
Also fixes this flaky situation cc @shawntabrizi  
https://github.com/paritytech/substrate/pull/4421#discussion_r359735286

frame/executive/src/lib.rs
```diff
@@ -711,9 +711,26 @@ mod tests {
 		});
 	}
 
+	#[test]
+	fn unsigned_weight_is_noted_when_applied() {
+		let xt = TestXt::new_unsigned(Call::Balances(BalancesCall::set_balance(33, 69, 69)));
+		let len = xt.clone().encode().len() as u32;
+		let mut t = new_test_ext(1);
+		t.execute_with(|| {
+			assert_eq!(<frame_system::Module<Runtime>>::all_extrinsics_weight(), 0);
```
NikVolf: same assert twice?

kianenigma: One should have been len 👍

frame/executive/src/lib.rs
```diff
@@ -711,9 +711,26 @@ mod tests {
...
+			assert_eq!(<frame_system::Module<Runtime>>::all_extrinsics_len(), 0);
+
+			// This is okay -- balances transfer will panic since it requires ensure_signed.
+			assert_eq!(Executive::apply_extrinsic(xt), Ok(Err(DispatchError::BadOrigin)));
+
+			assert_eq!(<frame_system::Module<Runtime>>::all_extrinsics_weight(), len);
```
NikVolf: but how comes that `weight` is equal to `len`?
maybe note it for future generations?

kianenigma: It is noted in the same file in another test 
https://github.com/paritytech/substrate/blob/d43b9f3ff94eb1f914ef77b00489bb42e5e40424/frame/support/src/weights.rs#L316-L327

PR #5001, opened 2020-02-20
Clean up the public API of sc-network-gossip

tomaka:
Let's also clean up the public-facing API of `sc-network-gossip` by turning the futures 0.1 into a futures 0.3.

I don't know if I should mark as "client-noteworthy" considering that this concerns a trait that exists purely in order to be able to mock the network for testing purposes. Polkadot in particular doesn't implement this trait.


expenses:
I did this in #4633 but that PR is on ice because I'm waiting for a new tokio version, so LGTM.

PR #5004, opened 2020-02-20
Wait for RPC server to cleanup on shutdown

arkpar:
RPC servers currently don't wait for worker threads to cleanup. This may result in RPC handlers, which contain a reference to the DB, to be still alive when the main thread terminates, leading to segfaults.

@tomusdrw perhaps `http::Server` should wait on `drop` internally?

### DIFF
--- a/client/service/src/lib.rs
+++ b/client/service/src/lib.rs
@@ -534,6 +534,30 @@ impl<TBl, TCl, TSc, TNetStatus, TNet, TTxPool, TOc> Drop for
 	}
 }
 
+#[cfg(not(target_os = "unknown"))]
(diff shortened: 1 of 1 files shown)


tomusdrw:
The WS server for sure does:
https://github.com/paritytech/jsonrpc/blob/ef44a62c543fd969dd5bedf5a4ece3f3814bb111/ws/src/server.rs#L150

I was sure that HTTP server does as well, but it seems not. Will make a PR upstream.

arkpar:
Should be reverted once jsonrpc with https://github.com/paritytech/jsonrpc/pull/543 is released

Issue #4419, opened 2019-12-17
Implement generic unsigned validation & remove `UnsignedValidator`


jimmychu0807:
Todo:

- Implement generic unsigned validation
- Get rid of the UnsignedValidator stuff

> I guess we could try to move a bunch of this stuff to a SignedExtension (Like InternalSignatureVerifier<T: HasInternalSignature> extension) and the key rotation and signing to some helper
@tomusdrw 

> Hmm, where do we create new keys on session change? That is only done when the user registers new keys. But the module does not need to do anything for that, the keys just need to be registered in the runtime in the SessionKeys object.
@bkchr 

Refer to:
- https://github.com/paritytech/substrate/blob/824009b48482a244f295d299e1a1dd9f358cec63/frame/im-online/src/lib.rs#L517

  Isn't this responsible for creating a separate set of keys? We definitely store them in ImOnline module.

- https://github.com/paritytech/substrate/blob/master/frame/session/src/lib.rs#L486 

  here you set the keys, that are passed to the trait you linked

@bkchr @tomusdrw Feel free to modify and see if I have captured all relevant info to work on this issue.

tomusdrw:
Isn't that a duplicate of #3419 ?

kianenigma:
Related, but no. Doing this PR should be done anyhow, leaving `SignedExtension` with 4 functions. The other issue would try to further shrink/generalize this into a generic set of two funcitons.

tomusdrw:
Right. I've been thinking about this lately a bit though, and I feel that having every pallet expose it's own `SignedExtension` seems like an overkill. I think that we will end up with a trait that pretty much looks exactly like `ValidateUnsigned`, but the way it's invoked will be a bit different. Currently it has a special meaning when declaring a runtime, and I think that instead there should be a generic `SignedExtension` that just calls into pallet's `ValidateUnsigned` (aggregating results from multiple pallets).

So to address points from the original description:
1. This is already possible via `validate_unsigned` in `SignedExtension`, so no change needed here.
2. That might not be necessary, my gut feeling that it would be more ergonomic to stay with `ValidateUnsigned` as described above.
3. That's definitely something we need.

erasmospunk:
I have been working on this issue and implemented some things:

```
Implement new SignedExtension logic to validate unsigned transactions

With the new rules SignedExtension::validate_unsigned will return an error
by default, specifically UnknownTransaction::NoUnsignedValidator to indicate
that unsigned validation was not implemented.
[...]

PR #5006, opened 2020-02-20
Implement ValidateUnsigned as SignedExtension

tomusdrw:
A follow up on #5003 
Related: #3419 
Closes: #4419 

This PR removes a special role `ValidateUnsigned` has in `executive`. Instead we have a `ValidateUnsigned` extension defined in `frame_system`, which dispatches the calls instead.

This is not properly tested yet, and possibly not something we want to include for 2.0, but still wanted to make the PR visible.

Let me know what you think, CC @bkchr @jimmychu0807


kianenigma:
I def. like the idea better as now we don't have two thing flying around doing the same thing: `ValidateUnsigned` and `SignedExtension`. But I think in the very long term, we want only the latter. Even with this PR, it is a bit confusing: You as a runtime dev have two options: 

- implement `ValidateUnsigned` and system will do the rest of the magic for you (if I have understood your code correctly -- had to read it really fast)
- Implement your own `SignedExtension`. 

If we can find a nice way to be able to have the nice features of `ValidateUnsigned` into `SignedExtension`, then I think that is much better. Based on the issue discussions the main problem with `SignedExtension` seem to be the glowing list of `Extra`.

tomusdrw:
> Based on the issue discussions the main problem with SignedExtension seem to be the glowing list of Extra.

Exactly this, so it's definitely possible for every pallet to expose it's own `SignedExtension`, but it seems like an overkill and it adds a lot of boilerplate - i.e. you need to change your `UncheckedExtrinsic::Extra` implementation every time you add a new module that has unsigned transactions which seems insane.

I think that `ValidateUnsigned` implementation for a pallet should be the way to go in the future. It's simple and does the job really well. We already aggregate all `ValidateUnsigned` impls from modules and implement `ValidateUnsigned` for the entire `Runtime`, and it's makes it definitely less effort.

If ever anyone needs the additional info passed to `SignedExtensions` they can definitely implement one for their module. Also we can revisit the set of methods of `SignedExtensions` as suggested in #3419

gavofyork:
@tomusdrw is this a draft or ready for review?
image

tomusdrw:
To be frank, I'm not actively working on this. I was hoping that we will merge that after 2.0 is released, as I didn't put a lot of work into testing it's correctness (and was quite surprised that none of our tests broke :)).

That said, we can review that and let @gnunicorn decide if we want to include in 2.0 or rather put on hold. IMHO this makes the `Applyable` API way cleaner which might help with #5130 as well.

frame/system/src/lib.rs
```diff
@@ -1282,6 +1282,59 @@ impl<T: Trait + Send + Sync> Debug for CheckWeight<T> {
 	}
 }
 
+
+/// Unsigned transactions validation.
+#[derive(Encode, Decode, Clone, Eq, PartialEq)]
+pub struct ValidateUnsigned<T: Trait + Send + Sync>(PhantomData<T>);
```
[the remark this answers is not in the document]
tomusdrw: Not sure what you mean. It doesn't have to be replicated, it's a system-level `SignedExtension` (like `CheckWeight`, etc) that can be plugged to the list of signed extensions and will delegate to `ValidateUnsigned` implementation of the `Runtime`.
Do you think it should be added to some other place?

frame/support/src/unsigned.rs
```diff
@@ -87,7 +91,7 @@ macro_rules! impl_outer_validate_unsigned {
 				#[allow(unreachable_patterns)]
 				match call {
 					$( Call::$module(inner_call) => $module::validate_unsigned(source, inner_call), )*
-					_ => $crate::unsigned::UnknownTransaction::NoUnsignedValidator.into(),
+					_ => $crate::unsigned::InvalidTransaction::NotFullyValidated.into(),
```
tomusdrw, line 94: This is a breaking change. Previously if you made an unsigned call to non-existent module it would merely return `UnknownTransaction`, which is interpreted by the txpool as "not valid now, might be valid at some point in the future". Which is sort of ok, but it has undersirable consequence of not banning the transaction - so basically you can spam the txpool with (one; the same) transaction to non-existent module to make it constantly re-verify it.
I think it's more wise to reject such transaction (pretty much the same would happen if you try to call a non-existent function (we return `InvalidTransaction::Call`)

tomusdrw:
Tests added, should be ready to re-review.

primitives/runtime/src/transaction_validity.rs
```diff
@@ -59,6 +59,13 @@ pub enum InvalidTransaction {
 	/// A transaction with a mandatory dispatch. This is invalid; only inherent extrinsics are
 	/// allowed to have mandatory dispatches.
 	MandatoryDispatch,
+	/// No validator has been able to fully verify this transaction.
+	///
+	/// The transactions are disallowed from being included by default. Some validator
+	/// (`SignedExtension`) needs to whitelist them first. This error usually means that either you
```
gavofyork, line 65: For signed transactions, `SignedExtension` works by default-pass (blacklisting), not whitelisting.

tomusdrw: That's only partially true, since the `ValidTransaction` produced will have no `provides` tags, which will be rejected by the transaction pool anyway.

Now `CheckedExtrinsic` will explicitly return `InvalidTransaction` error if no `SignedExtension` provides any tags in `validate` call. I also unified that with `pre_dispatch` now, so basically it's a whitelist now (in a sense that you need at least one `SignedExtension` to provide some tags, don't need to whitelist specific calls or anything).

gavofyork:
If we really want to combine `ValidateUnsigned` and `SignedExtension`, then I think a better direction would be first to consider the possibility of introducing some way to allow a pallet to state that a `SignedExtension` were mandatory for the runtime and have it auto-included without the need to collated `SignedExtensions` manually as is the case now. We might then revisit this PR in so much as to add the API that would allow `SignedExtension`s to be used for the various `ValidateUnsigned` use-cases that we have. We could then move the `ValidateUnsigned` usages over and retire the API.

tomusdrw:
> I see almost no improvement in the API; the only advatage here that I can see is that we get to lose one generic param in Executive (which could perhaps be lost in a better way by combining with AllModules) and one generic param elsewhere in validate/apply.

The main improvement imho is unified validation pipeline for signed and unsigned transactions, and decoulping unsigned transactions from `Executive` (the generic parameter you've mentioned).

> we have an additional param going into validate

I see that more as an improvement actually - having `TransactionSource` in `SignedExtension` is basically a missing part in my previous PR - I think it should simply be there for consistency.

> we have an additional datapath out of SignedExtension::pre_dispatch with the (rather opaque concept of) IsFullyValidated.

Only added to address Basti's critique about unsigned transactions being valid by default, removed in new approach. However I think this is a moot point anyway, since even though `validate` is going to return `Ok`
 the `ValidTransaction` has no `provides` tags which would cause the pool to reject it anyway.

> We also have an additional item in SignedExtra that must be there or random other stuff in a runtime will break.

I agree this is suboptimal, I want to get rid of all `()` in `SignedExtra` by introducing a single `SignedExtension` that is able to group a tuple of `SignedExtension<AdditionalData=()>`. So that all
adding/removing no-argument validator-like extensions would not alter `SignedExtra` type. But
that's rather a separate PR.

> It is highly questionable whether ValidateUnsigned should be subjugated to SignedExtension and whether the latter should see its API bloated in that cause.

The point is that currently `SignedExtension`s make an impression of being able to handle unsgined transactions
while in reality it's not possible at all. You can't create a `SignedExtension` and whitelist a call
  there, because `UnsignedValidator` takes precedence anyway. `SignedExtension` can only be used to further restrict unsigned transactions (or augment the `ValidTransaction` result). I think having two mechanisms that seem to address the same problem is confusing and that's why I believe this unification makes sense.

----
(Sorry for a wall of text above, tl;dr is below)


@gavofyork I've made another attempt to reduce the surface of API changes:
1. The `InvalidTransaction::NoUnsignedValidator` is special cases for `*_unsigned` pipeline so that it doesn't short-circuit, but rather all SignedExtensions are allowed to whitelist a transaction. This changes the semantics only for unsigned transactions (whitelist).
2. Same behavior for `signed` transactions is now reverted, so as previously they are valid by default (yet they will be rejected by the transaction pool anyway).

If that approach is going to be rejected as well I will split the PR into smaller improvements/fixes that don't change the semantics, namely:
- Extra tests and docs
- Passing `TransactionSource` to `SignedExtension`
- Returning `InvalidTransaction` instead of `UnknownTransaction` in case of unsigned validation failure.

gavofyork:
Happy for the incremental improvements. Also happy to continue with the approach to try to merge the APIs, but like I say in my previous comment, I think that the primary aspect of the approach must be to allow a pallet to state that a `SignedExtension` be mandatory for a runtime that includes the pallet since the API is entirely independent of the pallet APIs, and to handle it appropriately in any edge-cases (e.g. with instanceable pallets).

tomusdrw:
Closing. Will split into smaller PRs.

PR #5007, opened 2020-02-20
Revert "Unsigned transactions should also note weight (#4998)"

NikVolf:
Committed with old Cargo.lock format


NikVolf:
superseded by #5007

PR #5008, opened 2020-02-20
Revert "Update lock file" from "Unsigned transactions should also note weight (#4998)" 

NikVolf:
This reverts commit 50f867a71fed3b1391a1677de875b97acf9aaebc.

and updates Cargo.lock normally (with newer rust)


NikVolf:
it seems another one was merged :)

PR #5009, opened 2020-02-20
Add gitlab job for publishing draft releases

s3krit:
Idea is much the same as we currently do on polkadot - When a new tag is pushed (that fits our pattern for tags, e.g. `v1.2.3`), find a list of labelled changes and generate a changelog based on that. Create a draft release on Github and post about it on Matrix.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -524,7 +524,16 @@ publish-gh-doc:
   after_script:
     - rm -vrf ${HOME}/.gitconfig
 
-
+publish-draft-release:
+  stage:                           publish
+  image:                           parity/tools:latest
+  only:
+    - tags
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -524,7 +524,16 @@ publish-gh-doc:
   after_script:
     - rm -vrf ${HOME}/.gitconfig
 
-
+publish-draft-release:
+  stage:                           publish
+  image:                           parity/tools:latest
+  only:
+    - tags
+    - /^v[0-9]+\.[0-9]+\.[0-9]+.*$/
```
gnunicorn: If I am reading this right, it would not match in `v2.0.0-alpha.1`...

shawntabrizi: It should. The `.*` at the end will match any arbitrary string after `v#.#.#`

Issue #4591, opened 2020-01-10
The identity module use "Sudo" prefix incorrectly


thiolliere:
the prefix used by identity is "Sudo"

https://github.com/paritytech/substrate/blob/ae9c6eac90606c7e612df52d37e89ff1e6d9203b/frame/identity/src/lib.rs#L379

this is probably not intented and might leads to collision with sudo module, however fixing it require migrating Kusama identity module.

kianenigma:
Yeah; I saw this while scraping some data a while back and I hoped it was intentional in some way? I forgot to ask but I could see the author actually wanting to place them under the same tree path for some optimisation reasons?

bkchr:
We can not use `ModuleToIndex` for this, as it would break in the moment you insert a module in between. However, we probably could support this in construct runtime (by just taking the name that the user chooses), but this is a really huge change and requires a storage upgrade. So, nothing that will come in the near future.

PR #5010, opened 2020-02-20
Assign unique storage names to pallets.

shawntabrizi:
This is a PR to ensure that pallets included with Substrate 2.0 each have their own unique storage name so they are placed in separate prefix tries.

If two pallets do not have different storage prefixes, it may have unintended consequences for the end users. I worry if we do not address this before 2.0 release, we may cause trouble to a lot of people in the future who use our pallets.

Some of these changes will require a storage migration, which I will follow up with.
I can also try to improve the `construct_runtime!` macro to do this magic for us... may need some help here.

- [x] Correct Storage Names
- [x] Write Migration Code
- [x] Test Migration Code
- Update macros to prevent this in the future (Another PR)

Pallets in Kusama, need to create migration:

- [x] Identity
- [x] Finality Tracker
- [x] Transaction Payment

Pallets not in Kusama, no migration code written:

- Nicks
- EVM
- Example Offchain Worker
- Elections

Pallets that should probably be updated, but may affect other networks (like edgeware):

- [x] Contracts (`Contract` -> `Contracts`)

Closes https://github.com/paritytech/substrate/issues/4591


shawntabrizi:
FYI a subset of these changes which would not affect Kusama could probably be merged separately if there is any contention about renaming storage prefixes for modules which are being used.

* Nicks
* EVM
* Example Offchain Worker

frame/evm/src/lib.rs
```diff
@@ -131,7 +131,7 @@ pub trait Trait: frame_system::Trait + pallet_timestamp::Trait {
 }
 
 decl_storage! {
-	trait Store for Module<T: Trait> as Example {
+	trait Store for Module<T: Trait> as EVM {
```
sorpaas: Ah my bad!

shawntabrizi: no worries, you are not the only one obviously, and we should make this better. :)

shawntabrizi:
@gnunicorn I think this one needs to be in the 2.0 release

thiolliere:
> I can also try to improve the construct_runtime! macro to do this magic for us... may need some help here.

maybe we can do something similar to `ModuleToIndex` trait in frame-system::Trait.
But that means that every storage is generic over `T: Trait`

also we could at least provide a test in construct_runtime that ensures that every module has a different Prefix.

frame/identity/src/lib.rs
```diff
@@ -872,6 +877,36 @@ decl_module! {
...
+}
+
+// Migration code to update storage name
+use frame_support::storage::migration::{put_storage_value, take_storage_value, StorageIterator};
+impl<T: Trait> Module<T> {
+	pub fn do_upgrade() {
+		sp_runtime::print("Migrating Identity.");
+
+		for (hash, identity_of) in StorageIterator::<Option<Registration<BalanceOf<T>>>>::new(b"Sudo", b"IdentityOf").drain() {
+			put_storage_value(b"Identity", b"IdentityOf", &hash, identity_of);
```
thiolliere: note that we could make use of `<IdentityOf<T>>::storage_prefix()` instead of `b"IdentityOf"` to avoid typos but I'm ok with this if tested.

gnunicorn:
@shawntabrizi, just follow the [process to get PRs merged for the freeze](https://github.com/paritytech/substrate/wiki/Feature-freeze-towards-2.0#process-to-get-prs-merged-during-the-freeze).

gavofyork:
runtime version needs bumping.

shawntabrizi:
@gavofyork et al.

I have removed the `IsUpgraded` and omitted the `StorageVersion` storage item for these migrations because they will either do something, or do nothing.

There seems to be no need to introduce a storage version for these specific name changing upgrades.

Let me know if this is ill advised.

shawntabrizi:
We will wait until the next runtime upgrade to include this one

frame/contracts/src/lib.rs
```diff
@@ -923,7 +923,7 @@ decl_event! {
 }
 
 decl_storage! {
-	trait Store for Module<T: Trait> as Contract {
+	trait Store for Module<T: Trait> as Contracts {
```
gavofyork: This will break FF and EW, won't it?

gnunicorn:
what's the status here? There are open checkboxes in the description, but it seems to come to a close. Can we merge master, resolve the conflicts and get this ready to be merged? Or is there is still something pending, @shawntabrizi ?

shawntabrizi:
We need someone to test the migration. I can try to get to it today, but if we have any volunteers, that would be helpful for me.

frame/transaction-payment/Cargo.toml
```diff
@@ -9,6 +9,7 @@ repository = "https://github.com/paritytech/substrate/"
 description = "FRAME pallet to manage transaction payments"
 
 [dependencies]
+serde = { version = "1.0.101", optional = true }
```
bkchr: Why this change?

gavofyork:
@shawntabrizi i can have a go.

frame/transaction-payment/Cargo.toml
shawntabrizi: You are right, this is no longer needed. But it _was_ needed when I previously added:

```
pub RuntimeVersion build(|_| true): bool;
```

When I added build, and the module did not have serde, I got the following error:

```
error[E0463]: can't find crate for `serde`
  --> frame/transaction-payment/src/lib.rs:79:1
   |
79 | / decl_storage! {
80 | |     trait Store for Module<T: Trait> as TransactionPayment {
81 | |         pub NextFeeMultiplier get(fn next_fee_multiplier): Multiplier = Multiplier::from_parts(0);
82 | |
83 | |         pub RuntimeVersion build(|_| true): bool;
84 | |     }
85 | | }
   | |_^ can't find crate

```

Is this a bug in the macro? Should we auto import serde here if needed?

bkchr: I couldn't find the rustc issue (not sure if there even exists one). However, the problem is that when you have something like this:
```
#[derive(frame_support::serde::Serialize)]
struct Something {}
```

The procedural macro can not see this `frame_support::serde`. Which means it will try to search for `serde` in the current `Cargo.toml`... :( 

So, TLDR, nothing we can do about.

gavofyork:
There is a *LOT* of space indentation in this PR. Please root it out and eliminate it before my eyes burn.

gavofyork:
One systemic error here: any top-level `Option<>` in the type should be dropped when migrating with the migration tooling.

shawntabrizi:
> There is a _LOT_ of space indentation in this PR. Please root it out and eliminate it before my eyes burn.

Sorry, that was a new laptop, and default vs code settings. I have fixed it since.

> One systemic error here: any top-level Option<> in the type should be dropped when migrating with the migration tooling.

Ack.

gavofyork:
Ok - I've tested that as much as is possible, and it looks good. The test I cannot easily do is for migration of FinalityTracker, which I can't easily port across to a new genesis chain and expect to work.

gavofyork:
Ok - I managed to test Finality Tracker also. All good.

PR #5011, opened 2020-02-20
node, node-template: disable GRANDPA observer

andresilva:
Currently both the `node` and `node-template` run the GRANDPA observer protocol rather the full protocol if the node is not an authority. The observer currently has some issues with guaranteeing block and vote data availability, and having most nodes on a network run it could lead to finality stalls. We currently don't use the observer in Polkadot and as such it has not had much extensive testing either. This PR makes sure that both `node` and `node-template` always run the full GRANDPA protocol, the reason to do this is to prevent other projects from eventually using the observer on their network and running into issues (since they have no way to know about this and will mostly copy the service instantiation code from `node` / `node-template`.)

### DIFF
--- a/bin/node-template/node/src/service.rs
+++ b/bin/node-template/node/src/service.rs
@@ -168,6 +168,8 @@ pub fn new_full(config: Configuration<GenesisConfig>)
 			voting_rule: grandpa::VotingRulesBuilder::default().build(),
 		};
 
+		// the GRANDPA voter task is considered infallible, i.e.
+		// if it fails we take down the service with it.
 		service.spawn_essential_task(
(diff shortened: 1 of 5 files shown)


andresilva:
I added the  B1-clientnoteworthy tag because it might make sense to highlight this change, so that users update their service instantiation accordingly. We could also add a deprecation attribute to the observer so that they see the warning when compiling (it is not deprecated since we will keep supporting it, it is just not recommended to use it at this point)? Or maybe just making `run_grandpa_observer` private until it is ready to be used? If we're going to make it private probably makes sense to do it now before the freeze.

Issue #3492, opened 2019-08-27
wasm builder fails when running with a specific toolchain


jbg:
```
rustup target add wasm32-unknown-unknown --toolchain nightly-2019-08-22
cargo +nightly-2019-08-22 build
```

fails with

```
error: failed to run custom build command for `polkadot-runtime v0.6.0 (polkadot/runtime)`

Caused by:
  process didn't exit successfully: `polkadot/target/debug/build/polkadot-runtime-8e31e940c712ba37/build-script-build` (exit code: 1)
--- stderr
   Compiling wasm-build-runner-impl v1.0.0 (polkadot/target/debug/build/polkadot-runtime-eec245fcb90ce820/out/wasm_build_runner)
    Finished dev [unoptimized + debuginfo] target(s) in 0.90s
     Running `polkadot/target/debug/build/polkadot-runtime-eec245fcb90ce820/out/wasm_build_runner/target/debug/wasm-build-runner-impl`
Rust WASM toolchain not installed, please install it!
```

because https://github.com/paritytech/substrate/blob/master/core/utils/wasm-builder/src/lib.rs#L154 hardcodes `rustup run nightly cargo` rather than using the `CARGO` environment variable to find `cargo` from the toolchain that’s in use.

bkchr:
I had used `CARGO` in the beginning, but it would break for some use cases. Looks like I need to think again about it.

emkman:
I am unable to install the latest `subkey` because of this issue on OS X

PR #5021, opened 2020-02-22
Extend rust nightly detection in `wasm-builder`

bkchr:
Instead of just testing `cargo` and `rustup run nightly`, we now test
the `CARGO` env variable and also scan non default nightlies. The user
is also now able to select the toolchain with `WASM_BUILD_TOOLCHAIN`.

Fixes: https://github.com/paritytech/substrate/issues/3492

CC @xlc


gnunicorn:
what's the release story here for this crate? This has been released to crates.io already, right? So we need to bump the minor or patch before we can publish this?

bkchr:
Yeah, it is mostly me who releases this. However, I'm planing some other changes before releasing it.

gnunicorn:
@bkchr but we are going to release alpha.1 against the older (already) released version of the builder then? Then I might switch the version-num of this one to the `-dev`-prerelease.

PR #5024, opened 2020-02-22
Hide benchmarks behind a feature flag

### LINKED ISSUE #4873: Hide benchmarking pipeline behind a feature flag

DarkEld3r:
Closes #4873.


primitives/api/proc-macro/src/impl_runtime_apis.rs
```diff
@@ -163,7 +163,7 @@ fn generate_impl_calls(
 				)?;
 
 				impl_calls.push(
-					(impl_trait_ident.clone(), method.sig.ident.clone(), impl_call)
+					(impl_trait_ident.clone(), method.sig.ident.clone(), impl_call, impl_.attrs.clone())
```
bkchr: I think the best would be to filter the attributes and only propagate `cfg(something)` for now.

primitives/api/proc-macro/src/impl_runtime_apis.rs
```diff
@@ -745,3 +759,32 @@ fn impl_runtime_apis_impl_inner(api_impls: &[ItemImpl]) -> Result<TokenStream> {
 		)
 	)
 }
+
+// Filters all attributes except the cfg ones.
+fn filter_attrs(attrs: &[Attribute]) -> Vec<Attribute> {
```
marcio-diaz:
```suggestion
fn filter_cfg_attrs(attrs: &[Attribute]) -> Vec<Attribute> {
```

bin/node/runtime/Cargo.toml
```diff
@@ -132,3 +132,4 @@ std = [
 	"pallet-recovery/std",
 	"pallet-vesting/std",
 ]
+runtime_benchmarks = ["frame-benchmarking"]
```
expenses: For consistency, I'd prefer this feature to be called `runtime-benchmarks`.

DarkEld3r: Makes sense.

bkchr:
Sorry for the delay: https://github.com/paritytech/substrate/pull/5131

PR #5026, opened 2020-02-24
adding a ss58 format for Darwinia Network

yanganto:
In order to prevent the re-usage across networks, decreasing the value at risk if a private key is lost, and also to identify easily.  We want to registry the Subkey Address Type as 18 for [Darwinia](https://github.com/darwinia-network/darwinia) mainnet, and we will use 42 in our last testnet.  If there are any concerns please kindly tell us.  

Many thanks for your hard-working on Substrate.


yanganto:
[clabot:check].

PR #5029, opened 2020-02-24
Adds `vested_transfer` to Vesting pallet

### LINKED ISSUE #4780: Add a "VestingOrigin" to allow for the creation of vesting schedules indefinitely
It is foreseeable that a project will want to apply a `VestingSchedule` to accounts after the creation of the initial chain spec. Currently, there is not a way to do this without setting vesting in the chain spec or by implementing a new pallet.

The refactor to balances in https://github.com/paritytech/substrate/pull/4649 separates vesting into its own pallet. I propose an addition of a new dispatch that allows for a permissioned account (the `VestingOrigin`) to set new vesting schedules. This could be combined with a transfer of funds so that the vesting schedule cannot be completely arbitrarily set by the vesting origin, but only if it transfers the funds that would be vested at the same time.

Concretely, this is a problem for the W3F Grants team that would like to continue awarding grants with a schedule attached. However, other projects building on Substrate may benefit from this addition as well.

cc @shawntabrizi

lsaether:
closes #4780 

This PR implements a way to make a `vested_transfer` that will transfer currency and create a new vesting schedule for the target account. This means that the the target account must not already have a `VestingSchedule` attached to it. The `vested_transfer` function is permissionless, allowing for anyone in the system to create a new `VestingSchedule` for a target account but requires a minimum transfer amount (`VestingDeposit`) to do so. 

A further enhancement may be to add a maximum `MaxVestingLength` to prevent users from accidentally locking tokens indefinitely, however it's not critical to the logic in this PR.

I opened this PR to get initial review  and can follow up with the necessary changes to the Polkadot codebases if approved.


frame/vesting/src/lib.rs
```diff
@@ -206,6 +216,39 @@ decl_module! {
...
+		) -> DispatchResult {
+			let transactor = ensure_signed(origin)?;
+			let who = T::Lookup::lookup(target)?;
+
+			ensure!(schedule.locked >= T::VestingDeposit::get(), Error::<T>::AmountLow);
+			ensure!(!Vesting::<T>::contains_key(&who), Error::<T>::ExistingVestingSchedule);
+			
+			T::Currency::transfer(&transactor, &who, schedule.locked, ExistenceRequirement::AllowDeath)?;
+
+			Self::add_vesting_schedule(&who, schedule.locked, schedule.per_block, schedule.starting_block)?;
```
shawntabrizi:
```suggestion
			Self::add_vesting_schedule(&who, schedule.locked, schedule.per_block, schedule.starting_block)
			.expect("user does not have an existing vesting schedule; q.e.d.");
```

This line cannot/should not fail.

frame/vesting/src/lib.rs
```diff
@@ -613,4 +661,95 @@ mod tests {
...
+				assert_eq!(user3_free_balance_updated, 256 * 25);
+				let user4_free_balance_updated = Balances::free_balance(&4);
+				assert_eq!(user4_free_balance_updated, 256 * 45);
+				// Account 4 has 5 * 256 locked.
+				assert_eq!(Vesting::vesting_balance(&4), Some(256 * 5));
+
+				System::set_block_number(20);
+				assert_eq!(System::block_number(), 20);
+
+				// Account 4 has 5 * 64 unites vested by block 20.
```
shawntabrizi:
```suggestion
				// Account 4 has 5 * 64 units vested by block 20.
```

frame/vesting/src/lib.rs
```diff
@@ -206,6 +216,40 @@ decl_module! {
...
+		#[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
+		pub fn vested_transfer(
+			origin,
+			target: <T::Lookup as StaticLookup>::Source,
+			schedule: VestingInfo<BalanceOf<T>, T::BlockNumber>
+		) -> DispatchResult {
+			let transactor = ensure_signed(origin)?;
+			let who = T::Lookup::lookup(target)?;
+
+			ensure!(schedule.locked >= T::VestingDeposit::get(), Error::<T>::AmountLow);
```
bkchr:
```suggestion
			let who = T::Lookup::lookup(target)?;
```

frame/vesting/src/lib.rs
```diff
@@ -206,6 +216,40 @@ decl_module! {
...
+		///	   amount needed to succeed.
+		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
+		pub fn vested_transfer(
+			origin,
+			target: <T::Lookup as StaticLookup>::Source,
+			schedule: VestingInfo<BalanceOf<T>, T::BlockNumber>
+		) -> DispatchResult {
+			let transactor = ensure_signed(origin)?;
+			let who = T::Lookup::lookup(target)?;
```
bkchr:
```suggestion
			ensure!(schedule.locked >= T::VestingDeposit::get(), Error::<T>::AmountLow);
```

Just some small optimization ;)

frame/vesting/src/lib.rs
```diff
@@ -206,6 +216,40 @@ decl_module! {
...
+		///
+		/// # <weight>
+		/// - Creates a new storage entry, but is protected by a minimum transfer
+		///	   amount needed to succeed.
+		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedNormal(1_000_000)]
+		pub fn vested_transfer(
+			origin,
+			target: <T::Lookup as StaticLookup>::Source,
+			schedule: VestingInfo<BalanceOf<T>, T::BlockNumber>
```
bkchr:
```suggestion
			schedule: VestingInfo<BalanceOf<T>, T::BlockNumber>,
```

frame/vesting/src/lib.rs
```diff
@@ -206,6 +216,40 @@ decl_module! {
...
+			let transactor = ensure_signed(origin)?;
+			let who = T::Lookup::lookup(target)?;
+
+			ensure!(schedule.locked >= T::VestingDeposit::get(), Error::<T>::AmountLow);
+			ensure!(!Vesting::<T>::contains_key(&who), Error::<T>::ExistingVestingSchedule);
+			
+			T::Currency::transfer(&transactor, &who, schedule.locked, ExistenceRequirement::AllowDeath)?;
+
+			Self::add_vesting_schedule(&who, schedule.locked, schedule.per_block, schedule.starting_block)
+			.expect("user does not have an existing vesting schedule; q.e.d.");
```
bkchr: One tab please.

shawntabrizi:
@gnunicorn this be gucci

Issue #4452, opened 2019-12-19
traitify client


gnunicorn:
the `Client` from `sc-client` is a massive pull-it-all-together within the outer substrate infrastructure. While many features already depend on specific traits rather than the client directly, the client still has a few functions exclusive to it, that some other crates depend on, but some also depend on client without actually needing these functions. The goal is to make nothing in `sc-*` depend on client directly anymore and move client into the internals of `sc-service`.

The way forward is probably (in this order):

- [x] remove `Client<..>` dependencies where other traits are defining the use API well enough already
- [x] create proper `traits` for remaining functionality in `Client` in `sc-api`
- [x] replace remaining `Client`-dependencies with dependencies on `sc-api`
- [x] move `Client` into `sc-service`, make it internal–this probably needs some tweaking for tests (see test-client)
- [x] drop `sc-client`

This is an evolved refactor that probably has to touch a lot of (internal) code and `where`-clauses. Rust experience is required. Expect this to take a while. I am happy to mentor.

seunlanlege:
@gnunicorn Interested in picking this up.

arkpar:
>  drop `sc-client`

I'd suggest leaving it in a separate crate still.  `Client` currently is basically an abstraction over the blockchain database that can handle block import. While service is supposed to be a full-blown substrate node with networking, RPC, etc. which utilises `Client`. I can see `Client` being potentially useful for bridgers, or tools that only need database access. So it is important to keep the API boundary here. I believe some of the CLI commands don't actually create the `Service` instance, just the `Client`, because all they require is database access.

gnunicorn:
> I believe some of the CLI commands don't actually create the Service instance, just the Client, because all they require is database access.

Nope. `client-cli` only depends on `sc-client-api` and `sc-service`. `node-cli` does depend on it and node-template, but only for the `LongestChain`-Selector and `node-cli` only in interaction with `service`, not in direct interaction.

I see your point, though – arguably – that boundary is less clear cut as you make it out to be, `client` itself is already a full-blown thing, with telemetry and link to consensus, which I wouldn't expect to be living there if it was that boundary. If we wanted to make that inner object accessible, which is currently practically impossible to set up without a service, we should make define that boundary more clearly and trim down what that object should contain – and whether it shouldn't instead be a set of traits that we have alternative implementations for.

[...]

PR #5030, opened 2020-02-24
removes use of sc_client::Client from sc_finality_grandpa

seunlanlege:
See #4452

Introduces a new trait `ClientBackend` to `sc_client_api`, for methods that interact with the backend.
eg `fn lock_import_and_run`

### DIFF
--- a/client/api/src/backend.rs
+++ b/client/api/src/backend.rs
@@ -169,6 +169,14 @@ pub trait BlockImportOperation<Block: BlockT> {
 	fn mark_head(&mut self, id: BlockId<Block>) -> sp_blockchain::Result<()>;
 }
 
+/// trait for performing operations on the backend
+pub trait ClientBackend<Block: BlockT, B: Backend<Block>> {
+	/// Lock the import lock, and run operations inside.
+	fn lock_import_and_run<R, Err, F>(&self, f: F) -> Result<R, Err> where
+		F: FnOnce(&mut ClientImportOperation<Block, B>) -> Result<R, Err>,
+		Err: From<sp_blockchain::Error>;
+}
+
 /// Finalize Facilities
 pub trait Finalizer<Block: BlockT, B: Backend<Block>> {
 	/// Mark all blocks up to given as finalized in operation.
--- a/client/api/src/backend.rs
+++ b/client/api/src/backend.rs
@@ -172,9 +172,10 @@ pub trait BlockImportOperation<Block: BlockT> {
 /// trait for performing operations on the backend
 pub trait ClientBackend<Block: BlockT, B: Backend<Block>> {
 	/// Lock the import lock, and run operations inside.
-	fn lock_import_and_run<R, Err, F>(&self, f: F) -> Result<R, Err> where
-		F: FnOnce(&mut ClientImportOperation<Block, B>) -> Result<R, Err>,
-		Err: From<sp_blockchain::Error>;
+	fn lock_import_and_run<R, Err, F>(&self, f: F) -> Result<R, Err>
+		where
+			F: FnOnce(&mut ClientImportOperation<Block, B>) -> Result<R, Err>,
+			Err: From<sp_blockchain::Error>;
 }
 
 /// Finalize Facilities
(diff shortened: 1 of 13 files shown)


client/api/src/backend.rs
```diff
@@ -169,6 +169,14 @@ pub trait BlockImportOperation<Block: BlockT> {
 	fn mark_head(&mut self, id: BlockId<Block>) -> sp_blockchain::Result<()>;
 }
 
+/// trait for performing operations on the backend
```
NikVolf:
```suggestion
/// Interface for performing operations on the backend
```

client/api/src/backend.rs
```diff
@@ -169,6 +169,14 @@ pub trait BlockImportOperation<Block: BlockT> {
...
+pub trait ClientBackend<Block: BlockT, B: Backend<Block>> {
+	/// Lock the import lock, and run operations inside.
+	fn lock_import_and_run<R, Err, F>(&self, f: F) -> Result<R, Err> where
```
NikVolf: where part formatting should be aligned with the rest of the codebase

seunlanlege:
Pinging @gnunicorn @andresilva

client/api/src/backend.rs
```diff
@@ -169,6 +169,15 @@ pub trait BlockImportOperation<Block: BlockT> {
 	fn mark_head(&mut self, id: BlockId<Block>) -> sp_blockchain::Result<()>;
 }
 
+/// Interface for performing operations on the backend.
+pub trait ClientBackend<Block: BlockT, B: Backend<Block>> {
```
gnunicorn: The name of the trait and its only function don't really match up. May `LockedImportRun` would be a better naming?

Not super happy about that name but it is def cleaner, unless you intent to add other functionality to it.

```suggestion
pub trait LockedImportRun<Block: BlockT, B: Backend<Block>> {
```

client/finality-grandpa/src/lib.rs
```diff
@@ -56,15 +56,15 @@ use futures::prelude::*;
 use futures::StreamExt;
 use log::{debug, info};
 use futures::channel::mpsc;
-use sc_client_api::{BlockchainEvents, CallExecutor, backend::{AuxStore, Backend}, ExecutionStrategy};
-use sp_blockchain::{HeaderBackend, Error as ClientError};
+use sc_client_api::{ClientBackend, BlockchainEvents, CallExecutor, backend::{AuxStore, Backend}, ExecutionStrategy, Finalizer, TransactionFor};
```
gnunicorn: please wrap. no lines longer than 100 chars, please.

client/finality-grandpa/src/lib.rs
```diff
@@ -257,6 +257,28 @@ impl<B, E, Block: BlockT, RA> BlockStatus<Block> for Arc<Client<B, E, Block, RA>
 	}
 }
 
+/// A trait that includes all the client functionalities grandpa requires.
+/// Ideally this would be a trait alias, we're not there yet.
```
gnunicorn: do we have a remote github issue we could link to here?

client/finality-grandpa/src/tests.rs
```diff
@@ -18,10 +18,7 @@
 
 use super::*;
 use environment::HasVoted;
-use sc_network_test::{
-	Block, Hash, TestNetFactory, BlockImportAdapter, Peer,
-	PeersClient, PassThroughVerifier,
-};
+use sc_network_test::{Block, Hash, TestNetFactory, BlockImportAdapter, Peer, PeersClient, PassThroughVerifier, PeersFullClient};
```
gnunicorn: please wrap ;)

seunlanlege:
Failing tests pass on my machine, not sure what the problem is.

gnunicorn:
@seunlanlege please don't force push to any branches, especially when they are raised as PRs – it prevents github from picking up the difference between and thus reviewers can't just use the "changes since you last viewed" feature but have to review the entire PR again. We are squash merging in the end anyways, o no need to rebase or similar in between...

seunlanlege:
So sorry about that @gnunicorn, will merge from now on.

PR #5032, opened 2020-02-24
Weight Refunds mechanism RFC

### LINKED ISSUE #3730: Proposal: Allow Dynamic Weights

pepyakin:
This PR is a draft of the changes for weight refunds (aka [dynamic weights](https://github.com/paritytech/substrate/issues/3730), so fixes #3730 if merged).

This is not yet ready for the code related discussion, although design related feedback is welcome.

### DIFF
--- a/frame/system/src/lib.rs
+++ b/frame/system/src/lib.rs
@@ -108,6 +108,7 @@ use sp_runtime::{
 		self, CheckEqual, AtLeast32Bit, Zero, SignedExtension, Lookup, LookupError,
 		SimpleBitOps, Hash, Member, MaybeDisplay, EnsureOrigin, BadOrigin, SaturatedConversion,
 		MaybeSerialize, MaybeSerializeDeserialize, MaybeMallocSizeOf, StaticLookup, One, Bounded,
+		Dispatchable
 	},
 };
 
@@ -118,7 +119,7 @@ use frame_support::{
 		Contains, Get, ModuleToIndex, OnNewAccount, OnKilledAccount, IsDeadAccount, Happened,
 		StoredMap
 	},
-	weights::{Weight, DispatchInfo, DispatchClass, SimpleDispatchInfo, FunctionOf},
+	weights::{Weight, DispatchInfo, GetDispatchInfo, DispatchClass, SimpleDispatchInfo, FunctionOf},
 };
 use codec::{Encode, Decode, FullCodec, EncodeLike};
 
@@ -347,6 +348,9 @@ decl_storage! {
 
 		/// Total weight for all extrinsics put together, for the current block.
 		AllExtrinsicsWeight: Option<Weight>;
+
+		/// The weight taken by the current executing disptachable.
+		CurrentDispatchableWeight: Option<Weight>;
 
 		/// Total length (in bytes) for all extrinsics put together, for the current block.
 		AllExtrinsicsLen: Option<u32>;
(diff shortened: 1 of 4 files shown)


frame/system/src/lib.rs
```diff
@@ -923,6 +928,44 @@ impl<T: Trait> Module<T> {
...
+	}
+
+	/// Decrease the amount of weight consumed by the current dispatchable.
+	///
+	/// This function can only meaningfully called within `Self::dispatch`. Otherwise,
+	/// it does nothing and returns `Err`.
+	///
+	/// The total weight returned from the dispatchable cannot exceed its total requested weight.
+	pub fn return_unspent_weight(unspent: Weight) -> Result<(), ()> {
+		let weight = match CurrentDispatchableWeight::take() {
```
kianenigma:
```suggestion
		let weight = match CurrentDispatchableWeight::take().ok_or(())?
```

(I don't see much points in doing such reviews on a _draft_ -- this is an exception)

frame/transaction-payment/src/lib.rs
```diff
@@ -197,15 +232,30 @@ impl<T: Trait + Send + Sync> sp_std::fmt::Debug for ChargeTransactionPayment<T>
...
+	fn pre_dispatch(
+		self,
+		who: &Self::AccountId,
+		_call: &Self::Call,
+		info: Self::DispatchInfo,
+		len: usize,
+	) -> Result<Self::Pre, TransactionValidityError> {
+		let tip = self.0;
+		let fee = Self::compute_fee(len as u32, info, tip);
+		Self::withdraw_fee(who, tip, fee).map_err(|()| InvalidTransaction::Payment.into())
```
kianenigma: Let see if this is safe:

- If I panic within the code of the call itself, we have already charged the fee. That is good.
- But for example if the fee was meant to go to authors and treasury, we don't use the imbalance and it is just dropped, so we burn it. Is this desired? Or maybe it is addressed further down in the code. But we should def. discuss this scenario.

kianenigma: @thiolliere actually, you are also changing this code so pinging you

frame/system/src/lib.rs
```diff
@@ -923,6 +928,44 @@ impl<T: Trait> Module<T> {
...
+	///
+	/// Returns how much weight was unspent.
+	pub fn dispatch_call(
+		origin: <T::Call as Dispatchable>::Origin,
+		call: T::Call,
+	) -> (Weight, sp_runtime::DispatchResult) {
+		let original_weight = call.get_dispatch_info().weight;
+		CurrentDispatchableWeight::put(original_weight);
+		let result = call.dispatch(origin);
+		// TODO: This won't work with the nested calls.
```
kianenigma: didn't we come up with a way in out call that it did (I have to watch it again though.. damn memory)

frame/system/src/lib.rs
```diff
@@ -923,6 +928,44 @@ impl<T: Trait> Module<T> {
...
+		};
+		let next_weight = weight.checked_sub(unspent).ok_or(())?;
+		CurrentDispatchableWeight::put(next_weight);
+		Ok(())
+	}
+
+	/// Dispatch a given `T::Call` with a given origin.
+	///
+	/// Returns how much weight was unspent.
+	pub fn dispatch_call(
```
kianenigma:
```suggestion
	pub fn note_dispatch_call(
```
There should really be an emphasis on the fact that this some bookkeeping for you. Chose some other name if you want, but  `dispatch_call ` is too simple.

frame/system/src/lib.rs
pepyakin: Actually we came up! I assumed that this should be trivial to solve: we preserve the value of `CurrentDispatchableWeight` before and restore it after the dispatch.

frame/system/src/lib.rs
pepyakin: aha yeah agreed. I would not have been able to help myself as well if I saw something like this )

frame/transaction-payment/src/lib.rs
```diff
@@ -214,33 +264,41 @@ impl<T: Trait + Send + Sync> SignedExtension for ChargeTransactionPayment<T>
...
+		};
+
+		let actual_weight_fee = <frame_system::Module<T>>::spent_weight()
+			.map(|spent| T::WeightToFee::convert(spent))
+			.unwrap_or_default();
+		let refund = imbalance.peek().saturating_sub(actual_weight_fee);
+
+		let refund_imbalance = T::Currency::deposit_creating(&transactor, refund);
+		if let Ok(imbalance) = imbalance.offset(refund_imbalance) {
+			T::OnTransactionPayment::on_unbalanced(imbalance);
```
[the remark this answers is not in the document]
pepyakin: I am not sure what you mean here because panic is still a pretty-much a game over scenario atm since in case of a panic all changes are reverted, including `pre_dispatch`. There are some mitigations planned, but I wouldn't say that they are on the horizon yet.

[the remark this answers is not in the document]
kianenigma: yeah totally my misunderstanding ✋

frame/system/src/lib.rs
```diff
@@ -321,6 +322,9 @@ decl_storage! {
 		/// Total weight for all extrinsics put together, for the current block.
 		AllExtrinsicsWeight: Option<Weight>;
 
+		/// The weight taken by the current executing disptachable.
```
athei: typo: disptachable

frame/system/src/lib.rs
```diff
@@ -997,6 +1001,57 @@ impl<T: Trait> Module<T> {
...
+
+	/// Decrease the amount of weight consumed by the current dispatchable.
+	///
+	/// This function can only meaningfully called within `Self::dispatch`. Otherwise,
+	/// it does nothing and returns `Err`.
+	///
+	/// The total weight returned from the dispatchable cannot exceed its total requested weight.
+	pub fn return_unspent_weight(unspent: Weight) -> Result<(), ()> {
+		let weight = CurrentDispatchableWeight::take().ok_or(())?;
+		let next_weight = weight.checked_sub(unspent).ok_or(())?;
```
athei: Is Weight guaranteed to be always unsigned? Should we handle the case where `next_weight` becomes negative here?

frame/system/src/lib.rs
```diff
@@ -1221,6 +1276,16 @@ impl<T: Trait + Send + Sync> SignedExtension for CheckWeight<T> {
 	) -> TransactionValidity {
 		Self::do_validate(info, len)
 	}
+
+	fn post_dispatch(_pre: Self::Pre, info: Self::DispatchInfo, _len: usize) {
+		let spent = CurrentDispatchableWeight::get().unwrap_or(info.weight);
+		let reserved = info.weight;
+		let unspent = spent - reserved;
```
athei: I think a comment why unchecked math is OK here is in order.

frame/system/src/lib.rs
kianenigma: Yes weight is always unsigned and cannot be negative.

pepyakin:
As a status update on this, there is a problem with how to introduce a global central dispatch point.

athei:
Rebased onto current master and resolved conflicts with it. Removed some unrelated style changes that added to the noise of this PR.

Next steps will be to integrate it with the changes in #5130 as soon as there is some confidence on how this GCD will manifest.

athei:
Closed in favor of #5584

PR #5034, opened 2020-02-24
Add more metrics to prometheus

expenses:


client/tracing/src/lib.rs
```diff
@@ -287,3 +295,32 @@ fn send_telemetry(span_datum: SpanDatum) {
...
+				&["name", "target", "line", "values"]
+			)?, registry)?,
+		})
+	}
+}
+
+fn send_prometheus(span_datum: SpanDatum, metrics: &Metrics) {
+	metrics.profiling
+		.with_label_values(&[
+			span_datum.name, span_datum.target, &span_datum.line.to_string(), &span_datum.values.to_string()
```
mxinden: Do I understand correctly, that this would add a new metric label combination for each sample? That would result in a [cardinality explosion](https://www.robustperception.io/cardinality-is-key).

What questions would you like to answer with the information exposed here? Keep in mind that Prometheus is a metric system and not a tracing system.

client/network/src/service.rs
```diff
@@ -727,6 +732,26 @@ pub struct NetworkWorker<B: BlockT + 'static, H: ExHashT> {
 	light_client_rqs: Option<mpsc::UnboundedReceiver<RequestData<B>>>,
 	/// Senders for events that happen on the network.
 	event_streams: Vec<mpsc::UnboundedSender<Event>>,
+	/// Prometheus network metrics.
+	metrics: Option<Metrics>
+}
+
+struct Metrics {
+	is_major_syncing: Gauge<U64>,
+	peers_count: Gauge<U64>,
```
mxinden: Moving this to `client/network` looks a lot cleaner.

client/tracing/src/lib.rs
expenses: Oh, I see. I think we should probably ask @mattrutherford if this is needed.

mattrutherford: Not entirely sure how this will look on Prometheus side - but might be worth considering if we should collect spans (instead of sending each value) and provide Prometheus histogram instead. I guess it would not be practical to partition things by `values`, but including the line number would be nice. Maybe the metrics names could be derived by combining target + name + line eg: `pallet_society_on_initialize_680` and then have quantiles for that?

expenses: Thinking about it, I'm not sure there will be much of a cardinality problem given that the name, target and line (and values I think) are tied together.

client/service/src/builder.rs
```diff
@@ -887,6 +888,15 @@ ServiceBuilder<
 		let block_announce_validator =
 			Box::new(sp_consensus::block_validation::DefaultBlockAnnounceValidator::new(client.clone()));
 
+		// Prometheus registry and port
```
mxinden:
```suggestion
```

I don't think this comment is needed here.

expenses:
Agreed, I'll switch this PR to just be the networking tracing.

expenses:
This is ready for review, but I'm happy to accept more ideas of metrics to track.

client/finality-grandpa/src/environment.rs
```diff
@@ -400,6 +419,17 @@ impl<B, E, Block: BlockT, N: NetworkT<Block>, RA, SC, VR> Environment<B, E, Bloc
 		self.voter_set_state.with(|voter_set_state| {
 			if let Some(set_state) = f(&voter_set_state)? {
 				*voter_set_state = set_state;
+
+				if let Some(metrics) = self.metrics.as_ref() {
```
mxinden: Using `environment.rs` as a source of truth for the current round was a guess from my side. @andresilva could you approve that that is a good idea?

mxinden:
Given its changes to `Service` I am guessing this needs a corresponding Polkadot pull request @expenses?

mxinden:
:heavy_check_mark:  Tested with https://github.com/paritytech/substrate/blob/master/.maintain/sentry-node/docker-compose.yml.

client/finality-grandpa/src/lib.rs
```diff
@@ -670,6 +675,7 @@ where
 			authority_set: persistent_data.authority_set.clone(),
 			consensus_changes: persistent_data.consensus_changes.clone(),
 			voter_set_state: persistent_data.set_state.clone(),
+			metrics: prometheus_registry.map(|registry| Metrics::register(&registry).unwrap()),
```
NikVolf: At least `.expect("Reason/what failed")` instead of `unwrap`

And are we. sure we can panic here at all?

client/finality-grandpa/src/environment.rs
```diff
@@ -376,6 +377,23 @@ impl<Block: BlockT> SharedVoterSetState<Block> {
...
+pub(crate) struct Metrics {
+	finality_grandpa_round: Gauge<U64>,
+}
+
+impl Metrics {
+	pub(crate) fn register(registry: &prometheus_exporter::Registry) -> Result<Self, PrometheusError> {
+		Ok(Self {
+			finality_grandpa_round: register(Gauge::new(
+				"finality_grandpa_round", "Highest completed GRANDPA round.",
+			)?, registry)?,
```
NikVolf:
```suggestion
			finality_grandpa_round: register(
				Gauge::new("finality_grandpa_round", "Highest completed GRANDPA round.")?,
				registry,
			)?,
```

client/finality-grandpa/src/lib.rs
mxinden: > And are we. sure we can panic here at all?

Pretty sure we can't. One should be able to operate finality-grandpa without a Prometheus registry, in which case this would panic.

expenses: No, I don't believe we can panic here because we try to register other metrics before this.

PR #5035, opened 2020-02-24
Add storage root recalculation time to benchmarks

shawntabrizi:
This is a small PR which additionally adds storage root recalculation time to our pallet benchmarking setup.

Storage root recalculation is measured using:

```
sp_io::storage::root
```

Results now look like:

```
Pallet: "pallet-identity", Extrinsic: "add_registrar", Steps: 10, Repeat: 100
r,extrinsic_time,storage_root_time
1,305000,65000
1,97000,39000
1,281000,39000
1,102000,41000
```

### DIFF
--- a/frame/benchmarking/src/lib.rs
+++ b/frame/benchmarking/src/lib.rs
@@ -20,6 +20,8 @@
 
 mod utils;
 pub use utils::*;
+#[doc(hidden)]
+pub use sp_io::storage::root as storage_root;
 
 /// Construct pallet benchmarks for weighing dispatchables.
 ///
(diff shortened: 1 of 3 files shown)


shawntabrizi:
The two values should be loosely correlated. We need to run benchmarks to see.

The idea is you could have some very simple extrinsic, which makes a TON of changes to storage
while you have a very complicated extrinsic, which does touches only one storage item.

Before we were only measuring computation complexity, now we also look at the storage root complexity.

We will probably need to create yet another pipeline to do more accurate benchmarking of storage root calculation in general.

PR #5037, opened 2020-02-24
Remove CratesOrPath variant

cowboy-bebug:
I noticed that this lines in `build.rs` gives an error when `util/wasm-builder` is not found locally:
```rust
	WasmBuilder::new()
		.with_current_project()
		.with_wasm_builder_from_crates_or_path("1.0.9", "../../../utils/wasm-builder")
		.export_heap_base()
		.import_memory()
		.build()
```

Error:
```
Caused by:
  process didn't exit successfully: `/Users/ericlim/project/something/target/debug/build/something-runtime-b9cbca8cf16d835c/build-script-build` (exit code: 1)
--- stderr
error: failed to read `/Users/ericlim/utils/wasm-builder/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

This PR addresses one way to go around this problem (not the best solution but let me know!)

### DIFF
--- a/utils/wasm-builder-runner/src/lib.rs
+++ b/utils/wasm-builder-runner/src/lib.rs
@@ -139,11 +139,10 @@ impl WasmBuilderSelectSource {
 		version: &'static str,
 		path: &'static str,
 	) -> WasmBuilder {
-		WasmBuilder {
-			source: WasmBuilderSource::CratesOrPath { version, path },
-			rust_flags: Vec::new(),
-			file_name: None,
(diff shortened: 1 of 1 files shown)


bkchr:
We use this variant in Substrate and it works. This is normal `Cargo` behavior that the build fails, when the given path is invalid.

Nevertheless, ty for the pr. However, as this is no bug, I will close the pr.

(As documented, the path needs to be relative from `Cargo.toml` file)

PR #5041, opened 2020-02-25
Fix/div by zero

cowboy-bebug:
Hi guys. There might be a potential oversight on division by zero around `gas_price`.
For example, setting `gas_price = 0` here
https://github.com/paritytech/substrate/blob/41bb2193a267805e2093a081bc3e2aaccc64283a/frame/contracts/src/tests.rs#L233-L239

produces
```
---- tests::call_removed_contract stdout ----
thread 'tests::call_removed_contract' panicked at 'attempt to divide by zero',

---- tests::call_contract_removals stdout ----
thread 'tests::call_contract_removals' panicked at 'attempt to divide by zero',
...
```

If u guys are already aware of it, please ignore me 😄

### DIFF
--- a/frame/contracts/src/gas.rs
+++ b/frame/contracts/src/gas.rs
@@ -250,7 +250,11 @@ pub fn refund_unused_gas<T: Trait>(
 pub fn approx_gas_for_balance<Balance>(gas_price: Balance, balance: Balance) -> Gas
 	where Balance: AtLeast32Bit
 {
-	(balance / gas_price).saturated_into::<Gas>()
+	if gas_price.is_zero() {
+		Zero::zero()
+	} else {
+		(balance / gas_price).saturated_into::<Gas>()
+	}
 }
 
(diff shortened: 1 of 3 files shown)


frame/contracts/src/gas.rs
```diff
@@ -250,7 +250,11 @@ pub fn refund_unused_gas<T: Trait>(
 pub fn approx_gas_for_balance<Balance>(gas_price: Balance, balance: Balance) -> Gas
 	where Balance: AtLeast32Bit
 {
-	(balance / gas_price).saturated_into::<Gas>()
+	if gas_price.is_zero() {
+		Zero::zero()
+	} else {
+		(balance / gas_price).saturated_into::<Gas>()
+	}
```
thiolliere: alternatively you could write `balance.checked_div(gas_price).unwrap_or_else(|| Zero::zero()).saturating_into::<Gas>()`

PR #5042, opened 2020-02-25
client/network-gossip: Merge GossipEngine and GossipEngineInner

mxinden:
This pull request contains two commits, the latter supporting the former.

1. client/network-gossip: Merge GossipEngine and GossipEngineInner

    Given that GossipEngine and GossipEngineInner are not shared between threads
    anyone (see https://github.com/paritytech/substrate/pull/4767), neither depend
    on being Send or Sync. Thus one can merge the two as done in this patch. One
    only needs to wrap an `Arc<Mutex<>>` around the whole structure when the owner
    (e.g. finality-grandpa) needs to share the gossip engine between threads.

2. client/finality-grandpa: Wrap GossipEngine in Arc Mutex & lock it on use

    GossipEngine in itself has no need to be Send and Sync, given that it
    does not rely on separately spawned background tasks anymore. Given that
    finality-grandpa shares the `NetworkBridge` potentially between threads
    its components need to be clonable, thus this patch wraps `GossipEngine`
    in an `Arc<Mutex<>>`.

This would require changes within Polkadot which I have not yet prepared, therefore please don't merge this pull request just yet.


client/network-gossip/Cargo.toml
```diff
@@ -14,7 +14,6 @@ futures-timer = "3.0.1"
 libp2p = { version = "0.16.1", default-features = false, features = ["libp2p-websocket"] }
 log = "0.4.8"
 lru = "0.1.2"
-parking_lot = "0.10.0"
```
mxinden: This line probably summarizes the pull request the best.

tomaka:
Considering that each gossip protocol should have its own exclusive instance of `GossipEngine`, it looks good to me.

tomaka:
I didn't mark the PR as approved because of:

> This would require changes within Polkadot which I have not yet prepared, therefore please don't merge this pull request just yet.

But consider it approval.

tomaka:
This is technically an API breaking change, no? So we can't do it during the stabilization period.

mxinden:
> This is technically an API breaking change, no? So we can't do it during the stabilization period.

@tomaka yes, I am guessing so, given that `GossipEngine` is part of the public API and still `Send` but not `Sync` anymore after this patch.

@gnunicorn what is the right procedure here? Should I be adding the pull request to the 3.0 milestone?

gnunicorn:
if it can be raised during alpha-phase, we can consider still merging this before 2.0 – I consider not-too-involved API-clean-up PRs as "fixes". but it should have the corresponding polkadot PR pending. API changes will only be restricted once we reach beta-phase.

mxinden:
Corresponding Polkadot pull request: https://github.com/paritytech/polkadot/pull/890

@gnunicorn as far as I can tell we have not yet reached the _beta_ phase. Could you reconsider merging this?

gnunicorn:
sure, just follow the [process lined out to get PRs merged during the freeze](https://github.com/paritytech/substrate/wiki/Feature-freeze-towards-2.0#process-to-get-prs-merged-during-the-freeze).

andresilva:
I thought we had not stabilized the consensus related parts, isn't it why this crate is at verison `0.8` rather than `2.0`?

mxinden:
> sure, just follow the process lined out to get PRs merged during the freeze.

@gnunicorn thanks a bunch for the help and writing all of this down. I also added `B1-apinoteworthy` given that it changes the public facing Substrate API.

PR #5043, opened 2020-02-25
Fix: CI failing for some CLI tests

cecton:
There seems to be quite a lot of timeouts in the CI lately for stopping the nodes:
* https://gitlab.parity.io/parity/substrate/-/jobs/396630
* https://gitlab.parity.io/parity/substrate/-/jobs/396707
* https://gitlab.parity.io/parity/substrate/-/jobs/396726

Some tests are not using `--dev` and because of that they take time to connect.

This change will display a message when the process takes too long to exit (> 5 seconds) and use `--dev` everywhere.

### DIFF
--- a/bin/node/cli/tests/common.rs
+++ b/bin/node/cli/tests/common.rs
@@ -46,14 +46,11 @@ pub fn wait_for(child: &mut Child, secs: usize) -> Option<ExitStatus> {
 }
 
 /// Run the node for a while (30 seconds)
-pub fn run_command_for_a_while(base_path: &Path, dev: bool) {
+pub fn run_command_for_a_while(base_path: &Path) {
 	let mut cmd = Command::new(cargo_bin("substrate"));
 
-	if dev {
-		cmd.args(&["--dev", "--pruning", "archive"]);
-	}
-
 	let mut cmd = cmd
+		.args(&["--dev", "--pruning", "archive"])
 		.arg("-d")
 		.arg(base_path)
 		.spawn()
--- a/bin/node/cli/tests/common.rs
+++ b/bin/node/cli/tests/common.rs
@@ -27,13 +27,18 @@ use nix::unistd::Pid;
 ///
 /// Returns the `Some(exit status)` or `None` if the process did not finish in the given time.
 pub fn wait_for(child: &mut Child, secs: usize) -> Option<ExitStatus> {
-	for _ in 0..secs {
+	for i in 0..secs {
 		match child.try_wait().unwrap() {
-			Some(status) => return Some(status),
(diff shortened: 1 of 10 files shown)


client/cli/src/params/pruning_params.rs
```diff
@@ -0,0 +1,77 @@
...
+	pub pruning: Option<String>,
+
+	/// Force start with unsafe pruning settings.
+	///
+	/// When running as a validator it is highly recommended to disable state
+	/// pruning (i.e. 'archive') which is the default. The node will refuse to
+	/// start as a validator if pruning is enabled unless this option is set.
+	#[structopt(long = "unsafe-pruning")]
+	pub unsafe_pruning: bool,
+
```
bkchr:
```suggestion
```

client/cli/src/params/pruning_params.rs
```diff
@@ -0,0 +1,77 @@
...
+	#[structopt(long = "pruning", value_name = "PRUNING_MODE")]
+	pub pruning: Option<String>,
+
+	/// Force start with unsafe pruning settings.
+	///
+	/// When running as a validator it is highly recommended to disable state
+	/// pruning (i.e. 'archive') which is the default. The node will refuse to
+	/// start as a validator if pruning is enabled unless this option is set.
+	#[structopt(long = "unsafe-pruning")]
+	pub unsafe_pruning: bool,
```
bkchr: `--unsafe-pruning` does not makes any sense for `export-block` for example. You should have just added `pruning` to `export-block`.

bin/node/cli/tests/common.rs
```diff
@@ -27,28 +27,30 @@ use nix::unistd::Pid;
...
+	eprintln!("Took too long to exit (> {} seconds). Killing...", secs);
 	let _ = child.kill();
 	child.wait().unwrap();
 
 	None
 }
 
 /// Run the node for a while (30 seconds)
-pub fn run_command_for_a_while(base_path: &Path, dev: bool) {
+pub fn run_command_for_a_while(base_path: &Path) {
```
bkchr: `run_dev_node_for_a_while()`

bin/node/cli/tests/common.rs
```diff
@@ -27,28 +27,30 @@ use nix::unistd::Pid;
...
 	let mut cmd = Command::new(cargo_bin("substrate"));
 
-	if dev {
-		cmd.arg("--dev");
-	}
-
 	let mut cmd = cmd
+		.args(&["--dev", "--pruning", "archive"])
```
bkchr: `--dev` implies `--pruning archive`

PR #5045, opened 2020-02-25
Use a Kademlia instance per `ProtocolId`.

twittner:
This PR changes discovery to use one kademlia instance per protocol ID. The change is motivated by the fact that the single DHT we use now includes incompatible nodes with different protocols.

To retain backwards compatibility we currently use a fake `ProtocolId` matching the existing `/ipfs/kad/1.0.0` protocol string. If and when this change is rolled out we can remove it and work solely with `ProtocolId`s taken from the chain specification.

Currently the implementation relies on changes in libp2p master which have yet to be published so a patch section is added to `Cargo.toml` which should be removed before merging this PR.


tomaka:
I think libp2p 0.16.2 contains the methods that you're using, so we can remove the `[patch]` section.

client/network/src/behaviour.rs
```diff
@@ -86,6 +87,11 @@ impl<B: BlockT, H: ExHashT> Behaviour<B, H> {
 		}
 	}
 
+	/// Enable discovery for the given protocol.
+	pub fn add_discovery(&mut self, p: ProtocolId) {
```
tomaka: Since it doesn't make sense to add support for a protocol at runtime, I would instead pass a list of protocols as part of the `new` function.

client/network/src/discovery.rs
```diff
@@ -105,16 +119,9 @@ impl DiscoveryBehaviour {
...
-		let store = MemoryStore::new(local_id.clone());
-		let mut kademlia = Kademlia::new(local_id.clone(), store);
-		for (peer_id, addr) in &user_defined {
-			kademlia.add_address(peer_id, addr.clone());
-		}
-
 		DiscoveryBehaviour {
 			user_defined,
-			kademlia,
+			kademlias: HashMap::new(),
```
tomaka: I would always add the default Kademlia here, so that users don't have to worry about that.

client/network/src/discovery.rs
```diff
@@ -137,9 +144,42 @@ impl DiscoveryBehaviour {
...
+			kad.add_address(peer_id, addr.clone());
+		}
+
+		self.kademlias.insert(p, kad);
+	}
+
 	/// Returns the list of nodes that we know exist in the network.
 	pub fn known_peers(&mut self) -> impl Iterator<Item = &PeerId> {
-		self.kademlia.kbuckets_entries()
+		self.kademlias.values_mut().map(|k| k.kbuckets_entries()).flatten()
```
tomaka: This is probably going to produce a lot of duplicates.

client/network/src/discovery.rs
```diff
@@ -159,22 +199,28 @@ impl DiscoveryBehaviour {
...
+		for k in self.kademlias.values_mut() {
+			k.get_record(key, Quorum::One)
+		}
 	}
 
 	/// Start putting a record into the DHT. Other nodes can later fetch that value with
 	/// `get_value`.
 	///
 	/// A corresponding `ValuePut` or `ValuePutFailed` event will later be generated.
 	pub fn put_value(&mut self, key: record::Key, value: Vec<u8>) {
```
tomaka: `get_value` and `put_value` ultimately should be changed to pass a `ProtocolId` to designate which DHT to target.
I'm unsure if we should do this now though.

client/network/src/discovery.rs
```diff
@@ -271,45 +342,61 @@ impl NetworkBehaviour for DiscoveryBehaviour {
...
 	fn inject_node_event(
 		&mut self,
 		peer_id: PeerId,
-		event: <Self::ProtocolsHandler as ProtocolsHandler>::OutEvent,
+		(pid, event): <Self::ProtocolsHandler as ProtocolsHandler>::OutEvent,
 	) {
-		NetworkBehaviour::inject_node_event(&mut self.kademlia, peer_id, event)
+		if let Some(kad) = self.kademlias.get_mut(&pid) {
+			kad.inject_node_event(peer_id, event)
+		}
```
tomaka: Let's not let bugs go potentially undetected!

```suggestion
		} else {
			error!(target: "sub-libp2p", "State inconsistency: Kademlia event with no associated behaviour");
		}
```

Same for all the other `for` loops below.

client/network/src/discovery.rs
```diff
@@ -479,6 +573,170 @@ impl NetworkBehaviour for DiscoveryBehaviour {
...
+
+	fn listen_protocol(&self) -> SubstreamProtocol<Self::InboundProtocol> {
+		let upgrades = self.handlers.iter()
+			.map(|h| (h.0.clone(), h.1.listen_protocol().into_upgrade().1))
+			.collect();
+		SubstreamProtocol::new(KadUpgrade { upgrades })
+	}
+
+	fn inject_fully_negotiated_outbound
+		( &mut self
```
tomaka: As with the previous PR, I'm really not a fan of this code style. This isn't Haskell.

client/network/src/discovery.rs
```diff
@@ -479,6 +573,170 @@ impl NetworkBehaviour for DiscoveryBehaviour {
...
+		, protocol: <Self::OutboundProtocol as OutboundUpgrade<NegotiatedSubstream>>::Output
+		, (pid, rest): Self::OutboundOpenInfo
+		)
+	{
+		if let Some((_, kad)) = self.handlers.iter_mut().find(|p| &p.0 == &pid) {
+			kad.inject_fully_negotiated_outbound(protocol, rest)
+		}
```
tomaka:
```suggestion
		} else { error!(...) }
```

client/network/src/discovery.rs
```diff
@@ -479,6 +573,170 @@ impl NetworkBehaviour for DiscoveryBehaviour {
 	}
 }
 
+pub struct DiscoveryHandler {
```
tomaka: I would really prefer if ultimately this kind of combinators was in libp2p rather than Substrate.
What about opening a libp2p issue and adding a `TODO` here?

client/network/src/behaviour.rs
twittner: Is it conceivable to run discovery only with MDNS and bootnodes?

client/network/src/discovery.rs
twittner: Is the "default" not a temporary hack that should be removed?

client/network/src/discovery.rs
twittner: Do you want me to deduplicate here?

twittner: Well, nevermind, I deduplicate the peers now.

client/network/src/behaviour.rs
tomaka: In practice it doesn't really make sense, but I also don't see a reason to forbid that use case.

client/network/src/discovery.rs
tomaka: Yes, but "temporary" probably means a couple months, which is long enough to do it a bit properly.

twittner: What do you mean by "which is long enough to do it a bit properly"?

client/network/src/discovery.rs
```diff
@@ -45,85 +45,131 @@
...
+				// Temporary hack to retain backwards compatibility. Once this version
+				// has been rolled out we should remove the special handling of
+				// DEFAULT_PROTO_NAME and work only with proper protocol IDs.
+				Vec::from(p.as_bytes())
+			} else {
+				let mut v = vec![b'/'];
+				v.extend_from_slice(p.as_bytes());
+				v.extend_from_slice(b"/kad");
+				v
+			};
```
tomaka, line 144: I think it wouldn't be a bad idea to get rid of this hack.
What about something like this? (pseudocode)

```rust
impl DiscoveryConfig {
    pub fn new() {
        ...
        config.add_protocol_inner(libp2p::kad::protocol::DEFAULT_PROTO_NAME)
    }

    pub fn add_protocol(&mut self, p: ProtocolId) {
        let proto_name = "/" .. p .. "/kad";  // I know this is not Rust
        self.add_protocol_inner(proto_name)
    }

    // private
    fn add_protocol_inner(&mut self, name: Vec<u8>) { ... }
}
```

tomaka:
Please merge master in order for the Polkadot check to pass.

PR #5051, opened 2020-02-25
Fix documentation for "BlockBuilder::push_trusted" 

NikVolf:


client/block-builder/src/lib.rs
```diff
@@ -131,8 +131,7 @@ where
 
 	/// Push onto the block's list of extrinsics.
 	///
-	/// This will treat incoming extrinsic `xt` as untrusted and perform additional checks
-	/// (currenty checking signature).
+	/// This will treat incoming extrinsic `xt` as trusted and skip any signature checks within it.
```
bkchr: Hmm while this is true for signed transaction, it is not correct for unsigned ones where we have custom validation functions that will still check the signature.

NikVolf: rephrased

PR #5052, opened 2020-02-25
Sync: validate block responses for required data

arkpar:
Client API currently allows importing block without a body, skipping state verification and enactment.
When doing full sync we request block bodies from peers, but don't actually check if peers send them back, passing only headers to block import. The chain ends up being partially synced in "light mode". This PR ensures that blocks coming from the network always have bodies, when doing full client sync.

There's also come confusion in the Client API, where `best_block` does not necessarily means that the state for this block is available.  When doing full sync, the sync module assumes that the state for the best block is always available and therefore can't recover when that's not the case. This should be addressed in a separate PR.

Fixes https://github.com/paritytech/polkadot/issues/861, https://github.com/paritytech/polkadot/issues/859


client/network/src/protocol.rs
```diff
@@ -754,6 +758,11 @@ impl<B: BlockT, H: ExHashT> Protocol<B, H> {
 				message_queue: None,
 				justification,
 			};
+			// Stop if we don't have requested block body
+			if get_body && block_data.body.is_none() {
+				trace!(target: "sync", "Missing data for block request. BODY: {}", block_data.body.is_some());
```
bkchr:
```suggestion
				trace!(target: "sync", "Missing data for block request.");
```

This will only be printed when `body.is_none()`.

client/network/src/protocol.rs
```diff
@@ -784,7 +793,7 @@ impl<B: BlockT, H: ExHashT> Protocol<B, H> {
 		request: message::BlockRequest<B>,
 		response: message::BlockResponse<B>,
 	) -> CustomMessageOutcome<B> {
-		let blocks_range = match (
+		let blocks_range = || match (
```
bkchr: Is this change required?

arkpar: No, this is an optimization not to format strings when tracing is disabled.

bkchr: Ahh, didn't github folded the `trace!` call. Makes sense :)

client/network/src/protocol.rs
```diff
@@ -811,6 +820,19 @@ impl<B: BlockT, H: ExHashT> Protocol<B, H> {
...
 			}
 		} else {
+			// Validate fields against the request.
+			if request.fields.contains(message::BlockAttributes::HEADER) && response.blocks.iter().any(|b| b.header.is_none()) {
+				self.peerset_handle.report_peer(peer, rep::BAD_RESPONSE);
+				trace!(target: "sync", "Missing header for a block");
+				return CustomMessageOutcome::None
+			}
+			if request.fields.contains(message::BlockAttributes::BODY) && response.blocks.iter().any(|b| b.body.is_none()) {
+				self.behaviour.disconnect_peer(&peer);
```
andresilva: Why don't we disconnect when missing requested header above?

client/network/test/src/sync.rs
```diff
@@ -660,3 +660,24 @@ fn does_not_sync_announced_old_best_block() {
 	})).unwrap();
 	assert!(!net.peer(1).is_major_syncing());
 }
+
+#[test]
+fn full_sync_requires_block_body() {
+	// Check that we don't sync heders-only in full mode.
```
andresilva:
```suggestion
	// Check that we don't sync headers-only in full mode.
```

client/network/src/protocol.rs
arkpar: Good point

PR #5054, opened 2020-02-25
do not check unleash on every PR, only master and tags

gnunicorn:


gnunicorn:
@s3krit could you confirm I did these dependencies right: this will now only work _if_ the publish draft succeeded (thus preventing us to run on non-signed tags)!?!

s3krit:
`dependencies` is a badly named gitlab-ci directive. It actually refers to what artifacts a particular job depends on. For example, as it's written, your job will depend on some artifacts from `unleash-check` and `publish-draft-release`. Since neither of these jobs actually produce artifacts. I'm not sure what will happen if they fail - it might do what we want.

PR #5057, opened 2020-02-25
Make verbosity level mandatory with telemetry opt

### LINKED ISSUE #4372: CLI: Telemetry URL & Verbosity Flags
As it stands, the `--telemetry-url` flag expects both the URL of the `substrate-telemetry` server & the desired verbosity of logs to that service. This is confusing and fails silently when misconfigured.

For instance, running a node with `--telemetry-url 'ws://telemetry:8000/submit 5'` will appropriately set the verbosity to a level of `5`, but running a node with `--telemetry-url ws://telemetry:8000/submit 5` will set the verbosity to the default level (`0`) and will silently drop the `5`.

Ideally, a new flag like `--telemetry-verbosity` could be introduced here to add clarity around this configuration.

pscott:
- Closes #4372 

As mentioned in #4372 , this PR makes the verbosity level mandatory in the `--telemetry-urls` option, instead of defaulting to 0.

There are a couple of points I would like to clarify:
1. I'm not sure this is considering `breaking` the API? Or should I flag it as `B2-breaksapi`?
2. I don't think this PR changes anything to Polkadot but I'm not sure?
3. I haven't managed to find the place where we check that the `VERBOSITY_LEVEL` is not bigger than the maximum verbosity_level. Maybe there's no point in doing so, but it might be better practice to let the end user know that when he specifies a verbosity-level bigger than 9, it doesn't actually exist?
4. As for verbosity level, I've found that the constant SUBSTRATE_DEBUG is declared here https://github.com/paritytech/substrate/blob/db1ab7d18fbe7876cdea43bbf30f147ddd263f94/client/telemetry/src/lib.rs#L108 but I have not managed to find where it is USED in the code. Is it just here for export purposes?


client/cli/src/commands/runcmd.rs
```diff
@@ -562,13 +563,34 @@ fn interface_str(
...
+        write!(f, "{}", self.details)
+    }
+}
+
+impl Error for TelemetryParsingError {
+    fn description(&self) -> &str {
+        &self.details
+    }
+}
+
```
pscott: Is there a simpler way of having a simple String implement the std::error::Error trait?

pscott:
[clabot:check]

client/cli/src/commands/runcmd.rs
tomusdrw: Why not just `enum TelemetryError { MissingVerbosity }`? And `impl std::error::Error for TelemetryError {}` and `impl Display`?

client/cli/src/commands/runcmd.rs
```diff
@@ -562,13 +563,34 @@ fn interface_str(
...
 fn parse_telemetry_endpoints(s: &str) -> Result<(String, u8), Box<dyn std::error::Error>> {
```
tomaka:
```suggestion
fn parse_telemetry_endpoints(s: &str) -> Result<(String, u8), TelemetryParsingError> {
```

client/cli/src/commands/runcmd.rs
```diff
@@ -562,13 +563,34 @@ fn interface_str(
...
+}
+
+impl fmt::Display for TelemetryParsingError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.details)
+    }
+}
+
+impl Error for TelemetryParsingError {
+    fn description(&self) -> &str {
```
tomusdrw: I don't think that's necessary any more. They only relevant thing in `Error` trait is `fn source` I believe, the rest should be left as-is.

client/cli/src/commands/runcmd.rs
```diff
@@ -562,13 +563,34 @@ fn interface_str(
 	}
 }
 
-/// Default to verbosity level 0, if none is provided.
+/// Utility error for returning custom errors that implement the std::error::Error trait.
+#[derive(Debug)]
+struct TelemetryParsingError {
```
tomaka: This should be a strongly-typed enum. Errors that contain strings are generally not a good coding practice.
```suggestion
enum TelemetryParsingError {
    VerbosityLevelMissing,
}
```

client/cli/src/commands/runcmd.rs
tomaka: I think this method is deprecated

pscott: Cool, thanks for the tips! :D

client/cli/src/commands/runcmd.rs
pscott: I'm missing something here: what should we do with the result return by the `parse()` function ? Should I add something to the `TelemetryParsingError` enum?

tomaka: > Should I add something to the TelemetryParsingError enum?

Yes!

`TelemetryParsingError` by definition is an error that can happen when parsing an address, and should thus contain all the possible problems that can happen.

client/cli/src/commands/runcmd.rs
```diff
@@ -564,24 +564,30 @@ fn interface_str(
 
 #[derive(Debug)]
 enum TelemetryParsingError {
-	MissingVerbosity
+	MissingVerbosity,
+	VerbosityParsingError(std::num::ParseIntError),
+	UrlParsingError(std::convert::Infallible),
```
tomaka:
```suggestion
	UrlParsingError,
```

client/cli/src/commands/runcmd.rs
```diff
@@ -564,24 +564,30 @@ fn interface_str(
...
+fn parse_telemetry_endpoints(s: &str) -> Result<(String, u8), TelemetryParsingError> {
 	let pos = s.find(' ');
 	match pos {
-		None => Err(Box::new(TelemetryParsingError::MissingVerbosity)),
+		None => Err(TelemetryParsingError::MissingVerbosity),
 		Some(pos_) => {
-			let verbosity = s[pos_ + 1..].parse()?;
-			let url = s[..pos_].parse()?;
+			let verbosity = s[pos_ + 1..].parse().map_err(TelemetryParsingError::VerbosityParsingError)?;
+			let url = s[..pos_].parse().map_err(TelemetryParsingError::UrlParsingError)?;
```
tomaka:
```suggestion
			let url = s[..pos_].to_string();
```

I have no idea why we're using `parse()` here.

pscott: Very good point. Maybe the original idea was to trying and parse the string to make sure it's a valid url, and then the code author forgot to change it?

client/cli/src/commands/runcmd.rs
```diff
@@ -172,8 +173,7 @@ pub struct RunCmd {
 	///
 	/// This flag can be passed multiple times as a mean to specify multiple
 	/// telemetry endpoints. Verbosity levels range from 0-9, with 0 denoting
-	/// the least verbosity. If no verbosity level is specified the default is
-	/// 0.
+	/// the least verbosity.
```
bkchr: You should add an example how the expected format looks like.

client/cli/src/commands/runcmd.rs
```diff
@@ -562,16 +562,32 @@ fn interface_str(
...
+impl std::error::Error for TelemetryParsingError {}
+
+impl fmt::Display for TelemetryParsingError {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		match &*self {
+			TelemetryParsingError::MissingVerbosity => write!(f, "Verbosity level missing"),
+			TelemetryParsingError::VerbosityParsingError(e) => write!(f, "{}", e),
+			TelemetryParsingError::UrlParsingError(e) => write!(f, "{}", e),
+		}
+    }
```
bkchr: Spaces

client/cli/src/commands/runcmd.rs
bkchr: Checking that this is a valid url, sounds like a good idea. However, this will require more than just calling `parse()` here :) You will need a crate a provides support for url parsing/checking.

pscott: Indeed! I was thinking about using the [url](https://docs.rs/url/2.1.1/url/) crate, but I'm not sure what are the practices regarding adding crates... ?

bkchr: Maybe libp2p re-exports something you can use. @tomaka ?

tomaka: This can be either a libp2p `Multiaddr` or a URL.
We should add in libp2p-telemetry a function to parse this and returns a strongly-typed struct.

pscott:
Currently waiting on #5069 to be merged, and then will be able to address comments and get ready for review. :D

Issue #4056, opened 2019-11-08
Introduce on_runtime_upgrade similarly to on_initialize but executed before.


thiolliere:
With the recent introduction of [StorageValue::translate](https://substrate.dev/rustdocs/master/srml_support/storage/trait.StorageValue.html#tymethod.translate) a new pattern appeared in on_initialize:

```rust
fn on_initialize() {
    if new_runtime.get() == true {
        MyStorageValue::translate(...);
        new_runtime.put(false);
    }
}
```

The error proneness is that on_initialize are executed in the order declaration in construct_runtime. Thus if one upper module in on_initialize call your module and use this storage, then the value is still the old encoded one. And thus can leads to very error.

To solve this we can introduce a new function: on_runtime_upgrade_initialize which will be call before all on_initialize.
With this the code to check is just the code contained in on_runtime_upgrade_initialize of other module which logic is easier to audit.

cc @shawntabrizi

thiolliere:
note: even if we execute on_runtime_upgrade just before on_initialize of next block we still have some logic executing in between for validating transaction:

such logic is regular validate transaction but also the per-pallet defined signed extensions, this signed extensions logic could fail.

Could be interesting to know what happens if this validate_transaction from transactions pool panics.

thiolliere:
note: the staking module has update its storage one time, you can see an example of a migration code in frame/staking/src/migration.rs

This kind of transition work for staking but in case of SignedExtensions this has one caveat the validation of transaction in the queue could panic, thus probably having them removed from the queue.

shawntabrizi:
Expanding on this idea, it is very possible that post-runtime-upgrade functions may want to execute logic which would take more than a single block.

In magical dream land, `on_runtime_upgrade` would be special such that it could handle complex logic which could take multiple blocks, and Substrate would be able to correctly handle this.

For example: Let's imagine we want to update the balances module of a live change such that the balance type is u256 rather than u128 or something like that.

With thousands, maybe millions of accounts, a storage upgrade (https://github.com/paritytech/substrate/pull/4555) would not complete in just one block.

But, if a runtime developer could simply write in the `on_runtime_upgrade`: "Upgrade all the balances", and magically, it should be able to batch the storage migration over multiple blocks.

As long as there is still some `on_runtime_upgrade` logic to be run, the blockchain will not accept any other extrinsics or runtime tasks, but the blockchain will not halt or stop producing blocks.

thiolliere:
[...]

PR #5058, opened 2020-02-25
Introduce `on_runtime_upgrade`

shawntabrizi:
This is a basic implementation of the `on_runtime_upgrade` which triggers at the very beginning of block initialization, before `system::initialize` or `module::on_initialize`.

Introduces `OnRuntimeUpgrade` trait which exposes `on_runtime_upgrade` function which is triggered in the executor.

There is now an `on_runtime_upgrade()` function which can be defined in the `decl_module` macro.

We have also introduced a new storage item in FRAME System, `RuntimeUpgraded` which is a bool that expresses if a runtime upgrade occurred last block.

In the executive module, we read this storage item and execute the `on_runtime_upgrade` logic if needed.

Balances and Staking pallets have a new storage value `StorageVersion` which keeps track of a `Releases` enum, which will express the semantic version of the runtime storage, and should match 1:1 with the crate version.

Closes https://github.com/paritytech/substrate/issues/4056

I have also removed traces of `on_finalise` and `on_initialise` as they have been deprecated since v1.0.


drewstone:
Is this still planned?

shawntabrizi:
@drewstone Would like to, just need to prioritize 2.0 and avoid extra noise.

frame/executive/src/lib.rs
```diff
@@ -176,13 +178,21 @@ where
 		extrinsics_root: &System::Hash,
 		digest: &Digest<System::Hash>,
 	) {
+		<AllModules as OnRuntimeUpgrade>::on_runtime_upgrade();
```
thiolliere: Hmm actually in my branch I moved this to under system::initialize, so order would be:

system::initialize
all_modules::on_runtime_upgrade
all_modules::on_initialize

shawntabrizi: `system::initialize` does a bunch of stuff with the storage of the System module:

```
		// populate environment
		storage::unhashed::put(well_known_keys::EXTRINSIC_INDEX, &0u32);
		<Number<T>>::put(number);
		<Digest<T>>::put(digest);
		<ParentHash<T>>::put(parent_hash);
		<BlockHash<T>>::insert(*number - One::one(), parent_hash);
		<ExtrinsicsRoot<T>>::put(txs_root);

		if let InitKind::Full = kind {
			<Events<T>>::kill();
			EventCount::kill();
			<EventTopics<T>>::remove_all();
		}
```

From a principal standpoint, it seems we should do all the `on_runtime_upgrade()` stuff before any module (even the system) does storage things. For example, if we want to use `on_runtime_upgrade()` on the system module itself, right?

thiolliere: yes ok, I was thinking `system::initialize` is a bit different from `system::on_initialize` and might be needed for a module to behave correctly even in on_runtime_upgrade. And also in case we wanted system to upgrade we could directly call this upgrade from executive as it is frame-system.

But indeed I think I'm ok with having on_runtime_upgrade prior to system::initialize.

shawntabrizi: There actually is no `system:on_initialize`. We hardcode this `system::initialize. I don't know why, but my guess is so that we can guarantee that our stuff runs first...?

bkchr:
> In case a runtime wants to switch a module to another one or do big upgrade not self-contained, I think the logic should be written in node/runtime/src/lib.rs with an additional function. This function could call modules specific migration function for example:
> 

While this sounds reasonable.

> ```rust
> fn runtime_upgrade() {
> Balance::upgrade_from_other_currency_module(OldBalance);
> // or
> Balance::upgrade_path_for_kusama_1000();
> }
> ```
> 
> And pallet_balance could expose this function under some cargo features,

This sounds really error prone. Only you as a module developer know if you need to upgrade the storage. A runtime developer should not be required to call these functions manually IMHO.

frame/system/src/lib.rs
```diff
@@ -367,6 +367,9 @@ decl_storage! {
 		/// the `EventIndex` then in case if the topic has the same contents on the next block
 		/// no notification will be triggered thus the event might be lost.
 		EventTopics get(fn event_topics): map hasher(blake2_256) T::Hash => Vec<(T::BlockNumber, EventIndex)>;
+
+		/// Store the block number of last runtime upgrade.
+		LastRuntimeUpgrade get(fn last_runtime_upgrade): Option<T::BlockNumber>;
```
bkchr: Why isn't that just a `bool`?

We could just call `take()` on the value.

frame/balances/src/migration.rs
```diff
@@ -0,0 +1,87 @@
+use super::*;
+
+// Upgrade from the pre-#4649 balances/vesting into the new balances.
+pub fn on_runtime_upgrade<T: Trait<I>, I: Instance>() {
+	let current_version = ModuleVersion::<I>::get();
+
+	if current_version == 1 {
+		return
+	} else if current_version == 0 {
```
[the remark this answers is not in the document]
shawntabrizi: ?

IsUpgraded is still in staking. I could transition it to this upgrade pattern if needed.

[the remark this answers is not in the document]
bkchr: Sorry, I mean the storage item. You just removed it from `decl_storage`, but it will still be in the state. So you need to remove it from storage.

frame/executive/src/lib.rs
thiolliere: I don't know how much it is allowed but construct_runtime seems to accept having system not as first module indeed.

shawntabrizi: IMO should force System to be first :)

shawntabrizi: or we should remove it from the all modules list, and have it be hand-coded into the macros

frame/system/src/lib.rs
```diff
@@ -486,6 +489,7 @@ decl_module! {
 			}
 
 			storage::unhashed::put_raw(well_known_keys::CODE, &code);
+			RuntimeUpgraded::put(true);
```
gavofyork: technically this isn't the only way a runtime could be upgraded: any module can call `unhashed::set_storage(b":CODE", ...);` in order to change it. also, `set_storage(b":CODE", ...)` could be dispatched, which will also short-circuit an upgrade. i think allowing the first one to break things is fine (though should be included in the docs), but the second one should be prevented from setting the `:CODE` entry (or, if it does set it, then you should call `RuntimeUpgraded::put(true);`).

gavofyork:
just one thing to note: modules will not be keeping migration code in there indefinitely. indeed, if we could remove it immediately after use i would, however that's impractical.

nonetheless we will realistically support only one upgrade at a time - a module is either at the current version (the genesis config should default the version number to the current in the code or `IsUpgraded` flag to true) or it isn't (either because the version number is older or the `IsUpgraded` flag isn't there). if it's not yet upgraded, then we can so a sanity check on the version number (it should be exactly one less than the current in the code), if we're using version numbers (i kind of think an `IsUpgraded` flag might be cleaner, but don't really care).

once it's upgraded, then the flag or number can be changed to indicate it is up to date.

note that module-level upgrading won't always be enough anyway, as the recent balances refactoring shows, sometimes you need chain-specific upgrade strategies.

gavofyork:
> This sounds really error prone. Only you as a module developer know if you need to upgrade the storage. A runtime developer should not be required to call these functions manually IMHO.

error prone it might be, but in some cases (upgraded involving only certain module configurations or where modules need to be merged or split) there's nothing else that can be done.

frame/support/src/dispatch.rs
```diff
@@ -196,6 +196,10 @@ impl<T> Parameter for T where T: Codec + EncodeLike + Clone + Eq + fmt::Debug {}
 ///
 /// The following reserved functions also take the block number (with type `T::BlockNumber`) as an optional input:
 ///
+/// * `on_runtime_upgrade`: Executes at the beginning of a block prior to on_initialize when there
+/// is a runtime upgrade. This allow each module to upgrade their storage before the storage items are used.
+/// As such, **calling other module must be avoided**!! Using this function will implement the
+/// [`OnRuntimeUpgrade`](../sp_runtime/traits/trait.OnRuntimeUpgrade.html) trait.
```
apopiak: typos:
"allow" -> "allows"
"their" -> "its"
"calling other module" -> "calling other modules"

frame/support/test/tests/reserved_keyword/on_initialize.stderr
```diff
@@ -1,47 +1,39 @@
 error: Invalid call fn name: `on_finalize`, name is reserved and doesn't match expected signature, please refer to `decl_module!` documentation to see the appropriate usage, or rename it to an unreserved keyword.
   --> $DIR/on_initialize.rs:30:1
    |
-30 | reserved!(on_finalize on_initialize on_finalise on_initialise offchain_worker deposit_event);
```
apopiak: Is this removal of reserved functions intentional?

thiolliere: yes he wrote:

> I have also removed traces of on_finalise and on_initialise as they have been deprecated since v1.0.

frame/system/src/lib.rs
shawntabrizi: Good catch. I have added a check which sets the value to true and also a test.

frame/balances/src/migration.rs
```diff
@@ -0,0 +1,87 @@
...
+		ModuleVersion::<I>::put(1);
+		sp_runtime::print("Upgrading account balances...");
+		// First, migrate from old FreeBalance to new Account.
+		// We also move all locks across since only accounts with FreeBalance values have locks.
+		// FreeBalance: map T::AccountId => T::Balance
+		for (hash, free) in StorageIterator::<T::Balance>::new(b"Balances", b"FreeBalance").drain() {
+			let mut account = AccountData { free, ..Default::default() };
+			// Locks: map T::AccountId => Vec<BalanceLock>
+			let old_locks = get_storage_value::<Vec<OldBalanceLock<T::Balance, T::BlockNumber>>>(b"Balances", b"Locks", &hash);
```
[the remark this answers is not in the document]
shawntabrizi: it seems we have not

shawntabrizi:
Should be ready to review with the feedback given

frame/balances/src/lib.rs
```diff
@@ -347,6 +348,21 @@ impl<Balance: Saturating + Copy + Ord> AccountData<Balance> {
...
+// storage migration logic. This should match directly with the semantic versions of the Rust crate.
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, RuntimeDebug)]
+enum Releases {
+	V1_0_0,
+	V2_0_0,
+}
+
+impl Default for Releases {
+	fn default() -> Self {
+		Releases::V1_0_0
```
gavofyork: Shouldn't it default to the most recent version?

frame/balances/src/migration.rs
```diff
@@ -0,0 +1,83 @@
+use super::*;
+
+// Upgrade from the pre-#4649 balances/vesting into the new balances.
+pub fn on_runtime_upgrade<T: Trait<I>, I: Instance>() {
+	if StorageVersion::<I>::get() == Releases::V2_0_0 { return };
+
+	sp_runtime::print("Upgrading account balances...");
```
gavofyork: should really be an exhaustive match on `StorageVersion::<I>::get()`.

frame/balances/src/migration.rs
```diff
@@ -0,0 +1,88 @@
+use super::*;
+
+// Upgrade from the pre-#4649 balances/vesting into the new balances.
```
kianenigma:
```suggestion
```

frame/balances/src/migration.rs
```diff
@@ -0,0 +1,88 @@
...
+pub fn on_runtime_upgrade<T: Trait<I>, I: Instance>() {
+	match StorageVersion::<I>::get() {
+		Releases::V2_0_0 => return,
+		Releases::V1_0_0 => upgrade_v1_to_v2::<T, I>(),
+	}
+}
+
+fn upgrade_v1_to_v2<T: Trait<I>, I: Instance>() {
```
kianenigma:
```suggestion
// Upgrade from the pre-#4649 balances/vesting into the new balances.
fn upgrade_v1_to_v2<T: Trait<I>, I: Instance>() {
```

frame/staking/src/lib.rs
```diff
@@ -679,6 +680,21 @@ impl Default for Forcing {
 	fn default() -> Self { Forcing::NotForcing }
 }
 
+// A value placed in storage that represents the current version of the Balances storage.
```
kianenigma:
```suggestion
// A value placed in storage that represents the current version of the Staking storage.
```

frame/staking/src/lib.rs
```diff
@@ -679,6 +680,21 @@ impl Default for Forcing {
...
+// This value is used by the `on_runtime_upgrade` logic to determine whether we run
+// storage migration logic. This should match directly with the semantic versions of the Rust crate.
+#[derive(Encode, Decode, Clone, Copy, PartialEq, Eq, RuntimeDebug)]
+enum Releases {
```
kianenigma: Seeing this duplicated struct everywhere and with the problem of each module naming it whatever it wants, and as we had here confused docs, maybe it is a good idea to make it 

```
// support
trait RuntimeStorageVersion<V>: Parameter {
  fn get_version() -> V
}
```

and each module can implement whatever it wants for this

frame/staking/src/migration/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+/// Update storage from v1.0.0 to v2.0.0
+///
+/// In old version the staking module has several issue about handling session delay, the
+/// current era was always considered the active one.
+///
+/// After the migration the current era will still be considered the active one for the era of
+/// the upgrade. And the delay issue will be fixed when planning the next era.
```
kianenigma:
```suggestion
//! Update storage from v1.0.0 to v2.0.0
//!
//! In old version the staking module has several issue about handling session delay, the
//! current era was always considered the active one.
//!
//! After the migration the current era will still be considered the active one for the era of
//! the upgrade. And the delay issue will be fixed when planning the next era.
```

frame/staking/src/lib.rs
shawntabrizi: Let's keep this in mind, and see how much this is used and if it is really needed.

shawntabrizi:
@gnunicorn gucci gang

Issue #5047, opened 2020-02-25
Light client cache runtime code


bkchr:
Currently the runtime code is part of the `StorageProof` of the `execution_proof`. From a performance perspective, caching the runtime code on the light client would be a much better idea.

The rough idea would be the following:
1. Introduce a new digest `RuntimeCodeChanged`. This digest should be added to the header when the runtime code changes. 
2. Make light client listen for `RuntimeCodeChanged` digests.
3. Request runtime code in light client after `RuntimeCodeChanged` was detected and cache it locally.

CC @svyatonik

cheme:
Naïve question, why not storing a hash of `:code` in the state instead?

svyatonik:
> Naïve question, why not storing a hash of :code in the state instead?

Because light client has no local access to the state. It would need to fetch proof of read of that `:code_hash` at every block. Digest is available from headers => we'll know when we really need to fetch new `:code`. It is like polling vs subscription.

cheme:
Feels like `:code` is not needed anymore in the state since we got it from the headers (at least when managing a cht). Yet keeping the hash of the code in the state is probably harmless. I guess having a separate storage for runtimes is another issue, but maybe the light client specific runtime management could be generalized to full client for possible future changes.

svyatonik:
> Feels like :code is not needed anymore in the state since we got it from the headers (at least when managing a cht).

I may be missing something from your comment, but the idea was never to include the whole `:code` in the header. Just the signal that the code has been changed and we need to fetch the new version. I do not think that including the code itself is a good idea (given its size).

cheme:
yes I was just saying that we can fetch runtime hash from the headers (given on genesis we include a initial change), so it is not strictly needed in state as long as client got the cht.

I was not proposing to put those big blobs in header :) , quite the opposite (removing them from the state could be good but then we need to store them somewhere and a way to share them: so if there is a dedicated storage used by light, it could probably be use by full client to, but it needs a specific rpc to query it afterward).

Ok, I misread the description, I was thinking the `RuntimeCodeChange` digest would include a hash of the new runtime, but it does not have to. My comment only make sense if it does.

svyatonik:
OK, I got your idea now, thanks :) Yes - `hash(code)` should probably be a part of that digest. But what you're talking about is quite a significant change that, imo, deserves separate discussion/issue :)

seunlanlege:
interested in light client mechanics if anyone's willing to mentor this issue :eyes:

[...]

PR #5060, opened 2020-02-25
Don't include `:code` by default in storage proofs

bkchr:
This pr changes the behavior of automatically including `:code` in storage proofs. Before this pr, the `:code` was extracted by the wasm-executor before executing a function in the runtime. Now, the runtime code is passed as a parameter from the outside. This makes it possible to extract the `:code` from the backend, before the storage recorder tracks all accesses to the backend (which resulted in inclusion of `:code` in the proof).

However, the `execution_proof` function of `Client` still adds the `:code` to the proof, this is required by the light client, until https://github.com/paritytech/substrate/issues/5047 is implemented.

Closes: https://github.com/paritytech/cumulus/issues/8


client/src/client.rs
```diff
@@ -427,9 +427,19 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 		method: &str,
 		call_data: &[u8]
 	) -> sp_blockchain::Result<(Vec<u8>, StorageProof)> {
+		// Make sure we include the `:code` and `:heap_pages` in the execution proof to be
+		// backwards compatible.
+		//
+		// TODO: Remove when solved: https://github.com/paritytech/substrate/issues/5047
+		let code_proof = self.read_proof(
```
bkchr: @svyatonik here I'm currently forcing that the `:code` is still added to the proof.

client/block-builder/src/lib.rs
```diff
@@ -221,3 +221,35 @@ where
...
+
+		let backend = sp_state_machine::create_proof_check_backend::<Blake2Hasher>(
+			block.storage_changes.transaction_storage_root,
+			proof,
+		).unwrap();
+
+		assert!(
+			backend.storage(&sp_core::storage::well_known_keys::CODE)
+				.unwrap_err()
+				.contains("Database missing expected key"),
```
svyatonik: Probably just leave `.unwrap_err()` - comparing strings seems a bit fragile?

bkchr: Actually I wanted to match on the error, however it is a string already. The full error message is much longer, but this part is the important one (as long as no one changes the display impl for the error, it should be fine).

client/executor/src/wasm_runtime.rs
```diff
@@ -86,17 +86,18 @@ impl RuntimesCache {
 	///
 	/// # Parameters
 	///
-	/// `ext` - Externalities to use for the runtime. This is used for setting
-	/// up an initial runtime instance.
+	/// `ext` - Externalities to use for the runtime.
```
pepyakin: I think we need `ext` for executing `Core_version` call here...

```suggestion
	/// `ext` - Externalities to use for the getting the runtime's version call.
```

primitives/state-machine/src/backend.rs
```diff
@@ -359,3 +359,22 @@ pub(crate) fn insert_into_memory_db<H, I>(mdb: &mut MemoryDB<H>, input: I) -> Op
...
 }
+
+/// Get the runtime code from the given `backend`.
+///
+/// Returns an error if the `:code` could not be found.
+pub fn get_runtime_code<H: Hasher, B: Backend<H>>(backend: &B) -> Result<RuntimeCode, &'static str>
+	where H::Out: Encode,
+{
+	let code = backend.storage(well_known_keys::CODE)
+			.ok()
```
pepyakin: Isn't this overindented?

primitives/state-machine/src/backend.rs
```diff
@@ -359,3 +359,22 @@ pub(crate) fn insert_into_memory_db<H, I>(mdb: &mut MemoryDB<H>, input: I) -> Op
...
+			.flatten()
+			.ok_or_else(|| "`:code` not found")?;
```
pepyakin: nit: `.ok_or` can work here as well

```suggestion
			.ok_or("`:code` not found")?;
```

primitives/core/src/traits.rs
```diff
@@ -98,13 +98,78 @@ pub trait CodeExecutor: Sized + Send + Sync + CallInWasm + Clone + 'static {
...
+	}
+}
+
+/// Could not find the `:code` in the externalities while initializing the [`RuntimeCode`].
+#[derive(Debug)]
+pub struct CodeNotFound;
+
+impl std::fmt::Display for CodeNotFound {
+	fn fmt(&self, f: &mut std::fmt::Formatter) -> Result<(), std::fmt::Error> {
+		write!(f, "CodeNotFound")
```
pepyakin: This text would more suit a `Debug` implementation. `Display` could use a more human readable string.

```suggestion
		write!(f, "the storage entry `:code` doesn't have any code")
```

arkpar:
So now to make a call one is required to query code from storage and keep it in memory to avoid re-querying it each time? This just wastes memory. Why not just add filter to the proof recorder?

arkpar:
> This pr changes the behavior of automatically including :code in storage proofs. Before this pr, the :code was extracted by the wasm-executor before executing a function in the runtime. 

That's not true. It was only extracted if the hash does not match the available instance. When there's matching version of compiled code or wasmi instance we don't need to keep original WASM bytecode in memory.

bkchr:
> > This pr changes the behavior of automatically including :code in storage proofs. Before this pr, the :code was extracted by the wasm-executor before executing a function in the runtime.
> 
> That's not true. It was only extracted if the hash does not match the available instance. When there's matching version of compiled code or wasmi instance we don't need to keep original WASM bytecode in memory.

Yeah okay, after checking again the code, I found the code that caches the `storage_hash`. I thought that this wasn't done. However, given this code, the `:code` is also cached. 

I don't think that this wastes any more memory than the implementation before. We don't cache the code anywhere else as before.

> Why not just add filter to the proof recorder?

This doesn't work, as you also can have legit accesses from the runtime to `:code` and that requires that the proof recorder records them. From my point of view neither the state machine nor the executor should care where the runtime code comes from. It is just instructed to execute the given code and should not be responsible for fetching it from the backend.

arkpar:
> This doesn't work, as you also can have legit accesses from the runtime to :code and that requires that the proof recorder records them. 

Does it? Whoever verifies the proof has the code anyway and they can just return it when it is queried from the proof storage. Is that right @svyatonik?

> From my point of view neither the state machine nor the executor should care where the runtime code comes from. It is just instructed to execute the given code and should not be responsible for fetching it from the backend.

Needlessly fetching the code on each call is expensive. The executor decides when it actually needs the code and should use either externalities or some other interface to fetch it **only when needed**

svyatonik:
> Does it? Whoever verifies the proof has the code anyway and they can just return it when it is queried from the proof storage. Is that right @svyatonik?

No - there's no code on light client and that's why tthe ":code" was a part of execution proof since very beginning. There's an issue to add code caching #5047 , mentioned in the PR description.

arkpar:
> No - there's no code on light client and that's why tthe ":code" was a part of execution proof since very beginning. There's an issue to add code caching #5047 , mentioned in the PR description.

I'm asking about parachain validation here. The proof is useless without the code. So if it is excluded from the proof whoever is verifying it must have the code from some other source already. That also means when you have "legit accesses from the runtime to :code" they can just use the code they already have and there's no need to include it in the proof.  Is this technically correct?

bkchr:
For parachain validation, the runtime code is part of the validation blob that is stored on the relay chain. So, we don't need `:code` to be part of the proof.

PR #5063, opened 2020-02-26
removes use of sc_client::Client from sc-rpc

seunlanlege:
see #4452


bin/node/cli/src/service.rs
```diff
@@ -394,7 +395,7 @@ mod tests {
 	use sc_service::AbstractService;
 	use crate::service::{new_full, new_light};
 	use sp_runtime::traits::IdentifyAccount;
-
+gst
```
drewstone: should be removed

seunlanlege:
same failing ui test as #5030

client/api/src/proof_provider.rs
```diff
@@ -0,0 +1,67 @@
...
+	generic::BlockId,
+	traits::{Block as BlockT},
+};
+use crate::{StorageProof, ChangesProof};
+use sp_storage::{ChildInfo, StorageKey};
+
+/// Interface for providing block proving utilities.
+pub trait ProofProvider<Block: BlockT> {
+	/// Reads storage value at a given block + key, returning read proof.
+	fn read_proof(&self, id: &BlockId<Block>, keys: &[Vec<u8>]) -> sp_blockchain::Result<StorageProof>;
```
arkpar: It's unfortunate that we'll have to do temporary allocations for `keys` here. Maybe we could declare it like this?
```
fn read_proof(&self, id: &BlockId<Block>, keys: &mut dyn Iterator<Item=&[u8]>)
```
and then use `map` at call site to get `&[u8]` where needed?

seunlanlege: So I took a look at it and unfortunately, the `read_proof` is being used in contexts where a mutable reference to `keys` isn't available.

arkpar: This is a reference to iterator, not to keys which has to be mutable because `next` method for an iterator takes `&mut self`

arkpar:
```
let keys = [[1,2], [3, 4]];
x.read_proof(id, &mut key.iter().map(AsRef::as_ref)))
```
This should work even though the data itself is not mutable

seunlanlege: > This is a reference to iterator, not to keys

this is an example.

https://github.com/paritytech/substrate/blob/b5ec7d412d27feafad79a7c94b6f6451f0efc04f/client/network/src/protocol/light_client_handler.rs#L475

bin/node-template/node/src/service.rs
```diff
@@ -42,7 +44,7 @@ macro_rules! new_full_start {
 					.ok_or_else(|| sc_service::Error::SelectChainRequired)?;
 
 				let (grandpa_block_import, grandpa_link) =
-					grandpa::block_import(client.clone(), &*client, select_chain)?;
+					grandpa::block_import(client.clone(), &(client.clone() as Arc<_>), select_chain)?;
```
tomusdrw:
```suggestion
					grandpa::block_import(client.clone(), &(client.clone() as _), select_chain)?;
```

Have you tried this?

bin/node-template/node/src/service.rs
```diff
@@ -87,9 +89,11 @@ pub fn new_full(config: Configuration<GenesisConfig>)
 			.expect("Link Half and Block Import are present for Full Services or setup failed before. qed");
 
 	let service = builder
-		.with_finality_proof_provider(|client, backend|
-			Ok(Arc::new(GrandpaFinalityProofProvider::new(backend, client)) as _)
-		)?
+		.with_finality_proof_provider(|client, backend| {
+			// GenesisAuthoritySetProvider is implemented for StorageAndProofProvider
+			let provider = client as Arc<dyn StorageAndProofProvider<_, _>>;
+			Ok(Arc::new(GrandpaFinalityProofProvider::new(backend, provider)) as _)
```
tomusdrw:
```suggestion
			Ok(Arc::new(GrandpaFinalityProofProvider::new(backend, client.clone() as _)) as _)
```
This should probably work fine as well (without the upper line).

client/api/src/call_executor.rs
```diff
@@ -29,6 +29,16 @@ use sp_externalities::Extensions;
 use sp_core::NativeOrEncoded;
 
 use sp_api::{ProofRecorder, InitializeBlock, StorageTransactionCache};
+use crate::execution_extensions::ExecutionExtensions;
+
+/// Executor Provider
+pub trait ExecutorProvider<Block: BlockT, E: CallExecutor<Block>> {
+	/// Get call executor reference.
+	fn executor(&self) -> &E;
```
tomusdrw: Having this makes the bounds more complicated in so many places, in what context is it being used?
Maybe it's worth to consider returning a `dyn` here or making `CallExecutor` and associated type? Basically I'm curious if less code complication is worth the dynamic call - maybe we just use it once in RPC or sth.

client/network/src/protocol.rs
```diff
@@ -1473,7 +1473,10 @@ impl<B: BlockT, H: ExHashT> Protocol<B, H> {
 
 		trace!(target: "sync", "Remote read request {} from {} ({} at {})",
 			request.id, who, keys_str(), request.block);
-		let proof = match self.context_data.chain.read_proof(&request.block, &request.keys) {
+		let proof = match self.context_data.chain.read_proof(
+			&request.block,
+			&mut request.keys.iter().map(AsRef::as_ref)
```
seunlanlege: ah i see, a mutable reference to the iterator. Yeah, my brain couldn't make sense of it at the time.

bin/node-template/node/src/service.rs
seunlanlege: yup, here's the error you get.

```
  ::: bin/node-template/node/src/command.rs:35:32
   |
35 |                 |config: _| Ok(new_full_start!(config).0),
   |                                ----------------------- in this macro invocation

error[E0620]: cast to unsized type: `alloc::sync::Arc<sc_client::client::Client<sc_client_db::Backend<sp_runtime::generic::block::Block<sp_runtime::generic::header::Header<u32, sp_runtime::traits::BlakeTwo256>, sp_runtime::OpaqueExtrinsic>>, sc_client::call_executor::LocalCallExecutor<sc_client_db::Backend<sp_runtime::generic::block::Block<sp_runtime::generic::header::Header<u32, sp_runtime::traits::BlakeTwo256>, sp_runtime::OpaqueExtrinsic>>, sc_executor::native_executor::NativeExecutor<service::Executor>>, sp_runtime::generic::block::Block<sp_runtime::generic::header::Header<u32, sp_runtime::traits::BlakeTwo256>, sp_runtime::OpaqueExtrinsic>, node_template_runtime::RuntimeApi>>` as `dyn sc_finality_grandpa::GenesisAuthoritySetProvider<sp_runtime::generic::block::Block<sp_runtime::generic::header::Header<u32, sp_runtime::traits::BlakeTwo256>, sp_runtime::OpaqueExtrinsic>>`
  --> bin/node-template/node/src/service.rs:47:45
   |
47 |                     grandpa::block_import(client.clone(), &(client.clone() as _), select_chain)?;
   |                                                            ^^^^^^^^^^^^^^^^^^^^^
...
```

bin/node-template/node/src/service.rs
seunlanlege: the compiler cannot infer the type, needs the cast to `StorageAndProofProvider` -> `GenesisAuthoritySetProvider.`

client/api/src/call_executor.rs
seunlanlege: fixed this.

seunlanlege:
going to mark this as looks good if there are no objections :eyes:

client/finality-grandpa/src/lib.rs
```diff
@@ -56,12 +56,8 @@ use futures::prelude::*;
 use futures::StreamExt;
 use log::{debug, info};
 use futures::channel::mpsc;
-use sc_client_api::{
-	LockImportRun, BlockchainEvents, CallExecutor,
-	backend::{AuxStore, Backend}, ExecutionStrategy, Finalizer, TransactionFor,
-};
+use sc_client_api::{LockImportRun, BlockchainEvents, CallExecutor, backend::{AuxStore, Backend}, ExecutionStrategy, Finalizer, TransactionFor, ExecutorProvider};
```
tomusdrw: Looks like a pretty long line

gnunicorn:
sorry, @seunlanlege the merge order resulted in another conflict.

seunlanlege:
@gnunicorn fixed.

PR #5064, opened 2020-02-26
Fix substrate default storage location

chevdor:
This PR contains the following:
- same fix than in #5059 
- fix substrate default chain location (old issue finally fixed)
- add prometheus port to the list of hinted exposed ports

So this PR means we no longer *have to* provide the `-d /data` to substrate to make the volume mounting works.


chevdor:
I see #5059 has been merged. Let me rebase so the change set is a little smaller/cleaner.

chevdor:
Done, now the changes brought by #5059 are no longer part of this PR.

chevdor:
I don't think the failing test is coming from the PR. Can we manually trigger a re-run of the tests?

PR #5065, opened 2020-02-26
Fix revalidation not revalidating multiple times

NikVolf:
The problem was that revalidation queue wasn't accepting same transaction twice for revalidation, and transaction that were valid for some time, then became invalid, were stuck and not revalidated.

### DIFF
--- a/client/transaction-pool/src/revalidation.rs
+++ b/client/transaction-pool/src/revalidation.rs
@@ -113,7 +113,9 @@ async fn batch_revalidate<Api: ChainApi>(
 	}
 
 	pool.validated_pool().remove_invalid(&invalid_hashes);
-	pool.resubmit(revalidated);
+	if revalidated.len() > 0 {
+		pool.resubmit(revalidated);
+	}
 }
 
 impl<Api: ChainApi> RevalidationWorker<Api> {
@@ -149,6 +151,7 @@ impl<Api: ChainApi> RevalidationWorker<Api> {
 				} else {
 					for xt in &to_queue {
 						extrinsics.remove(xt);
+						self.members.remove(xt);
 					}
 				}
 				left -= to_queue.len();
@@ -163,6 +166,10 @@ impl<Api: ChainApi> RevalidationWorker<Api> {
 		queued_exts
 	}
 
+	fn len(&self) -> usize {
+		self.block_ordered.iter().map(|b| b.1.len()).sum()
+	}
+
 	fn push(&mut self, worker_payload: WorkerPayload<Api>) {
 		// we don't add something that already scheduled for revalidation
 		let transactions = worker_payload.transactions;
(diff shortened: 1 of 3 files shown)


client/transaction-pool/src/revalidation.rs
```diff
@@ -34,10 +34,15 @@ pub const BACKGROUND_REVALIDATION_INTERVAL: Duration = Duration::from_millis(5);
 
 const BACKGROUND_REVALIDATION_BATCH_SIZE: usize = 20;
 
-/// Payload from queue to worker.
-struct WorkerPayload<Api: ChainApi> {
-	at: NumberFor<Api>,
-	transactions: Vec<ExHash<Api>>,
+/// Background worker notification.
+enum WorkerNotify<Api: ChainApi> {
+	NewPayload {
```
tomusdrw:
```suggestion
	AddForRevalidation {
```

client/transaction-pool/src/revalidation.rs
```diff
@@ -34,10 +34,15 @@ pub const BACKGROUND_REVALIDATION_INTERVAL: Duration = Duration::from_millis(5);
...
+		at: NumberFor<Api>,
+		transactions: Vec<ExHash<Api>>,
+	},
+	Remove {
```
tomusdrw:
```suggestion
	RemoveFromRevalidation {
```
?

client/transaction-pool/src/revalidation.rs
```diff
@@ -275,6 +294,16 @@ where
...
 		}
 	}
+
+	/// Notify that some transactinos are no longer required to be revalidated.
+	pub fn notify_pruned(&self, transactions: Vec<ExHash<Api>>) {
+		if let Some(ref to_worker) = self.background {
+			if let Err(e) = to_worker.unbounded_send(WorkerNotify::Remove { transactions }) {
+				log::warn!(target: "txpool", "Failed to update background worker: {:?}", e);
+			}
+			return;
```
tomusdrw:
```suggestion
```

client/transaction-pool/src/revalidation.rs
NikVolf: It does only revalidation, so it is redundant to me

client/transaction-pool/src/revalidation.rs
NikVolf: Same, redundant, this module is dedicated only to revalidation

client/transaction-pool/src/revalidation.rs
tomusdrw: Okay, so let's just do `New/Remove` instead of `NewPayload/Remove`.

NikVolf:
I think I will run Kusama for a while with this

NikVolf:
Ok, ran it for a while but not sure if it fixed anything

But it still a good thing to have and also improved logging

client/transaction-pool/src/revalidation.rs
NikVolf: Changed to Add/Remove

client/transaction-pool/src/revalidation.rs
```diff
@@ -149,6 +158,7 @@ impl<Api: ChainApi> RevalidationWorker<Api> {
 				} else {
 					for xt in &to_queue {
 						extrinsics.remove(xt);
+						self.members.remove(xt);
```
NikVolf: Seems that only this required for a fix it.

The rest is defensive programming and logging :)

Issue #4272, opened 2019-12-02
Local nodes keep disconnecting.


arkpar:
Two nodes running on localhost keep disconnecting from each other. Disconnect is initiated by libp2p and not by sync or reputation change apparently. Logs have no useful information on why the disconnect has happened. 

Node 1 started as:

```
polkadot -d /tmp/polkadot/ -lpeerset=trace,sync=trace,sub-libp2p=trace,libp2p=trace --out-peers 0
```

Node2 started as:

```
polkadot -d /tmp/polkadot2 --reserved-nodes /ip4/127.0.0.1/tcp/30333/p2p/QmYiJFSrLQdvWJLLF64RnWGvk92zSjaN49EP9rtNLDbNEo --reserved-only -l peerset=trace,sub-libp2p=trace,libp2p=trace
```
Nodes stay connected for about 10 seconds before disconnect and reconnect happens.

* Nodes should stay connected.
* Logs should have disconnect reason.

tomaka:
For what it's worth, disabling the discovery mechanism fixes the issue. It's still unclear to me what is happening.

romanb:
What seems to happen is the following:

  1. Node 2 connects to node 1.
  2. Both nodes also perform regular (Kusama?) DHT discovery queries in regular (exponentially increasing) intervals, starting with seconds and capped at 60 seconds.
  3. It appears that local nodes connecting to the (Kusama?) DHT put their local address into the DHT, i.e. `/ip4/127.0.0.1/tcp/30333`. This is obviously necessary for nodes in the same local network to discover and talk to each other through the DHT, but these are also seen by remote nodes.
  4. As a result, Node 2 will discover a node with a different peer ID from Node 1 but with address `/ip4/127.0.0.1/tcp/30333`. It will try to connect to it as part of the DHT lookups.
  5. Thus Node 2 will make another connection attempt to Node 1, thinking it is some other node (i.e. expecting a different peer ID this time). It still has the existing connection to Node 1 for the moment.
  6. Node 1 receives the incoming connection from Node 2, replacing the existing connection (i.e. dropping it) due to the single-connection-per-peer policy. The reason(s) for always taking the new connection over the old, even if a node has the role of listener in both connections (i.e. it is not a "simultaneous connect" scenario) are not entirely clear to me (@tomaka?).
[...]

PR #5066, opened 2020-02-26
Adapt to rust-libp2p#1440.

romanb:
## Overview

These are the current changes necessary for adapting substrate to https://github.com/libp2p/rust-libp2p/pull/1440. As described in the libp2p PR, the underlying changes are primarily in `libp2p-core` and for the first iteration the impact on the `libp2p-swarm` API and thus substrate is relatively minimal since at this point the API of `libp2p-swarm` does not actually permit a `NetworkBehaviour` to explicitly request multiple connections per peer. That will change later. For the moment, realistically, a second connection to the same peer only occurs if two peers connect to each other "at the same time". As a side-effect, existing connections are also no longer closed in favour of new ones, which should implicitly address https://github.com/paritytech/substrate/issues/4272, though I didn't get around to verify that yet.

The approach to the integration of the libp2p changes taken here can be summarised as follows (also in the code comments).

## Details

  1. The enabled/disabled status is the same across all connections, as decided by the peerset manager. In order to make the `GenericProto` behaviour aware of all connection handlers (and thus connections), each handler now explicitly emits an `Init` event as the very first event, requesting initialisation (enable/disable) from the behaviour. This was previously implicit.
  2. `send_packet` and `write_notification` always send all data over the same connection to preserve the ordering provided by the transport, as long as that connection is open. If it closes, a second open connection may take over, if one exists, but that case should be no different than a single connection failing and being re-established in terms of potential reordering and dropped messages. Messages can be received on any connection.
  3. The behaviour reports `GenericProtoOut::CustomProtocolOpen` when the first connection reports `NotifsHandlerOut::Open`.
  4. The behaviour reports `GenericProtoOut::CustomProtocolClosed` when the last connection reports `NotifsHandlerOut::Closed`.

In this way, the number of actual established connections to the peer is an implementation detail of the `GenericProto` behaviour. As mentioned before, in practice and at the time of this writing, there may be at most two connections to a peer and only as a result of simultaneous dialing. However, the implementation accommodates for any number of connections.

## Noteworthy

During intermediate testing with the (by default disabled) integration tests `test_consensus`, `test_sync` and `test_connectivity` it was revealed that when run in release mode these tests were very often failing, with the common symptom that _the last node to start in a round of testing_ would often see no other peers (i.e. empty DHT routing table) and thus make no progress while all the others keep on running, causing the tests to time out waiting for the problematic peer to reach a certain state. The tests are mainly using `add_reserved_peer` on the network to initialise the topology, however, `add_reserved_peer` ultimately results in a call to `add_known_peer` on the `DiscoveryBehaviour` which did not actually add that address to the Kademlia routing table, though it adds it to the `user_defined` peers which, when passed in the constructor of the behaviour, _are_ added to the Kademlia routing table. I thus changed `add_known_peer` to also add the given address to the Kademlia routing table and that resolved the issues with these integration tests and the `test_connectivity` test seems to run notably faster (release mode). My current guess is that the tests were so far unknowingly relying on a timing assumption w.r.t. the initial discovery / connection setup and DHT queries in order for all peers to find each other, in particular when simultaneous connections attempts are in play, as often happens in release mode. Ultimately, the change of letting `add_known_peer` add the given address to the Kademlia routing table may be a patch worth extracting separately, because it does look like an oversight to me.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2630,8 +2630,7 @@ dependencies = [
 [[package]]
 name = "libp2p"
 version = "0.16.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bba17ee9cac4bb89de5812159877d9b4f0a993bf41697a5a875940cd1eb71f24"
+source = "git+https://github.com/romanb/rust-libp2p?branch=multicon#86cf85b9a53f2895819ee45f55e8adcdbbe7a72f"
 dependencies = [
  "bytes 0.5.4",
  "futures 0.3.4",
@@ -2657,7 +2656,7 @@ dependencies = [
  "libp2p-wasm-ext",
  "libp2p-websocket",
  "libp2p-yamux",
- "parity-multiaddr 0.7.2",
+ "parity-multiaddr 0.7.3",
  "parity-multihash 0.2.3",
  "parking_lot 0.10.0",
  "pin-project",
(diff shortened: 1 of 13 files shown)


romanb:
Superseded by https://github.com/paritytech/substrate/pull/5278.

PR #5068, opened 2020-02-26
Remove `Backend::destroy_state`

bkchr:
This removes the `destroy_state` function of `Backend` and instead moves
the functionality into the `Drop` implementation of the state. This
makes it much easier to work with the state, as the user no longer needs
to call` destroy_state` manually. However, it requires that we switch
from `RwLock` to `ReentrantMutex` as while importing a block we maybe
need to lock again in `drop`.


arkpar:
Using mutex instead of rwlock prevents parallel blockchain queries. Currently when there's no import in progress multiple threads can query blockchain data. This change seems to introduce contention for such queries.

client/db/src/storage_cache.rs
```diff
@@ -479,13 +496,47 @@ impl<S: StateBackend<HasherFor<B>>, B: BlockT> CachingState<S, B> {
...
+	/// This is just a convience method, because `cache` is a `Option`.
+	#[cfg(test)]
+	fn cache_mut(&mut self) -> &mut CacheChanges<B> {
+		self.cache.as_mut().expect("`cache` is always `Some(_)` for the lifetime of the object")
+	}
+
 	/// Dispose state and return cache data.
-	pub fn release(self) -> CacheChanges<B> {
-		self.cache
+	pub fn release(mut self) -> CacheChanges<B> {
```
arkpar: Is this still needed?

client/db/src/storage_cache.rs
```diff
@@ -302,12 +313,14 @@ pub struct CachingState<S: StateBackend<HasherFor<B>>, B: BlockT> {
 	/// Backing state.
 	state: S,
 	/// Cache data.
-	pub cache: CacheChanges<B>,
+	cache: Option<CacheChanges<B>>,
+	/// Data associated to the backend that created this instance.
+	backend: Option<CachingStateBackend<B>>,
```
arkpar: Do `cache ` and `backend `really need to be `Option`? It look slike `drop` implementation does not really require that.

arkpar:
You could probably use `RWLock::read_recursive` although this has a downside. Write starvation becomes possible. And we want to prioritize block import above any queries. So not sure if this is a good idea as well.

client/db/src/lib.rs
```diff
@@ -544,7 +544,7 @@ impl<Block: BlockT> BlockImportOperation<Block> {
 }
 
 impl<Block: BlockT> sc_client_api::backend::BlockImportOperation<Block> for BlockImportOperation<Block> {
-	type State = CachingState<RefTrackingState<Block>, Block>;
+	type State = SyncingCachingState<RefTrackingState<Block>, Block>;
```
bkchr: It would be really nice to have here `CachingState`, but that doesn't work and requires that we do some workarounds (`disable_syncing`)

client/db/src/storage_cache.rs
```diff
@@ -668,6 +668,213 @@ impl<S: StateBackend<HasherFor<B>>, B: BlockT> StateBackend<HasherFor<B>> for Ca
...
+	/// Reference to the meta db.
+	meta: Arc<RwLock<Meta<NumberFor<Block>, Block::Hash>>>,
+	/// Mutex to lock get exlusive access to the backend.
+	lock: Arc<RwLock<()>>,
+	/// The wrapped caching state.
+	///
+	/// This is required to be a `Option`, because sometimes we want to extract
+	/// the cache changes and Rust does not allow to move fields from types that
+	/// implement `Drop`.
+	caching_state: Option<CachingState<S, Block>>,
```
arkpar: Instead you could do something like 
```rust 
impl CachigState {
  fn drain(&mut self) -> LocalCache {
    mem::replace(&mut *self.local_cache.write(), Default::default())
  }
}

client/db/src/lib.rs
arkpar: `into_cache_changes` leaves the local cache empty. So `drop` would have nothing to sync anyway, and it would not have to take any locks in such case. You could probably remove the `disable_syncing` workaround and simply do nothing if the cache is empty.

client/db/src/storage_cache.rs
bkchr: We use `sync_cache` inside of `try_commit_operation`. So, this will not work.

client/db/src/lib.rs
bkchr: I had dead locks when I did not disable syncing for the state in block import. Did not really track it down. I think we should get rid of the `get_import_lock` function in the long term and than we should be able to maybe remove `disable_syncing`.

arkpar: The deadlock is because in `drop` you don't check if the cache is empty before taking the lock. If add that check first it should be fine, as you'd always empty the cache on import.
I agree that `get_import_lock` should be removed eventually.

arkpar: On the other hand `disable_syncing` make the code more explicit and easier to follow

gavofyork:
@bkchr needs resolving (and tests failing?)

PR #5069, opened 2020-02-26
TelemetryEndpoints must be valid MutliAddr URL

### LINKED ISSUE #2803: Creating a TelemetryEndpoints should parse the URL
If the telemetry URL is bad, we currently store it in `TelemetryEndpoints` and pass it to the telemetry, which will then print an error.

Instead, creating the `TelemetryEndpoints` should parse the URL and return an error if it is invalid.

pscott:
- Closes #2803

Previously, when creating a telemetry endpoint we would not verify the validity of the provided URL. If we the URL was invalid, we would just throw a `warning!`, saying that the URL was invalid.

With this PR, we move this parsing logic to the creation of the `TelemetryEndpoints` struct.


client/telemetry/src/lib.rs
```diff
@@ -96,14 +96,35 @@ pub struct TelemetryConfig {
...
+pub struct TelemetryEndpoints(Vec<(Multiaddr, u8)>);
 
 impl TelemetryEndpoints {
-	pub fn new(endpoints: Vec<(String, u8)>) -> Self {
-		TelemetryEndpoints(endpoints)
+	pub fn new(endpoints: Vec<(String, u8)>) -> Result<Self, libp2p::multiaddr::Error> {
+		let endpoints: Result<Vec<(Multiaddr, u8)>, libp2p::multiaddr::Error> = endpoints.iter()
+			.map(|e| Ok((url_to_multiaddr(&e.0)?, e.1)))
+			.collect();
+		endpoints.map(Self)
```
pscott: I first had a one-liner that was written like this:
```rust
	pub fn new(endpoints: Vec<(String, u8)>) -> Result<Self, libp2p::multiaddr::Error> {
        endpoints.iter()
	        .map(|e| Ok((url_to_multiaddr(&e.0)?, e.1)))
			.collect::<Result<Vec<Multiaddr, u8>, libp2p::multiaddr::Error>>()
		    map(Self)
```
but ultimately decided against it for code clarity.

client/cli/src/commands/runcmd.rs
```diff
@@ -686,7 +684,7 @@ mod tests {
 			"test-id",
 			|| (),
 			vec!["boo".to_string()],
-			Some(TelemetryEndpoints::new(vec![("foo".to_string(), 42)])),
+			TelemetryEndpoints::new(vec![("wss://foo/bar".to_string(), 42)]).ok(),
```
pscott: is "foo" on its own a valid URL or multiaddr? if it is, we can cancel part of this change :)

pscott:
@reviewers
Should I be adding a couple of unit tests to this? I didn't add them because the code is pretty straightforward and I felt like those tests wouldn't bring much but if you wish, I can add some.

client/telemetry/src/lib.rs
```diff
@@ -300,3 +297,34 @@ macro_rules! telemetry {
...
+		assert!(telem.is_err());
+	}
+
+	#[test]
+	fn valid_and_invalid_endpointsj() {
+		let endp = vec![("/ip4/80.123.90.4/tcp/5432".into(), 3), ("/ip4/no:!?;rlkqre;;::::///tcp/5432".into(), 4)];
+		let telem = TelemetryEndpoints::new(endp);
+		assert!(telem.is_err());
+	}
+}
```
pscott: Not sure those are useful or necessary.

client/cli/src/commands/runcmd.rs
NikVolf: Not more valid than `wss://foo/bar`

I thought `/wss/foo` will be valid?

pscott: All right. I took my example from here: https://github.com/paritytech/substrate/blob/5b41575012e935ffefee03d39ae5d925ac7992fd/client/chain-spec/res/chain_spec.json#L18-L20

Guess I'm missing something :p

client/telemetry/src/lib.rs
```diff
@@ -300,3 +297,34 @@ macro_rules! telemetry {
...
+
+#[cfg(test)]
+mod telemetry_endpoints_tests {
+	use libp2p::Multiaddr;
+	use super::TelemetryEndpoints;
+
+	#[test]
+	fn valid_endpoints() {
+		let endp = vec![("/ip4/80.123.90.4/tcp/5432".into(), 3), ("/ip4/80.123.90.4/tcp/5432".into(), 4)];
+		let telem = TelemetryEndpoints::new(endp.clone()).unwrap();
```
NikVolf: `.expect("what failed")` instead of `unwrap()`

saves time when tests fail

client/cli/src/commands/runcmd.rs
NikVolf: Have no idea also, but according to `url_to_multiaddr` function below probably both variants valid

client/cli/src/commands/runcmd.rs
```diff
@@ -416,9 +416,7 @@ impl RunCmd {
 		if self.no_telemetry {
 			config.telemetry_endpoints = None;
 		} else if !self.telemetry_endpoints.is_empty() {
-			config.telemetry_endpoints = Some(
-				TelemetryEndpoints::new(self.telemetry_endpoints.clone())
-			);
+			config.telemetry_endpoints = TelemetryEndpoints::new(self.telemetry_endpoints.clone()).ok();
```
tomaka: This should return an error and not just silently put a `None`, otherwise you can pass an invalid URL, it will not work, and you have no idea why

client/cli/src/commands/runcmd.rs
pscott: Yes, basically we either provide a valid url or a multiaddr and `url_to_multiaddr` transforms that into a valid multiaddr (when possible :p).
I believe that:
1. `wss://foo/bar` is a valid URL 
2. `/wss/foo` is a valid multiaddr

My question was whether:
```rust
"foo".is_valid_url() || "foo".is_valid_multiaddr()
```

tomaka:
Also this is technically a breaking change, so I guess we would wait for 3.0 to land it?
(cc @gnunicorn)

pscott:
Should we merge this one as well as #5057 into one bigger PR? Might be clearer?

pscott:
Tests are failing on CI but they succeeded locally on my machine. Last commits that got merged to master were failing CI too.. Not sure what the protocol is in this situation... :)

bin/node/cli/src/chain_spec.rs
```diff
@@ -159,7 +159,8 @@ pub fn staging_testnet_config() -> ChainSpec {
 		"staging_testnet",
 		staging_testnet_config_genesis,
 		boot_nodes,
-		Some(TelemetryEndpoints::new(vec![(STAGING_TELEMETRY_URL.to_string(), 0)])),
+		Some(TelemetryEndpoints::new(vec![(STAGING_TELEMETRY_URL.to_string(), 0)])
+			.expect("Staging telemetry url should be valid")),
```
bkchr:
```suggestion
			.expect("Staging telemetry url is valid; qed")),
```

client/telemetry/src/lib.rs
```diff
@@ -96,12 +96,45 @@ pub struct TelemetryConfig {
...
 /// The URL string can be either a URL or a multiaddress.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct TelemetryEndpoints(Vec<(String, u8)>);
+pub struct TelemetryEndpoints(
+	#[serde(deserialize_with = "url_or_multiaddr_deser")]
+	Vec<(Multiaddr, u8)>
+);
+
+/// Custom deserializer for TelemetryEndpoints, used to convert urls or multiaddr to multiaddr.
+fn url_or_multiaddr_deser<'de, D>(deserializer: D) -> Result<Vec<(Multiaddr, u8)>, D::Error> where D: Deserializer<'de> {
```
bkchr:
```suggestion
fn url_or_multiaddr_deser<'de, D>(deserializer: D) -> Result<Vec<(Multiaddr, u8)>, D::Error>
	where D: Deserializer<'de>
{
```

thiolliere:
polkadot PR https://github.com/paritytech/polkadot/pull/941

PR #5075, opened 2020-02-27
Propagate DispatchError for benchmarks.

marcio-diaz:
Currently `dispatch_benchmark` is returning `None` if there is an error in the call.
This produces the unhelpful output of "No results." (cc @drahnr ).

In order to facilitate the benchmarking process this PR propagates `DispatchError` (e.g. producing output `Error: "TooManyFields"`).


primitives/runtime/src/lib.rs
```diff
@@ -416,6 +416,13 @@ impl From<DispatchError> for &'static str {
 	}
 }
 
+impl From<DispatchError> for Vec<u8> {
+	fn from(err: DispatchError) -> Vec<u8> {
+		let s: &'static str = err.into();
+		s.as_bytes().to_vec()
+	}
```
marcio-diaz: Mmm, probably I should remove this and use it in `run_benchmarks`.

bin/node/runtime/src/lib.rs
```diff
@@ -821,14 +821,14 @@ impl_runtime_apis! {
...
 			match module.as_slice() {
-				b"pallet-balances" | b"balances" => Balances::run_benchmark(extrinsic, steps, repeat).ok(),
-				b"pallet-identity" | b"identity" => Identity::run_benchmark(extrinsic, steps, repeat).ok(),
-				b"pallet-timestamp" | b"timestamp" => Timestamp::run_benchmark(extrinsic, steps, repeat).ok(),
-				_ => None,
+				b"pallet-balances" | b"balances" => Balances::run_benchmark(extrinsic, steps, repeat),
+				b"pallet-identity" | b"identity" => Identity::run_benchmark(extrinsic, steps, repeat),
+				b"pallet-timestamp" | b"timestamp" => Timestamp::run_benchmark(extrinsic, steps, repeat),
+				_ => Err("Benchmark not found for this pallet.".as_bytes().to_vec()),
 			}
```
bkchr: Just put here a `.map_err(|e| e.encode())`

utils/frame/benchmarking-cli/src/lib.rs
```diff
@@ -109,34 +109,37 @@ impl BenchmarkCmd {
...
-				let parameters = &result.0;
-				parameters.iter().for_each(|param| print!("{:?},", param.1));
-				// Print extrinsic time and storage root time
-				print!("{:?},{:?}\n", result.1, result.2);
-			});
-
-			eprintln!("Done.");
-		} else {
-			eprintln!("No Results.");
+		let results = <Result<Vec<BenchmarkResults>, Vec<u8>> as Decode>::decode(&mut &result[..])
```
bkchr:
```suggestion
		let results = <Result<Vec<BenchmarkResults>, String> as Decode>::decode(&mut &result[..])
```

marcio-diaz: I get `Error: "�Benchmark not found for this pallet."` if we don't decode twice. Other option is to encode the whole thing?, but maybe you don't want that, given your other comment of encoding just the error.

marcio-diaz: I get `Error: "�Benchmark not found for this pallet."` if we don't decode twice. Other option is to encode the whole thing?, but maybe you don't want that, given your other comment of encoding just the error.

bkchr: Ahh yeah! Sorry!

Could you please use [`RuntimeString`](https://github.com/paritytech/substrate/blob/879e28ae38643c028fa09c157e7b2714bcbf02a1/primitives/runtime/src/runtime_string.rs#L23)

(You also need to change the `Benchmark` trait to directly return `Result<BenchmarkResult, RuntimeString>`

marcio-diaz: Nice, thanks!

utils/frame/benchmarking-cli/src/lib.rs
```diff
@@ -109,34 +109,36 @@ impl BenchmarkCmd {
...
-				// Print extrinsic time and storage root time
-				print!("{:?},{:?}\n", result.1, result.2);
-			});
-
-			eprintln!("Done.");
-		} else {
-			eprintln!("No Results.");
+
+		let results = <Result<Vec<BenchmarkResults>, String> as Decode>::decode(&mut &result[..])
+			.expect("failed to decode benchmark results");
```
bkchr:
```suggestion
			.map_err(|e| format!("Failed to decode benchmark results: {:?}", e))?;
```

I think that makes it cleaner if when it fails.

PR #5077, opened 2020-02-27
Adds more Democracy docs

lsaether:
The docs for the Democracy pallet were kinda lacking compared to some of the newer pallets. This PR brings them up to higher standards.


frame/democracy/src/lib.rs
```diff
@@ -14,7 +14,122 @@
...
+//!
+//! There are two different queues that a proposal can be added to before it
+//! becomes a referendum, 1) the proposal queue consisting of all public proposals
+//! and 2) the external queue consisting of a single proposal that originates
+//! from one of the _external_ origins (such as a collective group).
+//!
+//! Every launch period, a proposal will be taken from either the proposal queue
+//! or the external queue in turn. The taken proposal will become a referendum
+//! that can be voted on by any token holder in the system. The voting system
+//! uses timelock voting by allowing the token holder to set their _conviction_
```
```suggestion
//! uses time-lock voting by allowing the token holder to set their _conviction_
```

frame/democracy/src/lib.rs
```diff
@@ -14,7 +14,122 @@
...
+//! The Democracy pallet handles the administration of general stakeholder voting.
+//!
+//! There are two different queues that a proposal can be added to before it
+//! becomes a referendum, 1) the proposal queue consisting of all public proposals
+//! and 2) the external queue consisting of a single proposal that originates
+//! from one of the _external_ origins (such as a collective group).
+//!
+//! Every launch period, a proposal will be taken from either the proposal queue
+//! or the external queue in turn. The taken proposal will become a referendum
+//! that can be voted on by any token holder in the system. The voting system
```
joepetrowski: Define launch period and active voice where possible.
```suggestion
//! Every launch period - a length defined in the runtime - the Democracy pallet 
//! launches a referendum from a proposal that it takes from either the proposal
//! queue or the external queue in turn. Any token holder in the system can vote
//! on referenda. The voting system
```

frame/democracy/src/lib.rs
```diff
@@ -14,7 +14,122 @@
...
+//! uses time-lock voting by allowing the token holder to set their _conviction_
+//! behind a vote. The conviction will dictate the length of time the tokens
+//! will be locked, as well as the multiplier that scales the vote power.
+//!
+//! ### Terminology
+//!
+//! - **Conviction:** A multiplier that is added to a vote based on the length
+//!   of time the voter has opted to lock their tokens. Longer lock-up periods
+//!   grant a stronger vote.
```
joepetrowski: Are multipliers added? Also, this (_lightly_) implies that the lockup multiplier == the vote multiplier, which is not true (see impl). I think it should be clearer and more generic.
```
	/// The amount of time (in number of periods) that our conviction implies a successful voter's
	/// balance should be locked for.
	fn lock_periods(self) -> u32 {
		match self {
			Conviction::None => 0,
			Conviction::Locked1x => 1,
			Conviction::Locked2x => 2,
			Conviction::Locked3x => 4,
			Conviction::Locked4x => 8,
			Conviction::Locked5x => 16,
			Conviction::Locked6x => 32,
		}
	}
```

```suggestion
//! - **Enactment Period:** The minimum period of locking and the period between a proposal being
//! approved and enacted. 
//! - **Lock Period:** A period of time after proposal enactment that the tokens of _winning_ voters
//! will be locked.
//! - **Conviction:** An indication of a voter's strength of belief in their vote. An increase
//! of one in conviction indicates that a token holder is willing to lock their tokens for twice
//! as many lock periods after enactment.
```

frame/democracy/src/lib.rs
```diff
@@ -14,7 +14,122 @@
...
+//! - **Vote:** A value that can either be in approval ("Aye") or rejection ("Nay")
+//!   of a particular referendum.
+//! - **Proposal:** A submission to the chain that represents an action that is 
+//!   being suggested for the system to adopt.
```
joepetrowski:
```suggestion
//! - **Proposal:** A submission to the chain that represents an action that a proposer (either an
//! account or an external origin) suggests that the system adopt.
```

frame/democracy/src/lib.rs
```diff
@@ -14,7 +14,122 @@
...
+//! - **Referendum:** A proposal that is in the process of being voted on for 
+//!   either acceptance or rejection as a change to the system.
+//! - **Proxy:** An account that votes on behalf of a separate "Stash" account
+//!   that holds the funds.
+//! - **Delegation:** Granting your voting power to the decisions of another account.
```
joepetrowski:
```suggestion
//! - **Delegation:** The act of granting your voting power to the decisions of another account.
```

frame/democracy/src/lib.rs
```diff
@@ -14,7 +14,122 @@
...
+//! ## Interface
+//!
+//! ### Dispatchable Functions
+//!
+//! #### Public
+//!
+//! These calls can be made from any externally held account capable of creating
+//! a signed extrinsic.
+//!
+//! - `propose` - The submission of a sensitive action, represented as a hash.
```
joepetrowski: List parallelism. Personally, I would put the opening verb of this whole list in the infinitive (as the original function documentation does), i.e. "Submit a ...".
```suggestion
//! - `propose` - Submits a sensitive action, represented as a hash. 
```

frame/democracy/src/lib.rs
```diff
@@ -14,7 +14,122 @@
...
+//! - `second` - Signals agreement with a proposal, moves it higher on the
+//!   proposal queue, and requires a matching deposit to the original.
+//! - `vote` - Votes in a referendum, either the vote is "Aye" to enact the 
+//!   proposal or "Nay" to keep the status quo.
+//! - `proxy_vote` - Votes in a referendum on behalf of a stash account.
+//! - `activate_proxy` - Activates a proxy that is already open to the sender.
+//! - `close_proxy` - Clears the proxy status, called by the proxy.
+//! - `deactivate_proxy` - Deactivates a proxy back to the open status, called by
+//!   the stash.
+//! - `open_proxy` -  Opens a proxy account on behalf of the sender.
```
joepetrowski:
```suggestion
//! - `open_proxy` - Opens a proxy account on behalf of the sender.
```

frame/democracy/src/lib.rs
```diff
@@ -14,7 +14,122 @@
...
+//! This call can only be made by the `CancellationOrigin`.
+//!
+//! - `emergency_cancel` - Schedules an emergency cancellation of a referendum.
+//!   Can only happen once to a specific referendum.
+//!
+//! #### ExternalOrigin
+//!
+//! This call can only be made by the `ExternalOrigin`.
+//!
+//! - `external_propose` - Schedules a referendum to be tabled once it is is legal
```
joepetrowski: Let's avoid "table" since it means literally the opposite thing in US and British English.
```suggestion
//! - `external_propose` - Schedules a proposal to become a referendum once it is is legal
```

frame/democracy/src/lib.rs
```diff
@@ -14,7 +14,122 @@
...
+//!   for an externally proposed referendum.
+//!
+//! #### External Majority Origin
+//!
+//! This call can only be made by the `ExternalMajorityOrigin`.
+//!
+//! - `external_propose_majority` - Schedules a majority-carries referendum once
```
joepetrowski: Should probably introduce passage thresholds in terminology.

frame/democracy/src/lib.rs
```diff
@@ -14,7 +14,122 @@
...
+//!
+//! - `external_propose_default` - Schedules a negative-turnout-bias referendum
+//!   to be tabled once it is legal for an externally proposed referendum.
+//!
+//! #### Fast Track Origin
+//!
+//! This call can only be made by the `FastTrackOrigin`.
+//!
+//! - `fast_track` - Schedules the current externally proposed referendum that
+//!   is "majority-carries" to be tabled immediately.
```
joepetrowski: Tabled again.

frame/democracy/src/lib.rs
```diff
@@ -688,6 +900,10 @@ decl_module! {
 		///
 		/// NOTE: Used to be called `set_proxy`.
 		///
+		/// The dispatch origin of this call must be _Signed_.
+		///
+		/// - `proxy`: The account which will be activated as proxy.
```
joepetrowski:
```suggestion
		/// - `proxy`: The account that will be activated as proxy.
```

frame/democracy/src/lib.rs
```diff
@@ -729,6 +947,10 @@ decl_module! {
 		///
 		/// NOTE: Used to be called `remove_proxy`.
 		///
+		/// The dispatch origin of this call must be _Signed_.
+		///
+		/// - `proxy`: The account which will be deactivated as proxy.
```
joepetrowski: I will die on this hill.
```suggestion
		/// - `proxy`: The account that will be deactivated as proxy.
```

frame/democracy/src/lib.rs
lsaether: resolved in 7322662bca557071ee2aa896f53ad43b6195bd6d

frame/democracy/src/lib.rs
lsaether: resolved in 061432cec327e4145b82cac87ed9a6e20a7b1d52

frame/democracy/src/lib.rs
```diff
@@ -14,7 +14,140 @@
...
+//! - **Proxy:** An account that votes on behalf of a separate "Stash" account
+//!   that holds the funds.
+//! - **Delegation:** The act of granting your voting power to the decisions of another account.
+//!
+//! ### Adaptive Quorum Biasing
+//!
+//! A _referendum_ can be either simple majority-carries in which 50%+1 of the
+//! votes decide the outcome or _adaptive quorum biased_. Adaptive quorum biasing
+//! makes the threshold for passing or rejecting a referendum higher or lower
+//! depending on how the referendum was original proposed. There are two types of
```
joepetrowski:
```suggestion
//! depending on how the referendum was originally proposed. There are two types of 
```

PR #5082, opened 2020-02-27
Offence reporting returns a result

montekki:
Add a `Result` return type for `report_offence` in `offences` module. At this moment it's only purpose is  to report the caller if they are submitting a duplicate offence report.

This is a companion for paritytech/polkadot#840

### DIFF
--- a/frame/im-online/src/lib.rs
+++ b/frame/im-online/src/lib.rs
@@ -608,7 +608,12 @@ impl<T: Trait> pallet_session::OneSessionHandler<T::AccountId> for Module<T> {
 
 			let validator_set_count = keys.len() as u32;
 			let offence = UnresponsivenessOffence { session_index, validator_set_count, offenders };
-			T::ReportUnresponsiveness::report_offence(vec![], offence);
+			if let Err(e) = T::ReportUnresponsiveness::report_offence(vec![], offence) {
+				debug::info!(
+					target: "imonline",
+					"Failed to report an offence: {:?}", e,
+				);
+			}
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


frame/im-online/src/lib.rs
```diff
@@ -608,7 +608,12 @@ impl<T: Trait> pallet_session::OneSessionHandler<T::AccountId> for Module<T> {
 
 			let validator_set_count = keys.len() as u32;
 			let offence = UnresponsivenessOffence { session_index, validator_set_count, offenders };
-			T::ReportUnresponsiveness::report_offence(vec![], offence);
+			if let Err(e) = T::ReportUnresponsiveness::report_offence(vec![], offence) {
+				debug::info!(
```
bkchr: The problem with this, it will never be printed (in wasm). This functionality is only for offchain workers (it requires that the `RuntimeLogger` is intialized). Better to just use `sp_runtime::print`.

primitives/staking/src/offence.rs
```diff
@@ -91,14 +91,21 @@ pub trait Offence<Offender> {
 	) -> Perbill;
 }
 
+/// Errors that may happen on offence reports.
+#[derive(PartialEq, sp_runtime::RuntimeDebug)]
+pub enum OffenceError {
+	/// The report has already been sumbmitted.
+	DuplicateReport,
```
bkchr: Do we at least want to have a variant that may takes a `u8` or whatever? Otherwise we will need to update this for every error that may come u.

frame/im-online/src/lib.rs
montekki: Yes, but the rest of the code in this modules uses this format so I decided to keep it the same.

primitives/staking/src/offence.rs
montekki: I think that `Other(u8)` should be added when the need arises, r.n. it would be just an unused variant?

bkchr: I will remind you on your next pr :P

frame/im-online/src/lib.rs
bkchr: The code you are referring too is offchain, that is the reason it uses this format.

primitives/staking/src/offence.rs
montekki: Ok, added it here

frame/im-online/src/lib.rs
montekki: done

PR #5085, opened 2020-02-28
Add some proof comment

thiolliere:
The proof for `q_reduce < 2 * $max` is not completely obvious, this comment make the proof clear

### DIFF
--- a/primitives/arithmetic/src/per_things.rs
+++ b/primitives/arithmetic/src/per_things.rs
@@ -132,7 +132,8 @@ macro_rules! implement_per_thing {
 				let factor = (q.clone() / $max.into()).max((1 as Self::Inner).into());
 
 				// q cannot overflow: (q / (q/$max)) < 2 * $max. p < q hence p also cannot overflow.
-				// this implies that Self::Inner must be able to fit 2 * $max.
+				// this implies that Self::Inner must be able to fit 2 * $max. This is guarenteed
+				// by the macros tests.
 				//
 				// Proof for `q_reduce < 2 * $max`:
 				// i.e. `floor(q / max(floor(q/$max), 1)) < 2 * $max`:
--- a/primitives/arithmetic/src/per_things.rs
+++ b/primitives/arithmetic/src/per_things.rs
@@ -133,6 +133,22 @@ macro_rules! implement_per_thing {
 
 				// q cannot overflow: (q / (q/$max)) < 2 * $max. p < q hence p also cannot overflow.
 				// this implies that Self::Inner must be able to fit 2 * $max.
+				//
+				// Proof for `q_reduce < 2 * $max`:
+				// i.e. `floor(q / max(floor(q/$max), 1)) < 2 * $max`:
+				// ```
+				// if q <= $max:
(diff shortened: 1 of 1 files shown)


kianenigma:
the `round` can be implemented as such: 

```rust
let div_round_nearest = |x: N, f: N| -> N {
	let mut q = x / f;
	let r = x.rem(f);
	if r > (f / N::from(2)) {
		q = q + N::from(1);
	}
	q
};
```

kianenigma:
https://www.desmos.com/calculator/z34z6eaetw although it is only correct when you do `ceil`. Makes sense, `q / max` will always be equal or an epsilon more than its _real_ value. Hence `q / (q/max)` will always be less than `max`.

thiolliere:
> Sorry I got into thinking again :D
> 
> first, maybe it is better if you push to the offchian phragmne PR as there are some more changes over there?
> 
> And, this proof implies that we to further do
> 
> ```
> 				let part =
> 					p_reduce as $upper_type
> 					* <$upper_type>::from($max)
> 					/ q_reduce as $upper_type;
> ```
> 
> we need yet another type that can fit `2 * max * max`.

the comment state `// this implies that Self::Inner must be able to fit 2 * $max.`. If I remember correctly this was prooved by a test but I'll update the comment to mention the test if I find it.

> while if we change `factor = floor(q / max)` to `factor = round(q / max)`, then you can prove that:
> 
> `q_reduce` and `p_reduce` are always within `$max`, hence `let part = p_reduce as $upper_type * <$upper_type>::from($max) / ...` is also safe because `$upper_type` can fit `$max * $max`

I have to think if we loose some precision by using this factor, but maybe it is fine indeed.

thiolliere:
I think using ceil is a good idea and it will avoid using a new type for PerU16 or PerU32

thiolliere:
also mentionning maybe we should do ceil in the implementation of CurrencyToVoteHandler just to be more clean:
```rust
impl CurrencyToVoteHandler {
	fn factor() -> Balance { (Balances::total_issuance() / u64::max_value() as Balance).max(1) }
}

impl Convert<Balance, u64> for CurrencyToVoteHandler {
	fn convert(x: Balance) -> u64 { (x / Self::factor()) as u64 }
}

impl Convert<u128, Balance> for CurrencyToVoteHandler {
	fn convert(x: u128) -> Balance { x * Self::factor() }
}
```
it is fine because nobody has a balance close to total_issuance but actually if someone has its balance close to total_issuance then the conversion might not fit in u64 actually.

kianenigma:
Curretn totalIssuance is actually a big big problem. We use this to `convert()` very often and every time it reads from storage (caching sometimes helps but not always). This should be changed anyhow.

kianenigma:
@thiolliere can you check if https://github.com/paritytech/substrate/pull/4517 is now okay regards to documenting this section? and see if you prefer to update this PR or close it.

thiolliere:
https://github.com/paritytech/substrate/pull/4517 indeed superseed correctly this need for proof.

Close because https://github.com/paritytech/substrate/pull/4517 superseed.

PR #5090, opened 2020-02-28
Make export blocks default to json on stdout

### LINKED ISSUE #4378: [client cli]: export blocks on db with no blocks prints junk
```bash
$  ./target/debug/substrate export-blocks --from 0 --to 100
2019-12-12 11:57:58 DB path: /share/substrate/chains/flaming-fir/db
2019-12-12 11:57:59 Initializing Genesis block/state (state: 0xd733…e696, header-hash: 0xb653…1d25)
2019-12-12 11:57:59 Loading GRANDPA authority set from genesis on what appears to be first startup.
2019-12-12 11:57:59 Loaded block-time = BabeConfiguration { slot_duration: 3000, epoch_length: 200, c: (1, 4), genesis_authorities: [(Public(00299981a2b92f878baaf5dbeba5c18d4e70f2a1fcd9c61b32ea18daf38f4378 (5C4vDQxA...)), 1), (Public(482a3389a6cf42d8ed83888cfd920fec738ea30f97e44699ada7323f08c3380a (5DhKqkHR...)), 1), (Public(482dbd7297a39fa145c570552249c2ca9dd47e281f0c500c971b59c9dcdcd82e (5DhLtiaQ...)), 1), (Public(6e7e4eb42cbd2e0ab4cae8708ce5509580b8c04d11f6758dbf686d50fe9f9106 (5EZaeQ8d...)), 1)], randomness: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], secondary_slots: true } milliseconds from genesis on first-launch
2019-12-12 11:57:59 Creating empty BABE epoch changes on what appears to be first startup.
2019-12-12 11:57:59 Exporting blocks from #0 to #100
e3U_    ,1۝ T,g
              A]?
2019-12-12 11:57:59 #0
.uL9bW燆L% 
```

pscott:
- Fixes #4378 

As suggested by @arkpar , this PR will make export blocks default to json instead of encoded format when output is stdout.
The user can override this behaviour by using the `--json false` flag.

EDIT: now defaulting to json, and switched the flag `--json` to be `--binary`.


client/cli/src/commands/export_blocks_cmd.rs
```diff
@@ -51,7 +51,10 @@ pub struct ExportBlocksCmd {
 	pub to: Option<BlockNumber>,
 
 	/// Use JSON output rather than binary.
-	#[structopt(long = "json")]
+	#[structopt(long = "json",
+		parse(try_from_str),
+		default_value("true"),
+		default_value_if("output", None, "false"))]
 	pub json: bool,
```
bkchr: That is not intuitive. You should rename this to `binary` and print in binary format if this flag is given.

pscott: Yes, you're right, defaulting to JSON and requiring a `binary` flag is probably the best thing to do!

PR #5092, opened 2020-02-29
Task manager for background/async tasks in service

NikVolf:
This all started that I need a way to spawn background tasks inside `Client` (to be more precise in call executor) for #5023 - but this part better to be in separate PR.

With this pr
- `on_exit` is now managed inside TaskManager and not cloned everywhere 
- `TaskManagerBuilder` is created before client and client can schedule tasks, along with all other scheduling during service building.
- `TaskManager`/`TaskManagerBuilder` can spawn any future, no need to spam `Box::pin` everywhere
- generally cleaner code for dealing with async/background tasks requests


client/service/src/lib.rs
```diff
@@ -149,6 +144,77 @@ impl Spawn for SpawnTaskHandle {
...
+	/// Receiver for futures that must be spawned as background tasks.
+	to_spawn_rx: mpsc::UnboundedReceiver<(Pin<Box<dyn Future<Output = ()> + Send>>, Cow<'static, str>)>,
+}
+
+impl TaskManager {
+	/// New task manager for service.
+	pub fn new() -> Self {
+		let (signal, on_exit) = exit_future::signal();
+		let (to_spawn_tx, to_spawn_rx) =
+			mpsc::unbounded::<(Pin<Box<dyn Future<Output = ()> + Send>>, Cow<'static, str>)>();
```
bkchr: Isn't rust able to resolve this on its own?

client/service/src/lib.rs
```diff
@@ -149,6 +144,77 @@ impl Spawn for SpawnTaskHandle {
...
+	/// Process background task receiver
+	fn process_receiver(&mut self, executor: &ServiceTaskExecutor, cx: &mut Context) {
+		while let Poll::Ready(Some((task_to_spawn, name))) = Pin::new(&mut self.to_spawn_rx).poll_next(cx) {
+			(executor)(Box::pin(futures_diagnose::diagnose(name, task_to_spawn)));
+		}
+	}
+}
+
+
+impl Drop for TaskManager
```
bkchr:
```suggestion
impl Drop for TaskManager {
```

client/service/src/lib.rs
```diff
@@ -149,6 +144,77 @@ impl Spawn for SpawnTaskHandle {
...
+{
```
bkchr:
```suggestion
```

client/service/src/lib.rs
```diff
@@ -149,6 +144,77 @@ impl Spawn for SpawnTaskHandle {
...
+	fn drop(&mut self) {
+		debug!(target: "service", "Substrate service shutdown");
```
bkchr: Copy-pasta

client/service/src/lib.rs
```diff
@@ -149,6 +144,77 @@ impl Spawn for SpawnTaskHandle {
 	}
 }
 
+/// Helper struct to manage background/async tasks in Service.
+pub struct TaskManager {
```
bkchr: Could be moved into its own file.

NikVolf: Ok... I think I will improve on api then also, since it won't be internal part of service

client/service/src/task_manager.rs
```diff
@@ -0,0 +1,148 @@
...
+			futures::pin_mut!(task);
+			let _ = select(on_exit, task).await;
+		};
+		if self.to_spawn_tx.unbounded_send((Box::pin(future), name.into())).is_err() {
+			error!("Failed to send task to spawn over channel");
+		}
+	}
+
+	/// Get sender where background/async tasks can be sent.
+	pub fn scheduler(&self) -> TaskScheduler {
```
tomaka: This is an implementation detail that really shouldn't be exposed in the API.
If it's really necessary, I'm ok with `pub(super)` with a comment saying that this is an implementation detail.

client/service/src/task_manager.rs
```diff
@@ -0,0 +1,148 @@
...
+
+	/// Process background task receiver
+	pub(crate) fn process_receiver(&mut self, executor: &ServiceTaskExecutor, cx: &mut Context) {
+		while let Poll::Ready(Some((task_to_spawn, name))) = Pin::new(&mut self.to_spawn_rx).poll_next(cx) {
+			(executor)(Box::pin(futures_diagnose::diagnose(name, task_to_spawn)));
+		}
+	}
+
+	/// Clone on exit signal.
+	pub fn on_exit(&self) -> exit_future::Exit {
```
tomaka: Same here.

client/service/src/task_manager.rs
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+
+	/// Get sender where background/async tasks can be sent.
+	pub fn scheduler(&self) -> TaskScheduler {
+		self.to_spawn_tx.clone()
+	}
+
+	/// Process background task receiver
+	pub(crate) fn process_receiver(&mut self, executor: &ServiceTaskExecutor, cx: &mut Context) {
```
tomaka: This is also an implementation detail and generally a bad API. I'm really not sure it's a good idea to extract this struct in a separate module.

client/service/src/lib.rs
```diff
@@ -326,7 +264,7 @@ where
 	}
 
 	fn on_exit(&self) -> exit_future::Exit {
```
tomaka: We can probably deprecate this method?

client/service/src/task_manager.rs
NikVolf: Agree, for some reason I thought it was exposed by Service already, as `on_exit` was.

client/service/src/task_manager.rs
NikVolf: Totally agree,

Unfortunately, this is here because of the way actual ServiceTaskExecutor is instantiated (in `run_until_exit`. Otherwise TaskManager could've own executor and never expose it.

client/service/src/lib.rs
NikVolf: Fine with me.

client/service/src/task_manager.rs
```diff
@@ -0,0 +1,148 @@
...
+/// An handle for spawning tasks in the service.
+#[derive(Clone)]
+pub struct SpawnTaskHandle {
+	sender: TaskScheduler,
+	on_exit: exit_future::Exit,
+}
+
+impl SpawnTaskHandle {
+	/// Spawns the given task with the given name.
+	pub fn spawn(&self, name: impl Into<Cow<'static, str>>, task: impl Future<Output = ()> + Send + 'static) {
```
expenses: You could possibly have this take an `impl Future<Output = T>` instead if you wanted to avoid using `.map(drop)` when calling it.

client/service/src/task_manager.rs
NikVolf: Okay, I decoupled this

But it had required some more code.. :)

client/service/src/task_manager.rs
NikVolf: Hmm I am not sure I want something to be dropped silently, and this `T` cannot be returned in any way.

client/service/src/task_manager.rs
```diff
@@ -0,0 +1,193 @@
...
+pub type TaskExecutor = Arc<dyn Spawn + Send + Sync>;
+
+/// Type alias for service task executor (usually runtime).
+pub type ServiceTaskExecutor = Arc<dyn Fn(Pin<Box<dyn Future<Output = ()> + Send>>) + Send + Sync>;
+
+/// Type alias for the task scheduler.
+pub type TaskScheduler = mpsc::UnboundedSender<(Pin<Box<dyn Future<Output = ()> + Send>>, Cow<'static, str>)>;
+
+/// Helper struct to setup background tasks execution for service.
+pub struct TaskManagerSetup {
```
tomaka:
```suggestion
pub struct TaskManagerBuilder {
```

client/service/src/task_manager.rs
```diff
@@ -158,27 +155,27 @@ impl TaskManager {
...
 
 	/// Get sender where background/async tasks can be sent.
-	pub(crate) fn scheduler(&self) -> TaskScheduler {
+	pub fn scheduler(&self) -> TaskScheduler {
 		self.to_spawn_tx.clone()
 	}
 
 	/// Process background task receiver.
-	pub(crate) fn process_receiver(&mut self, cx: &mut Context) {
+	pub fn process_receiver(&mut self, cx: &mut Context) {
```
NikVolf: TaskManager is now not exposed and is internal within the `sc_service` crate

client/service/src/lib.rs
```diff
@@ -224,6 +174,7 @@ pub trait AbstractService: 'static + Future<Output = Result<(), Error>> +
 	fn transaction_pool(&self) -> Arc<Self::TransactionPool>;
 
 	/// Get a handle to a future that will resolve on exit.
+	#[deprecated]
```
expenses: Might be worth having a note about what to use instead.

client/service/src/task_manager.rs
tomaka: I don't know, I'd still mark these functions as `pub(super)`. I can totally imagine someone sneaking in a pull request exposing the task manager without thinking about it, and then we're stuck with this PR.

NikVolf:
>  NikVolf removed the B2-breaksapi  label now

Seems that changes touch (1) return type of new_full_parts (2) ServiceBuilder::new_light and (3) Service builder internals, and those are not part of public api

PR #5100, opened 2020-03-02
Fix benchmarks! macro for non dispatchable code

marcio-diaz:
@shawntabrizi  reported that benchmarking non dispatchable code using the `benchmarks!` macro doesn't work (and it should according to the doc written by gav). 

This fixes it and add some tests to `frame/example` (see `sort_vector` case).

Basically, instead of returning `(Call, RawOrigin)` from `BenchmarkingSetup::instance()` and then calling `dispatch` on it, now we return a closure to benchmark.


frame/example/src/lib.rs
```diff
@@ -258,10 +258,11 @@ use frame_support::{
 	dispatch::DispatchResult, decl_module, decl_storage, decl_event,
 	weights::{SimpleDispatchInfo, DispatchInfo, DispatchClass, ClassifyDispatch, WeighData, Weight, PaysFee},
 };
-use frame_system::{self as system, ensure_signed, ensure_root};
+use frame_benchmarking::{benchmarks, account};
```
shawntabrizi:
```suggestion
use sp_std::prelude::*;
use frame_benchmarking::{benchmarks, account};
```

frame/example/src/lib.rs
```diff
@@ -785,4 +824,62 @@ mod tests {
 		let info = custom_call.get_dispatch_info();
 		assert_eq!(info.weight, 2000);
 	}
+
+	#[test]
```
shawntabrizi: This test is weird to have in this crate.

You should keep tests for the macro with the macro code imo.

frame/example/src/lib.rs
```diff
@@ -642,12 +643,50 @@ impl<T: Trait + Send + Sync> SignedExtension for WatchDummy<T> {
...
+
+	// This will measure the execution time of `set_dummy` for b in [1..10] range.
+	another_set_dummy {
+		let b in 1 .. 10;
+		let caller = account("caller", 0, 0);
+	}: set_dummy (RawOrigin::Signed(caller), b.into())
+
+	// This will measure the execution time of sorting a vector.
+	sort_vector {
+		let x in 0 .. 1;
```
shawntabrizi:
```suggestion
		let x in 0 .. 10000;
```

This makes the benchmark actually do something :)

frame/example/src/lib.rs
marcio-diaz: Yeah, I was hesitating between writing it here or in the `benchmarking` crate. I choose here so I don't have to write an `example` code. But I can move if you think it's better.

frame/example/src/lib.rs
marcio-diaz: I didn't want it to spend much time in the test, it's just an example. But I guess 10000 it's not much either.

frame/benchmarking/src/lib.rs
```diff
@@ -463,28 +374,116 @@ macro_rules! selected_benchmark {
...
+			fn run_benchmark(
+				extrinsic: Vec<u8>,
+				lowest_range_values: Vec<u32>,
+				highest_range_values: Vec<u32>,
+				steps: Vec<u32>,
+				repeat: u32,
+			) -> Result<Vec<$crate::BenchmarkResults>, &'static str> {
+				// Map the input to the selected benchmark.
+				let extrinsic = sp_std::str::from_utf8(extrinsic.as_slice())
+					.map_err(|_| "Could not find extrinsic")?;
```
bkchr:
```suggestion
					.map_err(|_| "`extrinsic` is not a valid utf8 string!")?;
```

PR #5101, opened 2020-03-02
Use function trait objects for exchangeable host functions

coriolinus:
**Motivation**

The current state of exchangeable functions is that they all expect
raw function pointers: they're initialized with raw pointers to the
initial function, which means that every replacement must also be
a raw function pointer.

The problem with this arrangement is that this prevents us from
swapping in a closure. In Cumulus, we've discovered a case for which
it would be very convenient to swap in a closure which captures an
additional piece of data. This change enables this possibility.

**Drawbacks**

Function calls to a trait object pass through an additional layer
of indirection, which produces a negligable but non-zero additional
cost per function call.

**Breaking Changes**

Every call `foo.replace_implementation(new_impl)` must change to
`foo.replace_implementation(&new_impl)`, which is a simple but
possibly pervasive change.

### DIFF
--- a/primitives/runtime-interface/proc-macro/src/runtime_interface/host_function_interface.rs
+++ b/primitives/runtime-interface/proc-macro/src/runtime_interface/host_function_interface.rs
@@ -147,8 +147,8 @@ fn generate_exchangeable_host_function(method: &TraitItemMethod) -> Result<Token
 			#[allow(non_upper_case_globals)]
 			#[doc = #doc_string]
 			pub static #exchangeable_function : #crate_::wasm::ExchangeableFunction<
-				fn ( #( #arg_types ),* ) #output
-			> = #crate_::wasm::ExchangeableFunction::new(extern_host_function_impls::#function);
+				&dyn Fn ( #( #arg_types ),* ) #output
+			> = #crate_::wasm::ExchangeableFunction::new(&extern_host_function_impls::#function);
 		}
 	)
 }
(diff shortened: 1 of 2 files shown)


coriolinus:
As I understand it, storage calls are expensive; we need at least one storage write to inject the data we want, and then one read per `host_storage_set` call. If those calls are in fact expensive, than they will dwarf the cost of indirection of the function pointer.

bkchr:
I don't understand what you are talking about. Instead of having your closure, you just add the logic here: https://github.com/paritytech/cumulus/blob/master/runtime/src/validate_block/implementation.rs#L147

coriolinus:
Here's how it looks [with a closure](https://github.com/paritytech/cumulus/commit/f4f350929c6df679ce82d1e13e1eb5d3fa1432bf?diff=split) and [without a closure](https://github.com/paritytech/cumulus/commit/827b5cbeade0bbef5644bfc50ec8a2678c2e9cd2?diff=split). 

I'd rather not add a special exception to `WitnessStorage::(get|set)` for `CURRENT_BLOCK_KEY`, because that feels much too magical; it wouldn't implement the expected semantics for those methods anymore.

bkchr:
What is the current block? The current relay chain block?

Why isn't this just a field in the storage struct? I don't understand why you add this to the witness storage. `Storage` can also just be renamed to `State` or something similar.

bkchr:
Adding this to the witness storage will also break the storage root.

coriolinus:
Closing this to rework the cumulus branch instead.

PR #5103, opened 2020-03-02
Split ext_clear_storage out from ext_set_storage

Robbepop:
This PR splits the
```
ext_set_storage(key_ptr: u32, value_non_null: u32, value_ptr: u32, value_len: u32);
```
into two APIs:
```
ext_set_storage(key_ptr: u32, value_ptr: u32, value_len: u32);
ext_clear_storage(key_ptr: u32);
```

## Motivation

We perform this change to make it simpler for static analyzers to analyze the underlying code semantically.

The old API with its `value_non_null` parameter is theoretically impossible to analyze statically since the `value_non_null` parameter could be any runtime computed value. So it is potentially very hard for static analyzers to apply the correct gas costs in the case where we want to differentiate between setting and clearing contract storage entries.

**Why do we want to differentiate between those two operations?**

The operations have completely different semantics:
- The one API is updating a value of a storage entry,
- the other is clearing the value of a storage entry.

Also as described above this makes it far simpler for static analyzers (or in some cases even possible) to reason about the underlying code. It slightly reduces overhead in some situations but that should generally be negligible.

## Open Questions

While working on this PR one particular question arosed:

How do we currently treat empty storage entries? E.g. when setting `ext_set_storage` with a `value_len` of 0? This semantically is different from having a cleaned-up contract storage entry and thus should be taken into account for rental fees.

Since we were not aware of any smart contract use cases that could even profit from creating or having occupied empty storage entries we also disallowed having them in the adjusted `ext_set_storage`. Upon trying to set a contract storage entry to be empty this function will trigger a host error and the contract will trap.

We might change the semantics of this particular scenario once we are aware of a proper smart contracts use case.

## Task List of this PR

- [x] Introduced `ext_clear_storage`
- [x] Adjusted semantics of `ext_set_storage`
- [x] Adjusted `COMPLEXITY.md`
- [x] Adjusted SEAL tests

### DIFF
--- a/frame/contracts/src/wasm/runtime.rs
+++ b/frame/contracts/src/wasm/runtime.rs
@@ -348,30 +348,49 @@ define_env!(Env, <E: Ext>,
 		Ok(())
 	},
 
-	// Change the value at the given key in the storage or remove the entry.
-	// The value length must not exceed the maximum defined by the Contracts module parameters.
+	// Set the value at the given key in the contract storage.
 	//
-	// - key_ptr: pointer into the linear
-	//   memory where the location of the requested value is placed.
-	// - value_non_null: if set to 0, then the entry
-	//   at the given location will be removed.
-	// - value_ptr: pointer into the linear memory
-	//   where the value to set is placed. If `value_non_null` is set to 0, then this parameter is ignored.
-	// - value_len: the length of the value. If `value_non_null` is set to 0, then this parameter is ignored.
-	ext_set_storage(ctx, key_ptr: u32, value_non_null: u32, value_ptr: u32, value_len: u32) => {
-		if value_non_null != 0 && ctx.ext.max_value_size() < value_len {
+	// The value length must not exceed the maximum defined by the contracts module parameters.
+	// Storing an empty value is disallowed.
+	//
+	// # Parameters
+	//
+	// - `key_ptr`: pointer into the linear memory where the location to store the value is placed.
+	// - `value_ptr`: pointer into the linear memory where the value to set is placed.
+	// - `value_len`: the length of the value in bytes.
+	//
+	// # Errors
+	//
+	// - If value length exceeds the configured maximum value length of a storage entry.
+	// - Upon trying to set an empty storage entry (value length is 0).
+	ext_set_storage(ctx, key_ptr: u32, value_ptr: u32, value_len: u32) => {
+		if value_len > ctx.ext.max_value_size() {
+			// Bail out if value length exceeds the set maximum value size.
+			return Err(sp_sandbox::HostError);
+		}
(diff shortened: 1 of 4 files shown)


frame/contracts/src/wasm/runtime.rs
```diff
@@ -348,30 +348,49 @@ define_env!(Env, <E: Ext>,
...
+	//
+	// - If value length exceeds the configured maximum value length of a storage entry.
+	// - Upon trying to set an empty storage entry (value length is 0).
+	ext_set_storage(ctx, key_ptr: u32, value_ptr: u32, value_len: u32) => {
+		if value_len > ctx.ext.max_value_size() {
+			// Bail out if value length exceeds the set maximum value size.
+			return Err(sp_sandbox::HostError);
+		}
+		if value_len == 0 {
+			// Bail out on setting storage to an emptry entry.
```
pepyakin:
```suggestion
			// Bail out on setting storage to an empty entry.
```

frame/contracts/src/wasm/runtime.rs
```diff
@@ -348,30 +348,49 @@ define_env!(Env, <E: Ext>,
...
+	// # Errors
+	//
+	// - If value length exceeds the configured maximum value length of a storage entry.
+	// - Upon trying to set an empty storage entry (value length is 0).
+	ext_set_storage(ctx, key_ptr: u32, value_ptr: u32, value_len: u32) => {
+		if value_len > ctx.ext.max_value_size() {
+			// Bail out if value length exceeds the set maximum value size.
+			return Err(sp_sandbox::HostError);
+		}
+		if value_len == 0 {
```
pepyakin: Let's defer this to #5107

gavofyork:
The proposition that this PR is based on seems flawed. Storage rent should be all the incentivisation needed to ensure that storage clean-ups occur. Trying to warp the prices away from the true costs for the storage operations in order to try to incentivise cleaning up or disincentivise wasteful usage seems accordingly ill-considered.

Robbepop:
> The proposition that this PR is based on seems flawed. Storage rent should be all the incentivisation needed to ensure that storage clean-ups occur. Trying to warp the prices away from the true costs for the storage operations in order to try to incentivise cleaning up or disincentivise wasteful usage seems accordingly ill-considered.

You are right that the original proposition was flawed since we would never decouple gas prices of operations from their real performance impact on the system which the original proposition was based on. - Unfortunately I forgot to update after we had internal discussions about this. However, the code changes no longer reflect the original proposition.
However, there are other reasons for splitting this API up into 2 functions because they are semantically doing completely different things (free and set) and by having this split we can more easily analyze semantics of the underlying code statically.

PR #5104, opened 2020-03-02
Make sure we remove a peer on disconnect in gossip

bkchr:


client/network/src/protocol/sync/blocks.rs
```diff
@@ -165,20 +164,19 @@ impl<B: BlockT> BlockCollection<B> {
...
-					},
-					_ => break,
-				}
+
+		let mut prev = from;
+		for (start, range_data) in &mut self.blocks {
+			match range_data {
+				&mut BlockRangeState::Complete(ref mut blocks) if *start <= prev => {
+						prev = *start + (blocks.len() as u32).into();
+						drained.append(blocks);
```
tomaka: `append` isn't a very descriptive name, and I had to look up the documentation to check whether this was draining the content of `blocks`. Maybe a small comment mentioning that wouldn't hurt.

bkchr: Done.

arkpar: Documentation says `Moves all the elements of other into Self, leaving other empty.` but it does not clarify if "empty" means memory is deallocated.  `mem::replace` makes sure the memory is freed when `blocks` is dropped`

bkchr: Yeah, but this entry is going to be removed from the hashmap anyway. That will drop the vec.

PR #5108, opened 2020-03-02
Forget peerset manager nodes after 5mn

tomaka:
Fixes a small leak in the peerset manager.

When a node is in the "not connected" state, we additional store the last time we've been connected to it (or, if we've never been connected, when we discovered it).

If 300 seconds have elapsed, and the node's reputation is 0, we remove the node from the list.


tomaka:
I bumped the delay to one hour, and also made it so we update the value every time a node is discovered.

The discovery process discovers node randomly, so the delay has to be long enough for us to have a chance to randomly re-discover nodes we're not connected to.

PR #5109, opened 2020-03-02
Better wasm instance cache

arkpar:
Use fixed size pool for Wasm instances, rather than TLS based cache that grows indefinitely and depends on thread  context.

Upstream PR for wasmtime: https://github.com/bytecodealliance/wasmtime/pull/1227


client/executor/src/wasm_runtime.rs
```diff
@@ -39,15 +41,33 @@ pub enum WasmExecutionMethod {
...
-	/// Runtime version according to `Core_version`.
-	version: RuntimeVersion,
+	/// Runtime version according to `Core_version` if any.
+	version: Option<RuntimeVersion>,
+	/// Cached instance pool.
+	instances: RwLock<[Option<Arc<Mutex<Box<dyn WasmInstance>>>>; MAX_INSTANCES]>,
 }
 
+const MAX_RUNTIMES: usize = 2;
+const MAX_INSTANCES: usize = 8;
```
tomusdrw: Might make sense to expose as CLI params in subsequent PR.

client/executor/src/wasm_runtime.rs
```diff
@@ -104,87 +124,127 @@ impl RuntimesCache {
...
+				drop(instance);
+				let instance = runtime.runtime.new_instance()?;
+
+				let result = f(&*instance, runtime.version.as_ref(), ext);
+				let mut instance_pool = runtime.instances.write();
+				if let Some(ref mut slot) = instance_pool.iter_mut().find(|s| s.is_none()) {
+					**slot = Some(Arc::new(Mutex::new(instance)));
+					log::debug!(target: "wasm-cache", "Allocated WASM instance {}/{}", instance_pool.len(), MAX_INSTANCES);
+				} else {
+					log::warn!(target: "wasm-cache", "Ran out of free WASM instances");
```
tomusdrw:
```suggestion
					log::warn!(target: "wasm-cache", "Ran out of cacheable WASM instances");
```
?

client/executor/src/wasm_runtime.rs
```diff
@@ -104,87 +124,127 @@ impl RuntimesCache {
...
+		Ok(result)
+	}
+	/*
 	/// Invalidate the runtime for the given `wasm_method` and `code_hash`.
 	///
 	/// Invalidation of a runtime is useful when there was a `panic!` in native while executing it.
 	/// The `panic!` maybe have brought the runtime into a poisoned state and so, it is better to
 	/// invalidate this runtime instance.
-	pub fn invalidate_runtime(
+	fn invalidate_runtime(
```
tomusdrw: To remove.

client/executor/src/wasm_runtime.rs
```diff
@@ -224,21 +291,31 @@ fn create_versioned_wasm_runtime<E: Externalities>(
...
+		time.elapsed().as_millis()
+	);
 
 	Ok(VersionedRuntime {
+		code_hash,
 		runtime,
 		version,
 		heap_pages,
+		wasm_method,
+		instances: RwLock::new([None, None, None, None, None, None, None, None]),
```
tomusdrw: no `Default`?

client/executor/src/wasm_runtime.rs
```diff
@@ -104,87 +124,127 @@ impl RuntimesCache {
...
+
+		let result = {
+			// Find a free instance
+			let instance_pool = { runtime.instances.read().clone() };
+			let instance = instance_pool.iter().filter_map(|i| i.as_ref().and_then(|i| i.try_lock())).next();
+			if let Some(locked) = instance {
+				f(&**locked, runtime.version.as_ref(), ext)
+			} else {
+				// Allocate a new instance
+				drop(instance);
```
tomusdrw: This is needed to drop `instance_pool` borrow, right?

client/executor/wasmi/src/lib.rs
```diff
@@ -624,30 +624,96 @@ impl StateSnapshot {
...
+	module: Module,
+	/// The host functions registered for this instance.
+	host_functions: Arc<Vec<&'static dyn Function>>,
+	/// Enable stub generation for functions that are not available in `host_functions`.
+	/// These stubs will error when the wasm blob tries to call them.
+	allow_missing_func_imports: bool,
+	/// Numer of heap pages this runtime uses.
+	heap_pages: u64, 
+	/// Data segments created for each new instance.
+	data_segments: Vec<DataSegment>,
```
tomusdrw: Could this be shared via `Arc` as well? Or do we actually alter it?

client/executor/wasmtime/Cargo.toml
```diff
@@ -10,16 +10,15 @@ description = "Defines a `WasmRuntime` that uses the Wasmtime JIT to execute."
...
 parity-wasm = "0.41.0"
 codec = { package = "parity-scale-codec", version = "1.0.0" }
 sc-executor-common = { version = "0.8.0-alpha.2", path = "../common" }
 sp-wasm-interface = { version = "2.0.0-alpha.2", path = "../../../primitives/wasm-interface" }
 sp-runtime-interface = { version = "2.0.0-alpha.2", path = "../../../primitives/runtime-interface" }
 sp-core = { version = "2.0.0-alpha.2", path = "../../../primitives/core" }
 sp-allocator = { version = "2.0.0-alpha.2", path = "../../../primitives/allocator" }
-
-wasmtime = "0.11"
+wasmtime = { git = "https://github.com/paritytech/wasmtime", branch = "a-thread-safe-api" }
```
NikVolf: this will prevents publishing on crates.io (though not sure if it is a bad thing) for a while

client/executor/src/wasm_runtime.rs
arkpar: I believe it is no longer needed after a recent commit where I've reworked pool access.

client/executor/wasmi/src/lib.rs
arkpar: I'm fairly sure each instance needs to have it's own copy of the data. Right @pepyakin ?

pepyakin: I haven't made myself acquainted with the code yet, so here is my general answer:
`data_segments` are extracted from the corresponding `Module` and, like `Module`, doesn't change for a particular version of code.

client/executor/src/wasm_runtime.rs
arkpar: I've used "available" instead. Referred instances are cached already.

client/executor/wasmi/src/lib.rs
arkpar: As far as I can see each instantiation of `StateSnapshot` requires an instance of data segments. And each `WasmiInstance` requires a `StateSnapshot`

client/executor/src/native_executor.rs
```diff
@@ -176,87 +212,99 @@ impl<D: NativeExecutionDispatch> RuntimeInfo for NativeExecutor<D> {
...
+			>(
+				&self,
+				ext: &mut dyn Externalities,
+				method: &str,
+				data: &[u8],
+				use_native: bool,
+				native_call: Option<NC>,
+			) -> (Result<NativeOrEncoded<R>>, bool){
+				let mut used_native = false;
+				let result = self.wasm.with_instance(CodeSource::Externalities, ext, |instance, onchain_version, mut ext| {
```
pepyakin: Many lines in this file exceed, unnecessarily IMO, the soft limit of 100 characters. This line specifically exceeds the limit of 120. Can we wrap those?

client/executor/common/src/wasm_runtime.rs
```diff
@@ -17,18 +17,25 @@
...
 /// A trait that defines an abstract wasm runtime.
 ///
 /// This can be implemented by an execution engine.
-pub trait WasmRuntime {
-	/// Return the host functions that are registered for this Wasm runtime.
-	fn host_functions(&self) -> &[&'static dyn Function];
-
-	/// Call a method in the Substrate runtime by name. Returns the encoded result on success.
-	fn call(&mut self, method: &str, data: &[u8]) -> Result<Vec<u8>, Error>;
+pub trait WasmRuntime: Sync + Send {
```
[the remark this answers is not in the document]
arkpar: How about simple `WasmModule`?

client/executor/common/src/wasm_runtime.rs
```diff
@@ -17,18 +17,25 @@
...
-	fn get_global_val(&self, name: &str) -> Result<Option<Value>, Error>;
+/// A trait that defines an abstract wasm machine instance.
+///
+/// This can be implemented by an execution engine.
+pub trait WasmInstance: Send {
+	/// Call a method in the wasm code and reset this instance.
+	/// Returns the encoded result on success.
+	fn call(&self, method: &str, data: &[u8]) -> Result<Vec<u8>, Error>;
+
+	/// Get the value from a global with the given `name` on a fresh instance.
```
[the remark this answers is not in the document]
arkpar: I'd remove this altogether eventually, since it is only used in tests.

client/executor/wasmtime/Cargo.toml
gnunicorn: yes, it will prevent us from publishing this on crates.io . Current plan is to get this rolled out on kusama, confirm the findings, get upstream merged and switch back to the newly released version.

client/executor/src/wasm_runtime.rs
```diff
@@ -39,15 +41,33 @@ pub enum WasmExecutionMethod {
...
+pub enum CodeSource<'a> {
+	/// Take code from storage,
+	Externalities,
+	/// Use provided code,
+	Custom(&'a [u8]),
+}
+
 /// A Wasm runtime object along with its cached runtime version.
 struct VersionedRuntime {
+	/// Code hash.
```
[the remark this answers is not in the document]
arkpar: Sounds tautological. I'll put "Runtime code hash"

client/executor/src/wasm_runtime.rs
```diff
@@ -39,15 +41,33 @@ pub enum WasmExecutionMethod {
...
+	code_hash: Vec<u8>,
+	/// Wasm runtime type.
```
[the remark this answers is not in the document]
arkpar: Which is effectively the runtime type, isn't it?

client/executor/src/wasm_runtime.rs
```diff
@@ -104,86 +124,122 @@ impl RuntimesCache {
...
-					let heap_pages_changed = cached_runtime.heap_pages != heap_pages;
-					let host_functions_changed = cached_runtime.runtime.host_functions()
-						!= host_functions;
-					if heap_pages_changed || host_functions_changed {
-						let changed = if heap_pages_changed {
-							"heap_pages"
-						} else {
-							"host functions"
-						};
+		let runtimes = self.runtimes.upgradable_read(); // this must be released prior to calling f
```
bkchr: Why an upgradable read lock? As far as I can see, we never lock this for reading? 

`upgradable_read()` can only be hold by one thread at a time, like a write lock. So, we can use a write lock directly.

arkpar: Right, I'll replace it with a `Mutex`. It is only used here and unlocked for the actual call anyway.

arkpar:
All issues addressed

client/executor/wasmtime/src/state_holder.rs
```diff
@@ -15,63 +15,29 @@
...
 use crate::host::{HostContext, HostState};
-use std::cell::RefCell;
-use std::rc::Rc;
 
-/// A common place to store a reference to the `HostState`.
-///
-/// This structure is passed into each host function handler and retained in the implementation of
-/// `WasmRuntime`. Whenever a call into a runtime method is initiated, the host state is populated
-/// with the state for that runtime method call.
+scoped_tls::scoped_thread_local!(static HOST_STATE: HostState);
```
bkchr: You could have used the environmental crate for this.

This could also be problematic in the future when we wasmtime use for sandboxing for example. 

In a perfect world, we would pass the pointer to the host state as parameter to the functions, but yeah not sure if that works.

client/executor/wasmtime/src/runtime.rs
```diff
@@ -15,57 +15,85 @@
...
+}
 
-	fn call(&mut self, method: &str, data: &[u8]) -> Result<Vec<u8>> {
+// This is safe because `WasmtimeInstance` does not leak reference to `self.imports`
+// and all imports don't reference any anything, other than host functions and memory
+unsafe impl Send for WasmtimeInstance {}
+
+impl WasmInstance for WasmtimeInstance {
+	fn call(&self, method: &str, data: &[u8]) -> Result<Vec<u8>> {
+		// TODO: reuse the instance and reset globals after call
```
bkchr: We should add an issue for this.

client/executor/wasmtime/src/runtime.rs
```diff
@@ -15,57 +15,85 @@
...
+	fn new_instance(&self) -> Result<Box<dyn WasmInstance>> {
+		// Scan all imports, find the matching host functions, and create stubs that adapt arguments
+		// and results.
+		let imports = resolve_imports(
+			&self.module,
+			&self.host_functions,
+			self.heap_pages,
+			self.allow_missing_func_imports,
+		)?;
+
```
bkchr:
```suggestion
```

client/executor/src/wasm_runtime.rs
```diff
@@ -194,28 +268,43 @@ pub fn create_wasm_runtime_with_code(
...
-		.original_storage(well_known_keys::CODE)
-		.ok_or(WasmError::CodeNotFound)?;
-	let mut runtime = create_wasm_runtime_with_code(wasm_method, heap_pages, &code, host_functions, false)?;
+	let time = std::time::Instant::now();
+	let mut runtime = create_wasm_runtime_with_code(
+		wasm_method,
+		heap_pages,
+		&code,
+		host_functions,
+		allow_missing_func_imports
```
bkchr:
```suggestion
		allow_missing_func_imports,
```

client/executor/src/wasm_runtime.rs
```diff
@@ -224,21 +313,33 @@ fn create_versioned_wasm_runtime<E: Externalities>(
...
+			.map_err(|_|
+				WasmError::Instantiation("failed to decode \"Core_version\" result".into())
+			)?),
+		Err(_) => None,
+	};
+	log::debug!(
+		target: "wasm-runtime",
+		"Prepared new runtime version {:?} in {} ms.",
+		version,
+		time.elapsed().as_millis()
```
bkchr:
```suggestion
		time.elapsed().as_millis(),
```

client/executor/src/wasm_runtime.rs
```diff
@@ -60,130 +80,184 @@ struct VersionedRuntime {
...
+				match &result {
+					Ok(_) => {
+						let mut instance_pool = runtime.instances.write();
+						if let Some(ref mut slot) = instance_pool.iter_mut().find(|s| s.is_none()) {
+							**slot = Some(Arc::new(Mutex::new(instance)));
+							log::debug!(
+								target: "wasm-runtime",
+								"Allocated WASM instance {}/{}",
+								instance_pool.len(),
+								MAX_INSTANCES
```
bkchr:
```suggestion
								MAX_INSTANCES,
```

client/executor/src/wasm_runtime.rs
```diff
@@ -60,130 +80,184 @@ struct VersionedRuntime {
...
+						} else {
+							log::warn!(target: "wasm-runtime", "Ran out of free WASM instances");
+						}
+					}
+					Err(e) => {
+						log::warn!(
+							target:
+							"wasm-runtime",
+							"Fresh runtime instance failed with {:?}",
+							e
```
bkchr:
```suggestion
							e,
```

client/executor/src/native_executor.rs
```diff
@@ -127,46 +124,90 @@ impl<D: NativeExecutionDispatch> NativeExecutor<D> {
...
+	pub fn new(fallback_method: WasmExecutionMethod, default_heap_pages: Option<u64>) -> Self {
+		let mut host_functions = sp_io::SubstrateHostFunctions::host_functions();
+
+		// Add the custom host functions provided by the user.
+		host_functions.extend(D::ExtendHostFunctions::host_functions());
+		let wasm_executor = WasmExecutor::new(
+			fallback_method,
+			default_heap_pages,
+			host_functions,
+			false
```
bkchr:
```suggestion
			false,
```

client/executor/src/native_executor.rs
```diff
@@ -127,46 +124,90 @@ impl<D: NativeExecutionDispatch> NativeExecutor<D> {
...
+		&self,
+		wasm_blob: &[u8],
+		method: &str,
+		call_data: &[u8],
+		ext: &mut dyn Externalities,
+	) -> std::result::Result<Vec<u8>, String> {
+		self.with_instance(CodeSource::Custom(wasm_blob), ext, |instance, _, mut ext| {
+			with_externalities_safe(
+				&mut **ext,
+				move || instance.call(method, call_data)
```
bkchr:
```suggestion
				move || instance.call(method, call_data),
```

client/executor/src/native_executor.rs
```diff
@@ -176,87 +217,106 @@ impl<D: NativeExecutionDispatch> RuntimeInfo for NativeExecutor<D> {
...
-					used_native = true;
-					Ok(D::dispatch(&mut **ext, method, data).map(NativeOrEncoded::Encoded))
-				}
-			}
-		});
-		(result, used_native)
+		<
+			R: Decode + Encode + PartialEq,
+			NC: FnOnce() -> result::Result<R, String> + UnwindSafe,
+			>(
```
bkchr: I think all the new lines from here have one tab too much indentation.

client/executor/src/wasm_runtime.rs
```diff
@@ -60,130 +80,184 @@ struct VersionedRuntime {
...
+		let (code_hash, heap_pages) = match &code {
+			CodeSource::Externalities => {
+				(
+					ext
+						.original_storage_hash(well_known_keys::CODE)
+						.ok_or(Error::InvalidCode("`CODE` not found in storage.".into()))?,
+					ext
+						.storage(well_known_keys::HEAP_PAGES)
+						.and_then(|pages| u64::decode(&mut &pages[..]).ok())
+						.unwrap_or(default_heap_pages)
```
bkchr:
```suggestion
						.unwrap_or(default_heap_pages),
```

client/executor/wasmtime/src/state_holder.rs
arkpar: > In a perfect world, we would pass the pointer to the host state as parameter to the functions, but yeah not sure if that works.

This requires changes to wasmtime, again.

> You could have used the environmental crate for this.
Will do in a separate PR

bkchr: Yeah both of this stuff was more me writing down some considerations. Nothing that should be done now :)

arkpar: Turns out `environmental` also uses TLS under the hood, so I don't see much difference. At least with `scoped_thread_local!` you can check if the value has been set.

PR #5116, opened 2020-03-03
Ensure storage is upgraded for staking module.

thiolliere:
I think I forgot to add some ensure_storage_upgraded for some public function.


thiolliere:
until we have a better solution to do storage upgrade, we can also not support it.

PR #5120, opened 2020-03-03
Pass Prometheus `Registry` into `Client`

expenses:
We're probably going to want to instrument `Client` in the future, so this PR contains the refactorings needed to pass it the prometheus `Registry`.


client/service/src/builder.rs
```diff
@@ -92,6 +92,16 @@ impl ServiceMetrics {
 	}
 }
 
+fn get_registry_and_port(config: Option<&PrometheusConfig>) -> Result<Option<(Registry, SocketAddr)>, PrometheusError> {
+	Ok(match config {
+		Some(config) => match config.registry.as_ref() {
+			Some(registry) => Some((registry.clone(), config.port)),
+			None => Some((Registry::new_custom(Some("substrate".into()), None)?, config.port))
```
NikVolf:
```suggestion
			None => Some((Registry::new_custom(Some("substrate".into()), None)?, config.port)),
```

client/service/src/builder.rs
```diff
@@ -128,7 +138,7 @@ pub struct ServiceBuilder<TBl, TRtApi, TGen, TCSExt, TCl, TFchr, TSc, TImpQu, TF
 	remote_backend: Option<Arc<dyn RemoteBlockchain<TBl>>>,
 	marker: PhantomData<(TBl, TRtApi)>,
 	background_tasks: Vec<(&'static str, BackgroundTask)>,
-	prometheus_registry: Option<Registry>
+	prometheus_registry_and_port: Option<(Registry, SocketAddr)>,
```
NikVolf: I'd suggest adding separate data structure for this, `PrometheusLink` or similar

NikVolf: which already exists, `PrometheusConfig` ?

expenses: Done.

PR #5125, opened 2020-03-04
Adds a session getter to historical proofs

montekki:
Adds a session getter for `session::historical::Proof`. It is needed for offence handlers such as paritytech/polkadot#840 to determine the exact session of the commited equivocation and create the corresponding `Offence` struct.


rphmeier:
cc @gnunicorn this is technically a breaking change, should we wait to merge? Something that's definitely needed for slashing.

gnunicorn:
we are still in alpha-release-phase, we are accepting not-too-involved breaking changes.

montekki:
Also added some trait derives that are necessary for using the `Proof` type inside runtime.

gnunicorn:
failing test, merge conflict.

montekki:
@gnunicorn can we now get this in?

PR #5126, opened 2020-03-04
Add lots of networking metrics for Prometheus

tomaka:
I prefixed all the existing metrics with `sub_libp2p` to match the logging prefix, and added lots of new ones.

I suggest the following alerts to put in place: (I have no idea how to do that though)

- `max(rate(substrate_sub_libp2p_notifications[2m])) >= 500` (we're getting flooded; there's a bug somewhere or we're being DoS-attacked)
- `substrate_sub_libp2p_connections == 0` (no Internet connectivity)
- `substrate_sub_libp2p_peers_count == 0` (no Internet connectivity or can't handshake with nodes)
- `substrate_sub_libp2p_num_event_stream_channels >= 30` (there's a leak somewhere in the code)


client/network/src/service.rs
```diff
@@ -733,25 +733,116 @@ pub struct NetworkWorker<B: BlockT + 'static, H: ExHashT> {
...
+	peerset_num_requested: Gauge<U64>,
+	random_kademalia_queries: Counter<U64>,
 }
 
 impl Metrics {
 	fn register(registry: &Registry) -> Result<Self, PrometheusError> {
 		Ok(Self {
+			// This list is ordered alphabetically
+			average_download_per_sec: register(Gauge::new(
+				"sub_libp2p_average_download_per_sec", "Average number of bytes downloaded per second",
```
mxinden: We already prefix every metric with `substrate` here: https://github.com/paritytech/substrate/blob/master/client/service/src/builder.rs#L894

Would just `libp2p` be enough?

client/network/src/service.rs
```diff
@@ -733,25 +733,116 @@ pub struct NetworkWorker<B: BlockT + 'static, H: ExHashT> {
...
+			)?, registry)?,
+			average_upload_per_sec: register(Gauge::new(
+				"sub_libp2p_average_upload_per_sec", "Average number of bytes uploaded per second",
+			)?, registry)?,
+			import_queue_block_submits: register(Counter::new(
+				"import_queue_block_submits",
```
mxinden:
```suggestion
				"import_queue_blocks_submitted",
```
Sounds better to me. What do you think?

client/network/src/service.rs
```diff
@@ -733,25 +733,116 @@ pub struct NetworkWorker<B: BlockT + 'static, H: ExHashT> {
...
+			import_queue_justification_submits: register(Counter::new(
+				"import_queue_justification_submits",
+				"Number of justifications submitted to the import queue.",
+			)?, registry)?,
 			is_major_syncing: register(Gauge::new(
-				"is_major_syncing", "Whether the node is performing a major sync or not.",
+				"sub_libp2p_is_major_syncing", "Whether the node is performing a major sync or not.",
+			)?, registry)?,
+			libp2p_connections: register(Gauge::new(
+				"sub_libp2p_libp2p_connections", "Number of libp2p connections to a given node"
```
mxinden:
```suggestion
				"sub_libp2p_connections", "Number of libp2p connections to a given node"
```

mxinden:
```suggestion
				"sub_libp2p_libp2p_connections", "Number of libp2p connections"
```

As far as I can tell this is generic over the peer id, right?

client/network/src/service.rs
```diff
@@ -733,25 +733,116 @@ pub struct NetworkWorker<B: BlockT + 'static, H: ExHashT> {
...
+			)?, registry)?,
+			kbuckets_num_nodes: register(Gauge::new(
+				"sub_libp2p_kbuckets_num_nodes", "Number of nodes in the Kademlia k-buckets"
+			)?, registry)?,
+			notifications_in: register(CounterVec::new(
+				Opts::new(
+					"sub_libp2p_notifications_in",
```
mxinden:
```suggestion
					"sub_libp2p_notifications_in_total",
```

This is an ever increasing counter, right?

client/network/src/service.rs
```diff
@@ -733,25 +733,116 @@ pub struct NetworkWorker<B: BlockT + 'static, H: ExHashT> {
...
+					"Number of notification received from all nodes"
+				),
+				&["protocol"]
+			)?, registry)?,
+			notifications_out: register(CounterVec::new(
+				Opts::new(
+					"sub_libp2p_notifications_out",
```
mxinden:
```suggestion
					"sub_libp2p_notifications_out_total",
```

client/network/src/service.rs
```diff
@@ -733,25 +733,116 @@ pub struct NetworkWorker<B: BlockT + 'static, H: ExHashT> {
...
+				&["protocol"]
+			)?, registry)?,
+			num_event_stream_channels: register(Gauge::new(
+				"sub_libp2p_num_event_stream_channels",
+				"Number of internal active channels that broadcast network events",
+			)?, registry)?,
+			opened_notification_streams: register(GaugeVec::new(
+				Opts::new(
+					"sub_libp2p_opened_notification_streams",
+					"Number of notification substreams opens with a given node"
```
mxinden:
```suggestion
					"Number of notification substreams opened with a given node"
```

Is this generic over the peer id? If not, it should be.

client/network/src/service.rs
```diff
@@ -733,25 +733,116 @@ pub struct NetworkWorker<B: BlockT + 'static, H: ExHashT> {
...
+				"sub_libp2p_peers_count", "Number of network gossip peers",
+			)?, registry)?,
+			peerset_num_discovered: register(Gauge::new(
+				"sub_libp2p_peerset_num_discovered", "Number of nodes stored in the peerset manager",
+			)?, registry)?,
+			peerset_num_requested: register(Gauge::new(
+				"sub_libp2p_peerset_num_requested", "Number of nodes that the peerset manager wants us to be connected to",
+			)?, registry)?,
+			random_kademalia_queries: register(Counter::new(
+				"sub_libp2p_random_kademalia_queries", "Number of random Kademlia queries started",
```
mxinden:
```suggestion
				"sub_libp2p_random_kademalia_queries_total", "Number of random Kademlia queries started",
```

client/network/src/service.rs
mxinden: For accumulating counters it is a convention to postfix them with `_total`. See https://prometheus.io/docs/practices/naming/ for more details.

client/network/src/service.rs
```diff
@@ -733,25 +733,116 @@ pub struct NetworkWorker<B: BlockT + 'static, H: ExHashT> {
...
+			)?, registry)?,
 			is_major_syncing: register(Gauge::new(
-				"is_major_syncing", "Whether the node is performing a major sync or not.",
+				"sub_libp2p_is_major_syncing", "Whether the node is performing a major sync or not.",
+			)?, registry)?,
+			libp2p_connections: register(Gauge::new(
+				"sub_libp2p_libp2p_connections", "Number of libp2p connections to a given node"
+			)?, registry)?,
+			kbuckets_num_nodes: register(Gauge::new(
+				"sub_libp2p_kbuckets_num_nodes", "Number of nodes in the Kademlia k-buckets"
```
mxinden: Would it make sense to namespace this with `kad`?

```suggestion
				"sub_libp2p_kad_kbuckets_num_nodes", "Number of nodes in the Kademlia k-buckets"
```

client/network/src/service.rs
```diff
@@ -733,25 +733,116 @@ pub struct NetworkWorker<B: BlockT + 'static, H: ExHashT> {
...
 	event_streams: Vec<mpsc::UnboundedSender<Event>>,
 	/// Prometheus network metrics.
-	metrics: Option<Metrics>
+	metrics: Option<Metrics>,
 }
 
 struct Metrics {
+	// This list is ordered alphabetically
+	average_download_per_sec: Gauge<U64>,
+	average_upload_per_sec: Gauge<U64>,
```
expenses: It might be better to put these two together into a `GaugeVec` and use label values like we do here: https://github.com/paritytech/substrate/blob/c3ff85d13b8b660c2c4eb850209c06a9397bed3e/client/service/src/builder.rs#L1112-L1113

Same could be done for things like import queue submissions. @mxinden what do you think?

mxinden: Yes. Very good point. That makes later querying with PromQL a lot easier.

client/network/src/service.rs
tomaka: Yes

client/network/src/service.rs
tomaka: Yes, it's just the total number of connections

client/network/src/service.rs
tomaka: Ah I forgot to update the description

client/network/src/service.rs
tomaka: I used `sub_libp2p` so that it matches the logging target (i.e. we do `log::warn!(target: "sub-libp2p", "...")` everywhere.

client/network/src/service.rs
tomaka: I went without, as it seems redundant

tomaka:
For what it's worth I'll add more metrics in further PRs.

client/network/src/service.rs
```diff
@@ -733,25 +733,108 @@ pub struct NetworkWorker<B: BlockT + 'static, H: ExHashT> {
...
-	metrics: Option<Metrics>
+	metrics: Option<Metrics>,
 }
 
 struct Metrics {
+	// This list is ordered alphabetically
+	connections: Gauge<U64>,
+	import_queue_blocks_submitted: Counter<U64>,
+	import_queue_finality_proofs_submitted: Counter<U64>,
+	import_queue_justifications_submitted: Counter<U64>,
```
expenses: I think it might be better to put these all into one `CounterVec<U64>`.

tomaka: I'm not sure about this one, since we're measuring different things. It for example doesn't make sense to make the sum or the average of these three values.

expenses: Hmm, yeah not sure. @mxinden?

gavofyork:
probably needs a master merge to fix the CI. otherwise, @mxinden if you've no objections i would like to merge now.

client/network/src/service.rs
mxinden: Using a `CounterVec` here would ease displaying both finality proofs as well as justifications in the same graph within the Prometheus UI. Once one goes through Grafana it doesn't matter as Grafana can display arbitrary time series within the same graph.

PR #5129, opened 2020-03-04
Remove `sender` from `Applyable`.

pepyakin:
It looks like `Applyable::sender` along with its `Applyable::AccountId` are not used within primitives/FRAME. This PR just removes that.

The runtime behavior is not changed.

### DIFF
--- a/primitives/runtime/src/traits.rs
+++ b/primitives/runtime/src/traits.rs
@@ -876,17 +876,11 @@ impl SignedExtension for () {
 /// Also provides information on to whom this information is attributable and an index that allows
 /// each piece of attributable information to be disambiguated.
 pub trait Applyable: Sized + Send + Sync {
-	/// ID of the account that is responsible for this piece of information (sender).
-	type AccountId: Member + MaybeDisplay;
-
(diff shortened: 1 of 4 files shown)


gavofyork:
This is public API, so theoretically other projects could be depending on it. What's the rationale behind removing it?

pepyakin:
Right, theoretically that's possible. But I am not sure if that is the case since `Applyable` appears in `frame-executive`-like contexts and I would be surprised if somebody write their own. 

If you think that the risk is too high feel free to close the PR.

gavofyork:
Just wondering what the reason is for removing it?

primitives/runtime/src/traits.rs
```diff
@@ -876,18 +876,12 @@ impl SignedExtension for () {
 /// Also provides information on to whom this information is attributable and an index that allows
```
tomusdrw: This comment is not relevant any more.

pepyakin:
@tomusdrw @gavofyork yep, found that while working on that. I've decided to put it into a separate PR to try to sneak in this change while we can so we won't release (as in stable release) that and we don't get users of this API.

PR #5130, opened 2020-03-04
Grand Central Dispatch

pepyakin:
Related to #5032


primitives/runtime/src/testing.rs
```diff
@@ -427,19 +426,20 @@ impl<Origin, Call, Extra, Info> Applyable for TestXt<Call, Extra> where
...
-	fn apply<U: ValidateUnsigned<Call=Self::Call>>(
+	fn apply<U: ValidateUnsigned<Call = Self::Call>, D: crate::traits::Dispatcher<Self::Call>>(
 		self,
 		info: Self::DispatchInfo,
 		len: usize,
 	) -> ApplyExtrinsicResult {
-		let maybe_who = if let Some((who, extra)) = self.signature {
-			Extra::pre_dispatch(extra, &who, &self.call, info, len)?;
-			Some(who)
+		let (maybe_who, pre) = if let Some((who, extra)) = self.signature {
```
tomusdrw: Can we merge that with `fn apply` for `CheckedExtrinsic`?

frame/executive/src/lib.rs
```diff
@@ -100,54 +100,58 @@ pub type CheckedOf<E, C> = <E as Checkable<C>>::Checked;
...
-> Executive<System, Block, Context, UnsignedValidator, AllModules>
+		System: frame_system::Trait,
+		Block: traits::Block<Header = System::Header, Hash = System::Hash>,
+		Context: Default,
+		UnsignedValidator,
+		AllModules: OnInitialize<System::BlockNumber>
+			+ OnFinalize<System::BlockNumber>
+			+ OffchainWorker<System::BlockNumber>
+			+ WeighBlock<System::BlockNumber>,
+		Dispatcher: sp_runtime::traits::Dispatcher<CallOf<Block::Extrinsic, Context>>,
```
tomusdrw:
```suggestion
		Dispatcher: traits::Dispatcher<CallOf<Block::Extrinsic, Context>>,
```
^^ for all occurrences in that file

bin/node/runtime/src/lib.rs
```diff
@@ -159,6 +159,7 @@ impl pallet_utility::Trait for Runtime {
 	type MultisigDepositBase = MultisigDepositBase;
 	type MultisigDepositFactor = MultisigDepositFactor;
 	type MaxSignatories = MaxSignatories;
+	type Dispatcher = frame_system::Module<Runtime>;
```
athei: Why do we need one `Dispatcher` for every module that needs to dispatch. Couldn't we just add it once to the `frame_system::Trait`?

pepyakin: I didn’t put much thought into it TBH since it is the default approach. For example, we could have done the same with Call and others.


Having distinct types would allow customizing logic additionally. For example, you would be able to provide a dispatcher instance that would filter some calls or maybe add some logic. So maybe it is worth it to leave it that way?

athei: To make this possible `SignedExtension::post_dispatch` must be generic over the `Dispatcher` that belongs to the `Call` that we are post dispatching. We need this in order to bound it with something like `D: Dispatcher + WeightRefund`in order to get the spent weight from a generic Dispatcher. Right now in #5032 it hardcodes the system module there.

athei:
Rebased that PR onto the current master and fixed the test build. This was quiet a lot of work because the additional `Dispatchable`trait bound that was added to `frame_system::Trait::Call` broke almost every test build.

The test are successful with the exception of the `ui` test which apparently tests against compiler output. Have to look into that I guess.

bin/node/runtime/src/lib.rs
pepyakin: What I meant not the total customization but rather adding some logic for the downstream modules, like `pallet-contracts`.

pepyakin:
That's good news! I think ui tests could be fixed by merging in master.

bin/node/runtime/src/lib.rs
athei: Let's go with this example. `pallet_contracts` needs some additional logic in it's `Dispatcher`. Therefore the pallet adds some additional bounds onto it's `Dispatcher` type in it's pallet and the runtime impl supplies another `Dispatcher` as for other modules in order to fulfill the bound. This is then used for bookkeeping and all is fine.

**Until** some signed extension like the `ChargeTransactionPayment` needs to get an `Dispatcher` instance to get the unspent gas from. How would it determine which `Dispatcher` to query? It is only generic over the global overarching `Call`type. It does not generate code for every module.

For that reason in #5032 it just hardcodes the used dispatcher in `<ChargeTransactionPayment as SignedExtension>::post_dispatch`:
```
<frame_system::Module<T>>::spent_weight()
```

This won't work when the module that this extrinsic targets used another `Dispatcher` like in our example.

I therefore argue that we shouldn't allow a custom `Dispatcher` per module, because the point is that there is only one (specified by `frame_system::Trait::Dispatcher`.

Or would your custom `pallet_contract::Trait::Dispatcher` call into the `frame_system::Trait::Dispatcher`?

athei:
> That's good news! I think ui tests could be fixed by merging in master again.

It was just the missing documentation warning that made the UI test fail. Fixed that now. CI is finally happy and we are in sync with the current master.

athei:
> Would be interesting to see how this glues to the rest by
> 
>     * putting the docs there for the main trait
> 
>     * finishing bode/runtime/src/lib.rs

I added some documentation for the trait. Regarding your second point. Can you elaborate what is missing from that file?

primitives/runtime/src/testing.rs
athei: I de-duplicated this by factoring out `apply` to its own function. Alternative would be a blanket impl but I think that would have been even more ugly. Maybe you can have a look.

bin/node/runtime/src/lib.rs
athei: Summary a discussion with @pepyakin is that it should be implemented in the following way: 

The `SignedExtension` will always call into the "main dispatcher" which is determined by `frame_system::Trait::Dispatcher`. When runtime implementors decide to use a different `Dispatcher` for some of their modules these "sub dispatchers" must always delegate to the "main dispatcher". The `SignedExtension` therefore only deals with one `Dispatcher` implementation and does not need to be generic in that regard.

This delegation contract also must be explained in the `Dispatcher` rustdoc.

athei: I implemented the discussed changes as a `MainDispatcher` type alias on the system trait to make it more obvious that it has a special meaning and is the one that custom dispatchers for individual modules should feed into.

athei:
This PR is now ready for review. I think I addressed all comments and now are patiently waiting for further comments / review.

frame/contracts/src/lib.rs
```diff
@@ -367,6 +369,9 @@ pub trait Trait: frame_system::Trait {
 	/// trie id generator
 	type TrieIdGenerator: TrieIdGenerator<Self::AccountId>;
 
+	/// The means of dispatching the calls to the runtime.
+	type Dispatcher: Dispatcher<<Self as Trait>::Call>;
```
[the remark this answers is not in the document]
pepyakin: See the discussion [there](https://github.com/paritytech/substrate/pull/5130#discussion_r392571834). 

TL;DR: I started it without thinking too much (since we doo the same with all other things) and then I found some benefits, namely using that as an escape hatch. I.e. the chain author can implement their own dispatcher that e.g. filters some calls / does additional bookkeeping and then delegates control to the system module. @athei 's idea was to actually introduce `MainDispatcher` to avoid hard-coding `frame_system::Module` directly but to just delegate to `frame_system::Trait::MainDispatcher`.

bin/node-template/pallets/template/src/mock.rs
```diff
@@ -42,11 +48,13 @@ impl system::Trait for Test {
 	type AccountData = ();
 	type OnNewAccount = ();
 	type OnKilledAccount = ();
+	type MainDispatcher = ();
```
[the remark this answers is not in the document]
pepyakin: I actually have the same feeling. I don't feel strongly about this though.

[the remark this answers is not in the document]
athei: This is a great suggestion. I will go with it. Was never really satisfied with the name.

test-utils/runtime/src/lib.rs
```diff
@@ -183,6 +183,17 @@ impl ExtrinsicT for Extrinsic {
 	}
 }
 
+impl Dispatchable for Extrinsic {
```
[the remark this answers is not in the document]
athei: Because the test `Runtime` sets `Extrinsic` as its `pallet_system::Trait::Call` which is bound by `Dispatchable`since this PR.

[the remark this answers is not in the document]
athei: Because the test `Runtime` sets `Extrinsic` as its `pallet_system::Trait::Call which` is bound by `Dispatchable` since this PR.

primitives/runtime/src/testing.rs
```diff
@@ -365,19 +367,11 @@ impl<Origin, Call, Extra, Info> Applyable for TestXt<Call, Extra> where
 
 	/// Executes all necessary logic needed prior to dispatch and deconstructs into function call,
 	/// index and sender.
-	fn apply<U: ValidateUnsigned<Call=Self::Call>>(
+	fn apply<U: ValidateUnsigned<Call = Self::Call>, D: crate::traits::Dispatcher<Self::Call>>(
```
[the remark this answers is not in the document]
athei: Done.

primitives/runtime/src/generic/checked_extrinsic.rs
```diff
@@ -62,21 +60,42 @@ where
...
+	len: usize,
+) -> crate::ApplyExtrinsicResult
+where
+	Origin: From<Option<AccountId>>,
+	Call: Member + Dispatchable<Origin = Origin>,
+	AccountId: Member + MaybeDisplay,
+	Info: Clone,
+	Extra: SignedExtension<AccountId = AccountId, Call = Call, DispatchInfo = Info>,
+	U: ValidateUnsigned<Call = Call>,
+	D: crate::traits::Dispatcher<Call>,
```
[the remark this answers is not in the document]
athei: Done.

bin/node-template/pallets/template/src/mock.rs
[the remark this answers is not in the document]
athei: Done.

athei:
Resolved conflicts with master and fixed and incorporated the suggestions from the comments. Should be good for merge soon (praying to CI god).

bin/node/runtime/src/lib.rs
```diff
@@ -686,7 +694,14 @@ pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
...
 pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, Call, SignedExtra>;
 /// Executive: handles dispatch to the various modules.
-pub type Executive = frame_executive::Executive<Runtime, Block, frame_system::ChainContext<Runtime>, Runtime, AllModules>;
+pub type Executive = frame_executive::Executive<
+	Runtime,
+	Block,
+	frame_system::ChainContext<Runtime>,
+	Runtime,
+	AllModules,
+	(),
```
tomusdrw: Are there use cases for customizing the top-level (transaction-level) dispatcher?
1. If not, I'd remove this.
2. If yes, I'd rather see an `enum DefaultDispatcher {}` type that implements the dispatcher instead of `()`. While `()` works for associated types, here it lacks the context of what we are actually setting (You have to go to generic parameters and hope it has a nice readable name there).

bin/node-template/runtime/src/lib.rs
```diff
@@ -269,7 +271,14 @@ pub type UncheckedExtrinsic = generic::UncheckedExtrinsic<Address, Call, Signatu
...
 pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, Call, SignedExtra>;
 /// Executive: handles dispatch to the various modules.
-pub type Executive = frame_executive::Executive<Runtime, Block, system::ChainContext<Runtime>, Runtime, AllModules>;
+pub type Executive = frame_executive::Executive<
+	Runtime,
+	Block,
+	system::ChainContext<Runtime>,
+	Runtime,
+	AllModules,
+	()
```
kianenigma:
```suggestion
	(),
```

frame/collective/src/lib.rs
```diff
@@ -69,6 +69,9 @@ pub trait Trait<I=DefaultInstance>: frame_system::Trait {
 
 	/// The time-out for council motions.
 	type MotionDuration: Get<Self::BlockNumber>;
+
+    /// The means of dispatching the calls to the runtime.
```
kianenigma:
```suggestion
	/// The means of dispatching the calls to the runtime.
```

frame/collective/src/lib.rs
```diff
@@ -218,7 +221,7 @@ decl_module! {
 			ensure!(Self::is_member(&who), Error::<T, I>::NotMember);
 
 			let proposal_hash = T::Hashing::hash_of(&proposal);
-			let ok = proposal.dispatch(RawOrigin::Member(who).into()).is_ok();
+			let ok = T::Dispatcher::dispatch(*proposal, RawOrigin::Member(who).into()).is_ok();
```
kianenigma: I wonder if we can ever come up with an abstraction here to actually prevent one from calling `proposal.dispatch()` outside of the context of a `Dispatcher`. I think it should be possible by checking some global environmental value. But maybe it is not even needed.

frame/democracy/src/lib.rs
```diff
@@ -1426,7 +1432,11 @@ impl<T: Trait> Module<T> {
 				let _ = T::Currency::unreserve(&who, amount);
 				Self::deposit_event(RawEvent::PreimageUsed(proposal_hash, who, amount));
 
-				let ok = proposal.dispatch(frame_system::RawOrigin::Root.into()).is_ok();
+				let ok = T::Dispatcher::dispatch(
+					proposal,
+					frame_system::RawOrigin::Root.into()
```
kianenigma:
```suggestion
					frame_system::RawOrigin::Root.into(),
```

frame/example/src/lib.rs
```diff
@@ -718,7 +718,7 @@ mod tests {
 		type Index = u64;
 		type BlockNumber = u64;
 		type Hash = H256;
-		type Call = ();
+		type Call = Call<Test>;
```
kianenigma: Why are all of these suddenly needed? would be waaay better if you implement whatever you need for `()` to allow tests to ignore this.

frame/system/src/lib.rs
```diff
@@ -146,7 +147,7 @@ pub trait Trait: 'static + Eq + Clone {
 		+ Clone;
 
 	/// The aggregated `Call` type.
-	type Call: Debug;
+	type Call: Debug + Dispatchable<Origin = Self::Origin>;
```
kianenigma: Related to my other comment: Why Call has to be bounded here?

frame/system/src/lib.rs
```diff
@@ -1039,6 +1055,15 @@ impl<T: Trait> Module<T> {
 	}
 }
 
+impl<T: Trait> sp_runtime::traits::Dispatcher<T::Call> for Module<T> {
```
kianenigma: maybe you can move the `Call: Dispatchable` here (or implement `Dispatchable` for `()` that panics). 

Also, don't quote me on this, but I think there is a reason that we don't bound call to dispatchable so far. I have to think about it more but I am not sure if it complies with our abstractions.

kianenigma: Yeah even if we keep the trait as-is, an impl of dispatchable for `()` would be nice. 

or some other way to not force all tests to define a Call.

frame/utility/src/lib.rs
```diff
@@ -351,7 +355,11 @@ decl_module! {
 					}
 				}
 
-				let result = call.dispatch(frame_system::RawOrigin::Signed(id.clone()).into());
+				let result =
+					T::Dispatcher::dispatch(
+						*call,
+						frame_system::RawOrigin::Signed(id.clone()).into()
```
kianenigma:
```suggestion
						frame_system::RawOrigin::Signed(id.clone()).into(),
```

primitives/runtime/src/generic/checked_extrinsic.rs
```diff
@@ -62,21 +60,42 @@ where
...
-		let res = self.function.dispatch(Origin::from(maybe_who));
-		Extra::post_dispatch(pre, info.clone(), len);
-		Ok(res.map_err(Into::into))
+		apply::<U, D, _, _, _, _, _>(self.function, self.signed, info, len)
 	}
 }
+
+/// Apply a call together with its signature (if any).
+///
+/// This function is exposed in order for test code to make use of it. Production
```
kianenigma: which test codes? There is a `TestXt` in test-runtime that has its own implementation of `Applyable`. Can't that be used?

test-utils/runtime/src/lib.rs
```diff
@@ -180,6 +180,17 @@ impl ExtrinsicT for Extrinsic {
...
+	type Origin = Origin;
+	type Trait = ();
+
+	fn dispatch(self, _: Origin) -> sp_runtime::DispatchResult {
+		match self {
+			_ => panic!("the extrinsic is not supposed to be dispatched"),
+		}
```
kianenigma:
```suggestion
		panic!("the extrinsic is not supposed to be dispatched"),
```

If we want to panic then might as well do it brief :D

frame/collective/src/lib.rs
gavofyork: We could require the trait to accept a parameter of a type for which an instance can only be created from the `Dispatcher`'s module.

frame/system/src/lib.rs
```diff
@@ -225,6 +226,21 @@ pub trait Trait: 'static + Eq + Clone {
...
+	/// The `RootDispatcher` is the central dispatcher and all other dispatchers should
+	/// eventually delegate to it. The default runtime uses the `RootDispatcher` for all
+	/// modules that do dispatching themselves.
+	///
+	/// However, a runtime implementer can choose to provide a custom dispatcher to some
+	/// of the modules that is different from the `MainDispatcher´. When doing this it is
+	/// important to delegate to the `RootDispatcher`. There might be use cases where the
+	/// delegation is unwanted and the custom dispatcher dispatches the `Dispatchable`
+	/// itself. It is important to understand that by doing this the dispatch becomes
+	/// invisible to the rest of the runtime machinery which relies on the `RootDispatcher`
```
gavofyork: Anytime there is something that is "important to understand" written in docs then I shivers down my spine. 

This "API" (and I would use the term generously here, as it's far from codified) is dangerous.

primitives/runtime/src/generic/checked_extrinsic.rs
```diff
@@ -62,21 +60,42 @@ where
...
+/// code may use the `Applyable` implementation of `CheckedExtrinsic`.
+pub fn apply<U, D, Info, Call, Extra, Origin, AccountId>(
+	call: Call,
+	signature: Option<(AccountId, Extra)>,
+	info: Info,
+	len: usize,
+) -> crate::ApplyExtrinsicResult
+where
```
gavofyork:
```suggestion
) -> crate::ApplyExtrinsicResult where
```

test-utils/runtime/src/lib.rs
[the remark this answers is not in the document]
gavofyork: The existence of this would seem to indicate that the abstractions added are broken by design.

frame/system/src/lib.rs
athei: I agree with you. I assumed that we need a way to enforce the correct usage which triggered this [discussion](https://github.com/paritytech/substrate/pull/5130#discussion_r392571834). Bottom line was that the soft API or "escape hatch" was intentionally. You are rightfully unhappy about it.

athei:
> Not at all happy about this PR. It introduces needless constraints and soft APIs that will ultimately lead to mishap.

I agree. We must do better.

> I'd like to see a solid argument as to why needed changes for the contract-weight stuff cannot be kept isolated in the contracts module.

It is not only about contract-weight. As far as I understand it we want to be able to refund weight for every dispatch. For example a `Balance::transfer` which targets an existing account should be able to refund the weight that was reserved for the account creation.

primitives/runtime/src/generic/checked_extrinsic.rs
athei: And this implementation calls this function. It is the result of the change you requested.

frame/system/src/lib.rs
athei: I think that is a huge pain point of this PR. I try to work around it starting with your suggestion.

athei:
In response to the feedback we received I am planning the following larger changes apart from the minor issues that might or might not be resolved in the process:

- Remove the `Dispatchable` trait bound from `pallet_system::Trait::Call` as adding it is way to restrictive with regard to testing
- Redo the customization options for Runtime implementers so that it constitutes a hard to misuse API as it should be
- Prevent the direct call of a `Dispatchable` without a a `Dispatcher`

kianenigma:
>> I'd like to see a solid argument as to why needed changes for the contract-weight stuff cannot be kept isolated in the contracts module.

> It is not only about contract-weight. As far as I understand it we want to be able to refund weight for every dispatch. For example a Balance::transfer which targets an existing account should be able to refund the weight that was reserved for the account creation.

Regardless if this approach, just emphasising that this was about more than just the contracts module and I think a weight refund in the long term is a good feature to have (through any means -- central dispatch was the best that we and @pepyakin came up with in a call. We can do better). 

We might not use it at the time being since in Polkadot we want to stick to a worse case scenario. But a general substrate chain might want to be picky and have some means of refunding weight. 

an example that we might want to consider refund even in polkadot is the phragmen submission transactions. If it goes through the hot path of replacing a solution, it takes a lot of time and it should rightfully take a whole lot of block's weight. Otherwise, it would be quite a waste to take all of that weight when we, in some cases, almost immediately reject the solution. Ideally the fee should still be taken but the weight should be refunded to the system module.

bkchr:
TBH, I don't understand the purpose of this pr. Could someone sketch how that will help refunding weights?

athei:
> TBH, I don't understand the purpose of this pr. Could someone sketch how that will help refunding weights?

Weight refund will be done on a per dispatch basis as described [here](https://www.notion.so/Weight-Refunds-Proposal-357d9438885c4c3d8d1af3a974928087).

In order to do that we decided that going through a central instance that tracks the refunded gas per dispatch and deals with nested dispatches would be the best approach. #5032 already implements this central location as a `dispatch_call` method on the system pallet module.

The problem there is that this approach is purely advisory. Modules have to opt in by calling this new function instead of calling `Dispatchable::dispatch`. Because that approach is fragile  we establish a mandatory weight tracking with this PR by forcing all dispatches through a `Dispatcher`. The mentioned #5032 will then be changed to integrate the weight refund into the `Dispatcher`.

bkchr:
Hmm, so the problem you try to solve here isn't really solved IMHO. 

In the default case where we don't have a wrapped call, we don't require this pr.

For the other case, where we have a wrapped call this weight refund doesn't make really sense. Let me explain why. For example think about a democracy proposal. After the proposal is accepted and we reached the point it should be applied, we will dispatch it. The proposal is applied in `on_initialize`, while `on_initialize` has some weight, it is probably way less than the weight of the actual call that will be dispatched as proposal. So, in your call you will refund some weight that wasn't taken for your call. 

What we really want is probably an improved `SignedExtension` that is aware of `dispatch` and `post_dispatch`, but stuff like `TransactionPayment` should also not kick in for calls that are being dispatched from another module. I also already thought about making `SignedExtension` aware of wrapped calls, but didn't yet come to a proper solution. I'm open to discuss about this.

athei:
> Hmm, so the problem you try to solve here isn't really solved IMHO.

Yes. As i already stated the implementation is lacking. Working on it right now.


> For the other case, where we have a wrapped call this weight refund doesn't make really sense. Let me explain why. For example think about a democracy proposal. After the proposal is accepted and we reached the point it should be applied, we will dispatch it. The proposal is applied in on_initialize, while on_initialize has some weight, it is probably way less than the weight of the actual call that will be dispatched as proposal. So, in your call you will refund some weight that wasn't taken for your call.

No. It does not work as you describe. The caller of the nested or wrapped call gets the unspent gas as return value and is responsible for handling it. It is all written down in the notion page I linked.

> What we really want is probably an improved SignedExtension that is aware of dispatch and post_dispatch, but stuff like TransactionPayment should also not kick in for calls that are being dispatched from another module. I also already thought about making SignedExtension aware of wrapped calls, but didn't yet come to a proper solution. I'm open to discuss about this.

The other PR I linked takes care of that. It changes the SignedExtensions to be aware of weight refund.

frame/collective/src/lib.rs
```diff
@@ -69,6 +69,8 @@ pub trait Trait<I=DefaultInstance>: frame_system::Trait {
 
 	/// The time-out for council motions.
 	type MotionDuration: Get<Self::BlockNumber>;
+
+	type Dispatcher: sp_runtime::traits::Dispatcher<<Self as Trait<I>>::Proposal>;
```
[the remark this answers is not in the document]
athei: Fixed.

frame/contracts/src/lib.rs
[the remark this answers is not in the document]
athei: The reasoning is to give the runtime implementer the opportunity to customize the dispatch process for some Module. With the rewrite this is much safer in a way that dispatchers cannot manually call `Dispatchable::dispatch` but must specify a `RootDispatcher` in their implementation.

test-utils/runtime/src/lib.rs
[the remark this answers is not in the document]
athei: No longer needed and therefore removed with the rewrite.

bin/node/runtime/src/lib.rs
athei: There might be cases where someone want to provide another `RootDispatcher`. This is a far less intrusive change for a runtime implementer than modifying the `frame_system` which is the default `RootDispatcher`. Therefore I argue that it provides some benefit.

That said, I removed this parameter because we already specify the `RootDispatcher` through the `frame_system::trait::RootDispatcher`. No need for a redundant parameter.

bin/node-template/runtime/src/lib.rs
athei: No longer changing this line.

frame/collective/src/lib.rs
athei: The rewrite applies gavs suggestion in order to accomplish this.

frame/democracy/src/lib.rs
athei: No longer applies.

frame/example/src/lib.rs
athei: You are right. It is no longer needed with the rewrite.

frame/system/src/lib.rs
athei: It is no longer bounded.

test-utils/runtime/src/lib.rs
athei: This code was removed from the PR.

primitives/runtime/src/generic/checked_extrinsic.rs
athei: Done.

gavofyork:
It sounds like the only API change requiring is a minor one that just ensures dispatchables return the weight used.

athei:
> It sounds like the only API change required is a minor one that just ensures dispatchables (and on_initialize) return the weight consumed.
> 
> > we decided that going through a central instance
> 
> I do not want to see any new "central instances" of anything.

That is another approach to the weight refund task. One could modify the `Dispatchable` and the macros that generate its implementations so that it does the weight accounting for every call. This is clearly a more elegant approach.

It comes with its own drawbacks, though.

1. It would introduce a frame concept (weight) to `Dispatchable` that lives in primitives and should therefore be runtime agnostic. This can be worked around by making the return type of `Dispatchable` generic, I guess. So it is not necessarily weight but any additional data that a `Dispatchable` can deliver in addition to the `DispatchResult`.

2. What about frame chains that do not want to use weight refund? Generating the refund logic in macros makes it more difficult to make the concept pluggable (correct me if I am wrong). In its current state of this PR this would be just a matter of setting `frame_system::trait::RootDispatcher = SimpleDispatcher`. I did not do much thinking in this direction, yet. There might be a good solution for this. It might require an even larger change set than this PR, though.

@gavofyork Is this a direction you would rather see the weight refund stuff going?

kianenigma:
I think in the grand scheme of things, it is easier to think of it this way: 

Instead of asking dispatchables to __return the used weight__ (or signal it via some function or any other means), it should be assumed that they consume all of it unless if they return some value. So, if I have understood it correctly, I agree with 

> What about frame chains that do not want to use weight refund? Generating the refund logic in macros makes it more difficult to make the concept pluggable (correct me if I am wrong)

frame/collective/src/lib.rs
kianenigma: Do we have a `[should_panic]` test case around it? I'd love to see that 😄

frame/democracy/src/lib.rs
```diff
@@ -1539,7 +1545,11 @@ impl<T: Trait> Module<T> {
...
```
kianenigma:
```suggestion
					frame_system::RawOrigin::Root.into(),
```

frame/democracy/src/lib.rs
```diff
@@ -1663,4 +1673,4 @@ impl<T: Trait> Module<T> {
 		}
 		Ok(())
 	}
-}
+}
```
kianenigma:
```suggestion
}

```

frame/executive/src/lib.rs
```diff
@@ -312,7 +320,10 @@ where
 
 		// Decode parameters and dispatch
 		let dispatch_info = xt.get_dispatch_info();
-		let r = Applyable::apply::<UnsignedValidator>(xt, dispatch_info, encoded_len)?;
+		let r = Applyable::apply::<UnsignedValidator, System::RootDispatcher>(xt,
+			dispatch_info,
+			encoded_len
```
kianenigma:
```suggestion
			encoded_len,
```

frame/system/src/lib.rs
```diff
@@ -221,6 +222,14 @@ pub trait Trait: 'static + Eq + Clone {
 	///
 	/// All resources should be cleaned up associated with the given account.
 	type OnKilledAccount: OnKilledAccount<Self::AccountId>;
+
+	/// The central instance that dispatches all calls.
+	///
+	/// For `Dispatchable` calls it is usually the responsiblity of the frame_system
+	/// module to function as the root `RootDispatcher`. In order to customize the
+	/// dispatch process for a specific module the `Dispatcher` trait can be implemented
+	/// and supplied to to the respective `Trait`.
```
kianenigma:
```suggestion
	/// and supplied to the respective `Trait`.
```

primitives/runtime/src/lib.rs
```diff
@@ -434,6 +434,22 @@ impl traits::Printable for DispatchError {
 	}
 }
 
+/// This type describes the result of a module function call which is called through
```
kianenigma: nit: I'd put these guys next to `SimpleDispatcher`.

kianenigma: Or not. There's a lot more dispatch-related and sadly the ordering is a bit confused. nvmd.

primitives/runtime/src/traits.rs
```diff
@@ -902,11 +906,193 @@ pub trait Applyable: Sized + Send + Sync {
...
+		Self::new()
+	}
+}
+
+impl<T: UnconstructableToken> Copy for Unconstructable<T> {}
+
+/// A type that is responsible for customizing dispatching of a `Dispatchable`.
+///
+/// In order to provide a custom dispatcher this type should be implemented and then
+/// and then used by calling `dispatch` with a `Dispatchable`.
```
kianenigma:
```suggestion
/// In order to provide a custom dispatcher this type should be implemented
/// and then used by calling `dispatch` with a `Dispatchable`.
```

primitives/runtime/src/traits.rs
```diff
@@ -902,11 +906,193 @@ pub trait Applyable: Sized + Send + Sync {
...
+	where
+		D: Dispatchable<Origin = O>,
+	{
+		dispatchable.dispatch(origin, Unconstructable::new())
+	}
+
+	/// Implement your custom dispatch logic here.
+	///
+	/// Use `Self::raw_dspatch` in order to do the actual dispatch as direct call
+	/// of `Dispatchable::dispatch` is prevent by a token. This function should be called
```
kianenigma:
```suggestion
	/// Use `Self::raw_dspatch` in order to do the actual dispatch as direct call
	/// of `Dispatchable::dispatch`. ???? is prevent by a token. This function should be called
```

athei:
Closed in favor of #5458

PR #5133, opened 2020-03-05
grandpa: improve handling of global gossip messages

andresilva:
The GRANDPA gossip validator only tracked the height of the last commit we had processed and not the round/set_id at which it was committed. If multiple commits are available for the same best block then we'd keep multiple around (in particular commits at different rounds for the same block). Additionally, when considering whether to send a commit to a peer we wouldn't take into account the set id at which the peer was and would end up sending useless commit data to syncing peers. This PR fixes both issues.


andresilva:
Build failure seems unrelated.

PR #5136, opened 2020-03-05
Revert "Don't include `:code` by default in storage proofs (#5060)"

bkchr:
This reverts commit a193a19d57c894ddec17b95dc8f2b8f379be0d11.

This should make @arkpar life easier and will come back in another form later.


arkpar:
See discussion in #5060

PR #5137, opened 2020-03-05
Introduce default-setting prime for collective

gavofyork:
Allows the existence of a "prime member" of the collective whose vote acts as the default for any other members.

CC @jacogr


frame/collective/src/lib.rs
```diff
@@ -152,19 +177,21 @@ decl_module! {
...
 			ensure_root(origin)?;
 			let mut new_members = new_members;
 			new_members.sort();
-			<Members<T, I>>::mutate(|m| {
-				<Self as ChangeMembers<T::AccountId>>::set_members_sorted(&new_members[..], m);
-				*m = new_members;
-			});
+			let old = Members::<T, I>::get();
+			<Self as ChangeMembers<T::AccountId>>::set_members_sorted(&new_members[..], &old);
+			Prime::<T, I>::set(prime);
```
shawntabrizi: Updating `Members` storage with `new_members` has been lost with this change.

frame/collective/src/lib.rs
```diff
@@ -249,43 +277,110 @@ decl_module! {
...
+		///   - `L` is the encoded length of `proposal` preimage.
+		#[weight = SimpleDispatchInfo::FixedOperational(200_000)]
+		fn close(origin, proposal: T::Hash, #[compact] index: ProposalIndex) {
+			let _ = ensure_signed(origin)?;
+
+			let voting = Self::voting(&proposal).ok_or(Error::<T, I>::ProposalMissing)?;
+			ensure!(voting.index == index, Error::<T, I>::WrongIndex);
+			ensure!(system::Module::<T>::block_number() >= voting.end, Error::<T, I>::TooEarly);
+
+			// default to true only if there's a prime and they vpted in favour.
```
shawntabrizi:
```suggestion
			// default to true only if there's a prime and they voted in favour.
```

frame/collective/src/lib.rs
gavofyork: yeah. that was superfluous logic as it happens anyway in `<Self as ChangeMembers<T::AccountId>>::set_members_sorted(&new_members[..], &old)`.

shawntabrizi: I think this is an issue since other parts of logic rely on `Self::members()` which reads from the local members list.

We need some trait to tell the module where to read the correct list of members from, and use it here.

gavofyork: Not sure what you mean. `Members` is always kept up to date regardless of how the set is changed.

Issue #5112, opened 2020-03-03
Bad block cause runtime constantly panicking and DOS the network


xlc:
Again relates to this: https://github.com/AcalaNetwork/Acala/issues/135

Now it happens on every full node.

Here are my observations:

- For some reason a validator produced a block contains a transaction have invalid signature. (Haven't checked if it is indeed invalid, but is not accepted by other nodes)
- When other nodes trying to import this block, the importer panics.

https://github.com/paritytech/substrate/blob/7e383edb2f8ed9a912cc012905d752795b8a3201/frame/executive/src/lib.rs#L272

- The panic is kind of ignored and trying to continue importing new block
- And panics again

This happens on all nodes including other validators.

It should record this block is bad and ban it, instead of keep trying to import it and not able to do anything.

Some more logs

https://gist.github.com/xlc/8395857944b02fdad71c811e90daaded

The root cause be more serious as why a invalid block was produced by a honest node. The transaction is sent by one of our bot and it was always working (i.e. sending valid transaction).

The troublesome extrinsic:

0x310284ff9e22b64c980329ada2b46a783623bcf1f1d0418f6a2b5fbfb7fb68dbac5abf0f01b02802078d3fb8fb7d48170ec21129abe4f53c02da03aa5e0386384fb35dce6e1257c02d048f22d99cfeca9c5f58232c30d2519fc9c25b390dc76642145f6c808501311200190003b233dc04b300540c00000000000000000158ed63772ec9c5bec501000000000000

Signature:

0xb02802078d3fb8fb7d48170ec21129abe4f53c02da03aa5e0386384fb35dce6e1257c02d048f22d99cfeca9c5f58232c30d2519fc9c25b390dc76642145f6c80

Signer:

5Fe3jZRbKes6aeuQ6HkcTvQeNhkkRPTXBwmNkuAPoimGEv45

xlc:
What I don't understand is that some validators accepts the signature. If that's the case, it indicates some serious issue on sr25519.

burdges:
We changed sr25519 11 months ago in April 2019 to set the high bit of byte 31 of a signature, which marks the signature as sr25519 to speed up batch verification when used along side ed25519.

Your byte 31 is 6e so your using sr25519 from before April 2019.  We delayed yanking schnorrkel 0.1.1 for testnet reasons, but its yanked now.  If anyone knows any method to actually purge the old code from crates.io then please let me know.

xlc:
[...]

PR #5138, opened 2020-03-05
Update schnorrkel to 0.9, remove deprecated api usage

NikVolf:


primitives/core/src/sr25519.rs
```diff
@@ -776,6 +780,7 @@ mod test {
 	}
 
 	#[test]
+	#[ignore = "sr25519 signatures from 1.1 version are no longer supported"]
```
NikVolf: what can be used instead?

primitives/core/src/sr25519.rs
```diff
@@ -532,19 +532,22 @@ impl TraitPair for Pair {
...
-		// Match both schnorrkel 0.1.1 and 0.8.0+ signatures, supporting both wallets
-		// that have not been upgraded and those that have. To swap to 0.8.0 only,
-		// create `schnorrkel::Signature` and pass that into `verify_simple`
-		match PublicKey::from_bytes(pubkey.as_ref()) {
-			Ok(pk) => pk.verify_simple_preaudit_deprecated(
-				SIGNING_CTX, message.as_ref(), &sig,
-			).is_ok(),
-			Err(_) => false,
-		}
+		let signature = match schnorrkel::Signature::from_bytes(sig) {
```
andresilva: Maybe we should try to sync kusama with this, not sure if there's txs signed with schnorrkel 0.1.1 that we won't be able to verify anymore. cc @burdges

andresilva: Or BABE blocks for that matter.

NikVolf: > Maybe we should try to sync kusama with this, not sure if there's txs signed with schnorrkel 0.1.1 that we won't be able to verify anymore. cc @burdges

Yep, trying to do it right now

@burdges said something that 1.1 were only used on Alexander

burdges: We never supported VRFs from 0.1.1 because BABE did not work until after the audit.  ;)

Just send schnorrkel a PR reverting https://github.com/w3f/schnorrkel/commit/8f3706591dcf0648086b5dfb4bd832970483ac03 if you really need it, but if these exist then we maybe we could keep that pain isolated to kusama, and disable the `preaudit_deprecated` feature for polkadot.

primitives/core/src/sr25519.rs
bkchr: Remove? The test is useless now^^

primitives/core/src/sr25519.rs
```diff
@@ -609,6 +612,7 @@ mod compatibility_test {
 	}
 
 	#[test]
+	#[ignore = "sr25519 signatures from version 1.1 are no longer supported"]
```
bkchr: Same

NikVolf:
> Did you synced Kusama with it?

got error in block 3876

Also can confirm that concerns in #5112 are valid.
Node just keeps spamming about runtime panic when invalid block is encountered.

primitives/core/src/sr25519.rs
jacogr: There will be old txs (old blocks) signed with 0.1.1 - the JS side only moved to 0.8.5 in Jan once Alex was dropped.

burdges:
Any thoughts?  

I've temporarily pushed f4447cd7eb0c793905aa85af4d32c71a78ea020a that reverts  8f3706591dcf0648086b5dfb4bd832970483ac03 and restores the preaudit_deprecated feature, but..  

We should avoid support for preaudit_deprecated in polkadot because allowing it screws up batch verification for ed25519 without adding a key type field. 

We actually have some key type field I suppose, so maybe we could leverage that for kusama somehow.

burdges:
I've an idea:  Could kusama verify these old signature entirely from the runtime?  We'd drop support for them from the runtime?  Is that too annoying?

bkchr:
> I've an idea: Could kusama verify these old signature entirely from the runtime? We'd drop support for them from in the host in this case. Is that too annoying?

No, the problem is that we can not modify the old runtimes anymore. However, we could probably workaround and only accept from now on the new signatures when building a block but still support verifying the old ones.

burdges:
I'll need to publish schnorrkel 0.9.1 with https://github.com/w3f/schnorrkel/commit/f4447cd7eb0c793905aa85af4d32c71a78ea020a included I suppose?

NikVolf:
> I'll need to publish schnorrkel 0.9.1 with [w3f/schnorrkel@f4447cd](https://github.com/w3f/schnorrkel/commit/f4447cd7eb0c793905aa85af4d32c71a78ea020a) included I suppose?

Yeah, if it is an improvement on 0.8.*, we can update to it

burdges:
I've published 0.9.1 that brings back the preaudit_deprecated feature.  

We cannot do a logical or in a batch verification, so #5023 requires the old and new signatures be distinguished.  We need not distinguish them on-chain however.  Instead, we expose both old and new verification mechanisms to the runtime, so the runtime could switch at some block height or whatever.

DemiMarie-parity:
> I've published 0.9.1 that brings back the preaudit_deprecated feature.
> 
> We cannot do a logical or in a batch verification, so #5023 requires the old and new signatures be distinguished. We need not distinguish them on-chain however. Instead, we expose both old and new verification mechanisms to the runtime, so the runtime could switch at some block height or whatever.

Will we be able to get rid of the cruft at some point, once a certain block hash becomes hard-coded in the client?

burdges:
> Will we be able to get rid of the cruft at some point, once a certain block hash becomes hard-coded in the client?

Yes and no.  We'll ignore this entirely on Polkadot, so we can drop it from substrate builds that only support Polkadot.  We cannot drop it from substrate builds that support Kusama, at least not until some consensus upgrade forces some "harder" fork.  We should've gone through this pain for alexandre vs kusama, but we took the more relaxed approach there so we pay for it now.

NikVolf:
> I've published 0.9.1 that brings back the preaudit_deprecated feature.
> 
> We cannot do a logical or in a batch verification, so #5023 requires the old and new signatures be distinguished. We need not distinguish them on-chain however. Instead, we expose both old and new verification mechanisms to the runtime, so the runtime could switch at some block height or whatever.

Can this signatures be distinguished at runtime, at runtime?
We can just skip batch for them then, and forbid when block building.

burdges:
> Can this signatures be distinguished at runtime, at runtime?
> We can just skip batch for them then, and forbid when block building.

Yes should work.  

Signatures by 0.1.1 should fail `Signature::from_bytes`.  I published 0.1.1 on 2 April 2019, but only added the distinguishing marker checked by `Signature::from_bytes` on 19 April 2019:  https://github.com/w3f/schnorrkel/commits/master?after=2d004cffa5dc3bf6dafeebee8d4d40b9f696462f+34&path%5B%5D=src&path%5B%5D=sign.rs  

I published 0.2-0.7 in June-July, but yanked them all before Kusama's launch.  We kept support for 0.1.1 for the alexandre testnet and only screwed up by not supporting 0.1.1 in kusama.

bkchr:
Was done by: https://github.com/paritytech/substrate/pull/5316

NikVolf:
It was not done by #5316, doing it properly (with disabling generation of deprecated signatures) requires runtime api versioning I'm working on now

bkchr:
I know, but I see this as a new pr with a different scope.

burdges:
I'll amend my comment above:  We'll separate proper schnorrkel signatures from these deprecated signatures for batch verification using `schnorrkel::Signature::from_bytes`, but we've no similar mechanism for separating the 0.1.1 signatures from ed25519 signatures for ed25519 batch verification.  We could however just leave ed25519 batch verification off for kusma, while leaving schnorrkel 0.1.1 verification off for polkadot.  If this switch gets controlled in the runtime, then kusama could flip ed25519 batch verification on at some point too.

bkchr:
Batch verification will require runtime changes and when this comes we will have support for disabling the old signature verification.

bkchr:
Tldr, we should get this working.

kirushik:
@bkchr is there a tracking issue for this task somewhere? Or any other way of monitoring the progress (and, more importantly, of ensuring that we don't leave `preaudit_deprecated` feature of schnorrkel on for any Polkadot release builds)?

bkchr:
The proper solution was implemented here: https://github.com/paritytech/substrate/pull/5358

The problem is that we will never be able to disable `preaudit_deprecated`, because Kusama requires it when syncing from genesis. And the way we currently build Kusama, prevents us from disabling the feature for Polkadot, because it would automatically be enabled by Kusama (everything is build in the same env). However, newer runtimes don't expose the old function, this means we don't accept anymore signatures that are build with the 1.0 schnorrkel.

DemiMarie-parity:
@bkchr we can (and probably should) switch to separate environments instead.  We can also hardcode the hash of every block made on Kusama with the old signature.  We then don’t need to validate the signatures any more.

bkchr:
Why should we do this? We can not call this implementation anymore from newer runtimes. (it would work, but you would need to have control over the build process and get this runtime as the canonical one)

DemiMarie-parity:
@bkchr to remove cruft, reduce code size, and to ensure that this does not wind up in Polkadot.

PR #5143, opened 2020-03-05
Expose the `runtime-benchmarks` feature to the cli crate

bkchr:
This exposes the `runtime-benchmarks` feature via the cli crate and
makes sure the benchmarking can be enabled. This requires that the user
goes to `bin/node/cli` and runs `cargo build --features
runtime-benchmarks` to build a node that has the feature enabled.


tomusdrw:
Should we put some docs how to compile & run the benchmarks?

PR #5145, opened 2020-03-05
Add more prometheus metrics to `network::Protocol`.

twittner:
Context: #5106.

### DIFF
--- a/client/network/src/protocol/sync.rs
+++ b/client/network/src/protocol/sync.rs
@@ -1202,6 +1202,27 @@ impl<B: BlockT> ChainSync<B> {
 	fn is_already_downloading(&self, hash: &B::Hash) -> bool {
 		self.peers.iter().any(|(_, p)| p.state == PeerSyncState::DownloadingStale(*hash))
 	}
+
+	/// Return some key metrics.
+	pub(crate) fn metrics(&self) -> Metrics {
+		use std::convert::TryInto;
+		Metrics {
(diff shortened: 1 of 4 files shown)


client/network/src/protocol/sync.rs
```diff
@@ -1202,6 +1202,27 @@ impl<B: BlockT> ChainSync<B> {
...
+
+	/// Return some key metrics.
+	pub(crate) fn metrics(&self) -> Metrics {
+		use std::convert::TryInto;
+		Metrics {
+			queued_blocks: self.queue_blocks.len().try_into().unwrap_or(std::u32::MAX),
+			fork_targets: self.fork_targets.len().try_into().unwrap_or(std::u32::MAX),
+			finality_proofs: self.extra_finality_proofs.metrics(),
+			justifications: self.extra_justifications.metrics(),
+			_priv: ()
```
mxinden: @twittner am I understanding correctly that `_priv` is used to tell its consumer that the given structure might be extended in the future and that such an extension would not be a breaking change (see https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md)?

Is this really necessary given that this is not exposed across crate boundaries?

twittner: It was added to ensure that any `Metrics` value anywhere outside this module was created here. `#[non_exchaustive]` only works for downstream crates. It is probably not really necessary so if you feel strongly about it free to remove it.

PR #5147, opened 2020-03-05
removes use of sc_client::Client from sc_network

seunlanlege:
see #4452


client/api/src/client.rs
```diff
@@ -74,7 +75,7 @@ pub trait BlockchainEvents<Block: BlockT> {
 }
 
 /// Fetch block body by ID.
-pub trait BlockBody<Block: BlockT> {
+pub trait BlockProvider<Block: BlockT> {
```
tomusdrw: Could this be eventually moved to `primitives`? Like is there any reason why this is not part of `sp_blockchain`?

tomusdrw: Seems that we already have `HeaderBackend`, maybe this one should be called `BlockBackend` instead?

client/network/src/protocol/block_requests.rs
```diff
@@ -192,7 +192,7 @@ where
 
 		let mut blocks = Vec::new();
 		let mut block_id = from_block_id;
-		while let Some(header) = self.chain.header(&block_id).unwrap_or(None) {
+		while let Some(header) = self.chain.header(block_id).unwrap_or(None) {
```
tomusdrw: Why does `fn header` take ownership over `BlockId`, but methods in `BlockProvider` don't? Can't we unify this and use references everywhere?

client/src/client.rs
```diff
@@ -1848,6 +1843,33 @@ impl<B, E, Block, RA> BlockBody<Block> for Client<B, E, Block, RA>
 			_ => None,
 		})
 	}
+
+	fn block_status(&self, id: &BlockId<Block>) -> sp_blockchain::Result<BlockStatus> {
+		// this can probably be implemented more efficiently
```
tomusdrw: Was this comment copied from somewhere? Can you elaborate what we could do to improve it here? Or maybe create a separate issue?

client/consensus/babe/rpc/src/lib.rs
```diff
@@ -50,8 +50,6 @@ pub trait BabeApi {
 }
 
 /// Implements the BabeRPC trait for interacting with Babe.
-///
-/// Uses a background thread to calculate epoch_authorship data.
```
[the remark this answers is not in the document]
seunlanlege: yes, stale comment.

client/src/client.rs
seunlanlege: Yeah the comment was copied from the original function body

client/network/src/protocol/block_requests.rs
seunlanlege: yeah so this method comes from `HeaderBackend`, This would require refactoring the trait itself and it's usage. Happy to do this in a separate PR.

tomusdrw: Fine with me!

gnunicorn:
@seunlanlege needs merge and tests are failing.

seunlanlege:
@gnunicorn green check marks on tests.

PR #5151, opened 2020-03-05
Deferred Slashing/Offence for offchain Phragmen 

kianenigma:
Some subtle complications: 

- No Exposures should change when we are accepting solutions, as it invalidates them and breaks accounting. We currently have `SlashDefferDuration > 0` in Kusama which means that this won't happen, but we can't assume that in substrate. Some chains might have immediate slashing and that simply cannot work with offchain phragmen. 
- Additionally, the last_nonzero_slash of nominators should also preferably not change as it can have the same impact. Although our validators currently take this into account, but there could be a slashing in the short span in between a solution being computed and it being submitted. 

There are numerous approaches to these and the risks and attack vectors were not clear, so for now we do something very simple: All reports are ignored while election window is open. This buffering can happen in either staking or offences. For simplicity, and for the sake of staking's code not bloating even more, I do it for now in the offence side. 

All in all, it is not an elegant solution but it is rather cheap and assures that no accounting/slashing assumptions are broken. And I don't see it as permanent, we'll probably refactor it into staking once staking itself is in a better shape.


primitives/staking/src/offence.rs
```diff
@@ -142,19 +142,28 @@ pub trait OnOffenceHandler<Reporter, Offender> {
 	/// Zero is a valid value for a fraction.
 	///
 	/// The `session` parameter is the session index of the offence.
+	///
+	/// The receiver might decide to not accept this offence. In this case, the call site is
+	/// responsible for queuing the report and re-submitting again.
```
tomusdrw: If that's common we should probably extract this logic into some kind of a wrapper.

primitives/staking/src/offence.rs
```diff
@@ -142,19 +142,28 @@ pub trait OnOffenceHandler<Reporter, Offender> {
...
 	fn on_offence(
 		offenders: &[OffenceDetails<Reporter, Offender>],
 		slash_fraction: &[Perbill],
 		session: SessionIndex,
-	);
+	) -> Result<(), ()>;
+
+	/// Can an offence be reported now or not? This should always be false if a direct call to
+	/// [`on_offence`] was to return `Err`.
```
tomusdrw: We can't really guarantee that, so I think the comment here should rather say that this should only be considered an optimization.
Also it seem that passing the arguments (or at least `OffenceDetails`) would be useful for some cases, I'd consider including that too.

frame/offences/src/lib.rs
```diff
@@ -80,11 +90,22 @@ decl_event!(
...
 
 decl_module! {
-	/// Offences module, currently just responsible for taking offence reports.
 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
 		fn deposit_event() = default;
+
+		fn on_initialize(_now: T::BlockNumber) {
+			// only decode storage if we can actually submit anything again.
+			if T::OnOffenceHandler::can_report() {
+				// defensive only. Per definition, `OnOffenceHandler` should accept reports when
```
tomusdrw: I'd remove this comment, the assumption is pretty weak, we don't have a ways to enforce that if `can_report` returns `false` the actual reporting won't fail.

frame/offences/src/lib.rs
```diff
@@ -80,11 +90,22 @@ decl_event!(
...
+				// `can_report` is true.
+				<DeferredOffences<T>>::take()
+					.iter()
+					.for_each(|(o, p, s)| Self::report_or_store_offence(o, p, *s));
```
tomusdrw: Should we print a warning at least if it wasn't accepted twice?

frame/offences/src/lib.rs
```diff
@@ -118,17 +139,31 @@ where
...
+		slash_perbill: &[Perbill],
+		session_index: SessionIndex,
+	) {
+		let _ = T::OnOffenceHandler::on_offence(
+			&concurrent_offenders,
+			&slash_perbill,
+			session_index,
+		).map_err(|_| {
+			<DeferredOffences<T>>::mutate(|d|
+				d.push((concurrent_offenders.to_vec(), slash_perbill.to_vec(), session_index))
```
tomusdrw: This is unbounded. Are there any guarantees that it won't go over some reasonable size? Also should we limit that?

kianenigma: If a sane module implement `OnOffence`, since we check on every `on_initialise`, then it should be emptied pretty quickly. Yet, no we don't enforce any size. I would be happy to do that from a code PoV. But with slashing, I think it is a tricky situation. A bug here that allows someone that can report but not get slashed due to some buffer being full, IMO, is more catastrophic than temporarily storing lots of data.

primitives/staking/src/offence.rs
kianenigma: Yes exactly it is just an optimization. Arguments will be unused for now and since I see this whole PR as potentially temporary, then I won't add it for now.

frame/offences/src/lib.rs
kianenigma: Yes that'd be good 👍

frame/offences/src/lib.rs
```diff
@@ -54,7 +55,16 @@ pub trait Trait: frame_system::Trait {
 decl_storage! {
 	trait Store for Module<T: Trait> as Offences {
 		/// The primary structure that holds all offence records keyed by report identifiers.
-		Reports get(fn reports): map hasher(blake2_256) ReportIdOf<T> => Option<OffenceDetails<T::AccountId, T::IdentificationTuple>>;
+		Reports get(fn reports): map hasher(blake2_256) ReportIdOf<T>
+			=> Option<OffenceDetails<T::AccountId, T::IdentificationTuple>>;
```
rphmeier: I would prefer to keep the `IdentificationTuple` in a separate map if there are multiple reports on the same validator. They are quite large and could bloat the state.

primitives/staking/src/offence.rs
rphmeier: Agree with Tomek that there may be false negatives (in this case `true` return values).

frame/offences/src/lib.rs
kianenigma: yeah this can be separate issue since this is actually old storage and I don't want to do migration as well in the current PR. A follow up can migrate both this and `DeferredOffences` into something that can de-compose `T::IdentificationTuple` into a `Identification` and `ExposureData` (atm this is not possible and not trivial) and then store them as you explained.

primitives/staking/src/offence.rs
kianenigma: We have deferring in quite a few places (in Staking and slashing) but I can't see how you can abstract them away, since IMO each has its own taste and details. Do you have a more concrete example of what you are suggesting here?

kianenigma:
@tomusdrw I applied almost all reviews. I will merge this into the base branch. If you are unhappy with any of the patches, please report there 👍

PR #5152, opened 2020-03-05
Revert quote fixed version and upgrade failure

bkchr:


frame/staking/reward-curve/Cargo.toml
```diff
@@ -13,7 +13,7 @@ proc-macro = true
 
 [dependencies]
 syn = { version = "1.0.7", features = ["full", "visit"] }
-quote = "=1.0.2"
+quote = "1.0"
```
andresilva:
```suggestion
quote = "1.0.2"
```

PR #5157, opened 2020-03-06
Don't use skip signature check by default

NikVolf:
As discussed in #5112, seems there are some corner cases that make this optimisation rather unsafe.

We'll revisit it soon to make it safe or use batching/parallelization/caching(with runtime version awareness) instead.


bkchr:
No, we will remove this feature completely. It is to fragile.

PR #5158, opened 2020-03-06
removes use of sc_client::Client from node-transaction-factory

seunlanlege:
see #4452


bin/node/testing/src/bench.rs
```diff
@@ -56,6 +56,7 @@ use sc_client_api::{
 	execution_extensions::{ExecutionExtensions, ExecutionStrategies},
 };
 use sp_core::{Pair, Public, sr25519};
+use sc_block_builder::BlockBuilderProvider;
```
tomusdrw: Ideally `traits` should be defined in `sp_` crates, to avoid bringing the entire implementation together with the trait. Is it possible to move `trait BlockBuilderProvider` to `sp-block-builder` without violating dependency rules?

seunlanlege: Then we'd have to move the `BlockBuilder` struct there as well because it's used in the trait and seeing as crates in `primtives/*` can't depend on crates in `client/*`, I would just elect for us to merge `sc-block-builder` and `sp-block-builder`.

seunlanlege: Ok after messing around with this, I've discovered I can't move the trait to `sp_block_builder`.

moving it would require moving the `BlockBuilder` struct which would also require moving the `BuiltBlock` struct which is generic over a `StateBackend` from `sc_client_api`.

and `primitives` -> `client` dependencies are discouraged.

gnunicorn:
@seunlanlege tests failing, mind checking that?

seunlanlege:
looks like the standard library's prelude is missing from `frame_support`

```
error: cannot find macro `vec` in this scope
  --> /builds/parity/substrate/frame/support/src/storage/migration.rs:80:16
   |
80 |     let mut key = vec![0u8; 32 + hash.len()];
   |                   ^^^
error: cannot find macro `vec` in this scope
  --> /builds/parity/substrate/frame/support/src/storage/migration.rs:89:16
   |
89 |     let mut key = vec![0u8; 32 + hash.len()];
   |                   ^^^
error: cannot find macro `vec` in this scope
  --> /builds/parity/substrate/frame/support/src/storage/migration.rs:98:16
   |
98 |     let mut key = vec![0u8; 32 + hash.len()];
   |                   ^^^
warning: ignoring -C extra-filename flag due to -o flag
error: aborting due to 3 previous errors
error: could not compile `frame-support`.
warning: build failed, waiting for other jobs to finish...
error: build failed
warning: build failed, waiting for other jobs to finish...
error: build failed
Command exited with non-zero status 101
2627.12user 102.53system 3:19.54elapsed 1367%CPU (0avgtext+0avgdata 699304maxresident)k
143368inputs+2386264outputs (58major+17407220minor)pagefaults 0swaps
ERROR: Job failed: exit code 1
```

gavofyork:
it's a `no_std` environment. if you need the prelude, make sure you have `use sp_std::prelude::*;` at the top of the file.

seunlanlege:
the issue was `sc-block-builder` was being compiled in a `no_std` environment, @gnunicorn this is also ready.

PR #5159, opened 2020-03-06
Revert "Build block without checking signatures (#4916)"

bkchr:
After speaking with @tomusdrw, we came to the conclusion that we should remove this feature. It is nice and could bring some speedups, but the security problems are too large. 

- A transaction could sit in the tx-pool waiting to be included and the account index is reaped in the meantime, someone else starts using the account index and we include this transaction. The signature verification would clearly fail at this point.

### DIFF
--- a/client/rpc/src/state/tests.rs
+++ b/client/rpc/src/state/tests.rs
@@ -402,7 +402,7 @@ fn should_return_runtime_version() {
 
 	let result = "{\"specName\":\"test\",\"implName\":\"parity-test\",\"authoringVersion\":1,\
 		\"specVersion\":1,\"implVersion\":2,\"apis\":[[\"0xdf6acb689907609b\",2],\
-		[\"0x37e397fc7c91f5e4\",1],[\"0xd2bc9897eed08f15\",1],[\"0x40fe3ad401f8959a\",5],\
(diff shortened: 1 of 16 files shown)


client/rpc/src/state/tests.rs
```diff
@@ -402,7 +402,7 @@ fn should_return_runtime_version() {
 
 	let result = "{\"specName\":\"test\",\"implName\":\"parity-test\",\"authoringVersion\":1,\
 		\"specVersion\":1,\"implVersion\":2,\"apis\":[[\"0xdf6acb689907609b\",2],\
-		[\"0x37e397fc7c91f5e4\",1],[\"0xd2bc9897eed08f15\",1],[\"0x40fe3ad401f8959a\",5],\
+		[\"0x37e397fc7c91f5e4\",1],[\"0xd2bc9897eed08f15\",1],[\"0x40fe3ad401f8959a\",4],\
```
tomusdrw: Wouldn't this revert cause issues? Or do we feel believe it wasn't deployed yet? I'd be safe and just bump the version instead of reverting it.

bkchr: It wasn't deployed on Kusama yet and we are still in alpha. This is okay.

PR #5160, opened 2020-03-06
Fix NotificationStreamClosed reported when it shouldn't

tomaka:
For backwards-compatibility reasons, right now the lifetime of a "notifications substream" is still tied to the lifetime of a handshake on the main substream.

At the moment, we report that we opened substreams when a handshake succeeds, and we report that we closed substreams when we disconnected.
However this left a bug: if a node connects then disconnects before the handshake, then we'd report that we closed substreams even though we have never reported that they were open.

While this bug doesn't really have bad consequences, it is the reason why the `polkadot_sub_libp2p_opened_notification_streams` metric is underflowing at the moment.

### DIFF
--- a/client/network/src/protocol.rs
+++ b/client/network/src/protocol.rs
@@ -803,7 +803,6 @@ impl<B: BlockT, H: ExHashT> Protocol<B, H> {
 				remote: peer,
 				protocols: self.protocol_name_by_engine.keys().cloned().collect(),
 			}
-
 		} else {
 			CustomMessageOutcome::None
 		}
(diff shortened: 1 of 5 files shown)


tomaka:
Done

client/network/src/protocol.rs
```diff
@@ -796,7 +796,16 @@ impl<B: BlockT, H: ExHashT> Protocol<B, H> {
...
 				peerset: self.peerset_handle.clone(),
-			}, peer);
+			}, &peer);
+
+			// Notify all the notification protocols as closed.
+			CustomMessageOutcome::NotificationStreamClosed {
+				remote: peer,
+				protocols: self.protocol_name_by_engine.keys().cloned().collect(),
+			}
+
```
bkchr:
```suggestion
```

gavofyork:
CI failing.

bkchr:
@gavofyork CI is happy now :)

PR #5161, opened 2020-03-06
client: fix notification sinks leak during initial sync

andresilva:
During initial sync we won't send any import or finality notifications and we also won't cleanup any closed sinks in the process. This can lead to a temporary increase of unused notification sinks since they will only be cleaned up after sync finishes. This PR makes sure sinks are always pruned even if no notifications are sent.


client/src/client.rs
```diff
@@ -939,7 +946,27 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
+			None => {
+				// cleanup any closed import notification sinks since we won't
+				// be sending any notifications below which would remove any
+				// closed sinks. this is necessary since during initial sync we
+				// won't send any import notifications which could lead to a
+				// temporary leak of closed/discarded notification sinks (e.g.
+				// from consensus code).
+				self.import_notification_sinks
+					.lock()
+					.retain(|sink| !sink.is_closed());
```
tomusdrw: It feels a bit wasteful to iterate over all sinks on EVERY block import now, but I suppose the list should have at most tens of elements, right?

bkchr: Yes

andresilva: At first I was thinking of doing this only every `N` blocks, but since compared to block import time this operation should be one or two orders of magnitude faster I think it is OK. Yeah, the number of subscribers should be in the tens.

andresilva:
Added test.

PR #5162, opened 2020-03-06
Make `wasm-builder` select dependencies more sensitive

bkchr:
Up to now `wasm-builder` iterated over all packages that were not part
of the wasm workspace and added the files to the `rerun-if-changed`
list. However, this included to many files and resulted in needless
re-compilations in `Cumulus` for example.

### DIFF
--- a/utils/wasm-builder/src/wasm_project.rs
+++ b/utils/wasm-builder/src/wasm_project.rs
@@ -16,7 +16,10 @@
 
 use crate::write_file_if_changed;
 
-use std::{fs, path::{Path, PathBuf}, borrow::ToOwned, process, env};
+use std::{
+	fs, path::{Path, PathBuf}, borrow::ToOwned, process, env, collections::HashSet,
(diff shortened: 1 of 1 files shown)


utils/wasm-builder/src/wasm_project.rs
```diff
@@ -471,11 +513,35 @@ fn generate_rerun_if_changed_instructions(
...
+	// Start with the dependencies of the crate we want to compile for wasm.
+	let mut dependencies = metadata.packages
+		.iter()
+		.find(|p| p.manifest_path == cargo_manifest)
+		.expect("The crate package is contained in its own metadata; qed")
+		.dependencies
+		.iter()
+		.collect::<Vec<_>>();
+
+	// Collect all packages by follow the dependencies of all packages we find.
```
andresilva:
```suggestion
	// Collect all packages by following the dependencies of all packages we find.
```

PR #5166, opened 2020-03-06
Fix a runtime panic in the browser node

expenses:
Calling `std::time::Instant::now` panics in wasm.


client/executor/src/wasm_runtime.rs
```diff
@@ -297,6 +297,7 @@ fn create_versioned_wasm_runtime(
 	host_functions: Vec<&'static dyn Function>,
 	allow_missing_func_imports: bool,
 ) -> Result<VersionedRuntime, WasmError> {
+	#[cfg(not(target_os = "unknown"))]
```
bkchr: Isn't there any better way? Wasn't there a crate for platform in-depend timers?

expenses: Yeah, there's `wasm_timer`, but the _better_ way to do this is to not sure the runtime in the light client at all.

PR #5168, opened 2020-03-06
State metrics possible changes

cheme:
This is a small PR where I tried to implement my comments on #4562. CC\ @NikVolf 
Currently I am not very happy with the way things are registered for the state machine overlay writes, but I am not very sure it is that usefull to collect.
I feel like it is interesting as it can give an idea of the difference between the writes the runtime ask and the actual writes that propagates to rocksdb.
While writing this, I also ask myself if using directly prometheus atomic u64 could not be more straightforward.
So I am putting the PR in InProgress state for now, @pscott I think you are also touching this code.


client/db/src/lib.rs
```diff
@@ -1116,21 +1124,35 @@ impl<Block: BlockT> Backend<Block> {
...
 				for (key, (val, rc)) in operation.db_updates.drain() {
 					if rc > 0 {
 						ops += 1;
 						bytes += key.len() as u64 + val.len() as u64;
 
 						changeset.inserted.push((key, val.to_vec()));
 					} else if rc < 0 {
-						ops += 1;
-						bytes += key.len() as u64;
+						removal += 1;
```
NikVolf: why removal needs to be counted differently than just writes?

cheme: Due to the different ways it can be handled in https://github.com/paritytech/substrate/blob/a439a7aa5a9a3df2a42d9b25ea04288d3a0866e8/client/state-db/src/lib.rs#L255
but like I said in https://github.com/paritytech/substrate/pull/4562#discussion_r386535588 we could also use the same counter but only register the ops when we do not run in archive mode; I do not really have a preference: on one hand it is an additional counter, on the other it is a bit implementation related.

NikVolf: Yeah thanks, nice

bkchr:
@NikVolf @cheme what are we waiting for?

NikVolf:
@bkchr this is marked "in progress"?

bkchr:
There was no progress in 18 days. 🤷‍♂️

bkchr:
So I wanted to ask what the status is.

cheme:
Oh sorry, I was waiting for feedback on wether we want those indicators or not, and then I lost trace of the PR.
This may need more tests (real use testing).
@pscott , is this a blocker? I can close it if it is.

NikVolf:
I would eventually want to move all this stuff to prometheus for production code (but have some performance concerns).

But then we will have issue with measuring this for benchmarks

Perhaps we need to make some abstraction which can either push metrics to prometheus immediately or collect in some local struct and then consumed by benchmarks.

Anyway, this should be ok for now, the above is further considerations

bkchr:
Sounds like you want to create an issue @NikVolf? :D

client/db/src/storage_cache.rs
```diff
@@ -852,8 +861,13 @@ impl<S: StateBackend<HashFor<B>>, B: BlockT> StateBackend<HashFor<B>> for Syncin
 			.as_trie_backend()
 	}
 
+	fn register_overlay_stats(&mut self, stats: &sp_state_machine::StateMachineStats) {
+		self.caching_state().register_overlay_stats(stats);
+	}
+
 	fn usage_info(&self) -> sp_state_machine::UsageInfo {
 		self.caching_state().usage_info()
+
```
bkchr:
```suggestion
```

client/db/src/lib.rs
```diff
@@ -1676,6 +1700,7 @@ impl<Block: BlockT> sc_client_api::backend::Backend<Block> for Backend<Block> {
 	fn get_import_lock(&self) -> &RwLock<()> {
 		&*self.import_lock
 	}
+
```
```suggestion
```

pscott:
@cheme I don't think this PR is a blocker for my PR. :)

PR #5169, opened 2020-03-06
Add ext_transfer call

### LINKED ISSUE #5027: pallet-contracts: ext_transfer
Currently, the only way to transfer some funds from one contract to another is using a `ext_call`. For use-cases where only transfer is required this can be problematic.

One can object introduction of such a function since that would make it possible to send funds to a contract which actively rejects that. That however is reality: pallet-contracts is built on top of FRAME and its contract is basically an account with attached code and storage. It is possible that some other pallet, internally or as a result of an extrinsic execution, can potentially increase the balance of a contract and there is no obligation for invoking a callback on such a change. Moreover, it is not clear how such a callback invoke a contract, or namely, how it would fund its execution with gas/weight.

athei:
This PR adds the following contract syscall:

```
ext_transfer(account_ptr: u32, account_len: u32, value_ptr: u32, value_len: u32);
```

It behaves similarly to `ext_call`. The differences are:

- Saves three arguments `input_ptr, input_len, gas_limit`.
- Garanteed to never execute code at the destination account (in case it is a contract).

# Motivation

Using `ext_call` for balance transfers has one major flaw: You need to know which kind of contract you are calling because otherwise you do not know if you are actually triggering contract code execution for the input you are passing. Nothing stops a contract from acting on an empty input in an arbitrary way.

Sparing the contract the hassle of passing gas costs and saving some overhead is merely a bonus.

closes #5027

### DIFF
--- a/frame/contracts/src/exec.rs
+++ b/frame/contracts/src/exec.rs
@@ -120,6 +120,14 @@ pub trait Ext {
 		input_data: Vec<u8>,
 	) -> Result<(AccountIdOf<Self::T>, ExecReturnValue), ExecError>;
 
+	/// Transfer some amount of funds into the specified account.
+	fn transfer(
+		&mut self,
+		to: &AccountIdOf<Self::T>,
+		value: BalanceOf<Self::T>,
+		gas_meter: &mut GasMeter<Self::T>,
+	) -> Result<(), DispatchError>;
+
 	/// Call (possibly transferring some amount of funds) into the specified account.
 	fn call(
 		&mut self,
@@ -329,6 +337,24 @@ where
 			timestamp: self.timestamp.clone(),
 			block_number: self.block_number.clone(),
 		}
+	}
+
+	/// Transfer balance to `dest` without calling any contract code.
+	pub fn transfer(
+		&mut self,
+		dest: T::AccountId,
+		value: BalanceOf<T>,
+		gas_meter: &mut GasMeter<T>
+	) -> Result<(), DispatchError> {
+		let sender = self.self_account.clone();
+		transfer(
+			gas_meter,
+			TransferCause::Call,
+			&sender,
+			&dest,
+			value,
+			self,
+		)
 	}
 
 	/// Make a call to the specified address, optionally transferring some funds.
@@ -704,6 +730,15 @@ where
 		input_data: Vec<u8>,
 	) -> Result<(AccountIdOf<T>, ExecReturnValue), ExecError> {
 		self.ctx.instantiate(endowment, gas_meter, code_hash, input_data)
+	}
+
+	fn transfer(
+		&mut self,
+		to: &T::AccountId,
+		value: BalanceOf<T>,
+		gas_meter: &mut GasMeter<T>,
+	) -> Result<(), DispatchError> {
+		self.ctx.transfer(to.clone(), value, gas_meter)
 	}
 
 	fn call(
--- a/frame/contracts/src/exec.rs
+++ b/frame/contracts/src/exec.rs
@@ -346,11 +346,10 @@ where
 		value: BalanceOf<T>,
 		gas_meter: &mut GasMeter<T>
 	) -> Result<(), DispatchError> {
-		let sender = self.self_account.clone();
 		transfer(
 			gas_meter,
 			TransferCause::Call,
-			&sender,
+			&self.self_account.clone(),
 			&dest,
 			value,
 			self,
--- a/frame/contracts/src/wasm/runtime.rs
+++ b/frame/contracts/src/wasm/runtime.rs
@@ -405,6 +405,50 @@ define_env!(Env, <E: Ext>,
 		}
 	},
 
+	// Transfer some value to another account.
+	//
+	// If the value transfer was succesful zero is returned. Otherwise a non-zero value is returned.
+	// The scratch buffer is not touched. The receiver can be a plain account or a contract.
+	//
+	// - account_ptr: a pointer to the address of the beneficiary account
+	//   Should be decodable as an `T::AccountId`. Traps otherwise.
+	// - account_len: length of the address buffer.
+	// - value_ptr: a pointer to the buffer with value, how much value to send.
+	//   Should be decodable as a `T::Balance`. Traps otherwise.
+	// - value_len: length of the value buffer.
+	ext_transfer(
+		ctx,
+		account_ptr: u32,
+		account_len: u32,
+		value_ptr: u32,
+		value_len: u32
+	) -> u32 => {
+		let callee: <<E as Ext>::T as frame_system::Trait>::AccountId =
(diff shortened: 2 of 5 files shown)


frame/contracts/src/wasm/runtime.rs
```diff
@@ -393,6 +393,50 @@ define_env!(Env, <E: Ext>,
...
+		let ext = &mut ctx.ext;
+		let result = ctx.gas_meter.with_nested(ctx.gas_meter.gas_left(), |nested_meter| {
+			match nested_meter {
+				Some(nested_meter) => {
+					ext.transfer(
+						&callee,
+						value,
+						nested_meter,
+					)
+				}
```
NikVolf:
```suggestion
				},
```

frame/contracts/src/wasm/runtime.rs
```diff
@@ -393,6 +393,50 @@ define_env!(Env, <E: Ext>,
...
+	// - account_len: length of the address buffer.
+	// - value_ptr: a pointer to the buffer with value, how much value to send.
+	//   Should be decodable as a `T::Balance`. Traps otherwise.
+	// - value_len: length of the value buffer.
+	ext_transfer(
+		ctx,
+		account_ptr: u32,
+		account_len: u32,
+		value_ptr: u32,
+		value_len: u32
```
NikVolf:
```suggestion
		value_len: u32,
```

frame/contracts/src/exec.rs
```diff
@@ -331,6 +339,24 @@ where
...
+		dest: T::AccountId,
+		value: BalanceOf<T>,
+		gas_meter: &mut GasMeter<T>
+	) -> Result<(), DispatchError> {
+		let sender = self.self_account.clone();
+		transfer(
+			gas_meter,
+			TransferCause::Call,
+			&sender,
+			&dest,
```
NikVolf:
```suggestion
		transfer(
			gas_meter,
			TransferCause::Call,
			&self.self_account.clone(),
			&dest,
```

frame/contracts/src/wasm/runtime.rs
athei: I reverted that change because `define_env!` does not allow trailing comma.

frame/contracts/src/wasm/runtime.rs
```diff
@@ -393,6 +393,50 @@ define_env!(Env, <E: Ext>,
...
+					ext.transfer(
+						&callee,
+						value,
+						nested_meter,
+					)
+				},
+				// there is not enough gas to allocate for the nested call.
+				None => Err(sp_runtime::DispatchError::Other("Out of gas.")),
+			}
+		});
```
pepyakin: I think creating a nested gas meter passing whole available gas amount to it is essentially a no-op, isn't it? Can we use `ctx.gas_meter` without creating a nested meter?

frame/contracts/src/wasm/runtime.rs
```diff
@@ -393,6 +393,50 @@ define_env!(Env, <E: Ext>,
...
+
+		match result {
+			Ok(_) => Ok(STATUS_SUCCESS.into()),
+			Err(_) => Ok(TRAP_RETURN_CODE),
```
pepyakin: Ah, weirdly, those constants are for the return type from the `call`/`deploy` function. If we wanted to avoid we would need to create a new set of constants I think to avoid mixing up different constants and change all the other usages (note e.g. `ext_get_storage` just above - it deals with unnamed constants). However, I am OK just leaving it with magic numbers as for now, we can get rid of the magic numbers at once in a following refactoring PRs.

pepyakin: Having read the code closely, I think we don't even need errors here. 

A general strategy dealing with the out-of-gas (OOG) errors is to trap - generally because it is not feasible nor worthwhile to defend against out of gas with the exception of `call`s and `instantiate`s since they actually allow to devote gas budget.

frame/contracts/src/wasm/runtime.rs
athei: You are right. The nesting is not necessary as we are always using all the gas. I changed it.

frame/contracts/src/wasm/runtime.rs
athei: The return type is what I am particular uncertain about. There seems to be no consistent scheme around it.

Regarding your second comment: Do you think we should remove the return type and trap on any transfer error? There are more than OOG errors. These are the errors that can happen during a transfer:

> "not enough gas to pay transfer fee"
> "balance too low to send value"
> "value too low to create account"
> "destination balance too high to receive value"

You still want to trap?

athei: Result of a discussion with @pepyakin : We use explicit `0` and `1` as return values for now and re-iterate later when we refactor and unify error reporting. Already pushed that change.

frame/contracts/src/wasm/runtime.rs
```diff
@@ -393,6 +393,37 @@ define_env!(Env, <E: Ext>,
...
+		account_len: u32,
+		value_ptr: u32,
+		value_len: u32
+	) -> u32 => {
+		let callee: <<E as Ext>::T as frame_system::Trait>::AccountId =
+			read_sandbox_memory_as(ctx, account_ptr, account_len)?;
+		let value: BalanceOf<<E as Ext>::T> =
+			read_sandbox_memory_as(ctx, value_ptr, value_len)?;
+
+		let ext = &mut ctx.ext;
```
pepyakin: Can we omit this intermediate binding? I thought that it should work with just `ctx.ext.transfer`.

athei: You are right. Thanks. I removed it.

jonahb27:
Is there any documentation and/or example of using this function?

athei:
All the documentation for contract callable functions you can find inside the `define_env!` macro. This is the normative documentation. For `seal_transfer`:
https://github.com/paritytech/substrate/blob/3f49041b474f224ab08b29526364e77333ce62d9/frame/contracts/src/wasm/runtime.rs#L588-L617

That said, I guess you are interested in how to use this function from ink! rather. For that I can say that you can look [here](https://paritytech.github.io/ink/ink_core/env/fn.transfer.html). This is a flat transfer without any code executed at the receiver. The receiver doesn't even need to be a contract. It can be a plain account. A contract is neither notified nor can it prevent anyone from depositing balance into its account. Therefore what you rather want to do is **call** another contract and supply the amount of balance you want to transfer while calling it. The contract can then check how much value you you sent along with the call.

jonahb27:
Thank you!

drewstone:
@athei is this available on rc3 or only on rc5?

athei:
This PR predates `rc3`. So it is available there.

PR #5170, opened 2020-03-06
Fix staking bug

thiolliere:
the property `"Error: start_session_index must be set for active_era + 1"` is false, there can be no future era planned yet, there is a future era planned only one session before the end of the era is kusama.

Although this was a bug, the behavior wasn't broken because in case no era is planned, then index 0 is returned and the comparison:
```rust
if next_active_era_start_session_index == session_index + 1 {
```
would return false which is good because no era was planned thus, the session doesn't end an era.

### DIFF
--- a/frame/session/src/lib.rs
+++ b/frame/session/src/lib.rs
@@ -156,6 +156,8 @@ pub trait SessionManager<ValidatorId> {
 	/// `new_index` is strictly greater than from previous call.
 	///
 	/// The first session start at index 0.
+	///
+	/// `new_session(session)` is guaranteed to be called before `end_session(session-1)`.
 	fn new_session(new_index: SessionIndex) -> Option<Vec<ValidatorId>>;
 	/// End the session.
 	///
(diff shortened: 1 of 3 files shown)


frame/session/src/lib.rs
```diff
@@ -156,6 +156,8 @@ pub trait SessionManager<ValidatorId> {
 	/// `new_index` is strictly greater than from previous call.
 	///
 	/// The first session start at index 0.
+	///
+	/// `new_session(session)` is guaranteed to be called before `end_session(session-1)`.
```
thiolliere: this is not related to the bug, but as long as staking implement SessionManager with this constraint, I think it must stay here.

frame/staking/src/lib.rs
```diff
@@ -1659,17 +1659,12 @@ impl<T: Trait> Module<T> {
...
-						);
-						0
-					});
-
-			if next_active_era_start_session_index == session_index + 1 {
-				Self::end_era(active_era, session_index);
+			{
+				if next_active_era_start_session_index == session_index + 1 {
+					Self::end_era(active_era, session_index);
+				}
```
thiolliere: I don't know how to add a test for this, actually if I replace this unwrap_or_else by an expect the test-suite correctly fail.

To check other assumption like this I replaced occurence of unwrap_or.. by unwrap to check if test suite fail, and it succeed.
Maybe I should do some test_assert?

PR #5171, opened 2020-03-06
Increase node template max block weight

JoshOrndorff:
This PR increases the node template's maximum block weight from 1 million (`1_000_000`) to 1 billion (`1_000_000_000`) to match the main Substrate node. The former lower weight is the exact weight of a balance transfer which means that after a timestamp is in the block there isn't enough room left for even a single balance transfer.

A mystery that I haven't solved is why this just became a problem when none of those weight have changed in 9 months since https://github.com/paritytech/substrate/pull/3157

Cross reference: https://github.com/substrate-developer-hub/substrate-node-template/pull/25

Dumb question: Does this require a spec version bump? I think it does.

### DIFF
--- a/bin/node-template/runtime/src/lib.rs
+++ b/bin/node-template/runtime/src/lib.rs
@@ -120,7 +120,7 @@ pub fn native_version() -> NativeVersion {
 
 parameter_types! {
 	pub const BlockHashCount: BlockNumber = 250;
-	pub const MaximumBlockWeight: Weight = 1_000_000;
+	pub const MaximumBlockWeight: Weight = 1_000_000_000;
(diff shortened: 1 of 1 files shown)


JoshOrndorff:
Just want to double check on my question before I merge.

> Does this require a spec_version bump?

Initially I was thinking it did need one, but I'm thinking the node _template_ runtime should always stay at version 1 and when individual projects start building from the template, then they can start bumping the version.

riusricardo:
>Initially I was thinking it did need one, but I'm thinking the node template runtime should always stay at version 1 and when individual projects start building from the template, then they can start bumping the version.

@JoshOrndorff AFAIK you are correct here and there is no need to pump the version.

PR #5172, opened 2020-03-06
Fix possible corner case in revalidation

NikVolf:
It might be the case when revalidation is queued across multiple blocks


bkchr:
Can we get a test?

PR #5176, opened 2020-03-06
node: reset flaming fir testnet

andresilva:
Reset the flaming fir testnet with the latest runtime.


andresilva:
@ddorgan Note that I have changed the chain spec id to `flamingfir6` so you'll need to move the keystores accordingly.

PR #5177, opened 2020-03-06
Use CLI to configure max instances cache

### LINKED ISSUE #5127: Add cli to configure max instances of wasm instance cache.
expose https://github.com/paritytech/substrate/pull/5109#discussion_r386963152 to CLI rather than a const

DarkEld3r:
Closes #5127.


client/executor/src/wasm_runtime.rs
```diff
@@ -130,6 +131,7 @@ impl RuntimeCache {
 		default_heap_pages: u64,
 		host_functions: &[&'static dyn Function],
 		allow_missing_func_imports: bool,
+		max_runtime_instances: Option<usize>,
```
bkchr: Why isn't that parameter part of the struct + why is this an optional?

IMHO, the default value should be set in the cli.

arkpar:
There should be a hard cap. E.g. 256. Setting this too high will degrade performance and we don't want users to shot themselves in the foot. If they specify a value higher than the cap, there should be a warning that a capped value is used instead.

client/cli/src/commands/runcmd.rs
```diff
@@ -265,7 +265,11 @@ pub struct RunCmd {
 		parse(from_os_str),
 		conflicts_with_all = &[ "password-interactive", "password" ]
 	)]
-	pub password_filename: Option<PathBuf>
+	pub password_filename: Option<PathBuf>,
+
+	/// The size of the instances cache.
```
arkpar: Should probably clarify that this is per cached runtime code, not total.

client/executor/src/wasm_runtime.rs
```diff
@@ -62,7 +62,7 @@ struct VersionedRuntime {
 	/// Runtime version according to `Core_version` if any.
 	version: Option<RuntimeVersion>,
 	/// Cached instance pool.
-	instances: RwLock<[Option<Arc<Mutex<Box<dyn WasmInstance>>>>; MAX_INSTANCES]>,
+	instances: RwLock<Vec<Option<Arc<Mutex<Box<dyn WasmInstance>>>>>>,
```
arkpar: I used an array here to avoid an allocation for each call when this is cloned. Unfortunately with `parking_lot` I could not find a way to `try_lock` a mutex and return a locked guard for an clone of the `Arc<Mutex>`. An entire set of instances needs to be cloned so that the `RwLock` can be released for the duration of the call.
Consider using `SmallVec` here instead.

Also, with `Vec` or `SmallVec` you can use `RwLock<Vec<Arc<..` instead of `RwLock<Vec<Option<Arc<..`. Simply allocate a new instance by pushing it to the vector.

client/executor/src/wasm_runtime.rs
DarkEld3r: There are many places where executor is constructed without cli. I have moved default value into cli, but I'm not sure that it is better.

client/service/src/config.rs
```diff
@@ -123,6 +123,10 @@ pub struct Configuration<G, E = NoExtension> {
 	pub tracing_targets: Option<String>,
 	/// Tracing receiver
 	pub tracing_receiver: sc_tracing::TracingReceiver,
+	/// The size of the instances cache.
+	///
+	/// The default value is 2.
```
arkpar: Yet `default()` below sets it to zero

client/executor/src/wasm_runtime.rs
```diff
@@ -130,6 +133,7 @@ impl RuntimeCache {
 		default_heap_pages: u64,
 		host_functions: &[&'static dyn Function],
 		allow_missing_func_imports: bool,
+		max_runtime_instances: usize,
```
arkpar: Should rather be passed to `new` and stored in `self`.

client/cli/src/commands/runcmd.rs
```diff
@@ -265,7 +265,13 @@ pub struct RunCmd {
 		parse(from_os_str),
 		conflicts_with_all = &[ "password-interactive", "password" ]
 	)]
-	pub password_filename: Option<PathBuf>
+	pub password_filename: Option<PathBuf>,
+
+	/// The size of the instances cache for each runtime.
+	///
+	/// The default value is 2 and the values higher than 256 are ignored.
+	#[structopt(long = "max-runtime-instances", default_value = "2")]
```
arkpar: Any particular reason for setting the default to 2 instead of 8 as it was before?

DarkEld3r: I was looking at the `MAX_RUNTIMES` constant when setting the default value. :sweat_smile:

client/service/src/config.rs
DarkEld3r: Nice catch! One more reason why I don't like to set the default value in cli.

client/executor/src/wasm_runtime.rs
DarkEld3r: Constant can still be declared in the executor, but it will result in more dependencies. What do you think?

client/executor/src/wasm_runtime.rs
```diff
@@ -341,7 +349,8 @@ fn create_versioned_wasm_runtime(
 		version,
 		heap_pages,
 		wasm_method,
-		instances: Default::default(),
+		instances: RwLock::new(SmallVec::with_capacity(max_instances)),
```
bkchr: I don't see the advantage of using `SmallVec` when we use `with_capacity` here. And @arkpar comment about cloning the instances isn't solved either.

client/executor/src/wasm_runtime.rs
```diff
@@ -216,7 +223,7 @@ impl RuntimeCache {
 			let instance_pool = runtime.instances.read().clone();
```
bkchr: I think we should change `instances` to 
```
instances: Vec<RwLock<Option<Arc<VersionedRuntime>>>
```

When creating the `VersionedRuntime`, we do the following:
```
let mut instances = Vec::with_capacity(max_instances);
instances.resize(max_instances, Default::default());
```

What is the advantage of this solution?:
- `max_size` field isn't required anymore
- We don't need to clone here the instances anymore. The code would just be:
```
let instance = runtime.instances.iter().
```
- We need to lock on mutex less.

client/executor/src/wasm_runtime.rs
arkpar: There won't be allocation for `max_instances <=8` since that's the stack-based size for `SmallVec`. But yeah, better just start without setting capacity. If `max_size` is greater than 8 it will allocate right away, even though we might never actually need all those instances.

client/executor/src/wasm_runtime.rs
arkpar: Unfortunately this will require `RwLock` to be locked for the duration of the call. See https://github.com/paritytech/substrate/pull/5177#discussion_r389156381

client/executor/src/wasm_runtime.rs
bkchr: `with_capacity` allocates if `max_instances > 8`
https://docs.rs/smallvec/1.2.0/smallvec/struct.SmallVec.html#method.with_capacity

client/executor/src/wasm_runtime.rs
DarkEld3r: Unfortunately mutex is still needed because `Arc<T>` will only implement `Send` and `Sync` as long as the `T` implements `Send` and `Sync`.

Additionally, `RwLock` cannot be cloned, so `resize` won't work.

bkchr: Shouldn't a Box for wasminstance be enough? And also make sure the that wasminstance implements Send? 

Than just use resize_with...

arkpar: `WasmInstance` implements `Send` but not `Sync` indeed, so you can't concurrently access a collection of them.

bkchr: I have implemented it, please take a look.

PR #5179, opened 2020-03-07
Don't include `:code` by default in storage proofs

bkchr:
This pr changes the behavior of automatically including :code in storage proofs. Before this pr, the :code was extracted by the wasm-executor before executing a function in the runtime. Now, the runtime code is passed as a parameter from the outside. This makes it possible to extract the :code from the backend, before the storage recorder tracks all accesses to the backend (which resulted in inclusion of :code in the proof).

However, the execution_proof function of Client still adds the :code to the proof, this is required by the light client, until #5047 is implemented.

Closes: paritytech/cumulus#8

New version of: https://github.com/paritytech/substrate/pull/5060


client/executor/src/wasm_runtime.rs
```diff
@@ -168,19 +144,11 @@ impl RuntimeCache {
...
-				let code = match code {
-					CodeSource::Externalities => {
-						Cow::Owned(ext.original_storage(well_known_keys::CODE)
-							.ok_or(WasmError::CodeNotFound)?)
-					}
-					CodeSource::Custom(code) => {
-						Cow::Borrowed(code)
-					}
-				};
+				let code = runtime_code.fetch_runtime_code().ok_or(WasmError::CodeNotFound)?;
```
arkpar: Maybe `fetch_runtime_code` could return `Cow` so we don't have to clone here?

bkchr: Done

PR #5182, opened 2020-03-07
Offchain signing

### LINKED ISSUE #5213: New abstractions for signing for off-chain workers
Offchain workshop follow up task.

Current offchain primitives to create and send transactions are quite lacking. We have a bunch of patterns that are not covered (like sending unsigned transaction with signed payload). The Conversions between different crypto types are also very cumbersome.
Lastly current approach does not fit well with #4689 

Current work:
https://github.com/paritytech/substrate/tree/td-signed-unsigned

CC @rakanalh

rakanalh:
This PR is a new implementation to offchain worker signing related tasks where:
- The signing interface is simplified
- In addition to submitting a signed / (raw) unsigned transactions, this also introduces a way to submit unsigned transactions with signed payloads.

Resolves #5213

#### TODOs
- [x] Merge latest master
- [x] Remove usage of `CreateTransaction` in tests
- [x] Port the staking pallet if merged.
- [x] Implement companion Polkadot PR.

Polkadot companion: https://github.com/paritytech/polkadot/pull/985


bin/node/runtime/src/lib.rs
```diff
@@ -447,18 +446,81 @@ impl pallet_sudo::Trait for Runtime {
...
+	Call: From<C>,
+{
+
+	type OverarchingCall = Call;
+	type Extrinsic = UncheckedExtrinsic;
+}
+
+pub struct ImOnlineAuthId;
+impl frame_system::offchain::AppCrypto<<Signature as traits::Verify>::Signer, Signature> for ImOnlineAuthId {
+	// TODO [ToDr] Get rid of this trait and instead
```
bkchr: This is already supported and implemented:
https://crates.parity.io/sp_core/crypto/trait.Wraps.html

rakanalh: I will look into this, thanks for the pointer.

bin/node/executor/tests/submit_transaction.rs
```diff
@@ -70,17 +73,10 @@ fn should_submit_signed_transaction() {
...
-		assert_eq!(keys.len(), 3, "Missing keys: {:?}", keys);
-
-		let can_sign = <SubmitTransaction as SubmitSignedTransaction<Runtime, Call>>
-			::can_sign();
-		assert!(can_sign, "Since there are keys, `can_sign` should return true");
-
-		let call = pallet_balances::Call::transfer(Default::default(), Default::default());
-		let results =
-			<SubmitTransaction as SubmitSignedTransaction<Runtime, Call>>::submit_signed(call);
+		let results = Signer::<Runtime, ImOnlineAuthId>::all_accounts()
```
rakanalh: I should probably use a pallet `AppCrypto` AuthId here rather than im-online

bin/node/executor/tests/submit_transaction.rs
```diff
@@ -15,7 +15,7 @@
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
 use node_runtime::{
-	Call, Executive, Indices, Runtime, SubmitTransaction, UncheckedExtrinsic,
+	Executive, Indices, Runtime, UncheckedExtrinsic, ImOnlineAuthId
```
kianenigma:
```suggestion
	Executive, Indices, Runtime, UncheckedExtrinsic, ImOnlineAuthId,
```

bin/node/executor/tests/submit_transaction.rs
kianenigma: This is a test, I think it is fine. You should use some concrete crypto type and currently I think im-online is the only one atm. I introduce a new one here https://github.com/paritytech/substrate/pull/4517 and actually change these tests a bit. You should have a look.

bin/node/runtime/src/lib.rs
```diff
@@ -468,18 +467,81 @@ impl pallet_sudo::Trait for Runtime {
...
 	pub const SessionDuration: BlockNumber = EPOCH_DURATION_IN_SLOTS as _;
 }
 
+impl frame_system::offchain::SigningTypes for Runtime {
+	type Public = <Signature as traits::Verify>::Signer;
+	type Signature = Signature;
+}
+
+impl<LocalCall> frame_system::offchain::CreateSignedTransaction<LocalCall> for Runtime where
+	Call: From<LocalCall>,
```
kianenigma: @bkchr shan't we simply name the thing that `decl_outer_dispatch` created `enum OuterCall` or sth to make this a bit easier to grasp? Having conversations about calls is sometimes hard since you always have to specify which one you mean.

bin/node/runtime/src/lib.rs
```diff
@@ -468,18 +467,81 @@ impl pallet_sudo::Trait for Runtime {
...
+		let address = Indices::unlookup(account);
+		let (call, extra, _) = raw_payload.deconstruct();
+		Some((call, (address, signature.into(), extra)))
+	}
+}
+
+impl<C> frame_system::offchain::SendTransactionTypes<C> for Runtime where
+	Call: From<C>,
+{
+
```
kianenigma:
```suggestion
```

bin/node/runtime/src/lib.rs
```diff
@@ -468,18 +467,81 @@ impl pallet_sudo::Trait for Runtime {
...
+	type OverarchingCall = Call;
+	type Extrinsic = UncheckedExtrinsic;
+}
+
+pub struct ImOnlineAuthId;
```
kianenigma:
```suggestion
pub struct ImOnlineAuthorityId;
```

(I don't think it is a strict rule but I have been pointed numerous times to always use clear english for names except for closures or exceptional cases -- Generally a good practice)

bin/node/runtime/src/lib.rs
```diff
@@ -468,18 +467,81 @@ impl pallet_sudo::Trait for Runtime {
...
+impl frame_system::offchain::AppCrypto<<Signature as traits::Verify>::Signer, Signature> for ImOnlineAuthId {
```
kianenigma:
```suggestion
impl frame_system::offchain::AppCrypto<<Signature as traits::Verify>::Signer, Signature> for ImOnlineAuthorityId {
```

bin/node/runtime/src/lib.rs
```diff
@@ -468,18 +467,81 @@ impl pallet_sudo::Trait for Runtime {
...
+	// That way `AppCrypto` would just have a blanket impl for `RuntimeAppPublic`;
+	type RuntimeAppPublic = ImOnlineId;
+	type GenericSignature = sp_core::sr25519::Signature;
+	type GenericPublic = sp_core::sr25519::Public;
+}
+
+
 impl pallet_im_online::Trait for Runtime {
 	type AuthorityId = ImOnlineId;
+	type OffchainAuthorityId = ImOnlineAuthId;
```
kianenigma:
```suggestion
	type OffchainAuthorityId = ImOnlineAuthorityId;
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -40,20 +40,35 @@
...
+	ensure_none,
+	offchain::{
+		AppCrypto,
+		CreateSignedTransaction,
+		SendUnsignedTransaction,
+		SendRawUnsignedTransaction,
+		SignedPayload,
+		SigningTypes,
+		Signer,
+	}
```
kianenigma:
```suggestion
	self as system, ensure_signed, ensure_none,
	offchain::{
		AppCrypto, CreateSignedTransaction, SendUnsignedTransaction, SendRawUnsignedTransaction, SignedPayload, SigningTypes, Signer,
	}
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -105,6 +116,19 @@ pub trait Trait: frame_system::Trait {
 	type UnsignedInterval: Get<Self::BlockNumber>;
 }
 
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug)]
+pub struct PricePayload<Public, BlockNumber> {
+	block_number: BlockNumber,
+	price: u32,
+	public: Public
```
kianenigma:
```suggestion
	public: Public,
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -105,6 +116,19 @@ pub trait Trait: frame_system::Trait {
 	type UnsignedInterval: Get<Self::BlockNumber>;
 }
 
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug)]
+pub struct PricePayload<Public, BlockNumber> {
```
kianenigma: documentation?

frame/example-offchain-worker/src/lib.rs
```diff
@@ -186,6 +210,22 @@ decl_module! {
 			Ok(())
 		}
 
+		pub fn submit_price_unsigned_with_signed_payload(
+			origin,
+			price_payload: PricePayload<T::Public, T::BlockNumber>,
+			_signature: T::Signature
```
kianenigma:
```suggestion
			_signature: T::Signature,
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -377,13 +409,36 @@ impl<T: Trait> Module<T> {
...
 		// By default unsigned transactions are disallowed, so we need to whitelist this case
 		// by writing `UnsignedValidator`. Note that it's EXTREMELY important to carefuly
 		// implement unsigned validation logic, as any mistakes can lead to opening DoS or spam
 		// attack vectors. See validation logic docs for more details.
-		T::SubmitUnsignedTransaction::submit_unsigned(call)
-			.map_err(|()| "Unable to submit unsigned transaction.".into())
+		//
+		// Method 1: Unsigned transaction / Unsigned payload
+		let _result_raw: Result<(), String> = Signer::<T, T::AuthorityId>::send_raw_unsigned_transaction(call)
+    		.map_err(|()| "Unable to submit unsigned transaction.".into());
```
kianenigma:
```suggestion
    			.map_err(|()| "Unable to submit unsigned transaction.".into());
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -481,6 +536,58 @@ impl<T: Trait> Module<T> {
 			Some(prices.iter().fold(0_u32, |a, b| a.saturating_add(*b)) / prices.len() as u32)
 		}
 	}
+
+	fn validate_transaction_parameters(
+		block_number: &T::BlockNumber,
+		new_price: &u32
```
kianenigma:
```suggestion
		new_price: &u32,
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -481,6 +536,58 @@ impl<T: Trait> Module<T> {
...
+			// This transaction does not require anything else to go before into the pool.
+			// In theory we could require `previous_unsigned_at` transaction to go first,
+			// but it's not necessary in our case.
+			requires: vec![],
+			// We set the `provides` tag to be the same as `next_unsigned_at`. This makes
+			// sure only one transaction produced after `next_unsigned_at` will ever
+			// get to the transaction pool and will end up in the block.
+			// We can still have multiple transactions compete for the same "spot",
+			// and the one with higher priority will replace other one in the pool.
+			provides: vec![codec::Encode::encode(&(KEY_TYPE.0, next_unsigned_at))],
```
kianenigma:
```suggestion
			provides: vec![Encode::encode(&(KEY_TYPE.0, next_unsigned_at))],
```

frame/example-offchain-worker/src/tests.rs
```diff
@@ -71,22 +78,37 @@ impl frame_system::Trait for Test {
...
-	type Public = sp_core::sr25519::Public;
-	type Signature = sp_core::sr25519::Signature;
-
-	fn create_transaction<F: frame_system::offchain::Signer<Self::Public, Self::Signature>>(
-		call: <Extrinsic as ExtrinsicsT>::Call,
-		_public: Self::Public,
-		_account: <Test as frame_system::Trait>::AccountId,
-		nonce: <Test as frame_system::Trait>::Index,
-	) -> Option<(<Extrinsic as ExtrinsicsT>::Call, <Extrinsic as ExtrinsicsT>::SignaturePayload)> {
+type AccountId = <<Signature as Verify>::Signer as IdentifyAccount>::AccountId;
```
kianenigma: So this `Signature` is coming from sp_core::crypto directly. Not sure if we want to showcase this as an example? ain't this an anti-pattern and the mock file should create its own `app_crypto` and use it.

(don't quote me on this :D)

frame/example-offchain-worker/src/tests.rs
```diff
@@ -71,22 +78,37 @@ impl frame_system::Trait for Test {
...
+
+impl frame_system::offchain::SigningTypes for Test {
+	type Public = <Signature as Verify>::Signer;
+	type Signature = Signature;
+}
+
+impl<LocalCall> frame_system::offchain::SendTransactionTypes<LocalCall> for Test where
+	Call<Test>: From<LocalCall>,
+{
+
```
kianenigma:
```suggestion
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -377,13 +409,36 @@ impl<T: Trait> Module<T> {
...
+			},
+			|payload, signature| {
+				Call::submit_price_unsigned_with_signed_payload(payload, signature)
+			}
+		);
+
+
+		Ok(())
+		// T::SubmitUnsignedTransaction::submit_unsigned(call)
+		// 	.map_err(|()| "Unable to submit unsigned transaction.".into())
```
kianenigma:
```suggestion
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -481,6 +536,58 @@ impl<T: Trait> Module<T> {
...
+		//
+		// Note this doesn't make much sense when building an actual oracle, but this example
+		// is here mostly to show off offchain workers capabilities, not about building an
+		// oracle.
+		let avg_price = Self::average_price()
+			.map(|price| if &price > new_price { price - new_price } else { new_price - price })
+			.unwrap_or(0);
+
+		Ok(ValidTransaction {
+			// We set base priority to 2**20 to make sure it's included before any other
```
kianenigma: well... if I want to pedantic I would say this statement is wrong and accurately: 
```suggestion
			// We set base priority to 2**20 to hope it's included before any other
```

because you never know what the priority of other txs are. In a runtime that only has this module.. maybe :D

bin/node/runtime/src/lib.rs
bkchr: https://github.com/paritytech/substrate/blob/master/frame/support/src/dispatch.rs#L1658

You can name it whatever you like in the macro.

frame/example-offchain-worker/src/tests.rs
```diff
@@ -16,21 +16,28 @@
...
+	sr25519::Signature,
 	testing::KeyStore,
 	traits::KeystoreExt,
 };
 use sp_runtime::{
 	Perbill, RuntimeAppPublic,
 	testing::{Header, TestXt},
-	traits::{BlakeTwo256, IdentityLookup, Extrinsic as ExtrinsicsT},
+	traits::{
+		BlakeTwo256,
```
kianenigma: bad formatting here.

frame/im-online/src/lib.rs
```diff
@@ -517,9 +522,20 @@ impl<T: Trait> Module<T> {
...
+		// 1. A (ImOnline) public key to be used by a validator at index `idx` to send im-online
+		//          heartbeats.
 		let authorities = Keys::<T>::get();
+
+		// local keystore
+		//
+		// All `ImOnline` public (+private) keys currently in the local keystore.
 		let mut local_keys = T::AuthorityId::all();
+
+
```
kianenigma:
```suggestion
```

frame/im-online/src/mock.rs
```diff
@@ -40,25 +41,36 @@ impl_outer_dispatch! {
 }
 
 thread_local! {
-	pub static VALIDATORS: RefCell<Option<Vec<u64>>> = RefCell::new(Some(vec![1, 2, 3]));
+	pub static VALIDATORS: RefCell<Option<Vec<UintAuthorityId>>> = RefCell::new(Some(vec![
+		UintAuthorityId(1),
+		UintAuthorityId(2),
+		UintAuthorityId(3)
```
kianenigma:
```suggestion
		UintAuthorityId(3),
```

frame/im-online/src/mock.rs
```diff
@@ -40,25 +41,36 @@ impl_outer_dispatch! {
...
+	]));
+}
+
+pub struct ImOnlineAuthId;
```
kianenigma: name preferably fixed as explained above.

frame/im-online/src/mock.rs
```diff
@@ -161,13 +197,38 @@ impl pallet_authorship::Trait for Runtime {
...
-	type Call = Call;
-	type SubmitTransaction = SubmitTransaction;
 	type ReportUnresponsiveness = OffenceHandler;
 	type SessionDuration = Period;
 }
 
+impl<LocalCall> frame_system::offchain::SendTransactionTypes<LocalCall> for Runtime where
+	Call: From<LocalCall>,
+{
+
```
kianenigma:
```suggestion
```

frame/im-online/src/tests.rs
```diff
@@ -63,7 +63,14 @@ fn should_report_offline_validators() {
...
 		Session::rotate_session();
 		// enact the change and buffer another one
-		let validators = vec![1, 2, 3, 4, 5, 6];
+		let validators = vec![
+			UintAuthorityId(1),
+			UintAuthorityId(2),
+			UintAuthorityId(3),
+			UintAuthorityId(4),
+			UintAuthorityId(5),
+			UintAuthorityId(6)
```
kianenigma:
```suggestion
			UintAuthorityId(6),
```

kianenigma: and lots of other places.

primitives/runtime/src/testing.rs
```diff
@@ -17,11 +17,11 @@
 //! Testing utilities.
 
 use serde::{Serialize, Serializer, Deserialize, de::Error as DeError, Deserializer};
-use std::{fmt::Debug, ops::Deref, fmt, cell::RefCell};
+use std::{fmt::{self, Debug}, ops::Deref, cell::RefCell};
 use crate::codec::{Codec, Encode, Decode};
 use crate::traits::{
 	self, Checkable, Applyable, BlakeTwo256, OpaqueKeys,
-	SignedExtension, Dispatchable,
+	SignedExtension, Dispatchable
```
kianenigma:
```suggestion
	SignedExtension, Dispatchable,
```

frame/system/src/offchain.rs
```diff
@@ -17,349 +17,448 @@
...
-			TAnyAppPublic::from(public).sign(&payload)
-				.map(
-					<<TAnyAppPublic as RuntimeAppPublic>::Signature as AppSignature>
-					 ::Generic::from
-				)
-				.map(Signature::from)
-		})
+impl<T: SigningTypes, C: AppCrypto<T::Public, T::Signature>, X> Signer<T, C, X> {
+	pub fn all_accounts() -> Signer<T, C, ForAll> {
+		Default::default()
```
kianenigma: hmm are these mocks?

frame/im-online/src/mock.rs
```diff
@@ -90,6 +101,31 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
...
+#[derive(Clone, Debug, Eq, PartialEq, Encode, Decode)]
+pub struct PublicWrapper(Public);
+
+impl IdentifyAccount for PublicWrapper {
+	type AccountId = UintAuthorityId;
+	fn into_account(self) -> Self::AccountId {
+		let bytes = self.0.to_vec();
+		let mut truncated = [0u8; 8];
+		truncated.copy_from_slice(bytes.as_slice());
+		UintAuthorityId(u64::from_be_bytes(truncated))
```
kianenigma: why not keep accountId as u64 and return `self.0` here? would probably keep things simpler as working with u64 is easier.

frame/im-online/src/tests.rs
```diff
@@ -78,9 +85,9 @@ fn should_report_offline_validators() {
 				session_index: 2,
 				validator_set_count: 3,
 				offenders: vec![
-					(1, 1),
```
kianenigma: and all of these can be reverted, if you keep using `type AccountId = u64` and modify your impl of `IdentifyAccount`.

frame/system/src/offchain.rs
```diff
@@ -17,349 +17,448 @@
...
-	fn create_transaction<F: Signer<Self::Public, Self::Signature>>(
-		call: Extrinsic::Call,
-		public: Self::Public,
-		account: T::AccountId,
-		nonce: T::Index,
-	) -> Option<(Extrinsic::Call, Extrinsic::SignaturePayload)>;
+pub enum ForAll {}
+pub enum ForAny {}
+
+pub struct Signer<T: SigningTypes, C: AppCrypto<T::Public, T::Signature>, X = ForAny> {
```
kianenigma: documentation + this file in general is overhauled and has no documentation on the new code. Would also make the review process a bit easier to write a bit about them 👍

kianenigma:
I am very much looking forward to this. 

- Formatting issues are a bit too much and should resemble the rest of the code
- I think testing setups can be simplified. 
- The main file ` system/offchain` needs documentation to explain the API.

frame/example-offchain-worker/src/lib.rs
rakanalh: Is there a rule for such a format? i prefer my version TBH as the vertical version is more readable

frame/system/src/offchain.rs
rakanalh: `all_accounts` and `any_account` are part of the Signer interface where you can specify whether you want to sign with all keys or any key. `X` here denotes the mode on which the Signer operates. In `ForAll` case, the signer will operate on all keys (see calls to `for_all`) and the same goes for `for_any`

frame/im-online/src/mock.rs
rakanalh: I followed your suggestion. I initially had to change it all as i wanted AuthorityId to implement `RuntimeAppPublic` as required by the offchain module for `OffchainAuthorityId`  but since i implemented the `PublicWrapper`, the change of `AuthorityId` was no longer needed. Reverted the changes.

frame/im-online/src/mock.rs
```diff
@@ -161,13 +197,37 @@ impl pallet_authorship::Trait for Runtime {
...
+	fn create_transaction<C: frame_system::offchain::AppCrypto<Self::Public, Self::Signature>>(
+		call: Call,
+		_public: PublicWrapper,
+		_account: u64,
+		nonce: u64,
+	) -> Option<(Call, <Extrinsic as ExtrinsicT>::SignaturePayload)> {
+		Some((call, (nonce, ())))
+	}
+}
+
```
kianenigma: I am a bit confused why we need to specify things like signature when a module needs to just send unsigned transactions? the precious model was good in this aspect.

rakanalh: After looking at the stacking pallet, i agree that it shouldn't be required. I'll see what i can do about it.

rakanalh: Implemented `SubmitTransaction` without needing to specify any crypto related stuff (which is the case for the staking pallet)

frame/system/src/mock.rs
```diff
@@ -0,0 +1,43 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Test utilities
+
+use sp_core::{
```
bkchr, line 19: Why can this not be done by `UIntAuthorityId`?

Nevertheless, this file should be moved to somewhere else.

rakanalh: This mock is used in `bin/node/executor/tests/submit_transactions.rs` where the runtime used is the actual node runtime where the multisigner is used.

I tried having this code in `bin/node/executor/tests/common` but faced an issue that i'll talk to you offline about.

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,487 @@
...
+
+	pub fn submit_unsigned_transaction(
+		call: <T as SendTransactionTypes<LocalCall>>::OverarchingCall,
+	) -> Result<(), ()> {
+		SubmitTransaction::<T, LocalCall>::submit_transaction(call, None)
+	}
 }
 
-/// A trait responsible for signing a payload using given account.
+/// Provides an implementation for signing transaction payloads
```
kianenigma, line 60:
```suggestion
/// Provides an implementation for signing transaction payloads.
```

frame/system/src/offchain.rs
kianenigma: right. Doesn't phantomData have a default? in that case you can skip the `impl Default for Signer`.

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,487 @@
...
-		raw_payload.using_encoded(|payload| {
-			let public = public.try_into().ok()?;
-			TAnyAppPublic::from(public).sign(&payload)
-				.map(
-					<<TAnyAppPublic as RuntimeAppPublic>::Signature as AppSignature>
-					 ::Generic::from
-				)
-				.map(Signature::from)
-		})
+impl<T: SigningTypes, C: AppCrypto<T::Public, T::Signature>, X> Default for Signer<T, C, X> {
```
kianenigma, line 77: is this needed? Both option and PhranotomData have a default impl.

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,487 @@
...
+use sp_std::convert::{TryInto, TryFrom};
 use sp_std::prelude::Vec;
-use sp_runtime::app_crypto::{RuntimeAppPublic, AppPublic, AppSignature};
+use sp_runtime::app_crypto::RuntimeAppPublic;
 use sp_runtime::traits::{Extrinsic as ExtrinsicT, IdentifyAccount, One};
 use frame_support::{debug, storage::StorageMap};
 
-/// Creates runtime-specific signed transaction.
+/// Marker enum used to flag using all supported keys to sign a payload.
+pub enum ForAll {}
```
kianenigma, line 27: is there a difference between enum and struct here?

frame/system/src/offchain.rs
kianenigma: it is needed because of `X` I suppose.

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,487 @@
...
+	type Result = Option<(Account<T>, Result<(), ()>)>;
+
+	fn send_unsigned_transaction<TPayload, F>(
+		&self,
+		f: F,
+		f2: impl Fn(TPayload, T::Signature) -> LocalCall,
+	) -> Self::Result
+	where
+		F: Fn(&Account<T>) -> TPayload,
+		TPayload: SignedPayload<T>
```
kianenigma, line 245:
```suggestion
		TPayload: SignedPayload<T>, 
```

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,487 @@
...
+			self.submit_unsigned_transaction(call)
+		})
+	}
+}
+
+/// Account information used for signing payloads
+pub struct Account<T: SigningTypes> {
+	pub index: usize,
+	pub id: T::AccountId,
+	pub public: T::Public,
```
kianenigma, line 284:
```suggestion
	index: usize,
	id: T::AccountId,
	public: T::Public,
```

kianenigma: I don't see them being used externally and there is a `::new`

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,487 @@
...
+		+ IdentifyAccount<AccountId = Self::AccountId>
+		+ core::fmt::Debug
+		+ codec::Codec;
+	type Signature: Clone
+		+ PartialEq
+		+ core::fmt::Debug
+		+ codec::Codec;
+}
+
+/// A wrapper around the transaction and call types
```
kianenigma, line 385:
```suggestion
/// A wrapper around the transaction and call types.
```

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,487 @@
...
+pub trait SendTransactionTypes<LocalCall> {
+	type Extrinsic: ExtrinsicT<Call=Self::OverarchingCall> + codec::Encode;
+	type OverarchingCall: From<LocalCall>;
+}
+
+/// Create signed transaction
```
kianenigma, line 391:
```suggestion
/// Create signed transaction.
```

bin/node/executor/tests/submit_transaction.rs
```diff
@@ -66,23 +71,16 @@ fn should_submit_signed_transaction() {
 	t.register_extension(TransactionPoolExt::new(pool));
 
 	let keystore = KeyStore::new();
-	keystore.write().sr25519_generate_new(Key::ID, Some(&format!("{}/hunter1", PHRASE))).unwrap();
-	keystore.write().sr25519_generate_new(Key::ID, Some(&format!("{}/hunter2", PHRASE))).unwrap();
-	keystore.write().sr25519_generate_new(Key::ID, Some(&format!("{}/hunter3", PHRASE))).unwrap();
+	keystore.write().sr25519_generate_new(AuthorityId::ID, Some(&format!("{}/hunter1", PHRASE))).unwrap();
```
kianenigma, line 74: this code should actually use a real key, no? it is pulling in the substrate node runtime and should use a key from there (or create a custom real key with `app_crypto!`).

bin/node/executor/tests/submit_transaction.rs
```diff
@@ -98,23 +96,25 @@ fn should_submit_signed_twice_from_the_same_account() {
...
 	let keystore = KeyStore::new();
-	keystore.write().sr25519_generate_new(Key::ID, Some(&format!("{}/hunter1", PHRASE))).unwrap();
+	keystore.write().sr25519_generate_new(AuthorityId::ID, Some(&format!("{}/hunter1", PHRASE))).unwrap();
 	t.register_extension(KeystoreExt(keystore));
 
 	t.execute_with(|| {
-		let call = pallet_balances::Call::transfer(Default::default(), Default::default());
-		let results =
-			<SubmitTransaction as SubmitSignedTransaction<Runtime, Call>>::submit_signed(call);
+		let results = Signer::<Runtime, TestAuthorityId>::all_accounts()
```
kianenigma, line 103: yeah this is not correct. the first argument, `Runtime` which implements `SigningTypes` comes from the real runtime whereas the second one that defines the generic crypto types is a mock. I am surprised that it even works?

kianenigma: I would document that this is the signer with the default value for the third parameter, and make a new test with `ForAny` and `ForAll` and demonstrate how the behaviour is different.

frame/example-offchain-worker/src/tests.rs
```diff
@@ -16,21 +16,25 @@
...
 	testing::KeyStore,
 	traits::KeystoreExt,
 };
 use sp_runtime::{
 	Perbill, RuntimeAppPublic,
 	testing::{Header, TestXt},
-	traits::{BlakeTwo256, IdentityLookup, Extrinsic as ExtrinsicsT},
+	traits::{
+		BlakeTwo256, IdentityLookup, Extrinsic as ExtrinsicT,
+		IdentifyAccount, Verify
```
kianenigma, line 36:
```suggestion
		IdentifyAccount, Verify
```

frame/example-offchain-worker/src/tests.rs
kianenigma: ?

bin/node/executor/tests/submit_transaction.rs
```diff
@@ -98,23 +96,25 @@ fn should_submit_signed_twice_from_the_same_account() {
...
 		let len = results.len();
 		assert_eq!(len, 1);
 		assert_eq!(results.into_iter().filter_map(|x| x.1.ok()).count(), len);
 		assert_eq!(state.read().transactions.len(), 1);
 
 		// submit another one from the same account. The nonce should be incremented.
-		let call = pallet_balances::Call::transfer(Default::default(), Default::default());
-		let results =
-			<SubmitTransaction as SubmitSignedTransaction<Runtime, Call>>::submit_signed(call);
+		let results = Signer::<Runtime, TestAuthorityId>::all_accounts()
```
kianenigma, line 114: As mentioned in the discussion with Basti, I think this test file should use its own keys. No need to import system with features.

bin/node/executor/tests/submit_transaction.rs
kianenigma, line 103: So this is the Signer with the default `X`. Should preferably add tests to demonstrate the other two as well.

frame/example-offchain-worker/src/tests.rs
```diff
@@ -71,22 +75,36 @@ impl frame_system::Trait for Test {
...
+
+impl<LocalCall> frame_system::offchain::CreateSignedTransaction<LocalCall> for Test where
+	Call<Test>: From<LocalCall>,
+{
+	fn create_transaction<C: frame_system::offchain::AppCrypto<Self::Public, Self::Signature>>(
+		call: Call<Test>,
+		_public: <Signature as Verify>::Signer,
+		_account: AccountId,
+		nonce: u64,
+	) -> Option<(Call<Test>, <Extrinsic as ExtrinsicT>::SignaturePayload)> {
```
kianenigma, line 107: So here is what I read: 

- via `SigningTypes ` I define the runtime/chain's key types. This seems to happen once per runtime and it makes sense as we only have one crypto per chain. 

- via `SendTransactionTypes ` I declare the call type. This is once per runtime and it makes sense to stick it to the runtime object, `Test` in this case. 

- I think `TestAuthId` can be skipped and maybe implement this `offchain:: AppCrypto ` for `crypto::AuthorityId` in the `mod::crypto` and get it dealt with over there?

frame/im-online/src/lib.rs
```diff
@@ -219,19 +222,13 @@ pub struct Heartbeat<BlockNumber>
 	pub authority_index: AuthIndex,
 }
 
-pub trait Trait: frame_system::Trait + pallet_session::historical::Trait {
+pub trait Trait: SendTransactionTypes<Call<Self>> + pallet_session::historical::Trait {
```
kianenigma, line 225: How does this work? where did system go?

frame/system/src/mock.rs
kianenigma: I think it should go into `bin/node/executor/tests ` and it must be a real crypto used by `node-runtime`, not something dummy.

primitives/runtime/src/testing.rs
```diff
@@ -33,6 +33,12 @@ use crate::transaction_validity::{TransactionValidity, TransactionValidityError,
 #[derive(Default, PartialEq, Eq, Clone, Encode, Decode, Debug, Hash, Serialize, Deserialize, PartialOrd, Ord)]
 pub struct UintAuthorityId(pub u64);
 
+impl fmt::Display for UintAuthorityId {
```
kianenigma, line 36: It is totally fine since it is testing code, but I just wonder why? Debug is already derived.

rakanalh: Leftover from previous iterations... i removed this one.

frame/im-online/src/lib.rs
rakanalh: `pallet_session::historical_trait`'s super::super trait is `frame_system::trait`

rakanalh: Previously, `SendTransactionTypes`'s super trait was `SigningTypes` which in turn had the super trait of `frame_system::trait` but i removed this requirement to make submitting raw transactions easier by not requiring im-online to define `AppCrypto`.

frame/system/src/offchain.rs
rakanalh: These are used in some logging certain account information in the offchain worker example pallet.
https://gitlab.parity.io/parity/substrate/-/jobs/453647

I restored them to public

frame/example-offchain-worker/src/tests.rs
rakanalh: Signature actually comes from:
https://github.com/paritytech/substrate/blob/8c1858048148c1c717123b0e4d360f8ad1be5bf2/bin/node/primitives/src/lib.rs#L31

frame/system/src/mock.rs
rakanalh: Moved

frame/system/src/offchain.rs
rakanalh: You're right in the fact that we are implementing `Default` twice here where X could be either `ForAll` or `ForAny`

bin/node/executor/tests/submit_transaction.rs
rakanalh: I moved the implementation of `TestAuthorityId` to the executor tests

bin/node/executor/tests/submit_transaction.rs
rakanalh: Done.

frame/example-offchain-worker/src/tests.rs
rakanalh: I think this is more like a code organization suggestion. I moved `AppCrypto` implementation to `mod::crypto` but notice the clash in `Signature` naming. I would have preferred keeping it outside where the offchain definitions are in close proximity.

frame/example-offchain-worker/src/tests.rs
rakanalh: Sorry, my previous comment is wrong. I think the point of doing this is to have acces to a type that implements `IdentifyAccount`. The `Public` type generated by `app_crypto` does not implement this directly. `IdentifyAccount` is implemented for the generic `{sr/ed}25519::Public` types.

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,487 @@
...
+
+		x.verify(&payload, &signature)
+	}
+
+}
+
+/// A wrapper around the types which are used for signing transactions.
+/// This trait should be implemented on the runtime.
+pub trait SigningTypes: crate::Trait {
+	//type AccountId;
```
NikVolf, line 371: stray commented code?

bin/node/executor/tests/common.rs
```diff
@@ -15,10 +15,21 @@
...
+	NeverNativeValue, NativeOrEncoded,
+	crypto::KeyTypeId,
+	sr25519::Signature,
+	traits::{CodeExecutor, RuntimeCode},
+};
+use sp_runtime::{
+	ApplyExtrinsicResult,
+	MultiSigner,
+	MultiSignature,
+	traits::{Header as HeaderT, BlakeTwo256}
```
kianenigma, line 31:
```suggestion
	traits::{Header as HeaderT, BlakeTwo256},
```

frame/example-offchain-worker/src/tests.rs
kianenigma: Indeed, you need `IdentifyAccount `. But my point is, that a key type must be used here that resembles that most with the realistic scenario. And I am not sure if the core, untyped sr25519 is the way to go. Maybe you should try and mimic what is happening in `node/runtime`?

frame/example-offchain-worker/src/tests.rs
```diff
@@ -174,9 +185,16 @@ fn should_submit_signed_transaction_on_chain() {
 fn should_submit_unsigned_transaction_on_chain() {
 	let (offchain, offchain_state) = testing::TestOffchainExt::new();
 	let (pool, pool_state) = testing::TestTransactionPoolExt::new();
+
+	// Keystore is required here to be used
+	// if the signed payload / unsigned transaction
+	// method is used
```
kianenigma, line 191: This comment need either be removed or formatted properly.

frame/im-online/src/tests.rs
```diff
@@ -174,8 +174,7 @@ fn late_heartbeat_should_fail() {
 	new_test_ext().execute_with(|| {
 		advance_session();
 		// given
-		VALIDATORS.with(|l| *l.borrow_mut() = Some(vec![1, 2, 4, 4, 5, 6]));
-		assert_eq!(Session::validators(), Vec::<u64>::new());
+		VALIDATORS.with(|l| *l.borrow_mut() = Some(vec![1, 2, 3, 4, 5, 6]));		assert_eq!(Session::validators(), Vec::<u64>::new());
```
kianenigma, line 177:
```suggestion
		VALIDATORS.with(|l| *l.borrow_mut() = Some(vec![1, 2, 3, 4, 5, 6]));		
		assert_eq!(Session::validators(), Vec::<u64>::new());
```

or maybe github got confused here?

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,486 @@
...
-		account: T::AccountId,
-		nonce: T::Index,
-	) -> Option<(Extrinsic::Call, Extrinsic::SignaturePayload)>;
+impl<T, LocalCall> SubmitTransaction<T, LocalCall>
+where
+	T: SendTransactionTypes<LocalCall>,
+{
+	pub fn submit_transaction(
+		call: <T as SendTransactionTypes<LocalCall>>::OverarchingCall,
+		signature: Option<<T::Extrinsic as ExtrinsicT>::SignaturePayload>
```
kianenigma, line 47:
```suggestion
		signature: Option<<T::Extrinsic as ExtrinsicT>::SignaturePayload>,
```

primitives/runtime/src/lib.rs
```diff
@@ -181,18 +181,39 @@ impl From<ed25519::Signature> for MultiSignature {
 	}
 }
 
+impl TryFrom<MultiSignature> for ed25519::Signature {
```
kianenigma, line 184: where are these used?

gavofyork:
@tomusdrw just waiting for your thumbs-up before this can be merged...

primitives/runtime/src/lib.rs
rakanalh: I believe this is where the requirement comes from:
https://github.com/paritytech/substrate/blob/27f900ecad6f8b8851367d5382c2213920f01e8a/frame/system/src/offchain.rs#L330

frame/example-offchain-worker/src/lib.rs
```diff
@@ -196,6 +230,23 @@ decl_module! {
 			Ok(())
 		}
 
+		#[weight = SimpleDispatchInfo::FixedNormal(10_000)]
+		pub fn submit_price_unsigned_with_signed_payload(
+			origin,
+			price_payload: PricePayload<T::Public, T::BlockNumber>,
+			_signature: T::Signature,
+		) -> DispatchResult
```
tomusdrw, line 238:
```suggestion
		) -> DispatchResult {
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -196,6 +230,23 @@ decl_module! {
...
+		{
```
tomusdrw, line 239:
```suggestion
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -331,35 +382,28 @@ impl<T: Trait> Module<T> {
 
 	/// A helper function to fetch the price and send signed transaction.
 	fn fetch_price_and_send_signed() -> Result<(), &'static str> {
-		use system::offchain::SubmitSignedTransaction;
-		// Firstly we check if there are any accounts in the local keystore that are capable of
-		// signing the transaction.
-		// If not it doesn't even make sense to make external HTTP requests, since we won't be able
-		// to put the results back on-chain.
-		if !T::SubmitSignedTransaction::can_sign() {
```
tomusdrw, line 339: This check is missing, we should still show case how to avoid making unnecessary logic (fetching price) when there are no local accounts that can be used for signing.

frame/example-offchain-worker/src/lib.rs
```diff
@@ -385,14 +428,35 @@ impl<T: Trait> Module<T> {
...
+		//
 		// By default unsigned transactions are disallowed, so we need to whitelist this case
 		// by writing `UnsignedValidator`. Note that it's EXTREMELY important to carefuly
 		// implement unsigned validation logic, as any mistakes can lead to opening DoS or spam
 		// attack vectors. See validation logic docs for more details.
-		T::SubmitUnsignedTransaction::submit_unsigned(call)
-			.map_err(|()| "Unable to submit unsigned transaction.".into())
+		//
+		// Method 1: Unsigned transaction / Unsigned payload
+		let _result_raw: Result<(), &'static str> =
```
tomusdrw, line 442: Please don't ignore the error, but rather propagate it upstream via `?`

frame/example-offchain-worker/src/lib.rs
```diff
@@ -385,14 +428,35 @@ impl<T: Trait> Module<T> {
...
+			SubmitTransaction::<T, Call<T>>::submit_unsigned_transaction(call.into())
+				.map_err(|()| "Unable to submit unsigned transaction.");
+
+		// Method 2: Unsigned transction / signed payload
+		let _result_signed_payload = Signer::<T, T::AuthorityId>::all_accounts().send_unsigned_transaction(
```
tomusdrw, line 447: same here

frame/example-offchain-worker/src/tests.rs
tomusdrw: `AccountId` is never supposed to be `app_crypto`. This is a runtime-wide account id, app-crypto is used for pallet-specific cases.

In `node/runtime` we use `MultiSigner/MultiSignature`, but using a single crypto `sr25519` is totally fine.

frame/example-offchain-worker/src/tests.rs
```diff
@@ -71,22 +75,30 @@ impl frame_system::Trait for Test {
...
+}
+
+impl<LocalCall> frame_system::offchain::SendTransactionTypes<LocalCall> for Test where
+	Call<Test>: From<LocalCall>,
+{
+	type OverarchingCall = Call<Test>;
+	type Extrinsic = Extrinsic;
+}
+
+
```
tomusdrw, line 92:
```suggestion
```

frame/staking/src/lib.rs
```diff
@@ -726,7 +726,7 @@ impl<T: Trait> SessionInterface<<T as frame_system::Trait>::AccountId> for T whe
 	}
 }
 
-pub trait Trait: frame_system::Trait {
+pub trait Trait: frame_system::Trait + SendTransactionTypes<Call<Self>> {
```
tomusdrw, line 729:
```suggestion
pub trait Trait: SendTransactionTypes<Call<Self>> {
```

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,486 @@
...
-		call: Extrinsic::Call,
-		public: Self::Public,
-		account: T::AccountId,
-		nonce: T::Index,
-	) -> Option<(Extrinsic::Call, Extrinsic::SignaturePayload)>;
+impl<T, LocalCall> SubmitTransaction<T, LocalCall>
+where
+	T: SendTransactionTypes<LocalCall>,
+{
+	pub fn submit_transaction(
```
tomusdrw, line 45: Missing docs

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,486 @@
...
+		call: <T as SendTransactionTypes<LocalCall>>::OverarchingCall,
+		signature: Option<<T::Extrinsic as ExtrinsicT>::SignaturePayload>,
+	) -> Result<(), ()> {
+		let xt = T::Extrinsic::new(call.into(), signature).ok_or(())?;
+		sp_io::offchain::submit_transaction(xt.encode())
+	}
+
+	pub fn submit_unsigned_transaction(
```
tomusdrw, line 53: ditto

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,486 @@
...
-/// that has ability to sign arbitrary `Payloads`.
+/// Keys used for signing are defined when instantiating the signer object.
+/// Signing can be done using:
 ///
-/// NOTE: Most likely you don't need to implement this trait manually.
-/// It has a blanket implementation for all `RuntimeAppPublic` types,
-/// so it's enough to pass an application-specific crypto type.
+/// - All supported keys in the keystore
+/// - Any of the supported keys in the keystore
+/// - A list of provided keys
```
tomusdrw, line 67:
```suggestion
/// - An intersection of in-keystore keys and the list of provided keys
```

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,486 @@
...
-			::create_transaction::<Self::Signer>(call, public, id.clone(), account.nonce)
-			.ok_or(())?;
-		// increment the nonce. This is fine, since the code should always
-		// be running in off-chain context, so we NEVER persists data.
-		account.nonce += One::one();
-		super::Account::<T>::insert(&id, account);
-
-		let xt = Self::Extrinsic::new(call, Some(signature_data)).ok_or(())?;
-		sp_io::offchain::submit_transaction(xt.encode())
+	pub fn any_account() -> Signer<T, C, ForAny> {
```
tomusdrw, line 91: same

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,486 @@
...
 
-/// Retrieves a public key type for given `SignAndSubmitTransaction`.
-pub type PublicOf<T, Call, X> =
-<
-	<X as SignAndSubmitTransaction<T, Call>>::CreateTransaction
-	as
-	CreateTransaction<T, <X as SignAndSubmitTransaction<T, Call>>::Extrinsic>
->::Public;
+impl<T: SigningTypes, C: AppCrypto<T::Public, T::Signature>, X> Signer<T, C, X> {
+	pub fn all_accounts() -> Signer<T, C, ForAll> {
```
tomusdrw, line 87: Missing docs

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,486 @@
...
+
+/// Account information used for signing payloads
+pub struct Account<T: SigningTypes> {
+	pub index: usize,
+	pub id: T::AccountId,
+	pub public: T::Public,
+}
+
+impl<T: SigningTypes> Account<T> {
+	pub fn new(index: usize, id: T::AccountId, public: T::Public) -> Self {
```
tomusdrw, line 288: Another missing doc. Please check all the functions in that module (or even add `allow(missing_docs)`)

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,486 @@
...
-			!S::all().is_empty()
-		} else {
-			!Self::find_local_keys(accounts).is_empty()
+/// App specific crypto trait that provides sign/verify
+/// abilities to offchain workers. Implementations of this
+/// trait should specify the app-specific public/signature
+/// types.
+pub trait AppCrypto<Public, Signature> {
+	type RuntimeAppPublic: RuntimeAppPublic;
+	// TODO [ToDr] The conversions are messy, clean them up.
```
tomusdrw, line 312: Please move to a separate issue.

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,486 @@
...
+
+		x.verify(&payload, &signature)
+	}
+
+}
+
+/// A wrapper around the types which are used for signing transactions.
+/// This trait should be implemented on the runtime.
+pub trait SigningTypes: crate::Trait {
+	// TODO [ToDr] Could this be just `T::Signature as traits::Verify>::Signer`?
```
tomusdrw, line 371: Same here, either remove or create an issue and add a number for reference (fine with single issue with a checklist of things to address).

frame/system/src/offchain.rs
```diff
@@ -17,351 +17,486 @@
...
+		+ codec::Codec;
+	type Signature: Clone
+		+ PartialEq
+		+ core::fmt::Debug
+		+ codec::Codec;
+}
+
+/// A wrapper around the transaction and call types.
+pub trait SendTransactionTypes<LocalCall> {
+	type Extrinsic: ExtrinsicT<Call=Self::OverarchingCall> + codec::Encode;
```
tomusdrw, line 386: Missing docs for types

frame/staking/src/lib.rs
rakanalh: `SendTransactionTypes` is no longer has `frame_system::Trait` as a super trait to enable submitting unsigned transaction without having to define `AppCrypto`. This is why i have this change here.

rakanalh:
@kianenigma I have made vast improvements on the documentation. Please let me know if you have any feedback.

frame/example-offchain-worker/src/lib.rs
rakanalh: I introduced `can_sign` to `Signer` and re-added this check.

frame/example-offchain-worker/src/lib.rs
```diff
@@ -331,35 +381,35 @@ impl<T: Trait> Module<T> {
...
-		let results = T::SubmitSignedTransaction::submit_signed(call);
+		let results = signer.send_signed_transaction(
+			|_account| {
+				// Received price is wrapped into a call to `submit_price` public function of this pallet.
+				// This means that the transaction, when executed, will simply call that function passing
+				// `price` as an argument.
+				Call::submit_price(price)
+			}
+		);
+		// let results = T::SubmitSignedTransaction::submit_signed(call);
```
tomusdrw, line 408:
```suggestion
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -385,14 +434,49 @@ impl<T: Trait> Module<T> {
...
 		// implement unsigned validation logic, as any mistakes can lead to opening DoS or spam
 		// attack vectors. See validation logic docs for more details.
-		T::SubmitUnsignedTransaction::submit_unsigned(call)
-			.map_err(|()| "Unable to submit unsigned transaction.".into())
+		//
+		// Method 1: Unsigned transaction / Unsigned payload
+		SubmitTransaction::<T, Call<T>>::submit_unsigned_transaction(call.into())
+			.map_err(|()| "Unable to submit unsigned transaction.")?;
+
+		// Method 2: Unsigned transction / signed payload
```
tomusdrw, line 451:
```suggestion
		// Method 2: Unsigned transaction / signed payload
```

frame/example-offchain-worker/src/lib.rs
```diff
@@ -385,14 +434,49 @@ impl<T: Trait> Module<T> {
...
+		Signer::<T, T::AuthorityId>::all_accounts().send_unsigned_transaction(
+			|account| PricePayload {
+				price,
+				block_number,
+				public: account.public.clone()
+			},
+			|payload, signature| {
+				Call::submit_price_unsigned_with_signed_payload(payload, signature)
+			}
+		).into_iter().fold(Ok(()), |last_res, (_, res)| {
```
tomusdrw, line 474: I think it would be more readable if it was simply a loop. You can also easily collect a `Vec<Result<A, B>>` into `Result<Vec<A>, B>` afair, using `.collect()`. maybe that would work better here?

frame/staking/src/lib.rs
tomusdrw: Fair enough! I saw that `frame_system::Trait` is removed in some other place, but it was only because there is also a historical session bound.

frame/system/src/offchain.rs
```diff
@@ -15,353 +15,609 @@
...
+	/// vector of accounts will be intersected with the supported keys
+	/// in the keystore and the resulting list will be used for signing.
+	pub fn with_filter(mut self, accounts: Vec<T::Public>) -> Self {
+		self.accounts = Some(accounts);
+		self
+	}
+
+	/// Check if there are any keys that could be used for signing.
+	pub fn can_sign(&self) -> bool {
+		return self.accounts.is_some() && self.accounts.as_ref().unwrap_or(&vec![]).len() > 0
```
tomusdrw, line 181: That's not correct. We need to check if there are keys in the intersection. `self.accounts = None` is totally fine.

The simplest way to implement this logic correctly would be to use `for_any` function and just return `true` for any account that is found in the intersection (you would need to implement `for_any` for both `ForAll` and `ForAny`)

Also that reminds me that it would be good to have some tests for new traits in this pallet.

gavofyork:
Test failing...

kianenigma:
Test failures seem to be related.

PR #5183, opened 2020-03-08
Benchmark Staking and Session Pallet

shawntabrizi:
This PR benchmark all extrinsics from the Staking and Session pallet.

# Staking

- [x] Bond
- [x] Bond Extra
- [x] Unbond
- [x] Withdraw Unbonded
- [x] Validate
- [x] Nominate
- [x] Chill
- [x] Set Payee
- [x] Set Controller
- [x]  Set Validator Count
- [x] Force No Eras
- [x] Force New Era
- [x] Force New Era Always
- [x] Set Invulnerables
- [x] Force Unstake
- [x] Cancel Deferred Slash
- [x] Payout Validator
- [x] Payout Nominator
- [x] Rebond
- [x] Set History Depth
- [x] Reap Stash

We have also benchmarked other critical functionality in staking:

- [x] New Era
- [x] Do Slash

# Session

- [x] Set Keys
- [x] Purge Keys

Critically, we are missing the `on_initialize` benchmarks for Session. We will probably address this in a follow up.

Additionally, we had to separate the benchmarks for session into its own crate because of cyclic rust dependencies. So we don't follow the "normal" pattern for benchmarks in this case.


frame/staking/src/lib.rs
```diff
@@ -1284,6 +1286,17 @@ decl_module! {
 			ForceEra::put(Forcing::ForceNew);
 		}
 
+		/// Force there to be a new era at the end of sessions indefinitely.
```
kianenigma: I had a similar dilemma of wanting to re-order all of these but I ended up not doing it since it might mess with ongoing proposals with changing the call index. Should make sure that won't happen now. We usually have `staking.setValidatorCount` in the live proposals.

shawntabrizi: ah that is a good point. Can revert these changes if needed

shawntabrizi:
This is ready for review.

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+	let stash = create_funded_user::<T>("stash", n);
+	let controller = create_funded_user::<T>("controller", n);
+	let controller_lookup: <T::Lookup as StaticLookup>::Source = T::Lookup::unlookup(controller.clone());
+	let reward_destination = RewardDestination::Staked;
+	let amount = T::Currency::minimum_balance() * 10.into();
+	Staking::<T>::bond(RawOrigin::Signed(stash.clone()).into(), controller_lookup, amount, reward_destination)?;
+	return Ok((stash, controller))
+}
+
+fn create_stash_controller2<T: Trait>(n: u32) -> Result<(T::AccountId, T::AccountId), &'static str> {
```
kianenigma: I see only the strings are different and that can be a param to `format!`. Can we do better? probably.

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+use crate::Module as Staking;
+use frame_system::Module as System;
+
+const SEED: u32 = 0;
+
+fn create_funded_user<T: Trait>(string: &'static str, n: u32) -> T::AccountId {
+	let user = account(string, n, SEED);
+	let balance = T::Currency::minimum_balance() * 100.into();
+	T::Currency::make_free_balance_be(&user, balance);
+	return user
```
kianenigma:
```suggestion
	user
```

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+	for i in 0 .. max {
+		let (stash, controller) = create_stash_controller::<T>(i)?;
+		let validator_prefs = ValidatorPrefs {
+			commission: Perbill::from_percent(50),
+		};
+		Staking::<T>::validate(RawOrigin::Signed(controller).into(), validator_prefs)?;
+		let stash_lookup: <T::Lookup as StaticLookup>::Source = T::Lookup::unlookup(stash);
+		validators.push(stash_lookup);
+	}
+	return Ok(validators)
```
kianenigma:
```suggestion
	Ok(validators)
```

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+	let controller = create_funded_user::<T>("controller2", n);
+	let controller_lookup: <T::Lookup as StaticLookup>::Source = T::Lookup::unlookup(controller.clone());
+	let reward_destination = RewardDestination::Staked;
+	let amount = T::Currency::minimum_balance() * 10.into();
+	Staking::<T>::bond(RawOrigin::Signed(stash.clone()).into(), controller_lookup, amount, reward_destination)?;
+	return Ok((stash, controller))
+}
+
+fn create_validators<T: Trait>(max: u32) -> Result<Vec<<T::Lookup as StaticLookup>::Source>, &'static str> {
+	let mut validators: Vec<<T::Lookup as StaticLookup>::Source> = Vec::new();
```
kianenigma:
```suggestion
	let mut validators: Vec<<T::Lookup as StaticLookup>::Source> = Vec::with_capacity(max);
```

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+	for i in 0 .. max {
+		let (stash, controller) = create_stash_controller::<T>(i)?;
+		let validator_prefs = ValidatorPrefs {
+			commission: Perbill::from_percent(50),
+		};
```
kianenigma:
```suggestion
		let validator_prefs = Default::default()
```

(unless if it matters for some benches? maybe not)

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+		Staking::<T>::validate(RawOrigin::Signed(controller).into(), validator_prefs)?;
+		let stash_lookup: <T::Lookup as StaticLookup>::Source = T::Lookup::unlookup(stash);
+		validators.push(stash_lookup);
+	}
+	return Ok(validators)
+}
+
+// This function generates v validators each of whom have n nominators ready for a new era.
+pub fn create_validators_with_nominators_for_era<T: Trait>(v: u32, n: u32) -> Result<(), &'static str> {
+	let mut validators: Vec<T::AccountId> = Vec::new();
```
kianenigma:
```suggestion
	let mut validators: Vec<T::AccountId> = Vec::with_capacity(v);
```

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+	let controller_lookup: <T::Lookup as StaticLookup>::Source = T::Lookup::unlookup(controller.clone());
+	let reward_destination = RewardDestination::Staked;
+	let amount = T::Currency::minimum_balance() * 10.into();
+	Staking::<T>::bond(RawOrigin::Signed(stash.clone()).into(), controller_lookup, amount, reward_destination)?;
+	return Ok((stash, controller))
+}
+
+fn create_validators<T: Trait>(max: u32) -> Result<Vec<<T::Lookup as StaticLookup>::Source>, &'static str> {
+	let mut validators: Vec<<T::Lookup as StaticLookup>::Source> = Vec::new();
+	for i in 0 .. max {
```
kianenigma: this could also very well use a _map and collect_ instead of _for and push_ but that's up to you.

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+}
+
+// This function generates v validators each of whom have n nominators ready for a new era.
+pub fn create_validators_with_nominators_for_era<T: Trait>(v: u32, n: u32) -> Result<(), &'static str> {
+	let mut validators: Vec<T::AccountId> = Vec::new();
+
+	// Create v validators
+	for i in 0 .. v {
+		let (v_stash, v_controller) = create_stash_controller::<T>(i)?;
+		let validator_prefs = ValidatorPrefs {
```
kianenigma: same, if you agree.

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+			commission: Perbill::from_percent(50),
+		};
+		Staking::<T>::validate(RawOrigin::Signed(v_controller.clone()).into(), validator_prefs)?;
+		let stash_lookup: <T::Lookup as StaticLookup>::Source = T::Lookup::unlookup(v_stash.clone());
+		validators.push(v_controller.clone());
```
kianenigma: is this used anywhere?

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+	
+	ValidatorCount::put(v);
+
+	Ok(())
+}
+
+// This function generates one validator being nominated by n nominators.
+// It starts an era and creates pending payouts.
+pub fn create_validator_with_nominators<T: Trait>(n: u32, upper_bound: u32) -> Result<T::AccountId, &'static str> {
+	let mut validators: Vec<T::AccountId> = Vec::new();
```
kianenigma: based on the comment describing the function I am not sure if a vector and assering its length to be one is needed.

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+	};
+
+	let current_era = CurrentEra::get().unwrap();
+	ErasRewardPoints::<T>::insert(current_era, reward);
+
+	// Create reward pool
+	let total_payout = T::Currency::minimum_balance() * 1000.into();
+	<ErasValidatorReward<T>>::insert(current_era, total_payout);
+
+	Ok(validators[0].clone())
```
kianenigma: yeah the assertion has some sense but this is most wrong.

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+
+	new_era {
+		let v in 1 .. 10;
+		let n in 1 .. 100;
+		MinimumValidatorCount::put(0);
+		create_validators_with_nominators_for_era::<T>(v, n)?;
+		let session_index = SessionIndex::one();
+	}: { 
+		let maybe_validators = Staking::<T>::new_era(session_index).ok_or("`new_era` failed")?;
+		assert!(maybe_validators.len() == v as usize);
```
kianenigma:
```suggestion
		let validators = Staking::<T>::new_era(session_index).ok_or("`new_era` failed")?;
		assert!(validators.len() == v as usize);
```

(unwrapped option shall not be called _maybe_)

frame/staking/src/benchmarking.rs
```diff
@@ -0,0 +1,399 @@
...
+	new_era_vary {
+		let v in 1 .. 10;
+		let n in 1 .. 100;
+		let m in 1 .. 10;
+		MinimumValidatorCount::put(m);
+		create_validators_with_nominators_for_era::<T>(v, n)?;
+		let session_index = SessionIndex::one();
+	}: { 
+		let maybe_validators = Staking::<T>::new_era(session_index).ok_or("`new_era` failed")?;
+		assert!(maybe_validators.len() == m as usize);
```
kianenigma:
```suggestion
		let validators = Staking::<T>::new_era(session_index).ok_or("`new_era` failed")?;
		assert!(validators.len() == m as usize);
```

frame/staking/src/benchmarking.rs
shawntabrizi: Yeah you found some artifacts of refactored code. Fixed it.

frame/staking/src/benchmarking.rs
shawntabrizi: From percent 50 allows for nominators to get paid out with this generated validator.

kianenigma:
> A note is that this example is def. not representative since all nominators have one edge and their weight is 100%, this is too simple


I think this is worth fixing, but can also be a follow up fix.

shawntabrizi:
> > A note is that this example is def. not representative since all nominators have one edge and their weight is 100%, this is too simple
> 
> I think this is worth fixing, but can also be a follow up fix.

I have updated the logic so that we create `v` validators, `n` nominators, and then randomly assign `v.min(MAX_NOMINATIONS)` among the available validators. I have written a test too which tests this new functionality.

frame/staking/src/benchmarking.rs
shawntabrizi: my brain likes this better for this situation :)

frame/staking/src/benchmarking.rs
shawntabrizi: Updated so that where I was using `create_stash_controller2` instead I just count down from `u32::max_value()`. There will never be the conflict I was trying to avoid in this case :)

I have removed this redundant function

frame/session/benchmarking/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+
+//! Benchmarks for the Session Pallet.
+// This is separated into its own crate due to cyclic dependency issues.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use sp_std::prelude::*;
+use sp_std::vec;
+
+//use sp_runtime::traits::OnInitialize;
```
kianenigma:
```suggestion
```

frame/session/benchmarking/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+use frame_system::RawOrigin;
+use frame_benchmarking::benchmarks;
+
+use pallet_session::*;
+use pallet_session::Module as Session;
+
+use pallet_staking::{
+	MAX_NOMINATIONS,
+	benchmarking::create_validator_with_nominators,
+	//benchmarking::create_validators_with_nominators_for_era,
```
kianenigma:
```suggestion
```

or..?

bin/node/runtime/src/lib.rs
```diff
@@ -850,6 +850,8 @@ impl_runtime_apis! {
 			repeat: u32,
 		) -> Result<Vec<frame_benchmarking::BenchmarkResults>, sp_runtime::RuntimeString> {
 			use frame_benchmarking::Benchmarking;
+			use pallet_session_benchmarking::Module as SessionBench;
```
kianenigma: This seems out of place, why can't we use the normal `Session` which is already in scope? I suppose it is linked to the dependency of staking and Session. A note explaining why it is the way it is would be good.

frame/session/benchmarking/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+
+//! Benchmarks for the Session Pallet.
+// This is separated into its own crate due to cyclic dependency issues.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use sp_std::prelude::*;
+use sp_std::vec;
+
+//use sp_runtime::traits::OnInitialize;
```
kianenigma:
```suggestion
```

frame/staking/src/tests.rs
```diff
@@ -3039,3 +3039,98 @@ fn set_history_depth_works() {
 		assert!(!<Staking as Store>::ErasTotalStake::contains_key(10 - 5));
 	});
 }
+
+/* Benchmarking Tests */
```
kianenigma: Can this stuff live in the bottom of the benchmarking file?

frame/staking/src/tests.rs
```diff
@@ -3039,3 +3039,98 @@ fn set_history_depth_works() {
...
+
+#[test]
+fn create_validators_with_nominators_for_era_works() {
+	ExtBuilder::default().build().execute_with(|| {
+		let v = 10;
+		let n = 100;
+		let mut validator_iter = Validators::<Test>::enumerate();
+		let mut nominator_iter = Nominators::<Test>::enumerate();
+
+		// Clean up genesis validators and nominators to make test work
```
kianenigma: I added a `has_stakers` to mock builder in offchain phragmen PR, maybe you can do something similar.

shawntabrizi: yes, this is a good idea :)

bin/node/runtime/src/lib.rs
shawntabrizi: Yeah, I had to do some special tricks here because of cyclic dependencies. A comment is a good idea.

frame/staking/src/tests.rs
kianenigma: (I am going to hate myself when merging the conflicts later though :D)

gnunicorn:
tests failing

shawntabrizi:
@gnunicorn fixed!

PR #5186, opened 2020-03-08
Update logic to cancel_deferred_slash

shawntabrizi:
`cancel_deferred_slash` had a flaw where the vec of deferred slashes to remove could have duplicate values, and as a result, mess up the intended behavior.

This makes the extrinsic more restrictive, requiring the sender to submit a:

1. Non-empty
2. Sorted
3. Unique
4. In-Bound

List of deferred slashes to remove.

Tests have been updated so that all of these possible error conditions result in a noop.

### DIFF
--- a/frame/staking/src/lib.rs
+++ b/frame/staking/src/lib.rs
@@ -1329,7 +1329,7 @@ decl_module! {
 				.map(|_| ())
 				.or_else(ensure_root)?;
 
-			ensure!(slash_indices.len() > 0, Error::<T>::EmptyTargets);
+			ensure!(!slash_indices.is_empty(), Error::<T>::EmptyTargets);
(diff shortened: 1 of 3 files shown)


frame/staking/src/lib.rs
```diff
@@ -1460,6 +1456,11 @@ impl<T: Trait> Module<T> {
 		Self::bonded(stash).and_then(Self::ledger).map(|l| l.active).unwrap_or_default()
 	}
 
+	/// Check that list is sorted and has no duplicates.
+	fn is_sorted_and_unique(list: &Vec<u32>) -> bool {
```
rphmeier: 2 code smells: `&Vec` and should be free function

PR #5187, opened 2020-03-09
Remove `secp256k1` from WASM build

bkchr:


primitives/core/src/ecdsa.rs
```diff
@@ -133,10 +133,16 @@ impl sp_std::convert::TryFrom<&[u8]> for Public {
...
-				.map_err(|_| ())
+			#[cfg(feature = "std")]
+			{
+				secp256k1::PublicKey::parse_slice(data, None)
+					.map(|k| k.serialize_compressed())
+					.map(Self)
+					.map_err(|_| ())
+			}
+
+			#[cfg(not(feature = "std"))]
```
gavofyork: unless you can prove that this cannot be called on-chain, then this PR would create a consensus error and fork the network.

bkchr: Fixed.

Issue #5181, opened 2020-03-07
Prometheus metrics: add build info label


gabreal:
The new prometheus metics endpoint exposes metrics currently in the following format:
```
polkadot_block_height_number{endpoint="metrics",instance="XXX",job="job_name"",namespace="namespace",pod="pod_name",service="polkadot-metrics",status="best"}
```

These metrics give no information about which version/build these metrics are from. Therefore it would be very useful to have a default label added to all of those metrics e.g. like it is the case with the dotexporter:
```
dot_chain_block_number{block="head",build="0.7.22-dev-ab76ea7-x86_64-linux-gnu",chain="Kusama CC3",instance="XXX",job="jobb_name",name="parity-polkadot",namespace="namespace",pod="pod_name",service="polkadot-metrics",substrate_ref="a439a7a",version="0.7.22-dev"}
```

The relevant useful labels are:
- `version="0.7.22-dev"`
- `build="0.7.22-dev-ab76ea7-x86_64-linux-gnu"`
- `chain="Kusama CC3"`
also useful but not sure if so easy to get `substrate_ref="a439a7a"` which denotes the substrate build reference.

mxinden:
I would suggest we introduce a static info metric that each node exposes with a `version`, `build` and `chain` label. This metric can then be [grouped](https://prometheus.io/docs/prometheus/latest/querying/operators/#many-to-one-and-one-to-many-vector-matches) onto any other metric of the same target resulting in the above requested extended metric.

gabreal:
Sounds also good to me even though it will make queries more complex. What's the main argument against adding those labels to every metric?

mxinden:
https://github.com/paritytech/substrate/pull/5192 would add the `node_info` metric as suggested in https://github.com/paritytech/substrate/issues/5181#issuecomment-596423977.

> What's the main argument against adding those labels to every metric?

From a  gut feeling this would be something that should happen during relabeling or through service discovery. In addition adding a label to each metric on the target side bloats the overall output. 

I need to put some more thoughts into this. @gabreal could you explain your use cases when you need this information joined with other metrics? Would you like to alert among other things based on the version of Substrate/Polkadot?

Implementation note: One could do this within the [registry](https://docs.rs/prometheus/0.8.0/prometheus/struct.Registry.html#method.new_custom).

gabreal:
[...]

PR #5192, opened 2020-03-09
client/service/src/builder.rs: Add build_info metric

mxinden:
Add static Prometheus metric exposing the chain name, the version and
the commit.

```bash
curl localhost:9615/metrics | grep -E "build_info"

# HELP substrate_build_info A metric with a constant '1' value labeled by name, version, and commit.
# TYPE substrate_build_info gauge
substrate_build_info{commit="607e197f0",name="Substrate Node",version="2.0.0-alpha.3"} 1
```

Fixes https://github.com/paritytech/substrate/issues/5181.

In addition this pull request contains a second commit: 

- client/service/src/builder.rs: Move node_role to static metrics

    The Prometheus metrics `node_role` is static and thus there is no need
to keep a reference of it within `ServiceMetrics`. This follows the
example of the `build_info` metric.

### DIFF
--- a/client/service/src/builder.rs
+++ b/client/service/src/builder.rs
@@ -1013,6 +1013,16 @@ ServiceBuilder<
 
 		// Prometheus metrics
 		let metrics = if let Some(PrometheusConfig { port, registry }) = config.prometheus_config.clone() {
+			// Set static metrics.
+			register(Gauge::<U64>::with_opts(
+				Opts::new(
+					"build_info",
+					"A metric with a constant '1' value labeled by name, version, and commit."
+				).const_label("name", config.impl_name)
+					.const_label("version", config.impl_version)
+					.const_label("commit", config.impl_commit)
+			)?, &registry)?.set(1);
+
 			let metrics = ServiceMetrics::register(&registry)?;
 			metrics.node_roles.set(u64::from(config.roles.bits()));
 			spawn_handle.spawn(
@@ -1023,6 +1033,7 @@ ServiceBuilder<
 		} else {
 			None
 		};
+
 		// Periodically notify the telemetry.
 		let transaction_pool_ = transaction_pool.clone();
 		let client_ = client.clone();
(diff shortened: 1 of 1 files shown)


client/service/src/builder.rs
```diff
@@ -1011,18 +1007,33 @@ ServiceBuilder<
...
 		let metrics = if let Some(PrometheusConfig { port, registry }) = config.prometheus_config.clone() {
+			// Set static metrics.
+			register(Gauge::<U64>::with_opts(
+				Opts::new(
+					"build_info",
+					"A metric with a constant '1' value labeled by name, version, and commit."
+				).const_label("name", config.impl_name)
+					.const_label("version", config.impl_version)
+					.const_label("commit", config.impl_commit),
+			)?, &registry)?.set(1);
```
expenses: This indentation and brace placement here is a bit odd, but I'm not sure what the prettiest way do it would be.

mxinden:
> This indentation and brace placement here is a bit odd, but I'm not sure what the prettiest way do it would be.


@expenses Played around with it, but can't come up with anything better than https://github.com/paritytech/substrate/pull/5192/commits/59c8615565fa61038e8fdc564043ebb941ad1e0a. What do you think?

expenses:
I think it's good enough.

client/service/src/builder.rs
```diff
@@ -1011,18 +1007,34 @@ ServiceBuilder<
...
 
-		// Prometheus metrics
+		// Prometheus metrics.
 		let metrics = if let Some(PrometheusConfig { port, registry }) = config.prometheus_config.clone() {
+			// Set static metrics.
+			register(Gauge::<U64>::with_opts(
+				Opts::new(
+					"build_info",
+					"A metric with a constant '1' value labeled by name, version, and commit."
+				)
```
bkchr: If you remove one tab here, it will not be look that weird.

(All the other ones need to be moved to the left as well)

mxinden:
> Why not use the Registry::custom labels?

@bkchr see https://github.com/paritytech/substrate/issues/5181#issuecomment-596568120 for more details. In the 3 years I have been working with Prometheus I have never seen a `version` label duplicated on every metric. Thus I would like to put more thoughts into this before adding such a label to each metric.

client/service/src/builder.rs
mxinden: If I remove one tab in 1017 it would not align with the opening bracket in 1014. I am not sure I understand your suggestion @bkchr.

bkchr:
Okidoki 👍

PR #5197, opened 2020-03-09
Move runtime upgrade to `frame-executive`

bkchr:
Instead of storing the runtime upgraded in the space of `frame-system`,
this moves it to `frame-executive`. We also start storing the
`spec_version` and `impl_version` the last runtime upgrade was applied.

This scheme has multiple advantages:
- We don't need to make sure that runtime upgrade was set.
- This will work in the future when we want to test a runtime upgrade,
because the `on_runtime_upgrade` will be executed automatically when the
runtime changes.
- It works even when the `LastRuntimeUpgrade` storage value isn't set (expecting that the current version isn't `spec_version == 0`).


frame/executive/src/lib.rs
```diff
@@ -101,6 +98,37 @@ pub type CheckedOf<E, C> = <E as Checkable<C>>::Checked;
...
+	impl_version: u32,
+}
+
+impl LastRuntimeUpgrade {
+	/// Checks if in respect to `current` version there was a runtime upgrade.
+	fn was_upgraded(&self, current: &sp_version::RuntimeVersion) -> bool {
+		if current.spec_version > self.spec_version {
+			true
+		} else if current.spec_version == self.spec_version {
+			current.impl_version > self.impl_version
```
shawntabrizi: Should we run the runtime upgrade logic if impl version is bumped only?

bkchr: Good question. And probably you are right.

primitives/storage/src/lib.rs
```diff
@@ -98,6 +98,11 @@ pub mod well_known_keys {
 	/// Prefix of child storage keys.
 	pub const CHILD_STORAGE_KEY_PREFIX: &'static [u8] = b":child_storage:";
 
+	/// The last time the runtime upgrade happened.
+	///
+	/// Stores the the `impl_version` and the `spec_version` of this runtime.
+	pub const LAST_RUNTIME_UPGRADE: &'static [u8] = b":last_runtime_upgrade:";
```
pepyakin: This is a bit FRAME specific. Other runtimes might want to employ other techniques for detecting a runtime upgrade and they also might not have the notion of `impl_version` and `spec_version`. Probably it would be better to move somewhere under FRAME.

frame/executive/src/lib.rs
```diff
@@ -201,6 +229,25 @@ where
 		);
 	}
 
+	/// Returns if the runtime was upgraded since the last time this function was called.
```
pepyakin: Not sure if it is intentional or not, but this function will return `true` (IIUC) the first time it was called. If that's intentional worth to call it out?

frame/executive/src/lib.rs
bkchr: I thought again about this and I think it is okay to run runtime upgrade logic as well on `impl_version` bump.

I think there are multiple factors:
- Don't add an extra footgun for people that may forgot to increment the `spec_version`.
- With the old `RuntimeUpgraded` flag version, it would also execute the upgrade logic.
- You probably will not issue a runtime upgrade that just contains a bump of the `impl_version`. That is probably just for node releases. (Goes into the direction of point 1)

shawntabrizi: I can see you argument, but also think about a person who forgets to impl spec when they make a breaking change. Their network will fork. So we dont really protect them much here I think.

frame/executive/src/lib.rs
```diff
@@ -101,6 +98,43 @@ pub type CheckedOf<E, C> = <E as Checkable<C>>::Checked;
 pub type CallOf<E, C> = <CheckedOf<E, C> as Applyable>::Call;
 pub type OriginOf<E, C> = <CallOf<E, C> as Dispatchable>::Origin;
 
+
+/// The last time the runtime upgrade happened.
+///
+/// Stores the the `impl_version` and the `spec_version` of this runtime.
```
gavofyork: it should store `spec_name` and `spec_version`; `impl_` stuff is strictly unneeded according to our version semantics.

frame/executive/src/lib.rs
gavofyork: Agree. And as long as we have the semantic difference i think we should be using them.

gavofyork: To mitigate the potential foot-gun, we should ensure that `set_code` verifies that spec_version is different.

bkchr: Addressed.

frame/executive/src/lib.rs
shawntabrizi: So basically, when we run a brand new node, if there is any `on_runtime_upgrade`, it will happen. That seems like a potential issue. We should add into the genesis config generation the correct value then?

bkchr: It will run on genesis. What should be the potential issue? If some upgrade code would run out of nowhere, it means that this upgrade code is already not implemented properly.

gavofyork: I don't think it's especially expected for it to run on genesis. Once Kusama is upgraded to include this change, I would default to not running if there's not already a version stored.

frame/executive/src/lib.rs
```diff
@@ -101,6 +98,13 @@ pub type CheckedOf<E, C> = <E as Checkable<C>>::Checked;
 pub type CallOf<E, C> = <CheckedOf<E, C> as Applyable>::Call;
 pub type OriginOf<E, C> = <CallOf<E, C> as Dispatchable>::Origin;
 
+/// The last time the runtime upgrade happened.
+///
+/// Stores the the `spec_version` as [`LastRuntimeUpgrade`] of this runtime.
+const LAST_RUNTIME_UPGRADE: &'static [u8] = b":last_runtime_upgrade:";
+/// The type used to store the `spec_version` under [`LAST_RUNTIME_UPGRADE`].
+type LastRuntimeUpgrade = Compact<u32>;
```
gavofyork: `spec_name` may reasonably change, so should probably include it (or a hash of it) here.

frame/executive/src/lib.rs
bkchr: Yeah, it would run for the first block.

How should this work for other chains? We somehow need to store the version to know what was the last time we have upgraded the runtime.

I don't see the problem of running it for the first block. If there would be code that runs some weird migrations, it would mean that this would happen at the next upgrade anyway. Upgrade code should be protected in such a way that it can not be executed twice and especially it should not be executed on a fresh genesis.

frame/executive/src/lib.rs
bkchr: `set_code` currently prevents from changing the spec name. When would it happen that the spec name changes?

frame/system/src/lib.rs
```diff
@@ -367,9 +367,6 @@ decl_storage! {
 		/// the `EventIndex` then in case if the topic has the same contents on the next block
 		/// no notification will be triggered thus the event might be lost.
 		EventTopics get(fn event_topics): map hasher(blake2_256) T::Hash => Vec<(T::BlockNumber, EventIndex)>;
-
-		/// A bool to track if the runtime was upgraded last block.
-		pub RuntimeUpgraded: bool;
```
shawntabrizi: One last nit is that we should use your new feature to remove this storage item. Otherwise, it is looking good to me 👍

gavofyork:
Now, will `on_runtime_upgraded` get called on the upgrade that introduces this? :-)

bkchr:
> Now, will `on_runtime_upgraded` get called on the upgrade that introduces this? :-)

Yes, it will run always when it does not find the `:last_runtime_upgrade:` key in storage. (As long as the version of the active runtime is not `0` or the `spec_name` is not `""`, both should be valid for Kusama :D)

It is tested here: https://github.com/paritytech/substrate/pull/5197/files#diff-84f7353adcb1d8bf76b117d23280da44R844

frame/executive/src/lib.rs
gavofyork: The function is called `on_runtime_upgrade`. It's common sense that it shouldn't execute unless a runtime upgrade occured, which is not the case at genesis.

bkchr: We now don't run on genesis.

PR #5198, opened 2020-03-09
Make current extrinsic data accessible to pallets - Fixes #5153

### LINKED ISSUE #5153: Extrinsic ID (hash) of currently executing extrinsic should be accesible in the runtime
Someone submitting an extrinsic can easily obtain the id/hash of that extrinsic before submitting it to a validator by hashing it. (Side note: Why isn't this exposed as `.hash()` on [UncheckedExtrinsic](https://substrate.dev/rustdocs/pre-v2.0-3e65111/sp_runtime/generic/struct.UncheckedExtrinsic.html) while it is on `Block` and `Header`?)

However, from within the runtime, one cannot easily determine the hash of the extrinsic currently being processed. This is important if, for example, some external system refers to extrinsics by these IDs. In our case, some extrinsics create a pending state, which is later resolved by another extrinsic. We could create our own identifiers for this purpose (and is what we do currently), but given extrinsics already have a unique ID it would be desirable to re-use this. However, I can't create a map of <extrinsic id, data> in state without being able to access the ID, hence the need.

If there's some workaround for getting this now, I would love to know. Alternatively, if it can't be done, I'd be happy to look into making a PR for the functionality.

Sushisource:
Hello, this relates to https://github.com/paritytech/substrate/issues/5153 -- I've exposed the data in the most reasonable way I could find, basically sticking to the way the current extrinsic index is stored. If there's a better way to do this, by all means let me know and I'll adjust.


gavofyork:
If you're looking to expose the current extrinsic's hash (which is almost certainly not the solution to your problem), then you should pre-calculate it prior to it being decoded, while the extrinsic is still an opaque `Vec<u8>`; this will avoid unnecessary copies and encode/decode round-trips. While you're there, you could also pass the length in in order to avoid the current `.encode().len()` inefficiency where a superfluous `.encode` is needed in order to figure out the length.

frame/executive/src/lib.rs
```diff
@@ -260,42 +260,44 @@ where
...
-		match Self::apply_extrinsic_with_len(uxt, l, None) {
+		let encoded = uxt.encode();
+		match Self::apply_extrinsic_with_len(uxt, &encoded, false) {
 			Ok(_) => (),
 			Err(e) => { let err: &'static str = e.into(); panic!(err) },
 		}
 	}
 
-	/// Actually apply an extrinsic given its `encoded_len`; this doesn't note its hash.
+	/// Actually apply an extrinsic given its `encoded_len`
```
kianenigma: I am having a hard time following: 

What will be placed in `ExtrinsicData ` in system, and when? 
What will be placed in `EXTRINSIC_DATA ` directly, and when?

(to me it seems like they are duplicate and the latter should be a `EXTRINSIC_HASH` but I want to pose it as a question). 

And whatever is the conclusion should be documented in the function's header.

frame/executive/src/lib.rs
```diff
@@ -367,9 +369,13 @@ mod tests {
...
 	use sp_core::H256;
 	use sp_runtime::{
-		generic::Era, Perbill, DispatchError, testing::{Digest, Header, Block},
+		generic::Era,
+		Perbill,
+		DispatchError,
+		testing::{Digest, Header, Block},
 		traits::{Header as HeaderT, BlakeTwo256, IdentityLookup, ConvertInto},
 		transaction_validity::{InvalidTransaction, UnknownTransaction, TransactionValidityError},
+		traits::Hash
```
kianenigma:
```suggestion
		Perbill, DispatchError,
		generic::Era,
		testing::{Digest, Header, Block},
		traits::{Header as HeaderT, BlakeTwo256, IdentityLookup, ConvertInto, Hash},
		transaction_validity::{InvalidTransaction, UnknownTransaction, TransactionValidityError},
```

frame/executive/src/lib.rs
```diff
@@ -562,6 +589,24 @@ mod tests {
...
+			assert!(Executive::apply_extrinsic(xt).is_ok());
+			let eventrecs = System::events();
+            let found_saw_hash_event = eventrecs.iter().any(|rec| {
+                match rec.event {
+                    MetaEvent::custom(custom::Event::<Runtime>::SawExtHash(h)) => h == xt_hash,
+					_ => false
+				}
+			});
+
+			assert!(found_saw_hash_event)
```
kianenigma: I think reading a mock event is a very long way to check this. You can just read directly from storage at this point and check that correct hash is placed in storage after `Executive::apply_extrinsic(xt)`

Sushisource:
Closing this since we've figured another way to deal with this per the advice that we probably don't want to use extrinsic hashes as unique IDs.

Thanks!

frame/executive/src/lib.rs
Sushisource: Closed this but I wanted to point out I think this is actually a stronger test exactly because it goes a sort of roundabout way - it's demonstrating that the new interfaces I wrote for fetching the extrinsic data are available regardless of how they got there (ie: That it specifically happens because of apply_extrinsic). So the test should remain (relatively) stable in the face of internal changes.

Hopefully that's useful background.

PR #5200, opened 2020-03-10
feat/ocw/bookkeeping

drahnr:
Changes in order to allow writing data during block import to the offchain database, so that the offchain worker has access to this data.

The implementation introduces a new interface accessible from runtime.

Related #3722

polkadot companion: https://github.com/paritytech/polkadot/pull/1025


client/api/src/client.rs
```diff
@@ -63,6 +67,9 @@ pub trait BlockchainEvents<Block: BlockT> {
 	/// finalized block.
 	fn finality_notification_stream(&self) -> FinalityNotifications<Block>;
 
+	/// Get a stream of finality notification originating from the initial block import
```
tomusdrw: It's not `finality`, it's simply block import. I think we should also avoid `initial_sync` in naming. It should be `all_blocks_import_notification_stream` or sth like this and we should emphasise that this is triggered on every imported block, not like in case of regular `import_notification_stream`.

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,92 @@
...
+	#[structopt(
+		long = "offchain-worker",
+		value_name = "ENABLED",
+		possible_values = &OffchainWorkerEnabled::variants(),
+		case_insensitive = true,
+		default_value = "WhenValidating"
+    )]
+    pub enabled: OffchainWorkerEnabled,
+
+	/// Allow writing from the runtime to the offchain worker database directly (buffered).
```
tomusdrw: I'd call this: "Offchain workers indexing API". And emhpasize that if enabled, block import can write (buffered) data directly to offchain worker's DB.

client/service/src/builder.rs
```diff
@@ -924,17 +924,20 @@ ServiceBuilder<
 			let is_validator = config.roles.is_authority();
 
 			let (import_stream, finality_stream) = (
-				client.import_notification_stream().map(|n| ChainEvent::NewBlock {
-					id: BlockId::Hash(n.hash),
-					header: n.header,
-					retracted: n.retracted,
-					is_new_best: n.is_new_best,
-				}),
+				futures::stream::select(client.import_notification_stream(), client.initial_sync_import_notification_stream())
```
tomusdrw: I think we should only use `all_blocks_notification_stream`, not both. There should be an extra parameter being in the stream to know if it's initial sync or not. That in turn should be forwarded to off-chain workers.

client/src/client.rs
```diff
@@ -776,7 +779,24 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
 				retracted,
-			})
+			});
+		} else {
+			match origin {
+				// send a notification for the offchain workers
+				// which would not be notifice in the network initial sync
+				// case otherwise (see previous origin match statement)
+				BlockOrigin::NetworkInitialSync => {
+					self.notify_initial_sync_imported(ImportSummary {
```
tomusdrw: I think that `all_blocks` stream should not contain all this data, but rather since it's on a performance-critical path it should be just minimum necessary (so no storage changes, retracted blocks, etc).

client/src/client.rs
```diff
@@ -964,6 +984,31 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 		Ok(())
 	}
 
+	fn notify_initial_sync_imported(&self, notify_import: ImportSummary<Block>) -> sp_blockchain::Result<()> {
+		if let Some(storage_changes) = notify_import.storage_changes {
```
tomusdrw: We should not trigger storage change events here.

client/src/client.rs
```diff
@@ -1741,6 +1786,12 @@ where
 		stream
 	}
 
+	fn initial_sync_import_notification_stream(&self) -> InitialSyncBlockImportNotifications<Block> {
+		let (sink, stream) = mpsc::unbounded();
```
tomusdrw: We should consider using bounded channel here. If someone enables offchain workers running for initial sync, most likely we probably want to back-pressure block import.

client/cli/src/params/offchain_worker_params.rs
drahnr: Naming is hard. I am open to suggestions for the individual items, it feels like all of them get very lengthy (especially when taking windows cmdlines into consideration)

client/service/src/builder.rs
drahnr: Reverted that.

client/src/client.rs
drahnr: The API for bounded channels is different, as such it would require to have a future for capacity in the mpsc which would cause quite a bit of API change trickling up the call tree. Is that ok?

client/src/client.rs
drahnr: I solved this for now with a new `ChainEvent` variant (`ChainEvent::AnyNewBlock`) being that entity.

drahnr:
Note that this contains changes of #5455

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,85 @@
...
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
```
rphmeier: missing module docs

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,85 @@
+// Copyright 2018-2020 Parity Technologies (UK) Ltd.
```
rphmeier:
```suggestion
// Copyright 2020 Parity Technologies (UK) Ltd.
```

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,85 @@
...
+
+	/// Allow access to offchain workers indexing API
+	///
+	/// Enables a runtime to write directly to a offchain workers
+	/// DB during block import.
+    #[structopt(
+        long = "enable-offchain-worker-indexing",
+        value_name = "ENABLE_OFFCHAIN_WORKER_INDEXING"
+    )]
+	pub indexing_enabled: bool,
```
rphmeier: what's the default for this?

client/service/src/builder.rs
```diff
@@ -265,6 +266,10 @@ fn new_full_parts<TBl, TRtApi, TExecDisp>(
 			extensions,
 			Box::new(tasks_builder.spawn_handle()),
 			config.prometheus_config.as_ref().map(|config| config.registry.clone()),
+			ClientConfig {
+				offchain_worker_enabled : config.offchain_worker.enabled ,
+				offchain_indexing_api: config.offchain_worker.indexing_enabled,
+			}
```
rphmeier:
```suggestion
			},
```

client/service/src/builder.rs
```diff
@@ -888,11 +893,11 @@ ServiceBuilder<
 		let network_status_sinks = Arc::new(Mutex::new(status_sinks::StatusSinks::new()));
 
 		let offchain_storage = backend.offchain_storage();
-		let offchain_workers = match (config.offchain_worker, offchain_storage.clone()) {
-			(true, Some(db)) => {
+		let offchain_workers = match (config.offchain_worker.clone(), offchain_storage.clone()) {
+			(OffchainWorkerConfig{enabled : true, .. }, Some(db)) => {
```
rphmeier:
```suggestion
			(OffchainWorkerConfig {enabled : true, .. }, Some(db)) => {
```

client/service/src/builder.rs
```diff
@@ -888,11 +893,11 @@ ServiceBuilder<
...
 				Some(Arc::new(sc_offchain::OffchainWorkers::new(client.clone(), db)))
 			},
-			(true, None) => {
+			(OffchainWorkerConfig{enabled : true, .. }, None) => {
```
rphmeier:
```suggestion
			(OffchainWorkerConfig {enabled : true, .. }, None) => {
```

client/service/src/builder.rs
```diff
@@ -916,23 +921,23 @@ ServiceBuilder<
 			let is_validator = config.roles.is_authority();
 
 			let (import_stream, finality_stream) = (
-				client.import_notification_stream().map(|n| ChainEvent::NewBlock {
-					id: BlockId::Hash(n.hash),
-					header: n.header,
-					retracted: n.retracted,
-					is_new_best: n.is_new_best,
-				}),
+				client.all_blocks_notification_stream()
```
rphmeier: I am worried that when we are syncing >1k blk/s that this channel will clog up or become a bottleneck. Either that or it will race against state pruning.

Can you elaborate a little bit on the use case for running offchain workers on every block?

client/src/client.rs
```diff
@@ -146,7 +152,22 @@ pub fn new_in_mem<E, Block, S, RA>(
...
 {
-	new_with_backend(Arc::new(in_mem::Backend::new()), executor, genesis_storage, keystore, spawn_handle, prometheus_registry)
+	new_with_backend(
+		Arc::new(in_mem::Backend::new()),
+		executor,
+		genesis_storage,
+		keystore,
+		spawn_handle,
+		prometheus_registry,
+		config)
```
rphmeier:
```suggestion
		config,
```
and move close-paren to next line

client/src/client.rs
```diff
@@ -687,7 +714,19 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 
 				operation.op.update_cache(new_cache);
 
-				let (main_sc, child_sc, tx, _, changes_trie_tx) = storage_changes.into_inner();
+				let (main_sc,
+					child_sc,
+					offchain_sc,
+					tx, _,
+					changes_trie_tx) = storage_changes.into_inner();
```
rphmeier:
```suggestion
				let (
					main_sc,
					child_sc,
					offchain_sc,
					tx, _,
					changes_trie_tx,
				) = storage_changes.into_inner();
```

client/src/client.rs
```diff
@@ -742,21 +781,30 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
+			hash,
+			origin,
+			header,
+			is_new_best,
+			storage_changes,
+			retracted,
+		};
+
+		self.notify_any_block_imported(
+			&import_summary
```
rphmeier:
```suggestion
			&import_summary,
```

client/src/client.rs
```diff
@@ -971,6 +1019,21 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
+
+		let notification = AllBlocksNotification::<Block> {
+			hash: notify_import.hash.clone(),
+			origin: notify_import.origin.clone(),
+			header: notify_import.header.clone(),
+			is_new_best: notify_import.is_new_best,
+		};
+
+		self.all_blocks_notification_sinks.lock()
+			.retain(|sink| sink.unbounded_send(notification.clone()).is_ok());
```
rphmeier: At this stage I am wary of introducing new unbounded senders. @mxinden is already on a holy mission to remove all the existing ones. However, I don't see a way around this without executing this function asynchronously.

client/transaction-pool/src/lib.rs
```diff
@@ -465,6 +465,10 @@ impl<PoolApi, Block> MaintainedTransactionPool for BasicPool<PoolApi, Block>
 					}
 				}.boxed()
 			}
+			ChainEvent::BlockReceived{ .. } => {
```
rphmeier:
```suggestion
			ChainEvent::BlockReceived { .. } => {
```

primitives/api/src/lib.rs
```diff
@@ -354,30 +358,33 @@ pub enum InitializeBlock<'a, Block: BlockT> {
...
+    /// An optional native call that calls the `function`. This is an optimization to call into a
+    /// native runtime without requiring to encode/decode the parameters. The native runtime can
+    /// still be called when this value is `None`, we then just fallback to encoding/decoding the
+    /// parameters.
+    pub native_call: Option<NC>,
+    /// The encoded arguments of the function.
+    pub arguments: Vec<u8>,
+    /// The overlayed changes that are on top of the state.
+    pub overlayed_changes: &'a RefCell<OverlayedChanges>,
+    /// The overlayed changes that are there on top of whatever.
```
rphmeier: incomplete doc comment?

primitives/api/src/lib.rs
```diff
@@ -354,30 +358,33 @@ pub enum InitializeBlock<'a, Block: BlockT> {
...
+    pub offchain_changes: &'a RefCell<InMemOffchainStorage>,
```
rphmeier: This is overlayed over the offchain DB?

primitives/io/src/lib.rs
```diff
@@ -554,7 +554,18 @@ pub trait Hashing {
...
 }
 
+/// Interface that provides functions to access the offchain database access.
+#[runtime_interface]
+pub trait OffchainIndex {
+	/// Write a key value pair to the offchain worker database in a buffered fashion.
+	fn write_kv(&mut self, key : &[u8], value : &[u8]) {
+		self.local_ocw_storage_write_kv(key,value)
+	}
+}
```
rphmeier: for completeness a `remove` function would be good as well

primitives/io/src/lib.rs
```diff
@@ -554,7 +554,18 @@ pub trait Hashing {
 	}
 }
 
+/// Interface that provides functions to access the offchain database access.
+#[runtime_interface]
+pub trait OffchainIndex {
+	/// Write a key value pair to the offchain worker database in a buffered fashion.
+	fn write_kv(&mut self, key : &[u8], value : &[u8]) {
```
rphmeier:
```suggestion
	fn write_kv(&mut self, key: &[u8], value: &[u8]) {
```

primitives/state-machine/src/basic.rs
```diff
@@ -111,6 +111,10 @@ impl From<BTreeMap<StorageKey, StorageValue>> for BasicExternalities {
 }
 
 impl Externalities for BasicExternalities {
+	fn local_ocw_storage_write_kv(&mut self, _key: &[u8], _value: &[u8]) {
+		warn!("basic externalities does not support local_ocw_storage_write_kv")
```
rphmeier: Under what circumstances is this `warn!` triggered? Is this codepath hit whenever the `offchain_indexing` flag is false and we import a block whose runtime calls this function?

rphmeier:
@drahnr Is this still in-progress?

drahnr:
@rphmeier it's in progress due to #5455 / more specifically https://github.com/paritytech/substrate/pull/5455#discussion_r400083652 @tomusdrw 's comment regarding unbounded channels - other than that (and your review comments) it should be ready.

client/cli/src/params/offchain_worker_params.rs
drahnr: `bool` defaults to `false`, should this be implemented explicitly?

client/src/client.rs
drahnr: That was precisely the motivation to start with the approach as is. See #5455 for more details / discussion on how to get rid of it, this is still ongoing.

primitives/api/src/lib.rs
drahnr: essentially the changes to be applied to the ocw db

primitives/api/src/lib.rs
drahnr: Whops

primitives/state-machine/src/basic.rs
drahnr: I removed the warning, should not have been there in the first place.

primitives/io/src/lib.rs
drahnr: That would require a few adjustments, so it might make more sense to handle that in a separate PR?

client/service/src/builder.rs
drahnr: The initial idea was that for an import of all blocks and writing to the offchain worker db, we need a stream that also covers the initial sync - which to this point none did (there is an if case which excludes the initial import from being triggered for sync messages).

client/cli/src/params/offchain_worker_params.rs
drahnr: I did not find any of those `*_params.rs` had any, as such I skipped them.
Added a few lines which should suffice.

primitives/io/src/lib.rs
rphmeier: Sure thing

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,94 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+
+
+//! Offchain worker related configuration parameters.
+//!
+//! A subset of configuration parameters which are relevant to
+//! the inner working of offchain workers. The usage is soley
```
tomusdrw, line 22:
```suggestion
//! the inner working of offchain workers. The usage is solely
```

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,94 @@
...
+//! targetd at handling input parameter parsing providing
```
tomusdrw, line 23:
```suggestion
//! targeted at handling input parameter parsing providing
```

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,94 @@
...
+use sc_service::Configuration;
+
+use sc_service::config::OffchainWorkerConfig;
+use structopt::clap::arg_enum;
+
+
+use crate::error;
+
+arg_enum! {
+	/// Whether off-chain workers are enabled or not as a whole.
```
tomusdrw, line 36:
```suggestion
	/// Whether offchain workers are enabled or not as a whole.
```
Let's keep it consistent at least in this file :)

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,94 @@
...
+		value_name = "ENABLED",
+		possible_values = &OffchainWorkerEnabled::variants(),
+		case_insensitive = true,
+		default_value = "WhenValidating"
+    )]
+    pub enabled: OffchainWorkerEnabled,
+
+	/// Allow access to offchain workers indexing API
+	///
+	/// Enables a runtime to write directly to a offchain workers
```
tomusdrw, line 64:
```suggestion
	/// Enables runtime to write directly to the offchain worker's
```

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,94 @@
...
+	/// DB during block import.
+    #[structopt(
+        long = "enable-offchain-worker-indexing",
```
tomusdrw, line 67:
```suggestion
        long = "enable-offchain-indexing",
```
I'd rename it to "off-chain indexing" cause it's not strictly related to offchain workers and might be misleading in a way that this is some extra API available for OCW, while it's actually for the runtime.

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,94 @@
...
+        value_name = "ENABLE_OFFCHAIN_WORKER_INDEXING"
```
tomusdrw, line 68:
```suggestion
        value_name = "ENABLE_OFFCHAIN_INDEXING"
```

client/src/client.rs
```diff
@@ -687,7 +711,20 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 
 				operation.op.update_cache(new_cache);
 
-				let (main_sc, child_sc, tx, _, changes_trie_tx) = storage_changes.into_inner();
+				let (main_sc,
+					child_sc,
```
tomusdrw, line 715: Funky indentation here.

client/src/client.rs
```diff
@@ -687,7 +711,20 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
+					offchain_sc,
+					tx, _,
+					changes_trie_tx,
+				) = storage_changes.into_inner();
+
+				if let Some(mut offchain_storage) = self.backend.offchain_storage() {
+					offchain_sc.iter().for_each(|(k,v)| {
+						offchain_storage.set(b"block_import_info", k,v)
```
tomusdrw:
```suggestion
						offchain_storage.set(b"block_import_info",k, v)
```

Accessing this prefix will require changes to OCW storage APIs. Currently it's not supported to read arbitrary-prefix data.

tomusdrw: This also does not seem fully correct. We don't want to write directly to DB here, we should rather put this stuff to `opreation.op` or transaction, no so that it can be atomically flushed to the DB with all other changes.

primitives/api/src/lib.rs
```diff
@@ -354,30 +358,33 @@ pub enum InitializeBlock<'a, Block: BlockT> {
...
-	/// The cache for storage transactions.
-	pub storage_transaction_cache: &'a RefCell<StorageTransactionCache<Block, Backend>>,
-	/// Determines if the function requires that `initialize_block` should be called before calling
-	/// the actual function.
-	pub initialize_block: InitializeBlock<'a, Block>,
-	/// The context this function is executed in.
-	pub context: ExecutionContext,
-	/// The optional proof recorder for recording storage accesses.
-	pub recorder: &'a Option<ProofRecorder<Block>>,
+    /// A reference to something that implements the [`Core`] api.
```
tomusdrw, line 361: Spaces!

primitives/externalities/src/lib.rs
```diff
@@ -36,6 +36,9 @@ mod scope_limited;
 ///
 /// Provides access to the storage and to other registered extensions.
 pub trait Externalities: ExtensionStore {
+	/// Write a key value pair to the offchain storage database.
+	fn local_ocw_storage_write_kv(&mut self, key: &[u8], value: &[u8]);
```
tomusdrw, line 40: I think this should rather be called `fn set_offchain_storage(&mut self, key: &[u8], value: Option<&[u8]>)`

primitives/io/src/lib.rs
```diff
@@ -554,7 +554,18 @@ pub trait Hashing {
 	}
 }
 
+/// Interface that provides functions to access the offchain database access.
+#[runtime_interface]
+pub trait OffchainIndex {
+	/// Write a key value pair to the offchain worker database in a buffered fashion.
+	fn write_kv(&mut self, key: &[u8], value: &[u8]) {
```
tomusdrw, line 561: Same here, `write_kv` is not a good name.

primitives/state-machine/src/ext.rs
```diff
@@ -152,6 +157,11 @@ where
 	B: 'a + Backend<H>,
 	N: crate::changes_trie::BlockNumber,
 {
+
+	fn local_ocw_storage_write_kv(&mut self, key: &[u8], value: &[u8]) {
+		self.offchain_overlay.set(b"ocw", key, value);
```
tomusdrw, line 162: The prefix has to be shared

client/src/client.rs
drahnr: Removed.

client/service/src/builder.rs
drahnr: This queue was removed again

primitives/api/src/lib.rs
drahnr: Leftover from the rescue after running `cargo fmt`

client/src/client.rs
drahnr: There is not really a need to have arbitrary data, I assume we can use the currently unused `LOCAL` variant of `StorageKind` (unless that implies a different semantic).

tomusdrw: It is meant to have different semantics. The block import should ideally write to fork-aware DB (`LOCAL`) ,but for now we need to make it write to `PERSISTENT` one as that's the only one implemented.

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,63 @@ impl OffchainStorage for InMemOffchainStorage {
...
+		self.storage.into_iter()
+	}
+
+	/// Iterate over all key value pairs by reference.
+	pub fn iter<'a>(&'a self) -> impl Iterator<Item=(&'a Vec<u8>,&'a OffchainOverlayedChange)> {
+		self.storage.iter()
+	}
+
+	/// Drain all elements of changeset.
+	pub fn drain<'a,'d>(&'a mut self) -> impl Iterator<Item=(Vec<u8>, OffchainOverlayedChange)> + 'd where 'a : 'd {
```
tomusdrw, line 124:
```suggestion
	pub fn drain<'d, 'a: 'd>(&'a mut self) -> impl Iterator<Item=(Vec<u8>, OffchainOverlayedChange)> + 'd {
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,63 @@ impl OffchainStorage for InMemOffchainStorage {
...
+		self.storage.drain()
+	}
+
+	/// Remove a key and it's associated value from the offchain database.
+	pub fn remove(&mut self, prefix: &[u8], key: &[u8]) {
+		let key: Vec<u8> = prefix.iter().chain(key).cloned().collect();
+		let _ = self.storage.insert(key, OffchainOverlayedChange::Remove);
```
tomusdrw, line 131:
```suggestion
  self.storage.insert(key, OffchainOverlayedChange::Remove);
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,63 @@ impl OffchainStorage for InMemOffchainStorage {
...
+	}
+
+	/// Drain all elements of changeset.
+	pub fn drain<'a,'d>(&'a mut self) -> impl Iterator<Item=(Vec<u8>, OffchainOverlayedChange)> + 'd where 'a : 'd {
+		self.storage.drain()
+	}
+
+	/// Remove a key and it's associated value from the offchain database.
+	pub fn remove(&mut self, prefix: &[u8], key: &[u8]) {
+		let key: Vec<u8> = prefix.iter().chain(key).cloned().collect();
```
tomusdrw, line 130:
```suggestion
		let key = prefix.iter().chain(key).cloned().collect();
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,63 @@ impl OffchainStorage for InMemOffchainStorage {
...
+		let _ = self.storage.insert(key, OffchainOverlayedChange::Remove);
+	}
+
+	/// Set the value associated with a key under a prefix to the value provided.
+	pub fn set(&mut self, prefix: &[u8], key: &[u8], value: &[u8]) {
```
tomusdrw, line 135:
```suggestion
	pub fn set(&mut self, prefix: &[u8], key: &[u8], value: Vec<u8>) {
```
Maybe be explicit with the API, perhaps we can save some copying this way.

primitives/state-machine/src/basic.rs
```diff
@@ -111,6 +111,9 @@ impl From<BTreeMap<StorageKey, StorageValue>> for BasicExternalities {
 }
 
 impl Externalities for BasicExternalities {
+	fn set_offchain_storage(&mut self, _key: &[u8], _value: Option<&[u8]>) {
```
tomusdrw, line 114:
```suggestion
	fn set_offchain_storage(&mut self, _key: &[u8], _value: Option<&[u8]>) {}
```

primitives/state-machine/src/basic.rs
```diff
@@ -111,6 +111,9 @@ impl From<BTreeMap<StorageKey, StorageValue>> for BasicExternalities {
...
+	}
```
tomusdrw, line 115:
```suggestion
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,63 @@ impl OffchainStorage for InMemOffchainStorage {
...
+pub enum OffchainOverlayedChange {
+	/// Remove the data associated with the key
+	Remove,
+	/// Overwrite the value of an associated key
+	SetValue(Vec<u8>),
+}
+
+/// In-memory storage for offchain workers.
+#[derive(Debug, Clone)]
+pub struct OffchainOverlayedChanges {
```
tomusdrw, line 95: I think it would be best to implement this as an enum:
```rust
pub enum OffchainOverlayedChanges {
   Disabled,
   Enabled(HashMap<Vec<u8>>, OffchainOverlayedChange),
}
```
This way, in case we know the indexing API is defined all `set/remove` operations are simply no-ops instead of buffering all the data in memory, just to discard them later.

primitives/io/src/lib.rs
```diff
@@ -566,7 +566,23 @@ pub trait Hashing {
 	}
 }
 
+/// Interface that provides functions to access the offchain database access.
+#[runtime_interface]
+pub trait OffchainIndex {
+	/// Write a key value pair to the offchain worker database in a buffered fashion.
+	fn set_offchain_storage(&mut self, key: &[u8], value: &[u8]) {
```
tomusdrw, line 573: Are the generated APIs accessible via module, like so:
```
offchain_index::set_offchain_storage
```
?

If yes, I think we could reduce duplication and just do:
```
offchain_index::set(...)
offchain_index::clear(...)
```

primitives/state-machine/src/ext.rs
```diff
@@ -152,6 +157,15 @@ where
 	B: 'a + Backend<H>,
 	N: crate::changes_trie::BlockNumber,
 {
+
+	fn set_offchain_storage(&mut self, key: &[u8], value: Option<&[u8]>) {
+		const PREFIX : &'static [u8] = b"block-import-info";
```
tomusdrw, line 162: If we add a new prefix here, we need to make it accessible for OCW too. Please alter the `StorageKind` enum and add it there.
For simplicity I'd prefer to put it to `PERSISTENT` storage though than to introduce new variant (with the same semantics as `PERSISTENT`).

client/service/src/config.rs
```diff
@@ -171,6 +171,14 @@ pub enum DatabaseConfig {
 	/// A custom implementation of an already-open database.
 	Custom(Arc<dyn KeyValueDB>),
 }
+/// Configuration of the database of the client.
+#[derive(Clone, Default)]
+pub struct OffchainWorkerConfig {
+	/// If this is allowed.
+	pub enabled: bool,
+	/// allow writes from the runtime to the offchain worker database
+	pub indexing_enabled: bool,
```
rphmeier, line 180: @wpank @shawntabrizi  we're going to want validators to run with this flag. Are there documentation updates corresponding?

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,93 @@
...
+        let enabled = match (&self.enabled, role) {
+			(OffchainWorkerEnabled::WhenValidating, sc_service::Roles::AUTHORITY) => true,
+			(OffchainWorkerEnabled::Always, _) => true,
+			(OffchainWorkerEnabled::Never, _) => false,
+			(OffchainWorkerEnabled::WhenValidating, _) => false,
+		};
+
+        let indexing_enabled = enabled && self.indexing_enabled;
+
+        config.offchain_worker = OffchainWorkerConfig { enabled, indexing_enabled};
```
rphmeier, line 89:
```suggestion
        config.offchain_worker = OffchainWorkerConfig { enabled, indexing_enabled };
```

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,93 @@
...
+
+	/// Allow access to offchain workers indexing API
+	///
+	/// Enables runtime to write directly to the offchain worker's
+	/// DB during block import.
+    #[structopt(
+        long = "enable-offchain-indexing",
+        value_name = "ENABLE_OFFCHAIN_INDEXING"
+    )]
+	pub indexing_enabled: bool,
```
rphmeier, line 69: @wpank @shawntabrizi we're going to want validators to run with this flag. Are there documentation updates corresponding?

@drahnr @tomusdrw Otherwise, would it make sense to automatically set `indexing_enabled` if we are a validator?

client/src/client.rs
```diff
@@ -687,7 +711,22 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
...
+				let (
+					main_sc,
+					child_sc,
+					offchain_sc,
+					tx, _,
+					changes_trie_tx,
+				) = storage_changes.into_inner();
+
+				if self.config.offchain_indexing_api {
+					// if let Some(mut offchain_storage) = self.backend.offchain_storage() {
```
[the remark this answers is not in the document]
rphmeier: 👍 , we should not commit commented-out code

client/src/client.rs
rphmeier: I think this was my suggestion and it's generally the style I follow, except I keep it to only 1 field per line.

primitives/core/src/offchain/storage.rs
```diff
@@ -18,13 +18,32 @@
...
+	pub fn into_iter(self) -> impl Iterator<Item=(Vec<u8>,Vec<u8>)> {
+		self.storage.into_iter()
+	}
+
+	/// Iterate over all key value pairs by reference.
+	pub fn iter<'a>(&'a self) -> impl Iterator<Item=(&'a Vec<u8>,&'a Vec<u8>)> {
+		self.storage.iter()
+	}
+
+	/// Remove a key and it's associated value from the offchain database.
```
rphmeier, line 40:
```suggestion
	/// Remove a key and its associated value from the offchain database.
```

primitives/state-machine/src/basic.rs
```diff
@@ -111,6 +111,8 @@ impl From<BTreeMap<StorageKey, StorageValue>> for BasicExternalities {
 }
 
 impl Externalities for BasicExternalities {
+	fn set_offchain_storage(&mut self, _key: &[u8], _value: Option<&[u8]>) {}
```
rphmeier, line 114: `place_storage` is using the `StorageValue` and `StorageKey` types. Is there a reason to avoid the same here?

primitives/state-machine/src/ext.rs
```diff
@@ -152,6 +157,19 @@ where
 	B: 'a + Backend<H>,
 	N: crate::changes_trie::BlockNumber,
 {
+
+	fn set_offchain_storage(&mut self, key: &[u8], value: Option<&[u8]>) {
+		// write to storage kind `StorageKind::PERSISTENT`````` which
+		// is in turn mapped to `STORAGE_PREFIX` .
+		// MUST be kept in sync with  `sp_offchain::STORAGE_PREFIX`
+		// which can not be used directly, since it causes a circular dependency.
+		const STORAGE_PREFIX : &'static [u8] = b"storage";
```
rphmeier, line 166: Is there a way to test that these are still in-sync? Maybe a unit-test in `sp_offchain`?

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,63 @@ impl OffchainStorage for InMemOffchainStorage {
...
+#[derive(Debug, Clone)]
+pub struct OffchainOverlayedChanges {
+	storage: HashMap<Vec<u8>, OffchainOverlayedChange>,
+}
+
+/// Create an empty set of overlay changes.
+///
+/// The `Default` impl has to be done manually to avoid requirement of
+/// a `Default` impl for `OffchainOverlayedChange`
+impl Default for OffchainOverlayedChanges {
```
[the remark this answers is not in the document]
bkchr: You can not derive `Default` for an enum. The macro does not know what the default value is. (The comment above the `Default` implementation is also not required nor helpful)

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,206 @@ impl OffchainStorage for InMemOffchainStorage {
...
+        if let Some(ref mut iter) = self.inner {
+			iter.next()
+		} else {
+			None
+		}
+    }
+}
+
+impl<'i> OffchainOverlayedChangesIter<'i>
+{
```
bkchr, line 177:
```suggestion
impl<'i> OffchainOverlayedChangesIter<'i> {
```

primitives/runtime-interface/src/lib.rs
```diff
@@ -22,6 +22,8 @@
 //! maps to an external function call. These external functions are exported by the wasm executor
 //! and they map to the same implementation as the native calls.
 //!
+//! On a meta level this implies, the runtime calls the functions provided by the generated code.
```
bkchr, line 25: I don't get this comment? On which meta level?

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,206 @@ impl OffchainStorage for InMemOffchainStorage {
...
+}
+
+/// Iterate over all items while draining them from the collection.
+pub struct OffchainOverlayedChangesDrain<'d> {
+	inner: Option<hash_map::Drain<'d, Vec<u8>,OffchainOverlayedChange>>,
+}
+
+impl<'d> Iterator for OffchainOverlayedChangesDrain<'d> {
+	type Item = (Vec<u8>, OffchainOverlayedChange);
+    fn next(&mut self) -> Option<Self::Item> {
```
bkchr, line 225: Spaces.

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,206 @@ impl OffchainStorage for InMemOffchainStorage {
...
+
+
+/// Iterate by value over the prepared offchain worker storage changes.
+pub struct OffchainOverlayedChangesIntoIter {
+	inner: Option<hash_map::IntoIter<Vec<u8>,OffchainOverlayedChange>>,
+}
+
+impl Iterator for OffchainOverlayedChangesIntoIter {
+	type Item = (Vec<u8>, OffchainOverlayedChange);
+    fn next(&mut self) -> Option<Self::Item> {
```
bkchr, line 197: Spaces.

primitives/runtime-interface/src/lib.rs
drahnr: Should have written calling direction or birds eye view on the architecture.

bkchr: If you want to keep this, please give the sentence more love. First, I would not talk about "generated code" here. 

Maybe you could update the sentences above where you think information is missing?

drahnr: Removed.

primitives/state-machine/src/basic.rs
drahnr: None in particular, it was following `Externalities::set_offchain_storage` which also uses `&[u8]`

client/cli/src/commands/runcmd.rs
```diff
@@ -359,13 +343,7 @@ impl RunCmd {
...
 		config.sentry_mode = self.sentry;
 
-		config.offchain_worker = match (&self.offchain_worker, role) {
-			(OffchainWorkerEnabled::WhenValidating, sc_service::Roles::AUTHORITY) => true,
-			(OffchainWorkerEnabled::Always, _) => true,
-			(OffchainWorkerEnabled::Never, _) => false,
-			(OffchainWorkerEnabled::WhenValidating, _) => false,
-		};
-
+		self.offchain_worker.update_config(&mut config, role)?;
```
cecton, line 346: You did it right but I got a new update on the cli that has just been push sorry :sweat_smile:  .... and it changes things a lot.

1. You will need to check the default implementation in client/cli/src/config.rs and adapt to your needs
2. In runcmd.rs you will need to change the runcmd implementation for that field like you did here (similarly)
3. You can keep the implementation in OffchainWorkerParams **BUT** you will need to rename update_config to offchain_worker and return the value (you don't receive config in parameter anymore)

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,94 @@
...
+arg_enum! {
+	/// Whether offchain workers are enabled or not as a whole.
+	#[allow(missing_docs)]
+	#[derive(Debug, Clone)]
+	pub enum OffchainWorkerEnabled {
+		Always,
+		Never,
+		WhenValidating,
+	}
+}
```
cecton, line 44: arg_enums should be placed in arg_enum.rs if possible (because of the missing_docs thingy)

client/cli/src/params/offchain_worker_params.rs
drahnr: I don't see a particular reason enabled for validators should be disabled.

client/cli/src/commands/runcmd.rs
drahnr: Thanks for the heads up!

client/cli/src/commands/runcmd.rs
```diff
@@ -428,13 +414,12 @@ impl CliConfiguration for RunCmd {
...
-	fn offchain_worker(&self, role: &Role) -> Result<bool> {
-		Ok(match (&self.offchain_worker, role) {
-			(OffchainWorkerEnabled::WhenValidating, Role::Authority { .. }) => true,
-			(OffchainWorkerEnabled::Always, _) => true,
-			(OffchainWorkerEnabled::Never, _) => false,
-			(OffchainWorkerEnabled::WhenValidating, _) => false,
-		})
+	fn offchain_worker_config(&self) -> OffchainWorkerConfig {
+		let is_dev = self.is_dev();
+		let role = self.role(is_dev);
```
cecton, line 419: is_dev and role must be passed by arguments. You can change the signature of offchain_worker_config to provide them (in config.rs) and change create_configuration (in config.rs) so they are given in arguments.

client/cli/src/config.rs
```diff
@@ -278,13 +283,12 @@ pub trait CliConfiguration: Sized {
 		Ok(Default::default())
 	}
 
-	/// Returns `Ok(true)` if offchain worker should be used
+	/// Returns a offchain worker config wrapped in `Ok(_)`
 	///
-	/// By default this is `false`.
-	fn offchain_worker(&self, _role: &Role) -> Result<bool> {
+	/// By default offchain workers are disabled.
+	fn offchain_worker_config(&self) -> Result<OffchainWorkerConfig> {
```
cecton, line 289: Since offchain_worker_params is available in this trait, you need to change the default implementation of offchain_worker_config to make it try to use that instead.

In other words: you can move the function offchain_worker_config from runcmd.rs to config.rs. 

This will allow another command to use OffchainWorkerParams just by defining offchain_worker_params. Does that make sense?

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,80 @@
...
+			(OffchainWorkerEnabled::Always, _) => true,
+			(OffchainWorkerEnabled::Never, _) => false,
+			(OffchainWorkerEnabled::WhenValidating, _) => false,
+		};
+
+        let indexing_enabled = enabled && self.indexing_enabled;
+
+        Ok(OffchainWorkerConfig { enabled, indexing_enabled})
+	}
+}
```
cecton, line 80: This function should be renamed offchain_worker_config to match the one in the CliConfiguration trait.

You can also remove the Result from the signature here (but **not** in the trait CliConfiguration).

client/cli/src/commands/runcmd.rs
cecton: Example here with pruning: https://github.com/paritytech/substrate/blob/master/client/cli/src/config.rs#L400

You can totally inspire on pruning because it is the exact same use case than yours:
- it has a *Params structopt
- it defines a _params method in CliConfiguration to get this params
- the method pruning uses that _params method to provide the default

I hope this will help.

client/cli/src/config.rs
drahnr: Thanks!

client/cli/src/commands/runcmd.rs
```diff
@@ -292,6 +283,10 @@ impl CliConfiguration for RunCmd {
 		Some(&self.keystore_params)
 	}
 
+	fn offchain_worker_params(&self) -> Option<&OffchainWorkerParams> {
+		Some(&self.offchain_worker_params)
+	}
```
cecton, line 288: Don't use Option here if you don't need. The method name needs to be identical but not the signature.

cc @bkchr

cecton: ah wait, that's the trait. Nevermind!! :sweat_smile:

drahnr: whoops, then I'll undo that last commit :)

gavofyork:
@drahnr needs resolving... @rphmeier @tomusdrw you two ok with merging this?

client/db/src/offchain_indexing.rs
```diff
@@ -0,0 +1,119 @@
+//! Offchain Indexing specific validation.
```
tomusdrw, line 1: License Preamble missing in that file.

client/src/config.rs
```diff
@@ -0,0 +1,51 @@
+
```
tomusdrw, line 1: License preamble missing.

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,208 @@ impl OffchainStorage for InMemOffchainStorage {
...
+		if let Self::Enabled(ref storage) = self {
+			let key: Vec<u8> = prefix.iter().chain(key).cloned().collect();
+			storage.get(&key).cloned()
+		} else {
+			None
+		}
+	}
+}
+
+use std::collections::hash_map;
```
tomusdrw, line 161: This should go to the top of the file.

primitives/state-machine/src/ext.rs
```diff
@@ -121,6 +125,11 @@ where
 	fn mark_dirty(&mut self) {
 		self.storage_transaction_cache.reset();
 	}
+
+	/// Read only accessor for the scheduled overlay changes.
+	pub fn get_offchain_storage_changes(&self) -> &OffchainOverlayedChanges {
```
tomusdrw, line 130:
```suggestion
	pub fn offchain_storage_changes(&self) -> &OffchainOverlayedChanges {
```

drahnr:
Inspecting the failing tests already, so hopefully resolved at some point today.

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,77 @@
...
+		let enabled = match (&self.enabled, role) {
+			(OffchainWorkerEnabled::WhenValidating, Role::Authority { .. }) => true,
+			(OffchainWorkerEnabled::Always, _) => true,
+			(OffchainWorkerEnabled::Never, _) => false,
+			(OffchainWorkerEnabled::WhenValidating, _) => false,
+		};
+
+		let indexing_enabled = enabled && self.indexing_enabled;
+
+		Ok(OffchainWorkerConfig { enabled, indexing_enabled})
```
kianenigma, line 75:
```suggestion
		Ok(OffchainWorkerConfig { enabled, indexing_enabled })
```

client/api/src/execution_extensions.rs
```diff
@@ -202,4 +202,4 @@ impl<Block: traits::Block> offchain::TransactionPool for TransactionPoolAdapter<
 
 		self.pool.submit_at(&self.at, xt)
 	}
-}
+}
```
kianenigma, line 205:
```suggestion
}

```

client/executor/src/integration_tests/mod.rs
```diff
@@ -450,6 +450,28 @@ fn ordered_trie_root_should_work(wasm_method: WasmExecutionMethod) {
...
+fn offchain_index(wasm_method: WasmExecutionMethod) {
+	let mut ext = TestExternalities::default();
+	let (offchain, _state) = testing::TestOffchainExt::new();
+	ext.register_extension(OffchainExt::new(offchain));
+	call_in_wasm(
+			"test_offchain_index_set",
+			&[0],
+			wasm_method,
+			&mut ext.ext(),
+		).unwrap();
```
kianenigma, line 464:
```suggestion
	call_in_wasm(
		"test_offchain_index_set",
		&[0],
		wasm_method,
		&mut ext.ext(),
	).unwrap();
```

client/executor/runtime-test/src/lib.rs
```diff
@@ -183,6 +183,12 @@ sp_core::wasm_export_functions! {
 		}
 	}
 
+
+	fn test_offchain_index_set() {
```
kianenigma, line 187: what is the equivalent of this in a real runtime?

drahnr: `sp_io::offchain_index::set(b"k", b"v")`

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,77 @@
...
+	#[structopt(
+		long = "offchain-worker",
+		value_name = "ENABLED",
+		possible_values = &OffchainWorkerEnabled::variants(),
+		case_insensitive = true,
+		default_value = "WhenValidating"
+	)]
+	pub enabled: OffchainWorkerEnabled,
+
+	/// Allow access to offchain workers indexing API
```
kianenigma, line 48: where does the term indexing here comes from? It doesn't really connect for me to the description in the next line. Indexing implies some sort of caching similar to databases. Yet the description just says: `Enables a runtime to write directly to ..`. 

Maybe something like: `ENABLE_OFFCHAIN_IMPORT_DB`

client/service/src/config.rs
```diff
@@ -125,6 +125,14 @@ impl KeystoreConfig {
 		}
 	}
 }
+/// Configuration of the database of the client.
+#[derive(Clone, Default)]
+pub struct OffchainWorkerConfig {
+	/// If this is allowed.
+	pub enabled: bool,
+	/// allow writes from the runtime to the offchain worker database
```
kianenigma, line 133:
```suggestion
	/// allow writes from the runtime to the offchain worker database.
```

primitives/core/src/offchain/mod.rs
```diff
@@ -28,6 +28,9 @@ pub mod storage;
 #[cfg(feature = "std")]
 pub mod testing;
 
+/// Local storage prefix used by the Offchain Worker API to
+pub const STORAGE_PREFIX : &'static [u8] = b"storage";
```
kianenigma, line 32: @tomusdrw ain't this too generic? although I think it is too late to change this.

primitives/core/src/offchain/storage.rs
```diff
@@ -18,13 +18,32 @@
...
+
+	/// Iterate over all key value pairs by reference.
+	pub fn iter<'a>(&'a self) -> impl Iterator<Item=(&'a Vec<u8>,&'a Vec<u8>)> {
+		self.storage.iter()
+	}
+
+	/// Remove a key and it's associated value from the offchain database.
+	pub fn remove(&mut self, prefix: &[u8], key: &[u8]) {
+		let key: Vec<u8> = prefix.iter().chain(key).cloned().collect();
+		let _ = self.storage.remove(&key);
```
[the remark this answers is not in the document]
kianenigma: (I think generally suppressing either `Option` or `Result` in the production code is bad and even if there is no way to handle it, it is better to let the warning be there and gain attention in the future. `let _  = ...` basically means goodbye and this will never be seen again :D)

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+		OffchainOverlayedChangesIntoIter::new(self)
+	}
+
+	/// Iterate over all key value pairs by reference.
+	pub fn iter<'a>(&'a self) -> OffchainOverlayedChangesIter {
+		OffchainOverlayedChangesIter::new(&self)
+	}
+
+	/// Drain all elements of changeset.
+	pub fn drain<'a,'d>(&'a mut self) -> OffchainOverlayedChangesDrain<'d> where 'a: 'd {
```
kianenigma, line 130:
```suggestion
	pub fn drain<'a, 'd>(&'a mut self) -> OffchainOverlayedChangesDrain<'d> where 'a: 'd {
```

client/cli/src/params/offchain_worker_params.rs
drahnr: The data that is going to be written is in some form a cache for i.e. slashing based on this historical data. I am not fixed on the terminology here.

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+impl OffchainOverlayedChanges {
+	/// Create the disabled variant.
+	pub fn disabled() -> Self {
+		Self::Disabled
+	}
+
+	/// Create the enabled variant.
+	pub fn enabled() -> Self {
+		Self::Enabled(HashMap::new())
+	}
```
kianenigma, line 117: This is not an issue but rather an opinionated comment: What is the rational for having constructors for enum? for a struct with private field I can see the reason, but here I think the raw enum syntax is just fine.

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+			let key: Vec<u8> = prefix.iter().chain(key).cloned().collect();
+			let _ = storage.insert(key, OffchainOverlayedChange::Remove);
+		}
+	}
+
+	/// Set the value associated with a key under a prefix to the value provided.
+	pub fn set(&mut self, prefix: &[u8], key: &[u8], value: &[u8]) {
+		if let Self::Enabled(ref mut storage) = self {
+			let key = prefix.iter().chain(key).cloned().collect();
+			let _ = storage.insert(key, OffchainOverlayedChange::SetValue(value.to_vec()));
```
kianenigma, line 146: same as the other `let _ = ` comment.

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+			None
+		}
+	}
+}
+
+use std::collections::hash_map;
+
+/// Iterate by reference over the prepared offchain worker storage changes.
+pub struct OffchainOverlayedChangesIter<'i> {
+	inner: Option<hash_map::Iter<'i,Vec<u8>,OffchainOverlayedChange>>,
```
kianenigma, line 165:
```suggestion
	inner: Option<hash_map::Iter<'i, Vec<u8>, OffchainOverlayedChange>>,
```

or is there some convention here that I am missing?

drahnr: no, I am just still adjusting to formatting things manually

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+				inner: Some(inner.drain())
+			},
+			OffchainOverlayedChanges::Disabled => Self { inner: None, },
+		}
+	}
+}
+
+
+#[cfg(test)]
+mod test {
```
kianenigma, line 251: The iter api of this overlat seems very clean 👍

primitives/io/src/lib.rs
```diff
@@ -984,6 +1000,7 @@ pub type SubstrateHostFunctions = (
 	logging::HostFunctions,
 	sandbox::HostFunctions,
 	crate::trie::HostFunctions,
+	offchain_index::HostFunctions,
```
kianenigma, line 1003: I don't get what is happening here, but I can't imagine `offchain_index::HostFunctions,` twice in the list to be correct.

primitives/core/src/offchain/storage.rs
drahnr: Symmetry, if I have a constructor for one variant, I'd expect every sane user constructable variant to have one instead of mixing between enum Varaints, but this might be just my personal preference.

primitives/io/src/lib.rs
drahnr: huh, where do you see it twice?

primitives/core/src/offchain/storage.rs
kianenigma: >  if I have a constructor for one variant

Then the question is: why to have the constructor for the other variant in the first place? :D 

although, I can see here that `enabled` is basically `enabled_empty()`, which is a nicer syntax than explicitly building the empty HashMap.

primitives/core/src/offchain/storage.rs
drahnr: Even if it does not implement try, I tend to add these `let _ = ..` to be explicit that something is intentionally unused. Can be removed if desired, I have no strong opinion about this.

primitives/core/src/offchain/storage.rs
drahnr: Because the internal representation is not relevant to the user, so I don't want to clutter outer code which has no interest in knowing about it, to construct inner structure for the overlay if that makes sense :)

primitives/io/src/lib.rs
kianenigma: I am losing my eyesight, ignore :D

kianenigma: the only comment would be why not merge this with `offchain::HostFunctions`? won't they always be used and available together?

drahnr: Not necessarily, offchain workers can exist without the indexing API - in reality I don't know how correlated the usage will be.

client/cli/src/params/offchain_worker_params.rs
drahnr: @tomusdrw I don't have opinion about this, I would prefer to keep it mostly as is to drive this to completion.

primitives/core/src/offchain/mod.rs
tomusdrw: It is stored in a separate rocksdb column anyway. This is just an extra prefix to distinguish with fork-aware storage and potentially other things in the future.

client/cli/src/params/offchain_worker_params.rs
tomusdrw: So this API allows the block import to insert something to offchain DB, so that it's available later.
1. We want to emphasize, this is a write-only API, so I don't think that it should involve names like DB access or anything like that.
2. It allows you to produce some data that forms something like an index of imported blocks. IMHO indexing emphasizes that the data is going to be complete (i.e. we can produce an index of all transactions in blocks accesible externaly via RPC, etc).
3. I'm a bit reluctant to use "offchain" too much, to avoid the confusion with offchain workers. As @drahnr mentioned, these 2 concepts are pretty much disjoint (indexed data is already available via offchain-db RPCs, so you can access it externally). There are use cases however that will use both APIs - in such case whatever is put into the offchain index is consumed by offchain workers.

I'm not super happy with the naming, open to any suggestions. If we keep the current one, I'd like to request more attention to details and keep everything consistent, i.e.:
we should always use a full name Offchain Indexing API (or just Indexing API), but never mix it up with "workers", cause it has nothing to do.
The concepts we have:
1. Off-chain DB - database that is not part of the consensus (i.e not trie)
2. Off-chain Workers - an opt-in piece of code being run after block import, which has access to extended set of APIs (Off-chain (Worker) APIs), that includes access to Off-chain DB.
3. Off-chain indexing API - an opt-in API that allows Off-chain DB to be populated (write) by block import.

client/api/src/backend.rs
```diff
@@ -148,6 +148,14 @@ pub trait BlockImportOperation<Block: BlockT> {
 		child_update: ChildStorageCollection,
 	) -> sp_blockchain::Result<()>;
 
+	/// Set offchain storage changes.
```
tomusdrw, line 151:
```suggestion
	/// Write offchain storage changes to the database.
```

client/cli/src/config.rs
```diff
@@ -67,6 +67,11 @@ pub trait CliConfiguration: Sized {
 		None
 	}
 
+	/// Get the OffchainWorkerParams for this object
```
tomusdrw, line 70:
```suggestion
	/// Get a reference to `OffchainWorkerParams` for this object.
```

client/cli/src/config.rs
```diff
@@ -301,11 +306,13 @@ pub trait CliConfiguration: Sized {
 		Ok(Default::default())
 	}
 
-	/// Returns `Ok(true)` if offchain worker should be used
+	/// Returns a offchain worker config wrapped in `Ok(_)`
```
tomusdrw, line 309:
```suggestion
	/// Returns an offchain worker config wrapped in `Ok(_)`
```

client/cli/src/params/offchain_worker_params.rs
tomusdrw, line 48:
```suggestion
	/// Enable Offchain Indexing API, which allows block import to write to Offchain DB.
```

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,77 @@
...
+	///
+	/// Enables a runtime to write directly to a offchain workers
```
tomusdrw, line 50:
```suggestion
	/// Enables the runtime to write directly to Offchain DB during block import.
```

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,77 @@
...
+	/// DB during block import.
```
tomusdrw, line 51:
```suggestion
```

client/cli/src/params/offchain_worker_params.rs
```diff
@@ -0,0 +1,77 @@
...
+		let enabled = match (&self.enabled, role) {
+			(OffchainWorkerEnabled::WhenValidating, Role::Authority { .. }) => true,
+			(OffchainWorkerEnabled::Always, _) => true,
+			(OffchainWorkerEnabled::Never, _) => false,
+			(OffchainWorkerEnabled::WhenValidating, _) => false,
+		};
+
+		let indexing_enabled = enabled && self.indexing_enabled;
+
+		Ok(OffchainWorkerConfig { enabled, indexing_enabled })
```
tomusdrw, line 75: Not a fan of bundling the two settings together in one "worker" config. I think we should clearly separate this two things. But I can live with this, and would prefer to merge the PR earlier.

client/service/src/config.rs
```diff
@@ -125,6 +125,14 @@ impl KeystoreConfig {
 		}
 	}
 }
+/// Configuration of the database of the client.
+#[derive(Clone, Default)]
+pub struct OffchainWorkerConfig {
+	/// If this is allowed.
+	pub enabled: bool,
+	/// allow writes from the runtime to the offchain worker database.
```
tomusdrw, line 133:
```suggestion
	/// Allow writes from the runtime to the Offchain DB.
```

client/service/src/config.rs
```diff
@@ -125,6 +125,14 @@ impl KeystoreConfig {
 		}
 	}
 }
+/// Configuration of the database of the client.
+#[derive(Clone, Default)]
+pub struct OffchainWorkerConfig {
+	/// If this is allowed.
```
tomusdrw, line 131:
```suggestion
	/// Should we run Offchain Workers after every block?
```

client/service/src/config.rs
```diff
@@ -125,6 +125,14 @@ impl KeystoreConfig {
 		}
 	}
 }
+/// Configuration of the database of the client.
```
tomusdrw, line 128:
```suggestion
/// Configuration of Offchain APIs (workers and indexing).
```

client/src/client.rs
```diff
@@ -146,7 +148,24 @@ pub fn new_in_mem<E, Block, S, RA>(
...
+		executor,
+		genesis_storage,
+		keystore,
+		spawn_handle,
+		prometheus_registry,
+		config,
+	)
+}
+
+/// Relevant client configuration items relevant for the client.
```
tomusdrw, line 162:
```suggestion
/// Additional client configuration.
```

client/src/client.rs
```diff
@@ -146,7 +148,24 @@ pub fn new_in_mem<E, Block, S, RA>(
...
+#[derive(Debug,Clone,Default)]
+pub struct ClientConfig {
+	/// Enable the offchain worker db.
```
tomusdrw, line 165:
```suggestion
	/// Enable running Offchain Workers.
```

client/src/client.rs
```diff
@@ -146,7 +148,24 @@ pub fn new_in_mem<E, Block, S, RA>(
...
+	pub offchain_worker_enabled: bool,
+	/// If true, allows access from the runtime to write into offchain worker db.
```
tomusdrw, line 167:
```suggestion
	/// Enable Indexing API writes to Offchain DB.
```

client/src/client.rs
```diff
@@ -243,6 +264,7 @@ impl<B, E, Block, RA> Client<B, E, Block, RA> where
 	B: backend::Backend<Block>,
 	E: CallExecutor<Block>,
 	Block: BlockT,
+	Block::Header: Clone,
```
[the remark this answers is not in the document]
tomusdrw: ping

client/src/client.rs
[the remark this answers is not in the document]
tomusdrw: ping

primitives/core/src/offchain/mod.rs
```diff
@@ -28,6 +28,9 @@ pub mod storage;
 #[cfg(feature = "std")]
 pub mod testing;
 
+/// Local storage prefix used by the Offchain Worker API to
```
tomusdrw, line 31:
```suggestion
/// Database prefix use by persistent Offchain DB.
/// 
/// The data is stored in a separate RocksDB column anyway, the prefix is added to be able to distinguish
/// other kinds of entries in the same column in the future.
```

primitives/core/src/offchain/storage.rs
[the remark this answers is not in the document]
tomusdrw: ping

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
 		}
 	}
 }
+
+
+
+
+/// Change to be applied to the offchain worker db in regards to a key.
```
tomusdrw, line 84:
```suggestion
/// Change to be applied to the Offchain DB for a particular key.
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+#[derive(Debug,Clone,Hash,Eq,PartialEq)]
+pub enum OffchainOverlayedChange {
+	/// Remove the data associated with the key
```
tomusdrw, line 87:
```suggestion
	/// Remove the data associated with the key.
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+	Remove,
+	/// Overwrite the value of an associated key
```
tomusdrw, line 89:
```suggestion
	/// Overwrite the value of an associated key.
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+	SetValue(Vec<u8>),
+}
+
+/// In-memory storage for offchain workers recoding changes for the actual offchain storage implementation.
```
tomusdrw, line 93:
```suggestion
/// In-memory storage for Offchain DB buffering changes to be applied later to the underlying DB.
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+#[derive(Debug, Clone)]
+pub enum OffchainOverlayedChanges {
+	/// Writing overlay changes to the offchain worker database is disabled by configuration.
```
tomusdrw, line 96:
```suggestion
	/// Discard all changes right away when writing is disabled by configuration.
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+	pub fn iter<'a>(&'a self) -> OffchainOverlayedChangesIter {
+		OffchainOverlayedChangesIter::new(&self)
+	}
+
+	/// Drain all elements of changeset.
+	pub fn drain<'a, 'd>(&'a mut self) -> OffchainOverlayedChangesDrain<'d> where 'a: 'd {
+		OffchainOverlayedChangesDrain::new(self)
+	}
+
+	/// Remove a key and its associated value from the offchain database.
```
tomusdrw, line 134:
```suggestion
	/// Remove a key and its associated value from the Offchain DB.
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+	pub fn remove(&mut self, prefix: &[u8], key: &[u8]) {
+		if let Self::Enabled(ref mut storage) = self {
+			let key: Vec<u8> = prefix.iter().chain(key).cloned().collect();
+			let _ = storage.insert(key, OffchainOverlayedChange::Remove);
+		}
+	}
+
+	/// Set the value associated with a key under a prefix to the value provided.
```
tomusdrw, line 142:
```suggestion
	/// Insert the value under given key and prefix overwriting any previous data.
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+		if let Some(ref mut iter) = self.inner {
+			iter.next()
+		} else {
+			None
+		}
+	}
+}
+
+impl<'i> OffchainOverlayedChangesIter<'i> {
+	/// Create a new iterator based on a refernce to the parent container.
```
tomusdrw, line 180:
```suggestion
	/// Create a new iterator based on a reference to the parent container.
```

primitives/core/src/offchain/storage.rs
```diff
@@ -58,3 +77,212 @@ impl OffchainStorage for InMemOffchainStorage {
...
+			OffchainOverlayedChanges::Enabled(inner) => Self {
+				inner: Some(inner.iter())
+			},
+			OffchainOverlayedChanges::Disabled => Self { inner: None, },
+		}
+	}
+}
+
+
+/// Iterate by value over the prepared offchain worker storage changes.
```
tomusdrw, line 192:
```suggestion
/// Iterate by value over the prepared Offchain DB changes.
```

primitives/externalities/src/lib.rs
```diff
@@ -47,6 +47,9 @@ pub enum Error {
 ///
 /// Provides access to the storage and to other registered extensions.
 pub trait Externalities: ExtensionStore {
+	/// Write a key value pair to the offchain storage database.
```
tomusdrw, line 50:
```suggestion
	/// Write a key value pair to the Offchain DB.
```

primitives/io/src/lib.rs
```diff
@@ -602,13 +602,29 @@ pub trait Hashing {
 	}
 }
 
+/// Interface that provides functions to access the offchain database access.
```
tomusdrw, line 605:
```suggestion
/// Interface that provides functions to access the Offchain DB.
```

primitives/io/src/lib.rs
```diff
@@ -602,13 +602,29 @@ pub trait Hashing {
...
+#[runtime_interface]
+pub trait OffchainIndex {
+	/// Write a key value pair to the offchain worker database in a buffered fashion.
+	fn set(&mut self, key: &[u8], value: &[u8]) {
+		self.set_offchain_storage(key, Some(value));
+	}
+
+	/// Remove a key and its associated value from the offchain worker database.
```
tomusdrw, line 613:
```suggestion
	/// Remove a key and its associated value from the Offchain DB.
```

primitives/io/src/lib.rs
```diff
@@ -602,13 +602,29 @@ pub trait Hashing {
 	}
 }
 
+/// Interface that provides functions to access the offchain database access.
+#[runtime_interface]
+pub trait OffchainIndex {
+	/// Write a key value pair to the offchain worker database in a buffered fashion.
```
tomusdrw, line 608:
```suggestion
	/// Write a key value pair to the Offchain DB database in a buffered fashion.
```

primitives/state-machine/src/basic.rs
tomusdrw: I think in this case it's better to work with borrowed data, cause we can avoid a clone if indexing is disabled. That said we should probably unify the API (and introduce `Storage*Ref` to work with slices). I'd leave it as a separate PR though.

primitives/state-machine/src/ext.rs
```diff
@@ -74,6 +75,8 @@ pub struct Ext<'a, H, N, B>
 {
 	/// The overlayed changes to write to.
 	overlay: &'a mut OverlayedChanges,
+	/// The overlayed changes destined for the offchain worker database.
```
tomusdrw, line 78:
```suggestion
	/// The overlayed changes destined for the Offchain DB.
```

rphmeier:
@drahnr Please open a "Polkadot Companion" PR and then we can merge this and that shortly thereafter.

rphmeier:
Companion and this look good, I think we can merge once CI completes here. cc @TriplEight companion PR task may not be working correctly; I've tried restarting it but we'll see if it works.

TriplEight:
@rphmeier I've checked and it seems to be caused by GitHub availability issues.

cheme:
Quick question, should the overlay maintain a commit and prospective state?
(currently rejected extrinsic will index, but that may not be an issue (if reorg is not this is maybe not either).

rphmeier:
Good point - although at the moment our use-cases are in `on_initialize`/`on_finalize` only, we should be able to handle rollback better.

Issue #4917, opened 2020-02-13
Migrate storage maps to use twox64_concat or blake2_128_concat


shawntabrizi:
The various map pallets mostly use `blake2_256` hashers and sometimes `linked_map`s and empty double maps. All linked maps should be removed. Double maps where the initial item is expected to be constant (e.g. `pallet-session`'s `NextKeys`) should be altered to a simple map now. Hashers should be switched to either:
- `blake2_128_concat` in the case that the user is able to place map items whose keys are entirely at their discretion.
- `twox64_concat` in the case that the user cannot.

An example of the former would be the `frame_system`'s `Account` storage item, since any address (even one without a known corresponding private key!) may be sent an existential deposit and therefore be created.

An example of the latter would be `frame_system`'s `BlockHash` storage item, since its keys are limited to the block number integers. Users cannot populate it directly, nor can they influence the data.

A second example of the latter would be `pallet-staking`'s `Bonded` item. Though this has a corresponding key type/value to `frame_system`'s `Account` storage item, it has different semantics; no entry can be placed except through a valid signed transaction coming from the account of the key. As such its value is heavily constrained and cannot be selected at will (in general a public key would need to be known for the account and the distribution/correlation between public and secret keys are not such that a trie-imbalancing global griefing attack would be practical to execute).

### Tasks
- [x] Remove all the superfluous double-maps from Session & migrate.
- [x] Replace all `hasher(blake2_256)`s with `_concat` hashers & provide either static or dynamic migration.

kianenigma:
related: https://github.com/paritytech/substrate/issues/4610

PR #5202, opened 2020-03-10
Refactor session away from needless double_maps

gavofyork:
Part of https://github.com/paritytech/substrate/issues/4917


gavofyork:
Safely tested with the Kusama state.

frame/balances/src/migration.rs
```diff
@@ -1,3 +1,21 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
```
kianenigma:
```suggestion
// Copyright 2020 Parity Technologies (UK) Ltd.
```

PR #5203, opened 2020-03-10
CI: Check runtime version against last release, not master

s3krit:
we only need to bump the runtime version if there have been changes to the runtime since the last release (tracked by the `release` tag), not every time the runtime is changed.


gavofyork:
Good. @gnunicorn @s3krit worth being aware that we really *must* ensure `release` tag follows the latest release. If not, then our check runtime CI step will cease to do its job. Maybe we can CI this somehow?

PR #5204, opened 2020-03-10
Add dev chain as default chain for benchmarks.

### LINKED ISSUE #5154: Benchmarking CLI should use `--chain dev` if not provided.
Right now, if a user forgets to provide `--chain`, they get a nasty misleading error:

```
Error: Input("Error executing runtime benchmark: Wasmi(Function(\"Module doesn\\\'t have export Benchmark_dispatch_benchmark\"))")
```

Instead, we should choose <strike>a default chain just like if we were running the node</strike> the dev chain spec which contains the benchmarking apis.

marcio-diaz:
closes #5154. 

I went for removing the `update_config()` call, other options could be to add a `default_chain: Option<>` argument to it (maybe better), or to keep the call as it is and set the default before (imo worst).

### DIFF
--- a/utils/frame/benchmarking-cli/src/lib.rs
+++ b/utils/frame/benchmarking-cli/src/lib.rs
@@ -14,13 +14,13 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
+use std::{fmt::Debug, path::PathBuf};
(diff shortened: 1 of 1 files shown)


marcio-diaz:
> It's not ideal but I think it's okay and I don't see how we can do better.
> 
> Q: You really don't need the rest of the code in shared_params?
> 
> https://github.com/paritytech/substrate/blob/b70f191e944e6cf6235b18022c74f4ce1b62d50c/client/cli/src/params/shared_params.rs#L72-L90

Ok, thanks. Yeah, looks like we don't need it.

PR #5205, opened 2020-03-10
Revert #5025

shawntabrizi:
This reverts #5025 which removed the Indices pallet from the Substrate Node Template.

This change, while reasonable in the context of Substrate, had impact on the Polkadot JS API, which made it harder for new developers to build their apps against the front-end libraries.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3598,6 +3598,7 @@ dependencies = [
  "pallet-aura",
  "pallet-balances",
  "pallet-grandpa",
+ "pallet-indices",
  "pallet-randomness-collective-flip",
  "pallet-sudo",
  "pallet-template",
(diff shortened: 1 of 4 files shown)


jacogr:
I don't believe the template should work around the API. Since this same thing is applied in Kusama and will be in Polkadot, it is actually a sane change. So don't have a thumbs-up or thumbs-down to give here.

shawntabrizi:
@jacogr I think the context for an approval should be:

* If the polkadot-js-api will not support the node-template out of the box, that will be a large net negative for our potential developers. If adding indices will solve that, I think it is an overall win.

* We can always look to address a cleanup like this again in the future when there is less shit to fix.

If you agree, I think a thumbs up is the right response.

JoshOrndorff:
I prefer not to merge this revert. @shawntabrizi your reasons for removing indices in #5025 are good - makes the template simpler as a template should be. I'm willing to maintain Apps support for node template, and have made a PR to do so https://github.com/polkadot-js/apps/pull/2364

shawntabrizi:
Sounds good @JoshOrndorff 

Good on you here :)

PR #5207, opened 2020-03-10
Split fees and tips between author and treasury independently

gavofyork:
Alternative for #4364.
Closes https://github.com/paritytech/substrate/issues/4363

This avoids making too many changes by allowing `OnUnbalanced` to handle iterators of `Imbalance`s. A default and backwards compatible default implementation is provided. It is used by the transaction payment system in order to deal with fees and tips separately.

It should give strictly the same amount of storage accesses and runtime/native crossings as before, adding only a small amount of extra arithmetic, much of which should be optimisable away.

### DIFF
--- a/frame/support/src/traits.rs
+++ b/frame/support/src/traits.rs
@@ -339,8 +339,8 @@ pub trait Imbalance<Balance>: Sized + TryDrop {
 	/// Consume `self` and return two independent instances; the amounts returned will be in
 	/// approximately the same ratio as `first`:`second`.
 	///
-	/// NOTE: This requires up to `first + second` room for a multiply. Overflow will safely
-	/// saturate.
+	/// NOTE: This requires up to `first + second` room for a multiply, and `first + second` should
+	/// fit into a `u32`. Overflow will safely saturate in both cases.
 	fn ration(self, first: u32, second: u32) -> (Self, Self)
 		where Balance: From<u32> + Saturating + Div<Output=Balance>
 	{
--- a/frame/support/src/traits.rs
+++ b/frame/support/src/traits.rs
@@ -256,6 +256,13 @@ pub trait KeyOwnerProofSystem<Key> {
 /// - Someone got slashed.
 /// - Someone paid for a transaction to be included.
 pub trait OnUnbalanced<Imbalance: TryDrop> {
+	/// Handler for some imbalances. The different imbalances might have different origins or
+	/// meanings, dependent on the context. Will default to simply calling on_unbalanced for all
+	/// of them. Infallible.
+	fn on_unbalanceds<B>(amounts: impl Iterator<Item=Imbalance>) where Imbalance: crate::traits::Imbalance<B> {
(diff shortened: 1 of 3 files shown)


frame/transaction-payment/src/lib.rs
```diff
@@ -234,7 +259,8 @@ impl<T: Trait + Send + Sync> SignedExtension for ChargeTransactionPayment<T>
 				Ok(imbalance) => imbalance,
 				Err(_) => return InvalidTransaction::Payment.into(),
 			};
-			T::OnTransactionPayment::on_unbalanced(imbalance);
+			let (tip_imbalance, fee_imbalance) = imbalance.split(tip);
+			T::OnTransactionPayment::on_unbalanceds(PairIterator::with_two(fee_imbalance, tip_imbalance));
```
thiolliere: I think we should add to the doc of `type OnTransactionPayment` that unbalance will be iterator of fee and then tip

thiolliere: as we only use PairIterator::with_two note that we can just use: 

```suggestion
			T::OnTransactionPayment::on_unbalanceds(Some(fee_imbalance).into_iter().chain(Some(tip_imbalance)))
```

frame/support/src/traits.rs
```diff
@@ -331,10 +336,71 @@ pub trait Imbalance<Balance>: Sized + TryDrop {
 	/// is guaranteed to be at most `amount` and the second will be the remainder.
 	fn split(self, amount: Balance) -> (Self, Self);
 
+	/// Consume `self` and return two independent instances; the amounts returned will be in
+	/// approximately the same ratio as `first`:`second`.
+	///
+	/// NOTE: This requires up to `first + second` room for a multiply. Overflow will safely
+	/// saturate.
+	fn ration(self, first: u32, second: u32) -> (Self, Self)
```
thiolliere: if wanted we could implement it without saturation similarly to how perbill implements mul.

frame/support/src/traits.rs
```diff
@@ -331,10 +336,71 @@ pub trait Imbalance<Balance>: Sized + TryDrop {
 	/// is guaranteed to be at most `amount` and the second will be the remainder.
 	fn split(self, amount: Balance) -> (Self, Self);
 
+	/// Consume `self` and return two independent instances; the amounts returned will be in
+	/// approximately the same ratio as `first`:`second`.
+	///
+	/// NOTE: This requires up to `first + second` room for a multiply. Overflow will safely
+	/// saturate.
```
thiolliere: we also need first + second to fit into u32 not sure it worth mentionning

thiolliere:
except some comments everything looks good to me

frame/support/src/traits.rs
gavofyork: not immediately needed, any it would slow things down, and at least the polkadot and kusama balance types are designed specifically so that there is a u32's worth of space at the high end, so i'd prefer not to at this stage.

PR #5208, opened 2020-03-10
Staking improve payout_validator

thiolliere:
its adds some documentation + use clipped exposure for validator payout.

It shouldn't change the behavior of the runtime because for validator payout clipped exposure and exposure hold same information. It should just optimise the read of validator exposure.

### DIFF
--- a/frame/staking/src/lib.rs
+++ b/frame/staking/src/lib.rs
@@ -767,6 +767,7 @@ decl_storage! {
 		///
 		/// This is similar to [`ErasStakers`] but number of nominators exposed is reduce to the
 		/// `T::MaxNominatorRewardedPerValidator` biggest stakers.
+		/// (Note: the field `total` and `own` of the exposure remains unchanged).
 		/// This is used to limit the i/o cost for the nominator payout.
 		///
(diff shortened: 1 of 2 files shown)


frame/staking/src/lib.rs
```diff
@@ -1349,7 +1350,10 @@ decl_module! {
 		/// - `who` is the controller account of the nominator to pay out.
 		/// - `era` may not be lower than one following the most recently paid era. If it is higher,
 		///   then it indicates an instruction to skip the payout of all previous eras.
-		/// - `validators` is the list of all validators that `who` had exposure to during `era`.
+		/// - `validators` is the list of all validators that `who` had exposure to during `era`
+		///   alongside the index of the `who` in the clipped exposure of the validator.
+		///   i.e. each element is a tuple of
+		///   `(validator, index of `who` in clipped exposure of validator)`.
```
joepetrowski:
```suggestion
		/// - `validators` is the list of all validators that `who` had exposure to during `era`,
		///   alongside the index of `who` in the clipped exposure of the validator.
		///   I.e. each element is a tuple of
		///   `(validator, index of `who` in clipped exposure of validator)`.
```

PR #5216, opened 2020-03-10
Update sysinfo version to 0.11

### LINKED ISSUE #5163: Update sysinfo
We are still using `0.9` of the sysinfo crate. We should update this to latest.

DarkEld3r:
Closes #5163.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3719,6 +3719,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ntapi"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26e041cd983acbc087e30fcba770380cfa352d0e392e175b2344ebaf7ea0602"
+dependencies = [
+ "winapi 0.3.8",
+]
+
+[[package]]
 name = "num-bigint"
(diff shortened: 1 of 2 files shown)


gnunicorn:
@DarkEld3r web-wasm tests failing now, benches, too.

DarkEld3r:
@gnunicorn There is some strange issue with the benches:
```
error: failed to get `sysinfo` as a dependency of package `sc-service v0.8.0-alpha.3 (/builds/parity/substrate/client/service)`
 Caused by:
   failed to fetch `https://github.com/rust-lang/crates.io-index`
 Caused by:
   missing delta bases; class=Indexer (15)
```
I'm not sure that it is caused by my changes.

As for web-wasm tests: it is an issue in the `sysinfo` crate itself: https://github.com/GuillaumeGomez/sysinfo/pull/276

gnunicorn:
> As for web-wasm tests: it is an issue in the sysinfo crate itself: GuillaumeGomez/sysinfo#276

which means the current version isn't gonna work for us, but we need to wait for the next release including your patch. icing.

DarkEld3r:
@gnunicorn The new version has been released, so I have updated the pull request.

PR #5226, opened 2020-03-11
Refactor away from opaque hashes

### LINKED ISSUE #4917: Migrate storage maps to use twox64_concat or blake2_128_concat

gavofyork:
Closes #4917

One final addition, which will need to be kept for a future PR, is the ability to iterate over all key pairs of a double-map. This is currently not possible, since we don't have any way of determining which portion of the combined key belongs to K1 and which to K2. Since all practical applications of maps have a fixed size key, this is not insurmountable, however we would need to introduce into codec an additional trait:

```
pub trait FixedSizeEncoding: Encode { const SIZE: usize; }
```

which should then be implemented by all fixed size types that also implement `Encode`. This could then be required for the key arguments (well, the first, at least), in order to partition the key material (with the cooperation of the `ReversibleStorageHasher` trait, which knows the aggregate size) and decode the keys properly. @thiolliere maybe this is something you could work on?

TODO:
- [x] Docs to guard against using the wrong hasher
- [x] Rename opaque hashers accordingly
- [x] Introduce aliases `tainted`, `natural`, `prehashed` which can be used in `decl_storage` instead of `hasher(blake2_128_concat)`, `hasher(twox_64_concat)` and `hasher(identity)` respectively. (They still need `hasher` currently.)
- [x] Remove `linked_map`
- [x] Introduce K/V iteration for double maps

### With migration (account keys)
- [x] System.Account (blake2_128_concat)
- [x] Balances.Locks (twox_64_concat)
- [x] ElectionsPhragmen.VotesOf (twox_64_concat)
- [x] ElectionsPhragmen.StakeOf (twox_64_concat)
- [x] Identity.IdentityOf (twox_64_concat)
- [x] Identity.SubsOf (twox_64_concat)  
- [x] Society.Payouts (twox_64_concat)
- [x] Session.NextKeys (twox_64_concat)
- [x] Staking.Bonded (twox_64_concat)
- [x] Staking.Ledger (blake2_128_concat)
- [x] Staking.Payee (twox_64_concat)
- [x] Staking.Validators (twox_64_concat) 
- [x] Staking.Nominators (twox_64_concat)
- [x] Vesting.Vesting (blake2_128_concat)
- [x] Democracy.Proxy (twox_64_concat) (Hopefully, should check)
- [x] Democracy.Locks (twox_64_concat)
- [x] Democracy.Delegations (twox_64_concat)

### With migration (closed range)
- [x] System.BlockHash (twox_64_concat)
- [x] Babe.UnderConstruction (twox_64_concat)
- [x] Grandpa.SetIdSession (twox_64_concat)
- [x] Staking.ErasStartSessionIndex (twox_64_concat)
- [x] Staking.ErasValidatorReward (twox_64_concat)
- [x] Staking.ErasRewardPoints (twox_64_concat)
- [x] Staking.ErasTotalStake (twox_64_concat)
- [x] Staking.SlashingSpans (twox_64_concat)
- [x] Democracy.VotersFor (twox_64_concat)
- [x] Democracy.ReferendumInfoOf (twox_64_concat)
- [x] Democracy.DepositOf (twox_64_concat)
- [x] Democracy.Preimages (identity)

### With migration, others
- [x] Claims.Claims (twox_64_concat) (Ethereum addresses - will need manual insertion)
- [x] Claims.Vesting (twox_64_concat) (Ethereum addresses - will need manual insertion)
- [x] Identity.SuperOf (blake2_128_concat) (Needs to come from contents of every accounts' SubsOf)
- [x] Session.KeyOwner (blake2_128_concat) (Needs to come from contents of every account's NextKeys)
- [x] Democracy.VoteOf (twox_64_concat) (Pair map, combination of account ID and closed range)
- [x] Collective.ProposalOf (identity) (Hash preimage, from Proposals)
- [x] Collective.Voting (identity) (Hash preimage, from Proposals)
- [x] Imonline.ReceivedHeartbeats
- [x] Imonline.AuthoredBlocks
- [x] Recovery.Recoverable
- [x] Recovery.Proxy
- [x] Historical.HistoricalSessions
- [x] Historical.CachedObsolete
- [x] Treasury.Proposals

### Probably unneeded migration as empty in Kusama
- [x] Staking.ValidatorSlashInEra (twox_64_concat, twox_64_concat) (As long as it's quiet)
- [x] Staking.NominatorSlashInEra (twox_64_concat, twox_64_concat) (As long as it's quiet)
- [x] Staking.UnappliedSlashes (twox_64_concat)
- [x] Staking.SpanSlash (twox_64_concat)
- [x] Offences.Reports (twox_64_concat) (As long as it's quiet, we just clear the exiting entries)
- [x] Offences.ReportsByKindIndex (twox_64_concat) (As long as it's quiet, we just clear the exiting entries)
- [x] Offences.ConcurrentReportsIndex (twox_64_concat, twox_64_concat) (As long as it's quiet, we just clear the exiting entries)
- [x] Society.SuspendedCandidates (twox_64_concat)
- [x] Society.SuspendedMembers (twox_64_concat)
- [x] Society.Vouching (twox_64_concat)
- [x] Society.Strikes (twox_64_concat) (Should be empty)
- [x] Democracy.Blacklist (identity) (Should be empty - will clear)
- [x] Democracy.Cancellations (identity) (Should be empty - will clear)

### Without migration - unused in Kusama
- [x] System.EventTopics
- [x] Balances.Account
- [x] Nicks.NameOf
- [x] Assets.Balances
- [x] Assets.TotalSupply
- [x] Parachains...
- [x] Contracts...
- [x] Evm...
- [x] ScoredPool.CandidateExists (twox_64_concat)
- [x] Example...
- [x] Benchmark...


frame/balances/src/lib.rs
```diff
@@ -372,11 +372,11 @@ decl_storage! {
 		/// is ever zero, then the entry *MUST* be removed.
 		///
 		/// NOTE: This is only used in the case that this module is used to store balances.
-		pub Account: map hasher(blake2_256) T::AccountId => AccountData<T::Balance>;
+		pub Account: map hasher(blake2_128_concat) T::AccountId => AccountData<T::Balance>;
 
 		/// Any liquidity locks on some account balances.
 		/// NOTE: Should only be accessed when setting, changing and freeing a lock.
-		pub Locks get(fn locks): map hasher(blake2_256) T::AccountId => Vec<BalanceLock<T::Balance>>;
+		pub Locks get(fn locks): map hasher(twox_64_concat) T::AccountId => Vec<BalanceLock<T::Balance>>;
```
```suggestion
		pub Locks get(fn locks): map hasher(blake2_128_concat) T::AccountId => Vec<BalanceLock<T::Balance>>;
```

I believe a lock can be created against any target using the `vested_transfer` feature.

bin/node-template/pallets/template/src/mock.rs
```diff
@@ -40,7 +40,7 @@ impl system::Trait for Test {
 	type Version = ();
 	type ModuleToIndex = ();
 	type AccountData = ();
-	type OnNewAccount = ();
+	type MigrateAccount = (); type OnNewAccount = ();
```
bkchr: This pattern repeats over multiple files. Looks like some broken s&r

gavofyork: yeah, my editor doesn't properly support multiline s&r. but since all of this code will be removed 24 hours from now (and i know it's a trivial operation for my editor as long as there's no multiline), i don't really see the point in trying to fix it.

bkchr: k

gavofyork:
I've tested the migration code. It takes a while to run, but as far as I can tell it works as expected. Would be good to get @rphmeier's eyes on the staking migration stuff.

frame/elections/src/lib.rs
```diff
@@ -236,16 +236,16 @@ decl_storage! {
 		// [`all_approvals_of`]. Furthermore, each vector of scalars is chunked with the cap of
 		// `APPROVAL_SET_SIZE`.
 		pub ApprovalsOf get(fn approvals_of):
-			map hasher(blake2_256) (T::AccountId, SetIndex) => Vec<ApprovalFlag>;
+			map hasher(twox_64_concat) (T::AccountId, SetIndex) => Vec<ApprovalFlag>;
 		/// The vote index and list slot that the candidate `who` was registered or `None` if they
 		/// are not currently registered.
 		pub RegisterInfoOf get(fn candidate_reg_info):
-			map hasher(blake2_256) T::AccountId => Option<(VoteIndex, u32)>;
+			map hasher(blake2_128_concat) T::AccountId => Option<(VoteIndex, u32)>;
```
shawntabrizi:
```suggestion
			map hasher(twox_64_concat) T::AccountId => Option<(VoteIndex, u32)>;
```

As far as I can see, this key is set with `submit_candidacy`, and we insert `let who = ensure_signed(origin)`.

So safety by transaction here.

frame/elections/src/lib.rs
```diff
@@ -266,7 +266,7 @@ decl_storage! {
 
 		/// Who is able to vote for whom. Value is the fund-holding account, key is the
 		/// vote-transaction-sending account.
-		pub Proxy get(fn proxy): map hasher(blake2_256) T::AccountId => Option<T::AccountId>;
+		pub Proxy get(fn proxy): map hasher(blake2_128_concat) T::AccountId => Option<T::AccountId>;
```
shawntabrizi: This storage item is unused in the pallet. What am I missing?

gavofyork: It's not actually unused (search `Self::proxy()`) but it's never written to. Should put an issue up for it.

frame/system/src/lib.rs
```diff
@@ -337,10 +341,8 @@ impl From<sp_version::RuntimeVersion> for LastRuntimeUpgradeInfo {
 decl_storage! {
 	trait Store for Module<T: Trait> as System {
 		/// The full account information for a particular account ID.
-		// TODO: should be hasher(twox64_concat) - will need staged migration
-		// https://github.com/paritytech/substrate/issues/4917
 		pub Account get(fn account):
-			map hasher(blake2_256) T::AccountId => AccountInfo<T::Index, T::AccountData>;
+			map hasher(twox_64_concat) T::AccountId => AccountInfo<T::Index, T::AccountData>;
```
```suggestion
			map hasher(blake2_128_concat) T::AccountId => AccountInfo<T::Index, T::AccountData>;
```

This map can get populated with an arbitrary balance transfer, or any other arbitrary implementation of `AccountData`

bin/node/runtime/src/lib.rs
```diff
@@ -140,6 +140,7 @@ impl frame_system::Trait for Runtime {
 	type Version = Version;
 	type ModuleToIndex = ModuleToIndex;
 	type AccountData = pallet_balances::AccountData<Balance>;
+	type MigrateAccount = Balances;
```
shawntabrizi:
```suggestion
	type MigrateAccount = (Balances, Identity, Democracy, Elections, ImOnline, Recovery, Session, Society, Staking, Vesting);
```

frame/elections/src/lib.rs
shawntabrizi: Created https://github.com/paritytech/substrate/issues/5261

frame/staking/src/lib.rs
```diff
@@ -719,23 +717,23 @@ decl_storage! {
 		pub Invulnerables get(fn invulnerables) config(): Vec<T::AccountId>;
 
 		/// Map from all locked "stash" accounts to the controller account.
-		pub Bonded get(fn bonded): map hasher(blake2_256) T::AccountId => Option<T::AccountId>;
+		pub Bonded get(fn bonded): map hasher(twox_64_concat) T::AccountId => Option<T::AccountId>;
 
 		/// Map from all (unlocked) "controller" accounts to the info regarding the staking.
 		pub Ledger get(fn ledger):
-			map hasher(blake2_256) T::AccountId
+			map hasher(blake2_128_concat) T::AccountId
```
kianenigma: why is this one `blake2_128_concat`? I see that the user can set the controller, but logically the user can also decide the origin which dictates the key for `Bonded`, so what is the difference?

kianenigma: (controller is actually indirectly mapped to accountId through StaticLookup path, maybe that makes an impact. )

kianenigma:
Is there anything particular about staking's migration that makes it tricky, or you flagged it because it cannot be tested? it looked more or less okay to me but I will read more.

frame/staking/src/lib.rs
shawntabrizi: Because we know a stash account must call the functions to set the controller account, we know that the caller has a corresponding private key, and this imposes a restriction on the list of available addresses that can be used. A user would not be able to generate both the public and private key for an account with any arbitrary bytes. Hence a stash account would be okay to hash with `twox_64_concat`

However, when a stash sets a controller, they specify the controller account without any restrictions. A stash account could literally use any 32 bytes as the target for the controller. As a result, this could be a possible attack vector for creating an imbalanced trie, and thus we need to use `blake2_128_concat`

gavofyork:
> Is there anything particular about staking's migration that makes it tricky, or you flagged it because it cannot be tested? it looked more or less okay to me but I will read more.

Yeah, actually it's one of the simpler ones. Basically there are two key things to check:
- all items whose storage hasher has changed have actually been migrated
- the maps which have special key domains are treated correctly (democracy, grandpa, identity, im_online (this one's particularly), session & especially session.historical (@rphmeier).

frame/babe/src/lib.rs
```diff
@@ -188,6 +188,12 @@ decl_module! {
 				Self::deposit_vrf_output(&vrf_output);
 			}
 		}
+
+		fn on_runtime_upgrade() {
+			for i in 0..=SegmentIndex::get() {
+				UnderConstruction::migrate_key_from_blake(i);
```
gavofyork: @rphmeier is this correct?

frame/grandpa/src/lib.rs
```diff
@@ -174,15 +174,25 @@ decl_storage! {
...
 		config(authorities): AuthorityList;
 		build(|config| Module::<T>::initialize_authorities(&config.authorities))
 	}
 }
 
+mod migration {
+	use super::*;
+	pub fn migrate<T: Trait>() {
+		for i in 0..=CurrentSetId::get() {
+			SetIdSession::migrate_key_from_blake(i);
```
gavofyork: @rphmeier please confirm this is correct.

frame/im-online/src/lib.rs
```diff
@@ -301,12 +303,38 @@ decl_error! {
 	}
 }
 
+mod migration {
+	use super::*;
+	use frame_support::Blake2_256;
+	pub fn migrate<T: Trait>() {
```
gavofyork: @cmichi @tomusdrw please check this logic that of `migrate_account` below

frame/session/src/historical.rs
```diff
@@ -71,6 +71,20 @@ decl_module! {
...
+		fn on_runtime_upgrade() {
+			migration::migrate::<T>();
+		}
+	}
+}
+
+mod migration {
+	use super::*;
+	pub fn migrate<T: Trait>() {
+		if let Some((begin, end)) = StoredRange::get() {
```
gavofyork: @rphmeier please confirm this is correct.

frame/session/src/lib.rs
```diff
@@ -498,36 +499,23 @@ decl_module! {
...
-		}
-		for (hash, value) in StorageIterator::<T::ValidatorId>::with_suffix(b"Session", b"KeyOwner", &prefix[..]).drain() {
-			put_storage_value(b"Session", b"KeyOwner", &hash, value);
+impl<T: Trait> MigrateAccount<T::AccountId> for Module<T> {
+	fn migrate_account(a: &T::AccountId) {
+		use codec::Encode;
+		if let Ok(v) = a.using_encoded(|mut d| T::ValidatorId::decode(&mut d)) {
+			if let Some(keys) = NextKeys::<T>::migrate_key_from_blake(v) {
+				for id in T::Keys::key_ids() {
+					KeyOwner::<T>::migrate_key_from_blake((*id, keys.get_raw(*id)));
```
gavofyork: @rphmeier please confirm this is correct.

frame/babe/src/lib.rs
rphmeier: Looks like it

frame/grandpa/src/lib.rs
rphmeier: Looks fine to me

frame/session/src/historical.rs
rphmeier: Looks fine

frame/session/src/lib.rs
```diff
@@ -498,36 +499,23 @@ decl_module! {
...
-
-		for (hash, value) in StorageIterator::<T::Keys>::with_suffix(b"Session", b"NextKeys", &prefix[..]).drain() {
-			put_storage_value(b"Session", b"NextKeys", &hash, value);
-		}
-		for (hash, value) in StorageIterator::<T::ValidatorId>::with_suffix(b"Session", b"KeyOwner", &prefix[..]).drain() {
-			put_storage_value(b"Session", b"KeyOwner", &hash, value);
+impl<T: Trait> MigrateAccount<T::AccountId> for Module<T> {
+	fn migrate_account(a: &T::AccountId) {
+		use codec::Encode;
+		if let Ok(v) = a.using_encoded(|mut d| T::ValidatorId::decode(&mut d)) {
```
rphmeier: This should use `T::ValidatorIdOf::convert(a.clone())`. The code as-is attempts to interpret an account key as a `ValidatorId` (in practice - stash key) where what we really want to do is look up the stash key by account ID in the staking module. This is what `ValidatorIdOf` does in practice.

frame/session/src/lib.rs
rphmeier: All logic other than the piece I commented seems correct.

bin/node/runtime/src/lib.rs
```diff
@@ -140,6 +140,7 @@ impl frame_system::Trait for Runtime {
 	type Version = Version;
 	type ModuleToIndex = ModuleToIndex;
 	type AccountData = pallet_balances::AccountData<Balance>;
+	type MigrateAccount = (Balances, Identity, Democracy, Elections, ImOnline, Recovery, Session, Society, Staking, Vesting);
```
apopiak, line 143: Note for people using this as inspiration for migration:
This order is incorrect because the account migration for `Session` depends on the migration for `Staking`. It thus needs to be e.g. `... Recovery, Staking, Session, Society, ...`

frame/staking/src/lib.rs
```diff
@@ -1480,6 +1478,30 @@ decl_module! {
...
+	fn migrate_account(a: &T::AccountId) {
+		if let Some(controller) = Bonded::<T>::migrate_key_from_blake(a) {
+			Ledger::<T>::migrate_key_from_blake(controller);
+			Payee::<T>::migrate_key_from_blake(a);
+			Validators::<T>::migrate_key_from_blake(a);
+			Nominators::<T>::migrate_key_from_blake(a);
+			SlashingSpans::<T>::migrate_key_from_blake(a);
+		}
+	}
+}
```
apopiak, line 1491: Note for people using this as inspiration for migration:
This migration will likely be incorrect, as the `Session` account migration depends on the stash account migration to happen before the controller account migration which means that the order of account migration cannot be arbitrary (and thus just enumerating them is not enough to guarantee a successful migration).
See [here](https://github.com/hicommonwealth/substrate/pull/17/files#diff-32ce5961608780655bf3e00b2cc07d59R2222) for an order independent implementation.

Issue #5211, opened 2020-03-10
txpool resubmission race condition on fast block authoring


gnunicorn:
when authoring blocks fast (sub-second with POW), it can happen that the pruning process for the extrinsic in a block hasn't actually removed the blocks from the pool before an attempt is made to author another block. because it has already been included, it is then marked as `Invalid` and removed and the RPC is informed – rather than given then information that it has been included in the previous block.

Leading to the odd case, that over RPC we are informed the block was invalid, on chain, however, we can see that it has been applied.

bkchr:
@NikVolf is this solved now?

NikVolf:
@bkchr Yes

Issue #5139, opened 2020-03-05
Some invalid transactions removed from the pool should be revalidated


tomusdrw:
During block authorship we attempt to detect invalid transactions that are still in the pool.
Since pool is being cleaned asynchronously, it is a completely legal and valid situation - in the front of the ready queue we have transactions that just got included to recently imported/authored block.
However the transactions that are detected invalid are removed from the pool via `remove_invalid` call.
That in turn causes the entire subtree of transactions to be removed (i.e. all transactions that depend on the invalid ones), as we assume that since their predecessor is invalid they are going to be invalid too.

As an example, let's imagine we have a following set of transactions in the pool (coming from the same sender):
```
tx(nonce=1)
tx(2)
tx(3)
tx(4)
```
now, we import a block that contains `tx(1)`, but don't clear the pool.

If we attempt to produce another block, `tx(1)` will cause `InvalidTransaction::Stale` error to be returned, hence we will mark it as invalid.
Next we will try to execute `tx(2)`, but it will obviously succeed.
Then imagine that `tx(3)` does not fit to the block - we finish up block authoring and should resume the next block starting off with including `tx(3)`.

However `remove_invalid(tx(1))` will cause the entire subtree originating in `tx(1)` to be removed from the pool, so `tx(3)` and `tx(4)` will no longer be there.

There are two ways to solve that issue:
1. Either attempt to resubmit transactions that were not directly invalid, but just transitevly (`tx(2),tx(3),tx(4)` in the example)
2. Or remove only invalid transactions reported directly, not their subtrees - this may leave some more invalid transactions in the pool, but they should just be detected on next attempts.

I'd vote for option (2), cause it's more suited for frame-based runtimes, since most likely the transactions left in the pool are still valid. We can possibly implement both solutions and have that configurable as well.

tomusdrw:
CC @NikVolf let me know if you want to tackle this one.

NikVolf:
> now, we import a block that contains tx(1), but don't clear the pool.

(1) why we don't clear the pool, then?

(2) I think we go with variant 2, but we can differentiate and don't remove subtree when transaction is `Stale` and remove subtree when it is invalid for other reason.

tomusdrw:
(1) we do that asynchronously currently, and we don't really know when the notification task will be picked up.

(2) Yeah that sounds reasonable too.

NikVolf:
[...]

PR #5227, opened 2020-03-11
Produce block always on updated transaction pool state

NikVolf:
This mostly fixes pipeline problems we experience during block production (#5211, partially #5139, but the latter may require small follow-up) and makes so we always produce block with updated transaction pool state.

The idea is that pending set iterator returned only when transaction pool has processed previous block update.

This will:
- avoid validating transactions that are included in previous block
- avoid marking invalid transactions that made it to the previous block and reimported
- avoid sending invalid rpc updates associated with above behaviour
- generally make block import-production more correct

This might induce small blocking before starting block production, but generally should increase speed, since we won't validate something that is known to be invalid.


NikVolf:
> Can we get a test?

There are multiple tests altered and unaltered to show this should work

I'll add dedicated one to specify behaviour though, yes

client/basic-authorship/src/basic_authorship.rs
```diff
@@ -210,7 +210,18 @@ impl<A, B, Block, C> ProposerInner<B, Block, C, A>
 		let mut is_first = true;
 		let mut skipped = 0;
 		let mut unqueue_invalid = Vec::new();
-		let pending_iterator = self.transaction_pool.ready();
+		let pending_iterator = match executor::block_on(future::select(
+			self.transaction_pool.ready_at(Some(self.parent_number)),
+			futures_timer::Delay::new(deadline - (self.now)()),
```
tomusdrw: That might not give us enough time to produce anything. I think it should be half of the remaining time to at least produce an empty block.

client/basic-authorship/src/basic_authorship.rs
```diff
@@ -210,7 +210,18 @@ impl<A, B, Block, C> ProposerInner<B, Block, C, A>
...
+		)) {
+			Either::Left((iterator, _)) => iterator,
+			Either::Right(_) => {
+				debug!(
+					"Consensus deadline reached while waiting transaction pool to reutnr pending set. Returning error!",
```
tomusdrw: Wouldn't it be better to at least produce an empty block instead?
```suggestion
					"Consensus deadline reached while waiting for transaction pool to return the pending set. Returning error!",
```

client/basic-authorship/src/basic_authorship.rs
```diff
@@ -371,6 +398,17 @@ mod tests {
 			txpool.submit_at(&BlockId::number(0), vec![extrinsic(0)]),
 		).unwrap();
 
+		futures::executor::block_on(
+			txpool.maintain(
```
tomusdrw: This is repeated couple of times already, can we extract it into a helper function?

client/transaction-pool/src/lib.rs
```diff
@@ -54,6 +60,55 @@ pub struct BasicPool<PoolApi, Block>
...
+
+impl<T, Block> ReadyPoll<T, Block>
+where
+	Block: BlockT,
+{
+	fn trigger(&mut self, number: NumberFor<Block>, iterator_factory: impl Fn() -> T) {
+		self.updated_at = Some(number);
+		let pollers = self.pollers.drain(..).collect::<Vec<(_, _)>>();
+
+		for (poller_number, poller_sender) in pollers {
```
tomusdrw: That loop look like it could just be `self.pollers.retain()` no?

client/transaction-pool/src/lib.rs
```diff
@@ -196,8 +252,21 @@ impl<PoolApi, Block> TransactionPool for BasicPool<PoolApi, Block>
...
 
-	fn ready(&self) -> Box<dyn Iterator<Item=Arc<Self::InPoolTransaction>>> {
-		Box::new(self.pool.validated_pool().ready())
+	fn ready_at(&self, at: Option<NumberFor<Self::Block>>) -> PolledIterator<PoolApi> {
+		if self.ready_poll.lock().updated_at() >= at.unwrap_or(NumberFor::<Self::Block>::zero()) {
+			let iterator: ReadyIteratorFor<PoolApi> = Box::new(self.pool.validated_pool().ready());
+			Box::pin(futures::future::ready(iterator))
+		} else {
+			Box::pin(self.ready_poll.lock().add(at.expect("Should be Some(_) otherwise expression above is true.")).map(|received| {
+				match received {
```
tomusdrw:
```
received.unwrap_or_else(|| {
  log::warn!("Error receiving pending set: {:?}", e);
  Box::new(vec![].into_iter())
})
```

client/transaction-pool/src/lib.rs
```diff
@@ -348,6 +418,9 @@ impl<PoolApi, Block> MaintainedTransactionPool for BasicPool<PoolApi, Block>
 						}
 					}
 
+					let extra_pool = pool.clone();
+					ready_poll.lock().trigger(block_number, move || Box::new(extra_pool.validated_pool().ready()));
```
tomusdrw: As mentioned in private convos, the `ChainEvent::NewEvent` is not super reliable, so we might end up skipping block `trigger`. We should switch to `all_blocks_import_notification_stream` when #5200 lands

NikVolf: I am not sure, but we'll see, I'll add a reference/todo

Not sure because maybe `ChainEvent::NewBlock` is "unreliable" in a sense that we need here - so if we are catching up with network, we also shouldn't make attempts to produce blocks

client/basic-authorship/src/basic_authorship.rs
NikVolf: I somewhat agree

But also there is an argument that we must just fail here instead of trying to recover from some invalid system state (transaction pool does not receive new block notifications or cannot handle them fast for some reason). Timeout is therefore here (for now) just for avoiding thread hanging forever, not for some logic after it.

client/transaction-pool/src/lib.rs
NikVolf: oneshot `Sender::send` is by-value

NikVolf:
> Can we get a test?

@bkchr added!

client/basic-authorship/src/basic_authorship.rs
NikVolf: Updated to warning & continue

primitives/transaction-pool/src/pool.rs
```diff
@@ -210,8 +210,17 @@ pub trait TransactionPool: Send + Sync {
 	) -> PoolFuture<Box<TransactionStatusStreamFor<Self>>, Self::Error>;
 
 	// *** Block production / Networking
-	/// Get an iterator for ready transactions ordered by priority
-	fn ready(&self) -> Box<dyn Iterator<Item=Arc<Self::InPoolTransaction>>>;
+	/// Get an iterator for ready transactions ordered by priority.
+	///
+	/// Guarantees to return only when transaction pool got updated at `at` block.
+	/// Guarantees to return immediately when `None` is passed.
+	fn ready_at(&self, at: Option<NumberFor<Self::Block>>)
```
bkchr: Don't see the reason for `at` being an `Option` here.

If you don't want to have the `at` parameter, you can just call `ready`

client/transaction-pool/src/lib.rs
```diff
@@ -54,6 +60,57 @@ pub struct BasicPool<PoolApi, Block>
...
+}
+
+impl<T, Block> ReadyPoll<T, Block>
+where
+	Block: BlockT,
+{
+	fn trigger(&mut self, number: NumberFor<Block>, iterator_factory: impl Fn() -> T) {
+		self.updated_at = Some(number);
+
+		let pollers = self.pollers.drain(..).collect::<Vec<(_, _)>>();
```
bkchr: I don't like that we allocate each time we call this function :( 

```
let mut i = 0;
while i < self.pollers.len() {
    if self.pollers[i].0 <= number {
        let poller_sender = self.pollers.swap_remove(i);
				log::debug!(target: "txpool", "Sending ready signal at block {}", number);
				let _s = poller_sender.send(iterator_factory());
    } else {
        i += 1;
    }
}
```

bkchr: This would not require the extra allocation.

client/transaction-pool/src/lib.rs
```diff
@@ -54,6 +60,57 @@ pub struct BasicPool<PoolApi, Block>
...
+	}
+
+	fn add(&mut self, number: NumberFor<Block>) -> oneshot::Receiver<T> {
+		let (sender, receiver) = oneshot::channel();
+		self.pollers.push((number, sender));
+		receiver
+	}
+
+	fn updated_at(&self) -> NumberFor<Block> {
+		self.updated_at.unwrap_or(NumberFor::<Block>::zero())
```
bkchr: Why is `updated_at` an `Option`? By default you always return `0` anyway.

client/basic-authorship/src/basic_authorship.rs
```diff
@@ -210,7 +210,18 @@ impl<A, B, Block, C> ProposerInner<B, Block, C, A>
 		let mut is_first = true;
 		let mut skipped = 0;
 		let mut unqueue_invalid = Vec::new();
-		let pending_iterator = self.transaction_pool.ready();
+		let pending_iterator = match executor::block_on(future::select(
+			self.transaction_pool.ready_at(Some(self.parent_number)),
+			futures_timer::Delay::new((deadline - (self.now)()) / 8),
```
bkchr: Hmm this deadline seem really random.

Did you test this with a `dev` chain node and tx spamming?

client/consensus/manual-seal/src/lib.rs
```diff
@@ -432,14 +432,24 @@ mod tests {
...
 		let (tx1, rx1) = futures::channel::oneshot::channel();
 		assert!(sink.send(EngineCommand::SealNewBlock {
-			parent_hash: Some(created_block.hash.clone()),
+			parent_hash: Some(created_block.hash),
 			sender: Some(tx1),
 			create_empty: false,
 			finalize: false,
 		}).await.is_ok());
-		assert!(rx1.await.unwrap().is_ok());
+		assert_matches::assert_matches!(
```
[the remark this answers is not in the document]
NikVolf: is_ok is not okay here, it is impossible to tell what failed if it is not ok

client/basic-authorship/src/basic_authorship.rs
NikVolf: See [discussion](https://github.com/paritytech/substrate/pull/5227#discussion_r391520954) above with @tomusdrw.

This timeout is merely avoid thread locking if something is wrong with the pipeline. The future itself supposed to resolve almost instantly (or made to do so if not). If you have ideas what would be less random for this cases, pls share.

I always test prs

client/transaction-pool/src/lib.rs
NikVolf: I would've bothered if this happened more often than once per block

will update

client/consensus/manual-seal/src/lib.rs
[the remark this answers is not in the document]
NikVolf: `assert!(matches!(...))` is bad as well I suppose, it will just tell that expression was false, not what was the result that haven't matched

client/basic-authorship/src/basic_authorship.rs
bkchr: I don't wanted to say that you don't test stuff :D I just don't know how long it takes until the transaction pool is ready after producing a block? However, I would assume that with tx spamming on a dev-chain we see some problems (skipping blocks), but if you say that everything is fine, I'm happy :)

NikVolf: Everything is fine :)

client/basic-authorship/src/basic_authorship.rs
```diff
@@ -459,15 +501,26 @@ mod tests {
...
+				1,
+				client.header(&BlockId::Number(1)).expect("header get error").expect("there should be header")
+			))
+		);
 
-		// This is most likely incorrect, and caused by #5139
-		let tx_remaining = 0;
-		let block = propose_block(&client, 1, 2, tx_remaining);
+		// now let's make sure that we can still make some progress
+		let block = propose_block(&client, 1, 2, 5);
```
bkchr: Why did we had before `0` remaining and now `5`?

client/transaction-pool/src/lib.rs
```diff
@@ -54,6 +60,51 @@ pub struct BasicPool<PoolApi, Block>
 	api: Arc<PoolApi>,
 	revalidation_strategy: Arc<Mutex<RevalidationStrategy<NumberFor<Block>>>>,
 	revalidation_queue: Arc<revalidation::RevalidationQueue<PoolApi>>,
+	ready_poll: Arc<Mutex<ReadyPoll<ReadyIteratorFor<PoolApi>, Block>>>,
+}
+
+struct ReadyPoll<T, Block: BlockT>
```
bkchr:
```suggestion
struct ReadyPoll<T, Block: BlockT> {
```

client/transaction-pool/src/lib.rs
```diff
@@ -54,6 +60,51 @@ pub struct BasicPool<PoolApi, Block>
...
+{
```
bkchr:
```suggestion
```

client/transaction-pool/src/lib.rs
```diff
@@ -54,6 +60,51 @@ pub struct BasicPool<PoolApi, Block>
...
+	updated_at: NumberFor<Block>,
+	pollers: Vec<(NumberFor<Block>, oneshot::Sender<T>)>,
+}
+
+impl<T, Block: BlockT> Default for ReadyPoll<T, Block>
```
bkchr:
```suggestion
impl<T, Block: BlockT> Default for ReadyPoll<T, Block> {
```

client/transaction-pool/src/lib.rs
```diff
@@ -54,6 +60,51 @@ pub struct BasicPool<PoolApi, Block>
...
+{
```
bkchr:
```suggestion
```

client/transaction-pool/src/lib.rs
```diff
@@ -54,6 +60,51 @@ pub struct BasicPool<PoolApi, Block>
...
+	fn default() -> Self {
+		Self {
+			updated_at: NumberFor::<Block>::zero(),
+			pollers: Default::default(),
+		}
+	}
+}
+
+impl<T, Block: BlockT> ReadyPoll<T, Block>
```
bkchr:
```suggestion
impl<T, Block: BlockT> ReadyPoll<T, Block> {
```

client/transaction-pool/src/lib.rs
```diff
@@ -54,6 +60,51 @@ pub struct BasicPool<PoolApi, Block>
...
+{
```
bkchr:
```suggestion
```

client/basic-authorship/src/basic_authorship.rs
NikVolf: There were 7 initially, and 2 was included in previous block
(see deleted comment above that outlines that 0 was an incorrect observation caused by the bug this pr aims to fix among other things)

Issue #5223, opened 2020-03-11
Emit events for liquidity changes in Staking


joepetrowski:
Some infrastructure providers reconcile spendable balance against on-chain information (events, tx parameters). A number of functions in the Staking pallet update locks by an amount that is not known at dispatch, e.g. `unbond` will [unbond](https://github.com/paritytech/substrate/blob/e896cf3da7a4d691789d874551fbbc850232e6f8/frame/staking/src/lib.rs#L1083) `let mut value = value.min(ledger.active);`

We should either emit an event with the actual value (un)bonded or return an error if the input parameters are not valid for the operation.

PR #5228, opened 2020-03-11
Emit events to bonding and unbonding ops in Staking pallet

joepetrowski:
Closes https://github.com/paritytech/substrate/issues/5223

The reason to emit events is to find out the actual change to the staking ledger as users could call `bond`, `bond_extra`, `unbond`, or `withdraw_unbonded` with a different `value` than will actually update the staking ledger.

### DIFF
--- a/frame/staking/src/lib.rs
+++ b/frame/staking/src/lib.rs
@@ -1163,7 +1163,7 @@ decl_module! {
 			if ledger.total < old_total {
 				// Already checked that this won't overflow by entry condition.
 				let value = old_total - ledger.total;
-				Self::deposit_event(RawEvent::Withdrawn(stash, value))
+				Self::deposit_event(RawEvent::Withdrawn(stash, value));
 			}
 		}
 
--- a/frame/staking/src/lib.rs
+++ b/frame/staking/src/lib.rs
@@ -897,6 +897,16 @@ decl_event!(
 		/// An old slashing report from a prior era was discarded because it could
 		/// not be processed.
 		OldSlashingReportDiscarded(SessionIndex),
+		/// An account has bonded this amount.
+		///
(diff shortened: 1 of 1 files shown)


kianenigma:
I am slightly against because at least the example of unbond can easily be determined by the origin/service provider. Generally, (and preferably) the chain should be seen as lazy unit that does something (i.e. store and emit an event) only when it absolutely has to. I think this stuff can be computed offchain. 

Nonetheless, if it is worth the cost of an event, I am also okay with this.

frame/staking/src/lib.rs
```diff
@@ -1096,6 +1115,7 @@ decl_module! {
 				let era = Self::current_era().unwrap_or(0) + T::BondingDuration::get();
 				ledger.unlocking.push(UnlockChunk { value, era });
 				Self::update_ledger(&controller, &ledger);
+				Self::deposit_event(RawEvent::Unbonded(stash.clone(), value));
```
kianenigma:
```suggestion
				Self::deposit_event(RawEvent::Unbonded(ledger.stash.clone(), value));
```
I do see a reason to have a one time used variable `stash` here.

frame/staking/src/lib.rs
```diff
@@ -1119,23 +1141,30 @@ decl_module! {
 		fn withdraw_unbonded(origin) {
 			let controller = ensure_signed(origin)?;
 			let mut ledger = Self::ledger(&controller).ok_or(Error::<T>::NotController)?;
+			let (stash, old_total) = (ledger.stash.clone(), ledger.total.clone());
```
kianenigma:
```suggestion
			let (stash, old_total) = (ledger.stash.clone(), ledger.total);
```

PR #5229, opened 2020-03-11
Fix identity benchmark

### LINKED ISSUE #5217: Issues with Identity benchmarks
When compiling with `cargo build --release --features=runtime-benchmarks`, running the functions `set_subs`, `clear_identity` and `kill_identity` fail:

```bash
./target/release/substrate benchmark --chain dev --execution=wasm --wasm-execution=compiled --pallet pallet-identity --extrinsic set_subs --steps 1 --repeat 1
Error: "NotFound"

./target/release/substrate benchmark --chain dev --execution=wasm --wasm-execution=compiled --pallet pallet-identity --extrinsic clear_identity --steps 1 --repeat 1
Error: "NotFound"

./target/release/substrate benchmark --chain dev --execution=wasm --wasm-execution=compiled --pallet pallet-identity --extrinsic kill_identity --steps 1 --repeat 1
Error: "NotFound"
```

The error is probably this one:
```rust
decl_error! {
	/// Error for the identity module.
	pub enum Error for Module<T: Trait> {
		/// ...
		/// Account isn't found.
		NotFound
		/// ...
```

I looked through the code in `frame/identity/src/benchmarking.rs` and `frame/identity/src/lib.rs`, but I'm not familiar enough with the Pallet internals to (quickly) figure out what the correct fix is for this.

For example, removing:

```rust
ensure!(<IdentityOf<T>>::contains_key(&sender), Error::<T>::NotFound);
```

... from `fn set_subs` makes the old error disappear and replaces it with `Error: "InsufficientBalance"`.

marcio-diaz:
fixes #5217 

We weren't setting identity before calling `set_subs`.


frame/identity/src/benchmarking.rs
```diff
@@ -91,7 +97,7 @@ benchmarks! {
 	// These are the common parameters along with their instancing.
 	_ {
 		let r in 1 .. MAX_REGISTRARS => add_registrars::<T>(r)?;
-		let s in 1 .. T::MaxSubAccounts::get() => {
+		let s in 1 .. T::MaxSubAccounts::get() - 1 => {
```
bkchr: Why `-1`?

marcio-diaz: Because `set_subs` do one more (and then we get an error) and to keep a common range. Another option may be to remove it here and add it only at `set_subs` benchmark.

marcio-diaz: Mmm, looks like the check is wrong `ensure!(subs.len() <= T::MaxSubAccounts::get() as usize, Error::<T>::TooManySubAccounts);`, should be `>` according to the docs: `/// The maximum number of sub-accounts allowed per identified account.`

gavofyork: why does `set_subs` do one more?

marcio-diaz: I don't see a (strong) reason to push `n + 1`. The test was written in that way before the pr. I'll change it to push `n`.

shawntabrizi: Yeah seems like just an oversight or something

PR #5230, opened 2020-03-11
Update master benchmarks

NikVolf:
There were couple of benchmarks that were only in `nv-bench-base` branch and used by bot, they have some general use also.


bin/node/testing/src/bench.rs
```diff
@@ -215,7 +256,10 @@ impl BenchDb {
 					function: Call::Balances(
 						BalancesCall::transfer(
 							pallet_indices::address::Address::Id(receiver),
-							1*DOLLARS
+							match block_type {
+								BlockType::RandomTransfers(_) => 1*DOLLARS,
```
bkchr: I think it would be better to use the `ExistentialDeposit` constant directly.

bin/node/testing/src/bench.rs
```diff
@@ -215,7 +256,10 @@ impl BenchDb {
...
+								BlockType::RandomTransfersReaping(_) => 99*DOLLARS,
```
bkchr: Why 99?

bin/node/testing/benches/import.rs
```diff
@@ -104,7 +104,63 @@ fn bench_block_import(c: &mut Criterion) {
...
+
+	let mut bench_db = BenchDb::with_key_types(100, KeyTypes::Ed25519);
+	let block = bench_db.generate_block(BlockType::RandomTransfers(100));
+
+	c.bench_function_over_inputs("import-block-ed25519-B-0003",
+		move |bencher, profile| {
+			bencher.iter_batched(
+				|| {
+					let context = bench_db.create_context(*profile);
+					let _version = context.client.runtime_version_at(&BlockId::Number(0))
```
bkchr:
```suggestion
					context.client.runtime_version_at(&BlockId::Number(0))
```

bin/node/testing/benches/import.rs
```diff
@@ -104,7 +104,63 @@ fn bench_block_import(c: &mut Criterion) {
...
+						.expect("Failed to get runtime version")
```
bkchr:
```suggestion
						.expect("Failed to get runtime version");
```

bin/node/testing/benches/import.rs
```diff
@@ -104,7 +104,63 @@ fn bench_block_import(c: &mut Criterion) {
...
+						.spec_version;
```
bkchr:
```suggestion
```

bin/node/testing/benches/import.rs
```diff
@@ -104,7 +104,63 @@ fn bench_block_import(c: &mut Criterion) {
...
+	let block = bench_db.generate_block(BlockType::RandomTransfersReaping(100));
+
+	c.bench_function_over_inputs("import-block-reaping-B-0002",
+		move |bencher, profile| {
+			bencher.iter_batched(
+				|| {
+					let context = bench_db.create_context(*profile);
+
+					// mostly to just launch compiler before benching!
+					let _version = context.client.runtime_version_at(&BlockId::Number(0))
```
bkchr:
```suggestion
					context.client.runtime_version_at(&BlockId::Number(0))
```

bin/node/testing/benches/import.rs
```diff
@@ -104,7 +104,63 @@ fn bench_block_import(c: &mut Criterion) {
...
+						.expect("Failed to get runtime version")
+						.spec_version;
```
bkchr:
```suggestion
```

bin/node/testing/benches/import.rs
```diff
@@ -104,7 +104,63 @@ fn bench_block_import(c: &mut Criterion) {
...
+
+	c.bench_function_over_inputs("import-block-reaping-B-0002",
+		move |bencher, profile| {
+			bencher.iter_batched(
+				|| {
+					let context = bench_db.create_context(*profile);
+
+					// mostly to just launch compiler before benching!
+					let _version = context.client.runtime_version_at(&BlockId::Number(0))
+						.expect("Failed to get runtime version")
```
bkchr:
```suggestion
						.expect("Failed to get runtime version");
```

bin/node/testing/src/bench.rs
NikVolf: 1 fee+99 transfer = 100 (initial account endowment), so the balance should become 0

bin/node/testing/src/bench.rs
NikVolf: this is not related to existential deposit

gavofyork: it probably should be - it's important that this value is greater than it.

NikVolf: Right, it makes sense in case it is changed indeed

NikVolf: what is the easiest way to query it from runtime though?

bkchr: `runtime::ExistentialDeposit::get()`

gnunicorn:
PR conflicting, please  merge (unclear to me whether that other import is still needed).

PR #5233, opened 2020-03-12
Allow to expose a subset of unsafe RPCs

Xanewok:
Closes https://github.com/paritytech/substrate/issues/2303

Introduces machinery which allows us to deny/allow certain "unsafe" RPCs if we're exposing it to the network. The last bits (configuration) are fairly easy to add; in the meantime, I'd like to get some feedback on the initial implementation.

The aim is to provide a very simple filtering out-of-box and recommend using something like https://github.com/tomusdrw/jsonrpc-proxy for anything more complicated. We decided against a separate blacklist/whitelist with @tomusdrw since this would make things unnecessarily more scattered and harder to navigate.

Unresolved/To do:
- [ ] Should the configuration leverage existing `--unsafe-{rpc,ws}-external` flags?
- [ ] Should the setting be separate for HTTP and WS servers?
- [ ] Make sure every "unsafe" RPC is included and checked here
- [ ] Should tests be added for each API modified?

### DIFF
--- a/client/cli/src/commands/runcmd.rs
+++ b/client/cli/src/commands/runcmd.rs
@@ -93,6 +93,14 @@ pub struct RunCmd {
 	#[structopt(long = "unsafe-rpc-external")]
 	pub unsafe_rpc_external: bool,
 
+	/// Don't deny potentially unsafe RPCs when listening on external interfaces.
+	///
+	/// Default is false. This allows exposing RPC methods publicly (same as `--unsafe-{rpc,ws}-external` )
+	/// but will allow doing so even on validator nodes, which is prohibited by default.
+	/// Please do this if you know what you're doing.
+	#[structopt(long = "unsafe-rpc-expose")]
+	pub unsafe_rpc_expose: bool,
+
 	/// Listen to all Websocket interfaces.
 	///
 	/// Default is local. Note: not all RPC methods are safe to be exposed publicly. Use an RPC proxy
@@ -406,14 +414,29 @@ impl CliConfiguration for RunCmd {
 	}
 
 	fn rpc_http(&self) -> Result<Option<SocketAddr>> {
-		let interface = rpc_interface(self.rpc_external, self.unsafe_rpc_external, self.validator)?;
+		let interface = rpc_interface(
+			self.rpc_external,
+			self.unsafe_rpc_external,
+			self.unsafe_rpc_expose,
+			self.validator
+		)?;
+
 		Ok(Some(SocketAddr::new(interface, self.rpc_port.unwrap_or(9933))))
 	}
 
 	fn rpc_ws(&self) -> Result<Option<SocketAddr>> {
-		let interface = rpc_interface(self.ws_external, self.unsafe_ws_external, self.validator)?;
+		let interface = rpc_interface(
+			self.ws_external,
+			self.unsafe_ws_external,
+			self.unsafe_rpc_expose,
+			self.validator
+		)?;
 
 		Ok(Some(SocketAddr::new(interface, self.ws_port.unwrap_or(9944))))
+	}
+
+	fn unsafe_rpc_expose(&self) -> Result<bool> {
+		Ok(self.unsafe_rpc_expose)
 	}
 
 	fn offchain_worker(&self, role: &Role) -> Result<bool> {
@@ -459,9 +482,10 @@ pub fn is_node_name_valid(_name: &str) -> std::result::Result<(), &str> {
 fn rpc_interface(
 	is_external: bool,
 	is_unsafe_external: bool,
+	is_unsafe_rpc_expose: bool,
 	is_validator: bool,
 ) -> Result<IpAddr> {
-	if is_external && is_validator {
+	if is_external && is_validator && !is_unsafe_rpc_expose {
 		return Err(Error::Input(
 			"--rpc-external and --ws-external options shouldn't be \
 		used if the node is running as a validator. Use `--unsafe-rpc-external` if you understand \
--- a/client/cli/src/commands/runcmd.rs
+++ b/client/cli/src/commands/runcmd.rs
@@ -27,7 +27,7 @@ use sc_service::{
 	ChainSpec, Role,
 };
 use sc_telemetry::TelemetryEndpoints;
-use std::net::SocketAddr;
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 use structopt::{clap::arg_enum, StructOpt};
 
 arg_enum! {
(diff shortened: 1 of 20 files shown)


xlc:
Is it possible for `rpc_methods` to exclude unsafe rpc endpoints when they are disabled? so client can easily detect if a given RPC is truly supported.

Also please make sure we have an updated list of unsafe RPCs somewhere, https://github.com/paritytech/substrate/wiki/Public-RPC is outdated already.

Is `author_submitExtrinsic` safe or unsafe?

For the RPC node that I would like to have, it should deny all unsafe RPC endpoints but accept `author_submitExtrinsic` so public can connect to it and submit transactions via frontend.

client/rpc/src/state/tests.rs
```diff
@@ -177,7 +177,7 @@ fn should_send_initial_storage_changes_and_notifications() {
 
 	{
 		let mut client = Arc::new(substrate_test_runtime_client::new());
-		let api = new_full(client.clone(), Subscriptions::new(Arc::new(remote)));
+		let api = new_full(client.clone(), Subscriptions::new(Arc::new(remote)), DenyUnsafe::No);
```
NikVolf: I would suggest add method that creates with `DenyUnsafe::No` instead of passing it every time

client/rpc/src/system/mod.rs
```diff
@@ -91,21 +94,41 @@ impl<B: traits::Block> SystemApi<B::Hash, <B::Header as HeaderT>::Number> for Sy
 		Receiver(Compat::new(rx))
 	}
 
-	fn system_peers(&self) -> Receiver<Vec<PeerInfo<B::Hash, <B::Header as HeaderT>::Number>>> {
+	fn system_peers(&self)
+		-> Compat<BoxFuture<'static, rpc::Result<Vec<PeerInfo<B::Hash, <B::Header as HeaderT>::Number>>>>>
+	{
+		if let Err(err) = self.deny_unsafe.check_if_safe() {
```
NikVolf: I would suggest adding helper method that will contain this check and takes closure executed only if check passes

The check will be in one place then

client/cli/src/commands/runcmd.rs
```diff
@@ -559,9 +546,9 @@ fn interface_str(
 		log::warn!("It isn't safe to expose RPC publicly without a proxy server that filters \
 		available set of RPC methods.");
 
-		Ok("0.0.0.0")
+		Ok(Ipv4Addr::UNSPECIFIED.into())
```
NikVolf:
```suggestion
		Ok(Ipv4Addr::UNSPECIFIED)
```

client/cli/src/commands/runcmd.rs
```diff
@@ -559,9 +546,9 @@ fn interface_str(
...
 	} else {
-		Ok("127.0.0.1")
+		Ok(Ipv4Addr::LOCALHOST.into())
```
NikVolf:
```suggestion
		Ok(Ipv4Addr::LOCALHOST)
```

client/cli/src/commands/runcmd.rs
Xanewok: The return type is `IpAddr` (notice lack of protocol version), so I'd have to wrap that as:
```rust
Ok(IpAddr::V4(Ipv4Addr::UNSPECIFIED))
```
so to reduce the stutter I went for the `.into()` approach. Do you want me to use `IpAddr::V4` instead or is `into()` okay?

client/cli/src/commands/runcmd.rs
Xanewok: same as above

client/rpc/src/system/mod.rs
Xanewok: The error type is different for e.g. System and State API - should I add a separate helper method for each API?

The check is used to short-circuit, so it'd have to be a macro; I can't use `?` with a plain helper function since I'd have to materialize an `Ok` value if this succeeds. Is that still okay?

client/rpc/src/state/tests.rs
Xanewok: Added a `new_full_test` macro not to repeat numerous bounds from `new_full` again here

NikVolf:
@tomusdrw / @jacogr might have opinion about this, I don't know all context in which rpc is used

client/cli/src/commands/runcmd.rs
```diff
@@ -530,26 +550,13 @@ fn generate_node_name() -> String {
...
-fn interface_str(
+fn rpc_interface(
 	is_external: bool,
 	is_unsafe_external: bool,
+	is_unsafe_rpc_expose: bool,
 	is_validator: bool,
-) -> Result<&'static str, error::Error> {
-	if is_external && is_validator {
+) -> Result<IpAddr, error::Error> {
+	if is_external && (is_validator && !is_unsafe_rpc_expose) {
```
tomusdrw:
```suggestion
	if is_external && is_validator && !is_unsafe_rpc_expose {
```

client/rpc/src/system/mod.rs
```diff
@@ -120,6 +143,10 @@ impl<B: traits::Block> SystemApi<B::Hash, <B::Header as HeaderT>::Number> for Sy
 	fn system_remove_reserved_peer(&self, peer: String)
 		-> Compat<BoxFuture<'static, std::result::Result<(), rpc::Error>>>
 	{
+		if let Err(err) = self.deny_unsafe.check_if_safe() {
```
tomusdrw: The code repetition is a bit annoying, might be worth a macro actually.

bkchr:
What is the status of this pr?

Xanewok:
This is rebased against the current master and catches up with some CLI updates; I'll add macros to remove some error boilerplate and will need to think about safe way to inspect supported RPC calls.

client/rpc/src/state/mod.rs
```diff
@@ -266,6 +271,10 @@ impl<Block, Client> StateApi<Block::Hash> for State<Block, Client>
 		key_prefix: StorageKey,
 		block: Option<Block::Hash>,
 	) -> FutureResult<Vec<StorageKey>> {
+		if let Err(e) = self.deny_unsafe.check_if_safe() {
```
tomusdrw, line 274: These ones (the whole `state` module) are currently used on our public RPC node, and I think they should not be marked. While they can generate high load on the machine, I think we should keep the flag to thing that are more insecure than simply DDoS vector (for instance we did left out `author_submitExtrinsic` (Which is correct)).

Xanewok: Makes sense, unmarked the entire module.

Xanewok:
Rebased, added a macro (in `system`) to reduce some repetition when checking if a given RPC is safe. Let's see what CI says and this should be good to go :+1:

client/rpc/src/system/mod.rs
```diff
@@ -125,6 +152,10 @@ impl<B: traits::Block> SystemApi<B::Hash, <B::Header as HeaderT>::Number> for Sy
...
```
tomusdrw, line 155: Should be using `bail_if_unsafe!` too, right?

Xanewok: Oops! You're right, fixed. Thanks for spotting this!

gavofyork:
@Xanewok needs resolving.

Xanewok:
Rebased, everything is green now.

client/cli/src/commands/runcmd.rs
```diff
@@ -93,6 +93,14 @@ pub struct RunCmd {
 	#[structopt(long = "unsafe-rpc-external")]
 	pub unsafe_rpc_external: bool,
 
+	/// Don't deny potentially unsafe RPCs when listening on external interfaces.
```
xlc, line 96: ` Don't deny` => `Accept`?
Double negative is hard to understand

>  Default is false. This allows exposing RPC methods publicly (same as `--unsafe-{rpc,ws}-external` )

Default allows to expose or not?

> Please do this if you know what you're doing.

If you are running a development node?

xlc:
This makes CLI super confusing.

We have `--unsafe-rpc-external`, which is "unsafe" "rpc-external" and now
`--unsafe-rpc-expose`, which is "unsafe-rpc" "expose", not "unsafe" "rpc-expose"

I suggest rename it.

PR #5234, opened 2020-03-12
Add ext_terminate call

### LINKED ISSUE #4952: srml-contracts: ext_terminate
Due to upcoming changes with the FRAME balances pallet we won't be able to rely on the callback called upon decreasing a contract's balance below existential deposit / minimum balance.

Instead of that, we should not allow for a contract's balance to go below the minimum balance. That would prevent the contract to destroy itself, so to give a substitute we introduce a dedicated function tentatively called `ext_terminate` with the following signature:

```rust
fn ext_terminate(beneficiary: AccountId) → !;
```

This function would take an ID of an account that would receive all the remaining funds possessed by the terminated contract. Because self-termination and sending all funds implies removing the associated account, this function doesn't return control and finishes the contract execution. The function traps **with reversal** (i.e. not performing the termination) if the contract is already present on the execution stack.

Open questions:

- [ ] It is a bit weird that this function doesn't allow to return a value to the caller since it is similar to `ext_return`.
- [ ] In case that a contract is already present on execution stack should we return a status code? This will give an ability to the caller contract to handle this condition. However, I don't see a way to handle this meaningfully other than trap.

athei:
This PR adds the following contract syscall  ( closes #4952 )
```
ext_terminate(beneficiary_ptr: u32, beneficiary_len: u32);
```

It removes a contract from storage (terminates it) after sending all remaining balance to a beneficiary. The beneficiary can be a contract or a plain account. The call never returns: Neither on success (execution halts and changes are committed) nor on failure (trap and rollback).

This PR also changes the semantics of all other contract initiated balance transfers to fail if they would bring the caller below the existential deposit. This makes `ext_terminate` the only way for a contract to self-destruct.

# Motivation
Due to upcoming changes with the FRAME balances pallet we won't be able to rely on the callback called upon decreasing a contract's balance below existential deposit / minimum balance. Additionally we are also adding some safety by preventing accidental contract terminations.

### DIFF
--- a/frame/contracts/COMPLEXITY.md
+++ b/frame/contracts/COMPLEXITY.md
@@ -350,6 +350,20 @@ Loading `init_code` and `input_data` should be charged in any case.
 
 **complexity**: All complexity comes from loading buffers and executing `instantiate` executive function. The former component is proportional to the sizes of `init_code`, `value` and `input_data` buffers. The latter component completely depends on the complexity of `instantiate` executive function and also dominated by it.
 
+## ext_terminate
+
+This function receives the following arguments:
+
+- `beneficiary`, buffer of a marshaled `AccountId`
+
+It consists of the following steps:
+
+1. Loading `beneficiary` buffer from the sandbox memory (see sandboxing memory get) and then decoding it.
+
+Loading of the `beneficiary` buffer should be charged. This is because the sizes of buffers are specified by the calling code, even though marshaled representations are, essentially, of constant size. This can be fixed by assigning an upper bound for sizes of `AccountId`.
+
+**complexity**: All complexity comes from loading buffers and executing `terminate` executive function. The former component is proportional to the size of the `beneficiary` buffer. The latter component completely depends on the complexity of `terminate` executive function and also dominated by it.
+
 ## ext_return
 
 This function receives a `data` buffer as an argument. Execution of the function consists of the following steps:
--- a/frame/contracts/src/exec.rs
+++ b/frame/contracts/src/exec.rs
@@ -128,6 +128,13 @@ pub trait Ext {
 		gas_meter: &mut GasMeter<Self::T>,
 	) -> Result<(), DispatchError>;
 
+	/// Transfer all funds to `beneficiary` and delete the contract.
+	fn terminate(
+		&mut self,
+		beneficiary: &AccountIdOf<Self::T>,
+		gas_meter: &mut GasMeter<Self::T>,
+	) -> Result<(), DispatchError>;
+
 	/// Call (possibly transferring some amount of funds) into the specified account.
 	fn call(
 		&mut self,
@@ -428,20 +435,6 @@ where
 							gas_meter,
 						)?;
 
-					// Destroy contract if insufficient remaining balance.
-					if nested.overlay.get_balance(&dest) < nested.config.existential_deposit {
-						let parent = nested.parent
-							.expect("a nested execution context must have a parent; qed");
-						if parent.is_live(&dest) {
-							return Err(ExecError {
-								reason: "contract cannot be destroyed during recursive execution".into(),
-								buffer: output.data,
-							});
-						}
-
-						nested.overlay.destroy_contract(&dest);
-					}
-
 					Ok(output)
 				}
 				None => Ok(ExecReturnValue { status: STATUS_SUCCESS, data: Vec::new() }),
@@ -535,9 +528,37 @@ where
 		Ok((dest, output))
 	}
 
-	fn new_call_context<'b>(&'b mut self, caller: T::AccountId, value: BalanceOf<T>)
-		-> CallContext<'b, 'a, T, V, L>
-	{
+	pub fn terminate(
+		&mut self,
+		beneficiary: &T::AccountId,
+		gas_meter: &mut GasMeter<T>,
+	) -> Result<(), DispatchError> {
+		let self_id = self.self_account.clone();
+		let value = self.overlay.get_balance(&self_id);
+		if let Some(parent) = self.parent {
+			if parent.is_live(&self_id) {
+				return Err(DispatchError::Other(
+					"Cannot terminate a contract that is present on the call stack",
+				));
+			}
+		}
+		transfer(
+			gas_meter,
+			TransferCause::Terminate,
+			&self_id,
+			beneficiary,
+			value,
+			self,
+		)?;
+		self.overlay.destroy_contract(&self_id);
+		Ok(())
+	}
+
+	fn new_call_context<'b>(
+		&'b mut self,
+		caller: T::AccountId,
+		value: BalanceOf<T>,
+	) -> CallContext<'b, 'a, T, V, L> {
 		let timestamp = self.timestamp.clone();
 		let block_number = self.block_number.clone();
 		CallContext {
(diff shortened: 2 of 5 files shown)


frame/contracts/src/exec.rs
```diff
@@ -510,9 +502,35 @@ where
...
-		-> CallContext<'b, 'a, T, V, L>
-	{
+	pub fn terminate(
+		&mut self,
+		beneficiary: &T::AccountId,
+		gas_meter: &mut GasMeter<T>,
+	) -> Result<(), DispatchError> {
+		let self_id = self.self_account.clone();
+		let value = self.overlay.get_balance(&self_id);
+		if let Some(parent) = self.parent {
```
kianenigma: what is the parent of an account here? (totally random question)

frame/contracts/src/wasm/runtime.rs
```diff
@@ -571,6 +586,30 @@ define_env!(Env, <E: Ext>,
...
+	// which is considered fatal and results in a trap + rollback.
+	//
+	// - beneficiary_ptr: a pointer to the address of the beneficiary account where all
+	//   where all remaining funds of the caller are transfered.
+	//   Should be decodable as an `T::AccountId`. Traps otherwise.
+	// - beneficiary_len: length of the address buffer.
+	ext_terminate(
+		ctx,
+		beneficiary_ptr: u32,
+		beneficiary_len: u32
```
kianenigma:
```suggestion
		beneficiary_len: u32,
```

frame/contracts/src/wasm/runtime.rs
```diff
@@ -571,6 +586,30 @@ define_env!(Env, <E: Ext>,
...
+	) => {
+		let beneficiary: <<E as Ext>::T as frame_system::Trait>::AccountId =
+			read_sandbox_memory_as(ctx, beneficiary_ptr, beneficiary_len)?;
+
+		if let Ok(_) = ctx.ext.terminate(&beneficiary, ctx.gas_meter) {
+			ctx.special_trap = Some(SpecialTrap::Termination);
+		}
+		Err(sp_sandbox::HostError)
```
kianenigma: This function always returns an error?

frame/contracts/src/exec.rs
athei: It is the parent `ExecutionContext`. So basically the calling contract. We check if this contract is already on the call stack (re-entrance)  and if so deny the deletion.

frame/contracts/src/wasm/runtime.rs
athei: The macro where this code is in does not support trailing comma.

frame/contracts/src/wasm/runtime.rs
athei: No: It always terminates the execution of the current call (this is what the return of `Err` does). But in the case of a successful `terminate` the `special_trap` is set. This trap signals to the VM that the termination of the call was successful: No revert takes place and success is signaled to the caller.

The debatable reasoning behind that is that `ext_terminate` should never return. Neither on success nor on failure. Returning to a terminated contract feels wrong to me. Returning on failure when we doesn't on success feels like a recipe for disaster. I am willing to be swayed, though.

frame/contracts/COMPLEXITY.md
```diff
@@ -335,6 +335,20 @@ Loading `init_code` and `input_data` should be charged in any case.
...
+
+- `beneficiary`, buffer of a marshaled `AccountId`
+
+It consists of the following steps:
+
+1. Loading `beneficiary` buffer from the sandbox memory (see sandboxing memory get) and then decoding it.
+
+Loading of the `beneficiary` buffer should be charged. This is because the sizes of buffers are specified by the calling code, even though marshaled representations are, essentially, of constant size. This can be fixed by assigning an upper bound for sizes of `AccountId`.
+
+**complexity**: All complexity comes from loading buffers and executing `terminate` executive function. The former component is proportional to the size of the `beneficiary` buffer. The latter component completely depends on the complexity of `terminate` executive function and also dominated by it.
```
pepyakin: Then we need to define this function in the executive section of this file.

frame/contracts/src/wasm/runtime.rs
pepyakin: in hindsight, I think it is the time to actually add support for that feature. There was uncountable number of times when people suggested that, then optionally the author tried to add that or maybe applied the suggestion, and then had to report back and possibly revert the suggestion.

frame/contracts/src/wasm/mod.rs
```diff
@@ -182,6 +182,12 @@ mod tests {
 		gas_left: u64,
 	}
 
+	#[derive(Debug, PartialEq, Eq)]
```
pepyakin: I think this module has grown so that it deserves its own file.

frame/contracts/src/exec.rs
```diff
@@ -510,9 +502,35 @@ where
...
+			if parent.is_live(&self_id) {
+				return Err(DispatchError::Other("Cannot terminate a live contract"));
```
pepyakin: I think the users might be confused by this. 

How about "Cannot terminate a contract that is present on the call stack"?

frame/contracts/src/exec.rs
pepyakin: Might be a good sign to consider to rename this field to `parent_ctx` or smth.

frame/contracts/src/exec.rs
```diff
@@ -510,9 +502,35 @@ where
...
 
-	fn new_call_context<'b>(&'b mut self, caller: T::AccountId, value: BalanceOf<T>)
-		-> CallContext<'b, 'a, T, V, L>
-	{
+	pub fn terminate(
+		&mut self,
+		beneficiary: &T::AccountId,
+		gas_meter: &mut GasMeter<T>,
+	) -> Result<(), DispatchError> {
+		let self_id = self.self_account.clone();
```
pepyakin: Can we get away without this clone?

frame/contracts/src/exec.rs
```diff
@@ -1634,6 +1671,42 @@ mod tests {
...
+				ctx.overlay.set_balance(&ALICE, 1000);
+
+				assert_matches!(
+					ctx.instantiate(
+						100,
+						&mut GasMeter::<Test>::with_limit(10000, 1),
+						&terminate_ch,
+						vec![],
+					),
+					Err(ExecError { reason: DispatchError::Other("insufficient remaining balance"), buffer }) if buffer == Vec::<u8>::new()
```
pepyakin: This line exceeds the line width. Can we wrap it?

frame/contracts/src/exec.rs
athei: #5270

frame/contracts/src/wasm/mod.rs
athei: Not which module you are talking about. You mean the wasm unit tests which are located in the wasm root module?

frame/contracts/src/exec.rs
athei: Your are absolutely right. Nobody knows what "live" means. The function name `is_live` is equally confusing (but this is not the point here).

frame/contracts/src/exec.rs
athei: The problem here is that the `transfer` call borrows `self` mutably with the `ctx` arg. Therefore we cannot borrow `self.self_account` for the `transactor` arg.

In order to safe the clone we need to change how this function words.

frame/contracts/src/exec.rs
athei: Sure.

PR #5236, opened 2020-03-12
Allow changing the behavior for imported blocks

cecton:
This change is needed for cumulus to be able to implement https://github.com/paritytech/cumulus/issues/7

The goal is to allow the user to override the default behavior of substrate when a new block is imported. By default the original behavior is kept unchanged.

The changes are:
 - Add an option to disable the default block announce
 - Add an on_block_imported method on NetworkService that will allow the user to call the method manually

polkadot companion: paritytech/polkadot#955


client/network/src/service.rs
```diff
@@ -665,6 +665,11 @@ impl<B: BlockT + 'static, H: ExHashT> NetworkService<B, H> {
 	pub fn num_connected(&self) -> usize {
 		self.num_connected.load(Ordering::Relaxed)
 	}
+
+	/// You must call this when a new block is imported by the client.
+	pub fn on_block_imported(&self, header: B::Header, data: Vec<u8>, is_best: bool) {
```
bkchr: This should be announce_block

tomaka: It's not just about announcing the block. It also updates the state of `sync` so that it knows what's in the database.

tomaka:
I don't think there's a better way to document `on_block_imported` than "you must call this when a block is now in the database, otherwise things will break".

cecton:
> I don't think there's a better way to document `on_block_imported` than "you must call this when a block is now in the database, otherwise things will break".

Maybe it could be something like: "This method is normally called by substrate unless default_announce_block is set on false in the Configuration, in which case you must call this when a block is now in the database, otherwise things will break"

client/network/src/service.rs
cecton: There is already a method announce block there: https://github.com/paritytech/substrate/blob/61c5d4b6e0339acccfc37f5dbfa15cde0e4183ac/client/network/src/service.rs#L552-L558

bkchr: Yeah I know. For whatever we decide, `default_announce_block` should be named accordingly.

cecton:
I'm starting to wonder if it wouldn't be better to make a Configuration parameter "override_on_block_imported" that would be an `Option<Fn()>` that would be called if provided. And if not provided, fallbacking to the default behavior.

client/network/src/service.rs
cecton: Fair point

bkchr: > There is already a method announce block there:
> 
> https://github.com/paritytech/substrate/blob/61c5d4b6e0339acccfc37f5dbfa15cde0e4183ac/client/network/src/service.rs#L552-L558

I just realized this. This makes the whole story here even more confusing. We just want to announce a block later, but that would not update `sync` and would break everything... If we continue with this new method here, I would propose to remove this `announce_block` function.

@tomaka any ideas?

cecton:
Maybe a good example is better to explain my idea:

```rust
	let mut imported_blocks_stream = client.import_notification_stream().fuse();
	let mut finality_notification_stream = client.finality_notification_stream().fuse();

	futures::future::poll_fn(move |cx| {
		let before_polling = Instant::now();

		// We poll `imported_blocks_stream`.
		while let Poll::Ready(Some(notification)) = Pin::new(&mut imported_blocks_stream).poll_next(cx) {
			let maybe_import =
				if let Some(override_func) = maybe_override {
					override_func(notification)
				} else {
					Some(notification)
				};
			if let Some(notification) = maybe_import {
				network.on_block_imported(notification.header, Vec::new(), notification.is_new_best);
			}
		}
```

tomaka:
If the reason for this PR is to be able to announce a block later, then IMO this is a very wrong approach and we should instead configure the network to not announce a block by default when `on_block_import` is called, rather than not calling the method at all.

client/network/src/service.rs
rphmeier: Removing that function without replacing the functionality _exactly_ will break GRANDPA.

client/network/src/protocol.rs
```diff
@@ -1446,6 +1446,11 @@ impl<B: BlockT, H: ExHashT> Protocol<B, H> {
 			return;
 		}
 
+		// blocks are announced by default
+		if !self.config.default_block_announce {
+			return;
+		}
+
```
cecton: @tomaka something like this?

client/network/src/protocol.rs
```diff
@@ -451,13 +451,16 @@ pub struct ProtocolConfig {
 	pub roles: Roles,
 	/// Maximum number of peers to ask the same blocks in parallel.
 	pub max_parallel_downloads: u32,
+	/// Use default block announcement
+	pub default_announce_block: bool,
```
rphmeier: I agree with Pierre that this feels very hacky and fragile. The documentation is not good either.

rphmeier:
> As mentioned, there is already a function that lets you announce a block.

Right. We should definitely be using that, but come up with a cleaner way to disable the normal "announce new best" logic.

Usually this would be done via a trait.

Could something like this fit?
```rust
trait ImportedBlockAnnounceFilter {
    // Whether to announce a block that has just been imported to peers.
    fn announce_block(&self, some_block_info) -> bool;
}

struct AnnounceNewBestBlocks;
impl ImportedBlockAnnounceFilter for AnnounceNewBestBlocks {
    // ...
}

struct AnnounceNothing;
impl ImportBlockAnnounceFilter for AnnounceNothing {
    // ...
}
```

tomaka:
Giving more thoughts, to me the best solution would simply be to never announce blocks automatically. In other words, not even give option to disable it like this PR does, but always disable it.

Instead have [this code here](https://github.com/paritytech/substrate/blob/7b39077af588a6bcd426ba925c77aa571b7f93f5/client/service/src/lib.rs#L338) also call `announce_block`.

As an alternative we could also change the signature of `on_block_import` to have a boolean `also_announce`.

However this adds a bit more burden to the `sc_service` crate.

bkchr:
Sounds good to me!

cecton:
@tomaka I think I understand theoretically but not how it translates in practice :sweat_smile: can you check this?

client/service/src/lib.rs
```diff
@@ -335,7 +335,8 @@ fn build_network_future<
 
 		// We poll `imported_blocks_stream`.
 		while let Poll::Ready(Some(notification)) = Pin::new(&mut imported_blocks_stream).poll_next(cx) {
-			network.on_block_imported(notification.header, Vec::new(), notification.is_new_best);
+			network.on_block_imported(notification.header, Vec::new(), notification.is_new_best, false);
+			network.service().announce_block(notification.hash, Vec::new());
```
tomaka: Well, you can pass `true` here.
The entire point of passing an extra boolean parameter is that you don't have to call `announce_block` separately.

```suggestion
			network.on_block_imported(notification.header, Vec::new(), notification.is_new_best, true);
```

tomaka:
Although maybe not having a boolean is better?
If you're reading the code of sc_service trying to figure out when blocks are announced, this little extra `true` is easy to miss.

client/service/src/lib.rs
tomaka: Although maybe not having a boolean is better?
If you're reading the code of sc_service trying to figure out when blocks are announced, this little extra `true` is easy to miss.

rphmeier: `enum Announce { Yes, No }`

cecton: I finally get it :sweat_smile: calling the code explicitly is more visible than the boolean

client/network/src/protocol.rs
```diff
@@ -1434,24 +1434,11 @@ impl<B: BlockT, H: ExHashT> Protocol<B, H> {
 		}
 	}
 
-	/// Call this when a block has been imported in the import queue and we should announce it on
-	/// the network.
-	pub fn on_block_imported(&mut self, header: &B::Header, data: Vec<u8>, is_best: bool, also_announce: bool) {
+	/// Call this when a block has been imported in the import queue
+	pub fn on_block_imported(&mut self, header: &B::Header, is_best: bool) {
```
rphmeier: TBH, seems that this function now does nothing unless the new block is best. Maybe it should be `fn on_best_block_imported(&mut self, header: &B::Header)`

tomaka: But then we would need to reintroduce it if later we want to do something for non-best blocks.

I quite dislike that function to be honest, but its API seems to be "please call me when a block has been added to the client", and adding an "only if it is the new best block" clause seems more error-prone.

client/service/src/lib.rs
```diff
@@ -328,6 +328,7 @@ fn build_network_future<
 	status_sinks: Arc<Mutex<status_sinks::StatusSinks<(NetworkStatus<B>, NetworkState)>>>,
 	mut rpc_rx: mpsc::UnboundedReceiver<sc_rpc::system::Request<B>>,
 	should_have_peers: bool,
+	should_announce_imported_blocks: bool,
```
tomaka:
```suggestion
	announce_imported_blocks: bool,
```

client/service/src/lib.rs
```diff
@@ -337,7 +338,11 @@ fn build_network_future<
 
 		// We poll `imported_blocks_stream`.
 		while let Poll::Ready(Some(notification)) = Pin::new(&mut imported_blocks_stream).poll_next(cx) {
-			network.on_block_imported(notification.header, Vec::new(), notification.is_new_best);
+			network.on_block_imported(notification.header, notification.is_new_best);
+
+			if should_announce_imported_blocks {
```
tomaka:
```suggestion
			if announce_imported_blocks {
```

cecton:
> > Add an on_block_imported method on NetworkService that will allow the user to call the method manually
> 
> I think we need example (test) how this `on_block_imported` can be used, this will also ensure that api is actually consumable

@NikVolf I thought about it and I have no idea how this can be tested or how to give a meaningful example at this point because I don't fully understand the product. I'm open to suggestions but I propose to merge as it for now because it's technically tested (tests fail if the blocks are not announced right after being imported). When the feature in cumulus will be fully developed (this might take time) maybe I will find some idea.

bkchr:
Please merge master.

bkchr:
And also requires a polkadot pr.

cecton:
@bkchr done. Could you click the merge button please?

PR #5237, opened 2020-03-12
client/finality-grandpa: Add Prometheus metrics to GossipValidator

mxinden:
Instrument finality grandpa `GossipValidator` exposing count of messages
validated by message type and message action.

```
\# HELP substrate_finality_grandpa_communication_gossip_validator_messages Number of messages validated by the finality grandpa gossip validator.
\# TYPE substrate_finality_grandpa_communication_gossip_validator_messages counter
substrate_finality_grandpa_communication_gossip_validator_messages{action="discard",message="neighbor"} 39
substrate_finality_grandpa_communication_gossip_validator_messages{action="keep",message="vote"} 28
```

### DIFF
--- a/client/finality-grandpa/src/communication/gossip.rs
+++ b/client/finality-grandpa/src/communication/gossip.rs
@@ -91,6 +91,7 @@ use sp_finality_grandpa::AuthorityId;
 use sc_telemetry::{telemetry, CONSENSUS_DEBUG};
 use log::{trace, debug};
 use futures::channel::mpsc;
+use prometheus_endpoint::{CounterVec, Opts, PrometheusError, register, Registry, U64};
 use rand::seq::SliceRandom;
 
 use crate::{environment, CatchUp, CompactCommit, SignedMessage};
@@ -1197,11 +1198,34 @@ impl<Block: BlockT> Inner<Block> {
 	}
 }
 
+// Prometheus metrics for [`GossipValidator`].
+pub(crate) struct Metrics {
+	messages_validated: CounterVec<U64>,
+}
+
+impl Metrics {
+	pub(crate) fn register(registry: &prometheus_endpoint::Registry) -> Result<Self, PrometheusError> {
+		Ok(Self {
+			messages_validated: register(
+				CounterVec::new(
+					Opts::new(
+						"finality_grandpa_communication_gossip_validator_messages",
+						"Number of messages validated by the finality grandpa gossip validator."
+					),
+					&["message", "action"]
+				)?,
+				registry,
+			)?,
+		})
+	}
+}
+
 /// A validator for GRANDPA gossip messages.
 pub(super) struct GossipValidator<Block: BlockT> {
 	inner: parking_lot::RwLock<Inner<Block>>,
 	set_state: environment::SharedVoterSetState<Block>,
 	report_sender: mpsc::UnboundedSender<PeerReport>,
+	metrics: Option<Metrics>,
 }
 
 impl<Block: BlockT> GossipValidator<Block> {
@@ -1211,12 +1235,27 @@ impl<Block: BlockT> GossipValidator<Block> {
 	pub(super) fn new(
 		config: crate::Config,
 		set_state: environment::SharedVoterSetState<Block>,
+		prometheus_registry: Option<&Registry>,
 	) -> (GossipValidator<Block>, mpsc::UnboundedReceiver<PeerReport>)	{
+		let metrics = match prometheus_registry {
+			Some(registry) => {
+				match Metrics::register(registry) {
+					Ok(metrics) => Some(metrics),
+					Err(e) => {
+						debug!(target: "afg", "Failed to register metrics: {:?}", e);
+						None
+					}
+				}
+			},
+			None => None,
+		};
+
 		let (tx, rx) = mpsc::unbounded();
 		let val = GossipValidator {
 			inner: parking_lot::RwLock::new(Inner::new(config)),
 			set_state,
 			report_sender: tx,
+			metrics: metrics,
 		};
 
 		(val, rx)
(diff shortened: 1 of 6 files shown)


client/finality-grandpa/src/communication/gossip.rs
```diff
@@ -1211,12 +1235,27 @@ impl<Block: BlockT> GossipValidator<Block> {
 	pub(super) fn new(
 		config: crate::Config,
 		set_state: environment::SharedVoterSetState<Block>,
+		prometheus_registry: Option<&Registry>,
 	) -> (GossipValidator<Block>, mpsc::UnboundedReceiver<PeerReport>)	{
+		let metrics = match prometheus_registry {
```
bkchr:
```suggestion
		let metrics = match prometheus_registry.map(Metrics::register) {
```

Could reduce the code a little bit.

client/finality-grandpa/src/communication/gossip.rs
```diff
@@ -1320,6 +1372,16 @@ impl<Block: BlockT> GossipValidator<Block> {
...
 		};
 
+		// Prometheus metric recording.
+		if let (Some(metrics), Some(message_name)) = (&self.metrics, message_name) {
+			let action_name = match action {
+				Action::Keep(_, _) => "keep",
+				Action::ProcessAndDiscard(_, _) => "process_and_discard",
+				Action::Discard(_) => "discard",
+			};
+			metrics.messages_validated.with_label_values(&[message_name, action_name]).inc();
```
bkchr: I hope that this does not introduces performance problems?

mxinden: Valid concern. I am curious myself, so let's dig into this a bit.

Within the benchmark suit of the `prometheus` crate (we are using version `v0.7.0`) there is the [`bench_counter_with_label_values`](https://github.com/tikv/rust-prometheus/blob/v0.7.0/benches/counter.rs#L24) test which mirrors our use case here the best. This test increases a `CounterVec` with 3 labels in a tight loop. Each increase is one iteration.

On my CPU (`i7-8550U`) an iteration takes ~ `100ns` which one could interpret as: **Increasing a Prometheus counter with 3 labels takes around `100ns`**. (Given that the benchmark above runs in a tight loop on a single thread, one could argue that it does not mirror the amount of CPU cache invalidation and cache coherency traffic that we would see in  a real-world scenario.)

The `GosssipValidator::do_validate` method is called on each incoming gossip message. One could assume that each gossip message translates to an IP packet that needs to be received. Receiving (not the latency on the wire, but only the time within the machine) an IP packet would roughly be around `0.025 ms`. Ignoring all additional steps involved, which would probably each exceed the `0.025 ms`, e.g. transport layer decryption, SCALE decoding, ... one could say that: **The lower bound of the `do_validate` function is at least `0.025 ms`**.

Comparing the two assumptions: **Increasing the counter would make up `0.4 %` (`100 ns / 0.025 ms`) of the overall `do_validate` execution time.** I would say this is an overhead worth ignoring / paying.

Does this sound reasonable to you @bkchr?

bkchr: You had me with `100ns` :D

client/finality-grandpa/src/communication/gossip.rs
```diff
@@ -1279,12 +1318,21 @@ impl<Block: BlockT> GossipValidator<Block> {
 		let mut broadcast_topics = Vec::new();
 		let mut peer_reply = None;
 
+		// Message name for Prometheus metric recording.
+		let message_name;
+
 		let action = {
```
expenses: I think this would be better like 
```suggestion
		let (action, message_name) = {
```

mxinden: Are you suggesting the following?

```diff
diff --git a/client/finality-grandpa/src/communication/gossip.rs b/client/finality-grandpa/src/communication/gossip.rs
index 89c507f39..8751f4c9b 100644
--- a/client/finality-grandpa/src/communication/gossip.rs
+++ b/client/finality-grandpa/src/communication/gossip.rs
@@ -1314,21 +1314,13 @@ impl<Block: BlockT> GossipValidator<Block> {
                let mut broadcast_topics = Vec::new();
                let mut peer_reply = None;
 
-               // Message name for Prometheus metric recording.
-               let message_name;
-
-               let action = {
+               let (action, message_name) = {
                        match GossipMessage::<Block>::decode(&mut data) {
-                               Ok(GossipMessage::Vote(ref message)) => {
-                                       message_name = Some("vote");
-                                       self.inner.write().validate_round_message(who, message)
-                               },
-                               Ok(GossipMessage::Commit(ref message)) => {
-                                       message_name = Some("commit");
-                                       self.inner.write().validate_commit_message(who, message)
-                               },
+                               Ok(GossipMessage::Vote(ref message)) =>
+                                       (self.inner.write().validate_round_message(who, message), Some("vote")),
+                               Ok(GossipMessage::Commit(ref message)) =>
+                                       (self.inner.write().validate_commit_message(who, message), Some("commit")),
                                Ok(GossipMessage::Neighbor(update)) => {
-                                       message_name = Some("neighbor");
                                        let (topics, action, catch_up, report) = self.inner.write().import_neighbor_message(
                                                who,
                                                update.into_neighbor_packet(),
@@ -1340,14 +1332,12 @@ impl<Block: BlockT> GossipValidator<Block> {
 
                                        broadcast_topics = topics;
                                        peer_reply = catch_up;
-                                       action
+                                       (action, Some("neighbor"))
                                }
                                Ok(GossipMessage::CatchUp(ref message)) => {
-                                       message_name = Some("catch_up");
-                                       self.inner.write().validate_catch_up_message(who, message)
+                                       (self.inner.write().validate_catch_up_message(who, message), Some("catch_up"))
                                },
                                Ok(GossipMessage::CatchUpRequest(request)) => {
-                                       message_name = Some("catch_up_request");
                                        let (reply, action) = self.inner.write().handle_catch_up_request(
                                                who,
                                                request,
@@ -1355,15 +1345,14 @@ impl<Block: BlockT> GossipValidator<Block> {
                                        );
 
                                        peer_reply = reply;
-                                       action
+                                       (action, Some("catch_up_request"))
                                }
                                Err(e) => {
-                                       message_name = None;
                                        debug!(target: "afg", "Error decoding message: {}", e.what());
                                        telemetry!(CONSENSUS_DEBUG; "afg.err_decoding_msg"; "" => "");
 
                                        let len = std::cmp::min(i32::max_value() as usize, data.len()) as i32;
-                                       Action::Discard(Misbehavior::UndecodablePacket(len).cost())
+                                       (Action::Discard(Misbehavior::UndecodablePacket(len).cost()), None)
                                }
                        }
                };
```

I would not prefer this code style.

1. I see `action` as actual business logic whereas `message_name` a meta logic, combining the two in one tuple removes the differentiation. 

2. Just by looking at a line like `(action, Some("catch_up_request"))` the reader can not imply what the second element of the tuple is.

Do you feel strongly about this @expenses?

expenses: No, I don't.

Issue #4362, opened 2019-12-11
decl_storage: storage the key alongside the value to be able to iterate on key/value


thiolliere:
as follow up of https://github.com/paritytech/substrate/pull/4185#issuecomment-562122935

It would be interesting to introduce a way in decl_storage to tell to store the key alongside the value.
at the end the storage will have the data: `(value, key)`.
This way when iterating on prefix we could actually iterate on `(key, value)` instead of just `value`.

This could be just a new attribute.

Implementing the StorageMap and other trait is not straightforward, here are some difficultyies:
* StorageMap::append is implementing to work on the raw not decoded value. If the raw value is `(value, key)` we can't call scale_codec::EncodeAppend::append on it because we need to give only the raw data for the value.
  multiple solution here:
  * store actually `(key, value)`, decode the key, the remaining byte are the raw value.
     not very good because get function will also have to decode the key.
  * store `(value, key, key_length)`, so we can use key_length to split the raw data.
    (key_length is encoded on 4 byte or in a compact encoding but reversed so we can decode from the end)
  * implement a new EncodeAppendPreserving which does append on a raw data while preserving the suffix.
  * store `(value, key)` and just decode the value, thus there won't be any optimisation anymore, it will be same as using `mutate`.

* StorageMap::swap: we can't swap the raw data anymore, solutions:
  * if we store `(key, value)` then we have to decode the key only and then we can swap split and swap the raw values
  * if we store `(value, key)` we have to decode the value and then we can split and swap the raw value
  * if we store `(value, key, key_length)` then we decode key_length and then we can split and swap the raw value.

bkchr:
> for encode_append, we can't append on the raw data anymore as encode_append support only having the data as input.
we could extend encode_append to be able to append on some value while keeping the data in suffix.

I don't understand this. Could you explain it better? Why can we not append `(key, value)`?

thiolliere:
indeed I updated the message for more clarity, the initial idea was to store `(value, key)` so we don't have to decode the key on operation such as `get` and `mutate`.
If we store indeed `(key, value)` then append operation can work easily just decoding key and working the raw value.

bkchr:
Ahh, yeah, now I understand :) We could maybe thing of a `skip()` method for skipping data in the encoded blob. So, we could use `(key, data)`, skip the `key` and use `data` to append the stuff. Or we just prepend the length of the key.

thiolliere:
I think this should be close as we are now storing keys in hash

PR #5238, opened 2020-03-12
Add iteration with key information when using $hash_concat

thiolliere:
Fix https://github.com/paritytech/substrate/issues/4362

When iterating on a storage_map which use twox_64_concat we must be able to get the key alongside the value. This PR create a new function on StoragePrefixedMap to make it possible.

### In future PR:

DoubleMap::iter_prefix (which iter on all the second key associated to one first key) can't iterate with key information yet. I will do in a following PR

### Additionnal notes

The fact the the iteration stop immediately if a value fail to decode was made to follow the same behavior of StoragPrefixedMap::iter(). Though maybe we should move both implementation to just ignore failed to decode value.

### Alternative implementation

If having () instead of key when the hasher doesn't give information is too confusing we could also go in another direction by making such function available with a `where` statement.

Though to do so I would actually prefer merging StorageMap and generator::StorageMap and having the function `iter_with_info()... where Hasher: HasherWithInfo`.

### DIFF
--- a/frame/support/procedural/src/storage/storage_struct.rs
+++ b/frame/support/procedural/src/storage/storage_struct.rs
@@ -121,10 +121,25 @@ pub fn decl_and_impl(scrate: &TokenStream, def: &DeclStorageDefExt) -> TokenStre
 			},
 			StorageLineTypeDef::Map(map) => {
 				let hasher = map.hasher.to_storage_hasher_struct();
+				let key_type = &map.key;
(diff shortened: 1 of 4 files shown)


gavofyork:
This is already implemented in my PR.

thiolliere:
partially implemented in #5226
and superseeded by https://github.com/paritytech/substrate/pull/5335

PR #5245, opened 2020-03-13
Remove `extern crate proc_macro`

kianenigma:
as it looks like it is no longer needed https://blog.rust-lang.org/2020/03/12/Rust-1.42.html#use-proc_macro:tokenstream;-now-works

and some unused imports.


kianenigma:
(Probably ci is using the old version)

kianenigma:
(ci is now using the new version)

PR #5247, opened 2020-03-13
Don't lock if we're not voting with conviction

gavofyork:


frame/democracy/src/lib.rs
```diff
@@ -2853,6 +2855,28 @@ mod tests {
...
 
+	#[test]
+	fn no_locks_without_conviction_should_work() {
+		new_test_ext().execute_with(|| {
+			System::set_block_number(0);
+			let r = Democracy::inject_referendum(
+				2,
+				set_balance_proposal_hash_and_note(2),
+				VoteThreshold::SuperMajorityApprove,
+				0
```
kianenigma:
```suggestion
				0,
```

PR #5249, opened 2020-03-13
Task executor extension for runtime host

NikVolf:
This will essentially allow spawning tasks when handling host functions, using the main scheduler.

It is immediately needed for #5023, also for variant of #5023 for block production

And is generally useful in a sense that it will allow parallelise runtime execution (but should be used carefully and join spawned tasks deterministically - one-by-one / or in well-defined order if multiple are joined)

- [x] example: see #5023


primitives/externalities/src/lib.rs
```diff
@@ -173,6 +173,11 @@ pub trait Externalities: ExtensionStore {
 	/// Returns the SCALE encoded hash.
 	fn storage_changes_root(&mut self, parent: &[u8]) -> Result<Option<Vec<u8>>, ()>;
 
+	/// Return handle that can spawn parallel tasks.
+	///
+	/// Can be none if host does not support parallel tasks.
+	fn spawn_handle(&self) -> Option<&dyn ClonableSpawn>;
```
bkchr: I really don't want to add anymore functionality to `Externalities`. This was the reason `Extension`'s where introduced. I know that extension handling is currently a bit shitty, but please make this an extension.

NikVolf: Fine with me, but just to note and as discussed: this `Externalities` ability will be quite essential (modulo test implementations of it), while `Extension`-s are optional and generally not required for code execution.

kianenigma:
CC @tomusdrw ^^

primitives/externalities/src/lib.rs
NikVolf: Ok after some more disussion with @kianenigma there is a strong indication that this functionality will need to be rather configurable, and Extension will definitely help to achieve configurability much in much more convenient way.

kianenigma: Mot notably, if we can configure the child threads (or whatever we call them in this context) to not be able to write to storage and instead return the buffer changes, this will give us a big chunk of what we need for transaction parallelism.

NikVolf: @bkchr updated!

NikVolf:
I will squash / force push this pr, because it is used in another pr, so sorry about that

primitives/core/src/traits.rs
```diff
@@ -212,3 +212,21 @@ impl CallInWasmExt {
 		Self(Box::new(inner))
 	}
 }
+
+/// Something that can spawn tasks and also can be cloned.
+pub trait ClonableSpawn: futures::task::Spawn + Send + Sync {
```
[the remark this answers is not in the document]
NikVolf: Updated!

primitives/core/src/traits.rs
```diff
@@ -212,3 +212,21 @@ impl CallInWasmExt {
 		Self(Box::new(inner))
 	}
 }
+
+/// Something that can spawn tasks and also can be cloned.
+pub trait CloneableSpawn: futures::task::Spawn + Send + Sync {
```
tomusdrw: I'm not a fan of traits that just describe the expected programatic behavior instead of some more high-level code idea.
Also this causes `Box<dyn CloneableSpawn>` to be quite widespread.

Can't we have a `struct` that will just wrap this? Like so:
```rust
#[derive(Clone)]
struct TaskExecutor(Arc<dyn futures::Task::Spawn + Send + Sync>);
impl TaskExecutor {
  pub fn spawn(...) {}
}
```
We don't even need this extra trait cause it's in `Arc` and I feel it avoids much of code repetition.

primitives/core/src/lib.rs
```diff
@@ -70,6 +70,8 @@ mod changes_trie;
 #[cfg(feature = "std")]
 pub mod traits;
 pub mod testing;
+#[cfg(feature = "std")]
+pub mod tasks;
```
tomusdrw: I'm not 100% sure this fits into `sp-core`, but at the same time we don't really have a better place currently, so I'm fine with this (unless someone has a better idea where to put it).

primitives/core/src/traits.rs
NikVolf: I've considered something like that. The problem is we won't be able to have special implementation of clone in task executor this way. It might require little more than just shared reference clone in use cases I think about.

client/service/src/task_manager.rs
```diff
@@ -188,3 +195,9 @@ impl Drop for TaskManager {
 		}
 	}
 }
+
+#[test]
+fn can_make_dyn_clonable_spawn() {
```
bkchr: Why does this test not lives at the point where we declare the trait?

primitives/core/Cargo.toml
```diff
@@ -102,7 +103,9 @@ std = [
 	"sp-externalities",
 	"sp-storage/std",
 	"sp-runtime-interface/std",
-	"zeroize/alloc"
+	"zeroize/alloc",
+	"futures",
+	"futures/thread-pool"
```
bkchr:
```suggestion
	"futures/thread-pool",
```

primitives/core/src/tasks.rs
```diff
@@ -0,0 +1,56 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Module for low-level asynchronous processing.
```
bkchr: Isn't this just providing a single threaded executor?

primitives/core/src/lib.rs
bkchr: We probably could put everything tasks related (Also the `CloneableSpawn` trait) in its own crate, but the current amount of code isn't also that much.

primitives/state-machine/Cargo.toml
```diff
@@ -22,6 +22,7 @@ codec = { package = "parity-scale-codec", version = "1.2.0" }
 num-traits = "0.2.8"
 rand = "0.7.2"
 sp-externalities = { version = "0.8.0-alpha.2", path = "../externalities" }
+futures = "0.3.1"
```
bkchr: As far as I can see, the dependency is only used in tests? 
And you only need it to create a dummy executor, can we not use the single threaded executor you provide in `sp-core`?

primitives/externalities/Cargo.toml
```diff
@@ -13,3 +13,4 @@ documentation = "https://docs.rs/sp-externalities"
 sp-storage = { version = "2.0.0-alpha.2", path = "../storage" }
 sp-std = { version = "2.0.0-alpha.2", path = "../std" }
 environmental = { version = "1.1.1" }
+futures = { version = "0.3.1" }
```
bkchr:
```suggestion
```

client/service/src/task_manager.rs
NikVolf: It strayed...

primitives/core/src/tasks.rs
NikVolf: Well, at the moment, yes

PR #5250, opened 2020-03-13
Add statistics to benchmarking-cli output

pscott:
The goal of this PR is to add some useful information to the benchmarking-cli output.

- [x] Accurately track db reads / writes
- [x] Send the statistics to the benchmarking-cli and output it in CSV format

Before merging:
- [x] Remove the code that exposes the benchmarking pallet so that it doesn't get added to master.

In order to pass the `UsageInfo` data from `/frame/benchmarking` crate to the `/utils/frame/benchmarking-cli` crate, and use the `UsageInfo` struct in `sp-state-machine` and `sp-externalities`, I had to create a new primitive called `sp-stats` for this very purpose. Let me know if there's any other way of doing so.

I also merged in [Cheme's PR](https://github.com/paritytech/substrate/pull/5168/) , and then reverted the merge using `git revert -m 1 COMMIT`. Hope this is the proper way to do it.


client/db/src/bench.rs
```diff
@@ -132,7 +135,13 @@ impl<B: BlockT> StateBackend<HashFor<B>> for BenchmarkingState<B> {
 	type TrieBackendStorage = <DbState<B> as StateBackend<HashFor<B>>>::TrieBackendStorage;
 
 	fn storage(&self, key: &[u8]) -> Result<Option<Vec<u8>>, Self::Error> {
-		self.state.borrow().as_ref().ok_or_else(state_err)?.storage(key)
+		let state = self.state.borrow_mut();
+		let db_state = state.as_ref().ok_or_else(state_err)?;
+		let _ = self.usage_info.try_borrow_mut().map(|mut usage_info| {
+			usage_info.reads.ops += 1;
+			usage_info.reads.bytes += key.len() as u64;
```
pscott: This is NOT good, and I would appreciate some help on how to properly track those reads

client/db/src/bench.rs
```diff
@@ -132,7 +135,13 @@ impl<B: BlockT> StateBackend<HashFor<B>> for BenchmarkingState<B> {
 	type TrieBackendStorage = <DbState<B> as StateBackend<HashFor<B>>>::TrieBackendStorage;
 
 	fn storage(&self, key: &[u8]) -> Result<Option<Vec<u8>>, Self::Error> {
-		self.state.borrow().as_ref().ok_or_else(state_err)?.storage(key)
+		let state = self.state.borrow_mut();
+		let db_state = state.as_ref().ok_or_else(state_err)?;
+		let _ = self.usage_info.try_borrow_mut().map(|mut usage_info| {
```
pscott: Anti pattern... I would like to try_borrow_mut, and if it fails simply "silence" it out, as the usage_info is not primordial

client/db/src/bench.rs
```diff
@@ -254,15 +263,21 @@ impl<B: BlockT> StateBackend<HashFor<B>> for BenchmarkingState<B> {
...
 					db_transaction.put(0, &key, &val);
 				} else if rc < 0 {
 					db_transaction.delete(0, &key);
 				}
+				bytes += key.len() as u64;
 			}
 			db.write(db_transaction).map_err(|_| String::from("Error committing transaction"))?;
+			let _ = self.usage_info.try_borrow_mut().map(|mut usage_info| {
+				usage_info.writes.ops += 1;
+				usage_info.writes.bytes += bytes;
```
pscott: Inaccurate tracking of writes hear, would appreciate some feedback on how to track those properly

client/db/src/bench.rs
```diff
@@ -254,15 +263,21 @@ impl<B: BlockT> StateBackend<HashFor<B>> for BenchmarkingState<B> {
...
 			for (key, (val, rc)) in transaction.drain() {
 				if rc > 0 {
 					db_transaction.put(0, &key, &val);
 				} else if rc < 0 {
 					db_transaction.delete(0, &key);
 				}
+				bytes += key.len() as u64;
 			}
 			db.write(db_transaction).map_err(|_| String::from("Error committing transaction"))?;
+			let _ = self.usage_info.try_borrow_mut().map(|mut usage_info| {
```
pscott: Anti-pattern just like the one above

client/db/src/storage_cache.rs
```diff
@@ -852,7 +861,11 @@ impl<S: StateBackend<HashFor<B>>, B: BlockT> StateBackend<HashFor<B>> for Syncin
 			.as_trie_backend()
 	}
 
-	fn usage_info(&self) -> sp_state_machine::UsageInfo {
+	fn register_overlay_stats(&mut self, stats: &sp_stats::StateMachineStats) {
```
NikVolf: the signature/name of this function must somehow ensure that it consumes stats (i.e. we never add something twice

pscott: Yes so as explained in the PR, I've merged Cheme's WIP PR. I think it adds a lot of confusion to my PR, but it kind of touches the same areas of the code...
I think I can safely revert this merge. Will simply running `git revert -m 1 COMMIT` work, even if we merge his PR later on?

primitives/runtime-interface/src/impls.rs
```diff
@@ -532,3 +532,7 @@ impl PassBy for sp_wasm_interface::ValueType {
 impl PassBy for sp_wasm_interface::Value {
 	type PassBy = Codec<sp_wasm_interface::Value>;
 }
+
+impl PassBy for sp_stats::UsageInfo {
```
bkchr: That is wrong. `sp-stats` should include `runtime-interface` and use the appropriate derive macro.

pscott: This was what I did in the first place, but then I had some cyclic dependencies.
The problem is `sp-externalities` needs to use `UsageInfo`. But if I add `sp-stats` to `sp-externalities` we get a cyclic dependency (sp-externalities -> sp-stats -> sp-runtime-interface -> sp-externalities).

client/db/src/storage_cache.rs
pscott: I reverted the merge

client/db/src/bench.rs
bkchr: You have single threaded code and you can just use `borrow_mut()`. If that fails, it means your code is broken and somewhere keeps a mutable reference. Please remove the `try_borrow_mut`

client/db/src/bench.rs
bkchr: What you mean? What kind of help do you need?

primitives/state-machine/src/basic.rs
```diff
@@ -273,7 +274,13 @@ impl Externalities for BasicExternalities {
 		Ok(None)
 	}
 
-	fn wipe(&mut self) {}
+	fn wipe(&mut self) {
+		unimplemented!()
```
bkchr: Why this change?

primitives/stats/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Usage statistics for subtrate.
```
bkchr:
```suggestion
//! Usage statistics for Substrate.
```

bkchr: Please add more info about the "usage statistics" Nobody knows what this means.

primitives/stats/src/usage_info.rs
```diff
@@ -37,11 +38,16 @@ pub struct UsageInfo {
 	/// Cache read statistics.
 	pub cache_reads: UsageUnit,
 	/// Memory used.
-	pub memory: usize,
+	// Encoded as u32 because wasm's usize is u32.
+	pub memory: u32,
```
bkchr: Please use `u64` and update the comment.

client/db/src/bench.rs
pscott: Is the number of bytes read exactly equivalent to `key.len()`?
Is this counting exhaustive? (i.e. are there other places in the code where we should be incrementing those counters?)

Basically I just place this as a placeholder hoping that someone more knowledgeable than me would provide some guidelines as to how / where I could measure those metrics :)

bkchr: You need to add there the number of bytes of the value to the `reads.bytes`. You are not reading the key here.

pscott: Yeah I was kind of copying some code that I found elsewhere: [lib.rs](https://github.com/paritytech/substrate/blob/c50faf2395218e644859611d703d9fe3a4876f5b/client/db/src/lib.rs#L1127), [stats.rs](https://github.com/paritytech/substrate/blob/c50faf2395218e644859611d703d9fe3a4876f5b/client/db/src/stats.rs#L58).

From what I understand I can't really tell how many bytes I will read from this function?

pscott: Maybe @NikVolf could help here? :D

primitives/state-machine/src/basic.rs
pscott: My bad :)

client/db/src/bench.rs
NikVolf: What you copied from lib.rs is a deletion operation, and we track key length for deletion operation.

For read operation it is value length that should be counted, not key length (for example, what you can observe in `tally_key_read`)

pscott: Thanks for your help! I've added a commit that does just that! :)

pscott:
This PR is ready for review (and merge). I've removed the code that exposed the Benchmark pallet. @bkchr @NikVolf  :D

NikVolf:
can you rebase without any @cheme code in the diff?

client/db/src/stats.rs
```diff
@@ -77,7 +78,7 @@ impl StateUsageStats {
 	}
 
 	/// Merge state machine usage info.
-	pub fn merge_sm(&self, info: sp_state_machine::UsageInfo) {
```
NikVolf: is this something you changed @pscott 
or it was part of @cheme merged pr?

NikVolf:
Can we have at least simple test which shows that stats are being collected?

pscott:
Closed in favor of #5483 :)

PR #5251, opened 2020-03-14
Fork flaming fir

ddorgan:
Fork flaming fir after latest master upgrade.


ddorgan:
This is the spec used in current deployments so it would be good to merge.

PR #5252, opened 2020-03-14
Create a Benchmark Chainspec

shawntabrizi:
This PR adds a new `--chain benchmark` which is the same as the `--chain dev` chain specification, but additionally adds 50,000 funded users.

This also critically fixes the `wipe_db` function to reset the database to the genesis state.

We can continue to expand this chain spec to increase the complexity of of the genesis state of the benchmarks.

I have also included a `benchmark.json`, which is a pre-calculated version of the Benchmark genesis, and contains a Wasm runtime which includes the benchmarking runtime APIs.

### DIFF
--- a/frame/balances/src/benchmarking.rs
+++ b/frame/balances/src/benchmarking.rs
@@ -116,15 +116,4 @@ benchmarks! {
 		let balance_amount = existential_deposit.saturating_mul(e.into());
 		let _ = <Balances<T> as Currency<_>>::make_free_balance_be(&user, balance_amount);
 	}: set_balance(RawOrigin::Root, user_lookup, 0.into(), 0.into())
-
-	confirm_genesis {
-		let x in 1 .. 1;
-		let mut iter = frame_system::Account::<T>::iter();
-		let mut user_count = 0;
-		while let Some(_) = iter.next() {
-			user_count += 1;
-		}
-	}: {
-		assert!(user_count >= 1);
-	}
 }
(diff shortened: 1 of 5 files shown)


frame/balances/src/benchmarking.rs
```diff
@@ -116,4 +116,15 @@ benchmarks! {
...
-}
\ No newline at end of file
+
+	confirm_genesis {
+		let x in 1 .. 1;
+		let mut iter = frame_system::Account::<T>::iter();
+		let mut user_count = 0;
+		while let Some(_) = iter.next() {
+			user_count += 1;
+		}
```
gavofyork:
```suggestion
		let user_count = frame_system::Account::<T>::iter().count();
```

shawntabrizi: I've removed this test function from the benchmarks, but noted 👍 

ty

shawntabrizi:
Thinking about another approach with a script that generates a genesis. Will report back

shawntabrizi:
I was thinking just to load it with `--chain path/to/file`, but with my new approach, I think I might not including it at all. It would need to be updated constantly anyway, since the runtime Wasm will keep changing.

shawntabrizi:
I think https://github.com/shawntabrizi/substrate-benchmark-genesis/ will be a better approach than creating a new `--chain benchmark`.

Through the genesis config available in the CLI, we do not have enough flexibility to make it so that all pallets are correctly saturated with values since not all of them expose genesis configurations. Instead, using a secondary tool to generate random storage keys under the correct module/storage prefix seems better. I will create a new PR which just extracts the fix to the genesis state for benchmarking in general.

Issue #4876, opened 2020-02-10
Benchmarking for Pallets


marcio-diaz:
Issue to track progress on benchmarking dispatch calls.

In order of importance from Gav

- [x] Identity. @shawntabrizi 
- [x] Balances. @shawntabrizi 
- [x] Timestamp. @marcio-diaz  (#4891)
- [x] Polkadot/Claims. @shawntabrizi https://github.com/paritytech/polkadot/pull/876
- [x] Vesting. @marcio-diaz (#5048)
- [x] Im Online. @shawntabrizi 
- [x] Staking @shawntabrizi https://github.com/paritytech/substrate/pull/5183
- [x] Session @shawntabrizi https://github.com/paritytech/substrate/pull/5183
- [x] Elections Phragmen @kianenigma 
- [x] Democracy. @marcio-diaz (https://github.com/paritytech/substrate/pull/5257)
- [x] Collective. @marcio-diaz (https://github.com/paritytech/substrate/pull/5343)
- [x] Offences.@marcio-diaz (review #5411)
- [x] Treasury @shawntabrizi https://github.com/paritytech/substrate/pull/5287
- [x] Utility @shawntabrizi https://github.com/paritytech/substrate/pull/5384

shawntabrizi:
Updated with order of importance from gav

apopiak:
It seems `system` is missing from this?

PR #5257, opened 2020-03-16
Benchmark Democracy Pallet

marcio-diaz:
Part of #4876

It seems that some of the calls (like `veto_external`) can't be benchmarked without modification because they should be dispatched from other modules (like `collective` module) due to `origin` restrictions.

To solve this problem I could modify the code and guard it but it may be too dangerous.


marcio-diaz:
> Hm, github doesn't let me put in inline comments...
> 
> So a general comment, I would think that we should use the origins as designed and not cheat by using root. I think some of these origin check are doing some storage lookups that could influence the benchmark results.

Yeah, I agree. Do you have any idea on how to benchmark these calls?

bkchr:
Can you not construct the state in such a way that these origins are correct?

marcio-diaz:
> Can you not construct the state in such a way that these origins are correct?

The problem is that `run_benchmark` is called on `Democracy`, which requires origin from `frame_system`. So it doesn't compile if I use a `collective::RawOrigin`. If I use `system::Origin` it compiles but  `into()` returns an error as defined in the macro at `support/origin.rs`.

My best idea is to define a conversion between both types and guard it, by modifying that macro. But it sounds a bit dangerous as I said  in the description.

bkchr:
What is the error you get?

marcio-diaz:
> What is the error you get?

The error is mapped, so it says only `BadOrigin`, should be something like `Err(system_origin)`. It's coming from here https://github.com/paritytech/substrate/blob/master/frame/support/src/origin.rs#L215

bkchr:
Stupid question, can we not call the method like it is done here: https://github.com/paritytech/substrate/blob/master/frame/collective/src/lib.rs#L362-L363?

marcio-diaz:
> Stupid question, can we not call the method like it is done here: https://github.com/paritytech/substrate/blob/master/frame/collective/src/lib.rs#L362-L363?

My guess is that it's possible but from other module, something like `TechnicalCommittee::run_benchmarks()`. I'll check.

marcio-diaz:
> Stupid question, can we not call the method like it is done here: https://github.com/paritytech/substrate/blob/master/frame/collective/src/lib.rs#L362-L363?

I was missing some bounds as you said (thanks!!!), it's fixed now.

marcio-diaz:
Update: Gav suggested to use `Origin::Root`, it remains to see how to bench `fn veto_external`, because of `who` in `let who = T::VetoOrigin::ensure_origin(origin)?;`.

frame/benchmarking/Cargo.toml
```diff
@@ -17,6 +17,8 @@ sp-std = { version = "2.0.0-alpha.2", path = "../../primitives/std", default-fea
 sp-io = { path = "../../primitives/io", default-features = false, version = "2.0.0-alpha.2" }
 frame-support = { version = "2.0.0-alpha.2", default-features = false, path = "../support" }
 frame-system = { version = "2.0.0-alpha.2", default-features = false, path = "../system" }
+pallet-collective = { version = "2.0.0-alpha.2", default-features = false, path = "../collective" }
+node-primitives = { version = "2.0.0-alpha.2", default-features = false, path = "../../bin/node/primitives" }
```
thiolliere: I don't think having a pallet and a node crate as dependency of frame-benchmarking is suitable this is contradictory to their seperation IMO.

frame/democracy/src/benchmarking.rs
```diff
@@ -0,0 +1,383 @@
...
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Democracy pallet benchmarking.
+
+use super::*;
+
+use frame_benchmarking::{benchmarks, account};
+use frame_support::traits::{Currency, Get};
+use frame_system::{RawOrigin, Module as System, self};
+use pallet_collective::{RawOrigin as CollectiveOrigin, Instance1, Instance2};
```
thiolliere: and benchmark should be related to the node benchmarked not to our node in substrate.
Thus any Origin can be used here, not necessary collective.

I didn't dive into implementation but for me we should bound new trait on our origins associeted types like this:
```rust
	type CancellationOrigin: EnsureOrigin<Self::Origin> + BenchmarkOriginProvider<Self::Origin>;
```
(or directly into EnsureOrigin traits.)
```rust
trait BenchmarkOriginProvider {
  fn random_successful_origin() -> OriginKind;
  // And maybe this one but i'm not sure it is needed.
  fn successful_origin_from_success(success: EnsureOrigin::Success) -> OriginKind;
}
```

marcio-diaz: Yeah, I was thinking something like that, using collective is awful and `Root` doesn't work always. :+1:

thiolliere: or maybe something like what is done in https://github.com/paritytech/substrate/pull/5287 for the trait Contains

bin/node/runtime/Cargo.toml
```diff
@@ -146,5 +147,8 @@ runtime-benchmarks = [
 	"pallet-vesting/runtime-benchmarks",
 	"pallet-session-benchmarking",
 	"pallet-staking/runtime-benchmarks",
+	"pallet-society/runtime-benchmarks",
 	"pallet-im-online/runtime-benchmarks",
+	"pallet-democracy/runtime-benchmarks",
+	"pallet-democracy/runtime-benchmarks",
```
bkchr: Duplicate

bin/node/runtime/src/lib.rs
```diff
@@ -917,6 +916,13 @@ impl_runtime_apis! {
 					steps,
 					repeat,
 				),
+				b"pallet-democracy" | b"democracy" => Democracy::run_benchmark(
```
kianenigma: is `run_benchmark` coming from a trait and will it hence have the same signature for all modules? in that case it would be way easier to create a tuple of all modules that implement it and call `run_benchmark()` on the tuple. Or maybe even use `AllModules`.

frame/collective/Cargo.toml
```diff
@@ -34,3 +34,7 @@ std = [
 	"sp-runtime/std",
 	"frame-system/std",
 ]
+runtime-benchmarks = [
+	"sp-runtime/runtime-benchmarks",
+	"frame-system/runtime-benchmarks",
+]
```
kianenigma:
```suggestion
]

```

frame/democracy/src/benchmarking.rs
```diff
@@ -0,0 +1,393 @@
...
+		proposal_hash,
+		vote_threshold,
+		0.into(),
+	);
+
+	Ok(0u32.into())
+}
+
+fn account_vote<T: Trait>() -> AccountVote<BalanceOf<T>> {
+	let v = Vote {
```
kianenigma:
```suggestion
	let vote = Vote {
```

frame/democracy/src/benchmarking.rs
```diff
@@ -0,0 +1,393 @@
...
+		aye: true,
+		conviction: Conviction::Locked1x,
+	};
+
+	AccountVote::Standard {
+		vote: v,
```
kianenigma:
```suggestion
		vote,
```

primitives/runtime/src/traits.rs
```diff
@@ -156,6 +156,12 @@ pub trait EnsureOrigin<OuterOrigin> {
 	}
 	/// Perform the origin check.
 	fn try_origin(o: OuterOrigin) -> result::Result<Self::Success, OuterOrigin>;
+
+	/// Returns an outer origin capable of passing `try_origin` check.
+	/// 
+	/// ** Should be used for benchmarking only!!! **
```
kianenigma: I won't complain because I don't know an alternative which is better :D mocking this in a way that has less footprint would be good. But I read in the discussion that this is apparently as good as it gets so 👍

bin/node/runtime/src/lib.rs
marcio-diaz: It doesn't have the same signature for all modules, because of instances. I'll see if it I can refactor it, in another PR.

kianenigma: yeah might be a good improvement 👍  but not really impactful so maybe in the long future. having some duplicate code is fine for now.

PR #5264, opened 2020-03-16
Remove the #[doc(hidden)] reexports of sc_network

### LINKED ISSUE #5246: sc_network_gossip shouldn't import sc_network::message::generic
The content of `sc_network::message::generic` is public but is hidden behind a `#[doc(hidden)]` because we want to no longer expose it.

However `sc_network_gossip` still makes use of it, for no particular reason.

This is a small change.
I'm opening this issue mostly to remember to do it (or for someone else to do it).

tomaka:
Close #5246

Removes the last `#[doc(hidden)]` public items from `sc_network` that we were planning to clean up.

This replaces all the usages of `ConsensusMessage` with `(ConsensusEngineId, Vec<u8>)`, which are the two fields of `ConsensusMessage`.


gnunicorn:
@tomaka this needs a conflict resolution.

PR #5266, opened 2020-03-16
client/network/src/protocol: Refactor Prometheus metric logic

mxinden:
I have the following suggestions for the Prometheus metrics within `client/network/src/protocol.rs`:

1. Start Prometheus metric help with capital

2. Differentiate metric status as label 

    Prometheus query language is powerful through its multi-dimensional data
model. Metric names are hirarchical whereas labels enable data to become
multi-dimensional.

    Exposing the justification of finality-proof status as a label allows
for more powerful queries.

3. Remove 'Total' from non counter metric help

    The word 'total' is reserved for accumulating counters. Counters have to
be monotonically increasing. `obsolete_requests` can decrease, thereby
it is defined as a `Gauge` and not a `Counter`.

    For more details on metric naming see
https://prometheus.io/docs/practices/naming/

### DIFF
--- a/client/network/src/protocol.rs
+++ b/client/network/src/protocol.rs
@@ -38,7 +38,7 @@ use sp_arithmetic::traits::SaturatedConversion;
 use message::{BlockAnnounce, BlockAttributes, Direction, FromBlock, Message, RequestId};
 use message::generic::Message as GenericMessage;
 use light_dispatch::{LightDispatch, LightDispatchNetwork, RequestData};
-use prometheus_endpoint::{Registry, Gauge, register, PrometheusError, U64};
+use prometheus_endpoint::{Registry, Gauge, GaugeVec, PrometheusError, Opts, register, U64};
 use sync::{ChainSync, SyncState};
 use crate::service::{TransactionPool, ExHashT};
 use crate::config::{BoxFinalityProofRequestBuilder, Roles};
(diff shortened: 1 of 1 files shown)


client/network/src/protocol.rs
```diff
@@ -1917,15 +1875,23 @@ impl<B: BlockT, H: ExHashT> Protocol<B, H> {
...
-			metrics.justifications_pending.set(m.justifications.pending_requests.into());
-			metrics.justifications_active.set(m.justifications.active_requests.into());
-			metrics.justifications_failed.set(m.justifications.failed_requests.into());
-			metrics.justifications_importing.set(m.justifications.importing_requests.into());
-
-			metrics.finality_proofs_pending.set(m.finality_proofs.pending_requests.into());
-			metrics.finality_proofs_active.set(m.finality_proofs.active_requests.into());
-			metrics.finality_proofs_failed.set(m.finality_proofs.failed_requests.into());
-			metrics.finality_proofs_importing.set(m.finality_proofs.importing_requests.into());
+			metrics.justifications.with_label_values(&["pending"])
```
mxinden: Matching a string e.g. `"pending"` to a structure field e.g. `m.justifications.pending_requests` might be error prone. I am thinking about alternatives but would do so in a follow-up pull request.

PR #4981, opened 2020-02-19
Introduce Prometheus metric endpoint replacing Grafana endpoint

mxinden:
Superseding B-Harvest's pull request https://github.com/paritytech/substrate/pull/4511/ adding a Prometheus metric endpoint to Substrate, replacing the current Grafana endpoint. 

//CC @nodebreaker0-0

PR #5271, opened 2020-03-17
CLI: refactoring: remove Options from sc_service::Configuration's fields

### LINKED ISSUE #4776: Service's Configuration has optional fields that shouldn't be optional

cecton:
## Major changes

1. Removes `VersionInfo` and replace by a trait SubstrateCli that needs to be implemented on the top level structopt of the binary
2. Add a proc_macro `spec_factory` to put on a spec factory/load spec function to help the impl of SubstrateCli (will automatically fill version, executable name, etc... for you)
3. Remove Option on `sc_service::Configuration`'s fields where it doesn't make any sense.
4. Add a trait `CliConfiguration` that needs to be implemented on structopts that are considered as commands (subcommands and runcmd). This these impl are all provided by sc_cli by default but if the user add a subcommand, they will need to implement it.
5. Add a proc_macro `substrate_cli_params` to help impl the CliConfiguration
5. Replaced `impl_version` by full version and dropped `Configuration.full_version()`

## Minor changes

5. Merge Configuration.node and NetworkConfiguration.node_name because they are the same thing
6. Removed config_path in NetworkConfiguration because it's not used and it's redundant with net_config_path
7. Re-export `NoExtension` and `TracingReceiver` in `sc_service`
8. Added a default impl for WasmExecutionMethod to Intepreted
9. Renamed "role" to "roles" because you can be an authority and a full node
8. ... (more details will follow)

## Help needed

Please check and propose fixes for all the new documentation.

Closes #4776

polkadot companion: paritytech/polkadot#935


client/cli/src/config.rs
```diff
@@ -0,0 +1,315 @@
...
+		.expect("app directories exist on all supported platforms; qed");
+		let config_dir = self
+			.get_base_path()?
+			.unwrap_or(&default_config_dir)
+			.join("chains")
+			.join(chain_spec.id());
+		let net_config_dir = config_dir.join(DEFAULT_NETWORK_CONFIG_PATH);
+		let client_id = C::client_id();
+		// TODO: this parameter is really optional, shouldn't we leave it to None?
+		let database_cache_size = Some(self.get_database_cache_size()?.unwrap_or(1024));
```
cecton: @bkchr this is a parameter you set to 1024 in 2b716b5a96e2806a94200434c0f54a10114a126c

Should I:
- keep it that way
- move the default value of 1024 to sc_service (as suggested by @tomaka )
- delete it and leave it to none

client/service/src/builder.rs
```diff
@@ -1139,7 +1137,7 @@ ServiceBuilder<
 			let system_info = sc_rpc::system::SystemInfo {
 				chain_name: chain_spec.name().into(),
 				impl_name: config.impl_name.into(),
-				impl_version: config.impl_version.into(),
+				impl_version: config.impl_version.into(), // TODO: this is now using full version instead of just version. Is it correct?
```
cecton: Just to be sure everybody is aware: the impl_version use to be the "short" version (ex: 1.0.0) but now it is the full version (ex: 2.0.0-alpha.2-b950f731c-x86_64-linux-gnu)

client/service/src/config.rs
```diff
@@ -18,69 +18,41 @@
...
+pub use sc_telemetry::TelemetryEndpoints;
 
 /// Service configuration.
 pub struct Configuration<G, E = NoExtension> {
 	/// Implementation name
 	pub impl_name: &'static str,
 	/// Implementation version
 	pub impl_version: &'static str,
-	/// Git commit if any.
-	pub impl_commit: &'static str,
```
cecton: Now in impl_version already

client/service/src/config.rs
```diff
@@ -18,69 +18,41 @@
...
 	/// Size of internal state cache in Bytes
 	pub state_cache_size: usize,
 	/// Size in percent of cache size dedicated to child tries
 	pub state_cache_child_ratio: Option<usize>,
 	/// Pruning settings.
 	pub pruning: PruningMode,
 	/// Chain configuration.
-	pub chain_spec: Option<ChainSpec<G, E>>,
-	/// Node name.
-	pub name: String,
```
cecton: Deleted in favor of network.node_name

client/service/src/config.rs
```diff
@@ -144,8 +114,8 @@ impl KeystoreConfig {
 	/// Returns the path for the keystore.
 	pub fn path(&self) -> Option<&Path> {
 		match self {
-			Self::Path { path, .. } => Some(&path),
-			Self::None | Self::InMemory => None,
+			Self::Path { path, .. } => Some(path),
+			Self::InMemory => None,
```
cecton: None was some kind of Option here

client/cli/src/lib.rs
```diff
@@ -221,3 +253,18 @@ fn kill_color(s: &str) -> String {
...
+
+	generate_cargo_keys(ConstantsFlags::SHA_SHORT).expect("Failed to generate metadata files");
+}
+
+/// Make sure the calling `build.rs` script is rerun when `.git/HEAD` changed.
+///
+/// The file is searched from the `CARGO_MANIFEST_DIR` upwards. If the file can not be found,
+/// a warning is generated.
+pub fn rerun_if_git_head_changed() {
+	build_script_utils::rerun_if_git_head_changed();
```
cecton: Do we still need to implement this ourselves? This functionality exist already in vergen: https://docs.rs/vergen/3.1.0/vergen/struct.ConstantsFlags.html#associatedconstant.REBUILD_ON_HEAD_CHANGE

client/cli/src/config.rs
```diff
@@ -0,0 +1,315 @@
...
+	/// This method:
+	///
+	/// 1. Set the panic handler
+	/// 2. Raise the FD limit
+	/// 3. Initialize the logger
+	fn init<C: SubstrateCLI<G, E>, G, E>(&self) -> Result<()>
+	where
+		G: RuntimeGenesis,
+		E: ChainSpecExtension;
+}
```
cecton: Can the reviewers check and fix all the documentation added here?

client/cli/src/config.rs
cecton: Correction: this value has been changed to 32 later on c3ff85d13b8b660c2c4eb850209c06a9397bed3e

cecton:
@pepyakin @bkchr @tomaka this PR is finally ready for review \o/

arkpar:
What's the motivation for introducing `CliConfiguration` and the proc macro? I fear this brings a lot of unneeded complexity to the code base. Now if someone needs to introduce a new option they would need to change the proc macro as well? It is quite hard to follow what's going on already, and this does not make is simpler.

Same applies to `spec_factory` which btw has a bit of a misleading name, since it provides much more than the actual chain spec now. What's wrong with providing version info by a struct, instead of figuring out how to use an obscure attribute?

arkpar:
More of a generic node: we tend not to use `get_` prefix for getters. 

Also, this will require a PR for polkadot before merging.

client/cli/src/config.rs
```diff
@@ -0,0 +1,308 @@
...
+		Ok(false)
+	}
+
+	/// Get the roles
+	fn get_roles(&self, _is_dev: bool) -> Result<Roles> {
+		Ok(Roles::FULL)
+	}
+
+	/// Get the transaction pool options
+	fn get_transaction_pool(&self) -> Result<TransactionPoolOptions> {
```
seunlanlege: Hmm, this method name doesn't match return type.

You could either
1. rename this to `get_transaction_pool_options`
2. drop the `get_` prefix so you don't have long-ass method names.

client/cli/src/config.rs
```diff
@@ -0,0 +1,308 @@
...
+		Ok(Default::default())
+	}
+
+	/// Get maximum runtime instances
+	fn get_max_runtime_instances(&self) -> Result<Option<usize>> {
+		Ok(Default::default())
+	}
+
+	/// Create a Configuration object from the current object
+	fn create_configuration<C: SubstrateCLI>(
```
seunlanlege: tbh this is a really great change, the `update_config` _methodology_ was very weird.

client/service/src/builder.rs
seunlanlege: cc @tomusdrw

cecton:
> What's the motivation for introducing `CliConfiguration` and the proc macro? 

Closing #4776 :grin: The problem we have right now is there are a lot of expect in the code because some of the fields of sc_service::Configuraiton are Option despite they shouldn't be option. This is the whole point of this major change but this PR solves also the issue that if someone wants to insert a field, they can actually missed the subtility of the current implementation and break the point 2. at the end of this answer. This happens [here](https://github.com/paritytech/substrate/pull/4981/files#diff-ed4918dd0da951e344e205c78f972484R650) for example (line 650).

> I fear this brings a lot of unneeded complexity to the code base.

Tbh I fear it too. I think this PR is introducing more complex maintenance but an easier API to use. Though I think that it is more beneficial overall because more errors can be detected at compile time.

> Now if someone needs to introduce a new option they would need to change the proc macro as well? It is quite hard to follow what's going on already, and this does not make is simpler.

Yes. Adding a new field in configuration will automatically make the compilation fail at multiple different places: in config.rs where CliConfiguration must define it's interface on how to get the new field but also in RunCmd and Subcommand where you need to define how to get the field (if there is no default) which lead you to the proc_macro if the field is common (shared parameters, import parameters, etc...).

> Same applies to `spec_factory` which btw has a bit of a misleading name, since it provides much more than the actual chain spec now.

Naming things is hard. I put that name to the macro for now but I'm waiting for someone to find better names for things. I'm not satisfied with the struct "Runtime" either.

> What's wrong with providing version info by a struct, instead of figuring out how to use an obscure attribute?

It's a lot of boiler plate and it's annoying to pass the object around. Here the object is replaced by a trait on the structopt directly (one less object :+1:) and on top of that: the user don't need to write it by hand. Check [how the API is used here](https://github.com/paritytech/substrate/pull/5271/files#diff-1a052be6a59b59c68229c2ae6fe1eaf9) and you will see that it is actually much simpler to use. It is not so obscure at it is already implemented in the node-template as example.

I guess the price here is the maintenance but since there are less errors only detectable at runtime... I think overall it's better. It's a bit like using clap vs structopt: clap is more basic Rust code while structopt is just a fancy macro (btw: you can definitely ignore the macro and implement CliConfiguration yourself, the macro is only here to reduce the boilerplate, it is not required).

> It is quite hard to follow what's going on already,

I'm improving sc-cli's API by simplifying it's use while keeping the flexibility needed for polkadot and cumulus. Those 2 projects have 2 particularities:

1. polkadot needs to run its runtime by its own because the runtime can be kusama (you know it at runtime). Therefore we need a split between "loading the configuration" and "running the node":

https://github.com/paritytech/polkadot/blob/5c4dcfb257adae9f7c10f689c126e6aedb27ce97/cli/src/command.rs#L60-L81

With this new API you can use Runtime::sync_run or Runtime::async_run to run whatever you need in a custom way.

2. cumulus needs to parse the command line arguments for 2 different nodes that will be run together in the same process. Each node has its own default parameters (like default ports because you can't open 2 times the same port). Therefore we need a way to let the binary define its own defaults for each node. This is currently done by overriding in Configuration: 

https://github.com/paritytech/cumulus/blob/dd5b344274f148411e90fcbf162aa2839144b849/test/parachain/src/command.rs#L106-L112

But with this new API this will be done by implementing CliConfiguration properly.

client/cli-derive/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "sc-cli-derive"
+version = "0.8.0-alpha.4"
+authors = ["Cecile Tonglet <cecile@parity.io>"]
```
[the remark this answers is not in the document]
cecton: :scream:

bin/node/cli/src/command.rs
```diff
@@ -14,102 +14,100 @@
...
+use crate::{
+	factory_impl::FactoryState,
+	service, ChainSpec, Cli, FactoryCmd, Subcommand,
+};
 use node_transaction_factory::RuntimeAdapter;
-use crate::{Cli, service, ChainSpec, load_spec, Subcommand, factory_impl::FactoryState};
+use sc_cli::{spec_factory, substrate_cli_params, CliConfiguration, Result, SubstrateCLI};
+use sc_service::Configuration;
+
+//TODO: 3 different types with the same name: ChainSpec
```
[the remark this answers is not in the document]
cecton: Yeah sorry :sweat_smile: I had no clue how to resolve this so I was waiting for some feedback. Explicitly writing the module sounds like a good idea for me! 

(I don't have much opinion on this, I just notice there is a problem)

client/cli-derive/src/spec_factory.rs
```diff
@@ -0,0 +1,117 @@
...
+			fn get_impl_name() -> &'static str { #impl_name }
+			fn get_impl_version() -> &'static str { #impl_version }
+			fn get_support_url() -> &'static str { #support_url }
+			fn get_executable_name() -> &'static str { #executable_name }
+			fn get_author() -> &'static str { #author }
+			fn get_description() -> &'static str { #description }
+			fn get_copyright_start_year() -> i32 { #copyright_start_year }
+		}
+	}
+}
```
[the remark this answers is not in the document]
cecton: Thanks! This whole macro is meant only to remove all the boilerplate with the VersionInfo. I renamed VersionInfo to SubstrateCLI because it now also has the load_spec/spec_factory function but if you have a better name I will just change it. Naming is hard

client/cli-derive/src/spec_factory.rs
```diff
@@ -0,0 +1,117 @@
...
+		"{}-{}{}{}",
+		TARGET_ARCH.as_str(),
+		TARGET_OS.as_str(),
+		env_dash,
+		TARGET_ENV.map(|x| x.as_str()).unwrap_or(""),
+	)
+}
+
+fn get_version() -> String {
+	let impl_commit = std::env::var("VERGEN_SHA_SHORT").unwrap_or_default();
```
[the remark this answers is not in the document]
cecton: Nice catch :ok_hand:

client/cli/src/config.rs
cecton: I will drop the `get_` everywhere as suggested here: https://github.com/paritytech/substrate/pull/5271#issuecomment-602550473

client/cli/src/config.rs
cecton: Yes I think it did its job but this way is better even though more verbose.... Thanks!! Took me sometime to come up with this idea

cecton:
> Whew, :sweat_smile:, took some time to really look through, but this PR has a lot of welcome changes. 👏🏾. Just some minor nits to sort out before getting it merged.

Thanks!!

To all reviewers: don't hesitate to ask me to rename things. Naming is hard. I have a few questions btw:
1. SubstrateCLI or SubstrateCli?
2. I don't like sc_cli::Runtime, maybe there is a better name for this object
3. The macro `spec_factory` is not reflecting what it does (impl a CliConfiguration for you) but on what it applies: a load_spec/spec_factory function

bin/node/cli/src/command.rs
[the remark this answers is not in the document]
seunlanlege: Alright, I think this is last thing to be resolved. So we don't merge code with todos in them.

client/cli/src/commands/mod.rs
```diff
@@ -84,56 +84,70 @@ impl Subcommand {
...
 	/// 2. Raise the FD limit
 	/// 3. Initialize the logger
-	pub fn init(&self, version: &VersionInfo) -> error::Result<()> {
-		self.get_shared_params().init(version)
+	pub fn init<C: SubstrateCLI>(&self) -> Result<()> {
+		self.get_shared_params().init::<C>()
 	}
 }
+
+macro_rules! match_and_call {
```
tomaka: You're trying to combine a big list of subcommands with a big list of methods. But instead of putting the subcommands in the macro, I would have put the methods.

In other words, something like that:

```rust
macro_rules! impl_cli_config {
    ($struct:ident, $($variants:ident)*) => {
        impl CliConfiguration for $struct {
            fn base_path(&self) -> Result<Option<&PathBuf>> {
                match self { $($struct::$variants(cmd) => { cmd.base_path() })* }
            }

            // ... same for the other methods ...
        }
    }
}

impl_cli_config!(Subcommand, BuildSpec, ExportBlocks, ImportBlocks, CheckBlock, Revert, PurgeChain);
```

client/cli-derive/src/substrate_cli_params.rs
```diff
@@ -0,0 +1,231 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Configuration trait for a CLI based on substrate
+
+use proc_macro;
+use proc_macro_error::{abort, abort_call_site};
+use quote::{quote, ToTokens};
+use std::collections::{HashMap, HashSet};
+use syn;
```
tomaka:
```suggestion
```

client/cli-derive/src/substrate_cli_params.rs
```diff
@@ -0,0 +1,231 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Configuration trait for a CLI based on substrate
+
+use proc_macro;
```
tomaka:
```suggestion
```

client/cli-derive/examples/basic.rs
```diff
@@ -0,0 +1,25 @@
...
+	cli = MyCli,
+	support_url = "http://example.org",
+	copyright_start_year = 2020,
+	impl_version = "0.1.0",
+)]
+fn spec_factory(id: &str) -> Result<Box<dyn sc_service::ChainSpec>, String> {
+	Err("not implemented".into())
+}
+
+#[substrate_cli_params(shared_params = shared, import_params = import, keystore_params = keystore)]
```
tomaka: What screams to be when looking at this example is that the procedural macro should really be a custom derive. They are the idiomatic way to automatically implement a trait on a struct.
An API like this looks much more comprehensible to me:

```rust
#[derive(CliCommand)]
#[cli_command(spec_factory = spec_factory, support_url = "foo", ...)]
struct MyCli {
        #[cli_command(shared_params)]
	shared: SharedParams,
        #[cli_command(import_params)]
	import: ImportParams,
        #[cli_command(keystore_params)]
	keystore: KeystoreParams,
}

// No attribute here
fn spec_factory(id: &str) -> Result<Box<dyn sc_service::ChainSpec>, String> {
	Err("not implemented".into())
}

// No manual `impl`
```

client/cli/src/commands/mod.rs
cecton: Fair point!

client/cli-derive/examples/basic.rs
cecton: TL;DR: I will work more on this to see what I can do.

Long:

I actually tried this pattern but there are multiple problems with it:

1. it makes the cli.rs not agnostic of substrate. We currently use the cli.rs to build auto-completion following this pattern: https://clap.rs/2016/10/25/an-update/ I splitted cli.rs into cli.rs + command.rs explicitly for that reason. See also: https://github.com/paritytech/substrate/blob/26e37ff/bin/node/cli/build.rs

    (Note that I don't recall what exactly was the problem if we don't split... I think it was a compilation problem with WASM)

2. substrate_cli_params actually allows to complete and override any field. Example [here](https://github.com/paritytech/substrate/blob/dc7703c/utils/frame/benchmarking-cli/src/command.rs#L128-L135) where the chain_spec is loaded differently for that particular subcommand.

3. having spec_factory on the root structopt is actually easier so I don't have to pass the function around. See how [Runtime pass it around](https://github.com/paritytech/substrate/blob/dc7703c6bd653d8abf75bdd6046df7511c54e005/client/cli/src/runtime.rs#L121) to create the Configuration.

cecton:
> The procedural macro also feels very over-engineered to me.

I do not disagree with you. As I said:

> Tbh I fear it too. I think this PR is introducing more complex maintenance but an easier API to use. Though I think that it is more beneficial overall because more errors can be detected at compile time.

I think @seunlanlege really sees the upgrade here because this PR abandons the "update_config" pattern. As he said: 

> tbh this is a really great change, the update_config methodology was very weird.

But to me it reduces the errors we can do by adding a parameter or implementing a new cli.

For comparison:

Before:

```rust
let mut config = Configuration::default();
cmd.init();
cmd.update_config(&mut config); // update the configuration with the command line parameters
cmd.run(config, |config| ...)
```

After:

```rust
let runtime = Cli::parse_args();
runtime.run(|config| ...);

// ...

impl CliConfiguration for Cli {
    ... // everything declared and get independently
}
```

bin/node/cli/src/command.rs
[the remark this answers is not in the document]
gnunicorn: or record them in an issue and link the issue, then the TODo/FIXME can stay...

client/cli/src/commands/mod.rs
cecton: It looks like we could also do a derive macro but then it would have the same problems that I mentioned there: https://github.com/paritytech/substrate/pull/5271#discussion_r397117504

client/cli-derive/src/spec_factory.rs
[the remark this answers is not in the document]
cecton: I renamed the macro to substrate_cli_configuration

bin/node/cli/src/command.rs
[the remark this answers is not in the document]
cecton: I fixed it by removing one of the ChainSpec that was not really useful

client/service/src/builder.rs
tomusdrw: I think that's fine, just make sure to mention it to @jacogr and include in release notes.

cecton:
Can all the reviewers vote with a :+1: or :-1:  on this comment about this question:

Do you prefer the existing API because it's easier to maintain :-1: 
Or do you prefer this new API because it's easier to use :+1: 

(Don't hesitate to be honest!)

cc @gnunicorn @arkpar @bkchr @seunlanlege @tomaka

client/cli-derive/examples/basic.rs
cecton: I couldn't find a better idea for now. I do like the derive macro but I know for sure it doesn't work (tested that first) because mostly of the point 2. I will close it for now but I'm opened to suggestion.

Meanwhile I did change spec_factory macro:
1. It now must be used on an impl (exactly like substrate_cli_params, this is better for consistency), it's also more explicit
2. I renamed it to substrate_cli_configuration since what it does is to implement substrate_cli_configuration for you

client/cli/src/lib.rs
cecton: > yes we do. Our script does a parent-search for the .git/head, while the one in vergen might only work if the build.rs is in the folder the .git is, if I understand this correctly.

client/service/src/builder.rs
cecton: he said all fine ^_^

client/cli/src/params/import_params.rs
```diff
@@ -52,11 +52,11 @@ pub struct ImportParams {
...
 	#[allow(missing_docs)]
 	#[structopt(flatten)]
-	pub execution_strategies: ExecutionStrategies,
+	pub execution_strategies: ExecutionStrategiesParams,
 
 	/// Limit the memory the database cache can use.
-	#[structopt(long = "db-cache", value_name = "MiB", default_value = "128")]
-	pub database_cache_size: u32,
+	#[structopt(long = "db-cache", value_name = "MiB")]
+	pub database_cache_size: Option<usize>,
```
gnunicorn: as said before, we don't want to pass `None` to rocksdb but our own default. I think leaving `default_value` in for documentation purposes (and to show it in the cli help ... does this actually?), this is better than it being optional.

docs/CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog].
 
 ## Unreleased
 
+### Changed
+
+Client
+------
+
+* `sc_rpc::system::SystemInfo.impl_version` now returns the full version (2.0.0-alpha.2-b950f731c-x86_64-linux-gnu) instead of the short version (1.0.0)
```
gnunicorn: please add a link to the PR in the changelog entry

docs/CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog].
 
 ## Unreleased
 
+### Changed
```
gnunicorn:
```suggestion

```

Not necessary

client/cli/src/params/import_params.rs
cecton: Unfortunately it makes things less flexible because a binary cannot override the default value anymore. That is why the default value is now in config.rs.

Though maybe for this particular case it is fine if we keep it in the definition... Let me know what you prefer.

(Note: it does show the default value in the --help if we keep it in the definition)

bin/node/cli/src/command.rs
```diff
@@ -14,102 +14,102 @@
...
-			)
-		},
+			let runtime = cli.create_runtime(&cli.run)?;
+			runtime.run_node(service::new_light, service::new_full)
+		}
 		Some(Subcommand::Inspect(cmd)) => {
-			cmd.init(&version)?;
-			cmd.update_config(&mut config, load_spec, &version)?;
+			use node_executor::*;
+			use node_runtime::*;
```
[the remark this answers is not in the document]
cecton: It was just more handy that way. These stuff are used only there. I don't mind moving stuff around.

bin/node-template/node/src/main.rs
```diff
@@ -8,16 +8,5 @@ mod cli;
...
-		version: env!("CARGO_PKG_VERSION"),
-		executable_name: "node-template",
-		author: "Anonymous",
-		description: "Template Node",
-		support_url: "support.anonymous.an",
-		copyright_start_year: 2017,
-	};
-
-	command::run(version)
+	command::run()
```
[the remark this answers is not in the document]
cecton: main.rs is really supposed to only bootstrap a program (Rust convention, my memory might be wrong!) so it is best to keep it that way

client/cli/src/config.rs
```diff
@@ -0,0 +1,310 @@
...
+	/// 1. Set the panic handler
+	/// 2. Raise the FD limit
+	/// 3. Initialize the logger
+	fn init<C: SubstrateCli>(&self) -> Result<()>;
+}
+
+/// Generate a valid random name for the node
+pub fn generate_node_name() -> String {
+	let result = loop {
+		let node_name = Generator::with_naming(Name::Numbered).next().unwrap();
```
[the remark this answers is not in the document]
cecton: Ok will fix it! (That code was not from me actually.)

client/service/src/config.rs
```diff
@@ -18,70 +18,42 @@
...
-}
-
 /// Service configuration.
 pub struct Configuration {
 	/// Implementation name
 	pub impl_name: &'static str,
 	/// Implementation version
 	pub impl_version: &'static str,
-	/// Git commit if any.
-	pub impl_commit: &'static str,
```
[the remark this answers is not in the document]
cecton: I have thought it through for the code itself but I don't know if there are other implications. @jacogr told me it was fine but let me know if there is any problem.

Strictly from the code perspective there was no reason to keep the version and the commit separated (and the full version) so I store one version (the full version) and if we ever need a part of it we can just parse this string.

bin/node-template/node/Cargo.toml
```diff
@@ -37,5 +38,4 @@ sc-basic-authorship = { path = "../../../client/basic-authorship", version = "0.
 node-template-runtime = { version = "2.0.0-alpha.5", path = "../runtime" }
 
 [build-dependencies]
-vergen = "3.0.4"
-build-script-utils = { version = "2.0.0-alpha.5", package = "substrate-build-script-utils", path = "../../../utils/build-script-utils" }
+sc-cli = { version = "0.8.0-alpha.4", path = "../../../client/cli" }
```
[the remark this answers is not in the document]
cecton: You mean I should move it to substrate-build-script-utils? (That's fine by me)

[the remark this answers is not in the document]
cecton: Note: sc-cli OR structopt still needs to be in the build dependencies because that's how the shell completions are generated (in bin/node/cli).

Using structopt directly is fine (even if the version don't match) but if sc-cli gets a structopt upgrade to a major release, the structopt build dependency will need to get also the upgrade.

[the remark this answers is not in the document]
cecton: Note: actually we need:
- sc-cli for the ShareParams, ImportParams, etc...
- structopt because the derive cannot be imported from sc-cli https://github.com/TeXitoi/structopt/issues/339
- and now substrate-build-script-utils (that's the only one we could remove if we just re-export what we need)

client/cli/src/lib.rs
```diff
@@ -19,137 +19,164 @@
...
-pub fn try_from_iter<T, I>(iter: I, version: &VersionInfo) -> clap::Result<T>
-where
-	T: StructOpt + Sized,
-	I: IntoIterator,
-	I::Item: Into<std::ffi::OsString> + Clone,
+/// Substrate client CLI
+pub trait SubstrateCli: Sized
 {
-	let app = T::clap();
+	/// Implementation name.
```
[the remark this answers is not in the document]
cecton: Can you point me that again? (I don't remember and there are a lot of comments :sweat: ) Or even better: propose a fix

client/cli/Cargo.toml
```diff
@@ -23,6 +23,7 @@ tokio = { version = "0.2.9", features = [ "signal", "rt-core", "rt-threaded" ] }
 futures = "0.3.1"
 fdlimit = "0.1.4"
 serde_json = "1.0.41"
+sc-cli-derive = { version = "0.8.0-alpha.5", path = "../cli-derive" }
```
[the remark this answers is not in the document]
cecton: You can't, it's a proc-macro crate. I can't find a link to the doc that explains it but it's a limitation in Rust for now

bin/node-template/node/src/main.rs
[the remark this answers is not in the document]
gnunicorn: Well, I argued that what is in the `command.rs` is the boostrapping and thus these should be collapsed into one. But I care 2/10 about this.

client/cli/src/lib.rs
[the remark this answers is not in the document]
gnunicorn: I was referring to this comment https://github.com/paritytech/substrate/pull/5271#discussion_r397989558 

From looking at the code it is unclear to me which a user would discover first in the docs, so where this is the most appropriate to put here or on the other. The other should probably just point to this one so we have on single definition.

bin/node-template/node/src/main.rs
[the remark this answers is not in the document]
cecton: hmmm... :thinking: I guess you could say so. Though parsing the command line arguments is not exactly bootstrapping... or is it? I guess I don't have much opinion on this. What do you think is best? Should I proceed and merge the files?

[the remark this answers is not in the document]
cecton: Ok I will keep it as it for now as I'm a bit lazy to change and make it consistent with the other binaries

client/cli/src/params/import_params.rs
cecton: Done as suggested in DM

client/cli/src/lib.rs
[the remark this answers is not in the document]
cecton: Done as suggested in DM

Cargo.toml
```diff
@@ -21,6 +21,7 @@ members = [
 	"client/chain-spec",
 	"client/chain-spec/derive",
 	"client/cli",
+	"client/cli-derive",
```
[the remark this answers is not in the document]
cecton: I don't mind but I find weird to have 2 crates in a single crate directory:
```
cli/Cargo.toml
cli/src
cli/derive/Cargo.toml
cli/derive/src
```
It looks messy imo

cecton:
> so does this also change the cli-param from --name to --node-name? Is there any other --name-used or other good reason to make this currently short cli-params longer? If not, I'd prefer that we stick with --name as the cli-params (internal naming I don't care that much).

@gnunicorn sorry GitHub hided that comment very well from my eyes.

To answer your question: no, the parameter is still name. This change only applies to sc_service Configuration, no impact on the CLI itself.

bin/node-template/node/src/command.rs
```diff
@@ -14,36 +14,39 @@
...
 
-/// Parse and run command line arguments
-pub fn run(version: VersionInfo) -> sc_cli::Result<()> {
-	let opt = sc_cli::from_args::<Cli>(&version);
+#[substrate_cli(
+	impl_name = "Substrate Node",
+	support_url = "support.anonymous.an",
+	copyright_start_year = 2017
+)]
+impl SubstrateCli for Cli {
```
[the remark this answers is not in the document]
cecton: See https://github.com/paritytech/substrate/pull/5271#discussion_r397083720

bin/node-template/node/src/command.rs
```diff
@@ -14,36 +14,39 @@
...
-			opt.run.init(&version)?;
-			opt.run.update_config(&mut config, chain_spec::load_spec, &version)?;
-			opt.run.run(
-				config,
-				service::new_light,
-				service::new_full,
-				&version,
-			)
-		},
+			let runtime = cli.create_runtime(&cli.run)?;
```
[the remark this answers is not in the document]
cecton: No I guess it's a side effect to the fact that RunCmd is not part of Subcommand. We can solve this by adding a command "run" that we will need to type and that will run the node if you want :innocent: 

Or we can stop using structopt and use clap directly so we will have more flexibility. It will also solve the problem of the default values on the definitions (see [here](https://github.com/paritytech/substrate/pull/5271#discussion_r398017673)).

Or we can use structopt internals or debate with them if they should make augment_clap public or not.

Cargo.toml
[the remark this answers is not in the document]
cecton: done

cecton:
> I don't see tests for the macros?

@bkchr There is a doc test for each macro:
- https://github.com/paritytech/substrate/pull/5271/files#diff-f4464278640c8ff5fe6e441d14f3afcdR35-R55
- https://github.com/paritytech/substrate/pull/5271/files#diff-f4464278640c8ff5fe6e441d14f3afcdR72-R92
- https://github.com/paritytech/substrate/pull/5271/files#diff-f4464278640c8ff5fe6e441d14f3afcdR109-R132

client/cli/src/runner.rs
```diff
@@ -142,9 +142,9 @@ impl<C: SubstrateCli> Runner<C>
 			C::copyright_start_year(),
 			Local::today().year()
 		);
-		info!("Chain specification: {}", self.config.chain_spec.name());
-		info!("Node name: {}", self.config.network.node_name);
-		info!("Roles: {}", self.config.display_role());
+		info!("📋 Chain specification: {}", self.config.chain_spec.name());
+		info!("🏷 Node name: {}", self.config.network.node_name);
+		info!("👤 Roles: {}", self.config.display_role());
```
gnunicorn: I like this. Very cute!

cecton: huhu that's not me, it was on master. I just updated the code consistently with what is on master

client/cli/src/runner.rs
```diff
@@ -0,0 +1,230 @@
...
+	/// A helper function that runs a command with the configuration of this node
+	pub fn sync_run(self, runner: impl FnOnce(Configuration) -> Result<()>) -> Result<()> {
+		runner(self.config)
+	}
+
+	/// A helper function that runs a future with tokio and stops if the process receives
+	/// the signal SIGTERM or SIGINT
+	pub fn async_run<FUT>(self, runner: impl FnOnce(Configuration) -> FUT) -> Result<()>
+	where
+		FUT: Future<Output = Result<()>> + future::Future,
```
bkchr:
```suggestion
		FUT: Future<Output = Result<()>>,
```

client/cli/src/runner.rs
```diff
@@ -0,0 +1,230 @@
...
+
+/// A Substrate CLI runtime that can be used to run a node or a command
+pub struct Runner<C: SubstrateCli> {
+	config: Configuration,
+	tokio_runtime: tokio::runtime::Runtime,
+	phantom: PhantomData<C>,
+}
+
+impl<C: SubstrateCli> Runner<C>
+{
```
bkchr:
```suggestion
impl<C: SubstrateCli> Runner<C>
{
```
```suggestion
impl<C: SubstrateCli> Runner<C> {
```

client/cli/src/runner.rs
```diff
@@ -0,0 +1,230 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
```
bkchr:
```suggestion
// Copyright 2020 Parity Technologies (UK) Ltd.
```

client/db/src/lib.rs
```diff
@@ -277,7 +277,7 @@ pub enum DatabaseSettingsSrc {
 		/// Path to the database.
 		path: PathBuf,
 		/// Cache size in bytes. If `None` default is used.
```
```suggestion
		/// Cache size in bytes.
```

client/network/src/service.rs
```diff
@@ -163,9 +163,7 @@ impl<B: BlockT + 'static, H: ExHashT> NetworkWorker<B, H> {
 	pub fn new(params: Params<B, H>) -> Result<NetworkWorker<B, H>, Error> {
 		let (to_worker, from_worker) = mpsc::unbounded();
 
-		if let Some(ref path) = params.network_config.net_config_path {
-			fs::create_dir_all(Path::new(path))?;
-		}
+		fs::create_dir_all(Path::new(&params.network_config.net_config_path))?;
```
bkchr:
```suggestion
		fs::create_dir_all(&params.network_config.net_config_path)?;
```

client/cli/src/params/shared_params.rs
```diff
@@ -93,20 +88,26 @@ impl SharedParams {
...
-					author: version.author
-				}
-			).expect("app directories exist on all supported platforms; qed")
-		)
+	/// Get the database configuration object for the parameters provided
+	pub fn database_config(
+		&self,
+		base_path: &PathBuf,
+		cache_size: usize,
+	) -> Result<DatabaseConfig> {
```
bkchr: Why does this returns a result?

client/cli/src/params/shared_params.rs
```diff
@@ -39,51 +38,47 @@ pub struct SharedParams {
...
-					.in_chain_config_dir(DEFAULT_DB_CONFIG_PATH)
-					.expect("We provided a base_path/config_dir."),
-				cache_size: None,
-			});
-		}
-
-		Ok(config.expect_chain_spec())
+			None => {
+				if is_dev {
+					"dev".into()
```
bkchr: It is rather weird that this returns `dev`. I know that you can overwrite this function, but still weird. For polkadot it is for example `polkadot-dev` and `kusama-dev`.

client/cli/src/lib.rs
```diff
@@ -19,137 +19,167 @@
...
+		task_executor: Arc<dyn Fn(Pin<Box<dyn Future<Output = ()> + Send>>) + Send + Sync>,
+	) -> error::Result<Configuration>
+	{
+		command.create_configuration(self, task_executor)
+	}
+
+	/// Create a runner for the command provided in argument. This will create a Configuration and
+	/// a tokio runtime
+	fn create_runner<T: CliConfiguration>(&self, command: &T) -> error::Result<Runner<Self>>
+	{
```
```suggestion
	fn create_runner<T: CliConfiguration>(&self, command: &T) -> error::Result<Runner<Self>> {
```

client/cli/src/lib.rs
```diff
@@ -19,137 +19,167 @@
...
+	}
 
-	Ok(())
+	/// Only create a Configuration for the command provided in argument
+	fn create_configuration<T: CliConfiguration>(
+		&self,
+		command: &T,
+		task_executor: Arc<dyn Fn(Pin<Box<dyn Future<Output = ()> + Send>>) + Send + Sync>,
+	) -> error::Result<Configuration>
+	{
```
```suggestion
	) -> error::Result<Configuration> {
```

client/cli/src/lib.rs
```diff
@@ -19,137 +19,167 @@
...
-/// `AppSettings::ArgsNegateSubcommands` and `AppSettings::SubcommandsNegateReqs`.
-///
-/// Gets the struct from the command line arguments. Print the
-/// error message and quit the program in case of failure.
-pub fn from_args<T>(version: &VersionInfo) -> T
-where
-	T: StructOpt + Sized,
+/// Substrate client CLI
+pub trait SubstrateCli: Sized
 {
```
```suggestion
pub trait SubstrateCli: Sized {
```

client/cli/src/lib.rs
```diff
@@ -19,137 +19,167 @@
...
-///
-/// To allow running the node without subcommand, tt also sets a few more settings:
-/// `AppSettings::ArgsNegateSubcommands` and `AppSettings::SubcommandsNegateReqs`.
-///
-/// Gets the struct from the command line arguments. Print the
-/// error message and quit the program in case of failure.
-pub fn from_args<T>(version: &VersionInfo) -> T
-where
-	T: StructOpt + Sized,
+/// Substrate client CLI
```
bkchr: Please give more documentation what this is.

client/cli/derive/src/substrate_cli.rs
```diff
@@ -0,0 +1,175 @@
...
+		"{}-{}{}{}",
+		TARGET_ARCH.as_str(),
+		TARGET_OS.as_str(),
+		env_dash,
+		TARGET_ENV.map(|x| x.as_str()).unwrap_or(""),
+	)
+}
+
+fn get_version() -> String {
+	let impl_commit = std::env::var("VERGEN_SHA_SHORT")
```
bkchr: It is not guaranteed that the macro is called when this env variable changes!

client/cli/derive/src/substrate_cli.rs
```diff
@@ -0,0 +1,175 @@
...
+		.unwrap_or_else(|_| {
+			abort_call_site!("missing env variable VERGEN_SHA_SHORT. You might want to add a build.rs script that runs substrate_build_script_utils::generate_cargo_keys() or provide the macro attribute impl_version yourself (impl_version = \"1.0.0\")")
+		});
+	let commit_dash = if impl_commit.is_empty() { "" } else { "-" };
+
+	format!(
+		"{}{}{}-{}",
+		std::env::var("CARGO_PKG_VERSION").unwrap_or_default(),
```
bkchr: Same

cecton:
> I see another solution, at least for the implementation of all these functions.

@bkchr Almost all the functions in CliConfiguration cannot have a default implementation because the implementation depends on shared_params, import_params, etc... those *_params exist on the type only if they are defined (they are not all defined for all commands).

The only common one is shared_params but the binary could decide to not use shared_params and go for its own implementation.

You cannot generalize `fn shared_params()` in the CliConfiguration because its parameters must be overridable separately (for example: I want to be able to override the default RPC ports).

Maybe I should move CliConfiguration to sc_service and rename it to "CreateConfiguration" or something. It's not really CLI-dependant. It's a trait to generate a Configuration. (When I created it it was called "IntoConfiguration".)

client/cli/src/params/shared_params.rs
cecton: CliConfiguration is meant to be fallible for any field. This allows the binary to customize the behavior during the parsing and decide to fail if some conditions are met. It's just to be generic.

client/cli/src/params/shared_params.rs
cecton: Loading the spec probably requires more refactoring. It was already stringly-typed like that. In polkadot you can see that I actually removed an entire enum and a 2-step process (from && load) because it was unnecessary. I think we can open an issue for this.

client/cli/derive/src/substrate_cli.rs
cecton: The macro is used on the binary/cli. The binary depends of all its parts (service, etc...). I don't see how it wouldn't be re-triggered, it's on the top of the build. Can you find an example?

bkchr: Just because the build is triggered, doesn't mean the macro is called again. Rustc caches compilation outputs to not rebuild stuff every time.

bkchr:
`fn shared_params()` would not be generalized, it would be implemented for each CLI implementation. Can you please give an example on how to override the default RPC ports work? Not sure how that is related here.

client/cli/derive/src/substrate_cli.rs
cecton: Does `rerun_if_git_head_changed` force the rebuild? https://docs.rs/substrate-build-script-utils/2.0.0-alpha.3/substrate_build_script_utils/fn.rerun_if_git_head_changed.html

bkchr: That just makes sure that the `build` script is executed again.

cecton:
```rust
use sc_cli_derive::substrate_cli_params;

struct MyCommand {
	shared: sc_cli::SharedParams,
}

#[substrate_cli_params(
	shared_params = shared,
)]
impl sc_cli::CliConfiguration for MyCommand {
	fn rpc_http(&self) -> Result<Option<SocketAddr>> {
		unimplemented!()
	}

	fn rpc_ws(&self) -> Result<Option<SocketAddr>> {
		unimplemented!()
	}
}
```

client/cli/derive/src/substrate_cli.rs
cecton: Well I don't know. Can you show me an example or some doc? I'm not sure I understand what is the problem

cecton: (I do it exactly like they do in structopt: https://github.com/TeXitoi/structopt/blob/master/structopt-derive/src/attrs.rs#L330 )

bkchr: https://users.rust-lang.org/t/equivalent-of-cargo-rerun-if-changed-for-procedural-macros/37036 for example. Nothing says that the macro is re-run.

bkchr:
```
trait CliConfiguration {
    fn shared_params(&self) -> cli::SharedParams;
    fn rpc_http(&self) -> Result<Option<SocketAddr>> { Ok(Some(4455)) }
    fn rpc_ws(&self) -> Result<Option<SocketAddr>> { Ok(Some(2345)) }

   fn something_else(&self) -> String {
       self.shared_params().something_else()
   }
};

struct MyCommand {
	shared: sc_cli::SharedParams,
}

impl sc_cli::CliConfiguration for MyCommand {
	fn rpc_http(&self) -> Result<Option<SocketAddr>> {
		unimplemented!()
	}

	fn rpc_ws(&self) -> Result<Option<SocketAddr>> {
		unimplemented!()
	}
}
```

A pure trait solution, that works without the macro.

cecton:
It's already like that. The macro are just here to help with the boilerplate.

And you can't have import_params and the others because it means you need to define the getter for those but the commands don't necessarily have those. (Maybe shared_params is the exception though... I will look into it)

client/cli/derive/src/substrate_cli.rs
cecton: The macro won't rerun because vergen_sha_short has changed indeed, but this is fine.

The code in the macro is executed during the build of the binary. The binary is on top of everything so the code will be always executed. This works exactly the same than structopt works. That particular macro (substrate_cli) is only used in the binary, not in sc-cli.

bkchr:
No the macro just makes it more complicated and harder to understand. 

Maybe have different kind of traits, per parameter type?

cecton:
Do you think structopt makes clap more complicated and harder to understand too?

cecton:
(And no you need to have only one trait because there is one Configuration. You can maybe divide in multiple traits but then you will need a top-level trait to call everything)

bkchr:
You can not compare `structopt` with your macro here. This macro here is matching function names and decides on that if something is overridden or not. Just one small typo and the stuff is not overriden and you will go on a hunt for multiple hours for this small bug.

cecton:
> Just one small typo and the stuff is not overriden and you will go on a hunt for multiple hours for this small bug.

No, you put it on the implementation of a trait. If you make a typo it will tell you the function does not exist.

client/cli/derive/src/substrate_cli.rs
cecton: Tested with the new implementation and it works:

```
[130] [12:05:30] ~/r/s/b/n/cli cecton-the-revenge-of-the-cli > c run -- --version
   Compiling node-cli v2.0.0-alpha.5 (/home/cecile/repos/substrate/bin/node/cli)
    Finished dev [unoptimized + debuginfo] target(s) in 53.38s
     Running `/home/cecile/repos/substrate/target/debug/substrate --version`
substrate 2.0.0-alpha.5-41295e618-x86_64-linux-gnu

[0] [12:09:09] ~/r/s/b/n/cli cecton-the-revenge-of-the-cli > c run -- --version
error: no matching package named `node-cli` found
location searched: /home/cecile/repos/substrate/bin/node/cli
prerelease package needs to be specified explicitly
node-cli = { version = "2.0.1-alpha.5" }
required by package `chain-spec-builder v2.0.0-alpha.5 (/home/cecile/repos/substrate/bin/utils/chain-spec-builder)`
[101] [12:09:14] ~/r/s/b/n/cli cecton-the-revenge-of-the-cli > c run -- --version
   Compiling node-runtime v2.0.0-alpha.5 (/home/cecile/repos/substrate/bin/node/runtime)
   Compiling node-cli v2.0.1-alpha.5 (/home/cecile/repos/substrate/bin/node/cli)
   Compiling node-rpc v2.0.0-alpha.5 (/home/cecile/repos/substrate/bin/node/rpc)
   Compiling node-executor v2.0.0-alpha.5 (/home/cecile/repos/substrate/bin/node/executor)
    Finished dev [unoptimized + debuginfo] target(s) in 1m 38s
     Running `/home/cecile/repos/substrate/target/debug/substrate --version`
substrate 2.0.1-alpha.5-41295e618-x86_64-linux-gnu
```

client/cli/src/commands/runcmd.rs
```diff
@@ -142,6 +133,7 @@ pub struct RunCmd {
 	/// value). Value of `all` will disable origin validation. Default is to
 	/// allow localhost and https://polkadot.js.org origins. When running in
 	/// --dev mode the default is to allow all origins.
+	// NOTE: this is an option so implementations can set their own defaults
```
bkchr, line 136:
```suggestion
```

bin/node-template/node/src/command.rs
```diff
@@ -14,36 +14,64 @@
...
+	fn support_url() -> &'static str {
+		"support.anonymous.an"
+	}
+
+	fn copyright_start_year() -> i32 {
+		2017
+	}
+
+	fn executable_name() -> &'static str {
+		env!("CARGO_PKG_NAME")
```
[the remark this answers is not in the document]
cecton: Because that comes from runtime. structopt uses CARGO_PKG_NAME too and I think it's better but I can't pinpoint an example that shows why... I need to do some research.

If we would use current_exe then we could drop the whole executable_name function if we want.

client/cli/src/commands/runcmd.rs
```diff
@@ -240,259 +233,219 @@ pub struct RunCmd {
...
+	}
+
+	fn node_name(&self) -> Result<String> {
+		let name: String = match (self.name.as_ref(), self.get_keyring()) {
+			(Some(name), _) => name.to_string(),
+			(_, Some(keyring)) => keyring.to_string(),
+			(None, None) => crate::generate_node_name(),
+		};
+
+		if let Err(msg) = is_node_name_valid(&name) {
```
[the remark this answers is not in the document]
cecton: Ok will fix it (keep in mind that I just move code around, I try to avoid adding bugs by refactoring too much)

client/cli/src/commands/mod.rs
```diff
@@ -51,89 +40,343 @@ const DEFAULT_NETWORK_CONFIG_PATH : &'static str = "network";
...
+///	        }
+///	    }
+///
+///     // ...
+/// }
+/// ```
+#[macro_export]
+macro_rules! substrate_cli_subcommands {
+	($enum:ident => $($variant:ident),*) => {
+		impl ::sc_cli::CliConfiguration for $enum {
```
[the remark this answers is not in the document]
cecton: Ahh I didn't know that! I used `extern crate self as sc_cli` to make it work in both the current crate and a crate using it, I guess I can drop that now. Thanks!

client/cli/src/commands/runcmd.rs
```diff
@@ -240,259 +233,219 @@ pub struct RunCmd {
...
+			return Err(Error::Input(format!(
+				"Invalid node name '{}'. Reason: {}. If unsure, use none.",
+				name, msg
+			)));
+		}
+
+		Ok(name)
+	}
+
+	fn dev_key_seed(&self, is_dev: bool) -> Result<Option<String>> {
```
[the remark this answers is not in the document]
cecton: > However, I don't understand why all this infallible methods return a Result. This needs to be changed!

I think it's best to have a Result everywhere for the following use case:

If I want to implement an application that will return an error if certain arguments are passed, then I can return Err and it will fail early during the run time.

After all, the reason why it fails can be very arbitrary and dependent on your use case.

client/cli/src/commands/export_blocks_cmd.rs
```diff
@@ -14,39 +14,39 @@
...
 	///
 	/// Default is 1.
+	// NOTE: this is an option so implementations can set their own defaults
 	#[structopt(long = "from", value_name = "BLOCK")]
 	pub from: Option<BlockNumber>,
 
 	/// Specify last block number.
 	///
 	/// Default is best block.
+	// NOTE: this is an option so implementations can set their own defaults
```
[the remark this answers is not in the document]
cecton: wait no. why am I removing this. This has been requested by @gnunicorn . Can you check together and let met know if I should keep them or not?

client/cli/src/commands/export_blocks_cmd.rs
```diff
@@ -14,39 +14,39 @@
...
 #[derive(Debug, StructOpt, Clone)]
 pub struct ExportBlocksCmd {
 	/// Output file name or stdout if unspecified.
+	// NOTE: this is an option so implementations can set their own defaults
 	#[structopt(parse(from_os_str))]
 	pub output: Option<PathBuf>,
 
 	/// Specify starting block number.
 	///
 	/// Default is 1.
```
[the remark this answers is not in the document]
cecton: See conversation https://github.com/paritytech/substrate/pull/5271#discussion_r398017673

client/cli/src/commands/export_blocks_cmd.rs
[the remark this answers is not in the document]
bkchr: These are just code comments(aka not seen by anyone) and they don't bring any value. If the value is none, the implementation needs to provide a default value or ignore it 🤷‍♂️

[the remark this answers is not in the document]
bkchr: The comment you showed me from @gnunicorn is about showing default values in the cli. Which is not easy doable with the current implementation. But this comment does not bring any value.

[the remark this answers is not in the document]
gnunicorn: The comment she is referring to was in a personal conversation. I was making the point that why this is being an `Option` that always has a default is non-obvious and easily something that someone in the future on a refactor or otherwise touching this code might consider optimizing and thus breaking it without realising (and that is something that is also easily overlooked by a reviewer). However that is future relevant, yet undocumented at this point, and I don't want to have this regression because of simple oversight. So I proposed to put in a comment that makes clear to a future developer, why this is an `Option`. As we have no other safeguard to on this, I propose keeping the comment – it is an easy and quick safeguard.

bin/node-template/node/src/command.rs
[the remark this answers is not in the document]
bkchr: Yeah that was my point of dropping this function.

client/cli/src/commands/export_blocks_cmd.rs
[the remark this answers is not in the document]
gnunicorn: Reiterating from a conversation here – note that I care only 3 out of 10 about this. I don't think it is obvious from the code that having an `Option` is an explicitly wanted feature and thus the likelihood of this being refactored away unknowingly is pretty high. Vigilance on the side of the reviewer can prevent that from happening, sure, and the comment increases the chances of that happening. Another easy recommendation to ensure this feature is not being removed accidentally would also be to have a test case `test_that_XYZ_default_can_be_provided_by_implementation` rather than the comment. This would make it even more clear why this is the way it is. 

However, I do not insist on either of these.  If you believe vigilance from the reviewer is fine, alright with me, too. You are owners of this code ;) .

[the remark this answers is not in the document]
cecton: I don't have opinion on this tbh. It seems to me that the comment can easily be missed but it's not wrong to have one. Making a test is probably better but I'm not entirely sure it will be clear, someone can just update the test without seeing the point. So I will do whatever you guys decide.

The reason why it is here is because we use structopt in a library and it's not really made for that imo... Time will tell if what we choose here was the right decision or not.

bin/node-template/node/src/command.rs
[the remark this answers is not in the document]
cecton: Leave it as it for now as agreed

utils/build-script-utils/src/git.rs
```diff
@@ -0,0 +1,123 @@
...
+		manifest_dir.pop();
+	}
+
+	println!(
+		"cargo:warning=Could not find `.git/HEAD` searching from `{}` upwards!",
+		manifest_dir_copy.display(),
+	);
+}
+
+// Code taken from https://github.com/rustyhorde/vergen/blob/8d522db8c8e16e26c0fc9ea8e6b0247cbf5cca84/src/output/envvar.rs
```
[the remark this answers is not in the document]
cecton: We do it on our own at the moment AND we don't use the code from the crate! See here: https://github.com/paritytech/substrate/blob/master/utils/build-script-utils/src/lib.rs#L25-L43

I mixed the 2 implementations because the one of vergen is actually better because it handles git worktree and our was also better because it handled cargo workspaces.

It is not possible to use the one in the crate as this code is private. It won't mix.

PR #5273, opened 2020-03-17
Don't format addresses when telemetry is disabled

arkpar:
Heaptrack shows this is second hottest source for temporary allocations.
`voters.len() <= TELEMETRY_VOTERS_LIMIT` should not be the case on kusama as I understand.

### DIFF
--- a/client/finality-grandpa/src/communication/mod.rs
+++ b/client/finality-grandpa/src/communication/mod.rs
@@ -474,12 +474,12 @@ fn incoming_global<B: BlockT>(
 		gossip_validator: &Arc<GossipValidator<B>>,
 		voters: &VoterSet<AuthorityId>,
 	| {
-		let precommits_signed_by: Vec<String> =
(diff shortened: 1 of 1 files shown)


client/finality-grandpa/src/communication/mod.rs
```diff
@@ -474,12 +474,12 @@ fn incoming_global<B: BlockT>(
 		gossip_validator: &Arc<GossipValidator<B>>,
 		voters: &VoterSet<AuthorityId>,
 	| {
-		let precommits_signed_by: Vec<String> =
-			msg.message.auth_data.iter().map(move |(_, a)| {
-				format!("{}", a)
-			}).collect();
-
 		if voters.len() <= TELEMETRY_VOTERS_LIMIT {
+			let precommits_signed_by: Vec<String> =
```
andresilva: I'm dumb, I added this guard since you had previously said there were a lot of allocations in gossip telemetry, and apparently I forgot to move this inside the guard.

PR #5275, opened 2020-03-17
Update COMPLEXITY.md for ext_terminate

athei:
The `ext_terminate` [PR](https://github.com/paritytech/substrate/pull/5234) was missing some updates to the `COMPLEXITY.md`. This adds the missing explanations and other updates.

I am unsure about how many DB writes a contract deletions induces. Please review that part carefully.

### DIFF
--- a/frame/contracts/COMPLEXITY.md
+++ b/frame/contracts/COMPLEXITY.md
@@ -166,7 +166,8 @@ This function performs the following steps:
 
 In the course of the execution this function can perform up to 2 DB reads to `get_balance` of source and destination accounts. It can also induce up to 2 DB writes via `set_balance` if flushed to the storage.
 
-Moreover, if the source balance goes below `existential_deposit` then the account will be deleted along with all its storage which requires time proportional to the number of storage entries of that account.
+Moreover, if the source balance goes below `existential_deposit`  then the transfer is denied and
+returns with an error.
 
 Assuming marshaled size of a balance value is of the constant size we can neglect its effect on the performance.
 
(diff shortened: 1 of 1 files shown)


frame/contracts/COMPLEXITY.md
```diff
@@ -187,6 +188,20 @@ implementation they just involve a DB read.
...
+## Terminate
+
+This function performs the following steps:
+
+1. Check the calling contract is not already on the callstack by calling `is_live`.
+2. `transfer` funds from caller to the beneficiary.
+3. Flag the caller contract as deleted in the overlay.
+
+`is_live` does not do any database access nor does it allocate memory. It walks up the call
+stack and therefore executes in linear time depending on size of the call stack.
```
pepyakin: Which is also has upper bound (IIRC, 32 in size) so can be considered constant.

frame/contracts/COMPLEXITY.md
```diff
@@ -187,6 +188,20 @@ implementation they just involve a DB read.
...
+
+**complexity**: Database accesses as described in Transfer + 1 DB write that removes
```
pepyakin: The original text stated that it required a number of writes proportional to the number of the key value pairs that the contract had at the moment of removal.

That is not sound and the thinking has changed since then. So we are are using now child tries for storing the storage of a contract. To remove it we kill it. For now, this operation requires iteration over each value. However, in future we expect this to be constant-time (from the perspective of the runtime).

I think we can just explicitly state here that we are using child trie removal and we expect that it will become constant time (from PoV of the runtime)

frame/contracts/COMPLEXITY.md
athei: Added that information.

frame/contracts/COMPLEXITY.md
athei: Uff. Corrected the statement. Which upcoming changes will make this constant time?

pepyakin: I think we don't have an issue for that, but it seems we've got a PR now https://github.com/paritytech/substrate/pull/5280

PR #5277, opened 2020-03-17
Fixed a couple of syncing issues

arkpar:
This fixes a couple of issues observed on one of the validators:

* When syncing to old forks we don't check if the downloaded block is already in the queue. This may cause the same block to be added to the queue multiple times on restart. 

* When receiving announcement for some known block we set common block for that peer to be the announced block. The code assumes that if the block is known it is lower than our best blocks. This is not true since reorgs to previous (lower) blocks is possible. Fixed that check.

### DIFF
--- a/client/network/src/protocol/sync.rs
+++ b/client/network/src/protocol/sync.rs
@@ -85,7 +85,7 @@ mod rep {
 	pub const INCOMPLETE_HEADER: Rep = Rep::new(-(1 << 20), "Incomplete header");
 
 	/// Reputation change for peers which send us a block which we fail to verify.
-	pub const VERIFICATION_FAIL: Rep = Rep::new(-(1 << 20), "Block verification failed");
+	pub const VERIFICATION_FAIL: Rep = Rep::new(-(1 << 29), "Block verification failed");
 
 	/// Reputation change for peers which send us a known bad block.
 	pub const BAD_BLOCK: Rep = Rep::new(-(1 << 29), "Bad block");
@@ -138,6 +138,8 @@ pub struct ChainSync<B: BlockT> {
 	block_announce_validator: Box<dyn BlockAnnounceValidator<B> + Send>,
 	/// Maximum number of peers to ask the same blocks in parallel.
 	max_parallel_downloads: u32,
+	/// Total numbet of processed blocks (imported or failed).
+	processed_blocks: usize,
 }
 
 /// All the data we have about a Peer that we are trying to sync with
@@ -318,6 +320,7 @@ impl<B: BlockT> ChainSync<B> {
 			is_idle: false,
 			block_announce_validator,
 			max_parallel_downloads,
+			processed_blocks: 0,
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


client/network/test/src/sync.rs
```diff
@@ -657,3 +657,40 @@ fn full_sync_requires_block_body() {
...
 	assert_eq!(net.peer(1).client.info().best_number, 0);
 }
+
+#[test]
+fn imports_stale_once() {
+	let _ = ::env_logger::try_init();
+
+	fn import_with_announce(net: &mut TestNet, hash: H256) {
+		// Announce twice
+		net.peer(0).announce_block(hash, Vec::new());
```
bkchr: Why do we announce it twice?

arkpar: This is a test for the case when duplicate announcement lead to stale block being imported twice.

client/network/src/protocol/sync.rs
```diff
@@ -138,6 +138,8 @@ pub struct ChainSync<B: BlockT> {
 	block_announce_validator: Box<dyn BlockAnnounceValidator<B> + Send>,
 	/// Maximum number of peers to ask the same blocks in parallel.
 	max_parallel_downloads: u32,
+	/// Total numbet of processed blocks (imported or failed).
```
marcio-diaz:
```suggestion
	/// Total number of processed blocks (imported or failed).
```

client/network/src/protocol/sync.rs
```diff
@@ -1168,7 +1180,8 @@ impl<B: BlockT> ChainSync<B> {
 
 	/// Restart the sync process.
 	fn restart<'a>(&'a mut self) -> impl Iterator<Item = Result<(PeerId, BlockRequest<B>), BadPeer>> + 'a {
-		self.queue_blocks.clear();
+		//self.queue_blocks.clear();
```
bkchr:
```suggestion
```
?

client/network/src/protocol/sync.rs
```diff
@@ -1094,9 +1106,9 @@ impl<B: BlockT> ChainSync<B> {
 		if let PeerSyncState::AncestorSearch(_, _) = peer.state {
 			return OnBlockAnnounce::Nothing
 		}
-		// If the announced block is the best they have seen, our common number
+		// If the announced block is the best they have and is not ahead of us, our common number
 		// is either one further ahead or it's the one they just announced, if we know about it.
-		if is_best {
+		if is_best && number > self.best_queued_number {
```
bkchr: Why don't we only update the common number when the number is higher than your best queued number?

client/network/src/protocol/sync.rs
```diff
@@ -357,6 +360,11 @@ impl<B: BlockT> ChainSync<B> {
 		self.fork_targets.len()
 	}
 
+	/// Number of processed blocks.
+	pub fn num_processed_blocks(&self) -> usize {
```
bkchr: We only need this in tests? :thinking:

client/network/src/protocol/sync.rs
arkpar: This is explained in description and the comment above.

> When receiving announcement for some known block we set common block for that peer to be the announced block. The code assumes that if the block is known it is lower than our best blocks. This is not true since reorgs to previous (lower) blocks is possible. Fixed that check.

client/network/src/protocol/sync.rs
arkpar: Yes, unfortunately there's no way to access this private data since the tests are in a separate crate.

PR #5278, opened 2020-03-17
libp2p-next

romanb:
Supersedes https://github.com/paritytech/substrate/pull/5066. This PR upgrades substrate to `libp2p-0.17`. The following are the main libp2p changes which are visible in substrate:

  * Adapt to https://github.com/libp2p/rust-libp2p/pull/1440 and https://github.com/libp2p/rust-libp2p/pull/1519.
  * Adapt to https://github.com/libp2p/rust-libp2p/pull/1472.
  * Adapt to https://github.com/libp2p/rust-libp2p/pull/1467.

### Overview

As described in the libp2p PRs, the underlying changes are primarily in `libp2p-core`. Realistically, a second connection to the same peer only occurs if two peers connect to each other "at the same time". As a side-effect, existing connections are also no longer closed in favour of new ones, which addresses https://github.com/paritytech/substrate/issues/4272.

### Details

  1. The enabled/disabled status is the same across all connections, as decided by the peerset manager.
  2. `send_packet` and `write_notification` always send all data over the same connection to preserve the ordering provided by the transport, as long as that connection is open. If it closes, a second open connection may take over, if one exists, but that case should be no different than a single connection failing and being re-established in terms of potential reordering and dropped messages. Messages can be received on any connection and thus two peers which happened to connect to each other simultaneously may each a different connection for sending data.
  3. The behaviour reports `GenericProtoOut::CustomProtocolOpen` when the first connection reports `NotifsHandlerOut::Open`.
  4. The behaviour reports `GenericProtoOut::CustomProtocolClosed` when the last connection reports `NotifsHandlerOut::Closed`.

In this way, the number of actual established connections to the peer is an implementation detail of the network behaviours. As mentioned before, in practice and at the time of this writing, there may be at most two connections to a peer and only as a result of simultaneous dialing. The network service even configures a hard limit of 2 connections per peer. However, the implementation in principle accommodates for any number of connections.

### Noteworthy

During intermediate testing with the (by default disabled) integration tests `test_consensus`, `test_sync` and `test_connectivity` it was revealed that when run in release mode these tests were very often failing, with the common symptom that _the last node to start in a round of testing_ would often see no other peers (i.e. empty DHT routing table) and thus make no progress while all the others keep on running, causing the tests to time out waiting for the problematic peer to reach a certain state. The tests are mainly using `add_reserved_peer` on the network to initialise the topology, however, `add_reserved_peer` ultimately results in a call to `add_known_peer` on the `DiscoveryBehaviour` which did not actually add that address to the Kademlia routing table, though it adds it to the `user_defined` peers which, when passed in the constructor of the behaviour, _are_ added to the Kademlia routing table. I thus changed `add_known_peer` to also add the given address to the Kademlia routing table and that resolved the issues with these integration tests and the `test_connectivity` test seems to run notably faster (release mode). My current guess is that the tests were so far unknowingly relying on a timing assumption w.r.t. the initial discovery / connection setup and DHT queries in order for all peers to find each other, in particular when simultaneous connections attempts are in play, as often happens in release mode. Ultimately, the change of letting `add_known_peer` add the given address to the Kademlia routing table may be a patch worth extracting separately, because it does look like an oversight to me.

## https://github.com/libp2p/rust-libp2p/pull/1472

Insubstantial changes to adapt to the new APIs.

## https://github.com/libp2p/rust-libp2p/pull/1467

Insubstantial changes to add now necessary feature flags where there is a dependency on `libp2p` with `default-features = false`.


romanb:
CC @tomaka, @twittner, @mxinden - Feel free to review at your leisure. This is obviously not mergeable until the next libp2p release but you may already want to skim the PR description.

client/network/src/debug_info.rs
```diff
@@ -205,26 +205,15 @@ impl NetworkBehaviour for DebugInfoBehaviour {
...
-			EitherOutput::Second(event) => self.identify.inject_node_event(peer_id, event),
-		}
-	}
-
-	fn inject_replaced(&mut self, peer_id: PeerId, closed_endpoint: ConnectedPoint, new_endpoint: ConnectedPoint) {
-		self.ping.inject_replaced(peer_id.clone(), closed_endpoint.clone(), new_endpoint.clone());
-		self.identify.inject_replaced(peer_id.clone(), closed_endpoint, new_endpoint.clone());
-
-		if let Some(entry) = self.nodes_info.get_mut(&peer_id) {
-			entry.endpoint = new_endpoint;
```
tomaka: The fact that this line disappeared made me realize that the `NetworkBehaviour` has no way to know if the `ConnectedPoint` passed with `inject_connected` is still in use.

tomaka: The right thing to do theoretically speaking would be to remove the `ConnectedPoint` parameter from `inject_connected` entirely. Unfortunately the `GenericProto` behaviour uses this value at the moment.

romanb: > The fact that this line disappeared made me realize that the `NetworkBehaviour` has no way to know if the `ConnectedPoint` passed with `inject_connected` is still in use.

What do you mean by "is still in use"? By whom? Do you mean that when `inject_connected` is called, the endpoint might actually concurrently be closed by the background task e.g. on error and `inject_disconnected` may not even be called if there happens to be another connection? So `inject_connected` and `inject_disconnected` can actually be called with different endpoints? That is an interesting observation, though I'm not sure what problems that may cause, if any (since `inject_disconnected` means there is _no_ connection remaining, it obviously implies that endpoint originally passed to `inject_connected` is also closed, but if only the endpoint originally passed to `inject_connected` closes, with another connection remaining, the behaviour will indeed not know about that). The reason `inject_connected` and `inject_disconnected` are only called for the first and last connection, respectively, is of course intentional for backward compatibility. I thought we would later add new callbacks that are called for every connection, e.g. `inject_connection_established` and `inject_connection_closed`. Then we could move the `ConnectedPoint` parameter to these methods, where it is assured that the calls are always paired up for the same connected point and thus if an implementation of `NetworkBehaviour` wants to use these, it must use these callbacks.

client/network/src/protocol/generic_proto/behaviour.rs
```diff
@@ -415,9 +450,15 @@ impl GenericProto {
 		protocol_name: Cow<'static, [u8]>,
 		message: impl Into<Vec<u8>>,
 	) {
-		if !self.is_open(target) {
-			return;
-		}
+		let conn = match self.peers.get(target).and_then(|p| p.get_open()) {
+			None => {
+				debug!(target: "sub-libp2p",
+					"Tried to sent notification to {:?} without an open channel.",
```
mxinden: What differentiates a channel from a substream? If they are interchangeable, could we settle on one term?

romanb: "channel" is the more abstract term, i.e. could be a connection or a substream. I don't really mind about the terminology in debug logs, I just thought that the behaviour shouldn't really know / care about the distinction, i.e. if a connection handler reports "open", then the behaviour knows there is an "open (communication) channel" to the peer via that connection, whatever that means in terms of substreams.

client/network/src/debug_info.rs
tomaka: The consequence for `debug_info.rs` is that we're tracking, for each connection, its corresponding `ConnectedPoint`. This tracking is used only for debugging purposes, so it's not a big deal.

However, in `light_client_handler.rs` for example, this is more problematic. This module is responsible for requesting information about blocks/justifications/etc. from the network.
To give you a summary:

- When `LightClientHandler::request` is called, the request is added to a list of pending requests.
- In `poll()`, we then send this request to the `ProtocolsHandler` of a specific peer which we know we're connected to. This is done by emitting a `SendEvent`.
- If everything goes well, the `ProtocolsHandler` will later send back an event containing the response.
- However if `inject_disconnected()` is called with a peer we have sent a request to, then we have to add back the request to the list of pending requests.

Now that we have multiple connections per peer, it possible for the `ProtocolsHandler` to be silently dropped, in which case the `NetworkBehaviour` has no way to know that it has to try again the request with a different peer.

> I thought we would later add new callbacks that are called for every connection, e.g. inject_connection_established and inject_connection_closed.

I don't think there's a way to solve that problem without tackling this.

romanb: > Now that we have multiple connections per peer, it possible for the `ProtocolsHandler` to be silently dropped, in which case the request is silently lost and the `NetworkBehaviour` has no way to know that.

That generally shouldn't be a problem though, because any such code that expects a reply to some request necessarily needs a timeout (+ bounded retry) mechanism, as seems to be the case for the [`LightClientHandler`](https://github.com/paritytech/substrate/blob/master/client/network/src/protocol/light_client_handler.rs#L862-L887). Requests can always fail by not getting a (timely) response. And any code that sends something without expecting a reply is by definition accepting potential delivery failure of the request.

tomaka: You're essentially suggesting that, even when we *know* that a request has failed, we instead just wait for the timeout.

romanb: > You're essentially suggesting that, even when we _know_ that a request has failed, we instead just wait for the timeout.

It is not a suggestion, just an observation that I don't think this scenario should cause any problems. It is ultimately of course desirable to be aware of failed requests as soon as possible, and that can be done once the behaviour has callbacks for each connection open/close.

tomaka:
I opened https://github.com/paritytech/polkadot/pull/963, which should fix the Polkadot compilation.

client/network/src/debug_info.rs
romanb: Since `libp2p-0.17` included the new callbacks, I addressed this concern in https://github.com/paritytech/substrate/pull/5278/commits/31cdf4ff6465dcf4be743341d092616637b60f06. TL;DR: The `light_client_handler` tracks the connections a request has been sent with and re-enqueues any outstanding requests on a connection as pending requests when that connection closes.

client/network/src/service.rs
```diff
@@ -308,9 +308,16 @@ impl<B: BlockT + 'static, H: ExHashT> NetworkWorker<B, H> {
...
-			let mut builder = SwarmBuilder::new(transport, behaviour, local_peer_id.clone());
+			let mut builder = SwarmBuilder::new(transport, behaviour, local_peer_id.clone())
+				.peer_connection_limit(crate::MAX_CONNECTIONS_PER_PEER);
 			if let Some(spawner) = params.executor {
-				builder = builder.executor_fn(spawner);
+				struct SpawnImpl<F>(F);
+				impl<F: Fn(Pin<Box<dyn Future<Output = ()> + Send>>)> Executor for SpawnImpl<F> {
+					fn exec(&self, f: Pin<Box<dyn Future<Output = ()> + Send>>) {
+						(self.0)(f)
+					}
```
romanb, line 318: I copied this here from `libp2p-core` because I forgot to re-expose `SwarmBuilder::executor_fn` before `0.17`, there is only `SwarmBuilder::executor`, and I wanted to avoid another libp2p release just for that.

client/network/src/service.rs
```diff
@@ -1045,21 +1052,35 @@ impl<B: BlockT + 'static, H: ExHashT> Future for NetworkWorker<B, H> {
...
+					trace!(target: "sub-libp2p", "Libp2p => IncomingConnectionError({},{}): {}",
+						local_addr, send_back_addr, error),
+				Poll::Ready(SwarmEvent::BannedPeer { peer_id, endpoint }) =>
+					trace!(target: "sub-libp2p", "Libp2p => BannedPeer({}). Connected via {:?}.",
+						peer_id, endpoint),
+				Poll::Ready(SwarmEvent::UnknownPeerUnreachableAddr { address, error }) =>
+					trace!(target: "sub-libp2p", "Libp2p => UnknownPeerUnreachableAddr({}): {}",
+						address, error),
+				Poll::Ready(SwarmEvent::ListenerClosed { reason, addresses: _ }) =>
+					trace!(target: "sub-libp2p", "Libp2p => ListenerClosed: {:?}", reason),
```
tomaka, line 1081:
```suggestion
					warn!(target: "sub-libp2p", "Libp2p => ListenerClosed: {:?}", reason),
```

client/network/src/discovery.rs
```diff
@@ -315,9 +321,9 @@ impl NetworkBehaviour for DiscoveryBehaviour {
 		NetworkBehaviour::inject_listener_error(&mut self.kademlia, id, err);
 	}
 
-	fn inject_listener_closed(&mut self, id: ListenerId) {
-		error!(target: "sub-libp2p", "Libp2p listener {:?} closed", id);
-		NetworkBehaviour::inject_listener_closed(&mut self.kademlia, id);
+	fn inject_listener_closed(&mut self, id: ListenerId, reason: Result<(), &io::Error>) {
+		error!(target: "sub-libp2p", "Libp2p listener {:?} closed: {:?}", id, reason);
```
tomaka, line 325: We can now remove this hack, as it's redundant with `service.rs`.

```suggestion
```

client/network/src/protocol/generic_proto/handler/notif_out.rs
```diff
@@ -293,16 +293,22 @@ impl ProtocolsHandler for NotifsOutHandler {
 						});
 						self.state = State::Opening { initial_message };
 					},
-					State::Opening { .. } | State::Refused | State::Open { .. } =>
-						error!("☎️ Tried to enable notifications handler that was already enabled"),
-					State::Poisoned => error!("☎️ Notifications handler in a poisoned state"),
+					st @ State::Opening { .. } | st @ State::Refused | st @ State::Open { .. } => {
+						debug!(target: "sub-libp2p",
```
tomaka, line 297: :eyes: I don't care enough about that change to argue again

client/network/src/service.rs
```diff
@@ -1045,21 +1052,35 @@ impl<B: BlockT + 'static, H: ExHashT> Future for NetworkWorker<B, H> {
...
-						{
-							error!(
-								"Connecting to bootnode with peer id `{}` and address `{}` failed \
-								because it returned a different peer id!",
-								peer_id,
-								address,
-							);
-						}
+					if this.boot_node_ids.contains(&peer_id)
+						&& error.contains("Peer ID mismatch")
```
tomaka, line 1056: This is a hack because the error used to be a `Box<dyn Error>`, while in libp2p 0.17 it is now an enum (#5369).
Feel free to fix this now, or someone can fix later separately.

client/network/src/protocol/light_client_handler.rs
```diff
@@ -1457,10 +1531,12 @@ mod tests {
 		let pset = peerset();
 		let mut behaviour = make_behaviour(true, pset.1, make_config());
 
-		behaviour.inject_connected(peer.clone(), empty_dialer());
+		behaviour.inject_connection_established(&peer, &ConnectionId::new(1), &empty_dialer());
+		behaviour.inject_connected(&peer);
```
tomaka, line 1535: Should these two lines not be the other way around?

client/network/src/protocol/generic_proto/behaviour.rs
```diff
@@ -163,33 +187,37 @@ enum PeerState {
...
-			PeerState::Incoming { .. } => false,
+			PeerState::Disabled { open, .. } |
+			PeerState::DisabledPendingEnable { open, .. } |
+			PeerState::Enabled { open, .. } =>
+				if !open.is_empty() {
+					Some(open[0])
+				} else {
+					None
+				}
+			_ => None
```
tomaka, line 220: I prefer when we have all the variants explicitly written, as it makes it easier to see that we didn't forget something.

```suggestion
			PeerState::Poisoned => None,
			PeerState::Banned { .. } => None,
			PeerState::PendingRequest { .. } => None,
			PeerState::Requested => None,
			PeerState::Incoming { .. } => None,
```

tomaka:
Would you mind merging master, so that we can test Polkadot with it?

client/network/src/protocol/light_client_handler.rs
romanb: I think this is right, see the [Swarm](https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/lib.rs#L542-L545) and [the docs](https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/behaviour.rs#L79-L80) (though the docs should read "when the _first_ connection to the peer is established ...", there is a typo there). It is the same order for `inject_connection_closed` / `inject_disconnected` as well as `inject_addr_reach_failure` / `inject_dial_failure`, i.e. the more granular callback is invoked first, followed by a conditional invocation of the second.

client/network/src/service.rs
romanb: I can do that here, this was just a blind code adaptation as part of a rebase.

client/network/src/protocol/light_client_handler.rs
romanb: Strictly speaking, for these tests, the calls to `inject_connected` are not even necessary since the impl of the behaviour is empty, I just added them for completeness and to not make the tests rely on such an empty impl but only on the contract `libp2p-swarm` gives for these callbacks.

romanb:
> Would you mind merging master, so that we can test Polkadot with it?

Just did so. For what it is worth, I already ran polkadot with it a few weeks ago, just to check that https://github.com/paritytech/substrate/issues/4272 is resolved with these changes, which seemed to be the case.

tomaka:
Deployed both on my Google Cloud instance (with incoming connections) around 3pm and on `kusama-validator-ew4-0` around 5pm.

There are a couple warnings (notably about reserved nodes getting disconnected), but that was already the case before. In particular, comparing the logs before and after this PR, the rate of all warnings seem exactly the same.

In Grafana everything seems normal as well. The number of GrandPa notifications emitted has increased, but that is expected due to #5520.

So everything looks good to me so far. I suggest to take a look tomorrow morning again and if there's nothing weird happening, we merge this.

tomaka:
Well, the test is kind of inconclusive.
It seems that Parity's logs server has crashed during the night, and my Google Cloud node stopped shortly after I posted the previous message without me realizing :man_facepalming:  (probably because I closed the SSH connection, forgetting that this would close the node (I'm not a devops)).

On Grafana, however, everything looks normal.

tomaka:
I'd still be in favour of merging.
@twittner?

tomaka:
Needs a merge/rebase over master.

tomaka:
I think the test is failing because of the unused import warning.

> warning: unused import: `channel::mpsc`
>   --> client/network/src/service.rs:39:27
>    |
> 39 | use futures::{prelude::*, channel::mpsc};
>   |                           ^^^^^^^^^^^^^
>   |

PR #5280, opened 2020-03-17
bulk child trie deletion

cheme:
This Pr is built upon #4827.

It implements bulk child deletion: instead of deleting every element of a child trie one by one with `kill_child_storage`, the deletion is done as a whole.
This means:
- only a single item is stored when deleting a child trie (the identifier of the child trie).
- A single change trie item is produce for the whole child trie deletion, this is done in a breaking way but I do not think change trie is currently use. I do not know if the item should appear in the digest too. cc @svyatonik
- on pruning deletion is done by using the key prefix used by rocksdb to avoid key collision. This Pr is needed for this : https://github.com/paritytech/parity-common/pull/360
- pruning journal an noncanonical journal need to store this change, and a different format is use. This change of format means that it may be possible to use the first approach of #4827 to gain a bit of space (pushing the rocksdb prefix to key value later in order to reduce journal key redundancy by managing changes per child trie). cc @arkpar 

This PR depending on #4827 first, I put its status in progress, but it should be usable for contract testing cc @Robbepop @pepyakin .


client/rpc-api/src/state/mod.rs
```diff
@@ -77,8 +77,6 @@ pub trait StateApi<Hash> {
 	fn child_storage_keys(
 		&self,
 		child_storage_key: StorageKey,
-		child_info: StorageKey,
```
tomusdrw: This is a breaking change to the RPC, we should probably plan some transition or add versioning, os that the JS libs can figure out which API version it's operating on.

cheme: This is more related to this PR https://github.com/paritytech/substrate/pull/4857/files#r394314720 ,  there was a PR pending from @Stefie on the js side, I think the initial idea was that the api was never really well defined and do it in a breaking way, but the more time passes the more probable it will be needed to be versioned, I will look at how it can be versioned (I feel like as long as this makes it to 2.0 versioning is not needed but I may be wrong and #4857 may not make it to 2.0).

tomusdrw: Fair enough! We have a RPC call that returns a list of all available RPC methods (and their versions), but it's hardcoded to `1` for now. We don't really have a good mechanism to control that, but it's definitely possible.
If the API is not available on JS side yet (or not being used extensively) then I suppose it's probably fine to change without version change before 2.0), but if not then we need some migration path. I remember that JS team (notably @jacogr) had a really hard time maintaining different compatibility mechanisms.

jacogr: Ok, the issue is we cannot "just break" - the pending PR will not go in. We support different chains and cannot detect among them what params there are and what params there are not.

So as it stands, we will breaks some and not others. Sadly, not something that is quite acceptable.

jacogr: And by "fix", I mean the simplest is actually just a rename. No other magic required.

cheme: Versioning will be better, I will have a look later. It is a tricky question to know how long deprecated interface would need support, maybe there is something to do by putting old api in some compatibility crate to avoid new project using them.
(Unlike runtime api there is no need to keep supporting old api forever)

Stefie: My PR is now a month old, now, so I also wouldn't just merge it anymore. I haven't touched it since I opened it.  I'll close it to avoid confusion.

arkpar:
But what does this actually achieve? 
There seem to be a lot of complicated code introduced all over the place. Some abstractions start leaking. Pruning, change tries  and the low level key-value database should not have a concept of a "child". It would be a totally different architecture. Either we move subtree management entirely to the bottom layer, or keep it out of it entirely. 

> It implements bulk child deletion: instead of deleting every element of a child trie one by one with kill_child_storage, the deletion is done as a whole.

What's wrong with the current method and is it the new implementation really worth the additional complexity?

cheme:
Thanks, for looking at it.

> But what does this actually achieve?

With a db like rocksdb which manage its keys and allow iteration on it, the target was pretty clear: delete a range of keys instead of deleting all trie contents.

With parity-db, this is not doable (and I think the gain from not having key ordering/storage is better), so the goods for this approach are:
- streamed child trie deletion (no need to put the whole child trie in memory).
- no deathrow management and smaller journals.
- fast block processing (for contract rent), with no drawback when runing archive node.

The bads are:
- child trie btreemap everywhere
- no recording of every child trie key in change trie (not sure it is a bad).


It is clearly different, not sure if it is worth the complexity.
Note that I struggle a bit yesterday to bring trie iteration at db level without messing with the design, but next I would probably just need to put the child trie root with the keyspace in the journals (and some usual debugging) in order to make the test pass with parity-db.


> There seem to be a lot of complicated code introduced all over the place. Some abstractions start leaking. Pruning, change tries and the low level key-value database should not have a concept of a "child". It would be a totally different architecture. Either we move subtree management entirely to the bottom layer, or keep it out of it entirely.

Yes that was what I had in mind, long term, bringing child info to the db (I guess this is bottom layer).
I rollback the journalling change of #4827, but since this pr requires this change it is indeed half baked, it would be more consistent if I would restore #4827 unprefixed journals (but at this point we are considering dropping it iiuc).

So yes, it splits management per child trie (#4827) to allow operation on child trie at the db level.

> What's wrong with the current method and is it the new implementation really worth the additional complexity?

The new implementation is only for parity db and subdb, there is still the old active on rocksdb, but if the target is to drop rocksdb for trie state it does not really matter.
As mentioned above, I am not sure if it is worth it. I wanted to bring some as efficient as possible deletion (only request was from contract side, but it was the gist of the rent design), but at the same time I agree on the code complexity point.

arkpar:
> With a db like rocksdb which manage its keys and allow iteration on it, the target was pretty clear: delete a range of keys instead of deleting all trie contents.

Ah, I see now that it uses `delete_prefix`.

It it really not clear at this point what storage scheme for subtrees will be more efficient. We have two approaches here:

1. Separate all subtrees in the storage by using unique keys for each subtree. This allows for faster removal of subtrees. But you have common subtree data duplicated in the database. 

2. Use reference counting. This eliminates duplicates but requires walking the trie when deleting.

It seems to me that you'd want fast deletion when you have lots of subtrees in the system. On the other hand when you have lot of subtrees, you'd also wan't them to share the same data and not blow up the state (as it happens in parity-ethereum right now). I'm inclined to think that storage saving is more important here. Also, there are some potential optimizations in parity-db to speed up iterating the trie, which may bring down deletion time difference somewhat.

cheme:
After weekend, I got a clearer opinion about this.
I guess the rocksdb delete by prefix is nice to have, but shouldn't be the main pros in the story.
I think parity-db will perform well enough by parsing the trie.
And by extension rc removal by parsing the trie too.
This is specific to the way the backend db work, and thus I see the operation required at db level.

The main thing that should be avoided by having a specific child trie deletion mechanism is the bloat of an unbounded number of key values removal, and the specific scenario I did not have in mind is the parachain state transition proof verification: the current child trie deletion will create a huge proof of individual key value removal when what is really interesting is to know that the whole child trie got removed.
It is the same for child trie, I did put in cons the fact that it doesn't register every child trie key removal, but I think it is definitively a big pro.

Does it really require reorganizing splitting the data payload by child trie, maybe not, we could have same per child trie management by child trie, send a transaction payload that do not split child trie state nodes, but got an additional list of BulkTrieDeletion: so instead of splitting all the payload we only split for the optional child deletion info. That would be a way to remove #4827 from this PR.
This PR would still need additional journal info, but we could use a new journal containing child trie deletion without touching the existing ones.
That is a possibility.

arkpar:
> the current child trie deletion will create a huge proof of individual key value removal when what is really interesting is to know that the whole child trie got removed.

As far as I understand there's no need to include deleted tries in the parachain block proof  anyway. They don't really affect block validity. The runtime would not walk the trie to delete it, it would still call a host function (externality) for that. The host function to delete a trie on the parachain validator would simply assign an empty storage root, without bothering to actually delete anything.

@rphmeier could you please confirm that ^

cheme:
Oh yes that could work indeed, no need to know what stuff was deleted, same goes for clear prefix and clear child prefix. That also could be a neat optimization of the proof in current execution proof, I will add that on TODO list (just need to delete up to the first node containing prefix and not register the other ones and in fact no need go through them if a new primitive is created in trie crate, for child trie kill it is even more straight forward indeed). Would also need to make sure the proof check don't try to access those values.

cheme:
This makes me wish that partial key were stored in the parent branch so same optimization could be achievable at branch level (still doable if we know that deleted keys are direct descendant (0 length partial key) of a branch and we delete all the child in the bitmap, but it is probably rather small gain in this case), but it is starting to be late I am probably missing something here.

rphmeier:
> > the current child trie deletion will create a huge proof of individual key value removal when what is really interesting is to know that the whole child trie got removed.
> 
> As far as I understand there's no need to include deleted tries in the parachain block proof anyway. They don't really affect block validity. The runtime would not walk the trie to delete it, it would still call a host function (externality) for that. The host function to delete a trie on the parachain validator would simply assign an empty storage root, without bothering to actually delete anything.
> 
> @rphmeier could you please confirm that ^

That sounds right. For any stateless client we only need to compute the main trie root as part of the validation function. So the approach of setting the child trie root to an empty root will work. In this case it would happen via an emulated host function within the parachain Wasm blob.  The full nodes which generate this proof and follow the chain that is approved by Polkadot would need to delete the child trie keys from their DB (actually, schedule them for deletion in the state pruning system)

gnunicorn:

cheme:
I did start implementing the possibility to remove deletion from proof like describe in previous comment https://github.com/paritytech/substrate/compare/master...cheme:proof_backends .

The code is incomplete (no handle of child trie prefix deletion), and need many tests, but enough to formulate a few  intersting observation.


I end up with three kind of proof construction and two kind of proof verification, when with the bulk trie deletion apprach we would only consider not registering state change for performance, but proof will be fine by nature.


- proof that do not need to trace deletion that are bellow a parent deletion, and therefore can produce shorter proof:
	- A first variant that do not process the transaction. (`WithoutTransaction`)

This should be the default choice unless we need to use the transactions (performance).
The implementation uses a different logic at Ext level and additional filtering in the overlay change (deleted child trie and deleted prefix).
On root calculation we simply calculate child trie root on a empty backend.
Note that for the prefix case, we are using a hack to insert the unitary changes lazilly. That does not achieve the real goal (on proof calculation we will register the access to the unitary element), and new trie primitive would be needed to only change the prefix node. Still the lazy nature of this hack benefits proofs that do not call storage root.

- 
    - A second variant that process the transaction but do not record it. (`WithTransaction`)

The implementation simply switch of the proof node recording during prefixed key access and during child trie key access on delete prefix and delete child trie operation.
But internally for storage root calculation it needs to work the same way as the previous variant and suffer from the same issue on prefix case.


Both variant needs to be verified with a verifier that do not query child trie nodes on deletion, that is the same change in processsing than the first variant.
The drawback here is that an host function could not be use in the proof (except if we do not prove its value as it is currently doable with Core_execute_block behavior), changes_storage_root that compute the change trie root: **getting those deletion out of the proof we cannot check change trie**, that can be acceptable (note that with bulk deletion approach all will be fine because change trie do not expect those deletion entries).


- proof that need to all deletion access (the existing code where we register all trie node touched):

Interesting for testing.
Interesting for proving a change trie. Should parachain verify their change trie in polkadot, that does not seem very obvious to me (at first I was thinking 'no use', but we also probably want to trust change trie).


So my gist out of this branch would be to say that it is doable to skip unneeded deleted content in proof, but rely on the assumption that change trie root will never be read during block processing.
Generally I find it better to have a special change for trie deletion (and by extension prefix deletion), and do not see disadvantage from having it (except complexity but if we got some specific proof generation process that is not true at all).
Benefit being lighter change trie and smaller memory footprint.

In both scenario (the linked branch and this PR approach), delete by prefix optimization would require additional tools (radix trie map and specific trie update operation, plus some reverse deathrow logic on non archive node (this would be needed for bulk child trie deletion too if we would allow using a child trie again after deletion)).

athei:
What is the status here?
@cheme You need feedback (if yes by whom)?

I am asking because we need this for #6357.

cheme:
It is frozen, I need a feedback on wether we want to make child trie deletion (and by extension possibly prefix deletion) a first class citizen operation (currently we got insert key value and delete key in changesets).

I mean @rphmeier did reply that we would not include child trie key in proofs, so I started implemented that, but end up having complicated redundant code (state machine need to behave differently depending on context), and I still think that registering a specific child trie deletion operation is the easier and cleaner thing to do.


From a quick look at the changes, things that could be trimmed:
- remove the InnerHash code, it is an unrelated optimisation
- revert #4827 change (but we still need to add an additional child trie related info in client and organising by child trie makes thing a bit more evolutive (bringing child_info into the client db layer to allow different backend depending on child type)). This relates to discussion with @arkpar and what is referred too as a 'totally different architecture'.

Only way I see to avoid child info in client would be to create a new storage change that contains the bulk child deletion with either keyspace for rocksdb delete or fetched child trie node hash for parity db (but it would means we fetch child trie content at the time of block processing which makes things awkward with proof, and make use store useless journal of key when we only need root).
In this case that would be the db type that will be leaking in state machine, and we need to avoid fetching key when running a child deletion in statemachine for proof, so I think this idea is a bad move.

athei:
You outlined multiple possible solutions. You have problems deciding which to choose or why is it frozen? @rphmeier @arkpar can you comment which one you would prefer? I have to little knowledge on this topic to fully comprehend the answer.

arkpar:
I still think subtree deletion should not leak to state-db abstraction. It should really be backend agnostic, and not rely on prefixes or backend-provided identifiers. ParityDb uses reference-counting, so that same node could be shared by multiple tries. Saving a lot of space potentially. This does not play well with prefixed keys. The most generic way to implement it is to walk the tree and delete each individual key. Avoiding any special handling of this operation.  I'd suggest we don't do any DB-optimized deletion until it is shown to be a problem in practice.

Regarding proofs: Pausing recording during the deletion operation looks like the right thing to do. It is an optimization though. The validator would not access nodes when it validates deletion, so there's no need to provide them. I'm not sure I understand the issue with the change trie. How does skipping collecting proof for deletion affects the change trie root? As far as I know, change tries are only implemented for the main tree.

The long-term plan for ParityDB is to evolve to support trie-related operations. That would involve not just deletion, but also reference-counted insertion and iteration at the DB level. Until then we'd better keep it simple.

rphmeier:
I feel too disconnected from this issue and the underlying DB roadmap to meaningfully comment. I believe the only place I have weighed in is [here](https://github.com/paritytech/substrate/pull/5280#issuecomment-620371695) upon this request from @arkpar to clarify needs of some higher-level protocol. I defer to @arkpar for comments on the general approach.

athei:
> The most generic way to implement it is to walk the tree and delete each individual key. Avoiding any special handling of this operation. I'd suggest we don't do any DB-optimized deletion until it is shown to be a problem in practice.

This is basically what we have now, right? You argue for keeping child trie deletion a `O(n)` operation?

cheme:
> ParityDb uses reference-counting, so that same node could be shared by multiple tries. Saving a lot of space potentially. This does not play well with prefixed keys.

The PR do not use prefixed keys in this case: for rocksdb the drop tree operation contains both child trie root and keyspace, then only in the case of rocksdb we try to apply a delete by prefix, parity-db still iterate on trie but only when pruning is called.
At first I thought it would be useless with parity-db, but the good things are:
- only keeping trace of child root in the pruning journal.
- in case of an archive node it means that we never iterate the child trie at all
- smaller change trie (but a bit less info which in this case looks good to me)
Still those point are either optimization (I don't expect the size of the journal to be problematic), or breaking (change trie).

> I'd suggest we don't do any DB-optimized deletion until it is shown to be a problem in practice

Using trie iteration for rocksdb (the same way as parity-db) seems doable, it will then only helps with the previous point.

> Pausing recording during the deletion operation looks like the right thing to do. 

That is what I did in https://github.com/paritytech/substrate/compare/master...cheme:proof_backends , it means that there is different mode of execution for state-machine, one that produce the transaction payload, and one that don't (IIRC I did add a third one for doing both (tx payload and proof not recording ), I believe it is what is needed for cumulus).
I don't remember too well why I was really not happy with this branch, maybe bringing the same logic for delete_with_prefix was a bit too much, or I just end up seeing the new assumption with change trie and other things as more complicated than having a delete trie change set operation. 

> I'm not sure I understand the issue with the change trie.

Change trie root can be calculated from a storage host function so if a parachain uses it in its runtime, it would need to include all deletion in its proof (so running the proof add all the child trie deletion in change trie).
In practice the value of the change trie root would probably be directly send to the header without being use internally by the state machine and not stored in the state trie, but it is still possible to do.
Similarily including individual child trie deletion in change trie means that if we want to proof child trie the proof will allways be way bigger (as is the child trie)

> As far as I know, change tries are only implemented for the main tree.

I reply to quickly to previous point, change trie do also include child trie changes.

> The long-term plan for ParityDB is to evolve to support trie-related operations. That would involve not just deletion, but also reference-counted insertion and iteration at the DB level. Until then we'd better keep it simple.

Sounds interesting, I will try to keep looking at parity-db branches and issues if there is.

cheme:
> You argue for keeping child trie deletion a O(n) operation?

Note that in the parity-db case we still got a O(n) trie parsing removal operation. Regarding the perf gains the trade off seems fine.
The issue I see, and I think is your concern regarding rent system, is that we iterate during block evaluation when with this PR we only need to iterate on pruning (which makes sure this O(n) cost only happen for canonical and is easier to separate or reschedule).

athei:
The issue I have with `O(n)` child trie removal is when contracts are evicted due to rent. If the contract has a huge amount of accumulated storage this eviction might be even too large for one block. We then need to implement something like lazy removal that can distribute this across multiple blocks.

arkpar:
> > The most generic way to implement it is to walk the tree and delete each individual key. Avoiding any special handling of this operation. I'd suggest we don't do any DB-optimized deletion until it is shown to be a problem in practice.
> 
> This is basically what we have now, right? You argue for keeping child trie deletion a `O(n)` operation?

It is O(n) best case anyway. The only difference is with rocksdb you get O(n) on commit, and not on the actual delete operation. Unless you keep each subtree in a separate database. We can indeed optimize to perform it in the background on the database level but I'd argue it is too early for that.

arkpar:
I think it does not make much sense for change tries to collect all removed keys for subtries. If it does, there's your O(n) right there and there's little sense to optimize it in the database, since you have to do iteration anyway.  
There needs to be a single event recorded in the parent tree that would allow to track that all subtree data has been removed.

athei:
I see. Thanks for your answers. We need then cope with the fact that eviction is linear time operation.

Robbepop:
From my current understanding if we cannot get around the fact that contract eviction is linear time we should at least make it possible to perform the clean-up lazily or defer it when possible or break it up into chunks of clean-up. Otherwise we end up having no way to clean-up contract instances that are especially big. Might this even be an attack vector in certain cases? Maybe there are other operations that need to be deferred or chunked up into pieces to be distributed among several blocks so that we could find a generalized solution to this problem?

gnunicorn:

cheme:
> There needs to be a single event recorded in the parent tree that would allow to track that all subtree data has been removed.

Using a specific child trie delete operation does inherently cover this point.

Closing note:
bulk child trie deletion seems still very relevant to me.
Though it is more in line with an architecture where each child trie are managed independently (which is needed for multiple other things and open perspective for other kind of state).
So in line with #4827 I feel like this will be needed in the future (I feel like the switching off deletion when registering proof is more complex than this approach).

cheme:
Also for https://github.com/paritytech/substrate/issues/6594 having a single child trie delete info send back to the client makes the non O(1) cost only happen at pruning (or never for archive nodes), and it could be easier to implement this way.

cheme:
Note, this PR lacks a correct trie root calculation in state-machine, code such as https://github.com/cheme/trie/blob/379bf92720e23b2465badd4739c7d2ba75115c06/trie-db/src/traverse.rs#L658 (old code please don't use), mainly have a detach branch from trie action (can be implemented in triedbmut too, but more involve).

PR #5281, opened 2020-03-17
Extend PerThing + Saturating

sjeohp:
* Add `saturating_pow` to `Saturating`
* Add `mul_floor` to `PerThing` (renamed from `mul_collapse`)
* Add `mul_ceil` to `PerThing`
* Add `saturating_reciprocal_mul` & `_floor` / `_ceil` to `PerThing`
* Fix accuracy of `from_percent` when the type's denominator is not divisible by 100 (eg. `Peru16`)
* Provide default methods where possible


primitives/arithmetic/src/fixed64.rs
```diff
@@ -20,7 +20,8 @@ use sp_std::{
 };
 use codec::{Encode, Decode};
 use crate::{
-	Perbill,
+    Perbill,
```
bkchr: Spaces.

primitives/arithmetic/src/per_things.rs
```diff
@@ -159,39 +161,22 @@ macro_rules! implement_per_thing {
 			}
 		}
 
-		/// Implement const functions
-		impl $name {
-			/// From an explicitly defined number of parts per maximum of the type.
-			///
-			/// This can be called at compile time.
-			pub const fn from_parts(parts: $type) -> Self {
```
bkchr: The idea was that this is a const function

primitives/arithmetic/src/per_things.rs
```diff
@@ -159,39 +161,22 @@ macro_rules! implement_per_thing {
...
-				Self([parts, $max][(parts > $max) as usize])
-			}
-
-			/// Converts a percent into `Self`. Equal to `x / 100`.
-			///
-			/// This can be created at compile time.
-			pub const fn from_percent(x: $type) -> Self {
```
bkchr: Same.

primitives/arithmetic/src/per_things.rs
sjeohp: Yep I see that now

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +576,67 @@ macro_rules! implement_per_thing {
...
 				);
 			}
+
+			#[test]
+			fn saturating_pow_works() {
+				// x^0 == 1
+				assert_eq!(
+					$name::from_parts($max / 2).saturating_pow(0), 
+					$name::from_parts($max)
+					);
```
kianenigma:
```suggestion
				);
```

kianenigma: and all the rest

primitives/arithmetic/src/per_things.rs
```diff
@@ -200,6 +210,26 @@ macro_rules! implement_per_thing {
...
+					self
+				} else {
+					let p = <$name as PerThing>::Upper::from(self.deconstruct());
+					let q = <$name as PerThing>::Upper::from(Self::ACCURACY);
+					let mut s = Self::one();
+					for _ in 0..exp {
+						if s.is_zero() {
+							break;
+						} else {
+							s = Self::from_rational_approximation(<$name as PerThing>::Upper::from(s.deconstruct()) * p, q * q);
```
kianenigma: needs to comment why mul is not overflowing here.

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
...
+					$name::from_parts($max / 2).saturating_pow(2usize.pow(31)), 
+					$name::from_parts(0)
+					);
+			}
+
+			#[test]
+			fn saturating_reciprocal_mul_works() {
+				assert_eq!(
+					$name::from_parts($max).saturating_reciprocal_mul(<$type>::from(10u8)),
+					10
```
bkchr:
```suggestion
					10,
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
...
+					);
```
bkchr:
```suggestion
				);
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
...
+				assert_eq!(
+					$name::from_parts($max / 2).saturating_reciprocal_mul(<$type>::from(10u8)),
+					20
```
bkchr:
```suggestion
					20,
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
...
+					);
```
bkchr:
```suggestion
				);
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
...
+				assert_eq!(
+					$name::from_parts(1).saturating_reciprocal_mul($max),
+					<$type>::max_value()
+					);
```
bkchr:
```suggestion
					<$type>::max_value(),
				);
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
 					$name::from_percent(100)
 				);
 			}
+
+			#[test]
+			fn saturating_pow_works() {
+				// x^0 == 1
+				assert_eq!(
+					$name::from_parts($max / 2).saturating_pow(0), 
+					$name::from_parts($max)
```
bkchr:
```suggestion
					$name::from_parts($max),
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
...
+				);
+
+				// x^1 == x
+				assert_eq!(
+					$name::from_parts($max / 2).saturating_pow(1), 
+					$name::from_parts($max / 2)
+					);
```
bkchr:
```suggestion
					$name::from_parts($max / 2),
				);
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
...
+
+				// x^2
+				assert_eq!(
+					$name::from_parts($max / 2).saturating_pow(2), 
+					$name::from_parts($max / 2).square()
+					);
```
bkchr:
```suggestion
					$name::from_parts($max / 2).square(),
				);
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
...
+
+				// x^3
+				assert_eq!(
+					$name::from_parts($max / 2).saturating_pow(3), 
+					$name::from_parts($max / 8)
+					);
```
bkchr:
```suggestion
					$name::from_parts($max / 8),
				);
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
...
+
+				// 0^n == 0
+				assert_eq!(
+					$name::from_parts(0).saturating_pow(3), 
+					$name::from_parts(0)
+					);
```
bkchr:
```suggestion
					$name::from_parts(0),
				);
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
...
+
+				// 1^n == 1
+				assert_eq!(
+					$name::from_parts($max).saturating_pow(3), 
+					$name::from_parts($max)
+					);
```
bkchr:
```suggestion
					$name::from_parts($max),
				);
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -545,6 +581,67 @@ macro_rules! implement_per_thing {
...
+
+				// (x < 1)^inf == 0 (where 2.pow(31) ~ inf)
+				assert_eq!(
+					$name::from_parts($max / 2).saturating_pow(2usize.pow(31)), 
+					$name::from_parts(0)
+					);
```
bkchr:
```suggestion
					$name::from_parts(0),
				);
```

primitives/arithmetic/src/per_things.rs
```diff
@@ -30,29 +30,55 @@ pub trait PerThing: Sized + Saturating + Copy {
 	/// The data type used to build this per-thingy.
 	type Inner: BaseArithmetic + Copy;
 
-	/// accuracy of this type
-	const ACCURACY: Self::Inner;
+	/// A data type larger than `Self::Inner`, used to avoid overflow in some computations.
```
thiolliere:
```suggestion
	/// A data type larger than `Self::Inner`, used to avoid overflow in some computations.
	/// It must be able to compute `ACCURACY^2`.
```

This is needed in the implementation of `fn square`

primitives/arithmetic/src/per_things.rs
```diff
@@ -30,29 +30,55 @@ pub trait PerThing: Sized + Saturating + Copy {
...
+	fn square(self) -> Self {
+		let p = Self::Upper::from(self.deconstruct());
+		let q = Self::Upper::from(Self::ACCURACY); 
+		Self::from_rational_approximation(p * p, q * q)
+	}
+
+	/// Saturating reciprocal multiplication. Compute `x / self`, saturating at the numeric 
+	/// bounds instead of overflowing.
+	fn saturating_reciprocal_mul<N>(self, x: N) -> N
+	where N: Into<Self::Upper> + UniqueSaturatedFrom<Self::Upper> {
```
thiolliere: note that the bound `Into<Self::Upper>` is a strong constrain, for example usual arithmetic used in substrate are just implementing the trait AtLeast32Bit, and are generally more than u32.

On the other side Perbill or other per_things are usually using u64 as Upper type. so this is limiting the usage.

primitives/arithmetic/src/per_things.rs
```diff
@@ -30,29 +30,55 @@ pub trait PerThing: Sized + Saturating + Copy {
...
+		N::unique_saturated_from(x.into() * Self::Upper::from(Self::ACCURACY) / Self::Upper::from(self.deconstruct()))
```
thiolliere: `x.into() * Self::Upper::from(Self::ACCURACY)` can overflow

primitives/arithmetic/src/per_things.rs
```diff
@@ -30,29 +30,55 @@ pub trait PerThing: Sized + Saturating + Copy {
...
+	}
+
+	/// Saturating truncating multiplication. Compute `x * self`, truncating any fractional
+	/// remainder, IE rounding down to the nearest integer.
+	fn saturating_truncating_mul<N>(self, x: N) -> N 
+	where N: Into<Self::Upper> + UniqueSaturatedFrom<Self::Upper> {
+		N::unique_saturated_from(x.into() * Self::Upper::from(self.deconstruct()) / Self::Upper::from(Self::ACCURACY))
```
thiolliere: `x.into() * Self::Upper::from(self.deconstruct())` can overflow

primitives/arithmetic/src/per_things.rs
sjeohp: Changed to `UniqueSaturatedInto<Self::Inner>`. Better you think?

thiolliere: I don't think so, this saturation is very likely to happen for lots of value I guess.
for instance a currency amount can likely not fit into u64.

Maybe there is a way to solve it similarly to `fn mul` or maybe it is better to actually compute everything into the arithmetic space of `N`

thiolliere: I'm ok to merge with `Into<Self::Inner>` just I feel like maybe we could improve this

thiolliere: also I'm pretty we can make a similar implementation as `fn mul`

primitives/arithmetic/src/per_things.rs
```diff
@@ -30,29 +30,62 @@ pub trait PerThing: Sized + Saturating + Copy {
...
+		}
+	}
+
+	/// Saturating truncating multiplication. Compute `x * self`, truncating any fractional
+	/// remainder, IE rounding down to the nearest integer. 
+	fn saturating_truncating_mul<N>(self, x: N) -> N 
+	where N: UniqueSaturatedInto<Self::Inner> + UniqueSaturatedFrom<Self::Upper> {
+		let x = x.saturated_into::<Self::Inner>();
+		N::unique_saturated_from(Self::Upper::from(x) * Self::Upper::from(self.deconstruct()) / Self::Upper::from(Self::ACCURACY))
+	}
```
thiolliere: here as well I think `Into<Self::Inner>` would be better, also this is a fast subset implementation of trait `Mul` maybe we should put in the comment that for other usage one could use `Mul` instead.

primitives/arithmetic/src/per_things.rs
sjeohp: > I don't think so, this saturation is very likely to happen for lots of value I guess.
> for instance a currency amount can likely not fit into u64.

I don't think this is true. If you are handling values that regularly saturate `Inner` you should choose a bigger `Inner`, or implement one.

sjeohp: > also I'm pretty we can make a similar implementation as `fn mul`

What aspects of the `mul` implementation would improve this one?  The `mul` rounds to the nearest part, whereas these ones are meant to round down.

sjeohp: I could add various other operations, (eg. checked as well as saturating), if that would be useful, but I don't know if anyone needs them so could probably go in a later PR.

thiolliere: > > I don't think so, this saturation is very likely to happen for lots of value I guess.
> > for instance a currency amount can likely not fit into u64.
> 
> I don't think this is true. If you are handling values that regularly saturate `Inner` you should choose a bigger `Inner`, or implement one.

Perbill is used a lot in the code base with Currency and in kusama currency is u128 and usual amount does overflow u32 (`u32== Perbill::Inner`). And I don't think this is a wrong usage

thiolliere: > > also I'm pretty we can make a similar implementation as `fn mul`
> 
> What aspects of the `mul` implementation would improve this one? The `mul` rounds to the nearest part, whereas these ones are meant to round down.

implementation of `Mul` doesn't saturate the input `x: N` to u32 it computes an accurate multiplication (with rounding stuff) and doesn't saturate (as Per_things are for now in [0, 1]).

if I use some value in u128 and want to multiply with some Perbill the implementation Mul give a correct result whereas this implementation is very inaccurate if the input is more than `2^32`

primitives/arithmetic/src/per_things.rs
```diff
@@ -64,12 +112,43 @@ pub trait PerThing: Sized + Saturating + Copy {
 	/// `M` as the data type that can hold the maximum value of this per-thing (e.g. u32 for
 	/// perbill), this can only work if `N == M` or `N: From<M> + TryInto<M>`.
 	fn from_rational_approximation<N>(p: N, q: N) -> Self
-		where N: Clone + Ord + From<Self::Inner> + TryInto<Self::Inner> + ops::Div<N, Output=N>;
+	where N: Clone + Ord + From<Self::Inner> + TryInto<Self::Inner> + ops::Div<N, Output=N>;
+}
+
+/// Helper function to calculate a remainder used in `PerThing` multiplications.
+fn rem_helper<N, Inner, Upper>(x: N, part: Inner, acc: Inner, reciprocal: bool, truncate: bool) -> Upper
```
kianenigma: could use a better name that tells what this function does.

primitives/arithmetic/src/per_things.rs
```diff
@@ -64,12 +112,43 @@ pub trait PerThing: Sized + Saturating + Copy {
...
+}
+
+/// Helper function to calculate a remainder used in `PerThing` multiplications.
+fn rem_helper<N, Inner, Upper>(x: N, part: Inner, acc: Inner, reciprocal: bool, truncate: bool) -> Upper 
+where 
+	N: Clone + From<Inner> + UniqueSaturatedInto<Inner> + ops::Div<N, Output=N> + ops::Mul<N, Output=N> + ops::Add<N, Output=N> + ops::Rem<N, Output=N>,
+	Inner: BaseArithmetic + Copy,
+	Upper: BaseArithmetic + Copy + From<Inner> + TryInto<Inner>,
+{
+	let (numer_upper, denom_n, denom_upper): (Upper, N, Upper) = if reciprocal {
```
kianenigma:
```suggestion
	let (number_upper, denom_n, denom_upper): (Upper, N, Upper) = if reciprocal {
```

primitives/arithmetic/src/per_things.rs
kianenigma: Also, once you name it better it because clear what this function is doing, I imagine the parameters could also use a re-think. I don't think it is the responsibility of this function (if it is merely `rem`ing two numbers) to know about `reciprocal: bool`.

sjeohp: Good point. the caller could just swap `part` and `acc` to get the same result

primitives/arithmetic/src/per_things.rs
```diff
@@ -64,12 +112,43 @@ pub trait PerThing: Sized + Saturating + Copy {
...
+		(acc.into(), part.into(), part.into())
+	} else {
+		(part.into(), acc.into(), acc.into())
+	};
+	let rem = x.clone().rem(denom_n);
+	// `rem_sized` is inferior to $max, thus it fits into $type. This is assured by
```
kianenigma:
```suggestion
	// `rem_sized` is inferior to $max, thus it fits into `Inner`. This is assured by
```

kianenigma: These documents are now a bit outdated. I am not sure what's the best way to update them though.  Maybe replace all and then name that `Inner` is always `$type` in macro expansion and `Upper` is `$upper`. 

But then someone can mess with this function and call it like 

```
rem_helper::<N, u32, u16>(
	x.clone(), 
	self.deconstruct(), 
	Self::ACCURACY, 
	false, 
	truncate,
)
```

which breaks all the assumptions. 

Hence the function must accept a generic type `P: PerThing` and then use the types `P::Upper` and `P::Inner`. Then we can assume that these assumptions hold

sjeohp: >These documents are now a bit outdated.

I've rewritten them, they should make sense again. Let me know if they don't.

> Hence the function must accept a generic type P: PerThing and then use the types P::Upper and P::Inner. Then we can assume that these assumptions hold

👍

primitives/arithmetic/src/per_things.rs
sjeohp: I've changed `saturating_reciprocal_mul` to use the same remainder calculation as `mul` so it only saturates `N`.  Great suggestion, thanks

primitives/arithmetic/src/per_things.rs
```diff
@@ -200,6 +258,31 @@ macro_rules! implement_per_thing {
...
+			/// Saturating exponentiation. Computes `self.pow(exp)`, saturating at the numeric
+			/// bounds instead of overflowing. This operation is lossy.
+			fn saturating_pow(self, exp: usize) -> Self {
+				if self.is_zero() || self.is_one() {
+					self
+				} else {
+					let p = <$name as PerThing>::Upper::from(self.deconstruct());
+					let q = <$name as PerThing>::Upper::from(Self::ACCURACY);
+					let mut s = Self::one();
+					for _ in 0..exp {
```
thiolliere: I guess this could be improved by keeping intermediary result, like `x^10 == ((x^2)^2)^2*x^2` but anyway seems ok

kianenigma:
Thank you for your patience :D

bkchr:
Please merge master to fix tests.

PR #5286, opened 2020-03-18
Print non-output messages to stderr in subkey

lwshang:
This PR has a very simple motivation that the `subkey` utility should only print "real" output to `stdout` while direct other messages like errors or vanity progress to `stderr`.

I only modified some of `println!` to `eprintln!`. And I only checked `subkey`.

### DIFF
--- a/bin/utils/subkey/src/main.rs
+++ b/bin/utils/subkey/src/main.rs
@@ -135,7 +135,6 @@ trait Crypto: Sized {
 					);
 				},
 			}
-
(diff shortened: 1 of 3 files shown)


lwshang:
[clabot:check]

bin/utils/subkey/src/main.rs
```diff
@@ -135,7 +135,7 @@ trait Crypto: Sized {
 					);
 				},
 			}
-			
+
```
bkchr:
```suggestion
```

PR #5287, opened 2020-03-18
Benchmark Treasury Pallet

shawntabrizi:
This creates benchmarks for the Treasury pallet measuring extrinsics and the `on_finalize` function.

- [x] propose_spend
- [x] reject_proposal
- [x] approve_proposal
- [x] report_awesome
- [x] retract_tip
- [x] tip_new
- [x] tip
- [x] close_tip
- [x] on_finalize

To make this all work, I had to introduce an extra function (`add`) to the `Contains` trait which is enabled only with the `runtime-benchmarks` feature flag. This allowed me to add new members to `elections_phragmen` to pass the Origin checks in the pallet.

Additionally, I updated the runtime to allow some functions to be callable by Root, again, simplifying the benchmarks without breaking any authorization assumptions.

Finally, I moved tests to a separate file to keep things a bit cleaner.


frame/elections-phragmen/src/lib.rs
```diff
@@ -784,6 +784,18 @@ impl<T: Trait> Contains<T::AccountId> for Module<T> {
 		Self::is_member(who)
 	}
 	fn sorted_members() -> Vec<T::AccountId> { Self::members_ids() }
+
+	#[cfg(feature = "runtime-benchmarks")]
+	// A special function to populate members in this pallet for passing Origin
+	// checks in runtime benchmarking.
```
thiolliere: insubstantial but in the codebase I think comment always go first, this is to avoid having the attributes far from the function signature in case the comments get very long.

```suggestion
	// A special function to populate members in this pallet for passing Origin
	// checks in runtime benchmarking.
	#[cfg(feature = "runtime-benchmarks")]
```

frame/support/src/traits.rs
```diff
@@ -178,6 +178,13 @@ pub trait Contains<T: Ord> {
 
 	/// Get the number of items in the set.
 	fn count() -> usize { Self::sorted_members().len() }
+
+	#[cfg(feature = "runtime-benchmarks")]
+	/// Add an item that would satisfy `contains`. It does not make sure any other
+	/// state is correctly maintained or generated.
+	///
+	/// **Should be used for benchmarking only!!!**
```
thiolliere: same for same reason
```suggestion
	/// Add an item that would satisfy `contains`. It does not make sure any other
	/// state is correctly maintained or generated.
	///
	/// **Should be used for benchmarking only!!!**
	#[cfg(feature = "runtime-benchmarks")]
```

frame/treasury/src/lib.rs
```diff
@@ -386,10 +393,11 @@ decl_module! {
 		/// # </weight>
 		#[weight = SimpleDispatchInfo::FixedOperational(100_000)]
 		fn approve_proposal(origin, #[compact] proposal_id: ProposalIndex) {
-			T::ApproveOrigin::ensure_origin(origin)?;
+			T::ApproveOrigin::try_origin(origin)
+				.map(|_| ())
+				.or_else(ensure_root)?;
```
thiolliere: I see, using origin to be ApproveOrigin or root instead of just ApproveOrigin allows it be called during benchmark.

Note that this pattern can only works if we don't make any use of the Success type.

(Maybe a broader solution can be found that would also be usable for when Success type is used. because such a broader solution will have to be used anyway for democracy for instance)

frame/treasury/src/benchmarking.rs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+const MAX_BYTES: u32 = 16384;
+const MAX_TIPPERS: u32 = 100;
+
+benchmarks! {
+	_ { }
+
+	propose_spend {
+		let u in 0 .. 1000;
```
thiolliere: Just to confirm: Is 1000 the expected order of magnitude of the number of proposal in storage Proposal ?

I think the performance related to insert from and removing to `Proposals` is related to the number of proposal.
In propose_send I guess this cost is always negligable compared to balance upgrade.
but if this number gets very high then this might no longer be small enough.

frame/treasury/src/benchmarking.rs
```diff
@@ -0,0 +1,211 @@
...
+	approve_proposal {
+		let u in 0 .. 1000;
+		let (caller, value, beneficiary_lookup) = setup_proposal::<T>(u);
+		Treasury::<T>::propose_spend(
+			RawOrigin::Signed(caller).into(),
+			value,
+			beneficiary_lookup
+		)?;
+		let proposal_id = ProposalCount::get() - 1;
+	}: _(RawOrigin::Root, proposal_id)
```
thiolliere: (just a note that the Approvals::mutate performance is also related to the size of Approvals value, but as ProposalIndex is u32, decoding and encoding should be fast enough anyway)

frame/treasury/src/benchmarking.rs
```diff
@@ -0,0 +1,211 @@
...
+
+	report_awesome {
+		let r in 0 .. MAX_BYTES;
+		let (caller, reason, awesome_person) = setup_awesome::<T>(r);
+	}: _(RawOrigin::Signed(caller), reason, awesome_person)
```
thiolliere: This also depends on how much Tips and Reasons storage are full. I guess we should make the assumption that all those storage: Proposal, Reasons, Tips, have all negligible performance variation wether they are full or not because they will not hold that much value anyway.

thiolliere:
I'm not sure about what direction to take a bout making benchmark with some storage which their size is not supposed to be changing the weight.

Maybe a good direction would be to document what are the expected maximum size of those storages (by size I mean the number of keys in a StorageMap), and do the benchmark by filling those storage with this size.

shawntabrizi:
@thiolliere I actually created a project for just that:

https://github.com/shawntabrizi/substrate-benchmark-genesis/

This will allow us to generate a chain spec file which pre-populates the storage maps and double maps with values to simulate there being lots of state already on the chain. For each of those, we can configure how many storage items we want to add to the genesis.

Then, when running benchmarks, we will set each of these to their "practical max value" and see the results of the benchmarks.

This means that we need not set up any "existing" state unless we are trying to measure the change in performance over that change in state.

frame/treasury/src/lib.rs
shawntabrizi: This was a solution that @gavofyork said would be okay for these situations. Otherwise, we may need to extend traits like I have done for `Contains`. I dunno about any other better solutions :/

But I am very open to ideas!

frame/treasury/src/benchmarking.rs
shawntabrizi: u in this case only affects the seed of the user account that is making the call, which is negligible. This test does not take into account the number of pre-existing items in this storage trie, but as I mentioned below, we will take care of that via a special genesis file.

Relatively speaking (from within the perspective of the code) this function is constant time, so we benchmark it as so.

frame/treasury/src/benchmarking.rs
shawntabrizi: Can you clarify? You are saying larger u32 values take longer to decode?

frame/treasury/src/benchmarking.rs
shawntabrizi: See my other comments

frame/treasury/src/benchmarking.rs
thiolliere: `Vec<u32>` is shorter to decode and encode because it will just transmute to `Vec<u8>` whereas `Vec<Something different than uX or iX>` will probably take longer because it will decode/encode each element from/into another `Vec<u8>`

PR #5288, opened 2020-03-18
Regression algorithm for runtime benchmarks

marcio-diaz:
(In progress)

I think the code of Gav is good, so I'll just add tests/graphs to compare both algorithms and do some cleaning.
Also, we need to see the best output for the scripts that will run it.


shawntabrizi:
Have you tested this with some real benchmarks? Can you post the full results here?

PR #5289, opened 2020-03-18
Simplify BABE pre-digest definitions

sorpaas:
Rework of #4974. Extract VRF into a separate crate and unify `PreDigest` / `RawPreDigest` definition by using generic parameter.


frame/babe/src/lib.rs
```diff
@@ -216,7 +214,7 @@ impl<T: Trait> FindAuthor<u32> for Module<T> {
 	{
 		for (id, mut data) in digests.into_iter() {
 			if id == BABE_ENGINE_ID {
-				let pre_digest = RawPreDigest::decode(&mut data).ok()?;
+				let pre_digest: RawPreDigest = RawPreDigest::decode(&mut data).ok()?;
```
JoshOrndorff: Asking for my own learning: Why is the explicit type annotation necessary?

primitives/consensus/babe/src/digests.rs
```diff
@@ -18,175 +18,101 @@
...
-		/// VRF output
-		vrf_output: VRFOutput,
-		/// VRF proof
-		vrf_proof: VRFProof,
 		/// Authority index
 		authority_index: super::AuthorityIndex,
 		/// Slot number
 		slot_number: SlotNumber,
+		/// VRF output
+		vrf_output: VRFOutput,
```
rphmeier: This breaks compatibility with existing chain such as Kusama, no?

primitives/consensus/babe/src/digests.rs
```diff
@@ -18,175 +18,101 @@
...
 /// A BABE pre-runtime digest. This contains all data required to validate a
 /// block and for the BABE runtime module. Slots can be assigned to a primary
 /// (VRF based) and to a secondary (slot number based).
-#[cfg(feature = "std")]
-#[derive(Clone, Debug)]
-pub enum PreDigest {
+#[derive(Clone, Debug, Encode, Decode)]
+pub enum RawPreDigest<VRFOutput=schnorrkel::RawVRFOutput, VRFProof=schnorrkel::RawVRFProof> {
 	/// A primary VRF-based slot assignment.
+	#[codec(index = "1")]
```
rphmeier: It's good that this is added, but can you confirm that the format of what's deployed already has the same indices?

primitives/consensus/babe/src/digests.rs
rphmeier: I see - as long as `vrf_output` and `vrf_proof` encode the same as before (i.e. they are thin wrappers) then it should all be OK.

andresilva: But isn't changing the order of fields problematic?

frame/babe/src/lib.rs
```diff
@@ -373,17 +371,17 @@ impl<T: Trait> Module<T> {
...
 	}
 
-	fn deposit_vrf_output(vrf_output: &[u8; VRF_OUTPUT_LENGTH]) {
+	fn deposit_vrf_output(vrf_output: &schnorrkel::RawVRFOutput) {
 		let segment_idx = <SegmentIndex>::get();
 		let mut segment = <UnderConstruction>::get(&segment_idx);
 		if segment.len() < UNDER_CONSTRUCTION_SEGMENT_LENGTH {
 			// push onto current segment: not full.
-			segment.push(*vrf_output);
+			segment.push(vrf_output.clone());
```
andresilva: Shouldn't we be able to deref here?

primitives/consensus/babe/src/digests.rs
rphmeier: I had the same initial thought, but this is actually the same order of fields as the `RawBabePreDigest` which is removed in this PR.

frame/babe/src/lib.rs
sorpaas: It's due to the default type parameter of `RawPreDigest`. In expression they won't be auto-guessed, so if without the annotation it has to be explicit. We can also write:

```
let pre_digest = RawPreDigest::<RawVRFProof, RawVRFOutput>::decode(&mut data).ok()?;
```

primitives/consensus/babe/src/digests.rs
sorpaas: Yeah if I'm not making any mistakes, this is the same format as the previous `RawBabePreDigest`.

sorpaas:
Tried using this branch to sync Kusama for around ~1000 blocks. Not detecting any errors.

Issue #5248, opened 2020-03-13
Refactor Democracy


gavofyork:
Democracy module currently works in two stages: votes are registered initially, then funds possibly locked only at the point that votes are counted. In the case of a high number of votes, this can lead overweight `on_initialize` in blocks, which is bad for consensus.

A more robust design would be to lock funds immediately upon voting (a `cancel_vote` extrinsic could accordingly reverse that operation if done before the vote deadline) and keep a running tally of voting funds, thereby making the operation entirely incremental and avoiding any large compute/storage payloads at the end of blocks.

Unlocking should happen lazily, to avoid the need of keeping large Vecs of account IDs around. Votes and directions should remain in storage until either the vote is cancelled, the funds are unlocked because the vote was against the final decision or the funds are unlocked because the conviction lock period is over.

PR #5294, opened 2020-03-18
Redesign Democracy pallet

gavofyork:
Closes https://github.com/paritytech/substrate/issues/5248
- Lazy locking and progressive tallying:
  - Avoids large calculation/i/o access at the end of a vote.
  - Avoids unbounded storage.
- Repots various bits of code.
- Makes vote delegation efficient and safe.
- Introduces proxy delegating.
- Introduces non-locked split voting, good for when parachains want to vote.
- Introduces instant-scheduling, allowing a safe-guarded way for a proposal to make it through almost immediately.

TODO:
- [x] Update tests
- Tests for new functionality
  - [x] locking/unlocking/consolidation & moving between voting/delegating
  - [x] proxy delegation & proxy unvoting
  - [x] vote-reaping (third-party)
  - [x] split-voting
  - [x] instant scheduling
  - [x] delegation/undelegation
  - [x] over-voting should fail
- [x] Update runtimes
- Rejig turnouts
  - [x] Delegated capital should count
  - [x] Should use full capital even when there's no conviction

CC @jacogr @shawntabrizi


frame/democracy/src/vote_threshold.rs
```diff
@@ -38,7 +39,7 @@ pub trait Approved<Balance> {
 	/// Given `approve` votes for and `against` votes against from a total electorate size of
 	/// `electorate` (`electorate - (approve + against)` are abstainers), then returns true if the
 	/// overall outcome is in favor of approval.
-	fn approved(&self, approve: Balance, against: Balance, voters: Balance, electorate: Balance) -> bool;
+	fn approved(&self, tally: Tally<Balance>, electorate: Balance) -> bool;
```
marcio-diaz: update docs

frame/democracy/src/vote_threshold.rs
```diff
@@ -76,16 +77,16 @@ impl<Balance: IntegerSquareRoot + Zero + Ord + Add<Balance, Output = Balance> +
 	///
 	/// We assume each *voter* may cast more than one *vote*, hence `voters` is not necessarily equal to
 	/// `approve + against`.
-	fn approved(&self, approve: Balance, against: Balance, voters: Balance, electorate: Balance) -> bool {
-		let sqrt_voters = voters.integer_sqrt();
+	fn approved(&self, tally: Tally<Balance>, electorate: Balance) -> bool {
```
marcio-diaz: update docs

frame/democracy/src/lib.rs
```diff
@@ -820,10 +708,26 @@ decl_module! {
...
+			// - `InstantAllowed` is `true` and `origin` is `InstantOrigin`.
+			if let Some(ensure_instant) = if voting_period < T::FastTrackVotingPeriod::get() {
+				Some(origin)
+			} else {
+				if let Err(origin) = T::FastTrackOrigin::try_origin(origin) {
+					Some(origin)
+				} else {
+					None
+				}
+			} {
```
marcio-diaz: :thinking: looks ok, maybe extract it

frame/democracy/src/tests/public_proposals.rs
```diff
@@ -0,0 +1,104 @@
...
+	new_test_ext().execute_with(|| {
+		System::set_block_number(1);
+		assert_ok!(propose_set_balance_and_note(1, 2, 5));
+		assert_ok!(Democracy::second(Origin::signed(2), 0));
+		assert_ok!(Democracy::second(Origin::signed(5), 0));
+		assert_ok!(Democracy::second(Origin::signed(5), 0));
+		assert_ok!(Democracy::second(Origin::signed(5), 0));
+		assert_eq!(Balances::free_balance(1), 5);
+		assert_eq!(Balances::free_balance(2), 15);
+		assert_eq!(Balances::free_balance(5), 35);
```
shawntabrizi: Doesn't it make more sense to assert on the reserved balance here? I mean it is the same thing, but this is a test about deposits, it would be more clear imo.

frame/democracy/src/tests/public_proposals.rs
```diff
@@ -0,0 +1,104 @@
...
+fn poor_seconder_should_not_work() {
+	new_test_ext().execute_with(|| {
+		System::set_block_number(1);
+		assert_ok!(propose_set_balance_and_note(2, 2, 11));
+		assert_noop!(Democracy::second(Origin::signed(1), 0), BalancesError::<Test, _>::InsufficientBalance);
+	});
+}
+
+#[test]
+fn runners_up_should_come_after() {
```
shawntabrizi: A note to come back to this test which does not make sense at the time of reading.

frame/democracy/src/lib.rs
```diff
@@ -344,13 +238,23 @@ pub trait Trait: frame_system::Trait + Sized {
...
-	type EmergencyVotingPeriod: Get<Self::BlockNumber>;
+	/// Origin from which the next majority-carries (or more permissive) referendum may be tabled to
+	/// vote immediately and asynchronously in a similar manner to the emergency origin. It retains
+	/// its threshold method.
+	type InstantOrigin: EnsureOrigin<Self::Origin>;
+
+	/// Indicator for whether an emergency origin is even allowed to happen. Some chains may want
+	/// to set this permanently to `false`, others may want to condition it on things such as
+	/// an upgrade having happened recently.
+	type InstantAllowed: Get<bool>;
```
shawntabrizi: If I understand how this will be used, it makes more sense as a Runtime Storage Item.

A new upgrade occurs, we may want an instant origin for a little while. Then once we know things are stable, we turn off the instant origin. In this case, it would be nice to turn it off without needing to do another upgrade.

frame/democracy/src/tests/proxying.rs
```diff
@@ -0,0 +1,104 @@
...
+		assert_eq!(Democracy::proxy(10), Some(ProxyState::Active(1)));
+
+		// Can't set when already set.
+		assert_noop!(Democracy::activate_proxy(Origin::signed(2), 10), Error::<Test>::AlreadyProxy);
+
+		// But this works because 11 isn't proxying.
+		assert_ok!(Democracy::open_proxy(Origin::signed(11), 2));
+		assert_ok!(Democracy::activate_proxy(Origin::signed(2), 11));
+		assert_eq!(Democracy::proxy(10), Some(ProxyState::Active(1)));
+		assert_eq!(Democracy::proxy(11), Some(ProxyState::Active(2)));
```
shawntabrizi: What is the advantage to the two way handshake on proxy?

Why can't an account directly give another account proxy, and then leave it to the other account to use or not use their proxy vote?

shawntabrizi: Is it only for the reference counter?

frame/democracy/src/tests/public_proposals.rs
gavofyork: this code hasn't been changed from before...

frame/democracy/src/lib.rs
gavofyork: the idea is to let is be completely configurable without necessarily placing a (potentially unused) storage item that would need to be maintained separately.

e.g. it could be permanently on (like with kusama or westend) or off (like with a conservative mainnet) and it wouldn't need that storage lookup. or could be based on some other storage item (e.g. we may store the block at which the last runtime upgrade happened independently of this mechanism and it could check if it happened recently).

frame/democracy/src/tests/proxying.rs
gavofyork: because an account can only be a proxy for a single other account and once that account is a proxy it gets a data reference and thus you would be able to mildly grief any target account.

frame/democracy/src/tests/public_proposals.rs
gavofyork: but i disagree anyway. the main point is that they no longer have the tokens to spend, not that their reserved balanced has increased.

frame/democracy/src/tests/public_proposals.rs
gavofyork: this code hasn't been changed from before, either.

frame/democracy/src/lib.rs
shawntabrizi: I see, we just need to implement the `get()` function, which could be a storage lookup.

PR #5297, opened 2020-03-18
Update CODEOWNERS: adding gnunicorn for everything client related.

gnunicorn:


NikVolf:
Will you actually review all of the above?
You will be added as requested reviewer for every PR because of this.

gnunicorn:
> Will you actually review all of the above?
No. We are currently not enforcing that code owners have to sign off (and I don't see a value in doing that)

> You will be added as requested reviewer for every PR because of this.
Yes, the idea is to be pinged for all these to have a better overview of all the changes people are doing there.

pepyakin:
According to the docs in the header (not sure if they are still correct though), the last matching rule wins. That means that all rules before will be overridden and only you will get the review requests.

gnunicorn:
dropping the attempt in favor of a bot that will do a better job.

PR #5299, opened 2020-03-18
Add sp_std::prelude to offchain worker example 

NikVolf:
Not sure if it can work without it at all in runtime?

### DIFF
--- a/frame/example-offchain-worker/src/lib.rs
+++ b/frame/example-offchain-worker/src/lib.rs
@@ -54,6 +54,7 @@ use sp_runtime::{
 	traits::Zero,
 	transaction_validity::{InvalidTransaction, ValidTransaction, TransactionValidity},
 };
+use sp_std::prelude::*;
 
 #[cfg(test)]
 mod tests;


bkchr:
Why this pr?

I don't think that we should encourage people to blindly include stuff. Especially when it is not needed as in this case here. The code compiled, why do we need the prelude?

tomusdrw:
The module is not used in any runtime, which means that it's not being compiled for WASM. I'm pretty sure that this is going to be needed when you try to use it, but now I think about it it - yeah I probably approved too eagerly, cause it would be nice to make sure we are compiling the wasm version on the ci as well, to make sure it actually works and prevent breaking it in the future.

NikVolf:
> Why this pr?
> 
> I don't think that we should encourage people to blindly include stuff. Especially when it is not needed as in this case here. The code compiled, why do we need the prelude?

But you can't even reference this module and include it in node-template-runtime to try 🤷‍♀️

bkchr:
I thought it was compiled and tested. 

If that is not the case, we should add it to the CI as proposed by @tomusdrw. We should as well test the normal `example` pallet. We can just add a pass `cargo build --no-default-features --target wasm-unknown-whatever`

NikVolf:
will reopen with proper testing

PR #5301, opened 2020-03-18
client/network-gossip: Remove obsolete trait bounds

mxinden:
- client/network-gossip: Remove obsolete Send requirement for event_stream

  With 78cdd3b there is no need for the `Network` to be `Send` given that
  `GossipEngine` does not depend on background tasks anymore.

- client/network-gossip: Remove obsolete `Send` bound on `Network`

  With 78cdd3b there is no need for the `event_stream` to be `Send` given
 that `GossipEngine` does not depend on background tasks anymore.

### DIFF
--- a/client/network-gossip/src/bridge.rs
+++ b/client/network-gossip/src/bridge.rs
@@ -31,7 +31,7 @@ pub struct GossipEngine<B: BlockT> {
 	state_machine: ConsensusGossip<B>,
 	network: Box<dyn Network<B>>,
 	periodic_maintenance_interval: futures_timer::Delay,
-	network_event_stream: Pin<Box<dyn Stream<Item = Event> + Send>>,
+	network_event_stream: Pin<Box<dyn Stream<Item = Event>>>,
 	engine_id: ConsensusEngineId,
 }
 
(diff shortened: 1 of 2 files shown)


andresilva:
Tests need to be updated as well.

mxinden:
`GossipEngine` holds a `Box<dyn Network>`. Given that `GossipEninge` thereby uses dynamic dispatching to store a `Network` the compiler has no way to tell whether `Network`, `Box<dyn Network>` and thereby `GossipEngine` as a whole are `Send`.

With this patch `client/network-gossip` compiles fine, due to the fact that nothing within `client/network-gossip` requires `GossipEngine` to be `Send`.

`client/finality-grandpa` on the other hand does not compile fine. Given that a `NetworkBridge` holds a `GossipEngine` and given that `NetworkBridge` is required to be `Send` `GossipEngine` also needs to be `Send`. As a first thought I considered the `Arc<Mutex<_>>` that we wrap `GossipEngine` in within `client/finality-grandpa` to solve this issue.

An `Arc<T>` is only [`Send` and `Sync` when `T` is `Send` and `Sync`](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-Send). It does not make `GossipEngine` `Send`, but only allows us to have shared ownership.

A `Mutex<T>` is as well only [`Send` when `T` is `Send`](https://doc.rust-lang.org/std/sync/struct.Mutex.html#impl-Send). It does not make `GossipEngine` `Send`, but only allows shared mutability through mutual exclusion.

### Conclusion

With a `GossipEngine` that holds a dynamically dispatched `Network` one has to have a `Send` trait bound on `Network`.


(Thanks @tomaka for the help.)

Issue #4728, opened 2020-01-23
Light client finalized stuck at #28175 on Kusama chain


amaurym:
- It would help if you submit info about the system you are running, e.g.: operating system, kernel version, amount of available memory and swap, etc. **Tried in macOS 10.15.2, and with the wasm binary**
- Describe the role your node plays, e.g. validator, sentry, full node or light client. **Light Client**.
- Any command-line options were passed? `--light`

**Repro:**
- `polkadot --light`
- Wait for ~2h for it to sync to the tip

**Expected:**
Finalized block should match actual finalized

**Actual:**
```
[2020-01-23T16:59:01.069Z] info 2020-01-23 17:59:01 Imported #751213 (0xae26…ce92)

[2020-01-23T16:59:01.789Z] info 2020-01-23 17:59:01 Reorg from #751213,0xae26…ce92 to #751213,0xc566…ee6d, common ancestor #751212,0x7933…989f
2020-01-23 17:59:01 Imported #751213 (0xc566…ee6d)

[2020-01-23T16:59:04.645Z] info 2020-01-23 17:59:04 Idle (20 peers), best: #751213 (0xc566…ee6d), finalized #28175 (0x456c…19f0), ⬇ 8.9kiB/s ⬆ 5.6kiB/s

[2020-01-23T16:59:06.855Z] info 2020-01-23 17:59:06 Imported #751214 (0x4339…20ff)

[2020-01-23T16:59:09.649Z] info 2020-01-23 17:59:09 Idle (20 peers), best: #751214 (0x4339…20ff), finalized #28175 (0x456c…19f0), ⬇ 8.0kiB/s ⬆ 4.6kiB/s

[2020-01-23T16:59:13.243Z] info 2020-01-23 17:59:13 Imported #751215 (0x123b…fc6d)

[2020-01-23T16:59:14.654Z] info 2020-01-23 17:59:14 Idle (20 peers), best: #751215 (0x123b…fc6d), finalized #28175 (0x456c…19f0), ⬇ 8.5kiB/s ⬆ 6.2kiB/s

[2020-01-23T16:59:18.923Z] info 2020-01-23 17:59:18 Imported #751216 (0xdb1a…4a72)

[2020-01-23T16:59:19.661Z] info 2020-01-23 17:59:19 Idle (20 peers), best: #751216 (0xdb1a…4a72), finalized #28175 (0x456c…19f0), ⬇ 7.6kiB/s ⬆ 6.4kiB/s

[...]

PR #5306, opened 2020-03-18
grandpa: support for hard forking any pending standard changes

andresilva:
This PR adds support in GRANDPA for defining tuples `(SetId, (BlockHash, BlockNumber), AuthorityList)` which will be replace any existing pending authority set changes and additionally will also replace the signalled authority set change when importing a hard fork block (e.g. when syncing the chain after the fact).


andresilva:
cc @svyatonik We'll need to add support for this in the light client as well as it won't be able to sync Kusama after this.

svyatonik:
@andresilva Thanks, I've logged this in #5309 . Finalization on light clients is broken on Kusama anyway - see #4728

andresilva:
@svyatonik Yeah, I forgot that we need to change light block import to require storage proofs instead for checking authorities.

PR #5308, opened 2020-03-19
Prompt error if vanity pattern is not valid base58

lwshang:
The `vanity` subcommand of `subkey` will search for an address with a pattern. Since the address are subject to base58 encoding, the characters in the pattern should also narrow to base58. Otherwise the search will never end which is the current behaviour.

This PR fix the problem by prompting an error when user provide an invalid pattern.

### DIFF
--- a/bin/utils/subkey/src/vanity.rs
+++ b/bin/utils/subkey/src/vanity.rs
@@ -65,11 +65,7 @@ fn calculate_score(_desired: &str, key: &str) -> usize {
 /// Validate whether the char is allowed to be used in base58.
 /// num 0, lower l, upper I and O are not allowed.
 fn validate_base58(c :char) -> bool {
-	if !c.is_alphanumeric() || "0lIO".contains(c) {
-		false
-	} else {
-		true
-	}
+	c.is_alphanumeric() && !"0lIO".contains(c)
 }
 
 pub(super) fn generate_key<C: Crypto>(desired: &str) -> Result<KeyPair<C>, &'static str> where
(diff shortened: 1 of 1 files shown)


bin/utils/subkey/src/vanity.rs
```diff
@@ -62,13 +62,28 @@ fn calculate_score(_desired: &str, key: &str) -> usize {
...
 }
 
+/// Validate whether the char is allowed to be used in base58.
+/// num 0, lower l, upper I and O are not allowed.
+fn validate_base58(c :char) -> bool {
+	if !c.is_alphanumeric() || "0lIO".contains(c) {
+		false
+	} else {
+		true
+	}
```
bkchr:
```suggestion
	c.is_alphanumeric() && !"0lIO".contains(c)
```

No if required here

lwshang:
> Can you please add a test.

Sure, I will. 

In PolkadotJS, user can choose whether they want "case sensitive" or not. Do I need to check how will that be influenced by this PR?

bkchr:
If a capital `L` is forbidden, it should be tested as well.

lwshang:
Capital `L` is allowed in base58. So in `subkey vanity`, patterns contain `L` is valid.

From my understanding, `subkey` itself currently doesn't support case insensitive directly. I speculate that `PolkadotJS` has some mechanism to expand search cases. Therefore, I think it's meaningful to have a check on their code to see their implementation.

bkchr:
@jacogr can you shed some light on case sensitive here?

jacogr:
Add the UI does is to lowercase the whole thing when doing a match, if so requested. No magic there.

lwshang:
> All the UI does is to lowercase the whole thing when doing a match, if so requested. No magic there.

Just read your code, the `PolkadotJS` is more versatile. I would like to improve `subkey` with similar capabilities in the future.

PR #5314, opened 2020-03-19
Lazy Treasury Payouts

shawntabrizi:
This PR updated the Treasury Pallet to perform lazy payouts rather than automatic payouts with `on_finalize`.

There are a number of overall changes here:

- Create a new `payout_proposal` extrinsic that can be called by anyone and will payout an approved proposal. This function respects `KeepAlive` requirements of the treasury pot.
- Remove `Approved` storage vector, and added an `approved` tracking item to the `Proposal` struct. Since we no longer iterate over all the approved proposals, we do not need to keep track of a separate vector which lists them all. We simply track the approval status of a proposal on the proposal itself, and verify it is approved when asked to `payout_proposal`.
- Added a new `AmountApproved` storage item which simply tracks how much of the treasury pot has currently been approved, but not claimed. This is used so that we only burn funds when we are in excess of the amount approved.
- Change `on_finalize -> spend_funds` to `on_initialize -> burn_funds`. Now, at the beginning of every `BurnPeriod`, we check if we have any excess funds (comparing the `pot` with the `amount_approved`, and burn any of those excess funds in the same way we did before.
- Removed `Spending` and `Rollover` events, which do not provide unique information or value with this version of payouts.

**Additional Changes**

- For no reason I could tell, the tipping system was allowing death of the treasury account. I have updated the logic so that the max tip can be of size `pot` which will respect the existence requirement of the account, and the currency operations also now have `KeepAlive`.

(Note this is built on top of treasury benchmarks)


shawntabrizi:
This is now updated to the latest in Substrate master, with benchmarks updated.

frame/treasury/src/benchmarking.rs
```diff
@@ -94,7 +95,8 @@ fn create_approved_proposals<T: Trait>(n: u32) -> Result<(), &'static str> {
 		let proposal_id = ProposalCount::get() - 1;
 		Treasury::<T>::approve_proposal(RawOrigin::Root.into(), proposal_id)?;
 	}
-	ensure!(Approvals::get().len() == n as usize, "Not all approved");
+	ensure!(Proposals::<T>::iter().count() == n as usize, "Not all created");
+	ensure!(Proposals::<T>::iter().all(|(_, p)| p.approved == true), "Not all approved");
```
sjeohp:
```suggestion
	ensure!(Proposals::<T>::iter().all(|(_, p)| p.approved), "Not all approved");
```

frame/treasury/src/lib.rs
```diff
@@ -369,22 +374,55 @@ decl_module! {
...
+			let new_amount = amount_approved.checked_add(&proposal.value).ok_or(Error::<T>::Overflow)?;
+
+			proposal.approved = true;
+			AmountApproved::<T>::put(new_amount);
+			Proposals::<T>::insert(proposal_id, proposal);
+
+			Self::deposit_event(Event::<T>::Approved(proposal_id, amount_approved));
+		}
+
+		/// Payout an approved proposal, returning the deposit taken for making the proposal.
```
joepetrowski:
```suggestion
		/// Pay out an approved proposal, returning the deposit taken for making the proposal.
```

frame/treasury/src/lib.rs
```diff
@@ -369,22 +374,55 @@ decl_module! {
...
 		#[weight = SimpleDispatchInfo::FixedOperational(100_000)]
 		fn approve_proposal(origin, #[compact] proposal_id: ProposalIndex) {
 			T::ApproveOrigin::try_origin(origin)
 				.map(|_| ())
 				.or_else(ensure_root)?;
 
-			ensure!(<Proposals<T>>::contains_key(proposal_id), Error::<T>::InvalidProposalIndex);
-			Approvals::mutate(|v| v.push(proposal_id));
+			let mut proposal = Proposals::<T>::get(proposal_id).ok_or(Error::<T>::InvalidProposalIndex)?;
+			let amount_approved = Self::amount_approved();
```
gavofyork: would be better with a `mutate`.

gavofyork:
This alters the logic of this pallet. In particular, payout can happen immediately after approval, whereas before payouts all happened at a single point every payout period. Indeed, with this code change there's really no point in splitting up `approve` and `payout` into two separate functions.

frame/treasury/src/lib.rs
shawntabrizi: Ack, if you prefer the `try_mutate` syntax I can change it. But I want to confirm that logically/complexity it is the same thing as what I have written here.

shawntabrizi:
I can make it so that approved payouts have a delay which ends on the burn period. Should be relatively easy to calculate and add to the proposal structure.

frame/treasury/src/lib.rs
```diff
@@ -163,10 +163,24 @@ pub trait Trait: frame_system::Trait {
...
+pub enum ProposalStatus<BlockNumber> {
+	/// Proposal is pending approval.
+	Pending,
+	/// Proposal is approved for a one time payout.
+	Approved(BlockNumber),
+	/// Proposal is approved for a reoccurring payout.
+	/// First `BlockNumber` is the next payout block.
+	/// Second `BlockNumber` is the period between payouts.
+	/// `u32` represents the number of payouts left.
+	Reoccurring(BlockNumber, BlockNumber, u32),
```
shawntabrizi: Would need to actually integrate this option if we went down this path

joepetrowski:
```suggestion
	Recurring(BlockNumber, BlockNumber, u32),
```

frame/treasury/src/lib.rs
gavofyork: It's not that I "prefer" it. It is categorically worse.

Compare:
```
let foo = Foo::get();
foo = ...;
Foo::put(foo);
```

to:
```
Foo::mutate(|foo| {
    *foo = ...;
});
```

The second code has:
- fewer tokens (18 vs 15);
- far fewer identifiers (7 vs 4);
- no identifier duplication, whereas the first duplicates the `Foo` identifier.

These factors significantly impact readability, reviewability/auditability, maintainability and long-term, the likelihood of the code to be correct.

This is not abstract reasoning. I have seen bugs in the past within the pattern here. There are two bugs that are incipient to the first code and impossible in the second are a) mismatching identifiers for `put` and `get`, and b) forgetting the trailing `put`. I've definitely seen (b) in the wild.

gavofyork:
Yeah, I'm still unconvinced on the utility of this. It's not like it's a superlinear amount of work, nor that its uncontrolled (each item must be added only with council approval), nor that it's difficult to pre-compute the weight. These are the factors that are important in deciding whether to offload work. Making things lazy just for the sake of it actually introduces *extra* buren on the network through the additional transactions it needs to process (transaction queue logic, signature verification, block size, ...).

shawntabrizi:
Sure. Seems like we should create another PR though that switches to `on_initialize` and fixes `KeepAlive` issues.

PR #5317, opened 2020-03-19
Factor out can_set_code

coriolinus:
The motivation for this feature is parachain runtime upgrades, which
happen in two different moments: the initial call, when the checks
should be performed, and at the actual upgrade block. Calling 
`System::set_code` at the actual upgrade point would defer those
checks too much: errors wouldn't propagate back to the caller. It's much
better and more maintainable to just call `can_set_code` in the
parachain upgrade function than to manually copy those checks.

~~I'm not entirely thrilled with the interface, which requires
cloning the entire code vector in `set_code`.~~

Making this function non-dispatchable fixed the former interface issue.


frame/system/src/lib.rs
```diff
@@ -464,9 +464,15 @@ decl_module! {
...
+		/// Determine whether or not it is possible to update the code.
+		///
+		/// This function has no side effects and is idempotent, but is fairly
+		/// heavy. It is automatically called by `set_code`; in most cases,
+		/// a direct call to `set_code` is preferable. It is useful to call
+		/// `can_set_code` when it is desirable to perform the appropriate
+		/// runtime checks without actually changing the code yet.
 		#[weight = SimpleDispatchInfo::FixedOperational(200_000)]
-		pub fn set_code(origin, code: Vec<u8>) {
+		pub fn can_set_code(origin, code: Vec<u8>) {
```
bkchr: You only declare dispatchable functions inside `decl_module!`.

This is not an dispatchable function and you also don't want that the function sets `:code` in the storage.

Just put the function into the impl block of the module.

frame/system/src/lib.rs
```diff
@@ -573,6 +542,40 @@ decl_module! {
 	}
 }
 
+impl<T: Trait> Module<T> {
+	/// Determine whether or not it is possible to update the code.
```
bkchr: Please move to the other impl block: https://github.com/paritytech/substrate/blob/b3136234c9b41d630670ba8bb8f647ced214d7f2/frame/system/src/lib.rs#L702

coriolinus:
Done! I hadn't spotted the other impl block.

For now, I'm assuming that CI success is mandatory. Bumping the impl_version is no problem, as soon as I find where it's defined. However, the other failure has to do with Onion3 lifetimes somehow, and quite frankly I have no idea what has gone wrong there or how to fix it. Is that actually this PR's fault?

coriolinus:
Also: I note the breaksapi tag, which almost certainly comes from adding a public function. Does that tag imply any required action?

bkchr:
No, everything is okay. I merge when CI is green.

PR #5318, opened 2020-03-19
Benchmark Im Online Pallet

shawntabrizi:
This benchmarks the Im Online pallet.

Here we are interested in the time it takes to call the unsigned extrinsic `heartbeat` and to do the `verify_unsigned` logic.

```
./target/release/substrate benchmark --execution=wasm --wasm-execution=compiled --pallet im-online --extrinsic heartbeat

Pallet: "im-online", Extrinsic: "heartbeat", Lowest values: [], Highest values: [], Steps: [], Repeat: 1
k,extrinsic_time,storage_root_time
1,14298000,8867000
100,15021000,9025000
199,13722000,8417000
298,9905000,78000
397,9566000,79000
496,11492000,87000
595,10039000,88000
694,10631000,93000
793,9800000,102000
892,8296000,101000
991,11956000,2975000
Done.
```

I have some issues with benchmarking `validate_unsigned`. See my post below.

To get the benchmarks to work, I added the keystore extension to the benchmark state-machine.


shawntabrizi:
This is mostly done, however I was running into trouble when trying to benchmark `verify_unsigned` via the CLI.

I get the error:

> Error: "Transaction has a bad signature"

Which is caused by:

```
// check signature (this is expensive so we do it last).
let signature_valid = heartbeat.using_encoded(|encoded_heartbeat| {
	authority_id.verify(&encoded_heartbeat, &signature)
});

if !signature_valid {
	return InvalidTransaction::BadProof.into();
}
```

This means, for whatever reason, `authority_id.verify()` does not work with my signature + payload. However, the unit tests show that it works. I tried some alternative approaches, for example using a string as a Seed for generating the account, but I get the same error.

I am not sure why... any ideas?

frame/im-online/src/benchmarking.rs
```diff
@@ -0,0 +1,120 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Balances pallet benchmarking.
```
marcio-diaz:
```suggestion
//! I'm Online pallet benchmarking.
```

frame/im-online/src/benchmarking.rs
```diff
@@ -0,0 +1,120 @@
...
+{
+	let mut keys = Vec::new();
+	for _ in 0 .. k {
+		keys.push(T::AuthorityId::generate_pair(None));
+	}
+	Keys::<T>::put(keys);
+
+	let network_state = OpaqueNetworkState {
+		peer_id: OpaquePeerId::default(),
+		external_addresses: vec![OpaqueMultiaddr::new(vec![]); 10],
```
marcio-diaz: Maybe check if `external_addresses` is reasonable, or make it a parameter (I think we are encoding/decoding it)

frame/im-online/src/benchmarking.rs
```diff
@@ -0,0 +1,120 @@
...
+	Ok((input_heartbeat, signature))
+}
+
+benchmarks! {
+	_{ }
+
+	heartbeat {
+		let k in 1 .. MAX_KEYS;
+		let (input_heartbeat, signature) = create_heartbeat::<T>(k)?;
+	}: _(RawOrigin::None, input_heartbeat, signature)
```
marcio-diaz: Is it worth checking with previous heartbeats?

kianenigma:
Most likely, the key used to sign and the one used to verify don't match. The possibility of a bug in crypto is very low.

Can you reason about which key is being used to sign this payload? IIRC the offchain code looks up the local keystore and submits heartbeats for each key that matches with an entry in the local `Self::keys()` storage. 

https://github.com/paritytech/substrate/blob/5f99bcbef179931837d06346bfa7b37c6939dbf9/frame/im-online/src/lib.rs#L493-L506

kianenigma:
So, I don't know what 

```
T::AuthorityId::generate_pair(None);
```

Exactly generates, but it should generally be avoided IMO. but whatever it generates, unless if it is generating the same thing for all accounts, it is wrong. 

in your heartbeat payload: `authority_index: k-1,` Hence, you are claiming to own the key of this authority and can sign a message that can be verified with the on-chain keys. Hence, you encode with 

```
	let encoded_heartbeat = input_heartbeat.encode();
	let authority_id = Self::keys()[k-1];
	let signature = authority_id.sign(&encoded_heartbeat).ok_or("couldn't make signature")?;
```

frame/im-online/src/benchmarking.rs
```diff
@@ -0,0 +1,120 @@
...
+
+use crate::Module as ImOnline;
+
+const MAX_KEYS: u32 = 1000;
+
+pub fn create_heartbeat<T: Trait>(k: u32) ->
+	Result<(crate::Heartbeat<T::BlockNumber>, <T::AuthorityId as RuntimeAppPublic>::Signature), &'static str>
+{
+	let mut keys = Vec::new();
+	for _ in 0 .. k {
```
kianenigma:
```suggestion
	for _ in 0..k {
```

shawntabrizi:
@kianenigma you are 100% correct!

`generate_pair` is not consistent here, so we do need to take from the list I generated. I have updated the code and the benchmark works! Thank you!

frame/im-online/src/benchmarking.rs
shawntabrizi: Yeah, Should figure out what the right size is here. I will add some constants that we can tweak if needed.

frame/im-online/src/benchmarking.rs
shawntabrizi: The code shows that heartbeat should be constant time. We will test using the genesis generator and put storage items into the double map to simulate a more complex trie, but this can happen outside the benchmark

shawntabrizi:
Okay everything should be working 100%, would be happy to get any final feedback or a green checkmark :)

bkchr:
> So, I don't know what
> 
> ```rust
> T::AuthorityId::generate_pair(None);
> ```
> 
> Exactly generates, but it should generally be avoided IMO. but whatever it generates, **unless if it is generating the same thing for all accounts**, it is wrong.

@kianenigma we have documentation for it :P

utils/frame/benchmarking-cli/src/lib.rs
```diff
@@ -106,6 +111,9 @@ impl BenchmarkCmd {
 			2, // The runtime instances cache size.
 		);
 
+		let mut extensions: Extensions = Default::default();
```
bkchr:
```suggestion
		let mut extensions = Extensions::default();
```

PR #5324, opened 2020-03-20
EVM storage tracing

kirillt:
This PR adds storage tracing into EVM module. This is implemented as special event emitted every time a constructor/method of a smart contract is called. The event contains digest of all storage indices affected during the call, i.e. all indices storage was written by. When storage is reset another kind of Event is emitted.

### DIFF
--- a/frame/evm/src/backend.rs
+++ b/frame/evm/src/backend.rs
@@ -140,6 +140,10 @@ impl<'vicinity, T: Trait> ApplyBackend for Backend<'vicinity, T> {
 		I: IntoIterator<Item=(H256, H256)>,
 		L: IntoIterator<Item=evm::backend::Log>,
 	{
+		#[cfg(feature = "storage_tracing")] {
+			let mut affected_storage: Vec<(H160,H256)> = vec![];
+		}
+
 		for apply in values {
 			match apply {
 				Apply::Modify {
@@ -156,6 +160,9 @@ impl<'vicinity, T: Trait> ApplyBackend for Backend<'vicinity, T> {
 
 					if reset_storage {
 						AccountStorages::remove_prefix(address);
+						#[cfg(feature = "storage_tracing")] {
+							Module::<T>::deposit_event(Event::StorageReset(StorageReset(address)))
+						}
 					}
 
 					for (index, value) in storage {
@@ -163,6 +170,10 @@ impl<'vicinity, T: Trait> ApplyBackend for Backend<'vicinity, T> {
 							AccountStorages::remove(address, index);
 						} else {
 							AccountStorages::insert(address, index, value);
+						}
+
+						#[cfg(feature = "storage_tracing")] {
+							affected_storage.push((address, index));
 						}
 					}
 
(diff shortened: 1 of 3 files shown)


frame/evm/src/backend.rs
```diff
@@ -140,6 +140,10 @@ impl<'vicinity, T: Trait> ApplyBackend for Backend<'vicinity, T> {
 		I: IntoIterator<Item=(H256, H256)>,
 		L: IntoIterator<Item=evm::backend::Log>,
 	{
+		#[cfg(storage_tracing)] {
```
sorpaas: I think feature conditionals should be `#[cfg(feature = "storage-tracing")]`, or does this work?

frame/evm/src/backend.rs
```diff
@@ -140,6 +140,10 @@ impl<'vicinity, T: Trait> ApplyBackend for Backend<'vicinity, T> {
...
+			let mut affected_memory: Vec<(H160,H256)> = vec![];
```
sorpaas: Grumble.
```suggestion
			let mut affected_memory: Vec<(H160, H256)> = vec![];
```

frame/evm/src/backend.rs
```diff
@@ -185,3 +229,19 @@ impl<'vicinity, T: Trait> ApplyBackend for Backend<'vicinity, T> {
 		}
 	}
 }
+
+#[derive(Clone, Eq, PartialEq, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(Debug, Serialize, Deserialize))]
+pub struct StorageWritten {
```
sorpaas: Those structs should be hidden behind `storage-tracing` feature.

frame/evm/src/lib.rs
```diff
@@ -145,6 +145,11 @@ decl_event! {
 		Log(Log),
 		/// A contract has been created at given address.
 		Created(H160),
+
+		/// A batch of all indices written during last call (see "storage_tracing" feature)
+		StorageWritten(backend::StorageWritten),
```
sorpaas: Same as those two items. They should be hidden behind the `storage-tracing` feature.

frame/evm/src/backend.rs
```diff
@@ -176,6 +187,39 @@ impl<'vicinity, T: Trait> ApplyBackend for Backend<'vicinity, T> {
 			}
 		}
 
+		#[cfg(storage_tracing)] {
+			if affected_memory.len() > 0 {
```
sorpaas: Small grumble -- maybe better call it `affected_storage(s)` as this is about EVM storage, not EVM memory.

sorpaas:
Also just a small note -- in this PR we only traced the *result* storage value, but not anything intermediate. That means, if a storage changed from 0 to 1 and then to 0, we'll only record it in tracing as 0.

frame/evm/src/backend.rs
kirillt: Good point

kirillt:
This seems to be impossible to hide auxiliary events with current implementation of `decl_event` macro, which doesn't allow to introduce separate from `Event` type for events as well as conditionally compile branches of `Event` enum. At the same time, the events must be hidden if they are cannot be generated in a given features set.

Since this is temporary solution, it seems to be not rational to change logic of `decl_event` macro for it. I suggest to close this PR.

